PR #166, opened 2019-12-17
tabular kaggle tutorial

jwmueller:
*Description of changes:*

Notebook tutorial for running autogluon on Kaggle competitions with tabular data. This tutorial is slow since Kaggle dataset is very large, so it is not executed with {.python .input} blocks.

### DIFF
--- a/docs/tutorials/tabular_prediction/tabular-kaggle.md
+++ b/docs/tutorials/tabular_prediction/tabular-kaggle.md
@@ -1,0 +1,110 @@
+# How to easily enter Kaggle contests with AutoGluon  
+:label:`sec_tabularkaggle`
+
+This tutorial will teach you how to use AutoGluon to become a serious Kaggle competitor without writing lots of code.
+
+We first outline the general steps to use AutoGluon in Kaggle contests. Here, we assume the competition involves tabular data which are stored in one (or more) CSV files.
+
+1) Run Bash command: pip install kaggle
+
+2) Navigate to: https://www.kaggle.com/account, and create an account (if necessary).
+Then , click on "Create New API Token" and move downloaded file to this location on your machine: `~/.kaggle/kaggle.json`. For troubleshooting, see [Kaggle API instructions](https://www.kaggle.com/docs/api). 
+
+3) To download data programmatically: Execute this Bash command in your terminal: 
+
+`kaggle competitions download -c [COMPETITION]`
+
+Here, [COMPETITION] should be replaced by the name of the competition you wish to enter.
+Alternatively, you can download data manually: Just navigate to website of the Kaggle competition you wish to enter, click "Download All", and accept the competition's terms.
+
+4) If the competition's training data is comprised of multiple CSV files, use [pandas](https://pandas.pydata.org/pandas-docs/stable/user_guide/merging.html) to properly merge/join them into a single data table where rows = training examples, columns = features.
+
+5) Run autogluon `fit()` on the resulting data table.
+
+6) Load the test dataset from competition (again making the necessary merges/joins to ensure it is in the exact same format as the training data table), and then call autogluon `predict()`.  Subsequently use [`pandas.read_csv`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html) to load the competition's `sample_submission.csv` file into a Dataframe, put the AutoGluon predictions in the right column of this Dataframe, and finally save it as a CSV file via [`pandas.to_csv`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_csv.html). If the competition does not offer a sample submission file, you will need to create the submission file yourself by appropriately reformatting AutoGluon's test predictions.
+
+7) Submit your predictions via Bash command: 
+
+`kaggle competitions submit -c [COMPETITION] -f [FILE] -m ["MESSAGE"]`
+
+Here, [COMPETITION] again is the competition's name, [FILE] is the name of the CSV file you created with your predictions, and ["MESSAGE"] is a string message you want to record with this submitted entry. Alternatively, you can  manually upload your file of predictions on the competition website.
+
+8) Finally, navigate to competition leaderboard website to see how well your submission performed! 
+It may take time for your submission to appear.
+
+
+
+Below, we demonstrate how to do steps (4)-(6) in Python for a specific Kaggle competition: ieee-fraud-detection (https://www.kaggle.com/c/ieee-fraud-detection/). 
+This means you'll need to run the above steps with `[COMPETITION]` replaced by `ieee-fraud-detection` in each command.  Here, we assume you've already completed steps (1)-(3) and the data CSV files are available on your computer. To begin step (4), we first load the competition's training data into Python:
+
+```
+import pandas as pd
+import numpy as np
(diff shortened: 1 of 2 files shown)


szha:
Job PR-166-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-166/1/index.html

szha:
Job PR-166-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-166/2/index.html

szha:
Job PR-166-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-166/3/index.html

szha:
Job PR-166-4 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-166/4/index.html

szha:
Job PR-166-6 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-166/6/index.html

szha:
Job PR-166-7 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-166/7/index.html

docs/tutorials/tabular_prediction/tabular-kaggle.md
```diff
@@ -0,0 +1,114 @@
...
+`kaggle competitions submit -c ieee-fraud-detection -f sample_submission.csv -m "my first submission"`
+
+You can now play with different `fit()` arguments and feature-engineering techniques to try and maximize the rank of your submissions in the Kaggle Leaderboard!
+
+
+**Tips to maximize predictive performance:**
+
+   - Be sure to specify the appropriate evaluation metric if one is specified on the competition website! If you are unsure which metric is best, then simply do not specify this argument when invoking `fit()`; AutoGluon should still produce high-quality models by automatically inferring which metric to use.
+
+   - If the training examples are time-based and the competition test examples come from future data, we recommend you reserve the most recently-collected training examples as a separate validation dataset passed to `fit()`. Otherwise, you do not need to specify a validation set yourself and AutoGluon will automatically partition the competition training data into its own training/validation sets.
```
Innixma: Mention that they should avoid bagging and stacking if this is the case, since we will use the val data in bagging currently.

tests/unittests/test_tabular.py
```diff
@@ -28,75 +28,93 @@
...
+    # If True, run a faster benchmark (subsample training sets, less epochs, etc),
+    # otherwise we run full benchmark with default AutoGluon settings.
+    # performance_value warnings are disabled when fast_benchmark = True.
+
+    #### If fast_benchmark = True, can control model training time here. Only used if fast_benchmark=True ####
+    if fast_benchmark:
+        subsample_size = 100
+        nn_options = {'num_epochs': 3} 
+        gbm_options = {'num_boost_round': 30}
+        hyperparameters = {'GBM': gbm_options, 'NN': nn_options}
```
Innixma: Maybe we should start using the default hyperparameters to ensure all default models are working properly? This test wouldn't catch a crash in KNN for example.

tests/unittests/test_tabular_bagstack.py
```diff
@@ -0,0 +1,176 @@
...
+    if hyperparameters is not None:
+        fit_args['hyperparameters'] = hyperparameters
+    if time_limits is not None:
+        fit_args['time_limits'] = time_limits
+    ###################################################################
+    run_tabular_benchmarks(fast_benchmark=fast_benchmark, subsample_size=subsample_size, perf_threshold=perf_threshold,
+                           seed_val = seed_val, fit_args=fit_args)
+
+
+def test_tabular_stack2():
```
Innixma: This is thorough but perhaps we could just check if the stack_ensemble_levels is set to 0 because of num_bagging_folds=0 instead of actually running it. We could do this by splitting .fit internal logic into two stages: Handling params and the call to learner.fit(), we could avoid running the second stage and just check if stack_ensemble_levels was set to 0.

tests/unittests/test_tabular_bagstack.py
```diff
@@ -0,0 +1,176 @@
...
+    if hyperparameters is not None:
+        fit_args['hyperparameters'] = hyperparameters
+    if time_limits is not None:
+        fit_args['time_limits'] = time_limits
+    ###################################################################
+    run_tabular_benchmarks(fast_benchmark=fast_benchmark, subsample_size=subsample_size, perf_threshold=perf_threshold,
+                           seed_val=seed_val, fit_args=fit_args)
+
+
+def test_tabular_stack1():
```
Innixma: ditto to below comment

tests/unittests/test_tabular_bagstack.py
```diff
@@ -0,0 +1,176 @@
...
+    if hyperparameters is not None:
+        fit_args['hyperparameters'] = hyperparameters
+    if time_limits is not None:
+        fit_args['time_limits'] = time_limits
+    ###################################################################
+    run_tabular_benchmarks(fast_benchmark=fast_benchmark, subsample_size=subsample_size, perf_threshold=perf_threshold,
+                           seed_val=seed_val, fit_args=fit_args)
+
+
+if __name__ == '__main__':
```
Innixma: Seems like a lot of duplicate code here. I think it might be nice to reduce the duplication by having default parameter values and then only having to specify the non-standard parameters to the test func (stack_ensemble_levels and num_bagging_folds). This would avoid subtle bugs if we have to refactor this later.

This could be done by having another function run_tabular_benchmarks_template(), which would call run_tabular_benchmarks with all the code above, but having you able to specify overrides.

docs/tutorials/tabular_prediction/tabular-kaggle.md
jwmueller: Done in new commit.  I guess we also needed to mention this fact in the fit() docs, so have done so as well.

tests/unittests/test_tabular.py
jwmueller: will add one unit-test that runs with default hyperparams. The problem is default may be slow, depending on models/dataset

tests/unittests/test_tabular_bagstack.py
jwmueller: you don't think we'd ever support stacking without bagging in the future? If so I can just remove these tests {test_tabular_stack1() and test_tabular_stack2()} from being run at the moment.

tests/unittests/test_tabular_bagstack.py
jwmueller: I actually think we eventually want to change the defaults in the different tests, so we have more comprehensive test coverage of various scenarios. This is why the parameters are defined in every test, they just happen to be the same right now.  I'm not sure our our set of default base models is fully finalized, but once it is, I think these parameters should aim cover different subsets of base models in different unit tests. Because I'm confident GBM and NN will be in the official final set of base models, this is why these are currently the only ones tried for now.  I've gone and changed the seed for the different tests in new commit.

szha:
Job PR-166-9 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-166/9/index.html

Issue #168, opened 2019-12-19
Improve out-of-memory handling


bwerness:
Currently when running on a large dataset with limited compute, the model training can fail completely with out-of-memory errors, when it is possible for it to successfully train several of the models and produce strong predictions.  To make it as reliable as possible for new users, it would be good to instead catch errors when models fail, and simply omit them from the ensemble.   Since (in my experience) the models most likely to yield out of memory errors are the KNN models, this would likely also not have a large impact on final metrics.  These failures could then be reported to the user.

jwmueller:
Thanks Brent! This is also related to https://github.com/awslabs/autogluon/issues/83
If you encounter other OOM cases, please let us know in the linked issue.

Innixma:
Just merged #170 which should help significantly with this problem

Innixma:
#177 should help to resolve the majority of remaining memory issues

Innixma:
#177 is merged and should resolve this issue for the grand majority of cases. Please feel free to re-open this issue if you find OOM errors still occurring.

PR #170, opened 2019-12-21
Fixed model cleanup after training in low_memory mode

sgdread:
*Issue #, if available:*
Could be related to https://github.com/awslabs/autogluon/issues/168

*Description of changes:*
Models are not cleaned up after training in low-memory mode because `models` collection holds reference to the object.


Innixma:
Great defect find and fix! This defect appears to primarily impact non-bagging setting, but can also have a fairly massive impact on large multiclass classification datasets due to the large OOF predictions being held in the stacker/bag ensemble models that are still present in memory. This should help with OOM errors substantially!

Innixma:
Merging as changes do not impact the CI failure.

PR #177, opened 2020-01-07
AutoGluon Tabular Improvements

Innixma:
*Issue #, if available:*

*Description of changes:*

Major Additions:
- Repeated KFold Bagging
- Stacker max_models hyperparameters
- Memory Optimization / Checks
- Improved Logging, Leaderboard, Info
- Various Bug Fixes

### DIFF
--- a/autogluon/task/tabular_prediction/tabular_prediction.py
+++ b/autogluon/task/tabular_prediction/tabular_prediction.py
@@ -64,6 +64,7 @@ class TabularPrediction(BaseTask):
                                'KNN': {},
                                'custom': ['GBM'],
                               },
+            enable_fit_continuation=False,
             time_limits=None, num_trials=None, search_strategy='random', search_options={}, 
             nthreads_per_trial=None, ngpus_per_trial=None, dist_ip_addrs=[], visualizer='none',
             verbosity=2, **kwargs):
@@ -142,7 +143,10 @@ class TabularPrediction(BaseTask):
         stack_ensemble_levels : (int)
             Number of stacking levels to use in stack ensemble. Roughly increases model training time by factor of `stack_ensemble_levels+1` (set = 0 to disable stack ensembling). 
             Disabled by default, but we recommend values between 1-3 to maximize predictive performance. 
-            To prevent overfitting, this argument is ignored unless you have also set `num_bagging_folds >= 2`. 
+            To prevent overfitting, this argument is ignored unless you have also set `num_bagging_folds >= 2`.
+        enable_fit_continuation : (bool, default = False)
+            Whether to enable predictor to be able to be fit beyond the initial fit call.
+            When enabled, the training and validation data are saved to disk.
         time_limits : (int)
             Approximately how long `fit()` should run for (wallclock time in seconds). 
             `fit()` will stop training new models after this amount of time has elapsed (but models which have already started training will continue to completion). 
@@ -301,5 +305,5 @@ class TabularPrediction(BaseTask):
         learner.fit(X=train_data, X_test=tuning_data, scheduler_options=scheduler_options,
                       hyperparameter_tune=hyperparameter_tune, feature_prune=feature_prune,
                       holdout_frac=holdout_frac, num_bagging_folds=num_bagging_folds, stack_ensemble_levels=stack_ensemble_levels, 
-                      hyperparameters=hyperparameters, time_limit=time_limits_orig, verbosity=verbosity)
+                      hyperparameters=hyperparameters, time_limit=time_limits_orig, save_data=enable_fit_continuation, verbosity=verbosity)
         return TabularPredictor(learner=learner)
--- a/autogluon/task/tabular_prediction/tabular_prediction.py
+++ b/autogluon/task/tabular_prediction/tabular_prediction.py
@@ -142,7 +142,8 @@ class TabularPrediction(BaseTask):
             Disabled by default, but we recommend values between 5-10 to maximize predictive performance.
         num_bagging_sets : (int)
             Number of repeats of kfold bagging to perform. Total number of models trained during bagging = num_bagging_folds * num_bagging_sets.
-            Defaults to 1. Values greater than 1 will result in superior predictive performance, especially on smaller problems.
+            Defaults to 1 if time_limits is not specified, otherwise 10.
+            Values greater than 1 will result in superior predictive performance, especially on smaller problems and with stacking enabled.
             Disabled if num_bagging_folds is not specified.
             Value must be >= 1 if specified
         stack_ensemble_levels : (int)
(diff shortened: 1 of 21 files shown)


tests/unittests/test_tabular.py
```diff
@@ -128,13 +129,14 @@ def run_tabular_benchmarks(fast_benchmark, subsample_size, perf_threshold, seed_
...
-            directory = dataset['name'] + "/"
+            directory = directory_prefix + dataset['name'] + "/"
             train_file_path = directory + train_file
             test_file_path = directory + test_file
             if (not os.path.exists(train_file_path)) or (not os.path.exists(test_file_path)):
                 # fetch files from s3:
                 print("%s data not found locally, so fetching from %s" % (dataset['name'],  dataset['url']))
-                os.system("wget " + dataset['url'] + " -O temp.zip && unzip -o temp.zip && rm temp.zip")
+                os.makedirs(os.path.dirname(directory_prefix), exist_ok=True)
+                os.system("wget " + dataset['url'] + " -O temp.zip && unzip -o temp.zip -d " + directory_prefix + " && rm temp.zip")
```
zhanghang1989: you may use `foldername=ag.unzip(ag.download(filename))`

szha:
Job PR-177-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-177/2/index.html

tests/unittests/test_tabular.py
Innixma: Done, thanks!

autogluon/task/tabular_prediction/tabular_prediction.py
```diff
@@ -102,7 +103,12 @@ def fit(train_data, label, tuning_data=None, output_directory=None, problem_type
         hyperparameter_tune : (bool, default = False)
             Whether to tune hyperparameters or just use fixed hyperparameter values for each model. Setting as True will increase `fit()` runtimes.
         feature_prune : (bool, default = False)
-            Whether or not to perform feature selection. 
+            Whether or not to perform feature selection.
+        auto_stack : (bool, default = False)
+            Whether to have AutoGluon automatically attempt to select optimal num_bagging_folds and stack_ensemble_levels based on data properties.
+            Note: Overrides num_bagging_folds and stack_ensemble_levels values.
```
jwmueller: Assuming we keep `auto_stack` as is, does this override num_bagging_sets as well?  
It seems like this value should also be automatically set by `auto_stack`

autogluon/task/tabular_prediction/tabular_prediction.py
```diff
@@ -54,7 +54,7 @@ def load(output_directory, verbosity=2):
     
     @staticmethod
     def fit(train_data, label, tuning_data=None, output_directory=None, problem_type=None, eval_metric=None,
-            hyperparameter_tune=False, feature_prune=False, holdout_frac=None, num_bagging_folds=0, stack_ensemble_levels=0,
+            hyperparameter_tune=False, feature_prune=False, auto_stack=False, holdout_frac=None, num_bagging_folds=0, num_bagging_sets=None, stack_ensemble_levels=0,
```
jwmueller: Perhaps it will be cleaner in the future if we expand the scope of `auto_stack` beyond stacking to automatically decide on all fit() arguments?  

One could instead move it to be the first argument after `eval_metric` and call it something like:  'optimize_args', which when = True could overwrite all the accuracy-affecting arguments including:  `hyperparameters`, `hyperparameter_tune`, `feature_prune`, `holdout_frac`,  `num_bagging_folds`, `num_bagging_sets`, `stack_ensemble_levels`, `search_strategy`, `search_options`.  Of course, for now this argument can just overwrite the stacking-related arguments.  Essentially `optimize_args` would try to optimize absolutely everything for accuracy within the allotted time_limits (if `time_limits` is off, then fit() with optimize_args=True might be expected to run much slower).

autogluon/task/tabular_prediction/tabular_prediction.py
```diff
@@ -133,11 +139,20 @@ def fit(train_data, label, tuning_data=None, output_directory=None, problem_type
             Default value is 0.2 if `hyperparameter_tune = True`, otherwise 0.1 is used by default. 
         num_bagging_folds : (int)
             Number of folds used for bagging of models. When `num_bagging_folds = k`, training time is roughly increased by a factor of `k` (set = 0 to disable bagging). 
-            Disabled by default, but we recommend values between 5-10 to maximize predictive performance. 
+            Disabled by default, but we recommend values between 5-10 to maximize predictive performance.
+        num_bagging_sets : (int)
+            Number of repeats of kfold bagging to perform. Total number of models trained during bagging = num_bagging_folds * num_bagging_sets.
```
jwmueller: Can we add one more documentation sentence describing why `num_bagging_sets` > 1 is useful vs say just setting `num_bagging_folds` to a large number like 100?

setup.py
```diff
@@ -45,6 +45,7 @@ def run(self):
 MIN_PYTHON_VERSION = '>=3.6.*'
 
 requirements = [
+    'Pillow==6.2.1',
```
jwmueller: @zhanghang1989  told me this should be: 'Pillow<=6.2.1',

autogluon/task/tabular_prediction/tabular_prediction.py
Innixma: num_bagging_sets is extremely time consuming, which is why it isn't controlled by auto_stack currently. It is set to 1 by default and 10 by default if time_limits is set.

Innixma: The one argument for setting it is to set it above 1 when the data has few rows to help mitigate overfitting, but I think we should save that optimization for later, since this requires a more thorough investigation into how num_bagging_sets improves results

setup.py
Innixma: Added

autogluon/task/tabular_prediction/tabular_prediction.py
Innixma: Added

autogluon/task/tabular_prediction/tabular_prediction.py
Innixma: I agree we should introduce this parameter in future, however I don't think it should be in this PR. I think auto_stack can remain as is and this new optimize_args parameter could override it once it is introduced.

szha:
Job PR-177-4 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-177/4/index.html

szha:
Job PR-177-5 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-177/5/index.html

Issue #192, opened 2020-01-13
Tutorial example crash


ZhangSanMo:
**Working Env: Windows/Python 3.7.3/Jupyter NoteBook**
----
Runnig Code:
``` python
predictor_age = task.fit(train_data=train_data, output_directory="agModels-predictAge", label=age_column)
performance = predictor_age.evaluate(test_data)
```

Error Info:
``` 
Beginning AutoGluon training ...
Preprocessing data ...
Here are the first 10 unique label values in your data:  [25 23 46 55 36 51 33 18 43 41]
---------------------------------------------------------------------------
NotImplementedError                       Traceback (most recent call last)
<ipython-input-29-423783e2f952> in <module>
----> 1 predictor_age = task.fit(train_data=train_data, output_directory="agModels-predictAge", label="age")
      2 performance = predictor_age.evaluate(test_data)

C:\ProgramData\Miniconda3\lib\site-packages\autogluon\task\tabular_prediction\tabular_prediction.py in fit(train_data, label, tuning_data, output_directory, problem_type, eval_metric, hyperparameter_tune, feature_prune, holdout_frac, num_bagging_folds, stack_ensemble_levels, hyperparameters, time_limits, num_trials, search_strategy, search_options, nthreads_per_trial, ngpus_per_trial, dist_ip_addrs, visualizer, verbosity, **kwargs)
    288                       hyperparameter_tune=hyperparameter_tune, feature_prune=feature_prune,
    289                       holdout_frac=holdout_frac, num_bagging_folds=num_bagging_folds, stack_ensemble_levels=stack_ensemble_levels,
--> 290                       hyperparameters=hyperparameters, time_limit=time_limits_orig, verbosity=verbosity)
    291         return TabularPredictor(learner=learner)

C:\ProgramData\Miniconda3\lib\site-packages\autogluon\utils\tabular\ml\learner\default_learner.py in fit(self, X, X_test, scheduler_options, hyperparameter_tune, feature_prune, holdout_frac, num_bagging_folds, stack_ensemble_levels, hyperparameters, time_limit, verbosity)
     48         time_preprocessing_start = time.time()
     49         logger.log(20, 'Preprocessing data ...')
---> 50         X, y, X_test, y_test = self.general_data_processing(X, X_test)
     51         time_preprocessing_end = time.time()
     52         time_preprocessing = time_preprocessing_end - time_preprocessing_start

[...]

PR #197, opened 2020-01-14
Fixed problem_type detection on Windows

Innixma:
*Issue #, if available:*

#192 

*Description of changes:*

Made type detection generic using np.issubdtype, this enables AutoGluon to handle label dtypes of any kind of integer and float instead of only int64 and float64, which was causing issues on Windows.


szha:
Job PR-197-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-197/1/index.html

PR #257, opened 2020-02-01
PEP8 style, wrapping of long lines

larroy:
*Issue #, if available:*

*Description of changes:*

Cosmetic, we should add linting for PEP8


szha:
Job PR-257-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-257/1/index.html

szha:
Job PR-257-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-257/2/index.html

Innixma:
I'm somewhat on the fence on this.

Personally, I feel that having an entire idea/statement in a single line greatly helps the reader understand where an idea starts and ends. By wrapping lines like this, it can get confusing.

As an example:

```
y_pred : list or `numpy.array`
    The ordered collection of predictions. 
    For certain types of `eval_metric` (such as AUC), `y_pred` must be predicted-probabilities rather than predicted labels.
```

VS

```
y_pred : list or `numpy.array`
    The ordered collection of predictions.
    For certain types of `eval_metric` (such as AUC), `y_pred` must be predicted-probabilities rather than
    the predicted labels.
```

I personally find the first (PEP8 style violating) version to be more readable, especially if we establish a style where one-idea is present per line of comment, this way a reader can quickly glance to see that they need to be aware of N aspects of this functionality, where N is the number of lines of comments.

I am open to suggestions and counter-examples on this, but I find that aggressive line-wrapping, especially in code, to be very distracting and makes both writing and reading code more cumbersome.

larroy:
I agree with you in principle but I think we should set automatic linter check, the benefits outweight the negatives like the ones you mention.  On the documentation, the line wrapping should make no different when the docs are rendered right?

PR #279, opened 2020-02-10
Basic pep 8 corrections, bug fix in string formatting. 

razmikmelikbekyan:
*Issue #, if available:*

*Description of changes:*

Basic pep 8 corrections:
 - imports optimisation - https://www.python.org/dev/peps/pep-0008/#imports
 - surround top-level function and class definitions with two blank lines - https://www.python.org/dev/peps/pep-0008/#blank-lines
Bug fix in string formating.

### DIFF
--- a/autogluon/task/base/base_task.py
+++ b/autogluon/task/base/base_task.py
@@ -1,10 +1,11 @@
-import time
-import copy
 import collections
-import mxnet as mx
+import copy
+import time
 from abc import abstractmethod
+
+import mxnet as mx
+
 from ...scheduler import *
-from .base_predictor import *
 from ...utils import in_ipynb
 
 __all__ = ['BaseDataset', 'BaseTask']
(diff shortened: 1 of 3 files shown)


autogluon/task/base/base_task.py
```diff
@@ -66,6 +70,7 @@ def run_fit(cls, train_fn, search_strategy, scheduler_options):
                        args=record_args)
         return results
 
+    @classmethod
     @abstractmethod
```
Innixma: Shouldn't @abstractmethod be removed if it is a @classmethod?

razmikmelikbekyan: It is not obligatory, they can stay together. This means that `fit` method is a `abstractmethod` so all children must implement it and on the other hand with `classmethod` you indicate that it children should implement it as a `classmethod`. See the reference : https://docs.python.org/3.6/library/abc.html#abc.abstractmethod

szha:
Job PR-279-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-279/1/index.html

autogluon/task/base/base_task.py
Innixma: I see, sorry I misread abstractmethod to mean staticmethod. Thanks for the explaination!

PR #284, opened 2020-02-11
Basic pep 8 corrections, type checking correction.

razmikmelikbekyan:
*Issue #, if available:*

*Description of changes:*

Basic pep 8 corrections:
 -imports optimisation - https://www.python.org/dev/peps/pep-0008/#imports
 -Inline comments should be separated by at least two spaces from the statement - https://www.python.org/dev/peps/pep-0008/#inline-comments

Changing type checking to use `isinstance`.
Performing full check for given `dataset` object type.

### DIFF
--- a/autogluon/task/base/base_predictor.py
+++ b/autogluon/task/base/base_predictor.py
@@ -14,9 +14,11 @@
     - task object should have get_labels(Dataset) method
 """
 
+import json
 import logging
+import pickle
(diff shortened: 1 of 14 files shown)


szha:
Job PR-284-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-284/1/index.html

PR #294, opened 2020-02-13
remove redundant copies of dataset object in tabular neural net

jwmueller:
Makes tabular neural data processing more memory efficient. 
There still exist a duplicate copy of dataset inside the initialization of tabular_nn_dataset, when the data are converted to mx.nd.array, but that duplicate is not easy to remove.


szha:
Job PR-294-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-294/1/index.html

Issue #302, opened 2020-02-15
Multi-threaded Tabular embednet over 50% slower on MacOS Laptop than prior


Innixma:
Tested on MacOS Laptop with 2 physical cores, 4 threads

Multi-thread Speedup (#291) Benchmarking

adult dataset, 5000 sample rows:
Enabled:
Train time: 30.230026s
Inference time: 1.180679s

Disabled:
Train time: 18.27790s
Inference time: 0.809626s

adult dataset, 39073 rows (Full):
Enabled:
Train time: 229.882597s
Inference time: 2.202588

Disabled:
Train time: 138.790045s
Inference time: 0.977363s

Why is this? When does multi-threaded help? Can we add a toggle based on expected best option?

Also, while enabled on the full dataset, output is spammed with these log messages:

```
distributed.utils_perf - WARNING - full garbage collections took 12% CPU time recently (threshold: 10%)
distributed.utils_perf - WARNING - full garbage collections took 11% CPU time recently (threshold: 10%)
distributed.utils_perf - WARNING - full garbage collections took 11% CPU time recently (threshold: 10%)
```

Innixma:
Testing on m5.2xlarge, full adult dataset:
Enabled:
Train time: 163.131340s
Inference time: 1.02232s

Disabled:
Train time: 82.923477s
Inference time: 0.513330s

Neural Network Architecture:

```
Training data for neural network has: 36573 examples, 14 features (7 vector, 7 embedding, 0 language)
Training neural network for up to 500 epochs...
Neural network architecture:
EmbedNet(
  (numeric_block): NumericBlock(
    (body): Dense(None -> 160, Activation(relu))
  )
  (embed_blocks): HybridSequential(
    (0): EmbedBlock(
      (body): Embedding(10 -> 5, float32)
    )
    (1): EmbedBlock(
      (body): Embedding(17 -> 7, float32)
    )
    (2): EmbedBlock(
      (body): Embedding(8 -> 5, float32)
    )
    (3): EmbedBlock(
      (body): Embedding(16 -> 7, float32)
    )
    (4): EmbedBlock(
      (body): Embedding(7 -> 4, float32)
    )
    (5): EmbedBlock(
      (body): Embedding(6 -> 4, float32)
    )
    (6): EmbedBlock(
      (body): Embedding(43 -> 13, float32)
    )
  )
  (output_block): WideAndDeepBlock(
    (deep): FeedforwardBlock(
      (body): HybridSequential(
[...]

PR #291, opened 2020-02-13
[WIP] Accelerating Tabular NN

zhanghang1989:
*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #303, opened 2020-02-15
Revert #291

Innixma:
*Issue #, if available:*

#302

*Description of changes:*

Reverts #291 until performance evaluation can be more comprehensively tested.

### DIFF
--- a/autogluon/utils/tabular/ml/models/tabular_nn/embednet.py
+++ b/autogluon/utils/tabular/ml/models/tabular_nn/embednet.py
@@ -1,4 +1,3 @@
-import threading
 import numpy as np
 import mxnet as mx
 from mxnet import nd, gluon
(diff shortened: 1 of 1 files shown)


szha:
Job PR-303-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-303/1/index.html

zhanghang1989:
I am okay if you prefer to revert the code. But as I mentioned in the issue https://github.com/awslabs/autogluon/issues/302#issuecomment-586557126

If the machine does not have enough number of cores to handle multi-processing. There won't be any speed up. A mac book only has 4 cores.

@jwmueller

zhanghang1989:
The speedup only applies to the datasets with multiple embedding layers, which can be parallelized.

Issue #266, opened 2020-02-05
Improve Tabular Hyperparameter Tuning Stability


Innixma:
Master Issue for Tabular hyperparameter tuning improvement tracking.

If you are a user experiencing issues, we recommend to use **auto_stack=True**, **hyperparameter_tune=False** for best stable performance.

Currently hyperparameter tuning suffers from various issues, generally due to multithreading.

Related Issues:
#245 
#231 
#219 
#139 
#329

razmikmelikbekyan:
Hi, 

I am facing the following issue when running hyperparameters tuning. 
When there are  errors in all individual trial threads `autogluon.utils.tabular.ml.models.abstract.model_trial.model_trial` in `autogluon/scheduler/fifo.py:223` the main process is not finishing. I do not know exactly if someone else faced similar issues or not that is why I decided to add it as a comment. 
For reproducing it you can run the following: 
```
import autogluon as ag
from autogluon.task import TabularPrediction as task

hp_tune = True
cat_options = {
    'num_leaves': ag.space.Int(lower=26, upper=66, default=36),
}

hyperparameters = {
    'CAT': cat_options
}

time_limits = 2 * 60
num_trials = 2
search_strategy = 'skopt'
output_directory = 'AutogluonModels/agModels-predictTitanic-tuning'
data = task.Dataset(file_path='AutogluonModels/titanic/train.csv')
label_column = 'Survived'
id_columns = ['PassengerId']

predictor = task.fit(
    train_data=data,
    holdout_frac=0.3,
    label=label_column,
    output_directory=output_directory,
    time_limits=time_limits,
    num_trials=num_trials,
    hyperparameter_tune=hp_tune,
    hyperparameters=hyperparameters,
    search_strategy=search_strategy,
    auto_stack=False,
    stack_ensemble_levels=0
)

print(predictor.leaderboard(silent=True))
```

razmikmelikbekyan:
Basically I want to ask also, if someone is working on this issue on general and from where it is possible to find more detailed documentation about distributed resource management of AutoGluon to be able to solve this.

Innixma:
@razmikmelikbekyan According to https://catboost.ai/docs/concepts/python-reference_parameters-list.html, CatBoost num_leaves parameter is not valid on CPU.

The part about the main job not finishing is an issue however, it should simply fail on CatBoost and return.

[...]

PR #313, opened 2020-02-19
FIX: training will not hang if in task an error occurs

razmikmelikbekyan:
*Issue #266, if available:*

*Description of changes:*
Checking queue size before calling fetch. When we call fetch and the process which is inside the queue fails, then we have deadlock: process is raising exception, put the thread which opened this process is not raising an exception.
Basic PEP8 corrections.

### DIFF
--- a/autogluon/core/task.py
+++ b/autogluon/core/task.py
@@ -42,7 +42,7 @@ class Task:
 
     @classmethod
     def set_id(cls, taskid):
-        logger.info(f'Seting TASK ID: {taskid}')
+        logger.info(f'Setting TASK ID: {taskid}')
         cls.TASK_ID.value = taskid
 
     def __repr__(self):
--- a/autogluon/core/task.py
+++ b/autogluon/core/task.py
@@ -1,20 +1,21 @@
+import argparse
 import copy
 import logging
-import argparse
 import multiprocessing as mp
 
 logger = logging.getLogger(__name__)
 
 __all__ = ['Task']
 
-class Task(object):
-    """Individual training task, containing the lauch function, default arguments and
+
+class Task:
+    """Individual training task, containing the launch function, default arguments and
     required resources.
 
     Args:
-        fn (callable): Lauch function for the training task.
+        fn (callable): Launch function for the training task.
         args (argparse.ArgumentParser): Default function arguments.
-        resources (autogluon.scheduler.Resources): Required resources for lauching the task.
+        resources (autogluon.scheduler.Resources): Required resources for launching the task.
 
     Example:
         >>> def my_task():
@@ -24,6 +25,7 @@ class Task(object):
     """
     TASK_ID = mp.Value('i', 0)
     LOCK = mp.Lock()
+
     def __init__(self, fn, args, resources):
         self.fn = fn
         self.args = copy.deepcopy(args)
@@ -40,17 +42,20 @@ class Task(object):
 
     @classmethod
     def set_id(cls, taskid):
-        logger.info('Seting TASK ID: {}'.format(taskid))
+        logger.info(f'Seting TASK ID: {taskid}')
         cls.TASK_ID.value = taskid
 
     def __repr__(self):
-        reprstr = self.__class__.__name__ +  \
-            ' (' + 'task_id: ' + str(self.task_id) + \
-            ',\n\tfn: ' + str(self.fn) + \
-            ',\n\targs: {'
+        reprstr = (
+            f'{self.__class__.__name__} ('
+            f'task_id: {self.task_id},'
+            f'\n\tfn: {self.fn},'
+            f'\n\targs: {{'
+        )
+
         for k, v in self.args.items():
             data = str(v)
             info = (data[:100] + '..') if len(data) > 100 else data
-            reprstr +=  '{}'.format(k) + ': ' + info + ', '
-        reprstr +=  '},\n\tresource: ' + str(self.resources) + ')\n'
+            reprstr += f'{k}: {info}, '
+        reprstr += f'}},\n\tresource: {self.resources})\n'
         return reprstr
--- a/autogluon/scheduler/fifo.py
+++ b/autogluon/scheduler/fifo.py
@@ -239,6 +239,7 @@ class FIFOScheduler(TaskScheduler):
         rp.start()
         task_dict = self._dict_from_task(task)
         task_dict.update({'Task': task, 'Job': job, 'ReporterThread': rp})
+
         # checkpoint thread
         if self._checkpoint is not None:
             def _save_checkpoint_callback(fut):
@@ -257,6 +258,9 @@ class FIFOScheduler(TaskScheduler):
         last_result = None
         while not task_job.done():
             reported_result = reporter.fetch()
+            if reported_result is None:
+                continue
+
             if reported_result.get('done', False):
                 reporter.move_on()
                 break
@@ -264,6 +268,7 @@ class FIFOScheduler(TaskScheduler):
                 task.task_id, reported_result, config=task.args['config'])
             reporter.move_on()
             last_result = reported_result
+
         if last_result is not None:
             last_result['done'] = True
             searcher.update(
--- a/autogluon/scheduler/hyperband.py
+++ b/autogluon/scheduler/hyperband.py
@@ -1,14 +1,13 @@
-import pickle
 import logging
+import pickle
 import threading
+
 import numpy as np
-import multiprocessing as mp
 
 from .fifo import FIFOScheduler
-from .hyperband_stopping import HyperbandStopping_Manager
 from .hyperband_promotion import HyperbandPromotion_Manager
-from .reporter import DistStatusReporter, DistSemaphore
-from ..utils import DeprecationHelper
+from .hyperband_stopping import HyperbandStopping_Manager
+from .reporter import DistStatusReporter
 
 __all__ = ['HyperbandScheduler', 'HyperbandStopping_Manager', 'HyperbandPromotion_Manager']
 
@@ -219,6 +218,9 @@ class HyperbandScheduler(FIFOScheduler):
         last_updated = None
         while not task_job.done():
             reported_result = reporter.fetch()
+            if reported_result is None:
+                continue
+
             if reported_result.get('done', False):
                 reporter.move_on()
                 terminator.on_task_complete(task, last_result)
--- a/autogluon/scheduler/reporter.py
+++ b/autogluon/scheduler/reporter.py
@@ -64,6 +64,9 @@ class DistStatusReporter(object):
             raise AutoGluonEarlyStop('Stopping!')
 
     def fetch(self, block=True):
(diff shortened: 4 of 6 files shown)


szha:
Job PR-313-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-313/2/index.html

autogluon/core/task.py
```diff
@@ -40,17 +42,20 @@ def __init__(self, fn, args, resources):
 
     @classmethod
     def set_id(cls, taskid):
-        logger.info('Seting TASK ID: {}'.format(taskid))
+        logger.info(f'Seting TASK ID: {taskid}')
```
Innixma: Super minor, but I noticed old typo here 'Seting' -> 'Setting'

zhanghang1989:
This PR mainly addressed some coding style and looks good to me.

Why it fix training hang?

autogluon/scheduler/reporter.py
```diff
@@ -64,6 +64,9 @@ def __call__(self, **kwargs):
             raise AutoGluonEarlyStop('Stopping!')
 
     def fetch(self, block=True):
+        if not self._queue.qsize():
```
jwmueller: @zhanghang1989 
It appears this is the key part of the fix, are you able to verify this is safe?

zhanghang1989: Thanks @jwmueller  for reminding. This change is not safe. Basically it breaks all the search algorithms except for random search. Because we need the blocking call of `self._queue.get()` to wait for next trial finishes.

zhanghang1989:
Beyond the breaking of blocking call of `reporter.fetch()`, this PR looks okay for me.

@mseeger could you also review this PR and see if it breaks the backend?

autogluon/scheduler/fifo.py
```diff
@@ -257,13 +258,17 @@ def _run_reporter(self, task, task_job, reporter, searcher):
         last_result = None
         while not task_job.done():
             reported_result = reporter.fetch()
+            if reported_result is None:
```
razmikmelikbekyan: @zhanghang1989 this is the while loop, which will continue calling `queue.get` untill it will get something from there or `task_job.done() is True`

autogluon/scheduler/reporter.py
razmikmelikbekyan: @zhanghang1989 thanks for review. 

As far as I understand the logic is the following: for each training task we have a thread which itself opens a process, please correct we if I am wrong. Then we collect data in Queue from child process and pass to main process. So the issue occurs when we try to call `queue.get`, but the child process is already died due to some reason (for example user passed wrong parameters to particular model). At that moment we have a deadlock.  I assume in order to solve it, there will be a need to change some logic in the multiprocessing / multithreading part of the code. With this change I just avoid calling `queue.get` in order to not have the deadlock. And it is not blocking algorithms, since it will try to call `queue.get` until it will get something from `queue` or the job is done.. If you look at changes in searchers, you will see that it is in `while` loop. I checked it for `skopt` and it works.

@Innixma please see the minimal code example, which will cause deadlock, without these changes. 

Tabular Prediction: 

```
import autogluon as ag
from autogluon.task import TabularPrediction as task

import sklearn.datasets
import pandas as pd

cancer_data = sklearn.datasets.load_breast_cancer(return_X_y=False)
data = pd.DataFrame(cancer_data['data'], columns=cancer_data['feature_names'])
data['target'] = cancer_data['target']

hp_tune = True
cat_options = {
    'num_leaves': ag.space.Int(lower=26, upper=66, default=36),
    'depth': ag.space.Int(lower=4, upper=10, default=6),
}
gbm_options = {
    'num_boost_round': 100,
    'num_leaves': ag.space.Int(lower=26, upper=66, default=36),
}

hyperparameters = {
    'CAT': cat_options,
    'GBM': gbm_options,
}

time_limits = 2 * 60
num_trials = 3
output_directory = './agModels--tuningCancerDataset'
data = task.Dataset(data)
label_column = 'target'

# case 1: random search
search_strategy = 'random'
print('\n' * 2)
print('#' * 100)
print(f'Case 1: Running HPO suing search strategy = "{search_strategy}".')
print()
predictor = task.fit(
    train_data=data,
    holdout_frac=0.3,
    label=label_column,
    output_directory=output_directory,
    time_limits=time_limits,
    num_trials=num_trials,
    hyperparameter_tune=hp_tune,
    auto_stack=False,
    stack_ensemble_levels=0
)

print('\n' * 2)
print('#' * 100)
print('HPO results:')
print(predictor.leaderboard(silent=True))

# case 2: skopt search
search_strategy = 'skopt'
print('\n' * 2)
print('#' * 100)
print(f'Case 2: Running HPO suing search strategy = "{search_strategy}".')
print()
predictor = task.fit(
    train_data=data,
    holdout_frac=0.3,
    label=label_column,
    output_directory=output_directory,
    time_limits=time_limits,
    num_trials=num_trials,
    hyperparameter_tune=hp_tune,
    auto_stack=False,
    stack_ensemble_levels=0
)
print('\n' * 2)
print('#' * 100)
print('HPO results:')
print(predictor.leaderboard(silent=True))
```
Here I put wrong parameter `num_leaves` for `catboost`: `catboost` does not support changing it when you run on CPU.

Image Classification: 
```
import autogluon as ag
from autogluon import ImageClassification as task

train_data = task.Dataset(name='fashionmnist')

classifier = task.fit(train_data,
                      num_trials=1,
                      net=None,
                      epochs=5,
                      ngpus_per_trial=1,
                      verbose=False,
                      auto_search=False)
```
Here I put wrong parameter `net=None` .

szha:
Job PR-313-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-313/3/index.html

mseeger:
Hello @zhanghang1989, thanks for looping me in.
I am not fully on top of the implementation details of the backend.

First, I'd ask the author to split cosmetic changes into a separate PR, and focus this PR on the essential changes in question. The author should also comment on @zhanghang1989 objections.

autogluon/scheduler/fifo.py
mseeger: @zhanghang1989 It comes down to whether a call of reporter.fetch() is blocking, in the sense it waits until fetch() returns something. I kind assumed this was the case.

If the call is not blocking, our current code is broken, because it would try to add a signal from the reporter even if fetch() returned an empty dict (I think this is the issue, not so much None).

mseeger: The reason I assumed that reporter.fetch() is blocking is that I see reporter.move_on() afterwards. If the reporter would simply just continue, why we would need that call?

autogluon/scheduler/hyperband.py
```diff
@@ -219,6 +218,9 @@ def _run_reporter(self, task, task_job, reporter, searcher, terminator):
         last_updated = None
         while not task_job.done():
             reported_result = reporter.fetch()
+            if reported_result is None:
```
mseeger: Same issues as above. I do think reporter.fetch() would not return None. The question is whether it would return an empty dict.

autogluon/scheduler/reporter.py
mseeger: OK, I see now. @zhanghang1989 

I think this IS a problem. With this insertion, a call to fetch() becomes non-blocking, right? Before, it seems that fetch() would not return anything (and wait) until somethere is there to return (please confirm).

But now, it immediately returns None if nothing is there to report, and this is not what we want. This would really break things for the schedulers.

To be clear, there are three different things that could happen when call fetch():
A) Nothing new to be reported
B) Something new to be reported, but not a final metric at the end of training
C) Something new to be reported, namely final metric

Currently, schedulers assume that for A), reporter.fetch() blocks, until B) or C). And scheduler then has to distinguish between them.

Is this blocking the scheduler from doing other things? No, because we have a reporter thread, and only that is blocked (it runs _run_reporter)

razmikmelikbekyan: What if the process is killed, which is supposed to return something in queue ? Then we have deadlock.

mseeger:
I only have a bit of a high-level view on the workings here.
Please do discuss this with @zhanghang1989 

Maybe it is the case that changing fetch() to return None sometimes, and then insert continue if None is received in _run_reporter of several different schedulers keeps everything working, and on top serves a purpose I do not understand. If so, we should change it. For now, I do not see why reporter.fetch() should not wait until something can be returned. This may be my ignorance.

autogluon/scheduler/fifo.py
zhanghang1989: @mseeger Thanks for reviewing the PR.` reporter.fetch()` is a blocking call. This PR does break our code.

Innixma:
@zhanghang1989 Could you produce an example where this change breaks functionality that previously worked, as Razmik shows an example where his change has a positive impact on Tabular where previously it resulted in deadlock.

zhanghang1989:
> @zhanghang1989 Could you produce an example where this change breaks functionality that previously worked, as Razmik shows an example where his change has a positive impact on Tabular where previously it resulted in deadlock.

Basically the searcher state won't get updated in real example, because the results are not ready in reporter fetch. Only random searcher works as expected.

Innixma:
I understand the claim, but can you provide a code example of this not working as intended such that we can work towards a solution where both your example and Razmik's example work correctly?

zhanghang1989:
The easiest way to handle dead trials is adding a time_out, which I can help with.

Innixma:
We shouldn't be doing that, they aren't dead trials, they throw an exception which can be observed. We want to avoid the deadlock entirely.

mseeger:
Hello,

here is my view, which may be inaccurate:

I think the change does not break our code, because two things happen:
- reporter.fetch() returns None if nothing is to be reported, so the call becomes non-blocking
- But then, _run_reporter in the scheduler do continue if the result is None, so the blocking is
   simply moved to the _run_reporter loop

My real question would be why this is useful, because it clearly makes _run_reporter implementations a little more complex: we always have to remember to react to None in the right way.

@razmikmelikbekyan Could you explain to us (best with an example) why the added complexity actually helps over the previous code, where reporter.fetch() was blocking? I am not experienced in distributed systems.

razmikmelikbekyan:
> Hello,
> 
> here is my view, which may be inaccurate:
> 
> I think the change does not break our code, because two things happen:
> 
> * reporter.fetch() returns None if nothing is to be reported, so the call becomes non-blocking
> * But then, _run_reporter in the scheduler do continue if the result is None, so the blocking is
>   simply moved to the _run_reporter loop
> 
> My real question would be why this is useful, because it clearly makes _run_reporter implementations a little more complex: we always have to remember to react to None in the right way.
> 
> @razmikmelikbekyan Could you explain to us (best with an example) why the added complexity actually helps over the previous code, where reporter.fetch() was blocking? I am not experienced in distributed systems.

@Innixma , @mseeger , @zhanghang1989 
Please refer to PR #322

PR #322, opened 2020-02-21
FIX: tabular training hang due to error in model training

razmikmelikbekyan:
*Issue #266, if available:*

*Description of changes:*
Terminating reporter if tabular model failed to train due to some reason.

#### The code sample which was not working before the change:
```
import autogluon as ag
from autogluon.task import TabularPrediction as task

import sklearn.datasets
import pandas as pd

cancer_data = sklearn.datasets.load_breast_cancer(return_X_y=False)
data = pd.DataFrame(cancer_data['data'], columns=cancer_data['feature_names'])
data['target'] = cancer_data['target']

hp_tune = True
cat_options = {
    'num_leaves': ag.space.Int(lower=26, upper=66, default=36),
    'depth': ag.space.Int(lower=4, upper=10, default=6),
}
gbm_options = {
    'num_boost_round': 100,
    'num_leaves': ag.space.Int(lower=26, upper=66, default=36),
}

hyperparameters = {
    'CAT': cat_options,
    'GBM': gbm_options,
}

time_limits = 2 * 60
num_trials = 3
output_directory = './agModels--tuningCancerDataset'
data = task.Dataset(data)
label_column = 'target'

# case 1: random search
search_strategy = 'random'
print('\n' * 2)
print('#' * 100)
print(f'Case 1: Running HPO suing search strategy = "{search_strategy}".')
print()
predictor = task.fit(
    train_data=data,
    holdout_frac=0.3,
    label=label_column,
    output_directory=output_directory,
    time_limits=time_limits,
    num_trials=num_trials,
    hyperparameter_tune=hp_tune,
    auto_stack=False,
    stack_ensemble_levels=0
)
```

After detailed investigation I understood the core issue. So please find the description below. If I am wrong, please correct me. 

-  We call `autogluon.scheduler.scheduler.TaskScheduler._run_dist_job` method and by doing this we create a separate `process` for training our model. 
- For `Tabular Prediction` it calls the function `autogluon.utils.tabular.ml.models.abstract.model_trial.model_trial` https://github.com/awslabs/autogluon/blob/ddb167bdf835182bed4aec7752f21f80ae7dcc75/autogluon/utils/tabular/ml/models/abstract/model_trial.py#L12. 
-  It is passing also `reporter` object to this function.

##### Scenario A:  there is no any exception before calling reporter

- At the end of `model_trial` function we call `autogluon.scheduler.reporter.DistStatusReporter.__call__` https://github.com/awslabs/autogluon/blob/ddb167bdf835182bed4aec7752f21f80ae7dcc75/autogluon/scheduler/reporter.py#L42 or `autogluon.scheduler.reporter.LocalStatusReporter.__call__` https://github.com/awslabs/autogluon/blob/ddb167bdf835182bed4aec7752f21f80ae7dcc75/autogluon/scheduler/reporter.py#L113 which will put data into reporter queue. 
- After that we call `_run_rporter` of scheduler (for example `autogluon.scheduler.fifo.FIFOScheduler._run_reporter`). Which will try to get data from `reporter queue` with `lock`. 
- It will return some results from model training, since `Reporter.__call__` is called.

##### Scenario B:  there is an exception before calling reporter

- `autogluon.scheduler.reporter.DistStatusReporter.__call__` https://github.com/awslabs/autogluon/blob/ddb167bdf835182bed4aec7752f21f80ae7dcc75/autogluon/scheduler/reporter.py#L42 or `autogluon.scheduler.reporter.LocalStatusReporter.__call__` https://github.com/awslabs/autogluon/blob/ddb167bdf835182bed4aec7752f21f80ae7dcc75/autogluon/scheduler/reporter.py#L113 are **NOT** called, so queue is not filled: it will be **always empty**.
- After that we call `_run_rporter` of scheduler (for example `autogluon.scheduler.fifo.FIFOScheduler._run_reporter`). Which will try to get data from `reporter queue` with `lock`.
- It will cause `deadlock`, since `Reporter.__call__` is not called.

So in order to eliminate Scenario B I added `try except` and in case of exception terminated the reporter. 

I believe that this solution is much safer than returning `None`. The latter was also working as expected and as @mseeger noticed today, was not breaking block searcher update, but it was simply transferring it to `while` loop. 

If you are agree with this change, I assume that we should put similar `try except` calls in other training functions, for example : `autogluon.task.image_classification.pipeline.train_image_classification` https://github.com/awslabs/autogluon/blob/ddb167bdf835182bed4aec7752f21f80ae7dcc75/autogluon/task/image_classification/pipeline.py#L19.

### DIFF
--- a/autogluon/utils/tabular/ml/models/abstract/model_trial.py
+++ b/autogluon/utils/tabular/ml/models/abstract/model_trial.py
@@ -8,6 +8,7 @@ from autogluon.core import args
 
 logger = logging.getLogger(__name__)  # TODO: Currently unused
 
+
 @args()
 def model_trial(args, reporter):
     """ Training script for hyperparameter evaluation of an arbitrary model that subclasses AbstractModel.
@@ -19,27 +20,33 @@ def model_trial(args, reporter):
                 'num_threads', 'num_gpus' to set specific resources in model.fit()
             - model.save() must have return_filename, file_prefix, directory options
     """
-    trial_id = args.pop('task_id')  # Note may not start at 0 if HPO has been run for other models with same scheduler
-    directory = args.pop('directory')  # TODO: Separate model parameters vs HPO
-    file_prefix = f"trial_{trial_id}_"  # append to all file names created during this trial. Do NOT change!
-    model = args.pop('model')  # the model object must be passed into model_trial() here
+    try:
+        trial_id = args.pop('task_id')  # Note may not start at 0 if HPO has been run for other models with same scheduler
+        directory = args.pop('directory')  # TODO: Separate model parameters vs HPO
+        file_prefix = f"trial_{trial_id}_"  # append to all file names created during this trial. Do NOT change!
+        model = args.pop('model')  # the model object must be passed into model_trial() here
 
-    dataset_train_filename = args.pop('dataset_train_filename')
-    X_train, Y_train = pickle.load(open(directory + dataset_train_filename, 'rb'))
-    dataset_val_filename = args.pop('dataset_val_filename', None)
-    if dataset_val_filename is None:
-        raise NotImplementedError("must provide validation data for model_trial()")
-    X_val, Y_val = pickle.load(open(directory + dataset_val_filename, 'rb'))
-    model.params = model.params.copy()  # all hyperparameters must be stored in this dict
-    model.params.update(args)
-    if 'seed_value' in model.params:
-        random.seed(model.params['seed_value'])
-        np.random.seed(model.params['seed_value'])
+        dataset_train_filename = args.pop('dataset_train_filename')
+        X_train, Y_train = pickle.load(open(directory + dataset_train_filename, 'rb'))
+        dataset_val_filename = args.pop('dataset_val_filename', None)
+        if dataset_val_filename is None:
+            raise NotImplementedError("must provide validation data for model_trial()")
+        X_val, Y_val = pickle.load(open(directory + dataset_val_filename, 'rb'))
+        model.params = model.params.copy()  # all hyperparameters must be stored in this dict
+        model.params.update(args)
+        if 'seed_value' in model.params:
+            random.seed(model.params['seed_value'])
+            np.random.seed(model.params['seed_value'])
 
-        model.params.pop('seed_value')  # TODO: Should we be popping this?
+            model.params.pop('seed_value')  # TODO: Should we be popping this?
 
-    model.fit(X_train, Y_train, X_val, Y_val)
-    val_perf = model.score(X_val, Y_val)
-    trial_model_file = model.save(file_prefix=file_prefix, directory=directory, return_filename=True)
-    reporter(epoch=0, validation_performance=val_perf, directory=directory, file_prefix=file_prefix, 
-             trial_model_file=trial_model_file)  #  auxiliary_info = ?)
+        model.fit(X_train, Y_train, X_val, Y_val)
+        val_perf = model.score(X_val, Y_val)
+        trial_model_file = model.save(file_prefix=file_prefix, directory=directory, return_filename=True)
+
+    except Exception as e:
+        logger.exception(e, exc_info=True)
+        reporter.terminate()
+    else:
+        reporter(epoch=0, validation_performance=val_perf, directory=directory, file_prefix=file_prefix,
+                 trial_model_file=trial_model_file)  # auxiliary_info = ?)
--- a/autogluon/utils/tabular/ml/models/lgb/hyperparameters/lgb_trial.py
+++ b/autogluon/utils/tabular/ml/models/lgb/hyperparameters/lgb_trial.py
@@ -12,76 +12,81 @@ logger = logging.getLogger(__name__)  # TODO: Currently unused
 @args()
 def lgb_trial(args, reporter):
     """ Training script for hyperparameter evaluation of Gradient Boosting model """
-    try_import_lightgbm()
-    import lightgbm as lgb
-    # list of args which are not model hyperparameters:
-    nonparam_args = set(['directory', 'task_id', 'lgb_model', 'dataset_train_filename', 'dataset_val_filename'])
-    trial_id = args.task_id # Note may not start at 0 if HPO has been run for other models with same scheduler
-    directory = args.directory
-    file_prefix = "trial_"+str(trial_id)+"_" # append to all file names created during this trial. Do NOT change!
-    lgb_model = args.lgb_model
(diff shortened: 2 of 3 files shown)


szha:
Job PR-322-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-322/1/index.html

autogluon/utils/tabular/ml/models/abstract/model_trial.py
```diff
@@ -19,27 +20,33 @@ def model_trial(args, reporter):
...
-        np.random.seed(model.params['seed_value'])
-
-        model.params.pop('seed_value')  # TODO: Should we be popping this?
-
-    model.fit(X_train, Y_train, X_val, Y_val)
-    val_perf = model.score(X_val, Y_val)
-    trial_model_file = model.save(file_prefix=file_prefix, directory=directory, return_filename=True)
-    reporter(epoch=0, validation_performance=val_perf, directory=directory, file_prefix=file_prefix, 
-             trial_model_file=trial_model_file)  #  auxiliary_info = ?)
+    try:
```
Innixma: If 9 trials run successfully and then the next fails, will it keep training past the failure on new HPO configs or stop HPO entirely? If it stops HPO, are the 9 successfully trials returned or lost?

autogluon/utils/tabular/ml/models/abstract/model_trial.py
```diff
@@ -8,6 +8,7 @@
 
 logger = logging.getLogger(__name__)  # TODO: Currently unused
 
+
 @args()
 def model_trial(args, reporter):
```
Innixma: LightGBM and NN have their own custom trial functions for HPO that this would also need to be applied to:

utils/tabular/ml/models/lgb/hyperparameters/lgb_trial.py
utils/tabular/ml/models/tabular_nn/tabular_nn_trial.py

Looking at these very old functions, we will probably want to refactor them a bit beyond this HPO fix (though not in this PR) to be less disjointed from the usual .fit function.

autogluon/utils/tabular/ml/models/abstract/model_trial.py
razmikmelikbekyan: Hi @Innixma , 

Good question. So we are lucky, because for each HPO call corresponds a separate Task and also job which is doing this task. Separate job means that it has separate reporters. So it is terminating only single HPO call reporter and the trials which are finished with success are not lost. You can check for example `autogluon.scheduler.fifo.FIFOScheduler.schedule_next`  and corresponding `autogluon.scheduler.fifo.FIFOScheduler.add_job`.

Please see modified code sample, where I put wrong name of `activation` function for NN options. 

```
import autogluon as ag
from autogluon.task import TabularPrediction as task

import sklearn.datasets
import pandas as pd

cancer_data = sklearn.datasets.load_breast_cancer(return_X_y=False)
data = pd.DataFrame(cancer_data['data'], columns=cancer_data['feature_names'])
data['target'] = cancer_data['target']

hp_tune = True
nn_options = {
    'num_epochs': 10,
    'activation': ag.space.Categorical('relu', 'hello_my_friend'),
}

hyperparameters = {
    'NN': nn_options
}

time_limits = 2 * 60
num_trials = 5

output_directory = './agModels--tuningCancerDataset'
data = task.Dataset(data)
label_column = 'target'

# case 1: random search
search_strategy = 'random'
print('\n' * 2)
print('#' * 100)
print(f'Case 1: Running HPO suing search strategy = "{search_strategy}".')
print()
predictor = task.fit(
    train_data=data,
    holdout_frac=0.3,
    label=label_column,
    output_directory=output_directory,
    time_limits=time_limits,
    num_trials=num_trials,
    hyperparameter_tune=hp_tune,
    auto_stack=False,
    stack_ensemble_levels=0
)

print('\n' * 2)
print('#' * 100)
print('HPO results:')
print(predictor.leaderboard(silent=True))
```

As a result I got:
image
So instead of 5 trials it has only 3, so it is not loosing history.

autogluon/utils/tabular/ml/models/abstract/model_trial.py
razmikmelikbekyan: I added for GBM and NN.

Yep, you are right, they are old and a bit ugly :)

razmikmelikbekyan: Meanwhile I would like to ask, why there is no HPO support for RF and KNN and XT ? Do we plan to have them in the feature ?

autogluon/utils/tabular/ml/models/abstract/model_trial.py
Innixma: Great, thanks for the explanation!

autogluon/utils/tabular/ml/models/abstract/model_trial.py
Innixma: It is quite easy to add, but at the time HPO was unstable enough such that I didn't want to encourage such use. If it appears to be stable after this fix, I can quickly add the functionality in.

A more complex point is that some of these models might not benefit from HPO as much as the others, and would be taking valuable training time away from the other models if they were tuned.

In terms of disk space, RF, XT, and KNN take up much more space than the other models, so I didn't want AutoGluon to be taking 100+ GB of space while hyperparameter tuning, since we save every model (KNN can easily take up 10MB+ of space/second during HPO on big datasets) This is probably the biggest concern for me, I don't want people to run out of disk space using AutoGluon. We may have to check remaining disk space and ensure we don't go below some threshold.

autogluon/utils/tabular/ml/models/abstract/model_trial.py
Innixma: Regarding cases where all models fail (bad config), will it keep training (and immediately failing) models until time_limit or max_trials is reached? We may want to add some cutoff in future if the first 5-10 trials all fail.

razmikmelikbekyan: It will try all of them and fail very fast. So it will not wait until time_limit or max_trials, but will try all n_trials. We can think about cut-off, but it really fails fast in case of wrong params.

szha:
Job PR-322-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-322/2/index.html

Innixma:
@zhanghang1989 @mseeger @jwmueller Please take a look when you have time

szha:
Job PR-322-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-322/3/index.html

PR #326, opened 2020-02-22
Tabular: Fixed validation time values for bagged models

Innixma:
*Issue #, if available:*

*Description of changes:*

- Fixed validation time values for bagged models
- Added fit_time and predict_time variables to models
- Note: Validation time and fit time is not fixed for non-bagged HPO models, requires significant code refactor in model.hyperparameter_tune() that is beyond the scope of this PR.
- Note: Validation time reflects time for the model given its inputs, not the full stacked ensemble time including base models. This requires graph representation to change.

### DIFF
--- a/autogluon/task/tabular_prediction/predictor.py
+++ b/autogluon/task/tabular_prediction/predictor.py
@@ -254,6 +254,8 @@ class TabularPredictor(BasePredictor):
             print("*** Summary of fit() ***")
             print("Estimated performance of each model:")
             results['leaderboard'] = self._learner.leaderboard(silent=False)
+            results['leaderboard']['fit_time'] = results['leaderboard']['fit_time'].fillna(0)
+            results['leaderboard']['pred_time_val'] = results['leaderboard']['pred_time_val'].fillna(0)
             # self._summarize('model_performance', 'Validation performance of individual models', results)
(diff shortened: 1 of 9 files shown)


szha:
Job PR-326-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-326/3/index.html

autogluon/task/tabular_prediction/predictor.py
```diff
@@ -254,6 +254,8 @@ def fit_summary(self, verbosity=3):
             print("*** Summary of fit() ***")
             print("Estimated performance of each model:")
             results['leaderboard'] = self._learner.leaderboard(silent=False)
+            results['leaderboard']['fit_time'] = results['leaderboard']['fit_time'].fillna(0)
+            results['leaderboard']['pred_time_val'] = results['leaderboard']['pred_time_val'].fillna(0)
```
jwmueller: since you mentioned you didn't want to contaminate the stored statistics with wrong data, you could consider moving the NA imputation into the plotting function, since that's the only place where it's needed:

https://github.com/awslabs/autogluon/blob/master/autogluon/utils/plots.py#L92-L93

Innixma: Will submit separate PR for moving fillna calls to plotting

PR #333, opened 2020-02-25
Tabular: Fixed LightGBM HPO

Innixma:
*Issue #, if available:*

*Description of changes:*

Major Changes:
- Significantly simplified lgb_trial.py to call model.fit.
- Corrected HPO model path reporting and names
- Enabled time-based early stopping of HPO trials for LightGBM
- Enabled memory-based early stopping of HPO trials for LightGBM
- Significantly simplified LightGBM callback code, merged two existing callbacks into one
- Corrected LightGBM HPO hyperparameters (Previously was throwing away seed and num_boost_rounds, + altering original model hyperparameters incorrectly)
- Optimized early stopping for LightGBM HPO to use dynamic early stopping rounds

Note: These issues still exist for the other models. Because this is a major change, I will leave the code changes for other models to separate PR's.

### DIFF
--- a/autogluon/utils/tabular/ml/models/abstract/abstract_model.py
+++ b/autogluon/utils/tabular/ml/models/abstract/abstract_model.py
@@ -314,7 +314,7 @@ class AbstractModel:
         template.path = template.create_contexts(self.path + template.name + os.path.sep)
         return template
 
-    def hyperparameter_tune(self, X_train, X_test, Y_train, Y_test, scheduler_options=None, **kwargs):
+    def hyperparameter_tune(self, X_train, X_test, Y_train, Y_test, scheduler_options, **kwargs):
         # verbosity = kwargs.get('verbosity', 2)
         start_time = time.time()
         logger.log(15, "Starting generic AbstractModel hyperparameter tuning for %s model..." % self.name)
--- a/autogluon/utils/tabular/ml/models/lgb/callbacks.py
+++ b/autogluon/utils/tabular/ml/models/lgb/callbacks.py
@@ -74,6 +74,10 @@ def early_stopping_custom(stopping_rounds, first_metric_only=False, metrics_to_u
        Whether to use only the first metric for early stopping.
     verbose : bool, optional (default=True)
         Whether to print message with early stopping information.
+    train_loss_name : str, optional (default=None):
+        Name of metric that contains training loss value.
+    reporter : optional (default=None):
+        reporter object from AutoGluon scheduler.
 
     Returns
     -------
--- a/autogluon/utils/tabular/ml/models/lgb/callbacks.py
+++ b/autogluon/utils/tabular/ml/models/lgb/callbacks.py
@@ -6,7 +6,7 @@ from .....try_import import try_import_lightgbm
 
 logger = logging.getLogger(__name__)
 
-# callback
+
 def save_model_callback(path, latest_model_checkpoint, interval, offset):
     def _callback(env):
         if ((env.iteration - offset) % interval == 0) & (env.iteration != 0):
@@ -149,7 +149,7 @@ def early_stopping_custom(stopping_rounds, first_metric_only=False, metrics_to_u
         if not enabled[0]:
             return
         if train_loss_name is not None:
-            train_loss_evals = [eval for i, eval in enumerate(env.evaluation_result_list) if eval[0] == 'train_set' and eval[1] == train_loss_name]
+            train_loss_evals = [eval for eval in env.evaluation_result_list if eval[0] == 'train_set' and eval[1] == train_loss_name]
             train_loss_val = train_loss_evals[0][2]
         else:
             train_loss_val = 0.0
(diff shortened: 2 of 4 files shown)


Innixma:
@razmikmelikbekyan This may be of interest to you, feel free to leave comments.

szha:
Job PR-333-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-333/1/index.html

szha:
Job PR-333-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-333/2/index.html

autogluon/utils/tabular/ml/models/lgb/callbacks.py
```diff
@@ -142,14 +148,27 @@ def _callback(env):
             _init(env)
         if not enabled[0]:
             return
-
+        if train_loss_name is not None:
+            train_loss_evals = [eval for i, eval in enumerate(env.evaluation_result_list) if eval[0] == 'train_set' and eval[1] == train_loss_name]
```
razmikmelikbekyan: just small comment to old code: `enumerate` is not needed.

autogluon/utils/tabular/ml/models/lgb/lgb_model.py
```diff
@@ -277,23 +264,27 @@ def hyperparameter_tune(self, X_train, X_test, Y_train, Y_test, scheduler_option
...
         logger.log(15, "Best hyperparameter configuration for Gradient Boosting Model: ")
         logger.log(15, best_hp)
-        return (hpo_models, hpo_model_performances, hpo_results)
-        # TODO: do final fit here?
-        # args.final_fit = True
-        # final_model = scheduler.run_with_config(best_config)
-        # save(final_model)
+        return hpo_models, hpo_model_performances, hpo_results
+
+    def _get_train_loss_name(self):
```
razmikmelikbekyan: Here I have comment question. As I understand it is for early stopping. Why we are not using the same loss which is used for `fit`  ?

Innixma: This is not for early stopping actually, this is just recorded in reporter for info to the user when they call fit_summary() if I'm not mistaken, although that code was not written by me. The loss metric is often not equivalent to the evaluation metric, since an evaluation metric can often not be directly optimized for. Recording the training evaluation metric would make sense, although we currently don't do it. I personally don't see much benefit in general from observing training metrics, although @jwmueller would know the intent best.

razmikmelikbekyan: Thanks for the explanation, I got the purpose of doing it. But still I have question. Let me reformulate it.  What will happen if for regression I put `eval_metric="r2"` when I run `fit()` ? 
I tried to debug a bit and as far as I see that in this line https://github.com/awslabs/autogluon/blob/b9fb6671c1ab70ddd29e931ef163bee512a86e8c/autogluon/utils/tabular/ml/models/lgb/hyperparameters/lgb_trial.py#L77

the `train_params['params']['objective']` is equal to `regression` which is equivalent to `l2` loss according to LightGBM documentation (https://lightgbm.readthedocs.io/en/latest/Parameters.html#objective). Which means we run `lgb.train()` with loss `l2` rather than `r2`. Is this what we want to do ? So this means that regardless which `eval_metric` I will use for regression problem, we always train the model with `l2` loss. This will work since minimising one loss should minimise the other also, but want to be sure that it is the behaviour which went to have.

jwmueller: Yes this code is just to track the performance of the model at each round of iterative training, so the user can inspect the training runs later.

Re the objective mismatch: not every 3rd party model supports all the objectives that autogluon has, so we sometimes will have to make approximations.  However, we are fortunate in this case: minimizing `l2` is equivalent to maximizing `r2`.  This approximation would be less appropriate if we were doing something like quantile regression (say `MAE` loss to estimate median)  and telling `lgb.train()` to still minimize `l2` loss.  So we'd love to have your help if you notice any metric mismatches between what the user specified to autogluon in `eval_metric` and what we pass to each of the models as their objective.

autogluon/utils/tabular/ml/models/lgb/callbacks.py
Innixma: Thanks! Removed in latest update.

szha:
Job PR-333-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-333/3/index.html

autogluon/utils/tabular/ml/models/abstract/abstract_model.py
```diff
@@ -314,7 +314,7 @@ def convert_to_compressed_template(self):
         template.path = template.create_contexts(self.path + template.name + os.path.sep)
         return template
 
-    def hyperparameter_tune(self, X_train, X_test, Y_train, Y_test, scheduler_options=None, **kwargs):
+    def hyperparameter_tune(self, X_train, X_test, Y_train, Y_test, scheduler_options, **kwargs):
```
jwmueller: should every model's hyperparameter_tune require scheduler_options, when some don't even utilize a scheduler?

Innixma: For the ones that don't they can have it default to None, but for AbstractModel's version it is required:

```
scheduler_func, scheduler_options = scheduler_options  # Unpack tuple
        if scheduler_func is None or scheduler_options is None:
            raise ValueError("scheduler_func and scheduler_options cannot be None for hyperparameter tuning")
```

PR #334, opened 2020-02-26
FIX: random searcher for negative reward

razmikmelikbekyan:
*Issue #, if available:*

*Description of changes:*
- Basic PEP8 corrections and optimisations. 
- Fixing random searcher, when reward is negative value (for example r2).

### DIFF
--- a/autogluon/searcher/searcher.py
+++ b/autogluon/searcher/searcher.py
@@ -1,14 +1,14 @@
-import os
-import pickle
 import logging
-from collections import OrderedDict
 import multiprocessing as mp
+import pickle
+from collections import OrderedDict
 
-from ..utils import load, DeprecationHelper
+from ..utils import DeprecationHelper
 
 __all__ = ['BaseSearcher', 'RandomSearcher', 'RandomSampling']
 
 logger = logging.getLogger(__name__)
+
 
 class BaseSearcher(object):
     """Base Searcher (virtual class to inherit from if you are creating a custom Searcher).
@@ -20,6 +20,7 @@ class BaseSearcher(object):
         specification of the Hyperparameters with their priors
     """
     LOCK = mp.Lock()
+
     def __init__(self, configspace):
         self.configspace = configspace
         self._results = OrderedDict()
@@ -36,7 +37,7 @@ class BaseSearcher(object):
         returns: (config, info_dict)
             must return a valid configuration and a (possibly empty) info dict
         """
-        raise NotImplementedError('This function needs to be overwritten in %s.'%(self.__class__.__name__))
+        raise NotImplementedError(f'This function needs to be overwritten in {self.__class__.__name__}.')
 
     def update(self, config, reward, **kwargs):
         """Update the searcher with the newest metric report
@@ -59,8 +60,7 @@ class BaseSearcher(object):
                 config_pkl = pickle.dumps(config)
                 old_reward = self._results.get(config_pkl, reward)
                 self._results[config_pkl] = max(reward, old_reward)
-            logger.info('Finished Task with config: {} and reward: {}'.format(
-                config, reward))
+            logger.info(f'Finished Task with config: {config} and reward: {reward}')
 
     def register_pending(self, config, milestone=None):
         """
@@ -80,9 +80,8 @@ class BaseSearcher(object):
            Assumes higher reward values indicate better performance.
         """
         with self.LOCK:
-            if len(self._results) > 0:
-                config = max(self._results, key=self._results.get)
-                return self._results[config]
+            if self._results:
+                return max(self._results.values())
         return 0.0
 
     def get_reward(self, config):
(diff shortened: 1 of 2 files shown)


autogluon/searcher/searcher.py
```diff
@@ -155,25 +156,19 @@ def get_config(self, **kwargs):
...
             num_tries = 1
             while pickle.dumps(new_config) in self._results.keys():
                 assert num_tries <= self.MAX_RETRIES, \
-                    "Cannot find new config in BaseSearcher, even after {} trials".format(
-                        self.MAX_RETRIES)
+                    f"Cannot find new config in BaseSearcher, even after {self.MAX_RETRIES} trials"
                 new_config = self.configspace.sample_configuration().get_dictionary()
                 num_tries += 1
-            self._results[pickle.dumps(new_config)] = 0
+            self._results[pickle.dumps(new_config)] = float("-inf")
```
razmikmelikbekyan: Please pay attention here. When we are requesting new config from random searcher, it    already puts the config in the `self._results` dict with value `0`. Whenever reward is coming and it is negative, it will be not updated in results dict due to line 62: `self._results[config_pkl] = max(reward, old_reward)`. So I changed it to infinity to be able to update dict in any case.

Innixma: Could you give an example where this alters behavior from the existing code? Are certain eval metrics defective at present because of this?

Innixma: Is there an example eval metric which causes defective behavior without this change during HPO?

razmikmelikbekyan: yep, for example `r2`, which can be negative.

autogluon/searcher/searcher.py
```diff
@@ -80,9 +80,8 @@ def get_best_reward(self):
            Assumes higher reward values indicate better performance.
         """
         with self.LOCK:
-            if len(self._results) > 0:
-                config = max(self._results, key=self._results.get)
-                return self._results[config]
+            if self._results:
+                return max(self._results.values())
         return 0.0
```
jwmueller: should we also use -inf everywhere else there are zeros?  
It seems we should be consistent.

szha:
Job PR-334-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-334/1/index.html

autogluon/searcher/searcher.py
Innixma: I believe it also fixes logloss because for Tabular we flip the scores such that 0 is the max value and thus you can get scores like -0.25.

autogluon/searcher/searcher.py
razmikmelikbekyan: yep, definitely, thanks for comment.

szha:
Job PR-334-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-334/2/index.html

mseeger:
OK, so I also looked at this code once, and I think it is a bit broken what is done there.

We should simply not enter anything into self._results, as long as an evaluation has not finished. If you refactor this, why not try that?

It is anyway not clear to me who is using self._results. But we need to get clear about it semantics. To me, it would store confirmed results of jobs that ran to the end.

The difficulty here is that we support multi-fidelity methods, where the searcher also works with intermediate results. But then, it should deal with this on its own.

mseeger:
My favourite way of refactoring this would be to remove self._results altogether. If this is not possible, then I'd populate self._results only once a real metric value comes in, not before. This way, you do not have to choose a "default smallest" value at all.

razmikmelikbekyan:
> My favourite way of refactoring this would be to remove self._results altogether. If this is not possible, then I'd populate self._results only once a real metric value comes in, not before. This way, you do not have to choose a "default smallest" value at all.

@mseeger I agree with you, for me entering some default reward for config is a bit strange also, but I did not want to change that logic, since I was wandering that it is not safe enough. So I decided to put `inf` instead, in order to not change the current logic. But if it is possible to remove this line `self._results[pickle.dumps(new_config)] = float("-inf")` at all, it would be even better in my opinion. 

Besides that we have to somewhere mention in documentation, that we are doing `maximisation` task, but I do not now exactly where.

mseeger:
@razmikmelikbekyan 

OK, I have checked, unfortunately self._results is used all over the place.
I feel now that what you do make sense (using -infty as initial value). But could you please do it this way?
-  Define a static method _reward_while_pending, or _minimum_reward, which returns your choice, name -infty
- Grep the code and make sure every piece which is accessing self._results is refactored to use this method. I found searcher.py, skopt_searcher.py, rl_controller.py.
   In these codes, setting the entry to 0.0 is all over the place, this has to change

The problem is that self._results is used for TWO things. Not just does it record results, but it also is used to record which configs have been started, so that a random choice does not pick those again. It is also possible to call get_best_reward before any evaluation has produced a value, and then one would get the value of _reward_while_pending.

razmikmelikbekyan:
> @razmikmelikbekyan
> 
> OK, I have checked, unfortunately self._results is used all over the place.
> I feel now that what you do make sense (using -infty as initial value). But could you please do it this way?
> 
> * Define a static method _reward_while_pending, or _minimum_reward, which returns your choice, name -infty
> * Grep the code and make sure every piece which is accessing self._results is refactored to use this method. I found searcher.py, skopt_searcher.py, rl_controller.py.
>   In these codes, setting the entry to 0.0 is all over the place, this has to change
> 
> The problem is that self._results is used for TWO things. Not just does it record results, but it also is used to record which configs have been started, so that a random choice does not pick those again. It is also possible to call get_best_reward before any evaluation has produced a value, and then one would get the value of _reward_while_pending.

@mseeger 

Please see updated version.

szha:
Job PR-334-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-334/3/index.html

Innixma:
Received offline approval from @zhanghang1989 , merging.

PR #335, opened 2020-02-26
Internal functionality accessible via trainer

jwmueller:
*Description of changes:*

1) Internal functionality for refit_single_full (formally called "compressed"). Fits models using 100% of the data with early stopping point determined based on previous k-fold-bagging run.   

Example usage of refit_single_full:

```
predictor = task.fit(train_data=train_data, label=label_column, output_directory=savedir,
                     enable_fit_continuation=True, auto_stack=True)
learner = predictor._learner
trainer = learner.load_trainer()
best_baggedbase_model = trainer.best_single_model(stack_name='core', stack_level=0)
best_compressed_model = trainer.refit_single_full(models=[best_baggedbase_model])[0]
all_compressed_models = trainer.refit_single_full()
```

Remaining issues in refit_single_full:
- currently only works for bagged models, should be extended to arbitrary models.
- currently, the refit_single_full models are a type of stack-ensemble, but this may be misleading since they are single models.  In general, we want a clear way for users to distinguish between StackEnsemble models that are big bag of many model copies vs those that are just a single model.
- currently is not yet user-facing.  Should add refit_single_full interface in Predictor once the remaining kinks are ironed out.

2) Internal functionality for SOFTCLASS problem_type.  These are multi-class classification problems where the labels are not individual classes, but rather class-probability vectors.  This problem_type is encountered in model-distillation or crowdsourced datasets. 

- currently, neural net is the only type of model supported for this problem_type (RF can also be applied to distillation and this problem_type, although the current RF model class does not yet include any info about this problem_type).

- currently SOFTCLASS problem_type is not yet user-facing.  Need to add label_cleaner,  preset_models, documentation, and clear example script+dataset for this to be truly accessible to users.

3) Neural net model now offers support for learning-rate schedules as an additional  hyperparameter.  Not used by default, so neural net remains the same when user does not specify any hyperparameters.

Note: ensemble-distillation functionality will be added in a separate PR


szha:
Job PR-335-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-335/1/index.html

autogluon/utils/tabular/ml/models/tabular_nn/tabular_nn_model.py
```diff
@@ -433,15 +437,30 @@ def _predict_tabular_data(self, new_data, process=True, predict_proba=True):  #
         if self.problem_type == REGRESSION or not predict_proba:
             return preds.asnumpy().flatten() # return 1D numpy array
         elif self.problem_type == BINARY and predict_proba:
-            return preds[:,1].asnumpy() # for binary problems, only return P(Y==1)
+            preds = preds[:,1].asnumpy() # for binary problems, only return P(Y==+1)
+            if self.stopping_metric == log_loss or self.objective_func == log_loss: # Ensure nonzero predicted probabilities under log-loss:
```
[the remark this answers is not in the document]
Innixma: I'm somewhat concerned by this. I don't think we should be changing predict_proba functionality based off of the eval metric and stopping metric of the model. Perhaps there is a cleaner solution? When is this relevant? If its only during training, perform the correction after predict_proba returns, not inside the function. Then call score_from_pred_proba on the corrected pred_probas to get the score.

autogluon/utils/tabular/ml/models/tabular_nn/tabular_nn_model.py
```diff
@@ -433,15 +437,30 @@ def _predict_tabular_data(self, new_data, process=True, predict_proba=True):  #
...
+                min_pred = np.min(preds)
+                max_pred = np.max(preds)
+                if min_pred < EPS or max_pred > 1-EPS: # remap predicted probs to line that goes through: (min_y, EPS), (max_y, 1-EPS)
+                    preds =  EPS + ((1-2*EPS)/(max_pred-min_pred)) * (preds - min_pred)
+            return preds
+        elif (predict_proba and (self.problem_type == MULTICLASS or self.problem_type == SOFTCLASS) and
```
Innixma: ditto

autogluon/utils/tabular/ml/models/tabular_nn/tabular_nn_model.py
```diff
@@ -490,23 +509,43 @@ def process_train_data(self, df, labels):
...
+    def setup_trainer(self, train_dataset):
+        """ Set up stuff needed for training:
             optimizer, loss, and summary writer (for mxboard).
-            Network must first be initialized before this. 
+            Network must first be initialized before this.
         """
-        optimizer_opts = {'learning_rate': self.params['learning_rate'],  
+        optimizer_opts = {'learning_rate': self.params['learning_rate'],
             'wd': self.params['weight_decay'], 'clip_gradient': self.params['clip_gradient']}
+        if 'lr_scheduler' in self.params and self.params['lr_scheduler'] is not None:
```
Innixma: I'm a bit concerned that these extra hyperparameters feel tacked on such that they don't belong to the PR. I'm ok with approving it with these changes present this time, but in future we should avoid having disjoint functionality being added in a single PR, especially functionality which we have no way of automatically verifying it won't crash upon use, yet users could technically specify in their HPO spaces.

autogluon/utils/tabular/ml/trainer/abstract_trainer.py
```diff
@@ -128,6 +127,7 @@ def __init__(self, path: str, problem_type: str, scheduler_options=None, objecti
         self.num_cols_train = None
 
         self.is_data_saved = False
+        self.normalize_predprobs = False # whether or not probabilistic predictions may need to be renormalized (eg. distillation with BINARY -> REGRESSION)
```
Innixma: Shouldn't this live in the models? Whether we normalize depends on which model we are getting results from, this could be a distilled model or a normal model, or both.

autogluon/utils/tabular/ml/trainer/abstract_trainer.py
```diff
@@ -637,6 +637,22 @@ def predict_proba_model(self, X, model, level_start=0):
         if isinstance(model, str):
             model = self.load_model(model)
         X = self.get_inputs_to_model(model=model, X=X, level_start=level_start, fit=False)
+        EPS = 1e-8 # predicted probabilities can be at most this confident if we normalize predicted probabilities
```
Innixma: ditto on comment regarding normalization

autogluon/utils/tabular/ml/trainer/abstract_trainer.py
```diff
@@ -727,42 +744,39 @@ def compress(self, X=None, y=None, models=None):
...
-            raise NotImplementedError
-        models_distill = get_preset_models_distillation(path=self.path, problem_type=self.problem_type, objective_func=self.objective_func, stopping_metric=self.stopping_metric, num_classes=self.num_classes, hyperparameters=self.hyperparameters)
-        y_distill = pd.Series(model_best.oof_pred_proba)
-
-        # TODO: Do stratified for binary/multiclass, folds are not aligned!
-        models_trained = self.stack_new_level_core(X=X, y=y_distill, models=models_distill, level=0, stack_name='distilled', hyperparameter_tune=False, feature_prune=False)
-        self.compress(X=X, y=y_distill, models=models_trained)
-
-        self.save()
+    def best_single_model(self, stack_name, stack_level):
```
Innixma: Looking at the original intent of this PR (Although it seems the scope has expanded), this is the only function that is necessary for the original intended PR, correct?

autogluon/task/tabular_prediction/tabular_prediction.py
```diff
@@ -268,9 +268,9 @@ def fit(train_data, label, tuning_data=None, output_directory=None, problem_type
             # Currently disabled, needs to be updated to align with new model class functionality
             logger.log(30, 'Warning: feature_prune does not currently work, setting to False.')
 
-        if enable_fit_continuation:
-            enable_fit_continuation = False  # TODO: Add fit_continue function to enable this
-            logger.log(30, 'Warning: enable_fit_continuation does not currently work, setting to False.')
+        # if enable_fit_continuation:
```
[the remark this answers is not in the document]
jwmueller: If this is uncommented, there will be no way to use some of added functionality. 
Why do we need to set `enable_fit_continuation = False`?

To me it seems like the warning is the important part to keep currently, not whether or not this argument is internally set = False or not. I think it should only be uncommented out if `enable_fit_continuation = True` breaks something else I'm not aware of?

autogluon/utils/tabular/ml/models/tabular_nn/tabular_nn_dataset.py
```diff
@@ -114,29 +116,34 @@ def __init__(self, processed_array, feature_arraycol_map, feature_type_map, batc
...
-                    labels = labels.astype('float32') # Convert to proper float-type if not already
-            data_list.append(mx.nd.array(labels.reshape(len(labels),1)))
             self.data_desc.append("label")
-            self.label_index = len(data_list) - 1 # To access data labels, use: self.dataset._data[self.label_index]
-            if problem_type in [BINARY, MULTICLASS]:
-                self.num_classes = len(set(labels))
-        
+            self.label_index = len(data_list) # To access data labels, use: self.dataset._data[self.label_index]
+            self.num_classes = None
+            if self.problem_type == REGRESSION and labels.dtype != np.float32:
```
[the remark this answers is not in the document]
jwmueller: do you mean the `self.num_classes = None` or the `if` statement.  If you mean the latter, wouldn't the readability be most improved if I simply change L139 to: 

`elif self.problem_type == SOFTCLASS:`

autogluon/utils/tabular/ml/models/tabular_nn/tabular_nn_model.py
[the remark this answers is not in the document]
jwmueller: this can be relevant if for example during the early stages of training,

autogluon/utils/tabular/ml/models/tabular_nn/tabular_nn_model.py
jwmueller: This functionality can be useful for inference-constrained / distillation-like settings where the user's goal is just deploy a single neural network model. While the LRscheduler is typically not as robust across many diverse datasets by default, it can be used to improve the performance of the NN on a particular dataset, given substantial user-tweaking/HPO to tailor it to this data.  

Is your concern regarding code coverage in the unit tests? I fully support expanding our  coverage.  I can remove it from this PR if you'd like and add it in a separate PR?  but even if I add testing coverage for this functionality, there are still tons of other autogluon-options not being covered in the unit tests.

autogluon/utils/tabular/ml/trainer/abstract_trainer.py
jwmueller: If you mean getting "compressed" models, then yes this is the only function needed for that.
The rest of the PR contains other internally-accessible functionality that will be useful for the closely related task of distillation and other problem_types such as SOFTCLASS.

autogluon/utils/tabular/metrics/softclass_metrics.py
```diff
@@ -0,0 +1,29 @@
...
+
+import logging
+import numpy as np
+import mxnet as mx
+
+logger = logging.getLogger(__name__)
+
+
+EPS = 1e-9 # clipping threshold to prevent NaN
+softloss = mx.gluon.loss.SoftmaxCrossEntropyLoss(sparse_label=False, from_logits=True) # assumes predictions are already log-probabilities.
```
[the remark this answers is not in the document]
jwmueller: Yes I'll just define the mxnet training loss in the neural network file.

autogluon/utils/tabular/ml/trainer/abstract_trainer.py
jwmueller: This is like a final safety check that one may sometimes wish to apply (eg. when changing problem_types as in distillation).  I agree all individual models should be outputting normalized probabilities in the first place.  But example in distillation, the individual models currently think they are doing a regression task, and thus it might be weird to ask them to normalize in this case.  I think have this (optional) safety check for the final overall prediction may be useful in future applications as well.

autogluon/utils/tabular/ml/trainer/abstract_trainer.py
jwmueller: I will decrease the `EPS` factors so you can be assured they won't have any detrimental impact on performance ever.

autogluon/task/tabular_prediction/tabular_prediction.py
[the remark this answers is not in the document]
Innixma: Nothing breaks, so keeping the warning with it just mentioning that it persists the data to disk without enabling fit continuation at present will work.

autogluon/utils/tabular/ml/models/tabular_nn/tabular_nn_dataset.py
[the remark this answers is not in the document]
Innixma: if self.problem_type == REGRESSION and labels.dtype != np.float32: -> Line 143, since it will only be able to be triggered if the else statement would have triggered. I mean the if statement.

autogluon/utils/tabular/ml/models/tabular_nn/tabular_nn_model.py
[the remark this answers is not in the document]
Innixma: Ok, my concern is this:

predict_proba functionality now depends on stopping_metric and eval_metric variables.

No other model does this. This also means the stopping_metric a user chooses to help early stop the model will impact the model predict_proba results even after the training stage, despite stopping_metric being unused outside of .fit by every other model. At the very least, stopping_metric should not impact the predict_proba result outside of .fit() calls to .predict_proba().

My concern is not that EPS is too large, its that it is dependent on the eval metric/stopping metric

autogluon/utils/tabular/ml/models/tabular_nn/tabular_nn_model.py
Innixma: It is fine for now. Primarily I just think in future we should look towards more focused PR's. I believe these HPO settings should be their own PR.

autogluon/utils/tabular/ml/trainer/abstract_trainer.py
Innixma: Ok, although can you add a TODO above this that mentions long-term we should look to have each model do its own normalization, and Trainer can specify to normalize each model by checking the model's problem_type upon loading and calling predict_proba()

autogluon/utils/tabular/ml/trainer/abstract_trainer.py
Innixma: ditto on follow-up comment, add TODO to move normalization internal to models with Trainer dictating which models are normalized based on model problem_type

autogluon/utils/tabular/ml/models/tabular_nn/tabular_nn_model.py
jwmueller: Added partial coverage in unit test: test_tabularHPO_bagstack.py
It's hard to fully cover all hyperparameter combinations within reasonable time-limit in unit tests.

szha:
Job PR-335-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-335/2/index.html

szha:
Job PR-335-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-335/3/index.html

szha:
Job PR-335-4 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-335/4/index.html

PR #336, opened 2020-02-26
Tabular AbstractModel HPO Improvements

Innixma:
*Issue #, if available:*

*Description of changes:*

Major Changes:
- Simplified AbstractModel HPO
- Added support for early stopping with time for AbstractModel HPO
- Unified various HPO code under a single method
- Fixed minor defects in AbstractModel HPO
- Added fit_time and predict_time recording to HPO models (Not utilized yet)

This PR is part 2 of HPO Cleanup/Improvements.
A future PR will address tabular_nn HPO and remaining defects.

### DIFF
--- a/autogluon/utils/tabular/ml/models/abstract/abstract_model.py
+++ b/autogluon/utils/tabular/ml/models/abstract/abstract_model.py
@@ -92,6 +92,7 @@ class AbstractModel:
 
         self.fit_time = None  # Time taken to fit in seconds (Training data)
         self.predict_time = None  # Time taken to predict in seconds (Validation data)
+        self.val_score = None  # Score with eval_metric (Validation data)
 
         self.params = {}
         self._set_default_params()
--- a/autogluon/utils/tabular/ml/models/abstract/abstract_model.py
+++ b/autogluon/utils/tabular/ml/models/abstract/abstract_model.py
@@ -1,7 +1,6 @@
 import copy
 import logging
 import os
-import pickle
 import time
 
 import numpy as np
@@ -16,6 +15,7 @@ from ....utils.loaders import load_pkl
 from ....utils.savers import save_pkl
 from ......core import Space, Categorical, List, NestedSpace
 from ......task.base import BasePredictor
+from ......scheduler.scheduler import TaskScheduler
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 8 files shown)


szha:
Job PR-336-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-336/1/index.html

szha:
Job PR-336-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-336/2/index.html

szha:
Job PR-336-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-336/3/index.html

autogluon/utils/tabular/ml/models/abstract/abstract_model.py
```diff
@@ -389,23 +392,16 @@ def hyperparameter_tune(self, X_train, X_test, Y_train, Y_test, scheduler_option
...
-            trial_model_name = f"{self.name}_{file_id}"
-            trial_model_path = f"{self.path}{file_prefix}"
+            file_id = "trial_" + str(trial)  # unique identifier to files from this trial
+            trial_model_name = self.name + os.path.sep + file_id
+            trial_model_path = self.path_root + trial_model_name + os.path.sep
             hpo_models[trial_model_name] = trial_model_path
             hpo_model_performances[trial_model_name] = hpo_results['trial_info'][trial][scheduler._reward_attr]
 
         logger.log(15, "Time for %s model HPO: %s" % (self.name, str(hpo_results['total_time'])))
-        self.params.update(best_hp)
```
jwmueller: Don't we want to update this model's params with the best_hp?  For example, if a user just calls model.hyperparameter_tune() themselves (outside of task.fit()), they presumably expect the model now to have its optimal hyperparameter values.

autogluon/utils/tabular/ml/models/abstract/model_trial.py
```diff
@@ -21,32 +22,42 @@ def model_trial(args, reporter):
...
+        model.fit_time = time_fit_end - time_fit_start
+        model.predict_time = time_pred_end - time_fit_end
+        model.val_score = val_perf
+        model.save()
 
     except Exception as e:
-        logger.exception(e, exc_info=True)
+        if not isinstance(e, TimeLimitExceeded):
+            logger.exception(e, exc_info=True)
         reporter.terminate()
```
jwmueller: good to have here for now, but this kind of thing should really be handled in the scheduler class eventually.

autogluon/utils/tabular/ml/models/lgb/hyperparameters/lgb_trial.py
```diff
@@ -2,47 +2,62 @@
...
+        model.val_score = score
         model.save()
-        # TODO: add to reporter: time_of_trial without load/save time (isn't this just function of early-stopping point?), memory/inference ??
-        # TODO: add to reporter: Name/path of model
     except Exception as e:
         if not isinstance(e, TimeLimitExceeded):
             logger.exception(e, exc_info=True)
         reporter.terminate()
+
+    # FIXME: If stopping metric and eval metric differ, the previous reported scores will not align as they will be evaluated with stopping_metric, whereas this is evaluated with eval_metric
```
jwmueller: Will this also affect the search if it is adaptive?  Ie. when searcher selects next configuration based on the previous configuration's reported score.

If so, we can probably just manually overwrite the final reporter-score with a final `model.score()` call whenever `eval_metric != stopping_metric`.

Innixma: Yes it will, but manually overwriting might not be the best way to do things. If you stick with the same metric across the entire HPO it won't matter (but it will optimize for stopping metric instead of eval metric, which is kind of the intent anyways)

autogluon/utils/tabular/ml/models/abstract/abstract_model.py
Innixma: We don't, the model which hyperparameter_tune is called for is never used and shouldn't be treated as a real model. The return value of HPO contains the real models.

PR #337, opened 2020-02-27
Updated contribution instructions

jwmueller:
Provide more detailed tips to new contributors as suggested by @mseeger 
I think the more instructions & information we can provide the better.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,7 +21,7 @@ reported the issue. Please try to include as much information as you can. Detail
 
 
 ## Contributing via Pull Requests
-Contributions via pull requests are much appreciated. Before sending us a pull request, please ensure that:
+Code contributions via pull requests are much appreciated. Before sending us a pull request, please ensure that:
 
 1. You are working against the latest source on the *master* branch.
 2. You check existing open, and recently merged, pull requests to make sure someone else hasn't addressed the problem already.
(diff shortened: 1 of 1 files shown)


szha:
Job PR-337-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-337/1/index.html

mseeger:
OK, could you be explicit what to do in order to install the AG dependencies (other than MXNet, and apart from non-Python things)?

I suppose this just needs:
   python3 setup.py develop

But in case this does not work out of the box, could you provide hints how to still get it done?

jwmueller:
> OK, could you be explicit what to do in order to install the AG dependencies (other than MXNet, and apart from non-Python things)?


There is nothing that needs to be done beyond the installation instructions on our website: https://autogluon.mxnet.io/install.html
  
If these are producing issues for you and other folks, they need to be improved.

Or are you suggesting to relist the installation instructions from our website in CONTRIBUTING.md?  I find this could be redundant + information overload, given there are more multiple types of install depending on OS, gpu/cpu etc.  A contributor only needs to make sure to install autogluon from source rather than pip, as I've stated on L46 of CONTRIBUTING.md.  Happy to incorporate any suggestions you have to make this more clear.

Or do you mean we should mention all of the optional packages needed for auxiliary autogluon functionality like `bokeh`, `tensorboard`, etc?  I agree we need to gather a list of these and include them somewhere (for both users & contributors), will sync with the team to identify the complete list of optional packages.  This list should probably live in `setup.py` though with instructions on how to install the optional packages listed on our website.

AFAIK the dependencies to build the docs / CI are the only contributor-specific packages needed that a user would have no use for.  However, I don't think we can expect contributors to run these things themselves.  They should just rely on the CI system that will automatically run when they open their PR (we can rethink this is we encounter scaling issues down the line).

PR #338, opened 2020-02-28
Tabular: Refactored tabular_nn_model and tabular_nn_trial

Innixma:
*Issue #, if available:*

#266 

*Description of changes:*

Major Changes:
- Reworked TabularNN HPO to be consistent with other models, fixed various defects in TabularNN HPO.
- Reworked TabularNN to minimize unnecessary variables and avoid alteration of original self.params during .fit(). Simplified code structure.

### DIFF
--- a/autogluon/utils/tabular/ml/models/tabular_nn/tabular_nn_model.py
+++ b/autogluon/utils/tabular/ml/models/tabular_nn/tabular_nn_model.py
@@ -270,8 +270,10 @@ class TabularNeuralNetModel(AbstractModel):
             else:
                 params['loss_function'] = gluon.loss.SoftmaxCrossEntropyLoss(from_logits=self.model.from_logits)
 
+        loss_func = params['loss_function']
+        epochs_wo_improve = params['epochs_wo_improve']
         loss_scaling_factor = 1.0  # we divide loss by this quantity to stabilize gradients
-        loss_torescale = [key for key in self.rescale_losses if isinstance(params['loss_function'], key)]
+        loss_torescale = [key for key in self.rescale_losses if isinstance(loss_func, key)]
         if len(loss_torescale) > 0:
             loss_torescale = loss_torescale[0]
             if self.rescale_losses[loss_torescale] == 'std':
@@ -279,7 +281,7 @@ class TabularNeuralNetModel(AbstractModel):
             elif self.rescale_losses[loss_torescale] == 'var':
                 loss_scaling_factor = np.var(train_dataset.get_labels())/5.0 + EPS  # variance of labels
             else:
-                raise ValueError("Unknown loss-rescaling type %s specified for loss_func==%s" % (self.rescale_losses[loss_torescale], params['loss_function']))
+                raise ValueError("Unknown loss-rescaling type %s specified for loss_func==%s" % (self.rescale_losses[loss_torescale], loss_func))
 
         if self.verbosity <= 1:
             verbose_eval = -1  # Print losses every verbose epochs, Never if -1
@@ -298,7 +300,7 @@ class TabularNeuralNetModel(AbstractModel):
                 with autograd.record():
                     output = self.model(data_batch)
                     labels = data_batch['label']
-                    loss = params['loss_function'](output, labels) / loss_scaling_factor
+                    loss = loss_func(output, labels) / loss_scaling_factor
                     # print(str(nd.mean(loss).asscalar()), end="\r")  # prints per-batch losses
                 loss.backward()
                 self.optimizer.step(labels.shape[0])
@@ -319,7 +321,7 @@ class TabularNeuralNetModel(AbstractModel):
                 with autograd.record():
                     output = self.model(data_batch)
                     labels = data_batch['label']
-                    loss = params['loss_function'](output, labels) / loss_scaling_factor
+                    loss = loss_func(output, labels) / loss_scaling_factor
                     # print(str(nd.mean(loss).asscalar()), end="\r")  # prints per-batch losses
                 loss.backward()
                 self.optimizer.step(labels.shape[0])
@@ -349,9 +351,10 @@ class TabularNeuralNetModel(AbstractModel):
             if self.summary_writer is not None:
                 self.summary_writer.add_scalar(tag='train_loss', value=train_loss.asscalar(), global_step=e)  # TODO: do we want to keep mxboard support?
             if reporter is not None:
+                # TODO: Ensure reporter/scheduler properly handle None/nan values after refactor
                 if test_dataset is not None and (not np.isnan(val_metric)):  # TODO: This might work without the if statement
                     reporter(epoch=e, validation_performance=val_metric, train_loss=float(train_loss.asscalar()))  # Higher val_metric = better
-            if e - best_val_epoch > params['epochs_wo_improve']:
+            if e - best_val_epoch > epochs_wo_improve:
                 break
             if time_limit:
                 time_elapsed = time.time() - start_time
@@ -531,12 +534,10 @@ class TabularNeuralNetModel(AbstractModel):
                                 problem_type=self.problem_type, labels=labels, is_test=False)
 
     def setup_trainer(self, params, train_dataset=None):
-        """ Set up stuff needed for training:
-            optimizer, loss, and summary writer (for mxboard).
+        """ Set up optimizer needed for training.
             Network must first be initialized before this.
         """
-        optimizer_opts = {'learning_rate': params['learning_rate'],
-            'wd': params['weight_decay'], 'clip_gradient': params['clip_gradient']}
+        optimizer_opts = {'learning_rate': params['learning_rate'], 'wd': params['weight_decay'], 'clip_gradient': params['clip_gradient']}
         if 'lr_scheduler' in params and params['lr_scheduler'] is not None:
             if train_dataset is None:
                 raise ValueError("train_dataset cannot be None when lr_scheduler is specified.")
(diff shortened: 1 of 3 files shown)


szha:
Job PR-338-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-338/1/index.html

autogluon/utils/tabular/ml/models/tabular_nn/tabular_nn_model.py
```diff
@@ -253,31 +247,39 @@ def train_net(self, train_dataset, test_dataset=None,
...
             y_test = None
 
+        if params['loss_function'] is None:
+            if self.problem_type == REGRESSION:
+                params['loss_function'] = gluon.loss.L1Loss()
+            elif self.problem_type == SOFTCLASS:
+                params['loss_function'] = gluon.loss.SoftmaxCrossEntropyLoss(sparse_label=False, from_logits=self.model.from_logits)
+            else:
+                params['loss_function'] = gluon.loss.SoftmaxCrossEntropyLoss(from_logits=self.model.from_logits)
+
```
jwmueller: Is there an advantage to moving this outside of `setup_trainer`? I guess to not have to store loss_func as instance variable? 
The original philosophy was to separate `train_net` from `setup_trainer` so that `setup_trainer` could be called just once, and `train_net` could be called repeatedly to resume training on the same network (without having to setup_trainer again).  Of course moving this quick loss-function setup has little practical effect.  It is also unclear what should be the desired behavior if the user has changed `params` in between two `train_net` calls...

autogluon/utils/tabular/ml/models/tabular_nn/tabular_nn_model.py
```diff
@@ -345,7 +348,10 @@ def train_net(self, train_dataset, test_dataset=None,
                     logger.log(15, "Epoch %s.  Train loss: %s" % (e, train_loss.asscalar()))
             if self.summary_writer is not None:
                 self.summary_writer.add_scalar(tag='train_loss', value=train_loss.asscalar(), global_step=e)  # TODO: do we want to keep mxboard support?
-            if e - best_val_epoch > self.params['epochs_wo_improve']:
+            if reporter is not None:
+                if test_dataset is not None and (not np.isnan(val_metric)):  # TODO: This might work without the if statement
```
jwmueller: The `np.nan()` check is just here because the current scheduler/reporter will break if it records `nan`?
It's unclear to me what would be the best behavior in this case, but I think ignoring `nan` is a good idea for now.  However, we probably want to add a TODO here in case we later add a unified way to handle `nan` in the reporters.

autogluon/utils/tabular/ml/models/tabular_nn/tabular_nn_model.py
```diff
@@ -725,11 +727,13 @@ def save(self, file_prefix="", directory = None, return_name=False, verbose=None
             self.model.save_parameters(params_filepath)
         temp_model = self.model
         temp_sw = self.summary_writer
+        self._architecture_desc = self.model.architecture_desc
         self.model = None
         self.summary_writer = None
         save_pkl.save(path=modelobj_filepath, object=self, verbose=verbose)
         self.model = temp_model
         self.summary_writer = temp_sw
+        self._architecture_desc = None
```
jwmueller: This is to save space?  
Are we sure this won't break repeated saving+reloading of the same model multiple times?

autogluon/utils/tabular/ml/models/tabular_nn/tabular_nn_model.py
```diff
@@ -739,115 +743,72 @@ def load(cls, path, file_prefix="", reset_paths=False, verbose=True):
             If you want to load files with given prefix, can also pass arg: path = directory+file_prefix
         """
         path = path + file_prefix
-        obj = load_pkl.load(path = path + cls.model_file_name, verbose=verbose)
+        obj: TabularNeuralNetModel = load_pkl.load(path=path + cls.model_file_name, verbose=verbose)
         if reset_paths:
             obj.set_contexts(path)
-        obj.model = EmbedNet(architecture_desc=obj.architecture_desc, ctx=obj.ctx) # recreate network from architecture description
+        obj.model = EmbedNet(architecture_desc=obj._architecture_desc, ctx=obj.ctx)  # recreate network from architecture description
+        obj._architecture_desc = None
```
jwmueller: This is to save space?
Are we sure this won't break repeated saving+reloading of the same model multiple times?

autogluon/utils/tabular/ml/models/tabular_nn/tabular_nn_model.py
Innixma: Mostly it is to avoid hiding unnecessary param setting outside of the main code. loss_function is now fully defined and used in just a single function, rather than defined in one and used in another. loss_function has no dependence on setup_trainer, as I have changed it to simply return optimizer without setting any class variables.

Anything that takes virtually no time to do (such as creating the loss function) we should prioritize clarity and stability rather than speed.

I'd argue the 'unclear what would be desired behavior...' was also present before. This PR isn't aimed at answering that question.

autogluon/utils/tabular/ml/models/tabular_nn/tabular_nn_model.py
```diff
@@ -288,20 +290,21 @@ def train_net(self, train_dataset, test_dataset=None,
...
+        net_filename = self.path + self.temp_file_name
         if num_epochs == 0:  # use dummy training loop that stops immediately (useful for using NN just for data preprocessing / debugging)
             logger.log(20, "Not training Neural Net since num_epochs == 0.  Neural network architecture is:")
             for batch_idx, data_batch in enumerate(train_dataset.dataloader):
                 data_batch = train_dataset.format_batch_data(data_batch, self.ctx)
                 with autograd.record():
                     output = self.model(data_batch)
                     labels = data_batch['label']
-                    loss = self.loss_func(output, labels) / loss_scaling_factor
+                    loss = params['loss_function'](output, labels) / loss_scaling_factor
```
jwmueller: would be nicer to define:
```
loss_func = params['loss_function']
```
outside of the iterative loop over epochs
and then use:
```
loss = loss_func(output, labels) / loss_scaling_factor
```

autogluon/utils/tabular/ml/models/tabular_nn/tabular_nn_model.py
```diff
@@ -316,7 +319,7 @@ def train_net(self, train_dataset, test_dataset=None,
...
```
jwmueller: same comment as above

autogluon/utils/tabular/ml/models/tabular_nn/tabular_nn_model.py
Innixma: This won't break anything, it is primarily to avoid an unnecessary variable. It does not save disk space. It saves a very minor amount of memory space.

This works by:

before save ->
self._architecture_desc = None

on save ->
set self._architecture_desc

on load ->
self._architecture_desc = None

Since self._architecture_desc val lives in .model, it can be retrieved from .model after load. self._architecture_desc is never used outside of save/load to init model

autogluon/utils/tabular/ml/models/tabular_nn/tabular_nn_model.py
Innixma: ditto on previous comment

autogluon/utils/tabular/ml/models/tabular_nn/tabular_nn_model.py
Innixma: Uncertain if np.nan/None breaks it, but until reporter/scheduler is refactored I don't want to risk it.

autogluon/utils/tabular/ml/models/tabular_nn/tabular_nn_model.py
Innixma: Agreed, will update

autogluon/utils/tabular/ml/models/tabular_nn/tabular_nn_model.py
Innixma: Agreed, will update

autogluon/utils/tabular/ml/models/tabular_nn/tabular_nn_model.py
```diff
@@ -514,55 +527,46 @@ def process_train_data(self, df, labels):
         self.feature_type_map = self._get_feature_type_map() # OrderedDict of feature-name -> feature_type string (options: 'vector', 'embed', 'language')
         # print(self.feature_type_map)
         return TabularNNDataset(df, self.feature_arraycol_map, self.feature_type_map,
-                                batch_size=self.params['batch_size'], num_dataloading_workers=self.params['num_dataloading_workers'],
+                                batch_size=batch_size, num_dataloading_workers=num_dataloading_workers,
                                 problem_type=self.problem_type, labels=labels, is_test=False)
 
-    def setup_trainer(self, train_dataset=None):
+    def setup_trainer(self, params, train_dataset=None):
```
jwmueller: should remove "loss" from docstring of this function now to avoid confusion

autogluon/utils/tabular/ml/models/tabular_nn/tabular_nn_model.py
```diff
@@ -739,115 +743,72 @@ def load(cls, path, file_prefix="", reset_paths=False, verbose=True):
...
+        test_dataset.save(file_prefix=test_path)
+
         if not np.any([isinstance(params_copy[hyperparam], Space) for hyperparam in params_copy]):
             logger.warning("Warning: Attempting to do hyperparameter optimization without any search space (all hyperparameters are already fixed values)")
         else:
             logger.log(15, "Hyperparameter search space for Neural Network: ")
             for hyperparam in params_copy:
                 if isinstance(params_copy[hyperparam], Space):
                     logger.log(15, str(hyperparam)+ ":   "+str(params_copy[hyperparam]))
-        directory = self.path # path to directory where all remote workers store things
```
jwmueller: This will be needed for distributed setting, so at least keep a TODO here.  According to Hang, the working directory on remote may be different when `upload_files()` is invoked.  The current distributed usage of autogluon (eg. for image classification) does not utilize dask's `upload_files()`, but instead requires user to manually setup all worker machines with identical file-structure + dataset, but this means preprocessing will be redundantly executed on all of them.

autogluon/utils/tabular/ml/models/tabular_nn/tabular_nn_model.py
Innixma: Agreed, will update

autogluon/utils/tabular/ml/models/tabular_nn/tabular_nn_model.py
Innixma: Added todo

autogluon/utils/tabular/ml/models/tabular_nn/tabular_nn_model.py
Innixma: Added TODO

szha:
Job PR-338-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-338/2/index.html

PR #340, opened 2020-02-29
Tabular: Added get_info() to models, trainer, and learner

Innixma:
*Issue #, if available:*

*Description of changes:*

Added get_info() to models, trainer, and learner

### DIFF
--- a/autogluon/utils/tabular/ml/models/ensemble/greedy_weighted_ensemble_model.py
+++ b/autogluon/utils/tabular/ml/models/ensemble/greedy_weighted_ensemble_model.py
@@ -1,4 +1,5 @@
 import logging
+
 from ..abstract.abstract_model import AbstractModel
 from ...constants import MULTICLASS
 from ...tuning.ensemble_selection import EnsembleSelection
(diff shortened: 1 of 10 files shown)


szha:
Job PR-340-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-340/1/index.html

autogluon/utils/tabular/ml/models/ensemble/greedy_weighted_ensemble_model.py
```diff
@@ -67,3 +68,13 @@ def set_stack_columns(self, base_model_names):
...
         return stack_columns, num_pred_cols_per_model
+
+    def _get_model_weights(self):
+        num_models = len(self.base_model_names)
+        model_weight_dict = {self.base_model_names[i]: self.model.weights_[i] for i in range(num_models)}
+        return model_weight_dict
+
+    def get_info(self):
+        info = super().get_info()
+        info['model_weights'] = self._get_model_weights()
```
jwmueller: Should `_get_model_weights()` be a user-facing method instead?
Or is your recommendation the user instead just calls: `get_info()['model_weights']`?  

I still think it would be nice if there is similar functionality inside WeightedEnsembleModel, ie. something like:
```
def get_model_weights():
   return {child.name: child._get_model_weights() for child in self.models}
```

Otherwise the method for users to get model-weights out of  `WeightedEnsemble.get_info()` may be somewhat hard to remember (would it be: `WeightedEnsemble.get_info()["children_info"][CHILD_NAME]["model_weights"]?)

szha:
Job PR-340-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-340/2/index.html

autogluon/utils/tabular/ml/models/ensemble/greedy_weighted_ensemble_model.py
Innixma: We can discuss the specifics of this when we want to present weights to users. The tricky part is if the 'best_model' is not a weighted model, in which there are no weights (unless we say weights={best_model: 1.0}

Issue #329, opened 2020-02-24
Scheduler/searcher runs infinitely long (fails to handle error in individual training trials)


jwmueller:
HPO may currently run infinitely (due to deadlock) when any individual training trial has an error.

May be root cause of issues like this:  https://github.com/awslabs/autogluon/issues/245

A temporary fix is implemented for TabularPrediction in: https://github.com/awslabs/autogluon/pull/322/ 

But a more long-term fix should live in scheduler/searcher.

zhanghang1989:
Thanks @jwmueller 

As we discussed, this need to be implemented in the scheduler to handle the reporter when exception happens in the user/pipeline code.

We can prioritize it when we refactor searchers & schedulers with new algorithms added.

mseeger:
Hello,

this is something I am interested in. We will add new searchers that come with a certain complexity, and they may well fail with throwing an exception (even though this happens very rarely for us).

We'd love that in this case, the whole experiment is shut down.

In the longer term, we should consider alternatives. Maybe the very simplest is to catch exceptions and return some "invalid evaluation" code. The searcher/scheduler would then have the chance to react to this, for example they could just ignore the evaluation call (as there is quite some randomness involved, this may well help to recover).

We also have HPO variants which can deal with failures systematically. This is an exciting and useful direction.

Innixma:
Why not have a parameter to Searcher which dictates behavior upon receiving an exception: Shutdown, ignore+continue, crash, etc.?

Innixma:
Additionally, I'd like to comment that skopt Searcher randomly deadlocks even after #322 has been applied. In 3 runs, 2 of them deadlocked at different points and 1 succeeded. It's unknown why it does this.

jwmueller:
It seems the base RandomSearcher was updated with some locking functionality that was never added to the skopt searcher: https://github.com/awslabs/autogluon/blob/master/autogluon/searcher/searcher.py#L162

@zhanghang1989 Can you add this same functionality to the skopt searcher?
In general, we really need to ensure that whenever any of the searchers/schedulers is altered, the same changes are made to all the other searchers/schedulers where appropriate.

mseeger:
I may have put in the additional locking. While I lack the understanding of Hang, I'd rather err on the cautious side. Louis once told me that both read and write access to class members should be protected. If members are objects with methods, then maybe they should have their own locks. At least, this seems to me how the code works.

mseeger:
[...]

PR #345, opened 2020-03-01
FIX: fixing training hanging

razmikmelikbekyan:
*Issue #329 , if available:*

*Description of changes:*
- Implementation of `CustomProcess` class, which has property `exception`.
- Using `CustomProcess` class for individual training process.
- Checking individual training fails due to some Exception using `exception` property of `CustomProcess`.
- Before fetching from local reporter queue checking if it has some results, if not continue waiting for results or exception in `while self.process.is_alive()`. 
- If process is not fails,  then reported results are transferred to `dist_reporter` and so on.
- if process fails, terminating `local_reporter` and transferring exception traceback to `dist_reporter`.

The change is implemented only for `FIFO` scheduler, it will be approved, then we can implement also for other schedulers. 

Code sample # 1:
```
import autogluon as ag
import pandas as pd
import sklearn.datasets
from sklearn.ensemble import RandomForestRegressor
from sklearn.metrics import r2_score, mean_absolute_error

initial_data = sklearn.datasets.load_boston(return_X_y=False)
data = pd.DataFrame(initial_data['data'], columns=initial_data['feature_names'])
data['target'] = initial_data['target']


def train_rf(args, reporter):
    train_data = args.train_data
    target_variable = args.target_variable

    rf = RandomForestRegressor(
        n_estimators=args.n_estimators,
        max_depth=args.max_depth,
        min_samples_leaf=args.min_samples_leaf,
        min_samples_split=args.min_samples_split,
        min_weight_fraction_leaf=args.min_weight_fraction_leaf,
        random_state=34
    )

    rf.fit(train_data.drop(columns=target_variable), train_data[target_variable])
    pred = rf.predict(train_data.drop(columns=target_variable))
    mae = abs(mean_absolute_error(train_data[target_variable], pred))
    r2 = r2_score(train_data[target_variable], pred)
    reporter(mae=mae, r2=r2)


@ag.args(
    n_estimators=-10,  # this will raise an exception, since it must be >= 1
    max_depth=ag.space.Int(1, 10),
    min_samples_leaf=ag.space.Int(1, 5),
    min_samples_split=ag.space.Int(2, 5),
    min_weight_fraction_leaf=ag.space.Real(0.0, 1e-1),
    train_data=data,
    target_variable="target"
)
def ag_train_rf(args, reporter):
    return train_rf(args, reporter)


myscheduler = ag.scheduler.FIFOScheduler(
    ag_train_rf,
    resource={'num_cpus': 8, 'num_gpus': 0},
    num_trials=5,
    reward_attr='r2',
    searcher='skopt',  # currently supports "random", "skopt", "gird",
    max_reward=0,
)
print(myscheduler)
myscheduler.run()
myscheduler.join_jobs()
```

Code sample # 2:
```
import autogluon as ag
from autogluon import ImageClassification as task

train_data = task.Dataset(name='fashionmnist')

classifier = task.fit(
    train_data,
    num_trials=1,
    net=ag.Categorical('hello'),  # this will raise an exception
    epochs=5,
    ngpus_per_trial=1,
    verbose=False,
    auto_search=False,
)
```

### DIFF
--- a/autogluon/scheduler/fifo.py
+++ b/autogluon/scheduler/fifo.py
@@ -257,6 +257,10 @@ class FIFOScheduler(TaskScheduler):
         last_result = None
         while not task_job.done():
             reported_result = reporter.fetch()
+            if 'traceback' in reported_result:
+                logger.exception(reported_result['traceback'])
+                break
+
             if reported_result.get('done', False):
                 reporter.move_on()
                 break
--- a/autogluon/scheduler/reporter.py
+++ b/autogluon/scheduler/reporter.py
@@ -174,15 +174,29 @@ class Communicator(threading.Thread):
 
     def run(self):
         while self.process.is_alive():
+
+            # breaking communication if process raises exception
+            if self.process.exception is not None:
+                error, traceback = self.process.exception
+                self.local_reporter.terminate()
+                self.dist_reporter(done=True, traceback=traceback)
+                self.process.join()
+                break
+
             try:
+                # waiting until process reports results or raises exception
+                if self.local_reporter._queue.empty():
+                    continue
                 reported_result = self.local_reporter.fetch()
             except BrokenPipeError:
                 break
+
             try:
                 self.dist_reporter(**reported_result)
                 self.local_reporter.move_on()
             except AutoGluonEarlyStop:
                 self.local_reporter.terminate()
+
             if 'done' in reported_result and reported_result['done'] is True:
                 self.process.join()
                 break
--- a/autogluon/scheduler/scheduler.py
+++ b/autogluon/scheduler/scheduler.py
@@ -14,7 +14,7 @@ from .remote import RemoteManager
 from .resource import DistributedResourceManager
 from ..core import Task
 from .reporter import *
-from ..utils import AutoGluonWarning, AutoGluonEarlyStop
+from ..utils import AutoGluonWarning, AutoGluonEarlyStop, CustomProcess
 
 logger = logging.getLogger(__name__)
 
@@ -142,7 +142,7 @@ class TaskScheduler(object):
 
         try:
             # start local progress
-            p = mp.Process(target=_worker, args=(return_list,gpu_ids, args))
+            p = CustomProcess(target=_worker, args=(return_list, gpu_ids, args))
             p.start()
             if 'reporter' in args:
                 cp = Communicator.Create(p, local_reporter, dist_reporter)
--- a/autogluon/utils/__init__.py
+++ b/autogluon/utils/__init__.py
@@ -16,4 +16,4 @@ from .file_helper import *
 from .plot_network import plot_network
 from .defaultdict import keydefaultdict
 from .util_decorator import classproperty
-
+from .custom_process import CustomProcess
--- a/autogluon/utils/custom_process.py
+++ b/autogluon/utils/custom_process.py
@@ -1,0 +1,27 @@
+# Adapted from: https://stackoverflow.com/a/33599967/8199034
+import multiprocessing as mp
+import traceback
+
+
+class CustomProcess(mp.Process):
+    """Custom Process implementation which checks has Exception property."""
+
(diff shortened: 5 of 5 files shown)


szha:
Job PR-345-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-345/1/index.html

autogluon/scheduler/reporter.py
```diff
@@ -174,15 +174,29 @@ def __init__(self, process, local_reporter, dist_reporter):
 
     def run(self):
         while self.process.is_alive():
+
+            # breaking communication if process raises exception
+            if self.process.exception is not None:
+                error, traceback = self.process.exception
+                self.local_reporter.terminate()
+                self.dist_reporter(done=True, traceback=traceback)
```
mseeger: Not sure if done=True is OK here. In the end, the process aborted. done=True I think implies it stopped normally and published a metric, no?

autogluon/scheduler/reporter.py
```diff
@@ -174,15 +174,29 @@ def __init__(self, process, local_reporter, dist_reporter):
 
     def run(self):
         while self.process.is_alive():
+
+            # breaking communication if process raises exception
+            if self.process.exception is not None:
+                error, traceback = self.process.exception
```
mseeger: error is not being used

autogluon/scheduler/reporter.py
razmikmelikbekyan: In general I agree with you that `done=True` is a bit confusing. We can remove it. It should work without it also.

autogluon/scheduler/reporter.py
razmikmelikbekyan: yep, since its traceback already contains all the info. I can replace it with `_`, but maybe you have other ideas how we can use it here. I was thinking maybe we can raise this error somewhere.

mseeger: If the error tells me the type of exception, it should also be transmitted. It cannot hurt, even though most generic schedulers will not be able to use it.

razmikmelikbekyan: so where do you think is most appropriate to use it?

Innixma:
Would this enable Tabular to not have to wrap try/except in its HPO code?

razmikmelikbekyan:
> Would this enable Tabular to not have to wrap try/except in its HPO code?

Yes, I checked it with some examples from that fix.

autogluon/scheduler/reporter.py
mseeger: Well, normally a scheduler would just note that the evaluation failed, and it would react to any such failure in the same way. Only highly specific schedulers could do more than that.
But certainly, we should log everything we can about what happened, it may be useful.

autogluon/scheduler/fifo.py
```diff
@@ -257,6 +257,10 @@ def _run_reporter(self, task, task_job, reporter, searcher):
         last_result = None
         while not task_job.done():
             reported_result = reporter.fetch()
+            if 'traceback' in reported_result:
```
mseeger: I can confirm that the reaction to an exception being reported (the signal is whether 'traceback' is present) is OK. What happens is that the _run_reporter while loop is left. The current reported_result is not used as a datapoint for the model (this is correct -- it does not have any metric value). What *is* used is the last recently fetched reported_result which did not fail. If the process reports an exception before ever obtaining a metric value, then nothing is reported to the model.

For most searchers using FIFO scheduler, this should be a good reaction. We want to make sure that a config whose evaluation has failed, is not attempted again. At least for RandomSearcher, this currently works, because a config is never attempted twice in the first place.

mseeger:
Given this goes in, I will later add some other method to BaseSearcher, which the scheduler calls if an evaluation failed. While the current searchers maybe don't need that, more advanced searchers would then be able to react. For example, they could explicitly blacklist the config, or even better, they could model the success/failure probability, and then this would be data.

PR #346, opened 2020-03-02
Use pytest for unit tests

fela:
Tests can be run with $ python3 -m pytest tests

I disabled slow tests by default, but they will be run by Jenkins and
can be run manually by adding the --runslow option

All tabular benchmarks have been moved to the same file as they all
depend on run_tabular_benchmarks. This is the simplest way to share
a helper function between pytest tests.

To run the tests in a specific test use

$ python3 -m pytest path_to_file:test_mytest

:test_mytest can be omitted to run all tests in a specific file.


szha:
Job PR-346-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-346/2/index.html

mseeger:
Hello @jwmueller , @zhanghang1989 

Fela is an MLS from the Amper team, he wrote a lot of the HPO code which we intend to contribute.
This contribution is intended to put pytest in place for unit and other tests. It is needed because our contribution will have quite a number of such tests.

Could you put the person on the issue who oversees the Jenkins stuff? Would that be @szha ?

fela:
The reason I did not add it there is that it is not strictly a requirement for the pip package itself, and I did not want to add unneeded dependencies to our end users. It is a development dependency only, not sure if there is a canonical way to specify those. 

Instead I install it in the Jenkins shell commands. The Jenkins pipelines is green with this change, the run is [here](https://ci.mxnet.io/blue/organizations/jenkins/autogluon/detail/PR-346/2/pipeline/4). The only thing that we might still want to do is configuring Jenkins to use the generated results.xml for nicer reports. I don't think I have permissions for that, so can anyone off the other contributors do that after the merge?

fela:
But probably you are right, it's quite unlikely to cause a conflicts for our users, so will just add it for convenience.

szha:
Job PR-346-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-346/3/index.html

fela:
For skipping a test the standard way in pytest is to add `@pytest.mark.skip(reason="why the test should not be run")`, not sure it is worth adding this information anywhere. I checked the CI run of this PR and it appears to work well.

PR #354, opened 2020-03-05
Fix scheduler hpo

razmikmelikbekyan:
*Issue #329 , if available:*

*Description of changes:*
Continuation of #345 for hyperband and rl.

### DIFF
--- a/autogluon/scheduler/hyperband.py
+++ b/autogluon/scheduler/hyperband.py
@@ -219,6 +219,12 @@ class HyperbandScheduler(FIFOScheduler):
         last_updated = None
         while not task_job.done():
             reported_result = reporter.fetch()
+            if 'traceback' in reported_result:
+                logger.exception(reported_result['traceback'])
+                # reporter.move_on()
+                # terminator.on_task_remove(task)
+                break
+
             if reported_result.get('done', False):
                 reporter.move_on()
                 terminator.on_task_complete(task, last_result)
--- a/autogluon/scheduler/hyperband.py
+++ b/autogluon/scheduler/hyperband.py
@@ -221,8 +221,8 @@ class HyperbandScheduler(FIFOScheduler):
             reported_result = reporter.fetch()
             if 'traceback' in reported_result:
                 logger.exception(reported_result['traceback'])
-                # reporter.move_on()
-                # terminator.on_task_remove(task)
+                reporter.move_on()
+                terminator.on_task_remove(task)
                 break
 
             if reported_result.get('done', False):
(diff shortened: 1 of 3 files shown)


autogluon/scheduler/hyperband.py
```diff
@@ -219,6 +219,10 @@ def _run_reporter(self, task, task_job, reporter, searcher, terminator):
         last_updated = None
         while not task_job.done():
             reported_result = reporter.fetch()
+            if 'traceback' in reported_result:
+                logger.exception(reported_result['traceback'])
```
mseeger: Please insert
   terminator.on_task_remove(task)
here. This is necessary for the terminator to keep its datastructures in order.

autogluon/scheduler/hyperband.py
```diff
@@ -219,6 +219,10 @@ def _run_reporter(self, task, task_job, reporter, searcher, terminator):
...
+                break
```
mseeger: Looking at this, and also FIFOScheduler:
Don't we also need to call
   reporter.move_on()  ?
I am not deep in the details here, but I note this is done if the reporter signals 'done', so should it not also be done if the reporter signals an exception?

autogluon/scheduler/rl_scheduler.py
```diff
@@ -236,6 +236,11 @@ def _run_reporter(task, task_job, reporter):
             last_result = None
             config = task.args['config']
             while not task_job.done():
+                reported_result = reporter.fetch()
```
zhanghang1989: Why calling `reported_result = reporter.fetch()` twice?

autogluon/scheduler/hyperband.py
razmikmelikbekyan: yep, this is good catch, I added it.

autogluon/scheduler/hyperband.py
razmikmelikbekyan: done

szha:
Job PR-354-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-354/2/index.html

szha:
Job PR-354-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-354/3/index.html

Innixma:
@mseeger You're review still shows as requesting changes, can you change to approved so that the PR can be merged?

Innixma:
Merging into mainline, thanks again for your contribution!

PR #344, opened 2020-02-29
FIX: basic PEP8 fixes and improvements

razmikmelikbekyan:
*Issue #, if available:*

*Description of changes:*
Basic PEP8 corrections and improvements. 

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #355, opened 2020-03-10
Fixed defect causing LightGBM to fail on multiclass classification

Innixma:
*Issue #, if available:*

*Description of changes:*

Fixed defect causing LightGBM to fail on multiclass classification

Defect introduced in #344 

This highlights that we should alter tests to fail if **ANY** model failure occurs, rather than if training fails entirely.


Innixma:
Exception prior to fix:

```
lightgbm.basic.LightGBMError: Number of classes must be 1 for non-multiclass training
Number of classes must be 1 for non-multiclass training
```

szha:
Job PR-355-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-355/1/index.html

PR #362, opened 2020-03-17
Tabular: Improved ngram feature generation stability

Innixma:
*Issue #, if available:*

*Description of changes:*

Adjusted ngram max_memory_percentage from 0.2 -> 0.15


szha:
Job PR-362-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-362/1/index.html

Innixma:
> More broadly, we should consider allowing user to specify memory limits

I agree, but this is a more complex feature add, depending on how much freedom we want to give the users. Plus it would have to be another value passed to all models, which would get complicated. Longer term we should definitely have the option to specify (and ignore) memory limits.

PR #363, opened 2020-03-17
[WIP] Add custom tsv/csv dataset support for text classification

cgraywang:
*Description of changes:*
- Add custom dataset support class CustomTSVClassificationTask
- Add example of using custom dataset
- Add example of using glue dataset
- Add unittest
- Add docs


autogluon/task/text_classification/dataset.py
```diff
@@ -1,34 +1,37 @@
...
+import multiprocessing as mp
 from typing import AnyStr
+import pandas as pd
 
 from mxnet import gluon
 from mxnet.metric import Accuracy, F1, MCC, PearsonCorrelation, CompositeEvalMetric
 import gluonnlp as nlp
 from gluonnlp.data import GlueCoLA, GlueSST2, GlueSTSB, GlueMRPC
 from gluonnlp.data import GlueQQP, GlueRTE, GlueMNLI, GlueQNLI, GlueWNLI
+# from gluonnlp.data.utils import (Splitter, concat_sequence, line_splitter, whitespace_splitter)
```
mli: remove it

autogluon/task/text_classification/dataset.py
```diff
@@ -1,34 +1,37 @@
...
             Name describing which built-in popular text dataset to use (mostly from the GLUE NLP benchmark).
             Options include: 'mrpc', 'qqp', 'qnli', 'rte', 'sts-b', 'cola', 'mnli', 'wnli', 'sst', 'toysst'. 
             Detailed descriptions can be found in the file: `autogluon/task/text_classification/dataset.py`
-        train : bool
-            Whether this data will be used for training models.
     """
+    path = kwargs.get('filepath', None)
     if path is not None:
-        raise NotImplementedError
+        if '.tsv' or '.csv' in path:
```
mli: not `in`, use `path.endswith`

autogluon/task/text_classification/dataset.py
```diff
@@ -1,34 +1,37 @@
...
+            return CustomTSVClassificationTask(*args, **kwargs)
```
mli: why `get_dataset` returns `CustomTSVClassificationTask` which sounds like a task

autogluon/task/text_classification/dataset.py
```diff
@@ -119,28 +122,167 @@ def get_dataset(self, segment='train'):
...
-        metric = CompositeEvalMetric()
-        metric.add(F1())
-        metric.add(Accuracy())
-        super(TSVClassificationTask, self).__init__(class_labels, metric, is_pair)
-        dataset = nlp.data.TSVDataset(*self.args, **self.kwargs)
-        # do the split
-        train_sampler, val_sampler = get_split_samplers(dataset, split_ratio=0.8)
-        self.trainset = SampledDataset(dataset, train_sampler)
-        self.valset = SampledDataset(dataset, val_sampler)
+class CustomTSVClassificationTask(AbstractGlueTask):
```
mli: why it is a subclass of `AbstractGlueTask`, even that it has same the data format as the GLUE dataset, but semantically it's a not a GLUE task.

autogluon/task/text_classification/dataset.py
```diff
@@ -119,28 +122,167 @@ def get_dataset(self, segment='train'):
...
+    """
+
+    Parameters
+    ----------
+    filepath: str, path of the file
```
mli: how about multiple files?

autogluon/task/text_classification/dataset.py
```diff
@@ -119,28 +122,167 @@ def get_dataset(self, segment='train'):
...
+        Any valid string path is acceptable.
+    sep : str
+        Delimiter to use.
+    delimiter : str, default ``None``
```
mli: why needs an alias? it confuse users

autogluon/task/text_classification/dataset.py
```diff
@@ -119,28 +122,167 @@ def get_dataset(self, segment='train'):
...
+        Alias for sep.
+    header : int, list of int, default 'infer'
```
mli: the default data type is string, not a int or a list of int

autogluon/task/text_classification/dataset.py
```diff
@@ -119,28 +122,167 @@ def get_dataset(self, segment='train'):
...
+        Row number(s) to use as the column names, and the start of the
+        data.
+    names : array-like, optional
```
mli: prefer `columns_to_use`

autogluon/task/text_classification/dataset.py
```diff
@@ -119,28 +122,167 @@ def get_dataset(self, segment='train'):
...
+        List of column names to use.
+    index_col : int, str, sequence of int / str, or False, default ``None``
```
mli: what's that?

autogluon/task/text_classification/dataset.py
```diff
@@ -119,28 +122,167 @@ def get_dataset(self, segment='train'):
...
+    usecols : list-like or callable, optional
```
mli: what's that?

autogluon/task/text_classification/dataset.py
```diff
@@ -119,28 +122,167 @@ def get_dataset(self, segment='train'):
...
+        Return a subset of the columns.
+    squeeze : bool, default False
```
mli: what's that?

autogluon/task/text_classification/dataset.py
```diff
@@ -119,28 +122,167 @@ def get_dataset(self, segment='train'):
...
+    prefix : str, optional
```
mli: what's that?

mli: I stoped here, there a large amount of arguments that I cannot understand, or are they necessary to expose to users? the key idea of autogluon is to have users to specify the minimal arguments. Also, as the general rule, a function should not have too many arguments...

autogluon/task/text_classification/dataset.py
```diff
@@ -119,28 +122,167 @@ def get_dataset(self, segment='train'):
...
+        kwargs.pop("filepath")
+        dataset_df = pd.read_csv(**kwargs)
+        dataset_df_lst = dataset_df.values.tolist()
+        self.dataset = gluon.data.SimpleDataset(dataset_df_lst)
+        split_ratio = kwargs.get('split_ratio', None) if 'split_ratio' in kwargs else 0.8
+        self.train_sampler, self.dev_sampler = get_split_samplers(self.dataset, split_ratio=split_ratio)
 
     def dataset_train(self):
-        return 'train', self.trainset
+        """Get the training segment of the dataset for the task.
```
mli: why we need a function to do it?

examples/text_classification/example_custom_dataset.py
```diff
@@ -0,0 +1,7 @@
+from autogluon import TextClassification as task
+
+dataset = task.Dataset(filepath='/home/ubuntu/data/tsv/train.csv', usecols=['text', 'target'])
```
mli: what if the username is not ubuntu? where is the train.csv file?

examples/text_classification/example_glue_dataset.py
```diff
@@ -0,0 +1,14 @@
+import autogluon as ag
+from autogluon import TextClassification as task
+
+dataset = task.Dataset(name='SST')
+predictor = task.fit(dataset, epochs=5, warmup_ratio=0.1, log_interval=400, seed=2, num_trials=100, dev_batch_size=8,
```
mli: why we need to specify so many arguments? can we just `task.fit(dataset)`

tests/unittests/test_text_classification.py
```diff
@@ -0,0 +1,13 @@
+from autogluon import TextClassification as task
+
+def test_fit():
+    dataset = task.Dataset(name='ToySST')
```
mli: `CustomTSVClassificationTask` is not tested.

examples/text_classification/example_custom_dataset.py
Innixma: Agreed, all example files should be runnable, additionally they should have relative paths. Expect the user's working directory to be the autogluon root directory.

autogluon/task/text_classification/dataset.py
Innixma: If Tabular's load_pd.load() was used it would be able to handle multiple files without issue.

autogluon/task/text_classification/dataset.py
```diff
@@ -119,28 +122,167 @@ def get_dataset(self, segment='train'):
...
+    memory_map : bool, default False
+        See <https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html#pandas.read_csv`_ .
+    float_precision : str, optional
+        See <https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html#pandas.read_csv`_ .
+    """
+    def __init__(self, *args, **kwargs):
+        self._read(**kwargs)
+        self.is_pair = False
+        self.class_labels = list(set([sample[1] for sample in self.dataset]))
+        self.metric = Accuracy()
```
Innixma: What if I want to optimize for AUC? Logloss? F1?

autogluon/task/text_classification/dataset.py
```diff
@@ -119,28 +122,167 @@ def get_dataset(self, segment='train'):
...
+        elif segment == 'dev':
+            return SampledDataset(self.dataset, self.dev_sampler)
+        else:
+            raise NotImplementedError
+
+    def _read(self, **kwargs):
+        path = kwargs.get('filepath', None)
+        kwargs['filepath_or_buffer'] = path
+        kwargs.pop("filepath")
+        dataset_df = pd.read_csv(**kwargs)
```
Innixma: I recommend using Tabular's load_pd.load()

autogluon/task/text_classification/dataset.py
```diff
@@ -119,28 +122,167 @@ def get_dataset(self, segment='train'):
...
+        dataset_df_lst = dataset_df.values.tolist()
+        self.dataset = gluon.data.SimpleDataset(dataset_df_lst)
```
Innixma: Why not keep it as a DataFrame? What is gained by this SimpleDataset class?

autogluon/task/text_classification/dataset.py
```diff
@@ -119,28 +122,167 @@ def get_dataset(self, segment='train'):
...
+        split_ratio = kwargs.get('split_ratio', None) if 'split_ratio' in kwargs else 0.8
```
Innixma: Just have it be a default value in __init__, this is unnecessary

autogluon/task/text_classification/dataset.py
```diff
@@ -119,28 +122,167 @@ def get_dataset(self, segment='train'):
...
+        See <https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html#pandas.read_csv`_ .
+    memory_map : bool, default False
+        See <https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html#pandas.read_csv`_ .
+    float_precision : str, optional
+        See <https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html#pandas.read_csv`_ .
+    """
+    def __init__(self, *args, **kwargs):
+        self._read(**kwargs)
+        self.is_pair = False
+        self.class_labels = list(set([sample[1] for sample in self.dataset]))
```
Innixma: Should we be accessing by index here? Isn't this supposed to be a DataFrame? How was I supposed to know that sample[1] is the label? Label header parameter isn't passed in, so are you assuming users will have label as the 2nd element?

autogluon/task/text_classification/dataset.py
jwmueller: Agreed & to add to this, can we call it something like "TSVDataset" instead?  
Naming is important here since this is user-facing and will appear in the docs.
I especially don't think the word "Custom" needs to appear, whether or not "Task" or "Classification" should be there is up to you, but it clashes with the fact that the usual workflow starts with:
```
import TextClassification as task 
```
(I think "Task" remains ok for the other GLUE dataset objects since those datasets are all associated with a very particular task beyond generic TextClassification).

autogluon/task/text_classification/dataset.py
jwmueller: Right, I'd remove all these from the docs and just say something end the documentation for this method with something like: 

"Numerous other keyword arguments can be provided which will be passed into pandas.read_csv().  Some you might find most useful include: sep, delimiter, header, etc. [just list <5 or so of the most useful options here]."  

Also as Nick mentioned you can just replace `pd.read_csv()` with `load_pd.load()` used in TabularPrediction, or extend that method if you require more args specific to text datasets (load_pd.load uses pd.read_csv under the hood but brings numerous benefits)

autogluon/task/text_classification/dataset.py
```diff
@@ -119,28 +122,167 @@ def get_dataset(self, segment='train'):
...
-        self.valset = SampledDataset(dataset, val_sampler)
+class CustomTSVClassificationTask(AbstractGlueTask):
+    """
+
+    Parameters
+    ----------
+    filepath: str, path of the file
+        Any valid string path is acceptable.
+    sep : str
+        Delimiter to use.
```
jwmueller: Shouldn't `sep = '\t'` by default, given that this is called "TSV" dataset?

jwmueller:
Could you remove these warning statements given that TextClassification should now be complete?
  
https://github.com/cgraywang/autogluon-1/blob/3b7100354afc5d31c79c7a46814d177728d894c7/autogluon/task/text_classification/text_classification.py#L146-L148

examples/text_classification/example_custom_dataset.py
jwmueller: Could you also provide an example somewhere to show how to use a user-specified train/validation split?

examples/text_classification/example_custom_dataset.py
```diff
@@ -0,0 +1,7 @@
...
+predictor = task.fit(dataset, epochs=1)
+print('Top-1 val acc: %.3f' % predictor.results['best_reward'])
+test_acc = predictor.evaluate(dataset)
```
jwmueller: This example could be better if you also showed how to: 
- read in a separate test dataset and make the predictions/evaluation on it (instead of the same training data)
- reload the trained predictor from disk after task.fit()

examples/text_classification/example_custom_dataset.py
jwmueller: Also this `train.csv` should probably actually live in the autogluon s3 bucket, like our example datasets for TabularPrediction:

https://github.com/awslabs/autogluon/blob/master/examples/tabular/example_simple_tabular.py

jwmueller:
A high level concern I have is many text-prediction tasks involve regression (eg. predicting sentiment, polarity, subjectivity, relevance, etc), and regression also seems trivial to add to this code (I'd be happy to do it).  Thus are we unnecessarily bottlenecking autogluon's use-cases with the name "TextClassification"?  While that name might need to stay for legacy-sake, it's not clear to me that  the new `CustomTSVClassificationTask` dataset object should have the word "Classification" in it...

Innixma:
> Could you remove these warning statements given that TextClassification should now be complete?
> 
> https://github.com/cgraywang/autogluon-1/blob/3b7100354afc5d31c79c7a46814d177728d894c7/autogluon/task/text_classification/text_classification.py#L146-L148

FYI: I think we should keep this warning in for this PR since it seems many changes might have to occur, we can remove it in a follow-up PR

autogluon/task/text_classification/dataset.py
cgraywang: 1) conceptually it is not just a dataset it has evaluation metric, statics around dataset such as class_labels; 2) to align with the BERT family model training pipeline, especially the necessary data transformation, the information in the task are needed, not only the dataset; 3) to align with the other benchmark dataset implementation in gluonnlp, such as GLUE etc.

autogluon/task/text_classification/dataset.py
cgraywang: Please find my reply: https://github.com/awslabs/autogluon/pull/363#discussion_r394675221
Task is more proper for NLP pipeline. Dataset is not comprehensive enough

autogluon/task/text_classification/dataset.py
cgraywang: I am targeting on basic and popular case in this PR which can be found in most benchmark and competition, where each dataset is a single file. In future, we could support this. To leverage load_pd, we would need to refactory the code to make it compatible between modules.

autogluon/task/text_classification/dataset.py
cgraywang: This is for documentation. I can improve the documentation if necessary. For load_pd, I think we need detailed documention as well as certain amount of refactory to be shared between tabular and nlp.

autogluon/task/text_classification/dataset.py
cgraywang: to make it align with the bert family training pipeline, see https://github.com/awslabs/autogluon/pull/363#discussion_r394675221

autogluon/task/text_classification/dataset.py
cgraywang: We can always use a composite metric to do that. The goal of the PR is to support basic custom dataset in tsv/csv format for classification. Accuracy is common and must-have in nlp

autogluon/task/text_classification/dataset.py
cgraywang: see https://github.com/awslabs/autogluon/pull/363#discussion_r394677735
We need both refactory and documentation for load_pd...

autogluon/task/text_classification/dataset.py
cgraywang: to align with BERT-family training pipeline, details in https://github.com/awslabs/autogluon/pull/363#discussion_r394675221

autogluon/task/text_classification/dataset.py
cgraywang: I think having a default value means an additional required argument, which is actually unnecessary.

autogluon/task/text_classification/dataset.py
cgraywang: support text/target for now, which has been discussed in the meeting.

autogluon/task/text_classification/dataset.py
Innixma: Why do you need to refactor load_pd? It is a file loading function. It does not need to be refactored.

autogluon/task/text_classification/dataset.py
Innixma: We should not be documenting all of these options. At most we should maybe document the most critical ones and then have others be optionally specified in some data_loader_args dictionary parameter. You can just provide a link to the pd.read_csv() function which describes the valid parameters. Honestly, we should likely not even provide full functionality and access to all parameters of read_csv. If a user wants to use those functionalities, they can load their data prior to the fit call and pass it in as a DataFrame input to the task. This is how Tabular works.

autogluon/task/text_classification/dataset.py
Innixma: Why does the bert family training pipeline need a function for this?

autogluon/task/text_classification/dataset.py
Innixma: >We can always use a composite metric to do that.

Not with how the code is right now. At the very minimum it should support the user specifying the metric as a parameter to the call.

autogluon/task/text_classification/dataset.py
Innixma: A default value by definition means that it is not a required argument, because the user doesn't have to specify it (default is used if not specified).

autogluon/task/text_classification/dataset.py
Innixma: This is highly restrictive and arbitrary. A lot of users are going to be frustrated by this. This should be changed to require the user to specify the label column and then assume all other columns are text or require them to specify text column, something to that effect.

autogluon/task/text_classification/dataset.py
Innixma: Probably ',' , I think csv is generally more common. I don't think this should necessarily be called "TSV" if it supports all kinds of delimited files. I honestly don't think the name of the task should even mention the method of file loading.

autogluon/task/text_classification/dataset.py
Innixma: A dataset is not a task. Why does a dataset have an evaluation metric variable? This should not be the case.

jwmueller: Is "task" needed as opposed to "dataset" because you care about handling pairs of sentences for certain tasks like Q&A?  I think we should just hold off on these in the custom dataset case since it can only handle a single text-field for now.

Couldn't all the other things you mentioned be handled in TextClassification.fit(), where the user would specify: evaluation_metric, class_labels (although these should be ideally detected automatically; you can reuse our tabular code for this in `label_cleaner.py`), etc.  

Also I don't think we should try to mold the user-experience/API with a custom dataset after methods specific to GLUE/BERT/GluonNLP.  Rather, I'd say these specific methods should be molded to the custom dataset UX, which is far more important than any one specific benchmark/GLUE task.

autogluon/task/text_classification/dataset.py
jwmueller: you could easily just extend load_pd to accept kwargs and pass those to pd.read_csv (if needed)

autogluon/task/text_classification/dataset.py
jwmueller: I'm just saying if the name of it TSV, then "\t" should be the default (I'm happy for it also to be CSV with "," as the default).  I think TSV format is more common for text data because commas appear way more frequently than tabs in natural language.  But I agree the method of file-loading probably does not need to be in the name of the class

cgraywang: Would detail the data format and corresponding the sep in the doc.

autogluon/task/text_classification/dataset.py
cgraywang: The current design is to ask the user to both specify the text and target/label column.

autogluon/task/text_classification/dataset.py
cgraywang: What I mean is the following: def(split_ratio=0.8, **kwargs) comparing to def(**kwargs), there is an extra argument in the function definition.

autogluon/task/text_classification/dataset.py
cgraywang: Ok. It could be an argument of the abstract task class.

examples/text_classification/example_glue_dataset.py
cgraywang: definitely...

autogluon/task/text_classification/dataset.py
cgraywang: For the data transformation in the pipeline, to align with the processing data functionality we have for the benchmarking bert family model as in the implementation.

autogluon/task/text_classification/dataset.py
cgraywang: Yes, I think we agree to have both better documentation for nlp data and tabular data now.

autogluon/task/text_classification/dataset.py
cgraywang: three simple reasons: 1) load_pd is a tabular functionality in design and structure; 2) it can be refined from both implementation and documentation perspective; 3) minor thing: `load_pd` doesn't follow python naming practice.

autogluon/task/text_classification/dataset.py
cgraywang: Task contains dataset, metric and other information needed to train BERT family models both from concept and implementation point of views. If we don't agree on the BERT training processing pipeline, we could either refactor the code for benchmarking datasets. This could yield duplicate codes but function-wise the same. We will need extra amount of work to address this and to align with gluonnlp to enable future pretrained models available in gluonnlp could be directly adapted to autogluon text classification module.

autogluon/task/text_classification/dataset.py
Innixma: Yes, I think it is good to have instead of just only **kwargs. We should honestly minimize the amount of parameters that have to be specified through kwargs. For example, Tabular only has very niche parameters as kwargs.

cgraywang: Yes, this is a design preference. I would rather provide a simplified interface with a doc string explaining what's available. Agree on we should specify which keyword arguments are available to the customers.

Innixma:
Can you update this PR with changes to make it pass CI and move the data loading parameters to some data_loading_kwargs parameter? I'd also remove most of the documentation of the available data loading kwargs and just link to pd.read_csv documentation. Also please add a train_dataset parameter which can be used to avoid having to load the data if it is already loaded.

Regarding your comments on load_pd, load_pd is not a user facing function, so we don't hold the same bar in terms of documentation requirements. No user has ever complained about load_pd or was confused on how to load data for Tabular, nor have they asked for additional data loading parameter options, which seems to me that it isn't a major issue.

I don't believe load_pd requires any code changes to be used by NLP.

autogluon/task/text_classification/dataset.py
```diff
@@ -141,93 +138,7 @@ class CustomTSVClassificationTask(AbstractGlueTask):
...
-        See <https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html#pandas.read_csv`_ .
-    dialect : str or csv.Dialect, optional
-        See <https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html#pandas.read_csv`_ .
-    error_bad_lines : bool, default True
-        See <https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html#pandas.read_csv`_ .
-    warn_bad_lines : bool, default True
-        See <https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html#pandas.read_csv`_ .
-    delim_whitespace : bool, default False
-        See <https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html#pandas.read_csv`_ .
-    low_memory : bool, default True
```
Innixma: low_memory mode should always be False, it is harmful in many cases.

autogluon/task/text_classification/dataset.py
```diff
@@ -119,28 +121,79 @@ def get_dataset(self, segment='train'):
...
-        self.trainset = SampledDataset(dataset, train_sampler)
-        self.valset = SampledDataset(dataset, val_sampler)
+class CustomTSVClassificationTask(AbstractGlueTask):
+    """
+
+    Parameters
+    ----------
+    filepath: str, path of the file
+        Any valid string path is acceptable.
+    sep : str
```
Innixma: Default value?

autogluon/task/text_classification/dataset.py
```diff
@@ -119,28 +121,79 @@ def get_dataset(self, segment='train'):
...
-        # do the split
-        train_sampler, val_sampler = get_split_samplers(dataset, split_ratio=0.8)
-        self.trainset = SampledDataset(dataset, train_sampler)
-        self.valset = SampledDataset(dataset, val_sampler)
+class CustomTSVClassificationTask(AbstractGlueTask):
+    """
+
+    Parameters
+    ----------
+    filepath: str, path of the file
```
Innixma: Add option for user to give a pandas DataFrame as input (train_data)

autogluon/task/text_classification/dataset.py
```diff
@@ -119,28 +121,79 @@ def get_dataset(self, segment='train'):
...
+        Delimiter to use.
+    header : int, list of int, default 'infer'
+        Row number(s) to use as the column names, and the start of the
+        data.
+    names : array-like, optional
+        List of column names to use.
+    index_col : int, str, sequence of int / str, or False, default ``None``
+    usecols : list-like or callable, optional
+        Return a subset of the columns.
+    For our keyword argument, plesase see
```
Innixma: Typo and please rephrase to:

'For additional keyword arguments related to loading the data, please see'

autogluon/task/text_classification/dataset.py
```diff
@@ -119,28 +121,79 @@ def get_dataset(self, segment='train'):
...
+            raise NotImplementedError
+
+    def _read(self, **kwargs):
+        path = kwargs.get('filepath', None)
+        kwargs['filepath_or_buffer'] = path
+        kwargs.pop("filepath")
+        dataset_df = pd.read_csv(**kwargs)
+        dataset_df_lst = dataset_df.values.tolist()
+        self.dataset = gluon.data.SimpleDataset(dataset_df_lst)
+        split_ratio = kwargs.get('split_ratio', None) if 'split_ratio' in kwargs else 0.8
```
Innixma: split_ratio not documented and furthermore should be an explicit arg with a default value.

autogluon/task/text_classification/dataset.py
```diff
@@ -119,28 +121,79 @@ def get_dataset(self, segment='train'):
...
+    filepath: str, path of the file
+        Any valid string path is acceptable.
+    sep : str
+        Delimiter to use.
+    header : int, list of int, default 'infer'
+        Row number(s) to use as the column names, and the start of the
+        data.
+    names : array-like, optional
+        List of column names to use.
+    index_col : int, str, sequence of int / str, or False, default ``None``
```
Innixma: Why default None? What does that mean?

autogluon/task/text_classification/dataset.py
Innixma: Any valid string path? So 'file.custom_extension is valid? I don't think it is.

autogluon/task/text_classification/dataset.py
```diff
@@ -119,28 +121,79 @@ def get_dataset(self, segment='train'):
...
+        elif segment == 'dev':
+            return SampledDataset(self.dataset, self.dev_sampler)
+        else:
+            raise NotImplementedError
+
+    def _read(self, **kwargs):
+        path = kwargs.get('filepath', None)
+        kwargs['filepath_or_buffer'] = path
+        kwargs.pop("filepath")
+        dataset_df = pd.read_csv(**kwargs)
```
Innixma: Add data loading specific parameter that houses the params, don't use kwargs here.

autogluon/task/text_classification/dataset.py
Innixma: Dataset should be entirely separate from task. A task takes a dataset as input to .fit(). A dataset should not know things like evaluation metric, label, split_ratio, etc. Ideally, the dataset should just be a pandas DataFrame or as close to a DataFrame as possible. We shouldn't have to init CustomTSVClassificationTask with any parameters, it should all be passed in the .fit() call, as done in Tabular.

cgraywang: 1. The task here is an NLP task which aligns with the GLUE task both conceptually and implementation-wise (aligns with the NLP). 2. The nlp task contains dataset class, dataset class doesn't contain evaluation metric.... 3. CustomTSVClassificationTask is an internal implementation. The dataset is processed as pd dataframe but need to convert it to MXNet dataset in order to make it run with mxnet based training pipeline. In tabular, there is no HPO supported, this is different from the scope of NLP. NLP considers HPO as the most basic component in AutoML.

autogluon/task/text_classification/dataset.py
cgraywang: Again, this is design preference. I would prefer such design and provides more detailed documentation to specify the valid user argument. Besides, it is internal implementation, users don't interact directly with the class.

autogluon/task/text_classification/dataset.py
cgraywang: It is just a string indicating a path to the file, which is similar to the case in tabular.

autogluon/task/text_classification/dataset.py
cgraywang: We have already talked about it. See https://github.com/awslabs/autogluon/pull/363#discussion_r395928171

autogluon/task/text_classification/dataset.py
cgraywang: For the initial tsv support and as normal practice in NLP, I would prefer just support the limited file type. It introduces less dependency and concept (e.g., dataframe) to the user, as it is a user interface.

autogluon/task/text_classification/dataset.py
cgraywang: Please comment in the updated version. Thanks

Innixma: I am saying that by default pd.read_csv sets low_memory=True, but you should explicitly default it to False in your code.

autogluon/task/text_classification/dataset.py
Innixma: You are already loading it into a DataFrame internally. They don't need to use this. If you look at Tabular, train_data can be either str or DataFrame. I think this should act in the same way.

Innixma: I understand that, but in your code you raise an exception if the file name does not end in .tsv or .csv. This is not communicated in this documentation.

cgraywang: This is a user interface. Regardless what's been implemented internally, I think it is more important to expose to NLP users what they are used to: dataset path, less dependency (as we already have many dependencies internally, do not pass the burden to the user further).

cgraywang: If the user's data format is not in tsv or csv, or not benchmark dataset supported in gluonnlp, we would raise NotImplementedError.

tests/unittests/test_text_classification.py
```diff
@@ -0,0 +1,31 @@
...
+                         pretrained_dataset=ag.Categorical('book_corpus_wiki_en_uncased'),
+                         epochs=1,
+                         num_trials=1,
+                         batch_size=4,
+                         dev_batch_size=4,
+                         max_len=16)
+    test_acc = predictor.evaluate(dataset)
+    print('finished')
+
+def test_custom_dataset_fit():
```
Innixma: CI Failed:

tests/unittests/test_text_classification.py Fatal Python error: Fatal Python error: AbortedAborted

Can this be fixed? Also, could you try to update test to work on CPU? Once it does, I can try this locally to get a better idea of how the code is structured.

Innixma: This will probably work best since my understanding of the code may be somewhat incorrect according to this test. I thought that the user would import the CustomTSVTask and create it themselves, but it appears this is not the case.

cgraywang: I tried to fix the ci however it took 3h to make a pass, and though it is so strange we do not have a GPU machine to run the ci. We should have one. It can of course run on the CPU. I am not sure why you want to run the code locally in order to help you understand the code structure. If you really want to understand, you could actually leverage a GPU machine that could make the process faster.

cgraywang: I do not quite understand what you mean by " the code is incorrect" according the test. You have not ran the code yet based on your previous comment. For now the user just need to provide the dataset path to enable custom dataset usage. It should be easier than what you suggested, and it is the same with the tabular dataset design.

Innixma: I did not say the code is incorrect, I said that my understanding of the how the code worked was likely not accurate. Regarding running its locally, it is so I can use the python debugger from my Mac instead of debugging from a remote machine which is more difficult.

I believe that the CI machine does has a GPU, but you can ask @zhanghang1989 if this is the case.

szha:
Job PR-363-4 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-363/4/index.html

tests/unittests/test_text_classification.py
cgraywang: Actually you can leverage python pdb to debug on remote machines: https://docs.python.org/3/library/pdb.html. It could save time in lots of cases, e.g., especially when we are dealing with deep and large models. 

From the ci information, I think we have one gpu node "
NODE_NAME=linux-gpu-g4dn-2x-6", however it could be not enough to test some large models when there is a need with reasonable time limits.

autogluon/task/text_classification/dataset.py
Innixma: 2. When task.Dataset() is called, it returns a Task which you have named 'dataset' in your test. With how the code is structured, there is no such thing as a SimpleDataset without a Task that holds it, so effectively to load a dataset, it must also have a metric. This should not be necessary.

3. HPO is supported in Tabular, so I am uncertain what you mean in point 3. Nothing I mentioned was specific to HPO functionality.

cgraywang: Regarding the dataset discussion, please refer to the previous response and see the reasons there. We need to have mxnet dataset to enable the dataset to be used in mxnet training script, e.g., the BERT family training scripts in autogluon and gluonnlp. For the metric, please refer to the previous response and see the reasons there. 

It is good that you now support the HPO in tabular. Could you please point out where is the code to support the hyperparameter optimization for tabular nn training, such as model candidates and learning rate range. Thanks.

autogluon/task/text_classification/dataset.py
```diff
@@ -116,31 +120,168 @@ def get_dataset(self, segment='train'):
...
+    For other keyword arguments, please see
+        <https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html#pandas.read_csv`_ .
+    """
+    def __init__(self, *args, **kwargs):
+        self._read(**kwargs)
+        self.is_pair = False
+        self._label_column_id = self._index_label_column(**kwargs)
+        self.class_labels = list(set([sample[self._label_column_id] for sample in self.dataset]))
+        self.metric = Accuracy()
+        super(CustomTSVClassificationTask, self).__init__(self.class_labels, self.metric, self.is_pair)
```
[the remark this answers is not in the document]
cgraywang: There are at least two reasons: 1) it provides better semantic of the code. By specifying self.pair = False then pass it to parent class, it gives user the better understanding of what this argument is rather than a random variable False; 2) it provides better flexibility. The user could modify it based their usage.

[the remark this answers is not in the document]
cgraywang: Same reply. There are at least two reasons: 1) it provides better semantic of the code. By specifying self.class_labels then pass it to parent class, it gives user the better understanding of what this argument is rather than a random variable False; 2) it provides better flexibility. The user could modify it based their usage. 3) It also provides better code style, obviously, 
__init__(self.class_labels, self.metric, self.is_pair) is better regarding style compared to 
__init__(
list(set([sample[self._label_column_id] for sample in self.dataset]))
, self.metric, self.is_pair)

autogluon/task/text_classification/dataset.py
```diff
@@ -116,31 +120,168 @@ def get_dataset(self, segment='train'):
...
+        path = kwargs.get('filepath', None)
+        kwargs['filepath_or_buffer'] = path
+        kwargs.pop("filepath")
+        low_memory = kwargs.get('low_memory', None)
+        if low_memory is None:
+            kwargs['low_memory'] = False
+        dataset_df = pd.read_csv(**kwargs)
+        dataset_df_lst = dataset_df.values.tolist()
+        self.dataset = gluon.data.SimpleDataset(dataset_df_lst)
+        split_ratio = kwargs.get('split_ratio', None) if 'split_ratio' in kwargs else 0.8
```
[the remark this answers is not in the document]
cgraywang: Good catch.

autogluon/task/text_classification/dataset.py
```diff
@@ -116,31 +120,168 @@ def get_dataset(self, segment='train'):
...
+        self.train_sampler, self.dev_sampler = get_split_samplers(self.dataset, split_ratio=split_ratio)
```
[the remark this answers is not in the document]
cgraywang: Again, the scope of this PR is to support basic data and other functionalities needed to run custom dataset in text classification module. Now the current practice in kaggle nlp, they mostly only provide training dataset, I mainly support this common use case for now. And we would automatically generate the heldout dataset for validation purpose of hpo.

autogluon/task/text_classification/dataset.py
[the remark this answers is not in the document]
cgraywang: In the future, we would have more custom text format supported in the nlp module. 
AbstractCustomTask is the abstract class that every new dataset support inheriting from.
For example, now the tsv/csv dataset supports inherit from the abstract class.
As replied in the previous responses, the metric can be a composed metric including multiple metrics.
The current tsv dataset supports accuracy for now.

autogluon/task/text_classification/dataset.py
```diff
@@ -116,31 +120,168 @@ def get_dataset(self, segment='train'):
...
+
+    def _index_label_column(self, **kwargs):
+        label_column_id = kwargs.get('label', None)
+        if label_column_id is None:
+            return len(self.dataset[0]) - 1
```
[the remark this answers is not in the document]
cgraywang: Either way works for me. I prefer the current implementation as normally the last column is the label column. I would incorporate this in the documentation and add warning messages to the users.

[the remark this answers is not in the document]
Innixma: This is a blocking feature for me which needs to be changed for my PR approval. We need the user to be able to specify the label column, it is not safe to say that the last column will always be the label column, and it will be very tedious for the user to reformat their data to work in that case.

autogluon/task/text_classification/dataset.py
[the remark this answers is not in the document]
Innixma: I don't understand the points you are making here. This is Python 101 to avoid repeatedly declaring variables. That is one of the major points of using class inheritance.

autogluon/task/text_classification/dataset.py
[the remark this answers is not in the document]
Innixma: Until this is added, Tabular cannot make use of Text module. It is fine to not have in this PR at this point to avoid complicating things further.

szha:
Job PR-363-5 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-363/5/index.html

szha:
Job PR-363-6 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-363/6/index.html

autogluon/task/text_classification/dataset.py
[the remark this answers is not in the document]
Innixma: Tabular requires the ability to specify a specific validation set. Don't add that functionality to this PR though.

autogluon/task/text_classification/dataset.py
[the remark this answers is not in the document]
Innixma: `usecols` is documented as an optional parameter, yet is required if the user passes str value label_column_id. This isn't necessary. You can get it directly from the DataFrame itself by df.columns().

Regarding your comment: More flexible does not equate to more user-friendly. I think we should never infer label position, it should be a required parameter, as it is in all other tasks in AutoGluon and in virtually all well known ML algorithms such as XGBoost, LightGBM, CatBoost, sklearn, Keras, etc.

autogluon/task/text_classification/dataset.py
```diff
@@ -233,37 +240,53 @@ def get_dataset(self, segment='train'):
...
 
     def _read(self, **kwargs):
-        path = kwargs.get('filepath', None)
-        kwargs['filepath_or_buffer'] = path
-        kwargs.pop("filepath")
         low_memory = kwargs.get('low_memory', None)
         if low_memory is None:
             kwargs['low_memory'] = False
-        dataset_df = pd.read_csv(**kwargs)
+        split_ratio = kwargs.get('split_ratio', None) if 'split_ratio' in kwargs else 0.8
```
Innixma, line 249: Simplify these 5 lines to just:

`split_ratio = kwargs.pop('split_ratio', 0.8)`

autogluon/task/text_classification/dataset.py
```diff
@@ -233,37 +240,53 @@ def get_dataset(self, segment='train'):
...
+        if split_ratio is None:
+            split_ratio = 0.8
+        if 'split_ratio' in kwargs:
+            kwargs.pop('split_ratio')
+        path = kwargs.get('filepath', None)
+        kwargs['filepath_or_buffer'] = path
+        dataset_df = kwargs.get('df', None)
+        if path is not None:
```
Innixma, line 257: path should not silently override df, either raise exception if both are passed or prefer DF, since it is already loaded.

autogluon/task/text_classification/dataset.py
```diff
@@ -233,37 +240,53 @@ def get_dataset(self, segment='train'):
...
         low_memory = kwargs.get('low_memory', None)
         if low_memory is None:
             kwargs['low_memory'] = False
-        dataset_df = pd.read_csv(**kwargs)
+        split_ratio = kwargs.get('split_ratio', None) if 'split_ratio' in kwargs else 0.8
+        if split_ratio is None:
+            split_ratio = 0.8
+        if 'split_ratio' in kwargs:
+            kwargs.pop('split_ratio')
+        path = kwargs.get('filepath', None)
```
Innixma, line 254: pop instead to remove later redundant line

autogluon/task/text_classification/dataset.py
```diff
@@ -233,37 +240,53 @@ def get_dataset(self, segment='train'):
...
+        kwargs['filepath_or_buffer'] = path
+        dataset_df = kwargs.get('df', None)
```
Innixma, line 256: pop instead to remove later redundant line

autogluon/task/text_classification/dataset.py
```diff
@@ -198,13 +202,16 @@ class CustomTSVClassificationTask(AbstractCustomTask):
...
         Delimiter to use.
-    header : int, list of int, default 'infer'
+    header : int, list of int
         Row number(s) to use as the column names, and the start of the
         data.
     names : array-like, optional
         List of column names to use.
     usecols : list-like or callable, optional
         Return a subset of the columns.
+    split_ratio: float
```
Innixma, line 212: Mention it defaults to 0.8

autogluon/task/text_classification/dataset.py
```diff
@@ -233,37 +240,53 @@ def get_dataset(self, segment='train'):
...
+                raise ValueError('Please specify the `usecols` when specifying the label column.')
             else:
                 for i, col in enumerate(usecols):
                     if col == label_column_id:
                         return i
         elif isinstance(label_column_id, int):
             return label_column_id
         else:
-            raise NotImplementedError
+            raise ValueError('Please specify label column using either the column name (str) or index (int).')
```
Innixma, line 289: Just a comment, does not have to be resolved in this PR: A column name can be an integer. This breaks this functionality.

autogluon/task/text_classification/dataset.py
```diff
@@ -198,13 +202,16 @@ class CustomTSVClassificationTask(AbstractCustomTask):
         Any valid string path is acceptable.
     sep : str
         Delimiter to use.
-    header : int, list of int, default 'infer'
+    header : int, list of int
```
Innixma, line 205: Why did you remove 'infer' default?

autogluon/task/text_classification/dataset.py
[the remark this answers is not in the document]
Innixma: We will have to discuss this in a meeting. It appears there is some deeper misunderstanding going on.

autogluon/task/text_classification/dataset.py
cgraywang: This would fail when user specifies split_ratio=None, an additional if is needed to assign a non-none value.

autogluon/task/text_classification/dataset.py
cgraywang: good catch

autogluon/scheduler/remote/remote.py
```diff
@@ -80,7 +80,7 @@ def close(self, timeout=2):
         if self.service:
             self.service.shutdown()
         super().close(timeout)
-        self.shutdown()
+        # self.shutdown()
```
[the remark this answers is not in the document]
Innixma: Please get approvals regarding this change from @zhanghang1989 and @mseeger to ensure it is safe.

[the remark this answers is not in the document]
zhanghang1989: I don't think this change is necessary. Please make sure the correct versions for dask and distributed packages are installed.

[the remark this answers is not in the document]
cgraywang: @zhanghang1989 thanks, it looks like the version issue of disk and distributed.

szha:
Job PR-363-7 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-363/7/index.html

cgraywang:
Thanks @Innixma for the approval. @jwmueller and @mli , please feel free to let me know if you have more comments before merging the PR.

mli:
looks like my comment about the test is not addressed. in other words, just print the accuracy in the test is not effective, we need to at least check the accuracy if it's higher than a desired value. but i'm fine to merge this pr right now, as it took too long, and leave test to further PRs.

cgraywang:
Thanks @mli for the comments. I think I have already explained offline that the unittests needed to be aligned with all the modules, and some other issues need to be addressed first. For example, searcher would generate a random hyperparameter combination, and we could not ensure the performance has a lower bound given that the tests now run on cpu with a very limited number of trials and epochs. I agree that we do need to address this across the modules (not only text modules), so it should be addressed in a future separate PR (with possibily better CI setup as well). I will merge the PR now. But please let me know if you have more concerns.

PR #375, opened 2020-03-22
Linear models support

sgdread:
*Description of changes:*
This code adds linear models capabilities to the framework. Models are fully-integrated, but disabled by default: uncomment `LR` section in `fit()` method in  `autogluon/task/tabular_prediction/tabular_prediction.py`

The model has 3 variations, which are different by how in-model pre-processing and feature selection is performed. All operations are performed in sparse matrix form via Sklearn pipelines. Text features are controlled via `handle_text` hyper-parameter:
* `ignore` handles continuous and categorical features; features marked as NLP-capable are dropped. Preprocessing is similar to NNTabular: imputation, scaling, skew correction; categorical variables are one-hot encoded with a limit on max number of values in category.
* `only` handles only text features marked as NLP-capable. The fields are collapsed into a single text value, then TF-IDF tansformed.
* `include` combines both: NLP and regular features in a single model

For regression problems there are two options available via `penalty` optional arg: `L1` (Lasso regression) and `L2` (default - Ridge regressopm).

Example of usage:
```
from autogluon.utils.tabular.ml.trainer.auto_trainer import AutoTrainer
from autogluon.utils.tabular.ml.models.lr.lr_model import LinearModel

class LinearRegrssionTrainer(AutoTrainer):
    def get_models(self, hyperparameters={}, hyperparameter_tune=False):
        models = [
             LinearModel(
                 path=self.path, name='lr', problem_type=self.problem_type, objective_func=self.objective_func,
                 hyperparameters={'penalty': 'L2'},
             ),
            LinearModel(
                path=self.path, name='lr', problem_type=self.problem_type, objective_func=self.objective_func,
                hyperparameters={'handle_text': 'include'},
        ]
        return models

predictor = task.fit(train_data=train_data, label=label_column, output_directory=output_directory,
                     eval_metric=eval_metric, hyperparameter_tune=False, verbosity=3,
                     trainer_type = LinearRegrssionTrainer, 
reduce runtime
                    )
```

### DIFF
--- a/autogluon/task/tabular_prediction/tabular_prediction.py
+++ b/autogluon/task/tabular_prediction/tabular_prediction.py
@@ -180,7 +180,7 @@ class TabularPrediction(BaseTask):
                     Note: Hyperparameter tuning is disabled for this model.
                     Note: 'weights' parameter will be overridden. Both 'distance' and 'uniform' are used automatically, training two models.
                 LR: `autogluon/utils/tabular/ml/models/lr/hyperparameters/parameters.py`
-                    Note: a list of hyper-parameters should be passed; each set will create differen version of the model.
+                    Note: a list of hyper-parameters should be passed; each set will create different version of the model.
                     Note: Hyperparameter tuning is disabled for this model.
                     Note: 'penalty' parameter can be used for regression to specify regularization method: 'L1' and 'L2' values are supported.
 
--- a/autogluon/task/tabular_prediction/tabular_prediction.py
+++ b/autogluon/task/tabular_prediction/tabular_prediction.py
@@ -337,7 +337,6 @@ class TabularPrediction(BaseTask):
                'CAT': {'iterations': 10000},
                'RF': {'n_estimators': 300},
                'XT': {'n_estimators': 300},
-               'LR': {'C': 1},
                'KNN': {},
                'custom': ['GBM'],
                 'LR': {'C': 1},
--- a/autogluon/task/tabular_prediction/tabular_prediction.py
+++ b/autogluon/task/tabular_prediction/tabular_prediction.py
@@ -337,10 +337,10 @@ class TabularPrediction(BaseTask):
                'CAT': {'iterations': 10000},
                'RF': {'n_estimators': 300},
                'XT': {'n_estimators': 300},
-               'LR': {'C': 1},
                'KNN': {},
                'custom': ['GBM'],
-             }
+                'LR': {'C': 1},
+            }
 
         # Process kwargs to create feature generator, trainer, schedulers, searchers for each model:
         output_directory = setup_outputdir(output_directory)  # Format directory name
--- a/autogluon/task/tabular_prediction/tabular_prediction.py
+++ b/autogluon/task/tabular_prediction/tabular_prediction.py
@@ -337,6 +337,7 @@ class TabularPrediction(BaseTask):
                'CAT': {'iterations': 10000},
                'RF': {'n_estimators': 300},
                'XT': {'n_estimators': 300},
+               'LR': {'C': 1},
                'KNN': {},
                'custom': ['GBM'],
              }
--- a/autogluon/task/tabular_prediction/tabular_prediction.py
+++ b/autogluon/task/tabular_prediction/tabular_prediction.py
@@ -339,7 +339,7 @@ class TabularPrediction(BaseTask):
                'XT': {'n_estimators': 300},
                'KNN': {},
                'custom': ['GBM'],
-                'LR': {'C': 1},
+                # 'LR': {'C': 1}, # Enable linear models
             }
 
         # Process kwargs to create feature generator, trainer, schedulers, searchers for each model:
--- a/autogluon/task/tabular_prediction/tabular_prediction.py
+++ b/autogluon/task/tabular_prediction/tabular_prediction.py
@@ -179,6 +179,10 @@ class TabularPrediction(BaseTask):
                 KNN: See sklearn documentation: https://scikit-learn.org/stable/modules/generated/sklearn.neighbors.KNeighborsClassifier.html
                     Note: Hyperparameter tuning is disabled for this model.
                     Note: 'weights' parameter will be overridden. Both 'distance' and 'uniform' are used automatically, training two models.
+                LR: `autogluon/utils/tabular/ml/models/lr/hyperparameters/parameters.py`
+                    Note: a list of hyper-parameters should be passed; each set will create differen version of the model.
+                    Note: Hyperparameter tuning is disabled for this model.
+                    Note: 'penalty' parameter can be used for regression to specify regularization method: 'L1' and 'L2' values are supported.
 
         holdout_frac : float
             Fraction of train_data to holdout as tuning data for optimizing hyperparameters (ignored unless `tuning_data = None`, ignored if `num_bagging_folds != 0`).
--- a/autogluon/task/tabular_prediction/tabular_prediction.py
+++ b/autogluon/task/tabular_prediction/tabular_prediction.py
@@ -339,7 +339,6 @@ class TabularPrediction(BaseTask):
                'XT': {'n_estimators': 300},
                'KNN': {},
                'custom': ['GBM'],
-                # 'LR': {'C': 1}, # Enable linear models
             }
 
         # Process kwargs to create feature generator, trainer, schedulers, searchers for each model:
--- a/autogluon/task/tabular_prediction/tabular_prediction.py
+++ b/autogluon/task/tabular_prediction/tabular_prediction.py
@@ -180,7 +180,7 @@ class TabularPrediction(BaseTask):
                     Note: Hyperparameter tuning is disabled for this model.
                     Note: 'weights' parameter will be overridden. Both 'distance' and 'uniform' are used automatically, training two models.
                 LR: `autogluon/utils/tabular/ml/models/lr/hyperparameters/parameters.py`
-                    Note: a list of hyper-parameters should be passed; each set will create different version of the model.
+                    Note: a list of hyper-parameters dicts can be passed; each set will create different version of the model.
                     Note: Hyperparameter tuning is disabled for this model.
                     Note: 'penalty' parameter can be used for regression to specify regularization method: 'L1' and 'L2' values are supported.
 
--- a/autogluon/task/tabular_prediction/tabular_prediction.py
+++ b/autogluon/task/tabular_prediction/tabular_prediction.py
@@ -337,6 +337,7 @@ class TabularPrediction(BaseTask):
                'CAT': {'iterations': 10000},
                'RF': {'n_estimators': 300},
                'XT': {'n_estimators': 300},
+               'LR': {'C': 1},
                'KNN': {},
                'custom': ['GBM'],
                 'LR': {'C': 1},
--- a/autogluon/utils/tabular/ml/models/lr/lr_model.py
+++ b/autogluon/utils/tabular/ml/models/lr/lr_model.py
@@ -16,7 +16,7 @@ from autogluon.utils.tabular.ml.models.lr.lr_preprocessing_utils import NlpDataP
 logger = logging.getLogger(__name__)
 
 
-class LRModel(AbstractModel):
+class AbstractLinearModel(AbstractModel):
 
     def __init__(self, path: str, name: str, problem_type: str, objective_func, num_classes=None, hyperparameters=None, features=None,
                  feature_types_metadata=None, debug=0):
@@ -82,44 +82,41 @@ class LRModel(AbstractModel):
         X = X.copy()
         feature_types = self._get_types_of_features(X)
         if is_train:
-            transformer_list = []
-
-            if len(feature_types['language']) > 0:
-                pipeline = Pipeline(steps=[
-                    ("preparator", NlpDataPreprocessor(nlp_cols=feature_types['language'])),
-                    ("vectorizer",
-                     TfidfVectorizer(ngram_range=self.params['proc.ngram_range'], sublinear_tf=True, max_features=vect_max_features, tokenizer=self.tokenize))
-                ])
-                transformer_list.append(('vect', pipeline))
-
-            if len(feature_types['onehot']) > 0:
-                pipeline = Pipeline(steps=[
-                    ('generator', OheFeaturesGenerator(cats_cols=feature_types['onehot'])),
-                ])
-                transformer_list.append(('cats', pipeline))
-
-            if len(feature_types['continuous']) > 0:
-                pipeline = Pipeline(steps=[
-                    ('generator', NumericDataPreprocessor(cont_cols=feature_types['continuous'])),
-                    ('imputer', SimpleImputer(strategy=self.params['proc.impute_strategy'])),
-                    ('scaler', StandardScaler())
-                ])
-                transformer_list.append(('cont', pipeline))
-
-            if len(feature_types['skewed']) > 0:
-                pipeline = Pipeline(steps=[
-                    ('generator', NumericDataPreprocessor(cont_cols=feature_types['skewed'])),
-                    ('imputer', SimpleImputer(strategy=self.params['proc.impute_strategy'])),
-                    ('quantile', QuantileTransformer(output_distribution='normal')),  # Or output_distribution = 'uniform'
-                ])
-                transformer_list.append(('skew', pipeline))
-
-            self.pipeline = FeatureUnion(transformer_list=transformer_list)
-            self.pipeline.fit(X)
-
+            self.preprocess_train(X, feature_types, vect_max_features)
         X = self.pipeline.transform(X)
 
         return X
+
+    def preprocess_train(self, X, feature_types, vect_max_features):
+        transformer_list = []
+        if len(feature_types['language']) > 0:
+            pipeline = Pipeline(steps=[
+                ("preparator", NlpDataPreprocessor(nlp_cols=feature_types['language'])),
+                ("vectorizer",
+                 TfidfVectorizer(ngram_range=self.params['proc.ngram_range'], sublinear_tf=True, max_features=vect_max_features, tokenizer=self.tokenize))
+            ])
+            transformer_list.append(('vect', pipeline))
+        if len(feature_types['onehot']) > 0:
+            pipeline = Pipeline(steps=[
+                ('generator', OheFeaturesGenerator(cats_cols=feature_types['onehot'])),
+            ])
+            transformer_list.append(('cats', pipeline))
+        if len(feature_types['continuous']) > 0:
+            pipeline = Pipeline(steps=[
+                ('generator', NumericDataPreprocessor(cont_cols=feature_types['continuous'])),
+                ('imputer', SimpleImputer(strategy=self.params['proc.impute_strategy'])),
+                ('scaler', StandardScaler())
+            ])
+            transformer_list.append(('cont', pipeline))
+        if len(feature_types['skewed']) > 0:
+            pipeline = Pipeline(steps=[
+                ('generator', NumericDataPreprocessor(cont_cols=feature_types['skewed'])),
+                ('imputer', SimpleImputer(strategy=self.params['proc.impute_strategy'])),
+                ('quantile', QuantileTransformer(output_distribution='normal')),  # Or output_distribution = 'uniform'
+            ])
+            transformer_list.append(('skew', pipeline))
+        self.pipeline = FeatureUnion(transformer_list=transformer_list)
+        self.pipeline.fit(X)
 
     def _set_default_params(self):
         default_params = {
@@ -163,7 +160,7 @@ class LRModel(AbstractModel):
         return hpo_models, hpo_model_performances, hpo_results
 
 
-class LRModelAll(LRModel):
+class LinearModel(AbstractLinearModel):
     def _select_features(self, df, types_of_features, categorical_featnames, language_featnames, continuous_featnames):
         # continuous = numeric features to rescale
         # skewed = features to which we will apply power (ie. log / box-cox) transform before normalization
@@ -184,7 +181,7 @@ class LRModelAll(LRModel):
         return types_of_features
 
 
-class LRModelNLP(LRModel):
+class LinearModelOnlyTextFeatures(AbstractLinearModel):
     def _select_features(self, df, types_of_features, categorical_featnames, language_featnames, continuous_featnames):
         for feature in self.features:
             if feature in language_featnames:
@@ -192,7 +189,7 @@ class LRModelNLP(LRModel):
         return types_of_features
 
 
-class LRModelPlain(LRModel):
+class LinearModelNoTextFeatures(AbstractLinearModel):
     def _select_features(self, df, types_of_features, categorical_featnames, language_featnames, continuous_featnames):
         # continuous = numeric features to rescale
         # skewed = features to which we will apply power (ie. log / box-cox) transform before normalization
--- a/autogluon/utils/tabular/ml/models/lr/lr_model.py
+++ b/autogluon/utils/tabular/ml/models/lr/lr_model.py
@@ -159,6 +159,12 @@ class AbstractLinearModel(AbstractModel):
 
         return hpo_models, hpo_model_performances, hpo_results
 
+    def get_info(self):
+        # TODO: All AG-Tabular models now offer a get_info method:
+        # https://github.com/awslabs/autogluon/blob/master/autogluon/utils/tabular/ml/models/abstract/abstract_model.py#L474
+        # dict of weights?
+        return super().get_info()
+
 
 class LinearModel(AbstractLinearModel):
     def _select_features(self, df, types_of_features, categorical_featnames, language_featnames, continuous_featnames):
--- a/autogluon/utils/tabular/ml/models/lr/lr_model.py
+++ b/autogluon/utils/tabular/ml/models/lr/lr_model.py
@@ -1,0 +1,108 @@
+import re
+import time
+
+from pandas import DataFrame
+from sklearn.feature_extraction.text import TfidfVectorizer
+from sklearn.linear_model import LogisticRegression
+from sklearn.pipeline import FeatureUnion, Pipeline
+from sklearn.preprocessing import StandardScaler
+
+from autogluon.utils.tabular.ml.constants import BINARY, MULTICLASS
+from autogluon.utils.tabular.ml.models.abstract.abstract_model import AbstractModel
+from autogluon.utils.tabular.ml.models.lr.lr_preprocessing_utils import NlpDataPreprocessor, OheFeaturesGenerator, get_one_hot_features, NumericDataPreprocessor
+
+
+class LRModel(AbstractModel):
+
+    def __init__(self, path: str, name: str, problem_type: str, objective_func, num_classes=None, hyperparameters=None, features=None,
+                 feature_types_metadata=None, debug=0):
+        super().__init__(path=path, name=name, problem_type=problem_type, objective_func=objective_func, hyperparameters=hyperparameters, features=features,
+                         feature_types_metadata=feature_types_metadata, debug=debug)
+        self.pipeline = None
+        self.cat_one_hot = None
+
+    def tokenize(self, s):
+        return re.split('[ ]+', s)
+
+    def preprocess(self, X: DataFrame, is_train=False, vect_max_features=1000):
+        X = X.copy()
+        nlp_cols = self.feature_types_metadata['nlp']
+        cats = self.feature_types_metadata['object']
+        cats = [c for c in cats if c in self.cat_one_hot]
+
+        cont = self.feature_types_metadata['int'] + self.feature_types_metadata['float']
+        cont = [c for c in cont if '__nlp__' not in c]
+
+        X[cont] = X[cont].fillna(0)
+
+        # TODO: reuse the code from NNs - imputation and quantile transform
+        if is_train:
+            vectorizer_pipeline = Pipeline(steps=[
+                ("preparator", NlpDataPreprocessor(nlp_cols=nlp_cols)),
+                ("vectorizer", TfidfVectorizer(ngram_range=(1, 3), sublinear_tf=True, max_features=vect_max_features, tokenizer=self.tokenize))
+            ])
+
+            cats_pipeline = Pipeline(steps=[
+                ('generator', OheFeaturesGenerator(cats_cols=cats)),
+            ])
+
+            numeric_feats_pipeline = Pipeline(steps=[
+                ('generator', NumericDataPreprocessor(cont_cols=cont)),
+                ('scaler', StandardScaler())
+            ])
+
+            self.pipeline = FeatureUnion(transformer_list=[
+                ('cats', cats_pipeline),
+                ('vect', vectorizer_pipeline),
+                ('cont', numeric_feats_pipeline),
+            ])
+            self.pipeline.fit(X)
+
+        X = self.pipeline.transform(X)
+        return X
+
+    def _set_default_params(self):
+        default_params = {
+            'model_type': 'LR',
+            'C': 1,
+            'vectorizer_dict_size': 3000,
+        }
+        for param, val in default_params.items():
+            self._set_default_param_value(param, val)
+
+    def _get_default_searchspace(self, problem_type):
+        spaces = {
+            # 'C': Real(lower=1e-4, upper=1e5, default=1),
+            # 'tokenizer': Categorical('split', 'sentencepiece')
+        }
+        return spaces
+
+    def fit(self, X_train, Y_train, X_test=None, Y_test=None, time_limit=None, **kwargs):
+        time_start = time.time()
+        hyperparams = self.params.copy()
+
+        self.cat_one_hot = get_one_hot_features(X_train)
+
+        # See solver options here: https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LogisticRegression.html
+        if self.problem_type == BINARY:
+            Y_train = Y_train.astype(int).values
+            solver = 'liblinear'
+        elif self.problem_type == MULTICLASS:
+            solver = 'saga'  # another option is lbfgs
+        else:
+            solver = 'liblinear'
+
+        X_train = self.preprocess(X_train, is_train=True, vect_max_features=hyperparams['vectorizer_dict_size'])
+
+        # TODO: get seed from seeds provider
+        self.model = LogisticRegression(C=hyperparams['C'], random_state=17, solver=solver)
+        self.model.fit(X_train, Y_train)
+
+    def hyperparameter_tune(self, X_train, X_test, Y_train, Y_test, scheduler_options=None, **kwargs):
+        self.fit(X_train=X_train, X_test=X_test, Y_train=Y_train, Y_test=Y_test, **kwargs)
+        hpo_model_performances = {self.name: self.score(X_test, Y_test)}
+        hpo_results = {}
+        self.save()
+        hpo_models = {self.name: self.path}
+
+        return hpo_models, hpo_model_performances, hpo_results
--- a/autogluon/utils/tabular/ml/models/lr/lr_model.py
+++ b/autogluon/utils/tabular/ml/models/lr/lr_model.py
@@ -49,23 +49,10 @@ class LRModel(AbstractModel):
             self.features = list(df.columns)
 
         types_of_features = {'continuous': [], 'skewed': [], 'onehot': [], 'language': []}
-        # continuous = numeric features to rescale
-        # skewed = features to which we will apply power (ie. log / box-cox) transform before normalization
-        # onehot = features to one-hot encode (unknown categories for these features encountered at test-time are encoded as all zeros). We one-hot encode any features encountered that only have two unique values.
-        one_hot_threshold = 2 if len(language_featnames) > 0 else 10000
-        for feature in self.features:
-            feature_data = df[feature]
-            num_unique_vals = len(feature_data.unique())
-            if feature in language_featnames:
-                types_of_features['language'].append(feature)
-            elif feature in continuous_featnames:
-                if np.abs(feature_data.skew()) > self.params['proc.skew_threshold']:
-                    types_of_features['skewed'].append(feature)
-                else:
-                    types_of_features['continuous'].append(feature)
-            elif (feature in categorical_featnames) and (num_unique_vals <= one_hot_threshold):
-                types_of_features['onehot'].append(feature)
-        return types_of_features
+        return self._select_features(df, types_of_features, categorical_featnames, language_featnames, continuous_featnames)
+
+    def _select_features(self, df, types_of_features, categorical_featnames, language_featnames, continuous_featnames):
+        raise NotImplementedError()
 
     def __get_feature_type_if_present(self, feature_type):
         """ Returns crude categorization of feature types """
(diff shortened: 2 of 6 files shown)


szha:
Job PR-375-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-375/1/index.html

autogluon/utils/tabular/ml/models/lr/lr_model.py
```diff
@@ -0,0 +1,206 @@
...
+    def _select_features(self, df, types_of_features, categorical_featnames, language_featnames, continuous_featnames):
+        raise NotImplementedError()
+
+    def __get_feature_type_if_present(self, feature_type):
+        """ Returns crude categorization of feature types """
+        return self.feature_types_metadata[feature_type] if feature_type in self.feature_types_metadata else []
+
+    # TODO: handle collinear features - they will impact results quality
+    def preprocess(self, X: DataFrame, is_train=False, vect_max_features=1000):
+        X = X.copy()
```
Innixma: Should this be replaced with:

`X = super().preprocess(X)`

?

The code crashes for me unless this line is added because self.features is None

autogluon/utils/tabular/ml/models/lr/lr_model.py
```diff
@@ -0,0 +1,206 @@
...
+        feature_types = self._get_types_of_features(X)
+        logger.log(15, "Applying model-specific pre-processing")
```
Innixma: I'd recommend removing logging statements from preprocessing functions, this gets called a lot and during inference can be expensive to log.

autogluon/utils/tabular/ml/trainer/model_presets/presets.py
```diff
@@ -87,6 +89,10 @@ def get_preset_models_classification(path, problem_type, objective_func, stoppin
             TabularNeuralNetModel(path=path, name='NeuralNetClassifier', problem_type=problem_type,
                                   objective_func=objective_func, stopping_metric=stopping_metric, hyperparameters=nn_options.copy()),
         )
+    if lr_options is not None:
```
Innixma: Is there a reason this wasn't added for regression as well?

autogluon/utils/tabular/ml/models/lr/lr_model.py
```diff
@@ -0,0 +1,206 @@
...
+            solver = 'lbfgs'  # TODO use liblinear for smaller datasets
+        elif self.problem_type == MULTICLASS:
+            solver = 'saga'  # another option is lbfgs
+        else:
+            solver = 'lbfgs'
+
+        X_train = self.preprocess(X_train, is_train=True, vect_max_features=hyperparams['vectorizer_dict_size'])
+
+        # TODO: get seed from seeds provider
+        self.model = LogisticRegression(C=hyperparams['C'], random_state=17, solver=solver, n_jobs=-1)
```
Innixma: All random states should be 0 in AG unless they are a variable, just for consistency.

autogluon/utils/tabular/ml/models/lr/lr_model.py
```diff
@@ -0,0 +1,206 @@
...
+            self._set_default_param_value(param, val)
+
+    def _get_default_searchspace(self, problem_type):
+        spaces = {
+            # 'C': Real(lower=1e-4, upper=1e5, default=1),
+            # 'tokenizer': Categorical('split', 'sentencepiece')
+        }
+        return spaces
+
+    def fit(self, X_train, Y_train, X_test=None, Y_test=None, time_limit=None, **kwargs):
```
Innixma: Remove X_train, Y_train, and time_limit args, since they are never used.

autogluon/utils/tabular/ml/models/lr/lr_model.py
```diff
@@ -0,0 +1,206 @@
...
+        hyperparams = self.params.copy()
+
+        self.cat_one_hot = get_one_hot_features(X_train)
```
Innixma: This is never used

autogluon/utils/tabular/ml/models/lr/lr_model.py
```diff
@@ -0,0 +1,206 @@
...
+
+        # See solver options here: https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LogisticRegression.html
+        if self.problem_type == BINARY:
```
Innixma: solver should be set during init, not during fit

autogluon/utils/tabular/ml/models/lr/lr_model.py
Innixma: Only set self.model after .fit(), .fit() should return self

self.model = model.fit(X_train, Y_train)

Example is in KNN

autogluon/utils/tabular/ml/models/lr/lr_model.py
Innixma: Can you feasibly think of a scenario where this model results in OOM? Is there a good way to check for that scenario?

autogluon/utils/tabular/ml/models/lr/lr_model.py
```diff
@@ -0,0 +1,206 @@
...
+
+        return hpo_models, hpo_model_performances, hpo_results
+
+
+class LRModelAll(LRModel):
+    def _select_features(self, df, types_of_features, categorical_featnames, language_featnames, continuous_featnames):
+        # continuous = numeric features to rescale
+        # skewed = features to which we will apply power (ie. log / box-cox) transform before normalization
+        # onehot = features to one-hot encode (unknown categories for these features encountered at test-time are encoded as all zeros). We one-hot encode any features encountered that only have two unique values.
+        one_hot_threshold = 2 if len(language_featnames) > 0 else 10000
```
Innixma: Why? Won't this affect categorical non-language features?

autogluon/utils/tabular/ml/models/lr/lr_model.py
```diff
@@ -0,0 +1,206 @@
...
+            feature_data = df[feature]
+            num_unique_vals = len(feature_data.unique())
+            if feature in language_featnames:
+                types_of_features['language'].append(feature)
+            elif feature in continuous_featnames:
+                if np.abs(feature_data.skew()) > self.params['proc.skew_threshold']:
+                    types_of_features['skewed'].append(feature)
+                else:
+                    types_of_features['continuous'].append(feature)
+            elif (feature in categorical_featnames) and (num_unique_vals <= one_hot_threshold):
```
Innixma: else?

autogluon/utils/tabular/ml/models/lr/lr_model.py
```diff
@@ -0,0 +1,206 @@
...
+                types_of_features['language'].append(feature)
+        return types_of_features
+
+
+class LRModelPlain(LRModel):
+    def _select_features(self, df, types_of_features, categorical_featnames, language_featnames, continuous_featnames):
+        # continuous = numeric features to rescale
+        # skewed = features to which we will apply power (ie. log / box-cox) transform before normalization
+        # onehot = features to one-hot encode (unknown categories for these features encountered at test-time are encoded as all zeros). We one-hot encode any features encountered that only have two unique values.
+        one_hot_threshold = 2 if len(language_featnames) > 0 else 10000
```
Innixma: ditto

autogluon/utils/tabular/ml/models/lr/lr_model.py
```diff
@@ -0,0 +1,206 @@
...
+
+        logger.log(15, " - transforming inputs using pre-processing pipeline")
+        X = self.pipeline.transform(X)
+        logger.log(15, " - output shape %s" % str(X.shape))
+
+        return X
+
+    def _set_default_params(self):
+        default_params = {
+            'model_type': 'LR',
```
Innixma: Why is this a param?

autogluon/utils/tabular/ml/models/lr/lr_model.py
```diff
@@ -0,0 +1,206 @@
...
+        """ Returns crude categorization of feature types """
+        return self.feature_types_metadata[feature_type] if feature_type in self.feature_types_metadata else []
+
+    # TODO: handle collinear features - they will impact results quality
+    def preprocess(self, X: DataFrame, is_train=False, vect_max_features=1000):
+        X = X.copy()
+        feature_types = self._get_types_of_features(X)
+        logger.log(15, "Applying model-specific pre-processing")
+        logger.log(15, " - input shape %s" % str(X.shape))
+        if is_train:
```
Innixma: Can you move this to a preprocess_train() function? Longer term I want to consolidate this logic for stateful preprocessors and having it separated will simplify things.

autogluon/utils/tabular/ml/models/lr/lr_model.py
```diff
@@ -0,0 +1,206 @@
...
+    def tokenize(self, s):
+        return re.split('[ ]+', s)
+
+    def _get_types_of_features(self, df):
+        """ Returns dict with keys: : 'continuous', 'skewed', 'onehot', 'embed', 'language', values = ordered list of feature-names falling into each category.
+            Each value is a list of feature-names corresponding to columns in original dataframe.
+            TODO: ensure features with zero variance have already been removed before this function is called.
+        """
+        if self.types_of_features is not None:
+            Warning("Attempting to _get_types_of_features for TabularNeuralNetModel, but previously already did this.")
```
Innixma: log.warning?

autogluon/utils/tabular/ml/models/lr/lr_model.py
Innixma: Also: Is there any feasible way to track time_limit during fit or estimate the time it will take prior? If the model is always quick then it shouldn't matter much.

jwmueller:
Since these model names will be user facing, can we make them a bit more informative?

Something like:

LRModelPlain -> LinearModelNoTextFeatures
LRModelNLP -> LinearModelOnlyTextFeatures
LRModelAll -> LinearModel

As the names get displayed in various plots/summaries like `fit_summary`/`leaderboard`, they should be more interpretable in my opinion.

Also is there a reason why `LRModelAll` is not being used as the default model in presets instead of `LRModelPlain`? Is it because the text was already featurized? Wouldn't these two be equivalent in that case?

autogluon/utils/tabular/ml/trainer/model_presets/presets.py
jwmueller: More generally, I didn't see `sklearn.linear_model.LinearRegression` anywhere?  Is this current PR only covering classification use-cases for now?

jwmueller:
Something that can be saved for future PR: 

All AG-Tabular models now offer a `get_info` method:  
https://github.com/awslabs/autogluon/blob/master/autogluon/utils/tabular/ml/models/abstract/abstract_model.py#L474

Since a major reason to use linear models is interpretability, it would be nice if this `get_info` method can display their weights as a pandas DataFrame.  In particular, this DF could have row-indices = feature-names, a column called "weight_raw" with the raw weight (as produced by sklearn's `coef_` which unfortunately can be misleading due to feature-scaling), a column called "weight_normalized" (which normalizes the weights based on feature-scaling for better interpretability, i.e. correlation), and possibly another column "pvalue" that provides statistical significance that this weight is non-zero.

Innixma:
Regarding get_info, we probably don't want it to be a DataFrame. GetInfo is compatible with json currently, and it would be the only model which has a DataFrame as part of its info. Why not just a dict of weights?

jwmueller:
> Regarding get_info, we probably don't want it to be a DataFrame. GetInfo is compatible with json currently, and it would be the only model which has a DataFrame as part of its info. Why not just a dict of weights?

Dict is fine as well for json compatibility.  But the raw weights may be misleading for ML beginners, since a tiny-variance feature can have large weight irrespective of its relationship with the target.  I guess this may be ok though since in our case, the `coef_` from sklearn will generally be invoked only for normalized data (post-processing), although TF-IDF features may still be unnormalized...

autogluon/task/tabular_prediction/tabular_prediction.py
```diff
@@ -295,7 +295,8 @@ def fit(train_data, label, tuning_data=None, output_directory=None, problem_type
                'XT': {'n_estimators': 300},
                'KNN': {},
                'custom': ['GBM'],
-             }
+                # 'LR': {'C': 1}, # Enable linear models
```
jwmueller: can you update fit() documentation with 'LR' option, similar to the other sklearn models.

autogluon/utils/tabular/ml/models/lr/lr_model.py
jwmueller: It could be possible to adaptively set `max_iter`:  https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LogisticRegression.html#examples-using-sklearn-linear-model-logisticregression

to approximately respect `time_limit` based on sample-size, feature-dimensionality, and the `solver` used.

jwmueller: To better respect `time_limit`, can run at first with `max_iter=1` and then save, measuring the time required for one iteration (or a couple).  Can subsequently set `warm_start = True` and run with appropriately-set `max_iter`

https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LogisticRegression.html#examples-using-sklearn-linear-model-logisticregression

One difficulty here is standardizing things between regression & classification though, I'm not sure there are analogous functionalities for regression linear models in sklearn...

autogluon/utils/tabular/ml/models/lr/lr_preprocessing_utils.py
```diff
@@ -0,0 +1,74 @@
+from scipy.sparse import hstack
+from sklearn.base import BaseEstimator, TransformerMixin
+from sklearn.preprocessing import OneHotEncoder
+
+
+def get_one_hot_features(df):
+    ohe_cats = df.nunique()
+    ohe_cats = ohe_cats[ohe_cats < 50].index.values.tolist()
```
jwmueller: should 50 be a tracked argument here? perhaps eventually something to search over?

autogluon/utils/tabular/ml/trainer/model_presets/presets.py
jwmueller: One difficulty I think is sklearn offers little linear regressor functionality that is analogous to the args of its `LogisticRegression` class

autogluon/utils/tabular/ml/models/lr/lr_model.py
Innixma: Probably not necessary for first version but definitely something to keep in mind

autogluon/utils/tabular/ml/models/lr/lr_model.py
sgdread: Removed

autogluon/utils/tabular/ml/models/lr/lr_model.py
sgdread: Update to 0

autogluon/utils/tabular/ml/models/lr/lr_model.py
sgdread: Added TODO for this version

autogluon/utils/tabular/ml/models/lr/lr_model.py
sgdread: Removed

autogluon/utils/tabular/ml/models/lr/lr_model.py
sgdread: moved to init

autogluon/utils/tabular/ml/models/lr/lr_model.py
sgdread: Done

autogluon/utils/tabular/ml/models/lr/lr_model.py
sgdread: fixed

szha:
Job PR-375-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-375/2/index.html

autogluon/utils/tabular/ml/models/lr/lr_model.py
sgdread: The crash is unrelated to this call. It was related due to bug in other place. Fixed now.

autogluon/utils/tabular/ml/trainer/model_presets/presets.py
sgdread: Good catch. Added Linear regression for REGRESSION problem type.

autogluon/utils/tabular/ml/models/lr/lr_model.py
sgdread: It was creating memory issues when joined with TFIDF vectors from NLP features. Basically currently it's switching between 2 modes: NLP and Plain. Not the ideal solution, needs deep dive to find proper resolution.

autogluon/utils/tabular/ml/models/lr/lr_model.py
sgdread: removed

autogluon/utils/tabular/ml/models/lr/lr_model.py
sgdread: Done

autogluon/utils/tabular/ml/models/lr/lr_preprocessing_utils.py
sgdread: This method was removed - remainings of early prototype code.

autogluon/task/tabular_prediction/tabular_prediction.py
sgdread: I removed the line for now.

sgdread:
> Since these model names will be user facing, can we make them a bit more informative?

Renamed

szha:
Job PR-375-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-375/3/index.html

szha:
Job PR-375-4 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-375/4/index.html

szha:
Job PR-375-5 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-375/5/index.html

szha:
Job PR-375-6 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-375/6/index.html

szha:
Job PR-375-7 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-375/7/index.html

autogluon/utils/tabular/ml/trainer/model_presets/presets.py
```diff
@@ -87,6 +89,10 @@ def get_preset_models_classification(path, problem_type, objective_func, stoppin
...
+        models.append(
+            LinearModelNoTextFeatures(path=path, name='LinearModelNoTextFeatures', problem_type=problem_type, objective_func=objective_func, hyperparameters=lr_options.copy())
```
jwmueller, line 94: Why is this not `LinearModel` as the default?

Moreover, should the 'LR' hyperparams dict include a key 'handle_text' with options: 'include' (use `LinearModel`), 'ignore' (use `LinearModelNoText`), 'only' (use `LinearModelOnlyText`)?

Otherwise it seems the user cannot easily access these without delving into undocumented functionalities like Trainer.

autogluon/utils/tabular/ml/models/lr/lr_model.py
```diff
@@ -0,0 +1,242 @@
...
+                types_of_features['language'].append(feature)
+        return types_of_features
+
+
+class LinearModelNoTextFeatures(AbstractLinearModel):
+    def _select_features(self, df, types_of_features, categorical_featnames, language_featnames, continuous_featnames):
+        # continuous = numeric features to rescale
+        # skewed = features to which we will apply power (ie. log / box-cox) transform before normalization
+        # onehot = features to one-hot encode (unknown categories for these features encountered at test-time are encoded as all zeros). We one-hot encode any features encountered that only have two unique values.
+        one_hot_threshold = 2 if len(language_featnames) > 0 else 10000
```
jwmueller, line 229: Does this code even matter if there are no text features?

autogluon/utils/tabular/ml/models/lr/lr_model.py
```diff
@@ -0,0 +1,242 @@
...
+        self.pipeline = FeatureUnion(transformer_list=transformer_list)
+        self.pipeline.fit(X)
+
+    def _set_default_params(self):
+        default_params = {
+            'C': 1,
+            'vectorizer_dict_size': 75000,
+            'proc.ngram_range': (1, 5),
+            'proc.skew_threshold': 0.99,
+            'proc.impute_strategy': 'median',  # strategy argument of sklearn.SimpleImputer() used to impute missing numeric values
```
jwmueller, line 145: Each of these hyperparameters should be explained with comments.

Also could you put these in a separate file located in a hyperparameters/ subdirectory like we have for the other (non-sklearn) models?  This would make it more standardized, and a lot easier for users to quickly view the  hyperparameters they can specify.

Should 'regression_option' be another hyperparameter in the 'LR' dict? Otherwise how can a user choose between 'ridge', 'lasso' without delving into undocumented Trainer code?  Also consider standardizing this hyperparameter with the 'penalty' option in LogisticRegression which allows you to choose between 'l1', 'l2'. I would call it something more informative like 'penalty' instead of 'regression_option' (and you can just appropriately translate the 'l1' & 'l2' options into Lasso & Ridge internally).

Finally, what are your thoughts of exposing additional hyperparameters that can be shared

autogluon/utils/tabular/ml/models/lr/lr_preprocessing_utils.py
```diff
@@ -0,0 +1,67 @@
...
+class OheFeaturesGenerator(BaseEstimator, TransformerMixin):
+
+    def __init__(self, cats_cols):
+        self._feature_names = []
+        self.cats = cats_cols
+        self.ohe_encs = None
+        self.labels = None
+
+    def fit(self, X, y=None):
+        self.ohe_encs = {f: OneHotEncoder(handle_unknown='ignore') for f in self.cats}
```
jwmueller, line 15: Why not use the OneHotMergeRaresHandleUnknownEncoder:
https://github.com/awslabs/autogluon/blob/3f2e3b3e46cc10e01017257b46e8f5992cbc6c00/autogluon/utils/tabular/ml/models/tabular_nn/categorical_encoders.py#L321

You can see how it's used for the neural net here:
https://github.com/awslabs/autogluon/blob/79d2f7d60a2eebf0f9aaf1d43186b583f81e703a/autogluon/utils/tabular/ml/models/tabular_nn/tabular_nn_model.py#L689
(it can be be a drop-in replacement for the sklearn OneHotEncoder as it offers the same functionality).  It has the big advantage that it can prevent feature explosion with high-cardinality categorical features (it uses an additional `max_categories` hyperparameter and merges all the rarest categories beyond this value).

autogluon/utils/tabular/ml/models/lr/lr_model.py
sgdread: * Added comments
* Moved-out hyperparameters and searchspaces
* `regression_option` renamed into `penalty` passed as a hyper-parameter with options: `L1`/`L2`
* fit_intercept/solver/random_state can be passed through hyper-parameters already.

autogluon/utils/tabular/ml/trainer/model_presets/presets.py
sgdread: * Switched to single `LinearModel` class with `handle_text` hyperparameter option.

autogluon/utils/tabular/ml/models/lr/lr_model.py
sgdread: Updated

autogluon/utils/tabular/ml/models/lr/lr_preprocessing_utils.py
sgdread: I tried it, but it was causing memory issues on few datasets.

autogluon/utils/tabular/ml/models/lr/lr_model.py
```diff
@@ -0,0 +1,242 @@
...
+    def __init__(self, path: str, name: str, problem_type: str, objective_func, num_classes=None, hyperparameters=None, features=None,
+                 feature_types_metadata=None, debug=0, regression_option='ridge', **kwargs):
+        super().__init__(path=path, name=name, problem_type=problem_type, objective_func=objective_func, hyperparameters=hyperparameters, features=features,
+                         feature_types_metadata=feature_types_metadata, debug=debug)
+        self.types_of_features = None
+        self.regression_option = regression_option
+        self.pipeline = None
+
+        if self.problem_type == REGRESSION:
+            self._get_regression_model(kwargs)
```
Innixma, line 30: **kwargs?

autogluon/utils/tabular/ml/models/lr/lr_model.py
```diff
@@ -0,0 +1,242 @@
...
+        else:
+            self._model_type = LogisticRegression
+
+        self.model_params = None
+        self.set_default_params()
+
+    def _get_regression_model(self):
```
Innixma, line 37: This can't accept kwargs currently, but it is called with kwargs as a parameter in __init__

autogluon/utils/tabular/ml/trainer/model_presets/presets.py
Innixma, line 92: Regression is still not added, it can't be specified as a model in a regression problem. It has to be added in get_preset_models_regression()

autogluon/utils/tabular/ml/models/lr/lr_model.py
sgdread: fixed in the new version.

autogluon/utils/tabular/ml/trainer/model_presets/presets.py
```diff
@@ -157,6 +158,10 @@ def get_preset_models_regression(path, problem_type, objective_func, stopping_me
         if 'GBM' in custom_options:
             models += [LGBModel(path=path, name='LightGBMRegressorCustom', problem_type=problem_type, objective_func=objective_func, stopping_metric=stopping_metric, hyperparameters=get_param_baseline_custom(problem_type))]
         # SKLearnModel(path=path, name='DummyRegressor', model=DummyRegressor(), problem_type=problem_type, objective_func=objective_func),
+    if lr_options is not None:
```
Innixma, line 161: Order matters in the presets, the order it is appended is the order it is trained. I think we should have LR go directly after XT. (Both classification and regression

szha:
Job PR-375-8 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-375/8/index.html

szha:
Job PR-375-11 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-375/11/index.html

szha:
Job PR-375-12 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-375/12/index.html

autogluon/utils/tabular/ml/models/lr/lr_preprocessing_utils.py
jwmueller: Here you mean memory issues during the preprocessing transform, not during the model training, correct? Or is the issue due to lack of sparse matrix support in the OneHotMergeRaresHandleUnknownEncoder?

For now, it's fine with me to use the standard OHE for this PR.  I can later substitute the OHE with the OneHotMergeRaresHandleUnknownEncoder in a separate PR.

Is there a way for you to help me reproduce this issue for debugging (via email)?

Innixma:
Result of automlbenchmark test:

autogluon_4h -> Baseline AG (Stacking)
autogluon_lr_4h -> LR model only (with bagging+stacking)
autogluon_lr_full_4h -> Baseline AG with LR model added in (Stacking)

```
autogluon_4h VS all
              framework  > autogluon_4h  < autogluon_4h  = autogluon_4h  % Less Avg. Errors Than autogluon_4h  time_train_s  metric_error  time_infer_s       acc       auc   logloss  loss_rescaled      rank  rank=1_count  rank=2_count  rank=3_count  rank>3_count  error_count
0          autogluon_4h               0               0              39                               0.00000  11165.480645      0.202322    557.805080  0.865016  0.886292  0.325018       0.034667  1.370968            26            12             1             0            0
1  autogluon_lr_full_4h              10              21               3                              -6.51865  11013.158065      0.228615    516.420520  0.847958  0.872236  0.368255       0.148292  1.693548             9            25             0             0            5
2       autogluon_lr_4h               1              30               0                             -58.83001   5556.551613      0.476849      2.620599  0.735114  0.782671  0.836186       0.967742  2.935484             1             0            30             0            8

```

This indicates that LR is not fully stable (5 of the 39 datasets failed by adding LR as an option (Likely OOM), and LR failed completely on 8 datasets (OOM or some other issue)). Also, when LR does work, it does not generally seem to be beneficial to the stack.

Also: 24hrs later and LR has not finished training on the Dionis dataset, meaning it likely does not handle multiclass very well, and it can effectively deadlock the process in these situations.

I think if we can resolve the crashes and find some way to detect the expected training time required, it will be in a state to be added as a model option, but not to the default model set until we have a way to identify when it will be helpful.

autogluon/utils/tabular/ml/models/lr/lr_model.py
```diff
@@ -0,0 +1,197 @@
...
+from autogluon.utils.tabular.ml.models.lr.hyperparameters.parameters import get_param_baseline, get_model_params, get_default_params, INCLUDE, IGNORE, ONLY
+from autogluon.utils.tabular.ml.models.lr.hyperparameters.searchspaces import get_default_searchspace
+from autogluon.utils.tabular.ml.models.lr.lr_preprocessing_utils import NlpDataPreprocessor, OheFeaturesGenerator, NumericDataPreprocessor
+
+logger = logging.getLogger(__name__)
+
+
+class LinearModel(AbstractModel):
+
+    def __init__(self, path: str, name: str, problem_type: str, objective_func, hyperparameters=None, features=None,
```
Innixma, line 22: Should be changed to:

```
    def __init__(self, path: str, name: str, problem_type: str, objective_func, hyperparameters=None, features=None,
                 feature_types_metadata=None, debug=0, **kwargs):
        self.model_class, self.penalty, self.handle_text = get_model_params(problem_type, hyperparameters)
        name = name + '-' + self.model_class.__name__
        name = name + '-text_' + self.handle_text
        super().__init__(path=path, name=name, problem_type=problem_type, objective_func=objective_func, hyperparameters=hyperparameters, features=features,
                         feature_types_metadata=feature_types_metadata, debug=debug)
        self.types_of_features = None
        self.pipeline = None

        self.model_params, default_params = get_default_params(self.problem_type, self.penalty)
```

Otherwise self.name and self.path are misaligned and the model is saved in the root directory "." rather than in {dir}/models/

jwmueller:
> @jwmueller Perhaps we should have some differentiation between 'trusted algorithms' and 'experimental algorithms'? If so, we could add LR in the state it is in to the 'experimental' set and make it very clear to the user that experimental models do not have consistent performance or stability.

I think we can just do this case-by-case for now, unless you foresee there being many experimental models in the near future?  For any model that is not in the default set (specified as the default `hyperparameter` keys in `task.fit()`), we will anyway need to add some documentation to inform the user of its existence.  So this documentation could just be ordered such that the experimental models appear at the bottom in their own section prefaced by a CAVEAT warning.  And the same CAVEAT could be added as a comment at the top of each experimental model's source code file.  I guess a more systematic distinction would come into play regarding unit-testing though, we should think about adding individual unit tests for each model-type.

autogluon/utils/tabular/ml/models/lr/lr_model.py
sgdread: good catch

szha:
Job PR-375-13 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-375/13/index.html

szha:
Job PR-375-14 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-375/14/index.html

autogluon/utils/tabular/ml/trainer/model_presets/presets.py
```diff
@@ -90,9 +90,10 @@ def get_preset_models_classification(path, problem_type, objective_func, stoppin
                                   objective_func=objective_func, stopping_metric=stopping_metric, hyperparameters=nn_options.copy()),
         )
     if lr_options is not None:
-        models.append(
-            LinearModel(path=path, name='LinearModel', problem_type=problem_type, objective_func=objective_func, hyperparameters=lr_options.copy())
-        )
+        for lr_option in lr_options:
```
Innixma, line 93: We have to ensure models don't share the same name

autogluon/utils/tabular/ml/trainer/model_presets/presets.py
```diff
@@ -159,9 +160,10 @@ def get_preset_models_regression(path, problem_type, objective_func, stopping_me
             models += [LGBModel(path=path, name='LightGBMRegressorCustom', problem_type=problem_type, objective_func=objective_func, stopping_metric=stopping_metric, hyperparameters=get_param_baseline_custom(problem_type))]
         # SKLearnModel(path=path, name='DummyRegressor', model=DummyRegressor(), problem_type=problem_type, objective_func=objective_func),
     if lr_options is not None:
-        models.append(
-            LinearModel(path=path, name='LinearModel', problem_type=problem_type, objective_func=objective_func, hyperparameters=lr_options.copy())
-        )
+        for lr_option in lr_options:
```
Innixma, line 163: ditto

autogluon/utils/tabular/ml/trainer/model_presets/presets.py
sgdread: Addressed in the last push

szha:
Job PR-375-15 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-375/15/index.html

szha:
Job PR-375-16 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-375/16/index.html

Innixma:
I ran a benchmark on the latest version of this PR:

```
                 framework  time_train_s  metric_error  time_infer_s       acc       auc   logloss  bestdiff  loss_rescaled      rank  rank=1_count  rank=2_count  rank=3_count  rank>3_count  error_count
0  autogluon_v3_lr_full_1h   3285.048718      0.286284    192.416207  0.852953  0.877648  0.498430  0.035740       0.307692  1.346154            24            15             0             0            0
1          autogluon_v2_1h   3232.005128      0.284045    194.409595  0.852604  0.882683  0.499813  0.021865       0.615385  1.653846            12            27             0             0            0

                 framework  time_train_s  metric_error  time_infer_s       acc       auc   logloss  bestdiff  loss_rescaled      rank  rank=1_count  rank=2_count  rank=3_count  rank>3_count  error_count
0  autogluon_v3_lr_full_4h  11637.597436      0.269802    562.811775  0.854806  0.877743  0.460741  0.022004       0.461538  1.487179            18            21             0             0            0
1          autogluon_v2_4h  11625.187179      0.267092    563.867179  0.857582  0.884163  0.462834  0.017560       0.461538  1.512821            17            22             0             0            0
```

Comparing mainline AG (autogluon_v2, auto_stack=True), with autogluon + LR model (AG with LR model added in, auto_stack=True). I only did the first fold of the datasets, instead of 10, so the results are a bit noisy.

Good thing to note that no errors were encountered during the run, and no problem had an immensely poor result indicating that the prior defects have been resolved.

While LR appears quite dominant given only 1h, for whatever reason this lead is nullified after 4h of training time. I will probably have to rerun the benchmark on all 10 folds to get a less noisy estimate, but it is clear that LR is stable and can be helpful to the ensemble.

The primary remaining issue is that LR can still take a very long time and we don't have a way to stop it or detect when it is/will do so.

Below is the time taken for a single model training (so 10-fold bagging would take 10x this time):


```
dataset | time_train_s
-- | --
Dionis | 3783
Robert | 254.6
Fashion-MNIST | 224.3
riccardo | 150
guiellermo | 149.9
Covertype | 141.5
Helena | 122.1
Volkert | 73
dilbert | 71.8
Albert | 51
fabert | 44.7
APSFailure | 25.9
christine | 18.9
KDDCup09_appetency | 18.7
Jannis | 18
MiniBooNE | 15.7
Airlines | 12.6
connect-4 | 11.1
cnae-9 | 7.4
nomao | 6.5
higgs | 4.4
Shuttle | 4.3
mfeat-factors | 3.7
numerai28.6 | 3.2
adult | 2.8
bank-marketing | 2.8
jasmine | 2.5
Amazon_employee_access | 2
segment | 2
jungle_chess_2pcs_raw_endgame_complete | 1.9
kr-vs-kp | 1.8
sylvine | 1.6
vehicle | 1.6
car | 1.5
credit-g | 1.5
kc1 | 1.5
Australian | 1.4
blood-transfusion | 1.4
phoneme | 1.4
```

While runtimes < 120 seconds are generally not that impactful, going far beyond leads to an issue where AutoGluon can go over the time limit or waste time, particularly during bagging. For example, on Dionis, all 10 folds must be fit to be a valid bagged model, but it takes 1 hour to train only a single fold. That means that given 1h total training time, the entire time budget is wasted on training an incomplete bag of LR (Thus not being a valid model). In the case of 4 hour time budget, 25% of the total time is wasted on LR (Again, won't be a valid model).

This isn't a blocking issue for the PR, but it is a blocking issue to be added to the core default set of models trained by Tabular.

autogluon/utils/tabular/ml/models/lr/lr_model.py
```diff
@@ -0,0 +1,204 @@
...
+                 feature_types_metadata=None, debug=0, **kwargs):
+        self.model_class, self.penalty, self.handle_text = get_model_params(problem_type, hyperparameters)
+        name = name + '-' + self.model_class.__name__
+        name = name + '-text_' + self.handle_text
```
Innixma, line 26: Lets avoid altering the name based on the hyperparameters. If the user wants to know more about a given linear learner, they can look at the info metadata.

Proposing to just name it as LinearModel_0, LinearModel_1, etc. (Same name as passed in init)

sgdread: Updated

szha:
Job PR-375-17 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-375/17/index.html

autogluon/task/tabular_prediction/tabular_prediction.py
```diff
@@ -146,6 +146,10 @@ def fit(train_data, label, tuning_data=None, output_directory=None, problem_type
                 KNN: See sklearn documentation: https://scikit-learn.org/stable/modules/generated/sklearn.neighbors.KNeighborsClassifier.html
                     Note: Hyperparameter tuning is disabled for this model.
                     Note: 'weights' parameter will be overriden. Both 'distance' and 'uniform' are used automatically, training two models.
+                LR: `autogluon/utils/tabular/ml/models/lr/hyperparameters/parameters.py`
+                    Note: a list of hyper-parameters should be passed; each set will create differen version of the model.
```
jwmueller, line 150: typo: differen

sgdread: fixed

Innixma:
+1 to Jonas' comment. If you can do that while also resolving the file conflict, that would be ideal!

sgdread:
Please don't merge it yet, I'll add the List/dict option

szha:
Job PR-375-19 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-375/19/index.html

szha:
Job PR-375-20 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-375/20/index.html

PR #379, opened 2020-03-25
fix for balanced accuracy computation

### LINKED ISSUE #378: Balanced accuracy error
(this is with 6b247bf)

Consider (from `tabular-indepth` tutorial):

```python
>>> y_test[:5]
0               Sales
1               Sales
2     Exec-managerial
3     Exec-managerial
4      Prof-specialty
Name: occupation, dtype: object

>>> y_pred[:5]
array([' Other-service', ' Craft-repair', ' Exec-managerial', ' Sales',
       ' Other-service'], dtype=object)
```

**with ag**

```python
>>> ag.utils.tabular.metrics.accuracy(y_test, y_pred)
0.3393387245368001

>>> ag.utils.tabular.metrics.balanced_accuracy(y_test, y_pred)
[error see stacktrace]
```

**with sklearn**

```python
>>> sklearn.metrics.balanced_accuracy_score(y_test, y_pred)
0.21896145445995055
```

### Reason

I believe the issue stems from this line: https://github.com/awslabs/autogluon/blob/6b247bfea9d504381cc512e36ba1909e6c54c0c3/autogluon/utils/tabular/metrics/classification_metrics.py#L21 

which would be ok if `prediction` and `solution` have been encoded to integer values but not in general. The max should be over the number of unique elements. This requires a bit more than a one line refactoring though, I'll open a PR. 

### Stacktrace

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-28-ddcca1d23e69> in <module>
----> 1 ag.utils.tabular.metrics.balanced_accuracy(y_test, y_pred)

~/Desktop/github-aws/autogluon-public/autogluon/utils/tabular/metrics/__init__.py in __call__(self, y_true, y_pred, sample_weight)
     87         else:
     88             return self._sign * self._score_func(y_true, y_pred,
---> 89                                                  **self._kwargs)
     90 
     91 

~/Desktop/github-aws/autogluon-public/autogluon/utils/tabular/metrics/classification_metrics.py in balanced_accuracy(solution, prediction)
     19     elif y_type == 'multiclass':
     20         # Need to create a multiclass solution and a multiclass predictions
---> 21         max_class = int(np.max((np.max(solution), np.max(prediction))))
     22         solution_binary = np.zeros((len(solution), max_class + 1))
     23         prediction_binary = np.zeros((len(prediction), max_class + 1))

<__array_function__ internals> in amax(*args, **kwargs)

~/Desktop/github-aws/ghaws/lib/python3.7/site-packages/numpy/core/fromnumeric.py in amax(a, axis, out, keepdims, initial, where)
   2666     """
   2667     return _wrapreduction(a, np.maximum, 'max', axis, None, out,
-> 2668                           keepdims=keepdims, initial=initial, where=where)
   2669 
   2670 

~/Desktop/github-aws/ghaws/lib/python3.7/site-packages/numpy/core/fromnumeric.py in _wrapreduction(obj, ufunc, method, axis, dtype, out, **kwargs)
     88                 return reduction(axis=axis, out=out, **passkwargs)
     89 
---> 90     return ufunc.reduce(obj, axis, dtype, out, **passkwargs)
     91 
     92 

TypeError: cannot perform reduce with flexible type
```

tlienart:
closes #378

Before:

```python
>>> ag.utils.tabular.metrics.balanced_accuracy(y_test, y_pred)
[error]
```

After

```python
>>> ag.utils.tabular.metrics.balanced_accuracy(y_test, y_pred)
0.1826867855754422
```

Sklearn:

```python
>>> sklearn.metrics.balanced_accuracy_score(y_test, y_pred)
0.1826867855754422
```

## MWE

```python
import autogluon as ag
import pandas as pd
d = {'solution': ['a','b','c','a','b','c'], 'prediction': ['b','b','c','d','b','c']}
df = pd.DataFrame(data=d)
ag.utils.tabular.metrics.balanced_accuracy(df['solution'], df['prediction'])
```

with this PR, it returns `0.75`, without, you get an error. 

### Notes

* There may be more efficient ways to do the OHE.
* We could also just wrap around `sklearn.metrics.balanced_accuracy_score`, note however that it fails on the MWE above because that score expects the classes to match (afaict); for non-pathological cases where all classes are present in both vectors, the results are the same:

```python
import autogluon as ag
import pandas as pd
import sklearn as sk
d = {'solution': ['a','b','c','a','b','c','d','d'], 'prediction': ['b','b','c','d','b','c','a','d']}
df = pd.DataFrame(data=d)
ag.utils.tabular.metrics.balanced_accuracy(df['solution'], df['prediction']) # 0.625
sk.metrics.balanced_accuracy_score(df['solution'], df['prediction']) # 0.625
```

### DIFF
--- a/autogluon/utils/tabular/metrics/classification_metrics.py
+++ b/autogluon/utils/tabular/metrics/classification_metrics.py
@@ -17,15 +17,19 @@ def balanced_accuracy(solution, prediction):
         pass
 
     elif y_type == 'multiclass':
-        # Need to create a multiclass solution and a multiclass predictions
-        max_class = int(np.max((np.max(solution), np.max(prediction))))
-        solution_binary = np.zeros((len(solution), max_class + 1))
-        prediction_binary = np.zeros((len(prediction), max_class + 1))
-        for i in range(len(solution)):
-            solution_binary[i, int(solution[i])] = 1
-            prediction_binary[i, int(prediction[i])] = 1
-        solution = solution_binary
-        prediction = prediction_binary
+        n = len(solution)
+        unique_sol, encoded_sol = np.unique(solution, return_inverse=True)
+        unique_pred, encoded_pred = np.unique(prediction, return_inverse=True)
+        classes = np.unique(np.concatenate((unique_sol, unique_pred)))
+        map_sol = np.array([np.where(classes==c)[0][0] for c in unique_sol])
+        map_pred = np.array([np.where(classes==c)[0][0] for c in unique_pred])
+        # one hot encoding
+        sol_ohe = np.zeros((n, len(classes)))
+        pred_ohe = np.zeros((n, len(classes)))
+        sol_ohe[np.arange(n), map_sol[encoded_sol]] = 1
+        pred_ohe[np.arange(n), map_pred[encoded_pred]] = 1
+        solution = sol_ohe
+        prediction = pred_ohe
 
     elif y_type == 'multilabel-indicator':
         solution = solution.toarray()


szha:
Job PR-379-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-379/1/index.html

Innixma:
Interesting examples, does this mean that after the PR, AutoGluon's balanced_accuracy is more robust than sklearns?

autogluon/utils/tabular/metrics/classification_metrics.py
```diff
@@ -17,15 +17,19 @@ def balanced_accuracy(solution, prediction):
...
-        solution = solution_binary
-        prediction = prediction_binary
+        n = len(solution)
+        unique_sol, encoded_sol = np.unique(solution, return_inverse=True)
+        unique_pred, encoded_pred = np.unique(prediction, return_inverse=True)
+        classes = np.unique(np.concatenate((unique_sol, unique_pred)))
+        map_sol = np.array([np.where(classes==c)[0][0] for c in unique_sol])
+        map_pred = np.array([np.where(classes==c)[0][0] for c in unique_pred])
+        # one hot encoding
+        sol_ohe = np.zeros((n, len(classes)))
```
Innixma: A bit concerned with OHE, if num classes is very large (say 1000), this will become huge. Is there any way to avoid OHE? I think a performance benchmark example with runtimes of sklearn vs ag on this metric could be helpful (with artificially very large class count inputs)

Innixma: sklearn source code for reference:

```
    C = confusion_matrix(y_true, y_pred, sample_weight=sample_weight)
    with np.errstate(divide='ignore', invalid='ignore'):
        per_class = np.diag(C) / C.sum(axis=1)
    if np.any(np.isnan(per_class)):
        warnings.warn('y_pred contains classes not in y_true')
        per_class = per_class[~np.isnan(per_class)]
    score = np.mean(per_class)
    if adjusted:
        n_classes = len(per_class)
        chance = 1 / n_classes
        score -= chance
        score /= 1 - chance
    return score
```

Innixma: Furthermore, does it make sense to not error if a class is missing? Are we sure sklearn's implementation is not correct in this instance?

tlienart: 1. we could add logic to compute a confusion matrix, this would be useful because we could compute a whole bunch of derived metrics out of it, note that the confusion matrix in multi class is C x C where C is the number of classes so if you have 1000 classes, you still have to pay that price. Note also that the previous code, afaik, was basically OHE based as well so this is just a fix, but we could open another issue and fix this proper. I would be in favour of adding logic to compute the confusion matrix in general and then have methods on that
2. in terms of sklearn's implementation, this is a somewhat tricky topic, the behaviour now is what I believe the intended behaviour was before I made the change so I kept it that way. IMO it makes sense to have this (slightly) more general thing: you could train on 100 classes but then make a prediction on something where you only predict 50 classes and the ground truth has 45. This would not happen in a situation where you have scores for every class seen at training though. 

So basically I would suggest the following steps:

1. merge this for now (possibly with minor modification) as it's the past behaviour but actually correct in the general case (I think)
2. open another issue discussing confusion matrix + derived metrics

Innixma: Good points and I'm aligned on your next-steps. Thanks for the contribution!

PR #384, opened 2020-03-27
searcher improvement

zhanghang1989:
*Issue #, if available:*

*Description of changes:*
Thanks @mseeger for pointing this out.  The current searcher would suggest next configuration even there is no compute resources available. This will impact the performance of search algorithms. The last one configuration can not fully utilize all previous results.

As suggested by @mseeger , moving cls.RESOURCE_MANAGER._request(), before searcher.get_config()

### DIFF
--- a/autogluon/scheduler/fifo.py
+++ b/autogluon/scheduler/fifo.py
@@ -167,6 +167,11 @@ class FIFOScheduler(TaskScheduler):
         # Allow for the promotion of a previously chosen config. Also,
         # extra_kwargs contains extra info passed to both add_job and to
         # get_config (if no config is promoted)
+
+        # request resource first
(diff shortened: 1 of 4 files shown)


zhanghang1989:
> In general, the benefit is that you can run get_config with more data. The drawback could be that if get_config takes a while to run, this could mean that a worker who is ready, has to wait.

I agree.
This change should improve the performance of most of the searchers and does not impact rand search. I think the delay during the `get_config()` is okay, which is usually a small overhead for each trial.

szha:
Job PR-384-5 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-384/5/index.html

PR #389, opened 2020-03-30
Fix mousouver plot after bokeh dropped previewsave

jwmueller:
Update mousoverplot to match bokeh update to v2.0.1 which dropped compatibility with previewsave tool.

### DIFF
--- a/autogluon/utils/plots.py
+++ b/autogluon/utils/plots.py
@@ -138,7 +138,7 @@ def mousover_plot(datadict, attr_x, attr_y, attr_color=None, attr_size=None, sav
         bokeh_imported = False
     
     if not bokeh_imported:
-        warnings.warn('AutoGluon summary plots cannot be created because bokeh is not installed. To see plots, please do: "pip install bokeh"')
+        warnings.warn('AutoGluon summary plots cannot be created because bokeh is not installed. To see plots, please do: "pip install bokeh==2.0.1"')
         return None
     
     n = len(datadict[attr_x])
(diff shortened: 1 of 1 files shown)


Innixma:
Will this fix work with older versions of bokeh? If not we should specify version requirements.

Also, this might break 0.0.5 and 0.0.6 for users. Do we want to do a version update to 0.0.7? How widespread is the impact?

szha:
Job PR-389-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-389/1/index.html

jwmueller:
Most older versions of bokeh should still work.  Note that bokeh is not currently a listed requirement, we currently just print a warning for the user to install bokeh when the lazy import fails.  I'll update this print statement to direct the user to install a specific bokeh version, so we don't run into this scenario again in the future.

I'm not sure we should install a particular bokeh version on the CI server though, it's in some ways good to be able to catch issues like this. 

I'm also not sure the backwards-incompatible bokeh update should force us to release autogluon  0.0.7 in response, since we don't specify bokeh as an official requirement.  Only the `fit_summary()` method used by various tasks' `Predictor` objects is broken as a result of the update.  I think a user who encounters this error can just downgrade their bokeh version and then recall `fit_summary()`, so all their computation expended during training is not wasted (will post a github issue to clarify this).

@zhanghang1989 any thoughts?

jwmueller:
Issue to warn recent users of autogluon <= 0.0.6 about newest bokeh:  https://github.com/awslabs/autogluon/issues/390

PR #395, opened 2020-03-31
Sagemaker Compatibility

zhanghang1989:
*Description of changes:*

Many thanks to @tlienart !

Working with SageMaker example at https://github.com/zhanghang1989/AutoGluon-Docker

The change does not break any existing features.

### DIFF
--- a/autogluon/scheduler/remote/remote.py
+++ b/autogluon/scheduler/remote/remote.py
@@ -50,7 +50,7 @@ class Service(object):
         self.status = 'live'
 
     def shutdown(self):
-        self.proc.kill()
+        os.killpg(os.getpgid(self.proc.pid), signal.SIGTERM)
 
 def start_service(remote_ip, port):
     cmd = ['agremote', '--address', remote_ip, '--port', str(port)]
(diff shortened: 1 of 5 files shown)


autogluon/scheduler/remote/remote.py
```diff
@@ -68,6 +68,9 @@ def __init__(self, remote_ip=None, port=None, local=False, ssh_username=None,
         else:
             remote_addr = (remote_ip + ':{}'.format(port))
             self.service = start_service(remote_ip, port)
+            _set_global_remote_service(self.service)
+            import time
+            time.sleep(10)
```
Innixma: Can this be a lower value? What impact will this have on repeated HPO runs for many different models like in Tabular?

zhanghang1989: 1). This does not impact single machine training. 

2). The 10 seconds overhead is introduced only once at the starting of the remote nodes, which is usually acceptable.  There won't be any overhead during the job scheduling.

zhanghang1989: > Can this be a lower value? What impact will this have on repeated HPO runs for many different models like in Tabular?

There is no overhead for repeated run.

szha:
Job PR-395-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-395/1/index.html

autogluon/task/object_detection/data_parallel.py
```diff
@@ -53,6 +52,7 @@ def forward_backward(self, x):
             rcnn_l1_loss_metric = [[box_targets, box_masks], [box_pred]]
 
             if self.amp:
+                from mxnet.contrib import amp
```
Jerryzcn, line 55: is there a specific reason for this?

PR #396, opened 2020-03-31
Tabular: Added support for .tsv files automatically inferring delimiter

Innixma:
*Issue #, if available:*

*Description of changes:*

Tabular: Added support for .tsv files automatically inferring delimiter as '\t'.

- Enables code such as:

```
from autogluon import TabularPrediction as task
predictor = task.fit(train_data='train_data.tsv', label='class')
```

### DIFF
--- a/autogluon/utils/tabular/utils/loaders/load_pd.py
+++ b/autogluon/utils/tabular/utils/loaders/load_pd.py
@@ -11,7 +11,8 @@ from .load_s3 import list_bucket_prefix_suffix_s3
 
 logger = logging.getLogger(__name__)
 
-def load(path, delimiter=',', encoding='utf-8', columns_to_keep=None, dtype=None, error_bad_lines=True, header=0, 
+
+def load(path, delimiter=None, encoding='utf-8', columns_to_keep=None, dtype=None, error_bad_lines=True, header=0,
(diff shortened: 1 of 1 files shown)


autogluon/utils/tabular/utils/loaders/load_pd.py
```diff
@@ -35,6 +36,12 @@ def load(path, delimiter=',', encoding='utf-8', columns_to_keep=None, dtype=None
         format = 'parquet'
     else:
         format = 'csv'
+        if path.endswith('.tsv'):
+            if delimiter is None:
+                delimiter = '\t'
```
jwmueller: Should we add logger debug message stating the delimiter was inferred (at L41 here and L44)?  I imagine the official error message may be opaque if the delimiter is wrong.

szha:
Job PR-396-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-396/1/index.html

autogluon/utils/tabular/utils/loaders/load_pd.py
Innixma: Good point, added

szha:
Job PR-396-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-396/2/index.html

PR #400, opened 2020-04-02
Tabular: Fixed defect where model val_score was not updated properly

Innixma:
*Issue #, if available:*

*Description of changes:*

Issue caused when time_limit specified and repeated bagging occurs. Model will not update val_score beyond its initial score achieved after a single bagged fit. This fix makes it so the val_score is updated each time it is updated as a model in trainer. Before this fix, .leaderboard() would not show correct val_score for these models, however the best_model would still be chosen correctly.


szha:
Job PR-400-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-400/1/index.html

PR #403, opened 2020-04-04
Tabular: Added support for advanced stacker model interactions

Innixma:
*Issue #, if available:*

*Description of changes:*

Major Changes:
- Expanded stacker functionality to work outside of level-alignment
  - This means that a stacker model at L3 can now take models at L0, L1, and L2 as input.
  - Previously they could only take models at L2 as input.
- Upcoming PR will build upon this functionality to enable _FULL ensembles, resulting in 10-200x inference speedup with minimal accuracy loss for bag enabled ensembles.
- Enables functionality to swap model pred_proba inputs on an already fitted stacker model. (Tradeoff between bagged stacker and _FULL stacker on-the-fly for accuracy/inference speed.)

### DIFF
--- a/autogluon/utils/tabular/ml/learner/abstract_learner.py
+++ b/autogluon/utils/tabular/ml/learner/abstract_learner.py
@@ -212,71 +212,39 @@ class AbstractLearner:
             if (not trainer.objective_func_expects_y_pred) and (-1 in y.unique()):
                 # Log loss
                 raise ValueError(f'Multiclass scoring with eval_metric={self.objective_func.name} does not support unknown classes.')
-        # TODO: Move below into trainer, should not live in learner
 
-        max_level_to_check = trainer.get_max_level_all()
         scores = {}
-        pred_time_test_marginal = {}
-        pred_probas = None
-        stack_names = list(trainer.models_level.keys())
-        stack_names_not_core = [name for name in stack_names if name != 'core']
+        all_trained_models = trainer.get_model_names_all()
+        all_trained_models_original = all_trained_models.copy()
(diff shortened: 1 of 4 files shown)


szha:
Job PR-403-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-403/1/index.html

jwmueller:
Seems like the plan going forward is to keep the model names as: "<MODELTYPE>_STACKER_lX" where X is the stacker "level" (determined by which round of stacking this model was trained in)?

If arbitrary DAGs are supported (and eventually actually used), then the documentation will  probably need to clarify what "level" means precisely to avoid user confusion.  

If renaming eventually comes under consideration, I still think we might reconsider dropping the "_STACKER_*" suffix from the base models to more clearly distinguish them, as well as more clearly distinguishing between bagged vs not-bagged models (could be a separate suffix).

Innixma:
Ran this change on automlbenchmark, we match or outperform previous version, with no errors, so I have confirmed stability. We don't experience visible slowdown in inference time on automlbenchmark.

I don't intend to utilize the complex DAG functionality immediately, its implementation was a pre-requisite for _FULL ensembles due to intricacies in the code structure rather than _FULL ensembles actually utilizing complex DAG functionality. The most likely place it will be used is for weighted ensembles that can weight models at different stack levels.

Once this is the case, I will add to the documentation as you say. I also agree that we should make a tutorial notebook on these features. I think this can wait until after _FULL functionality and user-facing .info() are added. I also wanted to add a draw option to visualize the graph, but networkx makes weird drawings that are near impossible to make sense of. If we can figure out how to draw the graph nicely, I think that would be extremely helpful in the notebook to visualize the stackers and how they work.

PR #410, opened 2020-04-11
Fixes in tabular prediction

razmikmelikbekyan:
*Issue #, if available:*

*Description of changes:*

- Sklearn model moved to abstract model, since it is abstract model itself. 
- `abstract_spaces.py` for abstract model and lgb model are deleted, since they are not used. 
- `searchspaces.py` is created for catboost in order to keep the same convention for HPO relevant models (similar to lgb and NN)
- `_set_default_searchspace` method is deleted for lgb, since it was duplicating mother class method.
- other basic PEP8 corrections and small optimisations. 


szha:
Job PR-410-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-410/1/index.html

autogluon/utils/tabular/ml/models/knn/knn_model.py
```diff
@@ -26,6 +27,7 @@ def preprocess(self, X):
         cat_columns = X.select_dtypes(['category']).columns
         X = X.drop(cat_columns, axis=1)  # TODO: Test if crash when all columns are categorical
         X = super().preprocess(X).fillna(0)
+        X = normalize(X, copy=True)
```
Innixma, line 30: Will do a review of the rest of the changes later, but just wanted to comment on this:

I believe this has to be done in a stateful manner or else the normalization done on train/val/test will differ.

I would also prefer if this change be its own separate PR as it is a major functionality change for KNN (It is not necessarily a strict improvement on all problems and it slows down KNN for both training and prediction, so it needs to be evaluated through a benchmark)

szha:
Job PR-410-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-410/2/index.html

autogluon/utils/tabular/ml/models/lgb/lgb_model.py
```diff
@@ -272,18 +260,9 @@ def _get_train_loss_name(self):
         elif self.problem_type == REGRESSION:
             train_loss_name = 'l2'
         else:
-            raise ValueError("unknown problem_type for LGBModel: %s" % self.problem_type)
+            raise ValueError(f"unknown problem_type for LGBModel: {self.problem_type}")
         return train_loss_name
 
-    def _set_default_searchspace(self):
```
Innixma, line 278: LightGBM needs this function because it passes num_classes unlike AbstractModel. Also, it needs to call LightGBM's get_default_searchspace, not AbstractModel's.

Innixma: Perhaps we add _get_default_searchspace function which is called by _set_default_searchspace, so each model only needs to add a 1-liner _get_default_searchspace.

```
def _get_default_searchspace(self):
    return get_default_searchspace(problem_type=self.problem_type, num_classes=self.num_classes)
```

autogluon/utils/tabular/ml/models/rf/rf_model.py
```diff
@@ -130,7 +131,4 @@ def get_model_feature_importance(self):
             # TODO: Consider making this raise an exception
             logger.warning('Warning: get_model_feature_importance called when self.features is None!')
             return dict()
-        feature_names = self.features
-        importances = self.model.feature_importances_
-        importance_dict = {feature_name: importance for (feature_name, importance) in zip(feature_names, importances)}
-        return importance_dict
+        return dict(zip(self.features, self.model.feature_importances_))
```
Innixma, line 134: Didn't know you could dict a zip, cool!

autogluon/utils/tabular/ml/models/lgb/lgb_model.py
razmikmelikbekyan: > LightGBM needs this function because it passes num_classes unlike AbstractModel. Also, it needs to call LightGBM's get_default_searchspace, not AbstractModel's.

my mistake, I forgot that it is calling LGB's `get_default_searchspace`. The last suggestion is fine, will add the method.

szha:
Job PR-410-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-410/3/index.html

PR #412, opened 2020-04-13
Tabular: Added refit_full user-facing functionality

Innixma:
*Issue #, if available:*

*Description of changes:*

Added refit_full functionality to user-facing layer.

Benchmarking is still running, will update with results tomorrow.

### DIFF
--- a/autogluon/task/tabular_prediction/predictor.py
+++ b/autogluon/task/tabular_prediction/predictor.py
@@ -63,13 +63,14 @@ class TabularPredictor(BasePredictor):
             To access any trainer method `func()` from this `Predictor`, use: `predictor._trainer.func()`.
         """
         self._learner: Learner = learner  # Learner object
+        self._learner.persist_trainer(low_memory=True)
         self._trainer: AbstractTrainer = self._learner.load_trainer()  # Trainer object
         self.output_directory = self._learner.path
         self.problem_type = self._learner.problem_type
         self.eval_metric = self._learner.objective_func
         self.label_column = self._learner.label
         self.feature_types = self._trainer.feature_types_metadata
-        self.model_names = self._trainer.get_model_names_all()
+        self.model_names = self._trainer.get_model_names_all()  # TODO: Will be outdated if new models are trained through refit_full()
         self.model_performance = self._trainer.model_performance
         self.class_labels = self._learner.class_labels
 
--- a/autogluon/task/tabular_prediction/predictor.py
+++ b/autogluon/task/tabular_prediction/predictor.py
@@ -388,7 +388,7 @@ class TabularPredictor(BasePredictor):
 
         return self._learner.get_feature_importance(model=model, X=dataset, features=features, raw=raw, subsample_size=subsample_size, silent=silent)
 
-    def refit_full(self, model=None):
+    def refit_full(self, model='all'):
         """
         Retrain model on all of the data (training + validation).
         For bagged models:
@@ -412,8 +412,10 @@ class TabularPredictor(BasePredictor):
 
         Parameters
         ----------
-        model : str, default = None
-            Model to refit. If None then all models are refitted.
+        model : str, default = 'all'
+            Model name of model to refit.
+                If 'all' then all models are refitted.
+                If 'best' then the model with the highest validation score is refit.
             All ancestor models will also be refit in the case that the selected model is a weighted or stacker ensemble.
             Valid models are listed in this `predictor` by calling `predictor.model_names`.
 
--- a/autogluon/task/tabular_prediction/predictor.py
+++ b/autogluon/task/tabular_prediction/predictor.py
@@ -71,7 +71,7 @@ class TabularPredictor(BasePredictor):
         self.eval_metric = self._learner.objective_func
         self.label_column = self._learner.label
         self.feature_types = self._trainer.feature_types_metadata
-        self.model_names = self._trainer.get_model_names_all()  # TODO: Will be outdated if new models are trained through refit_full()
+        self.model_names = self._trainer.get_model_names_all()  # TODO: Consider making this a function instead of a variable: This should never be de-synced with the output of self._trainer.get_model_names_all()
         self.model_performance = self._trainer.model_performance
         self.class_labels = self._learner.class_labels
 
@@ -390,15 +390,25 @@ class TabularPredictor(BasePredictor):
 
     def refit_full(self, model=None):
         """
-        Optimizes a model's inference time by collapsing bagged ensembles into a single model fit on all of the training data.
-        For bagged models, this process will typically result in a slight accuracy reduction and a large inference speedup.
+        Retrain model on all of the data (training + validation).
+        For bagged models:
+            Optimizes a model's inference time by collapsing bagged ensembles into a single model fit on all of the training data.
+            This process will typically result in a slight accuracy reduction and a large inference speedup.
             The inference speedup will generally be between 10-200x faster than the original bagged ensemble model.
+                The inference speedup factor is equivalent to (k * n), where k is the number of folds (`num_bagging_folds`) and n is the number of finished repeats (`num_bagging_sets`) in the bagged ensemble.
             The runtime is generally 10% or less of the original fit runtime.
-        This function can also be called on non-bagged models for a slight accuracy increase and no change to inference time.
-            In this instance, the runtime will be approximately equal to the original fit runtime.
+                The runtime can be roughly estimated as 1 / (k * n) of the original fit runtime, with k and n defined above.
+        For non-bagged models:
+            Optimizes a model's accuracy by retraining on 100% of the data without using a validation set.
+            Will typically result in a slight accuracy increase and no change to inference time.
+            The runtime will be approximately equal to the original fit runtime.
         This process does not alter the original models, but instead adds additional models.
+        If stacker models are refit by this process, they will use the refit_full versions of the ancestor models during inference.
         Models produced by this process will not have validation scores, as they use all of the data for training.
-        Therefore, it is up to the user to determine if the models are of sufficient quality by including test data in `predictor.leaderboard(dataset=test_data)`.
+            Therefore, it is up to the user to determine if the models are of sufficient quality by including test data in `predictor.leaderboard(dataset=test_data)`.
+            If the user does not have additional test data, they should reference the original model's score for an estimate of the performance of the refit_full model.
+                Warning: Be aware that utilizing refit_full models without separately verifying on test data means that the model is untested, and has no guarantee of being consistent with the original model.
+        `cache_data` must have been set to `True` during the original training to enable this functionality.
 
         Parameters
         ----------
@@ -409,9 +419,11 @@ class TabularPredictor(BasePredictor):
 
         Returns
         -------
-        Dictionary of original model names -> refit full model names.
+        Dictionary of original model names -> refit_full model names.
         """
-        return self._learner.refit_ensemble_full(model=model)
+        refit_full_dict = self._learner.refit_ensemble_full(model=model)
+        self.model_names = self._trainer.get_model_names_all()
+        return refit_full_dict
 
     def get_model_full_dict(self):
         """
--- a/autogluon/task/tabular_prediction/predictor.py
+++ b/autogluon/task/tabular_prediction/predictor.py
@@ -1,3 +1,4 @@
+import copy
 import logging
 
 import pandas as pd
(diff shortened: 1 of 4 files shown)


szha:
Job PR-412-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-412/1/index.html

Innixma:
Benchmark Results:

```
autogluon_v2_1h VS all
                          framework  > autogluon_v2_1h  < autogluon_v2_1h  = autogluon_v2_1h  % Less Avg. Errors Than autogluon_v2_1h  time_train_s  metric_error  time_infer_s       acc       auc   logloss  loss_rescaled      rank  rank=1_count  rank=2_count  rank=3_count  rank>3_count  error_count
0                   autogluon_v2_1h                  0                  0                 38                                 0.000000   3230.091176      0.279153    169.679150  0.858036  0.896050  0.501077       0.214943  1.867647            17            10            10             1            1
1        autogluon_v2_compressed_1h                 11                 24                  1                                -0.172806   3644.944118      0.282663     15.444709  0.857795  0.894187  0.506674       0.377367  2.250000             9            17             4             7            2
2  autogluon_v2_nobag_compressed_1h                 11                 25                  0                               -12.366583   1082.573529      0.274612      7.523006  0.856385  0.895306  0.489845       0.548736  2.529412             7             8            18             4            2
3             autogluon_v2_nobag_1h                 10                 27                  0                               -14.895866    580.670588      0.280489      7.103622  0.855686  0.890711  0.497342       0.772690  3.352941             5             5             6            22            1

autogluon_v2_4h VS all
                          framework  > autogluon_v2_4h  < autogluon_v2_4h  = autogluon_v2_4h  % Less Avg. Errors Than autogluon_v2_4h  time_train_s  metric_error  time_infer_s       acc       auc   logloss  loss_rescaled      rank  rank=1_count  rank=2_count  rank=3_count  rank>3_count  error_count
0                   autogluon_v2_4h                  0                  0                 37                                 0.000000  11057.985294      0.259324    510.558659  0.868035  0.901404  0.462912       0.132871  1.602941            21             9             7             0            2
1        autogluon_v2_compressed_4h                  5                 28                  3                                -4.411548  12144.014706      0.263722     28.211233  0.864026  0.898227  0.468860       0.375229  2.264706             6            21             3             7            2
2  autogluon_v2_nobag_compressed_4h                  9                 26                  0                               -16.233976   1849.879412      0.270656     10.598040  0.860595  0.900316  0.487221       0.601134  2.764706             4            10            16             7            2
3             autogluon_v2_nobag_4h                  6                 30                  0                               -18.441443    962.850000      0.276271     10.173476  0.860412  0.895628  0.494008       0.824352  3.367647             5             2            10            21            1
```

As can be seen, _FULL compression of bagged models results in significant inference speedup (510s avg -> 28s avg) with only minor drops in overall accuracy and slight increase to runtime. This result leads to a highly desirable middle ground between full bagged ensembles and non-bagged ensembles in terms of accuracy and inference time. _FULL conversion on non-bagged models shows inference parity while demonstrating superior accuracy and doubling of training time, as predicted. All errors are unrelated to the _FULL logic.

autogluon/task/tabular_prediction/predictor.py
```diff
@@ -63,13 +64,14 @@ def __init__(self, learner):
...
         self._learner: Learner = learner  # Learner object
+        self._learner.persist_trainer(low_memory=True)
         self._trainer: AbstractTrainer = self._learner.load_trainer()  # Trainer object
         self.output_directory = self._learner.path
         self.problem_type = self._learner.problem_type
         self.eval_metric = self._learner.objective_func
         self.label_column = self._learner.label
         self.feature_types = self._trainer.feature_types_metadata
-        self.model_names = self._trainer.get_model_names_all()
+        self.model_names = self._trainer.get_model_names_all()  # TODO: Will be outdated if new models are trained through refit_full()
```
jwmueller, line 74: shouldn't this be updated by `refit_full()`? otherwise things might appear broken to user?

autogluon/task/tabular_prediction/predictor.py
```diff
@@ -386,6 +388,43 @@ def feature_importance(self, model=None, dataset=None, features=None, raw=True,
 
         return self._learner.get_feature_importance(model=model, X=dataset, features=features, raw=raw, subsample_size=subsample_size, silent=silent)
 
+    def refit_full(self, model=None):
+        """
+        Optimizes a model's inference time by collapsing bagged ensembles into a single model fit on all of the training data.
+        For bagged models, this process will typically result in a slight accuracy reduction and a large inference speedup.
+            The inference speedup will generally be between 10-200x faster than the original bagged ensemble model.
```
jwmueller, line 395: Why not define the speedup mathematically, is the approximation too crude?
E.g. I'd imagine with k-fold bagging, the full model will be ~k times faster than the bagged model (also for repeated bagging, etc.).

autogluon/task/tabular_prediction/predictor.py
```diff
@@ -386,6 +388,43 @@ def feature_importance(self, model=None, dataset=None, features=None, raw=True,
 
         return self._learner.get_feature_importance(model=model, X=dataset, features=features, raw=raw, subsample_size=subsample_size, silent=silent)
 
+    def refit_full(self, model=None):
+        """
+        Optimizes a model's inference time by collapsing bagged ensembles into a single model fit on all of the training data.
```
jwmueller, line 393: What do you think about extending this function (in separate PR) to also be applicable to non-bagged models trained via a standard train/val split? (FULL just refits on train+val in this case).

I'm not sure we want user to have to learn too many refit_full variants for different predictors generated via different task.fit() args.

autogluon/task/tabular_prediction/predictor.py
```diff
@@ -386,6 +388,43 @@ def feature_importance(self, model=None, dataset=None, features=None, raw=True,
...
+        For bagged models, this process will typically result in a slight accuracy reduction and a large inference speedup.
+            The inference speedup will generally be between 10-200x faster than the original bagged ensemble model.
+            The runtime is generally 10% or less of the original fit runtime.
+        This function can also be called on non-bagged models for a slight accuracy increase and no change to inference time.
+            In this instance, the runtime will be approximately equal to the original fit runtime.
+        This process does not alter the original models, but instead adds additional models.
+        Models produced by this process will not have validation scores, as they use all of the data for training.
```
jwmueller, line 400: Should we add statement addressed to users with no extra test data? (as this may be many of them)

autogluon/task/tabular_prediction/predictor.py
```diff
@@ -386,6 +388,43 @@ def feature_importance(self, model=None, dataset=None, features=None, raw=True,
...
+        Therefore, it is up to the user to determine if the models are of sufficient quality by including test data in `predictor.leaderboard(dataset=test_data)`.
+
+        Parameters
+        ----------
+        model : str, default = None
+            Model to refit. If None then all models are refitted.
```
jwmueller, line 406: should we refit the best_model by default instead, so this runs faster?  
Eg. if the WeightedEnsemble only depends on models #1 and #2, it probably is unnecessary to refit model #3 by default right?

autogluon/task/tabular_prediction/tabular_prediction.py
```diff
@@ -117,6 +117,20 @@ def fit(train_data, label, tuning_data=None, output_directory=None, problem_type
             Use `auto_stack` to maximize predictive accuracy; use `hyperparameter_tune` if you prefer to deploy just a single model rather than an ensemble.
         feature_prune : bool, default = False
             Whether or not to perform feature selection.
+        refit_full : bool, default = False
+            Whether to retrain all models on all of the data after the normal training procedure.
```
jwmueller, line 121: maybe clarify: "on all of the data (training+validation) after ..."

autogluon/task/tabular_prediction/tabular_prediction.py
```diff
@@ -117,6 +117,20 @@ def fit(train_data, label, tuning_data=None, output_directory=None, problem_type
...
+            The time taken by this process is not enforced by `time_limits`.
+            If `True`, optimizes a model's inference time by collapsing bagged ensembles into a single model fit on all of the training data.
+            This process will typically result in a slight accuracy reduction and a large inference speedup.
+            The inference speedup will generally be between 10-200x faster than the original bagged ensemble model.
+            The added fit runtime of this function is generally 10% or less of the original fit runtime.
```
jwmueller, line 126: Many of these statements are only relevant if user specified bagging should be used (via  `auto_stack`, `num_bagging_folds`...), so there should probably be a qualifying statement preceding them.

autogluon/task/tabular_prediction/tabular_prediction.py
```diff
@@ -117,6 +117,20 @@ def fit(train_data, label, tuning_data=None, output_directory=None, problem_type
...
+            This function can also be called on non-bagged models for a slight accuracy increase and no change to inference time.
+                In this instance, the added fit runtime will be approximately equal to the original fit runtime.
+            This process does not alter the original models, but instead fits additional models.
+            Models produced by this process will not have validation scores, as they use all of the data for training.
+                Therefore, it is up to the user to determine if the models are of sufficient quality by including test data in `predictor.leaderboard(dataset=test_data)`.
+            This is equivalent to calling `predictor.refit_full()` after training.
```
jwmueller, line 132: The documentation for `predictor.refit_full()` is missing the case of non-bagged ensembles.  In general both docs should probably be more clearly partitioned into two sections: bagging was used in task.fit() vs bagging was not used.

autogluon/task/tabular_prediction/tabular_prediction.py
```diff
@@ -212,6 +237,10 @@ def fit(train_data, label, tuning_data=None, output_directory=None, problem_type
             id_columns : list, default = []
                 Banned subset of column names that model may not use as predictive features (e.g. contains label, user-ID, etc).
                 These columns are ignored during `fit()`, but DataFrame of just these columns with appended predictions may be produced, for example to submit in a ML competition.
+            set_best_to_refit_full : bool, default = False
+                If True, will set Trainer.best_model = Trainer.full_model_dict[Trainer.best_model]
+                This will change the default model used in Predictor for predictions when model is not specified to the refit_full version of the best model.
```
jwmueller, line 242: This will change the default model used in Predictor for predictions when model is not specified to the refit_full version of the best model.
->
This will change the default model that Predictor uses for prediction when model is not specified to the refit_full version of the model that previously exhibit the highest validation score.

autogluon/task/tabular_prediction/predictor.py
Innixma: Added, clarified that the comment was towards a more general statement that self.model_names should probably be a function to ensure it is always aligned (It doesn't make sense to enable user to mutate it as a variable)

autogluon/task/tabular_prediction/predictor.py
Innixma: Originally I thought this would be too technical, but I think its manageable, added.

autogluon/task/tabular_prediction/predictor.py
Innixma: Clarified in documentation

autogluon/task/tabular_prediction/predictor.py
Innixma: Added

autogluon/task/tabular_prediction/predictor.py
Innixma: Good point, however in many cases a user may be using this functionality to try to find a good accuracy/latency tradeoff, which isn't necessarily the 'best model'. This is already quite fast relative to the normal training procedure. What we could do is add an option such as model='best', model='all', to enable those use-cases. I still think the default should be model='all' given how cheap it is to compute.

autogluon/task/tabular_prediction/tabular_prediction.py
Innixma: Added

autogluon/task/tabular_prediction/tabular_prediction.py
Innixma: Added

autogluon/task/tabular_prediction/tabular_prediction.py
Innixma: Added

autogluon/task/tabular_prediction/tabular_prediction.py
Innixma: Good language change, added

autogluon/task/tabular_prediction/predictor.py
jwmueller: switching to function sounds good to me

autogluon/task/tabular_prediction/predictor.py
jwmueller: Good point re accuracy/latency, I like the idea of offering model='best', model='all' options, with model='all' as the default.

szha:
Job PR-412-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-412/2/index.html

szha:
Job PR-412-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-412/3/index.html

Issue #411, opened 2020-04-11
Tabular: os.fork() called by MXNet NN model during prediction causing OOM


Innixma:
This OOM error was caused during training of Albert OpenML dataset (Large) on m5.2xlarge. This OOM error has never been encountered before. (Happened 3 times on latest benchmark run on Albert and Airlines datasets).

The error happened during the setup for training the weighted ensemble with bagging and stacking disabled. Therefore, each model needs to do predictions on the validation data in turn (The amount of validation data is very small, less than 10,000, and therefore should not cause issues). For whatever reason, TabularNN is forking the process, which likely leads to a doubling of memory usage.

Question: Why are we running OOM during a stage that should require very minimal memory usage, and why are we forking the process?

Stack Trace, note that the memory check shortly prior to this indicates 32% memory usage:

```
[INFO] [amlb.utils.process:01:07:58.029] CPU Utilization: 41.7%
[INFO] [amlb.utils.process:01:07:58.029] Memory Usage: 32.1%
[INFO] [amlb.utils.process:01:07:58.029] Disk Usage: 15.7%
[ERROR] [amlb.benchmark:01:08:55.237] [Errno 12] Cannot allocate memory
Traceback (most recent call last):
  File "/home/ubuntu/workspace/automlbenchmark/amlb/benchmark.py", line 391, in run
    meta_result = framework.run(self._dataset, task_config)
  File "/home/ubuntu/workspace/automlbenchmark/frameworks/autogluon_nobag/__init__.py", line 4, in run
    return run(*args, **kwargs)
  File "/home/ubuntu/workspace/automlbenchmark/frameworks/autogluon_nobag/exec.py", line 16, in run
    return exec_template.run(dataset=dataset, config=config, parameters=parameters)
  File "/home/ubuntu/workspace/automlbenchmark/frameworks/autogluon/exec_template.py", line 82, in run
    **fit_params
  File "/home/ubuntu/workspace/autogluon/autogluon/task/tabular_prediction/tabular_prediction.py", line 425, in fit
    hyperparameters=hyperparameters, time_limit=time_limits_orig, save_data=cache_data, verbosity=verbosity)
  File "/home/ubuntu/workspace/autogluon/autogluon/utils/tabular/ml/learner/default_learner.py", line 99, in fit
    hyperparameters=hyperparameters)
  File "/home/ubuntu/workspace/autogluon/autogluon/utils/tabular/ml/trainer/auto_trainer.py", line 39, in train
[...]

PR #416, opened 2020-04-16
change tabular-NN dataloader to use thread_pool

jwmueller:
Improve memory usage of tabular-NN DataLoader to prevent issues like this:  https://github.com/awslabs/autogluon/issues/411


szha:
Job PR-416-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-416/1/index.html

Innixma:
Please refer to #422, I believe it results in superior training and inference times then the thread_pool solution.

PR #419, opened 2020-04-17
Relaxing requirements & updating dask

tlienart:
I can confirm this runs on SageMaker for HPO experiments with dask and distributed 2.14, seems a shame to restrict it here to 2.6.

cc: @zhanghang1989 (it needs an update of the docker image, I can do that separately if/when this gets merged)

*Issue #, if available:*

*Description of changes:*


szha:
Job PR-419-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-419/1/index.html

tlienart:
@zhanghang1989 I've updated the requirements in the docker repo

zhanghang1989:
> @zhanghang1989 I've updated the requirements in the docker repo

Awesome. Thanks a lot! I will change the submodule to current master branch.

Innixma:
Why is parmiko now >= 2.4 when it was previously ~= 2.4? Are we ok with not allowing <2.4 and allowing 2.4? @zhanghang1989 @tlienart

tlienart:
>= 2.4 is more permissive than ~= 2.4 and ~=2.4 does not allow < 2.4 (also I was the one who put ~= 2.4). 

As far as I'm concerned it doesn't break AutoGluon and it's better practice in terms of security. It doesn't make sense to have needlessly constraining version bounds unless the more recent versions actually break something AG relies on which doesn't seem to be the case.

Innixma:
It appears it was a misunderstanding with how I interpreted `~=`, I thought it meant any version except 2.4, but I read the official documentation and I was mistaken. This looks good to me!

PR #422, opened 2020-04-20
Optimized tabular neural network inference speed

Innixma:
*Issue #, if available:*
#411 

*Description of changes:*

Optimizes inference time on tabular NN by restricting data loading to 1-core and avoiding os.fork() calls. This results in a flat speedup of ~0.05 - 0.10 seconds per call to inference. For small datasets and online-inference situations, this can lead to a 400% or larger inference speedup, while using less overall CPU with no impact to training times. Additionally, it is suspected this uses far less memory and will resolve potential memory issues during validation and test inference observed in #411.

Benchmark results (_mixed represents this change):

```
                               framework  time_train_s  metric_error  time_infer_s       acc       auc   logloss  bestdiff  loss_rescaled      rank  rank=1_count  rank=2_count  rank=3_count  rank>3_count  error_count
0    autogluon_ag_nn_nostack_numwork0_4h    290.444737      0.338697      1.174810  0.827697  0.842307  0.562289  0.003178       0.026316  2.381579             4             0            33             1            0
1             autogluon_ag_nn_nostack_4h    250.284211      0.338524      1.229560  0.827861  0.842256  0.561840  0.002062       0.054820  2.513158             0             2            36             0            0
2       autogluon_ag_nn_nostack_mixed_4h    243.205263      0.338736      1.173114  0.827857  0.842221  0.562271  0.002979       0.089021  2.539474             0             2            35             1            1
3  autogluon_ag_nn_nostack_threadpool_4h    905.784211      0.338425      1.859594  0.827830  0.842176  0.561519  0.001396       0.082441  2.565789             1             1            33             3            0

```

Note: The 1 error that occurred for _mixed was not related to _mixed, it was an SSH error and the trial was never ran.

### DIFF
--- a/autogluon/utils/tabular/ml/models/tabular_nn/tabular_nn_model.py
+++ b/autogluon/utils/tabular/ml/models/tabular_nn/tabular_nn_model.py
@@ -90,6 +90,7 @@ class TabularNeuralNetModel(AbstractModel):
         self.ctx = mx.cpu()
         self.batch_size = None
         self.num_dataloading_workers = None
+        self.num_dataloading_workers_inference = 0
         self.params_post_fit = None
         self.num_net_outputs = None
         self._architecture_desc = None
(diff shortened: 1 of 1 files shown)


szha:
Job PR-422-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-422/1/index.html

autogluon/utils/tabular/ml/models/tabular_nn/tabular_nn_model.py
```diff
@@ -90,6 +90,7 @@ def __init__(self, path: str, name: str, problem_type: str, objective_func, stop
         self.ctx = mx.cpu()
         self.batch_size = None
         self.num_dataloading_workers = None
+        self.num_dataloading_workers_inference = 0
```
jwmueller, line 93: In the future, should we make this a hyperparameter of the NN model instead?

Innixma: I agree that it could be, if you'd like to make it a hyperparameter we can do so in future.

Issue #425, opened 2020-04-20
Predict proba with changing seed


amirsani:
Is there a way to run the model or train the model with different random seeds?

Thanks

Innixma:
Do you have a particular task in mind (Tabular, Image, etc.)? Regarding Tabular, this is not added at present but could be if you believe there is value in it.

amirsani:
There's value in that it provides an avenue for generating bootstrap confidence intervals in the case of prediction and robustness in the case of training. Note that early stopping has a lot of value in the case of training.

Thanks for the great package and putting the time into this.

Innixma:
Just to confirm, are you referring to this functionality for Tabular? Also, how does early stopping relate to this exactly?

amirsani:
Yes, tabular

As you estimate the performance over seeds, you can use an online estimator to see if the mean is converging to a value that is better than the current performance threshold. You can also use forecasts to see if the rate of improvement in the estimated performance is likely to exceed the stopping threshold. You can further accelerate this estimate by bootstrapping.

amirsani:
How can I set the seed of a model that's been learned with AutoGluon?

Innixma:
Please see PR #429 for the addition of the random_seed parameter to the task.fit() call.

Regarding estimating performance, we have an optimized version of this if you are using bagging which is repeated bagged functionality. Using it will give you both better predictions and more accurate prediction estimates. It is enabled automatically if you specify time_limits and set auto_stack=True, or by specifying num_bagging_folds and num_bagging_sets. This will be far more efficient and produce better results than bootstrap aggregation with different random seeds as an outer loop to AutoGluon.

See also our tutorial on [maximizing model performance](https://autogluon.mxnet.io/tutorials/tabular_prediction/tabular-indepth.html#maximizing-predictive-performance) in Tabular.

Regarding calculating the confidence intervals inside of AutoGluon without an outer loop, this would require getting the predictions of the individual models within a bag rather than the averaged predictions of the bagged models. This is something users do not have access to from the user-facing API at present, partly because of the ease at which it could cause OOM errors on multiclass datasets.

Regarding 

> How can I set the seed of a model that's been learned with AutoGluon?

Could you explain your intent? Do you mean setting the seed prior to fitting the model (#429 implements this)? There is no randomness associated with an already trained model, and thus no seed to set after training.

amirsani:
Thank you for the quick turnaround and apologies for my poor explanation.

[...]

PR #429, opened 2020-04-20
Tabular: Added random_seed parameter

Innixma:
*Issue #, if available:*
#425 

*Description of changes:*

Added random_seed parameter to TabularPrediction. Functionality is identical to previous if random_seed is kept as default value.

### DIFF
--- a/autogluon/utils/tabular/ml/trainer/auto_trainer.py
+++ b/autogluon/utils/tabular/ml/trainer/auto_trainer.py
@@ -11,11 +11,11 @@ logger = logging.getLogger(__name__)
 # This Trainer handles model training details
 class AutoTrainer(AbstractTrainer):
     def __init__(self, path, problem_type, scheduler_options=None, objective_func=None, stopping_metric=None, num_classes=None,
-                 low_memory=False, feature_types_metadata={}, kfolds=0, n_repeats=1, stack_ensemble_levels=0, time_limit=None, save_data=False, verbosity=2):
+                 low_memory=False, feature_types_metadata={}, kfolds=0, n_repeats=1, stack_ensemble_levels=0, time_limit=None, save_data=False, random_seed=0, verbosity=2):
         super().__init__(path=path, problem_type=problem_type, scheduler_options=scheduler_options,
                          objective_func=objective_func, stopping_metric=stopping_metric, num_classes=num_classes, low_memory=low_memory,
                          feature_types_metadata=feature_types_metadata, kfolds=kfolds, n_repeats=n_repeats,
-                         stack_ensemble_levels=stack_ensemble_levels, time_limit=time_limit, save_data=save_data, verbosity=verbosity)
+                         stack_ensemble_levels=stack_ensemble_levels, time_limit=time_limit, save_data=save_data, random_seed=random_seed, verbosity=verbosity)
 
     def get_models(self, hyperparameters={'NN':{},'GBM':{}}, hyperparameter_tune=False, **kwargs):
         return get_preset_models(path=self.path, problem_type=self.problem_type, objective_func=self.objective_func, stopping_metric=self.stopping_metric,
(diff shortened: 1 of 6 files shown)


szha:
Job PR-429-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-429/1/index.html

autogluon/utils/tabular/ml/trainer/auto_trainer.py
```diff
@@ -35,7 +35,7 @@ def train(self, X_train, y_train, X_test=None, y_test=None, hyperparameter_tune=
             y_test = None
         else:
             if (y_test is None) or (X_test is None):
-                X_train, X_test, y_train, y_test = generate_train_test_split(X_train, y_train, problem_type=self.problem_type, test_size=holdout_frac)
+                X_train, X_test, y_train, y_test = generate_train_test_split(X_train, y_train, problem_type=self.problem_type, test_size=holdout_frac, random_state=self.random_seed)
```
jwmueller, line 38: Default `random_state` in `generate_train_test_split` is currently 42? Its default value should probably match the `random_seed` default

jwmueller:
This PR is not intended to fully address the reproducibility issue, right?  If so I'm happy to approve it as just a start.  But if this PR is intended to fully address the linked issue, then the `random_seed` arg of fit() needs to be passed to each of the models, no?

autogluon/utils/tabular/ml/trainer/auto_trainer.py
Innixma: The default value in the function itself is always overridden, so there is no impact, but I will change it to 0... @sgdread So long and thanks for all the fish.

Innixma:
> This PR is not intended to fully address the reproducibility issue, right? If so I'm happy to approve it as just a start. But if this PR is intended to fully address the linked issue, then the `random_seed` arg of fit() needs to be passed to each of the models, no?

Initially I wasn't sure what the exact request was in the issue. This is only intended to enable reproducible data splits (and the ability to randomize them each run, prior it was always the same splits). This does not fully address the linked issue, which is actually more complex than propagating the random_seed value to the models.

szha:
Job PR-429-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-429/2/index.html

jwmueller:
> This is only intended to enable reproducible data splits (and the ability to randomize them each run, prior it was always the same splits). This does not fully address the linked issue, which is actually more complex than propagating the random_seed value to the models.

Right, I'd just make sure to add a cautionary sentence to the documentation here:  
https://github.com/awslabs/autogluon/pull/429/files#diff-fedbb5dec476a83cdce171966541d6a7R257

Otherwise users who don't carefully read it will think this can achieve fully reproducible fit().  For example we could say:

"Caution: This seed only enables reproducible data splits (and the ability to randomize splits in  each run by changing seed values).  This seed is NOT used in the training of individual models, for that you need explicitly set the corresponding seed hyperparameter (usually called 'seed_value') of each individual model."

Innixma:
Added additional warning in documentation as requested.

szha:
Job PR-429-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-429/3/index.html

amirsani:
Would the following be the proper way to use this if I want to set the seeds for each fit,

``
i = 0

hyperparameters = {
               #'NN': {},
               'GBM': {"seed":i},
               'CAT': {"random_seed":i},
               'RF': {"random_state":i},
               'XT': {"random_state":i},
               'KNN': {"random_state":i},
               #'custom': ['GBM']
} 

task.fit(..., hyperparameters=hyperparameters, random_seed=i)
``

Thank you

Innixma:
@amirsani This looks correct to me

Issue #428, opened 2020-04-20
Export feature matrix


braaannigan:
I'd like to be able to export the final feature matrix that autogluon creates.  For example, in the tabular case I'd like to see if I can replicate the model performance using a synthetic dataset created from a tabular GAN model. Is this possible?

Innixma:
This is technically possible, but it depends on what level of preprocessing we are talking about.

Preprocessing in Tabular works at different stages:

Stage 1: The generic preprocessing stage which does automated FE, type inference, ngram generation, etc. This is then passed to the training logic as the 'synthetic dataset' which models are trained with.

Stage 2: Stacker preprocessing stage. This effectively is the same as Stage 1, but with the OOF (Out of Fold) predictions of lower layer stacker models added as features to the dataset.

Stage 3: Then there is the model specific preprocessing. Each ML model does an additional preprocessing stage to get the data in its preferred form. For the NN, this is converting the data to an MXNet representation, for LightGBM, it is converting it to a lgb.Dataset(), for KNN, it is dropping categorical columns and imputing NaNs.

For stage 1 (Generic preprocessing), this is trivial to add for both training data and test data.

For stage 2 (Stacker preprocessing), this is trivial to add for test data. For training data it is complex because between stage 1 and 2, AutoGluon has the ability to drop rows from the training set it views as 'invalid' (and in future for subsampling huge datasets). If you are ok with the training data returned to you missing some amount of rows, this is an easy add (effectively it would return to you the input dataset to the next stacker layer if one were to be trained), otherwise is becomes significantly more complex to implement. It is still possible, although it would be a mash up of OOF prediction generated data (non-bagged inference) and inference-time generated data (bagged inference).

For stage 3 (Model specific preprocessing), this is trivial to add for test data. For training data it shares all the same complexities with stage 2. Furthermore, the output of this stage is highly specialized to the particular model, and likely not very useful to the user who wishes to make their own custom model from the data.

Which of stage 1, stage 2, and stage 3 are you interested in? Is there anything that you are looking for that I missed?

Innixma:
I gave this a shot and have created #431 which should enable the retrieval of feature matrices for stage 1 and 2 as described above. I will notify you when it has been merged. Let me know if this addresses your request!

braaannigan:
Thanks!

Innixma:
Hi @braaannigan,

I believe I have added the functionality you requested.

As an example code:

```
predictor = task.load('my_predictor/')
[...]

Issue #249, opened 2020-01-28
Train data with new/updated features


prasiyer:
Is there a function or pre-defined approach to get a copy of the train data with new and transformed features?

Innixma:
This would be a good feature to have!

At the current time the data transformation is contained internally in AutoGluon, and thus you wouldn't be able to get the data in this form unless you called some internal functions that aren't well documented for users.

Note that there are some transformations of the data done internally that may not work very well if used outside of the context of AutoGluon. For example, we may duplicate rows, cut rows, etc. Additionally, each model has its own specialized preprocessing after the initial generic preprocessing.

Innixma:
This is now achievable via `predictor.transform_features` and `predictor.load_data_internal`

PR #434, opened 2020-04-23
Tabular: Added load_data_internal function to TabularPredictor

Innixma:
*Issue #, if available:*
#249, #428, #429 

*Description of changes:*

Added load_data_internal function to TabularPredictor.

Now when users use transform_features(), they can get the y labels as well by calling these functions.

### DIFF
--- a/autogluon/task/tabular_prediction/predictor.py
+++ b/autogluon/task/tabular_prediction/predictor.py
@@ -553,7 +553,7 @@ class TabularPredictor(BasePredictor):
 
     def load_data_internal(self, dataset='train', return_X=True, return_y=True):
         """
-        Loads the generic internal data representation used during model training.
+        Loads the internal data representation used during model training.
         Individual AutoGluon models like the neural network may apply additional feature transformations that are not reflected in this method.
         This method only applies universal transforms employed by all AutoGluon models.
         This will raise an exception if `cache_data=False` was set in `task.fit()`.
--- a/autogluon/task/tabular_prediction/predictor.py
+++ b/autogluon/task/tabular_prediction/predictor.py
@@ -361,10 +361,14 @@ class TabularPredictor(BasePredictor):
                 For non-bagged mode predictors:
                     The dataset used when not specified is the validation set.
                     This can either be an automatically generated validation set or the user-defined `tuning_data` if passed during fit().
+                    If all parameters are unspecified, then the output is equivalent to `predictor.load_X_val(with_label=False)`.
+                    To get the label values of the output, call `predictor.load_y_val()`.
                     If the original training set is desired, it can be passed in through `dataset`.
                         Warning: Do not pass the original training set if `model` or `base_models` are set. This will result in overfit feature transformation.
                 For bagged mode predictors:
                     The dataset used when not specified is the full training set.
+                    If all parameters are unspecified, then the output is equivalent to `predictor.load_X_train(with_label=False)`.
+                    To get the label values of the output, call `predictor.load_y_train()`.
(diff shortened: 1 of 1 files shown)


szha:
Job PR-434-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-434/1/index.html

jwmueller:
Might these methods (load_X_train, load_y_train, load_X_val, load_y_val) might be better as internal methods (e.g. in Learner)?  I'm just worried if the Predictor offers too many methods, then UX may become more complicated than necessary. 

If these methods become internal, there could just be certain args in `transform_features()` to invoke each (combination of) these.  For example: `dataset` arg in `transform_features()` could have special str options `train` & `validation`.  I also don't see why `transform_features()` shouldn't automatically return a tuple with both X and y (the latter whenever it's available in data, otherwise returned `y = None`; or could even have a boolean arg `transform_label` in `transform_features()` which determines whether y is returned or not). 

Another related point that the documentation might explicitly state somewhere is:   `transform_features()` (and these new load_* methods) allow you to train other 3rd party ML models using the same data representation used internally by AutoGluon. 

If this is the main intended use-case, I'd also consider rethinking whether `transform_features()` returns the train or validation DataFrame by default (when `dataset=None`), as I think returning train might make more sense in this context (or even both train and val).  If the above suggestion is adopted with special `dataset` options `train` & `validation`, then you could also just set `dataset=train` as the default in `transform_features()`.

Innixma:
Addressed comments based on feedback, combined functionality into a single function.

szha:
Job PR-434-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-434/2/index.html

autogluon/task/tabular_prediction/predictor.py
```diff
@@ -361,10 +361,14 @@ def transform_features(self, dataset=None, model=None, base_models=None):
                 For non-bagged mode predictors:
                     The dataset used when not specified is the validation set.
                     This can either be an automatically generated validation set or the user-defined `tuning_data` if passed during fit().
+                    If all parameters are unspecified, then the output is equivalent to `predictor.load_data_internal(dataset='val', return_X=True, return_y=False)[0]`.
```
jwmueller, line 364: why is the validation set the default? I think I'm missing what the use-case would be there.   I thought the main use case would be for users to train 3rd party models on the same data representation, in which case the training set would be the default?

autogluon/task/tabular_prediction/predictor.py
```diff
@@ -547,6 +551,56 @@ def save(self):
         self._learner.save()
         logger.log(20, "TabularPredictor saved. To load, use: TabularPredictor.load(\"%s\")" % self.output_directory)
 
+    def load_data_internal(self, dataset='train', return_X=True, return_y=True):
+        """
+        Loads the generic internal data representation used during model training.
```
jwmueller, line 556: unsure what 'generic' means here?  could be confusing for users

autogluon/task/tabular_prediction/predictor.py
Innixma: It must default to ‘val’ for non-bagged because it would crash if it was ‘train’ and the user asked for weighted ensemble input (because it doesn’t exist)

autogluon/task/tabular_prediction/predictor.py
Innixma: Good catch, removed.

Issue #432, opened 2020-04-21
How to get a trained model instance by its name  in Tabular task?


razmikmelikbekyan:
What is the most suitable way to get the trained model instance including all data preprocessing steps in Tabular prediction? 

Image you have trained autogluon tabular task and now you want to use one of the autogluon models as a separate model instance. The idea is the following. 

```
import autogluon as ag
from autogluon.task import TabularPrediction as task
import pandas as pd

df_train = pd.read_csv('AutogluonModels/titanic/train.csv')

output_directory = 'AutogluonModels/my_models'
predictor = task.fit(
    train_data=df_train,
    holdout_frac=None, 
    label='Survived',
    output_directory=output_directory,
    time_limits=60,
    auto_stack=False,
    stack_ensemble_levels=1
)
```

So now we have a predictor object. It would be nice to have some `get_model(model_name)` method which will return the specified model including its data preprocessing steps. So you can use only this model for refit or something else. This question arise because I wanted to use only the ensemble model from autogluon for cross validation. So I had to refit all the models.

Innixma:
Could you explain the last part, specifically: 

> This question arise because I wanted to use only the ensemble model from autogluon for cross validation. So I had to refit all the models.

What do you mean for cross-validation, are you running AutoGluon 5 times to get a CV score?  Would the ability to return OOF (Out of Fold) predictions solve your issue? This is added in #431. Could you provide pseudo-code that achieves what you require?

To get the model object at present, you would do:

`model = predictor._trainer.load_model('my_model')`

To get X into the form required for inference (Does not work for the training set):
```
X_raw = task.Dataset(...)
X_preprocessed = predictor._learner.transform_features(X_raw)
X = predictor._trainer.get_inputs_to_model(model='my_model', X=X_preprocessed, preprocess=False)
y_pred_proba = model.pred_proba(X)
```

However, this y_pred_proba is still the internal pred proba representation, so you would need further transformation code to make it return equivalent to `predictor.pred_proba(X_raw, model='my_model')`

This will get significantly simpler after #431 is merged.

Innixma:
Note also that the model object does not know all the context required to transform the raw data into its internal representation. To return a model with that knowledge, it would be to effectively return a copy of the Learner object which has all other models removed from it.

razmikmelikbekyan:
[...]

PR #435, opened 2020-04-24
Tabular: Added transform_labels function to TabularPredictor

Innixma:
*Issue #, if available:*
#428, #432 

*Description of changes:*

Added transform_labels function to TabularPredictor

Now users can also transform the labels of the data to the internal representation and back, including the probabilities.

### DIFF
--- a/autogluon/task/tabular_prediction/predictor.py
+++ b/autogluon/task/tabular_prediction/predictor.py
@@ -357,6 +357,7 @@ class TabularPredictor(BasePredictor):
         Transforms dataset features through the AutoGluon feature generator.
         This is useful to gain an understanding of how AutoGluon interprets the dataset features.
         The output of this function can be used to train further models, even outside of AutoGluon.
+        This can be useful for training your own models on the same data representation as AutoGluon.
         Individual AutoGluon models like the neural network may apply additional feature transformations that are not reflected in this method.
         This method only applies universal transforms employed by all AutoGluon models.
         When `dataset=None`, `base_models=[{best_model}], and bagging was enabled during fit():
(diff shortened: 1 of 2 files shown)


szha:
Job PR-435-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-435/1/index.html

autogluon/task/tabular_prediction/predictor.py
```diff
@@ -412,6 +424,50 @@ def transform_features(self, dataset=None, model=None, base_models=None):
...
+            When `True`, the input labels are treated as being in the internal representation and the original representation is outputted.
+        proba : boolean, default = False
+            When `True`, the input labels are treated as probabilities and the output will be the internal representation of probabilities.
+                In this case, it is expected that `labels` be a `numpy.ndarray`.
+                If the `problem_type` is multiclass:
+                    The input column order must be equal to `predictor.class_labels`.
+                    The output column order will be equal to `predictor.class_labels_internal`.
+                    if `inverse=True`, the same logic applies, but with input and output columns interchanged.
+            When `False`, the input labels are treated as actual labels and the output will be the internal representation of the labels.
+                In this case, it is expected that `labels` be a `pandas.Series` or `numpy.ndarray`.
```
jwmueller, line 452: Can you clarify what is a use-case where `proba=True`?

Innixma: Imagine someone uses transform_features and transform_labels to generate the internal representation, then trains a model on that data. They then pass in their test data through transform_features to get prediction probabilities. The pred_proba is currently in the internal representation and they want the probabilities in the original label representation, so they pass it in as predictor.transform_labels(internal_pred_proba, inverse=True, proba=True) to get the original label pred_proba.

proba=True, inverse=False would be quite niche, but for example if the user had their own custom model predictions and they wanted to ensemble them with an internal model to create yet another model, they could convert their external proba estimates to the internal one to align.

PR #445, opened 2020-04-30
Refactoring of schedulers and searchers

mseeger:
- Cleanup of BaseSearcher API, in particular update
- Improve checkpoint/resume, support searcher objects which cannot be pickled
- Improved documentation of FIFOScheduler, HyperbandScheduler

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/autogluon/contrib/enas/enas_scheduler.py
+++ b/autogluon/contrib/enas/enas_scheduler.py
@@ -51,8 +51,8 @@ class ENAS_Scheduler(object):
         self.baseline = None
         self.ema_decay = ema_baseline_decay
         self.searcher = RLSearcher(
-            kwspaces, reward_attribute='accuracy',
-            controller_type=controller_type, prefetch=4, num_workers=4)
+            kwspaces, controller_type=controller_type, prefetch=4,
+            num_workers=4)
         # controller setup
         self.controller = self.searcher.controller
         self.controller_optimizer = mx.gluon.Trainer(
--- a/autogluon/contrib/enas/enas_scheduler.py
+++ b/autogluon/contrib/enas/enas_scheduler.py
@@ -50,8 +50,9 @@ class ENAS_Scheduler(object):
         # create RL searcher/controller
         self.baseline = None
         self.ema_decay = ema_baseline_decay
-        self.searcher = RLSearcher(kwspaces, controller_type=controller_type,
-                                   prefetch=4, num_workers=4)
+        self.searcher = RLSearcher(
+            kwspaces, reward_attribute='accuracy',
+            controller_type=controller_type, prefetch=4, num_workers=4)
         # controller setup
         self.controller = self.searcher.controller
         self.controller_optimizer = mx.gluon.Trainer(
--- a/autogluon/core/task.py
+++ b/autogluon/core/task.py
@@ -40,7 +40,7 @@ class Task(object):
 
     @classmethod
     def set_id(cls, taskid):
-        logger.info('Seting TASK ID: {}'.format(taskid))
+        logger.info('Setting TASK ID: {}'.format(taskid))
         cls.TASK_ID.value = taskid
 
     def __repr__(self):
--- a/autogluon/scheduler/fifo.py
+++ b/autogluon/scheduler/fifo.py
@@ -228,7 +228,7 @@ class FIFOScheduler(TaskScheduler):
         if checkpoint is None:
             checkpoint = self._checkpoint
         if checkpoint is not None:
-            with self.LOCK:
+            with self.log_lock:
                 mkdir(os.path.dirname(checkpoint))
                 save(self.state_dict(), checkpoint)
 
@@ -325,7 +325,7 @@ class FIFOScheduler(TaskScheduler):
         if self._checkpoint is not None or \
                 self.training_history_callback is not None:
             self._add_checkpointing_to_job(job)
-        with self.LOCK:
+        with self.obj_lock:
             self.scheduled_tasks.append(task_dict)
 
     def _clean_task_internal(self, task_dict):
--- a/autogluon/scheduler/fifo.py
+++ b/autogluon/scheduler/fifo.py
@@ -30,40 +30,50 @@ class FIFOScheduler(TaskScheduler):
 
     Parameters
     ----------
-    train_fn : callable
+    train_fn: callable
         A task launch function for training.
-    args : object (optional)
+    args: object (optional)
         Default arguments for launching train_fn.
-    resource : dict
+    resource: dict
         Computation resources. For example, `{'num_cpus':2, 'num_gpus':1}`
-    searcher : str or BaseSearcher
+    searcher: str or BaseSearcher
         Searcher (get_config decisions). If str, this is passed to
         searcher_factory along with search_options.
-    search_options : dict
+    search_options: dict
         If searcher is str, these arguments are passed to searcher_factory.
-    checkpoint : str
+    checkpoint: str
         If filename given here, a checkpoint of scheduler (and searcher) state
         is written to file every time a job finishes.
         Note: May not be fully supported by all searchers.
-    resume : bool
+    resume: bool
         If True, scheduler state is loaded from checkpoint, and experiment
         starts from there.
         Note: May not be fully supported by all searchers.
-    num_trials : int
+    num_trials: int
         Maximum number of jobs run in experiment.
-    time_out : float
+    time_out: float
         If given, jobs are started only until this time_out (wall clock time)
-    reward_attr : str
+    reward_attr: str
         Name of reward (i.e., metric to maximize) attribute in data obtained
         from reporter
-    time_attr : str
+    time_attr: str
         Name of resource (or time) attribute in data obtained from reporter.
         This attribute is optional for FIFO scheduling, but becomes mandatory
         in multi-fidelity scheduling (e.g., Hyperband).
         Note: The type of resource must be int.
-    dist_ip_addrs : list of str
+    dist_ip_addrs: list of str
         IP addresses of remote machines.
-    delay_get_config : bool
+    training_history_callback: callable
+        Callback function func called every time a job finishes, if at least
+        training_history_callback_delta_secs seconds passed since the last
+        recent call. The call has the form:
+            func(self.training_history, self._start_time)
+        Here, self._start_time is time stamp for when experiment started.
+        Use this callback to serialize self.training_history after regular
+        intervals.
+    training_history_callback_delta_secs: float
+        See training_history_callback.
+    delay_get_config: bool
         If True, the call to searcher.get_config is delayed until a worker
         resource for evaluation is available. Otherwise, get_config is called
         just after a job has been started.
@@ -100,6 +110,8 @@ class FIFOScheduler(TaskScheduler):
                  time_out=None, max_reward=None, reward_attr='accuracy',
                  time_attr='epoch',
                  visualizer='none', dist_ip_addrs=None,
+                 training_history_callback=None,
+                 training_history_callback_delta_secs=60,
                  delay_get_config=True):
         super().__init__(dist_ip_addrs)
         if resource is None:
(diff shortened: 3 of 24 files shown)


tlienart:
@zhanghang1989 btw could you make me a contributor so that I could be assigned to such issues?

zhanghang1989:
> @zhanghang1989 or @Innixma btw could you add me as a contributor so that I could be assigned to such issues?

Hey @tlienart , looks like you are not a member of awslabs. Could you submit a ticket to join the organization? I will add you to the maintainer then.

tlienart:
@mseeger looks like a silly error in the skopt unit test:

```
=================================== FAILURES ===================================
______________________________ test_skoptsearcher ______________________________

    def test_skoptsearcher():
        logger.debug('Start testing SKoptSearcher')
        random.seed(1)
        # Create configuration space:
        cs = CS.ConfigurationSpace()
        a = CSH.UniformFloatHyperparameter('a', lower=1e-4, upper=1e-1, log=True) # log-scale float
        b = CSH.UniformFloatHyperparameter('b', lower=-2, upper=0) # float with uniform prior
        c = CSH.UniformIntegerHyperparameter('c', lower=0, upper=1000) # integer
        d = CSH.CategoricalHyperparameter('d', choices=['good','neutral','bad']) # categorical
        cs.add_hyperparameters([a,b,c,d])
        # Determine reward of optimal config:
        optimal_config = cs.sample_configuration()
        optimal_config['a'] = 1e-1
        optimal_config['b'] = 0
        optimal_config['c'] = 1000
        optimal_config['d'] = 'good'
        optimal_reward = toy_reward(optimal_config) # should ~= 7025.58
        # Compare skopt searchers VS random sampling searcher:
        num_configs_totry = 15
>       skopt_searcher = SKoptSearcher(cs) # skopt searcher with all default arguments
E       TypeError: __init__() missing 1 required positional argument: 'reward_attribute'
```

mseeger:
Ah I am sorry, beginner's mistake!

tlienart:
Same error but with the RandomSearcher:

```
------------------
searcher = ag.searcher.RandomSearcher(train_fn.cs)
config = searcher.get_config()
print(config)
------------------

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-20-783528b41f8d> in <module>
----> 1 searcher = ag.searcher.RandomSearcher(train_fn.cs)
      2 config = searcher.get_config()
      3 print(config)

TypeError: __init__() missing 1 required positional argument: 'reward_attribute'
TypeError: __init__() missing 1 required positional argument: 'reward_attribute'
```

mseeger:
Damn! There are all sorts of examples, where RandomSearcher (or maybe others) are created directly, without any scheduler around them.

My cleanup would require you to specify the reward_attribute to the searcher at construction, so that we can just pass reported_result from the reporter to searcher.update directly.

I should rethink this refactoring. I think I can maybe do this without the user having to specify reward_attribute at searcher creation explicitly.

Still, I find it pretty dumb to even show users how to create searcher objects on their own, as they are usually created with the scheduler.

I certainly do not want to roll back to the "old" BaseSearcher.update, since this was a real mess.

mseeger:
This CR also fixes another issue, namely that searcher.configure_scheduler has to be called passing the scheduler, since some searchers need to know stuff from their scheduler. This call is now made in FIFOScheduler.run.
I am also using this in order to inform the searcher what the reward_attribute is.

Apart, we should maybe change the tutorials, and strongly discourage users from creating searcher objects by hand, as this is not necessary, and can go wrong. In the end, a searcher is always used inside a scheduler.

mseeger:
OK, CI now fails for a "timeout" reason, this is not due to my code I suppose

zhanghang1989:
Thanks @mseeger I just re-triggered the CI doc build

mseeger:
@zhanghang1989 I have no idea where these timeouts come from. Is there a way to debug this?

zhanghang1989:
@mseeger the error looks like running image classification hpo example taking too much time. Let me trigger it again. If it still fails, I can test it on my machine.

zhanghang1989:
I have tested this tutorial locally, https://autogluon.mxnet.io/tutorials/image_classification/hpo.html

It deadlocks when running the following code 
```python
search_strategy = 'hyperband'

classifier = task.fit(dataset,
                      net=net,
                      optimizer=optimizer,
                      epochs=epochs,
                      num_trials=2,
                      verbose=False,
                      plot_results=True,
                      ngpus_per_trial=1,
                      grace_period=1)

print('Top-1 val acc: %.3f' % classifier.results[classifier.results['reward_attr']])
```

Looks like some changes in the hyperband may lead to deadlock.

mseeger:
I'll do a diff and see whether any change looks funny.

mseeger:
@zhanghang1989 Could you also look at the changes in hyperband.py and hyperband_stopping.py, whether anything could cause this?

I am not so firm on this locking.

mseeger:
I am in particular confused in *which* methods of HyperbandScheduler I have to lock-protect access to members?
I noted that I did not in add_job, and will do that, but I am just not sure which of all these methods are critical.

zhanghang1989:
Sorry. This looks wired. I will take a look again and get back to you.

mseeger:
We have never experienced such problems with Hyperband.

mseeger:
@zhanghang1989 

There is something odd I found, and maybe this is creating the issue?
- In TaskScheduler, self.LOCK is defined
- In FIFOScheduler, self.log_lock is defined
- In HyperbandScheduler, no extra lock is defined

I never really understood why log_lock is needed in FIFOScheduler, and why we cannot use self.LOCK there?

In any case, in the new HyperbandScheduler code, self.LOCK is used to protect HyperbandScheduler members. Could that be the issue, that I am using LOCK of TaskScheduler for this purpose?

If so, what is the fix? Can I use self.log_lock (defined in FIFOScheduler) to protect HyperbandScheduler member access? Or do I need to define another lock in HyperbandScheduler?

mseeger:
@zhanghang1989
The following is also odd: self.LOCK in TaskScheduler is a single object across all schedulers. It still is meant to protect access to a member, namely self.scheduled_tasks.

In contrast, self.log_lock in FIFOScheduler is a member, so it probably is local to the scheduler object, no?

And that example in that notebooks is creating two schedulers, one after the other. Maybe they share self.LOCK, and this is creating a problem?

But if that is the case, why did we never see issues before?

mseeger:
I will rewrite the HyperbandScheduler code to use log_lock instead of LOCK.

But I am still puzzled with TaskScheduler.LOCK is not a member, but it is supposed to protect members.

mseeger:
@zhanghang1989
Thinking more about this locking situation: I feel the code has too many locks all over the place.

One proposal would be to define just three locks:
- LOCK (class-level) and obj_lock (object-level) in TaskScheduler
- obj_lock (object-level) in BaseSearcher
All other locks would be removed (FIFOScheduler.log_lock, locks in Hyperband terminators).

Maybe it is even possible to cut the lock in BaseSearcher, and get by with locks in schedulers only?

zhanghang1989:
> One proposal would be to define just three locks:

Thanks @mseeger ! Your thoughts on refactoring locking sounds good to me. Sorry that I got a bunch of meetings today. I will take a look at the code.

We do need a lock for the searcher, because we have multi-threading logic in the reporter.

zhanghang1989:
I figured out what is the issue. The deadlock is because `self.LOCK` is used inside the `run_reporter` thread, which is not expected. I am updating the code and hope it will fix the issues.

mseeger:
In my latest commit, I replaced all self.LOCK with self.log_lock, also inside _run_reporter.

I am pretty confused by all this. For example, if you look at the code in HyperbandScheduler._run_reporter, we have:
- Calls of methods of searcher, which internally are protected by a lock in the searcher object
- Calls of methods of terminator, which internally are protected by a lock in the terminator object
- Pieces of code protected by self.log_lock
- Call of _add_training_result, protected by self.log_lock

Naively, I'd have expected the whole body of the while loop in _run_reporter to be protected by a single lock.

But I think I understand too little about multiprocessing.

It would help me to understand which are the methods in FIFOScheduler and HyperbandScheduler which can in fact run concurrently.

zhanghang1989:
The actual cause of the deadlock is because the update `state_dict` function use the `LOCK`, while the saving function `save` inherited from FIFOScheduler also require the same lock. Therefore there is a deadlock. I have updated the code. Hopefully the issue will be resolved.

mseeger:
When you look at commit e602893004cd8a5ca8961a43841169cd4db86bfa

I already back then reverted FIFOScheduler and HyperbandScheduler to their old versions, and this did not solve the issue either ...

zhanghang1989:
> When you look at commit [e602893](https://github.com/awslabs/autogluon/commit/e602893004cd8a5ca8961a43841169cd4db86bfa)
> 
> I already back then reverted FIFOScheduler and HyperbandScheduler to their old versions, and this did not solve the issue either ...

Yes, despite some use cases of the `LOCK` in `run_reporter` are not idea. The actually cause of the deadlock is as I mentioned in previous reply, during the saving and state_dict. This is a really hard to figure out.

zhanghang1989:
> > When you look at commit [e602893](https://github.com/awslabs/autogluon/commit/e602893004cd8a5ca8961a43841169cd4db86bfa)
> > I already back then reverted FIFOScheduler and HyperbandScheduler to their old versions, and this did not solve the issue either ...
> 
> Yes, despite some use cases of the `LOCK` in `run_reporter` are not idea. The actually cause of the deadlock is as I mentioned in previous reply, during the saving and state_dict. This is a really hard to figure out.

To be specific:
https://github.com/awslabs/autogluon/blob/e602893004cd8a5ca8961a43841169cd4db86bfa/autogluon/scheduler/fifo.py#L208-L210
https://github.com/awslabs/autogluon/blob/e602893004cd8a5ca8961a43841169cd4db86bfa/autogluon/scheduler/hyperband.py#L331-L340

mseeger:
Ah OK, what you say about state_dict makes sense. That explains things.

The root cause seems that we define too few locks, and use them in too many different places.

Was this save method called in the tutorial?

mseeger:
Do you think once you fixed this, the code is sound? Or does it need to be refactored further, turning some of the static locks into object locks?

I'd like to avoid having to do this, since I am just not enough of an expert on this. But I also do not like to run into trouble down the line.

zhanghang1989:
I agree. We can make some separate object locks to protect specific variables. 

This PR looks good to me and I believe it should pass the CI.

Would you prefer to do the lock refactoring now or we should do it later?

szha:
Job PR-445-15 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-445/15/index.html

mseeger:
I'll do some further things, in particular move the locks in the terminator classes. But I will not refactor the lock situation in the searchers.

I still lack understanding here. What is the policy for using locks?
- I could say I protect each member by having a (object) lock in each subclass (as we do right now for FIFOScheduler and HyperbandScheduler). This would avoid such deadlocks
- But should we not still be worried about code paths which contain several different pieces of lock-protected code, sometimes even different locks (because code is partly in super classes)? When I write such code, I think of the whole code path as atomic, so the members do not change. This would call for a single lock acquired at the start of the code path, and released at the end

mseeger:
One thing I will do:
- Remove LOCK in Hyperband*_Manager
- Instead, use the object lock in HyperbandScheduler to protect all calls to self.terminator

This is along the lines that less locks are better than more locks

mseeger:
@zhanghang1989 
I will also do one thing I always wanted to do: I'll remove the searcher argument in _run_reporter.
Namely, _run_reporter has access to searcher via self.searcher anyway, right? On the other hand, it cannot be made independent of the scheduler, it also needs access to _training_history.

szha:
Job PR-445-16 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-445/16/index.html

mseeger:
OK, if there are no further objections, this can go in.

szha:
Job PR-445-17 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-445/17/index.html

Issue #399, opened 2020-04-01
LightGBM error special JSON characters in feature name


braaannigan:
Hi - great library everyone!

I'm doing tabular prediction with news article data.  The LightGBM model doesn't run with the following error:
```
LightGBMError: Do not support special JSON characters in feature name.
Do not support special JSON characters in feature name.
```
The upstream issue in LightGBM is here:https://github.com/microsoft/LightGBM/issues/2455

Basically they check for special json characters .e.g ```[],{}":``` in features names and throw an error if found. Could autogluon check for these characters and remove any offending features?

Liam

Innixma:
Good catch, thanks for the info! Just to clarify, LightGBM fails in this scenario, but the rest of the models train successfully and you are able to get a result from AutoGluon?

If so, I think I can look into fixing this, the question is if I can get it to still use the features rather than dropping them entirely (I think I can by clever renaming).

braaannigan:
Yep, everything else works fine and I get a result at the end

Innixma:
To aid in testing to ensure I am covering all of the failure cases, could you provide a sample training data of the news article data you are experiencing the problem with or the location which you obtained the data? If you wish to keep it private that is fine, but it could help accelerating the fix on my end.

braaannigan:
Sure - first 3 rows of csv are enough to trigger it - see attached zip. 
[err.tar.gz](https://github.com/awslabs/autogluon/files/4425329/err.tar.gz)

Is there a way to access errors in the underlying models using the %debug in ipython?

oguzhangur96:
Hi, 

I have encountred the same problem. Fixed it before the training begins with a regular expression and lambda function (pandas). Object named "_data_" must be a dataframe.

```python
import re
data = data.rename(columns = lambda x:re.sub('[^A-Za-z0-9_]+', '', x))
```

Could provide data/column headers if needed.

Innixma:
Thanks for the info @oguzhangur96! Right now the main difficulty here is to ensure that column names are always converted when entering the model (whether for feature importances, fit, inference) and also inversely converted back to the original upon leaving. This seems to be something that LightGBM itself should handle, and I don't have a good understanding of why they haven't done so.

[...]

PR #451, opened 2020-05-06
Tabular: Fixed crash in LightGBM with special characters in column names

Innixma:
*Issue #, if available:*
#399 

*Description of changes:*

Fixed crash in LightGBM with special characters in column names.

This fix makes LightGBM work for problems with text features, as previously the feature engineered column names from AutoGluon when text was present contained these special characters and crashed LightGBM.

### DIFF
--- a/autogluon/utils/tabular/ml/models/lgb/lgb_model.py
+++ b/autogluon/utils/tabular/ml/models/lgb/lgb_model.py
@@ -2,6 +2,7 @@ import gc
 import logging
 import os
 import random
+import re
 import time
 import warnings
 
@@ -34,6 +35,7 @@ class LGBModel(AbstractModel):
 
         self.eval_metric_name = self.stopping_metric.name
         self.is_higher_better = True
+        self._internal_feature_map = None
 
     def _set_default_params(self):
         default_params = get_param_baseline(problem_type=self.problem_type, num_classes=self.num_classes)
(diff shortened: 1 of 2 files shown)


szha:
Job PR-451-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-451/1/index.html

autogluon/utils/tabular/ml/models/lgb/lgb_model.py
```diff
@@ -144,14 +146,36 @@ def predict_proba(self, X, preprocess=True):
             else:  # Should this ever happen?
                 return y_pred_proba[:, 1]
 
+    def preprocess(self, X, is_train=False):
+        X = super().preprocess(X=X)
+
+        if is_train:
+            requires_internal_map = False
+            for column in X.columns:
+                new_column = re.sub(r'[",:{}[\]]', '', column)
```
jwmueller, line 155: This is the full set of lGBM-breaking chars?  If not consider adding TODO that there may be others as well

autogluon/utils/tabular/ml/models/lgb/lgb_model.py
```diff
@@ -144,14 +146,36 @@ def predict_proba(self, X, preprocess=True):
...
+                if new_column != column:
+                    requires_internal_map = True
+                if requires_internal_map:
```
jwmueller, line 158: this line not strictly necessary

autogluon/utils/tabular/ml/models/lgb/lgb_model.py
```diff
@@ -144,14 +146,36 @@ def predict_proba(self, X, preprocess=True):
...
+                    break
+            if requires_internal_map:
+                self._internal_feature_map = {feature: i for i, feature in enumerate(list(X.columns))}
```
jwmueller, line 161: Is renaming features as integers `i` the best strategy if user wants to work with lightGBM themselves post-autogluon fit?  Eg. they might want to use lightGBM's feature-importance functionalities.  So perhaps it's better to use map that changes names as little as possible such as your previous: `column` -> `new_column = re.sub(r'[",:{}[\]]', '', column)` (if name uniqueness gets violated, could additionally paste index `i` to the end of `new_column`).

autogluon/utils/tabular/ml/models/lgb/lgb_model.py
Innixma: This is the code in LGBM that checks, and I've tested many other characters without issue: https://github.com/microsoft/LightGBM/pull/2557/files#diff-c199c652de48f026fa699c5775b6eab1R934

Innixma: non-ASCII are also not supported, but that is probably something we should handle at a higher level, as many models may have issues in that regard.

autogluon/utils/tabular/ml/models/lgb/lgb_model.py
Innixma: Functionality is purely internal, permutation shuffling feature importance output is still in the original names if you look at how I altered the feature importance function.

I have added parameter use_original_feature_names to get_model_feature_importance (the LightGBM specialized output), which advanced users can leverage if necessary. If they want to use LightGBM's specialized feature importance, that is a niche case that isn't representative of most users (Permutation shuffling is better). Also, the amount of problems where this conversion occurs is rare (I've only seen it for NLP problems).

I'd rather have it clear that the column names were altered rather than subtly changing the names and giving people the wrong idea if they are planning to do something programmatically. By giving them the option to convert back to exact original columns, this enables a more seamless experience between different models.

autogluon/utils/tabular/ml/models/lgb/lgb_model.py
Innixma: Good catch, cleaned.

szha:
Job PR-451-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-451/2/index.html

PR #455, opened 2020-05-10
[WIP] autogluon_core separation

Innixma:
*Issue #, if available:*

*Description of changes:*

Separating core functionality into autogluon_core directory.

TODO:

1. Ensure full separation is achieved
2. Fix documentation
3. Create separated setup.py files
4. Ensure autogluon_core works in isolation
5. Ensure files are in proper location


szha:
Job PR-455-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-455/2/index.html

szha:
Job PR-455-7 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-455/7/index.html

autogluon/contrib/enas/enas.py
```diff
@@ -2,7 +2,7 @@
 import collections
 import mxnet as mx
 from mxnet import gluon
-from ...core.space import Categorical, Space, _strip_config_space
+from autogluon_core.core.space import Categorical, Space, _strip_config_space
```
mseeger, line 5: +1 for absolute instead of relative imports, the latter never really worked for me.

Are you sure you want to rename autogluon -> autogluon_core?
Would it not rather be useful to put the non-core parts a little elsewhere, or in a submodule?

Just thinking the core should be mainly what AG is about (and maybe it should include most of AG tabular, except esoteric base learners).
If so, it would be good to call it just "autogluon", no?

It maybe does not matter so much, since the user does not see this ...

autogluon/utils/tabular/ml/models/abstract/abstract_model.py
```diff
@@ -16,8 +16,8 @@
 from .... import metrics
 from ....utils.loaders import load_pkl
 from ....utils.savers import save_pkl, save_json
-from ......core import Space, Categorical, List, NestedSpace
```
mseeger, line 19: Did you ever consider changing all relative to absolute imports?

Maybe I am not good enough in Python, but every time I try to use relative imports, it does not work for me, so for example all our (still to be contributed) HPO code uses absolute imports.

autogluon/contrib/enas/enas.py
Innixma: I am not switching away from relative imports, rather I am separating autogluon into autogluon and autogluon_core.

Things in autogluon will relative import things in autogluon, and absolute import things in autogluon_core, and autogluon_core will relative import things in autogluon_core.

I was originally against relative imports when it was first proposed to me, but I am in support of them now. They make refactoring much simpler and cleaner, and reduce the impact of name collisions.

Regarding what core is: Core is just core functionality for HPO / Searchers / etc. Tabular will not exist in it (nor CV or NLP) at least initially.

autogluon/utils/tabular/ml/models/abstract/abstract_model.py
Innixma: See my above comment, relative imports should be present at all times when possible. I am still keeping all relative imports with this change, but treating autogluon and autogluon_core as separate packages from an import standpoint (thus absolute rather than relative).

tlienart:
@Innixma what is the status on this?

Innixma:
> @Innixma what is the status on this?

Currently doing a feature generator refactor, then I will be prioritizing this for end of July / early-mid August.

Innixma:
Closing as #667 continues this effort.

PR #470, opened 2020-05-16
Unify scheduler creation in different tasks. Basic fix to FIFOSchedul…

mseeger:
…er, HyperbandScheduler

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/autogluon/scheduler/hyperband.py
+++ b/autogluon/scheduler/hyperband.py
@@ -3,11 +3,14 @@ import logging
 import threading
 import numpy as np
 import multiprocessing as mp
+import os
 
 from .fifo import FIFOScheduler
 from .hyperband_stopping import HyperbandStopping_Manager
 from .hyperband_promotion import HyperbandPromotion_Manager
 from .reporter import DistStatusReporter
+from ..utils import load
+
 
 __all__ = ['HyperbandScheduler',
            'HyperbandStopping_Manager',
@@ -186,9 +189,11 @@ class HyperbandScheduler(FIFOScheduler):
         _search_options['resource_attribute'] = time_attr
         _search_options['min_epochs'] = grace_period
         _search_options['max_epochs'] = max_t
+        # Pass resume=False here. Resume needs members of this object to be
+        # created
         super().__init__(
             train_fn=train_fn, args=args, resource=resource, searcher=searcher,
-            search_options=_search_options, checkpoint=checkpoint, resume=resume,
+            search_options=_search_options, checkpoint=checkpoint, resume=False,
             num_trials=num_trials, time_out=time_out, max_reward=max_reward,
             reward_attr=reward_attr, time_attr=time_attr,
             visualizer=visualizer, dist_ip_addrs=dist_ip_addrs,
@@ -236,6 +241,16 @@ class HyperbandScheduler(FIFOScheduler):
         # This lock protects both _running_tasks and terminator, the latter
         # does not define its own lock
         self._hyperband_lock = mp.Lock()
+        if resume:
+            assert checkpoint is not None, \
+                "Need checkpoint to be set if resume = True"
+            if os.path.isfile(checkpoint):
+                self.load_state_dict(load(checkpoint))
+            else:
+                msg = f'checkpoint path {checkpoint} is not available for resume.'
+                logger.exception(msg)
+                raise FileExistsError(msg)
+
 
     def add_job(self, task, **kwargs):
         """Adding a training task to the scheduler.
@@ -330,12 +345,9 @@ class HyperbandScheduler(FIFOScheduler):
     def _run_reporter(self, task, task_job, reporter):
         last_result = None
         last_updated = None
+        task_key = str(task.task_id)
         while not task_job.done():
             reported_result = reporter.fetch()
-            # Time since start of experiment
-            elapsed_time = self._elapsed_time()
-            reported_result['time_since_start'] = elapsed_time
-            task_key = str(task.task_id)
             if 'traceback' in reported_result:
                 # Evaluation has failed
                 logger.exception(reported_result['traceback'])
@@ -353,6 +365,13 @@ class HyperbandScheduler(FIFOScheduler):
                     # Cleanup
                     del self._running_tasks[task_key]
                 break
+            if len(reported_result) == 0:
+                # An empty dict should just be skipped
+                logger.warning("Skipping empty dict received from reporter")
+                continue
+            # Time since start of experiment
+            elapsed_time = self._elapsed_time()
+            reported_result['time_since_start'] = elapsed_time
 
             # Call before _add_training_results, since we may be able to report
             # extra information from the bracket
@@ -431,14 +450,6 @@ class HyperbandScheduler(FIFOScheduler):
                 # variant. It means that the *task* terminates, while the
                 # evaluation of the config is just paused
                 last_result['terminated'] = True
-                resource = int(reported_result[self._time_attr])
-                if self.type == 'stopping' or resource >= self.max_t:
-                    act_str = 'terminating'
-                else:
-                    act_str = 'pausing'
-                logger.debug(
-                    'Stopping task ({} evaluation, resource = {}):\n{}'.format(
-                        act_str, resource, task))
                 with self._hyperband_lock:
                     self.terminator.on_task_remove(task)
                     # Cleanup
--- a/autogluon/scheduler/hyperband.py
+++ b/autogluon/scheduler/hyperband.py
@@ -150,7 +150,7 @@ class HyperbandScheduler(FIFOScheduler):
     ...     print('lr: {}, wd: {}'.format(args.lr, args.wd))
     ...     for e in range(10):
     ...         dummy_accuracy = 1 - np.power(1.8, -np.random.uniform(e, 2*e))
-    ...         reporter(epoch=e, accuracy=dummy_accuracy, lr=args.lr, wd=args.wd)
+    ...         reporter(epoch=e+1, accuracy=dummy_accuracy, lr=args.lr, wd=args.wd)
     >>> scheduler = ag.scheduler.HyperbandScheduler(train_fn,
     ...                                             resource={'num_cpus': 2, 'num_gpus': 0},
     ...                                             num_trials=20,
--- a/autogluon/scheduler/hyperband.py
+++ b/autogluon/scheduler/hyperband.py
@@ -26,6 +26,11 @@ class HyperbandScheduler(FIFOScheduler):
     using multiple brackets, task allocation to bracket is done randomly 
     based on a softmax probability.
 
+    Note: This scheduler requires both reward and resource (time) to be
+    returned by the reporter. Here, resource (time) values must be positive
+    int. If time_attr == 'epoch', this should be the number of epochs done,
+    starting from 1 (not the epoch number, starting from 0).
+
     Parameters
     ----------
     train_fn : callable
@@ -56,7 +61,7 @@ class HyperbandScheduler(FIFOScheduler):
         from reporter
     time_attr: str
         Name of resource (or time) attribute in data obtained from reporter.
-        Note: The type of resource must be int.
+        Note: Resource values must be positive int.
     max_t: int
         Maximum resource (see time_attr) to be used for a job. Together with
         grace_period and reduction_factor, this is used to determine rung
--- a/autogluon/scheduler/hyperband.py
+++ b/autogluon/scheduler/hyperband.py
@@ -12,9 +12,7 @@ from .reporter import DistStatusReporter
 from ..utils import load
 
 
-__all__ = ['HyperbandScheduler',
-           'HyperbandStopping_Manager',
-           'HyperbandPromotion_Manager']
+__all__ = ['HyperbandScheduler']
 
 logger = logging.getLogger(__name__)
 
@@ -42,31 +40,31 @@ class HyperbandScheduler(FIFOScheduler):
     searcher : str or BaseSearcher
         Searcher (get_config decisions). If str, this is passed to
         searcher_factory along with search_options.
-    search_options: dict
+    search_options : dict
         If searcher is str, these arguments are passed to searcher_factory.
-    checkpoint: str
+    checkpoint : str
         If filename given here, a checkpoint of scheduler (and searcher) state
         is written to file every time a job finishes.
         Note: May not be fully supported by all searchers.
-    resume: bool
+    resume : bool
         If True, scheduler state is loaded from checkpoint, and experiment
         starts from there.
         Note: May not be fully supported by all searchers.
-    num_trials: int
+    num_trials : int
         Maximum number of jobs run in experiment.
-    time_out: float
+    time_out : float
         If given, jobs are started only until this time_out (wall clock time)
-    reward_attr: str
+    reward_attr : str
         Name of reward (i.e., metric to maximize) attribute in data obtained
         from reporter
-    time_attr: str
+    time_attr : str
         Name of resource (or time) attribute in data obtained from reporter.
-        Note: Resource values must be positive int.
-    max_t: int
+        Note: The type of resource must be positive int.
+    max_t : int
         Maximum resource (see time_attr) to be used for a job. Together with
         grace_period and reduction_factor, this is used to determine rung
         levels in Hyperband brackets.
-    grace_period: int
+    grace_period : int
         Minimum resource (see time_attr) to be used for a job.
     reduction_factor : int (>= 2)
         Parameter to determine rung levels in successive halving (Hyperband).
@@ -74,7 +72,7 @@ class HyperbandScheduler(FIFOScheduler):
         Number of brackets to be used in Hyperband. Each bracket has a different
         grace period, all share max_t and reduction_factor.
         If brackets == 1, we just run successive halving.
-    training_history_callback:
+    training_history_callback : callable
         Callback function func called every time a job finishes, if at least
         training_history_callback_delta_secs seconds passed since the last
         recent call. The call has the form:
@@ -82,9 +80,9 @@ class HyperbandScheduler(FIFOScheduler):
         Here, self._start_time is time stamp for when experiment started.
         Use this callback to serialize self.training_history after regular
         intervals.
-    training_history_callback_delta_secs: float
+    training_history_callback_delta_secs : float
         See training_history_callback.
-    delay_get_config: bool
+    delay_get_config : bool
         If True, the call to searcher.get_config is delayed until a worker
         resource for evaluation is available. Otherwise, get_config is called
         just after a job has been started.
@@ -122,7 +120,7 @@ class HyperbandScheduler(FIFOScheduler):
         the acquisition function (for model-based variant), which operates
         at level max_t. On the other hand, it decreases the variance of the
         latent process there.
-    searcher_data: str
+    searcher_data : str
         Relevant only if a model-based searcher is used, and if train_fn is such
         that we receive results (from the reporter) at each successive resource
         level, not just at the rung levels.
@@ -183,8 +181,8 @@ class HyperbandScheduler(FIFOScheduler):
                  dist_ip_addrs=None,
                  keep_size_ratios=False,
                  maxt_pending=False,
-                 searcher_data = 'rungs',
-                 do_snapshots = False):
+                 searcher_data='rungs',
+                 do_snapshots=False):
         # Adjoin information about scheduler to search_options
         if search_options is None:
             _search_options = dict()
--- a/autogluon/task/base/base_task.py
+++ b/autogluon/task/base/base_task.py
@@ -91,19 +91,47 @@ searcher_for_hyperband_strategy = {
 
 
 def compile_scheduler_options(
-        search_strategy, nthreads_per_trial, ngpus_per_trial, checkpoint,
-        num_trials, time_out, resume, visualizer, time_attr, reward_attr,
-        search_options, dist_ip_addrs, **kwargs):
+        scheduler_options, search_strategy, search_options, nthreads_per_trial,
+        ngpus_per_trial, checkpoint, num_trials, time_out, resume, visualizer,
+        time_attr, reward_attr, dist_ip_addrs, epochs):
+    """
+    Updates a copy of scheduler_options (scheduler-specific options, can be
+    empty) with general options. The result can be passed to __init__ of the
+    scheduler.
+
+    :param scheduler_options:
+    :param search_strategy:
+    :param search_options:
+    :param nthreads_per_trial:
+    :param ngpus_per_trial:
+    :param checkpoint:
+    :param num_trials:
+    :param time_out:
+    :param resume:
+    :param visualizer:
+    :param time_attr:
+    :param reward_attr:
+    :param dist_ip_addrs:
+    :param kwargs:
+    :param epochs:
+    :return:
+
+    """
+    if scheduler_options is None:
+        scheduler_options = dict()
+    else:
+        assert isinstance(scheduler_options, dict)
     assert isinstance(search_strategy, str)
+    if search_options is None:
+        search_options = dict()
     if visualizer is None:
         visualizer = 'none'
     if time_attr is None:
         time_attr = 'epoch'
     if reward_attr is None:
         reward_attr = 'accuracy'
-    if search_options is None:
-        search_options = dict()
-    scheduler_options = {
+    scheduler_options = copy.copy(scheduler_options)
+    scheduler_options.update({
         'resource': {
             'num_cpus': nthreads_per_trial, 'num_gpus': ngpus_per_trial},
         'searcher': search_strategy,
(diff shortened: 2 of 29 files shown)


mseeger:
The arguments checkpoint (str), resume (bool) have a fixed role in schedulers. In object_detection, resume (str) is used for something else, renamed this one resume_model.

szha:
Job PR-470-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-470/2/index.html

autogluon/task/tabular_prediction/tabular_prediction.py
```diff
@@ -65,7 +65,8 @@ def fit(train_data, label, tuning_data=None, time_limits=None, output_directory=
             holdout_frac=None, num_bagging_folds=0, num_bagging_sets=None, stack_ensemble_levels=0,
             hyperparameters=None, num_trials=None, search_strategy='random', search_options=None,
             nthreads_per_trial=None, ngpus_per_trial=None, dist_ip_addrs=None, visualizer='none',
-            verbosity=2, **kwargs):
+            verbosity=2, epochs=None, grace_period=None, reduction_factor=None, brackets=None,
```
Innixma, line 68: ditto previous comment. Please add a new argument scheduler_options, these arguments are too niche to be their own dedicated arguments.

autogluon/task/text_classification/text_classification.py
```diff
@@ -49,14 +49,18 @@ def fit(dataset='SST',
...
+            search_options=None,
             time_limits=None,
             resume=False,
             checkpoint='checkpoint/exp1.ag',
             visualizer='none',
             num_trials=2,
-            dist_ip_addrs=[],
+            dist_ip_addrs=None,
             grace_period=None,
+            reduction_factor=None,
```
Innixma, line 60: ditto previous comment

autogluon/task/base/base_task.py
```diff
@@ -17,8 +21,18 @@
...
+    'rl': RLScheduler}
+
+
+def create_scheduler(train_fn, search_strategy, scheduler_options):
+    if isinstance(search_strategy, str):
+        scheduler_cls = schedulers[search_strategy.lower()]
+    else:
+        assert callable(search_strategy)
+        scheduler_cls = search_strategy
+        scheduler_options = copy.copy(scheduler_options)
```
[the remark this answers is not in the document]
mseeger: The idea here is just to avoid the original scheduler_options to be modified.

autogluon/task/base/base_task.py
```diff
@@ -74,3 +82,70 @@ def run_fit(cls, train_fn, search_strategy, scheduler_options):
...
+        search_strategy, nthreads_per_trial, ngpus_per_trial, checkpoint,
+        num_trials, time_out, resume, visualizer, time_attr, reward_attr,
+        search_options, dist_ip_addrs, **kwargs):
+    assert isinstance(search_strategy, str)
+    if visualizer is None:
+        visualizer = 'none'
+    if time_attr is None:
+        time_attr = 'epoch'
+    if reward_attr is None:
+        reward_attr = 'accuracy'
```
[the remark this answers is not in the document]
mseeger: Actually, I will follow your advice below and use a dict scheduler_options passed into run, instead of all these single ones. Then, defaults are dealt with by the scheduler itself.

autogluon/task/base/base_task.py
```diff
@@ -74,3 +82,70 @@ def run_fit(cls, train_fn, search_strategy, scheduler_options):
...
+        'dist_ip_addrs': dist_ip_addrs,
+        'delay_get_config': kwargs.get('delay_get_config', True)}
+    searcher = searcher_for_hyperband_strategy.get(search_strategy)
+    if searcher is not None:
+        # Additional arguments for HyperbandScheduler
+        # Note: We can have epochs=None in kwargs, or epochs
+        # missing in kwargs
+        epochs = kwargs.get('epochs')
+        if epochs is None:
+            epochs = 20
```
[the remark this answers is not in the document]
mseeger: Yes I will change this, the defaults should be dealt with in the scheduler

autogluon/task/image_classification/image_classification.py
```diff
@@ -153,24 +156,34 @@ def fit(dataset,
...
             Options include: 'random' (random search), 'skopt' (SKopt Bayesian optimization), 'grid' (grid search), 'hyperband' (Hyperband), 'rl' (reinforcement learner)
         search_options : dict
             Auxiliary keyword arguments to pass to the searcher that performs hyperparameter optimization.
         resume : bool
-            If a model checkpoint file exists, model training will resume from there when specified.
+            If True, the hyperparameter search is started from state loaded from
+            os.path.join(output_directory, 'exp1.ag')
         dist_ip_addrs : list
             List of IP addresses corresponding to remote workers, in order to leverage distributed computation.
+        grace_period : int
```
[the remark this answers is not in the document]
mseeger: I actually wanted to suggest that. I'll do this, it is a lot better, also because then the scheduler deals with defaults.

mseeger:
My changes to the schedulers:
- _run_reporter reacts to empty result from reporter.fetch(), by skipping the result
- resume=True did not properly work in HyperbandScheduler. This is fixed, I have a unit test
   which I will contribute with our large contribution

szha:
Job PR-470-5 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-470/5/index.html

autogluon/task/image_classification/image_classification.py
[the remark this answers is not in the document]
mseeger: done

autogluon/task/base/base_task.py
```diff
@@ -17,8 +21,18 @@
...
+
+
+def create_scheduler(train_fn, search_strategy, scheduler_options):
+    if isinstance(search_strategy, str):
+        scheduler_cls = schedulers[search_strategy.lower()]
+    else:
+        assert callable(search_strategy)
+        scheduler_cls = search_strategy
+        scheduler_options = copy.copy(scheduler_options)
+        scheduler_options['searcher'] = 'random'
```
jwmueller, line 34: why is this always set to 'random'?

autogluon/task/tabular_prediction/tabular_prediction.py
```diff
@@ -264,6 +285,10 @@ def fit(train_data, label, tuning_data=None, time_limits=None, output_directory=
             Higher levels correspond to more detailed print statements (you can set verbosity = 0 to suppress warnings).
             If using logging, you can alternatively control amount of information printed via `logger.setLevel(L)`,
             where `L` ranges from 0 to 50 (Note: higher values of `L` correspond to fewer print statements, opposite of verbosity levels)
+        epochs : int
```
jwmueller, line 288: `epochs` should live in scheduler_options as it is an argument for the Hyperband Scheduler. It also should be more precisely named (e.g. `hyperband_epochs`), as epochs could refer to many other things inside autogluon.

autogluon/task/tabular_prediction/tabular_prediction.py
```diff
@@ -264,6 +285,10 @@ def fit(train_data, label, tuning_data=None, time_limits=None, output_directory=
...
+            Maximum number of epochs (max_t of HyperbandScheduler)
+        scheduler_options : dict
+            Extra arguments passed to __init__ of scheduler
```
jwmueller, line 291: I think documentation of `scheduler_options` could briefly explain what scheduler even is, as many users may be unfamiliar with it, eg: 

"Extra arguments passed to __init__ of scheduler to configure the orchestration of training jobs during hyperparameter-tuning.  This is ignored if `hyperparameter_tune=False`.

docs/tutorials/image_classification/hpo.md
```diff
@@ -124,6 +124,10 @@ Here is an example of using an early stopping scheduler :class:`autogluon.schedu
 
 ```{.python .input}
 search_strategy = 'hyperband'
+scheduler_options = {
+    'grace_period': 1,
```
jwmueller, line 128: add brief comment in the tutorial explaining what is 'grace_period', 'reduction_factor', 'brackets'.

jwmueller:
I agree with most of Nick's comments.  My only other concern is that the documentation for the tasks (e.g. TabularPrediction, ObjectDetection, ...) should be more clear about what the scheduler is.   For example: "Extra arguments passed to `__init__` of Scheduler (subclasses of `autogluon.scheduler.TaskScheduler`) used to configure the orchestration of training jobs during hyperparameter-tuning."

autogluon/task/base/base_task.py
mseeger: This was like that in the current code, I did not change it.
I do not know who would call 'fit' with search_strategy being not a str (then, it must be a scheduler class).

We could remove this altogether, because passing a non-str as search_strategy would fail before in all these cases (and did so in the old code). Shall I throw this out, and insist on search_strategy being a str? The comment says so anyway.

autogluon/task/tabular_prediction/tabular_prediction.py
mseeger: epochs is not just used in the (HB) scheduler, but also passed to the training function, both in image_classification and in object_detection.

I could remove epochs in those task cases where it is not used outside of the (HB) scheduler, OK?

I just did not want to break external code by changing things.

The argument in HyperbandScheduler is called max_t, not epochs. But for DNN training, this is equal to epochs

autogluon/task/tabular_prediction/tabular_prediction.py
mseeger: OK

docs/tutorials/image_classification/hpo.md
mseeger: OK

autogluon/task/tabular_prediction/tabular_prediction.py
mseeger: There is one danger to this. If somebody codes up a new module under task, has an epochs argument, and forgets to copy that to scheduler_options['max_t'], then the scheduler would be inconsistent with the training function.

I'll make the change requested, and add a comment, so hopefully this will not happen.

mseeger:
@jwmueller Have a look at my change, I think it does what you recommend.
This means the epochs argument is removed in tabular_prediction. If you want ton configure HB there, you'd have to use scheduler_options['max_t'].

The other 3 cases are NN training, and epochs is used outside of HB scheduler, so they pass epochs into compile_scheduler_options, where 'max_t' is set accordingly.

I also added a comment to compile_scheduler_options. If somebody codes a new thing under task, they'll hopefully start from the 3 examples already there, and understand they need to pass epochs to compile_scheduler_options.

szha:
Job PR-470-6 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-470/6/index.html

autogluon/scheduler/hyperband.py
```diff
@@ -236,6 +241,16 @@ def __init__(self, train_fn, args=None, resource=None,
         # This lock protects both _running_tasks and terminator, the latter
         # does not define its own lock
         self._hyperband_lock = mp.Lock()
+        if resume:
```
zhreshold, line 244: shall we move this code block before parent init with resume being the validated boolean?

autogluon/task/image_classification/image_classification.py
```diff
@@ -153,24 +152,27 @@ def fit(dataset,
...
+        scheduler_options : dict
+            Extra arguments passed to __init__ of scheduler, to configure the
+            orchestration of training jobs during hyperparameter-tuning.
         verbose : bool
             Whether or not to print out intermediate information during training.
         plot_results : bool
             Whether or not to generate plots summarizing training process.
         visualizer : str
             Describes method to visualize training progress during `fit()`. Options: ['mxboard', 'tensorboard', 'none'].
-        grace_period : int
```
zhreshold, line 172: why are we removing this option?

autogluon/task/object_detection/object_detection.py
```diff
@@ -43,15 +43,16 @@ def fit(dataset='voc',
             ngpus_per_trial=1,
             hybridize=True,
             search_strategy='random',
-            search_options={},
+            search_options=None,
             time_limits=None,
             verbose=False,
             transfer='coco',
-            resume='',
+            resume_model='',
```
zhreshold, line 50: any reason to change the name? This will cause backward incompatibility even though we are in early stages

autogluon/scheduler/hyperband.py
mseeger: The resume block, calling self.load_state_dict, needs to be down here, after everything else has been created. This is because of the way that load_state_dict is done now.

Previously, it relied in pickling complex objects like the searcher. Now, it works differently. The checkpoint only stores what I call "mutable state" of the searcher, and of other objects which would not pickle well. And load_state_dict is first running __init__ for all these objects, and then just puts in their resp mutable state from the checkpoint.

This is the reason I had to move the resume block here, and also that we cannot just pass resume to FIFOScheduler, because it could not do the job properly.

autogluon/task/image_classification/image_classification.py
mseeger: Ask by @jwmueller . Now, all scheduler-specific options are passed in scheduler_options.

autogluon/task/object_detection/object_detection.py
mseeger: Yes, but it was a bug before. You passed resume (a str) to the scheduler, which needs resume to be a bool. This fixes that bug.

autogluon/scheduler/hyperband.py
mseeger: This sounds a little complex, but we need that for our more complex searchers we will contribute. Note that the default for any new searcher is just what has been done before, namely pickling the whole searcher object.

autogluon/task/object_detection/object_detection.py
mseeger: The schedulers currently can only resume from the file whose name is passed in checkpoint. This is a little funny, because that same file would then be overwritten by subsequent checkpoints. But it is simple and works, and I'd not change it for now.

autogluon/task/base/base_task.py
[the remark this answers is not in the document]
Innixma: That is the purpose of deepcopy, copy.copy is only a shallow copy.

copy:

```
import copy
d = {'a': ['hello']}
d_copy = copy.copy(d)
d_copy['a'].append('world')
d['a']
Out[6]: ['hello', 'world']
```

vs deepcopy:

```
d = {'a': ['hello']}
d_copy = copy.deepcopy(d)
d_copy['a'].append('world')
d['a']
Out[11]: ['hello']
```

autogluon/task/image_classification/image_classification.py
Innixma: Consider moving as a kwargs option with deprecation warning that it will be removed in future versions. If present in kwargs, override value in scheduler_options for grace_period. This avoids a breaking change. See 'enable_fit_continuation' argument in tabular as an example.

autogluon/task/image_classification/image_classification.py
```diff
@@ -268,30 +270,15 @@ def fit(dataset,
...
-            scheduler_options.update(
-                {
-                    'searcher': 'random',
-                    'max_t': epochs,
-                    'grace_period': grace_period if grace_period else epochs // 4
-                }
-            )
-
-        results = BaseTask.run_fit(train_image_classification, search_strategy, scheduler_options)
+        scheduler_options = compile_scheduler_options(
```
Innixma, line 273: keyword arguments should be used with this many inputs, it is very hard to notice if an arg is in the wrong position currently.

autogluon/task/image_classification/image_classification.py
```diff
@@ -268,30 +270,15 @@ def fit(dataset,
...
+            scheduler_options, search_strategy, search_options,
+            nthreads_per_trial, ngpus_per_trial, checkpoint, num_trials,
+            time_out=time_limits, resume=resume, visualizer=visualizer,
+            time_attr='epoch', reward_attr='classification_reward',
+            dist_ip_addrs=dist_ip_addrs, epochs=epochs)
+        results = BaseTask.run_fit(
```
Innixma, line 279: super().run_fit is cleaner

autogluon/task/object_detection/object_detection.py
Innixma: The defect/bug is unrelated to the name change, we should keep it as resume, to be consistent with image_classification and avoid backwards breaking change.

Innixma: Why not have the boolean resume you require specified in scheduler_options instead?

autogluon/task/tabular_prediction/tabular_prediction.py
```diff
@@ -59,13 +59,33 @@ def load(output_directory, verbosity=2):
...
+            hyperparameters=None,
+            num_trials=None,
+            search_strategy='random',
+            search_options=None,
+            nthreads_per_trial=None,
+            ngpus_per_trial=None,
+            dist_ip_addrs=None,
+            visualizer='none',
+            verbosity=2,
+            scheduler_options=None,
```
Innixma, line 87: prefer to move scheduler_options to be directly before search_strategy so it is near similar arguments.

autogluon/task/tabular_prediction/tabular_prediction.py
```diff
@@ -497,31 +521,14 @@ def fit(train_data, label, tuning_data=None, time_limits=None, output_directory=
...
-            # but this will be different in the future.
-            if search_strategy == 'hyperband':
-                # Currently, HyperbandScheduler only supports random searcher
-                scheduler_options['searcher'] = 'random'
-        else:
-            # TODO: Check that search_strategy is a subclass of TaskScheduler
-            assert callable(search_strategy)
-            scheduler_cls = search_strategy
-            scheduler_options['searcher'] = 'random'
+        scheduler_options = compile_scheduler_options(
```
Innixma, line 524: ditto previous comment on using keyword parameter names, rather than implicit based on position.

autogluon/task/text_classification/text_classification.py
```diff
@@ -176,24 +178,12 @@ def fit(dataset='SST',
...
-            'dist_ip_addrs': dist_ip_addrs,
-            'searcher': search_strategy,
-            'search_options': search_options,
-        }
-        if search_strategy == 'hyperband':
-            scheduler_options.update({
-                'searcher': 'random',
-                'max_t': epochs,
-                'grace_period': grace_period if grace_period else epochs//4})
+        scheduler_options = compile_scheduler_options(
```
Innixma, line 181: ditto

autogluon/task/object_detection/object_detection.py
```diff
@@ -242,28 +247,16 @@ def fit(dataset='voc',
...
-            'dist_ip_addrs': dist_ip_addrs,
-            'searcher': search_strategy,
-            'search_options': search_options,
-        }
-        if search_strategy == 'hyperband':
-            scheduler_options.update({
-                'searcher': 'random',
-                'max_t': epochs,
-                'grace_period': grace_period if grace_period else epochs // 4})
+        scheduler_options = compile_scheduler_options(
```
Innixma, line 254: ditto

autogluon/task/image_classification/image_classification.py
mseeger: OK, good idea!

autogluon/task/object_detection/object_detection.py
mseeger: Well, the thing is for all other task's, resume is bool and is passed to the scheduler.

My guess is this code would have never worked, certainly not as intended. In all other task's, you checkpoint the searcher state and resume from there. Here, somehow, you resume from an NN model.

I don't really understand this code. Maybe the best would be to pass resume=False to the scheduler? I can also pass (len(resume) > 0) to the resume (bool) argument of the scheduler, maybe that was intended here? Let me try that.

But I think whoever wrote this task, needs to say what is intended here, because the behaviour differs from the other ones, and that is confusing to the user.

autogluon/task/image_classification/image_classification.py
mseeger: OK

autogluon/task/image_classification/image_classification.py
mseeger: OK

autogluon/task/tabular_prediction/tabular_prediction.py
mseeger: OK

mseeger:
BTW: I noted a general bug in the documentation. When 'epoch' is used as time_attr in HyperbandScheduler, you need to report the number of epochs done, i.e. start with 1. The examples report the number of current epoch, i.e. start with 0. This leads to mistakes in HyperbandScheduler.

I'll see that I can fix this everywhere.

mseeger:
OK, I tried to fix this issue with epoch passed to reporter (must start with 1, not 0) almost everywhere.

I have not touched the following (because I am not familiar enough with tabular, and I do not want to break things):

./autogluon/utils/tabular/ml/models/abstract/model_trial.py:        reporter(epoch=0, validation_performance=model.val_score)
./autogluon/utils/tabular/ml/models/tabular_nn/tabular_nn_model.py:                    reporter(epoch=e, validation_performance=val_metric, train_loss=float(train_loss.asscalar()))  # Higher val_metric = better
./autogluon/utils/tabular/ml/models/lgb/callbacks.py:                    reporter(epoch=env.iteration, validation_performance=validation_perf, train_loss=best_trainloss[i],
./autogluon/utils/tabular/ml/models/lgb/hyperparameters/lgb_trial.py:    # reporter(epoch=model.params_trained['num_boost_round'] + 1, validation_performance=score)

autogluon/task/image_classification/image_classification.py
mseeger: Sorry this does not work, these are class and static methods. Reverting to how it was before.

szha:
Job PR-470-9 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-470/9/index.html

autogluon/task/object_detection/object_detection.py
mseeger: OK, I am now passing len(resume) > 0 to resume (bool) of the scheduler. Note what this means: If resume is a non-empty file name, the train_fn will somehow load model pars from that file, and then the scheduler/searcher will resume from the checkpoint loaded from the file name `checkpoint`.

szha:
Job PR-470-10 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-470/10/index.html

autogluon/task/object_detection/object_detection.py
Innixma: @Jerryzcn Could you weigh in here?

Innixma:
> OK, I tried to fix this issue with epoch passed to reporter (must start with 1, not 0) almost everywhere.
> 
> I have not touched the following (because I am not familiar enough with tabular, and I do not want to break things):
> 
> ./autogluon/utils/tabular/ml/models/abstract/model_trial.py: reporter(epoch=0, validation_performance=model.val_score)
> ./autogluon/utils/tabular/ml/models/tabular_nn/tabular_nn_model.py: reporter(epoch=e, validation_performance=val_metric, train_loss=float(train_loss.asscalar())) # Higher val_metric = better
> ./autogluon/utils/tabular/ml/models/lgb/callbacks.py: reporter(epoch=env.iteration, validation_performance=validation_perf, train_loss=best_trainloss[i],
> ./autogluon/utils/tabular/ml/models/lgb/hyperparameters/lgb_trial.py: # reporter(epoch=model.params_trained['num_boost_round'] + 1, validation_performance=score)

Please see my response to the github issue, perhaps we can instead refactor HyperbandScheduler to use e+1. I'm not certain which solution is more 'correct' in this instance. Perhaps @zhanghang1989 could weigh in.

mseeger:
I am in favour of keep HyperbandScheduler as it is.

To, me 'epoch' means number of epochs done. It is a resource attribute, you want to count how much resource (or time) has been spent. So 0 makes no sense to me.

BTW: Some of the code did this correct already, just not all of it.

Innixma:
Regarding Tabular changes:

./autogluon/utils/tabular/ml/models/abstract/model_trial.py: Set to 1 from 0

./autogluon/utils/tabular/ml/models/tabular_nn/tabular_nn_model.py: e+1

./autogluon/utils/tabular/ml/models/lgb/callbacks.py: env.iteration+1

./autogluon/utils/tabular/ml/models/lgb/hyperparameters/lgb_trial.py: Don't change

mseeger:
OK: Did the changes in tabular as well.

BTW: image_classification/pipeline.py did it correctly already, looping epoch starting from 1.

mseeger:
I am on holiday starting from tomorrow, until June 2.

szha:
Job PR-470-11 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-470/11/index.html

szha:
Job PR-470-12 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-470/12/index.html

szha:
Job PR-470-13 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-470/13/index.html

szha:
Job PR-470-14 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-470/14/index.html

mseeger:
Hello, I am off now for holiday until June 2.

If you think this is fine now, please do merge it in.

If not, will have to wait until I am back. This PR fixes quite a few things, including the docstrings of searchers (sorry, I messed them up somehow in earlier PR), and the epoch issue with HB.

PR #472, opened 2020-05-16
Tabular: Fixed refit_full models without bagging

Innixma:
*Issue #, if available:*

*Description of changes:*

Fixed refit_full models without bagging being in subdirectory of their original model.

This previously caused 'optimize_for_deployment' preset to crash when auto_stack=False, refit_full=True, and set_best_to_refit_full=True. This now works correctly.


szha:
Job PR-472-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-472/1/index.html

PR #474, opened 2020-05-19
Tabular: Fixed exception raised when calling fit_summary after deleting models.

Innixma:
*Issue #, if available:*

*Description of changes:*

Fixed exception raised when calling fit_summary after deleting models.

### DIFF
--- a/autogluon/task/tabular_prediction/tabular_prediction.py
+++ b/autogluon/task/tabular_prediction/tabular_prediction.py
@@ -128,6 +128,7 @@ class TabularPrediction(BaseTask):
                     Often can reduce disk usage by ~2-4x with no negatives to model accuracy or inference speed.
                     This will disable numerous advanced functionality, but has no impact on inference.
                     This will make certain functionality less informative, such as `predictor.leaderboard()` and `predictor.fit_summary()`.
+                        Because unused models will be deleted under this preset, methods like `predictor.leaderboard()` and `predictor.fit_summary()` will no longer show the full set of models that were trained during `fit()`.
                     Recommended for applications where the inner details of AutoGluon's training is not important and there is no intention of manually choosing between the final models.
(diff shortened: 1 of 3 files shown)


szha:
Job PR-474-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-474/1/index.html

autogluon/task/tabular_prediction/tabular_prediction.py
```diff
@@ -127,6 +127,7 @@ def fit(train_data, label, tuning_data=None, time_limits=None, output_directory=
                     Optimizes result immediately for deployment by deleting unused models and removing training artifacts.
                     Often can reduce disk usage by ~2-4x with no negatives to model accuracy or inference speed.
                     This will disable numerous advanced functionality, but has no impact on inference.
+                    This will make certain functionality less informative, such as `predictor.leaderboard()` and `predictor.fit_summary()`.
```
jwmueller, line 130: can we explain why, eg something like:

"Because unused models will be deleted under this preset, methods like `predictor.leaderboard()` and `predictor.fit_summary()` will no longer show the full set of models that were trained during `fit()`

jwmueller:
Can we add unit tests to cover all the presets?
There're so many combinations of options now that we really need more unit tests during CI.

szha:
Job PR-474-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-474/2/index.html

autogluon/task/tabular_prediction/tabular_prediction.py
Innixma: Added

Innixma:
> Can we add unit tests to cover all the presets?
> There're so many combinations of options now that we really need more unit tests during CI.

This is probably best left to an issue. The main concern is that it would significantly increase the runtime of CI.

Issue #480, opened 2020-05-21
'pac_score' eval_metric does not work


jwmueller:
Using eval_metric = 'pac_score' produces error: 
```
IndexError: index 1 is out of bounds for axis 1 with size 1
```

Can reproduce by simply running examples/tabular/example_simple_tabular.py with eval_metric set to 'pac_score'.

Innixma:
Fixed in #483

PR #483, opened 2020-05-23
Tabular: Fixed pac_score metric

Innixma:
*Issue #, if available:*
#480 

*Description of changes:*

- Fixed pac_score metric for both binary and multiclass classification
- Fixed defective differences between learner.evaluate(), learner.score() and learner.score_debug().

### DIFF
--- a/autogluon/utils/tabular/ml/learner/abstract_learner.py
+++ b/autogluon/utils/tabular/ml/learner/abstract_learner.py
@@ -219,15 +219,10 @@ class AbstractLearner:
         trainer = self.load_trainer()
         if self.problem_type == MULTICLASS:
             y = y.fillna(-1)
-            if trainer.objective_func_expects_y_pred:
-                return trainer.score(X=X, y=y, model=model)
-            else:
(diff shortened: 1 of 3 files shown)


szha:
Job PR-483-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-483/1/index.html

autogluon/utils/tabular/ml/learner/abstract_learner.py
```diff
@@ -383,7 +379,15 @@ def evaluate(self, y_true, y_pred, silent=False, auxiliary_metrics=False, detail
                 y_true_cleaned = y_true_cleaned.fillna(-1)  # map unknown classes to -1
```
jwmueller, line 379: BTW should we output warning in this case where there's NA labels?

Innixma: Probably not necessary, if it causes a problem with scoring an exception is raised.

PR #501, opened 2020-06-05
Gaussian process based Bayesian optimization for FIFO and Hyperband s…

mseeger:
…chedulers

*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #504, opened 2020-06-07
Fix default for num_trials: If time_limits given, this can be None

mseeger:
*Issue #, if available:*

*Description of changes:*
If time_limits is specified, but num_trials is not, you want to keep num_trials = None, in which case the scheduler starts jobs until the time budget is spent. In the old code, it would still only start 2 trials.

One thing to note: If None is passed to FIFOScheduler, but time_limits is given, it will set num_trials to 100000, so that only time_limits is used. This leads to some funny output ("Number of pending trials: 100000"), but it works.

### DIFF
--- a/autogluon/scheduler/fifo.py
+++ b/autogluon/scheduler/fifo.py
@@ -142,7 +142,6 @@ class FIFOScheduler(TaskScheduler):
         if num_trials is None:
             assert time_out is not None, \
                 "Need stopping criterion: Either num_trials or time_out"
-            num_trials = 100000  # time_out is what matters
         self.num_trials = num_trials
         self.time_out = time_out
         self.max_reward = max_reward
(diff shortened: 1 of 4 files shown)


szha:
Job PR-504-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-504/1/index.html

szha:
Job PR-504-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-504/2/index.html

mseeger:
Fixed behaviour of FIFOScheduler if num_trials = None (there will be no progress bar then).

mseeger:
BTW: This PR is doing a few things which I'll also put into #501 , but it is independent and very small, and should hopefully not be a problem to be merged rapidly.

mseeger:
This is already contained in #501, so can be closed.

PR #511, opened 2020-06-11
Tabular: Improved LGBM training speed by 100%-200% on most datasets

Innixma:
*Issue #, if available:*

*Description of changes:*

Tabular: Improved LGBM training speed by 100%-200% on most datasets

This is achieved by using the built-in evaluation metrics of LGBM when possible, as they are much faster than the sklearn versions AutoGluon natively uses, while resulting in identical final models. I have added support for all of the most common metrics in this PR, and have done a benchmark which confirmed the speedup, which stays consistent on both large and small datasets.


szha:
Job PR-511-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-511/1/index.html

autogluon/utils/tabular/ml/models/lgb/hyperparameters/parameters.py
```diff
@@ -32,7 +32,6 @@ def get_param_multiclass_baseline_custom(num_classes):
         'num_boost_round': DEFAULT_NUM_BOOST_ROUND,
         'num_threads': -1,
         'objective': 'multiclass',
-        'metric': 'multi_error,multi_logloss',
```
jwmueller, line 35: Since 'metric' is now removed from this file, can we add comment in this file that users can also specify additional hyperparameters listed here: 
https://lightgbm.readthedocs.io/en/latest/Parameters.html#

I think this should be done in both this file and the other models' hyperparameters/ files (especially catboost), otherwise users do not know which hyperparameters are even available for them to play with (since we have no documentation of this anywhere outside of these files).

Innixma: Agreed, will do in a separate PR dedicated to documentation of model hyperparameters

PR #512, opened 2020-06-11
[NLP Refactor] Revise the dependency structure of gluonnlp to lazy dependency

sxjscience:
The first step of refactoring the text part in the code base.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -16,6 +16,7 @@ stage("Unit Test") {
         export LD_LIBRARY_PATH=/usr/local/cuda-10.0/lib64
         export MPLBACKEND=Agg
         export MXNET_CUDNN_AUTOTUNE_DEFAULT=0
+        pip install --force-reinstall --upgrade gluonnlp==0.8.1
         pip install --upgrade --force-reinstall -e .
         pip install pytest
         python3 -m pytest --junitxml=results.xml --runslow tests
@@ -42,6 +43,7 @@ stage("Build Docs") {
         export AG_DOCS=1
         git clean -fx
         pip install git+https://github.com/zhanghang1989/d2l-book
+        pip install --force-reinstall --upgrade gluonnlp==0.8.1
         pip install --upgrade --force-reinstall -e .
         cd docs && bash build_doc.sh
         if [[ ${env.BRANCH_NAME} == master ]]; then
--- a/autogluon/utils/try_import.py
+++ b/autogluon/utils/try_import.py
@@ -1,5 +1,5 @@
 __all__ = ['try_import_catboost', 'try_import_lightgbm', 'try_import_mxboard', 'try_import_mxnet',
-           'try_import_cv2']
+           'try_import_cv2', 'try_import_gluonnlp']
 
 def try_import_catboost():
     try:
@@ -48,3 +48,13 @@ def try_import_cv2():
         raise ImportError(
             "Unable to import dependency cv2. "
             "A quick tip is to install via `pip install opencv-python`. ")
+
+def try_import_gluonnlp():
+    try:
+        import gluonnlp
+    except ImportError:
+        raise ImportError(
+            "Unable to import dependency gluonnlp. The NLP model won't be available "
+            "without installing gluonnlp. "
+            "A quick tip is to install via `pip install gluonnlp`. ")
+    return gluonnlp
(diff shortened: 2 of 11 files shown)


autogluon/utils/try_import.py
```diff
@@ -48,3 +48,13 @@ def try_import_cv2():
         raise ImportError(
             "Unable to import dependency cv2. "
             "A quick tip is to install via `pip install opencv-python`. ")
+
+def try_import_gluonnlp():
```
Innixma, line 52: Because AutoGluon imports virtually all files on import autogluon, this will make autogluon raise ImportError on import autogluon. NLP code either needs to be removed from AutoGluon or all usage of gluonnlp has to move to lazy initialization.

sxjscience: I see. Should we revise the import structure accordingly? I find that we have https://github.com/awslabs/autogluon/blob/3c0c51630530a2381ac34383d9da8acbc0a9c580/autogluon/__init__.py#L25

sxjscience: Okay... Just noticed that the whole code base use `from xxx import *`

szha:
Job PR-512-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-512/3/index.html

Jenkinsfile
```diff
@@ -16,6 +16,7 @@ stage("Unit Test") {
         export LD_LIBRARY_PATH=/usr/local/cuda-10.0/lib64
         export MPLBACKEND=Agg
         export MXNET_CUDNN_AUTOTUNE_DEFAULT=0
+        pip install --force-reinstall --upgrade gluonnlp==0.8.1
```
Innixma, line 19: Unit tests must be ran without gluonnlp if it isn't part of the dependencies. This PR appears to have passed CI, but in reality will not work for users.

If we don't plan to use the existing gluonnlp code, we can remove all of the autogluon nlp code.

Otherwise we need to move all imports of gluonnlp to lazy initialization and disable the nlp test until we have a more thorough CI capable of running tests with different dependencies installed for different tests.

sxjscience: I think we need to revise the import structure to avoid importing everything when call `import autogluon`. For the testing purpose, it may be fine to install all required dependencies in the test.

Innixma: Yes, I agree that the importing structure should be revised. This is planned as part of the modularization effort. Unfortunately a lot of the existing tutorials and docs won't work anymore if this is changed right now.

szha:
Job PR-512-4 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-512/4/index.html

Jenkinsfile
sxjscience: @Innixma Revised every appearance to lazy. Let me see if the CI can pass.

szha:
Job PR-512-6 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-512/6/index.html

autogluon/utils/try_import.py
```diff
@@ -48,3 +48,20 @@ def try_import_cv2():
...
+        from pkg_resources import parse_version  # pylint: disable=import-outside-toplevel
+        gluonnlp_version = parse_version(gluonnlp.__version__)
+        assert gluonnlp_version >= parse_version('0.8.1') and\
+               gluonnlp_version <= parse_version('0.8.3'), \
+            'Currently, we only support 0.8.1<=gluonnlp<=0.8.3'
+    except ImportError:
+        raise ImportError(
+            "Unable to import dependency gluonnlp. The NLP model won't be available "
+            "without installing gluonnlp. "
+            "A quick tip is to install via `pip install gluonnlp`. ")
```
Innixma, line 66: change to `pip install gluonnlp==0.8.1`

Innixma: Also add a comment to the text tutorial telling the user to `pip install gluonnlp==0.8.1` prior to running.

szha:
Job PR-512-7 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-512/7/index.html

szha:
Job PR-512-11 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-512/11/index.html

PR #517, opened 2020-06-12
Tabular: Fixed defect in transform_features + new arg

Innixma:
*Issue #, if available:*

*Description of changes:*

- Fixed defect in transform_features when training data's indices are not 0-n
- Added return_original_features argument to simplify obtaining pure OOF predictions

### DIFF
--- a/autogluon/task/tabular_prediction/predictor.py
+++ b/autogluon/task/tabular_prediction/predictor.py
@@ -366,7 +366,7 @@ class TabularPredictor(BasePredictor):
             print("*** End of fit() summary ***")
         return results
 
-    def transform_features(self, dataset=None, model=None, base_models=None):
+    def transform_features(self, dataset=None, model=None, base_models=None, return_original_features=True):
         """
         Transforms dataset features through the AutoGluon feature generator.
         This is useful to gain an understanding of how AutoGluon interprets the dataset features.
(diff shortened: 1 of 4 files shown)


szha:
Job PR-517-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-517/2/index.html

szha:
Job PR-517-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-517/3/index.html

autogluon/task/tabular_prediction/predictor.py
```diff
@@ -412,6 +412,10 @@ def transform_features(self, dataset=None, model=None, base_models=None):
             Valid models are listed in this `predictor` by calling `predictor.model_names`.
             If a stacker model S exists with `base_models=M`, then setting `base_models=M` is equivalent to setting `model=S`.
             `model=None` is a requirement when specifying `base_models`.
+        return_original_features : bool, default = True
```
jwmueller, line 415: `return_original_features` option could be clarified a bit more, eg by precisely stating what sort of DF will be returned when it is true vs when it is false, as well as describing for what purposes one might want it to be True vs False.

Innixma:
> Just clarify the documentation in `transform_features`.
> 
> Also, why not have `use_orig_features` be an option of the stacker models that can be specified by user in task.fit() or optimized in HPO?

This is planned to be added in future

autogluon/task/tabular_prediction/predictor.py
Innixma: Will clarify in future PR

PR #523, opened 2020-06-15
Tests and files

mseeger:
*Issue #, if available:*

*Description of changes:*
- OpenML dataset used in docs/tutorials/course/mlp.md downloaded from S3, not from OpenML
- Stuff from this tutorial also added as unit test
- Comparison against GPy unit test, which also depends on files in S3 bucket


autogluon/searcher/bayesopt/autogluon/openml_download.py
```diff
@@ -0,0 +1,60 @@
+import os
```
Innixma, line 1: Should this file live somewhere else? It doesn't seem to be particular to bayesopt. Maybe in autogluon/utils?

mseeger:
The CI always breaks with:

tests/test_check_style.py .                                              [  0%]

tests/unittests/test_bayesopt_searcher.py ..                             [  1%]

tests/unittests/test_checkpoint_resume.py ..                             [  2%]

tests/unittests/test_classification_tricks.py Fatal Python error: Aborted


Thread 0x00007ff32e933700 (most recent call first):



Does anybody know how to get more information? Which of the tests is failing, and where?
All of this runs fine locally on my Mac.

autogluon/searcher/bayesopt/autogluon/openml_download.py
mseeger: It is some pretty specific piece of code that was previously in Thibaut's tutorial, and is now also used in a unit test. Do you think it would be more generally useful?

I could also use a file collecting other helpers for unit tests of bayesopt, and put it there.

mseeger:
@Innixma @zhanghang1989 Could it be the CI system is running several tests on different CPUs on the same instance?

This is something that gaves us trouble in the past. The test test_bayesopt_searcher.py, which is really just Thibaut's tutorial, is trying to claim all CPUs on the instance for this reason.

I can only say:
- This test runs fine when run separately on an instance
- Thibaut's tutorial also seems to work as part of the CI, at least it did in the past

mseeger:
@szha As you seem to be most knowledgable about the CI system, could you have a look at this one?
Question: Is the CI system using multiple processes (or even threads) to run several tests on different CPUs on the same instance?

If so, this is something that tends to clash with MXNet, even if things are running on CPU only. If this is possible, it would be great to check what happens if the CI system is forced to run one test at a time, so that we do not run into these issues that we know affect AutoGluon in any case.

mseeger:
Latest update:
- test_comparison_gpy goes through
- test_bayesopt_search seems to cause the failure, and is this is likely because several tests are run at the same time on one instance, which we know is a major issue with MXNet

I see the following alternatives:
- Remove test_bayesopt_search: I am MUCH against that, this would be very risky
- Make sure that at least for test_bayesopt_search, but maybe for all tests, the CI system does
   not run it (or them) in parallel
- Split tests into such which can be run in parallel, and such which cannot. I'd strongly recommend all test which really run an experiment (i.e., scheduler.run(); scheduler.join_jobs()) not to be run in parallel, as this issue will otherwise creep in later

szha:
Job PR-523-5 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-523/5/index.html

mseeger:
One more datapoint:
The code in the failing unit test test_bayesopt_searcher.py is pretty much exactly the same code that is in the tutorial in docs/tutorials/course/mlp.md.

And the latter works in the CI system. As you can see, when I removed the test_bayesopt_searcher.py, everything went through, and when I look at the output of
Job PR-523-5, then the tutorial outputs looks all fine. It even ran more trials of HPO, because it has a time limit of 2 mins.

I browsed around, and usually you have to be pretty explicit to tell pytest to run things in parallel (and the command in Jenkinsfile does not do that), so I really do not know what is going on here

Innixma:
I'm not very well versed in how the CI functions, perhaps @zhreshold has more insight?

autogluon/searcher/bayesopt/autogluon/openml_download.py
zhreshold: I would agree with nick's suggestion, put the download utils somewhere in utils is better IMO

zhreshold:
> Latest update:
> 
> * test_comparison_gpy goes through
> * test_bayesopt_search seems to cause the failure, and is this is likely because several tests are run at the same time on one instance, which we know is a major issue with MXNet
> 
> I see the following alternatives:
> 
> * Remove test_bayesopt_search: I am MUCH against that, this would be very risky
> * Make sure that at least for test_bayesopt_search, but maybe for all tests, the CI system does
>   not run it (or them) in parallel
> * Split tests into such which can be run in parallel, and such which cannot. I'd strongly recommend all test which really run an experiment (i.e., scheduler.run(); scheduler.join_jobs()) not to be run in parallel, as this issue will otherwise creep in later

If you suspect it's the parallel test who caused the problem, probably you can try `pytest -m "serial"` to verify? For the ci instance, I don't think there's something we can control for resources that request ALL CPU, if I remember it correctly, the cpu resource manager does not even care about it?

mseeger:
https://stackoverflow.com/questions/56659700/pytest-pytest-benchmark-serial-or-parallel

https://github.com/pytest-dev/pytest-xdist/issues/84

All are saying pytest never runs tests in parallel. To do so, one would have to use pytest-xdist, we are not doing that.
@zhreshold According to docs, -m "serial" just means pytest will only execute tests marked as serial.

Given that pytest by default executes tests sequentially, I am very puzzled about the CI out with the percentages. It just looks like things are happening in parallel.

@zhanghang1989 It would be great if you had a look at this.

Otherwise, the only option I see for now would be to remove this test. We'd then rely on the docs generation to break if anything is wrong. I find this pretty dangerous moving forward.

zhreshold:
One guess is that internally in the test, mxnet is run in async mode causing the bug harder to catch.
So if it fails inside of test_bayesopt_searcher.py it will very likely cause the following error

```
tests/test_check_style.py .                                              [  0%]

tests/unittests/test_bayesopt_searcher.py ...               [  1%].   <- failed here

tests/unittests/test_checkpoint_resume.py ..                             [  2%]

tests/unittests/test_classification_tricks.py Fatal Python error: Aborted
```

When I search for `Fatal Python error: Aborted` issue, there's similar problem here and there pointing to memory allocation error, which sounds relavant since the CI instances have 32G memory in total. @mseeger Can you verify if it's the case? 

I would also try to synchronize fully for mxnet with `mx.nd.waitall()` after each test.

szha:
You can try either adding waitall between tests (e.g. we did so in mxnet with pytest in [conftest.py](https://github.com/apache/incubator-mxnet/blob/master/conftest.py#L195-L199)) or setting the `MXNET_ENGINE_TYPE=NaiveEngine` in the CI to get unblocked on the asynchronous execution and see which test breaks.

mseeger:
This still fails. Maybe @zhanghang1989 has an idea?

Otherwise, I will remove this test.

szha:
Job PR-523-11 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-523/11/index.html

mseeger:
OK, when I remove the unit test, the CI passes now.

Innixma:
> OK, when I remove the unit test, the CI passes now.

Could you open an issue relating to this test since it is still unknown why it fails? Also, I will let @zhreshold decide on merging, as I lack context for this PR.

mseeger:
@Innixma Should I wait with raising the issue until this is merged? The test code is part of this PR, but it is marked as pytest.mark.skip, so is not executed. Without the code, it'd be hard to state the issue.

This PR is also modifying the mlp.md tutorial, and adds a unit test for the GP code.

Innixma:
> @Innixma Should I wait with raising the issue until this is merged? The test code is part of this PR, but it is marked as pytest.mark.skip, so is not executed. Without the code, it'd be hard to state the issue.
> 
> This PR is also modifying the mlp.md tutorial, and adds a unit test for the GP code.

That is fine, it can be created after the PR is merged.

mseeger:
@zhreshold Could we merge this? It does a few things @jwmueller asked for, in particular removes dependence on openml (which can lead to CI failures) and also adds a unit test for our GP code.

The issue with the end-2-end bayesopt unit test remains, but I'd open another issue.

PR #527, opened 2020-06-19
CatBoost - disabled RSM if GPU training is enabled

sgdread:
*Description of changes:*
The RSM optimization works only when trained on CPU - added check if GPU training is active

### DIFF
--- a/autogluon/utils/tabular/ml/models/catboost/catboost_model.py
+++ b/autogluon/utils/tabular/ml/models/catboost/catboost_model.py
@@ -142,9 +142,10 @@ class CatboostModel(AbstractModel):
                 # Reduce sample iterations to avoid taking unreasonable amounts of time
                 num_sample_iter_max = max(round(num_sample_iter_max/2), 2)
             # Subsample columns to speed up training
-            params['colsample_bylevel'] = max(min(1.0, 1000 / num_features), 0.05)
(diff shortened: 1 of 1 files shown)


szha:
Job PR-527-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-527/1/index.html

sgdread:
> One thing to note is that given colsample_by_level does not work in GPU, should we also add a warning if it is set and automatically set it to 1.0 when GPU is used?

Agreed, updated.

Innixma:
The else statement added never has any affect, since it only occurs in the if statement when it wasn't set already. It has to occur as an if prior to the outer if statement:

```
if params.get('task_type', None) == 'GPU':
    if 'colsample_bylevel' in params:
        params.pop('colsample_bylevel')
        logger.log(30, f'\t\'colsample_bylevel\' is not supported on GPU, using default value (Default = 1).')
    if 'rsm' in params:
        params.pop('rsm')
        logger.log(30, f'\t\'rsm\' is not supported on GPU, using default value (Default = 1).')
if self.problem_type == MULTICLASS and 'rsm' not in params and 'colsample_bylevel' not in params and num_features > 1000:
    ...
    ...
```

szha:
Job PR-527-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-527/2/index.html

szha:
Job PR-527-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-527/3/index.html

PR #534, opened 2020-06-23
Tabular: Moved CatBoost import into model fit call

Innixma:
*Issue #, if available:*

*Description of changes:*

Moved CatBoost import into model fit call, this enables CatBoost to be an optional dependency without causing an overall crash if it is included in the model list.


szha:
Job PR-534-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-534/1/index.html

PR #540, opened 2020-06-27
Clean up argument checking and default value imputation for schedulers

mseeger:
*Issue #, if available:*

*Description of changes:*
Extends something I did for GP searchers to schedulers. Arguments are losely type-checked, and default values are maintained consistently (e.g., do not have to copy&paste default values from superclass).
Also, whenever a default value is imputed, this is written into the log, so you later on know what argument values were used.

### DIFF
--- a/autogluon/scheduler/fifo.py
+++ b/autogluon/scheduler/fifo.py
@@ -19,10 +19,35 @@ from ..core.decorator import _autogluon_method
 from ..searcher import BaseSearcher
 from ..searcher.searcher_factory import searcher_factory
 from ..utils import save, load, mkdir, try_import_mxboard
+from ..utils.default_arguments import check_and_merge_defaults, \
+    Float, Integer, String, Boolean
 
 __all__ = ['FIFOScheduler']
 
 logger = logging.getLogger(__name__)
+
+
+_DEFAULT_OPTIONS = {
+    'resource': {'num_cpus': 1, 'num_gpus': 0},
+    'searcher': 'random',
+    'resume': False,
+    'reward_attr': 'accuracy',
+    'time_attr': 'epoch',
+    'visualizer': 'none',
+    'training_history_callback_delta_secs': 60,
+    'delay_get_config': True}
+
+_CONSTRAINTS = {
+    'checkpoint': String(),
+    'resume': Boolean(),
+    'num_trials': Integer(1, None),
+    'time_out': Float(0.0, None),
+    'max_reward': Float(),
+    'reward_attr': String(),
+    'time_attr': String(),
+    'visualizer': String(),
+    'training_history_callback_delta_secs': Integer(1, None),
+    'delay_get_config': Boolean()}
 
 
 class FIFOScheduler(TaskScheduler):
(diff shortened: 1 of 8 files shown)


szha:
Job PR-540-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-540/2/index.html

szha:
Job PR-540-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-540/3/index.html

autogluon/scheduler/fifo.py
```diff
@@ -105,29 +130,23 @@ class FIFOScheduler(TaskScheduler):
...
-                 training_history_callback_delta_secs=60,
-                 delay_get_config=True):
-        super().__init__(dist_ip_addrs)
-        if resource is None:
-            resource = {'num_cpus': 1, 'num_gpus': 0}
-        self.resource = resource
-
-        if searcher is None:
-            searcher = 'random'  # RandomSearcher
+    def __init__(self, train_fn, **kwargs):
```
Innixma, line 133: Since this function takes a lot of parameters which have now been moved to kwargs, do we have any intention to raise an exception if an unknown parameter is passed, or will we just ignore them?

autogluon/scheduler/fifo.py
```diff
@@ -19,12 +19,37 @@
...
 from ..utils import save, load, mkdir, try_import_mxboard
+from ..utils.default_arguments import check_and_merge_defaults, \
+    Float, Integer, String, Boolean
 
 __all__ = ['FIFOScheduler']
 
 logger = logging.getLogger(__name__)
 
 
+_DEFAULT_OPTIONS = {
```
Innixma, line 30: If the primary purpose of using kwargs is to have these defaults set and the type checks (_CONSTRAINTS), I think we could instead use a decorator function to __init__, with the decorator calling the check_and_merge_defaults function. This output is then passed to the __init__ as input, avoiding the need to alter the input type to be kwargs if desired. See TabularPrediction fit() as an example of this, where we validate and clean the inputs without having to change fit() to be kwargs input.

autogluon/scheduler/fifo.py
mseeger: It only works the (simple) way I did it, if such extra parameters are ignored, because I am just passing all parameters also of the subclass. For example, this is called by HyperbandScheduler.__init__, passing things like max_t, grace_period.

By doing these "imputation" outputs into the log, I feel the user can see what is going on.

autogluon/scheduler/fifo.py
mseeger: Do you insist on this? Since I am not familiar with decorators, I'd love to not do this now.

autogluon/scheduler/fifo.py
mseeger: The only thing I can see which could happen now, is that a user misspells an argument, say passes 'searchr' instead of 'searcher'. In this case, the code would not break, but it would generate an output in the log, saying that 'searcher' is imputed. So, it certainly does not pass silently.

Hmm, in fact maybe I can think of something ...

mseeger: OK, I put in a mechanism to catch invalid parameters.

szha:
Job PR-540-4 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-540/4/index.html

autogluon/scheduler/fifo.py
Innixma: It is fine for now, I am just mentioning this as an option if things start to get confusing down the road.

autogluon/scheduler/fifo.py
Innixma: Looks great! I am fine for you to merge.

PR #518, opened 2020-06-13
Tabular: Fixed predict defect in NN, optimized memory usage in NN

Innixma:
*Issue #, if available:*

*Description of changes:*

This change theoretically could enable 50% larger datasets to be preprocessed in NN without going OOM.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #542, opened 2020-06-28
Tabular: Fixed major defect in NN causing crashes with categoricals

Innixma:
*Issue #, if available:*

*Description of changes:*

Defect introduced in #518 

Fixed major defect in NN causing crashes with categoricals. Since 0.0.11 release, any dataset with categoricals that cause embeddings to be created in the NN will cause the NN to fail during training.


szha:
Job PR-542-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-542/1/index.html

PR #548, opened 2020-07-04
Tabular: Refactored feature types metadata

Innixma:
*Issue #, if available:*

*Description of changes:*

- Refactored feature_types_metadata to be a custom class object instead of a dictionary for increased extensibility.
- Refactored feature_generator to have improved readability and extensibility.
- KNN now drops NLP features to avoid large inaccurate KNN models with massive inference time in NLP problems.

### DIFF
--- a/autogluon/task/tabular_prediction/predictor.py
+++ b/autogluon/task/tabular_prediction/predictor.py
@@ -31,8 +31,10 @@ class TabularPredictor(BasePredictor):
         label_column : str
             Name of table column that contains data from the variable to predict (often referred to as: labels, response variable, target variable, dependent variable, Y, etc).
         feature_types : :class:`autogluon.utils.tabular.features.feature_types_metadata.FeatureTypesMetadata`
-            Inferred data type of each predictive variable (i.e. column of training data table used to predict `label_column`).
-            Contains both raw dtype and special dtype information. Special dtypes are AutoGluon specific feature types.
+            Inferred data type of each predictive variable after preprocessing transformation (i.e. column of training data table used to predict `label_column`).
+            Contains both raw dtype and special dtype information. Each feature has exactly 1 raw dtype (such as 'int', 'float', 'category') and zero to many special dtypes (such as 'datetime', 'text', 'text_ngram').
+            Special dtypes are AutoGluon specific feature types that are used to identify features with meaning beyond what the raw dtype can convey.
+                `feature_types.get_feature_types_raw_flattened()`: Dictionary of feature name -> raw dtype mappings.
                 `feature_types.feature_types_raw`: Dictionary of lists of raw feature names, grouped by raw feature dtype.
                 `feature_types.feature_types_special`: Dictionary of lists of special feature names, grouped by special feature dtype.
         model_names : list
--- a/autogluon/task/tabular_prediction/predictor.py
+++ b/autogluon/task/tabular_prediction/predictor.py
@@ -30,8 +30,11 @@ class TabularPredictor(BasePredictor):
             What metric is used to evaluate predictive performance.
         label_column : str
             Name of table column that contains data from the variable to predict (often referred to as: labels, response variable, target variable, dependent variable, Y, etc).
-        feature_types : dict
+        feature_types : :class:`autogluon.utils.tabular.features.feature_types_metadata.FeatureTypesMetadata`
             Inferred data type of each predictive variable (i.e. column of training data table used to predict `label_column`).
+            Contains both raw dtype and special dtype information. Special dtypes are AutoGluon specific feature types.
+                `feature_types.feature_types_raw`: Dictionary of lists of raw feature names, grouped by raw feature dtype.
+                `feature_types.feature_types_special`: Dictionary of lists of special feature names, grouped by special feature dtype.
         model_names : list
             List of model names trained during `fit()`.
         model_performance : dict
--- a/autogluon/task/tabular_prediction/tabular_prediction.py
+++ b/autogluon/task/tabular_prediction/tabular_prediction.py
@@ -399,6 +399,13 @@ class TabularPrediction(BaseTask):
                 Note: The file containing your `FeatureGenerator` class must be imported into current Python session in order to use a custom class.
             feature_generator_kwargs : dict, default={}
                 Keyword arguments to pass into the `FeatureGenerator` constructor.
+                Valid `AutoMLFeatureGenerator` kwargs:
+                    enable_nlp_vectorizer_features : bool, default = True
+                        If True, the vectorizer argument value is used to generate 'nlp_ngram' features from text features if present.
+                    enable_nlp_ratio_features : bool, default = True
+                        If True, generate 'nlp_special' features from text features if present.
+                    vectorizer : `sklearn.feature_extraction.text.CountVectorizer`, default = `CountVectorizer(min_df=30, ngram_range=(1, 3), max_features=10000, dtype=np.uint8)`
+                        Determines the count vectorizer used during feature generation if text features are detected.
             trainer_type : `Trainer` class, default=`AutoTrainer`
                 A class inheriting from `autogluon.utils.tabular.ml.trainer.abstract_trainer.AbstractTrainer` that controls training/ensembling of many models.
                 Note: In order to use a custom `Trainer` class, you must import the class file that defines it into the current Python session.
--- a/autogluon/task/tabular_prediction/tabular_prediction.py
+++ b/autogluon/task/tabular_prediction/tabular_prediction.py
@@ -148,7 +148,7 @@ class TabularPrediction(BaseTask):
                     This preset pairs well with the other presets such as `good_quality_faster_inference_only_refit` to make a very compact final model.
                     Identical to calling `predictor.delete_models(models_to_keep='best', dry_run=False)` and `predictor.save_space()` directly after `fit()`.
 
-                ignore_text={'feature_generator_kwargs': {'enable_nlp_vectorizer_features': False, 'enable_nlp_ratio_features': False}}
+                ignore_text={'feature_generator_kwargs': {'enable_text_ngram_features': False, 'enable_text_special_features': False}}
                     Disables automated feature generation when text features are detected.
                     This is useful to determine how beneficial text features are to the end result, as well as to ensure features are not mistaken for text when they are not.
 
@@ -393,19 +393,21 @@ class TabularPrediction(BaseTask):
                 If True, will set Trainer.best_model = Trainer.full_model_dict[Trainer.best_model]
                 This will change the default model that Predictor uses for prediction when model is not specified to the refit_full version of the model that previously exhibited the highest validation score.
                 Only valid if `refit_full` is set.
-            feature_generator_type : `FeatureGenerator` class, default=`AutoMLFeatureGenerator`
+            feature_generator_type : :class:`autogluon.utils.tabular.features.auto_ml_feature_generator.AbstractFeatureGenerator` class, default = :class:`autogluon.utils.tabular.features.auto_ml_feature_generator.AutoMLFeatureGenerator`
                 A `FeatureGenerator` class specifying which feature engineering protocol to follow
-                (see autogluon.utils.tabular.features.abstract_feature_generator.AbstractFeatureGenerator).
                 Note: The file containing your `FeatureGenerator` class must be imported into current Python session in order to use a custom class.
             feature_generator_kwargs : dict, default={}
                 Keyword arguments to pass into the `FeatureGenerator` constructor.
-                Valid `AutoMLFeatureGenerator` kwargs:
-                    enable_nlp_vectorizer_features : bool, default = True
-                        If True, the vectorizer argument value is used to generate 'nlp_ngram' features from text features if present.
-                    enable_nlp_ratio_features : bool, default = True
-                        If True, generate 'nlp_special' features from text features if present.
+                Valid :class:`autogluon.utils.tabular.features.auto_ml_feature_generator.AutoMLFeatureGenerator` kwargs:
+                    enable_text_ngram_features : bool, default = True
+                        If True, the vectorizer argument value is used to generate 'text_ngram' features from text features if present.
+                        Try setting this to False if you encounter memory issues running AutoGluon on text data and cannot access a machine with more memory.
+                    enable_text_special_features : bool, default = True
+                        If True, generate 'text_special' features from text features if present.
+                        Examples of 'text_special' features include the number of whitespaces and the average word length in a text feature.
                     vectorizer : `sklearn.feature_extraction.text.CountVectorizer`, default = `CountVectorizer(min_df=30, ngram_range=(1, 3), max_features=10000, dtype=np.uint8)`
                         Determines the count vectorizer used during feature generation if text features are detected.
+                        If your data contain text fields and you encounter memory issues running AutoGluon (and cannot access a machine with more memory), then consider reducing max_features or setting n_gram_range=(1, 2).
             trainer_type : `Trainer` class, default=`AutoTrainer`
                 A class inheriting from `autogluon.utils.tabular.ml.trainer.abstract_trainer.AbstractTrainer` that controls training/ensembling of many models.
                 Note: In order to use a custom `Trainer` class, you must import the class file that defines it into the current Python session.
--- a/autogluon/utils/tabular/features/abstract_feature_generator.py
+++ b/autogluon/utils/tabular/features/abstract_feature_generator.py
@@ -100,6 +100,7 @@ class AbstractFeatureGenerator:
             unique_value_count = len(X_features[column].unique())
             if unique_value_count == 1:
                 self.features_to_remove_post.append(column)
+            # TODO: Consider making 0.99 a parameter to FeatureGenerator
             elif 'object' in self.feature_type_family and column in self.feature_type_family['object'] and (unique_value_count / X_len > 0.99):
                 self.features_to_remove_post.append(column)
 
@@ -327,7 +328,7 @@ class AbstractFeatureGenerator:
                 logger.debug(f'date: {column}')
                 logger.debug(unique_counts.head(5))
             elif self.check_if_nlp_feature(col_val):
-                type_family = 'nlp_text'
+                type_family = 'text'
                 logger.debug(f'nlp: {column}')
                 logger.debug(unique_counts.head(5))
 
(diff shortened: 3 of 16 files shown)


Innixma:
Example code output during preprocessing + example of KNN not having any valid features to train with (all int/float are from nlp ngrams/special):

```
Feature Generator processed 988 data points with 511 features
Original Features (raw dtypes):
	object features: 9
Original Features (inferred dtypes):
	object features: 6
	nlp_text features: 3
Generated Features (special dtypes):
	nlp_category features: 3
	nlp_special features: 87
	nlp_ngram features: 416
Final Features (raw dtypes):
	category features: 8
	int features: 503
Final Features:
	nlp_category features: 3
	nlp_special features: 87
	nlp_ngram features: 416
	category features: 5
	Data preprocessing and feature engineering runtime = 3.51s ...
AutoGluon will gauge predictive performance using evaluation metric: accuracy
To change this, specify the eval_metric argument of fit()
AutoGluon will early stop models using evaluation metric: accuracy
Fitting model: KNeighborsClassifier ...
	No valid features to train KNeighborsClassifier... Skipping this model.
Fitting model: NeuralNetClassifier ...
	0.5192	 = Validation accuracy score
	17.21s	 = Training runtime
	0.1s	 = Validation runtime
```

szha:
Job PR-548-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-548/2/index.html

szha:
Job PR-548-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-548/3/index.html

autogluon/task/tabular_prediction/predictor.py
```diff
@@ -30,8 +30,11 @@ class TabularPredictor(BasePredictor):
             What metric is used to evaluate predictive performance.
         label_column : str
             Name of table column that contains data from the variable to predict (often referred to as: labels, response variable, target variable, dependent variable, Y, etc).
-        feature_types : dict
+        feature_types : :class:`autogluon.utils.tabular.features.feature_types_metadata.FeatureTypesMetadata`
             Inferred data type of each predictive variable (i.e. column of training data table used to predict `label_column`).
+            Contains both raw dtype and special dtype information. Special dtypes are AutoGluon specific feature types.
```
jwmueller, line 35: can add clarification: 

"Special dtypes are AutoGluon-specific feature types, whereas raw dtypes refer to pandas column-types from the original DataFrame".

autogluon/task/tabular_prediction/predictor.py
```diff
@@ -30,8 +30,11 @@ class TabularPredictor(BasePredictor):
...
+                `feature_types.feature_types_raw`: Dictionary of lists of raw feature names, grouped by raw feature dtype.
+                `feature_types.feature_types_special`: Dictionary of lists of special feature names, grouped by special feature dtype.
```
jwmueller, line 37: Can you add example one-liner to docs to show how user can print the inferred feature_types from the predictor?

autogluon/task/tabular_prediction/tabular_prediction.py
```diff
@@ -399,6 +399,13 @@ def fit(train_data,
                 Note: The file containing your `FeatureGenerator` class must be imported into current Python session in order to use a custom class.
             feature_generator_kwargs : dict, default={}
                 Keyword arguments to pass into the `FeatureGenerator` constructor.
+                Valid `AutoMLFeatureGenerator` kwargs:
+                    enable_nlp_vectorizer_features : bool, default = True
+                        If True, the vectorizer argument value is used to generate 'nlp_ngram' features from text features if present.
```
jwmueller, line 404: Maybe add: "Try setting this to False if you encounter memory issues running AutoGluon and cannot access a machine with more memory".

autogluon/task/tabular_prediction/tabular_prediction.py
```diff
@@ -399,6 +399,13 @@ def fit(train_data,
...
+                    enable_nlp_ratio_features : bool, default = True
+                        If True, generate 'nlp_special' features from text features if present.
```
jwmueller, line 406: why is the argument named `nlp_ratio_features` but the docstring refers to `nlp_special` features?  It would also be good to (briefly) list what these features are (perhaps in English rather than variable-names)

autogluon/task/tabular_prediction/tabular_prediction.py
```diff
@@ -399,6 +399,13 @@ def fit(train_data,
...
+                    vectorizer : `sklearn.feature_extraction.text.CountVectorizer`, default = `CountVectorizer(min_df=30, ngram_range=(1, 3), max_features=10000, dtype=np.uint8)`
+                        Determines the count vectorizer used during feature generation if text features are detected.
```
jwmueller, line 408: Maybe add some trouble-shooting tips here in docstring?
E.g.
"If your data contain text fields and you encounter memory issues running AutoGluon (and cannot access a machine with more memory), then consider reducing `max_features` or setting `n_gram_range=(1,2)`."

autogluon/task/tabular_prediction/tabular_prediction.py
```diff
@@ -399,6 +399,13 @@ def fit(train_data,
                 Note: The file containing your `FeatureGenerator` class must be imported into current Python session in order to use a custom class.
             feature_generator_kwargs : dict, default={}
                 Keyword arguments to pass into the `FeatureGenerator` constructor.
+                Valid `AutoMLFeatureGenerator` kwargs:
```
jwmueller, line 402: maybe provide full path to `AutoMLFeatureGenerator`

jwmueller:
When printing outputs to user, should we replace `nlp` with `text` in the feature names?
Also many users may find `nlp_categorical`, `nlp_special` confusing when they scan the output.
Could perhaps name the features:
`nlp_text` -> `text`
`nlp_category` -> `text_as_category`
`nlp_ngram` -> `text_as_ngram`
`nlp_special` -> `text_as_?` (not sure why special and ratio are currently used interchangeably)

autogluon/utils/tabular/features/abstract_feature_generator.py
```diff
@@ -127,13 +92,15 @@ def fit_transform(self, X: DataFrame, y=None, banned_features=None, fix_categori
...
+        self.feature_type_family_init_raw = get_type_groups_df(X)
+
         X.reset_index(drop=True, inplace=True)
         X_features = self.generate_features(X)
         for column in X_features:
             unique_value_count = len(X_features[column].unique())
             if unique_value_count == 1:
                 self.features_to_remove_post.append(column)
-            elif column in self.feature_type_family['object'] and (unique_value_count / X_len > 0.99):
+            elif 'object' in self.feature_type_family and column in self.feature_type_family['object'] and (unique_value_count / X_len > 0.99):
```
jwmueller, line 103: consider adding TODO to make 0.99 here a kwarg of FeatureGenerator

jwmueller:
(In this or future PR) can you add documentation to `auto_ml_feature_generator.py` that briefly describes all the steps/transformations that are being applied?  I believe this is currently opaque for most users, and also subclassing the `FeatureGenerator` may be harder than it should be without an example that has documentation.

autogluon/utils/tabular/features/feature_types_metadata.py
```diff
@@ -0,0 +1,54 @@
...
+
+class FeatureTypesMetadata:
+    """
+    Contains feature type metadata information such as type family groups (feature_types_raw) and special feature type groups (feature_types_special)
+
+    feature_types_raw is the dictionary computed as output to :function:`autogluon.utils.tabular.features.utils.get_type_family_groups_df`
+    feature_types_special is an optional dictionary to communicate special properties of features to downstream models that have special handling functionality for those feature types.
+        As an example, feature_types_special might contain a key 'nlp_ngram' indicating that the list of values are all features which were generated from a nlp vectorizer and represent ngrams.
+        A downstream model such as a K-Nearest-Neighbor model could then check if 'nlp_ngram' is present in feature_types_special and drop those features if present, to speed up training and inference time.
+    """
```
jwmueller, line 15: Should we list all the possible feature_types here?
Otherwise it is not easy for contributors to add new models (and they may forget to handle some feature-types).

jwmueller:
In future PR, should we add support for user to declare `TabularDataset.feature_types`?
At least we should remove the argument in this PR if it'll never be used now that `feature_types` is a custom object.

In order for a user to declare their own feature_types, there will need to be a clearer description of the possible types which we can put at the top of the class `FeatureTypesMetadata` (or in the TabularDataset documentation).

jwmueller:
In TabularDataset documentation, it could also be nice to add brief statement for users of the `df` arg (who already have a pandas DataFrame object) on tips regarding what dtypes to use for best efficiency.

autogluon/utils/tabular/ml/models/abstract/abstract_model.py
```diff
@@ -153,6 +154,8 @@ def _set_default_auxiliary_params(self):
             # ignore_hpo=False,
             # max_early_stopping_rounds=None,
             # use_orig_features=True,  # TODO: Only for stackers
+            ignored_feature_types_special=None,  # List, drops any features in `self.feature_types_metadata.feature_types_special[type]` for type in `ignored_feature_types_special`. | Currently undocumented in task.
+            ignored_feature_types_raw=None,  # List, drops any features in `self.feature_types_metadata.feature_types_raw[type]` for type in `ignored_feature_types_raw`. | Currently undocumented in task.
```
jwmueller, line 158: perhaps default = [] is clearer than None that these should be lists.

autogluon/utils/tabular/ml/models/abstract/abstract_model.py
```diff
@@ -58,7 +59,7 @@ class AbstractModel:
     model_info_name = 'info.pkl'
     model_info_json_name = 'info.json'
 
-    def __init__(self, path: str, name: str, problem_type: str, objective_func, num_classes=None, stopping_metric=None, model=None, hyperparameters=None, features=None, feature_types_metadata=None, debug=0, **kwargs):
+    def __init__(self, path: str, name: str, problem_type: str, objective_func, num_classes=None, stopping_metric=None, model=None, hyperparameters=None, features=None, feature_types_metadata: FeatureTypesMetadata = None, debug=0, **kwargs):
```
jwmueller, line 62: It would be good to add brief documentation about `FeatureTypesMetadata`, since this class is important for users to understand in order to add a new model.

autogluon/utils/tabular/ml/models/knn/knn_model.py
```diff
@@ -37,6 +37,15 @@ def _set_default_params(self):
         for param, val in default_params.items():
             self._set_default_param_value(param, val)
 
+    def _set_default_auxiliary_params(self):
+        default_auxiliary_params = dict(
+            ignored_feature_types_special=['nlp_ngram', 'nlp_special'],
+            ignored_feature_types_raw=['category', 'object'],
```
jwmueller, line 43: Add TODO not to ignore categoricals

autogluon/utils/tabular/ml/models/lr/lr_model.py
```diff
@@ -38,10 +38,12 @@ def _get_types_of_features(self, df):
...
-        categorical_featnames = self.__get_feature_type_if_present('object') + self.__get_feature_type_if_present('bool')
-        continuous_featnames = self.__get_feature_type_if_present('float') + self.__get_feature_type_if_present('int') + self.__get_feature_type_if_present(
-            'datetime')
-        language_featnames = self.feature_types_metadata['nlp']
+
+        feature_types = self.feature_types_metadata.feature_types_raw
+
+        categorical_featnames = feature_types['category'] + feature_types['object'] + feature_types['bool']
+        continuous_featnames = feature_types['float'] + feature_types['int']  # + self.__get_feature_type_if_present('datetime')
+        language_featnames = []  # TODO: Disabled currently, have to pass raw text data features here to function properly
```
jwmueller, line 46: Should we merge this PR before this is fixed
Seems like a quality-of-life PR shouldn't be breaking functionality, or is there some urgency to merge this?

autogluon/utils/tabular/ml/models/abstract/abstract_model.py
jwmueller: In the future, we may want the ability for different models to eg consider the top k of the provided ngrams rather than ignoring them entirely, so consider adding TODO here.

autogluon/utils/tabular/ml/models/tabular_nn/tabular_nn_model.py
```diff
@@ -573,15 +573,10 @@ def _get_types_of_features(self, df, skew_threshold, embed_min_categories, use_n
...
-            vectorizers_featnames = self.__get_feature_type_if_present('vectorizers')
-            nlp_featnames = self.__get_feature_type_if_present('nlp')
-            self.feature_types_metadata['int'] = [feature for feature in self.__get_feature_type_if_present('int') if feature not in vectorizers_featnames]
-            self.feature_types_metadata['object'] = [feature for feature in self.__get_feature_type_if_present('object') if feature not in nlp_featnames]
+        feature_types = self.feature_types_metadata.feature_types_raw
 
-        categorical_featnames = self.__get_feature_type_if_present('object') + self.__get_feature_type_if_present('bool')
-        continuous_featnames = self.__get_feature_type_if_present('float') + self.__get_feature_type_if_present('int') + self.__get_feature_type_if_present('datetime')
+        categorical_featnames = feature_types['category'] + feature_types['object'] + feature_types['bool']
+        continuous_featnames = feature_types['float'] + feature_types['int']  # + self.__get_feature_type_if_present('datetime')
```
jwmueller, line 579: why ignore datetime?

Innixma:
> looks good, but if the goal of this is to make things clearer/more extensible for the user, the docs could be clarified to make this more helpful (as well as the printed names of the derived text features).
> 
> Second concern: why does this PR break passing text to LR model? And does NN model still receive datetime, or is that ignored now too?

Will try to clarify docs/names

Regarding second concern:

1. LR was already broken for text by the categorical optimization done in 0.0.11, since the text is no longer present in the categorical and LR needs the text to generate sparse ngrams. This fixes that so it doesn't crash, but a true fix will have to be done later as a larger effort, as we don't want to always have text in memory, especially when it takes up a lot of space.

2. NN still receives datetime, datetime was always transformed into int/float anyways, and NN didn't have any special handling of datetime, so it doesn't need to explicitly reference it with the new metadata layout (uses raw instead of special).

autogluon/utils/tabular/ml/models/tabular_nn/tabular_nn_model.py
Innixma: Already in float/int. The way metadata is structured has changed, raw feature_types are the actual raw DataFrame dtypes (after combining into families). Raw has no context of the special meaning behind features, that is what Special is for.

autogluon/task/tabular_prediction/predictor.py
Innixma: Added

autogluon/task/tabular_prediction/predictor.py
Innixma: Added more clarification

autogluon/task/tabular_prediction/tabular_prediction.py
Innixma: I had the same urge to change it originally. Reason I didn't was because of backwards compatibility, but I think very few people probably set this value beforehand and it was essentially internal, so its probably safe to change. Done. Also improved description.

autogluon/task/tabular_prediction/tabular_prediction.py
Innixma: Added

autogluon/task/tabular_prediction/tabular_prediction.py
Innixma: Added

autogluon/task/tabular_prediction/tabular_prediction.py
Innixma: Added

autogluon/utils/tabular/features/abstract_feature_generator.py
Innixma: Added

autogluon/utils/tabular/ml/models/abstract/abstract_model.py
Innixma: Added both

autogluon/utils/tabular/features/feature_types_metadata.py
Innixma: feature types metadata isn't made to be an authority on what the dtypes mean or what ones are valid. That is between the feature generator and the models, the metadata is just the generic link between them.

Added TODO to AutoMLFeatureGenerator to do this.

Innixma:
> In future PR, should we add support for user to declare `TabularDataset.feature_types`?
> Otherwise we should remove the argument in this PR if it'll never be used now that `feature_types` is a custom object.
> 
> Assuming it remains, in order for a user to declare their own feature_types, there will need to be a clearer description of the possible types which we can put at the top of the class `FeatureTypesMetadata` (or in the TabularDataset documentation).

Yes in future we should allow this.

autogluon/utils/tabular/ml/models/lr/lr_model.py
Innixma: As explained offline, this is not breaking.

autogluon/utils/tabular/ml/models/knn/knn_model.py
Innixma: Added

autogluon/utils/tabular/ml/models/abstract/abstract_model.py
Innixma: Agreed, added. Also improved documentation on some of the other arguments.

Innixma:
> When printing outputs to user, should we replace `nlp` with `text` in the feature names? as ML beginners might not understand otherwise.
> Also many users may find names like `nlp_categorical`, `nlp_special` confusing when they scan the output.
> Could perhaps name the features:
> `nlp_text` -> `text`
> `nlp_category` -> `text_as_category`
> `nlp_ngram` -> `text_as_ngram`
> `nlp_special` -> `text_as_?` (not sure why special and ratio are currently used interchangeably)

Good point, changed.

szha:
Job PR-548-4 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-548/4/index.html

Issue #544, opened 2020-06-29
Function on AutoGluon that returns a confusion matrix corresponding to the predict


calebeaires:
I have done some tests with this lib. Is is fast and accurate. I was wondering how to call some function from evaluate objects that returns a **confusion matrix corresponding to the  predict** while using the processo of TabularPrediction

Innixma:
This is not currently implemented, but I think would be a great addition. Contributions are welcome!

alan-aipe:
@Innixma : May I take this task up? I would like to start contributing to this project and this seems like a good first issue as you have rightfully pointed out.

Innixma:
@alan-aipe : Certainly! I've sent you an email with some tips on getting started.

alan-aipe:
Seems that I did not receive the email yet. Can you forward it to alanaipe@outlook.com?

alan-aipe:
Thanks for the mail. I will start working on this :)

alan-aipe:
@Innixma : Can you have a look at the PR. It seems the CI failed during BUILD_DOC phase because of `ImportError: cannot import name '_psutil_linux' from 'psutil' (/var/lib/jenkins/miniconda3/envs/autogluon_docs/lib/python3.7/site-packages/psutil/__init__.py)` . Let me know if this was caused because of my code as I am not able to figure this out.

Innixma:
@alan-aipe Sorry for the confusion, this started happening 2 days ago due to a psutil update (I believe), I will be debugging it to try to fix the CI. Once the fix is merged I'll update this thread and you can rebase your PR onto the fix to get CI to pass.

jwmueller:
Added in:  https://github.com/awslabs/autogluon/pull/571

PR #571, opened 2020-07-20
Added confusion matrix metric for binary and multi-class classification

alan-aipe:
*Issue #544 :*

**Motivation**
Added confusion matrix metric into tabular classification metrics library with corresponding unit tests

**Implementation**
- Checked if both true and predicted targets are of same shape and are of type binary or multi-class
- Checked if labels (if passed as a parameter) is a 1-D binary or multi-class array without duplicate elements
- Checked if weights (if passed as a parameter) is a 1-D binary, multi-class or continuous array
- Removed indexes having targets that are not included in the desired label array
- Returned coordinate matrix with the processed targets, labels and weights

**Testing**
Unit tests passed 

### DIFF
--- a/autogluon/utils/tabular/metrics/classification_metrics.py
+++ b/autogluon/utils/tabular/metrics/classification_metrics.py
@@ -2,6 +2,9 @@ import logging
 
 import numpy as np
 import pandas as pd
+from scipy.sparse import coo_matrix
+from sklearn.utils.multiclass import unique_labels
+from sklearn.utils import check_consistent_length
 try:
     from sklearn.metrics._classification import _check_targets, type_of_target
 except:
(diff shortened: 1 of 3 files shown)


Innixma:
Thanks for the PR and quick turn around! The CI is a bit unstable due to what we think is a psutil update, so I'll be trying to get it fixed in the next few days: I don't believe the failure is due to your code changes.

Innixma:
@alan-aipe I believe that the CI is fixed on mainline. Can you rebase your PR onto mainline to get the CI fix commit?

alan-aipe:
@Innixma  It does not seem to fix the issue. My current branch is up-to-date with the mainline. It seems that this issue is a known one and is being faced in other projects as well. Some ways others solved it are as follows :
1. [Here](https://github.com/giampaolo/psutil/issues/1288#issuecomment-401303133), the resolver uninstalled psutils then ran a [purge script](https://github.com/giampaolo/psutil/blob/master/scripts/internal/purge_installation.py) to remove left-over files from site-packages and reinstalled it.
2. [Here](https://github.com/giampaolo/psutil/issues/1689#issuecomment-594964580), the resolver solved it by just reinstalling it using `python3.7 -m pip install --user psutil` but he is not sure whether this solution will work for cron-jobs

Hope the above info helps. Please retrigger CI build once its fixed. Let me know if you need any action from my side.

jwmueller:
Thanks for your PR!
Could you briefly clarify what part of your `confusion_matrix()` differs from sklearn's:
https://github.com/scikit-learn/scikit-learn/blob/fd237278e/sklearn/metrics/_classification.py#L199
and why these differences are better?  For instance, why remove the `normalize` argument?

jwmueller:
Could you compute confusion-matrix here:
https://github.com/awslabs/autogluon/blob/master/autogluon/utils/tabular/ml/learner/abstract_learner.py#L431
and correspondingly print it here: 
https://github.com/awslabs/autogluon/blob/master/autogluon/utils/tabular/ml/learner/abstract_learner.py#L444
so the user will see the confusion matrix when they call `Predictor.evaluate_predictions()`

alan-aipe:
@jwmueller : 

> Could you briefly clarify what part of your confusion_matrix() differs from sklearn's:
https://github.com/scikit-learn/scikit-learn/blob/fd237278e/sklearn/metrics/_classification.py#L199
and why these differences are better? For instance, why remove the `normalize` argument?

I had two reasons for writing my own function rather than using sklearn:
1) In sklearn, labels and weights are not thoroughly tested before applying to the core algorithm. For example, in sklearn, length of the weight array is checked with length of the targets. But this is not sufficient. if weights = [[1.2, 0.5], 0.7] and targets = ["cat", "dog"], sklearn will go ahead and try to calculate the confusion matrix with this and results in error `ValueError: setting an array element with a sequence.` in coo_matrix calculation. For the novice autogluon user, the error may be confusing. So I added appropriate checks for labels and weights
2) In sklearn's implementation,  target values are converted to integers before eliminating items not in labels. This is sub-optimal. In my opinion, the unwanted items should be taken off before converting to integers because we iterate over each element in the numpy array for that conversion. This will be beneficial for large datasets

Regarding `normalize`,  I think that idea of confusion matrix evolved out of TP, FP, TN, FN values which are all numbers. Normalization is not a part of the core algorithm to compute confusion matrix. We should have a separate function called `normalize_matrix` so that the code is concise and more reusable. 

These are my personal opinions. Let me know you comments on this

alan-aipe:
@jwmueller 
> Could you compute confusion-matrix here:
https://github.com/awslabs/autogluon/blob/master/autogluon/utils/tabular/ml/learner/abstract_learner.py#L431
and correspondingly print it here:
https://github.com/awslabs/autogluon/blob/master/autogluon/utils/tabular/ml/learner/abstract_learner.py#L444
so the user will see the confusion matrix when they call Predictor.evaluate_predictions()

Added it in the auxillary metric for binary and multi-class problem type. There is no need to print it explicitly as perf_dict as a whole is dumped to log

Innixma:
@alan-aipe Could you try rebasing again? I believe #577 fixed the CI issue (heres hoping)

jwmueller:
@alan-aipe 

> I had two reasons for writing my own function rather than using sklearn
> 
> Regarding `normalize`, I think that idea of confusion matrix evolved out of TP, FP, TN, FN values which are all numbers. Normalization is not a part of the core algorithm to compute confusion matrix. We should have a separate function called `normalize_matrix` so that the code is concise and more reusable.
> 
> These are my personal opinions. Let me know you comments on this

Your reasons sounds good to me, I agree normalize isn't really critical, just figured it's nice have as an option.

jwmueller:
> @jwmueller
> 
> > Could you compute confusion-matrix here:
> > https://github.com/awslabs/autogluon/blob/master/autogluon/utils/tabular/ml/learner/abstract_learner.py#L431
> > and correspondingly print it here:
> > https://github.com/awslabs/autogluon/blob/master/autogluon/utils/tabular/ml/learner/abstract_learner.py#L444
> > so the user will see the confusion matrix when they call Predictor.evaluate_predictions()
> 
> Added it in the auxillary metric for binary and multi-class problem type. There is no need to print it explicitly as perf_dict as a whole is dumped to log

Does the confusion matrix still look good when printed with `json.dumps()`? This is used to print out `perf_dict` in a nice format
I'm concerned it'll look ugly if there are many classes (eg. 300+), in which case it could still be preferable to print the confusion matrix separately

szha:
Job PR-571-5 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-571/5/index.html

alan-aipe:
@Innixma : CI passed. Thanks for the deep dive and fixing it. Can you also review the PR and let me know your comments so that I can address both yours and @jwmueller  comments in a single push

autogluon/utils/tabular/metrics/classification_metrics.py
```diff
@@ -254,3 +257,49 @@ def prior_log_loss(frac_pos, task):
     score = (pac - base_pac) / np.maximum(eps, (1 - base_pac))
 
     return score
+
+
+def confusion_matrix(solution, prediction, labels=None, weights=None):
```
Innixma, line 262: Regarding the 'normalize' argument, I would highly recommend we add this, as most of the time this is more important than the raw numbers for analytics. This should live in this function and be applied after the standard confusion matrix is constructed.

autogluon/utils/tabular/metrics/classification_metrics.py
```diff
@@ -254,3 +257,49 @@ def prior_log_loss(frac_pos, task):
...
+    y_type, solution, prediction = _check_targets(solution, prediction)
+    if solution.size == 0 or prediction.size == 0:
+        return np.zeros((n_labels, n_labels), dtype=np.int)
```
daveharmon, line 265: has `n_labels` been set at this point? Wouldn't we want to raise an Exception here?

daveharmon: Should add a unit test for this as well

szha:
Job PR-571-7 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-571/7/index.html

alan-aipe:
@Innixma Can you review and approve the PR if it looks ok

PR #578, opened 2020-07-22
Tabular: Refactor FeatureGenerator (Part 1)

Innixma:
*Issue #, if available:*
#579

*Description of changes:*

- Fixed datetime features to avoid repeated appending to feature name list (no impact on predictions, only on memory usage after many calls to predict).
- Fixed similar defect in text_as_category features.
- Removed inplace=True from reset_index calls to avoid outward propagation of index changes.
- Disabled default df.copy(deep=True) call in TabularDataset init, to avoid unnecessarily doubling memory usage.
- Various code cleanup

### DIFF
--- a/autogluon/task/tabular_prediction/dataset.py
+++ b/autogluon/task/tabular_prediction/dataset.py
@@ -34,6 +34,7 @@ class TabularDataset(pd.DataFrame):
     copy : bool (optional, default=False)
         If True and `df` is passed, then `df` will be deep copied, resulting in two instances of the provided data in memory.
         If False, the original `df` and the new `TabularDataset` will share the same underlying data.
+        Set `copy=True` if you don't want AutoGluon to be able to modify the original DataFrame during it's training process (and your machine has enough memory).
 
     Attributes
     ----------
--- a/autogluon/task/tabular_prediction/dataset.py
+++ b/autogluon/task/tabular_prediction/dataset.py
@@ -30,7 +30,10 @@ class TabularDataset(pd.DataFrame):
     subsample : int (optional)
         If specified = k, we only keep first k rows of the provided dataset.
     name : str (optional)
-         Optional name to assign to dataset (has no effect beyond being accessible via `TabularDataset.name`).
+        Optional name to assign to dataset (has no effect beyond being accessible via `TabularDataset.name`).
+    copy : bool (optional, default=False)
+        If True and `df` is passed, then `df` will be deep copied, resulting in two instances of the provided data in memory.
+        If False, the original `df` and the new `TabularDataset` will share the same underlying data.
 
     Attributes
     ----------
@@ -71,6 +74,7 @@ class TabularDataset(pd.DataFrame):
         feature_types = kwargs.get('feature_types', None)
         df = kwargs.get('df', None)
         subsample = kwargs.get('subsample', None)
+        copy = kwargs.get('copy', False)
         construct_from_df = False  # whether or not we are constructing new dataset object from scratch based on provided DataFrame.
         # if df is None and file_path is None: # Cannot be used currently!
         #     raise ValueError("Must specify either named argument 'file_path' or 'df' in order to construct tabular Dataset")
@@ -80,7 +84,8 @@ class TabularDataset(pd.DataFrame):
                 raise ValueError("'df' must be existing pandas DataFrame. To read dataset from file instead, use 'file_path' string argument.")
             if file_path is not None:
                 warnings.warn("Both 'df' and 'file_path' supplied. Creating dataset based on DataFrame 'df' rather than reading from file_path.")
-            df = df.copy(deep=True)
+            if copy:
+                df = df.copy(deep=True)
         elif file_path is not None:  # Read from file to create dataset
             construct_from_df = True
             df = load_pd.load(file_path)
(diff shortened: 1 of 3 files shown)


szha:
Job PR-578-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-578/1/index.html

szha:
Job PR-578-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-578/2/index.html

szha:
Job PR-578-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-578/3/index.html

autogluon/task/tabular_prediction/dataset.py
```diff
@@ -30,7 +30,10 @@ class TabularDataset(pd.DataFrame):
     subsample : int (optional)
         If specified = k, we only keep first k rows of the provided dataset.
     name : str (optional)
-         Optional name to assign to dataset (has no effect beyond being accessible via `TabularDataset.name`).
+        Optional name to assign to dataset (has no effect beyond being accessible via `TabularDataset.name`).
+    copy : bool (optional, default=False)
+        If True and `df` is passed, then `df` will be deep copied, resulting in two instances of the provided data in memory.
```
jwmueller, line 35: Should we add advice where this copy might be useful?  Eg: 
"Set `copy=True` if you don't want AutoGluon to modify your DataFrame during it's training process."

autogluon/utils/tabular/features/abstract_feature_generator.py
```diff
@@ -441,5 +295,128 @@ def drop_duplicate_features(X):
...
+        logger.log(20, 'Original Features (raw dtypes):')
+        for key, val in self.feature_type_family_init_raw.items():
+            if val: logger.log(20, '\t%s features: %s' % (key, len(val)))
+        logger.log(20, 'Original Features (inferred dtypes):')
+        for key, val in self.feature_type_family.items():
+            if val: logger.log(20, '\t%s features: %s' % (key, len(val)))
+        logger.log(20, 'Generated Features (special dtypes):')
+        for key, val in self.feature_types_metadata.feature_types_special.items():
+            if val: logger.log(20, '\t%s features: %s' % (key, len(val)))
+        logger.log(20, 'Final Features (raw dtypes):')
```
jwmueller, line 414: consider substitution: "Final" -> "Processed"

autogluon/utils/tabular/features/abstract_feature_generator.py
```diff
@@ -441,5 +295,128 @@ def drop_duplicate_features(X):
...
+        for key, val in self.feature_types_metadata.feature_types_raw.items():
+            if val: logger.log(20, '\t%s features: %s' % (key, len(val)))
+        logger.log(20, 'Final Features:')
```
jwmueller, line 417: I think all 4 print statements ("... Features ... :") should make it more clear what the difference between these 4 feature groups is.
Eg. user may not understand what's the difference between "Final Features (raw dtypes)" and "Final Features" or the difference between "raw dtypes" and "inferred dtypes"

autogluon/utils/tabular/features/auto_ml_feature_generator.py
```diff
@@ -84,22 +77,29 @@ def generate_features(self, X: DataFrame):
...
+                self.feature_type_family_generated['text_special'] += list(X_text_special_combined.columns)
+            X_features = X_features.join(X_text_special_combined)
 
         if self.feature_transformations['datetime']:
             for datetime_feature in self.feature_transformations['datetime']:
                 X_features[datetime_feature] = pd.to_datetime(X[datetime_feature])
                 X_features[datetime_feature] = pd.to_numeric(X_features[datetime_feature])  # TODO: Use actual date info
-                self.feature_type_family_generated['datetime'].append(datetime_feature)
+                if not self.fit:
+                    self.feature_type_family_generated['datetime'].append(datetime_feature)
```
jwmueller, line 102: Can you confirm all models, like the NeuralNet, are still actually utilizing the datetime feature after these changes?

Innixma: This does not affect downstream models, it only means that the feature name is appended to self.feature_type_family_generated['datetime'] during training, instead of at every call.

So if you have a datetime feature 'date', then after fit

`self.feature_type_family_generated['datetime'] == ['date']`

if you called predict 5 times:

`self.feature_type_family_generated['datetime'] == ['date', 'date', 'date', 'date', 'date', 'date']`

This wouldn't impact anything except take increasingly more memory to store the long list of repeated 'date' strings, particularly important if doing live-inference over an extended period of time on single-rows. This is the defect I fixed (same with text_as_category).

autogluon/utils/tabular/features/abstract_feature_generator.py
Innixma: I gave it a shot but couldn't figure out a nice way to communicate it without making the logs overly verbose. Feel free to adjust these logs in a separate PR if you have suggestions though.

autogluon/utils/tabular/features/abstract_feature_generator.py
Innixma: Added

autogluon/task/tabular_prediction/dataset.py
Innixma: Added

szha:
Job PR-578-4 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-578/4/index.html

PR #587, opened 2020-07-29
Fix wrong function overloading.

yongzhengqi:
I was using RL NAS part of Autogluon. But I found that `rl_scheduler.get_best_config()` wasn't returning the best config.

So I check the code of Autogluon and found wrong overloading. The `RLSearcher` is inherited from `BaseSearcher`. While `BaseSearcher` has the right `get_best_config()`, `RLSearcher` overloaded this function with simply `inference()` which is obviously wrong. 

So I remove the wrong overloading and everything works fine again.

### DIFF
--- a/autogluon/searcher/rl_controller.py
+++ b/autogluon/searcher/rl_controller.py
@@ -60,9 +60,6 @@ class RLSearcher(BaseSearcher):
     def get_config(self, **kwargs):
         return self.controller.sample()[0]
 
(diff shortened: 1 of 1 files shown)


szha:
Job PR-587-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-587/1/index.html

PR #588, opened 2020-07-29
Tabular: Optimized OOF memory usage

Innixma:
*Issue #, if available:*

*Description of changes:*

- Optimized OOF memory usage by converting from float64 -> float32
- Optimized KNN memory usage / disk usage / speed by converting to float32 and removing unnecessary categorical removal (Already handled by AG_args_fit defaults). This typically reduces disk/memory usage of KNN by ~20-25%.

### DIFF
--- a/autogluon/utils/tabular/ml/models/knn/knn_model.py
+++ b/autogluon/utils/tabular/ml/models/knn/knn_model.py
@@ -3,6 +3,7 @@ import pickle
 import sys
 import time
 
+import numpy as np
 import psutil
 from sklearn.neighbors import KNeighborsClassifier, KNeighborsRegressor
 from .knn_utils import FAISSNeighborsClassifier, FAISSNeighborsRegressor
@@ -28,9 +29,8 @@ class KNNModel(AbstractModel):
             return KNeighborsClassifier
 
     def preprocess(self, X):
-        cat_columns = X.select_dtypes(['category']).columns
-        X = X.drop(cat_columns, axis=1)  # TODO: Test if crash when all columns are categorical
         X = super().preprocess(X).fillna(0)
+        X = X.to_numpy(dtype=np.float32)
         return X
 
     def _set_default_params(self):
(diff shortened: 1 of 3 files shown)


szha:
Job PR-588-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-588/1/index.html

szha:
Job PR-588-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-588/2/index.html

autogluon/utils/tabular/ml/models/knn/knn_model.py
```diff
@@ -28,9 +27,8 @@ def _get_model_type(self):
             return KNeighborsClassifier
 
     def preprocess(self, X):
-        cat_columns = X.select_dtypes(['category']).columns
-        X = X.drop(cat_columns, axis=1)  # TODO: Test if crash when all columns are categorical
         X = super().preprocess(X).fillna(0)
+        X = X.to_numpy(dtype=np.float32)
```
jwmueller, line 31: What happens if `X` originally only contained `int` types say?

Innixma: They would be converted to float32 and the accuracy/predictions would be the same. However, if they were all uint8, it would be slightly less efficient to convert them. This is a level of optimization that we could consider in future, but would have to be tested.

Innixma:
> looks good. Not sure it matters practically, but could consider adding optional arg: `AbstractModel.predict_proba(..., convert_type = True)` and then inside the function have:
> 
> ```
> if convert_type:
>     y_pred_proba = y_pred_proba.astype(np.float32)
> ```

I think we would rather have it be an init argument to the model, such that it is consistent. I suppose you intend for this flag to give an option to avoid the conversion call (speedup inference)?

PR #592, opened 2020-07-30
[HPO][BugFix] Fix the bug in scheduler

sxjscience:
According to the document of dask: https://distributed.dask.org/en/latest/api.html#distributed.Future.result

`.result(timeout)` will raise a `distributed.TimeoutError`. However, in the code base, the try + except just handles the `TimeoutError`.

We may use the following tiny example to understand the issue. If the exception is handled correctly, it will print 123. However, it actually raises the Exception.

```python
import distributed
try:
    raise(distributed.TimeoutError)
except TimeoutError:
    print(123)
```

Output:
```
TimeoutError: 
```

@zhreshold

### DIFF
--- a/autogluon/scheduler/scheduler.py
+++ b/autogluon/scheduler/scheduler.py
@@ -2,6 +2,8 @@
 import os
 import pickle
 import logging
+import sys
+import distributed
 from warnings import warn
 import multiprocessing as mp
 from collections import OrderedDict
@@ -196,8 +198,11 @@ class TaskScheduler(object):
         for task_dict in self.scheduled_tasks:
             try:
                 task_dict['Job'].result(timeout=timeout)
-            except TimeoutError as e:
+            except distributed.TimeoutError as e:
                 logger.error(str(e))
+            except:
+                print("Unexpected error:", sys.exc_info()[0])
+                raise
             self._clean_task_internal(task_dict)
         self._cleaning_tasks()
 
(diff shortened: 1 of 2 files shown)


szha:
Job PR-592-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-592/1/index.html

autogluon/scheduler/scheduler.py
```diff
@@ -196,8 +198,11 @@ def join_jobs(self, timeout=None):
         for task_dict in self.scheduled_tasks:
             try:
                 task_dict['Job'].result(timeout=timeout)
-            except TimeoutError as e:
+            except distributed.TimeoutError as e:
                 logger.error(str(e))
+            except:
+                print("Unexpected error:", sys.exc_info()[0])
```
mli, line 204: while not use logger but print?

sxjscience:
@mseeger Yes, I think it led to increased stability. For me, my tuning job will occasionally fail if I do not fix this. Let me also add a testcase.

szha:
Job PR-592-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-592/3/index.html

zhreshold:
@sxjscience  good catch,  if you find other dask specific exception handling issue I think it significantly improve the failure rate
@tlienart This might be relevant to sagemaker failure rate issue as well.

tlienart:
When this is in I can try relaunching the same experiments and report

Issue #595, opened 2020-07-30
`NeuralNetClassifier` took too long to train


mli:
sometimes the `NeuralNetClassifier` took way long time than others. for example, here is an example to train on a small scale dataset  

```python
!kaggle d download mlg-ulb/creditcardfraud -p ../input/
!cd ../input/; unzip creditcardfraud.zip -d creditcardfraud

import pandas as pd 
from autogluon import TabularPrediction as task

data = pd.read_csv('../input/creditcardfraud/creditcard.csv')
model = task.fit(train_data=task.Dataset(data), label='Class')
```

The normal time to train a classifier is about 5sec, while the `NeuralNetClassifier` needs >1h (didn't finish). The only specialty I feel is that it's a unbalanced dataset, with ~0.1% positive examples. Maybe negative sampling is used here?

An easy way to fix it is using `time_limits`. It would be good if we can hint to use it for a beginner users. E.g. when we detected a classifier used way more time than others, e.g. `10 * avg_time` and `time_limits` isn't set, we could print a message saying that you can use `time_limits`, or instructions to filter out this classifier. Also it could be great if we can show the training progress of a classifier it needs a long time to train.

```
No output_directory specified. Models will be saved in: AutogluonModels/ag-20200730_190141/
Beginning AutoGluon training ...
AutoGluon will save models to AutogluonModels/ag-20200730_190141/
AutoGluon Version:  0.0.12
Train Data Rows:    284807
Train Data Columns: 31
Preprocessing data ...
Here are the 2 unique label values in your data:  [0, 1]
AutoGluon infers your prediction problem is: binary  (because only two unique label-values observed).
If this is wrong, please specify `problem_type` argument in fit() instead (You may specify problem_type as one of: ['binary', 'multiclass', 'regression'])

Selected class <--> label mapping:  class 1 = 1, class 0 = 0
Train Data Class Count: 2
Feature Generator processed 284807 data points with 30 features
Original Features (raw dtypes):
	float64 features: 30
Original Features (inferred dtypes):
	float features: 30
Generated Features (special dtypes):
Final Features (raw dtypes):
	float features: 30
Final Features:
	float features: 30
	Data preprocessing and feature engineering runtime = 1.68s ...
AutoGluon will gauge predictive performance using evaluation metric: accuracy
To change this, specify the eval_metric argument of fit()
AutoGluon will early stop models using evaluation metric: accuracy
Fitting model: RandomForestClassifierGini ...
[...]

PR #598, opened 2020-07-31
more aggressive NN early stopping

jwmueller:
Change NN to early-stop if validation-score hasn't *strictly* improved in `epochs_wo_improve` consecutive epochs.
Suppose epoch `e1` = epoch where validation-score first hits max-value and validation-score plateaus at this point. 
After this change, we now consider the "best" model the one from epoch `e2 = e1 + epochs_wo_improve`. 
Thus we still train a little bit beyond the point where validation-score plateaus (perhaps margin of decision-boundary can still be improved), but the amount of extra-training is strictly limited by the user-controllable hyperparameter `epochs_wo_improve` (rather than before where the NN could continue training for a long time while validation-score remained plateaued).

So users who want to be sure they are achieving maximal accuracy (at potential cost of long training-time) can still do this by appropriately increasing the `epochs_wo_improve` hyperparameter. 

*Issue #, if available:*

https://github.com/awslabs/autogluon/issues/595

### DIFF
--- a/autogluon/utils/tabular/ml/models/tabular_nn/tabular_nn_model.py
+++ b/autogluon/utils/tabular/ml/models/tabular_nn/tabular_nn_model.py
@@ -262,6 +262,7 @@ class TabularNeuralNetModel(AbstractModel):
         best_val_metric = -np.inf  # higher = better
         val_metric = None
         best_val_epoch = 0
+        val_improve_epoch = 0  # most recent epoch where validation-score strictly improved
         num_epochs = params['num_epochs']
         if val_dataset is not None:
             y_val = val_dataset.get_labels()
@@ -334,11 +335,12 @@ class TabularNeuralNetModel(AbstractModel):
                 cumulative_loss += loss.sum()
             train_loss = cumulative_loss/float(train_dataset.num_examples)  # training loss this epoch
             if val_dataset is not None:
-                # val_metric = self.evaluate_metric(test_dataset)  # Evaluate after each epoch
                 val_metric = self.score(X=val_dataset, y=y_val, eval_metric=self.stopping_metric, metric_needs_y_pred=self.stopping_metric_needs_y_pred)
             if (val_dataset is None) or (val_metric >= best_val_metric) or (e == 0):  # keep training if score has improved
                 if val_dataset is not None:
                     if not np.isnan(val_metric):
+                        if val_metric > best_val_metric:
+                            val_improve_epoch = e
                         best_val_metric = val_metric
                 best_val_epoch = e
                 # Until functionality is added to restart training from a particular epoch, there is no point in saving params without test_dataset
@@ -361,8 +363,8 @@ class TabularNeuralNetModel(AbstractModel):
                 if val_dataset is not None and (not np.isnan(val_metric)):  # TODO: This might work without the if statement
                     # epoch must be number of epochs done (starting at 1)
                     reporter(epoch=e+1, validation_performance=val_metric, train_loss=float(train_loss.asscalar()))  # Higher val_metric = better
-            if e - best_val_epoch > epochs_wo_improve:
-                break
+            if e - val_improve_epoch > epochs_wo_improve:   
+                break  # early-stop if validation-score hasn't strictly improved in `epochs_wo_improve` consecutive epochs
             if time_limit:
                 time_elapsed = time.time() - start_time
                 time_left = time_limit - time_elapsed
--- a/autogluon/utils/tabular/ml/models/tabular_nn/tabular_nn_model.py
+++ b/autogluon/utils/tabular/ml/models/tabular_nn/tabular_nn_model.py
@@ -363,7 +363,7 @@ class TabularNeuralNetModel(AbstractModel):
                 if val_dataset is not None and (not np.isnan(val_metric)):  # TODO: This might work without the if statement
                     # epoch must be number of epochs done (starting at 1)
                     reporter(epoch=e+1, validation_performance=val_metric, train_loss=float(train_loss.asscalar()))  # Higher val_metric = better
-            if e - val_improve_epoch > epochs_wo_improve:   
+            if e - val_improve_epoch > epochs_wo_improve:
                 break  # early-stop if validation-score hasn't strictly improved in `epochs_wo_improve` consecutive epochs
             if time_limit:
                 time_elapsed = time.time() - start_time


jwmueller:
I believe this strategy is preferred vs the simple approach of just returning the model from epoch `e1` (defined above).
And does not involve extra computation as we anyway need to train the extra `epochs_wo_improve` epochs to determine that the validation-loss is no longer improving.

Innixma:
> @Innixma I believe this strategy is preferred vs the simple approach of just returning the model from epoch `e1` (defined above).
> And does not involve extra computation as we anyway need to train the extra `epochs_wo_improve` epochs to determine that the validation-loss is no longer improving.

Say `epochs_wo_improve=5`

```
E0: 0.5 auc
E1: 0.6 auc
E2: 0.7 auc
E3: 0.8 auc
E4: 0.8 auc
E5: 0.8 auc
E6: 0.7 auc
E7: 0.6 auc
E8: 0.6 auc
```

In this case with the proposed logic, `e1=3`, and `e2=8`

Selecting `e2=8` is not ideal. e2 should either be 3, 4, or 5. I propose 5, such that we keep track of the last epoch which matched the validation score of `e1`, and select that epoch when early stopping.

szha:
Job PR-598-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-598/1/index.html

szha:
Job PR-598-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-598/2/index.html

jwmueller:
> If you still say that `e2=8` is optimal, are you saying that we should not trust the validation score, and that we intrinsically trust that the model is improving?

No sorry, my example was only referring to the case where validation score plateaus (is totally constant for all epochs past `e1`).
If your example were instead modified to have: val-score = 0.8 for all epochs >= 3, then yes I'm suggesting `e2=8` is the one we want to return with `epochs_wo_improve=5`.

In the case of your actual example, the code in this PR should return: `e2= 5`, which I do think is better than `e2=3` for reasons of potential margin-increase.  To translate your example into the code of the PR, basically `best_val_epoch = 5` but  `val_improve_epoch` would remain stuck at 3, so training would still `break` at E8 with `epochs_wo_improve=5`.

Innixma:
> @Innixma
> 
> > If you still say that `e2=8` is optimal, are you saying that we should not trust the validation score, and that we intrinsically trust that the model is improving?
> 
> No sorry, my example was only referring to the case where validation score plateaus (is totally constant for all epochs past `e1`).
> If your example were instead modified to have: val-score = 0.8 for all epochs >= 3, then yes I'm suggesting `e2=8` is the one we want to return with `epochs_wo_improve=5`.
> 
> In the case of your actual example, the code in this PR should return: `e2= 5`, which I do think is better than `e2=3` for reasons of potential margin-increase. To translate your example into the code of the PR, basically `best_val_epoch = 5` (and this is the model that's saved/reloaded for prediction) but `val_improve_epoch` would remain stuck at 3, so training would still `break` at E8 with `epochs_wo_improve=5`.

I see, makes sense. This differs from what your original description was so I originally thought it was doing `e2=8` because of `e2 = e1 + epochs_wo_improve`, which is not quite accurate to the code.

I think this code is correct, but before I approve I will benchmark before/after versions to ensure we don't experience degradation of performance.

szha:
Job PR-598-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-598/3/index.html

szha:
Job PR-598-4 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-598/4/index.html

szha:
Job PR-598-5 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-598/5/index.html

szha:
Job PR-598-6 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-598/6/index.html

Innixma:
Did a benchmark of this change. Looks to be mostly matching previous scores with 1 major change in the `kr-vs-kp` dataset.

`Prior score: 0.999961 AUC`
`New score: 0.999883 AUC`

Average epochs trained by NN went from 300 -> 27
The NN is the best model here, and it gets a very very high validation score (often 1). So after the change, the model likely wasn't fully fit even though it had a perfect validation AUC score, which results in a negligible (absolute) and major (relative) error increase.

On the upside, the NN training took only 1/10th of the time for `kr-vs-kp`, and the NN was previously taking 90%+ of the total training time among all models.

I think this will only ever occur on extremely easy problems, where perhaps the difference really isn't that important, and it could be argued we shouldn't be benchmarking on such easy datasets to begin with.

Innixma:
Here is the run for creditfraud on my laptop to show that it early stops reasonably:

```
Epoch 0.  Train loss: 0.06224216, Val accuracy: 0.9996489996489997
Epoch 1.  Train loss: 0.0044396217, Val accuracy: 0.9996489996489997
Epoch 2.  Train loss: 0.0030911337, Val accuracy: 1.0
Epoch 3.  Train loss: 0.0026924242, Val accuracy: 0.9996489996489997
Epoch 4.  Train loss: 0.0022474146, Val accuracy: 0.9996489996489997
Epoch 5.  Train loss: 0.0021056526, Val accuracy: 0.9992979992979993
Epoch 6.  Train loss: 0.0018715413, Val accuracy: 1.0
Epoch 7.  Train loss: 0.0017852911, Val accuracy: 0.9996489996489997
Epoch 8.  Train loss: 0.0015773426, Val accuracy: 1.0
Epoch 9.  Train loss: 0.0014420052, Val accuracy: 0.9996489996489997
Epoch 10.  Train loss: 0.0013045863, Val accuracy: 1.0
Epoch 11.  Train loss: 0.0011904336, Val accuracy: 1.0
Epoch 12.  Train loss: 0.0011358258, Val accuracy: 1.0
Epoch 13.  Train loss: 0.0010043315, Val accuracy: 1.0
Epoch 14.  Train loss: 0.00096173154, Val accuracy: 1.0
Epoch 15.  Train loss: 0.0009741376, Val accuracy: 1.0
Epoch 16.  Train loss: 0.00090411986, Val accuracy: 1.0
Epoch 17.  Train loss: 0.00074588344, Val accuracy: 1.0
Epoch 18.  Train loss: 0.0008922371, Val accuracy: 0.9996489996489997
Epoch 19.  Train loss: 0.0007041674, Val accuracy: 1.0
Epoch 20.  Train loss: 0.00066614343, Val accuracy: 1.0
Epoch 21.  Train loss: 0.00063342834, Val accuracy: 1.0
Epoch 22.  Train loss: 0.00059692626, Val accuracy: 1.0
Epoch 23.  Train loss: 0.0006864279, Val accuracy: 1.0
Best model found in epoch 23. Val accuracy: 1.0
Saving AutogluonModels/ag-20200828_021941/models/NeuralNetClassifier/model.pkl
	1.0	 = Validation accuracy score
	808.1s	 = Training runtime
	0.28s	 = Validation runtime
```

jwmueller:
> I think this will only ever occur on extremely easy problems, where perhaps the difference really isn't that important, and it could be argued we shouldn't be benchmarking on such easy datasets to begin with.

Right this is what I've been saying all along regarding the NN continuing to expand the margin (for datasets where classes are entirely separable so 100% test accuracy is possible).  So there'll always be some tension between runtime and test accuracy for certain datasets, but at least now the user can adjust it themselves via `epochs_wo_improve` hyperparameter. We may consider different default value of this hyperparameter going forward as well, the current choice of 20 is fairly arbitrary...

PR #618, opened 2020-08-14
Add mqcnn

yx1215:
*Issue #, if available:*
This PR aims to enable mqcnn estimator in gluonts for autogluon. Currently it is an attempt and has some bugs. It would be very appreciated if anyone can give me some suggestions. 

There is a testing script ```examples/tabular/example_simple_tabular_forecasting.py```.

The script will fail when it comes to the ```Fitting model: weighted_ensemble_k0_l1 ...```.

### DIFF
--- a/autogluon/utils/tabular/ml/models/mqcnn/mqcnn_model.py
+++ b/autogluon/utils/tabular/ml/models/mqcnn/mqcnn_model.py
@@ -1,0 +1,44 @@
+import logging
+import re
+
+import numpy as np
+import pandas as pd
+from pandas import DataFrame
+from gluonts.model.seq2seq import MQCNNEstimator
+from sklearn.feature_extraction.text import TfidfVectorizer
+from sklearn.impute import SimpleImputer
+from sklearn.pipeline import FeatureUnion, Pipeline
+from sklearn.preprocessing import StandardScaler, QuantileTransformer
+from gluonts.model.seq2seq import MQCNNEstimator
+# from .hyperparameters.parameters import get_param_baseline, get_model_params, get_default_params, INCLUDE, IGNORE, ONLY
+# from .hyperparameters.searchspaces import get_default_searchspace
+# from .lr_preprocessing_utils import NlpDataPreprocessor, OheFeaturesGenerator, NumericDataPreprocessor
+from ...constants import BINARY, REGRESSION
+from ....ml.models.abstract.abstract_model import AbstractModel
+from .hyperparameters.parameters import get_default_parameters
+from gluonts.dataset.common import ListDataset
+from gluonts.dataset.field_names import FieldName
+
+
+class MQCNNModel(AbstractModel):
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.params = get_default_parameters()
+
+    def _fit(self, X_train, y_train, X_val=None, y_val=None, time_limit=None, **kwargs):
+        X_train = self.preprocess(X_train)
+        print(X_train)
+        estimator = MQCNNEstimator.from_hyperparameters(**self.params)
+        self.model = estimator.train(X_train)
+
+    def preprocess(self, X):
+        # TODO: transform dataframe into gluon-ts ListDataset
+        target_values = X.drop("index", axis=1).values
+        processed_X = ListDataset([
+            {
+                FieldName.TARGET: target,
+                FieldName.START: pd.Timestamp("2020-01-01", freq="1D"),
+            }
+            for (target, ) in zip(target_values,)
+        ], freq="D")
+        return processed_X
--- a/autogluon/utils/tabular/ml/models/mqcnn/mqcnn_model.py
+++ b/autogluon/utils/tabular/ml/models/mqcnn/mqcnn_model.py
@@ -3,42 +3,83 @@ import re
 
 import numpy as np
 import pandas as pd
-from pandas import DataFrame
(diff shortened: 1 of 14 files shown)


Innixma:
Thanks for creating a PR! One thing I notice is that you have tried to first get it working in task.fit, which is going to be a fair amount of work and should be the last thing we eventually do once the model and its API are nearly finalized.

Try to get it working as in the example code I referenced as a stand-alone model first: https://github.com/awslabs/autogluon/blob/master/examples/tabular/example_custom_model_tabular.py .

From there, we can talk about what exactly the input and output of this model is, and how it differs from what already exists in AutoGluon. That will determine the work outside of the model that needs to be done.

Also, please include an example from a real forecasting dataset, as a dummy dataset makes it hard to know if things are working properly. If you can provide me a public forecasting dataset I can make it available in S3 as is done for the dataset in the custom model example.

Innixma:
Also: Please rebase this PR with mainline to fix the CI failure caused by incorrect mxnet version.

yx1215:
> Also: Please rebase this PR with mainline to fix the CI failure caused by incorrect mxnet version.

I've just rebased and add a script ```examples/tabular/basic_mqcnn_model.py``` following the example you given. I was using a real COV19 data set from kaggle. I can send you the dataset through chime.

examples/tabular/basic_mqcnn_model.py
```diff
@@ -0,0 +1,111 @@
...
+    def predict(self, X, index_column="index", date_column="date", target_column="target", preprocess=True):
+        if preprocess:
+            X = self.preprocess(X=X, index_column=index_column, date_column=date_column, target_column=target_column)
+
+        forecast_it, ts_it = make_evaluation_predictions(
+            dataset=X,
+            predictor=self.model,
+            num_samples=100
+        )
+        return list(tqdm(forecast_it, total=len(X))), list(tqdm(ts_it, total=len(X))), len(X)
```
Innixma, line 87: Can the predictions not just be in the same form as the labels?

examples/tabular/basic_mqcnn_model.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def score(self, X, y=None, quantiles=[0.5, 0.9], eval_metric=None, metric_needs_y_pred=None, preprocess=True, index_column="index", date_column="date", target_column="target"):
```
Innixma, line 89: This seems like quite a complicated scoring method that shouldn't be hard-coded into the model. Please extract these kinds of code logic outside of the model. Why not start with RMSE to keep things simple?

autogluon/utils/tabular/ml/models/mqcnn/mqcnn_model.py
```diff
@@ -0,0 +1,44 @@
...
+from gluonts.dataset.common import ListDataset
+from gluonts.dataset.field_names import FieldName
+
+
+class MQCNNModel(AbstractModel):
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.params = get_default_parameters()
+
+    def _fit(self, X_train, y_train, X_val=None, y_val=None, time_limit=None, **kwargs):
```
Innixma, line 28: X_train has len(X_train)=1 from the example, but there are 200 rows. If you are going to re-order the rows and columns, do it inside of fit instead of outside of it.

examples/tabular/basic_mqcnn_model.py
yx1215: I am not sure if I understand what you mean by this..

examples/tabular/basic_mqcnn_model.py
```diff
@@ -0,0 +1,111 @@
...
+    ts_index = []
+    for i in range(ts_n):
+        for j in range(ts_length):
+            ts_index.append(i)
+    n = ts_n * ts_length
+    date = pd.date_range(start="2020", freq=freq, periods=ts_length)
+    return pd.DataFrame({"index": ts_index,"date": date, "target": targets, "freq": [freq for i in range(n)], "prediction_length": [prediction_length for i in range(n)]})
+
+
+class MQCNNModel(AbstractModel):
```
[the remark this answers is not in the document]
yx1215: basic_mqcnn_model.py is more detailed than models/mqcnn_model.py, and I think finally models/mqcnn_model.py should be like basic_mqcnn_model.py.

autogluon/utils/tabular/ml/models/mqcnn/mqcnn_model.py
```diff
@@ -0,0 +1,44 @@
...
+        print(X_train)
+        estimator = MQCNNEstimator.from_hyperparameters(**self.params)
+        self.model = estimator.train(X_train)
+
+    def preprocess(self, X):
+        # TODO: transform dataframe into gluon-ts ListDataset
+        target_values = X.drop("index", axis=1).values
+        processed_X = ListDataset([
+            {
+                FieldName.TARGET: target,
```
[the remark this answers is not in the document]
yx1215: I think that is what confused me. If the target here is implemented as y, then what is X?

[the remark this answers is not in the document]
Innixma: What is required is to output predictions of length N for N rows of data in standard form. Currently this does not seem to be the case, but lets focus on basic_mqcnn_model from now on.

Issue #614, opened 2020-08-12
Tabular: output_directory overwrite issue


Innixma:
If two predictors are trained within the same second, they can overwrite each-other's output. This can be fixed by adding a check to see if the directory already exists, and then adding an _N to make a new unique directory (but only if the directory was not specified by user prior). N would be a number >=1, which would be checked repeatedly until a number that results in a unique directory is chosen.

Note that race conditions might still occur in this setup if they are extremely close in start time. A lock could potentially be added on the directory creation process to ensure this does not happen.

Contributions are welcome!

songqiang:
I can take care of this. Seems a simple one.

Innixma:
Awesome, sounds good!

Innixma:
Resolved by #614, thanks Qiang for the very helpful contribution!

PR #620, opened 2020-08-17
more robust method to automatically set up output directory (#614)

songqiang:
avoid crashes when two autogluon jobs launched in the same second.

*Issue #, if available:* #614 

*Description of changes:*
As described in the issue, repeatedly add a increasing sequential number to the output directory until we find one that is available. As a guardrail, we limit the number of tries to 1000.

Tests:

Change savedir in the example script to None, and run the following command:
```
for i in $(seq 4); do python examples/tabular/example_simple_tabular.py & done
```

Before the change, two or three jobs will fail due to FileExistsError.

After the change, all four jobs succeed, with jobs launched in the same second go to different output directories
```
ag-20200817_004754
ag-20200817_004754-001
ag-20200817_004754-002
```


szha:
Job PR-620-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-620/1/index.html

Issue #376, opened 2020-03-23
TabularPredictor's predict is very slow!


elife33:
I had trained a TabularPredictor model, when I called  evaluate() at 50k+ samples, it took 3.991s. But when I called predict() at a sample for 100 times,  it took 151.334s. Why predict() is so slow?!

Innixma:
Could you provide a code sample? Predictor.evaluate() should take at least as long as Predictor.predict(), as evaluate calls predict internally.

Can you also clarify what you mean by 'called predict() at a sample for 100 times', are you saying that you are only passing in 1 row, and that the average runtime was 1.5s?

If so, you can speed-up these kinds of predictions with the following line of code:

`predictor._learner.persist_trainer(low_memory=False)`

**Be warned**: This is not memory safe, and may run OOM if the AutoGluon ensemble is large and consists of many models. Inference speed optimizations are an ongoing effort and will be improved in future.

elife33:
Thanks Innixma! Yes, I mean only passing in 1 row, and that the average runtime was 1.5s.
After I add predictor._learner.persist_trainer(low_memory=False), the average runtime reduced to 0.6s.

The code is below:
fn='/Users/elife/work/ai/ml/dl/autogluon/dtlQT/NNQT-Bn/AutogluonModels/ag-20200321_133804'
predictor_0= ag.task.tabular_prediction.predictor.TabularPredictor.load(fn)
fn='/Users/elife/work/ai/ml/dl/autogluon/dtlQT/NNQT-Bn/AutogluonModels/ag-20200321_134232'
predictor_1= ag.task.tabular_prediction.predictor.TabularPredictor.load(fn)
predictor_1._learner.persist_trainer(low_memory=False)

t = pd.read_csv('NNQT-Bn2-2010-2018.csv')
print(t.values)
print(t.describe())
N=int(t.describe().values[0][0])
print("N="+ str(N))
#print(t.values[0][1])
#print(t.values[N-1][45])
t1 = time.time()
performance = predictor_1.evaluate(task.Dataset(df = t))
t2 = time.time()
print("Time elapsed: %.3fs" % (t2 - t1))

test = pd.read_csv('test2.csv')     #only a fieldname row and a data row
print(test)

def test3():
    iModel=1
    
    t1 = time.time()
    N=10
    for i in range(0,N):
        for j in range(0,45):
            test.iloc[0][j]=t.values[i+1][j]
        test_data = task.Dataset(df = test) # convert to AutoGluon dataset
[...]

PR #640, opened 2020-08-28
Tabular: Added persist_models, unpersist_models methods to TabularPredictor

Innixma:
*Issue #, if available:*
#376 

*Description of changes:*

Added persist_models, unpersist_models methods to TabularPredictor

### DIFF
--- a/autogluon/task/tabular_prediction/predictor.py
+++ b/autogluon/task/tabular_prediction/predictor.py
@@ -622,20 +622,22 @@ class TabularPredictor(BasePredictor):
 
         return self._learner.get_feature_importance(model=model, X=dataset, features=features, feature_stage=feature_stage, subsample_size=subsample_size, silent=silent)
 
-    def persist_models(self, models=None, with_ancestors=True, max_memory=0.1) -> list:
+    def persist_models(self, models='best', with_ancestors=True, max_memory=0.1) -> list:
         """
         Persist models in memory for reduced inference latency. This is particularly important if the models are being used for online-inference where low latency is critical.
         If models are not persisted in memory, they are loaded from disk every time they are asked to make predictions.
 
         Parameters
         ----------
-        models : list of str or str, default = None
+        models : list of str or str, default = 'best'
             Model names of models to persist.
-            If None then all models are persisted.
-            If 'best' then the model with the highest validation score is persisted.
+            If 'best' then the model with the highest validation score is persisted (this is the model used for prediction by default).
+            If 'all' then all models are persisted.
             Valid models are listed in this `predictor` by calling `predictor.get_model_names()`.
         with_ancestors : bool, default = True
             If True, all ancestor models of the provided models will also be persisted.
+            If False, stacker models will not have the models they depend on persisted unless those models were specified in `models`. This will slow down inference as the ancestor models will still need to be loaded from disk for each predict call.
+            Only relevant for stacker models.
         max_memory : float, default = 0.1
             Proportion of total available memory to allow for the persisted models to use.
             If the models' summed memory usage requires a larger proportion of memory than max_memory, they are not persisted. In this case, the output will be an empty list.
@@ -647,16 +649,17 @@ class TabularPredictor(BasePredictor):
         """
         return self._learner.persist_trainer(low_memory=False, models=models, with_ancestors=with_ancestors, max_memory=max_memory)
 
-    def unpersist_models(self, models=None) -> list:
+    def unpersist_models(self, models='all') -> list:
         """
         Unpersist models in memory for reduced memory usage.
         If models are not persisted in memory, they are loaded from disk every time they are asked to make predictions.
+        Note: Another way to reset the predictor and unpersist models is to reload the predictor from disk via `predictor = TabularPredictor.load(predictor.output_directory)`.
 
         Parameters
         ----------
-        models : list of str, default = None
+        models : list of str or str, default = 'all'
             Model names of models to unpersist.
-            If None then all models are unpersisted.
+            If 'all' then all models are unpersisted.
             Valid models are listed in this `predictor` by calling `predictor.get_model_names_persisted()`.
 
         Returns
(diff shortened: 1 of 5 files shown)


szha:
Job PR-640-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-640/1/index.html

szha:
Job PR-640-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-640/2/index.html

autogluon/task/tabular_prediction/predictor.py
```diff
@@ -622,6 +622,49 @@ def feature_importance(self, dataset=None, model=None, features=None, feature_st
...
+
+        Parameters
+        ----------
+        models : list of str or str, default = None
+            Model names of models to persist.
+            If None then all models are persisted.
+            If 'best' then the model with the highest validation score is persisted.
+            Valid models are listed in this `predictor` by calling `predictor.get_model_names()`.
+        with_ancestors : bool, default = True
+            If True, all ancestor models of the provided models will also be persisted.
```
jwmueller, line 638: Clarify what downsides arise if False. Also may want to clarify setting where ancestors are even relevant (stacking).

autogluon/task/tabular_prediction/predictor.py
```diff
@@ -622,6 +622,49 @@ def feature_importance(self, dataset=None, model=None, features=None, feature_st
...
+        """
+        Persist models in memory for reduced inference latency. This is particularly important if the models are being used for online-inference where low latency is critical.
+        If models are not persisted in memory, they are loaded from disk every time they are asked to make predictions.
+
+        Parameters
+        ----------
+        models : list of str or str, default = None
+            Model names of models to persist.
+            If None then all models are persisted.
+            If 'best' then the model with the highest validation score is persisted.
```
jwmueller, line 635: Shouldn't the default be 'best'?  I imagine this `persist_models()` is primarily used before inference

autogluon/task/tabular_prediction/predictor.py
```diff
@@ -622,6 +622,49 @@ def feature_importance(self, dataset=None, model=None, features=None, feature_st
...
+            Valid models are listed in this `predictor` by calling `predictor.get_model_names()`.
+        with_ancestors : bool, default = True
+            If True, all ancestor models of the provided models will also be persisted.
+        max_memory : float, default = 0.1
+            Proportion of total available memory to allow for the persisted models to use.
+            If the models' summed memory usage requires a larger proportion of memory than max_memory, they are not persisted. In this case, the output will be an empty list.
+            If None, then models are persisted regardless of estimated memory usage. This can cause out-of-memory errors.
```
jwmueller, line 642: Maybe point user to chunked inference trick here as well?

autogluon/task/tabular_prediction/predictor.py
```diff
@@ -622,6 +622,49 @@ def feature_importance(self, dataset=None, model=None, features=None, feature_st
...
+        Returns
+        -------
+        List of persisted model names.
+        """
+        return self._learner.persist_trainer(low_memory=False, models=models, with_ancestors=with_ancestors, max_memory=max_memory)
+
+    def unpersist_models(self, models=None) -> list:
+        """
+        Unpersist models in memory for reduced memory usage.
+        If models are not persisted in memory, they are loaded from disk every time they are asked to make predictions.
```
jwmueller, line 653: Should we clarify another way to reset the predictor / remove its models from memory is simply to reload it?

autogluon/task/tabular_prediction/predictor.py
jwmueller: May also clarify: "If 'best' then the model with the highest validation score is persisted (this is the model used for prediction by default)."

szha:
Job PR-640-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-640/3/index.html

autogluon/utils/tabular/ml/trainer/abstract_trainer.py
```diff
@@ -983,27 +984,60 @@ def save_model(self, model, reduce_memory=True):
...
-    def load_models_into_memory(self, model_names=None):
+    def persist_models(self, model_names: list = None, with_ancestors=False, max_memory=None) -> list:
         if model_names is None:
             model_names = self.get_model_names_all()
+        elif model_names is 'best':
+            if self.model_best is not None:
+                model_names = [self.model_best]
+            else:
+                model_names = [self.get_model_best(can_infer=True)]
+        if with_ancestors:
```
jwmueller, line 1002: Should there be a check that `model_names` is valid argument before this moving on with this line?

autogluon/task/tabular_prediction/predictor.py
Innixma: added

autogluon/task/tabular_prediction/predictor.py
Innixma: added

autogluon/task/tabular_prediction/predictor.py
Innixma: I think this is better saved for a tutorial, they aren't very related.

autogluon/task/tabular_prediction/predictor.py
Innixma: added

autogluon/utils/tabular/ml/trainer/abstract_trainer.py
Innixma: added

Innixma:
> Looks good to me, just have minor comments about clarity. I also think 'best' should be the default for models in `predictor.persist_models()` since most users probably don't require all other models persisted as well.
> 
> Should also consider if we want to persist at the end of `task.fit()` by default or not. I imagine if the models occupy hardly any of the available memory, then persisting them by default is reasonable.

Regarding persisting by default, this is planned for a future PR to ensure stability, since this is a major change in functionality and we need to be careful about the % of memory to allow by default.

szha:
Job PR-640-4 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-640/4/index.html

PR #645, opened 2020-08-30
Tabular: Removed prediction cache functionality

Innixma:
*Issue #, if available:*

*Description of changes:*

Removed prediction cache functionality
- Prediction cache didn't work with a lot of more complex functionality and is something that ideally should be done by users instead of by AutoGluon.
- Technically this is an API breaking change if users were specifying the cache arguments in predictor.predict.
- Fixed defect in as_pandas output of predict_proba not using correct index.
- Removed old and unused variables and methods from abstract_learner.py


szha:
Job PR-645-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-645/1/index.html

Issue #581, opened 2020-07-26
Tabular: Add compression options for pickle file saving/loading


adrienatallah:
Currently pickle files saved to disk are uncompressed and can take up large amounts of disk space.  Some users may wish to have slower inference time/model loading for a large reduction in disk space usage.

**Proposal:**

Add a compression level parameter to task.fit that defaults to uncompressed (existing behavior) and is applied in [save_pkl.py](https://github.com/awslabs/autogluon/blob/master/autogluon/utils/tabular/utils/savers/save_pkl.py#L9) and [load_pkl.py](https://github.com/awslabs/autogluon/blob/master/autogluon/utils/tabular/utils/loaders/load_pkl.py#L8).

**Notes:**

-  Make sure pickles >4G are supported - this is needed for extreme classification (large number of classes + large datasets)

**Resources:**
-  https://docs.python.org/3/library/gzip.html

Innixma:
I was curious about this and gave it an initial test:

```
import os, pickle, tempfile, logging, boto3
import gzip
import bz2
import lzma

logger = logging.getLogger(__name__)


compression_fn_map = {
    None: open,
    'gzip': gzip.open,
    'bz2': bz2.open,
    'lzma': lzma.open,
}


def save(path, object, compression_fn=None, format=None, verbose=True):
    pickle_fn = lambda o, buffer: pickle.dump(o, buffer, protocol=4)
    save_with_fn(path, object, pickle_fn, compression_fn=compression_fn, format=format, verbose=verbose)


def save_with_fn(path, object, pickle_fn, compression_fn=None, format=None, verbose=True):
    if verbose:
        logger.log(15, 'Saving '+str(path))
    if s3_utils.is_s3_url(path):
        format = 's3'
    if format == 's3':
        save_s3(path, object, pickle_fn, verbose=verbose)
    else:
        os.makedirs(os.path.dirname(path), exist_ok=True)
        if compression_fn in compression_fn_map:
            with compression_fn_map[compression_fn](path, 'wb') as fout:
                pickle_fn(object, fout)
        else:
            raise ValueError(f'compression_fn={compression_fn} is not a valid compression_fn. Valid values: {compression_fn_map.keys()}')


def load(path, format=None, compression_fn=None, verbose=True):
    if path.endswith('.pointer'):
        format = 'pointer'
    elif s3_utils.is_s3_url(path):
        format = 's3'
    if format == 'pointer':
        content_path = load_pointer.get_pointer_content(path)
[...]

PR #654, opened 2020-09-07
Tabular pickle compression

adrienatallah:
*[Issue # 581](https://github.com/awslabs/autogluon/issues/581)*

*Description of changes:*

This PR implements two extra keyword args `compression_fn` and `compression_fn_kwargs` in `task.fit()` that are used when saving/loading pkl files to disk.  Valid compression functions are `gzip`, `bz2` and `lzma`; valid `compresion_fn_kwargs` are accepted arguments for each compression function's `open` function.  See each description here: 

[https://docs.python.org/3/library/gzip.html#gzip.open](gzip.open)
[https://docs.python.org/3/library/lzma.html#lzma.open](lzma.open)
[https://docs.python.org/3/library/bz2.html#bz2.open](bz2.open) 

A sample use:

```
predictor = task.fit(train_data=train_data, label='class',
                     compression_fn='gzip',
                     compression_fn_kwargs={'compresslevel': 3}
                     )
```

**TODO: benchmarking results**
---


Innixma (commented):
Looks good, thanks for the contribution! Could you rebase with mainline so that the CI is triggered to ensure that tests pass?

autogluon/utils/tabular/utils/compression_utils.py
```diff
@@ -0,0 +1,37 @@
...
+        'extension': 'bz2',
+    },
+    'lzma': {
+        'open': lzma.open,
+        'extension': 'lzma',
+    },
+}
+
+
+def get_validated_path(filename, compression_fn=None):
```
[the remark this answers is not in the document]
adrienatallah: Yes, definitely, good call.

autogluon/utils/tabular/ml/learner/abstract_learner.py
```diff
@@ -563,8 +566,9 @@ def get_info(self, include_model_info=False):
     def infer_problem_type(y: Series):
         return infer_problem_type(y=y)
 
-    def save(self):
```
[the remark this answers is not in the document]
adrienatallah: I'll update to use self.compression_fn and self.compression_fn_kwargs.

szha:
Job PR-654-4 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-654/4/index.html

adrienatallah:
Rebased and added a unit test for `compression_utils.py`

daveharmon (approved):

tests/unittests/utils/tabular/test_compression_utils.py
```diff
@@ -0,0 +1,70 @@
...
+
+    single_extension_filepath = 'dummy_file.pkl'
+    assert(compression_utils.get_validated_path(single_extension_filepath, compression_fns[0])
+           == expected_gzip_filepath)
+    assert(compression_utils.get_validated_path(single_extension_filepath, compression_fns[1])
+           == expected_lzma_filepath)
+    assert(compression_utils.get_validated_path(single_extension_filepath, compression_fns[2])
+           == expected_bz2_filepath)
+
+    multiple_extension_filepath = 'dummy_file.fake-foo.zip.pkl'
```
Innixma, line 42: happy to discuss, but is this what we want? If someone specifies they want to save a file named 'abc.def.pkl', this will currently save it as 'abc.pkl' with no compression. This seems a bit counter-intuitive. Is the idea to avoid having multiple extensions? If so, why? Would it make more sense to throw an exception if sent multiple extensions/any extensions as input?

autogluon/task/tabular_prediction/tabular_prediction.py
```diff
@@ -429,6 +429,14 @@ def fit(train_data,
                     The seed used for stack level L is equal to `seed+L`.
                     This means `random_seed=1` will have the same split indices at L=0 as `random_seed=0` will have at L=1.
                 If `random_seed=None`, a random integer is used.
+            compression_fn : str, default = None
+                Compression algorithm used to compress pickle files saved/loaded on disk.
```
Innixma, line 433: Also mention that None is an option and will do no compression

autogluon/utils/tabular/ml/models/abstract/abstract_model.py
```diff
@@ -297,7 +297,7 @@ def score_with_y_pred_proba(self, y, y_pred_proba, eval_metric=None, metric_need
         else:
             return eval_metric(y, y_pred_proba)
 
-    def save(self, path: str = None, verbose=True) -> str:
+    def save(self, path: str = None, verbose=True, compression_fn=None, compression_fn_kwargs=None) -> str:
```
Innixma, line 300: use **kwargs instead and pass **kwargs into save_pkl.save()?

autogluon/utils/tabular/ml/models/abstract/abstract_model.py
```diff
@@ -320,11 +320,12 @@ def save(self, path: str = None, verbose=True) -> str:
...
             path = self.path
         file_path = path + self.model_file_name
-        save_pkl.save(path=file_path, object=self, verbose=verbose)
+        save_pkl.save(path=file_path, object=self, verbose=verbose, compression_fn=compression_fn,
+                      compression_fn_kwargs=compression_fn_kwargs)
         return path
 
     @classmethod
-    def load(cls, path: str, reset_paths=True, verbose=True):
+    def load(cls, path: str, reset_paths=True, verbose=True, compression_fn=None, compression_fn_kwargs=None):
```
Innixma, line 328: ditto regarding kwargs

autogluon/utils/tabular/ml/models/ensemble/bagged_ensemble_model.py
```diff
@@ -373,8 +373,8 @@ def _add_child_times_to_bag(self, model):
             self.predict_time += model.predict_time
 
     @classmethod
-    def load(cls, path: str, reset_paths=True, low_memory=True, load_oof=False, verbose=True):
-        model = super().load(path=path, reset_paths=reset_paths, verbose=verbose)
+    def load(cls, path: str, reset_paths=True, low_memory=True, load_oof=False, verbose=True, compression_fn=None, compression_fn_kwargs=None):
```
Innixma, line 376: ditto regarding kwargs

autogluon/utils/tabular/ml/models/ensemble/bagged_ensemble_model.py
```diff
@@ -415,7 +415,7 @@ def load_model_base(self):
     def save_model_base(self, model_base):
         save_pkl.save(path=self.path + 'utils' + os.path.sep + 'model_template.pkl', object=model_base)
 
-    def save(self, path=None, verbose=True, save_oof=True, save_children=False) -> str:
+    def save(self, path=None, verbose=True, save_oof=True, save_children=False, compression_fn=None, compression_fn_kwargs=None) -> str:
```
Innixma, line 418: ditto regarding kwargs

autogluon/utils/tabular/ml/models/ensemble/stacker_ensemble_model.py
```diff
@@ -218,13 +219,13 @@ def hyperparameter_tune(self, X, y, k_fold, scheduler_options=None, compute_base
         # TODO: hpo_results likely not correct because no renames
         return stackers, stackers_performance, hpo_results
 
-    def load_base_model(self, model_name):
+    def load_base_model(self, model_name, compression_fn=None, compression_fn_kwargs=None):
```
Innixma, line 222: consider replacing with **kwargs to send to the model_type.load(model_path, **kwargs)

Innixma (changes requested):
Looks pretty good, thanks for the contribution! Mostly minor comments, once those are resolved I'll give it a benchmark run to ensure things are working properly.

autogluon/utils/tabular/ml/models/ensemble/weighted_ensemble_model.py
```diff
@@ -12,7 +12,9 @@
 # TODO: Optimize predict speed when fit on kfold, can simply sum weights
 class WeightedEnsembleModel(StackerEnsembleModel):
     def __init__(self, base_model_names, base_model_paths_dict, base_model_types_dict, **kwargs):
-        model_0 = base_model_types_dict[base_model_names[0]].load(path=base_model_paths_dict[base_model_names[0]], verbose=False)
+        compression_fn = kwargs.get('compression_fn', '')
```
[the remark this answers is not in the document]
adrienatallah: Yep, thanks.

autogluon/utils/tabular/ml/models/ensemble/weighted_ensemble_model.py
```diff
@@ -12,7 +12,9 @@
...
+        compression_fn_kwargs = kwargs.get('compression_fn_kwargs', {})
```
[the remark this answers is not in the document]
adrienatallah: Yep

autogluon/utils/tabular/ml/models/tabular_nn/tabular_nn_model.py
```diff
@@ -697,14 +697,14 @@ def _create_preprocessor(self, impute_strategy, max_category_levels):
             raise NotImplementedError("language_features cannot be used at the moment")
         return ColumnTransformer(transformers=transformers)  # numeric features are processed in the same order as in numeric_features vector, so feature-names remain the same.
 
-    def save(self, path: str = None, verbose=True) -> str:
+    def save(self, path: str = None, verbose=True, compression_fn=None, compression_fn_kwargs=None) -> str:
```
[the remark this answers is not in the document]
adrienatallah: Yes, good point.

autogluon/utils/tabular/ml/trainer/abstract_trainer.py
```diff
@@ -160,19 +163,23 @@ def load_y_val(self):
 
     def save_X_train(self, X, verbose=True):
```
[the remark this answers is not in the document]
adrienatallah: Thought about doing something like that, will do.

autogluon/utils/tabular/utils/compression_utils.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+compression_fn_map = {
+    None: {
+        'open': open,
+        'extension': 'pkl',
+    },
+    'gzip': {
+        'open': gzip.open,
+        'extension': 'gz',
```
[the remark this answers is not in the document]
adrienatallah: Considered that also, it does make sense, will update.

tests/unittests/utils/tabular/test_compression_utils.py
adrienatallah: No, I misunderstood your previous comment about testing the `abc.def.pkl` filename, I think it does make sense to just append the compression function extensions and keep multiple extensions if they are used. So, `abc.def.pkl.gz` for a gzip compressed file named `abc.def.pkl`.  The extension shouldn't be changed if there is no compression though, the original filename should be saved.

adrienatallah:
Thanks for the feedback, happy to be contributing!

tests/unittests/utils/tabular/test_compression_utils.py
Innixma: Ok makes sense! Regarding your last statement, I agree that original name should be saved, so instead of changing the dict for gz/bzip/etc., just change None to be '' instead of '.pkl'. I believe that is what you are saying?

Innixma:
Once you address comments, please also rebase with mainline to get latest changes so I can compare benchmark runs easily and so CI can execute.

tests/unittests/utils/tabular/test_compression_utils.py
adrienatallah: Yes, the way I had it now was in `get_validated_path` if compression_fn is None, it just returns the path unchanged: https://github.com/adrienatallah/autogluon/blob/3447591a2c6f8517cb3b4493181f034efb4abde9/autogluon/utils/tabular/utils/compression_utils.py#L30, so really the '.pkl' for None in the dict wasn't being used, which isn't the best.  I'll change the extension for None to be ' ' instead of '.pkl' and just always append the extensions.

Innixma:
@adrienatallah Just a heads up that we will be having a code-freeze starting Friday for ~1 week in order to merge in a major modularization refactor. This refactor moves the location of all files in the code-base, which will mean the PR will likely have to be rebased manually to the new file locations. To avoid this, please submit revisions by Wednesday to leave time for review if you want this PR to get in for 0.0.14 release.

adrienatallah:
Okay, thanks for the heads up.  It's almost there, I'll try to get it finished up and rebased tonight.

szha:
Job PR-654-5 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-654/5/index.html

autogluon/utils/tabular/utils/loaders/load_pkl.py
```diff
@@ -22,8 +27,19 @@ def load(path, format=None, verbose=True):
...
+    validated_path = compression_utils.get_validated_path(path, compression_fn=compression_fn)
+
+    if compression_fn_kwargs is None:
+        compression_fn_kwargs = {}
+
+    if compression_fn in compression_fn_map:
+        with compression_fn_map[compression_fn]['open'](validated_path, 'rb', **compression_fn_kwargs) as fin:
+            object = pickle.load(fin)
+    else:
+        raise ValueError(f'compression_fn={compression_fn} or compression_fn_kwargs={compression_fn_kwargs} are not valid. Valid function values: {compression_fn_map.keys()}')
```
adrienatallah, line 41: We are still not validating `compression_fn_kwargs` here, do you think it is worth doing?  I decided not to since a change to the various compression function open API's could break things.

szha:
Job PR-654-7 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-654/7/index.html

autogluon/utils/tabular/utils/loaders/load_pkl.py
Innixma: It's fine to not validate kwargs, as that is hard to do.

Innixma (commented):
Looks good to me! I will still need to benchmark and validate that it is stable, and I think that since 0.0.14 is our last 0.0.x release before 0.1, I will probably wait for 0.0.14 to release on Thursday before merging this just to ensure this PR doesn't introduce some form of instability and it has time to be battle tested in mainline as we build towards 0.1.

Thanks for the contribution and the revisions, much appreciated!

Innixma (changes requested):
I sent out a PR with some fixes and cleanup. After PR is merged into this, the following TODO's remain:

TODO: HPO does not work yet.
TODO: models should have parameter indicating compression method used in model.save(). Trainer should maintain what compression to use for each model separately in DAG. Models will have compression method set upon init. This will automatically solve HPO without increasing complexity.
TODO: Unit test end-to-end task.fit with compression (Consider adding to advanced_functionality unit test).
TODO: Benchmark

adrienatallah:
Sounds good, happy to make the contribution!

Innixma:
@adrienatallah Given the time gap in this contribution, I think it makes sense to break it into smaller chunks so we can get the simplest parts in first. If you plan to work on this, could you open a new PR with just the changes to save_pkl and load_pkl, along with a unit test that demonstrates that the new options work correctly? Then after that is merged, we will have a stable foundation to gradually add in the compression options.

My first thought is we can prioritize files which are not loaded during inference by the user, such as the cached train/val data and various util files that are only needed during training. This would essentially be free space reduction, since we don't care to load these files often.

adrienatallah:
@Innixma Yes, definitely, that very much makes sense to me. I did start on the rebase, but there is quite a bit more to go.  I will open a separate PR with just `save_pkl` and `load_pkl` changes with unit tests.

adrienatallah:
Opened a part 1 PR to help break up the rebase: https://github.com/awslabs/autogluon/pull/940

Innixma:
Closing this PR as #940 was contributed in its place.

PR #661, opened 2020-09-13
Tabular: Feature Generator Refactor (Part 3)

Innixma:
*Issue #, if available:*

*Description of changes:*

Major changes:
- Added chained feature usage detection to highlight to users which features were not utilized in the feature generator and to automatically optimize generators to avoid generating features which are later dropped in the pipeline.
- General logging cleanup/improvements. Now generators which have no valid input features do not log that they are being fit to avoid confusion.


szha:
Job PR-661-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-661/1/index.html

autogluon/utils/tabular/features/generators/abstract.py
```diff
@@ -461,11 +461,21 @@ def _remove_features_in(self, features: list):
...
             self.feature_metadata_in = self.feature_metadata_in.remove_features(features=features)
             self.features_in = self.feature_metadata_in.get_features()
+            if self._pre_astype_generator:
+                self._pre_astype_generator.remove_features_out(features)
 
-    # TODO: Use this to prune features post-fit
-    def _remove_features_out(self, features: list):
+    def remove_features_out(self, features: list):
         """
         Removes features from all relevant objects which represent the content of the output data.
```
jwmueller, line 480: maybe clarify: "This is used for XYZ".

autogluon/utils/tabular/features/generators/abstract.py
```diff
@@ -560,6 +567,97 @@ def _log(self, level, msg, log_prefix=None, verb_min=None):
     def is_fit(self):
         return self._is_fit
 
+    def is_valid_metadata_in(self, feature_metadata_in: FeatureMetadata):
+        """True if input data with feature metadata of feature_metadata_in could result in non-empty output."""
```
jwmueller, line 571: Could clarify further why empty output might arise.

autogluon/utils/tabular/features/generators/abstract.py
```diff
@@ -571,6 +669,7 @@ def print_generator_info(self, log_level: int = 20):
         """
         if self.fit_time:
             self._log(log_level, f'\t{round(self.fit_time, 3)}s\t= Fit runtime')
+            self._log(log_level, f'\t({len(self.features_in)}, {len(self.features_out)}) = (input feature count, output feature count)')
```
jwmueller, line 672: I worry some novices may confuse "fit" with model-fitting and would hope the output remains clear to somebody who has no idea what a "FeatureGenerator" is.  I think this would be clearer if printed as for example:
```
Fitting AutoMLPipelineFeatureGenerator to preprocess the data...
```
and later:
```
30s = Fitting data preprocessors runtime
13 features in original data, 15 features in preprocessed data
```

autogluon/utils/tabular/features/generators/category.py
```diff
@@ -124,3 +124,13 @@ def _generate_category_map(self, X: DataFrame) -> (DataFrame, dict):
...
+
+    def _remove_features_in(self, features: list):
+        super()._remove_features_in(features)
+        if self.category_map:
+            for feature in features:
+                if feature in self.category_map:
+                    self.category_map.pop(feature)
+
+    def _more_tags(self):
+        return {'feature_interactions': False}
```
jwmueller, line 136: many newlines missing at bottom of files

autogluon/utils/tabular/features/generators/pipeline.py
```diff
@@ -108,9 +108,9 @@ def print_feature_metadata_info(self, log_level=20):
             self._log(log_level, f'\t\tThese features do not need to be present at inference time for this FeatureGenerator.')
         if self._feature_metadata_in_unused.get_features():
             self._log(log_level, f'\tUnused Original Features (Count: {len(self._feature_metadata_in_unused.get_features())}): {self._feature_metadata_in_unused.get_features()}')
-            self._log(log_level, f'\t\tThese features were not valid input to any of the feature generators. Add a feature generator compatible with these features to utilize them.')
+            self._log(log_level, f'\t\tThese features were not used to generate any of the output features. Add a feature generator compatible with these features to utilize them.')
```
jwmueller, line 111: IMO the print statements should be more from the perspective of the overall ML pipeline, rather than from the perspective of a FeatureGenerator (because many users who just skim docs may have no idea what this is or what "output features" are).  So this could say for example:
```
These features will be ignored by all models. Add a feature generator compatible with these features to utilize them.
```

autogluon/utils/tabular/features/generators/pipeline.py
```diff
@@ -108,9 +108,9 @@ def print_feature_metadata_info(self, log_level=20):
...
             self._log(log_level, f'\t\tThese features do not need to be present at inference time for this FeatureGenerator.')
```
jwmueller, line 112: To reduce confusion, would replace: "These features do not need to be present at inference time for this FeatureGenerator"
with:
"These features do not need to be present at inference time."

autogluon/utils/tabular/features/generators/rename.py
```diff
@@ -0,0 +1,72 @@
...
+from pandas import DataFrame
+
+from .abstract import AbstractFeatureGenerator
+
+logger = logging.getLogger(__name__)
+
+
+class RenameFeatureGenerator(AbstractFeatureGenerator):
+    """
+    RenameFeatureGenerator renames the columns without altering the features.
```
jwmueller, line 13: "altering the features" -> "altering their values"?

jwmueller: maybe clarify to users why column names might matter?

autogluon/utils/tabular/features/generators/pipeline.py
```diff
@@ -108,9 +108,9 @@ def print_feature_metadata_info(self, log_level=20):
...
-            self._feature_metadata_in_unused.print_feature_metadata_full(self.log_prefix + '\t', log_level=log_level)
+            self._feature_metadata_in_unused.print_feature_metadata_full(self.log_prefix + '\t\t', log_level=log_level)
         self._log(20, '\tOriginal Features (exact raw dtype, raw dtype):')
```
jwmueller, line 114: Upon playing with the new feature generators more, I think users will find these outputs confusing:

```
Original Features (exact raw dtype, raw dtype):
            ('int64', 'int')     : 6 | ['age', 'fnlwgt', 'education-num', ...]
            ('object', 'object') : 8 | ['workclass', 'education', 'marital-status', ...]
    Original Features (raw dtype, special dtypes):
            ('int', [])    : 6 | ['age', 'fnlwgt', 'education-num', ...]
            ('object', []) : 8 | ['workclass', 'education', 'marital-status', ...]
    Processed Features (exact raw dtype, raw dtype):
            ('category', 'category') : 8 | ['workclass', 'education', 'marital-status', ...]
            ('int64', 'int')         : 6 | ['age', 'fnlwgt', 'education-num', ...]
    Processed Features (raw dtype, special dtypes):
            ('category', []) : 8 | ['workclass', 'education', 'marital-status', ...]
            ('int', [])      : 6 | ['age', 'fnlwgt', 'education-num', ...]
```

and think this could be clarified via a print-structure like this:
```
Types of features in original data:
...

Types of features in preprocessed data:
...
```

Also the reasoning behind currently presenting the types in terms of pairs `(a, b)` seems unclear from user's perspective (could instead say eg: "Originally of type a, will be treated as type b"). 

It also may be confusing that  'special dtypes' can corresponds to empty list, isn't it better just to not print anything if there's no special dtype?

Also only listing up to 3 features of each type seems quite few (and almost not worth the print statement at all for such little information).

jwmueller: I also think this statement may confuse users: 
```
AutoMLPipelineFeatureGenerator processed 499 data points with 14 features
```
when the numbers don't match the earlier output `Train Data Rows:    500`

I'd vote to suppress this statement in general. If we want to print something here, I'd only do it whenever these values don't match the original # of train data rows/columns, and would also provide possible explanations (eg. datapoints with rare class-label may be deleted).

autogluon/utils/tabular/features/generators/pipeline.py
jwmueller: Also if we cannot easily track the reason why each feature was ignored, perhaps should clarify here why features may be ignored in general (eg. due to taking constant value across all rows).

jwmueller (changes requested):
Code looks good, but I'm still worried the numerous output statements will confuse novice users, who I don't think we should expect to even know what a "feature generator" is.

autogluon/utils/tabular/features/generators/category.py
Innixma: good catch, added newlines

autogluon/utils/tabular/features/generators/rename.py
Innixma: clarified

autogluon/utils/tabular/features/generators/pipeline.py
Innixma: added

autogluon/utils/tabular/features/generators/abstract.py
Innixma: added clarification and made private again since this is advanced functionality that shouldn't be used by many users.

autogluon/utils/tabular/features/generators/abstract.py
Innixma: clarified with example

autogluon/utils/tabular/features/generators/abstract.py
Innixma: Added 2nd half of comment, tried to add 'preprocess the data', but it looked messy.

autogluon/utils/tabular/features/generators/pipeline.py
Innixma: Reworked this logging somewhat, and removed the statement you mentioned as confusing since it is redundant.

autogluon/utils/tabular/features/generators/pipeline.py
Innixma: I'd prefer to keep it in context of generator to avoid assumptions about what the data is used for. Happy to tweak this later though. Added some extra logs on common reasons for being unused and TODO to try to add logic to determine why a feature was unused.

szha:
Job PR-661-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-661/2/index.html

jwmueller (approved):
LGTM, I still have the following remaining logging suggestions you might add before merging:

1) There is weird indentation in output here:
```
Stage 3 Generators:
            Fitting IdentityFeatureGenerator...
            Fitting CategoryFeatureGenerator...
                    Fitting CategoryMemoryMinimizeFeatureGenerator... 
```

2) Before "Fitting AutoMLPipelineFeatureGenerator...", add statement like:
```
Using Feature Generators to preprocess the data ...
```
which'll make it much clearer what is going on (again I don't think users will understand what a feature generator is otherwise).

3) Add some similar qualifying statement to precede `feature_metadata_in` in the logs as I also don't think users will know what this is. Eg something like: 
```
Inferring data type of each feature based on column values. Set feature_metadata_in to manually specify special dtypes of the features.
```

4) Consider omitting this print statement for binary problems (redundant with other prints):
```
Train Data Class Count: 2
```

Innixma:
> LGTM, I still have the following remaining logging suggestions you might add before merging:
> 
> 1. There is weird indentation in output here:
> 
> ```
> Stage 3 Generators:
>             Fitting IdentityFeatureGenerator...
>             Fitting CategoryFeatureGenerator...
>                     Fitting CategoryMemoryMinimizeFeatureGenerator... 
> ```
> 

I'm not able to reproduce this. I get the following:

```
Fitting AutoMLPipelineFeatureGenerator...
	Available Memory:                    6013.36 MB
	Train Data (Original)  Memory Usage: 1.19 MB (0.0% of available memory)
	Inferring feature_metadata_in based on data. Specify feature_metadata_in to control the special dtypes of the input data.
	Stage 1 Generators:
		Fitting AsTypeFeatureGenerator...
	Stage 2 Generators:
		Fitting FillNaFeatureGenerator...
	Stage 3 Generators:
		Fitting IdentityFeatureGenerator...
		Fitting CategoryFeatureGenerator...
			Fitting CategoryMemoryMinimizeFeatureGenerator...
	Stage 4 Generators:
		Fitting DropUniqueFeatureGenerator...
```

> 1. Before "Fitting AutoMLPipelineFeatureGenerator...", add statement like:
> 
> ```
> Using Feature Generators to preprocess the data ...
> ```
> 
> which'll make it much clearer what is going on (again I don't think users will understand what a feature generator is otherwise).
> 

Added

> 1. Add some similar qualifying statement to precede `feature_metadata_in` in the logs as I also don't think users will know what this is. Eg something like:
> 
> ```
> Inferring data type of each feature based on column values. Set feature_metadata_in to manually specify special dtypes of the features.
> ```
> 

Added

> 1. Consider omitting this print statement for binary problems (redundant with other prints):
> 
> ```
> Train Data Class Count: 2
> ```

Omitted when binary

szha:
Job PR-661-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-661/3/index.html

PR #667, opened 2020-09-15
[WIP, DO NOT MERGE] Autogluon Modularization - separating core, cv, text into namespaces

gradientsky:
Please add your comments/suggestions for proposed layout. If you want to try this version, here's how to build:
```
python3 -m venv $VENV_LOCATION
source $VENV_LOCATION/bin/activate
pip install -U pip
pip install mxnet-cu102
pip install gluonnlp==0.8.3
pip install numpy

cd $PROJECT_ROOT/autogluon_core/
pip install -e .
cd $PROJECT_ROOT/autogluon_tabular/
pip install -e .
cd $PROJECT_ROOT/autogluon_mxnet/
pip install -e .
cd $PROJECT_ROOT/autogluon_extra/
pip install -e .
cd $PROJECT_ROOT/autogluon_text/
pip install -e .
cd $PROJECT_ROOT/autogluon_vision/
pip install -e .
```
Test can be started by running `pytest` from the `autogluon_core`/`autogluon_tabular`/`autogluon_extra`.
CV and Text modules will be separated this week


Innixma:
Awesome work!

A few comments:

For this first PR, lets try to keep code structure as similar as possible, so for example tabular's structure is utils/tabular/..., which we want to change after modularization, but we should do that in a follow-up PR so that these changes are more notice-able and reviewable.

Since this change is so massive, anything that isn't a file move change should be highlighted in some capacity so that we are aware of what has been changed in order to minimize bugs.

try_import can stay in core for now, and we can see if we want to split it later. The imports only occur within the functions.

Once CV and NLP are moved to separate modules, we will have to see what remains in extra and discuss what to do with it.

gradientsky:
> For this first PR, lets try to keep code structure as similar as possible, so for example tabular's structure is utils/tabular/..., which we want to change after modularization, but we should do that in a follow-up PR so that these changes are more notice-able and reviewable.

Note taken. Few comments here:
1) some utils classes needs to be split to break cross-modules dependencies. One way we can address it: do the split as a precursor PR (this would be less invasive than the big refactoring).
2) we will get a strange package names until next PRs: similar to `autogluon.tabular.utils.tabular.features...`

autogluon_extra/src/autogluon/extra/task/__init__.py
```diff
@@ -4,8 +4,9 @@
 from autogluon.core.task.base import BaseTask
 from .image_classification import ImageClassification
 from .object_detection import ObjectDetection, Detector
-from .text_classification import TextClassification
-from .text_prediction import TextPrediction
+from autogluon.text.task.text_classification import TextClassification
```
Innixma, line 7: These can still be (and should be) relatively imported, such as `from ...text.task.text_classification import TextClassification`

autogluon_text/README.md
```diff
@@ -0,0 +1,91 @@
+
+
+<div align="left">
+  <img src="https://user-images.githubusercontent.com/16392542/77208906-224aa500-6aba-11ea-96bd-e81806074030.png" width="350">
+</div>
+
+## AutoML Toolkit for Deep Learning
```
Innixma, line 7: Why have readme in autogluon_text? Why not in root?

gradientsky: README.md is used as a long description in setup.py:
```
long_description = open('README.md').read()
```
Open for any suggestions what should we do here

autogluon_extra/src/autogluon/extra/task/__init__.py
gradientsky: not if they are in different module

gradientsky:
Updated the code: separated text/vision/mxnet (utils shared by vision and text)

autogluon_extra/src/autogluon/extra/task/__init__.py
Innixma: Yes I believe they can be.

Innixma:
I think autogluon.extra.model_zoo can go into autogluon.mxnet, and I think enas can also go in autogluon.mxnet. I think this should make autogluon.extra completely empty.

mseeger (commented):
Maybe I am missing something. But almost everything in autogluon/searcher/bayesopt currently still depends on MXNet (it will later only depend on autograd, but this has not been brought in yet). You are moving this into the core. I think it should first go into the MXNet-dependent part, right?

PR #673, opened 2020-09-18
add example to run ABOHB on pytorch model

aaronkl:
adds an example python script to show how to optimize the hyperparameters of a pytorch models with ABOHB

### DIFF
--- a/examples/hnas/abohb_example.py
+++ b/examples/hnas/abohb_example.py
@@ -21,7 +21,7 @@ from torchvision import transforms
     learning_rate=ag.space.Real(lower=1e-6, upper=1, log=True),
     batch_size=ag.space.Int(lower=8, upper=128),
     wd=ag.space.Real(lower=1e-8, upper=1, log=True),
-    epochs=27,
+    epochs=27
 )
 def objective_function(args, reporter, **kwargs):
     ts_start = time.time()
@@ -111,10 +111,10 @@ def objective_function(args, reporter, **kwargs):
         ts_now = time.time()
         eval_time = ts_start - ts_now
         reporter(
-            epoch=epoch + 1,
+            epochs=epoch + 1,
             performance=float(curr_best),
             eval_time=eval_time,
-            time_step=ts_now)
+            time_step=ts_now, **config)
 
 
 def callback(training_history, start_timestamp):
@@ -211,9 +211,8 @@ if __name__ == "__main__":
                                                 reward_attr='performance',
                                                 # The metric along we make scheduling decision. Needs to be also
                                                 # reported back to AutoGluon in the objective function.
-                                                time_attr='epoch',
+                                                time_attr='epochs',
                                                 brackets=brackets,
-                                                max_t=27,  # maximum number of epochs to train the network
                                                 checkpoint=None,
                                                 searcher="bayesopt",  # Defines searcher for new configurations
                                                 dist_ip_addrs=dist_ip_addrs,
--- a/examples/hnas/abohb_example.py
+++ b/examples/hnas/abohb_example.py
@@ -111,7 +111,7 @@ def objective_function(args, reporter, **kwargs):
         ts_now = time.time()
         eval_time = ts_start - ts_now
         reporter(
-            epochs=epoch + 1,
+            epoch=epoch + 1,
             performance=float(curr_best),
             eval_time=eval_time,
             time_step=ts_now, **config)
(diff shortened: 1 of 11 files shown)


examples/hnas/abohb_example.py
```diff
@@ -0,0 +1,229 @@
...
+        config["dropout_2"] = dropout_2
+        config["batch_size"] = batch_size
+        config["learning_rate"] = learning_rate
+        config["wd"] = wd
+
+        ts_now = time.time()
+        eval_time = ts_start - ts_now
+        reporter(
+            epoch=epoch + 1,
+            performance=float(curr_best),
```
mseeger, line 115: Strictly speaking, this is not the accuracy after epoch epochs, but the best accuracy over these epochs. Do we want that?

examples/hnas/abohb_example.py
```diff
@@ -0,0 +1,229 @@
...
+        # We also report the hyperparameter configuration back to AutoGluon,
+        # so we can process it later in the callback function, for example for visualization.
+        config = dict()
+        config["n_units_1"] = n_units_1
+        config["n_units_2"] = n_units_2
+        config["dropout_1"] = dropout_1
+        config["dropout_2"] = dropout_2
+        config["batch_size"] = batch_size
+        config["learning_rate"] = learning_rate
+        config["wd"] = wd
```
mseeger, line 109: This is not passed back to reporter, right?

examples/hnas/abohb_example.py
```diff
@@ -0,0 +1,229 @@
+import autogluon as ag
```
mseeger, line 1: Very cool. Let us also replace the other examples/tutorials/etc (MXNet code) by PyTorch.

examples/hnas/abohb_example.py
```diff
@@ -0,0 +1,229 @@
...
+                                                num_trials=args.num_trials,
+                                                time_out=args.timeout,
+                                                # This argument defines the metric that will be maximized.
+                                                # Make sure that you report this back in the objective function.
+                                                reward_attr='performance',
+                                                # The metric along we make scheduling decision. Needs to be also
+                                                # reported back to AutoGluon in the objective function.
+                                                time_attr='epoch',
+                                                brackets=brackets,
+                                                max_t=27,  # maximum number of epochs to train the network
```
mseeger, line 216: You do not need to set this, it will get it from the train_fn (from args.epochs).

examples/hnas/abohb_example.py
```diff
@@ -0,0 +1,229 @@
...
+                                                checkpoint=None,
```
mseeger, line 217: The default is None anyway

Innixma:
To fix the CI, please rebase to mainline. awscli update broke CI recently.

szha:
Job PR-673-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-673/2/index.html

examples/hnas/abohb_example.py
aaronkl: It's a matter of perspective. The intuition is, that, in practice, you wouldn’t trust that the performance at the final epochs is the best. As practitioner you would checkpoint your model after each epoch if the validation performance has improved, and afterwards take the last checkpointed model rather than the final weights.

examples/hnas/abohb_example.py
aaronkl: you are right, I will fix it

szha:
Job PR-673-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-673/3/index.html

Innixma:
@aaronkl Are you looking for reviews from anyone in particular at this time, or are you working on this with Matthias still?

aaronkl:
I fixed all of Matthias comments, so if he is fine with it we can merge it

Innixma:
@aaronkl Why does this PR show that it is adding so many lines? Most of these changes appear to be changes made by Jonas that were already merged into mainline. Do you know why this is the case?

aaronkl:
@Innixma  I have no idea to be honest, I guess something went wrong with the rebase. Do you want me to create a new PR to fix this?

Innixma:
@aaronkl Yeah, I think it would be best to make a new PR just to clean up the changes. It is definitely weird, not sure why it looks like that.

aaronkl:
ok new PR is now #687

PR #677, opened 2020-09-22
Tabular: Simplified handling of time_limit in Trainer

Innixma:
*Issue #, if available:*

*Description of changes:*

Simplified handling of time_limit in Trainer. Now it is managed by two private variables instead of 4 public variables. This will help better support future fit calls to work properly.

### DIFF
--- a/autogluon/utils/tabular/ml/trainer/abstract_trainer.py
+++ b/autogluon/utils/tabular/ml/trainer/abstract_trainer.py
@@ -41,7 +41,7 @@ class AbstractTrainer:
 
     def __init__(self, path: str, problem_type: str, scheduler_options=None, eval_metric=None, stopping_metric=None,
                  num_classes=None, low_memory=False, feature_metadata=None, kfolds=0, n_repeats=1,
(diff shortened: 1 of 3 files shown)


autogluon/utils/tabular/ml/trainer/abstract_trainer.py
```diff
@@ -1782,14 +1785,13 @@ def distill(self, X_train=None, y_train=None, X_val=None, y_val=None,
             if self.problem_type != REGRESSION:
                 self.regress_preds_asprobas = True
 
-        self.time_train_start = time.time()
```
Innixma, line 1785: safe to remove because the private variables only matter for multi-layer model training, and are not needed for single-layer model training (which distilled models are)

szha:
Job PR-677-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-677/1/index.html

daveharmon (approved):

gradientsky (approved):

Issue #675, opened 2020-09-21
fit_summary Error KeyError: 'trial_info'


amybachir:
I'm trying autogluon for the first time with a dataset of 75,000 data points and labels.
autogluon version 0.0.13
OS: 
Distributor ID: Ubuntu
Description:    Ubuntu 18.04.4 LTS
Release:        18.04
Codename:       bionic

Here is my predictor, it is a TabularPrediction with hyperparameter_tune turned on:
`predictor = task.fit(train_data=train_data, label=label_column, output_directory=savedir, problem_type='binary', eval_metric='accuracy', hyperparameter_tune=True )`

I tried with hyperparameter_tune  off and the task finished with no problem. Once I turned on hyperparameter_tune, it errored with the following error:
```
*** Details of Hyperparameter optimization ***
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-2-1b709fc952c0> in <module>
     11 predictor = task.fit(train_data=train_data, label=label_column, output_directory=savedir, problem_type='binary', eval_metric='accuracy', hyperparameter_tune=True )
---> 13 results = predictor.fit_summary()
     14 
     15 # Inference time:

/usr/local/lib/python3.6/dist-packages/autogluon/task/tabular_prediction/predictor.py in fit_summary(self, verbosity)
    414                     hpo_model = hpo_results[model_type]
    415                     print("HPO for %s model:  Num. configurations tried = %s, Time spent = %s, Search strategy = %s"
--> 416                           % (model_type, len(hpo_model['trial_info']), hpo_model['total_time'], hpo_model['search_strategy']))
    417                     print("Best hyperparameter-configuration (validation-performance: %s = %s):"
    418                           % (self.eval_metric, hpo_model['validation_performance']))

KeyError: 'trial_info'

```

jwmueller:
Thanks for pointing out this bug. I believe it should be fixed in this PR:  https://github.com/awslabs/autogluon/pull/678 
if you'd like to install that version of autogluon locally and try it out. 

Also just FYI:  `hyperparameter_tune` is currently not typically the best way to maximize accuracy in TabularPrediction. Instead consider specifying: `task.fit(..., presets='best_quality')` which utilizes stack-ensembling instead of HPO.

https://autogluon.mxnet.io/tutorials/tabular_prediction/tabular-quickstart.html#maximizing-predictive-performance

amybachir:
[...]

PR #678, opened 2020-09-22
bugfix fit_summary with HPO

jwmueller:
Should fix Issue: https://github.com/awslabs/autogluon/issues/675

### DIFF
--- a/autogluon/task/tabular_prediction/predictor.py
+++ b/autogluon/task/tabular_prediction/predictor.py
@@ -421,11 +421,10 @@ class TabularPredictor(BasePredictor):
                 print("*** Details of Hyperparameter optimization ***")
                 for model_type in hpo_results:
                     hpo_model = hpo_results[model_type]
-                    print("HPO for %s model:  Num. configurations tried = %s, Time spent = %s, Search strategy = %s"
-                          % (model_type, len(hpo_model['trial_info']), hpo_model['total_time'], hpo_model['search_strategy']))
-                    print("Best hyperparameter-configuration (validation-performance: %s = %s):"
-                          % (self.eval_metric, hpo_model['validation_performance']))
-                    print(hpo_model['best_config'])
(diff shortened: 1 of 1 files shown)


szha:
Job PR-678-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-678/1/index.html

autogluon/task/tabular_prediction/predictor.py
```diff
@@ -421,11 +421,10 @@ def fit_summary(self, verbosity=3):
                 print("*** Details of Hyperparameter optimization ***")
                 for model_type in hpo_results:
                     hpo_model = hpo_results[model_type]
-                    print("HPO for %s model:  Num. configurations tried = %s, Time spent = %s, Search strategy = %s"
-                          % (model_type, len(hpo_model['trial_info']), hpo_model['total_time'], hpo_model['search_strategy']))
-                    print("Best hyperparameter-configuration (validation-performance: %s = %s):"
-                          % (self.eval_metric, hpo_model['validation_performance']))
-                    print(hpo_model['best_config'])
+                    if 'trial_info' in hpo_model:
```
daveharmon, line 424: We should probably just wrap the first print statement in the `if` statement

jwmueller: sorry can you clarify what you mean?

Innixma (approved):

autogluon/task/tabular_prediction/predictor.py
daveharmon: The `trial_info` key is only used in the first `print` statement, so you don't need to protect the other two print statements with this check

jwmueller: Oh actually those other print statements implicitly still depend on the first one to make sense, so that's why they're in there as well

PR #682, opened 2020-09-24
Tabular: Added LightGBM DART model retrain support, added GPU support for CatBoost

Innixma:
*Issue #, if available:*

*Description of changes:*

Added LightGBM DART model retrain support, added GPU support for CatBoost.

TODO: Benchmark

### DIFF
--- a/autogluon/utils/tabular/ml/models/catboost/catboost_model.py
+++ b/autogluon/utils/tabular/ml/models/catboost/catboost_model.py
@@ -50,7 +50,7 @@ class CatboostModel(AbstractModel):
 
     # TODO: Use Pool in preprocess, optimize bagging to do Pool.split() to avoid re-computing pool for each fold! Requires stateful + y
     #  Pool is much more memory efficient, avoids copying data twice in memory
-    def _fit(self, X_train, y_train, X_val=None, y_val=None, time_limit=None, **kwargs):
+    def _fit(self, X_train, y_train, X_val=None, y_val=None, time_limit=None, num_gpus=0, **kwargs):
         try_import_catboost()
         from catboost import CatBoostClassifier, CatBoostRegressor, Pool
         if self.problem_type == SOFTCLASS:
@@ -146,6 +146,11 @@ class CatboostModel(AbstractModel):
 
         params = self.params.copy()
         num_features = len(self.features)
+        if num_gpus != 0:
+            if 'task_type' not in params:
+                params['task_type'] = 'GPU'
+                # TODO: Confirm if GPU is used in HPO (Probably not)
+                # TODO: Adjust max_bins to 254?
 
         if params.get('task_type', None) == 'GPU':
             if 'colsample_bylevel' in params:
--- a/autogluon/utils/tabular/ml/models/lgb/lgb_model.py
+++ b/autogluon/utils/tabular/ml/models/lgb/lgb_model.py
@@ -52,7 +52,7 @@ class LGBModel(AbstractModel):
             eval_metric_name = eval_metric
         return eval_metric, eval_metric_name
 
-    def _fit(self, X_train=None, y_train=None, X_val=None, y_val=None, dataset_train=None, dataset_val=None, time_limit=None, **kwargs):
+    def _fit(self, X_train=None, y_train=None, X_val=None, y_val=None, dataset_train=None, dataset_val=None, time_limit=None, num_gpus=0, **kwargs):
         start_time = time.time()
         params = self.params.copy()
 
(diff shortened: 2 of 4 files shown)


szha:
Job PR-682-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-682/1/index.html

szha:
Job PR-682-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-682/2/index.html

jwmueller:
Later we should consider keeping `num_gpus = None` as default in `task.fit()`, and then allowing each model to autodetect GPU and utilize it when it sees `num_gpus=None`

Innixma:
> Later we should consider keeping `num_gpus = None` as default in `task.fit()`, and then allowing each model to autodetect GPU and utilize it when it sees `num_gpus=None`

Yes, I had a similar idea. Either `num_gpus=None` or `num_gpus='auto'`. We need to very solid in terms of fallback options if the GPU doesn't work before we go this route though, so likely try/except checks for each model in case GPU training fails. This gets more tricky with models like LightGBM that need special installations to use GPU, but hopefully try/except will still work.

Innixma:
What makes this even trickier is that GPU isn't simply a training speedup, it fundamentally changes how the models are trained and their resulting accuracy, often for the worse.

Innixma:
Benchmark results show that while DART does help slightly when stack ensembling is disabled, it adds very little if anything to the stack ensemble results (weakest average results of the four LGBM variants). Since it is an expensive model to train, I have removed it from the default model set.

szha:
Job PR-682-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-682/3/index.html

jwmueller:
> What makes this even trickier is that GPU isn't simply a training speedup, it fundamentally changes how the models are trained and their resulting accuracy, often for the worse.

Right but that's why each model could choose whether or not to utilize the GPU based on auto-detection of GPU availability + possibly some dataset characteristics.  That flexibility seems a much better default.

autogluon/utils/tabular/ml/models/catboost/catboost_model.py
```diff
@@ -146,6 +146,11 @@ def _fit(self, X_train, y_train, X_val=None, y_val=None, time_limit=None, **kwar
 
         params = self.params.copy()
         num_features = len(self.features)
+        if num_gpus != 0:
```
jwmueller, line 149: add comment that only 1 GPU can be utilized?

autogluon/utils/tabular/ml/models/catboost/catboost_model.py
```diff
@@ -146,6 +146,11 @@ def _fit(self, X_train, y_train, X_val=None, y_val=None, time_limit=None, **kwar
...
+            if 'task_type' not in params:
+                params['task_type'] = 'GPU'
+                # TODO: Confirm if GPU is used in HPO (Probably not)
```
jwmueller, line 152: any idea why GPU isn't utilized during HPO?

autogluon/utils/tabular/ml/models/lgb/lgb_model.py
```diff
@@ -74,6 +74,15 @@ def _fit(self, X_train=None, y_train=None, X_val=None, y_val=None, dataset_train
         gc.collect()
 
         num_boost_round = params.pop('num_boost_round', 1000)
+        dart_retrain = params.pop('dart_retrain', False)  # Whether to retrain the model to get optimal iteration if model is trained in 'dart' mode.
```
jwmueller, line 77: For both of these hyperparams, I don't think we should have specify the default value here?  Shouldn't we just ensure all default values come from `hyperparameters/parameters.py`, so that they're very clearly documented?

jwmueller (approved):
Looks good, just have minor comments

autogluon/utils/tabular/ml/models/catboost/catboost_model.py
Innixma: I believe CatBoost can use multi-gpu.

autogluon/utils/tabular/ml/models/catboost/catboost_model.py
Innixma: Disjoint code calls, I don't think the num_gpus param is passed with HPO. In general, I'm in favor of moving num_gpus into params instead of fit, that gives us a lot more flexibility in terms of per-model GPU usage, and keeps model .fit args clean. It could be part of AG_args_fit.

autogluon/utils/tabular/ml/models/lgb/lgb_model.py
Innixma: dart_retrain is very niche, and probably will only be for internal usage. I agree in general however, and I think we can maybe clean that up. Do you suggest that if num_boost_round doesn't exist at this point in the code to throw an exception?

jwmueller: Yep could just do `params['num_boost_round']` instead of `.pop()`

PR #721, opened 2020-10-21
Remove all MXNet dependencies from core

mseeger:
*Issue #, if available:*

*Description of changes:*
I've also pulled all try_import_XYZ into core, which is where they may be needed.


mseeger:
What is missing here are changes to Jenkins etc in order to not require MXNet for core. I lack the skills to do that

mseeger:
All tests in core/tests/unittests pass in a venv without mxnet being installed.

Innixma:
Thanks for the PR! Could you rebase with mainline? Files were moved slightly in Tabular recently, but they should stay stable for awhile now.

The change that occurred is that files in `tabular/utils/ml/` and `tabular/utils/` were moved directly to `tabular/`

PR #737, opened 2020-10-23
[CI refactoring] Updated docs publication process

gradientsky:
*Description of changes:*
* master branch will no longer publish docs into the root https://autogluon.mxnet.io/ 
* master branch will publish all docs to https://autogluon.mxnet.io/dev/
* master branch will publish root_index.html (new file) into root index.html. The only thing this file is going to do is to immediately redirect into https://autogluon.mxnet.io/stable/
* docs updated
* restricted internal cross-module versions to a specific version
* doc links in headers are now replaced dynamically to a correct branch during the build time

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -96,8 +96,9 @@ stage("Build Docs") {
             cacheControl = ''
         } else {
             isMaster = env.BRANCH_NAME == 'master'
+            isDev = env.BRANCH_NAME == 'dev'
             bucket = 'autogluon.mxnet.io'
-            path = isMaster ? 'dev' : "${env.BRANCH_NAME}"
+            path = isMaster ? 'dev' : isDev ? 'dev-branch' : "${env.BRANCH_NAME}"
             site = "${bucket}/${path}"
             flags = isMaster ? '' : '--delete'
             cacheControl = '--cache-control max-age=7200'
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -167,7 +167,7 @@ stage("Build Docs") {
         python3 -m pip install --upgrade -e .
         cd ..
 
-        sed -i -e 's/###_PLACEHOLDER_WEB_CONTENT_ROOT_###/https:\\/\\/${escaped_context_root}/g' docs/config.ini
+        sed -i -e 's/###_PLACEHOLDER_WEB_CONTENT_ROOT_###/http:\\/\\/${escaped_context_root}/g' docs/config.ini
 
         cd docs && bash build_doc.sh
         aws s3 sync ${flags} _build/html/ s3://${bucket}/${path} --acl public-read ${cacheControl}
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -87,20 +87,28 @@ stage("Build Docs") {
         checkout scm
         VISIBLE_GPU=env.EXECUTOR_NUMBER.toInteger() % 8
 
+        index_update_str = ''
         if (env.BRANCH_NAME.startsWith("PR-")) {
             bucket = 'autogluon-staging'
-            path = "${env.BRANCH_NAME}/${env.BUILD_NUMBER}/"
-            site = "${bucket}.s3-website-us-west-2.amazonaws.com/${path}index.html"
+            path = "${env.BRANCH_NAME}/${env.BUILD_NUMBER}"
+            site = "${bucket}.s3-website-us-west-2.amazonaws.com/${path}"
             flags = '--delete'
             cacheControl = ''
         } else {
             isMaster = env.BRANCH_NAME == 'master'
             bucket = 'autogluon.mxnet.io'
-            path = isMaster ? '' : "${env.BRANCH_NAME}/"
+            path = isMaster ? 'dev' : "${env.BRANCH_NAME}"
             site = "${bucket}/${path}"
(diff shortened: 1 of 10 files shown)


szha:
Job PR-737-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-737/1/index.html

Jenkinsfile
```diff
@@ -96,10 +97,16 @@ stage("Build Docs") {
         } else {
             isMaster = env.BRANCH_NAME == 'master'
             bucket = 'autogluon.mxnet.io'
-            path = isMaster ? '' : "${env.BRANCH_NAME}/"
+            path = isMaster ? 'dev/' : "${env.BRANCH_NAME}/"
```
Innixma, line 100: what happens if the branch is 'dev'? Do we just avoid having a branch with that name?

docs/ReleaseInstructions.md
```diff
@@ -0,0 +1,27 @@
+### Release process
+
+* Update version `title` in `root_index.html`.
```
Innixma, line 3: We will need to somehow lock all released submodules to only allow installations of the other submodules in the same version.

For example,
in v0.1 release, autogluon.tabular==0.1 must have 'autogluon.core==0.1' in setup.py, otherwise release 0.1 could break when we release 0.2 since autogluon.tabular==0.1 would then install autogluon.core==0.2

Innixma: This might need to be done at the time of PyPi release, that way it isn't manual. Somehow we would need to edit the setup.py files specifically in the PyPi release, otherwise pre-releases might also suffer from the same issue (0.2 tabular pre-release installing 0.1 core)

Jenkinsfile
gradientsky: If we have dev branch, it will overwrite this space. So the recommendation is to avoid this branch name.

docs/config.ini
```diff
@@ -20,9 +20,9 @@ release = 0.0.15
 # items: name, URL, and a fontawesome icon
 # (https://fontawesome.com/icons?d=gallery). Items are separated by commas.
 header_links = Stable Version Documentation, https://autogluon.mxnet.io/stable/index.html, fas fa-book,
```
Innixma, line 22: Ideally we will also want a More Versions button that has a hardcoded link to 
```http://autogluon.mxnet.io/dev/versions.html```
and the page contains hardcoded links to the other known documentation versions such as 0.0.14, dev, stable, 0.x.y, etc.

docs/ReleaseInstructions.md
gradientsky: Alternatively we can lock version of the dependencies to the same version as the current library

szha:
Job PR-737-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-737/3/index.html

docs/ReleaseInstructions.md
Innixma: that is genius actually, I didn't think about the fact that we would know the version of the package during setup.py

Innixma: about 100x easier than what I was thinking ;)

gradientsky: done

Innixma: You'll need to rebase with mainline, I updated the setup.py files recently

Jenkinsfile
Innixma: for safety, could we have isDev and then set path to 'dev-branch/' just to avoid worst case scenario? It will be hard for all team members to always remember this pitfall.

docs/config.ini
Innixma: We will have to backport this (and other changes) to 0.0.14/stable so they can link properly, then things should work properly going forward.

szha:
Job PR-737-4 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-737/4/index.html

Jenkinsfile
gradientsky: good point, added `dev-branch` convention

docs/ReleaseInstructions.md
gradientsky: Done

szha:
Job PR-737-6 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-737/6/index.html

szha:
Job PR-737-7 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-737/7/index.html

szha:
Job PR-737-8 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-737/8/index.html

szha:
Job PR-737-9 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-737/9/index.html

docs/config.ini
gradientsky: Fixed the links, now they are branch-specific

Innixma (approved):

Innixma:
Website still doesn't redirect to stable or dev, it is by default the old doc build.

Issue #488, opened 2020-05-27
AttributeError: 'NoneType' object has no attribute 'update'


HerbertArthur:
I am trying to run this code like below:

```python
#-*- coding:utf-8 -*-

import autogluon as ag
from autogluon import ImageClassification as task

#filename=ag.download("https://autogluon.s3.amazonaws.com/datasets/shopee-iet.zip")
#ag.unzip(filename)
dataset = task.Dataset('data/train')
test_dataset = task.Dataset('data/test', train=False)
classifier = task.fit(dataset, epochs=3, ngpus_per_trial=1, verbose=True)
print(">>>>")
HBox(children=(FloatProgress(value=0.0, max=2.0), HTML(value="")))
print("Top 1 val accuracy: %.3f" % classifier.results["best_reward"])
```

Then I got error, the output is following:

>(mxnet) [root@46acffeb04ca autogluon]# python classfication_opt.py 
>scheduler: FIFOScheduler(
>DistributedResourceManager{
>(Remote: Remote REMOTE_ID: 0, 
><Remote: 'inproc://172.17.0.15/14336/1' processes=1 threads=28, memory=202.71 GB>, Resource: NodeResourceManager(28 CPUs, 4 GPUs))
>})
>
>Starting Experiments
>Num of Finished Tasks is 0
>Num of Pending Tasks is 2
>[Epoch 3] Validation: 0.388: 100%|████████████████████████████████████████████████████████████████████████████████████| 3/3 [00:27<00:00,  9.06s/it]
>100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 2/2 [00:31<00:00, 15.95s/it]
>[Epoch 3] Validation: 0.425: 100%|████████████████████████████████████████████████████████████████████████████████████| 3/3 [00:27<00:00,  9.03s/it]
>[Epoch 3] training: accuracy=0.441: 100%|█████████████████████████████████████████████████████████████████████████████| 3/3 [00:27<00:00,  9.30s/it]
>Traceback (most recent call last):
>  File "classfication_opt.py", line 10, in <module>
>    classifier = task.fit(dataset, epochs=3, ngpus_per_trial=1, verbose=False)
>  File "/root/anaconda3/envs/mxnet/lib/python3.7/site-packages/autogluon/task/image_classification/image_classification.py", line 294, in fit
>    results = BaseTask.run_fit(train_image_classification, search_strategy, scheduler_options)
[...]

PR #744, opened 2020-10-27
[Bug Fix] BaseTask: fix silent exception, handle None in best config result

zhreshold:
*Issue #, if available:*
https://github.com/awslabs/autogluon/issues/488

*Description of changes:*
Found the root error during another experiment. There are two issues with existing BaseTask
- If logging level not properly set, the stacktrace of failed task won't be displayed
- If training config with best config failed, `None` object is returned, causing #488 

The solution is to ensure the stacktrace will be printed, and catch the `None.update` issue

### DIFF
--- a/core/src/autogluon/core/scheduler/fifo.py
+++ b/core/src/autogluon/core/scheduler/fifo.py
@@ -225,7 +225,7 @@ class FIFOScheduler(TaskScheduler):
                 self.load_state_dict(load(checkpoint))
             else:
                 msg = f'checkpoint path {checkpoint} is not available for resume.'
-                logger.exception(msg)
+                logger.critical(msg)
                 raise FileExistsError(msg)
 
     def run(self, **kwargs):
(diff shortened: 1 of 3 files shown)


szha:
Job PR-744-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-744/1/index.html

core/src/autogluon/core/scheduler/fifo.py
```diff
@@ -399,7 +399,7 @@ def _run_reporter(self, task, task_job, reporter):
             reported_result = reporter.fetch()
             if 'traceback' in reported_result:
                 # Evaluation has failed
-                logger.exception(reported_result['traceback'])
+                logger.critical(reported_result['traceback'])
```
mseeger, line 402: Do we should use critical instead of exception? Did you check whether this has to be replaced elsewhere?

mseeger (approved):

core/src/autogluon/core/scheduler/fifo.py
zhreshold: setting to critical is to make sure the logger message won't be discard due to logger.level == `NOTSET` by default, of course we can change the default logger level, but it's just safer to ensure that the exception message has the highest logging level.

jwmueller (approved):

Issue #580, opened 2020-07-25
Add fastText classification model to tabularprediction


songqiang:
Tracking discussions on adding fasttext model to TabularPrediction

Innixma:
FastText has been added, thanks again @songqiang for the contribution!

PR #756, opened 2020-10-31
Tabular: Added dropping of object type features in tabular models

Innixma:
*Issue #, if available:*
#580 

*Description of changes:*

- Added object dtype to the ignored type list of all tabular models, as they should be converted to category first.
- This enables text features in object type form to not crash tabular models, they will be dropped instead.
- Added get_features_kwargs and get_features_kwargs_extra aux model params to enable full flexibility of which feature types a model accepts.

### DIFF
--- a/tabular/src/autogluon/tabular/features/feature_metadata.py
+++ b/tabular/src/autogluon/tabular/features/feature_metadata.py
@@ -94,7 +94,7 @@ class FeatureMetadata:
         """
         Returns a list of features held within the feature metadata object after being pruned through the available parameters.
 
-         Parameters
+        Parameters
         ----------
         valid_raw_types : list, default None
             If a feature's raw type is not in this list, it is pruned.
@@ -121,7 +121,7 @@ class FeatureMetadata:
             If special_types is None, then any feature will satisfy the special type requirement (including those with no special types).
         required_exact : bool, default False
             If True, then if a feature does not have the exact same special types (with no extra special types) as required_special_types, it is pruned.
-            This also applied to required_raw_special_pairs if specified.
+            This is also applied to required_raw_special_pairs if specified.
             Has no effect if required_special_types and required_raw_special_pairs are None.
         required_at_least_one_special : bool, default False
             If True, then if a feature has zero special types, it is pruned.
--- a/tabular/src/autogluon/tabular/models/abstract/abstract_model.py
+++ b/tabular/src/autogluon/tabular/models/abstract/abstract_model.py
@@ -132,10 +132,14 @@ class AbstractModel:
         for key, value in default_auxiliary_params.items():
             self._set_default_param_value(key, value, params=self.params_aux)
 
+    # TODO: v0.1 consider adding documentation to each model highlighting which feature dtypes are valid
     def _get_default_auxiliary_params(self) -> dict:
-        """Dictionary of auxiliary parameters that dictate various model-agnostic logic."""
+        """
+        Dictionary of auxiliary parameters that dictate various model-agnostic logic, such as:
+            Which column dtypes are filtered out of the input data, or how much memory the model is allowed to use.
+        """
         default_auxiliary_params = dict(
-            max_memory_usage_ratio=1.0,  # Ratio of memory usage allowed by the model. Values > 1.0 have an increased risk of causing OOM errors.
+            max_memory_usage_ratio=1.0,  # Ratio of memory usage allowed by the model. Values > 1.0 have an increased risk of causing OOM errors. Used in memory checks during model training to avoid OOM errors.
             # TODO: Add more params
             # max_memory_usage=None,
             # max_disk_usage=None,
--- a/tabular/src/autogluon/tabular/models/abstract/abstract_model.py
+++ b/tabular/src/autogluon/tabular/models/abstract/abstract_model.py
@@ -64,7 +64,7 @@ class AbstractModel:
 
         if self.eval_metric.name in OBJECTIVES_TO_NORMALIZE:
             self.normalize_pred_probas = True
-            logger.debug(self.name +" predicted probabilities will be transformed to never =0 since eval_metric=" + self.eval_metric.name)
+            logger.debug(f"{self.name} predicted probabilities will be transformed to never =0 since eval_metric='{self.eval_metric.name}'")
         else:
             self.normalize_pred_probas = False
 
@@ -118,11 +118,22 @@ class AbstractModel:
     def is_fit(self) -> bool:
         return self.model is not None
 
+    # TODO: v0.1 update to be aligned with _set_default_auxiliary_params(), add _get_default_params()
     def _set_default_params(self):
         pass
 
     def _set_default_auxiliary_params(self):
+        """
+        Sets the default aux parameters of the model.
+        This method should not be extended by inheriting models, instead extend _get_default_auxiliary_params.
+        """
         # TODO: Consider adding to get_info() output
+        default_auxiliary_params = self._get_default_auxiliary_params()
+        for key, value in default_auxiliary_params.items():
+            self._set_default_param_value(key, value, params=self.params_aux)
+
+    def _get_default_auxiliary_params(self) -> dict:
+        """Dictionary of auxiliary parameters that dictate various model-agnostic logic."""
         default_auxiliary_params = dict(
             max_memory_usage_ratio=1.0,  # Ratio of memory usage allowed by the model. Values > 1.0 have an increased risk of causing OOM errors.
             # TODO: Add more params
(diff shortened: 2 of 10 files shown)


szha:
Job PR-756-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-756/1/index.html

tabular/src/autogluon/tabular/features/feature_metadata.py
```diff
@@ -121,7 +121,7 @@ def get_features(self, valid_raw_types: list = None, valid_special_types: list =
             If special_types is None, then any feature will satisfy the special type requirement (including those with no special types).
         required_exact : bool, default False
             If True, then if a feature does not have the exact same special types (with no extra special types) as required_special_types, it is pruned.
-            This also applied to required_raw_special_pairs if specified.
+            This is also applied to required_raw_special_pairs if specified.
             Has no effect if required_special_types and required_raw_special_pairs are None.
```
jwmueller, line 125: I'd also clarify here the fact that: columns with object dtype will be ignored by all models.

tabular/src/autogluon/tabular/models/abstract/abstract_model.py
```diff
@@ -118,11 +118,22 @@ def can_infer(self) -> bool:
...
+        Sets the default aux parameters of the model.
+        This method should not be extended by inheriting models, instead extend _get_default_auxiliary_params.
+        """
         # TODO: Consider adding to get_info() output
+        default_auxiliary_params = self._get_default_auxiliary_params()
+        for key, value in default_auxiliary_params.items():
+            self._set_default_param_value(key, value, params=self.params_aux)
+
+    def _get_default_auxiliary_params(self) -> dict:
+        """Dictionary of auxiliary parameters that dictate various model-agnostic logic."""
```
jwmueller, line 136: I'd provide some examples in docstring.

"logic, such as: which dtypes are columns that are ignored by the model, or how much memory the model is allowed to use."

tabular/src/autogluon/tabular/models/catboost/catboost_model.py
```diff
@@ -12,15 +12,15 @@
...
 logger = logging.getLogger(__name__)
 
 
-# TODO: Catboost crashes on multiclass problems where only two classes have significant member count.
-#  Question: Do we turn these into binary classification and then convert to multiclass output in Learner? This would make the most sense.
-# TODO: Consider having Catboost variant that converts all categoricals to numerical as done in RFModel, was showing improved results in some problems.
+# TODO: Consider having CatBoost variant that converts all categoricals to numerical as done in RFModel, was showing improved results in some problems.
+# TODO: v0.1 rename to CatBoostModel and rename model name default to CatBoost (instead of Catboost)
 class CatboostModel(AbstractModel):
     def __init__(self, **kwargs):
```
jwmueller, line 25: I think it would be good if each model has some minimal documentation that at least states which dtypes will lead to a column being ignored by this model.  I don't think most users/contributors know about this key fact otherwise as it's buried deep in the code.

tabular/src/autogluon/tabular/models/knn/knn_model.py
```diff
@@ -39,14 +39,14 @@ def _set_default_params(self):
         for param, val in default_params.items():
             self._set_default_param_value(param, val)
 
-    def _set_default_auxiliary_params(self):
-        default_auxiliary_params = dict(
+    def _get_default_auxiliary_params(self) -> dict:
+        default_auxiliary_params = super()._get_default_auxiliary_params()
+        extra_auxiliary_params = dict(
             ignored_type_group_raw=[R_CATEGORY, R_OBJECT],  # TODO: Eventually use category features
             ignored_type_group_special=[S_TEXT_NGRAM, S_TEXT_SPECIAL, S_DATETIME_AS_INT],
```
jwmueller, line 46: why does KNN need to ignore S_DATETIME_AS_INT? Because of lack of feature-rescaling?

jwmueller (approved):
Looks good, but I think it needs to be made more clear to users/contributors what is going on regarding these ignored columns.

In particular, I think the NaiveBayes custom model example should include some clarification of this point as well, at least a brief comment explaining to rely on `categorical` instead of `object` dtypes for discrete features.

tabular/src/autogluon/tabular/models/abstract/abstract_model.py
gradientsky: 1) if this method is expected to be extended in child classes, should it be marked as internal?
2) since this is effectively is implemented in each model, it's probably a good idea to add example in docstring.

tabular/src/autogluon/tabular/models/abstract/abstract_model.py
```diff
@@ -118,11 +118,22 @@ def can_infer(self) -> bool:
...
         default_auxiliary_params = dict(
             max_memory_usage_ratio=1.0,  # Ratio of memory usage allowed by the model. Values > 1.0 have an increased risk of causing OOM errors.
```
gradientsky, line 138: More documentation on how his parameter should be used needed.

tabular/src/autogluon/tabular/models/catboost/catboost_model.py
```diff
@@ -280,3 +280,11 @@ def get_model_feature_importance(self):
         importance_series = importance_df.set_index('Feature Id')['Importances']
         importance_dict = importance_series.to_dict()
         return importance_dict
+
+    def _get_default_auxiliary_params(self) -> dict:
+        default_auxiliary_params = super()._get_default_auxiliary_params()
+        extra_auxiliary_params = dict(
+            ignored_type_group_raw=[R_OBJECT],
+        )
+        default_auxiliary_params.update(extra_auxiliary_params)
```
gradientsky, line 289: If the merge approach is the expected pattern of usage to add additional parameters, this would be error-prone: please consider generalizing the merge logic in the abstract class and allow extension to return just `extra_auxiliary_params`.

gradientsky: [Discussion point] Also, does it makes sense to force sub-classes to implement this method? If so, then abstract class can throw a `NotImplementedError` by default. This would guide model author to override all the necessary methods.

gradientsky (changes requested):

tabular/src/autogluon/tabular/models/abstract/abstract_model.py
Innixma: 1. _ just means that it shouldn't be called outside of the object itself. I don't think it has any relation to user extensions. For example, we encourage users to extend _fit, _predict_proba, and _preprocess

2. Perhaps example would live best in the example custom model code? Not sure it belongs in the docstring.

tabular/src/autogluon/tabular/models/catboost/catboost_model.py
Innixma: How would this work if the model extended a class which then extended AbstractModel? We would need to handle chained calls to extra_auxiliary_params.

tabular/src/autogluon/tabular/models/knn/knn_model.py
Innixma: Yes, currently the datetime as int is a massive integer that would overpower all the other features.

tabular/src/autogluon/tabular/models/abstract/abstract_model.py
Innixma: Added extra examples to docstring

tabular/src/autogluon/tabular/features/feature_metadata.py
Innixma: I don't think FeatureMetadata should have comments talking about model implementations, as it is used in various contexts.

tabular/src/autogluon/tabular/models/catboost/catboost_model.py
Innixma: added TODO

tabular/src/autogluon/tabular/models/abstract/abstract_model.py
Innixma: Added, although this is generally not meant for most users to use.

tabular/src/autogluon/tabular/models/catboost/catboost_model.py
gradientsky: Yes, the class hierarchy would make it more complicated. I think this alternative would work better: perform overwrites in `__init__`, instead - this would preserve hierarchy chain overwrites:
```python
class AbstractModel(object):

    def __init__(self):
        self.auxiliary_params = dict(
            max_memory_usage_ratio=1.0,  # Ratio of memory usage allowed by the model. Values > 1.0 have an increased risk of causing OOM errors. Used in memory checks during model training to avoid OOM errors.
            max_time_limit_ratio=1.0,  # ratio of given time_limit to use during fit(). If time_limit == 10 and max_time_limit_ratio=0.3, time_limit would be changed to 3.
        )

    def register_auxiliary_params(self, extra_auxiliary_params: dict):
        self.auxiliary_params.update(extra_auxiliary_params)


class SubModel(AbstractModel):

    def __init__(self):
        super().__init__()
        self.register_auxiliary_params(dict(
            max_time_limit_ratio=2.0,
            new_param='a',
        ))


class SubSubModel(SubModel):

    def __init__(self):
        super().__init__()
        self.register_auxiliary_params(dict(
            max_memory_usage_ratio=3.0,
            new_sub_param='c',
        ))

if __name__ == '__main__':
    print(f'AbstractModel: {AbstractModel().auxiliary_params}')
    print(f'SubModel     : {SubModel().auxiliary_params}')
    print(f'SubSubModel  : {SubSubModel().auxiliary_params}')
```
Output:
```
AbstractModel: {'max_memory_usage_ratio': 1.0, 'max_time_limit_ratio': 1.0}
SubModel     : {'max_memory_usage_ratio': 1.0, 'max_time_limit_ratio': 2.0, 'new_param': 'a'}
SubSubModel  : {'max_memory_usage_ratio': 3.0, 'max_time_limit_ratio': 2.0, 'new_param': 'a', 'new_sub_param': 'c'}
```

Innixma: The problem comes that the user can override the defaults by the params argument. This can't be done properly if done in init since this logic happens in abstract_model's init, and thus this order will overwrite user specified overrides.

szha:
Job PR-756-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-756/2/index.html

gradientsky (approved):

PR #762, opened 2020-11-08
Tabular: Ensure unique model names

Innixma:
*Issue #, if available:*

*Description of changes:*

- Addressed duplicate model name issue, now when models share the same name, one will have a unique suffix appended to it. This is a pre-req for repeated fit functionality to work smoothly. Prior, the model would overwrite the earlier model it shared a name with, leading to major bugs with stackers having their base models overwritten. This bug did not happen in current usage, but could start to happen if users were allowed to call fit repeatedly.
- Removed duplicate stack trace print on model exception, before the stack trace was being printed twice.
- Simplified hyperparameters argument in `Trainer.train_multi_levels` to handle the same input format as users provide in task.fit(hyperparameters), in preparation for repeated fit functionality.

### DIFF
--- a/tabular/src/autogluon/tabular/trainer/abstract_trainer.py
+++ b/tabular/src/autogluon/tabular/trainer/abstract_trainer.py
@@ -16,7 +16,6 @@ from ..utils import get_pred_from_proba, generate_train_test_split, infer_eval_m
 from ..models.abstract.abstract_model import AbstractModel
 from ..metrics import log_loss, scorer_expects_y_pred
 from ..models.ensemble.bagged_ensemble_model import BaggedEnsembleModel
-from ..trainer.model_presets.presets import get_preset_models
 from ..trainer.model_presets.presets_custom import get_preset_custom
 from ..trainer.model_presets.presets_distill import get_preset_models_distillation
 from ..models.ensemble.stacker_ensemble_model import StackerEnsembleModel
@@ -193,7 +192,7 @@ class AbstractTrainer:
         return max_level
 
     # TODO: Rename method for v0.1
-    def get_models(self, hyperparameters: dict, hyperparameter_tune=False, **kwargs) -> List[AbstractModel]:
+    def get_models(self, hyperparameters: dict, **kwargs) -> List[AbstractModel]:
         """Constructs a list of unfit models based on the hyperparameters dict."""
         raise NotImplementedError
 
@@ -248,6 +247,8 @@ class AbstractTrainer:
             base_model_names = []
         if hyperparameters is None:
             hyperparameters = self.hyperparameters
+        else:
+            hyperparameters = self._process_hyperparameters(hyperparameters=hyperparameters)
         hyperparameters = copy.deepcopy(hyperparameters)
 
         # TODO: Hack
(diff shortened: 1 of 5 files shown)


szha:
Job PR-762-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-762/2/index.html

tabular/src/autogluon/tabular/trainer/abstract_trainer.py
```diff
@@ -336,13 +337,24 @@ def stack_new_level_core(self, X, y, models: List[AbstractModel], X_val=None, y_
...
+            for model in models:
+                # Ensure name is unique
+                model_stack_name = f'{model.name}_STACKER_l{level}'
+                num_increment = 2
+                while model_stack_name in invalid_model_names:  # Ensure name is unique
+                    model_stack_name = f'{model.name}_{num_increment}_STACKER_l{level}'
+                    num_increment += 1
+                invalid_model_names.add(model_stack_name)
+
+                model_stack = stacker_type(
```
jwmueller, line 351: should we print some output (with unimportant logger level) to alert user of the new model name and that duplicate model exists?

tabular/src/autogluon/tabular/trainer/abstract_trainer.py
```diff
@@ -336,13 +337,24 @@ def stack_new_level_core(self, X, y, models: List[AbstractModel], X_val=None, y_
                     return []
             else:
                 raise AssertionError(f'Stack level cannot be negative! level = {level}')
-            models = [
-                stacker_type(
-                    path=self.path, name=model.name + '_STACKER_l' + str(level), model_base=model, base_model_names=base_model_names,
+            invalid_model_names = set(self.get_model_names_all())
```
jwmueller, line 340: What about use-case where somebody does want to overwrite an existing model?

tabular/src/autogluon/tabular/trainer/model_presets/presets.py
```diff
@@ -203,7 +204,10 @@ def get_preset_models_softclass(path, hyperparameters, num_classes=None, hyperpa
         # TODO: add support for per-stack level hyperparameters
     models = get_preset_models(path=path, problem_type=SOFTCLASS, eval_metric=soft_log_loss, stopping_metric=soft_log_loss,
                                hyperparameters=hyperparameters_standard, num_classes=num_classes, hyperparameter_tune=hyperparameter_tune,
-                                name_suffix=name_suffix, default_priorities=DEFAULT_SOFTCLASS_PRIORITY)
+                                name_suffix=name_suffix, default_priorities=DEFAULT_SOFTCLASS_PRIORITY, invalid_model_names=invalid_model_names)
+    if invalid_model_names is None:
+        invalid_model_names = []
+    invalid_model_names = invalid_model_names + [model.name for model in models]
```
jwmueller, line 210: could ensure name uniqueness by also adding to `invalid_names` here the new model names after the rename that occurs later on:

`model.rename(model.name.replace('Regressor', 'Classifier'))`

jwmueller (approved):

tabular/src/autogluon/tabular/trainer/model_presets/presets.py
Innixma: revised

tabular/src/autogluon/tabular/trainer/abstract_trainer.py
Innixma: added TODO for v0.1, don't want to touch args for these methods until TabTransformer PR is merged to avoid conflicts

tabular/src/autogluon/tabular/trainer/abstract_trainer.py
Innixma: We can come back to this question at a later time if we find users being confused about names, but I think in general it should be fine, I don't expect users to care too much about the exact name of models, and if they are specifying them explicitly then its likely it won't be a duplicate.

szha:
Job PR-762-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-762/3/index.html

PR #764, opened 2020-11-09
Tabular: Added unit tests for models

Innixma:
*Issue #, if available:*

*Description of changes:*

- Added unit tests for models
- FastAI NN unit tests are disabled at present since required dependencies are not installed during test runs on CI.

### DIFF
--- a/tabular/tests/conftest.py
+++ b/tabular/tests/conftest.py
@@ -29,9 +29,33 @@ def pytest_collection_modifyitems(config, items):
 
 
 class DatasetLoaderHelper:
+    dataset_info_dict = dict(
+        # Binary dataset
+        adult={
+            'url': 'https://autogluon.s3.amazonaws.com/datasets/AdultIncomeBinaryClassification.zip',
+            'name': 'AdultIncomeBinaryClassification',
+            'problem_type': BINARY,
+            'label_column': 'class',
+        },
+        # Multiclass big dataset with 7 classes, all features are numeric. Runs SLOW.
+        covertype={
+            'url': 'https://autogluon.s3.amazonaws.com/datasets/CoverTypeMulticlassClassification.zip',
+            'name': 'CoverTypeMulticlassClassification',
+            'problem_type': MULTICLASS,
+            'label_column': 'Cover_Type',
+        },
+        # Regression with mixed feature-types, skewed Y-values.
+        ames={
+            'url': 'https://autogluon.s3.amazonaws.com/datasets/AmesHousingPriceRegression.zip',
+            'name': 'AmesHousingPriceRegression',
+            'problem_type': REGRESSION,
+            'label_column': 'SalePrice',
+        },
+    )
+
     @staticmethod
     def load_dataset(name: str, directory_prefix: str = './datasets/'):
-        dataset_info = DatasetLoaderHelper.dataset_info_dict()[name]
+        dataset_info = DatasetLoaderHelper.dataset_info_dict[name]
         train_file = dataset_info.pop('train_file', 'train_data.csv')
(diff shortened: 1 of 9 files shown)


szha:
Job PR-764-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-764/1/index.html

tabular/tests/conftest.py
```diff
@@ -19,3 +26,118 @@ def pytest_collection_modifyitems(config, items):
...
+            test_file=test_file,
+            name=name_inner,
+            url=url,
+        )
+
+        return train_data, test_data, dataset_info
+
+    @staticmethod
+    def dataset_info_dict():
+        dataset_dict = dict(
```
gradientsky, line 51: nit: can be just a constant instead of method

tabular/tests/conftest.py
```diff
@@ -19,3 +26,118 @@ def pytest_collection_modifyitems(config, items):
...
+        predictor.leaderboard(test_data, extra_info=True)
+        assert savedir == predictor.output_directory
+        if delete_directory:
+            shutil.rmtree(savedir, ignore_errors=True)  # Delete AutoGluon output directory to ensure runs' information has been removed.
+        return predictor
+
+    @staticmethod
+    def fit_dataset(train_data, fit_args, sample_size=None):
+        if sample_size is not None and sample_size < len(train_data):
+            train_data = train_data.sample(n=sample_size, random_state=0)
```
gradientsky, line 132: should this be stratifed by target variable to ensure cases when in multiclass we could actually test binary classification?

gradientsky (changes requested):

tabular/tests/conftest.py
Innixma: stratification won't actually solve that since we still drop classes with low count if eval_metric allows it (aka accuracy). Regardless, I don't think we need to worry about this edge case, since I am using large enough samples (1000) that it won't occur. The way to ensure it works would be via targeting log_loss + stratified sampling, but we probably don't need to go that far by having reasonable sample sizes.

tabular/tests/conftest.py
Innixma: Updated

tabular/tests/conftest.py
```diff
@@ -19,3 +27,115 @@ def pytest_collection_modifyitems(config, items):
...
+            os.remove(zip_name)
+
+        train_data = task.Dataset(file_path=train_file_path)
+        test_data = task.Dataset(file_path=test_file_path)
+        return train_data, test_data
+
+
+class FitHelper:
+    @staticmethod
+    def fit_and_validate_dataset(dataset_name, fit_args, sample_size=1000, refit_full=True, delete_directory=True):
```
jwmueller, line 95: do we want time_limits for potentially slow models?

jwmueller:
Not sure if it's worthwhile, but we could also test individual models' HPO & bagging capabilities.

jwmueller (approved):

szha:
Job PR-764-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-764/3/index.html

tabular/tests/conftest.py
Innixma: I considered it, but I think we should only do this if unit tests become slow, since the alternative is the test could fail because the model didn't have enough time to train and didn't return a result.

Innixma:
> Not sure if it's worthwhile, but we could also test individual models' HPO & bagging capabilities.

I think this is reasonable, but we need to be careful about the time testing all of these things requires.

gradientsky (approved):

PR #768, opened 2020-11-10
[Part 1] Tabular: Added TabularPredictorV2

Innixma:
*Issue #, if available:*

*Description of changes:*

- Added TabularPredictorV2
- Numerous internal cleanup and extensibility improvements

Note:
- I have various TODO's of logic that I plan to update after a few other major PRs are merged to avoid requiring rebases from other contributors.

This PR is part 1 of the API refactor logic:

Part 1: Majority of logic changes
Part 2: Migrate unit tests to use new API
Part 3: Remainder of logic changes + deprecate old API (Old API incompatible with the more complex logic changes without significant effort)
Part 4: Documentation of new API
Part 5: Removal of old Tabular API
Part 6: Update Text API to align with new API
Part 7: Update Vision API to align with new API

### DIFF
--- a/core/src/autogluon/core/task/base/base_task.py
+++ b/core/src/autogluon/core/task/base/base_task.py
@@ -98,9 +98,9 @@ searcher_for_hyperband_strategy = {
 
 
 def compile_scheduler_options(
-        scheduler_options, search_strategy, nthreads_per_trial,
-        ngpus_per_trial, num_trials, time_out, search_options=None, checkpoint=None, resume=False, visualizer=None,
-        time_attr=None, reward_attr=None, dist_ip_addrs=None, epochs=None):
+        scheduler_options, search_strategy, search_options, nthreads_per_trial,
+        ngpus_per_trial, checkpoint, num_trials, time_out, resume, visualizer,
+        time_attr, reward_attr, dist_ip_addrs, epochs=None):
     """
     Updates a copy of scheduler_options (scheduler-specific options, can be
     empty) with general options. The result can be passed to __init__ of the
@@ -134,8 +134,6 @@ def compile_scheduler_options(
         scheduler_options = dict()
     else:
         assert isinstance(scheduler_options, dict)
-    if dist_ip_addrs is None:
-        dist_ip_addrs = []
     assert isinstance(search_strategy, str)
     if search_options is None:
         search_options = dict()
@@ -144,7 +142,7 @@ def compile_scheduler_options(
     if time_attr is None:
         time_attr = 'epoch'
     if reward_attr is None:
-        reward_attr = 'validation_performance'
+        reward_attr = 'accuracy'
     scheduler_options = copy.copy(scheduler_options)
     scheduler_options.update({
         'resource': {
--- a/core/src/autogluon/core/task/base/base_task.py
+++ b/core/src/autogluon/core/task/base/base_task.py
@@ -10,6 +10,7 @@ from ...utils import in_ipynb, try_import_mxnet
 __all__ = [
     'BaseTask',
     'compile_scheduler_options',
+    'compile_scheduler_options_v2',
     'create_scheduler']
 
 Results = collections.namedtuple('Results', 'model reward config time metadata')
@@ -97,9 +98,9 @@ searcher_for_hyperband_strategy = {
 
 
 def compile_scheduler_options(
-        scheduler_options, search_strategy, search_options, nthreads_per_trial,
-        ngpus_per_trial, checkpoint, num_trials, time_out, resume, visualizer,
-        time_attr, reward_attr, dist_ip_addrs, epochs=None):
+        scheduler_options, search_strategy, nthreads_per_trial,
+        ngpus_per_trial, num_trials, time_out, search_options=None, checkpoint=None, resume=False, visualizer=None,
+        time_attr=None, reward_attr=None, dist_ip_addrs=None, epochs=None):
     """
     Updates a copy of scheduler_options (scheduler-specific options, can be
     empty) with general options. The result can be passed to __init__ of the
@@ -133,6 +134,8 @@ def compile_scheduler_options(
         scheduler_options = dict()
     else:
         assert isinstance(scheduler_options, dict)
+    if dist_ip_addrs is None:
+        dist_ip_addrs = []
     assert isinstance(search_strategy, str)
     if search_options is None:
         search_options = dict()
@@ -141,7 +144,7 @@ def compile_scheduler_options(
     if time_attr is None:
         time_attr = 'epoch'
     if reward_attr is None:
-        reward_attr = 'accuracy'
+        reward_attr = 'validation_performance'
     scheduler_options = copy.copy(scheduler_options)
     scheduler_options.update({
         'resource': {
@@ -162,3 +165,97 @@ def compile_scheduler_options(
         if epochs is not None:
             scheduler_options['max_t'] = epochs
     return scheduler_options
+
+
+# TODO: Migrate TextPrediction to use this version, delete old version
+def compile_scheduler_options_v2(
+        scheduler_options, nthreads_per_trial,
+        ngpus_per_trial, num_trials, time_out, search_strategy=None, search_options=None, checkpoint=None, resume=False, visualizer=None,
+        time_attr=None, reward_attr=None, dist_ip_addrs=None, epochs=None):
+    """
+    Updates a copy of scheduler_options (scheduler-specific options, can be
+    empty) with general options. The result can be passed to __init__ of the
+    scheduler.
+
+    Special role of epochs for HyperbandScheduler: If the search_strategy
+    involves HyperbandScheduler and epochs is given, then this value is
+    copied to scheduler_options['max_t']. Pass epochs for applications
+    where the time_attr is epoch, and epochs is the maximum number of
+    epochs.
+
+    :param scheduler_options:
+    :param search_strategy:
+    :param search_options:
+    :param nthreads_per_trial:
+    :param ngpus_per_trial:
+    :param checkpoint:
+    :param num_trials:
+    :param time_out:
+    :param resume:
+    :param visualizer:
+    :param time_attr:
+    :param reward_attr:
+    :param dist_ip_addrs:
+    :param kwargs:
+    :param epochs: See above. Optional
+    :return: Copy of scheduler_options with updates
+
+    """
+    if scheduler_options is None:
+        scheduler_options = dict()
+    else:
+        assert isinstance(scheduler_options, dict)
+    if dist_ip_addrs is None:
+        dist_ip_addrs = []
+    if search_strategy is None:
+        search_strategy = 'random'
+    assert isinstance(search_strategy, str)
+    if search_options is None:
+        search_options = dict()
+    if visualizer is None:
+        visualizer = 'none'
+    if time_attr is None:
+        time_attr = 'epoch'
+    if reward_attr is None:
+        reward_attr = 'validation_performance'
+    _scheduler_options = {
+        'resource': {
+            'num_cpus': nthreads_per_trial, 'num_gpus': ngpus_per_trial},
+        'searcher': search_strategy,
+        'search_options': search_options,
+        'checkpoint': checkpoint,
+        'resume': resume,
+        'num_trials': num_trials,
+        'time_out': time_out,
+        'reward_attr': reward_attr,
+        'time_attr': time_attr,
+        'visualizer': visualizer,
+        'dist_ip_addrs': dist_ip_addrs,
+    }
+    _scheduler_options.update(copy.copy(scheduler_options))
+
+    searcher = searcher_for_hyperband_strategy.get(_scheduler_options['searcher'])
+    if searcher is not None:
+        _scheduler_options['searcher'] = searcher
+        if epochs is not None:
+            _scheduler_options['max_t'] = epochs
+    required_options = [
+        'resource',
+        'searcher',
+        'search_options',
+        'checkpoint',
+        'resume',
+        'num_trials',
+        'time_out',
+        'reward_attr',
+        'time_attr',
+        'visualizer',
+        'dist_ip_addrs',
+    ]
+    missing_options = []
+    for option in required_options:
+        if option not in _scheduler_options:
+            missing_options.append(option)
+    if missing_options:
+        raise AssertionError(f'Missing required keys in scheduler_options: {missing_options}')
+    return _scheduler_options
--- a/examples/tabular/example_tabular_predictor_v2.py
+++ b/examples/tabular/example_tabular_predictor_v2.py
@@ -1,6 +1,5 @@
 
 from autogluon.tabular import TabularPrediction as task
-from autogluon.tabular.task.tabular_prediction.predictor import TabularPredictor
 from autogluon.tabular.task.tabular_prediction.predictor_v2 import TabularPredictorV2
 
 
@@ -11,6 +10,7 @@ from autogluon.tabular.task.tabular_prediction.predictor_v2 import TabularPredic
 train_data = task.Dataset(file_path='https://autogluon.s3.amazonaws.com/datasets/Inc/train.csv')
 test_data = task.Dataset(file_path='https://autogluon.s3.amazonaws.com/datasets/Inc/test.csv')
 label = 'class'
+eval_metric = 'roc_auc'
 hyperparameters = {'RF': {}}
 train_data = train_data.head(1000)  # subsample for faster demo
 
@@ -18,30 +18,16 @@ train_data = train_data.head(1000)  # subsample for faster demo
 # Fitting with the old Predictor #
 ##################################
 
-predictor1 = task.fit(train_data=train_data, label=label, hyperparameters=hyperparameters, num_bagging_folds=2)
+predictor1 = task.fit(train_data, label=label, eval_metric=eval_metric, hyperparameters=hyperparameters, num_bagging_folds=2)
 predictor1.leaderboard(test_data)
 
 ##################################
 # Fitting with the new Predictor #
 ##################################
 
-predictor2 = TabularPredictorV2(label=label)
-
-# Showing that the new Predictor can be saved and loaded prior to calling .fit()
-output_dir = predictor2.output_directory
-predictor2.save()
-del predictor2
-predictor2 = TabularPredictorV2.load(output_dir)
-
-predictor2.fit_feature_generator(data=train_data)  # Can also fit the feature generator prior to calling fit
-# train_data_transformed = predictor2.transform_features(train_data)  # FIXME: Make this work
+predictor2 = TabularPredictorV2(label, eval_metric=eval_metric)
+predictor2.fit_feature_generator(train_data)  # Can also fit the feature generator prior to calling fit
 predictor2.fit(train_data, hyperparameters=hyperparameters, num_bagging_folds=2)
-predictor2.leaderboard(test_data)
-
-# Showing that the new Predictor can be reloaded after fit in the same fashion as the old Predictor
-predictor2.save()
-del predictor2
-predictor2 = TabularPredictorV2.load(output_dir)
 predictor2.leaderboard(test_data)
 
 ####################################
@@ -50,7 +36,7 @@ predictor2.leaderboard(test_data)
 
 # Fit extra models at level 0, with 30 second time limit
 hyperparameters_extra1 = {'GBM': {}, 'NN': {}}
-predictor2.fit_extra(hyperparameters_extra1, time_limits=30)
+predictor2.fit_extra(hyperparameters_extra1, time_limit=30)
 
 # Fit new level 1 stacker models that use the level 0 models from the original fit and the previous fit_extra call as base models
 hyperparameters_extra2 = {'CAT': {}, 'NN': {}}
--- a/examples/tabular/example_tabular_predictor_v2.py
+++ b/examples/tabular/example_tabular_predictor_v2.py
@@ -33,6 +33,8 @@ predictor2.save()
 del predictor2
 predictor2 = TabularPredictorV2.load(output_dir)
 
+predictor2.fit_feature_generator(data=train_data)  # Can also fit the feature generator prior to calling fit
+# train_data_transformed = predictor2.transform_features(train_data)  # FIXME: Make this work
 predictor2.fit(train_data, hyperparameters=hyperparameters, num_bagging_folds=2)
 predictor2.leaderboard(test_data)
 
--- a/examples/tabular/example_tabular_predictor_v2.py
+++ b/examples/tabular/example_tabular_predictor_v2.py
@@ -26,8 +26,7 @@ predictor1.leaderboard(test_data)
 ##################################
 
 predictor2 = TabularPredictorV2(label, eval_metric=eval_metric)
-predictor2.fit_feature_generator(train_data)  # Can also fit the feature generator prior to calling fit
-predictor2.fit(train_data, hyperparameters=hyperparameters, num_bagging_folds=2)
+predictor2.fit(train_data, hyperparameters=hyperparameters, num_bag_folds=2)
 predictor2.leaderboard(test_data)
 
 ####################################
--- a/examples/tabular/example_tabular_predictor_v2.py
+++ b/examples/tabular/example_tabular_predictor_v2.py
@@ -18,11 +18,7 @@ train_data = train_data.head(1000)  # subsample for faster demo
 # Fitting with the old Predictor #
 ##################################
 
-predictor1 = task.fit(
-    train_data=train_data,
-    label=label,
-    hyperparameters=hyperparameters,
-)
+predictor1 = task.fit(train_data=train_data, label=label, hyperparameters=hyperparameters, num_bagging_folds=2)
 predictor1.leaderboard(test_data)
 
 ##################################
@@ -35,28 +31,37 @@ predictor2 = TabularPredictorV2(label=label)
 output_dir = predictor2.output_directory
 predictor2.save()
 del predictor2
-predictor2 = TabularPredictorV2.load(output_directory=output_dir)
+predictor2 = TabularPredictorV2.load(output_dir)
 
-predictor2.fit(
-    train_data=train_data,
-    hyperparameters=hyperparameters,
-)
+predictor2.fit(train_data, hyperparameters=hyperparameters, num_bagging_folds=2)
 predictor2.leaderboard(test_data)
 
 # Showing that the new Predictor can be reloaded after fit in the same fashion as the old Predictor
 predictor2.save()
 del predictor2
-predictor2 = TabularPredictorV2.load(output_directory=output_dir)
+predictor2 = TabularPredictorV2.load(output_dir)
 predictor2.leaderboard(test_data)
 
-#######################################################################
-# Constructing old and new Predictor objects using the Learner object #
-#######################################################################
+####################################
+# Advanced fit_extra functionality #
+####################################
 
-learner = predictor2._learner
+# Fit extra models at level 0, with 30 second time limit
+hyperparameters_extra1 = {'GBM': {}, 'NN': {}}
+predictor2.fit_extra(hyperparameters_extra1, time_limits=30)
 
-predictor_new_loaded_from_learner = TabularPredictorV2.from_learner(learner=learner)
-predictor_new_loaded_from_learner.leaderboard(test_data)
+# Fit new level 1 stacker models that use the level 0 models from the original fit and the previous fit_extra call as base models
+hyperparameters_extra2 = {'CAT': {}, 'NN': {}}
+base_model_names = predictor2.get_model_names(stack_name='core', level=0)
+predictor2.fit_extra(hyperparameters_extra2, base_model_names=base_model_names)
 
-predictor_old_loaded_from_learner = TabularPredictor(learner=learner)
-predictor_old_loaded_from_learner.leaderboard(test_data)
+# Fit a new 3-layer stack ensemble on top of level 1 stacker models
+hyperparameters_extra3 = {
+    0: {'XT': {}},
+    1: {'NN': {}, 'RF': {}},
+    2: {'XGB': {}, 'custom': ['GBM']}
+}
+base_model_names = predictor2.get_model_names(stack_name='core', level=1)
+predictor2.fit_extra(hyperparameters_extra3, base_model_names=base_model_names)
+
+predictor2.leaderboard(test_data)
--- a/examples/tabular/example_tabular_predictor_v2.py
+++ b/examples/tabular/example_tabular_predictor_v2.py
@@ -1,0 +1,62 @@
+
+from autogluon.tabular import TabularPrediction as task
+from autogluon.tabular.task.tabular_prediction.predictor import TabularPredictor
+from autogluon.tabular.task.tabular_prediction.predictor_v2 import TabularPredictorV2
+
+
+################
+# Loading data #
+################
+
+train_data = task.Dataset(file_path='https://autogluon.s3.amazonaws.com/datasets/Inc/train.csv')
+test_data = task.Dataset(file_path='https://autogluon.s3.amazonaws.com/datasets/Inc/test.csv')
+label = 'class'
+hyperparameters = {'RF': {}}
+train_data = train_data.head(1000)  # subsample for faster demo
+
+##################################
+# Fitting with the old Predictor #
+##################################
+
+predictor1 = task.fit(
+    train_data=train_data,
+    label=label,
+    hyperparameters=hyperparameters,
+)
+predictor1.leaderboard(test_data)
+
+##################################
+# Fitting with the new Predictor #
+##################################
+
+predictor2 = TabularPredictorV2(label=label)
+
+# Showing that the new Predictor can be saved and loaded prior to calling .fit()
+output_dir = predictor2.output_directory
+predictor2.save()
+del predictor2
+predictor2 = TabularPredictorV2.load(output_directory=output_dir)
+
+predictor2.fit(
+    train_data=train_data,
+    hyperparameters=hyperparameters,
+)
+predictor2.leaderboard(test_data)
+
+# Showing that the new Predictor can be reloaded after fit in the same fashion as the old Predictor
+predictor2.save()
+del predictor2
+predictor2 = TabularPredictorV2.load(output_directory=output_dir)
+predictor2.leaderboard(test_data)
+
+#######################################################################
+# Constructing old and new Predictor objects using the Learner object #
+#######################################################################
+
+learner = predictor2._learner
+
+predictor_new_loaded_from_learner = TabularPredictorV2.from_learner(learner=learner)
+predictor_new_loaded_from_learner.leaderboard(test_data)
+
+predictor_old_loaded_from_learner = TabularPredictor(learner=learner)
+predictor_old_loaded_from_learner.leaderboard(test_data)
--- a/tabular/src/autogluon/tabular/models/ensemble/bagged_ensemble_model.py
+++ b/tabular/src/autogluon/tabular/models/ensemble/bagged_ensemble_model.py
@@ -91,7 +91,7 @@ class BaggedEnsembleModel(AbstractModel):
         else:
             return X
 
-    def _fit(self, X, y, k_fold=5, k_fold_start=0, k_fold_end=None, n_repeats=1, n_repeat_start=0, time_limit=None, **kwargs):
+    def _fit(self, X_train, y_train, k_fold=5, k_fold_start=0, k_fold_end=None, n_repeats=1, n_repeat_start=0, time_limit=None, **kwargs):
         if k_fold < 1:
             k_fold = 1
         if k_fold_end is None:
@@ -131,11 +131,11 @@ class BaggedEnsembleModel(AbstractModel):
                 raise ValueError(f'n_repeats must equal 0 when fitting a single model with k_fold < 2, values: ({self._n_repeats}, {k_fold})')
             model_base.set_contexts(path_context=self.path + model_base.name + os.path.sep)
             time_start_fit = time.time()
-            model_base.fit(X_train=X, y_train=y, time_limit=time_limit, **kwargs)
+            model_base.fit(X_train=X_train, y_train=y_train, time_limit=time_limit, **kwargs)
             model_base.fit_time = time.time() - time_start_fit
             model_base.predict_time = None
-            self._oof_pred_proba = model_base.predict_proba(X=X)  # TODO: Cheater value, will be overfit to valid set
-            self._oof_pred_model_repeats = np.ones(shape=len(X), dtype=np.uint8)
+            self._oof_pred_proba = model_base.predict_proba(X=X_train)  # TODO: Cheater value, will be overfit to valid set
+            self._oof_pred_model_repeats = np.ones(shape=len(X_train), dtype=np.uint8)
             self._n_repeats = 1
             self._n_repeats_finished = 1
             self._k_per_n_repeat = [1]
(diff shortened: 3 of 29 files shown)


szha:
Job PR-768-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-768/1/index.html

tabular/src/autogluon/tabular/task/tabular_prediction/predictor_v2.py
```diff
@@ -0,0 +1,733 @@
...
+
+
+class TabularPredictorV2(TabularPredictor):
+    """
+    AutoGluon Task for predicting values in column of tabular dataset (classification or regression)
+    """
+    Dataset = TabularDataset
+    predictor_file_name = 'predictor.pkl'
+
+    def __init__(
```
jwmueller, line 34: What's the logic to determine what goes here vs in `fit()`?
I'd imagine stuff that affects the models' quality (and could be viewed as a form of "hyperparameter") would go in `fit()`, but it seems some arguments like this are here instead, including: `label_count_threshold`, `feature_generator`, `feature_metadata`

examples/tabular/example_tabular_predictor_v2.py
```diff
@@ -0,0 +1,62 @@
+
+from autogluon.tabular import TabularPrediction as task
+from autogluon.tabular.task.tabular_prediction.predictor import TabularPredictor
```
Innixma, line 3: Consider renaming to Predictor

examples/tabular/example_tabular_predictor_v2.py
```diff
@@ -0,0 +1,62 @@
...
+    label=label,
+    hyperparameters=hyperparameters,
+)
+predictor1.leaderboard(test_data)
+
+##################################
+# Fitting with the new Predictor #
+##################################
+
+predictor2 = TabularPredictorV2(label=label)
```
mli, line 32: how we decide which args to the Predictor, which to the fit. `label` is related to `train_data`. If we change the data later, we may change `label` as well

examples/tabular/example_tabular_predictor_v2.py
```diff
@@ -0,0 +1,62 @@
...
+predictor2.save()
+del predictor2
+predictor2 = TabularPredictorV2.load(output_directory=output_dir)
+predictor2.leaderboard(test_data)
+
+#######################################################################
+# Constructing old and new Predictor objects using the Learner object #
+#######################################################################
+
+learner = predictor2._learner
```
mli, line 56: we should have a `learner` property, to avoid users to access `_` attribute directly.

examples/tabular/example_tabular_predictor_v2.py
```diff
@@ -0,0 +1,62 @@
...
+    train_data=train_data,
+    hyperparameters=hyperparameters,
+)
+predictor2.leaderboard(test_data)
+
+# Showing that the new Predictor can be reloaded after fit in the same fashion as the old Predictor
+predictor2.save()
+del predictor2
+predictor2 = TabularPredictorV2.load(output_directory=output_dir)
+predictor2.leaderboard(test_data)
```
mli, line 50: why we call it `leaderboard` instead of `predict`

examples/tabular/example_tabular_predictor_v2.py
```diff
@@ -0,0 +1,62 @@
...
+
+#######################################################################
+# Constructing old and new Predictor objects using the Learner object #
+#######################################################################
+
+learner = predictor2._learner
+
+predictor_new_loaded_from_learner = TabularPredictorV2.from_learner(learner=learner)
```
mli, line 58: roughly predictor = learner + data?, so the above just copy the model from `predictor2`? it's a little confusion if learner is not related to data, then how about the the data preprocessing pipeline? maybe we just demonstrate `TabularPredictorV2(predictor2)` or `predictor2.copy()`

examples/tabular/example_tabular_predictor_v2.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+################
+# Loading data #
+################
+
+train_data = task.Dataset(file_path='https://autogluon.s3.amazonaws.com/datasets/Inc/train.csv')
+test_data = task.Dataset(file_path='https://autogluon.s3.amazonaws.com/datasets/Inc/test.csv')
+label = 'class'
+hyperparameters = {'RF': {}}
```
mli, line 14: shall we call it `RandomForest` to avoid confusion, given only a small set of users will construct  `hyperparameters`, making it longer but clearer is ok

examples/tabular/example_tabular_predictor_v2.py
```diff
@@ -0,0 +1,62 @@
...
+from autogluon.tabular.task.tabular_prediction.predictor import TabularPredictor
+from autogluon.tabular.task.tabular_prediction.predictor_v2 import TabularPredictorV2
+
+
+################
+# Loading data #
+################
+
+train_data = task.Dataset(file_path='https://autogluon.s3.amazonaws.com/datasets/Inc/train.csv')
+test_data = task.Dataset(file_path='https://autogluon.s3.amazonaws.com/datasets/Inc/test.csv')
```
mli, line 12: can we avoid `file_path`, instead checking the arg if it is a `str` or a `dataframe`

examples/tabular/example_tabular_predictor_v2.py
```diff
@@ -0,0 +1,62 @@
...
+predictor2 = TabularPredictorV2(label=label)
+
+# Showing that the new Predictor can be saved and loaded prior to calling .fit()
+output_dir = predictor2.output_directory
+predictor2.save()
+del predictor2
+predictor2 = TabularPredictorV2.load(output_directory=output_dir)
+
+predictor2.fit(
+    train_data=train_data,
```
mli, line 41: we can get ride of `train_data=` by using args directly, as the variable name shows it's training data. in general, our code examples should use args as much as possible if the variance names have meanings, such as `fit(train_data, hyperparamters, number_worker=4)`, the last one should use kwargs as it's not easy to infer the meaning from 4.

examples/tabular/example_tabular_predictor_v2.py
```diff
@@ -0,0 +1,62 @@
...
+    hyperparameters=hyperparameters,
+)
+predictor2.leaderboard(test_data)
+
+# Showing that the new Predictor can be reloaded after fit in the same fashion as the old Predictor
+predictor2.save()
+del predictor2
+predictor2 = TabularPredictorV2.load(output_directory=output_dir)
```
mli, line 49: similar to above, we can hide `output_directory=`. besides, we may call it `directory`, the reason is that `output_directory` means the  output directory of the previous predictor when we save it, while `load` focus on the current predictor

examples/tabular/example_tabular_predictor_v2.py
mli, line 3: per discussed, we will rename it `from autogluon.tabular import Predictor`, so other tasks can have a similar import sentence

examples/tabular/example_tabular_predictor_v2.py
mli: btw, i did a search over `file_path` vs `filepath`, and found

According to PEP8 you should use _ between each meaningful words for variable names, Similarly we use capital case for class names. by searching about the word filepath I should say that there is not such a word in English, it means that it is not a single word, it contains two separate word(file, path), so it is correct to use file_path instead of 'filepath', although both of them is being used by developers these days.

tabular/src/autogluon/tabular/task/tabular_prediction/predictor_v2.py
Innixma: My thought process (happy to discuss) is currently:

In general, arguments which could break model fitting in repeated fit calls if changed go in the init, such as feature_generator, feature_metadata, and label_count_threshold (threshold determines which classes are used and how the data is split).

As they are optional arguments, I think we can also have a `.process_features` or `.fit_feature_generator` method to fit the generator prior to calling `.fit`.

1. label_count_threshold will likely live in init kwargs as it is (somewhat) tabular/stacking specific.
2. train_data could potentially live in init, but for now it lives in fit.
3. Once we support multiple feature generators, we could consider having feature_generator and feature_metadata live in fit. The logic to ensure that the stacker can work properly in these scenarios is complex though, particularly if we allow different train_data features in separate fit calls.

examples/tabular/example_tabular_predictor_v2.py
Innixma: `predictor2.predict` also works, this file is purely to demonstrate identical functionality, but I can revise it to be more demonstrating generic usage.

jwmueller:
Perhaps we want to explicitly write out here all the common arguments to `fit()` for each task?

Also all the common instance-methods & attributes of the Predictor object for each task?

examples/tabular/example_tabular_predictor_v2.py
```diff
@@ -0,0 +1,62 @@
...
+
+##################################
+# Fitting with the new Predictor #
+##################################
+
+predictor2 = TabularPredictorV2(label=label)
+
+# Showing that the new Predictor can be saved and loaded prior to calling .fit()
+output_dir = predictor2.output_directory
+predictor2.save()
```
zhreshold, line 36: an open question I would like to discuss is the serialization of TabularPredictor, so for now my understanding it that it's using a directory to save possibly multiple models and to reduce memory cost as you can dump intermediate buffers on disk. 

However, a directory isn't a good choice for serialization, as it's vulnerable(integrity, file system permissions, ...), cannot be safely copied directly or through communications. In constrast, a binary file can be pickled, base64 encoded and passed through https request, or directly shared through queues or pipes.

Any thought on this before we dig into more complex use cases, e.g., multi model prediction tasks?

tabular/src/autogluon/tabular/task/tabular_prediction/predictor_v2.py
```diff
@@ -0,0 +1,733 @@
...
+    @unpack(set_presets)
+    def fit(self,
+            train_data,
+            tuning_data=None,
+            time_limits=None,
+            presets=None,
+            eval_metric=None,
+            stopping_metric=None,
+            auto_stack=False,
+            hyperparameter_tune=False,
```
zhreshold, line 87: I found `num_trials` is sufficient to switch HPO on/off, `hyperparameter_tune == Ture` is equivalent to `num_trials > 1`

examples/tabular/example_tabular_predictor_v2.py
sxjscience: A simple solution is to use the zipfile to group the artifacts.

tabular/src/autogluon/tabular/task/tabular_prediction/predictor_v2.py
Innixma: How does time_limits play into this? If someone want to train for 10 minutes, then they might not care about setting num_trials, and num_trials is less obviously linked to hyperparameter_tune. I foresee most users specifying time_limits rather than num_trials, and few if any specifying both.

examples/tabular/example_tabular_predictor_v2.py
Innixma: Will plan to do this (and retain option for 'RF' as additional alias) when we rework the model names (ref #782)

examples/tabular/example_tabular_predictor_v2.py
Innixma: Removed this section of example, as it was mostly demonstrating internal functionality equivalence rather than user experience.

examples/tabular/example_tabular_predictor_v2.py
Innixma: Will plan to address this question once pre model-fit methods are added such as fit_feature_generator so the utility of the placement of such arguments can be demonstrated and reviewed.

examples/tabular/example_tabular_predictor_v2.py
Innixma: Removed this section of example, as it was mostly demonstrating internal functionality equivalence rather than user experience.

Essentially, Predictor is merely a wrapper around Learner that is more user friendly and limited. Learner is a more complicated object that houses the even more complicated Trainer object, which we generally don't want users to access to avoid confusion and to allow us freedom to update the internal API without affecting the external Predictor API. Learner actually has access to the data, so Predictor adds nothing on top of Learner except for a more friendly and simple API.

The main question should be if we want to make Learner and Trainer private classes such that they have to be imported with a _ import (ex `from autogluon.tabular._trainer import AbstractTrainer`)

examples/tabular/example_tabular_predictor_v2.py
Innixma: Simplified, was overly verbose originally to demonstrate equivalence, but it was not necessary.

examples/tabular/example_tabular_predictor_v2.py
Innixma: Updated

examples/tabular/example_tabular_predictor_v2.py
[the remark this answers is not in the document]
Innixma: Agreed, this is to match the existing API at present. I think we should either have both be `label` or both be `label_column`. I'd personally recommend `label` as it is shorter and makes more sense if later we have `label` be multiple columns.

examples/tabular/example_tabular_predictor_v2.py
Innixma: I agree with this simplification.

If a single argument shares both functionality, what would you recommend we call it?

Ideas:

`data`
`df`
`file_path_or_data`
`file_path_or_df`
`file_path`

examples/tabular/example_tabular_predictor_v2.py
Innixma: In later meeting we came to the conclusion that calling every task's predictor `Predictor` would be confusing, especially for tasks with multiple predictor variants.

Instead planning to revise name to:

```
predictor = TabularPredictor(...)
```

szha:
Job PR-768-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-768/3/index.html

szha:
Job PR-768-4 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-768/4/index.html

szha:
Job PR-768-5 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-768/5/index.html

tabular/src/autogluon/tabular/task/tabular_prediction/predictor.py
```diff
@@ -90,9 +91,59 @@ def __init__(self, learner):
         self.eval_metric = self._learner.eval_metric
         self.label_column = self._learner.label
         self.feature_metadata = self._trainer.feature_metadata
-        self.class_labels = self._learner.class_labels
-        self.class_labels_internal = self._learner.label_cleaner.ordered_class_labels_transformed
-        self.class_labels_internal_map = self._learner.label_cleaner.inv_map
+
+    @property
+    def class_labels(self):
+        return self._learner.class_labels
```
sxjscience, line 97: What will be the behavior of `class_labels` if it's a regression task?

Innixma: It will return None

szha:
Job PR-768-7 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-768/7/index.html

szha:
Job PR-768-8 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-768/8/index.html

szha:
Job PR-768-9 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-768/9/index.html

szha:
Job PR-768-10 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-768/10/index.html

szha:
Job PR-768-12 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-768/12/index.html

szha:
Job PR-768-14 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-768/14/index.html

szha:
Job PR-768-15 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-768/15/index.html

szha:
Job PR-768-16 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-768/16/index.html

szha:
Job PR-768-17 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-768/17/index.html

szha:
Job PR-768-20 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-768/20/index.html

Innixma:
Ran PR on automlbenchmark V2 (104 datasets), slightly outperformed mainline (likely due to the stopping metric logic change in NN) and had fewer errors.

szha:
Job PR-768-25 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-768/25/index.html

szha:
Job PR-768-26 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-768/26/index.html

szha:
Job PR-768-27 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-768/27/index.html

examples/tabular/example_tabular_predictor_v2.py
Innixma: Agreed that zipfile or binary file would be the ideal format for transporting the predictor. Will plan to address this in Part 2/Part 3 as a follow-up

szha:
Job PR-768-28 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-768/28/index.html

szha:
Job PR-768-30 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-768/30/index.html

szha:
Job PR-768-31 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-768/31/index.html

tabular/src/autogluon/tabular/models/ensemble/bagged_ensemble_model.py
```diff
@@ -597,3 +597,71 @@ def _construct_empty_oof(self, X, y):
...
+            y_pred_proba = child.predict_proba(X_val_fold)
+
+            # TODO: Create new Ensemble Here
+            bag = copy.deepcopy(self)
+            bag.name = bag.name + os.path.sep + str(i)
+            bag.set_contexts(self.path_root + bag.name + os.path.sep)
+
+            oof_pred_proba, oof_pred_model_repeats = self._construct_empty_oof(X=X_train, y=y_train)
+            oof_pred_proba[test_index] += y_pred_proba
+            oof_pred_model_repeats[test_index] += 1
```
gradientsky, line 639: This might not work correctly if AUC is the optimization metric since AUC cares only about rank. [Here](https://scikit-learn.org/stable/auto_examples/model_selection/plot_roc_crossval.html) are some ideas on how to better handle it.
I don't expect this to be fixed in this PR, but let's add some TODO to address it later.

tabular/src/autogluon/tabular/task/tabular_prediction/predictor.py
```diff
@@ -91,14 +91,43 @@ def __init__(self, learner):
...
+        return self._trainer.feature_metadata
+
+    @property
+    def feature_metadata_in(self):
+        return self._learner.feature_generator.feature_metadata_in
+
+    # TODO: v0.1 rename to label
+    @property
+    def label_column(self):
+        return self._learner.label
```
gradientsky, line 126: nit: This implies single label classification. Do we expect to add multi-label classification support at some point?

sxjscience: I think we may consider to support multi-label prediction. I'm adding https://www.kaggle.com/c/google-quest-challenge to the multimodal benchmark and it has 30 labels.

Innixma: Yes, label_column will be changed to label, this is only to support legacy API during transition process.

tabular/src/autogluon/tabular/models/ensemble/bagged_ensemble_model.py
Innixma: This is already a TODO at top of Trainer (I've known about it for a long time)

tabular/src/autogluon/tabular/task/tabular_prediction/predictor_v2.py
Innixma: Support for continuing training a model with additional epochs won't be available in v0.1 since it is very hard to universally apply and it is likely quite niche. I do agree than a way to overwrite an existing model would be useful (currently it will just rename the new model so it doesn't overwrite). However, that is also very niche and probably not necessary for v0.1.

2nd fit() will break if label_count_threshold changes because this impacts how many classes are actually outputted and the training data used. This will conflict with OOF preds and various other assumptions. Metadata and generator are not conflicting however, and can be changed as we discussed prior.

tabular/src/autogluon/tabular/task/tabular_prediction/predictor_v2.py
```diff
@@ -0,0 +1,737 @@
...
+    AutoGluon Task for predicting values in column of tabular dataset (classification or regression)
+    """
+    Dataset = TabularDataset
+    predictor_file_name = 'predictor.pkl'
+
+    def __init__(
+            self,
+            label,
+            problem_type=None,
+            output_directory=None,
```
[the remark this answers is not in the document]
Innixma: Good point, added

tabular/src/autogluon/tabular/task/tabular_prediction/predictor_v2.py
```diff
@@ -0,0 +1,737 @@
...
+            eval_metric=None,
+            stopping_metric=None,
+            feature_metadata_in=None,
```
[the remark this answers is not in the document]
Innixma: We can consider changing this, but the reasoning is that FeatureGenerator output metadata is called feature_metadata, and input feature_metadata is called feature_metadata_in. We could potentially change FeatureGenerator to be feature_metadata_out instead to make _in be feature_metadata.

tabular/src/autogluon/tabular/task/tabular_prediction/predictor_v2.py
```diff
@@ -0,0 +1,737 @@
...
+            verbosity=2,
+            id_columns=None,
+            label_count_threshold=10,
+            cache_data=True,
+            feature_generator=None,
+            learner_type=DefaultLearner,
+            trainer_type=AutoTrainer,
```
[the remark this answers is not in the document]
Innixma: This is in kwargs now

core/src/autogluon/core/task/base/base_task.py
```diff
@@ -162,3 +163,97 @@ def compile_scheduler_options(
...
+        'resource',
+        'searcher',
+        'search_options',
+        'checkpoint',
+        'resume',
+        'num_trials',
+        'time_out',
+        'reward_attr',
+        'time_attr',
+        'visualizer',
```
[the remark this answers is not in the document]
Innixma: I currently have it available via both, but only for backwards compatibility purposes. I am happy to restrict it to only HPO if we think there isn't much value add. It currently can be specified via an ag_args_fit kwarg.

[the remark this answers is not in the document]
Innixma: I think we really need a solid tutorial example of using a visualizer however, since it is hard for me to gauge what users are really gaining from this.

examples/tabular/example_tabular_predictor_v2.py
```diff
@@ -0,0 +1,54 @@
+
+from autogluon.tabular import TabularPrediction as task
+from autogluon.tabular.task.tabular_prediction.predictor_v2 import TabularPredictorV2
```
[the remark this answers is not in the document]
Innixma: This will be shortened, the reason it is long is because we don't want it easily visible to users until it is ready.

We could potentially have the dataset loader be in predictor, but it is indeed a bit strange. We can discuss in future PRs.

examples/tabular/example_tabular_predictor_v2.py
```diff
@@ -0,0 +1,54 @@
...
+base_model_names = predictor2.get_model_names(stack_name='core', level=0)
+predictor2.fit_extra(hyperparameters_extra2, base_model_names=base_model_names)
+
+# Fit a new 3-layer stack ensemble on top of level 1 stacker models
+hyperparameters_extra3 = {
+    0: {'XT': {}},
+    1: {'NN': {}, 'RF': {}},
+    2: {'XGB': {}, 'custom': ['GBM']}
+}
+base_model_names = predictor2.get_model_names(stack_name='core', level=1)
```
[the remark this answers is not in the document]
Innixma: We can leave this for a future PR, I think only advanced users will be aware of these names to begin with, and 'stacker' is not the right term, since 'core' models can also be stackers (such as LightGBM_BAG_L1)

tabular/src/autogluon/tabular/task/tabular_prediction/predictor.py
```diff
@@ -91,14 +91,43 @@ def __init__(self, learner):
...
+
+    @property
+    def output_directory(self):
```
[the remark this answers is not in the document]
Innixma: Added TODO, this will be done once V1 is removed.

tabular/src/autogluon/tabular/task/tabular_prediction/predictor_v2.py
```diff
@@ -0,0 +1,614 @@
...
+
+    def _set_feature_generator(self, feature_generator='auto', feature_metadata=None):
+        if self._learner.feature_generator is not None:
+            if isinstance(feature_generator, str) and feature_generator == 'auto':
+                feature_generator = self._learner.feature_generator
+            else:
+                raise AssertionError('FeatureGenerator already exists!')
+        self._learner.feature_generator = self._get_default_feature_generator(feature_generator=feature_generator, feature_metadata=feature_metadata)
+
+    # TODO: Move out of predictor?
```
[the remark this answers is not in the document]
Innixma: Moved out of predictor

tabular/src/autogluon/tabular/task/tabular_prediction/predictor_v2.py
```diff
@@ -0,0 +1,614 @@
...
+        return predictor
+
+    @classmethod
+    def from_learner(cls, learner: AbstractLearner):
+        predictor = cls(label=learner.label, output_directory=learner.path)
+        predictor._set_post_fit_vars(learner=learner)
+        return predictor
+
+    @staticmethod
+    def _validate_init_kwargs(kwargs):
```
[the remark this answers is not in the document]
Innixma: I believe in most packages such as sklearn, they keep the sanitization and validation methods internal to the class. We can consider removing in future PR.

tabular/src/autogluon/tabular/task/tabular_prediction/predictor_v2.py
```diff
@@ -0,0 +1,614 @@
...
+            'learner_kwargs',
+        }
+        invalid_keys = []
+        for key in kwargs:
+            if key not in valid_kwargs:
+                invalid_keys.append(key)
+        if invalid_keys:
+            raise ValueError(f'Invalid kwargs passed: {invalid_keys}\nValid kwargs: {list(valid_kwargs)}')
+
+    def _validate_fit_kwargs(self, kwargs):
```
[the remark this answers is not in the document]
Innixma: I believe in most packages such as sklearn, they keep the sanitization and validation methods internal to the class. We can consider removing in future PR.

tabular/src/autogluon/tabular/task/tabular_prediction/predictor_v2.py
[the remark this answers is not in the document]
Innixma: Also, `_` are private, meaning it won't be visible to user.

tabular/src/autogluon/tabular/task/tabular_prediction/predictor_v2.py
```diff
@@ -0,0 +1,614 @@
...
+
+    def _set_post_fit_vars(self, learner: AbstractLearner = None):
+        if learner is not None:
+            self._learner: AbstractLearner = learner
+        self._learner_type = type(self._learner)
+        if self._learner.trainer_path is not None:
+            self._learner.persist_trainer(low_memory=True)
+            self._trainer: AbstractTrainer = self._learner.load_trainer()  # Trainer object
+
+    def _set_logger_verbosity(self, verbosity=None):
```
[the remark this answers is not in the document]
Innixma: Moved out of predictor

tabular/src/autogluon/tabular/task/tabular_prediction/predictor_v2.py
```diff
@@ -0,0 +1,614 @@
...
+            logger.log(30, 'Warning: Specified num_trials == 1 or time_limit is too small for hyperparameter_tune, disabling HPO.')
+            return None  # FIXME
+
+        scheduler_cls = schedulers[scheduler_options['searcher'].lower()]
+        if scheduler_options['time_out'] is None:
+            scheduler_options.pop('time_out', None)
+        scheduler_options = (scheduler_cls, scheduler_options)  # wrap into tuple
+        return scheduler_options
+
+    def _set_post_fit_vars(self, learner: AbstractLearner = None):
```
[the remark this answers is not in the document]
Innixma: I think this should stay as it is specific to predictor.

tabular/src/autogluon/tabular/task/tabular_prediction/predictor_v2.py
```diff
@@ -0,0 +1,614 @@
...
+        self._post_fit(
+            keep_only_best=kwargs['keep_only_best'],
+            refit_full=kwargs['refit_full'],
+            set_best_to_refit_full=kwargs['set_best_to_refit_full'],
+            save_space=kwargs['save_space'],
+        )
+        self.save()
+
+    # TODO: Move to generic, migrate all tasks to same kwargs logic
+    def _init_scheduler(self, hyperparameter_tune_kwargs, time_limit, hyperparameters, num_cpus, num_gpus, num_bag_folds, num_stack_levels):
```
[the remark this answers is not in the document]
Innixma: Will do this in follow-up PR, as it is more disruptive.

tabular/src/autogluon/tabular/task/tabular_prediction/predictor_v2.py
```diff
@@ -0,0 +1,614 @@
...
+        )
+
+        kwargs = self._validate_fit_extra_kwargs(kwargs, extra_valid_keys=list(fit_kwargs_default.keys()))
+
+        kwargs_sanitized = fit_kwargs_default.copy()
+        kwargs_sanitized.update(kwargs)
+
+        return kwargs_sanitized
+
+    def _validate_fit_extra_kwargs(self, kwargs, extra_valid_keys=None):
```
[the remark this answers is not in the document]
Innixma: Refer to comments on other `_validate` methods. Regarding nit: It is actually both. `fit_extra` kwargs is a subset of `fit` kwargs.

tabular/src/autogluon/tabular/task/tabular_prediction/predictor_v2.py
```diff
@@ -0,0 +1,614 @@
...
+        refit_full = kwargs_sanitized['refit_full']
+        set_best_to_refit_full = kwargs_sanitized['set_best_to_refit_full']
+        if refit_full and not self._learner.cache_data:
+            raise ValueError('`refit_full=True` is only available when `cache_data=True`. Set `cache_data=True` to utilize `refit_full`.')
+        if set_best_to_refit_full and not refit_full:
+            raise ValueError('`set_best_to_refit_full=True` is only available when `refit_full=True`. Set `refit_full=True` to utilize `set_best_to_refit_full`.')
+
+        return kwargs_sanitized
+
+    def _validate_fit_data(self, train_data, tuning_data=None, unlabeled_data=None):
```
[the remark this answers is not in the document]
Innixma: Refer to comments on other _validate methods.

tabular/src/autogluon/tabular/task/tabular_prediction/predictor_v2.py
```diff
@@ -0,0 +1,614 @@
...
+                raise ValueError("Column names must match between training and tuning data")
+        if unlabeled_data is not None:
+            train_features = sorted(np.array([column for column in train_data.columns if column != self.label_column]))
+            unlabeled_features = sorted(np.array([column for column in unlabeled_data.columns]))
+            if np.any(train_features != unlabeled_features):
+                raise ValueError("Column names must match between training and unlabeled data.\n"
+                                 "Unlabeled data must have not the label column specified in it.\n")
+        return train_data, tuning_data, unlabeled_data
+
+    def _set_feature_generator(self, feature_generator='auto', feature_metadata=None):
```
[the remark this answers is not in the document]
Innixma: This is private and is needed for checking that the generator is already present.

tabular/src/autogluon/tabular/task/tabular_prediction/predictor_v2.py
```diff
@@ -0,0 +1,614 @@
...
+            else:
+                raise ValueError(f"Unknown feature_generator preset: '{feature_generator}', valid presets: {['auto']}")
+        if feature_metadata is not None:
+            if feature_generator.feature_metadata_in is None and not feature_generator.is_fit():
+                feature_generator.feature_metadata_in = copy.deepcopy(feature_metadata)
+            else:
+                raise AssertionError('`feature_metadata_in` already exists in `feature_generator`.')
+        return feature_generator
+
+    def _sanitize_stack_args(self, num_bag_folds, num_bag_sets, num_stack_levels, time_limit, auto_stack, num_train_rows):
```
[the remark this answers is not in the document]
Innixma: Refer to comments on other _validate methods.

examples/tabular/example_tabular_predictor_v2.py
```diff
@@ -0,0 +1,54 @@
...
+predictor2.fit_extra(hyperparameters_extra3, base_model_names=base_model_names)
+
+predictor2.leaderboard(test_data)
```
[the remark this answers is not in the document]
Innixma: Examples will be fleshed out in future PRs as tutorials. This example is primarily for the dev team to understand the differences between the two APIs.

tabular/src/autogluon/tabular/trainer/model_presets/presets.py
```diff
@@ -75,6 +76,7 @@
     TRANSF=TabTransformerModel,
     TEXT_NN_V1=TextPredictionV1Model,
     FASTTEXT=FastTextModel,
+    ENS_GREEDY=GreedyWeightedEnsembleModel,
```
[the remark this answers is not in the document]
Innixma: Agreed, updated to ENS_WEIGHTED

tabular/src/autogluon/tabular/task/tabular_prediction/predictor_v2.py
```diff
@@ -0,0 +1,614 @@
...
+            num_trials=num_trials,
+            time_out=time_limits_hpo,
+        )
+        if scheduler_options is None:
+            return None
+
+        assert scheduler_options['searcher'] != 'bayesopt_hyperband', "searcher == 'bayesopt_hyperband' not yet supported"
+        # TODO: Fix or remove in v0.1
+        if scheduler_options.get('dist_ip_addrs', None):
+            logger.log(30, 'Warning: dist_ip_addrs does not currently work. Distributed instances will not be utilized.')
```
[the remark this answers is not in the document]
Innixma: Added

tabular/src/autogluon/tabular/task/tabular_prediction/predictor_v2.py
```diff
@@ -0,0 +1,614 @@
...
+        self._trainer = None
+
+    @unpack(set_presets)
+    def fit(self,
+            train_data,
+            tuning_data=None,
+            time_limit=None,
+            presets=None,
+            hyperparameters=None,
+            feature_generator="auto",  # TODO: kwargs?
```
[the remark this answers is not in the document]
Innixma: Moved to kwargs

tabular/src/autogluon/tabular/task/tabular_prediction/predictor_v2.py
```diff
@@ -0,0 +1,614 @@
...
+        kwargs_sanitized = fit_kwargs_default.copy()
+        kwargs_sanitized.update(kwargs)
+
+        return kwargs_sanitized
+
+    def _validate_fit_extra_kwargs(self, kwargs, extra_valid_keys=None):
+        fit_extra_kwargs_default = dict(
+            # data split / ensemble architecture kwargs -> Don't nest but have nested documentation -> Actually do nesting
+            num_bag_sets=None,
+            num_stack_levels=None,
```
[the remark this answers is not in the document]
Innixma: num_bag_sets and num_stack_levels are free to change between fit calls, auto_stack is only available in the first fit call however.

`num_bag_folds` is very tricky, because it dictates `label_count_threshold`, which can't be changed. For example, if someone used `num_bag_folds=5` during first fit and a class only appeared 3 times and eval_metric=log_loss, it would be duplicated so it appeared 5 times so 1 example existed in each fold.

However, if they then called `num_bag_folds=10` in the 2nd call, there would be not enough samples for that class to exist in each fold, causing an exception. We could possibly avoid this by raising an exception pre-emptively if we determine a certain num_bag_folds value will cause this issue.

For now I've added a TODO to num_bag_folds to consider moving to fit_extra kwargs.

tabular/src/autogluon/tabular/task/tabular_prediction/predictor_v2.py
```diff
@@ -0,0 +1,733 @@
...
+from ...features import AutoMLPipelineFeatureGenerator
+from ...learner import AbstractLearner, DefaultLearner
+from ...metrics import get_metric
+from ...trainer import AbstractTrainer, AutoTrainer
+
+
+logger = logging.getLogger()  # return root logger
+
+
+class TabularPredictorV2(TabularPredictor):
```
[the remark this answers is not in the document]
Innixma: Good point, added TODO. This will be handy if users are calling fit_extra many times.

szha:
Job PR-768-32 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-768/32/index.html

szha:
Job PR-768-33 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-768/33/index.html

jwmueller (approved):

examples/tabular/example_tabular_predictor_v2.py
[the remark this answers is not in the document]
jwmueller: I just meant some examples of what the API cannot do will also help dev team understand the intent behind its design better

PR #767, opened 2020-11-10
[WIP - DO NOT MERGE][Scheduler][Stability] Enable forkserver as a default forking mode

gradientsky:
- fixes issues with threads locking - switched to forkserver, refactored code to be picklable
- fixes issues with CUDA - [#620](https://github.com/awslabs/autogluon/issues/602)

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #751, opened 2020-10-29
[Tabular][FastText] add fastText custom model for classification (#580)

songqiang:
*Issue #, if available:* #580 

*Description of changes:*
Added the fastText custom model. See  examples/tabular/tabular-fasttext.md for a demo. 

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #770, opened 2020-11-10
Integrate Text Prediction model to Tabular

sxjscience:
*Issue #, if available:*

*Description of changes:*

Major functionality improvement:
- [x] Integrate Text Prediction model to Tabular
- [x] Add extract_embeddings to TextPrediction model

Other changes:
- [x] Add alias to metrics. For example `mean_squared_error` and `mse` can point to the same scorer.


tabular/src/autogluon/tabular/models/text_prediction/text_prediction_v1_model.py
```diff
@@ -0,0 +1,270 @@
...
+logger = logging.getLogger(__name__)
+
+
+AG_TEXT_IMPORT_ERROR = 'autogluon.text has not been installed. ' \
+                       'You may try to install "autogluon.text" first by running. ' \
+                       '`python3 -m pip install autogluon.text`'
+
+
+class TextPredictionV1Model(AbstractModel):
+    def __init__(self, path: str, name: str, problem_type: str,
```
Innixma, line 24: Why not simplify to:

```
def __init__(self, **kwargs):
     super().__init__(**kwargs)
     self._label_column_name = None
```

tabular/src/autogluon/tabular/models/text_prediction/text_prediction_v1_model.py
```diff
@@ -0,0 +1,270 @@
...
+                                                label_shapes=[label_shape],
+                                                stopping_metric=stopping_metric,
+                                                log_metrics=log_metrics,
+                                                output_directory=os.path.join(self.path, self.name),
+                                                logger=logger,
+                                                base_config=None,
+                                                search_space=search_space)
+        return column_properties
+
+    def _predict_proba(self, X, **kwargs):
```
Innixma, line 129: why not just use the parent? If changes have been made, then you need to add the call to

`X = self.preprocess(X, **kwargs)`

which is currently missing and will cause the model to fail.

tabular/src/autogluon/tabular/models/text_prediction/text_prediction_v1_model.py
```diff
@@ -0,0 +1,270 @@
...
+        try:
+            from autogluon.text.text_prediction.dataset import TabularDataset
+        except ImportError:
+            raise ImportError(AG_TEXT_IMPORT_ERROR)
+
+        # Get arguments from kwargs
+        verbosity = kwargs.get('verbosity', 2)
+        num_cpus = kwargs.get('num_cpus')
+        num_gpus = kwargs.get('num_gpus')
+        X_train = X_train[['sentence_raw_text']]
```
Innixma, line 213: missing call to `self.preprocess`?

tabular/src/autogluon/tabular/models/text_prediction/text_prediction_v1_model.py
```diff
@@ -0,0 +1,270 @@
...
+        try:
+            from autogluon.text.text_prediction.text_prediction \
+                import ag_text_prediction_params, merge_params, get_column_properties, \
+                infer_problem_type, infer_eval_stop_log_metrics
+            from autogluon.text.text_prediction.models.basic_v1 import BertForTextPredictionBasic
+        except ImportError:
+            raise ImportError(AG_TEXT_IMPORT_ERROR)
+
+        # Decide the name of the label column
+        if 'label' in X_train.columns:
```
Innixma, line 85: Would be simpler if self._label_column_name was set to 'label' before this if statement to avoid needing the `else:`

tabular/src/autogluon/tabular/models/text_prediction/text_prediction_v1_model.py
```diff
@@ -0,0 +1,270 @@
...
+            label_col_id = 0
+            while True:
+                self._label_column_name = 'label{}'.format(label_col_id)
+                if self._label_column_name not in X_train.columns:
+                    break
+                label_col_id += 1
+        else:
+            self._label_column_name = 'label'
+        feature_column_properties = get_column_properties(
```
Innixma, line 94: consider using self.feature_metadata here, it already contains the type information (and additional metadata information)

tabular/src/autogluon/tabular/models/text_prediction/text_prediction_v1_model.py
```diff
@@ -0,0 +1,270 @@
...
+        )
+        label_column_property = get_column_properties(
+            pd.DataFrame({self._label_column_name: pd.concat([y_train, y_val])}),
+            metadata=None,
+            label_columns=None,
+            provided_column_properties=None
+        )
+        column_properties = collections.OrderedDict(list(feature_column_properties.items()) +
+                                                    list(label_column_property.items()))
+        problem_type, label_shape = infer_problem_type(column_properties=column_properties,
```
Innixma, line 108: self.problem_type contains the problem type already, if that is what you are looking for. Also, self.num_classes contains the number of classes.

tabular/src/autogluon/tabular/models/text_prediction/text_prediction_v1_model.py
```diff
@@ -0,0 +1,270 @@
...
+        column_properties = self._build_model(X_train=X_train,
+                                              y_train=y_train,
+                                              X_val=X_val,
+                                              y_val=y_val,
+                                              hyperparameters=self.params)
+        print('X_train=', X_train)
+        # Insert the label column
+        X_train.insert(len(X_train.columns), self._label_column_name, y_train)
+        X_val.insert(len(X_val.columns), self._label_column_name, y_val)
+        scheduler_options = self.params['hpo_params']['scheduler_options']
```
Innixma, line 224: why is hyperparameter tuning logic in `_fit`? This should probably not be present. First lets get it working without HPO and then we can add HPO in properly (reference other model implementations for an idea of what is needed)

tabular/src/autogluon/tabular/models/text_prediction/text_prediction_v1_model.py
```diff
@@ -0,0 +1,270 @@
...
+                                   'num_gpus': num_gpus},
+                         time_limits=time_limit,
+                         search_strategy=search_strategy,
+                         search_options=self.params['hpo_params']['search_options'],
+                         scheduler_options=scheduler_options,
+                         num_trials=self.params['hpo_params']['num_trials'],
+                         console_log=verbosity >= 2,
+                         ignore_warning=verbosity < 2)
+
+    def save(self, path: str = None, verbose=True) -> str:
```
Innixma, line 254: Missing `super().save` call

tabular/src/autogluon/tabular/models/text_prediction/text_prediction_v1_model.py
```diff
@@ -0,0 +1,270 @@
...
+        if path is None:
+            path = self.path
+        logger.log(15, f'Save to {path}.')
+        self.model.save(path)
+        return path
+
+    @classmethod
+    def load(cls, path: str, reset_paths=True, verbose=True):
```
Innixma, line 262: Missing `super().load` call, this does not return the correct object. load should load the `TextPredictionV1Model` object.

tabular/src/autogluon/tabular/models/text_prediction/text_prediction_v1_model.py
Innixma: You can save self.model separately and then load it separately to avoid serialization issues. Refer to `TabularNeuralNetModel.save` and `TabularNeuralNetModel.load` for an example.

tabular/src/autogluon/tabular/models/text_prediction/text_prediction_v1_model.py
sxjscience: I've revised to build the object based on the loaded model_info and the model.

tabular/src/autogluon/tabular/models/text_prediction/text_prediction_v1_model.py
```diff
@@ -0,0 +1,270 @@
...
+        X_train = X_train[['sentence_raw_text']]
+        X_val = X_val[['sentence_raw_text']]
+        column_properties = self._build_model(X_train=X_train,
+                                              y_train=y_train,
+                                              X_val=X_val,
+                                              y_val=y_val,
+                                              hyperparameters=self.params)
+        print('X_train=', X_train)
+        # Insert the label column
+        X_train.insert(len(X_train.columns), self._label_column_name, y_train)
```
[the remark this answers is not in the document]
sxjscience: Revised.

tabular/src/autogluon/tabular/models/text_prediction/text_prediction_v1_model.py
```diff
@@ -0,0 +1,270 @@
...
+
+        """
+        try:
+            from autogluon.text.text_prediction.dataset import TabularDataset
+        except ImportError:
+            raise ImportError(AG_TEXT_IMPORT_ERROR)
+
+        # Get arguments from kwargs
+        verbosity = kwargs.get('verbosity', 2)
+        num_cpus = kwargs.get('num_cpus')
```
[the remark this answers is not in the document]
sxjscience: Revised to infer the num_cpus + num_gpus.

tabular/src/autogluon/tabular/models/text_prediction/text_prediction_v1_model.py
sxjscience: Revised.

tabular/src/autogluon/tabular/models/text_prediction/text_prediction_v1_model.py
sxjscience: Removed to directly call the parent.

tabular/src/autogluon/tabular/models/text_prediction/text_prediction_v1_model.py
Innixma: I think it will be more maintainable if we save this object as a pkl instead of reconstructing upon load.

See this code snippet taken from `NNFastAiTabularModel` :

```
        __model = self.model
        self.model = None
        path_final = super().save(path=path, verbose=verbose)
        self.model = __model

        # Export model
        save_pkl.save_with_fn(f'{path_final}{self.model_internals_file_name}', self.model, lambda m, buffer: m.export(buffer), verbose=verbose)
        return path_final

    @classmethod
    def load(cls, path: str, reset_paths=True, verbose=True):
        from fastai.basic_train import load_learner
        model = super().load(path, reset_paths=reset_paths, verbose=verbose)
        model.model = load_pkl.load_with_fn(f'{model.path}{model.model_internals_file_name}', lambda p: load_learner(model.path, p), verbose=verbose)
        return model
```

tabular/src/autogluon/tabular/models/text_prediction/text_prediction_v1_model.py
```diff
@@ -0,0 +1,258 @@
...
+        X_val = self.preprocess(X_val)
+        column_properties = self._build_model(X_train=X_train,
+                                              y_train=y_train,
+                                              X_val=X_val,
+                                              y_val=y_val,
+                                              hyperparameters=self.params)
+        # Insert the label column
+        X_train[self._label_column_name] = y_train
+        X_val[self._label_column_name] = y_val
+        scheduler_options = self.params['hpo_params']['scheduler_options']
```
Innixma, line 186: consider doing params = copy.deepcopy(self.params) to avoid updating objects such as scheduler_options inplace. We don't generally want to update the input self.params values in the fit call.

tabular/src/autogluon/tabular/models/text_prediction/text_prediction_v1_model.py
```diff
@@ -0,0 +1,258 @@
...
+    def _set_default_params(self):
+        try:
+            from autogluon.text.text_prediction.dataset import TabularDataset
+            from autogluon.text.text_prediction.text_prediction import ag_text_prediction_params
+        except ImportError:
+            raise ImportError(AG_TEXT_IMPORT_ERROR)
+        super()._set_default_params()
+        self.params = ag_text_prediction_params.create('default')
+
+    def _fit(self, X_train: pd.DataFrame, y_train: pd.Series,
```
Innixma, line 140: This will currently crash if X_val, y_val are None, you can test this by adding `refit_full=True` to task.fit args. Can you update the code to properly handle this case?

Additionally, for refit_full to work properly, the model needs to be trained with the same number of epochs and same hyperparameters as it was for the final model prior to refit_full.

As an example, in CatBoost at the end of `_fit`:

```
self.params_trained['iterations'] = self.model.tree_count_
```

This ensures that the refit_full model sets the hyperparameter `self.params['iterations'] == self.model.tree_count_`

sxjscience:
The reason that I chose to not use pickle is that it's not a very good data sharing format. It has potential security risk and incompatibility issues: See https://docs.python.org/3/library/pickle.html. Thus, usually we may rely on `json` and other format for serialization. Pickle may be used for serialization in the distributed /multiprocessing setting.

Innixma:
> The reason that I chose to not use pickle is that it's not a very good data sharing format. It has potential security risk and incompatibility issues: See https://docs.python.org/3/library/pickle.html. Thus, usually we may rely on `json` and other format for serialization. Pickle may be used for serialization in the distributed /multiprocessing setting.
> 
> From python3 documentation:
> 
> """
> Warning
> The pickle module is not secure. Only unpickle data you trust.
> 
> It is possible to construct malicious pickle data which will execute arbitrary code during unpickling. Never unpickle data that could have come from an untrusted source, or that could have been tampered with.
> 
> Consider signing data with hmac if you need to ensure that it has not been tampered with.
> 
> Safer serialization formats such as json may be more appropriate if you are processing untrusted data. See Comparison with json.
> """

I don't see how this security concern is any different from the security concern of importing any module including autogluon itself. Additionally, we only expect compatibility for save/load using the same dependency versions as was used during fit, which eliminates the incompatibility issue. We use pickle in a variety of situations in Tabular and it is very effective, so I would have to be convinced as to why we would move away from it. @gradientsky , @jwmueller feel free to weigh in.

full_install.sh
```diff
@@ -1,7 +1,7 @@
-pip install -e core/
-pip install -e tabular/
-pip install -e mxnet/
-pip install -e extra/
-pip install -e text/
-pip install -e vision/
-pip install -e autogluon/
+python3 -m pip install -e core/
```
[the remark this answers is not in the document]
sxjscience: Yes, it ensures that we are using python3. In some systems (for example those installed via the Deep Learning Base AMI), python3 will point to the correct version of python while python can sometimes point to python2.

sxjscience:
It's really not a good practice to rely on pickle to save the trained model. Apart from the security problem, the pickle protocol is also related to the python version. It may happen that what you saved via a newer version of python can not be loaded with an older version of python. I think it's fine to have our own way of serialization. For example, the `torch.save()` functionality has recently been moved to use a `zip` based format: https://pytorch.org/docs/stable/generated/torch.save.html

jwmueller:
> I think it's fine to have our own way of serialization. 

I agree it's ok for different models to implement their own serialization and we don't need to require every model to use pickle, as long as their save()/load() is highly reliable. Of course,  pickle is unpopular for NN model serialization. 

I do think @Innixma's example can be more clear like this, and you should still consider it as an easy-to-maintain save/load strategy:
```
    def save(self, path, verbose=True):
        __model = self.model
        self.model = None
        path_final = super().save(path=path, verbose=verbose) # save this AbstractModel object without NN weights
        weights_path = path + WEIGHTS_FILENAME # where to save NN weights
        mxnet_save(__model, weights_path) # serialize NN weights/architecture using  MXNet/pytorch framework-specific saving mechanism
       self.model = __model
        return path_final

    @classmethod
    def load(cls, path, verbose=True):
        model = super().load(path, reset_paths=reset_paths, verbose=verbose) # model.model will be None at first
        weights_path = path + WEIGHTS_FILENAME
        model.model = mxnet_load(weights_path) # some framework-specific mechanism to load NN weights/architecture
        return model
```

Note that this strategy does not rely on pickle to save NN weights, only the AbstractModel object is pickled (which seems ok to me since users will probably not use this outside of autogluon), and the NN weights/architecture can be serialized however you prefer (so users can use the NN outside of autogluon if they wish).

Innixma:
Right, thanks Jonas for clarifying my point with a code example. I am not asking to save the model weights to pickle, just the wrapper AbstractModel object.

tabular/src/autogluon/tabular/features/generators/auto_ml_pipeline.py
```diff
@@ -62,8 +62,10 @@ class AutoMLPipelineFeatureGenerator(PipelineFeatureGenerator):
     >>>
     >>> X_test_transformed = feature_generator.transform(test_data)
     """
-    def __init__(self, enable_numeric_features=True, enable_categorical_features=True, enable_datetime_features=True,
-                 enable_text_special_features=True, enable_text_ngram_features=True, vectorizer=None, **kwargs):
+    def __init__(self, enable_numeric_features=True, enable_categorical_features=True,
```
Innixma, line 65: Could you split this change into a separate PR so #751 can benefit from the change? Ideally, could you also set `enable_raw_text_feature=False` by default so existing users don't have extra object features being generated/in-memory by default (we can smartly enable it in future). Then, whenever you are testing a text model, simply specify

```
task.fit(..., feature_generator=AutoMLPipelineFeatureGenerator(enable_raw_text_feature=True))
```

Innixma: Also, could you update the name from `enable_raw_text_feature` to `enable_raw_text_features` to be inline with the other parameters?

tabular/src/autogluon/tabular/features/generators/drop_unique.py
```diff
@@ -36,10 +36,16 @@ def _drop_unique_features(X: DataFrame, feature_metadata: FeatureMetadata, max_u
         max_unique_value_count = X_len * max_unique_ratio
         for column in X:
             unique_value_count = len(X[column].unique())
+            # Drop features that are always the same
             if unique_value_count == 1:
                 features_to_drop.append(column)
-            elif feature_metadata.get_feature_type_raw(column) in [R_CATEGORY, R_OBJECT] and (unique_value_count > max_unique_value_count):
-                features_to_drop.append(column)
+            elif feature_metadata.get_feature_type_raw(column) in [R_CATEGORY, R_OBJECT]\
```
Innixma, line 42: Please also split this change out to the same PR as mentioned in the above comment.

core/setup.py
```diff
@@ -58,7 +58,8 @@ def create_version_file():
     'boto3',
     'pandas>=1.0.0,<2.0',
     'scikit-learn>=0.22.0,<0.24',
-    'autograd>=1.3'
+    'autograd>=1.3',
+    'dill==0.3.3',
```
zhreshold, line 62: it dill added to solve some unpickable object?

sxjscience: This is cherry-picked from https://github.com/awslabs/autogluon/pull/767

Innixma:
@sxjscience Due to the complexity of the moved files and thus changing imports, do you think it makes sense to split that into a separate PR? I'm guessing this is what is currently causing the majority of the errors in CI build. This would also help to focus this PR on the new model code.

szha:
Job PR-770-90 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-770/90/index.html

szha:
Job PR-770-107 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-770/107/index.html

szha:
Job PR-770-108 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-770/108/index.html

core/src/autogluon/core/metrics/__init__.py
```diff
@@ -2,23 +2,44 @@
...
 class Scorer(object, metaclass=ABCMeta):
     def __init__(self, name, score_func, optimum, sign, kwargs):
         self.name = name
         self._kwargs = kwargs
         self._score_func = score_func
         self._optimum = optimum
         self._sign = sign
+        self.alias = set()
+
+    def add_alias(self, alias):
```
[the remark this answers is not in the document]
sxjscience: I checked it in the later and will raise a value error.

szha:
Job PR-770-111 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-770/111/index.html

sxjscience:
I'm not sure what's wrong but after the change, the tabular prediction test will sometimes hang.

szha:
Job PR-770-115 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-770/115/index.html

core/src/autogluon/core/utils/utils.py
```diff
@@ -64,14 +64,28 @@ def setup_outputdir(output_directory):
...
+            num_parallel_jobs = get_gpu_count() // ngpus_per_trial
+            nthreads_per_trial = max(get_cpu_count() // num_parallel_jobs, 1)
+        else:
+            nthreads_per_trial = min(get_cpu_count(), 4)
+    nthreads_per_trial = min(nthreads_per_trial, get_cpu_count())
+    ngpus_per_trial = min(ngpus_per_trial, get_gpu_count())
+    if ngpus_per_trial > 1:
         ngpus_per_trial = 1
-        logger.debug("tabular_prediction currently doesn't use >1 GPU per training run. ngpus_per_trial set = 1")
+        logger.debug("tabular_prediction currently doesn't use >1 GPU per training run. "
```
jwmueller, line 87: Can TextPrediction use >1 GPU per training run?

jwmueller:
allow verbosity to control amount of printed output 
(default verbosity=2 should print almost no output)

jwmueller:
In the tutorials, I'd try to reduce use of the phrase "multimodal data" as this is non-specific, and instead try to use phrases like "data tables with text fields" or "tabular data with text and numerical/categorical features"

core/src/autogluon/core/utils/utils.py
sxjscience: Yes, it can use more than 1 GPU.

tabular/src/autogluon/tabular/task/tabular_prediction/tabular_prediction.py
```diff
@@ -603,7 +606,10 @@ def fit(train_data,
         if _feature_generator_kwargs:
             if 'feature_generator' in kwargs:
                 logger.log(30, "WARNING: `feature_generator` was specified and will override any presets that alter feature generation (such as 'ignore_text')")
-        feature_generator = kwargs.get('feature_generator', AutoMLPipelineFeatureGenerator(**_feature_generator_kwargs))
+        if 'TEXT_NN_V1' in hyperparameters:
+            _feature_generator_kwargs['enable_raw_text_features'] = True
```
sxjscience, line 610: @jwmueller @Innixma Here's the logic for setting the `enable_raw_text_features` flag.

zhreshold: So I assume that for every new argument we will need to touch the tabular `hyperparameters`? My concern is that will this become extremely hard to maintain in the long run?

Innixma: This will only be temporary. Likely I will focus on smarter logic as this solution is not comprehensive (It doesn't work on the complex hyperparameters format that incorporates level-wise models).

Because it will be temporary, can you add a clear TODO comment regarding this to replace prior to v0.1 release?

tabular/src/autogluon/tabular/models/text_prediction/text_prediction_v1_model.py
Innixma: This comment still applies.

tabular/src/autogluon/tabular/models/text_prediction/text_prediction_v1_model.py
Innixma: This comment still applies.

tabular/src/autogluon/tabular/models/text_prediction/text_prediction_v1_model.py
Innixma: This comment still applies.

tabular/src/autogluon/tabular/models/text_prediction/text_prediction_v1_model.py
Innixma: This comment still applies. Cleanup in this regard might be easier in a follow-up PR though, so I'd recommend making a TODO.

tabular/src/autogluon/tabular/models/text_prediction/text_prediction_v1_model.py
Innixma: This comment still applies.

tabular/src/autogluon/tabular/models/text_prediction/text_prediction_v1_model.py
```diff
@@ -0,0 +1,280 @@
...
+            Other keyword arguments
+
+        """
+        try:
+            import mxnet as mx
+            from autogluon.text.text_prediction.dataset import TabularDataset
+            from autogluon.text.text_prediction.text_prediction import get_recommended_resource
+        except ImportError:
+            raise ImportError(AG_TEXT_IMPORT_ERROR)
+
```
Innixma, line 180: Add TODO: Raise exception if no GPU (With added functionality to override and use CPU anyway as an AG_args_fit parameter)

Innixma:
@sxjscience could you try rebasing from mainline? This might fix CI due to #816 being merged.

tabular/src/autogluon/tabular/models/tabular_nn/tabular_nn_model.py
```diff
@@ -191,10 +191,6 @@ def _fit(self, X_train, y_train, X_val=None, y_val=None, time_limit=None, report
                len(train_dataset.feature_groups['language']) ))
         # self._save_preprocessor()  # TODO: should save these things for hyperparam tunning. Need one HP tuner for network-specific HPs, another for preprocessing HPs.
 
-        if 'num_gpus' in kwargs and kwargs['num_gpus'] >= 1:
```
Innixma, line 194: Why is this deleted?

tabular/src/autogluon/tabular/models/text_prediction/text_prediction_v1_model.py
sxjscience: Should be addressed.

tabular/src/autogluon/tabular/models/tabular_nn/tabular_nn_model.py
sxjscience: I think previously I removed it to try to make tabular nn work but let me add back.

tabular/src/autogluon/tabular/models/text_prediction/text_prediction_v1_model.py
sxjscience: In fact, I just noticed that it may also be reasonable for users to try out BERT on CPU (not completely disable it). May be we can just raise a warning.

tabular/src/autogluon/tabular/models/tabular_nn/tabular_nn_model.py
sxjscience: Added back.

tabular/src/autogluon/tabular/models/text_prediction/text_prediction_v1_model.py
sxjscience: I think the reason is that TextPrediction is not using the same problem type as tabular... So I'm keeping the two implementations. Merging these two will take some time.

szha:
Job PR-770-123 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-770/123/index.html

szha:
Job PR-770-124 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-770/124/index.html

szha:
Job PR-770-126 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-770/126/index.html

core/src/autogluon/core/metrics/__init__.py
```diff
@@ -303,33 +362,69 @@ def rmse_func(predictions, targets):
 recall = make_scorer('recall',
                      sklearn.metrics.recall_score)
 
+
+def customized_log_loss(y_true, y_pred):
```
Innixma, line 366: Could you describe the need for this? How does this impact Tabular overall if the eval_metric is log_loss?

sxjscience: Basically, it solves an issue in sklearn log_loss. The following script will raise an error for sklearn log_loss:

```python
from sklearn.metrics import log_loss

log_loss([0, 1, 0], [[0.1, 0.2, 0.7], [0.2, 0.7, 0.1], [0.7, 0.1, 0.2]])

```

Error:

```
-> 2232                                                   lb.classes_))
   2233         else:
   2234             raise ValueError('The number of classes in labels is different '

ValueError: y_true and y_pred contain different number of classes 2, 3. Please provide the true labels explicitly through the labels argument. Classes found in y_true: [0 1]
```

The root cause is that sklearn will try to infer the label set by looking at the ground truth labels. This is not needed because we have already inferred the label set in AutoGluon.

Innixma: Found a case where this still errors:

```
from autogluon.tabular import TabularPrediction as task

LABEL = 'class'
SAMPLE = 1000


if __name__ == '__main__':
    path_prefix = 'https://autogluon.s3.amazonaws.com/datasets/AdultIncomeBinaryClassification/'

    train_data = task.Dataset(file_path=path_train)
    test_data = task.Dataset(file_path=path_test)

    import pandas as pd
    with pd.option_context('display.max_columns', None, 'display.width', 1000):
        print(train_data)

    if SAMPLE is not None:
        train_data = train_data.sample(n=SAMPLE, random_state=0)

    predictor = task.fit(train_data=train_data, label=LABEL, hyperparameters='toy', eval_metric='log_loss')

    leaderboard = predictor.leaderboard(test_data, extra_info=True)

    test_data_only_pos = test_data[test_data[LABEL] == ' <=50K']

    # FIXME: This crashes
    predictor.leaderboard(test_data_only_pos)
```

sxjscience: Nice catch. I previously only fixed the multi-class case. It seems that the binary case also contains this issue and I just submitted the fix.

Innixma: This code change appears to fix the issue:
```
def customized_log_loss(y_true, y_pred):
    """

    Parameters
    ----------
    y_true : array-like or label indicator matrix
        Ground truth (correct) labels for n_samples samples.

    y_pred : array-like of float
        The predictions. shape = (n_samples, n_classes) or (n_samples,)

    Returns
    -------
    loss
        The negative log-likelihood
    """
    assert y_true.ndim == 1
    if y_pred.ndim == 1:
        labels = [0, 1]
    else:
        assert y_pred.ndim == 2, 'Only ndim=2 is supported'
        labels = np.arange(y_pred.shape[1], dtype=np.int32)
    return sklearn.metrics.log_loss(y_true.astype(np.int32), y_pred, labels=labels)
```

Is there a reason we are converting to int32?

szha:
Job PR-770-130 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-770/130/index.html

core/src/autogluon/core/metrics/__init__.py
sxjscience: I think both implementations are fine. For converting to int32, sometimes y_true can be float so I made the conversion.

text/src/autogluon/text/text_prediction/text_prediction.py
```diff
@@ -54,12 +54,33 @@ def default() -> dict:
...
-        }
+            'num_trials': 3,               # The number of trials
+        },
+        'seed': None,                      # The seed value
     }
     return ret
 
 
+@ag_text_prediction_params.register()
+def default_no_hpo() -> dict:
```
jwmueller, line 65: Do we want to provide alternate HPs if the user specifies:
`hyperparameters='light'` or `'very_light'` or `'toy'`?

jwmueller (changes requested):
mostly have minor text changes. Bigger remaining questions include: default value for num-gpus across all autogluon models, HPO / 'best_quality'-preset missing in tutorial, allow user to specify different-sized TextPrediction models based on `hyperparameters='default'` vs `'light'`/`'very_light'`

text/src/autogluon/text/text_prediction/text_prediction.py
Innixma: I believe that this is unrelated to those presets, as this is in text_prediction

Innixma: Currently TextPrediction model must be specified explicitly, there is no default that uses it. I think that is the best approach for the moment until we hone in on when exactly it should be used.

core/src/autogluon/core/utils/utils.py
```diff
@@ -64,14 +64,28 @@ def setup_outputdir(output_directory):
 
 
 def setup_compute(nthreads_per_trial, ngpus_per_trial):
```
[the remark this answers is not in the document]
sxjscience: How about to just get all the available resources (`num_gpus` + `num_cpus`) and each model can still decide whether to use all resources or just use part of the resources.

core/src/autogluon/core/utils/utils.py
```diff
@@ -64,14 +64,28 @@ def setup_outputdir(output_directory):
...
+            logger.debug(
+                "tabular_prediction currently doesn't use >1 GPU per training run. "
+                "ngpus_per_trial set = 1")
+        if ngpus_per_trial != 0:
+            num_parallel_jobs = get_gpu_count() // ngpus_per_trial
+            nthreads_per_trial = max(get_cpu_count() // num_parallel_jobs, 1)
+        else:
+            nthreads_per_trial = min(get_cpu_count(), 4)
+    nthreads_per_trial = min(nthreads_per_trial, get_cpu_count())
+    ngpus_per_trial = min(ngpus_per_trial, get_gpu_count())
```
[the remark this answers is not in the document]
Innixma: +1, we should

tabular/src/autogluon/tabular/models/text_prediction/text_prediction_v1_model.py
```diff
@@ -0,0 +1,280 @@
...
+logger = logging.getLogger(__name__)
+
+
+AG_TEXT_IMPORT_ERROR = 'autogluon.text has not been installed. ' \
+                       'You may try to install "autogluon.text" first by running. ' \
+                       '`python3 -m pip install autogluon.text`'
+
+
+class TextPredictionV1Model(AbstractModel):
```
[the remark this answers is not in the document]
sxjscience: My intention is to later upgrade to V2 while still keep V1.

tabular/src/autogluon/tabular/models/text_prediction/text_prediction_v1_model.py
```diff
@@ -0,0 +1,280 @@
+"""Text Prediction Model based on BERT"""
```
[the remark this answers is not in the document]
sxjscience: Good point. Let me change it to pretrained language model, which is more precise.

szha:
Job PR-770-131 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-770/131/index.html

szha:
Job PR-770-132 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-770/132/index.html

szha:
Job PR-770-133 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-770/133/index.html

docs/tutorials/tabular_prediction/tabular-multimodal-text-others.md
```diff
@@ -0,0 +1,267 @@
...
+                                         stopping_metric='accuracy',
+                                         hyperparameters='default',
+                                         output_directory='model5')
+```
+
+
+```{.python .input}
+predictor_model5.leaderboard(dev_df.join(pd.DataFrame(dev_sentence_multimodal_embeddings)))
+```
+
```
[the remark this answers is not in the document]
sxjscience: How about for me to do it in a following PR. This one is getting too big.

docs/tutorials/tabular_prediction/tabular-multimodal-text-others.md
```diff
@@ -0,0 +1,264 @@
...
+```{.python .input}
+predictor_model6.leaderboard(dev_df)
+```
+
+## Major Take-aways
+
+After performing these comparisons, we have the following takeaways:
+- The multimodal text neural network structure used in TextPrediction, which is based on pretrained language model, is a good network strcuture for dealing with multi-modal data.
+- K-fold bagging / stacking is helpful
+- We need a larger backbone. This aligns with the observation in recent papers, e.g., [Scaling Laws for Autoregressive Generative Modeling](https://arxiv.org/abs/2010.14701).
```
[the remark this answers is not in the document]
sxjscience: @jwmueller How about for me to do it in a later PR (after I obtained more results).

szha:
Job PR-770-134 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-770/134/index.html

szha:
Job PR-770-135 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-770/135/index.html

docs/tutorials/tabular_prediction/tabular-multimodal-text-others.md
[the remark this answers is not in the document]
jwmueller: I can add a sentence for each model in followup PR. I think it's important to do this, otherwise this tutorial will confuse beginner users with all of the different options being presented

jwmueller (approved):
looks great, rest of issues can be addressed in followup PRs

szha:
Job PR-770-137 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-770/137/index.html

Innixma (approved):
Awesome work!

PR #777, opened 2020-11-13
Add missing __init__.py to tab_transformer

jwmueller:


Innixma:
We will really want to find a way to have CI detect this and properly error, this is a tricky issue to spot in PR contributions before they are merged.

- [ ] CI fails when init file is missing that would cause imports to fail when using pip

Innixma (approved):

PR #780, opened 2020-11-15
Update ConfigSpace dependency

jonashaag:
Fixes install when NumPy isn't already installed, see https://github.com/automl/ConfigSpace/pull/115

### DIFF
--- a/core/setup.py
+++ b/core/setup.py
@@ -52,7 +52,7 @@ requirements = [
     'paramiko>=2.4',
     'dask>=2.6.0',
     'distributed>=2.6.0',
-    'ConfigSpace<=0.4.10',
+    'ConfigSpace<=0.4.16',
     'graphviz<0.9.0,>=0.8.1',
     'scikit-optimize',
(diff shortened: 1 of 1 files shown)


szha:
Job PR-780-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-780/1/index.html

Innixma:
Looks good to me! @jwmueller , @zhreshold was there any special reason we chose to limit the version in the past?

jwmueller (approved):

jwmueller:
> was there any special reason we chose to limit the version in the past?

Not that I can recall. Perhaps one of the intermediate versions between 0.4.11 and 0.4.15 didn't work, but I think it should be ok to stick with 0.4.16 as the new limit regardless given this version appears to work.

zhreshold (approved):

Innixma (approved):
Thanks for the contribution!

PR #786, opened 2020-11-17
adding stack ensemble visualization

muhyun:
*Issue #, if available:*

NA

*Description of changes:*

It is to add a function to Tabular predictor which visualizes the stacking ensemble model. It does not take any parameter, and it saves the generated graph into the model directory.

This function relies on graphviz and pygraphviz. These libraries can be installed via;

```
sudo apt-get install graphviz
pip install graphviz
```

### DIFF
--- a/tabular/src/autogluon/tabular/task/tabular_prediction/predictor.py
+++ b/tabular/src/autogluon/tabular/task/tabular_prediction/predictor.py
@@ -1,7 +1,11 @@
 import copy
 import logging
 
+import os
 import pandas as pd
+from PIL import Image
+
+import networkx as nx
 
 from .dataset import TabularDataset
 from .hyperparameter_configs import get_hyperparameter_config
@@ -1086,6 +1090,47 @@ class TabularPredictor(BasePredictor):
         return self._learner.distill(X=train_data, X_val=tuning_data, time_limits=time_limits, hyperparameters=hyperparameters, holdout_frac=holdout_frac,
                                      verbosity=verbosity, models_name_suffix=models_name_suffix, teacher_preds=teacher_preds,
                                      augmentation_data=augmentation_data, augment_method=augment_method, augment_args=augment_args)
+
+    def plot_ensemble_model(self):
+        """
+            Output the visualized stack ensemble architecture of a model trained by `fit()`. 
+            The plot is stored to a file, `ensemble_model.png` in folder `Predictor.output_directory` 
+
+            This function requires `graphviz` and `pygraphviz` to be installed because this visualization depends on those package.
+            Unless this function will raise `ImportError` without being able to generate the visual of the ensemble model.
+
+        """
+        try:
+            import pygraphviz
+        except:
+            raise ImportError('Visualizing ensemble network architecture requires pygraphviz library')
+            
+        G = self._trainer.model_graph
+        remove = [node for node,degree in dict(G.degree()).items() if degree < 1]
+        G.remove_nodes_from(remove)
+        root_node = [n for n,d in G.out_degree() if d == 0]
+        best_model_node = self.get_model_best()
+
+        A = nx.nx_agraph.to_agraph(G)
+        
+        A.graph_attr.update(rankdir='BT')
+        A.node_attr.update(fontsize=10)
+        A.node_attr.update(shape='rectangle')
+
+        for node in A.iternodes():
+            node.attr['label'] = f"{node.name}\nVal score: {float(node.attr['val_score']):.4f}"
+
+            if node.name == best_model_node:
+                node.attr['style'] = 'filled'
+                node.attr['fillcolor'] = '#ff9900'
+                node.attr['shape'] = 'box3d'
+            elif nx.has_path(G, node.name, best_model_node):
+                node.attr['style'] = 'filled'
+                node.attr['fillcolor'] = '#ffcc00'
+
+        model_image_fname = os.path.join(self.output_directory, 'ensemble_model.png')
+
+        A.draw(model_image_fname, format='png', prog='dot')
 
     @staticmethod
     def _summarize(key, msg, results):
--- a/tabular/src/autogluon/tabular/task/tabular_prediction/predictor.py
+++ b/tabular/src/autogluon/tabular/task/tabular_prediction/predictor.py
@@ -1106,8 +1106,13 @@ class TabularPredictor(BasePredictor):
             raise ImportError('Visualizing ensemble network architecture requires pygraphviz library')
             
         G = self._trainer.model_graph
-        remove = [node for node,degree in dict(G.degree()).items() if degree < 1]
-        G.remove_nodes_from(remove)
+
+        nodes_without_outedge = [node for node,degree in dict(G.degree()).items() if degree < 1]
+        nodes_no_val_score = [node for node in G if G.nodes[node]['val_score'] == None]
+        
+        G.remove_nodes_from(nodes_without_outedge)
+        G.remove_nodes_from(nodes_no_val_score)
+
         root_node = [n for n,d in G.out_degree() if d == 0]
         best_model_node = self.get_model_best()
 
--- a/tabular/src/autogluon/tabular/task/tabular_prediction/predictor.py
+++ b/tabular/src/autogluon/tabular/task/tabular_prediction/predictor.py
@@ -1091,7 +1091,7 @@ class TabularPredictor(BasePredictor):
                                      verbosity=verbosity, models_name_suffix=models_name_suffix, teacher_preds=teacher_preds,
                                      augmentation_data=augmentation_data, augment_method=augment_method, augment_args=augment_args)
 
-    def plot_ensemble_model(self):
+    def plot_ensemble_model(self, prune_unused_nodes=True):
         """
             Output the visualized stack ensemble architecture of a model trained by `fit()`. 
             The plot is stored to a file, `ensemble_model.png` in folder `Predictor.output_directory` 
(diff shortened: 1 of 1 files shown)


szha:
Job PR-786-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-786/1/index.html

Innixma:
Thanks for the PR, really excited for this contribution! One bug I've found is that if I specify 'refit_full=True' in task.fit, it will crash when I try to plot since the val_score for _FULL models is None:

`ValueError: could not convert string to float: 'None'`

Could this be changed to handle None values (ideally ignore them from consideration when determining the best model)?

Additionally, would it be possible to add `dataset` as an argument so the test score can also be outputted (internally call `self.leaderboard(dataset)` to get the test scores)? This would be a helpful way to visualize also the performance of the _FULL models. Finally, similar to how val_score is displayed, could you add optional argument with a list of extra info the user can request be displayed (such as fit_time, pred_time_val, fit_order, hyperparameters, etc., which are available either in the graph itself or via predictor.leaderboard(extra_info=True).

Innixma:
ensemble_model

Pretty cool stuff! Turn it sideways and you see a MLP ;) . We may want to play around with the edge line thickness + arrow thickness, I think it may be useful to make them thinner.

tabular/src/autogluon/tabular/task/tabular_prediction/predictor.py
```diff
@@ -1087,6 +1091,47 @@ def distill(self, train_data=None, tuning_data=None, augmentation_data=None, tim
...
+
+            if node.name == best_model_node:
+                node.attr['style'] = 'filled'
+                node.attr['fillcolor'] = '#ff9900'
+                node.attr['shape'] = 'box3d'
+            elif nx.has_path(G, node.name, best_model_node):
+                node.attr['style'] = 'filled'
+                node.attr['fillcolor'] = '#ffcc00'
+
+        model_image_fname = os.path.join(self.output_directory, 'ensemble_model.png')
```
Innixma, line 1131: Would be nice to add a logging statement saying where the image can be found.

tabular/src/autogluon/tabular/task/tabular_prediction/predictor.py
```diff
@@ -1087,6 +1091,47 @@ def distill(self, train_data=None, tuning_data=None, augmentation_data=None, tim
                                      verbosity=verbosity, models_name_suffix=models_name_suffix, teacher_preds=teacher_preds,
                                      augmentation_data=augmentation_data, augment_method=augment_method, augment_args=augment_args)
 
+    def plot_ensemble_model(self):
+        """
+            Output the visualized stack ensemble architecture of a model trained by `fit()`. 
+            The plot is stored to a file, `ensemble_model.png` in folder `Predictor.output_directory` 
+
+            This function requires `graphviz` and `pygraphviz` to be installed because this visualization depends on those package.
```
Innixma, line 1099: Can you add more detailed install instructions for users (Linux and Mac instructions)?

tabular/src/autogluon/tabular/task/tabular_prediction/predictor.py
```diff
@@ -1087,6 +1091,47 @@ def distill(self, train_data=None, tuning_data=None, augmentation_data=None, tim
...
+            Unless this function will raise `ImportError` without being able to generate the visual of the ensemble model.
+
+        """
+        try:
+            import pygraphviz
```
Innixma, line 1104: Would graphviz work instead of pygraphviz? What are the advantages of pygraphviz? I mention this because we already use graphviz elsewhere in the project.

tabular/src/autogluon/tabular/task/tabular_prediction/predictor.py
```diff
@@ -1087,6 +1091,47 @@ def distill(self, train_data=None, tuning_data=None, augmentation_data=None, tim
...
+        except:
+            raise ImportError('Visualizing ensemble network architecture requires pygraphviz library')
+            
+        G = self._trainer.model_graph
+        remove = [node for node,degree in dict(G.degree()).items() if degree < 1]
```
Innixma, line 1109: Why remove these nodes? What if the model with the best test score isn't used by any other model or I simply want to see all the models which were trained?

tabular/src/autogluon/tabular/task/tabular_prediction/predictor.py
```diff
@@ -1087,6 +1091,47 @@ def distill(self, train_data=None, tuning_data=None, augmentation_data=None, tim
...
+        G.remove_nodes_from(remove)
```
Innixma, line 1110: isn't this editing the model graph inplace? Won't this break the predictor?

tabular/src/autogluon/tabular/task/tabular_prediction/predictor.py
```diff
@@ -1087,6 +1091,47 @@ def distill(self, train_data=None, tuning_data=None, augmentation_data=None, tim
...
+        A.graph_attr.update(rankdir='BT')
+        A.node_attr.update(fontsize=10)
+        A.node_attr.update(shape='rectangle')
+
+        for node in A.iternodes():
+            node.attr['label'] = f"{node.name}\nVal score: {float(node.attr['val_score']):.4f}"
+
+            if node.name == best_model_node:
+                node.attr['style'] = 'filled'
+                node.attr['fillcolor'] = '#ff9900'
```
Innixma, line 1125: could we have variable names for these colors so it is more interpretable: i.e. `gold = '#ff9900'`

jwmueller:
Awesome contribution!  

Since currently every stacker model in level 2 depends on every model in level 1, here's one possible way to make the graph more visually digestable: 

Add a special "All level X models" node between stack-layers (perhaps next to the corresponding weighted ensemble), where this node is connected to all models in the stack layer X, and stacker models in layer X+1. 

This might also make the graph better vertically spaced, it currently looks a bit squashed.

szha:
Job PR-786-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-786/2/index.html

tabular/src/autogluon/tabular/task/tabular_prediction/predictor.py
muhyun: I will return the image location. I think it is better to print out it.

tabular/src/autogluon/tabular/task/tabular_prediction/predictor.py
muhyun: How about having links to graphviz(https://graphviz.org/download/) and pygraphviz(https://pygraphviz.github.io/documentation/pygraphviz-1.6/install.html)? Also, I want to have apt-get install and pip command as an example.

tabular/src/autogluon/tabular/task/tabular_prediction/predictor.py
muhyun: Dose graphviz provide Python front-end? If we have graphviz in the package already, then installing pygraphviz should be easily done by adding pip install pygraphviz.

tabular/src/autogluon/tabular/task/tabular_prediction/predictor.py
muhyun: I removed models which are not used by the best model. Unless, we will have several isolated nodes in the visual. I think it will be better to provide an option like prune_nodes (True or False).

tabular/src/autogluon/tabular/task/tabular_prediction/predictor.py
muhyun: It should return the reference rather than the copy. You are right. I will make a copy for graph visualization.

tabular/src/autogluon/tabular/task/tabular_prediction/predictor.py
muhyun: That makes sense. I will reflect this to the code as well.

muhyun:
> Thanks for the PR, really excited for this contribution! One bug I've found is that if I specify 'refit_full=True' in task.fit, it will crash when I try to plot since the val_score for _FULL models is None:
> 
> `ValueError: could not convert string to float: 'None'`
> 
> Could this be changed to handle None values (ideally ignore them from consideration when determining the best model)?
> 
> Additionally, would it be possible to add `dataset` as an argument so the test score can also be outputted (internally call `self.leaderboard(dataset)` to get the test scores)? This would be a helpful way to visualize also the performance of the _FULL models. Finally, similar to how val_score is displayed, could you add optional argument with a list of extra info the user can request be displayed (such as fit_time, pred_time_val, fit_order, hyperparameters, etc., which are available either in the graph itself or via predictor.leaderboard(extra_info=True).

1. ValueError: I fixed this by removing those nodes from the graph.

2. test score: Do you mean that it is to put the test score of each model (or node in the graph) along with val score? I need your opinion on this.

3. what do you mean by "_FULL models"?

4. extra info: Yes, this is very doable, but the node rectangle might be messy. Can you narrow down the options?

szha:
Job PR-786-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-786/3/index.html

szha:
Job PR-786-4 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-786/4/index.html

Innixma (approved):
Looks good, majority of remaining asks are added functionality which can be included in future PRs that extend this method. Thanks for the contribution!

Innixma:
> > Thanks for the PR, really excited for this contribution! One bug I've found is that if I specify 'refit_full=True' in task.fit, it will crash when I try to plot since the val_score for _FULL models is None:
> > `ValueError: could not convert string to float: 'None'`
> > Could this be changed to handle None values (ideally ignore them from consideration when determining the best model)?
> > Additionally, would it be possible to add `dataset` as an argument so the test score can also be outputted (internally call `self.leaderboard(dataset)` to get the test scores)? This would be a helpful way to visualize also the performance of the _FULL models. Finally, similar to how val_score is displayed, could you add optional argument with a list of extra info the user can request be displayed (such as fit_time, pred_time_val, fit_order, hyperparameters, etc., which are available either in the graph itself or via predictor.leaderboard(extra_info=True).
> 
> 1. ValueError: I fixed this by removing those nodes from the graph.
> 2. test score: Do you mean that it is to put the test score of each model (or node in the graph) along with val score? I need your opinion on this.
> 3. what do you mean by "_FULL models"?
> 4. extra info: Yes, this is very doable, but the node rectangle might be messy. Can you narrow down the options?

2. Yes, but I think that is best left as a future contribution (I might add it along with 4 since it will probably be easier than through code review).
3. models that are created via `predictor.refit_full()`.
4. Same as 2, likely I can contribute this addition as there are many possible values to print out.

tabular/src/autogluon/tabular/task/tabular_prediction/predictor.py
Innixma: I meant the python package [graphviz](https://github.com/xflr6/graphviz) (separate from graphviz the non-python package).

PR #787, opened 2020-11-17
parallel and stashing tutorials

zhreshold:
*Issue #, if available:*

*Description of changes:*

Parallel jenkins build_docs stage

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -513,7 +513,7 @@ stage("Build Tutorials") {
         stash includes: 'docs/_build/torch/*', name: 'torch'
       }
     }
-  },
+  }
 }
 
 stage("Build Docs") {
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -275,6 +275,240 @@ stage("Unit Test") {
   }
 }
 
+stage("Build Tutorials") {
+  parallel 'course': {
+    node('linux-gpu') {
+      ws('workspace/autogluon-tutorial-course') {
+        checkout scm
+        VISIBLE_GPU=env.EXECUTOR_NUMBER.toInteger() % 8
+        sh """#!/bin/bash
+        set -ex
+        conda env update -n autogluon_tutorial_course -f docs/build_contrib.yml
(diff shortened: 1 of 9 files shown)


szha:
Job PR-787-11 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-787/11/index.html

Jenkinsfile
```diff
@@ -275,6 +275,198 @@ stage("Unit Test") {
   }
 }
 
+stage("Build Tutorials") {
```
szha, line 278: looking at this file, I think you'd start benefiting from refactoring the pipelines into functions.

szha:
Job PR-787-12 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-787/12/index.html

Jenkinsfile
zhreshold: any working example? I tried global functions but hit something similar to https://github.com/jenkins-docs/creating-a-pipeline-in-blue-ocean/issues/19

szha:
Job PR-787-16 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-787/16/index.html

zhreshold:
The parallel tutorial build is working.
However, per @szha suggestion, it can be improved for readability if the shared functions can be used. 
Please advice if we should wait for solutions or proceed.

Innixma (approved):
CI end-to-end in under an hour?! Count me in!

I think we can do a cleanup to avoid code duplication as a follow-up PR, which I would recommend we do soon so it isn't forgotten.

Another thing I see is that the Text module is taking ~40 minutes for its doc build, @sxjscience do you think you could reduce the tutorial runtime to ~10 minutes as a follow-up PR? This would allow our end-to-end CI to finish in just **_30 minutes_**, which would be awesome.

sxjscience:
Yes, we may change to use less trials.

PR #791, opened 2020-11-19
TabTransformer final fixes PR

Chudbrochil:
Fix HPO CPU, bug fix for unlabeled/train mismatch, TT hyperparameter changes, train/pretrain split time_limits.

### DIFF
--- a/tabular/src/autogluon/tabular/models/tab_transformer/tab_transformer_model.py
+++ b/tabular/src/autogluon/tabular/models/tab_transformer/tab_transformer_model.py
@@ -11,9 +11,9 @@ from tqdm import tqdm
 from .hyperparameters.parameters import get_default_param
 from .hyperparameters.searchspaces import get_default_searchspace
 from ..abstract.abstract_model import AbstractNeuralNetworkModel
-from ...constants import BINARY, REGRESSION, MULTICLASS
 from autogluon.core.utils import try_import_torch
 from ...features.feature_metadata import R_OBJECT, S_TEXT_NGRAM, S_TEXT_AS_CATEGORY
+from autogluon.core.constants import BINARY, REGRESSION, MULTICLASS
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


szha:
Job PR-791-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-791/2/index.html

tabular/src/autogluon/tabular/models/tab_transformer/tab_transformer_model.py
```diff
@@ -385,6 +385,9 @@ def _fit(self, X_train, y_train, X_val=None, y_val=None, X_unlabeled=None, time_
         self.model = self._get_model()
 
         if X_unlabeled is not None:
+            # Can't spend all the time in pretraining, have to split it up.
+            if time_limit is not None:
+                time_limit = time_limit // 2
```
Innixma, line 390: why not just `time_limit / 2`?

tabular/src/autogluon/tabular/task/tabular_prediction/tabular_prediction.py
```diff
@@ -295,7 +295,7 @@ def fit(train_data,
         stack_ensemble_levels : int, default = 0
             Number of stacking levels to use in stack ensemble. Roughly increases model training time by factor of `stack_ensemble_levels+1` (set = 0 to disable stack ensembling).
             Disabled by default, but we recommend values between 1-3 to maximize predictive performance.
-            To prevent overfitting, this argument is ignored unless you have also set `num_bagging_folds >= 2`.
+            To prevent overfitting, this argument is ignored unless you have also set `num_bagging_folds >= 2`.
```
Innixma, line 298: What is this out of curiousity? I also see this edit when I delete this space and add it back.

Innixma (approved):

tabular/src/autogluon/tabular/task/tabular_prediction/tabular_prediction.py
Chudbrochil: Was the weirdest thing. I think pasting it into vim made some sort of slight formatting change? I couldn't "fix it", but the space is there when I look at the code so it seems fine?

tabular/src/autogluon/tabular/models/tab_transformer/tab_transformer_model.py
Innixma: Also, if pretraining takes 2 minutes but 10 minutes was given as total time limit, would it make sense to use 8 minutes instead of 5 for the train fit?

Chudbrochil: Can do time_limit / 2, at the time I was thinking it should be an int.

Chudbrochil: > Also, if pretraining takes 2 minutes but 10 minutes was given as total time limit, would it make sense to use 8 minutes instead of 5 for the train fit?

Absolutely.

Chudbrochil: New commit has the new time_limit logic in it.

szha:
Job PR-791-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-791/3/index.html

Innixma (approved):

PR #794, opened 2020-11-20
Added AG_ARGS_STACK functionality

Innixma:
*Issue #, if available:*

*Description of changes:*

Added functionality to specify that a model should not be used as a stacker and that a model should not use the original features.

Example:

```
hyperparameters = {
    'GBM': [
        # Normal model
        {},

        # Will train in the first level but will not train in stack levels (level > 0)
        {'AG_args': {'valid_stacker': False}},

        # Will not train in the first level but will train in stack levels (level > 0)
        {'AG_args': {'valid_base': False}},

        # Will use original features in the first level but will not use original features in stack levels (level > 0) and instead only use OOF features.
        {'AG_args_ensemble': {'use_orig_features': False}}
    ],
}
```

TODO:

- [ ] Add unit test
- [ ] Add documentation + argument to task.fit
- [x] Add ability for models to have default AG_ARGS values without specifying in hyperparameters dict (such as BERT defaulting to `'valid_stacker'=False`)

### DIFF
--- a/core/src/autogluon/core/constants.py
+++ b/core/src/autogluon/core/constants.py
@@ -14,5 +14,6 @@ REFIT_FULL_SUFFIX = "_FULL"  # suffix appended to model name for refit_single_fu
 # AG_ARGS variables are key names in model hyperparameters to dictionaries of custom AutoGluon arguments.
 AG_ARGS = 'AG_args'  # Contains arguments to control model name, model priority, and the valid configurations which it can be used in.
 AG_ARGS_FIT = 'AG_args_fit'  # Contains arguments that impact model training, such as early stopping rounds, #cores, #gpus, max time limit, max memory usage  # TODO
+AG_ARGS_STACK = 'AG_args_stack'  # Contains arguments that impact model stacking, such as if the model is allowed to be a stacker, or if it is allowed to use the original features  # TODO: v0.1 add to documentation
 
 OBJECTIVES_TO_NORMALIZE = ['log_loss', 'pac_score', 'soft_log_loss']  # do not like predicted probabilities = 0
--- a/core/src/autogluon/core/constants.py
+++ b/core/src/autogluon/core/constants.py
@@ -12,8 +12,26 @@ REFIT_FULL_NAME = 'refit_single_full'  # stack-name used for refit_single_full (
 REFIT_FULL_SUFFIX = "_FULL"  # suffix appended to model name for refit_single_full (aka "compressed") models
 
 # AG_ARGS variables are key names in model hyperparameters to dictionaries of custom AutoGluon arguments.
+# TODO: Have documentation for all AG_ARGS values
 AG_ARGS = 'AG_args'  # Contains arguments to control model name, model priority, and the valid configurations which it can be used in.
 AG_ARGS_FIT = 'AG_args_fit'  # Contains arguments that impact model training, such as early stopping rounds, #cores, #gpus, max time limit, max memory usage  # TODO
-AG_ARGS_STACK = 'AG_args_stack'  # Contains arguments that impact model stacking, such as if the model is allowed to be a stacker, or if it is allowed to use the original features  # TODO: v0.1 add to documentation
+AG_ARGS_ENSEMBLE = 'AG_args_ensemble'  # Contains arguments that impact model ensembling, such as if an ensemble model is allowed to use the original features.  # TODO: v0.1 add to documentation
 
 OBJECTIVES_TO_NORMALIZE = ['log_loss', 'pac_score', 'soft_log_loss']  # do not like predicted probabilities = 0
+
+# TODO: Add docs to dedicated page, or should it live in AbstractModel?
+# TODO: How to reference correct version of docs?
+# TODO: Add error in AG_ARGS if unknown key present
+"""
+AG_ARGS: Dictionary of customization options related to meta properties of the model such as its name, the order it is trained, and the problem types it is valid for.
+    name: (str) The name of the model. This overrides AutoGluon's naming logic and all other name arguments if present.
+    name_main: (str) The main name of the model. In 'RandomForestClassifier', this is 'RandomForest'.
+    name_prefix: (str) Add a custom prefix to the model name. Unused by default.
+    name_type_suffix: (str) Override the type suffix of the model name. In 'RandomForestClassifier', this is 'Classifier'. This comes before 'name_suffix'.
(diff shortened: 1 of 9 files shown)


szha:
Job PR-794-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-794/1/index.html

core/src/autogluon/core/constants.py
```diff
@@ -14,5 +14,6 @@
 # AG_ARGS variables are key names in model hyperparameters to dictionaries of custom AutoGluon arguments.
 AG_ARGS = 'AG_args'  # Contains arguments to control model name, model priority, and the valid configurations which it can be used in.
 AG_ARGS_FIT = 'AG_args_fit'  # Contains arguments that impact model training, such as early stopping rounds, #cores, #gpus, max time limit, max memory usage  # TODO
+AG_ARGS_STACK = 'AG_args_stack'  # Contains arguments that impact model stacking, such as if the model is allowed to be a stacker, or if it is allowed to use the original features  # TODO: v0.1 add to documentation
```
jwmueller, line 17: How advantageous is it to have 3 different dicts here vs just one `AG_ARGS` dict that holds all of these options?  To me the `hyperparameters` argument now seems have an extremely deep amount of dictionary/list nesting.

jwmueller:
For the future:  What about another `AG_ARGS_STACK` key  similar to `valid_stacker` that enables a model to *only* be used as stacker? One could imagine two different RF models (with different hyperparameters), one used as stacker, one used as base model.  

Also what about HPO for stacker models?

core/src/autogluon/core/constants.py
sxjscience: I agree. Is it possible to merge some options together? May be in a later PR since it will also impact `AG_ARGS_FIT`.

Innixma: AG_ARGS_FIT should definitely remain its own separate thing since it is information which is visible to the model. The main question is if AG_ARGS_STACK should be combined with AG_ARGS.

Looking deeper into it, I think that is a fair request.

ex:

In AG_ARGS there exist these options:
-  problem_types: (list) List of valid problem types for the model. `problem_types=['binary']` will result in the model only being trained if `problem_type` is 'binary'.
- disable_in_hpo: (bool) If True, the model will only be trained if `hyperparameter_tune=False`.

These are very similar to the 'valid_stacker' option, so I think it makes sense.

However, there is one aspect which isn't covered here that I haven't implemented and that is specification of the stack/bagged ensemble hyperparameters (not the inner model hyperparameters), which ideally could be different for every model. This probably lives best in some kind of AG_ARGS_STACK key. Thoughts?

Innixma:
> For the future: What about another `AG_ARGS_STACK` key similar to `valid_stacker` that enables a model to _only_ be used as stacker? One could imagine two different RF models (with different hyperparameters), one used as stacker, one used as base model.
> 
> Also what about HPO for stacker models?

Yep, I plan to do that. The real question is how flexible we make it. i.e.: valid_levels, invalid_levels, min_level, max_level, only_stacker, valid_base, etc.

Regarding HPO, are you referring to simply enabling HPO on stackers > level 0, or tuning the stack model hyperparameters (max_models_per_type, etc.)?

core/src/autogluon/core/constants.py
jwmueller: I see your point re additional ensembling hyperparameters, so I guess having the extra dict is ok to maintain separation of concerns. But I'd consider renaming this dict to `AG_ARGS_ENSEMBLE` so it at least encompasses all ensembling hyperparams, and we don't need separate dicts for bagging hyperparams, stacking hyperparams, etc.

I think the `fit()` docs should only have one sentence describing these `AG_ARGS*` dicts, with a link to another static page on our website that describes these dicts in detail. Otherwise it's information overload.

jwmueller:
> Regarding HPO, are you referring to simply enabling HPO on stackers > level 0, or tuning the stack model hyperparameters (max_models_per_type, etc.)?

Simply HPO on the stacker model hyperparams for now, but we may want to revisit this design anyway if it's nontrivial to add. Eg. the current HPO design with just bagging may already not be optimal.

Innixma:
Updated to address comments:

1. Added valid_base option as requested (Usage shown in updated PR description).
2. Renamed to AG_ARGS_ENSEMBLE.
3. Added ability to specify default AG_ARGS values in models (see code change in `greedy_weighted_ensemble_model.py`).
4. Limited AG_ARGS_ENSEMBLE scope to the hyperparameters of the ensemble model (`use_orig_features`, `max_models`, and `max_models_per_type` ), all other logic will live in AG_ARGS or AG_ARGS_FIT.

Added TODO comments regarding documentation, not sure how we want to go about static pages and ensuring they are matching in version.

Re stacker HPO: This is very easy to enable as it already works, just need to remove the line of code disabling it. I'd prefer to do this in a separate PR since it is a logic change that impacts the results of training.

core/src/autogluon/core/constants.py
Innixma: Added AG_ARGS_ENSEMBLE. Regarding the docs, do you have a suggestion for how we ensure they stay aligned such that an old version doesn't link to a new versions docs?

tabular/src/autogluon/tabular/models/ensemble/greedy_weighted_ensemble_model.py
```diff
@@ -81,3 +81,10 @@ def get_info(self):
         info = super().get_info()
         info['model_weights'] = self._get_model_weights()
         return info
+
+    @classmethod
+    def _get_default_ag_args(cls) -> dict:
+        default_ag_args = super()._get_default_ag_args()
+        extra_ag_args = {'valid_base': False}
```
Innixma, line 88: @sxjscience This would be how to specify the stack logic for a model. For BERT, it might make sense for the default to be `{'valid_stacker': False}`.

szha:
Job PR-794-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-794/2/index.html

jwmueller (approved):
looks good

sxjscience (approved):

Innixma:
Merging, will plan to address remaining TODO's in follow-up PR after API refactor since doc changes will be best handled together with the API refactor.

PR #828, opened 2020-12-16
[Vision] Improvements to documentation, inconsistent API, and fix legacy examples

zhreshold:
*Issue #, if available:*
Continue #733

*Description of changes:*

#### remaining unsolved comments from previous PR
it's still unclear what's the right choice or happens to be on the plan for migration towards v0.1

- Using `random_state` vs `random_seed` in `fit` function, I see both terms being heavily used in autogluon
- How to align with Tabular for `predict` and `predict_proba`
- `time_limit` vs `time_limits`
- `def predict(self, x)` or `def predict(self, dataset)`


extra/setup.py
```diff
@@ -46,9 +46,9 @@ def create_version_file():
 
 requirements = [
     'numpy>=1.16.0',
-    'gluoncv==0.9.0',
+    'gluoncv==0.10.0b20201216',
```
Innixma, line 49: Why is gluoncv present twice in this requirements list?

vision/src/autogluon/vision/image_prediction/image_prediction.py
```diff
@@ -134,26 +136,27 @@ def fit(self,
             return
 
         # new HPO task
+        if time_limit is None and num_trials is None:
```
Innixma, line 139: Is this check missing in object detection?

Innixma (approved):
Looks good! In regards to the remaining issues, we can plan on addressing them once the Tabular API refactor PR is merged in.

extra/setup.py
Innixma: Also, how necessary is it to be using pre-releases from gluoncv? What if v0.10 isn't released by the time we want to release v0.1 AG?

szha:
Job PR-828-5 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-828/5/index.html

PR #838, opened 2020-12-28
Tabular: optimized memory efficiency of rf/xt during training

Innixma:
*Issue #, if available:*

*Description of changes:*

Optimized memory efficiency of RF/XT during training.

This helps in the case where the trees are massive in memory and training 8 trees + pickling 8 trees led to 16 trees in memory which raised a memory error during the memory safe check. This change makes it so only 5 trees are in memory at a time during the memory safe check, making it take ~3x less memory to avoid OOM errors.

### DIFF
--- a/tabular/src/autogluon/tabular/models/rf/rf_model.py
+++ b/tabular/src/autogluon/tabular/models/rf/rf_model.py
@@ -69,10 +69,7 @@ class RFModel(AbstractModel):
         n_estimators_final = hyperparams['n_estimators']
 
         n_estimators_minimum = min(40, n_estimators_final)
-        if n_estimators_minimum < 40:
-            n_estimators_test = max(1, math.floor(n_estimators_minimum/5))
-        else:
-            n_estimators_test = 8
+        n_estimators_test = min(4, max(1, math.floor(n_estimators_minimum/5)))
 
         X_train = self.preprocess(X_train)
         n_estimator_increments = [n_estimators_final]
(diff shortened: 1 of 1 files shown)


tabular/src/autogluon/tabular/models/rf/rf_model.py
```diff
@@ -113,8 +110,9 @@ def _fit(self, X_train, y_train, time_limit=None, **kwargs):
             self.model = self.model.fit(X_train, y_train)
             if (i == 0) and (len(n_estimator_increments) > 1):
                 time_elapsed = time.time() - time_train_start
-
-                model_size_bytes = sys.getsizeof(pickle.dumps(self.model))
+                model_size_bytes = 0
+                for estimator in self.model.estimators_:  # Uses far less memory than pickling the entire forest at once
+                    model_size_bytes += sys.getsizeof(pickle.dumps(estimator))
```
gradientsky, line 115: Do we need to get through all of the estimators? One of the option take just 30-ish trees and average them out to get the projection.

gradientsky (approved):

tabular/src/autogluon/tabular/models/rf/rf_model.py
Innixma: calculation based on sample, only at most 4 trees are in this loop

szha:
Job PR-838-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-838/1/index.html

PR #843, opened 2021-01-01
Tabular: Optimized NN category embedding logic

Innixma:
*Issue #, if available:*

*Description of changes:*

- Optimized NN category embedding logic from O(n) -> O(1) where n is the number of categories.
- Added improved time limit control to NN.
- Optimized category handling for FeatureGenerator when minimum_cat_count is set from O(n) -> O(1).

This change results in a 4x speedup on small-medium category features between 5-15, and 100x+ speedup for extreme category count features during transformation.

The [albert](https://www.openml.org/d/41147) dataset had categorical features with 150,000 unique values:
Previously, NN took 1.5 hrs to transform the categorical features prior to fit. (ignoring time limits while doing so)
Now, NN takes only 30 seconds on the same data.

### DIFF
--- a/tabular/src/autogluon/tabular/features/generators/category.py
+++ b/tabular/src/autogluon/tabular/features/generators/category.py
@@ -111,7 +111,8 @@ class CategoryFeatureGenerator(AbstractFeatureGenerator):
                     if len(category_list) > 1:
                         if self.cat_order == 'original':
                             original_cat_order = list(X_category[column].cat.categories)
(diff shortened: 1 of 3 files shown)


gradientsky (approved):

szha:
Job PR-843-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-843/1/index.html

PR #850, opened 2021-01-05
[Fix] Disable TextPrediction when GPU is not available + Fix Tensorboard Integration

sxjscience:
- [x] Disable TextPrediction when GPU is not available
- [x] Add the `visualizer` option to TextPrediction
- [x] Fix Tensorboard. After the fix, we can use the following code to log to tensorboard.

```
predictor = TextPrediction.fit(train_data, label='Sentiment', visualizer='tensorboard')
```

Screen Shot 2021-01-05 at 10 06 39 PM

### DIFF
--- a/text/src/autogluon/text/text_prediction/models/basic_v1.py
+++ b/text/src/autogluon/text/text_prediction/models/basic_v1.py
@@ -587,8 +587,6 @@ class BertForTextPredictionBasic:
         os.makedirs(self._output_directory, exist_ok=True)
         search_space_reg = args(search_space=space.Dict(**self.search_space))
         # Scheduler and searcher for HPO
-        if search_strategy.endswith('hyperband') and time_limits is None:
-            time_limits = 5 * 60 * 60  # 5 hours
         if scheduler_options is None:
             scheduler_options = dict()
         stopping_metric_scorer = get_metric(self._stopping_metric)
--- a/text/src/autogluon/text/text_prediction/models/basic_v1.py
+++ b/text/src/autogluon/text/text_prediction/models/basic_v1.py
@@ -598,7 +598,7 @@ class BertForTextPredictionBasic:
             search_options=search_options,
             nthreads_per_trial=resource['num_cpus'],
             ngpus_per_trial=resource['num_gpus'],
-            checkpoint=None,
+            checkpoint=os.path.join(self._output_directory, 'checkpoint.ag'),
             num_trials=num_trials,
             time_out=scheduler_options.get('time_out'),
             resume=False,
--- a/text/src/autogluon/text/text_prediction/text_prediction.py
+++ b/text/src/autogluon/text/text_prediction/text_prediction.py
@@ -488,7 +488,7 @@ class TextPrediction(BaseTask):
 
         if recommended_resource['num_gpus'] == 0:
             if 'AUTOGLUON_TEXT_TRAIN_WITHOUT_GPU' in os.environ:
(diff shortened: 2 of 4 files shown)


szha:
Job PR-850-4 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-850/4/index.html

szha:
Job PR-850-6 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-850/6/index.html

text/src/autogluon/text/text_prediction/models/basic_v1.py
```diff
@@ -598,7 +596,7 @@ def train(self, train_data, tuning_data, resource,
             search_options=search_options,
             nthreads_per_trial=resource['num_cpus'],
             ngpus_per_trial=resource['num_gpus'],
-            checkpoint=None,
+            checkpoint=os.path.join(self._output_directory, 'checkpoint.ag'),
```
jwmueller, line 599: could this possibly cause issue if 2 models are trained simultaneously in parallel during HPO?

text/src/autogluon/text/text_prediction/text_prediction.py
```diff
@@ -475,17 +475,20 @@ def fit(cls, train_data,
...
                 'grace_period', 10)
             scheduler_options['max_t'] = scheduler_options.get(
-                'max_t', 10)
+                'max_t', 50)
 
         if recommended_resource['num_gpus'] == 0:
-            warnings.warn('Recommend to use GPU to run the TextPrediction task!')
+            raise RuntimeError('Only CPU is detected and we will not proceed to run TextPrediction.'
+                               ' GPU is required to run the model! You may set '
+                               '`ngpus_per_trial` to enable GPU.')
```
jwmueller, line 491: May clarify: 
"we will not proceed to run TextPrediction"
->
"we will not fit TextPrediction models because they will train too slowly"

jwmueller: Should there be a hyperparam flag that the user can specify if they really want to run TextPrediction on CPU?

jwmueller (approved):
just have minor questions, some of which can be addressed in future PRs if at all

text/src/autogluon/text/text_prediction/models/basic_v1.py
sxjscience: No, the results will be displayed correct (I've tried that).

Innixma: Wouldn't they be overwriting each-others checkpoint files?

text/src/autogluon/text/text_prediction/text_prediction.py
Innixma: +1 on flag, but I recommend not adding it right now, probably better to wait until we have per-model cpu/gpu control.

sxjscience: Haha, just added the environment variable support.

text/src/autogluon/text/text_prediction/models/basic_v1.py
Innixma: bumping this in case it wasn't seen due to being resolved.

szha:
Job PR-850-11 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-850/11/index.html

PR #851, opened 2021-01-06
Tabular: Fixed balanced_accuracy scoring issue with missing classes

Innixma:
*Issue #, if available:*

*Description of changes:*

- Changes internal learner scorer to use external label values in eval_metric calls rather than internal label values. This fixes edge case issues in `balanced_accuracy` when multiple new test time classes are provided and are all converted to class `-1` internally, causing an incorrect score to be returned.

### DIFF
--- a/tabular/src/autogluon/tabular/learner/abstract_learner.py
+++ b/tabular/src/autogluon/tabular/learner/abstract_learner.py
@@ -170,30 +170,26 @@ class AbstractLearner:
     def score(self, X: DataFrame, y=None, model=None):
         if y is None:
             X, y = self.extract_label(X)
-        X = self.transform_features(X)
-        y = self.label_cleaner.transform(y)
-        trainer = self.load_trainer()
-        if self.problem_type == MULTICLASS:
-            y = y.fillna(-1)
-            if (not trainer.eval_metric_expects_y_pred) and (-1 in y.unique()):
-                # log_loss / pac_score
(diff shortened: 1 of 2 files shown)


tabular/src/autogluon/tabular/learner/abstract_learner.py
```diff
@@ -170,30 +170,26 @@ def transform_features(self, X):
...
         X = self.transform_features(X)
-        y = self.label_cleaner.transform(y)
-        trainer = self.load_trainer()
-        if self.problem_type == MULTICLASS:
-            y = y.fillna(-1)
-            if (not trainer.eval_metric_expects_y_pred) and (-1 in y.unique()):
-                # log_loss / pac_score
-                raise ValueError(f'Multiclass scoring with eval_metric=\'{self.eval_metric.name}\' does not support unknown classes.')
+        y_internal = self.label_cleaner.transform(y)
+        y_internal = y_internal.fillna(-1)
```
jwmueller, line 190: doesn't balanced_accuracy bug still exist in score_debug() if two different unknown classes are replaced by -1?

Innixma: No, because y_internal is only used in pred_proba metrics, but balanced_accuracy is a pred metric.

szha:
Job PR-851-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-851/2/index.html

jwmueller (approved):
looks good

Innixma:
Benchmarking shows no errors or accuracy drop.

PR #861, opened 2021-01-12
[Bug][Fix] Fix TextPrediction + Other bugs

sxjscience:
Fix bugs revealed from internal benchmarking:

- [x] Update the autogluon-contrib-nlp package to use tokenizers 0.9.4. (Prepare for the upgrade to huggingface/transformers)
- [x] Handling of categorical columnns (pandas categorical type)
- [x] Model is saved as `model.pklmodel.pkl`. Change to `model.pkl`
- [x] Resolve the error when we called refit_full
- [x] Handle missing values in numerical columns in TextPrediction

In the process, fixed two other issues:
- [x] Use `uuid.uuid4()` to generate the directory names so that it will always be unique. Otherwise, we may see issues if we are testing multiple runs simultaneously
- [x] We have one issue when we generate the local paths for saving the ensemble models. Basically, we need to first use `os.realpath()` to expand the paths before doing the prefix match. For example, `./autogluon` is equivalent to `/home/ubuntu/autogluon/dataset`, but the current logic will give an empty prefix.

TODO (In a later PR in which we will support different ways for filling missing values):
- [ ] Store the best iteration and improve refit_full

### DIFF
--- a/core/src/autogluon/core/metrics/__init__.py
+++ b/core/src/autogluon/core/metrics/__init__.py
@@ -386,7 +386,7 @@ def customized_log_loss(y_true, y_pred, eps=1e-15):
     if y_pred.ndim == 1:
         # First clip the y_pred which is also used in sklearn
         y_pred = np.clip(y_pred, eps, 1 - eps)
-        return (y_true * np.log(y_pred) + (1 - y_true) * np.log(1 - y_pred)).mean()
+        return - (y_true * np.log(y_pred) + (1 - y_true) * np.log(1 - y_pred)).mean()
     else:
         assert y_pred.ndim == 2, 'Only ndim=2 is supported'
         labels = np.arange(y_pred.shape[1], dtype=np.int32)
--- a/core/src/autogluon/core/metrics/__init__.py
+++ b/core/src/autogluon/core/metrics/__init__.py
@@ -29,21 +29,6 @@ class Scorer(object, metaclass=ABCMeta):
         self.alias.add(alias)
 
     @property
-    def reward_attr(self) -> str:
-        """Get the reward attribute that matches the scorer.
-
-        Returns
-        -------
-        reward_attr
-        """
-        if self._sign == 1:
-            return self.name
-        elif self._sign == -1:
-            return f'- {self.name}'
-        else:
-            raise NotImplementedError('Only supports sign = -1 or 1.')
-
-    @property
     def greater_is_better(self) -> bool:
         """Return whether the score is greater the better.
 
--- a/extra/setup.py
+++ b/extra/setup.py
@@ -49,7 +49,7 @@ requirements = [
     'gluoncv>=0.9.1,<1.0',
     'scipy>=1.3.3,<1.5.0',
     'graphviz<0.9.0,>=0.8.1',
-    'bokeh',
+    'bokeh<=2.2.2',
     'pandas>=1.0.0,<2.0',
     'scikit-learn>=0.22.0,<0.24',
     f'autogluon.core=={version}'
--- a/extra/setup.py
+++ b/extra/setup.py
@@ -49,7 +49,7 @@ requirements = [
     'gluoncv>=0.9.1,<1.0',
     'scipy>=1.3.3,<1.5.0',
     'graphviz<0.9.0,>=0.8.1',
-    'bokeh<=2.2.2',
+    'bokeh',
     'pandas>=1.0.0,<2.0',
     'scikit-learn>=0.22.0,<0.24',
     f'autogluon.core=={version}'
--- a/tabular/src/autogluon/tabular/models/ensemble/stacker_ensemble_model.py
+++ b/tabular/src/autogluon/tabular/models/ensemble/stacker_ensemble_model.py
@@ -118,13 +118,11 @@ class StackerEnsembleModel(BaggedEnsembleModel):
         super()._fit(X_train=X_train, y_train=y_train, k_fold=k_fold, k_fold_start=k_fold_start, k_fold_end=k_fold_end, n_repeats=n_repeats, n_repeat_start=n_repeat_start, time_limit=time_limit, **kwargs)
 
     def set_contexts(self, path_context):
-        path_root_orig = os.path.realpath(self.path_root)
+        path_root_orig = self.path_root
         super().set_contexts(path_context=path_context)
         for model, model_path in self.base_model_paths_dict.items():
-            model_path = os.path.realpath(model_path)
-            prefix = os.path.commonprefix([path_root_orig, model_path])
-            model_local_path = model_path[len(prefix):]
-            self.base_model_paths_dict[model] = os.path.join(self.path_root, model_local_path)
+            model_local_path = model_path.split(path_root_orig, 1)[1]
+            self.base_model_paths_dict[model] = self.path_root + model_local_path
 
     def set_stack_columns(self, stack_column_prefix_lst):
         if self.problem_type == MULTICLASS:
--- a/tabular/src/autogluon/tabular/models/ensemble/stacker_ensemble_model.py
+++ b/tabular/src/autogluon/tabular/models/ensemble/stacker_ensemble_model.py
@@ -124,7 +124,7 @@ class StackerEnsembleModel(BaggedEnsembleModel):
             model_path = os.path.realpath(model_path)
             prefix = os.path.commonprefix([path_root_orig, model_path])
             model_local_path = model_path[len(prefix):]
-            self.base_model_paths_dict[model] = self.path_root + model_local_path
+            self.base_model_paths_dict[model] = os.path.join(self.path_root, model_local_path)
 
     def set_stack_columns(self, stack_column_prefix_lst):
         if self.problem_type == MULTICLASS:
--- a/tabular/src/autogluon/tabular/models/ensemble/stacker_ensemble_model.py
+++ b/tabular/src/autogluon/tabular/models/ensemble/stacker_ensemble_model.py
@@ -118,10 +118,12 @@ class StackerEnsembleModel(BaggedEnsembleModel):
         super()._fit(X_train=X_train, y_train=y_train, k_fold=k_fold, k_fold_start=k_fold_start, k_fold_end=k_fold_end, n_repeats=n_repeats, n_repeat_start=n_repeat_start, time_limit=time_limit, **kwargs)
 
     def set_contexts(self, path_context):
-        path_root_orig = self.path_root
+        path_root_orig = os.path.realpath(self.path_root)
         super().set_contexts(path_context=path_context)
         for model, model_path in self.base_model_paths_dict.items():
-            model_local_path = model_path.split(path_root_orig, 1)[1]
+            model_path = os.path.realpath(model_path)
+            prefix = os.path.commonprefix([path_root_orig, model_path])
+            model_local_path = model_path[len(prefix):]
             self.base_model_paths_dict[model] = self.path_root + model_local_path
 
     def set_stack_columns(self, stack_column_prefix_lst):
--- a/tabular/src/autogluon/tabular/models/text_prediction/text_prediction_v1_model.py
+++ b/tabular/src/autogluon/tabular/models/text_prediction/text_prediction_v1_model.py
@@ -266,14 +266,15 @@ class TextPredictionV1Model(AbstractModel):
     def save(self, path: str = None, verbose=True) -> str:
         if path is None:
             path = self.path
-        model_path = os.path.join(path, self.model_file_name)
         text_nn_path = os.path.join(path, self.nn_model_name)
-        logger.log(15, f'Save Model Hyperparams to {model_path}.')
-        logger.log(15, f'Save Model Text NN weights to {text_nn_path}')
+        model_path = text_nn_path + os.sep + self.model_file_name
+        logger.log(15, f'Save Model Text NN weights and model hyperparameters'
+                       f' to {text_nn_path}. '
+                       f'The model hyper-parameters are saved as {model_path}')
         model = self.model
         self.model = None
         # save this AbstractModel object without NN weights
-        super().save(path=model_path, verbose=verbose)
+        super().save(path=text_nn_path + os.sep, verbose=verbose)
         model.save(text_nn_path)
         self.model = model
         return path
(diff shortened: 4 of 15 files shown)


szha:
Job PR-861-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-861/2/index.html

szha:
Job PR-861-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-861/3/index.html

szha:
Job PR-861-18 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-861/18/index.html

tabular/src/autogluon/tabular/models/ensemble/stacker_ensemble_model.py
```diff
@@ -118,11 +118,13 @@ def _fit(self, X_train, y_train, k_fold=5, k_fold_start=0, k_fold_end=None, n_re
...
-        path_root_orig = self.path_root
+        path_root_orig = os.path.realpath(self.path_root)
         super().set_contexts(path_context=path_context)
         for model, model_path in self.base_model_paths_dict.items():
-            model_local_path = model_path.split(path_root_orig, 1)[1]
-            self.base_model_paths_dict[model] = self.path_root + model_local_path
+            model_path = os.path.realpath(model_path)
+            prefix = os.path.commonprefix([path_root_orig, model_path])
+            model_local_path = model_path[len(prefix):]
+            self.base_model_paths_dict[model] = os.path.join(self.path_root, model_local_path)
```
sxjscience, line 127: @Innixma Would you take a look at this part? I first use `realpath` to ensure that the paths are expanded. Then, I use commonprefix to extract the common prefix so the local_path is `path[len(prefix):]`

Innixma: Why are you making this change? What does this fix?

Innixma: All changes should be reverted in this file, I believe the issue comes from the TextPrediction save and load methods, which I have code fixes for in a separate comment.

tabular/src/autogluon/tabular/models/text_prediction/text_prediction_v1_model.py
```diff
@@ -266,14 +258,15 @@ def _fit(self, X_train: pd.DataFrame, y_train: pd.Series,
     def save(self, path: str = None, verbose=True) -> str:
         if path is None:
             path = self.path
-        model_path = os.path.join(path, self.model_file_name)
         text_nn_path = os.path.join(path, self.nn_model_name)
-        logger.log(15, f'Save Model Hyperparams to {model_path}.')
-        logger.log(15, f'Save Model Text NN weights to {text_nn_path}')
+        model_path = text_nn_path + os.sep + self.model_file_name
```
Innixma, line 262: I don't think this is the right way to do saving and loading. Instead do the following (I have tested it works):

```
        model = self.model
        self.model = None
        # save this AbstractModel object without NN weights
        path = super().save(path=path, verbose=verbose)
        self.model = model

        text_nn_path = os.path.join(path, self.nn_model_name)
        model.save(text_nn_path)
        logger.log(15, f"\tSaved Text NN weights and model hyperparameters to '{text_nn_path}'.")

        return path
        

    @classmethod
    def load(cls, path: str, reset_paths=True, verbose=True):
        try:
            from autogluon.text.text_prediction.dataset import TabularDataset
            from autogluon.text.text_prediction.models.basic_v1 import BertForTextPredictionBasic
        except ImportError:
            raise ImportError(AG_TEXT_IMPORT_ERROR)

        model = super().load(path=path, reset_paths=reset_paths, verbose=verbose)
        model.model = BertForTextPredictionBasic.load(os.path.join(path, cls.nn_model_name))
        return model

```

As a general rule, all model.pkl files should be located in `{self.path}/model.pkl`.

extra/setup.py
```diff
@@ -49,7 +49,7 @@ def create_version_file():
     'gluoncv>=0.9.1,<1.0',
     'scipy>=1.3.3,<1.5.0',
     'graphviz<0.9.0,>=0.8.1',
-    'bokeh',
+    'bokeh<=2.2.2',
```
Innixma, line 52: Why?

tabular/tests/conftest.py
```diff
@@ -103,13 +104,11 @@ def fit_and_validate_dataset(dataset_name, fit_args, sample_size=1000, refit_ful
...
         label_column = dataset_info['label_column']
-        directory = directory_prefix + dataset_name + "/"
-        savedir = directory + 'AutogluonOutput/'
-
-        shutil.rmtree(savedir, ignore_errors=True)  # Delete AutoGluon output directory to ensure runs' information has been removed.
+        savedir = os.path.join(directory_prefix, dataset_name, f'AutogluonOutput_{uuid.uuid4()}')
         fit_args['label'] = label_column
         fit_args['output_directory'] = savedir
-        predictor = FitHelper.fit_dataset(train_data=train_data, fit_args=fit_args, sample_size=sample_size)
+        predictor = FitHelper.fit_dataset(train_data=train_data, fit_args=fit_args,
```
Innixma, line 110: nit: avoid edits like this to make the PR more readable.

tabular/tests/conftest.py
```diff
@@ -103,13 +104,11 @@ def fit_and_validate_dataset(dataset_name, fit_args, sample_size=1000, refit_ful
         directory_prefix = './datasets/'
         train_data, test_data, dataset_info = DatasetLoaderHelper.load_dataset(name=dataset_name, directory_prefix=directory_prefix)
         label_column = dataset_info['label_column']
-        directory = directory_prefix + dataset_name + "/"
-        savedir = directory + 'AutogluonOutput/'
-
-        shutil.rmtree(savedir, ignore_errors=True)  # Delete AutoGluon output directory to ensure runs' information has been removed.
+        savedir = os.path.join(directory_prefix, dataset_name, f'AutogluonOutput_{uuid.uuid4()}')
```
Innixma, line 107: Why is this necessary? A conflict could only occur if we were running tests in parallel.

Innixma (changes requested):

tabular/src/autogluon/tabular/models/text_prediction/text_prediction_v1_model.py
sxjscience: The problem of the original code is that it will be problematic if we are mixing relative path and absolute path. For example:
```python
model_path=./autogluon/SOME_MODEL
path_root_orig=/home/ubuntu/autogluon
```
The previous code will fail to recognize that `./autogluon` is actually equivalent to `/home/ubuntu/autogluon`. Thus, we should use `realpath` to normalize the path data.

tabular/tests/conftest.py
sxjscience: This is a good practice. Usually, we'd like to parallelize the pytest testing so we may want to ensure that each individual test is isolated.

Innixma:
Can we set a fixed version for autogluon-contrib-nlp? Otherwise we risk breaking old releases in future if we ever have a breaking change in autogluon-contrib-nlp.

tabular/src/autogluon/tabular/models/text_prediction/text_prediction_v1_model.py
sxjscience: For the `save()` + `load()` issue, I wasn't aware of this requirement (I think that all we need is consistency). I will change it to `{self.path}/file_name`. 

Also, I noticed that we are using `+` to join the directory and file: https://github.com/awslabs/autogluon/blob/947890e9f48ebd792407517af163d069cb150a27/tabular/src/autogluon/tabular/models/abstract/abstract_model.py#L376

This can be problematic if `path` does not end up with `os.sep`.

tabular/src/autogluon/tabular/models/ensemble/stacker_ensemble_model.py
sxjscience: save load works fine. I met this issue when I updated the savedir to `savedir = os.path.join(directory_prefix, dataset_name, f'AutogluonOutput_{uuid.uuid4()}')`.

tabular/src/autogluon/tabular/models/text_prediction/text_prediction_v1_model.py
Innixma: Please use the code I provided. It does everything correctly. I believe all the issues you are referring to are only present because of the current save/load methods in TextPrediction.

Innixma: Your concern about mixing relative and absolute path is only true if it is possible to happen. The current code makes this impossible. If you still think there is a failure case, please provide a working code snippet showcasing the issue.

sxjscience: I used the code that you shared with me initially but then noticed that we are always using `path + file_name` to join directory + filenames. Would you try to read again my comment about the issue (not this one, the one in ensemble model).

tabular/src/autogluon/tabular/models/ensemble/stacker_ensemble_model.py
Innixma: All paths are relative to the working directory in AutoGluon, and we don't support absolute directly paths. Are you saying that this adds support for absolute directory paths?

extra/setup.py
sxjscience: I met a dependency conflict problem reported by CI so I added this.

tabular/src/autogluon/tabular/models/text_prediction/text_prediction_v1_model.py
sxjscience: Okay, you are correct. I haven't seen issues after revising the save + load logic of the text_nn_model in tabular.

Innixma: Could you use my TextModel save/load code instead of the code currently in this PR? It is cleaner, as the current code is saving the model not in `{self.path}/model.pkl` but in `{self.path}/{text_nn_path}/model.pkl`, which we don't want

extra/setup.py
sxjscience: I turns out that this is not needed .

szha:
Job PR-861-21 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-861/21/index.html

szha:
Job PR-861-22 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-861/22/index.html

szha:
Job PR-861-23 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-861/23/index.html

tabular/tests/conftest.py
sxjscience: OK. (Tried to make the sentence shorter)

szha:
Job PR-861-25 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-861/25/index.html

tabular/src/autogluon/tabular/models/text_prediction/text_prediction_v1_model.py
```diff
@@ -239,19 +231,19 @@ def _fit(self, X_train: pd.DataFrame, y_train: pd.Series,
                 'grace_period', 10)
             scheduler_options['max_t'] = scheduler_options.get(
                 'max_t', 50)
+        if X_val is None:
+            holdout_frac = default_holdout_frac(len(X_train), True)
+            X_train, X_val = random_split_train_val(X_train, valid_ratio=holdout_frac)
```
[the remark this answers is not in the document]
sxjscience: Added.

szha:
Job PR-861-26 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-861/26/index.html

core/src/autogluon/core/metrics/__init__.py
```diff
@@ -386,7 +386,7 @@ def customized_log_loss(y_true, y_pred, eps=1e-15):
     if y_pred.ndim == 1:
         # First clip the y_pred which is also used in sklearn
         y_pred = np.clip(y_pred, eps, 1 - eps)
-        return (y_true * np.log(y_pred) + (1 - y_true) * np.log(1 - y_pred)).mean()
+        return - (y_true * np.log(y_pred) + (1 - y_true) * np.log(1 - y_pred)).mean()
```
sxjscience, line 389: @Innixma The root cause of the error should be this line.

sxjscience: Let me double check if it will fix the issue.

Innixma: Please test this on Tabular with the standard models, I suspect this is not the correct solution.

Innixma: Nevermind, this is indeed a bug.

szha:
Job PR-861-29 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-861/29/index.html

szha:
Job PR-861-34 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-861/34/index.html

szha:
Job PR-861-35 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-861/35/index.html

szha:
Job PR-861-40 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-861/40/index.html

szha:
Job PR-861-49 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-861/49/index.html

tabular/src/autogluon/tabular/models/text_prediction/text_prediction_v1_model.py
```diff
@@ -143,9 +132,7 @@ def _get_default_auxiliary_params(self) -> dict:
         default_auxiliary_params = super()._get_default_auxiliary_params()
         extra_auxiliary_params = dict(
             get_features_kwargs=dict(
-                valid_raw_types=[R_INT, R_FLOAT,
-                                 # R_CATEGORY,  # FIXME: Add R_CATEGORY features
-                                 R_OBJECT],
+                valid_raw_types=[R_INT, R_FLOAT, R_CATEGORY, R_OBJECT],
```
Innixma, line 135: last thing before merging: Can you comment out category until we resolve some of the issues with overfitting?

Innixma (approved):

szha:
Job PR-861-50 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-861/50/index.html

PR #877, opened 2021-01-20
Text: Improved time_limits adherence

Innixma:
*Issue #, if available:*

*Description of changes:*

If time_limits=50s, and model. takes ~40s to train fully

Prior:

```
model 1: train 40s
model 2: train 40s (30 seconds overtime)
model 3: train 40s (70 seconds overtime, occurs due to separate bug in scheduler)
model 4: doesn't train due to time_out being met. (never triggers because of time_out)
```

Now:

```
model 1: train 40s
model 2: train 10s
model 3: doesn't train due to train_function checking that no time is left. (0 seconds overtime, triggered due to scheduler bug)
model 4: doesn't train due to time_out being met. (never triggers because of time_out)
```

Additionally, fixed a bug that failed to record the last trained iteration result when time_limits ran out.
This fixes a case where low time_limits would cause a model crash due to `best_report_items` being `None`, and improves results in general when constrained by time_limits.

### DIFF
--- a/text/src/autogluon/text/text_prediction/models/basic_v1.py
+++ b/text/src/autogluon/text/text_prediction/models/basic_v1.py
@@ -239,10 +239,16 @@ def calculate_metric(scorer, ground_truth, predictions, problem_type):
 
 @use_np
 def train_function(args, reporter, train_df_path, tuning_df_path,
-                   time_limits, base_config, problem_types,
+                   time_limits, time_start, base_config, problem_types,
                    column_properties, label_columns, label_shapes,
                    log_metrics, stopping_metric, console_log,
                    ignore_warning=False):
+    if time_limits is not None:
+        start_train_tick = time.time()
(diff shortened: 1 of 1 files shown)


text/src/autogluon/text/text_prediction/models/basic_v1.py
```diff
@@ -619,6 +632,7 @@ def train(self, train_data, tuning_data, resource,
         train_fn = search_space_reg(functools.partial(train_function,
                                                       train_df_path=train_df_path,
                                                       time_limits=time_limits,
+                                                      time_start=start_tick,
```
sxjscience, line 635: I think we haven not initialized start_tick.

sxjscience: Okay, we have that initialized to time.time().

PR #878, opened 2021-01-20
[Scheduler][Bugfix] Added resources availability check before time limit is checked

gradientsky:
*Description of changes:*

Added  a quick check if resources are available before we check the time limit. This is required to prevent booking of the next job while we are waiting for a resource, which results in going over the time limit.

### DIFF
--- a/core/src/autogluon/core/scheduler/fifo.py
+++ b/core/src/autogluon/core/scheduler/fifo.py
@@ -7,6 +7,7 @@ import pickle
 import threading
 import time
 from collections import OrderedDict
+from time import sleep
 
 import numpy as np
 from tqdm.auto import tqdm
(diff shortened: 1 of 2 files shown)


szha:
Job PR-878-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-878/1/index.html

Innixma (approved):
Looks good!

core/src/autogluon/core/scheduler/fifo.py
```diff
@@ -253,6 +254,13 @@ def run(self, **kwargs):
         if time_out is not None:
             logger.info(f'Time out (secs) is {time_out}')
         for _ in tbar:
+            # Quick check if resources are available before we check the time limit
+            # This is to prevent booking next job while we are waiting for a resource, which
+            # results in going over the time limit
+            resources = DistributedResource(**self.resource)
+            while not FIFOScheduler.managers.check_availability(resources):
```
mseeger, line 261: Are we sure this carries no risks? What if this is waiting forever?

PR #882, opened 2021-01-23
[v0.1] align APIs for vision/ImagePredictor/ObjectDetector

zhreshold:
*Issue #, if available:*

*Description of changes:*
- API adjustments for vision part, to align with TabularPredictor
- GPU memory warning for too large custom batch sizes

Update: re-ran sanity fitting batch test with all d8 datasets over the weekend, all passed with reasonable validation scores for all presets with `epochs=1` 


vision/src/autogluon/vision/image_prediction/image_prediction.py
```diff
@@ -246,17 +272,17 @@ def predict_feature(self, x):
             raise RuntimeError('Classifier is not initialized, try `fit` first.')
         return self._classifier.predict_feature(x)
 
-    def evaluate(self, val_data):
+    def evaluate(self, tuning_data):
```
Innixma, line 275: recommend it just be called `data`, since this isn't necessarily used for tuning.

vision/src/autogluon/vision/object_detection/object_detection.py
```diff
@@ -13,32 +14,36 @@ class ObjectDetector(object):
...
+            time_limit=None,
             num_trials=1,
             hyperparameters=None,
             search_strategy='random',
             scheduler_options=None,
-            nthreads_per_trial=None,
-            ngpus_per_trial=None,
-            dist_ip_addrs=None,
-            verbosity=3):
+            num_cpus=None,
```
Innixma, line 45: same comments on num_cpus and num_gpus as in ImagePredictor

vision/src/autogluon/vision/object_detection/object_detection.py
```diff
@@ -13,32 +14,36 @@ class ObjectDetector(object):
...
 
     def fit(self,
             train_data,
-            val_data=None,
+            tuning_data=None,
             holdout_frac=0.1,
             random_state=None,
-            time_limit=12*60*60,
+            time_limit=None,
             num_trials=1,
```
Innixma, line 41: Same comment regarding hyperparameter_tune_kwargs

vision/src/autogluon/vision/object_detection/object_detection.py
```diff
@@ -78,37 +84,44 @@ def fit(self,
...
             return
 
         # new HPO task
+        if time_limit is None and num_trials is None:
+            raise ValueError('`time_limit` and `num_trials` can not be `None` at the same time, '
+                             'otherwise the training will not be terminated gracefully.')
         config={'log_dir': self._log_dir,
                 'num_trials': 99999 if num_trials is None else max(1, num_trials),
-                'time_limits': time_limit,
+                'time_limits': 2147483647 if time_limit is None else max(1, time_limit),
```
Innixma, line 105: It might be nicer if downstream None was handled instead of setting extremely high values. Not necessary for v0.1, more an observation.

vision/src/autogluon/vision/object_detection/object_detection.py
```diff
@@ -150,17 +163,17 @@ def predict(self, x):
             raise RuntimeError('Detector is not initialized, try `fit` first.')
         return self._detector.predict(x)
 
-    def evaluate(self, val_data):
+    def evaluate(self, tuning_data):
```
Innixma, line 166: ditto

vision/src/autogluon/vision/object_detection/object_detection.py
```diff
@@ -13,32 +14,36 @@ class ObjectDetector(object):
...
+        Higher levels correspond to more detailed print statements (you can set verbosity = 0 to suppress warnings). 
+        If using logging, you can alternatively control amount of information printed via logger.setLevel(L), 
+        where L ranges from 0 to 50 (Note: higher values of L correspond to fewer print statements, opposite of verbosity levels)
     """
     # Dataset is a subclass of `pd.DataFrame`, with `image` and `bbox` columns.
     Dataset = _ObjectDetection.Dataset
 
-    def __init__(self, log_dir=None):
-        self._log_dir = log_dir
+    def __init__(self, path=None, verbosity=2):
```
Innixma, line 29: will we ever have a reason to have different eval_metric or problem_type for this?

Innixma (commented):
Looks good, just a few minor tweaks to simplify the hyperparameter tuning arguments

szha:
Job PR-882-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-882/2/index.html

vision/src/autogluon/vision/object_detection/object_detection.py
Innixma: Also, why is config specifying 'time_limits'? Can we make it 'time_limit'?

vision/src/autogluon/vision/object_detection/object_detection.py
jwmueller: seems important to allow for other eval_metric in the future

vision/src/autogluon/vision/image_prediction/image_prediction.py
```diff
@@ -51,17 +65,19 @@ def fit(self,
             For more details of how to construct a dataset for image predictor, check out:
             `http://preview.d2l.ai/d8/main/image_classification/getting_started.html`.
             If a string is provided, will search for k8 built-in datasets.
-        val_data : pd.DataFrame or str, default = None
-            Training data, can be a dataframe like image dataset.
+        tuning_data : pd.DataFrame or str, default = None
+            Another dataset containing validation data reserved for tuning processes,
```
jwmueller, line 69: "reserved for tuning processes" -> "reserved for model selection and hyperparameter-tuning"

vision/src/autogluon/vision/object_detection/object_detection.py
```diff
@@ -48,17 +53,18 @@ def fit(self,
             For more details of how to construct a object detection dataset, please checkout:
             `http://preview.d2l.ai/d8/main/object_detection/getting_started.html`.
             If a string is provided, will search for k8 datasets.
-        val_data : pd.DataFrame or str, default = None
-            Training data, can be a dataframe like image dataset.
+        tuning_data : pd.DataFrame or str, default = None
+            Holdout tuning data for validation, can be a dataframe like image dataset.
```
jwmueller, line 57: "Holdout tuning data for validation" -> "Held-out validation data reserved for model selection and hyperparameter-tuning"

vision/src/autogluon/vision/object_detection/object_detection.py
```diff
@@ -48,17 +53,18 @@ def fit(self,
...
             If a string is provided, will search for k8 datasets.
-            If `None`, the validation dataset will be randomly split from `train_data`.
+            If `None`, the validation dataset will be randomly split from `train_data` according to `holdout_frac`.
         holdout_frac : float, default = 0.1
-            The random split ratio for `val_data` if `val_data==None`.
+            The random split ratio for `tuning_data` if `tuning_data==None`.
         random_state : numpy.random.state, default = None
-            The random_state for shuffling, only used if `val_data==None`.
+            The random_state for shuffling, only used if `tuning_data==None`.
```
jwmueller, line 63: "shuffling" -> "shuffling data"

vision/src/autogluon/vision/image_prediction/image_prediction.py
```diff
@@ -51,17 +65,19 @@ def fit(self,
...
+            can be a dataframe like image dataset.
             If a string is provided, will search for k8 datasets.
-            If `None`, the validation dataset will be randomly split from `train_data`.
+            If `None`, the validation dataset will be randomly split from `train_data` according to `holdout_frac`.
         holdout_frac : float, default = 0.1
-            The random split ratio for `val_data` if `val_data==None`.
+            The random split ratio for `tuning_data` if `tuning_data==None`.
         random_state : numpy.random.state, default = None
-            The random_state for shuffling, only used if `val_data==None`.
+            The random_state for shuffling, only used if `tuning_data==None`.
```
jwmueller, line 76: "shuffling" -> "shuffling data"

vision/src/autogluon/vision/image_prediction/image_prediction.py
```diff
@@ -144,13 +160,23 @@ def fit(self,
...
+        if num_gpus is not None:
+            config['ngpus_per_trial'] = num_gpus
         if isinstance(hyperparameters, dict):
+            if 'batch_size' in hyperparameters:
+                bs = hyperparameters['batch_size']
+                if num_gpus is not None and num_gpus > 1 and bs > 64:
+                    # using gpus, check batch size vs. available gpu memory
+                    free_gpu_memory = get_gpu_free_memory()
+                    if not free_gpu_memory:
+                        warnings.warn('Unable to detect free GPU memory, we are unable to verify your batch_size setting.')
```
jwmueller, line 174: "your batch_size setting" -> "whether your data mini-batches will fit on the GPU for the specified batch_size"

vision/src/autogluon/vision/image_prediction/image_prediction.py
```diff
@@ -144,13 +160,23 @@ def fit(self,
...
+                    elif len(free_gpu_memory) < num_gpus:
+                        warnings.warn(f'Detected GPU memory for {len(free_gpu_memory)} gpus but {num_gpus} is requested.')
+                    elif sum(free_gpu_memory[:num_gpus]) / bs < 128:
+                        warnings.warn(f'batch-size: {bs} is potentially larger than what your gpus can support ' +
```
jwmueller, line 178: "batch-size" -> "batch_size"

jwmueller: add to warning: "Try reducing "batch_size" if you encounter memory issues"

vision/src/autogluon/vision/object_detection/object_detection.py
zhreshold: this internal naming and behavior change will need to be addressed in gluoncv and it may not be necessary for 0.1 release

szha:
Job PR-882-6 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-882/6/index.html

vision/src/autogluon/vision/detector/detector.py
```diff
@@ -0,0 +1,270 @@
...
+                logging.getLogger(logger_name).propagate = False
+        task = _ObjectDetection(config=config)
+        task._logger.setLevel(log_level)
+        task._logger.propagate = True
+        self._detector = task.fit(train_data, tuning_data, 1 - holdout_frac, random_state)
+        self._detector._logger.setLevel(log_level)
+        self._detector._logger.propagate = True
+        self._fit_summary = task.fit_summary()
+
+    def _validate_kwargs(self, kwargs):
```
Innixma, line 165: consider using `from autogluon.core.scheduler.scheduler_factory import scheduler_factory` to simplify this code and align with Tabular.

Innixma, line 165: Perhaps sanity check that no unknown kwargs were passed.

vision/src/autogluon/vision/detector/detector.py
```diff
@@ -0,0 +1,270 @@
...
+                raise ValueError(f"Insufficient detected # gpus {detected_gpu} vs requested {kwargs['ngpus_per_trial']}")
+        # tune kwargs
+        hpo_tune_args = kwargs.get('hyperparameter_tune_kwargs', {})
+        hpo_tune_args['num_trials'] = hpo_tune_args.get('num_trials', 1)
+        hpo_tune_args['search_strategy'] = hpo_tune_args.get('search_strategy', 'random')
+        if not hpo_tune_args['search_strategy'] in ('random', 'bayesopt', 'grid'):
+            raise ValueError(f"Invalid search strategy: {hpo_tune_args['search_strategy']}, supported: ('random', 'bayesopt', 'grid')")
+        hpo_tune_args['max_reward'] = hpo_tune_args.get('max_reward', 0.9)
+        if hpo_tune_args['max_reward'] < 0:
+            raise ValueError(f"Expected `max_reward` to be a positive float number between 0 and 1.0, given hpo_tune_args['max_reward']")
```
Innixma, line 185: This does not print out the value of max_reward. Missing `{}`. Also, the if statement doesn't check for `> 1` as the error states.

vision/src/autogluon/vision/predictor/predictor.py
```diff
@@ -0,0 +1,418 @@
...
+                Trainer learning rate for optimization process.
+            You can get the list of accepted hyperparameters in `config.yaml` saved by this predictor.
+        **kwargs :
+            holdout_frac : float, default = 0.1
+                The random split ratio for `tuning_data` if `tuning_data==None`.
+            random_state : int, default = None
+                The random_state(seed) for shuffling data, only used if `tuning_data==None`.
+                Note that the `random_state` only affect the splitting process, not model training.
+                If not specified(None), will leave the original random sampling intact.
+            nthreads_per_trial : int, default = (# cpu cores)
```
Innixma, line 94: same comment as in ObjectDetection

vision/src/autogluon/vision/predictor/predictor.py
```diff
@@ -0,0 +1,418 @@
...
+                raise ValueError(f"Insufficient detected # gpus {detected_gpu} vs requested {kwargs['ngpus_per_trial']}")
+        # tune kwargs
+        hpo_tune_args = kwargs.get('hyperparameter_tune_kwargs', {})
+        hpo_tune_args['num_trials'] = hpo_tune_args.get('num_trials', 1)
+        hpo_tune_args['search_strategy'] = hpo_tune_args.get('search_strategy', 'random')
+        if not hpo_tune_args['search_strategy'] in ('random', 'bayesopt', 'grid'):
+            raise ValueError(f"Invalid search strategy: {hpo_tune_args['search_strategy']}, supported: ('random', 'bayesopt', 'grid')")
+        hpo_tune_args['max_reward'] = hpo_tune_args.get('max_reward', 0.95)
+        if hpo_tune_args['max_reward'] < 0:
+            raise ValueError(f"Expected `max_reward` to be a positive float number between 0 and 1.0, given hpo_tune_args['max_reward']")
```
Innixma, line 246: Ditto comment: Missing `{}` to print actual value

vision/src/autogluon/vision/predictor/predictor.py
```diff
@@ -0,0 +1,418 @@
...
+            logging.getLogger("ImageClassificationEstimator").setLevel(log_level)
+        task = _ImageClassification(config=config)
+        task._logger.setLevel(log_level)
+        task._logger.propagate = True
+        self._classifier = task.fit(train_data, tuning_data, 1 - holdout_frac, random_state)
+        self._classifier._logger.setLevel(log_level)
+        self._classifier._logger.propagate = True
+        self._fit_summary = task.fit_summary()
+
+    def _validate_kwargs(self, kwargs):
```
Innixma, line 226: ditto same comment: Consider using `scheduler_factory`

vision/src/autogluon/vision/predictor/predictor.py
```diff
@@ -0,0 +1,418 @@
...
+        ----------
+        file_name : str
+            The file name of saved copy.
+
+        """
+        with open(file_name, 'wb') as fid:
+            pickle.dump(self, fid)
+
+    @classmethod
+    def load(cls, file_name, verbosity=2):
```
Innixma, line 375: `file_name` -> `path`, ideally it should be the same path value as specified by user or available via `predictor.path`.

vision/src/autogluon/vision/detector/detector.py
```diff
@@ -0,0 +1,270 @@
...
+                Trainer learning rate for optimization process.
+            You can get the list of accepted hyperparameters in `config.yaml` saved by this predictor.
+        **kwargs :
+            holdout_frac : float, default = 0.1
+                The random split ratio for `tuning_data` if `tuning_data==None`.
+            random_state : int, default = None
+                The random_state(seed) for shuffling data, only used if `tuning_data==None`.
+                Note that the `random_state` only affect the splitting process, not model training.
+                If not specified(None), will leave the original random sampling intact.
+            nthreads_per_trial : int, default = (# cpu cores)
```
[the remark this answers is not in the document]
zhreshold: introducing another `resource` concept is adding additional complexity IMO, however, the value `nthreads_per_trial` or `num_cpus` is useful even with HPO turned off.

vision/src/autogluon/vision/detector/detector.py
```diff
@@ -0,0 +1,270 @@
...
+                Number of CPU threads for each trial, if `None`, will detect the # cores on current instance.
+            ngpus_per_trial : int, default = (# gpus)
+                Number of GPUs to use for each trial, if `None`, will detect the # gpus on current instance.
+            hyperparameter_tune_kwargs: dict, default = None
+                num_trials : int, default = 1
+                    The number of HPO trials when `time_limit` is None.
```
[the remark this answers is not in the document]
zhreshold: the logic is to stop HPO when either `time_limit` or `num_trials` reaches the limit, whichever comes first. What do you suggest for the docstring?

vision/src/autogluon/vision/detector/detector.py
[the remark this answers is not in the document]
Innixma: Why is it valuable if HPO is turned off? That isn't what is communicated by the documentation:

```
Number of CPU threads for each trial, if `None`, will detect the # cores on current instance.
```

Are you referring to the single model training if HPO is turned off? If that is the case, shouldn't it be specified in the model hyperparameters?

vision/src/autogluon/vision/detector/detector.py
```diff
@@ -0,0 +1,270 @@
...
+            ngpus_per_trial : int, default = (# gpus)
+                Number of GPUs to use for each trial, if `None`, will detect the # gpus on current instance.
+            hyperparameter_tune_kwargs: dict, default = None
+                num_trials : int, default = 1
+                    The number of HPO trials when `time_limit` is None. 
+                search_strategy : str, default = 'random'
+                    Searcher strategy for HPO, 'random' by default.
+                    Options include: ‘random’ (random search), ‘bayesopt’ (Gaussian process Bayesian optimization),
+                    ‘grid’ (grid search).
+                max_reward : float, default = 0.9
```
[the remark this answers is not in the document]
zhreshold: Yes, theoretically it's almost impossible to reach 0.9 mean average precision for bonding boxes. Even if a model does, the prediction difference between 0.9 mAP and 1.0 mAP is neglected. 

My input of 0.9 here is to avoid the searcher desperately exploring the best configuration even if the result is already saturated(good for the average users, bad for competition?) What I can do is to put 1.0 back as default, or print a message to tell if the `max_reward` is met but you can possibly improve it.

@Innixma What's you suggestion?

vision/src/autogluon/vision/detector/detector.py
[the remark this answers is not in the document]
Innixma:
```
The maximum amount of HPO trials to perform, without going over the specified `time_limit`.
```

Additionally, I'd recommend adding a link to FIFOScheduler documentation in the overall hyperparameter_tune_kwargs docs.

```
For valid dictionary keys, refer to :class:`autogluon.core.scheduler.FIFOScheduler` documentation.
    The 'searcher' key is required when providing a dict.
```

Example, scroll to kwargs hyperparameter_tune_kwargs to see link in doc: http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-889/8/api/autogluon.task.html#autogluon.tabular.TabularPredictor.fit

vision/src/autogluon/vision/detector/detector.py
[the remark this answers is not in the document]
zhreshold: Adding `presets` though, where `best_quality` mode will override `max_reward` to 1 though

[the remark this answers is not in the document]
Innixma: I'll leave this to your decision since you are more familiar with what 0.9 mAP means, but I would suggest that so long as the user specified a long enough time_limit, there is no reason to stop training when improvements in score could still be gained. Perhaps with a higher quality preset it can be set to 1?

vision/src/autogluon/vision/predictor/predictor.py
```diff
@@ -0,0 +1,418 @@
...
+            ngpus_per_trial : int, default = (# gpus)
+                Number of GPUs to use for each trial, if `None`, will detect the # gpus on current instance.
+            hyperparameter_tune_kwargs: dict, default = None
+                num_trials : int, default = 1
+                    The number of HPO trials when `time_limit` is None. 
+                search_strategy : str, default = 'random'
+                    Searcher strategy for HPO, 'random' by default.
+                    Options include: ‘random’ (random search), ‘bayesopt’ (Gaussian process Bayesian optimization),
+                    ‘grid’ (grid search).
+                max_reward : float, default = 0.95
```
Innixma, line 105: I feel that this should default to 1 or effectively not be set. I think there is little reason to stop HPO early when achieving 95% accuracy, especially for datasets that are imbalanced and it is trivial to get 95% accuracy without much meaningful benefits.

Innixma: Another reason for not setting this at all would be if in future we had `log_loss` as an evaluation metric. A max_reward of 0.95 wouldn't make sense in this case.

vision/src/autogluon/vision/configs/presets_configs.py
```diff
@@ -0,0 +1,184 @@
...
+            return f(*gargs, **gkwargs)
+        return _call
+    return _unpack_inner
+
+
+# Dictionary of preset fit() parameter configurations for ImagePredictor.
+preset_image_predictor = dict(
+    # Best predictive accuracy with little consideration to inference time or model size. Achieve even better results by specifying a large time_limit value.
+    # Recommended for applications that benefit from the best possible model accuracy.
+    best_quality={
```
Innixma, line 22: This will end very poorly if someone does not specify a time_limit. We should ensure that a time limit is set for this, otherwise it will run 1024 trials.

Innixma: Same comments for all of the presets. Any time num_trials is specified automatically, we probably want to ensure time_limit is specified by user

vision/src/autogluon/vision/configs/presets_configs.py
```diff
@@ -0,0 +1,184 @@
...
+        'hyperparameters': {
+            'model': 'resnet50_v1b',
+            'lr': 0.01,
+            'batch_size': 64,
+            'epochs': 50
+            },
+        'hyperparameter_tune_kwargs': {
+            'num_trials': 8,
+            'search_strategy': 'random',
+            'max_reward': 0.85
```
Innixma, line 64: Why 0.85? This seems extremely low. I generally think we should avoid any specification of max_reward, since an acceptable accuracy differs drastically between problems.

vision/src/autogluon/vision/configs/presets_configs.py
```diff
@@ -0,0 +1,184 @@
...
+        'hyperparameter_tune_kwargs': {
+            'num_trials': 32,
+            'search_strategy': 'bayesopt',
+            'max_reward': 0.95
+        }
+    },
+)
+
+# Dictionary of preset fit() parameter configurations for ObjectDetector.
+preset_object_detector = dict(
```
Innixma, line 86: same comments here as in image predictor presets

vision/src/autogluon/vision/configs/presets_configs.py
```diff
@@ -0,0 +1,184 @@
...
+
+)
+
+_PRESET_DICTS = {
+    'image_predictor': preset_image_predictor, 
+    'object_detector': preset_object_detector,
+}
+
+
+def set_presets(preset_name, *args, **kwargs):
```
Innixma, line 161: Do you think tabular's preset functionality is general enough to move to core so we don't have to copy paste these functions?

zhreshold: moving to core is nice, see my modifications, I prefer a named tuple or mapping so that I won't need to import specific preset dict from ag.core

vision/src/autogluon/vision/configs/presets_configs.py
zhreshold: here's my confusing part, if running things without a wall clock time limit is scary, should we at least safeguard the default time_limit with a reasonable value, say 1 day or 3 days?

Innixma: Perhaps we can set default to 4 hours in the presets and warn the user if they did not set it themselves. Whatever solution we go with, we need to ensure that the user doesn't unknowingly start a job by specifying a preset that will take weeks+ to finish when they never touched time_limit.

I would personally be in favor of raising an exception if the user specified a preset but did not specify time_limit, or otherwise set num_trials=1 if they only specified a preset but no time limit. Alternatively, we could set time_limit default value in the preset itself (and user can freely override if they wish).

Example:

```
    best_quality={
        'hyperparameters': {
            'model': Categorical('resnet50_v1b', 'resnet101_v1d', 'resnest200'),
            'lr': Real(1e-2, 1e-6, log=True),
            'epochs': 200
            },
        'hyperparameter_tune_kwargs': {
            'num_trials': 1024,
            'search_strategy': 'bayesopt',
            'max_reward': 1.0
        },
        'time_limit': 86400,
    },
```

szha:
Job PR-882-9 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-882/9/index.html

szha:
Job PR-882-10 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-882/10/index.html

mxnet/src/autogluon/mxnet/task/classifier.py
```diff
@@ -35,9 +35,8 @@ class Classifier(BasePredictor):
     >>>                        test_path='data/test')
     >>> classifier = task.fit(dataset,
```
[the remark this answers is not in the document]
jwmueller: All the documentation should drop the usage of the term `task` since ImagePrediction will be dropped and ImagePredictor is not a task.

vision/src/autogluon/vision/predictor/predictor.py
jwmueller: Agreed I think this should default to infinity and we should not assume metrics will always lie in [0,1].

vision/src/autogluon/vision/configs/presets_configs.py
jwmueller: Alternative suggestion:

- Do not set time-limits by default (since users expect things to take longer for larger datasets anyway).

- Set num_trials default (probably much lower than 1024 though. Perhaps for the different presets we could have: num_trials = 1 (for medium), 10 (for good), 100 (for best).

- Whenever user specifies time_limit, this overrides num_trials. So we don't run for min{time_limit, num_trials}, but instead just run for whatever time_limit the user specified (which may involve more trials than the default num_trials). If user specifies both time_limit and num_trials, then throw Exception.

szha:
Job PR-882-12 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-882/12/index.html

szha:
Job PR-882-15 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-882/15/index.html

szha:
Job PR-882-16 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-882/16/index.html

szha:
Job PR-882-17 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-882/17/index.html

szha:
Job PR-882-19 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-882/19/index.html

szha:
Job PR-882-21 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-882/21/index.html

vision/src/autogluon/vision/detector/detector.py
```diff
@@ -0,0 +1,379 @@
...
+            The input data, can be str(filepath), pd.DataFrame with 'image' column, or raw ndarray input.
+        as_pandas : bool, default = True
+            Whether to return the output as a pandas object (True) or list of numpy array(s) (False).
+            Pandas object is a DataFrame.
+
+        Returns
+        -------
+
+        pd.DataFrame
+            The returned dataframe will contain (`pred_score`, `pred_bbox`, `pred_id`).
```
Innixma, line 308: Should `pred_id` just be the index? Does it need to be its own column?

Innixma (commented):
Once PR has been rebased I'll be happy to approve. I think some of the comments can be discussed after PR is merged.

vision/src/autogluon/vision/detector/detector.py
zhreshold: this is the id of each predicted object

szha:
Job PR-882-22 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-882/22/index.html

Innixma (approved):

Issue #899, opened 2021-02-02
Add support for token pre-selection


sflender:
Add option to pre-filter the tokens generated with CountVectorizer, for instance with sklearn's SelectKBest, and only keep the K tokens that are most correlated with the label column. This should improve both training speed and model performance.

piece of the code to be changed:
https://github.com/awslabs/autogluon/blob/ff6830416e88affa75d5550e9a5875708223fc5a/features/src/autogluon/features/generators/text_ngram.py

PR #907, opened 2021-02-04
NLP token prefilter

sflender:
*Issue #, if available:* 
https://github.com/awslabs/autogluon/issues/899

*Description of changes:*
add option to pre-filter tokens according to correlation with the label: if `prefilter_tokens=True`, then only keep the `prefilter_token_count` most correlated tokens.

Use this option by parsing `AutoMLPipelineFeatureGenerator(prefilter_tokens=True, prefilter_token_count=10)`
into .fit().

Note that in the current form this will not work for multi-class classification problems.

### DIFF
--- a/features/src/autogluon/features/generators/auto_ml_pipeline.py
+++ b/features/src/autogluon/features/generators/auto_ml_pipeline.py
@@ -38,7 +38,7 @@ class AutoMLPipelineFeatureGenerator(PipelineFeatureGenerator):
         Appends TextSpecialFeatureGenerator() to the generator group.
     enable_text_ngram_features : bool, default True
         Whether to use 'object' features identified as 'text' features to generate 'text_ngram' features.
-        Appends TextNgramFeatureGenerator(vectorizer=vectorizer) to the generator group.
+        Appends TextNgramFeatureGenerator(vectorizer=vectorizer, text_ngram_params) to the generator group. See text_ngram.py for valid parameters.
     enable_raw_text_features : bool, default False
         Whether to use the raw text features. The generated raw text features will end up with '_raw_text' suffix.
         For example, 'sentence' --> 'sentence_raw_text'
@@ -102,6 +102,6 @@ class AutoMLPipelineFeatureGenerator(PipelineFeatureGenerator):
         if self.enable_text_special_features:
             generator_group.append(TextSpecialFeatureGenerator())
         if self.enable_text_ngram_features:
-            generator_group.append(TextNgramFeatureGenerator(**self.text_ngram_params))
+            generator_group.append(TextNgramFeatureGenerator(vectorizer=vectorizer, **self.text_ngram_params))
         generators = [generator_group]
         return generators
--- a/features/src/autogluon/features/generators/auto_ml_pipeline.py
+++ b/features/src/autogluon/features/generators/auto_ml_pipeline.py
@@ -69,7 +69,7 @@ class AutoMLPipelineFeatureGenerator(PipelineFeatureGenerator):
     def __init__(self, enable_numeric_features=True, enable_categorical_features=True,
                  enable_datetime_features=True,
                  enable_text_special_features=True, enable_text_ngram_features=True,
-                 enable_raw_text_features=False, vectorizer=None, **kwargs):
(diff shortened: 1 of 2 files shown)


szha:
Job PR-907-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-907/1/index.html

features/src/autogluon/features/generators/text_ngram.py
```diff
@@ -73,6 +84,8 @@ def _transform(self, X: DataFrame) -> DataFrame:
             return DataFrame(index=X.index)
         try:
             X_out = self._generate_ngrams(X=X)
+            if self.prefilter_tokens:
```
Innixma, line 87: add `# TODO: Optimize for inference`

This can be optimized heavily to be much faster, but is more complex than necessary for initial implementation.
The idea is to update the vectorizer to only produce the chosen set of ngrams, removing the need for any change at all to the `transform` method.

features/src/autogluon/features/generators/text_ngram.py
```diff
@@ -60,9 +61,19 @@ def __init__(self, vectorizer=None, vectorizer_strategy='combined', max_memory_r
             raise ValueError(f"vectorizer_strategy must be one of {['combined', 'separate', 'both']}, but value is: {vectorizer_strategy}")
         self.vectorizer_strategy = vectorizer_strategy
         self.vectorizer_features = None
+        self.prefilter_tokens = prefilter_tokens
+        self.prefilter_token_count = prefilter_token_count
+        self.token_mask = None
 
-    def _fit_transform(self, X: DataFrame, **kwargs) -> (DataFrame, dict):
+    def _fit_transform(self, X: DataFrame, y: Series, **kwargs) -> (DataFrame, dict):
```
Innixma, line 68: Please add a unit test showcasing these changes that also tests edge cases.

features/src/autogluon/features/generators/text_ngram.py
sflender: agree. I'll add the comment for now.

features/src/autogluon/features/generators/text_ngram.py
sflender: need some help with this. what exactly should we test for?

Innixma: Tests should be added to this file: https://github.com/awslabs/autogluon/blob/master/tabular/tests/unittests/features/generators/test_text_ngram.py

Essentially, we want to test that using the functionality works in the way we expect it to. Edge cases include if y contains NaN values in rows (A valid situation if we passed concatenated train and unlabeled data to increase number of ngrams we generate) and k > ngram count.

features/src/autogluon/features/generators/auto_ml_pipeline.py
```diff
@@ -69,7 +69,7 @@ class AutoMLPipelineFeatureGenerator(PipelineFeatureGenerator):
     def __init__(self, enable_numeric_features=True, enable_categorical_features=True,
                  enable_datetime_features=True,
                  enable_text_special_features=True, enable_text_ngram_features=True,
-                 enable_raw_text_features=False, vectorizer=None, **kwargs):
+                 enable_raw_text_features=False, vectorizer=None, prefilter_tokens=False, prefilter_token_count=10, **kwargs):
```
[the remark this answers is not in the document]
sflender: agree, I changed this in the new revision.

sflender:
perhaps checks failed because vectorizer=vectorizer needs to parsed in this line: https://github.com/awslabs/autogluon/pull/907/files#diff-bbbc40d9dd88e368a7208d22b898e303b8383d416de6ca6812532c784610ed43R105

Innixma:
Yes, you'll want to check if vectorizer is not None and if so then add as key to text_ngram_params.

```
if text_ngram_params is None:
    text_ngram_params = {}
if vectorizer is not None:
    text_ngram_params['vectorizer'] = vectorizer
```

szha:
Job PR-907-5 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-907/5/index.html

Innixma (approved):
Looks good, thanks for the contribution!

PR #908, opened 2021-02-04
[CI] Updates to build scripts to use clean venvs for each build

gradientsky:
*Description of changes:*
- rework of build scripts to use clean venv for each build - this should improve libraries version compatibility checks so old dependencies are not masking such problems.
- version updates to ConfigSpace/numpy/pyarrow so they are compatible
- `batch_size` set to 8 for object detection to prevent GPU memory issues

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,36 +1,58 @@
 max_time = 180
 
+setup_pip_venv = """
+    rm -rf venv
+    conda list
+
+    python3 -m venv venv
+    source venv/bin/activate
+    python3 -m pip install -U pip
+    python3 -m pip install -U setuptools wheel
+
+    python3 -m pip install 'graphviz'
+    python3 -m pip install 'jupyter-sphinx>=0.2.2'
+    python3 -m pip install 'portalocker'
+    python3 -m pip install 'scipy>=1.3.3,<1.5.0'
+    python3 -m pip install 'nose'
+    python3 -m pip install 'docutils'
+    python3 -m pip install 'mu-notedown'
+    python3 -m pip install 'flake8'
+    python3 -m pip install 'awscli>=1.18.140'
+
+    export MPLBACKEND=Agg
+"""
+
+setup_mxnet_gpu = """
+    python3 -m pip install mxnet-cu101==1.7.0
+    export LD_LIBRARY_PATH=/usr/local/cuda-10.1/lib64
+    export MXNET_CUDNN_AUTOTUNE_DEFAULT=0
+"""
+
+cleanup_venv = """
+    deactivate
+    rm -rf venv
+"""
+
 stage("Unit Test") {
   parallel 'core': {
     node('linux-cpu') {
-      ws('workspace/autogluon-core-py3') {
+      ws('workspace/autogluon-core-py3-v3') {
         timeout(time: max_time, unit: 'MINUTES') {
           checkout scm
           VISIBLE_GPU=env.EXECUTOR_NUMBER.toInteger() % 8
           sh """#!/bin/bash
           set -ex
-          conda env update -n autogluon_core_py3 -f docs/build.yml
-          conda activate autogluon_core_py3
+          conda env update -n autogluon-core-py3-v3 -f docs/build.yml
+          conda activate autogluon-core-py3-v3
           conda list
-          export CUDA_VISIBLE_DEVICES=${VISIBLE_GPU}
+          ${setup_pip_venv}
+          python3 -m pip install 'mxnet==1.7.0.*'
           env
-          export LD_LIBRARY_PATH=/usr/local/cuda-10.1/lib64
-          export MPLBACKEND=Agg
-          export MXNET_CUDNN_AUTOTUNE_DEFAULT=0
-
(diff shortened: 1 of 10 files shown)


core/setup.py
```diff
@@ -53,7 +53,7 @@ def create_version_file():
     'paramiko>=2.4',
     'dask>=2.6.0',
     'distributed>=2.6.0',
-    'ConfigSpace<=0.4.16',  # TODO v0.1: Upgrade?
+    'ConfigSpace<=0.4.17',
```
Innixma, line 56: Should this be `>=`?

text/setup.py
```diff
@@ -55,7 +55,7 @@ def create_version_file():
 ]
 
 text_requirements = [
-    'pyarrow<=1.0.0',
+    'pyarrow<=3.0.0',
```
Innixma, line 58: Do we really want `<=` here?

core/setup.py
```diff
@@ -43,7 +43,7 @@ def create_version_file():
 MIN_PYTHON_VERSION = '>=3.6.*'  # TODO v0.1: Also specify max version? (We don't support 3.8)
 
 requirements = [
-    'numpy>=1.16.0,<1.20.0',
+    'numpy>=1.16.0',
```
Innixma, line 46: Will this still work if numpy is 1.16.0?

Jenkinsfile
```diff
@@ -1,150 +1,140 @@
 max_time = 180
 
+setup_pip_venv = """
+    rm -rf venv
+    conda list
+
+    python3 -m venv venv
+    source venv/bin/activate
+    python3 -m pip install -U pip setuptools wheel
```
Innixma, line 9: might need to separate the install pip line with the others, since pip version dictates the way setuptools and wheel are installed:

```
python3 -m pip install -U pip
python3 -m pip install -U setuptools wheel
```

Jenkinsfile
```diff
@@ -1,150 +1,140 @@
...
+cleanup_venv = """
+    deactivate
+    rm -rf venv
+"""
+
 stage("Unit Test") {
   parallel 'core': {
     node('linux-cpu') {
-      ws('workspace/autogluon-core-py3') {
+      ws('workspace/autogluon-core-py3-v3') {
```
Innixma, line 38: Why `-v3`?

Jenkinsfile
```diff
@@ -1,150 +1,140 @@
...
         timeout(time: max_time, unit: 'MINUTES') {
           checkout scm
           VISIBLE_GPU=env.EXECUTOR_NUMBER.toInteger() % 8
```
Innixma, line 41: What is the purpose of this line now?

Jenkinsfile
```diff
@@ -1,150 +1,140 @@
...
           sh """#!/bin/bash
           set -ex
-          conda env update -n autogluon_core_py3 -f docs/build.yml
-          conda activate autogluon_core_py3
+          conda env update -n autogluon-core-py3-v3 -f docs/build.yml
+          conda activate autogluon-core-py3-v3
           conda list
-          export CUDA_VISIBLE_DEVICES=${VISIBLE_GPU}
+          ${setup_pip_venv}
+          python3 -m pip install 'mxnet==1.7.0.*'
```
Innixma, line 48: Why are we using CPU mxnet?

core/setup.py
gradientsky: I think we need to fix it to 1.20.x, 1.16 has compatibility issues with a combination of `ConfigSpace` and `pyarrow`

Jenkinsfile
gradientsky: because this is `linux-cpu` node (2 tasks run on it)

Jenkinsfile
gradientsky: just some unique namespace not used before (-v1/v2 are already used)

Jenkinsfile
gradientsky: good point -- updated to be on a separate line

tabular/tests/unittests/features/generators/test_category.py
```diff
@@ -36,15 +36,15 @@ def test_category_feature_generator(generator_helper, data_helper):
         [0, 1, 0, 3, 3, 3, 2, np.nan, np.nan],
         [0, np.nan, 0, 1, 1, 1, np.nan, np.nan, np.nan],
         [np.nan, np.nan, np.nan, 0, 0, 0, np.nan, np.nan, np.nan],
-        [2, 1, 2, 3, 3, 3, 0, np.nan, np.nan],
+        [2, 0, 2, 3, 3, 3, 1, np.nan, np.nan],
```
Innixma, line 39: What caused this?

szha:
Job PR-908-5 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-908/5/index.html

Innixma (approved):

tabular/tests/unittests/features/generators/test_category.py
gradientsky: Likely pandas upgrade

PR #912, opened 2021-02-08
Tabular: HPO bag fix

Innixma:
*Issue #, if available:*

*Description of changes:*

- [x] During bagging, first fold of model is called `trial*_fold_0/` while other folds are called `trial*_F2/` etc. This has been fixed.
- [x] AutoGluon was giving far too much time to HPO when bagging due to a bug (9x longer than intended). This has been fixed.
- [x] Trainer became very large during HPO sometimes, such as >5GB. This caused memory errors. Trainer should never be >1 MB in general. This was caused by `trainer.hpo_results` being extremely large after many HPO trials (>>5 GB). I have removed this object for now, fixing the issue.
- [x] Simplified model name logic such that `_FULL` is always at the end of a model name, rather than inserted before the model level. This makes the model name construction logic much easier to understand.
- [x] Improved naming of fold models. Now they are simply `SxFy`, where x is the repeat set, and y is the fold in the repeat set. This reduces redundant naming and separates the naming logic of the child models from the naming logic of the BaggedEnsemble.
- [x] Other minor fixes and improvements

### DIFF
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -609,7 +609,7 @@ class AbstractModel:
         hpo_model_performances = {}
         for trial in sorted(hpo_results['trial_info'].keys()):
             # TODO: ignore models which were killed early by scheduler (eg. in Hyperband). How to ID these?
-            file_id = "trial_" + str(trial)  # unique identifier to files from this trial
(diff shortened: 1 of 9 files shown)


gradientsky (approved):

szha:
Job PR-912-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-912/2/index.html

PR #919, opened 2021-02-11
Tabular: Fixed rare crash on leaderboard

Innixma:
*Issue #, if available:*

*Description of changes:*

- Fixes crash on leaderboard when multiclass problem is converted to binary internally, label cleaner was incorrectly using the internal problem type instead of the external problem type when converting internal pred_proba to external pred in leaderboard.
- Added support in save_pd to save without needing a directory. Aka: `save_pd.save('file.csv')` did not work before, but now does, because before it was trying to create a directory at `''`, which wasn't valid.

### DIFF
--- a/core/src/autogluon/core/data/label_cleaner.py
+++ b/core/src/autogluon/core/data/label_cleaner.py
@@ -126,7 +126,7 @@ class LabelCleanerMulticlass(LabelCleaner):
         else:
             y_transformed = y
         if as_pred:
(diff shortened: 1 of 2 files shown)


szha:
Job PR-919-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-919/1/index.html

gradientsky (approved):

PR #924, opened 2021-02-13
Tabular: Refined HPO code / edge case handling

Innixma:
*Issue #, if available:*

*Description of changes:*

- Moved HPO artifacts when bagging under the bagged model directory so they never conflict with future models, and the resulting directory layout is less cluttered.
- Added extra edge case handling for banned model names to avoid `.fit_extra` calls overwriting model names if both original `fit` and `fit_extra` used HPO with the same model types. Now this will never happen.
- If resource not specified in HPO kwargs, it will no longer override resources set in the original model to `'auto'`, but instead use the original model's values (could be values other than `'auto'`).

### DIFF
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -519,7 +519,15 @@ class AbstractModel:
 
     def hyperparameter_tune(self, scheduler_options, time_limit=None, **kwargs):
         scheduler_options = copy.deepcopy(scheduler_options)
-        scheduler_options[1]['resource'] = self._preprocess_fit_resources(silent=True, **scheduler_options[1]['resource'])
+        resource = copy.deepcopy(scheduler_options[1]['resource'])
+        if 'num_cpus' in resource:
+            if resource['num_cpus'] == 'auto':
+                resource.pop('num_cpus')
+        if 'num_gpus' in resource:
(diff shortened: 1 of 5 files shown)


szha:
Job PR-924-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-924/1/index.html

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -519,7 +519,15 @@ def _get_init_args(self):
...
     def hyperparameter_tune(self, scheduler_options, time_limit=None, **kwargs):
         scheduler_options = copy.deepcopy(scheduler_options)
-        scheduler_options[1]['resource'] = self._preprocess_fit_resources(silent=True, **scheduler_options[1]['resource'])
+        resource = copy.deepcopy(scheduler_options[1]['resource'])
+        if 'num_cpus' in resource:
+            if resource['num_cpus'] == 'auto':
+                resource.pop('num_cpus')
+        if 'num_gpus' in resource:
+            if resource['num_gpus'] == 'auto':
+                resource.pop('num_gpus')
```
gradientsky, line 528: nit: duplicated code between `num_cpus` and `num_gpus` handling, can be put in a small function

gradientsky (approved):

PR #936, opened 2021-02-17
[Major Upgrade] TextPrediction --> TextPredictor

sxjscience:
*Issue #, if available:*

*Description of changes:*

- TextPrediction --> TextPredictor
- Update tutorials
- Update examples to run experiments on Machine Hack hackathons and Kaggle competition + GLUE
- Move MXNet to lazy dependency to pave the road of multi-backend support.


examples/text_prediction/run_text_prediction.py
```diff
@@ -49,46 +55,66 @@ def get_parser():
...
+    test_df = load_pd.load(args.test_file)
+    train_df = train_df[feature_columns + [label_column]]
+    dev_df = dev_df[feature_columns + [label_column]]
+    test_df = test_df[feature_columns]
+    if args.use_tabular:
+
+        predictor = TabularPredictor(label=label_column,
+                                     eval_metric=eval_metric,
+                                     path=args.exp_dir,
+                                     presets='multimodal')
```
Innixma, line 93: there is no 'multimodal' preset. It is hyperparameters='multimodal'. If you also want stacking, you'd do: presets='best_quality'

core/src/autogluon/core/_setup_utils.py
```diff
@@ -10,7 +10,7 @@
     os.path.join(os.path.dirname(os.path.abspath(__file__)), '..', '..', '..', '..')
 )
 
-PYTHON_REQUIRES = '>=3.6, <3.8'
+PYTHON_REQUIRES = '>=3.6, <=3.8'
```
Innixma: I don't think we support 3.8. Is it necessary for this contribution? This kind of change should really be left to a separate PR dedicated to enabling 3.8 support.

sxjscience: Not necessary, I enabled it for the purpose of working around the installation on Conda, which will install python3.8 by default: https://docs.conda.io/en/latest/miniconda.html

core/src/autogluon/core/metrics/__init__.py
```diff
@@ -307,7 +307,7 @@ def make_scorer(name, score_func, optimum=1, greater_is_better=True,
 
 
 def rmse_func(predictions, targets):
-    return np.sqrt(((predictions - targets) ** 2).mean())
+    return sklearn.metrics.mean_squared_error(targets, predictions, squared=False)
```
Innixma, line 310: Why? This gives the same result but is slower and depends on sklearn. For predictions of small amounts of rows, this is 20x+ slower.

```

import sklearn.metrics
import numpy as np


if __name__ == '__main__':
    y_true = np.random.rand(100)
    y_pred = np.random.rand(100)

    def rmse_func(predictions, targets):
        return np.sqrt(((predictions - targets) ** 2).mean())

    def rmse_func2(predictions, targets):
        return sklearn.metrics.mean_squared_error(targets, predictions, squared=False)

    import time
    start_time = time.time()
    for i in range(10000):
        a = rmse_func(y_pred, y_true)
    end_time = time.time()
    print(f'numpy version:   {end_time - start_time}s')
    start_time = time.time()
    for i in range(10000):
        b = rmse_func2(y_pred, y_true)
    end_time = time.time()
    print(f'sklearn version: {end_time - start_time}s')

```

Output:

```
numpy version:   0.11647295951843262s
sklearn version: 2.302345037460327s
```

Additionally, changes like this should be in separate PRs as they are unrelated to refactoring text API.

sxjscience: I think this is caused by some shape mismatch issues. Let me change it back and verify again.

Innixma: If you can find a situation where there is an issue caused by the current function that is resolved in this fix, then that would be good to share.

Innixma:
@sxjscience could you try to make the changes which make CI pass? Tutorials are still referring to old API.

sxjscience:
@Innixma Updating the tutorial part. Will update soon.

core/src/autogluon/core/metrics/__init__.py
sxjscience: I think previously it's caused by a shape mismatch error: (B, 1) - (B,) will have shape (B, B). Confirmed that the change is not necessary.

szha:
Job PR-936-75 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-936/75/index.html

szha:
Job PR-936-77 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-936/77/index.html

sxjscience:
Some large files are based on the previous version:

- text/src/autogluon/text/text_prediction/mx/models.py [1505 LOC]
   - Based on `text/src/autogluon/text/text_prediction/models/basic_v1.py`
- text/src/autogluon/text/text_prediction/mx/modules.py [641 LOC]
   - Based on `text/src/autogluon/text/text_prediction/modules/basic_prediction.py`
- text/src/autogluon/text/text_prediction/mx/preprocessing.py [523 LOC]
   - New file
- text/src/autogluon/text/text_prediction/predictor/predictor.py [382 LOC]
   - partly based on `text/src/autogluon/text/text_prediction/text_prediction.py `. This is almost new.

text/src/autogluon/text/text_prediction/mx/preprocessing.py
```diff
@@ -0,0 +1,523 @@
...
+from mxnet.gluon.data import ArrayDataset
+from autogluon_contrib_nlp.utils.config import CfgNode
+from autogluon_contrib_nlp.models import get_backbone
+from autogluon_contrib_nlp.data.batchify import Pad, Stack, Tuple
+from autogluon.features import CategoryFeatureGenerator
+
+from .. import constants as _C
+from ..utils import parallel_transform, get_trimmed_lengths
+
+os.environ["TOKENIZERS_PARALLELISM"] = "false"
```
Innixma, line 19: We should add comments for all environment variable modifications.

text/src/autogluon/text/text_prediction/mx/models.py
```diff
@@ -0,0 +1,1505 @@
...
+    cfg.nbest = 1                   # Keep the top K performed models
+    cfg.per_device_batch_size = 16  # Per-device batch-size
+    cfg.auto_per_device_batch_size = True  # Whether to automatically determine the runnable
+                                           # per-device batch_size.
+    cfg.val_batch_size_mult = 2  # By default, we 2X the batch size for validation
+    cfg.lr = 1E-4
+    cfg.final_lr = 0.0
+    cfg.num_train_epochs = 10
+    cfg.warmup_portion = 0.1
+    cfg.layerwise_lr_decay = 0.8  # The layer_wise decay
```
gradientsky, line 194: How this number is selected? [In this paper](https://arxiv.org/pdf/1801.06146.pdf) they use ηl−1=ηl/2.6 (Discriminative  fine-tuning) section.

text/src/autogluon/text/text_prediction/mx/models.py
```diff
@@ -0,0 +1,1505 @@
...
+                plt.xlabel('report_idx')
+                plt.title("Performance vs Training-Time")
+                plt.plot(local_results['report_idx'].iloc[:-1],
+                         local_results[local_results['eval_metric'][0]].iloc[:-1], label=f'task0')
+                plt.legend(loc='best')
+                plt.savefig(plot_training_curves)
+                plt.show()
+            self._results = local_results
+        else:
+            force_forkserver()
```
Innixma, line 1049: Add FIXME: Don't use forkserver if using local sequential HPO

text/src/autogluon/text/text_prediction/mx/models.py
```diff
@@ -0,0 +1,1505 @@
...
+    cfg.per_device_batch_size = 16  # Per-device batch-size
+    cfg.auto_per_device_batch_size = True  # Whether to automatically determine the runnable
+                                           # per-device batch_size.
+    cfg.val_batch_size_mult = 2  # By default, we 2X the batch size for validation
+    cfg.lr = 1E-4
+    cfg.final_lr = 0.0
+    cfg.num_train_epochs = 10
+    cfg.warmup_portion = 0.1
+    cfg.layerwise_lr_decay = 0.8  # The layer_wise decay
+    cfg.wd = 0.01  # Weight Decay
```
gradientsky, line 195: this is probably a good candidate for HPO. [Paper](https://arxiv.org/pdf/1803.09820.pdf)

> Weight decay (WD): This requires a grid search to determine the proper magnitude butusually  does  not  require  more  than  one  significant  figure  accuracy.   Use  your  knowl-edge of the dataset and architecture to decide which values to test.  For example, a morecomplex dataset requires less regularization so test smaller weight decay values, such as10−4,10−5,10−6,0.   A  shallow  architecture  requires  more  regularization  so  test  largerweight decay values, such as10−2,10−3,10−4.

text/src/autogluon/text/text_prediction/mx/models.py
```diff
@@ -0,0 +1,1505 @@
...
+                use_warning = int(os.environ['AUTOGLUON_TEXT_TRAIN_WITHOUT_GPU'])
+            else:
+                use_warning = False
+            if use_warning:
+                warnings.warn('No GPU is detected in the machine and we will recommend you to '
+                              'use TextPrediction on a GPU-enabled instance. Currently, '
+                              'training on CPU is slow.')
+            else:
+                raise RuntimeError('No GPU is detected in the machine and we will '
+                                   'not proceed to run TexPrediction because they will train '
```
Innixma, line 958: TexPrediction -> TextPrediction

text/src/autogluon/text/text_prediction/mx/models.py
```diff
@@ -0,0 +1,1505 @@
...
+                                   'too slowly with only CPU. You may try to set `ngpus_per_trial` '
```
Innixma, line 959: ngpus_per_trial -> num_gpus

text/src/autogluon/text/text_prediction/mx/models.py
```diff
@@ -0,0 +1,1505 @@
...
+    assert num_trainable_layers <= len(all_layers)
+    for i in range(len(all_layers) - num_trainable_layers):
+        for p in all_layers[i].collect_params().values():
+            p.grad_req = 'null'
+    return
+
+
+def base_optimization_config():
+    """The basic optimization phase"""
+    cfg = CfgNode()
```
gradientsky, line 175: Is the optimizer momentum customizable? [Paper](https://arxiv.org/pdf/1803.09820.pdf)
> Momentum:  Short runs with momentum values of 0.99, 0.97, 0.95, and 0.9 will quicklyshow the best value for momentum. If using the 1cycle learning rate schedule, it is better touse a cyclical momentum (CM) that starts at this maximum momentum value and decreaseswith increasing learning rate to a value of 0.8 or 0.85 (performance is almost independentof the minimum momentum value).  Using cyclical momentum along with the LR rangetest stabilizes the convergence when using large learning rate values more than a constantmomentum does.

text/src/autogluon/text/text_prediction/mx/models.py
```diff
@@ -0,0 +1,1505 @@
...
+            nthreads_per_trial = get_cpu_count()
+    nthreads_per_trial = min(nthreads_per_trial, get_cpu_count())
+    ngpus_per_trial = min(ngpus_per_trial, get_gpu_count())
+    assert nthreads_per_trial > 0 and ngpus_per_trial >= 0,\
+        'Invalid number of threads and number of GPUs.'
+    return nthreads_per_trial, ngpus_per_trial
+
+
+@use_np
+class MultiModalTextModel:
```
Innixma, line 785: Is this a public API? If so, rename various things to align closer to predictors, such as:

output_directory -> path, dir_path -> path, test_data -> data

text/src/autogluon/text/text_prediction/mx/models.py
```diff
@@ -0,0 +1,1505 @@
+import numpy as np
```
Innixma, line 1: As a general comment on this PR and future PRs: Always try to separate file moving and logic changes into separate PRs. As it stands, this file (along with modules.py) is extremely difficult to review. Due to the specific circumstances of this PR, I won't make this a blocking issue, but this needs to be considered in future PRs, as I will be reluctant to approve PRs that do this going forward.

Additionally, we should try to keep PRs to below 1000 LOC change if possible (and ideally under 500 LOC). We should try to modularize our code changes as much as possible as it becomes harder to test and properly review as the PR gets larger.

Innixma (commented):
I have focused my review on the text code changes and have mostly skipped over unit tests, examples, and tutorials. I will take a look at those once my initial comments have been addressed.

text/src/autogluon/text/text_prediction/metrics.py
```diff
@@ -1,15 +1,51 @@
...
+    log_metrics
+        The updated logging metric
+    """
+    if problem_type == MULTICLASS:
+        if eval_metric is None:
+            eval_metric = 'acc'
+        log_metrics = ['acc', 'log_loss']
+    elif problem_type == BINARY:
+        if eval_metric is None:
+            eval_metric = 'roc_auc'
```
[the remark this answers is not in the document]
gradientsky: `roc_auc` is probably a better metric for binary classification.

text/src/autogluon/text/text_prediction/infer_types.py
```diff
@@ -0,0 +1,239 @@
...
+            num_train_missing = train_df[col_name].isnull().sum()
+            num_valid_missing = valid_df[col_name].isnull().sum()
+            if num_train_missing > 0:
+                raise ValueError(f'Label column "{col_name}" contains missing values in the '
+                                 f'training data frame. You may want to filter your data because '
+                                 f'missing label is currently not supported.')
+            if num_valid_missing > 0:
+                raise ValueError(f'Label column "{col_name}" contains missing values in the '
+                                 f'validation data frame. You may want to filter your data because '
+                                 f'missing label is currently not supported.')
```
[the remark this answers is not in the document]
Innixma: I'd say correct response is to error for now. NaNs are always tricky and we don't want to assume what the user wants to do with them. Let them clean their data and remove the NaNs.

text/src/autogluon/text/text_prediction/metrics.py
```diff
@@ -1,15 +1,51 @@
...
+        log_metrics = ['f1', 'mcc', 'roc_auc', 'acc', 'log_loss']
```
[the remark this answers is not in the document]
sxjscience: Currently this is for the purpose of logging so we will report lots of metrics.

text/src/autogluon/text/text_prediction/metrics.py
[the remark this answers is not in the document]
sxjscience: I thought that we have been using roc_auc in Tabular... Since this one is more reasonable to look at if it's binary classification.

text/src/autogluon/text/text_prediction/infer_types.py
```diff
@@ -0,0 +1,239 @@
+import collections
```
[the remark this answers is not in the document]
sxjscience: Yeah, I previously tried to merge these two but later noticed that it's not very easy so I changed it like this.

text/src/autogluon/text/text_prediction/mx/models.py
sxjscience: Yes, just need to change these values.

tabular/src/autogluon/tabular/trainer/model_presets/presets.py
```diff
@@ -154,7 +154,11 @@ def get_preset_models(path, problem_type, eval_metric, hyperparameters, feature_
                 logger.log(20, f"\tFound '{model_type}' model in hyperparameters, but '{model_type}' is present in `excluded_model_types` and will be removed.")
                 continue  # Don't include excluded models
             if isinstance(model_cfg, str):
-                model_cfgs_to_process += get_preset_custom(name=model_cfg, problem_type=problem_type, num_classes=num_classes)
+                if model_type == 'TEXT_NN_V1':
```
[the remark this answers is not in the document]
sxjscience: I'm not sure. What's your opinion?

core/src/autogluon/core/_setup_utils.py
```diff
@@ -10,7 +10,7 @@
     os.path.join(os.path.dirname(os.path.abspath(__file__)), '..', '..', '..', '..')
 )
 
-PYTHON_REQUIRES = '>=3.6, <3.8'
+PYTHON_REQUIRES = '>=3.6, <3.9'
```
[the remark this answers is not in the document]
sxjscience: Reverted.

core/src/autogluon/core/utils/loaders/load_pd.py
```diff
@@ -32,7 +32,7 @@ def load(path, delimiter=None, encoding='utf-8', columns_to_keep=None, dtype=Non
         format = 'multipart_s3'
     elif path[-1] == '/' and not s3_utils.is_s3_url(path):  # and path[:2] != 's3'
         format = 'multipart_local'
-    elif '.parquet' in path or path[-1] == '/':
+    elif '.parquet' in path or '.pq' in path or path[-1] == '/':
```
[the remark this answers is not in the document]
sxjscience: Do you think that we can use `.pq` for parquet? For example, in the pyarrow example, they have 

```
import pyarrow.parquet as pq
```

examples/text_prediction/run_competition.py
```diff
@@ -0,0 +1,209 @@
...
+    elif args.task == 'price_of_books':
+        train_df, test_df, label_column = load_price_of_books(args.train_file, args.test_file)
+    elif args.task == 'data_scientist_salary':
+        train_df, test_df, label_column = load_data_scientist_salary(args.train_file, args.test_file)
+    else:
+        raise NotImplementedError
+    if args.mode == 'stacking':
+        predictor = TabularPredictor(label=label_column,
+                                     eval_metric=args.eval_metric,
+                                     path=args.exp_dir)
```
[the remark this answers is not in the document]
sxjscience: OK, I think this may be easier to edit if we'd like to later add another branch.

tabular/src/autogluon/tabular/configs/hyperparameter_configs.py
```diff
@@ -79,23 +79,9 @@
...
-            {'criterion': 'gini', 'ag_args': {'name_suffix': 'Gini', 'problem_types': ['binary', 'multiclass']}},
-            {'criterion': 'entropy', 'ag_args': {'name_suffix': 'Entr', 'problem_types': ['binary', 'multiclass']}},
-            {'criterion': 'mse', 'ag_args': {'name_suffix': 'MSE', 'problem_types': ['regression']}},
-        ],
-        'KNN': [
-            {'weights': 'uniform', 'ag_args': {'name_suffix': 'Unif'}},
-            {'weights': 'distance', 'ag_args': {'name_suffix': 'Dist'}},
-        ],
-        'TEXT_NN_V1': ['ELECTRA_BASE'],
+        # 'FASTAI': {}, FastAI gets killed if the dataset is large (400K rows).
```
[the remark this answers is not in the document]
sxjscience: For the issue I experienced, it may not be related to forkserver code because I disabled that. It's just that the process gots killed when I'm training the model on the 430K dataset.

text/setup.py
```diff
@@ -27,7 +27,7 @@
     f'autogluon.core=={version}',
     f'autogluon.mxnet=={version}',
     'pyarrow>=3.0.0',
-    'autogluon-contrib-nlp==0.0.1b20210112',
+    'autogluon-contrib-nlp==0.0.1b20210201',
```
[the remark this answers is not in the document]
sxjscience: Currently not, this is not an official package.

tabular/src/autogluon/tabular/trainer/model_presets/presets_custom.py
```diff
@@ -11,22 +11,5 @@ def get_preset_custom(name, problem_type, num_classes):
...
         return [model]
-    elif name == 'ELECTRA_SMALL':
-        return _get_preset_electra('google_electra_small', 'default_no_hpo', 'Small')
-    elif name == 'ELECTRA_BASE':
-        return _get_preset_electra('google_electra_base', 'default_no_hpo', 'Base')
-    elif name == 'ELECTRA_BASE_HPO':
-        return _get_preset_electra('google_electra_base', 'default', 'BaseHPO')
-    elif name == 'ELECTRA_LARGE':
-        return _get_preset_electra('google_electra_large', 'default_no_hpo', 'Large')
     else:
```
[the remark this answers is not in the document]
sxjscience: Yes, I changed the code to rely on the presets logic in TextPrediction.

text/src/autogluon/text/text_prediction/mx/models.py
```diff
@@ -0,0 +1,1505 @@
...
+    """
+    if nthreads_per_trial is None and ngpus_per_trial is None:
+        nthreads_per_trial = get_cpu_count()
+        ngpus_per_trial = 1
+    elif nthreads_per_trial is not None and ngpus_per_trial is None:
+        ngpus_per_trial = 1
+    elif nthreads_per_trial is None and ngpus_per_trial is not None:
+        if ngpus_per_trial != 0:
+            num_parallel_jobs = get_gpu_count() // ngpus_per_trial
+            nthreads_per_trial = max(get_cpu_count() // num_parallel_jobs, 1)
```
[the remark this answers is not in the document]
sxjscience: I think it will be int since it's using the `//`, basically `3 // 2 --> 1`.

text/src/autogluon/text/text_prediction/infer_types.py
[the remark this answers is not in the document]
sxjscience: I think we should not allow missing values in the label column. This will make it really difficult to handle.

text/src/autogluon/text/text_prediction/mx/models.py
sxjscience: No HPO has been conducted yet. I'm going to run the HPO on the benchmark so we will see the best number.

sxjscience: Thus, this one is currently just based on the small product sentiment analysis dataset.

sxjscience: I did a basic HPO on that dataset. But we may need to change this.

text/src/autogluon/text/text_prediction/mx/models.py
sxjscience: We can specify the value in presets like in http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-936/77/tutorials/text_prediction/customization.html#change-hyperparameter

sxjscience: Basically, all data in the yml config are configurable.

tabular/src/autogluon/tabular/trainer/model_presets/presets.py
```diff
@@ -154,7 +154,11 @@ def get_preset_models(path, problem_type, eval_metric, hyperparameters, feature_
...
+                    from autogluon.text import ag_text_presets
```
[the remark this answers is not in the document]
sxjscience: @gradientsky I added the code here to load customized config.

text/src/autogluon/text/text_prediction/presets.py
```diff
@@ -0,0 +1,214 @@
...
+
+import copy
+import logging
+import functools
+from autogluon_contrib_nlp.utils.registry import Registry
+from autogluon.core import space
+
+logger = logging.getLogger(__name__)  # return root logger
+
+ag_text_presets = Registry('ag_text_presets')
```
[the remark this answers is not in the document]
sxjscience: The advantage of the current one is that the users can define their own presets easily and list the keys very easily. This is based on the Registry that is borrowed from GluonNLP: https://github.com/dmlc/gluon-nlp/blob/master/src/gluonnlp/utils/registry.py

core/src/autogluon/core/metrics/__init__.py
```diff
@@ -294,14 +294,23 @@ def make_scorer(name, score_func, optimum=1, greater_is_better=True,
                                     optimum=0,
                                     greater_is_better=False)
 
+
+def local_spearmanr(predictions, ground_truth):
```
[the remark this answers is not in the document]
sxjscience: Yes

core/src/autogluon/core/_setup_utils.py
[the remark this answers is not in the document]
sxjscience: Reverted.

text/src/autogluon/text/text_prediction/predictor/predictor.py
```diff
@@ -0,0 +1,382 @@
...
+            The training data
+        tuning_data
+            The tuning data
+        time_limit
+            The time limits
+        presets
+            The user can specify the presets of the hyper-parameters.
+        hyperparameters
+            The hyper-parameters
+        feature_columns
```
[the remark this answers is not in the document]
sxjscience: Do you think that we should just remove it?

szha:
Job PR-936-78 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-936/78/index.html

szha:
Job PR-936-80 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-936/80/index.html

text/src/autogluon/text/text_prediction/predictor/predictor.py
```diff
@@ -0,0 +1,382 @@
...
+        # Get tuning data
+        if tuning_data is not None:
+            if not isinstance(tuning_data, pd.DataFrame):
+                tuning_data = load_pd.load(tuning_data)
+            tuning_data = tuning_data[all_columns]
+        else:
+            if hyperparameters['misc']['holdout_frac'] is not None:
+                holdout_frac = hyperparameters['misc']['holdout_frac']
+            else:
+                num_trials = hyperparameters['hpo_params']['num_trials']
```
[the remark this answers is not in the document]
sxjscience: This won't happen usually. Because the user are asked to generate the new hparam based on an existing hparam.

text/src/autogluon/text/text_prediction/predictor/predictor.py
```diff
@@ -0,0 +1,382 @@
...
+        num_cpus
+            The number of CPUs to use for each trial
+        num_gpus
+            The number of GPUs to use for each trial
+        num_trials
+            The number of trials. By default, we will use the provided number of trials in the
+            hyperparameters or presets. This will overwrite the provided value.
+        plot_results
+            Whether to plot results.
+        seed
```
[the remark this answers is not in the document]
sxjscience: I think that usually the default value for random_state / seed will be None. (But specifying the seed is also reasonable).

text/src/autogluon/text/text_prediction/mx/models.py
sxjscience: This should be mostly private I think.

core/src/autogluon/core/metrics/__init__.py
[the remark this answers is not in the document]
jwmueller: doesn't matter here, but for future metrics the `predictions` arg should come after `ground_truth` like the existing metrics?

examples/text_prediction/run_text_prediction.py
```diff
@@ -46,49 +54,90 @@ def get_parser():
...
+                      num_bag_folds=5,
+                      num_stack_levels=1)
+    elif args.mode == 'weighted':
+        predictor = TabularPredictor(label=label_column,
+                                     eval_metric=eval_metric,
+                                     path=args.exp_dir)
+        predictor.fit(train_data=real_train_df,
+                      tuning_data=real_dev_df,
+                      hyperparameters='multimodal')
+    elif args.mode == 'single':
```
jwmueller, line 114: May want to add comment that highlights why TextPredictor is used for 'single', TabularPredictor for 'weighted', 'stacking'. Otherwise this is easy to miss.

tabular/src/autogluon/tabular/trainer/model_presets/presets.py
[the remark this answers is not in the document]
jwmueller: I'd remove it unless you have plans to have `_V2, _V3, ...` together at the same time.  Even in that case, more informative names would be better.

text/src/autogluon/text/text_prediction/metrics.py
```diff
@@ -1,15 +1,51 @@
...
+    elif problem_type == REGRESSION:
+        if eval_metric is None:
+            eval_metric = 'rmse'
+        log_metrics = ['mse', 'rmse', 'mae']
```
jwmueller, line 41: I don't think you should report both 'mse' and 'rmse'. Something like r2 or pearson/spearman would be much more informative to replace say MSE.

jwmueller: Also one downside of logging these alternative metrics, is users may then not specify them in `eval_metric` argument even when they only really care about one of the logged metrics. So may want to somehow prompt user to consider specifying `eval_metric` when they are looking through the log_metrics.

text/src/autogluon/text/text_prediction/utils.py
```diff
@@ -0,0 +1,156 @@
...
+                    trimmed_lengths[nonzero_idx[i]] += 1
+            else:
+                increment = min(min(nonzero_budgets), remainder // len(nonzero_idx))
+                trimmed_lengths[nonzero_idx] += increment
+        return trimmed_lengths
+    else:
+        return np.minimum(lengths, max_length)
+
+
+def logging_config(folder: Optional[str] = None,
```
jwmueller, line 98: This should go in core/ right, so other modules can also use it?

text/src/autogluon/text/text_prediction/presets.py
```diff
@@ -0,0 +1,214 @@
...
+
+
+def base() -> dict:
+    """The default hyperparameters.
+
+    It will have a version key and a list of candidate models.
+    Each model has its own search space inside.
+    """
+    ret = {
+        'version': 1,                     # Version of TextPrediction Model
```
jwmueller, line 26: Do we really need this `version`, isn't it tied to the autogluon version already?
If it's really needed, I think you may want to use more informative names rather than simply version 1, 2, ...

text/src/autogluon/text/text_prediction/metrics.py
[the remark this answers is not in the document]
Innixma: @jwmueller Thoughts? roc_auc is a more reasonable metric in general than accuracy (I find it more useful), but I suspect most people would expect accuracy as the default (and newcomers to ML wouldn't even know what roc_auc is).

core/src/autogluon/core/metrics/__init__.py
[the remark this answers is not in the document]
Innixma: agreed, it should be (y_true, y_pred)

text/src/autogluon/text/text_prediction/predictor/predictor.py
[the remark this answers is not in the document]
Innixma: I would recommend removing it since it can trivially be done outside of AG.

text/src/autogluon/text/text_prediction/mx/preprocessing.py
```diff
@@ -0,0 +1,523 @@
...
+    cfg.text.auto_max_length_quantile = 0.95  # We will ensure that the new max_length is around the quantile of the lengths of all samples
+    cfg.text.auto_max_length_round_to = 32    # We will ensure that the automatically determined max length will be divisible by round_to
+    cfg.categorical = CfgNode()
+    cfg.categorical.minimum_cat_count = 100   # The minimal number of data per categorical group
+    cfg.categorical.maximum_num_cat = 20      # The minimal number of data per categorical group
+    cfg.categorical.convert_to_text = False   # Whether to convert the feature to text
+
+    cfg.numerical = CfgNode()
+    cfg.numerical.convert_to_text = False     # Whether to convert the feature to text
+    cfg.numerical.impute_strategy = 'mean'    # Whether to use mean to fill in the missing values.
```
jwmueller, line 41: list options in the comment here

text/src/autogluon/text/text_prediction/predictor/predictor.py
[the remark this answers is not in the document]
Innixma: I'd prefer setting it to 0 (Tabular's models are all fixed seed by default), this will make benchmarking more straightforward and it will be easier to tell if we've made a change which impacts the model training, plus if we share a script users will get the same result in subsequent runs.

text/src/autogluon/text/text_prediction/predictor/predictor.py
```diff
@@ -0,0 +1,398 @@
...
+        If `problem_type = None`, the prediction problem type is inferred based on the
+        label-values in provided dataset.
+    eval_metric : function or str, default = None
+        Metric by which predictions will be ultimately evaluated on test data.
+        AutoGluon tunes factors such as hyper-parameters, early-stopping, ensemble-weights, etc.
+        in order to improve this metric on validation data.
+
+        If `eval_metric = None`, it is automatically chosen based on `problem_type`.
+        Defaults to 'roc_auc' for binary, 'accuracy' for multiclass classification, and
+        'root_mean_squared_error' for regression.
```
jwmueller, line 42: standardize defaults across all tasks. I am fine with using 'roc_auc' as default for binary on all tasks as long as it's standardized across tasks.

text/src/autogluon/text/text_prediction/predictor/predictor.py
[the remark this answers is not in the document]
jwmueller: Do you have example of this? It also seems confusing to me

text/src/autogluon/text/text_prediction/predictor/predictor.py
```diff
@@ -0,0 +1,382 @@
...
+        if not has_text_column:
+            raise AssertionError('No Text Column is found! This is currently not supported by '
+                                 'the TextPrediction task. You may try to use '
+                                 'autogluon.tabular.TabularPredictor.\n'
+                                 'The inferred column properties of the training data is {}'
+                                 .format(train_data))
+        logger.log(25, 'Problem Type="{}"'.format(problem_type))
+        logger.log(25, printable_column_type_string(column_types))
+        self._problem_type = problem_type
+        model_hparams = hyperparameters['models']['MultimodalTextModel']
```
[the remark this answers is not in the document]
jwmueller: I agree, 'models' should not be the name of a key in `hyperparameters`

text/src/autogluon/text/text_prediction/predictor/predictor.py
```diff
@@ -0,0 +1,398 @@
...
+        -------
+        output
+        """
+        assert self._model is not None, 'Model does not seem to have been constructed. ' \
+                                        'Have you called fit(), or load()?'
+        output = self._model.extract_embedding(dataset,
+                                               stochastic_chunk=stochastic_chunk,
+                                               num_repeat=num_repeat)
+        if as_pandas:
+            output = pd.DataFrame({self.label: output})
```
jwmueller, line 354: This is for extracting embedding of the features right, so why is `self.label` the name used here?  
And why would this embedding only be 1 column? Shouldn't it be many columns?

text/src/autogluon/text/text_prediction/predictor/predictor.py
```diff
@@ -0,0 +1,398 @@
...
+            The directory path to save the model artifacts
+
+        """
+        assert self._model is not None, 'Model does not seem to have been constructed.' \
+                                        ' Have you called fit(), or load()?'
+        os.makedirs(dir_path, exist_ok=True)
+        with open(os.path.join(dir_path, 'text_predictor_assets.json'), 'w') as of:
+            json.dump({'backend': self._backend,
+                       'label': self._label}, of)
+        self._model.save(os.path.join(dir_path, 'saved_model'))
```
jwmueller, line 375: set `self._model` to None and then save `TextPredictor`. 
For loading, make sure to load `self._model after loading `TextPredictor`.

jwmueller: Or is there nothing to save in TextPredictor besides `text_predictor_assets`?

docs/tutorials/text_prediction/customization.md
```diff
@@ -1,274 +1,204 @@
...
-            },
-    },
-    'hpo_params': {
-        'search_strategy': 'random'  # perform HPO via simple random search
-    }
-}
+from autogluon.text import TextPredictor
+predictor = TextPredictor(path='ag_text_sst_electra_small', eval_metric='acc', label='label')
+predictor.set_verbosity(0)
+predictor.fit(train_data, presets='electra_small_fuse_late', time_limit=60, seed=123)
```
jwmueller, line 51: what is the advantage of allowing user to specify backbone-type, fusion-type as presets? why not as hyperparameters?

Having less presets may make things simpler for novice users, while expert users can still customize via `hyperparameters`?

jwmueller: This also deviates from Tabular, where users choose which models to train via `hyperparameters`, not `presets`.

docs/tutorials/text_prediction/customization.md
```diff
@@ -1,274 +1,204 @@
...
 ```
 
-We can now call `fit()` with hyperparameter-tuning over our custom search space. 
-Below `num_trials` controls the maximal number of different hyperparameter configurations for which AutoGluon will train models (5 models are trained under different hyperparameter configurations in this case). To achieve good performance in your applications, you should use larger values of `num_trials`, which may identify superior hyperparameter values but will require longer runtimes.
 
 ```{.python .input}
```
jwmueller, line 55: Add comment that if users really care about one of these metrics, they should specify it as `eval_metric` before training so that AutoGluon can optimize for it.

docs/tutorials/text_prediction/customization.md
```diff
@@ -1,274 +1,204 @@
...
-predictor_mrpc = task.fit(train_data,
-                          label='label',
-                          hyperparameters=hyperparameters,
-                          num_trials=2,  # increase this to achieve good performance in your applications
-                          time_limits=60 * 2,
-                          ngpus_per_trial=1,
-                          seed=123,
-                          output_directory='./ag_mrpc_random_search')
+predictor.evaluate(dev_data, metrics=['f1', 'acc'])
```
jwmueller, line 56: Add comment that if users really care about one of these metrics, they should specify it as eval_metric before training so that AutoGluon can optimize for it.

docs/tutorials/text_prediction/customization.md
```diff
@@ -1,274 +1,204 @@
...
-print('Prediction = "{}"'.format(prediction1[0] == 1))
-print('Prob = "{}"'.format(prediction1_prob[0]))
-print('')
-prediction2 = predictor_mrpc.predict({'sentence1': [sentence1], 'sentence2': [sentence3]})
-prediction2_prob = predictor_mrpc.predict_proba({'sentence1': [sentence1], 'sentence2': [sentence3]})
-print('A = "{}"'.format(sentence1))
-print('B = "{}"'.format(sentence3))
-print('Prediction = "{}"'.format(prediction2[0] == 1))
-print('Prob = "{}"'.format(prediction2_prob[0]))
+predictor.fit(train_data, hyperparameters=ag_text_presets.create('electra_small_fuse_late'),
```
jwmueller, line 70: can you show more customization here? Eg. specifying the learning-rate or regularization

docs/tutorials/text_prediction/customization.md
```diff
@@ -1,274 +1,204 @@
...
-print('B = "{}"'.format(sentence3))
-print('Prediction = "{}"'.format(prediction2[0] == 1))
-print('Prob = "{}"'.format(prediction2_prob[0]))
+@ag_text_presets.register()
+def electra_small_fuse_late_train5():
+    hyperparameters = ag_text_presets.create('electra_small_fuse_late')
+    hyperparameters['models']['MultimodalTextModel']['search_space']['optimization.num_train_epochs'] = 5
+    return hyperparameters
+
+predictor.fit(train_data, presets='electra_small_fuse_late_train5', time_limit=60, seed=123)
```
jwmueller, line 98: This seems like a nice UX. 
@Innixma @gradientsky  Should we add this to core so we can use in TabularPredictor as well?

szha:
Job PR-936-84 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-936/84/index.html

szha:
Job PR-936-85 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-936/85/index.html

tabular/src/autogluon/tabular/trainer/model_presets/presets.py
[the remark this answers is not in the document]
sxjscience: Added try catch statement.

text/src/autogluon/text/text_prediction/infer_types.py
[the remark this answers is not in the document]
sxjscience: Added.

tabular/src/autogluon/tabular/trainer/model_presets/presets.py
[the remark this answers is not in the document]
sxjscience: Should we call it `MULTIMODAL_NN_V1`? (Because I may later like to update to V2).

text/src/autogluon/text/text_prediction/predictor/predictor.py
Innixma: Lets set to accuracy for now, since vision doesn't support any metrics beyond accuracy (if I'm not mistaken), and because most models / automl packages default to accuracy outside of AG.

docs/tutorials/text_prediction/customization.md
Innixma: It is quite reasonable, but probably won't make it for v0.1 unless we have spare time or someone wants to take it up.

Also, not sure how much code is in this register logic, we'd need to extract it from gluonnlp I think.

text/src/autogluon/text/text_prediction/mx/models.py
sxjscience: Revised.

docs/tutorials/text_prediction/customization.md
sxjscience: Added.

szha:
Job PR-936-88 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-936/88/index.html

text/src/autogluon/text/text_prediction/predictor/predictor.py
sxjscience: Yes, there's nothing to save besides those in assets.

text/src/autogluon/text/text_prediction/predictor/predictor.py
sxjscience: Changed to accuracy.

text/src/autogluon/text/text_prediction/presets.py
sxjscience: Nice catch. I removed this version flag.

text/src/autogluon/text/text_prediction/mx/preprocessing.py
sxjscience: Listed.

text/src/autogluon/text/text_prediction/predictor/predictor.py
sxjscience: Revised.

text/src/autogluon/text/text_prediction/utils.py
sxjscience: May be in a later PR.

text/src/autogluon/text/text_prediction/metrics.py
sxjscience: Replaced with r2.

examples/text_prediction/run_text_prediction.py
sxjscience: added the comment.

docs/tutorials/text_prediction/customization.md
sxjscience: Currently both are supported.

text/src/autogluon/text/text_prediction/mx/models.py
sxjscience: Mark as future work.

core/src/autogluon/core/metrics/__init__.py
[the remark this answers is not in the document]
sxjscience: fixed.

docs/tutorials/text_prediction/customization.md
sxjscience: I tried to update both number of training epochs and lr.

szha:
Job PR-936-106 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-936/106/index.html

szha:
Job PR-936-107 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-936/107/index.html

sxjscience:
Most comments are addressed and the CI has passed @Innixma @jwmueller @gradientsky

szha:
Job PR-936-108 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-936/108/index.html

examples/text_prediction/README.md
```diff
@@ -1,15 +1,144 @@
...
+ 
 ### Prepare the data
 ```bash
 python3 prepare_glue.py --benchmark glue
 ```
 
 ### Run the benchmark
-Run on all datasets
+Run on all datasets with either a single `TextPredictor` model or the `multimodal` configuration 
+in AutoGluon Tabular that will use the `TextPredictor` model in the 5-fold-1-layer stacking.
```
jwmueller, line 124: "use the `TextPredictor` model in the 5-fold-1-layer stacking."
-> 
"combine the `TextPredictor` model with tabular models from AutoGluon-Tabular via a single layer of  stack ensembling with 5-fold bagging."

tabular/src/autogluon/tabular/configs/hyperparameter_configs.py
```diff
@@ -79,23 +79,9 @@
...
         'custom': ['GBM'],
+        'AG_TEXT_NN': ['medium_quality'],
```
jwmueller, line 84: If user specifies `presets` in TabularPredictor.fit(), will these also affect the creation of AG_TEXT_NN?  I think it should, eg. `best_quality` can use larger AG_TEXT_NN than the default.

jwmueller: This could be done in future PR though, if you mark as TODO

tabular/src/autogluon/tabular/models/text_prediction/text_prediction_v1_model.py
```diff
@@ -1,19 +1,16 @@
 """Text Prediction Model based on Pretrained Language Model. Version 1"""
```
jwmueller, line 1: Why not clarify in docstring this is simply a wrapper around AutoGluon's TextPredictor?

tabular/src/autogluon/tabular/models/text_prediction/text_prediction_v1_model.py
```diff
@@ -195,92 +114,49 @@ def _fit(self, X: pd.DataFrame, y: pd.Series,
...
-                         search_strategy=search_strategy,
-                         search_options=self.params['hpo_params']['search_options'],
-                         scheduler_options=scheduler_options,
-                         num_trials=self.params['hpo_params']['num_trials'],
-                         console_log=verbosity >= 3,
-                         ignore_warning=verbosity < 3,
-                         verbosity=verbosity-1)
+        assert self.params['hpo_params']['num_trials'] == 1 \
+               or self.params['hpo_params']['num_trials'] is None,\
+            'Currently, you cannot nest the hyperparameter search in text neural network ' \
```
jwmueller, line 151: Can we clarify this error msg? The msg should also tell user what to do instead.

text/src/autogluon/text/text_prediction/utils.py
jwmueller: maybe add TODO?
Ideally, different tasks should not have different logging options, it should be standardized and this seems like a nice standard.

text/src/autogluon/text/text_prediction/presets.py
```diff
@@ -0,0 +1,239 @@
...
+                                                   fusion_strategy=fusion_strategy))
+
+
+@ag_text_presets.register()
+def default() -> dict:
+    return ag_text_presets.create('electra_base_fuse_late')
+
+
+@ag_text_presets.register()
+def fast_train() -> dict:
```
jwmueller, line 180: Can we standardize preset names with Tabular?

If standardized, this preset should be called: `medium_quality_faster_train`

text/src/autogluon/text/text_prediction/presets.py
```diff
@@ -0,0 +1,239 @@
...
+    """Configuration that supports fast training and inference.
+
+    By default, we use the late-fusion aggregator with electra-small
+    """
+    cfg = ag_text_presets.create('electra_small_fuse_late')
+    return cfg
+
+
+@ag_text_presets.register()
+def medium_quality() -> dict:
```
jwmueller, line 190: Can we standardize preset names with Tabular?

If standardized, this preset should be called: `medium_quality_faster_train`

text/src/autogluon/text/text_prediction/predictor/predictor.py
```diff
@@ -0,0 +1,382 @@
...
+            num_cpus=None,
+            num_gpus=None,
+            num_trials=None,
+            plot_results=None,
+            seed=None):
+        """Fit the predictor
+
+        Parameters
+        ----------
+        train_data
```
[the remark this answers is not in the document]
jwmueller: I'll go over docs/tutorials in a fast followup

text/src/autogluon/text/text_prediction/predictor/predictor.py
```diff
@@ -0,0 +1,407 @@
...
+        Parameters
+        ----------
+        dataset
+            The dataset to extract the embedding. It can either be a pandas dataframe or
+            the path to the pandas dataframe.
+
+        Returns
+        -------
+        embeddings
+            The output will have shape (#sample, #embedding dimension)
```
jwmueller, line 354: "The output will have shape (#sample, #embedding dimension)"
->
"The output will be numpy array of shape (len(dataset), D), where the embedding dimension D is determined by the neural network's architecture."

text/src/autogluon/text/text_prediction/predictor/predictor.py
```diff
@@ -0,0 +1,407 @@
...
+        """
+        assert self._model is not None, 'Model does not seem to have been constructed. ' \
```
jwmueller, line 356: maybe add "TODO: add `as_pandas` option (where the each dimension of embedding is a different column in returned DataFrame)".

sxjscience: I think this is not very useful so I removed it.

text/src/autogluon/text/text_prediction/predictor/predictor.py
```diff
@@ -0,0 +1,382 @@
...
+            The tuning data
+        time_limit
+            The time limits
+        presets
+            The user can specify the presets of the hyper-parameters.
+        hyperparameters
+            The hyper-parameters
+        feature_columns
+            Specify which columns in the data
+        column_types
```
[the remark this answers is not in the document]
jwmueller: At least, can we rename this arg `feature_metadata` to standardize with Tabular?

text/src/autogluon/text/text_prediction/predictor/predictor.py
```diff
@@ -0,0 +1,407 @@
...
+                                for training and inference
+                - medium_quality: Model with the medium quality
+                - fast_train: Model that can be trained and inferenced very fast compared to the other two options.
+            You may try to list all presets via `autogluon.text.ag_text_presets.list_keys()`.
+        hyperparameters
+            The hyper-parameters of the fit function. This can be used to specify the
+            search space and the configuration of the network.
+        column_types
+            The provided type of the columns. It will be a dictionary that maps the column name
+            to the type of the column.
```
jwmueller, line 163: what are the possible values in this dict?

text/src/autogluon/text/text_prediction/predictor/predictor.py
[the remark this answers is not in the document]
sxjscience: Improved.

text/src/autogluon/text/text_prediction/predictor/predictor.py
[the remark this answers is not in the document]
sxjscience: Improved for now. Need to create a separate PR to enhance the description.

text/src/autogluon/text/text_prediction/predictor/predictor.py
```diff
@@ -0,0 +1,382 @@
...
+                    'all columns = "{}"'.format(col, train_data.columns)
+            all_columns = feature_columns + label_columns
+        train_data = train_data[all_columns]
+        # Get tuning data
+        if tuning_data is not None:
+            if not isinstance(tuning_data, pd.DataFrame):
+                tuning_data = load_pd.load(tuning_data)
+            tuning_data = tuning_data[all_columns]
+        else:
+            if hyperparameters['misc']['holdout_frac'] is not None:
```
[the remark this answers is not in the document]
jwmueller: IMO `hyperparameters` are supposed to be options we choose from based on validation-score.  By this reasoning, `holdout_frac` does not belong in `hyperparameters`.

text/src/autogluon/text/text_prediction/predictor/predictor.py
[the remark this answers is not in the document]
sxjscience: I think previously we are considering to do a mini-ensemble in text also.

tabular/src/autogluon/tabular/configs/hyperparameter_configs.py
Innixma: Don't do this change now, just have as TODO.

sxjscience: Anyway, I changed it.

sxjscience: Oh, I misread the issue. Yes, this is marked as a TODO.

tabular/src/autogluon/tabular/models/text_prediction/text_prediction_v1_model.py
```diff
@@ -28,7 +25,7 @@ class TextPredictionV1Model(AbstractModel):
     nn_model_name = 'text_nn'
 
     def __init__(self, **kwargs):
-        """The TextPredictionV1Model.
+        """Wrapper of the Text Prediction model in AutoGluon Text.
```
Innixma, line 28: Wrapper of autogluon.text.TextPredictor

tabular/src/autogluon/tabular/trainer/model_presets/presets.py
```diff
@@ -154,7 +154,18 @@ def get_preset_models(path, problem_type, eval_metric, hyperparameters, feature_
...
                 continue  # Don't include excluded models
             if isinstance(model_cfg, str):
-                model_cfgs_to_process += get_preset_custom(name=model_cfg, problem_type=problem_type, num_classes=num_classes)
+                if model_type == 'AG_TEXT_NN':
+                    AG_TEXT_IMPORT_ERROR = 'autogluon.text has not been installed. ' \
+                                           'You may try to install "autogluon.text" ' \
+                                           'first by running. ' \
+                                           '`python3 -m pip install autogluon.text`'
+                    try:
+                        from autogluon.text import ag_text_presets
```
Innixma, line 163: Better to add a new try_except function where all the other try_except functions are, will look cleaner.

text/src/autogluon/text/text_prediction/predictor/predictor.py
```diff
@@ -0,0 +1,408 @@
...
+
+        Returns
+        -------
+        ret : a single number or a dict of metric --> metric scores
+            The output metrics. It will be a single value if there is only one metric and
+            will be a dictionary of {metric_name --> value} if there are multiple metrics to report.
+        """
+        return self._model.evaluate(data, metrics=metrics)
+
+    def predict(self, dataset, as_pandas=True):
```
Innixma, line 296: dataset -> data

text/src/autogluon/text/text_prediction/predictor/predictor.py
```diff
@@ -0,0 +1,408 @@
...
+            Whether to convert the output to a pandas dataframe
+
+        """
+        assert self._model is not None, 'Model does not seem to have been constructed. Have you called fit(), or load()?'
+        output = self._model.predict(dataset)
+        if as_pandas:
+            output = pd.DataFrame({self.label: output})[self.label]
+        return output
+
+    def predict_proba(self, dataset, as_pandas=True):
```
Innixma, line 320: dataset -> data

text/src/autogluon/text/text_prediction/predictor/predictor.py
```diff
@@ -0,0 +1,408 @@
...
+        """
+        assert self._model is not None,\
+            'Model does not seem to have been constructed. ' \
+            'Have you called fit(), or load()?'
+        output = self._model.predict_proba(dataset)
+        if as_pandas:
+            output = pd.DataFrame(output)
+        return output
+
+    def extract_embedding(self, dataset):
```
Innixma, line 343: dataset -> data

tabular/src/autogluon/tabular/models/text_prediction/text_prediction_v1_model.py
sxjscience: Changed.

text/src/autogluon/text/text_prediction/predictor/predictor.py
sxjscience: Changed.

text/src/autogluon/text/text_prediction/predictor/predictor.py
sxjscience: Changed.

text/src/autogluon/text/text_prediction/predictor/predictor.py
sxjscience: Changed.

szha:
Job PR-936-110 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-936/110/index.html

text/src/autogluon/text/text_prediction/predictor/predictor.py
jwmueller: This is useful if users want to merge these embeddings with their dataset, and train some of their own tabular models (outside of autogluon).

szha:
Job PR-936-113 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-936/113/index.html

szha:
Job PR-936-114 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-936/114/index.html

text/src/autogluon/text/text_prediction/predictor/predictor.py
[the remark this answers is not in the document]
sxjscience: Revised.

text/src/autogluon/text/text_prediction/predictor/predictor.py
```diff
@@ -0,0 +1,382 @@
...
+            hyperparameters['seed'] = seed
+        seed = hyperparameters['seed']
+        if num_trials is not None:
+            hyperparameters['hpo_params']['num_trials'] = num_trials
+        if isinstance(self._label, str):
+            label_columns = [self._label]
+        else:
+            label_columns = list(self._label)
+        # Get the training and tuning data as pandas dataframe
+        if not isinstance(train_data, pd.DataFrame):
```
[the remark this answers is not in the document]
sxjscience: I think it should be fine for that and we can leave it to load_pd.load().

text/src/autogluon/text/text_prediction/predictor/predictor.py
sxjscience: I think numpy is good enough and there is no need to convert to pandas for this actually. I really think that users can do it by themselves.

text/src/autogluon/text/text_prediction/predictor/predictor.py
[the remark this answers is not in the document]
sxjscience: Let me rename it for now.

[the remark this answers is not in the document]
sxjscience: After a second thought, I think using a separate name may enhance the backward-compatibility.

text/src/autogluon/text/text_prediction/predictor/predictor.py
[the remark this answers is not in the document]
sxjscience: What if the HPO algorithm itself needs to be searched?

text/src/autogluon/text/text_prediction/predictor/predictor.py
sxjscience: Added.

text/src/autogluon/text/text_prediction/presets.py
```diff
@@ -0,0 +1,234 @@
...
+    """
+    ret = {
+        'models': {
+            'MultimodalTextModel': {
+                'backend': 'gluonnlp_v0',
+                'search_space': {
+                    'model.backbone.name': 'google_electra_base',
+                    'optimization.batch_size': 128,
+                    'optimization.per_device_batch_size': 8,
+                    'optimization.num_train_epochs': 10,
```
jwmueller, line 34: Shouldn't `num_train_epochs` be increased for `'best_quality'` preset?
In the future, we should add a `early_stopping_patience_epochs` hyperparameter instead and allow `num_train_epochs` to be much larger by default.

Innixma: I believe the reasoning is it isn't easy to adjust the triangular learning rate, and having a high num_train_epochs would impact the triangular learning rate. Ideally post v0.1 we want to have LR cap off at a certain point and then stay constant to enable more epochs without impacting most training runs.

sxjscience: Ideally we should actually set to some values like 50 for number of epochs but it may take really long for training some models.

sxjscience: Let me change to train for 20 epochs for best quality?

text/src/autogluon/text/text_prediction/predictor/predictor.py
```diff
@@ -0,0 +1,419 @@
...
+            search space and the configuration of the network.
+        column_types
+            The provided type of the columns. It will be a dictionary that maps the column name
+            to the type of the column.
+        num_cpus
+            The number of CPUs to use for each trial
+        num_gpus
+            The number of GPUs to use for each trial
+        num_trials
+            The number of trials. By default, we will use the provided number of trials in the
```
jwmueller, line 178: Docs should clarify this is the number of hyperparameter-tuning trials.
Also should clarify if this overrides `hyperparameters['tune_kwargs']['num_trials']`

text/src/autogluon/text/text_prediction/predictor/predictor.py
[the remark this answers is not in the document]
jwmueller: At least you should print informative error message if `['models']['MultimodalTextModel']` are missing from `hyperparameters`

text/src/autogluon/text/text_prediction/predictor/predictor.py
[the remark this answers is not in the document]
sxjscience: I won't change the name for this PR and I think it is more reasonable to later add another key to ensure backward incompatibility.

text/src/autogluon/text/text_prediction/predictor/predictor.py
```diff
@@ -0,0 +1,382 @@
...
+            The time limits
+        presets
+            The user can specify the presets of the hyper-parameters.
+        hyperparameters
+            The hyper-parameters
+        feature_columns
+            Specify which columns in the data
+        column_types
+            The provided type of the columns
+        num_cpus
```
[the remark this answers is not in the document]
sxjscience: I think it's reasonable to directly treat them as arguments in `.fit()`. Basically, it's convenient to set these hparams.

text/src/autogluon/text/text_prediction/predictor/predictor.py
```diff
@@ -0,0 +1,382 @@
...
+            The number of CPUs to use for each trial
+        num_gpus
+            The number of GPUs to use for each trial
+        num_trials
+            The number of trials. By default, we will use the provided number of trials in the
+            hyperparameters or presets. This will overwrite the provided value.
+        plot_results
```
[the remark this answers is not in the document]
sxjscience: I think this is fine to be changed later so I won't change it now.

text/src/autogluon/text/text_prediction/predictor/predictor.py
```diff
@@ -0,0 +1,382 @@
...
+                assert col in train_data.columns,\
+                    'Feature columns must be in the pandas dataframe! Received col = "{}", ' \
+                    'all columns = "{}"'.format(col, train_data.columns)
+            all_columns = feature_columns + label_columns
+        train_data = train_data[all_columns]
+        # Get tuning data
+        if tuning_data is not None:
+            if not isinstance(tuning_data, pd.DataFrame):
+                tuning_data = load_pd.load(tuning_data)
+            tuning_data = tuning_data[all_columns]
```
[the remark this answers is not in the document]
sxjscience: I think relying on the default pandas raise mechanism is good enough. Basically, the user can know that they are not using the correct dataframe. We can even have issues for that.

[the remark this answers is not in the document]
sxjscience: For open source projects, the user can actually rely on some other resources for debugging, e.g., Github issues, stackoverflow and there is no need to have nice message for all these exceptions.

text/src/autogluon/text/text_prediction/predictor/predictor.py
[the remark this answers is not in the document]
sxjscience: It usually won't happen according to the current design. We are always recommending the users to build the hyperparameters based on some presets.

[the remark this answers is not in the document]
sxjscience: Nevertheless, I added one check to raise a more meaningful message.

text/src/autogluon/text/text_prediction/predictor/predictor.py
```diff
@@ -0,0 +1,382 @@
...
+        -------
+        output
+            Array of predictions. One element corresponds to the prediction value of one
+        as_pandas
+            Whether to convert the output to a pandas dataframe
+        """
+        assert self._model is not None, 'Model does not seem to have been constructed. Have you called fit(), or load()?'
+        output = self._model.predict(dataset)
+        if as_pandas:
+            output = pd.DataFrame({self.label: output})[self.label]
```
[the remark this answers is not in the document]
sxjscience: It's sliced to be a Series.

text/src/autogluon/text/text_prediction/predictor/predictor.py
```diff
@@ -0,0 +1,382 @@
...
+        return output
+
+    def predict_proba(self, dataset, as_pandas=False):
+        """Predict the probability from the input
+
+        Parameters
+        ----------
+        dataset
```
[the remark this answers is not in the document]
sxjscience: Changed.

jwmueller (approved):
For me, is now in good enough state to merge (once CI passes). 
But I recommend followup PRs before v0.1 to address some of the other remaining comments, and I can help improve docs/tutorials in a separate PR.

szha:
Job PR-936-131 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-936/131/index.html

PR #937, opened 2021-02-18
Tabular: Refactored distillation

Innixma:
*Issue #, if available:*

*Description of changes:*

- Refactored distillation
- Now distillation also creates a weighted ensemble
- Fixed rare bug related to NaN validation scores in Neural Network

### DIFF
--- a/tabular/src/autogluon/tabular/trainer/model_presets/presets.py
+++ b/tabular/src/autogluon/tabular/trainer/model_presets/presets.py
@@ -299,59 +299,31 @@ def model_factory(
 
 
 # TODO: v0.1 cleanup and avoid hardcoded logic with model names
-def get_preset_models_softclass(path, hyperparameters, feature_metadata, num_classes=None, name_suffix='', ag_args=None, invalid_model_names: list = None):
+def get_preset_models_softclass(hyperparameters, invalid_model_names: list = None, **kwargs):
     # TODO v0.1: This import depends on mxnet, consider refactoring to avoid mxnet
     from autogluon.core.metrics.softclass_metrics import soft_log_loss
-    model_types_standard = ['GBM', 'NN', 'CAT']
+    model_types_standard = ['GBM', 'NN', 'CAT', 'ENS_WEIGHTED']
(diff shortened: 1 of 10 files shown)


szha:
Job PR-937-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-937/1/index.html

tabular/src/autogluon/tabular/trainer/model_presets/presets.py
```diff
@@ -299,59 +299,31 @@ def model_factory(
...
+    models, model_args_fit = get_preset_models(problem_type=SOFTCLASS, eval_metric=soft_log_loss,
+                                               hyperparameters=hyperparameters_standard,
+                                               default_priorities=DEFAULT_SOFTCLASS_PRIORITY, invalid_model_names=invalid_model_names, **kwargs)
     if len(models) == 0:
         raise ValueError("At least one of the following model-types must be present in hyperparameters: ['GBM','CAT','NN','RF'], "
                          "These are the only supported models for softclass prediction problems. "
                          "Softclass problems are also not yet supported for fit() with per-stack level hyperparameters.")
     for model in models:
-        model.normalize_pred_probas = True
+        model.normalize_pred_probas = True  # FIXME: Do we need to do this for child models too?
```
jwmueller, line 327: I believe so unless they inherit this from parent model

jwmueller (approved):
looks good! If we're going to automatically construct weighted ensemble in distill(), could it also be worth automatically constructing alternative ensembles in distill() as well via: `fit_weighted_ensemble(base_models = DSTL_models, expand_pareto_frontier=True)`

Innixma:
> looks good! If we're going to automatically construct weighted ensemble in distill(), could it also be worth automatically constructing alternative ensembles in distill() as well via: `fit_weighted_ensemble(base_models = DSTL_models, expand_pareto_frontier=True)`

I had the same idea, but we can leave that for a future PR. It is definitely worth adding as an option.

PR #939, opened 2021-02-18
fix GPU bug in mxnet neural net for regression

jwmueller:
Fix for bug in mxnet tabular neural net that appeared when GPU was used for certain regression tasks


szha:
Job PR-939-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-939/1/index.html

gradientsky (approved):

Innixma (approved):

PR #940, opened 2021-02-18
Part 1 of Tabular pkl compression PR: https://github.com/awslabs/autogluon/pull/654 

adrienatallah:
*[Issue # 581](https://github.com/awslabs/autogluon/issues/581)*

*Description of changes:*
In an effort to break up the larger, outdated [Tabular pkl compression PR](https://github.com/awslabs/autogluon/pull/654) into smaller chunks, this PR takes the first step by adding `compression_fn` and `compression_fn_kwargs` to `load_pkl.py` and `save_pkl.py` as well as unit tests and utility functions.

### DIFF
--- a/core/src/autogluon/core/utils/compression_utils.py
+++ b/core/src/autogluon/core/utils/compression_utils.py
@@ -1,0 +1,33 @@
+import gzip
+import bz2
+import lzma
+
+
+compression_fn_map = {
+    None: {
+        'open': open,
+        'extension': '',
+    },
+    'gzip': {
+        'open': gzip.open,
+        'extension': 'gz',
+    },
+    'bz2': {
+        'open': bz2.open,
+        'extension': 'bz2',
+    },
+    'lzma': {
+        'open': lzma.open,
+        'extension': 'lzma',
+    },
+}
+
+def get_validated_path(filename, compression_fn=None):
+    if compression_fn is not None:
+        validated_path = f"{filename}.{compression_fn_map[compression_fn]['extension']}"
+    else:
+        validated_path = filename
+    return validated_path
+
+def get_compression_map():
+    return compression_fn_map
--- a/core/src/autogluon/core/utils/loaders/load_pkl.py
+++ b/core/src/autogluon/core/utils/loaders/load_pkl.py
@@ -1,11 +1,16 @@
 import io, logging, pickle, boto3
 
-from . import load_pointer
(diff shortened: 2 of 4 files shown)


szha:
Job PR-940-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-940/1/index.html

core/src/autogluon/core/utils/compression_utils.py
```diff
@@ -0,0 +1,33 @@
...
+        'open': bz2.open,
+        'extension': 'bz2',
+    },
+    'lzma': {
+        'open': lzma.open,
+        'extension': 'lzma',
+    },
+}
+
+def get_validated_path(filename, compression_fn=None):
```
Innixma, line 25: nit: PEP8: 2 newlines before a function def

core/src/autogluon/core/utils/compression_utils.py
```diff
@@ -0,0 +1,33 @@
...
+    if compression_fn is not None:
+        validated_path = f"{filename}.{compression_fn_map[compression_fn]['extension']}"
+    else:
+        validated_path = filename
+    return validated_path
+
+def get_compression_map():
+    return compression_fn_map
```
Innixma, line 33: PEP8: newline at EOF

core/src/autogluon/core/utils/compression_utils.py
```diff
@@ -0,0 +1,33 @@
...
+}
+
+def get_validated_path(filename, compression_fn=None):
+    if compression_fn is not None:
+        validated_path = f"{filename}.{compression_fn_map[compression_fn]['extension']}"
+    else:
+        validated_path = filename
+    return validated_path
+
+def get_compression_map():
```
Innixma, line 32: nit: PEP8: 2 newlines before a function def

Innixma (approved):
Looks good! Thanks for the contribution.

I've added some minor nit comments regarding PEP8 warnings. If you use PyCharm IDE you should be able to see them highlighted while coding. If you want to clean them up, feel free to submit a follow-up PR, otherwise I can clean them up later, it isn't urgent.

One thing we will want to think about is if we want to be renaming the files depending on the compression used. I'd somewhat lean towards us not renaming them, but we can see how it looks once we start using it in Trainer.

Issue #33, opened 2019-10-15
Tabular row weight support


Innixma:
Add support for row-wise weights in a similar fashion to LightGBM:

https://lightgbm.readthedocs.io/en/latest/Parameters.html#weight_column

We can have a dedicated weight column that is checked for presence such as '__weight', and potentially give the user the option to specify the column in the call to learner init.

With this column we would additionally need to add support for weight in all objective functions and models.

kmedved:
Thanks for this very intriguing library. i wanted to check if there was any update on allowing `sample_weights` for tabular regression problems? Sample weight support is core functionality of every major tabular machine learning library (AutoML and otherwise), e.g., sklearn, xgboost, catboost, lightgbm, ngboost, TPot, H2O, etc... It would be a major enhancement to add support.

Sample weight support is an essential part of many types of analysis (anytime you have multiple observations rolled up already, or when sensors capture multiple observations simultaneously). I work with a lot of sports data as well, and sample weights are an essential part of functionally all analyses.

Thanks again for releasing this library - looking forward to using it!

Innixma:
Hi @kmedved,

Thanks for your interest! This is on our roadmap and we intend to add support for row weights by the end of the year and will notify in this issue when progress has been made.

Best,
Nick

kmedved:
Thanks @Innixma - that's great to hear. This is a very impressive looking library - I'm getting excellent results on my data in limited testing so far, and am excited to deploy it into production.

jamescolless:
Any updates on this? Sample weights are critical in regression problems when you care more about getting certain predictions right than others! As kmedved pointed out this is already implemented in all the other major ML libraries!

Innixma:
Row weight support is planned for our next release (0.1) which is scheduled for release in December.

Venge42:
Any update on this one or the 0.1 release?

Innixma:
v0.1 release has been shifted to February 2021 to allow for more polishing. Row weight support is a stretch goal for v0.1 and will be a high priority follow-up if it doesn't make the v0.1 release.

Venge42:
> v0.1 release has been shifted to February 2021 to allow for more polishing. Row weight support is a stretch goal for v0.1 and will be a high priority follow-up if it doesn't make the v0.1 release.

Great! Thank's for the update. Desperately waiting for it... :)

Thanks for the great work, guys!

Innixma:
sample_weight support is now in PR: #942

[...]

PR #942, opened 2021-02-19
sample weights functionality

jwmueller:
Related Issues: #33 

Sample-weights functionality added. Example script:

```
import numpy as np
import pandas as pd

subsample_size = 100  # subsample subset of data for faster demo, try setting this to much larger values
train_data_noweight = train_data.copy()  # training data without weights

label = 'class'
pos_class = ' >50K'
sample_weight = 'weight'

weights = np.zeros((len(train_data),))
weights[train_data[label]==pos_class] = 1  # models that can handle sample weights will only learn to predict positive class this way.
train_data[sample_weight] = weights
print(train_data.head())



## Train predictor with sample-weighting but without reporting weighted metrics:

predictor = TabularPredictor(label=label, sample_weight=sample_weight).fit(train_data)
performance = predictor.evaluate(train_data_noweight)  # weights not needed in test data because weight_evaluation not specified.
predictor.leaderboard(train_data_noweight)

## Output (models which ignore sample-weights do the best since we used extreme sample weights here but are considering original metric):

                  model  score_test  score_val  pred_time_test  pred_time_val  fit_time  pred_time_test_marginal  pred_time_val_marginal  fit_time_marginal  stack_level  can_infer  fit_order
0        KNeighborsDist        0.93       0.65        0.106977       0.107285  0.001523                 0.106977                0.107285           0.001523            0       True          6
1       NeuralNetFastAI        0.84       0.90        0.170466       0.118562  8.476372                 0.170466                0.118562           8.476372            0       True         12
2   WeightedEnsemble_L1        0.84       0.90        0.174107       0.121302  9.203398                 0.003641                0.002740           0.727026            1       True         14
3        NeuralNetMXNet        0.81       0.85        0.116029       0.023758  3.213731                 0.116029                0.023758           3.213731            0       True         11
4        KNeighborsUnif        0.75       0.70        0.108753       0.106495  0.002271                 0.108753                0.106495           0.002271            0       True          5
5               XGBoost        0.26       0.25        0.023181       0.013916  0.366666                 0.023181                0.013916           0.366666            0       True         10
6            LightGBMXT        0.26       0.25        0.024338       0.030474  0.286936                 0.024338                0.030474           0.286936            0       True          8
7         LightGBMLarge        0.26       0.25        0.028656       0.018944  0.321018                 0.028656                0.018944           0.321018            0       True         13
8              LightGBM        0.26       0.25        0.029913       0.017574  0.245766                 0.029913                0.017574           0.245766            0       True          7
9              CatBoost        0.26       0.25        0.049470       0.019500  1.075555                 0.049470                0.019500           1.075555            0       True          9
10     RandomForestEntr        0.26       0.25        0.129918       0.117492  0.571822                 0.129918                0.117492           0.571822            0       True          2
11     RandomForestGini        0.26       0.25        0.131269       0.113353  0.591314                 0.131269                0.113353           0.591314            0       True          1
12       ExtraTreesGini        0.26       0.25        0.132937       0.116305  0.472288                 0.132937                0.116305           0.472288            0       True          3
13       ExtraTreesEntr        0.26       0.25        0.134349       0.115802  0.465550                 0.134349                0.115802           0.465550            0       True          4
                  model  score_test  score_val  pred_time_test  ...  fit_time_marginal  stack_level  can_infer  fit_order
0        KNeighborsDist        0.93       0.65        0.106977  ...           0.001523            0       True          6
1       NeuralNetFastAI        0.84       0.90        0.170466  ...           8.476372            0       True         12
2   WeightedEnsemble_L1        0.84       0.90        0.174107  ...           0.727026            1       True         14
3        NeuralNetMXNet        0.81       0.85        0.116029  ...           3.213731            0       True         11
4        KNeighborsUnif        0.75       0.70        0.108753  ...           0.002271            0       True          5
5               XGBoost        0.26       0.25        0.023181  ...           0.366666            0       True         10
6            LightGBMXT        0.26       0.25        0.024338  ...           0.286936            0       True          8
7         LightGBMLarge        0.26       0.25        0.028656  ...           0.321018            0       True         13
8              LightGBM        0.26       0.25        0.029913  ...           0.245766            0       True          7
9              CatBoost        0.26       0.25        0.049470  ...           1.075555            0       True          9
10     RandomForestEntr        0.26       0.25        0.129918  ...           0.571822            0       True          2
11     RandomForestGini        0.26       0.25        0.131269  ...           0.591314            0       True          1
12       ExtraTreesGini        0.26       0.25        0.132937  ...           0.472288            0       True          3
13       ExtraTreesEntr        0.26       0.25        0.134349  ...           0.465550            0       True          4

## Train predictor with sample-weighting and report weighted metrics:

predictor = TabularPredictor(label=label, sample_weight=sample_weight, weight_evaluation=True).fit(train_data)
performance = predictor.evaluate(train_data)  # weights are needed now in the test data
predictor.leaderboard(train_data)

## Output (models which ignore sample-weights do worse now since the metric we care about is now similarly weighted):

                  model  score_test  score_val  pred_time_test  ...  fit_time_marginal  stack_level  can_infer  fit_order
0               XGBoost    1.000000        1.0        0.016571  ...           0.217509            0       True         10
1         LightGBMLarge    1.000000        1.0        0.018913  ...           0.639915            0       True         13
2              CatBoost    1.000000        1.0        0.019209  ...           1.001294            0       True          9
3            LightGBMXT    1.000000        1.0        0.019859  ...           0.250556            0       True          8
4              LightGBM    1.000000        1.0        0.022115  ...           0.320440            0       True          7
5   WeightedEnsemble_L1    1.000000        1.0        0.023454  ...           1.200252            1       True         14
6        ExtraTreesGini    1.000000        1.0        0.123344  ...           0.547743            0       True          3
7        ExtraTreesEntr    1.000000        1.0        0.129012  ...           0.532536            0       True          4
8      RandomForestEntr    1.000000        1.0        0.131815  ...           0.720166            0       True          2
9      RandomForestGini    1.000000        1.0        0.132052  ...           0.606035            0       True          1
10       KNeighborsDist    0.846154        0.2        0.109156  ...           0.001933            0       True          6
11      NeuralNetFastAI    0.846154        0.6        0.186419  ...           7.150270            0       True         12
12       KNeighborsUnif    0.192308        0.0        0.109623  ...           0.002280            0       True          5
13       NeuralNetMXNet    0.192308        0.4        0.120182  ...           2.294746            0       True         11
```

### DIFF
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -267,6 +267,9 @@ class AbstractModel:
             raise NoValidFeatures
 
     def _preprocess_fit_args(self, **kwargs):
+        sample_weight = kwargs.get('sample_weight', None)
+        if sample_weight is not None and isinstance(sample_weight, str):
+            raise ValueError("In model.fit(), sample_weight should be array of sample weight values, not string.")
         time_limit = kwargs.get('time_limit', None)
         max_time_limit_ratio = self.params_aux.get('max_time_limit_ratio', 1)
         if time_limit is not None:
@@ -346,30 +349,24 @@ class AbstractModel:
         else:
             return y_pred_proba[:, 1]
 
-    def score(self, X, y, metric=None, sample_weights=None, **kwargs):
+    def score(self, X, y, metric=None, sample_weight=None, **kwargs):
         if metric is None:
             metric = self.eval_metric
         if metric.needs_pred:
             y_pred = self.predict(X=X, **kwargs)
         else:
             y_pred = self.predict_proba(X=X, **kwargs)
-        if sample_weights is None:
-            return metric(y, y_pred)
-        else:
-            return compute_weighted_metric(y, y_pred, metric, sample_weights)
+        return compute_weighted_metric(y, y_pred, metric, sample_weight)
 
 
-    def score_with_y_pred_proba(self, y, y_pred_proba, metric=None, sample_weights=None):
+    def score_with_y_pred_proba(self, y, y_pred_proba, metric=None, sample_weight=None):
         if metric is None:
             metric = self.eval_metric
         if metric.needs_pred:
             y_pred = get_pred_from_proba(y_pred_proba=y_pred_proba, problem_type=self.problem_type)
         else:
             y_pred = y_pred_proba
-        if sample_weights is None:
-            return metric(y, y_pred)
-        else:
-            return compute_weighted_metric(y, y_pred, metric, sample_weights)
+        return compute_weighted_metric(y, y_pred, metric, sample_weight)
 
     def save(self, path: str = None, verbose=True) -> str:
         """
@@ -562,8 +559,8 @@ class AbstractModel:
                     logger.log(15, f"{hyperparam}:   {params_copy[hyperparam]}")
 
         fit_kwargs=scheduler_params['resource'].copy()
-        fit_kwargs['sample_weights'] = kwargs.get('sample_weights', None)
-        fit_kwargs['sample_weights_val'] = kwargs.get('sample_weights_val', None)
+        fit_kwargs['sample_weight'] = kwargs.get('sample_weight', None)
+        fit_kwargs['sample_weight_val'] = kwargs.get('sample_weight_val', None)
         util_args = dict(
             dataset_train_filename=dataset_train_filename,
             dataset_val_filename=dataset_val_filename,
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -18,7 +18,7 @@ from ...features.feature_metadata import FeatureMetadata
 from ...features.types import R_CATEGORY, R_OBJECT, R_FLOAT, R_INT
 from ...scheduler import FIFOScheduler
 from ...task.base import BasePredictor
-from ...utils import get_cpu_count, get_pred_from_proba, normalize_pred_probas, infer_eval_metric, compute_permutation_feature_importance
+from ...utils import get_cpu_count, get_pred_from_proba, normalize_pred_probas, infer_eval_metric, compute_permutation_feature_importance, compute_weighted_metric
 from ...utils.exceptions import TimeLimitExceeded, NoValidFeatures
 from ...utils.loaders import load_pkl
 from ...utils.savers import save_json, save_pkl
@@ -346,24 +346,30 @@ class AbstractModel:
         else:
             return y_pred_proba[:, 1]
 
-    def score(self, X, y, metric=None, **kwargs):
+    def score(self, X, y, metric=None, weights=None, **kwargs):
         if metric is None:
             metric = self.eval_metric
         if metric.needs_pred:
             y_pred = self.predict(X=X, **kwargs)
+        else:
+            y_pred = self.predict_proba(X=X, **kwargs)
+        if weights is None:
             return metric(y, y_pred)
         else:
-            y_pred_proba = self.predict_proba(X=X, **kwargs)
-            return metric(y, y_pred_proba)
+            return compute_weighted_metric(y, y_pred, metric, weights)
 
-    def score_with_y_pred_proba(self, y, y_pred_proba, metric=None):
+
+    def score_with_y_pred_proba(self, y, y_pred_proba, metric=None, weights=None):
         if metric is None:
             metric = self.eval_metric
         if metric.needs_pred:
             y_pred = get_pred_from_proba(y_pred_proba=y_pred_proba, problem_type=self.problem_type)
+        else:
+            y_pred = y_pred_proba
+        if weights is None:
             return metric(y, y_pred)
         else:
-            return metric(y, y_pred_proba)
+            return compute_weighted_metric(y, y_pred, metric, weights)
 
     def save(self, path: str = None, verbose=True) -> str:
         """
@@ -555,6 +561,9 @@ class AbstractModel:
                 if isinstance(params_copy[hyperparam], Space):
                     logger.log(15, f"{hyperparam}:   {params_copy[hyperparam]}")
 
+        fit_kwargs=scheduler_params['resource'].copy()
+        fit_kwargs['weights'] = kwargs.get('weights', None)
+        fit_kwargs['weights_val'] = kwargs.get('weights_val', None)
         util_args = dict(
             dataset_train_filename=dataset_train_filename,
             dataset_val_filename=dataset_val_filename,
@@ -562,7 +571,7 @@ class AbstractModel:
             model=self,
             time_start=time_start,
             time_limit=scheduler_params['time_out'],
-            fit_kwargs=scheduler_params['resource'],
+            fit_kwargs=fit_kwargs,
         )
 
         model_trial.register_args(util_args=util_args, **params_copy)
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -346,30 +346,30 @@ class AbstractModel:
         else:
             return y_pred_proba[:, 1]
 
-    def score(self, X, y, metric=None, weights=None, **kwargs):
+    def score(self, X, y, metric=None, sample_weights=None, **kwargs):
         if metric is None:
             metric = self.eval_metric
         if metric.needs_pred:
             y_pred = self.predict(X=X, **kwargs)
         else:
             y_pred = self.predict_proba(X=X, **kwargs)
-        if weights is None:
+        if sample_weights is None:
             return metric(y, y_pred)
         else:
-            return compute_weighted_metric(y, y_pred, metric, weights)
+            return compute_weighted_metric(y, y_pred, metric, sample_weights)
 
 
-    def score_with_y_pred_proba(self, y, y_pred_proba, metric=None, weights=None):
+    def score_with_y_pred_proba(self, y, y_pred_proba, metric=None, sample_weights=None):
         if metric is None:
             metric = self.eval_metric
         if metric.needs_pred:
             y_pred = get_pred_from_proba(y_pred_proba=y_pred_proba, problem_type=self.problem_type)
         else:
             y_pred = y_pred_proba
-        if weights is None:
+        if sample_weights is None:
             return metric(y, y_pred)
         else:
-            return compute_weighted_metric(y, y_pred, metric, weights)
+            return compute_weighted_metric(y, y_pred, metric, sample_weights)
 
     def save(self, path: str = None, verbose=True) -> str:
         """
@@ -562,8 +562,8 @@ class AbstractModel:
                     logger.log(15, f"{hyperparam}:   {params_copy[hyperparam]}")
 
         fit_kwargs=scheduler_params['resource'].copy()
-        fit_kwargs['weights'] = kwargs.get('weights', None)
-        fit_kwargs['weights_val'] = kwargs.get('weights_val', None)
+        fit_kwargs['sample_weights'] = kwargs.get('sample_weights', None)
+        fit_kwargs['sample_weights_val'] = kwargs.get('sample_weights_val', None)
         util_args = dict(
             dataset_train_filename=dataset_train_filename,
             dataset_val_filename=dataset_val_filename,
--- a/core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
+++ b/core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
@@ -160,12 +160,7 @@ class BaggedEnsembleModel(AbstractModel):
             self._add_child_times_to_bag(model=model_base)
             return
 
-        weights = kwargs.get('weights', None)
-        if weights is not None:
-            temp_weight_name = '__sample_weights__'
-            while temp_weight_name in X_train.columns:
-                temp_weight_name += '_'
-            X_train[temp_weight_name] = weights
+        sample_weights = kwargs.get('sample_weights', None)
         # TODO: Preprocess data here instead of repeatedly
         kfolds = generate_kfold(X=X_train, y=y_train, n_splits=k_fold, stratified=self.is_stratified(), random_state=self._random_state, n_repeats=n_repeats)
 
@@ -205,12 +200,11 @@ class BaggedEnsembleModel(AbstractModel):
                 fold_model = copy.deepcopy(model_base)
                 fold_model.name = f'{fold_model.name}S{j+1}F{fold_num_in_repeat+1}'  # S5F3 = 3rd fold of the 5th repeat set
                 fold_model.set_contexts(self.path + fold_model.name + os.path.sep)
-                if weights is not None:
-                    X_train_fold, weights_train = extract_column(X_train_fold, temp_weight_name)
-                    X_val_fold, weights_val = extract_column(X_val_fold, temp_weight_name)
-                    kwargs['weights'] = weights_train
-                    kwargs['weights_val'] = weights_val
-                fold_model.fit(X_train=X_train_fold, y_train=y_train_fold, X_val=X_val_fold, y_val=y_val_fold, time_limit=time_limit_fold, **kwargs)
+                kwargs_fold = kwargs.copy()
+                if sample_weights is not None:
+                    kwargs_fold['sample_weights'] = sample_weights[train_index]
+                    kwargs_fold['sample_weights_val'] = sample_weights[val_index]
+                fold_model.fit(X_train=X_train_fold, y_train=y_train_fold, X_val=X_val_fold, y_val=y_val_fold, time_limit=time_limit_fold, **kwargs_fold)
                 time_train_end_fold = time.time()
                 if time_limit is not None:  # Check to avoid unnecessarily predicting and saving a model when an Exception is going to be raised later
                     if i != (fold_end - 1):
@@ -273,16 +267,14 @@ class BaggedEnsembleModel(AbstractModel):
     def _predict_proba(self, X, normalize=False, **kwargs):
         return self.predict_proba(X=X, normalize=normalize, **kwargs)
 
-    def score_with_oof(self, y, weights=None):
+    def score_with_oof(self, y, sample_weights=None):
         self._load_oof()
         valid_indices = self._oof_pred_model_repeats > 0
         y = y[valid_indices]
         y_pred_proba = self.oof_pred_proba[valid_indices]
-        if weights is None:
-            return self.score_with_y_pred_proba(y=y, y_pred_proba=y_pred_proba)
-        else:
-            weights = weights[valid_indices]
-            return self.score_with_y_pred_proba(y=y, y_pred_proba=y_pred_proba, weights=weights)
+        if sample_weights is not None:
+            sample_weights = sample_weights[valid_indices]
+        return self.score_with_y_pred_proba(y=y, y_pred_proba=y_pred_proba, sample_weights=sample_weights)
 
     # TODO: Augment to generate OOF after shuffling each column in X (Batching), this is the fastest way.
     # TODO: v0.1 Reduce logging clutter during OOF importance calculation (Currently logs separately for each child)
--- a/core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
+++ b/core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
@@ -98,7 +98,7 @@ class BaggedEnsembleModel(AbstractModel):
         else:
             return X
 
-    def _fit(self, X_train, y_train, k_fold=5, k_fold_start=0, k_fold_end=None, n_repeats=1, n_repeat_start=0, time_limit=None, **kwargs):
+    def _fit(self, X_train, y_train, k_fold=5, k_fold_start=0, k_fold_end=None, n_repeats=1, n_repeat_start=0, time_limit=None, sample_weight=None, **kwargs):
         if k_fold < 1:
             k_fold = 1
         if k_fold_end is None:
@@ -160,7 +160,6 @@ class BaggedEnsembleModel(AbstractModel):
             self._add_child_times_to_bag(model=model_base)
             return
 
-        sample_weights = kwargs.get('sample_weights', None)
         # TODO: Preprocess data here instead of repeatedly
         kfolds = generate_kfold(X=X_train, y=y_train, n_splits=k_fold, stratified=self.is_stratified(), random_state=self._random_state, n_repeats=n_repeats)
 
@@ -201,9 +200,9 @@ class BaggedEnsembleModel(AbstractModel):
                 fold_model.name = f'{fold_model.name}S{j+1}F{fold_num_in_repeat+1}'  # S5F3 = 3rd fold of the 5th repeat set
                 fold_model.set_contexts(self.path + fold_model.name + os.path.sep)
                 kwargs_fold = kwargs.copy()
-                if sample_weights is not None:
-                    kwargs_fold['sample_weights'] = sample_weights[train_index]
-                    kwargs_fold['sample_weights_val'] = sample_weights[val_index]
+                if sample_weight is not None:
+                    kwargs_fold['sample_weight'] = sample_weight[train_index]
+                    kwargs_fold['sample_weight_val'] = sample_weight[val_index]
                 fold_model.fit(X_train=X_train_fold, y_train=y_train_fold, X_val=X_val_fold, y_val=y_val_fold, time_limit=time_limit_fold, **kwargs_fold)
                 time_train_end_fold = time.time()
                 if time_limit is not None:  # Check to avoid unnecessarily predicting and saving a model when an Exception is going to be raised later
@@ -267,14 +266,14 @@ class BaggedEnsembleModel(AbstractModel):
     def _predict_proba(self, X, normalize=False, **kwargs):
         return self.predict_proba(X=X, normalize=normalize, **kwargs)
 
-    def score_with_oof(self, y, sample_weights=None):
+    def score_with_oof(self, y, sample_weight=None):
         self._load_oof()
         valid_indices = self._oof_pred_model_repeats > 0
         y = y[valid_indices]
         y_pred_proba = self.oof_pred_proba[valid_indices]
-        if sample_weights is not None:
-            sample_weights = sample_weights[valid_indices]
-        return self.score_with_y_pred_proba(y=y, y_pred_proba=y_pred_proba, sample_weights=sample_weights)
+        if sample_weight is not None:
+            sample_weight = sample_weight[valid_indices]
+        return self.score_with_y_pred_proba(y=y, y_pred_proba=y_pred_proba, sample_weight=sample_weight)
 
     # TODO: Augment to generate OOF after shuffling each column in X (Batching), this is the fastest way.
     # TODO: v0.1 Reduce logging clutter during OOF importance calculation (Currently logs separately for each child)
--- a/core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
+++ b/core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
@@ -13,7 +13,7 @@ from ...constants import MULTICLASS, REGRESSION, SOFTCLASS, REFIT_FULL_SUFFIX
 from ...utils.exceptions import TimeLimitExceeded
 from ...utils.loaders import load_pkl
 from ...utils.savers import save_pkl
-from ...utils.utils import generate_kfold, _compute_fi_with_stddev
+from ...utils.utils import generate_kfold, _compute_fi_with_stddev, extract_column
 
 from ..abstract.abstract_model import AbstractModel
 
@@ -160,6 +160,12 @@ class BaggedEnsembleModel(AbstractModel):
             self._add_child_times_to_bag(model=model_base)
             return
 
+        weights = kwargs.get('weights', None)
+        if weights is not None:
+            temp_weight_name = '__sample_weights__'
+            while temp_weight_name in X_train.columns:
+                temp_weight_name += '_'
+            X_train[temp_weight_name] = weights
         # TODO: Preprocess data here instead of repeatedly
         kfolds = generate_kfold(X=X_train, y=y_train, n_splits=k_fold, stratified=self.is_stratified(), random_state=self._random_state, n_repeats=n_repeats)
 
@@ -199,6 +205,11 @@ class BaggedEnsembleModel(AbstractModel):
                 fold_model = copy.deepcopy(model_base)
                 fold_model.name = f'{fold_model.name}S{j+1}F{fold_num_in_repeat+1}'  # S5F3 = 3rd fold of the 5th repeat set
                 fold_model.set_contexts(self.path + fold_model.name + os.path.sep)
+                if weights is not None:
+                    X_train_fold, weights_train = extract_column(X_train_fold, temp_weight_name)
+                    X_val_fold, weights_val = extract_column(X_val_fold, temp_weight_name)
+                    kwargs['weights'] = weights_train
+                    kwargs['weights_val'] = weights_val
                 fold_model.fit(X_train=X_train_fold, y_train=y_train_fold, X_val=X_val_fold, y_val=y_val_fold, time_limit=time_limit_fold, **kwargs)
                 time_train_end_fold = time.time()
                 if time_limit is not None:  # Check to avoid unnecessarily predicting and saving a model when an Exception is going to be raised later
@@ -262,13 +273,16 @@ class BaggedEnsembleModel(AbstractModel):
     def _predict_proba(self, X, normalize=False, **kwargs):
         return self.predict_proba(X=X, normalize=normalize, **kwargs)
 
-    def score_with_oof(self, y):
+    def score_with_oof(self, y, weights=None):
         self._load_oof()
         valid_indices = self._oof_pred_model_repeats > 0
         y = y[valid_indices]
         y_pred_proba = self.oof_pred_proba[valid_indices]
-
-        return self.score_with_y_pred_proba(y=y, y_pred_proba=y_pred_proba)
+        if weights is None:
+            return self.score_with_y_pred_proba(y=y, y_pred_proba=y_pred_proba)
+        else:
+            weights = weights[valid_indices]
+            return self.score_with_y_pred_proba(y=y, y_pred_proba=y_pred_proba, weights=weights)
 
     # TODO: Augment to generate OOF after shuffling each column in X (Batching), this is the fastest way.
     # TODO: v0.1 Reduce logging clutter during OOF importance calculation (Currently logs separately for each child)
--- a/core/src/autogluon/core/models/greedy_ensemble/ensemble_selection.py
+++ b/core/src/autogluon/core/models/greedy_ensemble/ensemble_selection.py
@@ -53,7 +53,7 @@ class EnsembleSelection:
 
     # TODO: Consider having a removal stage, remove each model and see if score is affected, if improves or not effected, remove it.
     def _fit(self, predictions, labels, time_limit=None, **kwargs):
-        sample_weights = kwargs.get('weights', None)
+        sample_weights = kwargs.get('sample_weights', None)
         ensemble_size = self.ensemble_size
         self.num_input_models_ = len(predictions)
         ensemble = []
--- a/core/src/autogluon/core/models/greedy_ensemble/ensemble_selection.py
+++ b/core/src/autogluon/core/models/greedy_ensemble/ensemble_selection.py
@@ -6,7 +6,7 @@ import numpy as np
 
 from ...constants import PROBLEM_TYPES
 from ...metrics import log_loss
-from ...utils import get_pred_from_proba
+from ...utils import get_pred_from_proba, compute_weighted_metric
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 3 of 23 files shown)


szha:
Job PR-942-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-942/1/index.html

szha:
Job PR-942-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-942/3/index.html

core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
```diff
@@ -262,13 +273,16 @@ def predict_proba(self, X, normalize=None, **kwargs):
...
 
-    def score_with_oof(self, y):
+    def score_with_oof(self, y, weights=None):
         self._load_oof()
         valid_indices = self._oof_pred_model_repeats > 0
         y = y[valid_indices]
         y_pred_proba = self.oof_pred_proba[valid_indices]
-
-        return self.score_with_y_pred_proba(y=y, y_pred_proba=y_pred_proba)
+        if weights is None:
```
Innixma, line 281: Can be simplified to avoid code duplication:
```
if weights is not None:
    weights = weights[valid_indices]
return self.score_with_y_pred_proba(y=y, y_pred_proba=y_pred_proba, weights=weights)
```

core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
```diff
@@ -262,13 +273,16 @@ def predict_proba(self, X, normalize=None, **kwargs):
     def _predict_proba(self, X, normalize=False, **kwargs):
         return self.predict_proba(X=X, normalize=normalize, **kwargs)
 
-    def score_with_oof(self, y):
+    def score_with_oof(self, y, weights=None):
```
Innixma, line 276: as a general question, not sure how this is handled in other packages, but should we name it `weights, weights_val` or `w, w_val`, considering we have `y, y_val`.

core/src/autogluon/core/utils/utils.py
```diff
@@ -663,6 +680,6 @@ def get_gpu_free_memory():
         memory_free_info = _output_to_list(subprocess.check_output(COMMAND.split()))[1:]
         memory_free_values = [int(x.split()[0]) for i, x in enumerate(memory_free_info)]
     except:
-        # could fail due to 
+        # could fail due to
```
Innixma, line 683: comment fragment?

tabular/src/autogluon/tabular/learner/abstract_learner.py
```diff
@@ -138,6 +138,21 @@ def predict(self, X: DataFrame, model=None, as_pandas=False):
     def _validate_fit_input(self, X: DataFrame, **kwargs):
         if self.label not in X.columns:
             raise KeyError(f"Label column '{self.label}' is missing from training data. Training data columns: {list(X.columns)}")
+        self.weight_column = kwargs.get('weight_column', None)
```
Innixma, line 141: I think it might be cleaner to just have `sample_weight` be the name since it is widely used in other packages. We can simply explain that it must be a str column name via ValueError:

```
if sample_weight is not None and not isinstance(sample_weight, str):
    raise ValueError(f'sample_weight must be of string type indicating the column name of sample weight in the data. Instead, sample_weight was of type: {type(sample_weight)}')
```

tabular/src/autogluon/tabular/learner/abstract_learner.py
```diff
@@ -138,6 +138,21 @@ def predict(self, X: DataFrame, model=None, as_pandas=False):
...
+            if self.weight_column not in X.columns:
+                raise KeyError(f"Weight column '{self.weight_column}' is missing from training data. Training data columns: {list(X.columns)}")
+            weight_vals = X[self.weight_column]
+            if weight_vals.isna().sum() > 0:
+                raise ValueError(f"Sample weights in column '{self.weight_column}' cannot be nan")
+            if weight_vals.dtype.kind not in 'biuf':
+                raise ValueError(f"Sample weights in column '{self.weight_column}' must be numeric values")
+            if weight_vals.min() < 0:
+                raise ValueError(f"Sample weights in column '{self.weight_column}' must be nonnegative")
+            if self.weight_evaluation and X_val is not None and self.weight_column not in X_val.columns:
```
Innixma, line 154: Does it make any sense to have sample_weight in unlabelled data?

tabular/src/autogluon/tabular/learner/abstract_learner.py
```diff
@@ -232,10 +254,15 @@ def score_debug(self, X: DataFrame, y=None, extra_info=False, compute_oracle=Fal
         for model_name, y_pred_proba_internal in model_pred_proba_dict.items():
             if self.eval_metric.needs_pred:
                 y_pred = self.label_cleaner.inverse_transform_proba(y_pred_proba_internal, as_pred=True)
+            else:
+                y_pred = self.label_cleaner.inverse_transform_proba(y_pred_proba_internal, as_pred=False)
+                y = y_internal
```
Innixma, line 259: bug: this is changing y in outer scope. Should use a temporary variable in here such as:

```
if ...:
    y_tmp = y
else:
    y_tmp = y_internal
scores = self.eval_metric (y_tmp, y_pred)
```

tabular/src/autogluon/tabular/learner/default_learner.py
```diff
@@ -187,6 +192,17 @@ def general_data_processing(self, X: DataFrame, X_val: DataFrame, X_unlabeled: D
...
+                if w_val is not None:
+                    X_val[self.weight_column] = w_val
+                elif not self.weight_evaluation:
+                    nan_vals = np.empty((len(X_val),))
+                    nan_vals[:] = np.nan
+                    X_val[self.weight_column] = nan_vals
+                else:
+                    raise ValueError(f"Weight column '{self.weight_column}' cannot be missing from X_val with weight_evaluation specified.")
 
         return X, y, X_val, y_val, X_unlabeled, holdout_frac, num_bag_folds
```
Innixma, line 207: I feel that this should instead return w, w_val instead of keeping in X. We don't want trainer knowing anything about the weight column name, in the same way trainer doesn't know about the label column name.

tabular/src/autogluon/tabular/trainer/abstract_trainer.py
```diff
@@ -37,13 +37,15 @@ class AbstractTrainer:
...
                  num_classes=None, low_memory=False, feature_metadata=None, k_fold=0, n_repeats=1,
-                 save_data=False, random_state=0, verbosity=2):
+                 weight_column=None, weight_evaluation=False, save_data=False, random_state=0, verbosity=2):
         self.path = path
         self.problem_type = problem_type
         self.feature_metadata = feature_metadata
         self.save_data = save_data
         self.random_state = random_state  # Integer value added to the stack level to get the random_state for kfold splits or the train/val split if bagging is disabled
         self.verbosity = verbosity
+        self.weight_column = weight_column
```
Innixma, line 47: I don't think it is necessary for trainer to know about what the weight column name is, same way it doesn't know about label column name.

tabular/src/autogluon/tabular/trainer/abstract_trainer.py
```diff
@@ -364,6 +367,12 @@ def stack_new_level_aux(self, X, y, base_model_names: List[str], level, fit=True
         Auxiliary models never use the original features and only train with the predictions of other models as features.
         """
         X_train_stack_preds = self.get_inputs_to_stacker(X, base_models=base_model_names, fit=fit, use_orig_features=False)
+        if self.weight_evaluation:
+            X, w = extract_column(X, self.weight_column)
+            if w is not None:
+                sample_weights = w.values/w.sum() * len(w)
+                X_train_stack_preds[self.weight_column] = sample_weights
```
Innixma, line 374: I don't think weight column should impact this function. Shouldn't be necessary if w is its own variable.

tabular/src/autogluon/tabular/trainer/abstract_trainer.py
```diff
@@ -600,6 +618,7 @@ def refit_single_full(self, X=None, y=None, X_val=None, y_val=None, X_unlabeled=
                 if issubclass(stacker_type, WeightedEnsembleModel):
                     # TODO: Technically we don't need to re-train the weighted ensemble, we could just copy the original and re-use the weights.
                     if self.bagged_mode:
+                        # TODO: add sample_weights here
```
Innixma, line 621: Marking as TODO to ensure this is updated. I believe this mostly fixes itself if we save/load w separately, like we do with y.

tabular/src/autogluon/tabular/trainer/abstract_trainer.py
```diff
@@ -1019,6 +1047,19 @@ def _train_single_full(self, X_train, y_train, model: AbstractModel, X_unlabeled
             time_limit=time_limit,
             verbosity=self.verbosity,
         )
+        if self.weight_column is not None:
+            X_train, w_train = extract_column(X_train, self.weight_column)
+            if w_train is not None:  # may be None for ensemble
+                weights_normalized = w_train.values/w_train.sum() * len(w_train)  # normalization can affect gradient algorithms like boosting
```
Innixma, line 1053: Why here? Wouldn't it make more sense to do this in the actual model code? Such as in the generic `model.fit` function that is applied before going to a model's `model._fit` implementation?

tabular/src/autogluon/tabular/trainer/abstract_trainer.py
```diff
@@ -1946,7 +1987,7 @@ def distill(self, X_train=None, y_train=None, X_val=None, y_val=None, X_unlabele
 
         distilled_model_names = []
         for model_name in models:  # finally measure original metric on validation data and overwrite stored val_scores
-            model_score = self.score(X_val, y_val_og, model=model_name)
+            model_score = self.score(X_val, y_val_og, model=model_name)  # TODO: consider if weight_evaluation may be appropriate.
```
Innixma, line 1990: Marking as TODO

tabular/src/autogluon/tabular/models/xgboost/xgboost_model.py
```diff
@@ -76,7 +76,9 @@ def _fit(self, X_train, y_train, X_val=None, y_val=None, time_limit=None, num_gp
         else:
             verbose = True
             verbose_eval = 1
-        
+
+        weights = kwargs.get('weights', None)
+        weights_val = kwargs.get('weights_val', None)
```
Innixma, line 81: log about weights_val not being used?

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -521,6 +521,13 @@ def fit(self,
                 data. e.g. 500 hand-labeled samples (perhaps a hard human task), whole data set (unlabeled) is thousands/millions.
                 However, this isn't the only use case. Given enough unlabeled data(millions of rows), you may see improvements
                 to any amount of labeled data.
+            weight_column : str, default = None
```
Innixma, line 524: I'm in favor of renaming to sample_weight to align with other packages. I think the fact we use a str instead of the actual values isn't too big of a deal.

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -621,6 +630,14 @@ def fit(self,
         if holdout_frac is None:
             holdout_frac = default_holdout_frac(len(train_data), ag_args.get('hyperparameter_tune_kwargs', None) is not None)
 
+        if self.weight_column is not None:
+            prefix = f"Values in column '{self.weight_column}' used as sample weights instead of predictive features."
+            if self.weight_evaluation:
+                suffix = " Evaluation will report weighted metrics, so ensure same column exists in test data."
+            else:
+                suffix = " Evaluation metrics will ignore sample weights, specify weight_evaluation to instead report weighted metrics."
```
Innixma, line 638: `weight_evaluation=True`

tabular/src/autogluon/tabular/predictor/predictor.py
Innixma, line 524: It might make sense for this to actually live in init:

i.e.:

```
predictor = TabularPredictor(label='class', sample_weight='__weight__', weight_evaluation=True, eval_metric='log_loss').fit(train_data)
```

Additionally, these parameters can be sent to Learner init instead of Learner fit.

Innixma: Also, I think we should support `sample_weight='auto'`, where we automatically set weights to whatever we think will get the best score for the eval_metric.

core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
Innixma: I'm in favor of w, w_val internally, since it is shorter while still clear.

tabular/src/autogluon/tabular/models/tabular_nn/tabular_nn_model.py
```diff
@@ -166,8 +166,12 @@ def _fit(self, X_train, y_train, X_val=None, y_val=None, time_limit=None, num_cp
         start_time = time.time()
         try_import_mxnet()
         import mxnet as mx
-        params = self.params.copy()
         self.verbosity = kwargs.get('verbosity', 2)
+        weights = kwargs.get('weights', None)
```
Innixma, line 170: as a future follow-up, how much work would it be to use weights in NN models?

tabular/src/autogluon/tabular/learner/abstract_learner.py
jwmueller: I don't think at this time, can add support once there is method that can exploit it.

tabular/src/autogluon/tabular/models/xgboost/xgboost_model.py
jwmueller: will add TODO instead, it can be used when scoring this model, so saying it's unused is inappropriate.

tabular/src/autogluon/tabular/learner/abstract_learner.py
jwmueller: good catch!

tabular/src/autogluon/tabular/models/tabular_nn/tabular_nn_model.py
jwmueller: added TODOs for neural net models. Not sure how hard it will be for fastAI neural network. Is relatively straightforward for MXNet neural network in future PR, but will likely require extensive benchmarking to catch edge cases.

core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
jwmueller: Naming is now: `sample_weights, sample_weights_val` for the model fit kwargs, and left as it was for internal trainer/learner functions to be updated as we see fit when documenting those functions.

tabular/src/autogluon/tabular/predictor/predictor.py
jwmueller: renamed, and added TODO to support 'auto' in the future

tabular/src/autogluon/tabular/trainer/abstract_trainer.py
jwmueller: Will keep current internal implementation with weights bundled inside X for now as discussed. Future contributors may otherwise not remember to handle sample-weights when they are adding new functionality, so alternative may need careful design.

szha:
Job PR-942-4 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-942/4/index.html

tabular/src/autogluon/tabular/learner/default_learner.py
jwmueller: Will keep current internal implementation with weights bundled inside X for now as discussed. Future contributors may otherwise not remember to handle sample-weights when they are adding new functionality, so alternative may need careful design. Added TODO for this issue in Trainer.

tabular/src/autogluon/tabular/trainer/abstract_trainer.py
jwmueller: Added TODO to consider not bundling weights inside of X in future internal refactor

tabular/src/autogluon/tabular/trainer/abstract_trainer.py
jwmueller: Added TODO to consider moving this into AbstractModel.fit().
However I'm not sure why it's preferred there, this weight normalization is not really part of the model fitting. Also it would look a bit odd to add this in current AbstractModel.fit() which does not even list `X_train` as an argument right now.

tabular/src/autogluon/tabular/trainer/abstract_trainer.py
jwmueller: addressed

tabular/src/autogluon/tabular/trainer/abstract_trainer.py
jwmueller: addressed

jwmueller:
@Innixma  Addressed comments, except for ones we discussed can be done later

szha:
Job PR-942-6 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-942/6/index.html

Innixma:
@jwmueller Could you rebase when you get a chance? f1 fix PR had a minor conflict since it updated scoring functions in Learner.

core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
```diff
@@ -199,7 +200,11 @@ def _fit(self, X_train, y_train, k_fold=5, k_fold_start=0, k_fold_end=None, n_re
                 fold_model = copy.deepcopy(model_base)
                 fold_model.name = f'{fold_model.name}S{j+1}F{fold_num_in_repeat+1}'  # S5F3 = 3rd fold of the 5th repeat set
                 fold_model.set_contexts(self.path + fold_model.name + os.path.sep)
-                fold_model.fit(X_train=X_train_fold, y_train=y_train_fold, X_val=X_val_fold, y_val=y_val_fold, time_limit=time_limit_fold, **kwargs)
+                kwargs_fold = kwargs.copy()
+                if sample_weights is not None:
+                    kwargs_fold['sample_weights'] = sample_weights[train_index]
```
Innixma, line 205: bug:

`sample_weights.iloc`, otherwise could be misaligned:

```
X_train_fold, X_val_fold = X_train.iloc[train_index, :], X_train.iloc[val_index, :]
y_train_fold, y_val_fold = y_train.iloc[train_index], y_train.iloc[val_index]
```

core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
```diff
@@ -160,6 +160,7 @@ def _fit(self, X_train, y_train, k_fold=5, k_fold_start=0, k_fold_end=None, n_re
             self._add_child_times_to_bag(model=model_base)
             return
 
+        sample_weights = kwargs.get('sample_weights', None)
```
Innixma, line 163: `sample_weight` , `sample_weight_val` instead of `sample_weights`, `sample_weights_val`: we call it `sample_weight` elsewhere, and it is `sample_weight` in sklearn.

This comment applies to all usages of `sample_weights`

core/src/autogluon/core/models/greedy_ensemble/ensemble_selection.py
```diff
@@ -45,14 +45,15 @@ def fit(self, predictions, labels, time_limit=None, identifiers=None):
...
-        self._fit(predictions=predictions, labels=labels, time_limit=time_limit)
+        self._fit(predictions=predictions, labels=labels, time_limit=time_limit, **kwargs)
         self._calculate_weights()
         logger.log(15, 'Ensemble weights: ')
         logger.log(15, self.weights_)
         return self
 
     # TODO: Consider having a removal stage, remove each model and see if score is affected, if improves or not effected, remove it.
-    def _fit(self, predictions, labels, time_limit=None):
+    def _fit(self, predictions, labels, time_limit=None, **kwargs):
```
Innixma, line 55: general coding practice:

I think it makes sense to not use kwargs here, since sample_weight is the only valid kwarg, and it isn't supposed to receive extra arguments beyond that.

core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
Innixma, line 163: For clarity, explicitly have `sample_weight` as an argument. This is to show what optional functionality each model uses in fit, and kwargs is generally ignored. This applies to all models that use `sample_weight` and `sample_weight_val`.

core/src/autogluon/core/models/greedy_ensemble/greedy_weighted_ensemble_model.py
```diff
@@ -38,7 +38,7 @@ def _fit(self, X_train, y_train, X_val=None, y_val=None, time_limit=None, **kwar
         X_train = self.preprocess(X_train)
 
         self.model = self.model_base(ensemble_size=self.params['ensemble_size'], problem_type=self.problem_type, metric=self.stopping_metric)
-        self.model = self.model.fit(X_train, y_train, time_limit=time_limit)
+        self.model = self.model.fit(X_train, y_train, time_limit=time_limit, **kwargs)
```
Innixma, line 41: shouldn't pass kwargs here. just pass the minimum necessary (`sample_weight`)

tabular/src/autogluon/tabular/learner/abstract_learner.py
```diff
@@ -138,6 +140,23 @@ def predict(self, X: DataFrame, model=None, as_pandas=False):
     def _validate_fit_input(self, X: DataFrame, **kwargs):
         if self.label not in X.columns:
             raise KeyError(f"Label column '{self.label}' is missing from training data. Training data columns: {list(X.columns)}")
+        self.sample_weight = kwargs.get('sample_weight', None)
+        self.weight_evaluation = kwargs.get('weight_evaluation', False)
+        X_val = kwargs.get('X_val', None)
+        if self.sample_weight is not None:
+            if not isinstance(self.sample_weight, str):
+                raise ValueError("sample_weight must be a string indicating the name of column that contains sample weights. If you have a vector of sample weights, first add these as an extra column to your data.")
```
Innixma, line 148: Consider moving this check to init, so it is raised immediately for user.

tabular/src/autogluon/tabular/learner/abstract_learner.py
```diff
@@ -138,6 +140,23 @@ def predict(self, X: DataFrame, model=None, as_pandas=False):
...
+            if self.sample_weight not in X.columns:
+                raise KeyError(f"sample_weight column '{self.sample_weight}' is missing from training data. Training data columns: {list(X.columns)}")
+            weight_vals = X[self.sample_weight]
+            if weight_vals.isna().sum() > 0:
```
Innixma, line 152: given how many checks we have here, probably good to move to its own private method to make code cleaner.

tabular/src/autogluon/tabular/learner/abstract_learner.py
```diff
@@ -232,10 +258,16 @@ def score_debug(self, X: DataFrame, y=None, extra_info=False, compute_oracle=Fal
...
+                y_pred = self.label_cleaner.inverse_transform_proba(y_pred_proba_internal, as_pred=False)
+                y_tmp = y_internal
+            if not self.weight_evaluation:
+                scores[model_name] = self.eval_metric(y_tmp, y_pred)
             else:
-                y_pred_proba = self.label_cleaner.inverse_transform_proba(y_pred_proba_internal, as_pred=False)
-                scores[model_name] = self.eval_metric(y_internal, y_pred_proba)
+                if sample_weights is None:
+                    raise ValueError("Sample weights cannot be None when weight_evaluation is specified.")
+                scores[model_name] = compute_weighted_metric(y_tmp, y_pred, self.eval_metric, sample_weights)
```
Innixma, line 270: probably cleaner if this function is modified to also work if sample_weight is None, so we can simplify most of this code block to just a single call to

```
compute_weighted_metric(y_tmp, y_pred, self.eval_metric, sample_weights, self.weight_evaluation)
```

That would be a drop-in replacement for lines 265-270

tabular/src/autogluon/tabular/learner/default_learner.py
```diff
@@ -187,6 +192,17 @@ def general_data_processing(self, X: DataFrame, X_val: DataFrame, X_unlabeled: D
...
             del X_super
+        if w is not None:
+            X[self.sample_weight] = w
+            if X_val is not None:
+                if w_val is not None:
+                    X_val[self.sample_weight] = w_val
+                elif not self.weight_evaluation:
+                    nan_vals = np.empty((len(X_val),))
+                    nan_vals[:] = np.nan
+                    X_val[self.sample_weight] = nan_vals
```
Innixma, line 203: This seems possibly dangerous. Are we sure we want to keep it / have NaNs?

I'd recommend at least adding a TODO here of potentially removing sample_weight column from X_val

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -74,7 +74,14 @@ class TabularPredictor(TabularPredictorV1):
         Verbosity levels range from 0 to 4 and control how much information is printed.
         Higher levels correspond to more detailed print statements (you can set verbosity = 0 to suppress warnings).
         If using logging, you can alternatively control amount of information printed via `logger.setLevel(L)`,
-        where `L` ranges from 0 to 50 (Note: higher values of `L` correspond to fewer print statements, opposite of verbosity levels)
+        where `L` ranges from 0 to 50 (Note: higher values of `L` correspond to fewer print statements, opposite of verbosity levels).
+    sample_weight : str, default = None
+        If specified, this column-name indicates which column of the data should be treated as sample weights. This column will NOT be considered as a predictive feature.
+        Sample weights should be non-negative (and cannot be nan), with larger values indicating which rows are more important than others.
```
Innixma, line 80: Should we suggest that the user make the values sum to 1 as a best practice? Otherwise they could get inconsistencies using the sample_weight outside of predictor.

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -158,18 +165,22 @@ def __init__(
             eval_metric=None,
             path=None,
             verbosity=2,
+            sample_weight=None,
+            weight_evaluation=False,
             **kwargs
     ):
         self.verbosity = verbosity
         set_logger_verbosity(self.verbosity, logger=logger)
+        self.sample_weight = sample_weight  # TODO: add support for sample_weight='auto'
```
Innixma, line 174: also TODO: support 'balanced'

Innixma: I think we should do a follow-up PR to add 'balanced' support (and default 'auto' to 'balanced'), as it will be a nice option for users to try out without needing to understand how to construct a proper sample_weight column, and probably will be better than nothing. We can just construct based off of class frequency for binary/multiclass, and raise exception on regression.

tabular/src/autogluon/tabular/trainer/abstract_trainer.py
```diff
@@ -405,20 +414,29 @@ def get_inputs_to_model(self, model, X, model_pred_proba_dict=None, fit=False, p
...
+    def score(self, X, y, model=None, weights=None) -> float:
         if self.eval_metric.needs_pred:
-            y_pred_ensemble = self.predict(X=X, model=model)
-            return self.eval_metric(y, y_pred_ensemble)
+            y_pred = self.predict(X=X, model=model)
         else:
-            y_pred_proba_ensemble = self.predict_proba(X=X, model=model)
-            return self.eval_metric(y, y_pred_proba_ensemble)
+            y_pred = self.predict_proba(X=X, model=model)
+        if not self.weight_evaluation:
```
Innixma, line 422: same comment as earlier, these 5 lines can be simplified to 1 by expanding `compute_weighted_metric` functionality.

tabular/src/autogluon/tabular/trainer/abstract_trainer.py
```diff
@@ -405,20 +414,29 @@ def get_inputs_to_model(self, model, X, model_pred_proba_dict=None, fit=False, p
...
-    def score_with_y_pred_proba(self, y, y_pred_proba) -> float:
+
+    def score_with_y_pred_proba(self, y, y_pred_proba, weights=None) -> float:
         if self.eval_metric.needs_pred:
             y_pred = get_pred_from_proba(y_pred_proba=y_pred_proba, problem_type=self.problem_type)
-            return self.eval_metric(y, y_pred)
         else:
-            return self.eval_metric(y, y_pred_proba)
+            y_pred = y_pred_proba
+        if not self.weight_evaluation:
```
Innixma, line 434: ditto

tabular/src/autogluon/tabular/trainer/abstract_trainer.py
```diff
@@ -599,12 +617,20 @@ def refit_single_full(self, X=None, y=None, X_val=None, y_val=None, X_unlabeled=
...
+                            X, w = extract_column(X, self.sample_weight)
                         X_train_stack_preds = self.get_inputs_to_stacker(X, base_models=base_model_names, fit=True, use_orig_features=False)
                         y_input = y
                     else:
+                        if self.weight_evaluation:
+                            X_val, w = extract_column(X_val, self.sample_weight)
                         X_train_stack_preds = self.get_inputs_to_stacker(X_val, base_models=base_model_names, fit=False, use_orig_features=False)  # TODO: May want to cache this during original fit, as we do with OOF preds
                         y_input = y_val
+                    if w is not None:
+                        sample_weights = w.values/w.sum() * len(w)
```
Innixma, line 632: wouldn't this just be `w.values/w.mean()` ?

tabular/tests/unittests/test_tabular.py
```diff
@@ -486,6 +486,38 @@ def test_tabular_bag():
                            seed_val=seed_val, fit_args=fit_args)
 
 
+def test_sample_weights():
```
Innixma, line 489: rename: `test_sample_weight`

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -346,24 +346,30 @@ def _predict_proba(self, X, **kwargs):
...
 
-    def score_with_y_pred_proba(self, y, y_pred_proba, metric=None):
+    def score_with_y_pred_proba(self, y, y_pred_proba, metric=None, sample_weights=None):
         if metric is None:
             metric = self.eval_metric
         if metric.needs_pred:
             y_pred = get_pred_from_proba(y_pred_proba=y_pred_proba, problem_type=self.problem_type)
+        else:
+            y_pred = y_pred_proba
+        if sample_weights is None:
```
Innixma, line 369: ditto on other comments, simplify to 1 line

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -346,24 +346,30 @@ def _predict_proba(self, X, **kwargs):
...
 
-    def score(self, X, y, metric=None, **kwargs):
+    def score(self, X, y, metric=None, sample_weights=None, **kwargs):
         if metric is None:
             metric = self.eval_metric
         if metric.needs_pred:
             y_pred = self.predict(X=X, **kwargs)
+        else:
+            y_pred = self.predict_proba(X=X, **kwargs)
+        if sample_weights is None:
```
Innixma, line 356: ditto, simplify to 1 line

tabular/src/autogluon/tabular/trainer/abstract_trainer.py
```diff
@@ -364,6 +367,12 @@ def stack_new_level_aux(self, X, y, base_model_names: List[str], level, fit=True
         Auxiliary models never use the original features and only train with the predictions of other models as features.
         """
         X_train_stack_preds = self.get_inputs_to_stacker(X, base_models=base_model_names, fit=fit, use_orig_features=False)
+        if self.weight_evaluation:
+            X, w = extract_column(X, self.sample_weight)  # TODO: consider redesign with w as separate arg instead of bundled inside X
+            if w is not None:
+                sample_weights = w.values/w.sum() * len(w)
```
Innixma, line 373: ditto other comment on simplifying. 2 lines can also be simplified to 1

tabular/src/autogluon/tabular/trainer/abstract_trainer.py
Innixma: also, 2 lines can be made into 1

tabular/src/autogluon/tabular/trainer/abstract_trainer.py
```diff
@@ -804,7 +830,7 @@ def unpersist_models(self, model_names='all') -> list:
             logger.log(30, f'No valid persisted models were specified to be unpersisted, so no change in model persistence was performed.')
         return unpersisted_models
 
-    def generate_weighted_ensemble(self, X, y, level, base_model_names, k_fold=0, n_repeats=1, stack_name=None, hyperparameters=None, time_limit=None, name_suffix: str = None, save_bag_folds=None, check_if_best=True, child_hyperparameters=None, get_models_func=None) -> List[str]:
+    def generate_weighted_ensemble(self, X, y, level, base_model_names, k_fold=0, n_repeats=1, stack_name=None, hyperparameters=None, time_limit=None, name_suffix: str = None, save_bag_folds=None, check_if_best=True, child_hyperparameters=None, get_models_func=None, sample_weights=None) -> List[str]:
```
Innixma, line 833: sample_weights is unused?

tabular/src/autogluon/tabular/trainer/abstract_trainer.py
```diff
@@ -1019,6 +1054,20 @@ def _train_single_full(self, X_train, y_train, model: AbstractModel, X_unlabeled
             time_limit=time_limit,
             verbosity=self.verbosity,
         )
+        if self.sample_weight is not None:
+            X_train, w_train = extract_column(X_train, self.sample_weight)
+            if w_train is not None:  # may be None for ensemble
+                # TODO: consider moving weight normalization into AbstractModel.fit()
+                weights_normalized = w_train.values/w_train.sum() * len(w_train)  # normalization can affect gradient algorithms like boosting
```
Innixma, line 1061: ditto again on simplifying and 2 lines to 1

tabular/src/autogluon/tabular/trainer/abstract_trainer.py
```diff
@@ -1019,6 +1054,20 @@ def _train_single_full(self, X_train, y_train, model: AbstractModel, X_unlabeled
...
+                model_fit_kwargs['sample_weights'] = weights_normalized
+            if X_val is not None:
+                X_val, w_val = extract_column(X_val, self.sample_weight)
+                if self.weight_evaluation and w_val is not None:  # ignore validation weights unless weight_evaluation specified
+                    weights_val_normalized = w_val.values / w_val.sum() * len(w_val)
```
Innixma, line 1066: ditto

szha:
Job PR-942-7 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-942/7/index.html

tabular/src/autogluon/tabular/learner/default_learner.py
jwmueller: added TODO, this is necessary as part of the current approach where weights are bundled inside X, but can be removed is we later unbundle

core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
jwmueller: Actually sample_weight here is numpy array, so the integer indexing should already be correct

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -158,18 +166,22 @@ def __init__(
             eval_metric=None,
             path=None,
             verbosity=2,
+            sample_weight=None,
+            weight_evaluation=False,
             **kwargs
     ):
         self.verbosity = verbosity
         set_logger_verbosity(self.verbosity, logger=logger)
+        self.sample_weight = sample_weight  # TODO: add support for sample_weight= 'auto', 'balanced'
```
Innixma, line 175: sample_weight and weight_evaluation can both be properties that link to self._learner.sample_weight, self._learner.weight_evaluation. Currently it is defined twice which isn't ideal.

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -621,6 +632,14 @@ def fit(self,
         if holdout_frac is None:
             holdout_frac = default_holdout_frac(len(train_data), ag_args.get('hyperparameter_tune_kwargs', None) is not None)
 
+        if self.sample_weight is not None:
```
Innixma, line 635: Maybe this code block makes more sense to live in learner than in predictor?

Innixma (approved):
Looks good, excited to get this in! I think having an 'auto'/'balanced' setting would add a lot of value as a follow-up.

szha:
Job PR-942-8 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-942/8/index.html

tabular/src/autogluon/tabular/predictor/predictor.py
jwmueller: not exactly sure what to do here as I didn't see any examples, so just marked as TODO

Innixma: Sure, it isn't urgent. Just a minor detail that I can cleanup afterwards

szha:
Job PR-942-9 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-942/9/index.html

PR #947, opened 2021-02-20
Tabular: Fixed errors related to 'f1' metric.

Innixma:
*Issue #, if available:*

*Description of changes:*

- Fixed CatBoost crashing when time_limit != None and stopping_metric = 'f1'.
- Fixed leaderboard, evaluate, and evaluate_predictions methods crashing with 'f1' metric when raw class labels are not integers.


szha:
Job PR-947-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-947/1/index.html

jwmueller (approved):
Perhaps we also want a unit test that tries all possible metrics on a toy dataset?

Innixma:
> Perhaps we also want a unit test that tries all possible metrics on a toy dataset?

Sure, only concern is how long it will take and if we can actually test all the edge cases.

PR #948, opened 2021-02-20
Tabular: Improved exception handling

Innixma:
*Issue #, if available:*

*Description of changes:*

- Verbose exception traceback will only be printed on verbosity>2 for ImportErrors
- Replaced overly verbose logger.exception calls (that include traceback) with logger.error calls.

### DIFF
--- a/core/src/autogluon/core/utils/try_import.py
+++ b/core/src/autogluon/core/utils/try_import.py
@@ -45,9 +45,13 @@ def try_import_mxnet():
 def try_import_catboost():
     try:
         import catboost
+    except ImportError as e:
+        raise ImportError("`import catboost` failed."
+                          "A quick tip is to install via `pip install catboost`.")
     except ValueError as e:
         raise ImportError("Import catboost failed. Numpy version may be outdated, "
-                          "Please ensure numpy version >=1.16.0. If it is not, please try 'pip uninstall numpy; pip install numpy>=1.17.0' Detailed info: {}".format(str(e)))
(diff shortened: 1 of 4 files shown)


szha:
Job PR-948-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-948/1/index.html

core/src/autogluon/core/utils/try_import.py
```diff
@@ -45,34 +45,41 @@ def try_import_mxnet():
 def try_import_catboost():
     try:
         import catboost
+    except ImportError as e:
+        raise ImportError("`import catboost` failed."
+                          "A quick tip is to install via `pip install catboost`.")
     except ValueError as e:
         raise ImportError("Import catboost failed. Numpy version may be outdated, "
-                          "Please ensure numpy version >=1.16.0. If it is not, please try 'pip uninstall numpy; pip install numpy>=1.17.0' Detailed info: {}".format(str(e)))
+                          "Please ensure numpy version >=1.16.0. If it is not, please try 'pip uninstall numpy -y; pip install numpy>=1.17.0' "
```
gradientsky, line 53: Just noticed: `==1.19.5`

gradientsky (approved):

core/src/autogluon/core/utils/try_import.py
Innixma: Will keep as is for now since this is specific to catboost's minimum numpy version. Not sure the context of when this error used to occur, but doubt it will ever be triggered at this point since we require more recent numpy version.

Issue #949, opened 2021-02-21
TextSpecialFeatureGenerator post_drop_dulicatesd default True


j2schmit:
https://github.com/awslabs/autogluon/blob/63fa1da164a3243de214a84067bbfa7d5f89fc51/features/src/autogluon/features/generators/text_special.py#L30

The default value for post_drop_duplicates is True, but for large datasets this activity seems to take a lot of time. Is there a recommended pythonic way to set this as False?

Innixma:
Refer to this example of creating custom feature generators: https://github.com/awslabs/autogluon/blob/master/examples/tabular/example_custom_feature_generator.py

You would need to specify a `feature_generator = PipelineFeatureGenerator(...)` and pass it into `predictor.fit(..., feature_generator)`

Since you'd be using [PipelineFeatureGenerator](https://auto.gluon.ai/dev/api/autogluon.features.html#pipelinefeaturegenerator), you would have full control over all init args of all feature generators used, including [TextSpecialFeatureGenerator](https://auto.gluon.ai/dev/api/autogluon.features.html#textspecialfeaturegenerator). Unfortunately this is a bit involved in order to change one boolean from True to False, but this is because of the vast amount of options available in the feature generators. If you have any ideas for how to make this process (of changing any parameter of any generator) simpler, feel free to propose a solution.

Alternatively, you could update inplace the default AutoMLPipelineFeatureGenerator(), but the code would be brittle if anything changed in future:

```
from autogluon.features.generators import AutoMLPipelineFeatureGenerator, TextSpecialFeatureGenerator
feature_generator = AutoMLPipelineFeatureGenerator()
# Replace the default TextSpecialFeatureGenerator with a custom TextSpecialFeatureGenerator
feature_generator.generators[2][3] = TextSpecialFeatureGenerator(post_drop_duplicates=False)
```

As a question, how large of data do you have (rows and total size in MB), and how long is post_drop_duplicates taking to compute? I would suspect that ngram generation would take more time than drop_duplicates.

Innixma:
I decided to look into this, and I found a way to heavily optimize the DropDuplicatesFeatureGenerator, now it should be 10x-1000x+ faster once this PR is merged: #950

I believe once this optimization is merged, you shouldn't need to disable duplicate dropping.

j2schmit:
[...]

PR #950, opened 2021-02-21
Features: Optimize DropDuplicates fit speed by 100x+ on large datasets.

Innixma:
*Issue #, if available:*
#949 

*Description of changes:*

- Added sample_size_init and sample_size_final params to DropDuplicatesFeatureGenerator.
- This enables virtually identical results as before, but with 100x+ fit time speedup for large datasets (1MM+).
- 1MM rows with 100 near-duplicate and 100 exact-duplicate features gets same result, but fit time reduced from 1500s -> 10s
- Added advanced detection of duplicate categoricals, even if their class names differ superficially.

### DIFF
--- a/features/src/autogluon/features/generators/drop_duplicates.py
+++ b/features/src/autogluon/features/generators/drop_duplicates.py
@@ -1,4 +1,5 @@
 import logging
+from typing import Union
 from collections import defaultdict
 
 from pandas import DataFrame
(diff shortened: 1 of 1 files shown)


features/src/autogluon/features/generators/drop_duplicates.py
```diff
@@ -12,9 +13,35 @@
...
+    """
+    Drops features which are exact duplicates of other features, leaving only one instance of the data.
+
+    Parameters
+    ----------
+    sample_size_init : int, default 1000
+        The number of rows to sample when doing an initial filter of duplicate feature candidates.
+        Usually, the majority of features can be filtered out using this smaller amount of rows which greatly speeds up the computation of the final check.
+        If None or greater than the number of rows, no initial filter will occur. This may increase the time to fit immensely for large datasets.
+    sample_size_final : int, default 20000
```
Innixma, line 25: This could probably be even lower (say 5000), but I made it 20K to ensure that only under the most absurd situations would it incorrectly detect duplicates. We can easily drop it to 5000 if users find it to still be too slow in certain situations.

j2schmit: Great idea, I had something similar in mind, but this is better.

szha:
Job PR-950-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-950/1/index.html

features/src/autogluon/features/generators/drop_duplicates.py
```diff
@@ -29,26 +56,28 @@ def get_default_infer_features_in_args() -> dict:
         return dict()
 
     # TODO: optimize categorical/object handling
-    # TODO: Potentially use subsamples?
-    def _drop_duplicate_features(self, X: DataFrame) -> list:
+    @staticmethod
```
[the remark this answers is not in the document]
Innixma: Added support for this in most recent commit

szha:
Job PR-950-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-950/2/index.html

gradientsky (approved):

PR #952, opened 2021-02-22
Add shebang to full_install.sh

larroy:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/full_install.sh
+++ b/full_install.sh
@@ -1,3 +1,5 @@
+#!/usr/bin/env bash
+set -euo pipefail
 python3 -m pip install -e core/
 python3 -m pip install -e features/
 python3 -m pip install -e tabular/[all]


full_install.sh
```diff
@@ -1,3 +1,5 @@
+#!/usr/bin/env bash
```
Innixma, line 1: Looks good! As a question, any ideas on making this dev setup process more flexible? i.e., perhaps having a root level setup.py that installs all the submodules so someone can do:

```
git clone https://github.com/awslabs/autogluon
cd autogluon && python3 setup.py develop
```

This was the code required in v0.0.15 prior to namespace modules, but doesn't work anymore since there isn't a root setup.py

Innixma (approved):

szha:
Job PR-952-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-952/1/index.html

full_install.sh
larroy: if something comes to mind I will let you know. You can include urls in setup.py so maybe it would work to use local paths to the submodules from a parent setup.py as you describe.

Issue #782, opened 2020-11-16
v0.1 Tabular model name refinement discussion


Innixma:
For v0.1 release, we are looking to refine and improve the readability of the model names created by Tabular.

Below is a list of proposals. We are looking for community feedback, so please add your thoughts and suggestions!

1. Weighted ensemble renaming:
Current -> weighted_ensemble_k0_l2
Proposed -> WeightedEnsemble_l2
Reasoning: All other models are in CamelCase format except for the weighted ensemble. Additionally, k0 is a remnant from before we were able to handle duplicate model names. (k0 represented the special case = kfold 0, aka train with all the data and use train data as validation data as well).
Alternative 1: 'Weighted_l2'

2. Stacker model renaming:
Current -> {model_name}_STACKER_l1
Proposed -> {model_name}_S_l1
Reasoning: Stacker model names often become very long, such that they are difficult to read and type in general usage. Since we now store information that identifies a model as a stacker elsewhere (such as in .leaderboard or .info), we don't need to be as verbose in the model name itself.
Alternative 1: '_STACK_'
Alternative 2: '_B_' or '_BAG_' due to '_S_' being defined when level != 0 (level > 0 and _S_ is redundant together), and additionally _STACKER_l0 is technically misleading, as the model is not a stacker but actually a bagged ensemble.
Alternative 3: No suffix, since the state of the stack and bag can be found through other means.

3. Classifier / Regressor suffix removal:
Current -> RandomForestClassifier
Proposed -> RandomForest
Reasoning: Classifier / Regressor suffix is a remnant from when we did not have such information available elsewhere. Since it is generally well known by the user what problem type they are dealing with, the extra info added to the model names is generally redundant and complicates the code, especially when handling duplicated model name resolution.

4. Stack level rework:
Current -> {model_name}_l{n}
Proposed -> {model_name}_l{n+1}
Reasoning: level=0 is confusing to users due to an l2 model being the head of a model stack which is 3 layers deep. With this change, an l2 model would be the head of a model stack which is 2 layers deep.

5. FULL rename:
Current -> {model_name}_FULL
Proposed -> {model_name}_F
Reasoning: Same reasoning as Proposal 2 in that it contributes to very long model names.

6. DSTL rename:
Current -> {model_name}_DSTL
Proposed -> {model_name}_D
Reasoning: Same reasoning as Proposal 5.

7. CatBoost rename:
Current -> Catboost
Proposed -> CatBoost
[...]

PR #960, opened 2021-02-24
Tabular: Added +1 to all stack level values (No functional change)

Innixma:
*Issue #, if available:*
#782 -> Implements Proposal 4. 

*Description of changes:*

- Added +1 to all stack level values in code
- This means base models are now considered 'L1' instead of 'L0'. This improves clarity and aligns with common naming schemes outside of AG for multi-layer stacking.


szha:
Job PR-960-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-960/1/index.html

jwmueller (approved):
Can we provide some information about the stack layers and naming suffix on the autogluon.tabular.models docs page?

It would be especially good to explain when these are actually bagged models as well there.

Innixma:
> Can we provide some information about the stack layers and naming suffix on the autogluon.tabular.models docs page?
> 
> It would be especially good to explain when these are actually bagged models as well there.

Added GitHub issue: https://github.com/awslabs/autogluon/issues/961

PR #981, opened 2021-02-28
Improved logging verbosity when loading predictors

Innixma:
*Issue #, if available:*

*Description of changes:*

- Changed predictor loading so it doesn't alter logger verbosity by default.
- Previously TextPredictor would forcibly change root logger to verbosity 3 when loaded, now it can be controlled (and by default it won't change logger on load).

### DIFF
--- a/tabular/src/autogluon/tabular/predictor/predictor.py
+++ b/tabular/src/autogluon/tabular/predictor/predictor.py
@@ -2076,7 +2076,7 @@ class TabularPredictor:
         logger.log(20, f'TabularPredictor saved. To load, use: predictor = TabularPredictor.load("{self.path}")')
 
     @classmethod
-    def load(cls, path, verbosity=2):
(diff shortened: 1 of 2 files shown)


szha:
Job PR-981-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-981/1/index.html

sxjscience:
Apart from the change, would it make sense to also change `logger = logging.getLogger()  # return root logger` to `logger = logging.getLogger(__name__)`? So that each predictor module will use its own logger.

sxjscience (approved):

Innixma:
> Apart from the change, would it make sense to also change `logger = logging.getLogger() # return root logger` to `logger = logging.getLogger(__name__)`? So that each predictor module will use its own logger.

Probably yes, although we'd want to do that post release since I think its a bit complicated.

PR #982, opened 2021-03-01
Create pythonpublish_testpypi.yml

Innixma:
*Issue #, if available:*

*Description of changes:*

- Adds file used to manually trigger a test pypi upload simulating official release.
- This should only be trigged once during prep for release.


sxjscience (approved):
LGTM. This is a useful feature to trigger a workflow that will release to testpypi.

jwmueller (approved):
Would be good to document this in release instructions, otherwise may be overlooked in the future.

Innixma:
> Would be good to document this in release instructions, otherwise may be overlooked in the future.

Good point, added issue: https://github.com/awslabs/autogluon/issues/984

PR #1015, opened 2021-03-15
Fixed exception when providing a dataset to ENAS_Scheduler train_set …

lukemorrill:
…parameter caused by local dataset_name variable being assigned the passed in dataset and then being treated like a string

*Issue #, if available:*
New issue just found.

*Description of changes:*
Now casting the passed in "train_set" argument to function "initialize_miscs" to a string when assigning it to "dataset_name" local variable. This should leave the demo functionality with "imagenet" unchanged, but fixes the problem when passing in an actual data set. The dataset_name variable gets compared to a string on line 105, which raises an exception if the train_set argument to the function is a numpy array or Autogluon Dataset. 


szha:
Job PR-1015-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1015/1/index.html

Innixma (approved):
Thanks for the contribution! This does indeed look like a bug. Our ENAS functionality is less explored and tested than most of our other components, so fixes like these are very valuable.

lukemorrill:
Great! I'm trying it out for a work project, so I may have more contributions coming.

PR #1022, opened 2021-03-17
Tabular: Added LOO-based bagging alternative to KNN

Innixma:
*Issue #, if available:*

*Description of changes:*

Added LOO-based bagging alternative to KNN (enabled by default).

Now, during bagging KNN will only train a single model and use LOO to get unbiased training predictions for out-of-fold.

This results in a **10x-200x** reduction in training time, inference time, and disk space usage during bagging for KNN, with no statistical drop in model quality.

Code: (New version vs old version, new version is 100x faster to train/infer, takes up 100x less disk space)

```

LABEL = 'class'
SAMPLE = 10000


if __name__ == '__main__':
    path_prefix = 'https://autogluon.s3.amazonaws.com/datasets/AdultIncomeBinaryClassification/'

    train_data = TabularDataset(path_train)
    test_data = TabularDataset(path_test)

    if SAMPLE is not None:
        train_data = train_data.sample(n=SAMPLE, random_state=0)

    predictor = TabularPredictor(label=LABEL).fit(
        train_data=train_data,
        num_bag_folds=10,
        num_bag_sets=10,
        hyperparameters={
            'KNN': [
                {},
                {'ag_args_ensemble': {'use_child_oof': False}}
            ],
        },
    )

    leaderboard = predictor.leaderboard(test_data)

```

Output:

```
                 model  score_test  score_val  pred_time_test  pred_time_val  fit_time  pred_time_test_marginal  pred_time_val_marginal  fit_time_marginal  stack_level  can_infer  fit_order
0    KNeighbors_BAG_L1    0.769884     0.7653        0.114250       0.114678  0.035231                 0.114250                0.114678           0.035231            1       True          1
1  KNeighbors_2_BAG_L1    0.769884     0.7659       11.144715      10.744937  2.262136                11.144715               10.744937           2.262136            1       True          2
2  WeightedEnsemble_L2    0.769884     0.7660       11.262506      10.877384  3.217055                 0.003541                0.017769           0.919687            2       True          3
```

### DIFF
--- a/core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
+++ b/core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
@@ -143,8 +143,6 @@ class BaggedEnsembleModel(AbstractModel):
         if self._k is not None and self._k != k_fold:
             raise ValueError(f'k_fold must equal previously fit k_fold value for the current n_repeat, values: (({k_fold}, {self._k})')
 
-        time_start = time.time()
-
         model_base = self._get_model_base()
         model_base.rename(name='')
         if self.features is not None:
@@ -156,44 +154,11 @@ class BaggedEnsembleModel(AbstractModel):
             self.model_base = None
 
         if k_fold == 1:
-            if self._n_repeats != 0:
-                raise ValueError(f'n_repeats must equal 0 when fitting a single model with k_fold < 2, values: ({self._n_repeats}, {k_fold})')
-            model_base.name = f'{model_base.name}S1F1'
-            model_base.set_contexts(path_context=self.path + model_base.name + os.path.sep)
-            time_start_fit = time.time()
-            model_base.fit(X=X, y=y, time_limit=time_limit, sample_weight=sample_weight, **kwargs)
-            model_base.fit_time = time.time() - time_start_fit
-            model_base.predict_time = None
-            if use_child_oof:
-                logger.log(15, '\t`use_child_oof` was specified for this model. It will function similarly to a bagged model, but will only fit one child model.')
-                time_start_predict = time.time()
-                if model_base._get_tags().get('valid_oof', False):
-                    self._oof_pred_proba = model_base.get_oof_pred_proba(X=X)
-                else:
-                    logger.warning('\tWARNING: `use_child_oof` was specified but child model does not have a dedicated `get_oof_pred_proba` method. This model may have heavily overfit validation scores.')
-                    self._oof_pred_proba = model_base.predict_proba(X=X)
-                self._child_oof = True
-                model_base.predict_time = time.time() - time_start_predict
-                model_base.val_score = model_base.score_with_y_pred_proba(y=y, y_pred_proba=self._oof_pred_proba)
-            else:
-                self._oof_pred_proba = model_base.predict_proba(X=X)  # TODO: Cheater value, will be overfit to valid set
-            self._oof_pred_model_repeats = np.ones(shape=len(X), dtype=np.uint8)
-            self._n_repeats = 1
-            self._n_repeats_finished = 1
-            self._k_per_n_repeat = [1]
-            self._bagged_mode = False
-            model_base.reduce_memory_size(remove_fit=True, remove_info=False, requires_save=True)
-            if not self.params.get('save_bag_folds', True):
-                model_base.model = None
-            if self.low_memory:
-                self.save_child(model_base, verbose=False)
-                self.models = [model_base.name]
-            else:
-                self.models = [model_base]
-            self._add_child_times_to_bag(model=model_base)
+            self._fit_single(X=X, y=y, model_base=model_base, use_child_oof=use_child_oof, time_limit=time_limit, sample_weight=sample_weight, **kwargs)
             return
 
         # TODO: Preprocess data here instead of repeatedly
+        time_start = time.time()
         kfolds = generate_kfold(X=X, y=y, n_splits=k_fold, stratified=self.is_stratified(), random_state=self._random_state, n_repeats=n_repeats)
 
         oof_pred_proba, oof_pred_model_repeats = self._construct_empty_oof(X=X, y=y)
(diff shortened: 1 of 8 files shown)


szha:
Job PR-1022-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1022/1/index.html

Innixma:
Before merging I plan to:

- [ ] Run AutoMLBenchmark

jwmueller:
Should this new model still have the _BAG suffix in its name? It's not being bagged so that seems misleading. As I understand, this new KNeighbors model would be no different than the one we produce with single train/val split, so it should have the same name?

Innixma:
> Should this new model still have the _BAG suffix in its name? It's not being bagged so that seems misleading. As I understand, this new KNeighbors model would be no different than the one we produce with single train/val split, so it should have the same name?

It is identical to the `KNeighbors_FULL` model in reality. It uses 100% of data for train, and its OOF is from LOO. We may consider removing the `_BAG` suffix, but it would have to still have some kind of suffix, either `_FULL` or something new.

core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
```diff
@@ -39,7 +39,8 @@ def __init__(self, model_base: AbstractModel, random_state=0, **kwargs):
         self._k_per_n_repeat = []  # k-fold used for each n_repeat. == [5, 10, 3] if first kfold was 5, second was 10, and third was 3
         self._random_state = random_state
         self.low_memory = True
-        self.bagged_mode = None
+        self._bagged_mode = None
+        self._child_oof = False  # Whether the OOF preds were taken from a single child model (Assumes child can produce OOF preds without bagging).
```
jwmueller, line 43: Should there eventually be a separate class for models that can produce OOF predictions without bagging? (RF/XT could eventually also do this in addition to KNN). Seems like these could just be of the same class as our regular models (single-copy, not bagged), adding them under BaggedEnsemble might be misleading since these are neither  ensembles nor using bagging?

tabular/src/autogluon/tabular/models/knn/_knn_loo_variants.py
```diff
@@ -0,0 +1,130 @@
...
+class KNeighborsRegressor(_KNeighborsRegressor):
+    def predict_loo(self):
+        """Predict the target for the training data via leave-one-out.
+
+        Returns
+        -------
+        y : ndarray of shape (n_queries,) or (n_queries, n_outputs), dtype=int
+            Target values.
+        """
+        neigh_dist, neigh_ind = self.kneighbors()
```
jwmueller, line 109: I'm not seeing where you are discarding the closest neighbor?

core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
jwmueller: Alternatively could make BaggedEnsemble a subclass of some shared parent-class (eg. OOFModel) that's capable of producing held-out prediction for any requested datapoint?

Innixma: These are all reasonable, just will take quite a bit of thought and coding to make sure its done right.

Innixma: The main thing is we want to keep the option available to bag these models that could also get OOF, for the purposes of consistency and comparisons. Furthermore, we should have the option to get training preds from models that don't support LOO and just have the predictions be overfit. For large enough datasets, it might be the case that we can get away with this if the training vs val scores are not much different.

For these reasons, I don't think it's the right idea to subclass the models themselves. A tag + implementation of the get_oof method should suffice. I think `OOFModel` may be the direction to go, such that `BaggedEnsemble` inherits from it. This can also mean that all `_FULL` models instead become `OOFModel` instead of `BaggedEnsemble`.

Innixma: The difficulty from this is that StackerEnsembleModel inherits from BaggedEnsemble, but in fact these single models can also be stackers. Unless we have multiple StackerEnsembleModels, which is ugly, we would have to re-think how our stacker models operate.

Innixma: One option is to delete stacker ensemble model and instead move all stacking logic to trainer, therefore completely hiding stacking logic from models.

jwmueller: I agree with "tag + implementation of the get_oof method should suffice".  Perhaps with the addition of OOFModel, StackerEnsembleModel could simply inherit from OOFModel instead of BaggedEnsemble? 

Understand it'll take significant code to add OOFModel, so may be only worthwhile once we have another model like RF/XT/LR also not requiring bagging to get OOF.  At least for this PR, maybe just add comment in BaggedEnsembleModel that describes the special exception for using single child model for all OOF-predictions like in the case of KNeighbors.

Innixma: The problem with "StackerEnsembleModel could simply inherit from OOFModel instead of BaggedEnsemble" is that then BaggedEnsembleModel wouldn't be part of StackerEnsembleModel.

jwmueller: Right it may be annoying to change StackerEnsemble too much.
Another option could be to simply change the name of the class `BaggedEnsemble` to be broader like `OOFModel`, with some attribute for distinguishing whether it's based on bagging or not.

Or could keep BaggedEnsemble and adapt Stacker to work with both BaggedEnsemble base models as well as regular base models that offer  `predict_oof()` method (where these are not subclass of BaggedEnsemble). But heterogeneity of stacking BaggedEnsemble and individual models together may be annoying as well...

szha:
Job PR-1022-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1022/2/index.html

core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
Innixma: Added comment about child_oof, added tags. Will leave `OOFModel` to a separate PR as it only becomes important if we have widespread usage of custom `OOF` generation strategies.

szha:
Job PR-1022-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1022/3/index.html

szha:
Job PR-1022-4 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1022/4/index.html

core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
jwmueller: Sounds good, I'd also edit the comment at the top of BaggedEnsemble class, as the current description is no longer accurate.

core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
```diff
@@ -108,6 +114,13 @@ def _fit(self,
              time_limit=None,
              sample_weight=None,
              **kwargs):
+        use_child_oof = self.params.get('use_child_oof', False)
+        if use_child_oof:
```
jwmueller, line 118: Will BaggedEnsemble with `_child_oof=True` be unnecessarily retrained during refit_full()? Or is this also covered by the `self.is_fit()` check here?

Innixma: It will be unnecessarily retrained. The code to make this not occur is complex, so it can be left to a future PR.

Innixma: Added a FIXME comment regarding this

core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
Innixma: Edited the doc

szha:
Job PR-1022-5 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1022/5/index.html

szha:
Job PR-1022-6 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1022/6/index.html

Innixma:
@jwmueller @gradientsky Benchmark results:

```
AutoGluon_bestquality_1h_2021_03_20_knn_v3 VS all
                                    framework  > AutoGluon_bestquality_1h_2021_03_20_knn_v3  < AutoGluon_bestquality_1h_2021_03_20_knn_v3  = AutoGluon_bestquality_1h_2021_03_20_knn_v3  % Less Avg. Errors Than AutoGluon_bestquality_1h_2021_03_20_knn_v3  time_train_s  metric_error  time_infer_s  loss_rescaled  time_infer_s_rescaled      rank  rank=1_count  rank=2_count  rank=3_count  rank>3_count  error_count
0  AutoGluon_bestquality_1h_2021_03_20_knn_v3                                             0                                             0                                          1026                                           0.000000                    3433.135231  69836.896504     81.431207       0.448479               1.255433  1.472031           516           510             0             0            1
1         AutoGluon_bestquality_1h_2021_02_26                                           452                                           509                                            58                                          -0.121582                    3470.376644  69749.006433    107.480373       0.505397               1.845392  1.527969           453           567             0             0            7
```

`AutoGluon_bestquality_1h_2021_02_26` is the v0.1.0 release results. This change reduces average inference time by ~33%. Refer to the `time_infer_s_rescaled` column for details. (equivalent to aggregated (framework time / best time) for each dataset)

This change also improves our robustness, we fail on fewer datasets as I've added a new check to ensure prediction times will not take extremely long relative to available time left. This fixes situations where KNN may take many hours to predict but only a few seconds to train. Now these cases will cause KNN to error with TimeLimitExceeded before making the long-running prediction.

Additionally, due to the time saved from avoiding bagging KNN, the overall results are also improved.

Finally, while not shown in these results, disk space usage will have also been reduced heavily since only one KNN model will be persisted on disk, instead of 10+ for each bag.

jwmueller (approved):

Issue #1030, opened 2021-03-24
autogluon.mxnet is missing dependency on autogluon.extra


petergtz:
`autogluon.mxnet` uses `autogluon.extra`:
https://github.com/awslabs/autogluon/blob/b403079bcc17171d8b5366096f7d7f9e7eb1b99a/mxnet/src/autogluon/mxnet/task/nets.py#L8
but it doesn't list it as a dependency:
https://github.com/awslabs/autogluon/blob/b403079bcc17171d8b5366096f7d7f9e7eb1b99a/mxnet/setup.py#L19-L28

This breaks when doing a `python3 -m pip install autogluon.mxnet` and then importing `autogluon.mxnet` with the following error message:
```
ModuleNotFoundError: No module named 'autogluon.extra'
```

This should probably be listed either as a required or optional dependency `setup.py`.

Innixma:
Thanks for spotting this! #1032 fixes the issue and will be available in our next release.

PR #1032, opened 2021-03-24
Fix autogluon.mxnet crash on submodule install

Innixma:
*Issue #, if available:*
#1030 

*Description of changes:*

If installed in isolation, `autogluon.mxnet` would crash on import due to unintentionally depending on `autogluon.extra`. This PR fixes this and now there is no dependency between them.


szha:
Job PR-1032-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1032/1/index.html

mxnet/tests/unittests/test_import.py
```diff
@@ -0,0 +1,3 @@
+
```
jwmueller, line 1: Should CI also run these individual `test_import()` calls with just the minimal sub-package installed?  Would that have caught the issue with importing only autogluon.mxnet without autogluon.extra being installed?

Without such an individual test per sub-package, there might again be unintended dependencies re-introduced between sub-packages in future commits?

extra/src/autogluon/extra/model_zoo/_nets.py
```diff
@@ -0,0 +1,54 @@
...
+from mxnet import gluon, init
+
+from .model_zoo import get_model
+
+logger = logging.getLogger(__name__)
+
+__all__ = ['get_network', 'get_built_in_network']
+
+
+# TODO: Consider deleting this file, it is only used in a single tutorial and shouldn't be a part the codebase
```
jwmueller, line 12: agreed, this could just be defined in the tutorial if it's not used anywhere else

jwmueller (approved):

extra/src/autogluon/extra/model_zoo/_nets.py
zhreshold: could affect the nas functionalities, though I agree current structure is not the best way to do it

zhreshold (approved):

mxnet/tests/unittests/test_import.py
Innixma: CI runs unit tests with minimal install (mostly, tabular installs text), so this should catch it.

extra/src/autogluon/extra/model_zoo/_nets.py
Innixma: @zhreshold if you can find time, it would be nice if you could rework the tutorial that uses this so that this code can be deleted. A lot of it appears to be largely outdated.

Innixma: Also, we should consider if we want to keep the NAS functionality or not, it isn't well maintained by us at present, and I'm concerned that it will confuse users and developers in future. I see a fair amount of old code that is only used in a few places in NAS code, and could be deleted if NAS wasn't present or if NAS was refactored.

PR #1034, opened 2021-03-25
Constrained Bayesian optimization

ValerioPerrone:
*Description of changes:*

Extends the acquisition function to handle multiple outputs and implements constrained Bayesian optimization.

This is useful to tackle problems such as optimizing accuracy with an upper bound on memory requirements or on unfairness.

The main change is meanstd_acqfunc.py. The acquisition function now internally operates with a dictionary Dict[str, SurrogateModel], allowing for a model for each output metric. The standard case is a dictionary with a single model but the change is backward compatible: if a plain SurrogateModel is passed, the acquisition function automatically turns it into a one-key dictionary.

Other key changes: 

* Implemented CEIAcquisitionFunction and EIpuAcquisitionFunction, two multi-output acquisition functions. These require the predictions from both the main model and, respectively, the constraint or cost model, and return the head gradients wrt to the mean and standard deviation of each model.

* Introduced 'constrained_bayesopt', a new searcher to run end-to-end experiments with constrained BO.

* Extended GPModelPendingCandidateStateTransformer to support multiple models with a shared state.

Testing

* All previous tests pass. The behavior of existing code is unaffected by the acquisition function refactoring.

* test_bayesopt_eipu.py: tests cost-aware EI, an example of multi-output acquisition function.

* test_bayesopt_cei.py: tests the constrained EI. This includes tests with/without MCMC, with/without fantasizing, and with/without feasible candidates.

* test_multi_output_searcher.py: tests the constrained BO searcher.

### DIFF
--- a/core/src/autogluon/core/searcher/bayesopt/autogluon/constrained_gp_fifo_searcher.py
+++ b/core/src/autogluon/core/searcher/bayesopt/autogluon/constrained_gp_fifo_searcher.py
@@ -1,0 +1,262 @@
+from typing import Callable, Dict, Type, Optional, Any
+import copy
+import logging
+
+from .gp_fifo_searcher import GPFIFOSearcher, decode_state, \
+    MapReward, GET_CONFIG_RANDOM_RETRIES, create_initial_candidates_scorer
+from ..datatypes.common import CandidateEvaluation, Candidate, \
+    candidate_for_print
+from ..datatypes.hp_ranges import HyperparameterRanges
+from ..datatypes.tuning_job_state import TuningJobState
+from ..models.gp_model import GPModel
+from ..models.gpmodel_skipopt import SkipOptimizationPredicate
+from ..models.gpmodel_transformers import GPModelPendingCandidateStateTransformer, GPModelArgs
+from ..tuning_algorithms.base_classes import DEFAULT_METRIC, DEFAULT_CONSTRAINT_METRIC, \
+    LocalOptimizer, AcquisitionFunction
+from ..tuning_algorithms.bo_algorithm import BayesianOptimizationAlgorithm
+from ..tuning_algorithms.common import compute_blacklisted_candidates
+from ..tuning_algorithms.defaults import DEFAULT_LOCAL_OPTIMIZER_CLASS, DEFAULT_NUM_INITIAL_CANDIDATES, \
+    DEFAULT_NUM_INITIAL_RANDOM_EVALUATIONS
+from ..utils.debug_log import DebugLogPrinter
+from ..utils.duplicate_detector import DuplicateDetectorIdentical
+from ..utils.simple_profiler import SimpleProfiler
+
+logger = logging.getLogger(__name__)
+
+
+class ConstrainedGPFIFOSearcher(GPFIFOSearcher):
+    """
+    Supports constrained GP-based hyperparameter optimization (to be used with a FIFO scheduler).
+
+    """
+    def __init__(self, hp_ranges: HyperparameterRanges, random_seed: int,
+                 output_gpmodels: Dict[str, GPModel],
+                 output_models_args: Dict[str, GPModelArgs],
+                 map_reward: MapReward,
+                 acquisition_class: Type[AcquisitionFunction],
+                 init_state: TuningJobState = None,
+                 local_minimizer_class: Type[LocalOptimizer] = DEFAULT_LOCAL_OPTIMIZER_CLASS,
+                 skip_optimization: SkipOptimizationPredicate = None,
+                 num_initial_candidates: int = DEFAULT_NUM_INITIAL_CANDIDATES,
+                 num_initial_random_choices: int = DEFAULT_NUM_INITIAL_RANDOM_EVALUATIONS,
+                 initial_scoring: Optional[str] = None,
+                 first_is_default: bool = True,
+                 debug_log: Optional[DebugLogPrinter] = None,
+                 cost_metric_name: Optional[str] = None,
+                 profiler: Optional[SimpleProfiler] = None,
+                 getconfig_callback: Optional[Callable[[dict], Any]] = None):
+        """
+        Note that the SurrogateModel is created on demand (by the state
+        transformer) in get_config, along with components needed for the BO
+        algorithm.
+
+        The searcher is supposed to maximize reward, while internally, the
+        criterion is minimized. map_reward maps reward to internal criterion, it
+        must be strictly decreasing.
+
+        :param hp_ranges: Configuration space without resource attribute
+        :param random_seed:
+        :param output_gpmodels: Dict with the GP regression model for the active metric and constraint metric
+        :param output_models_args: Dict with arguments for GPMXNet model creation for the active and constraint metric
+        :param map_reward: Function mapping reward to criterion to be minimized
+        :param acquisition_class: Type for acquisition function
+        :param init_state: TuningJobState to start from (default is empty)
+        :param local_minimizer_class: Type for local minimizer
+        :param skip_optimization: Predicate, see
+            GPMXNetPendingCandidateStateTransformer
+        :param num_initial_candidates: See BayesianOptimizationAlgorithm
+        :param num_initial_random_choices: Configs are sampled at random until
+            this many candidates received label feedback
+        :param initial_scoring: Scoring function to rank initial candidates.
+            Default: thompson_indep (independent Thompson sampling)
+        :param first_is_default: If true, the first config to be evaluated is
+            the default of the search space. Otherwise, the first is sampled
+            at random
+        :param debug_log: DebugLogPrinter for debug logging (optional)
+        :param cost_metric_name: If an `elapsed_time` record is passed to
+            `update`, the value is entered in the state using this key
+        :param profiler: If given, HPO computations are profiled
+        :param getconfig_callback: If given, this callback function is called
+            at the end of each get_config. It receives the chosen config
+            (as dict) as single argument. One use case for this callback is
+            to store profiler records.
+        """
+        self.active_metric_name = DEFAULT_METRIC
+        self.constraint_metric_name = DEFAULT_CONSTRAINT_METRIC
+        super().__init__(hp_ranges, random_seed, output_gpmodels[self.active_metric_name],
+                         output_models_args[self.active_metric_name], map_reward, acquisition_class, init_state,
+                         local_minimizer_class, skip_optimization, num_initial_candidates, num_initial_random_choices,
+                         initial_scoring, first_is_default, debug_log, cost_metric_name, profiler, getconfig_callback)
+
+        self.skip_optimization = skip_optimization
+        self.state_transformer = GPModelPendingCandidateStateTransformer(
+            gpmodel=output_gpmodels,
+            init_state=self._init_state,
+            model_args=output_models_args,
+            skip_optimization=skip_optimization,
+            profiler=profiler,
+            debug_log=debug_log)
+
+        if debug_log is not None:
+            deb_msg = "[ConstrainedGPFIFOSearcher.__init__]\n"
+            deb_msg += ("- acquisition_class = {}\n".format(acquisition_class))
+            deb_msg += ("- local_minimizer_class = {}\n".format(local_minimizer_class))
+            deb_msg += ("- num_initial_candidates = {}\n".format(num_initial_candidates))
+            deb_msg += ("- num_initial_random_choices = {}\n".format(num_initial_random_choices))
+            deb_msg += ("- initial_scoring = {}\n".format(self.initial_scoring))
+            deb_msg += ("- first_is_default = {}".format(first_is_default))
+            logger.info(deb_msg)
+
+    def update(self, config: Candidate, reward: float, constraint: float, **kwargs):
+        """
+        Registers new datapoint at config, with reward reward.
+        Note that in general, config should previously have been registered as
+        pending (register_pending). If so, it is switched from pending
+        to labeled. If not, it is considered directly labeled.
+        """
+        crit_val = self.map_reward(reward)
+        constr_val = constraint
+        metrics = {self.active_metric_name: crit_val,
+                   self.constraint_metric_name: constr_val}
+        if 'elapsed_time' in kwargs:
+            metrics[self.cost_metric_name] = kwargs['elapsed_time']
+        self.state_transformer.label_candidate(CandidateEvaluation(
+            candidate=copy.deepcopy(config), metrics=metrics))
+        if self.debug_log is not None:
+            config_id = self.debug_log.config_id(config)
+            msg = "Update for config_id {}: reward = {}, crit_val = {}, constr_val = {}".format(
+                config_id, reward, crit_val, constr_val)
+            logger.info(msg)
+
+    def get_config(self) -> Candidate:
+        """
+        Runs Bayesian optimization in order to suggest the next config to evaluate.
+
+        :return: Next config to evaluate at
+        """
+        state = self.state_transformer.state
+        if self.do_profile:
+            # Start new profiler block
+            meta = {
+                'fit_hyperparams': not self.state_transformer.skip_optimization(
+                    state),
+                'num_observed': len(state.candidate_evaluations),
+                'num_pending': len(state.pending_evaluations)
+            }
+            self.profiler.begin_block(meta)
+            self.profiler.start('all')
+        blacklisted_candidates = compute_blacklisted_candidates(state)
+        pick_random = (len(blacklisted_candidates) < self.num_initial_random_choices) or \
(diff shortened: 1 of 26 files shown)


szha:
Job PR-1034-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1034/1/index.html

core/src/autogluon/core/searcher/bayesopt/autogluon/constrained_gp_fifo_searcher.py
```diff
@@ -0,0 +1,262 @@
+from typing import Callable, Dict, Type, Optional, Any
```
Innixma, line 1: A general thought: This is a lot of changes, do we think it makes sense to have an entire submodule `autogluon.search` dedicated to this code / searcher code in general? This could be an optional dependency for other modules which would allow to plug-in searcher backends. This would lighten the `autogluon.core` code and ensure that the blast radius of searcher changes are contained within the searcher module.

Innixma: This would also make migration easier if we eventually move the searcher logic to ray or some similar package.

mseeger: In fact, we will move all this code out of the AutoGluon codebase and bring it back as a dependence. We are currently working on this full steam, our code would then work on AutoGluon, but also on Ray Tune. In the end, this is the cleanest solution.

This will happen likely in April, latest in May.

Of course, we'll make sure everything will work just as before.

Innixma: That's great, I think that is the cleanest approach as well.

szha:
Job PR-1034-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1034/2/index.html

szha:
Job PR-1034-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1034/3/index.html

core/src/autogluon/core/searcher/bayesopt/autogluon/constrained_gp_fifo_searcher.py
```diff
@@ -0,0 +1,262 @@
...
+        :param debug_log: DebugLogPrinter for debug logging (optional)
+        :param cost_metric_name: If an `elapsed_time` record is passed to
+            `update`, the value is entered in the state using this key
+        :param profiler: If given, HPO computations are profiled
+        :param getconfig_callback: If given, this callback function is called
+            at the end of each get_config. It receives the chosen config
+            (as dict) as single argument. One use case for this callback is
+            to store profiler records.
+        """
+        self.active_metric_name = DEFAULT_METRIC
```
mseeger, line 84: These two seem to be parameters the user can pass elsewhere, e.g. in searcher_factory.py.

I think either way is OK (user can choose, or we fix them). Since these are internal names, it is fine to fix them. But please only do one of them.

core/src/autogluon/core/searcher/bayesopt/autogluon/searcher_factory.py
```diff
@@ -101,7 +102,7 @@ def _create_common_objects(**kwargs):
     model_args = GPModelArgs(
         num_fantasy_samples=kwargs['num_fantasy_samples'],
         random_seed=random_seed,
-        active_metric=DEFAULT_METRIC,
+        active_metric=kwargs.get('active_metric', DEFAULT_METRIC),
```
mseeger, line 105: Since active_metric is just an internal name (right?), we can just fix it to DEFAULT_METRIC, no?

core/src/autogluon/core/searcher/bayesopt/autogluon/searcher_factory.py
```diff
@@ -167,6 +168,63 @@ def _resource_for_acquisition(kwargs, hp_ranges_cs):
...
+    :return: ConstrainedGPFIFOSearcher object
+
+    """
+    assert kwargs['scheduler'] == 'fifo', \
+        "This factory needs scheduler = 'fifo' (instead of '{}')".format(
+            kwargs['scheduler'])
+    # Common objects
+    hp_ranges_cs, random_seed, gpmodel, model_args, profiler, _map_reward, \
+    skip_optimization, debug_log = _create_common_objects(**kwargs)
+    constraint_metric_name = kwargs.get('constraint_metric', DEFAULT_CONSTRAINT_METRIC)
```
mseeger, line 186: These are internal names, no?

Of course, the user-facing names (appearing in train_fn) must be configurable.

core/src/autogluon/core/searcher/bayesopt/autogluon/searcher_factory.py
```diff
@@ -167,6 +168,63 @@ def _resource_for_acquisition(kwargs, hp_ranges_cs):
...
+    active_metric_name = kwargs.get('active_metric', DEFAULT_METRIC)
+    initial_scoring = kwargs.get('initial_scoring', 'acq_func')
+    opt_warmstart = kwargs.get('opt_warmstart', False)
+    # Constrained GP regression model
+    kernel = Matern52(dimension=hp_ranges_cs.ndarray_size(), ARD=True)
```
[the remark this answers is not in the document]
mseeger: BTW: You could make this really clean by making a temporary function which creates gpmodel, model_args, and then call that both in _create_common_objects and here. This way, we make sure this is done in the same way.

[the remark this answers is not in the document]
mseeger: Please do this, let us avoid copy&paste

core/src/autogluon/core/searcher/bayesopt/autogluon/searcher_factory.py
```diff
@@ -234,6 +292,10 @@ def _common_defaults(is_hyperband: bool) -> (Set[str], dict, dict):
         default_options['gp_resource_kernel'] = 'matern52'
         default_options['resource_acq'] = 'bohb'
         default_options['num_init_random'] = 10
+    if is_constrained:
+        default_options['active_metric'] = DEFAULT_METRIC
```
mseeger, line 296: See above. If these are internal, there is no value having the user set them. Correct me if I am wrong.

core/src/autogluon/core/searcher/bayesopt/autogluon/constrained_gp_fifo_searcher.py
ValerioPerrone: I've now fixed them as they are internal only.

core/src/autogluon/core/searcher/bayesopt/autogluon/searcher_factory.py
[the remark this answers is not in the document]
ValerioPerrone: I've factored it out as a common function.

core/src/autogluon/core/searcher/bayesopt/autogluon/searcher_factory.py
ValerioPerrone: Done.

core/src/autogluon/core/searcher/bayesopt/autogluon/searcher_factory.py
ValerioPerrone: Done.

core/src/autogluon/core/searcher/bayesopt/autogluon/searcher_factory.py
ValerioPerrone: Done. They are internal and fixed.

core/src/autogluon/core/searcher/bayesopt/autogluon/gp_fifo_searcher.py
```diff
@@ -49,13 +45,16 @@ def __call__(self, x: float) -> float:
...
-        random_state: np.random.RandomState) -> ScoringFunction:
+        random_state: np.random.RandomState, active_output: str = None) -> ScoringFunction:
     if initial_scoring == 'thompson_indep':
+        if isinstance(model, Dict):
+            assert active_output in model
+            model = model[active_output]
         return IndependentThompsonSampling(model, random_state=random_state)
     else:
-        return acquisition_class(model)
+        return acquisition_class(model, active_output)
```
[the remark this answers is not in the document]
ValerioPerrone: Done.

core/src/autogluon/core/searcher/bayesopt/autogluon/constrained_gp_fifo_searcher.py
```diff
@@ -0,0 +1,262 @@
...
+            if self.do_profile:
+                self.profiler.push_prefix('getconfig')
+                self.profiler.start('all')
+                self.profiler.start('gpmodel')
+            # Create the active and constraint models
+            output_models = self.state_transformer.model()  # Asking for the model triggers the posterior computation
+            if self.do_profile:
+                self.profiler.stop('total_update')
+            # Create BO algorithm
+            initial_candidates_scorer = create_initial_candidates_scorer(
```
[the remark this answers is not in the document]
ValerioPerrone: Done.

szha:
Job PR-1034-5 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1034/5/index.html

core/src/autogluon/core/searcher/bayesopt/models/gpmodel_transformers.py
```diff
@@ -200,24 +216,58 @@ def _compute_model(self, skip_optimization=None):
...
-            fit_parameters=fit_parameters,
-            num_fantasy_samples=args.num_fantasy_samples,
-            normalize_targets=args.normalize_targets,
-            profiler=self._profiler,
-            debug_log=self._debug_log,
-            debug_fantasy_values=self._debug_fantasy_values)
-        # DEBUG: Supplied values are used only once
-        self._debug_fantasy_values = None
+
+        if self._use_single_model:
```
mseeger, line 220: nitpick:
This could be done without copy&paste, something like this:

if self._use_single_model:
   gpmodels = dictionarize_objective(self._gpmodel)
else:
   gpmodels = self._gpmodel
output_models = {
   GaussProcSurrogateModel(...)
   for metric_name, gpmodel in gpmodels
}
if self._use_single_model:
   self._model = output_models[self.active_metric]
else:
   self._model = output_models

You are using dictionarize_objective elsewhere for this. That's elegant, but maybe the name is misleading then.

core/src/autogluon/core/searcher/bayesopt/tuning_algorithms/base_classes.py
```diff
@@ -1,10 +1,37 @@
...
 
 from ..datatypes.common import Candidate
 from ..datatypes.tuning_job_state import TuningJobState
 
+DEFAULT_METRIC = 'active_metric'
+DEFAULT_COST_METRIC = 'cost_metric'
+DEFAULT_CONSTRAINT_METRIC = 'constraint_metric'
+
+
+def dictionarize_objective(x):
```
mseeger, line 13: OK, is there not a potential problem here?
- In the code, you allow the user to define the active metric name, and only if not is DEFAULT_METRIC used
- But in dictionarize_objective, DEFAULT_METRIC is hardcoded.

I'd prefer if the hardcoded names are used everywhere, no? There seems no advantage of the user choosing these names, because they are internal anyway.

core/src/autogluon/core/searcher/bayesopt/tuning_algorithms/base_classes.py
```diff
@@ -116,6 +143,10 @@ def backward_gradient(
         pass
 
 
+# Useful type that allows for a dictionary mapping each output name to a SurrogateModel
```
mseeger, line 146: Can you expand this docstring? This type def is used all over the place. Please do explain why this is sometimes a dict.

core/src/autogluon/core/searcher/bayesopt/tuning_algorithms/defaults.py
```diff
@@ -1,13 +1,9 @@
 from .bo_algorithm_components import LBFGSOptimizeAcquisition
-from ..models.meanstd_acqfunc import EIAcquisitionFunction
+from ..models.meanstd_acqfunc_impl import EIAcquisitionFunction
 
 
 DEFAULT_ACQUISITION_FUNCTION = EIAcquisitionFunction
 DEFAULT_LOCAL_OPTIMIZER_CLASS = LBFGSOptimizeAcquisition
 DEFAULT_NUM_INITIAL_CANDIDATES = 250
 DEFAULT_NUM_INITIAL_RANDOM_EVALUATIONS = 3
-DEFAULT_METRIC = 'active_metric'
```
mseeger, line 9: Is there a good reason for moving this to base_classes, and not leave it here?

I've never been a big fan of this defaults.py, but the intention was to collect constants here, and why not? Can't you import from base_classes?

core/src/autogluon/core/searcher/gp_searcher.py
```diff
@@ -208,6 +209,37 @@ def _to_config_cs(self, config):
         return _to_config_cs(self.gp_searcher.hp_ranges.config_space, config)
 
 
+class ConstrainedGPFIFOSearcher(GPFIFOSearcher):
+    def __init__(self, configspace, **kwargs):
+        _gp_searcher = kwargs.get('_constrained_gp_searcher')
+        if _gp_searcher is None:
+            kwargs['configspace'] = configspace
+            if 'initial_scoring' in kwargs:
```
mseeger, line 217: Why not just:
self.initial_scoring = 'acq_func'
assert (...)

core/src/autogluon/core/searcher/gp_searcher.py
```diff
@@ -208,6 +209,37 @@ def _to_config_cs(self, config):
...
+        # can lead to deadlocks when superclass methods are called
+        self._gp_lock = mp.Lock()
+
+    def update(self, config, **kwargs):
+        BaseSearcher.update(self, config, **kwargs)
+        with self._gp_lock:
+            config_cs = self._to_config_cs(config)
+            self.gp_searcher.update(
+                config_cs, reward=kwargs[self._reward_attribute],
+                constraint=kwargs[DEFAULT_CONSTRAINT_METRIC])
```
mseeger, line 240: Again, DEFAULT_CONSTRAINT_METRIC hardcoded here. Please make it hardcoded everywhere then, do not let the user choose.

mseeger: Oh wait, no! This name is user facing. You cannot use DEFAULT_CONSTRAINT_METRIC here, you need to use something like self._constraint_attribute !

OK, there are:
- Internal names, in internal searcher code: You can fix them and hardcode
- User facing names, which appear in train_fn. These have to be arguments, of course

core/src/autogluon/core/searcher/bayesopt/models/gpmodel_transformers.py
mseeger: Actually, please do this. We don't want c&p.

mseeger:
I am OK once my comments from yesterday are addressed. In particular, the scheduler needs a _constraint_attribute attribute which can be passed in. I'd also strongly prefer the internal names of these metrics not to be user-selectable, this is just creating confusion.

core/src/autogluon/core/searcher/bayesopt/tuning_algorithms/defaults.py
ValerioPerrone: The reason is that EIAcquisitionfunction requires it, resulting in a circular import.

core/src/autogluon/core/searcher/gp_searcher.py
ValerioPerrone: I've now created a self._constraint_attribute, which is user facing, and hardcoded the rest.

szha:
Job PR-1034-7 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1034/7/index.html

mseeger:
OK,

mseeger (approved):

mseeger:
Any cycles for this one? This code is the basis for a tutorial on Fair HPO, which has been published now, and which would also go into a blog post, creating visibility for AG.

Innixma (approved):
Looks good in terms of not altering much existing functionality that is actively used by submodules. I don't have context in terms of the actual changes themselves, so I'm deferring to @mseeger for that.

I'm ok to merge, @mseeger feel free to merge whenever you are ready.

Issue #1033, opened 2021-03-25
GBM doesn't like ":" in column name


jmrichardson:
Using the [tutorial](https://auto.gluon.ai/stable/tutorials/tabular_prediction/tabular-indepth.html) with a different data set produced the error below.  Changing the column names to exclude a ":" resolved the issue however would like to report it as I prefer them and the other models i have tested so far don't seem to have this issue.  Also, this only occurs with something similar to the the following in fit (ie commenting them out does not produce the error).

```
    num_bag_folds=5,
    num_bag_sets=2,  
    num_stack_levels=2,
```

This resolved the problem:

`train_data.columns = [re.sub(r':', '_', col) for col in train_data.columns]`

Here is the error:

```
Fitting model: LightGBM_BAG_L1\T0 ... Training model for up to 30.39s of the 96.97s of remaining time.
	Warning: Exception caused LightGBM_BAG_L1\T0 to fail during training... Skipping this model.
		"None of [Index(['close', 'pta.hilo_high_length:1019_low_length:6646_0',\n       'pta.hilo_high_length:1019_low_length:6646_1',\n       'pta.hilo_high_length:1019_low_length:6646_2',\n       'tta.MIDPOINT_timeperiod:5952', 'pta.pwma_length:6042',\n       'pta.pvi_length:5903_initial:1241', 'fta.SMM_period:6867',\n       'fta.VAMA_period:6760', 'fta.EVWMA_period:4723',\n       'tta.T3_timeperiod:4868', 'pta.ttm_trend_length:6119',\n       'pta.vidya_length:416', 'tta.BBANDS_timeperiod:6238_0',\n       'tta.BBANDS_timeperiod:6238_1', 'tta.BBANDS_timeperiod:6238_2',\n       'fta.BBANDS_period:4424_0', 'fta.BBANDS_period:4424_1',\n       'fta.BBANDS_period:4424_2',\n       'fta.KAMA_er:4985_ema_fast:1498_ema_slow:152_period:7420',\n       'pta.decreasing_length:3031', 'fta.EFI_period:1761'],\n      dtype='object')] are in the [columns]"
Detailed Traceback:
Traceback (most recent call last):
  File "D:\Projects\tactory\autogluon\tabular\trainer\abstract_trainer.py", line 911, in _train_and_save
    model = self._train_single(X, y, model, X_val, y_val, **model_fit_kwargs)
[...]

PR #1036, opened 2021-03-27
Tabular: Fixed bug with preprocess_nonadaptive during stacking

Innixma:
*Issue #, if available:*
#1033 

*Description of changes:*

When `preprocess` was renamed to `preprocess_nonadaptive`, this usage was not updated to the new name, meaning it would default to `preprocess_nonadaptive=True` during stacking, which would cause exceptions in LightGBM during repeated bagging with time_limit set when invalid column names existed.

This PR fixes this bug.


szha:
Job PR-1036-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1036/1/index.html

jwmueller (approved):

PR #1037, opened 2021-03-28
Tabular: Added support for specifying early stopping rounds in GBM, CAT, XGB

Innixma:
*Issue #, if available:*

*Description of changes:*

Now users can specify how many rounds of early stopping should be used via `ag.es` parameter in models.

This is not yet documented to user as it is very advanced and likely isn't something we want users to be worrying about. Especially since long term we probably want to move towards adaptive learning rate strategies.

The plan after merge is to begin experimenting with alternative early stopping strategies. Currently for large datasets, we only have **10** early stopping rounds. This is far too few. For example, in the Airlines dataset (115M rows), Airlines required an early stopping rounds value of **10,000** to get to the best num_boost_rounds of **~275,000**. With only 10 early stopping rounds, it stopped around iteration 2000, with a full 5% lower accuracy than the fully trained one.

A simple adaptive strategy could be to have a minimum early stop round number, say 10, and a maximum early stop number, say 10000, and finally an early stop round ratio, say 0.1 . During training, the early stopping rounds value is dictated per iteration by:

```
# max_early_stop = 10000
# min_early_stop = 10
# early_stop_ratio = 0.1

for cur_round in range(num_boost_rounds):
    ...
    ...
    ...
    early_stopping_rounds = min(max_early_stop, max(min_early_stop, round(cur_best_round*early_stop_ratio)))
    if cur_round - cur_best_round > early_stopping_rounds:
        # early stop
        break
```

This would ensure the added cost of early stopping would only be the greater of 10 rounds or 1/10th of the total training time. This avoids the scenario where we set early_stop_rounds = 1000 but the best iteration is found on iteration 50 and then steeply gets worse, causing training to take 20x longer than with the adaptive strategy. 

### DIFF
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -705,7 +705,7 @@ class AbstractModel:
         time_start = time.time()
         logger.log(15, "Starting generic AbstractModel hyperparameter tuning for %s model..." % self.name)
         self._set_default_searchspace()
-        params_copy = self.params.copy()
+        params_copy = self._get_params()
         directory = self.path  # also create model directory if it doesn't exist
         # TODO: This will break on S3. Use tabular/utils/savers for datasets, add new function
         scheduler_cls, scheduler_params = scheduler_options  # Unpack tuple
@@ -918,7 +918,6 @@ class AbstractModel:
         """
         return {}
 
-
     def _get_default_stopping_metric(self):
         """
         Returns the default stopping metric to use for early stopping.
@@ -931,6 +930,31 @@ class AbstractModel:
             stopping_metric = self.eval_metric
         stopping_metric = metrics.get_metric(stopping_metric, self.problem_type, 'stopping_metric')
         return stopping_metric
+
+    def _get_params(self) -> dict:
+        """Gets all params."""
+        return self.params.copy()
+
+    def _get_ag_params(self) -> dict:
+        """Gets params that are not passed to the inner model, but are used by the wrapper."""
+        ag_param_names = self._ag_params()
+        if ag_param_names:
+            return {key: val for key, val in self.params.items() if key in ag_param_names}
+        else:
+            return dict()
+
+    def _get_model_params(self) -> dict:
+        """Gets params that are passed to the inner model."""
+        ag_param_names = self._ag_params()
+        if ag_param_names:
+            return {key: val for key, val in self.params.items() if key not in ag_param_names}
+        else:
+            return self._get_params()
+
+    # TODO: Add documentation for valid args for each model. Currently only `ag.es`
+    def _ag_params(self) -> set:
+        """Set of params that are not passed to the inner model, but are used by the wrapper."""
+        return set()
 
     def _get_tags(self):
         collected_tags = {}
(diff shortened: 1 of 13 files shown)


szha:
Job PR-1037-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1037/1/index.html

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -932,6 +931,31 @@ def _get_default_stopping_metric(self):
         stopping_metric = metrics.get_metric(stopping_metric, self.problem_type, 'stopping_metric')
         return stopping_metric
 
+    def _get_params(self) -> dict:
+        """Gets all params."""
+        return self.params.copy()
+
+    def _get_ag_params(self) -> dict:
+        """Gets params that are not passed to the inner model, but are used by the wrapper."""
```
jwmueller, line 939: 'inner model' may not be clear language for contributors

jwmueller:
Maybe we should have more informative parameter name for `ag.es`, like `ag.earlystop`?
Also currently this parameter doesn't appear to control NN models, which should be clarified?

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -932,6 +931,31 @@ def _get_default_stopping_metric(self):
...
+
+    def _get_model_params(self) -> dict:
+        """Gets params that are passed to the inner model."""
+        ag_param_names = self._ag_params()
+        if ag_param_names:
+            return {key: val for key, val in self.params.items() if key not in ag_param_names}
+        else:
+            return self._get_params()
+
+    # TODO: Add documentation for valid args for each model. Currently only `ag.es`
```
jwmueller, line 954: probably should add documentation before merging this PR? Otherwise the meaning of `ag.es` will appear cryptic to others?

jwmueller:
You'll benchmark this change?  It's not obvious to me whether it'll be better/worse depending on time-limits.

core/src/autogluon/core/models/abstract/abstract_model.py
Innixma: any recommendation for improved wording?

Innixma:
> You'll benchmark this change? It's not obvious to me whether it'll be always be better depending on time-limits.

This PR doesn't change the functionality, only cleans up code + adds option for specification. I will be benchmarking the follow-up PR that will add adaptive early stopping.

core/src/autogluon/core/models/abstract/abstract_model.py
jwmueller: maybe just: inner model (`self.model`)?

jwmueller (approved):
Just have minor comments re clarity.  Consider `ag.es` -> `ag.earlystop` as well for easier usage.

core/src/autogluon/core/models/abstract/abstract_model.py
Innixma: Added

core/src/autogluon/core/models/abstract/abstract_model.py
Innixma: added

Innixma:
will rename to ag.earlystop in follow-up PR.

PR #1039, opened 2021-03-29
[Vision] add early stopping strategies

zhreshold:
*Issue #, if available:*

*Description of changes:*
Add early stopping control for ImagePredictor and ObjectDetector.

New hyperparameters that controls the early stopping bahavior:

```python
"""
early_stop_patience : int, default=-1
    Number of epochs with no improvement after which train is early stopped. Use -1 to disable.
early_stop_min_delta : float, default=1e-4
    The small delta value to ignore when evaluating the metric. A large delta helps stablize the early
    stopping strategy against tiny fluctuation, e.g. 0.5->0.49->0.48->0.499->0.500001 is still considered as
    a good timing for early stopping.
early_stop_baseline : float, default=0.0
    The minimum(baseline) value to trigger early stopping. For example, with `early_stop_baseline=0.5`,
    early stopping won't be triggered if the metric is less than 0.5 even if plateau is detected.
early_stop_max_value : float, default=1.0
    The max value for metric, early stop training instantly once the max value is achieved.
You can get the list of accepted hyperparameters in `config.yaml` saved by this predictor.
"""
```

### DIFF
--- a/vision/src/autogluon/vision/detector/detector.py
+++ b/vision/src/autogluon/vision/detector/detector.py
@@ -101,7 +101,8 @@ class ObjectDetector(object):
                         'transfer': Categorical('faster_rcnn_fpn_resnet101_v1d_coco'),
                         'lr': Real(1e-5, 1e-3, log=True),
                         'batch_size': Categorical(4, 8),
-                        'epochs': 30
+                        'epochs': 30,
+                        'early_stop_patience': -1
                         },
                     'hyperparameter_tune_kwargs': {
                         'num_trials': 128,
@@ -117,7 +118,8 @@ class ObjectDetector(object):
                                                 'center_net_resnet50_v1b_coco'),
                         'lr': Real(1e-4, 1e-2, log=True),
                         'batch_size': Categorical(8, 16, 32, 64),
-                        'epochs': 50
+                        'epochs': 50,
+                        'early_stop_patience': 20
                         },
                     'hyperparameter_tune_kwargs': {
                         'num_trials': 512,
@@ -131,7 +133,8 @@ class ObjectDetector(object):
                         'transfer': Categorical('ssd_512_resnet50_v1_coco'),
                         'lr': 0.01,
                         'batch_size': Categorical(8, 16),
-                        'epochs': 30
+                        'epochs': 30,
+                        'early_stop_patience': 5
                         },
                     'hyperparameter_tune_kwargs': {
                         'num_trials': 16,
@@ -147,6 +150,7 @@ class ObjectDetector(object):
                         'lr': Categorical(0.01, 0.005, 0.001),
                         'batch_size': Categorical(32, 64, 128),
                         'epochs': Categorical(30, 50),
+                        'early_stop_patience': 10
                         },
                     'hyperparameter_tune_kwargs': {
                         'num_trials': 32,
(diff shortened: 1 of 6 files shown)


szha:
Job PR-1039-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1039/2/index.html

vision/src/autogluon/vision/detector/detector.py
```diff
@@ -164,6 +168,17 @@ def fit(self,
                 Mini batch size
             lr : float
                 Trainer learning rate for optimization process.
+            early_stop_patience : int, default=-1
+                Number of epochs with no improvement after which train is early stopped. Use -1 to disable.
+            early_stop_min_delta : float, default=1e-4
+                The small delta value to ignore when evaluating the metric. A large delta helps stablize the early
+                stopping strategy against tiny fluctuation, e.g. 0.5->0.49->0.48->0.499->0.500001 is still considered as
+                a good timing for early stopping.
+            early_stop_baseline : float, default=0.0
```
Innixma, line 177: Should we have these disabled (None) by default? This seems metric specific. RMSE / r2 / AUC / custom metrics may differ in min/max values.

vision/src/autogluon/vision/detector/detector.py
```diff
@@ -164,6 +168,17 @@ def fit(self,
                 Mini batch size
             lr : float
                 Trainer learning rate for optimization process.
+            early_stop_patience : int, default=-1
```
Innixma, line 171: Should we standardize that None is the way to disable? Since `-1` doesn't make sense for things like `early_stop_baseline` since metrics like r2 can have negative values.

vision/src/autogluon/vision/predictor/predictor.py
```diff
@@ -107,7 +107,8 @@ def fit(self,
                         'model': Categorical('resnet50_v1b', 'resnet101_v1d', 'resnest200'),
                         'lr': Real(1e-5, 1e-2, log=True),
                         'batch_size': Categorical(8, 16, 32, 64, 128),
-                        'epochs': 200
+                        'epochs': 200,
+                        'early_stop_patience': -1
```
Innixma, line 111: Are we sure this is useful? Why not have it be 50? Is it realistic that it would take 100+ epochs to show improvement?

Innixma (approved):
Looks good!

sxjscience (approved):
LGTM!

szha:
Job PR-1039-4 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1039/4/index.html

vision/src/autogluon/vision/predictor/predictor.py
zhreshold: changed best_quality to use 50 epochs, it's very reasonable for now but somewhat suboptimal comparing to `as-is` because currently the learning rate scheduler is not affected by the early stopping, in the future a LRSchedulerOnPlateau or similar stuff can offer better tradeoffs

zhreshold:
@Innixma Addressed the comments, after merging this it should improve the overall training time by a large margin especially when the dataset is not that hard.

Innixma:
Awesome! Feel free to merge whenever you are ready.

PR #1042, opened 2021-03-31
Tabular: Added Adaptive Early Stopping

Innixma:
*Issue #, if available:*

*Description of changes:*

- Added Adaptive Early Stopping to LightGBM.

TODO:

- [x] Benchmark

### DIFF
--- a/core/src/autogluon/core/utils/early_stopping.py
+++ b/core/src/autogluon/core/utils/early_stopping.py
@@ -1,0 +1,103 @@
+
+class AbstractES:
+    def update(self, cur_round, is_best=False) -> bool:
+        raise NotImplementedError
+
+    def early_stop(self, cur_round, is_best=False) -> bool:
+        raise NotImplementedError
+
+
+class SimpleES(AbstractES):
+    def __init__(self, stopping_rounds=10):
+        self.stopping_rounds = stopping_rounds
+        self.best_round = 0
+
+    def update(self, cur_round, is_best=False):
+        if is_best:
+            print(f'new best: \t{cur_round}\t{self.best_round}\t{cur_round - self.best_round}\t{(cur_round - self.best_round) / self.stopping_rounds}')
+            self.best_round = cur_round
+        return self.early_stop(cur_round, is_best=is_best)
+
+    def early_stop(self, cur_round, is_best=False):
+        if is_best:
+            return False
+        return cur_round - self.best_round >= self.stopping_rounds
+
+
+# TODO: Add time component
+#  if given a large amount of time and training is fast, should check more rounds before early stopping
+# TODO: Incorporate score, rolling window
+class AdaptiveES(AbstractES):
+    """
+    Implements Adaptive Early Stopping
+
+    stopping_rounds is adaptively adjusted across training instead of being a fixed value.
+
+    Parameters
+    ----------
+    adaptive_rate : float, default 0.2
+        The rate of increase in stopping_rounds.
+        Set to 0 to disable, or negative to shrink the early stopping rounds during training.
+    adaptive_offset : int, default 10
+        The initial stopping_rounds when cur_round is 0.
+    min_rounds : int, default 10
+        The minimum value of stopping_rounds.
+    max_rounds : int, default 10000
+        The maximum value of stopping_rounds.
+
+    Attributes
+    ----------
+    best_round : int
+        The most recent round passed to self.update with `is_best=True`.
+        Dictates stopping_rounds and is used to determine if self.early_stop() returns True.
+    stopping_rounds : int
+        If no improvement occurs in `stopping_rounds` rounds or greater, self.early_stop will return True.
+        stopping_rounds is dictated by the following formula:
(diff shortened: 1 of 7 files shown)


szha:
Job PR-1042-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1042/1/index.html

szha:
Job PR-1042-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1042/2/index.html

core/src/autogluon/core/utils/early_stopping.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+# TODO: Add time component
+#  if given a large amount of time and training is fast, should check more rounds before early stopping
+# TODO: Incorporate score, rolling window
+class AdaptiveES(AbstractES):
+    """
+    Implements Adaptive Early Stopping
+
+    stopping_rounds is adaptively adjusted across training instead of being a fixed value.
```
jwmueller, line 34: Would be good to explain what this does and its advantages since this docstring is the sole description of this.

jwmueller: Also would be good to include example of how user can in predictor.fit() specify some of these hyperparameters like `min_rounds`. This seems pretty unclear with the current documentation.

core/src/autogluon/core/utils/early_stopping.py
```diff
@@ -0,0 +1,103 @@
...
+class AbstractES:
+    def update(self, cur_round, is_best=False) -> bool:
+        raise NotImplementedError
+
+    def early_stop(self, cur_round, is_best=False) -> bool:
+        raise NotImplementedError
+
+
+class SimpleES(AbstractES):
+    def __init__(self, stopping_rounds=10):
```
jwmueller, line 11: Documentation would be helpful, and example of how user can specify this in predictor.fit().
Docstring may also want to explain stopping_rounds is equivalent to early-stopping "patience" as it's called in many libraries.

tabular/src/autogluon/tabular/models/lgb/callbacks.py
```diff
@@ -69,8 +29,9 @@ def early_stopping_custom(stopping_rounds, first_metric_only=False, metrics_to_u
 
     Parameters
     ----------
-    stopping_rounds : int
-       The possible number of rounds without the trend occurrence.
+    stopping_rounds : int or tuple
+       If int, The possible number of rounds without the trend occurrence.
```
jwmueller, line 33: "trend occurrence" should be explained in more detail

core/src/autogluon/core/utils/early_stopping.py
jwmueller: The example of how this can be specified in predictor.fit() could additionally be converted into a unit-test which could also be good to have.

core/src/autogluon/core/utils/early_stopping.py
```diff
@@ -0,0 +1,103 @@
+
```
jwmueller, line 1: Should we caveat these are strategies only being used by certain tree models, and the NN models use different early-stopping strategies?

jwmueller (changes requested):
Functionality looks good, but I think it won't be user-friendly without more detailed documentation & examples.

tabular/src/autogluon/tabular/models/lgb/callbacks.py
Innixma: Would prefer not to edit since this is copy pasted from LightGBM's early stopping function.

core/src/autogluon/core/utils/early_stopping.py
Innixma: The plan is to only mention them on a model-by-model basis. Each model would implement support for it separately. We do want the NN models to eventually use these strategies however. Added mention of it in LightGBM docs.

core/src/autogluon/core/utils/early_stopping.py
Innixma: Renamed stopping_rounds to patience. Added more documentation. Will leave the info on how to specify in .fit for a future PR, the goal is to have defaults strong enough that it should be very rare for a user to need to specify them manually.

core/src/autogluon/core/utils/early_stopping.py
Innixma: Added more documentation and explanation. I think we should separate the `predictor.fit` API documentation from the general low-level components, otherwise the code looks like it is tied to predictor.fit when it is in fact generic. I'd prefer to add support for this functionality to some more models before making the user aware of it.

szha:
Job PR-1042-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1042/3/index.html

jwmueller (approved):
LGTM. The only thing I'd still consider is an example somewhere of how an advanced user could specify some of these ES args via the predictor.fit() API. Otherwise we may be limiting AG's customizability.

PR #1049, opened 2021-04-06
Tabular: Added initial daal4py support

Innixma:
*Issue #, if available:*

*Description of changes:*

Added daal4py support to KNN: https://intelpython.github.io/daal4py/

This results in an average of **25x speed-up** of KNN inference on intel CPUs compared to native sklearn with minimal to no accuracy loss. Combined with the optimized LOO stacking strategy, mainline with daal will be **250x+** faster than v0.1.0 for KNN models.

All that is needed to get the speedup is to install daal4py:

```
pip install autogluon
pip install daal4py
```

Benchmark result (only KNN models):

```
valid_tasks: 550
                                           framework  time_train_s  metric_error  time_infer_s  bestdiff  loss_rescaled  time_infer_s_rescaled      rank  rank=1_count  rank=2_count  rank=3_count  rank>3_count  error_count
0  AutoGluon_bestquality_1h_2021_03_20_knn_v3_KNe...      0.959381      1.196857     13.041495  0.009988       0.235755              25.233313  2.009091           101           310            68            71           54
1  AutoGluon_bestquality_1h_2021_03_31_daal_KNeig...      0.346045      1.209153      0.790459  0.015466       0.288210               1.054045  2.068182           132           259            39           120           11
2  AutoGluon_bestquality_1h_2021_03_20_knn_v3_KNe...      0.912026      1.191187     13.051080  0.030370       0.709032              25.411448  2.907273            74            88            80           308           42
3  AutoGluon_bestquality_1h_2021_03_31_daal_KNeig...      0.523245      1.195930      0.789863  0.033754       0.751086               1.241980  3.015455            29           104           118           299            2
```

Very slight accuracy loss probably from floating points, but a **25x speedup** on inference. This speed-up is even larger on big datasets, where non-daal didn't even train KNN due to running out of time.

### DIFF
--- a/tabular/src/autogluon/tabular/models/knn/_knn_loo_variants.py
+++ b/tabular/src/autogluon/tabular/models/knn/_knn_loo_variants.py
@@ -4,13 +4,7 @@ logger = logging.getLogger(__name__)
 
 import numpy as np
 from scipy import stats
-try:
-    # TODO: Difficult to add user switch for daal usage without duplicating code (Sklearn and daal versions of KNN for predict_loo
-    from daal4py.sklearn.neighbors import KNeighborsClassifier as _KNeighborsClassifier, KNeighborsRegressor as _KNeighborsRegressor
-    # daal backend for KNN seems to be 20-40x+ faster than native sklearn with no downsides.
-    logger.log(15, '\tUsing daal4py KNN backend...')
-except:
-    from sklearn.neighbors import KNeighborsClassifier as _KNeighborsClassifier, KNeighborsRegressor as _KNeighborsRegressor
+from sklearn.neighbors import KNeighborsClassifier as _KNeighborsClassifier, KNeighborsRegressor as _KNeighborsRegressor
 from sklearn.neighbors._base import _get_weights
 from sklearn.utils.extmath import weighted_mode
 
--- a/tabular/src/autogluon/tabular/models/knn/_knn_loo_variants.py
+++ b/tabular/src/autogluon/tabular/models/knn/_knn_loo_variants.py
@@ -1,6 +1,16 @@
+import logging
+
+logger = logging.getLogger(__name__)
+
 import numpy as np
 from scipy import stats
-from sklearn.neighbors import KNeighborsClassifier as _KNeighborsClassifier, KNeighborsRegressor as _KNeighborsRegressor
+try:
+    # TODO: Difficult to add user switch for daal usage without duplicating code (Sklearn and daal versions of KNN for predict_loo
+    from daal4py.sklearn.neighbors import KNeighborsClassifier as _KNeighborsClassifier, KNeighborsRegressor as _KNeighborsRegressor
+    # daal backend for KNN seems to be 20-40x+ faster than native sklearn with no downsides.
+    logger.log(15, '\tUsing daal4py KNN backend...')
+except:
+    from sklearn.neighbors import KNeighborsClassifier as _KNeighborsClassifier, KNeighborsRegressor as _KNeighborsRegressor
 from sklearn.neighbors._base import _get_weights
 from sklearn.utils.extmath import weighted_mode
 
--- a/tabular/src/autogluon/tabular/models/rf/rf_model.py
+++ b/tabular/src/autogluon/tabular/models/rf/rf_model.py
@@ -34,7 +34,9 @@ class RFModel(AbstractModel):
         if self.params_aux.get('use_daal', False):
             # Disabled by default because it appears to degrade performance
             try:
+                # TODO: Use sklearnex instead once a suitable toggle option is provided that won't impact future models
                 from daal4py.sklearn.ensemble import RandomForestClassifier, RandomForestRegressor
+                logger.log(15, '\tUsing daal4py RF backend...')
                 self._daal = True
             except:
                 from sklearn.ensemble import RandomForestClassifier, RandomForestRegressor
--- a/tabular/src/autogluon/tabular/models/rf/rf_model.py
+++ b/tabular/src/autogluon/tabular/models/rf/rf_model.py
@@ -6,8 +6,6 @@ import time
 
 import numpy as np
 import psutil
-from sklearn.ensemble import RandomForestClassifier, RandomForestRegressor
-from .rf_quantile import RandomForestQuantileRegressor
 
 from autogluon.core.constants import BINARY, MULTICLASS, REGRESSION, SOFTCLASS, QUANTILE
 from autogluon.core.utils.exceptions import NotEnoughMemoryError, TimeLimitExceeded
@@ -27,12 +25,25 @@ class RFModel(AbstractModel):
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
         self._feature_generator = None
+        self._daal = False  # Whether daal4py backend is being used
 
     def _get_model_type(self):
+        if self.problem_type == QUANTILE:
+            from .rf_quantile import RandomForestQuantileRegressor
+            return RandomForestQuantileRegressor
+        if self.params_aux.get('use_daal', False):
+            # Disabled by default because it appears to degrade performance
+            try:
+                from daal4py.sklearn.ensemble import RandomForestClassifier, RandomForestRegressor
+                self._daal = True
+            except:
+                from sklearn.ensemble import RandomForestClassifier, RandomForestRegressor
+                self._daal = False
+        else:
+            from sklearn.ensemble import RandomForestClassifier, RandomForestRegressor
+            self._daal = False
         if self.problem_type in [REGRESSION, SOFTCLASS]:
             return RandomForestRegressor
-        elif self.problem_type == QUANTILE:
-            return RandomForestQuantileRegressor
         else:
             return RandomForestClassifier
 
@@ -74,6 +85,9 @@ class RFModel(AbstractModel):
              sample_weight=None,
              **kwargs):
         time_start = time.time()
+
+        model_cls = self._get_model_type()
+
         max_memory_usage_ratio = self.params_aux['max_memory_usage_ratio']
         params = self._get_model_params()
         n_estimators_final = params['n_estimators']
@@ -111,19 +125,27 @@ class RFModel(AbstractModel):
                     params['warm_start'] = True
 
         params['n_estimators'] = n_estimator_increments[0]
-        self.model = self._get_model_type()(**params)
+        if self._daal:
+            if params.get('warm_start', False):
+                params['warm_start'] = False
+
+        model = model_cls(**params)
 
         time_train_start = time.time()
         for i, n_estimators in enumerate(n_estimator_increments):
             if i != 0:
-                self.model.n_estimators = n_estimators
-            self.model = self.model.fit(X, y, sample_weight=sample_weight)
+                if params.get('warm_start', False):
+                    model.n_estimators = n_estimators
+                else:
+                    params['n_estimators'] = n_estimators
+                    model = model_cls(**params)
+            model = model.fit(X, y, sample_weight=sample_weight)
             if (i == 0) and (len(n_estimator_increments) > 1):
                 time_elapsed = time.time() - time_train_start
                 model_size_bytes = 0
-                for estimator in self.model.estimators_:  # Uses far less memory than pickling the entire forest at once
+                for estimator in model.estimators_:  # Uses far less memory than pickling the entire forest at once
                     model_size_bytes += sys.getsizeof(pickle.dumps(estimator))
-                expected_final_model_size_bytes = model_size_bytes * (n_estimators_final / self.model.n_estimators)
+                expected_final_model_size_bytes = model_size_bytes * (n_estimators_final / model.n_estimators)
                 available_mem = psutil.virtual_memory().available
                 model_memory_ratio = expected_final_model_size_bytes / available_mem
 
@@ -150,6 +172,7 @@ class RFModel(AbstractModel):
                     if n_estimator_increments[j] > n_estimators_ideal:
                         n_estimator_increments[j] = n_estimators_ideal
 
+        self.model = model
         self.params_trained['n_estimators'] = self.model.n_estimators
 
     # TODO: Remove this after simplifying _predict_proba to reduce code duplication. This is only present for SOFTCLASS support.
--- a/tabular/setup.py
+++ b/tabular/setup.py
@@ -47,9 +47,13 @@ extras_require = {
         'torch>=1.0,<2.0',
         'fastai>=2.0,<3.0',
     ],
+    'skex': [
+        'scikit-learn-intelex<2021.3',
+    ],
 }
 
 all_requires = []
+# TODO: Consider adding 'skex' to 'all'
 for extra_package in ['lightgbm', 'catboost', 'xgboost', 'fastai']:
     all_requires += extras_require[extra_package]
 all_requires = list(set(all_requires))
--- a/tabular/src/autogluon/tabular/models/knn/knn_model.py
+++ b/tabular/src/autogluon/tabular/models/knn/knn_model.py
@@ -25,6 +25,16 @@ class KNNModel(AbstractModel):
         self._X_unused_index = None  # Keeps track of unused training data indices, necessary for LOO OOF generation
 
     def _get_model_type(self):
+        if self.params_aux.get('use_daal', True):
+            try:
+                # TODO: Add more granular switch, currently this affects all future KNN models even if they had `use_daal=False`
+                from sklearnex import patch_sklearn
+                patch_sklearn("knn_classifier")
+                patch_sklearn("knn_regressor")
+                # daal backend for KNN seems to be 20-40x+ faster than native sklearn with no downsides.
+                logger.log(15, '\tUsing daal4py KNN backend...')
+            except:
+                pass
         try:
             from ._knn_loo_variants import KNeighborsClassifier, KNeighborsRegressor
         except:
--- a/tabular/src/autogluon/tabular/models/lr/lr_model.py
+++ b/tabular/src/autogluon/tabular/models/lr/lr_model.py
@@ -38,6 +38,17 @@ class LinearModel(AbstractModel):
 
     def _get_model_type(self):
         penalty = self.params.get('penalty', 'L2')
+        if self.params_aux.get('use_daal', False):
+            # Disabled by default until more testing is done, appears to give 20x training speedup when enabled
+            try:
+                # TODO: Add more granular switch, currently this affects all future LR models even if they had `use_daal=False`
+                from sklearnex import patch_sklearn
+                patch_sklearn("ridge")
+                patch_sklearn("lasso")
+                patch_sklearn("logistic")
+                logger.log(15, '\tUsing daal4py LR backend...')
+            except:
+                pass
         from sklearn.linear_model import LogisticRegression, Ridge, Lasso
         if self.problem_type == REGRESSION:
             if penalty == 'L2':
--- a/tabular/src/autogluon/tabular/models/xt/xt_model.py
+++ b/tabular/src/autogluon/tabular/models/xt/xt_model.py
@@ -1,7 +1,7 @@
-from sklearn.ensemble import ExtraTreesClassifier, ExtraTreesRegressor
-from ..rf.rf_quantile import ExtraTreesQuantileRegressor
-from ..rf.rf_model import RFModel
+
 from autogluon.core.constants import REGRESSION, QUANTILE
+
+from ..rf.rf_model import RFModel
 
 
 class XTModel(RFModel):
@@ -10,8 +10,11 @@ class XTModel(RFModel):
     """
     def _get_model_type(self):
         if self.problem_type == REGRESSION:
+            from sklearn.ensemble import ExtraTreesRegressor
             return ExtraTreesRegressor
         elif self.problem_type == QUANTILE:
+            from ..rf.rf_quantile import ExtraTreesQuantileRegressor
             return ExtraTreesQuantileRegressor
         else:
+            from sklearn.ensemble import ExtraTreesClassifier
             return ExtraTreesClassifier


szha:
Job PR-1049-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1049/1/index.html

tabular/src/autogluon/tabular/models/knn/_knn_loo_variants.py
```diff
@@ -1,6 +1,16 @@
...
 import numpy as np
 from scipy import stats
-from sklearn.neighbors import KNeighborsClassifier as _KNeighborsClassifier, KNeighborsRegressor as _KNeighborsRegressor
+try:
+    # TODO: Difficult to add user switch for daal usage without duplicating code (Sklearn and daal versions of KNN for predict_loo
+    from daal4py.sklearn.neighbors import KNeighborsClassifier as _KNeighborsClassifier, KNeighborsRegressor as _KNeighborsRegressor
+    # daal backend for KNN seems to be 20-40x+ faster than native sklearn with no downsides.
+    logger.log(15, '\tUsing daal4py KNN backend...')
+except:
+    from sklearn.neighbors import KNeighborsClassifier as _KNeighborsClassifier, KNeighborsRegressor as _KNeighborsRegressor
```
jwmueller, line 13: maybe print tip to install daal4py in order to get faster KNN models?

tabular/src/autogluon/tabular/models/rf/rf_model.py
```diff
@@ -27,12 +25,25 @@ class RFModel(AbstractModel):
...
+        if self.problem_type == QUANTILE:
+            from .rf_quantile import RandomForestQuantileRegressor
+            return RandomForestQuantileRegressor
+        if self.params_aux.get('use_daal', False):
+            # Disabled by default because it appears to degrade performance
+            try:
+                from daal4py.sklearn.ensemble import RandomForestClassifier, RandomForestRegressor
+                self._daal = True
+            except:
+                from sklearn.ensemble import RandomForestClassifier, RandomForestRegressor
```
jwmueller, line 40: maybe print tip to install daal4py in order to get faster RF models?

tabular/src/autogluon/tabular/models/rf/rf_model.py
```diff
@@ -27,12 +25,25 @@ class RFModel(AbstractModel):
...
 
     def _get_model_type(self):
+        if self.problem_type == QUANTILE:
+            from .rf_quantile import RandomForestQuantileRegressor
+            return RandomForestQuantileRegressor
+        if self.params_aux.get('use_daal', False):
+            # Disabled by default because it appears to degrade performance
+            try:
+                from daal4py.sklearn.ensemble import RandomForestClassifier, RandomForestRegressor
+                self._daal = True
```
jwmueller, line 38: Missing analogous statement you had for KNN:

`logger.log(15, '\tUsing daal4py RF backend...')`

jwmueller:
How are users supposed to know to install daal4py?

tabular/src/autogluon/tabular/models/rf/rf_model.py
Innixma: daal4py actually slowed RF models down from the tests I've ran. This is why its disabled by default until more testing is done.

SmirnovEgorRu:
Hi @Innixma, I'm one of daal4py maintainers. I appreciate that you are adding this into the project, I'm happy to help you with the activity :)

You mentioned that you have some issues with kNN and RF, it's interesting. What version of daal4py did you use? These issues are expected to be solved in the latest version 2021.2. Do you have some tests/benchmarks to reproduce the issues?

Also, I want to notice that we are extracting scikit-learn acceleration from daal4py to [**Intel Extension for Scikit-learn**](https://github.com/intel/scikit-learn-intelex) package. At the moment daal4py contains such functionality, but it will be removed in the future and will be available in Intel Extension for Scikit-learn only. There is no difference in functionality and performance between daal4py and Intel Extension for Scikit-learn (simply uses daal4py underneath). I prefer to use this package instead of daal4py, hopefully, it's easy to do :)

Installation:
```sh
pip install scikit-learn-intelex
```

Also, the main feature of Intel Extension for Scikit-learn - you don't need to change existing Scikit-learn code, but just call ```patch_sklearn``` function. Without arguments (```patch_sklearn()```) it patches all Scikit-learn algorithms which can be accelerated, but we can specify concrete estimators what we want to accelerate. The patches are just runtime, they don't change scikit-learn package permanently. It might look like this:
```py
from sklearnex import patch_sklearn
patch_sklearn("knn_classifier")
patch_sklearn("knn_regressor")
from sklearn.neighbors import KNeighborsClassifier as _KNeighborsClassifier, KNeighborsRegressor as _KNeighborsRegressor
```
[List of available algorithm](https://github.com/intel/scikit-learn-intelex/blob/f8c21bd424d2a249f5858cf69c250979f97bdaa9/daal4py/sklearn/monkeypatch/dispatcher.py#L77) to patch this way

Innixma:
Hi @SmirnovEgorRu,

Thanks to you and your team for the excellent package! I will have to take a look at `scikit-learn-intelex` and likely update this PR to use it.

I have a few questions regarding DAAL:
1. Say I wanted to train 2 models side-by-side, one with DAAL and one without. I'm assuming I would want to avoid monkey-patching in this case, is that correct?
2. Can I convert a trained DAAL model to a sklearn model? (say I am deploying to a machine that lacks DAAL's requirements)
3. What is the best way to check within python if the machine satisfies the requirements for DAAL acceleration? (How do I tell if I have a valid CPU?)
4. Is there a way to dynamically install `scikit-learn-intelex` from `pip install autogluon` iff the user has a valid intel CPU? If not, would you recommend an optional dependency option such as `pip install autogluon[daal]`?
5. Is there any benefit of `scikit-learn-intelex` when running on alternative CPUs? Does it error?

I also have a few questions regarding the acceleration:
1. I noticed random forest used sklearn version for predict_proba. Why is this? Generally, we only call predict_proba in AutoGluon for the individual models as we do a variety of post-processing steps such as weighted averaging.
2. When I did a raw test of `model.predict` on DAAL vs non-DAAL random-forest, DAAL was slower. Is this expected? Do you have an example notebook / script that shows DAAL speeding up RF? This would be of great interest to us.
3. I noticed that KNN had slightly worse accuracy when accelerated with DAAL (see main comment). Is this expected? Is it due to floating point error?

SmirnovEgorRu:
@Innixma, thank you!

> Is there any benefit of scikit-learn-intelex when running on alternative CPUs? Does it error?

It works on any x86 architectures, so it works on AMD as well as on Intel CPUs and also provides performance gains, but expected to be better on Intel CPUs (we have AVX-512 branches of code and other staff). ARM isn't supported and we don't provide built packages for this, so we just can't import Intel Extension for Scikit-learn package on ARM. 
Does this behavior satisfy your requirements?

> What is the best way to check within python if the machine satisfies the requirements for DAAL acceleration? (How do I tell if I have a valid CPU?)

Our patching is a "seamless way" to accelerate scikit-learn. If we don't accelerate some Scikit-learn functionality or don't support some algorithms parameters - we just use original Scikit-learn implementation inside the patching. So, all Scikit-learn algorithms work always as expected, but acceleration is achieved if parameters allow this.
We have [this table](https://intel.github.io/scikit-learn-intelex/algorithms.html) which show what we are able to accelerate and some limitations.
To check an algorithm was accelerated or not - we provide [a env variable](https://intel.github.io/scikit-learn-intelex/verbose.html) (```SKLEARNEX_VERBOSE=INFO```) which enables printing info such:
```
INFO: sklearn.cluster.DBSCAN.fit: uses Intel(R) oneAPI Data Analytics Library solver
or
INFO: sklearn.cluster.DBSCAN.fit: uses original Scikit-learn solver
```

Do you want to know what implementation is called exactly in python code rather than printing?

> Can I convert a trained DAAL model to a sklearn model? (say I am deploying to a machine that lacks DAAL's requirements)

At the moment - no. If you use pickle to serialize an estimator - it's expected to have Intel Extension for Scikit-learn on deploy stage as well. Is it critical?

> Is there a way to dynamically install scikit-learn-intelex from pip install autogluon iff the user has a valid intel CPU? If not, would you recommend an optional dependency option such as pip install autogluon[daal]?

If you split python packages to x86 and ARM specific: the extension works on any x86 architectures and can be used installed for this for x86 packages only.
Instead of ```autogluon[daal]``` I think better to use ```autogluon[skex]``` or ```autogluon[onedal]```. Because oneDAL is a new name of DAAL and extend it to run this on Intel GPUs.

> I noticed random forest used sklearn version for predict_proba. Why is this? Generally, we only call predict_proba in AutoGluon for the individual models as we do a variety of post-processing steps such as weighted averaging.

We want to add patching of predict_proba in the next release :)

To better answer on questions about kNN and RF - let me provide run some benchmarks and provide results. But my expectation is that mentioned issues have been solved in version 2021.2 recently released. The same for training 2 models side-by-side, let me think how to better to do it.

Innixma:
@SmirnovEgorRu 

I've created [a script](https://pastebin.com/raw/FdUkxrq9) which shows the behavior I was seeing in regards to performance for random forest. I also tried to create a [Colab notebook](https://colab.research.google.com/drive/1NKWfV1JUCii-_MR8CG-ckeZirv_V-rBS?usp=sharing), but for some reason the notebook failed to detect daal4py / sklearnex.

From the output below, the difference in score between DAAL and sklearn random forest is much larger than what would be caused by chance. Also, logistic regression appears to not use DAAL during training.

Output (ran on a p3.2xlarge, not using GPU, latest versions of daal/sklearnex/sklearn):

```
Sklearn RF (Native):
Score:             0.81
Training Time (s): 1.4285519123077393
Infer Time (s):    0.20345067977905273

Sklearn RF (DAAL):
INFO:root:sklearn.ensemble.RandomForestClassifier.fit: uses Intel(R) oneAPI Data Analytics Library solver
INFO:root:sklearn.ensemble.RandomForestClassifier.predict: uses Intel(R) oneAPI Data Analytics Library solver
Score:             0.7842
Training Time (s): 0.6954841613769531
Infer Time (s):    0.09650731086730957

Sklearn LR (Native):
Score:             0.704
Training Time (s): 120.02899146080017
Infer Time (s):    0.0036160945892333984

Sklearn LR (DAAL):
INFO:root:sklearn.linear_model.LogisticRegression.predict: uses Intel(R) oneAPI Data Analytics Library solver
Score:             0.704
Training Time (s): 119.7056519985199
Infer Time (s):    0.00391840934753418

Sklearn KNN (Native):
Score:             0.759
Training Time (s): 0.0025246143341064453
Infer Time (s):    5.370389938354492

Sklearn KNN (DAAL):
INFO:root:sklearn.neighbors.KNeighborsMixin.kneighbors: uses Intel(R) oneAPI Data Analytics Library solver
INFO:root:sklearn.neighbors.KNeighborsClassifier.predict: uses Intel(R) oneAPI Data Analytics Library solver
Score:             0.759
Training Time (s): 0.005175352096557617
Infer Time (s):    0.07428693771362305

```

Innixma:
In regards to your comments:

> Does this behavior satisfy your requirements?

Yes, we don't expect ARM support yet, and it seems most other CPUs should be supported.

Regarding `How do I tell if I have a valid CPU?`, it seems the previous answer that sklearnex works on nearly all CPU types is enough to answer that question, I don't think we need to check if they have an ARM CPU.

> At the moment - no. If you use pickle to serialize an estimator - it's expected to have Intel Extension for Scikit-learn on deploy stage as well. Is it critical?

This is fine. The fact that daal works on most CPUs means this isn't much of a concern.

> We want to add patching of predict_proba in the next release :)

Great! Are you also planning to add [ExtraTrees](https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.ExtraTreesClassifier.html#sklearn.ensemble.ExtraTreesClassifier) support? The sklearn models we most frequently use are RF, KNN, and XT (ExtraTrees). Also, sklearn now has a [MLP model](https://scikit-learn.org/stable/modules/generated/sklearn.neural_network.MLPClassifier.html#sklearn.neural_network.MLPClassifier). Are there any plans for CPU acceleration of the MLP model in sklearn?

tabular/src/autogluon/tabular/models/rf/rf_model.py
Innixma: Added

tabular/src/autogluon/tabular/models/knn/_knn_loo_variants.py
Innixma: will do this in future PR once we have a better idea of how prominently we want to encourage normal users to install daal. It's possible we add it to `autogluon.tabular[all]` and the default `autogluon` install, although the size of the package and its dependencies are somewhat large (~500MB - 1 GB).

Innixma:
@jwmueller 

> How are users supposed to know to install daal4py?

They aren't yet. This is the initial PR adding support, we'd want to do more benchmarking and add more algorithms before mentioning to users. It is possible this could be added to the default installation.

szha:
Job PR-1049-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1049/2/index.html

szha:
Job PR-1049-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1049/3/index.html

SmirnovEgorRu:
@Innixma, I reproduced the issue with RF. It's unexpected. I ran my several benchmarks for RF and accuracy is similar to scikit-learn. Let me investigate the case more precisely. Results what I have on my benchmarks:

Dataset | Original Scikit-learn, accuracy | Intel Extesnion for Scikit-learn, accuracy | Ratio
-- | -- | -- | --
ijcnn | 99.515 | 99.460 | 0.9994
a9a | 82.967 | 82.956 | 0.9999
gisette | 95.000 | 95.500 | 1.0053
klaverjas | 88.027 | 88.011 | 0.9998
connect | 82.564 | 82.889 | 1.0039
mnist | 95.770 | 95.880 | 1.0011
sensit | 84.396 | 84.441 | 1.0005
skin_segmentation | 99.929 | 99.929 | 1.0000
codrnanorm | 98.328 | 98.230 | 0.9990
Average |  | | 1.0010045


About Logistic Regression acceleration. I used this code to accelerate Scikit-learn and if works as expected (training is optimized):
```py
from sklearnex import patch_sklearn
patch_sklearn("logistic")
from sklearn.linear_model import LogisticRegression as LogisticRegression # estimator is optimized now
fit_model(LogisticRegression, **fit_args)
````

To enable default behavior of scikit-learn back - you can use:
```py  
from sklearnex import unpatch_sklearn
unpatch_sklearn()                                   # or unpatch_sklearn("logistic") for log reg only 
from sklearn.linear_model import LogisticRegression # re-importing to use original Scikit-learn
fit_model(LogisticRegression, **fit_args)
```

> Great! Are you also planning to add ExtraTrees support? The sklearn models we most frequently use are RF, KNN, and XT (ExtraTrees). Also, sklearn now has a MLP model. Are there any plans for CPU acceleration of the MLP model in sklearn?

ExtraTrees were planned for the second part of the year, MLP is out of plans at the moment. But we are able re-prioritize plans.  My understanding that ExtraTrees are quite fast already (if compare with RF for example). Are these algorithms hotspots in your automl in terms of execution time?

Also, we have quite fast SVM algorithm. It's not represented in your framework, but anyway it's powerful algorithm. This [blog](https://medium.com/intel-analytics-software/from-hours-to-minutes-600x-faster-svm-647f904c31ae) contains some insights of our implementation.

Innixma:
@SmirnovEgorRu, regarding the patching, I believe the re-import of LogisticRegression after unpatching does not actually have any effect, as the previously imported LogisticRegression was updated. my understanding is:

```
from sklearnex import patch_sklearn
patch_sklearn("logistic")
from sklearn.linear_model import LogisticRegression # estimator is optimized now
model_daal = fit_model(LogisticRegression, **fit_args)
from sklearnex import unpatch_sklearn
unpatch_sklearn("logistic")
from sklearn.linear_model import LogisticRegression # re-importing to use original Scikit-learn
model_sklearn = fit_model(LogisticRegression, **fit_args)

# model_daal is accelerated
# model_sklearn is not accelerated
# this is ideal
```

However,

```
from sklearnex import patch_sklearn
patch_sklearn("logistic")
from sklearn.linear_model import LogisticRegression # estimator is optimized now
unpatch_sklearn("logistic") # I want to limit the scope of patching, for example maybe I don't want to init the model right away.
# estimator is no longer optimized despite not re-importing
model_daal = fit_model(LogisticRegression, **fit_args) 
from sklearn.linear_model import LogisticRegression # re-importing to use original Scikit-learn
model_sklearn = fit_model(LogisticRegression, **fit_args)

# model_daal is not accelerated
# model_sklearn is not accelerated
# this is not ideal
```

A code example of this:

```
from sklearn.linear_model import LogisticRegression

print('Sklearn LR (Native):')
fit_model(LogisticRegression, **fit_args)

patch_sklearn('logistic')
from sklearn.linear_model import LogisticRegression

print('Sklearn LR (DAAL):')
fit_model(LogisticRegression, **fit_args)

print('Sklearn LR 2 (DAAL):')
fit_model(LogisticRegression, **fit_args)

unpatch_sklearn('logistic')
print('Sklearn LR 3 (Native -> Should be DAAL):')
fit_model(LogisticRegression, **fit_args)

from sklearn.linear_model import LogisticRegression
print('Sklearn LR 4 (Native):')
fit_model(LogisticRegression, **fit_args)
```

and output:

```
Sklearn LR (Native):
Score:             0.7256
Training Time (s): 6.198185443878174
Infer Time (s):    0.0037772655487060547

Intel(R) Extension for Scikit-learn* enabled (https://github.com/intel/scikit-learn-intelex)
Sklearn LR (DAAL):
INFO:root:sklearn.linear_model.LogisticRegression.fit: uses Intel(R) oneAPI Data Analytics Library solver
Score:             0.7256
Training Time (s): 0.5205447673797607
Infer Time (s):    0.010569334030151367

Sklearn LR 2 (DAAL):
INFO:root:sklearn.linear_model.LogisticRegression.fit: uses Intel(R) oneAPI Data Analytics Library solver
Score:             0.7256
Training Time (s): 0.47805118560791016
Infer Time (s):    0.0067026615142822266

Sklearn LR 3 (Native -> Should be DAAL):
Score:             0.7256
Training Time (s): 6.1781089305877686
Infer Time (s):    0.003770589828491211

Sklearn LR 4 (Native):
Score:             0.7256
Training Time (s): 6.188448429107666
Infer Time (s):    0.0038619041442871094
```

In this sense, I want to be able to have an equivalent to `from sklearnex.foo import LogisticRegression` (DAAL accelerated)

However, it seems it is different:

Currently:
```
# This is effectively identical to just importing sklearn LogisticRegression

patch_sklearn("logistic")
from sklearn.linear_model import LogisticRegression # daal
unpatch_sklearn("logistic")  # -> silently reverts LogisticRegression to sklearn
```

```
from sklearn.linear_model import LogisticRegression # sklearn
patch_sklearn("logistic")  # -> does nothing until LogisticRegression is re-imported
```

Ideally:
```
# unpatch should function in the same manner as patch: does not alter already imported modules
patch_sklearn("logistic")
from sklearn.linear_model import LogisticRegression # estimator is optimized now
unpatch_sklearn("logistic")  # -> does nothing until LogisticRegression is re-imported
model_daal = fit_model(LogisticRegression, **fit_args)  # fits DAAL accelerated
from sklearn.linear_model import LogisticRegression # sklearn version
model_sklearn = fit_model(LogisticRegression, **fit_args) 
```

This would be nice for our use-case, as we may do steps inbetween determining the model class and actually initializing the model. If we were to run into an exception between those two steps, we wouldn't have had a chance to call `unpatch`, meaning the DAAL patch will propagate to future model training, even if they didn't intend to use DAAL.

This could be solved by having an explicit `from sklearnex.foo import LogisticRegression` as an alternative to the patching API (personally I would prefer this in most cases).

Innixma:
@SmirnovEgorRu Regarding model priorities,

We are definitely interested in trying out algorithms which have gotten significant speedups via DAAL (SVM, linear) which we had previously discarded / disabled due to being too slow.

In terms of CPU acceleration, the priority of model speedups are as follows:

Neural Networks > LightGBM > KNN > CatBoost > XGBoost > RandomForest > ExtraTrees > Linear > Others

KNN has been optimized by 250x since our last release (25x by DAAL, 10x by a separate optimization), so it went from the most costly inference model to one of the least.

Now, the most costly models are neural networks (training speed / inference speed) and RandomForest/ExtraTrees (inference speed). Of course, any improvements to the GBM models' training speed would be highly important, but I suspect they are already heavily optimized. While we don't yet use the sklearn NN model, it is apparently [quite good](https://mljar.com/blog/tensorflow-vs-scikit-learn/) (2x faster training than TensorFlow equivalent). If something were to somehow give a 25x speedup (or even a 2x speedup!) to tabular NN models which don't benefit much from GPUs due to their small dimensions, this would dramatically improve AutoGluon.

Of course, given a big enough speedup (such as 100x+ with SVM), other models could obviously jump in priority as well.

jwmueller (approved):
LGTM!

SmirnovEgorRu:
Thank you, for the comprehensive report! We have never looked at MLP in sklearn. Let us do it and understand feasibility to improve this.
&nbsp;

Hopefully, my team is accelerating training of CatBoost (2 recent releases [0.24.4](https://github.com/catboost/catboost/releases/tag/v0.24.4) and [0.25](https://github.com/catboost/catboost/releases/tag/v0.25)) and XGBoost ([blog](https://medium.com/intel-analytics-software/new-optimizations-for-cpu-in-xgboost-1-1-81144ea21115)) by commits in mainlines as well :)

For faster prediction of LGBM and XGBoost - yes, we have model convertors from daal4py ([blog](https://medium.com/intel-analytics-software/improving-the-performance-of-xgboost-and-lightgbm-inference-3b542c03447e), [xgboost example](https://github.com/intel/scikit-learn-intelex/blob/master/examples/gbt_cls_model_create_from_xgboost_batch.py), [lgbm example](https://github.com/intel/scikit-learn-intelex/blob/master/examples/gbt_cls_model_create_from_lightgbm_batch.py)). Current limitation - we don't support missing values and categorical features.

 &nbsp;
>This could be solved by having an explicit from sklearnex.foo import LogisticRegression as an alternative to the patching API (personally I would prefer this in most cases).

Yep, we are planning to move this feature from daal4py to Intel Extension for Scikit-learn in the next release as well :)

napetrov:
Couple additional comments
1. Intel Extension for Scikit-learn also provides Intel GPU support - it's might be not so relevant for cloud environment at the moment(although all current GPU devices are supported already), but going forward this would be easiest way  to get Scikit-learn enabled there.

2. For NN cases - there is [oneDNN](https://github.com/oneapi-src/oneDNN) library that provides optimizations for TensorFlow and PyTorch. So they can be used for massive perf gains
[Intel® Optimization for TensorFlow](https://software.intel.com/content/www/us/en/develop/articles/intel-optimization-for-tensorflow-installation-guide.html)
[Intel® Optimization for PyTorch](https://software.intel.com/content/www/us/en/develop/articles/getting-started-with-intel-optimization-of-pytorch.html)

PR #1052, opened 2021-04-09
Removed version specification on pyarrow to allow RAPIDS install

Innixma:
*Issue #, if available:*

*Description of changes:*

RAPIDS depends on a special version of pyarrow not available via pip install, so if autogluon.text was installed after RAPIDS, RAPIDS would break. This change allows the RAPIDS version of pyarrow to remain.

@sxjscience What do you think about removing pyarrow as a dependency? Could it be moved to optional?

### DIFF
--- a/text/setup.py
+++ b/text/setup.py
@@ -26,7 +26,7 @@ requirements = [
 
     f'autogluon.core=={version}',
     f'autogluon.mxnet=={version}',
-    'pyarrow>=3.0.0',
+    'pyarrow',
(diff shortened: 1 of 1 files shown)


sxjscience:
We can move it to optional I think.

szha:
Job PR-1052-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1052/1/index.html

sxjscience (approved):
LGTM

Issue #1051, opened 2021-04-09
Memory error in official kaggle tutorial


qo4on:
`TabularPredictor().fit` from [this tutorial](https://auto.gluon.ai/stable/tutorials/tabular_prediction/tabular-kaggle.html) doesn't work in `Colab` or in `Kaggle` because of the memory error.
Kaggle has 16.81 GB of available RAM. Autogluon log shows that it needs much less memory: `Train Data (Original)  Memory Usage: 2715.97 MB`
What is the problem?

Innixma:
To clarify, the data taking 2.7 GB of memory doesn't mean that AutoGluon won't need more than that to actually train models. You'll likely need 32 GB of memory (such as an m5.2xlarge).

I think it might be good for the tutorial to be updated so it can be run on 16 GB of memory. For example, we have a kaggle kernel that gets top 1% in the Otto competition: https://www.kaggle.com/innixma/autogluon-rapids-top-1

qo4on:
Thank you for your answer. I understand that, but the error occures in `Fitting AsTypeFeatureGenerator...`
Are you sure that 11.3 GB of available Colab memory is not enough for feature generation? I successfully trained on 200 000 sampled examples and most of the time memory usage was far from the limit.

Also it's not clear what is the right way for installing AutoGluon. For `Colab` it kind of works with some warnings:
```
!python3 -m pip install -U pip > /dev/null
!python3 -m pip install -U setuptools wheel > /dev/null
!python3 -m pip install -U "mxnet<2.0.0" > /dev/null
!python3 -m pip install autogluon > /dev/null

ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
datascience 0.10.6 requires folium==0.2.1, but you have folium 0.8.3 which is incompatible.
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
datascience 0.10.6 requires folium==0.2.1, but you have folium 0.8.3 which is incompatible.
albumentations 0.1.12 requires imgaug<0.2.7,>=0.2.5, but you have imgaug 0.2.9 which is incompatible.
```
The same commands in `Kaggle` return more warnings:
```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
[...]

PR #1055, opened 2021-04-10
Remove bokeh dependency

Innixma:
*Issue #, if available:*
#1051

*Description of changes:*

Removes bokeh dependency in autogluon.extra, which was only used optionally for plots. Since it caused numerous issues related to tornado dependencies in Kaggle (crash on importing AutoGluon), better to have it be purely optional, as a warning is shown to users who try to make plots telling them to install it.

### DIFF
--- a/extra/setup.py
+++ b/extra/setup.py
@@ -25,7 +25,6 @@ requirements = [
     'gluoncv',
     'graphviz',
 
-    'bokeh',  # TODO: Remove bokeh, it should be purely optional and specific to core, not extra
     f'autogluon.core=={version}',
 ]
 


szha:
Job PR-1055-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1055/1/index.html

jwmueller:
Should there be some easy way for users to simultaneously install all optional dependencies like bokeh and others?
It can be inconvenient to have to first encounter the error message instructing how to install each one of these optional dependencies.  Might be nicer to have a full install (including all optional dependencies) available via something like this: 

```
pip install autogluon  # installs everything except autogluon.extra
pip install autogluon.extra
```

If we want the ENAS code to be non-optional, then it could be in `autogluon.nas` subpackage, which anyway is a clearer name than `.extra`

jwmueller (approved):

Innixma:
Will merge for now, and we can refactor how we want to deal with generic optional dependencies in future

PR #1066, opened 2021-04-16
Tabular: Added efficient OOF functionality to RF/XT models

Innixma:
*Issue #, if available:*

*Description of changes:*

- Added efficient OOF functionality to RF/XT models
- This should give a 10x train/infer speed-up and 10x disk space reduction in bagged modes, with roughly equivalent model quality.
- DAAL and RAPIDS don't support the necessary methods to do this, so only raw sklearn is accelerated at present.
- Fixed efficient OOF KNN crashing on regression problems.

TODO:

- [x] Benchmark

### DIFF
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -483,7 +483,12 @@ class AbstractModel:
         For multiclass and softclass classification, keeps y_pred_proba as a 2 dimensional array of prediction probabilities for each class.
         For regression, converts y_pred_proba to a 1 dimensional array of predictions.
         """
-        if self.problem_type == BINARY:
+        if self.problem_type == REGRESSION:
+            if len(y_pred_proba.shape) == 1:
+                return y_pred_proba
+            else:
+                return y_pred_proba[:, 1]
+        elif self.problem_type == BINARY:
             if len(y_pred_proba.shape) == 1:
                 return y_pred_proba
             elif y_pred_proba.shape[1] > 1:
@@ -492,7 +497,7 @@ class AbstractModel:
                 return y_pred_proba
         elif y_pred_proba.shape[1] > 2:  # Multiclass, Softclass
             return y_pred_proba
-        else:  # Regression
+        else:  # Unknown problem type
             return y_pred_proba[:, 1]
 
     def score(self, X, y, metric=None, sample_weight=None, **kwargs):
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -498,7 +498,7 @@ class AbstractModel:
         elif y_pred_proba.shape[1] > 2:  # Multiclass, Softclass
             return y_pred_proba
         else:  # Unknown problem type
-            return y_pred_proba[:, 1]
+            raise AssertionError(f'Unknown y_pred_proba format for `problem_type="{self.problem_type}"`.')
 
     def score(self, X, y, metric=None, sample_weight=None, **kwargs):
         if metric is None:
--- a/tabular/src/autogluon/tabular/models/rf/rf_model.py
+++ b/tabular/src/autogluon/tabular/models/rf/rf_model.py
@@ -245,8 +245,9 @@ class RFModel(AbstractModel):
 
         # TODO: Regression does not return NaN for missing rows, instead it sets them to 0. This makes life hard.
         #  The below code corrects the missing rows to NaN instead of 0.
-        # Don't bother if >40 trees, near impossible to have missing
-        if self.problem_type == REGRESSION and self.model.n_estimators <= 40:
+        # Don't bother if >60 trees, near impossible to have missing
+        # If using 68% of data for training, chance of missing for each row is 1 in 11 billion.
+        if self.problem_type == REGRESSION and self.model.n_estimators <= 60:
             from sklearn.ensemble._forest import _get_n_samples_bootstrap, _generate_unsampled_indices
             n_samples = len(y)
 
@@ -292,7 +293,7 @@ class RFModel(AbstractModel):
     @classmethod
     def _get_default_ag_args_ensemble(cls) -> dict:
         default_ag_args_ensemble = super()._get_default_ag_args_ensemble()
-        extra_ag_args_ensemble = {'use_child_oof': True}
+        extra_ag_args_ensemble = {'use_child_oof': False}
         default_ag_args_ensemble.update(extra_ag_args_ensemble)
         return default_ag_args_ensemble
 
--- a/tabular/src/autogluon/tabular/models/rf/rf_model.py
+++ b/tabular/src/autogluon/tabular/models/rf/rf_model.py
@@ -10,6 +10,7 @@ import psutil
 from autogluon.core.constants import BINARY, MULTICLASS, REGRESSION, SOFTCLASS, QUANTILE
 from autogluon.core.utils.exceptions import NotEnoughMemoryError, TimeLimitExceeded
 from autogluon.core.features.types import R_OBJECT
+from autogluon.core.utils.utils import normalize_pred_probas
 
 from autogluon.core.models.abstract.model_trial import skip_hpo
 from autogluon.core.models import AbstractModel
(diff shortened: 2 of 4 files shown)


szha:
Job PR-1066-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1066/1/index.html

Innixma:
Benchmark Results:

```
                                           framework  > AutoGluon_bestquality_1h_2021_04_15_rf_oof_opt_v2  < AutoGluon_bestquality_1h_2021_04_15_rf_oof_opt_v2  = AutoGluon_bestquality_1h_2021_04_15_rf_oof_opt_v2  % Less Avg. Errors Than AutoGluon_bestquality_1h_2021_04_15_rf_oof_opt_v2  time_train_s  metric_error  time_infer_s  loss_rescaled  time_train_s_rescaled  time_infer_s_rescaled      rank  rank=1_count  rank=2_count  rank=3_count  rank>3_count  error_count
0    AutoGluon_bestquality_1h_2021_04_14_upgrade_cat                                                499                                                  465                                                   55                                            -0.046997                           3403.159470  69921.773190     73.566928       0.463199               1.046596               2.757984  1.483317           500           520             0             0            2
1  AutoGluon_bestquality_1h_2021_04_15_rf_oof_opt_v2                                                  0                                                    0                                                 1021                                             0.000000                           3346.319431  69072.775867     34.807949       0.499509               1.019751               1.185171  1.516683           467           554             0             0            1
```

While we had a slight model quality drop, we reduced average latency by 2.3x and disk usage by likely ~5-10x.

The reasons for the drop are likely due to 2 causes:

1. Only 300 trees are used instead of 3000+ trees for RF and XT models in bagging. This means if RF/XT are the dominant model on a problem, it's likely the old method produces a stronger result.
- While each tree does get more data with the new method, it is unlikely that it will be enough to overcome having 3000+ trees from the old method. This is evident when looking deeper into the results: The new method is stronger on larger datasets because those datasets didn't have time to train the 3000+ trees (or it did but then neglected other models), whereas it is weaker on smaller datasets where repeated bagging often caused RF and XT to have 10000+ trees compared to only 300.
- A possible solution is to train more trees, such as 600, by default.
- Another option is to enable repeated-bagging for this model, adding more trees to the model each time (but not actually bagging).

2. XT is now using bootstrapping, where it previously was not. Bootstrapping is required for the efficient OOF calculation, and this means XT may be weaker in general than before, as it uses less data to train its trees.
- A possible solution would be to re-train XT without bootstrapping and use the bootstrapped version for training, and the non-bootstrapped version for inference.
- Alternatively, adjust the bootstrapping fraction to 10% instead of 38% to get similar strength trees as before. Downside: OOF score will be weaker due to fewer OOB samples. This drawback could be mitigated if we increased num trees to 600.

Innixma:
Overall, despite the slight accuracy drop, I recommend we merge this. The follow-ups can be investigated and even as is, the inference speed and disk usage improvements are enough to make this change worthwhile.

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -492,7 +497,7 @@ def _convert_proba_to_unified_form(self, y_pred_proba):
                 return y_pred_proba
         elif y_pred_proba.shape[1] > 2:  # Multiclass, Softclass
             return y_pred_proba
-        else:  # Regression
+        else:  # Unknown problem type
```
jwmueller, line 500: maybe should be marked with TODO?

tabular/src/autogluon/tabular/models/rf/rf_model.py
```diff
@@ -191,6 +195,81 @@ def _predict_proba(self, X, **kwargs):
...
+            normalize = self.normalize_pred_probas
+        if normalize:
+            y_oof_pred_proba = normalize_pred_probas(y_oof_pred_proba, self.problem_type)
+        y_oof_pred_proba = y_oof_pred_proba.astype(np.float32)
+        return y_oof_pred_proba
+
+    # FIXME: Unknown if this works with quantile regression
+    def _get_oof_pred_proba(self, X, y, **kwargs):
+        if self._daal:
+            raise AssertionError('DAAL forest backend does not support out-of-bag predictions.')
```
jwmueller, line 211: These errors may be more helpful if they describe an alternative for the user.

tabular/src/autogluon/tabular/models/rf/rf_model.py
```diff
@@ -191,6 +195,81 @@ def _predict_proba(self, X, **kwargs):
...
+                    self.model.n_classes_ = self.model.n_classes_[0]
+
+        if getattr(self.model, "oob_decision_function_", None) is not None:
+            y_oof_pred_proba = self.model.oob_decision_function_
+            self.model.oob_decision_function_ = None  # save memory
+        elif getattr(self.model, "oob_prediction_", None) is not None:
+            y_oof_pred_proba = self.model.oob_prediction_
+            self.model.oob_prediction_ = None  # save memory
+        else:
+            raise AssertionError(f'Model class {type(self.model)} does not support out-of-fold prediction generation.')
```
jwmueller, line 244: These errors may be more helpful if they describe an alternative for the user.

jwmueller (approved):
LGTM. I'll check with Taesup regarding possible Quantile Regression functionality issues. 

A remaining concern I have is the user will not understand what special behavior is happening with this RF model / KNN model. It may be worth at least adding a section to the Tabular FAQ page describing this, and possibly explain how the user could instead use regular bagging for these models instead of this special leave-some-out functionality (even though this is not typically recommended).

tabular/src/autogluon/tabular/models/rf/rf_model.py
```diff
@@ -191,6 +195,81 @@ def _predict_proba(self, X, **kwargs):
...
+        if self.problem_type == REGRESSION and self.model.n_estimators <= 40:
+            from sklearn.ensemble._forest import _get_n_samples_bootstrap, _generate_unsampled_indices
+            n_samples = len(y)
+
+            n_predictions = np.zeros(n_samples)
+            n_samples_bootstrap = _get_n_samples_bootstrap(n_samples, self.model.max_samples)
+            for estimator in self.model.estimators_:
+                unsampled_indices = _generate_unsampled_indices(estimator.random_state, n_samples, n_samples_bootstrap)
+                n_predictions[unsampled_indices] += 1
+            missing_row_mask = n_predictions == 0
```
gradientsky, line 258: what about legit zeroes? For example you predict a temperature between -100 and 100?

gradientsky: Another point: do we want exact 0 here? Float point errors can do this:
```
0.0003 - 0.0002 - 0.0001 == 0
```
This results in `False`

core/src/autogluon/core/models/abstract/abstract_model.py
Innixma: updated

tabular/src/autogluon/tabular/models/rf/rf_model.py
```diff
@@ -191,6 +195,81 @@ def _predict_proba(self, X, **kwargs):
...
+        elif getattr(self.model, "oob_prediction_", None) is not None:
+            y_oof_pred_proba = self.model.oob_prediction_
+            self.model.oob_prediction_ = None  # save memory
+        else:
+            raise AssertionError(f'Model class {type(self.model)} does not support out-of-fold prediction generation.')
+
+        # TODO: Regression does not return NaN for missing rows, instead it sets them to 0. This makes life hard.
+        #  The below code corrects the missing rows to NaN instead of 0.
+        # Don't bother if >40 trees, near impossible to have missing
+        if self.problem_type == REGRESSION and self.model.n_estimators <= 40:
```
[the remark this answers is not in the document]
Innixma: updated to 60 trees, now 1 in 11 billion chance

tabular/src/autogluon/tabular/models/rf/rf_model.py
Innixma: will plan to add more info when DAAL gets further along.

tabular/src/autogluon/tabular/models/rf/rf_model.py
Innixma: Generally user should never see this unless they are doing very advanced things.

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -492,7 +497,7 @@ def _convert_proba_to_unified_form(self, y_pred_proba):
...
             return y_pred_proba[:, 1]
```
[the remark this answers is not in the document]
Innixma: Added

Innixma:
I've disabled the OOB OOF opt for now until we pin down the impact related to stack-overfitting.

PR #1075, opened 2021-04-20
Tabular: Adjusted per-level stack time

Innixma:
*Issue #, if available:*

*Description of changes:*

- Added more time to early levels and less to later levels.
- For 2 level stacking, level 1 gets 2/3rds of the time, and level 2 gets 1/3rd of the time.
- This helps ensure good models are trained in level 1, which also improves level 2 models.

Benchmark results show that this is an improvement:

```
                                           framework  > AutoGluon_bestquality_1h_2021_04_16_stack_time  < AutoGluon_bestquality_1h_2021_04_16_stack_time  = AutoGluon_bestquality_1h_2021_04_16_stack_time
0     AutoGluon_bestquality_1h_2021_04_16_stack_time                                                 0                                                 0                                              1028
1  AutoGluon_bestquality_1h_2021_04_15_rf_oof_opt_v2                                               437                                               515                                                68
```

### DIFF
--- a/core/src/autogluon/core/_setup_utils.py
+++ b/core/src/autogluon/core/_setup_utils.py
@@ -14,11 +14,11 @@ PYTHON_REQUIRES = '>=3.6, <3.9'
 
 # Only put packages here that would otherwise appear multiple times across different module's setup.py files.
 DEPENDENT_PACKAGES = {
-    'numpy': '==1.19.5',
+    'numpy': '==1.19.5',  # TODO: v0.3 consider upgrading
     'pandas': '>=1.0.0,<2.0',
     'scikit-learn': '>=0.22.0,<0.25',
-    'scipy': '==1.5.4',
-    'gluoncv': '>=0.10.1,<0.12.0',
+    'scipy': '>=1.5.4,<1.7',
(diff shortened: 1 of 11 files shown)


szha:
Job PR-1075-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1075/1/index.html

jwmueller (approved):
Seems like better setting of `level_time_modifier` may be achieved by also taking into account the dataset-size (ie. time required to fully fit 1 model)? 

For example, `level_time_modifier = 1` would probably better than 0.333 if given small time-limit relative to dataset size?

Innixma:
> Seems like better setting of `level_time_modifier` may be achieved by also taking into account the dataset-size (ie. time required to fully fit 1 model)?
> 
> For example, `level_time_modifier = 1` would probably be better than 0.333 if given small time-limit relative to dataset size?

This is definitely true, however we currently don't have a concept within AutoGluon of an 'expected training time' for a model given the properties of a dataset. We would need this concept (either learned or heuristic based) to adjust the value sensibly.

PR #1080, opened 2021-04-24
Tabular: Refactored evaluate/evaluate_predictions

Innixma:
*Issue #, if available:*

*Description of changes:*

- Refactored evaluate and evaluate_predictions to cause less confusion with users and act more consistently across eval metrics and problem types.
- evaluate is now purely a wrapper around evaluate_predictions, making the logic much less disjoint.
- evaluate_predictions now can always handle pred_proba input, and can evaluate pred and pred_proba metrics at the same time (previously this didn't work).
- Made the return types always dictionaries to reduce confusion.

### DIFF
--- a/core/src/autogluon/core/utils/utils.py
+++ b/core/src/autogluon/core/utils/utils.py
@@ -412,7 +412,11 @@ def compute_weighted_metric(y, y_pred, metric, weights, weight_evaluation=None, 
     try:
         weighted_metric = metric(y, y_pred, sample_weight=weights, **kwargs)
     except (ValueError, TypeError, KeyError):
-        logger.log(30, f"WARNING: eval_metric='{metric.name}' does not support sample weights so they will be ignored in reported metric.")
+        if hasattr(metric, 'name'):
+            metric_name = metric.name
+        else:
+            metric_name = metric
+        logger.log(30, f"WARNING: eval_metric='{metric_name}' does not support sample weights so they will be ignored in reported metric.")
         weighted_metric = metric(y, y_pred, **kwargs)
     return weighted_metric
 
--- a/docs/tutorials/tabular_prediction/tabular-indepth.md
+++ b/docs/tutorials/tabular_prediction/tabular-indepth.md
@@ -81,7 +81,7 @@ We again demonstrate how to use the trained models to predict on the test data.
 ```{.python .input}
 y_pred = predictor.predict(test_data_nolabel)
 print("Predictions:  ", list(y_pred)[:5])
-perf = predictor.evaluate_predictions(y_true=y_test, y_pred=y_pred, auxiliary_metrics=False)
+perf = predictor.evaluate(test_data, auxiliary_metrics=False)
 ```
 
 Use the following to view a summary of what happened during fit. Now this command will show details of the hyperparameter-tuning process for each type of model:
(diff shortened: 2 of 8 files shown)


szha:
Job PR-1080-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1080/2/index.html

docs/tutorials/tabular_prediction/tabular-indepth.md
```diff
@@ -206,20 +206,16 @@ predictor_information = predictor.info()
 The `predictor` also remembers what metric predictions should be evaluated with, which can be done with ground truth labels as follows:
 
 ```{.python .input}
-y_pred = predictor.predict(test_data_nolabel)
-perf = predictor.evaluate_predictions(y_true=y_test, y_pred=y_pred, auxiliary_metrics=True)
+y_pred_proba = predictor.predict_proba(test_data_nolabel)
+perf = predictor.evaluate_predictions(y_true=y_test, y_pred=y_pred_proba)
 ```
```
jwmueller, line 212: May be worthwhile to add clarifying statement:

Note that `evaluate_predictions()` requires you to appropriately set `y_pred` as predicted class labels (`predict()` output) or probabilities (`predict_proba()` output) depending on which type of predictions are needed by the `eval_metric`. In contrast, `evaluate()` will automatically handle this for you by internally producing the correct type of predictions for the `eval_metric`.

tabular/src/autogluon/tabular/learner/abstract_learner.py
```diff
@@ -474,84 +490,100 @@ def evaluate_predictions(self, y_true, y_pred, silent=False, auxiliary_metrics=F
...
-                        performance_dict[metric_name] = metric_function(y_true, y_pred, **metric_kwargs)
-                    else:
-                        performance_dict[metric_name] = metric_function(y_true, y_pred)
-                except ValueError:
-                    pass
+                flipped = False
+            if not silent:
+                logger.log(20, f"Evaluation: {self.eval_metric.name} on test data: {score_eval}")
+                if flipped:
+                    logger.log(20, f"\tNote: Scores are always higher_is_better. This metric score can be multiplied by -1 to get the metric value.")
```
jwmueller, line 586: since difference between score/value may be unclear, might clarify:
 
"metric score" -> "reported metric score"
"metric value" -> "actual {self.eval_metric.name} metric value"

jwmueller (approved):
looks good!

core/src/autogluon/core/utils/utils.py
```diff
@@ -412,7 +412,11 @@ def compute_weighted_metric(y, y_pred, metric, weights, weight_evaluation=None,
     try:
         weighted_metric = metric(y, y_pred, sample_weight=weights, **kwargs)
     except (ValueError, TypeError, KeyError):
-        logger.log(30, f"WARNING: eval_metric='{metric.name}' does not support sample weights so they will be ignored in reported metric.")
+        if hasattr(metric, 'name'):
+            metric_name = metric.name
+        else:
+            metric_name = metric
```
gradientsky, line 418: str(metric)?

tabular/src/autogluon/tabular/learner/abstract_learner.py
gradientsky: @jwmueller this sounds more confusing to me compared to the one in the log message.

gradientsky (approved):
Looks good to me, minor comments.

core/src/autogluon/core/utils/utils.py
Innixma: I don't think it matters in the context I am using it.

docs/tutorials/tabular_prediction/tabular-indepth.md
Innixma: This is no longer necessary because of the refactor. Now there is no downside of always passing prediction probabilities as they will be smartly converted to predictions internally.

tabular/src/autogluon/tabular/learner/abstract_learner.py
Innixma: Will keep current for now, feel free to send a doc cleanup PR if you think it can be worded more clearly. I think we should get used to showing the negative values to users, as otherwise it just makes things more confusing in the long run.

jwmueller: I unfortunately don't have a much better solution to this confusion. Perhaps we should also print this same message near the start of `fit()` if a flipped `eval_metric` is being used (right after where it prints "AutoGluon will gauge predictive performance using evaluation metric: ..."), because otherwise users viewing the training-progress get confused and think there is an error during training.

An alternative could be to append `negative ` (or `neg `) prefix to `eval_metric.name` for each flipped metric, but this is may get complex to deal with...

docs/tutorials/tabular_prediction/tabular-indepth.md
jwmueller: I meant that they cannot pass predicted classes for metric like AUC, but I guess the error message that pops up should be clear enough.

PR #1092, opened 2021-05-01
TextPredictor fails if eval_metric = 'average_precision'

rxjx:
Issue # 1085

Check predictions shape and use second column if more than one for binary problems.

### DIFF
--- a/text/src/autogluon/text/text_prediction/mx/models.py
+++ b/text/src/autogluon/text/text_prediction/mx/models.py
@@ -187,7 +187,7 @@ def base_optimization_config():
     cfg.nbest = 1                   # Keep the top K performed models
     cfg.per_device_batch_size = 16  # Per-device batch-size
     cfg.auto_per_device_batch_size = True  # Whether to automatically determine the runnable
-                                           # per-device batch_size.
+    # per-device batch_size.
     cfg.val_batch_size_mult = 2  # By default, we 2X the batch size for validation
     cfg.lr = 1E-4
     cfg.final_lr = 0.0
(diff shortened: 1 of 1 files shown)


text/src/autogluon/text/text_prediction/mx/models.py
```diff
@@ -322,7 +322,7 @@ def _classification_regression_predict(net, dataloader, problem_type, label_scal
 
 
 def calculate_metric(scorer, ground_truth, predictions, problem_type):
-    if problem_type == BINARY and scorer.name == 'roc_auc':
+    if problem_type == BINARY and predictions.ndim > 1:
```
Innixma, line 325: Crashes on f1 metric. This may be a separate issue from the one being solved in this PR however, and this change merely surfaced it. @sxjscience we may in a future PR want to unify Tabular and Text's usage of metrics and calculating scores, I think they somewhat differ.

rxjx: Let me take a look. Yeah, I was thinking about Tabular vs Text but I figured we could do that after this is merged and used for a bit.

rxjx: matthews_corrcoef will also fail. I think I'll just be conservative and enable only for 'roc_auc' and 'average_precision' for now.

sxjscience: I think we may first turn it on for roc_auc and average_precision. The reason is that some other metrics need the probability for pos + neg classes and do not support array that has (B,) shape.

rxjx: Great minds think alike :-). I did just that a few minutes before your comment.

szha:
Job PR-1092-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1092/2/index.html

sxjscience (approved):
LGTM

PR #1102, opened 2021-05-07
Clean RF quantile ensemble logic

Innixma:
*Issue #, if available:*

*Description of changes:*

- Make special RF quantile ensemble logic live in model code

### DIFF
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -916,7 +916,7 @@ class AbstractModel:
         return {}
 
     @classmethod
-    def _get_default_ag_args_ensemble(cls) -> dict:
+    def _get_default_ag_args_ensemble(cls, **kwargs) -> dict:
         """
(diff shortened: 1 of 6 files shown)


szha:
Job PR-1102-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1102/1/index.html

jwmueller (approved):
In the future could consider some check function that verifies all the ag_args*, hyperparameters, and other dict inputs are correctly configured. I imagine wrong specification of these can otherwise produce opaque errors, but at least hopefully the users messing with these are more advanced.

Innixma:
> In the future could consider some check function that verifies all the ag_args*, hyperparameters, and other dict inputs are correctly configured. I imagine wrong specification of these can otherwise produce opaque errors, but at least hopefully the users messing with these are more advanced.

Yes, we mostly need to be mindful that we aren't limiting flexibility. A method to track unused ag_args to a particular model might be good as a logged warning.

Issue #1121, opened 2021-05-19
No version property


wlongxiang:
```
>>> import autogluon
>>> autogluon.__version__
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'autogluon' has no attribute '__version__'
```
For most python packages, the above should print out the version number. This is useful for quick version checking and automation in CI pipelines.

Innixma:
AutoGluon is a namespace package, which means you cannot perform any imports at the `autogluon` level as it is not technically a package.

Instead, you will need to get the version of a particular submodule of autogluon, such as:

```
from autogluon.core import __version__
from autogluon.tabular import __version__
from autogluon.text import __version__
from autogluon.vision import __version__
# etc...
```

Innixma:
Ah, you are correct, it is under:

```
from autogluon.tabular.version import __version__
print(__version__)
```

wlongxiang:
Thanks both for your quick response @sxjscience @Innixma !
I would still like to add that it is more or less conventional for a package, in this case, tabular, text, etc. under autogluon namespace to have `__version__`

This should ideally give back the version number:
```
>>> from autogluon import tabular
>>> tabular.__version__
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'autogluon.tabular' has no attribute '__version__'
```
I can make the pull request to add this property to the package level, if you guys think it could be a small improvement too.

Innixma:
Yes, I had the same thought @wlongxiang, I have created a PR which adds the requested functionality: #1122

Innixma:
#1122 is merged and fixes this issue.

PR #1122, opened 2021-05-19
Improved access method for `__version__`

Innixma:
*Issue #, if available:*
#1121 

*Description of changes:*

- Now users can access version via `autogluon.{submodule}.__version__`
- Open question: Do we want to refactor `version.py` to be `_version.py` so users don't get confused? Should it be a python file at all? LightGBM uses `VERSION.txt` and loads from file in `__init__.py`.

### DIFF
--- a/core/src/autogluon/core/__init__.py
+++ b/core/src/autogluon/core/__init__.py
@@ -18,3 +18,8 @@ __logging.getLogger("distributed.logging.distributed").setLevel(__logging.ERROR)
 __logging.getLogger("distributed.worker").setLevel(__logging.ERROR)
 
 __logging.basicConfig(format='%(message)s')  # just print message in logs
+
+try:
+    from .version import __version__
+except ImportError:
+    pass
--- a/core/src/autogluon/core/__init__.py
+++ b/core/src/autogluon/core/__init__.py
@@ -19,7 +19,4 @@ __logging.getLogger("distributed.worker").setLevel(__logging.ERROR)
 
 __logging.basicConfig(format='%(message)s')  # just print message in logs
 
-try:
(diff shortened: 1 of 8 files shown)


szha:
Job PR-1122-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1122/1/index.html

core/src/autogluon/core/__init__.py
```diff
@@ -16,3 +16,8 @@
 __logging.getLogger("distributed.utils_perf").setLevel(__logging.ERROR)
 __logging.getLogger("distributed.logging.distributed").setLevel(__logging.ERROR)
 __logging.getLogger("distributed.worker").setLevel(__logging.ERROR)
+
+try:
+    from .version import __version__
+except ImportError:
+    pass
```
sxjscience, line 23: I'm not sure if we need to pass here.

Innixma: Are you suggesting we raise an exception if the version file does not exist? If not, could you mention the code you are suggesting it be changed to?

sxjscience: I mean, we can directly change to 

```
```

szha:
Job PR-1122-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1122/2/index.html

core/src/autogluon/core/__init__.py
Innixma: updated

sxjscience (approved):
LGTM

Issue #1158, opened 2021-06-04
Does FastAI model usually not train well on regression data?


SangbumChoi:
According to the result of benchmark in kaggle dataset (mercedes, allstate, houses, etc...)

I think that fastainn model does not well-training (which means early stopping in 0 epoch with 20 patience) in regression kaggle datasets (except for allstate).

Since mxnet models usually train well on various regression and binary datasets and also fastainn models train well on binary and multi-class dataset. I assumed that loss normalization or scaling on regression dataset was the main reason of this phenomena (or learning-rate problem). Because as far as I searched fastainn library there was no such functions (ex: loss_scaling_factor) but mxnet does. However, all of the custom trial to enhance fastainn model in regression dataset failed.

Does my experiment (with default hyperparameter such as time, auto-stack, etc...) went wrong or does your group have faced this same problem?

SangbumChoi:
0
TensorBase(5.7082e+08)
1
TensorBase(1.3430e+10)
2
TensorBase(2.8899e+11)
3
TensorBase(3.9408e+12)
4
TensorBase(3.4854e+13)
5
TensorBase(2.3243e+14)
6
TensorBase(6.1357e+14)
7
TensorBase(8.1096e+14)
8
TensorBase(4.5594e+13)
9
TensorBase(5.2544e+13)
10
TensorBase(1.4245e+13)
11
TensorBase(1.0146e+14)
12
TensorBase(2.2767e+14)
13
TensorBase(8.1047e+12)
14
TensorBase(1.3270e+14)
15
TensorBase(3.3033e+14)
16
TensorBase(1.0559e+14)
17
TensorBase(9.1204e+13)
18
TensorBase(1.6465e+14)
19
TensorBase(3.5654e+14)
20
TensorBase(3.4201e+14)

This is the example of Fastainn loss with respective epoch in mercedes dataset

Innixma:
Thanks for noticing this! This appears to be a serious bug that we weren't aware of. At some point it must have been introduced as I remember a long time ago FastAI working correctly in regression problems.

Code showing the bug:

```
from autogluon.tabular import TabularDataset, TabularPredictor
from autogluon.core.utils import generate_train_test_split

LABEL = 'y'
SAMPLE = 1000


if __name__ == '__main__':
    path_prefix = 'datasets/mercedes-benz-greener-manufacturing/'
    path_train = path_prefix + 'train.csv'
    path_test = path_prefix + 'test.csv'

[...]

PR #1161, opened 2021-06-05
[Bugfix][FASTAI] - continuous variables with single values in fold splits producing outliers when unseen values are passed.

gradientsky:
Single-value inputs are usually dropped in feature generators on a dataset level, but it is still can happen within single fold split. When an unseen data gets unseen value in such continuous variables, this creates unexpected results.

*Issue #, if available:*
https://github.com/awslabs/autogluon/issues/1158

*Description of changes:*
+ Dropping single-value features on fold level
+ Upgraded fastai to 2.3.1 and removed code necessary to patch layers defect https://github.com/fastai/fastai/pull/3314
+ Enabled results scaling for regression tasks to stabilize learning rates requirements for wide ranges of datasets 

### DIFF
--- a/tabular/setup.py
+++ b/tabular/setup.py
@@ -45,7 +45,7 @@ extras_require = {
     ],
     'fastai': [
         'torch>=1.0,<2.0',
-        'fastai>=2.0,<3.0',
+        'fastai>=2.3.1,<3.0',
     ],
     'skex': [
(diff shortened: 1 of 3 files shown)


gradientsky:
Closing - the changes are added to https://github.com/awslabs/autogluon/pull/1162

PR #1166, opened 2021-06-09
Tabular: Avoid showing plot by default in fit_summary()

Innixma:
*Issue #, if available:*
#1062 

*Description of changes:*

- Disables showing plot by default to avoid halting scripts, particularly if the user is using command line.

### DIFF
--- a/tabular/src/autogluon/tabular/predictor/predictor.py
+++ b/tabular/src/autogluon/tabular/predictor/predictor.py
@@ -1097,7 +1097,7 @@ class TabularPredictor:
             Set <= 0 for no output printing, 1 to print just high-level summary,
             2 to print summary and create plots, >= 3 to print all information produced during `fit()`.
         show_plot : bool, default = False
-            If True, shows the plot in browser when verbosity > 1.
+            If True, shows the model summary plot in browser when verbosity > 1.
 
         Returns
         -------
--- a/tabular/src/autogluon/tabular/predictor/predictor.py
+++ b/tabular/src/autogluon/tabular/predictor/predictor.py
@@ -1085,7 +1085,7 @@ class TabularPredictor:
         return self._learner.leaderboard(X=data, extra_info=extra_info, extra_metrics=extra_metrics,
                                          only_pareto_frontier=only_pareto_frontier, silent=silent)
 
-    def fit_summary(self, verbosity=3):
+    def fit_summary(self, verbosity=3, show_plot=False):
         """
         Output summary of information about models produced during `fit()`.
         May create various generated summary plots and store them in folder: `predictor.path`.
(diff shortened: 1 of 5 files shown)


gradientsky (approved):

szha:
Job PR-1166-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1166/2/index.html

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -1085,17 +1085,19 @@ def leaderboard(self, data=None, extra_info=False, extra_metrics=None, only_pare
...
 
         Parameters
         ----------
         verbosity : int, default = 3
-            Controls how detailed of a summary to ouput.
+            Controls how detailed of a summary to output.
             Set <= 0 for no output printing, 1 to print just high-level summary,
             2 to print summary and create plots, >= 3 to print all information produced during `fit()`.
+        show_plot : bool, default = False
+            If True, shows the plot in browser when verbosity > 1.
```
jwmueller, line 1100: could clarify: "plot" -> "model summary plot"

jwmueller (approved):
Looks good, but show_plot should be set = True in the tutorials right?
Otherwise most users would probably not even know these plots exist.

Innixma:
> Looks good, but show_plot should be set = True in the tutorials right?
> Otherwise most users would probably not even know these plots exist.

Currently they don't show in tutorials anyways because bokeh isn't installed.

If you'd like you could send a PR which enables them in tutorials and visualizes them in our documentation.

jwmueller:
> > Looks good, but show_plot should be set = True in the tutorials right?
> > Otherwise most users would probably not even know these plots exist.
> 
> Currently they don't show in tutorials anyways because bokeh isn't installed.
> 
> If you'd like you could send a PR which enables them in tutorials and visualizes them in our documentation.

The bokeh plot would not show on website regardless because it is in new browser, but would show for user who runs the tutorial locally in interactive python/notebook with bokeh installed (as should the other plots).  One option is the tutorial has: show_plot = True with a comment to set this =False to avoid halting in non-interactive Python scripts.  Or tutorials can just have plots disabled (I'm just noting then we can't expect most users will ever see plots)

szha:
Job PR-1166-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1166/3/index.html

tabular/src/autogluon/tabular/predictor/predictor.py
Innixma: added

Innixma:
Added `show_plot=True` in tutorials, although it will look a bit odd since no plots will be shown on the website.

szha:
Job PR-1166-4 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1166/4/index.html

Issue #1147, opened 2021-05-28
Why "valid_acc" always better than "train_acc"?


Fanxingye:
In the summary of different task logs,  "valid_acc" always better than "train_acc", it seems unreasonable.
These are logs of different tasks:
![image](https://user-images.githubusercontent.com/50837470/119929306-d5146e00-bfaf-11eb-86d5-9f2bf13881dc.png)
![image](https://user-images.githubusercontent.com/50837470/119929360-f1180f80-bfaf-11eb-89ae-acd166a8f356.png)

talhaanwarch:
I did not observe this behavior in Vision task
![image](https://user-images.githubusercontent.com/37379131/120020007-98259700-c002-11eb-8268-69c455905e62.png)

Fanxingye:
> I did not observe this behavior in Vision task
> ![image](https://user-images.githubusercontent.com/37379131/120020007-98259700-c002-11eb-8268-69c455905e62.png)

But it always appears in all my Vision tasks, which dataset did you use?

talhaanwarch:
Are you defining tuning set explicitly in predictor function or AG split it self

jianzhnie:
Yes, I did not defin my own val_data, just use the tuning data in predictor function split it self

jianzhnie:
Yes, I did not defin my own val_data, just use the tuning data in predictor function split it self

Fanxingye:
> Are you defining tuning set explicitly in predictor function or AG split it self

Yes, I did not define my own val_data, just use the tuning data in predictor function split it self, Is there any different between these two mode？

kevin666aa:
![image](https://user-images.githubusercontent.com/32823396/120131774-a559a600-c1fb-11eb-850a-b06ff89f5400.png)
I used a dataset from kaggle, [the yoga posts]( https://www.kaggle.com/shrutisaxena/yoga-pose-image-classification-dataset?select=dataset). And I have observed the same problem. 

As you can see, the training accuracy is more than 20% less than validation accuracy, and when I used a self-split test set, I got accuracy similar to the training.

Would you please look into the problem? Maybe test with this kaggle dataset?
PS: My autogluon version is 0.2.0.

Innixma:
Thanks for reporting! This is definitely strange and might have to do with some kind of index misalignment in how training data is scored, hard to tell without deep diving. We will plan to look into this issue and resolve it for v0.3.

Innixma:
[...]

PR #1168, opened 2021-06-11
Vision: Fix major bug when tuning_data is not specified

Innixma:
*Issue #, if available:*
#1147 

*Description of changes:*

- Fixed bug causing train_data and tuning_data to both use all the training data if tuning_data is not specified in ImagePredictor.
- Improved usage of data splitting. Now it is deterministic and splits the exact specified proportion of data (rather than applying that proportion as a probability independently to each sample of being included in train vs val).
- Data splitting now performs stratification if applicable.

NOTE:
- A **major** bug still exists in `ImageDataset.to_mxnet()` that causes severe issues downstream by ignoring upstream data splits. This PR does not fix this bug, but rather avoids it for the specific instance of `ImagePredictor.fit()`.

### DIFF
--- a/vision/src/autogluon/vision/predictor/predictor.py
+++ b/vision/src/autogluon/vision/predictor/predictor.py
@@ -311,12 +311,14 @@ class ImagePredictor(object):
             tuning_data = tuning_data.reset_index(drop=True)
 
         train_data = self._validate_data(train_data)
-        train_data = self.Dataset(train_data, classes=train_data.classes)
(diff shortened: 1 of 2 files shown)


zhreshold (approved):

szha:
Job PR-1168-4 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1168/4/index.html

PR #1182, opened 2021-06-13
Tabular: Fixed datetime crash, fixed LightGBM on text

Innixma:
*Issue #, if available:*

*Description of changes:*

- Fixed cases where datetime that is out-of-bounds will cause incorrect behavior:
- If found in train, dtype was incorrectly set to categorical. This is fixed.
- If found in test, crashes during feature pre-processing. This is fixed.

- Fixed a recently introduced bug related to text feature names in LightGBM causing LightGBM to crash during training.


szha:
Job PR-1182-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1182/1/index.html

tabular/tests/unittests/features/conftest.py
```diff
@@ -122,20 +122,20 @@ def generate_datetime_as_object_feature() -> Series:
...
+                '',  # nan
+                np.nan,  # nan
                 '4/20/2018 15:37',
                 '4/20/2018 15:37',
                 '1/01/1800 00:00',
                 '12/31/2200 23:59',
-                '8/15/2020 7:12',
-                '12/18/2020 2:12',
+                '8/15/2700 7:12',  # nan
+                '12/18/1000 2:12',  # nan
```
gradientsky, line 132: Not sure this should be done. How far from now this would be considered an error?

Innixma: Could you clarify? This is a limitation of the pandas datetime data type. It can't represent these values in int64 since it is counting them as nanoseconds, so when you call `pd.to_datetime()` it will either crash or set them as NaN if `errors='coerece'`. We have to set to NaN at inference time or else it will crash (what happened before). The only other option is to use some different method to convert the dates to numeric. I'd prefer a version that doesn't use nanosecond but instead millisecond or second, but I haven't been able to find one.

In general, I believe the edges are year ~1740 and ~2300

gradientsky (approved):

PR #1193, opened 2021-06-20
Improved loop processing speed

daikikatsuragawa:
I improved the speed by replacing the following process with another method.

```py
df.iterrows()
```

- Unnecessary variables(`i`, `idx`) are being deleted.
- Using a dictionary seems to have little impact on code comprehension.

ps:

There is a similar code below, but this has been excluded from this fix.

https://github.com/awslabs/autogluon/blob/c377908da22ff8794b68247f623cc678953eab48/core/src/autogluon/core/utils/utils.py#L147

The reason is as follows.
- `index` is used in this part and cannot be applied as it is
- This is because the cognitive load may increase even if `zip` is used.

For reference, there is a library called [df4loop](https://github.com/daikikatsuragawa/df4loop)
 that supports such problems, so please consider it if you like. (I think "don't use" is also a good decision, so ignore it if it's not particularly attractive.)

### DIFF
--- a/examples/text_prediction/prepare_glue.py
+++ b/examples/text_prediction/prepare_glue.py
@@ -437,7 +437,7 @@ def read_record(dir_path):
         df = read_jsonl_superglue(jsonl_path)
         df_dict[fold] = df
         out = []
-        for i, row in df.iterrows():
+        for row in df.to_dict(orient="records"):
             source = row['source']
             passage = row['passage']
             text = passage['text']
--- a/examples/text_prediction/prepare_glue.py
+++ b/examples/text_prediction/prepare_glue.py
@@ -351,7 +351,7 @@ def read_wic(dir_path):
         jsonl_path = os.path.join(dir_path, '{}.jsonl'.format(fold))
         df = read_jsonl_superglue(jsonl_path)
         out = []
-        for idx, row in df.iterrows():
+        for row in df.to_dict(orient="records"):
             sentence1 = row['sentence1']
             sentence2 = row['sentence2']
             start1 = row['start1']


szha:
Job PR-1193-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1193/1/index.html

Innixma:
Thanks for the contribution! Since this is a relatively simple change and the idea is that it improves the code speed, could you comment a small script that shows that this change results in a speedup?

sxjscience:
Thanks for the contribution! I used iterrow a lot in my code and the major advantage is that it won’t generate a full copy of the data frame. For large datasets, .to_dict() may double the memory usage while .iterrow() is accessing the rows via an iterator. Thus, it will be great if you can provide the speedup v.s. Memory usage benchmark so we can know which standard to follow later.

szha:
Job PR-1193-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1193/3/index.html

daikikatsuragawa:
Sorry for the late reply.

> .iterrow() is accessing the rows via an iterator.

Thanks for your comment. I didn't have a "memory" perspective.

However, I think that this change is unnecessary from the viewpoint of memory in the first place, so you can have it closed. I'm sorry to send you a pull request with lack of knowledge.

szha:
Job PR-1193-4 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1193/4/index.html

szha:
Job PR-1193-5 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1193/5/index.html

szha:
Job PR-1193-8 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1193/8/index.html

Innixma:
Apologies for not closing based on your initial request. Thanks @daikikatsuragawa!

PR #1197, opened 2021-06-21
[Vision] fix duplicate class mapping

zhreshold:
*Issue #, if available:*

https://github.com/awslabs/autogluon/pull/1168

*Description of changes:*

Fix a bug when `tuning_data` is not specified, the validation acc will always be 0. This doesn't affect the trained model, nor the final test accuracy, but the searching metric will be messed since the searcher will always get 0 as feedback.

This is caused by applying validation to the (already) validated `tuning_data` split from `train_data`


Innixma (approved):
Approved assuming CI passes

szha:
Job PR-1197-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1197/1/index.html

PR #1199, opened 2021-06-22
Vision: Fix searcher presets and GPU memory check

Innixma:
*Issue #, if available:*

*Description of changes:*

- Presets specified `search_strategy` which is not a valid key, therefore searcher was always `random`.
- This PR removes the incorrect keys, replacing with `searcher`. `searcher` cannot be `bayesopt` due to a bug causing the run to crash.
- Fixed missing `Int` import causing GPU memory check to fail.
- Standardized usage of `ImageDataset` to align with `TabularDataset`.

### DIFF
--- a/docs/tutorials/image_prediction/beginner.md
+++ b/docs/tutorials/image_prediction/beginner.md
@@ -5,7 +5,7 @@ In this quick start, we'll use the task of image classification to illustrate ho
 
 ```{.python .input}
 import autogluon.core as ag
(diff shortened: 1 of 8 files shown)


szha:
Job PR-1199-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1199/1/index.html

zhreshold (approved):
lgtm

PR #1205, opened 2021-06-23
Create a method that repeatedly fits base learners on different feature subsets for tabular data

truebluejason:
*Issue #, if available:*

*Description of changes:* Created `fit_with_prune` method for AbstractModel that repeatedly alternates between model fitting and feature selection using permutation feature importance score until the validation score performance drops. This functionality can be tentatively accessed by providing a boolean flag `fit_with_prune` to TabularPredictor.

### DIFF
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -498,6 +498,67 @@ class AbstractModel:
             logger.warning(f'\tWarning: Model has no time left to train, skipping model... (Time Left = {round(kwargs["time_limit"], 1)}s)')
             raise TimeLimitExceeded
 
+    def fit_with_prune(self, max_num_fit=3, prune_threshold=0., num_shuffle_sets=50, **kwargs):
+        """
+        Functionally identical to `fit` method, but repeats feature importance based pruning until
+        validation set performance degrades or `max_num_fit` iterations have passed.
+
+        Parameters
+        ----------
+        max_num_fit : int, default = 3
+            Maximum number of time feature selection and fitting are performed
+        prune_threshold : float, default = 0.
+            Feature importance threshold that features must meet in order to not be dropped
+        num_shuffle_sets : int, default = 3
+            Number of shuffles to evaluate when computing permutation feature importance
+        """
+        if kwargs['X_val'] is None or kwargs['y_val'] is None:
+            logger.error(f'fit_with_prune must be called with validation data.')
+            raise Exception
+
+        untrained_model_path = os.path.join(self.path, 'untrained_')
+        best_model_path = os.path.join(self.path, 'best_')
+        best_score = -float("inf")
+        self.save(untrained_model_path)  # save untrained model to disk to avoid complications from refitting the same model object
+
+        try:
+            for i in range(max_num_fit):
+                self.fit(**kwargs)
+                score = self.score(X=kwargs['X_val'], y=kwargs['y_val'])
+                if score > best_score:
+                    # keep: replace best model with candidate model on disk
+                    logger.log(20, f"\tFit {i+1}: Current score {score} is better than best score {best_score}. Keeping model.")
+                    best_score = score
+                    self.save(best_model_path)
+                else:
+                    logger.log(20, f"\tCurrent score {score} is not better than best score {best_score}. Ending prune loop after {i+1} iterations.")
+                    break
+
+                # compute all feature importance and remove ones that don't meet the threshold
+                importance_df = self.compute_feature_importance(X=kwargs['X_val'], y=kwargs['y_val'], num_shuffle_sets=num_shuffle_sets)
+                cols_to_drop, cols_to_drop_importance = [], []
+                for column_name, row in importance_df.iterrows():
+                    if row['importance'] <= prune_threshold:
+                        cols_to_drop.append(column_name)
+                        cols_to_drop_importance.append(row['importance'])
+                logger.log(20, f"\tWill try fit after pruning these columns that failed to meet importance threshold {prune_threshold}:")
+                logger.log(20, f"\t{list(zip(cols_to_drop, cols_to_drop_importance))}")
+                # reset model but give it a new set of features
+                new_features = [feat for feat in self.features if feat not in cols_to_drop]
+
+                # reset model to pre-fit state
+                self = self.load(path=untrained_model_path)
+                self.features = new_features
+        finally:
+            # cleanup
+            if os.path.exists(untrained_model_path + self.model_file_name):
+                os.remove(untrained_model_path + self.model_file_name)
+            if os.path.exists(best_model_path + self.model_file_name):
+                # restore best model
+                self = self.load(path=best_model_path)
+                os.remove(best_model_path + self.model_file_name)
+            logger.log(20, f"\tFinal Features: {self.features}")
+
     def _fit(self,
              X,
              y,
--- a/tabular/src/autogluon/tabular/predictor/predictor.py
+++ b/tabular/src/autogluon/tabular/predictor/predictor.py
@@ -644,6 +644,7 @@ class TabularPredictor:
         ag_args_fit = kwargs['ag_args_fit']
         ag_args_ensemble = kwargs['ag_args_ensemble']
         excluded_model_types = kwargs['excluded_model_types']
+        fit_with_prune = kwargs['fit_with_prune']
 
         if ag_args is None:
             ag_args = {}
@@ -702,6 +703,7 @@ class TabularPredictor:
             'ag_args_ensemble': ag_args_ensemble,
             'ag_args_fit': ag_args_fit,
             'excluded_model_types': excluded_model_types,
+            'fit_with_prune': fit_with_prune
         }
         self._learner.fit(X=train_data, X_val=tuning_data, X_unlabeled=unlabeled_data,
                           holdout_frac=holdout_frac, num_bag_folds=num_bag_folds, num_bag_sets=num_bag_sets, num_stack_levels=num_stack_levels,
@@ -2205,7 +2207,8 @@ class TabularPredictor:
             # other
             feature_generator='auto',
             unlabeled_data=None,
-            _feature_generator_kwargs=None
+            _feature_generator_kwargs=None,
+            fit_with_prune=False
         )
 
         kwargs = self._validate_fit_extra_kwargs(kwargs, extra_valid_keys=list(fit_kwargs_default.keys()))
--- a/tabular/src/autogluon/tabular/trainer/abstract_trainer.py
+++ b/tabular/src/autogluon/tabular/trainer/abstract_trainer.py
@@ -896,6 +896,9 @@ class AbstractTrainer:
         """
         if isinstance(model, BaggedEnsembleModel):
             model.fit(X=X, y=y, **model_fit_kwargs)
+        elif 'fit_with_prune' in model_fit_kwargs and model_fit_kwargs['fit_with_prune']:
+            # only available for non-ensemble models
+            model.fit_with_prune(X=X, y=y, X_val=X_val, y_val=y_val, **model_fit_kwargs)
         else:
             model.fit(X=X, y=y, X_val=X_val, y_val=y_val, **model_fit_kwargs)
         return model
@@ -1063,8 +1066,10 @@ class AbstractTrainer:
             n_repeats = self.n_repeats
         model_fit_kwargs = dict(
             time_limit=time_limit,
+            fit_with_prune=kwargs['fit_with_prune'] if 'fit_with_prune' in kwargs else False,
             verbosity=self.verbosity,
         )
+
         if self.sample_weight is not None:
             X, w_train = extract_column(X, self.sample_weight)
             if w_train is not None:  # may be None for ensemble
@@ -1074,7 +1079,7 @@ class AbstractTrainer:
                 X_val, w_val = extract_column(X_val, self.sample_weight)
                 if self.weight_evaluation and w_val is not None:  # ignore validation sample weights unless weight_evaluation specified
                     model_fit_kwargs['sample_weight_val'] = w_val.values/w_val.mean()
-            ens_sample_weight =  kwargs.get('ens_sample_weight', None)
+            ens_sample_weight = kwargs.get('ens_sample_weight', None)
             if ens_sample_weight is not None:
                 model_fit_kwargs['sample_weight'] = ens_sample_weight  # sample weights to use for weighted ensemble only
 


core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -498,6 +498,67 @@ def fit(self, **kwargs):
...
+        untrained_model_path = os.path.join(self.path, 'untrained_')
+        best_model_path = os.path.join(self.path, 'best_')
+        best_score = -float("inf")
+        self.save(untrained_model_path)  # save untrained model to disk to avoid complications from refitting the same model object
+
+        try:
+            for i in range(max_num_fit):
+                self.fit(**kwargs)
+                score = self.score(X=kwargs['X_val'], y=kwargs['y_val'])
+                if score > best_score:
```
Innixma, line 528: `>=` will be a bit better since fewer features = more desirable, so on a tie prefer fewer features.

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -498,6 +498,67 @@ def fit(self, **kwargs):
...
+            raise Exception
+
+        untrained_model_path = os.path.join(self.path, 'untrained_')
+        best_model_path = os.path.join(self.path, 'best_')
+        best_score = -float("inf")
+        self.save(untrained_model_path)  # save untrained model to disk to avoid complications from refitting the same model object
+
+        try:
+            for i in range(max_num_fit):
+                self.fit(**kwargs)
```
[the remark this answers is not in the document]
Innixma: I see you already kind of do this by saving the unfit version and then reloading, but it would be better to instead do `self_copy.fit()` instead of `self.fit()`, otherwise strange things can happen, especially if an exception is raised as the outer context `self` object is being modified in `AbstractTrainer`.

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -498,6 +498,67 @@ def fit(self, **kwargs):
             logger.warning(f'\tWarning: Model has no time left to train, skipping model... (Time Left = {round(kwargs["time_limit"], 1)}s)')
             raise TimeLimitExceeded
 
+    def fit_with_prune(self, max_num_fit=3, prune_threshold=0., num_shuffle_sets=50, **kwargs):
```
Innixma, line 501: Add early_stopping_rounds, currently the functionality acts like `early_stopping_rounds=1`, but likely better results will come from `early_stopping_rounds=2`

szha:
Job PR-1205-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1205/1/index.html

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -498,6 +498,67 @@ def fit(self, **kwargs):
...
+                    best_score = score
+                    self.save(best_model_path)
+                else:
+                    logger.log(20, f"\tCurrent score {score} is not better than best score {best_score}. Ending prune loop after {i+1} iterations.")
+                    break
+
+                # compute all feature importance and remove ones that don't meet the threshold
+                importance_df = self.compute_feature_importance(X=kwargs['X_val'], y=kwargs['y_val'], num_shuffle_sets=num_shuffle_sets)
+                cols_to_drop, cols_to_drop_importance = [], []
+                for column_name, row in importance_df.iterrows():
```
Innixma, line 540: There are faster ways to do this in pandas, such as:

```
cols_to_drop_df = importance_df[importance_df['importance'] <= prune_threshold]
cols_to_drop += list(cols_to_drop_df.index)
cols_to_drop_importance += list(cols_to_drop_df['importance'])
```

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -498,6 +498,67 @@ def fit(self, **kwargs):
...
+                logger.log(20, f"\t{list(zip(cols_to_drop, cols_to_drop_importance))}")
+                # reset model but give it a new set of features
+                new_features = [feat for feat in self.features if feat not in cols_to_drop]
+
+                # reset model to pre-fit state
+                self = self.load(path=untrained_model_path)
+                self.features = new_features
+        finally:
+            # cleanup
+            if os.path.exists(untrained_model_path + self.model_file_name):
```
Innixma, line 554: Add TODO: Consider loading models and calling `model.delete` to ensure auxiliary files are removed

Innixma: Add FIXME: Does not work on with S3 paths

This isn't necessary to fix currently, but we eventually want all functionality to work from non-local disk storage (where os.remove / os.path.exists does not work)

core/src/autogluon/core/models/abstract/abstract_model.py
Innixma: Add TODO: Consider returning list of models, similar to HPO, as it is theoretically possible that ensembling the different iterations could result in a better model.

We don't need to do this right away, as it is not necessary to get the primary benefit. It will however help greatly in debugging and testing all of the iterations against test data rather than val data to see if we are overfitting in this method.

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -498,6 +498,67 @@ def fit(self, **kwargs):
...
+        Parameters
+        ----------
+        max_num_fit : int, default = 3
+            Maximum number of time feature selection and fitting are performed
+        prune_threshold : float, default = 0.
+            Feature importance threshold that features must meet in order to not be dropped
+        num_shuffle_sets : int, default = 3
+            Number of shuffles to evaluate when computing permutation feature importance
+        """
+        if kwargs['X_val'] is None or kwargs['y_val'] is None:
```
Innixma, line 515: If parameters in kwargs are required, they shouldn't be in kwargs.

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -498,6 +498,67 @@ def fit(self, **kwargs):
...
+                    break
+
+                # compute all feature importance and remove ones that don't meet the threshold
+                importance_df = self.compute_feature_importance(X=kwargs['X_val'], y=kwargs['y_val'], num_shuffle_sets=num_shuffle_sets)
+                cols_to_drop, cols_to_drop_importance = [], []
+                for column_name, row in importance_df.iterrows():
+                    if row['importance'] <= prune_threshold:
+                        cols_to_drop.append(column_name)
+                        cols_to_drop_importance.append(row['importance'])
+                logger.log(20, f"\tWill try fit after pruning these columns that failed to meet importance threshold {prune_threshold}:")
```
Innixma, line 544: If no features are dropped, then does it make sense to keep going?

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -498,6 +498,67 @@ def fit(self, **kwargs):
...
+        Functionally identical to `fit` method, but repeats feature importance based pruning until
+        validation set performance degrades or `max_num_fit` iterations have passed.
+
+        Parameters
+        ----------
+        max_num_fit : int, default = 3
+            Maximum number of time feature selection and fitting are performed
+        prune_threshold : float, default = 0.
+            Feature importance threshold that features must meet in order to not be dropped
+        num_shuffle_sets : int, default = 3
```
Innixma, line 512: default says 3 but its actually 50

core/src/autogluon/core/models/abstract/abstract_model.py
Innixma: Additionally, if X_val was not specified, this would crash instead of logging the error due to invalid key lookup

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -498,6 +498,67 @@ def fit(self, **kwargs):
...
+            Number of shuffles to evaluate when computing permutation feature importance
+        """
+        if kwargs['X_val'] is None or kwargs['y_val'] is None:
+            logger.error(f'fit_with_prune must be called with validation data.')
+            raise Exception
```
Innixma, line 517: Just raise exception with the message: `raise AssertionError(f'fit_with_prune must be called with validation data.')`

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -498,6 +498,67 @@ def fit(self, **kwargs):
...
+
+        untrained_model_path = os.path.join(self.path, 'untrained_')
+        best_model_path = os.path.join(self.path, 'best_')
+        best_score = -float("inf")
```
Innixma, line 521: In my personal experience, better to have this be None and have special logic if it is None in the loop. Otherwise can lead to strange bugs or weird logging messages.

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -644,6 +644,7 @@ def fit(self,
         ag_args_fit = kwargs['ag_args_fit']
         ag_args_ensemble = kwargs['ag_args_ensemble']
         excluded_model_types = kwargs['excluded_model_types']
+        fit_with_prune = kwargs['fit_with_prune']
```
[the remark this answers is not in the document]
truebluejason: To be clear, do you want `fit_with_prune` to be part of `ag_args` or `hyperparameter_tune_kwargs`?

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -498,6 +498,67 @@ def fit(self, **kwargs):
...
+        self.save(untrained_model_path)  # save untrained model to disk to avoid complications from refitting the same model object
+
+        try:
+            for i in range(max_num_fit):
```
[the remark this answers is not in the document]
truebluejason: I assumed the `fit` method in the loop would throw TimeoutException. Do you think a separate treatment is necessary? Perhaps I should clarify whether the model training failed to run even once or the model has been trained at least once.

core/src/autogluon/core/models/abstract/abstract_model.py
truebluejason: Adding early stopping with rounds >= 1 would mean we would have to adjust the threshold too, right? Otherwise if feature pruning sucked once we shouldn't take away even more features from that point.

tabular/src/autogluon/tabular/predictor/predictor.py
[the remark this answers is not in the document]
Innixma: part of ag_args

core/src/autogluon/core/models/abstract/abstract_model.py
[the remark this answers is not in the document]
Innixma: if time_limit is 600s, currently it would spend 600 seconds in each iteration instead of ensuring the entire for-loop takes at most 600s.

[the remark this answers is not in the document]
Innixma: you are passing the same time_limit value to each iteration

core/src/autogluon/core/models/abstract/abstract_model.py
Innixma: That is an implementation detail, in general we want to be adjusting the threshold but that is a separate piece of the overall functionality.

szha:
Job PR-1205-5 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1205/5/index.html

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -520,6 +520,84 @@ def fit(self, **kwargs):
...
+                if best_score is None or score >= best_score:
+                    logger.log(20, f"\tFit {index+1}: Current score {score} is better than best score {best_score}.")
+                    best_score = score
+                    best_index = index
+                if index - best_index == stop_threshold:
+                    logger.log(20, f"\tEnding prune loop after {index+1} iterations. Best score: {best_score}")
+                    break
+
+                # compute all feature importance and remove ones that don't meet the threshold
+                importance_df = self_copy.compute_feature_importance(X=X_val, y=y_val, num_shuffle_sets=num_shuffle_sets)
```
Innixma, line 574: To enable for bagging: BaggedEnsembleModel has an extra parameter in `.compute_feature_importance` that is `is_oof`. If `is_oof=True`, then it will expect `X=X, y=y` instead of `X=X_val, y=y_val`.

Therefore, `is_oof=True` if X_val=None, y_val=None : This is not invalid for bagging.

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -520,6 +520,84 @@ def fit(self, **kwargs):
...
+            for index in range(max_num_fit):
+                # Fit a deepcopied model on current_features
+                curr_model_path = os.path.join(self.path, f'prune_{index}', '')
+                self_copy = self.convert_to_template()
+                self_copy.features = current_features
+                self_copy.path = curr_model_path
+                self_copy.fit(X=X, y=y, X_val=X_val, y_val=y_val, **kwargs)
+                self_copy.save()
+
+                score = self_copy.score(X=X_val, y=y_val)
```
Innixma, line 563: To enable for bagging:

If X_val=None, y_val=None -> is_oof=True, then call self_copy.score_with_oof(y=y)

Innixma: Add TODO: if self.weight_evaluation==True pass sample_weight param for scoring

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -520,6 +520,84 @@ def fit(self, **kwargs):
...
+
+        TODO
+        1. Respect time limit
+        2. Consider returning list of models
+
+        FIXME
+        1. Does not work with S3 paths (no problem locally)
+        """
+        if X_val is None or y_val is None:
+            raise AssertionError(f'fit_with_prune must be called with validation data.')
```
Innixma, line 547: Not an error if its a bagged model

tabular/src/autogluon/tabular/trainer/model_presets/presets.py
```diff
@@ -104,6 +104,7 @@
     'valid_stacker',
     'valid_base',
     'hyperparameter_tune_kwargs',
+    'fit_with_prune'
```
Innixma, line 107: nit: This is a matter of preference, but I've found always ending list elements with a `,` even if its the last element saves a lot of pain over time.

tabular/src/autogluon/tabular/trainer/abstract_trainer.py
```diff
@@ -929,7 +933,11 @@ def _train_single(self, X, y, model: AbstractModel, X_val=None, y_val=None, **mo
         Trains model but does not add the trained model to this Trainer.
         Returns trained model object.
         """
-        model.fit(X=X, y=y, X_val=X_val, y_val=y_val, **model_fit_kwargs)
+        if 'fit_with_prune' in model_fit_kwargs and model_fit_kwargs['fit_with_prune'] is not None:
+            # This method trains a multiple copy of models and returns the best performing one
+            model = model.fit_with_prune(X=X, y=y, X_val=X_val, y_val=y_val, **model_fit_kwargs['fit_with_prune'], **model_fit_kwargs)
```
Innixma, line 938: This is unintentionally passing `fit_with_prune` as a kwarg.

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -520,6 +520,84 @@ def fit(self, **kwargs):
...
+                importance_df = self_copy.compute_feature_importance(X=X_val, y=y_val, num_shuffle_sets=num_shuffle_sets)
+                cols_to_drop_df = importance_df[importance_df['importance'] <= prune_threshold]
+                cols_to_drop = list(cols_to_drop_df.index)
+                cols_to_drop_importance = list(cols_to_drop_df['importance'])
+                if 0 < len(cols_to_drop) < len(self_copy.features):
+                    logger.log(20, f"\tWill try fit after pruning these columns that failed to meet importance threshold {prune_threshold}:")
+                    logger.log(20, f"\t{list(zip(cols_to_drop, cols_to_drop_importance))}")
+                    current_features = [feat for feat in self_copy.features if feat not in cols_to_drop]
+                else:
+                    break
```
Innixma, line 583: Add log message describing this

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -520,6 +520,84 @@ def fit(self, **kwargs):
...
+        best_index = None
+        fitted_copies_info = []
+        current_features = None
+        try:
+            for index in range(max_num_fit):
+                # Fit a deepcopied model on current_features
+                curr_model_path = os.path.join(self.path, f'prune_{index}', '')
+                self_copy = self.convert_to_template()
+                self_copy.features = current_features
+                self_copy.path = curr_model_path
```
Innixma, line 559: probably better to rename the model copy rather than hardcode a new path.

Innixma: The name is automatically added as a suffix to the full path

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -520,6 +520,84 @@ def fit(self, **kwargs):
...
+        finally:
+            # We can alternatively return a list of fitted models
+            if len(fitted_copies_info) > 0:
+                best_model_path = fitted_copies_info[best_index][1]
+                best_model = self.load(best_model_path)
+                # original_model_path = self.path
+                # self.__dict__.update(best_model.__dict__)
+                # self.path = original_model_path
+            else:
+                best_model = self
```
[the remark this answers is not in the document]
truebluejason: I thought yes because things are added to `fitted_copies_info` only after `fit` + `score` completes and if AutoGluon for example fails in `score` we should still say model training succeeded.

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -539,61 +539,73 @@ def fit_with_prune(self, X, y, X_val, y_val, max_num_fit=3, prune_threshold=0.,
...
                     best_score = score
                     best_index = index
-                if index - best_index == stop_threshold:
+                if index - best_index >= stop_threshold:
                     logger.log(20, f"\tEnding prune loop after {index+1} iterations. Best score: {best_score}")
                     break
 
                 # compute all feature importance and remove ones that don't meet the threshold
-                importance_df = self_copy.compute_feature_importance(X=X_val, y=y_val, num_shuffle_sets=num_shuffle_sets)
+                importance_df = self_copy.compute_feature_importance(X=X_fi, y=y_fi, num_shuffle_sets=num_shuffle_sets)
```
Innixma, line 586: bug: is_oof must be specified here if it is True

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -539,61 +539,73 @@ def fit_with_prune(self, X, y, X_val, y_val, max_num_fit=3, prune_threshold=0.,
...
-                # original_model_path = self.path
-                # self.__dict__.update(best_model.__dict__)
-                # self.path = original_model_path
+                best_model = self.__class__.load(best_model_path)
             else:
-                best_model = self
-            # Cleanup saved models
-            for info in fitted_copies_info:
-                tmp_model = AbstractModel.load(info[1])
+                best_model = self_copy
```
Innixma, line 603: When this is returned, what is going to happen?

tabular/src/autogluon/tabular/trainer/abstract_trainer.py
```diff
@@ -933,9 +934,10 @@ def _train_single(self, X, y, model: AbstractModel, X_val=None, y_val=None, **mo
         Trains model but does not add the trained model to this Trainer.
         Returns trained model object.
         """
-        if 'fit_with_prune' in model_fit_kwargs and model_fit_kwargs['fit_with_prune'] is not None:
+        if model_fit_kwargs.get('fit_with_prune_kwargs', None) is not None:
             # This method trains a multiple copy of models and returns the best performing one
-            model = model.fit_with_prune(X=X, y=y, X_val=X_val, y_val=y_val, **model_fit_kwargs['fit_with_prune'], **model_fit_kwargs)
+            fit_with_prune_kwargs = model_fit_kwargs.pop('fit_with_prune_kwargs')
```
Innixma, line 939: careful with popping, unless you really know what you are doing you could create serious bugs.

You may want to first do a shallow copy of the dictionary followed by a pop of the copy.

Innixma: In this particular case I believe there isn't an issue though, just something to be aware of.

tabular/src/autogluon/tabular/trainer/abstract_trainer.py
```diff
@@ -375,10 +375,11 @@ def stack_new_level_core(self, X, y, models: Union[List[AbstractModel], dict], X
                     for model_name in model_args_fit if 'hyperparameter_tune_kwargs' in model_args_fit[model_name]
                 }
                 kwargs['hyperparameter_tune_kwargs'] = hyperparameter_tune_kwargs
-            if ag_args and 'fit_with_prune' in ag_args:
-                kwargs['fit_with_prune'] = ag_args['fit_with_prune']
-            else:
-                kwargs['fit_with_prune'] = None
+                fit_with_prune_kwargs = {
+                    model_name: model_args_fit[model_name]['fit_with_prune']
```
Innixma, line 379: Why does 'fit_with_prune' suddenly become 'fit_with_prune_kwargs'? It should be one or the other, not both.

core/src/autogluon/core/models/abstract/abstract_model.py
Innixma: otherwise you will overfit massively by treating training data as validation data (not using OOF scores)

szha:
Job PR-1205-6 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1205/6/index.html

core/src/autogluon/core/models/abstract/abstract_model.py
truebluejason: Good catch. I should've checked myself rather than implementing what I thought you meant

core/src/autogluon/core/models/abstract/abstract_model.py
truebluejason: I had that there in case `fit` errors out at the very end or `score` errors out. If the model can still generate predictions, a trained model object will be returned. Otherwise it will return an incomplete model.

tabular/src/autogluon/tabular/trainer/abstract_trainer.py
truebluejason: I popped it on purpose because I didn't want `fit_with_prune_kwargs` to also be in `model_fit_kwargs`.

tabular/src/autogluon/tabular/trainer/abstract_trainer.py
truebluejason: I'll change everything to `fit_with_prune_kwargs`. I made that change because having kwargs at the end made it more clear that it is a dictionary not some boolean flag.

truebluejason:
I made all the changes but made it so that if the pruned refit performance is equal to original performance we don't use the pruned model since in certain cases, we can over-prune but have same validation set performance (ex. small dataset where validation set is also small).

szha:
Job PR-1205-7 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1205/7/index.html

core/src/autogluon/core/models/abstract/abstract_model.py
Innixma: Do we want an incomplete model to be returned?

tabular/src/autogluon/tabular/trainer/abstract_trainer.py
Innixma: I'm aware, I'm mostly saying that this would have impacted the upstream dictionary of model_fit_kwargs if it wasn't used as a kwarg arg in the method you popped it in. Because it was used as the kwarg arg, there isn't an issue however.

szha:
Job PR-1205-10 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1205/10/index.html

szha:
Job PR-1205-11 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1205/11/index.html

szha:
Job PR-1205-12 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1205/12/index.html

szha:
Job PR-1205-13 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1205/13/index.html

szha:
Job PR-1205-14 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1205/14/index.html

szha:
Job PR-1205-23 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1205/23/index.html

szha:
Job PR-1205-24 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1205/24/index.html

szha:
Job PR-1205-25 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1205/25/index.html

szha:
Job PR-1205-26 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1205/26/index.html

Issue #1207, opened 2021-06-24
ImagePredictor fails if any row contains NaN as its image.


Innixma:
ImagePredictor currently crashes if a NaN is present in a row for the image (row contains no image)

Ideally, it should either use a dummy image or predict_proba the mean of NaN rows in train during test, or otherwise predict_proba the overall mean of the training data if no NaN rows were in train. NaN rows in train can be dropped.

```
               'mxnet.base.MXNetError: Traceback (most recent call last):\n'
               '  File "src/io/image_io.cc", line 223\n'
               "Imread: Check failed: file.is_open(): '' couldn't open file: "
               'No such file or directory\n',
```

sxjscience:
@sallypannn We have also seen this error. The solution was to use a blank image with average size.

Innixma:
@sxjscience would it have been better to instead predict average (as in a dummy model) for these rows? Logically the blank image would not be able to do better than a dummy model that predicts average label of NaN rows during train, and the blank image could corrupt the training process of the valid images

sxjscience:
Yeah, that's reasonable. For me, I'm actually thinking about the multi-modal scenario in which we will just use a blank image to fill in the missing modality. For the uni-modal predictor, predicting the mean should work better.

gradientsky:
Quick note on fills: DO NOT use blank images. From my experience in multiple challenges it was negatively impacting model quality. I think the most common image is the best simple fill option here.

sxjscience:
@gradientsky Thanks for the suggestion! How do you usually determine the most frequent image? I can think of two algorithms:

- 1. Measure the checksum of all images and randomly pick one from those that have the highest counts. 
- 2. Extract embeddings from these images via ResNet50 and build the cluster with K-means. After that, we can pick the image that are closest to the center of the largest cluster.

gradientsky:
Another option for ensembles: get oof predictions from non-image models, then fill with random image from the same class.

PR #1210, opened 2021-06-24
Vision: Fixed crash when image is null

Innixma:
*Issue #, if available:*
#1207 

*Description of changes:*

- Fixed crash when images are null in train or test

### DIFF
--- a/tabular/src/autogluon/tabular/models/image_prediction/image_predictor.py
+++ b/tabular/src/autogluon/tabular/models/image_prediction/image_predictor.py
@@ -4,9 +4,10 @@ import logging
 import os
 import pickle
 
+import numpy as np
 import pandas as pd
 
-from autogluon.core.constants import BINARY, MULTICLASS, REGRESSION
+from autogluon.core.constants import BINARY, MULTICLASS, REGRESSION, QUANTILE, SOFTCLASS
 from autogluon.core.features.types import R_OBJECT, S_IMAGE_PATH
 from autogluon.core.models import AbstractModel
 from autogluon.core.utils import get_cpu_count, get_gpu_count, try_import_mxnet, try_import_autogluon_vision
@@ -31,6 +32,7 @@ class ImagePredictorModel(AbstractModel):
         self._load_model = None
 
         self._internal_feature_map = None
+        self._dummy_pred_proba = None  # Dummy value to predict if image is NaN
 
     def _get_default_auxiliary_params(self) -> dict:
         default_auxiliary_params = super()._get_default_auxiliary_params()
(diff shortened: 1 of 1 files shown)


gradientsky (approved):

tabular/src/autogluon/tabular/models/image_prediction/image_predictor.py
```diff
@@ -127,9 +146,44 @@ def _fit(self,
...
+                y_pred_proba = np.zeros(len(X))
+            else:
+                y_pred_proba = np.zeros((len(X), self.num_classes))
+            X = X.reset_index(drop=True)
+            null_indices = X['image'] == ''
+            X = X[~null_indices]
+            null_indices_rows = list(null_indices[null_indices].index)
+            non_null_indices_rows = list(null_indices[~null_indices].index)
+            y_pred_proba[null_indices_rows] = self._dummy_pred_proba
+            y_pred_proba[non_null_indices_rows] = self.model.predict_proba(X, as_pandas=False)
```
gradientsky, line 162: is it necessary to create lists? I think updates can be done directly with index values using iloc

tabular/src/autogluon/tabular/models/image_prediction/image_predictor.py
```diff
@@ -127,9 +146,44 @@ def _fit(self,
...
+    def _compute_dummy_pred_proba(self, y):
+        num_classes = self.num_classes
+        if self.problem_type in [BINARY, MULTICLASS]:
+            dummies = pd.get_dummies(y)
+            dummy_columns = set(list(dummies.columns))
+            if len(dummies.columns) < num_classes:
+                for c in range(num_classes):
+                    if c not in dummy_columns:
+                        dummies[c] = 0
+            dummies = dummies[list(range(num_classes))]
```
gradientsky, line 178: dummies = dummies.iloc[range(num_classes)]

tabular/src/autogluon/tabular/models/image_prediction/image_predictor.py
Innixma: y_pred_proba is not a DataFrame, its a numpy ndarray, it doesn't have an index or iloc.

tabular/src/autogluon/tabular/models/image_prediction/image_predictor.py
Innixma: That isn't what I am trying to do, I am re-ordering the column names so it ensures they are in the same order as in predict_proba output.

szha:
Job PR-1210-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1210/1/index.html

PR #1211, opened 2021-06-25
Features: Added IsNanFeatureGenerator

Innixma:
*Issue #, if available:*

*Description of changes:*

- Adds IsNanFeatureGenerator which adds boolean flag features on if values are null.
- Currently enabled only for image path features, to indicate to tabular models whether an image is present or not.

### DIFF
--- a/features/src/autogluon/features/generators/isnan.py
+++ b/features/src/autogluon/features/generators/isnan.py
@@ -1,0 +1,69 @@
+import logging
+
+import numpy as np
+import pandas as pd
+from pandas import DataFrame
+
+from autogluon.core.features.types import R_OBJECT
+
+from .abstract import AbstractFeatureGenerator
+
+logger = logging.getLogger(__name__)
+
+
+class IsNanFeatureGenerator(AbstractFeatureGenerator):
+    """
+    Transforms features into isnull flags.
+
+    Parameters
+    ----------
+    null_map : dict, default {'object': ''}
+        Map which dictates the values to consider as NaN.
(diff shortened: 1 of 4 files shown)


features/src/autogluon/features/generators/isnan.py
```diff
@@ -0,0 +1,69 @@
...
+
+    # TODO: Try returning bool type instead of uint8
+    def _transform(self, X: DataFrame) -> DataFrame:
+        is_nan_features = dict()
+        for feature in self.features_in:
+            if feature in self._null_feature_map:
+                null_val = self._null_feature_map[feature]
+                is_nan_features['__nan__.' + feature] = (X[feature] == null_val).astype(np.uint8)
+            else:
+                is_nan_features['__nan__.' + feature] = X[feature].isnull().astype(np.uint8)
```
gradientsky, line 54: why not keep it `bool`?

gradientsky (approved):

features/src/autogluon/features/generators/isnan.py
Innixma: Because we haven't tested using bools anywhere in Tabular models yet, so it would need testing

Innixma: also: uint8 is same size as bool: 1 byte

szha:
Job PR-1211-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1211/1/index.html

PR #1216, opened 2021-06-29
Tabular: Added experimental refit_folds option to BaggedEnsembleModel

Innixma:
*Issue #, if available:*

*Description of changes:*

- Added experimental refit_folds option to BaggedEnsembleModel. This enables performing refit_full automatically in a single .fit call. This can be used as a more expensive (in training time) alternative to efficient OOB / LOO bagging that produces less information leakage in OOF predictions.
- Fixed bug in BaggedEnsembleModel that did not respect `save_bag_folds=False` (10x disk-space reduction after fix for situations where refit was going to be called later, such as high_quality and good_quality presets).

### DIFF
--- a/core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
+++ b/core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
@@ -56,6 +56,7 @@ class BaggedEnsembleModel(AbstractModel):
         default_params = {
             # 'use_child_oof': False,  # [Advanced] Whether to defer to child model for OOF preds and only train a single child.
             'save_bag_folds': True,
+            # 'refit_folds': False,  # [Advanced, Experimental] Whether to refit bags immediately to a refit_full model in a single .fit call.
         }
         for param, val in default_params.items():
             self._set_default_param_value(param, val)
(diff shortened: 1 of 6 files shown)


szha:
Job PR-1216-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1216/1/index.html

core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
```diff
@@ -164,64 +162,37 @@ def _fit(self,
...
+                    # Reserve time for final refit model
+                    kwargs['time_limit'] = kwargs['time_limit'] * folds_to_fit / (folds_to_fit + 1.2)
+            self._fit_folds(X=X, y=y, model_base=model_base, k_fold=k_fold, k_fold_start=k_fold_start, k_fold_end=k_fold_end,
+                            n_repeats=n_repeats, n_repeat_start=n_repeat_start, save_folds=save_bag_folds, **kwargs)
+            # FIXME: Don't save folds except for refit
+            # FIXME: Cleanup self
+            # FIXME: Don't add `_FULL` to name
+            if refit_folds:
+                refit_template = self.convert_to_refit_full_template()
+                refit_template.params['use_child_oof'] = False
```
gradientsky, line 186: nit: should this be in `convert_to_refit_full_template()` I assume this is what we always do on refit

gradientsky (approved):

core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
Innixma: not necessarily because usually we don't directly hardcode the parent's OOF preds to the refit child, it is totally possible for the child to get its own OOF via LOO / efficient OOB. With that said, in future we may want to simplify and just always hardcode parent OOF to refit child OOF, as the nuance isn't necessarily all that helpful in practice.

Issue #1120, opened 2021-05-17
How can I add my own CV split?


YirunKCL:
Hi, I am working on a time series problem. Apparently, random k fold is not proper so I would like to use my own timeseries cv split. How can I add this to the models?

Innixma:
Custom CV splits are not yet supported, but are planned for v0.3. See existing issues: #617, #901, #999

Closing as duplicate for now, will communicate in linked issues when custom CV splits are implemented.

gauravbrills:
This will be very useful .. similar to something h2o automl does

Innixma:
@gauravbrills This was added in v0.3 release with #1224. Feel free to respond further if this is not what you are looking for.

gauravbrills:
> @gauravbrills This was added in v0.3 release with #1224. Feel free to respond further if this is not what you are looking for.

yes just saw that and updated looks good Thanks .

eromoe:
Group doesn't fit many usecase, for example I want to use sklearn TimeSeriesSpliter , or apply Purged Cross-Validation (https://gist.github.com/Burntt/f26e5414205542207949aeb9e9cc1ddb) ,  need control generation of train_index and test_index .

Issue #901, opened 2021-02-02
GroupKFold crossvalidation in AutoGluon Tabular?


davidabek1:
Hello, 
Thanks for your work on this AutoML package, 
**Is GroupKFold-cross-validation supported in AutoGluon Tabular?**
As AutoGluon creating validation sets for internal search of best model and hyperparameters, it has to create the train/val sets based on allowed groups, or training will be "contaminated".
currently with other models I'm using nested cross-validation approach, where in my inner cv using randomized/baysian parameters search, the lines of code considering the groups indices:
```
cv_inner = GroupKFold(n_splits=3)
clf = RandomizedSearchCV(model, param_distributions=param_dist, cv=cv_inner, n_iter=200, scoring='accuracy')
search = clf.fit(X_train, y_train, groups=groups_train)
```
Thank you.

Innixma:
This is not supported at present, but has been requested before and is a reasonable enhancement to add.

AutoGluon currently supports KFold splits by the following code:

```
def generate_kfold(X, y=None, n_splits=5, random_state=0, stratified=False, n_repeats=1):
    if stratified and (y is not None):
        if n_repeats > 1:
            kf = RepeatedStratifiedKFold(n_splits=n_splits, n_repeats=n_repeats, random_state=random_state)
        else:
            kf = StratifiedKFold(n_splits=n_splits, shuffle=True, random_state=random_state)

        kf.get_n_splits(X, y)
        return [[train_index, test_index] for train_index, test_index in kf.split(X, y)]
    else:
        if n_repeats > 1:
            kf = RepeatedKFold(n_splits=n_splits, n_repeats=n_repeats, random_state=random_state)
        else:
            kf = KFold(n_splits=n_splits, shuffle=True, random_state=random_state)

        kf.get_n_splits(X)
        return [[train_index, test_index] for train_index, test_index in kf.split(X)]
```

Note that `n_repeats` allows for repeated bagging / cv. It seems that `GroupKFold` would not support `n_repeats>1`, so we would need to force `n_repeats=1` in this scenario. Similarly, `stratified` and `random_state` arguments would be ignored.

Beyond updates to this function, we will need to add an additional argument to `TabularPredictor.fit` which allows the passing of the `groups` argument downstream, and correctly update it when we perform actions such as data augmentation / dropping rows.

I will plan to look into this when I find time, as it shouldn't be overly difficult to add.

davidabek1:
Thank you, for the fast response, 
[...]

Issue #617, opened 2020-08-14
Time-based crossvalidation in AutoGluon Tabular?


schinto:
Is [time-based-cross-validation](https://towardsdatascience.com/time-based-cross-validation-d259b13d42b8) supported in AutoGluon Tabular?

Innixma:
Not at present, however this is an active area where we are investigating to expand AutoGluon's functionality to time-series problems.

As a clarification, this cross-validation technique would not produce predictions on all training rows since the earliest in time rows would never be part of a validation set during cross-val, is that correct? Thus, with 5-fold cross-val only 4 folds (80%) of the data will receive out-of-fold predictions. That is the primary difficulty, as with multi-layer stacking, future layers would have to remove this data.

kmedved:
That's correct, yes @Innixma - the earliest in time rows would never be included in the validation data. Depending on the size of your training data, you can do 10 folds to ensure that 90% of your data receives a prediction, etc..., but to my knowledge there's no way around the core issue you raise.

Innixma:
As an update to this feature:

We are currently working on time-series functionality. The PR of the WIP time-series support is available here: #684

At present, it's recommended to not use bagging / stacking for time series problems in AutoGluon and instead use a train/val split and pass both into fit as `train_data`, `tuning_data`. Once #684 is merged, the plan is to investigate adding the inner cross-val time-series support for time-series ensembling.

Typically, this technique is used for HPO, so it's unclear if it will give a performance boost in the same way we get a boost in traditional supervised learning, as the amount of training data at each layer will shrink. With enough folds I'd bet it will still be an improvement. Either way, we are interested to try it out.

schinto:
A generalized form of the Splitter concept for CV as used here would be great
https://github.com/deepchem/deepchem/blob/master/deepchem/splits/splitters.py
So that time/clusterid/integer columns can be selected, which contains custom splits, but the columns are excluded from prediction.

jwmueller:
@Innixma time-based crossvalidation or train/val split does not necessarily correspond to time-series. It may still be useful for regular supervised learning where rows are independent, just not identically distributed over time.

I think users who want a time-based single train/val split should just do it themselves though and pass their val data as `tuning_data`.  For bagging/stacking, we could consider supporting arbitrary user-defined k-fold split, although this still doesn't resolve issue of being not fully compatible with a time-based crossvalidation schema due to the issue you mentioned.

[...]

PR #1224, opened 2021-07-05
Tabular: Added user-specified groups parameter, fixed HPO for RF and KNN

Innixma:
*Issue #, if available:*

#1174, #1120, #617, #901

*Description of changes:*

- Added user-specified groups parameter to enable custom fold-splits.
- Fixed HPO for RF and KNN (previously it would crash and not train the models if HPO is specified).

### DIFF
--- a/core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
+++ b/core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
@@ -777,6 +777,7 @@ class BaggedEnsembleModel(AbstractModel):
             y_val_fold = None
             train_index = list(range(len(X)))
             test_index = train_index
+            cv_splitter = None
         else:
             cv_splitter = self._get_cv_splitter(n_splits=k_fold, n_repeats=1, groups=groups)
             if k_fold != cv_splitter.n_splits:
@@ -838,6 +839,8 @@ class BaggedEnsembleModel(AbstractModel):
                 bag.models.append(child)
             bag.val_score = child.val_score
             bag._add_child_times_to_bag(model=child)
+            if cv_splitter is not None:
+                bag._cv_splitters = [cv_splitter]
 
             bag.save()
             bags[bag.name] = bag.path
--- a/core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
+++ b/core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
@@ -14,7 +14,7 @@ from ...constants import MULTICLASS, REGRESSION, SOFTCLASS, QUANTILE, REFIT_FULL
 from ...utils.exceptions import TimeLimitExceeded
 from ...utils.loaders import load_pkl
 from ...utils.savers import save_pkl
-from ...utils.utils import generate_kfold, _compute_fi_with_stddev
+from ...utils.utils import CVSplitter, _compute_fi_with_stddev
 
 logger = logging.getLogger(__name__)
 
@@ -46,6 +46,7 @@ class BaggedEnsembleModel(AbstractModel):
         # TODO: Consider moving `_child_oof` logic to a separate class / refactor OOF logic.
         # FIXME: Avoid unnecessary refit during refit_full on `_child_oof=True` models, just re-use the original model.
         self._child_oof = False  # Whether the OOF preds were taken from a single child model (Assumes child can produce OOF preds without bagging).
+        self._cv_splitters = []  # Keeps track of the CV splitter used for each bagged repeat.
 
         eval_metric = kwargs.pop('eval_metric', self.model_base.eval_metric)
         stopping_metric = kwargs.pop('stopping_metric', self.model_base.stopping_metric)  # FIXME: Has to be moved to post-model_base initialization, otherwise could be misaligned.
@@ -113,16 +114,20 @@ class BaggedEnsembleModel(AbstractModel):
         else:
             return X
 
+    def _get_cv_splitter(self, n_splits, n_repeats, groups=None):
+        return CVSplitter(n_splits=n_splits, n_repeats=n_repeats, groups=groups, stratified=self.is_stratified(), random_state=self._random_state)
+
     def _fit(self,
              X,
              y,
              X_val=None,
              y_val=None,
-             k_fold=5,
+             k_fold=None,
              k_fold_start=0,
              k_fold_end=None,
              n_repeats=1,
              n_repeat_start=0,
+             groups=None,
              **kwargs):
         use_child_oof = self.params.get('use_child_oof', False)
         if use_child_oof:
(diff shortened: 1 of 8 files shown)


tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -91,6 +91,13 @@ class TabularPredictor:
         If True, then weighted metrics will be reported based on the sample weights provided in the specified `sample_weight` (in which case `sample_weight` column must also be present in test data).
         In this case, the 'best' model used by default for prediction will also be decided based on a weighted version of evaluation metric.
         Note: we do not recommend specifying `weight_evaluation` when `sample_weight` is 'auto_weight' or 'balance_weight', instead specify appropriate `eval_metric`.
+    groups : str, default = None
+        [Experimental] If specified, AutoGluon will use the column named the value of groups in `train_data` during `.fit` as the data splitting indices for the purposes of bagging.
```
jwmueller, line 95: may be better left as kwarg while it is experimental?

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -91,6 +91,13 @@ class TabularPredictor:
...
+        This column will not be used as a feature during model training.
+        The data will be split via `sklearn.model_selection.LeaveOneGroupOut`.
+        Use this option to control the exact split indices AutoGluon uses.
```
jwmueller, line 98: probably need to clarify what the `groups` column values should look like. And how it works with bagging vs no-bagging

jwmueller: could be helpful to provide a motivating example: "For example, if you want your data folds to preserve adjacent rows in the table (without shuffling), then the `groups` column should look like ..."

szha:
Job PR-1224-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1224/3/index.html

tabular/src/autogluon/tabular/predictor/predictor.py
Innixma: added example and more clarifications

tabular/src/autogluon/tabular/predictor/predictor.py
Innixma: Since this feature is asked about so much, I'd rather it be highly visible so it isn't missed. By being experimental it does fully work without issue, it's just if the user specifies an invalid or weird grouping, edge cases can occur that may result in strange errors (hard to detect if groups are invalid or not prior to trying). This is mostly related to cases with very small amounts of data.

szha:
Job PR-1224-4 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1224/4/index.html

szha:
Job PR-1224-5 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1224/5/index.html

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -91,6 +91,16 @@ class TabularPredictor:
...
         Note: we do not recommend specifying `weight_evaluation` when `sample_weight` is 'auto_weight' or 'balance_weight', instead specify appropriate `eval_metric`.
+    groups : str, default = None
+        [Experimental] If specified, AutoGluon will use the column named the value of groups in `train_data` during `.fit` as the data splitting indices for the purposes of bagging.
+        This column will not be used as a feature during model training.
+        This parameter is ignored if bagging is not enabled. To instead specify a custom validation set with bagging disabled, specify `tuning_data` in `.fit`.
+        The data will be split via `sklearn.model_selection.LeaveOneGroupOut`.
+        Use this option to control the exact split indices AutoGluon uses.
+        It is not recommended to use this option unless it is required for very specific situations.
+        Bugs may arise from edge cases if the provided groups are not valid to properly train models, such as if not all classes are present during training in multiclass classification. It is up to the user to sanitize their groups.
+
```
gradientsky, line 102: There are couple cases to worry here:
1) Heavy classes imbalances between groups would results with incorrect time limits estimates and potentially memory errors on 'heavy' folds. Sanity check for such imbalance might be helpful.
2) sub-folds might have just a subset of training labels and bags need to be aligned on label space if some form of label encoding is used.
```
...,0,A
...,0,B
...,0,C
...,1,C
...,1,D
...,1,D
```
First fold would know only about A,B,C; 2nd about C,D. Is there handling which would align the fold predictions into an aggregate prediciton?

core/src/autogluon/core/utils/utils.py
```diff
@@ -40,40 +40,66 @@ def get_gpu_count():
...
+        self.groups = groups
+        if splitter_cls is None:
+            splitter_cls = self._get_splitter_cls()
+        self._splitter = self._get_splitter(splitter_cls)
+
+    def _get_splitter_cls(self):
+        if self.groups is not None:
+            num_groups = len(self.groups.unique())
+            if self.n_repeats != 1:
+                raise AssertionError(f'n_repeats must be 1 when split groups are specified. (n_repeats={self.n_repeats})')
```
gradientsky, line 64: Does it makes sense to raise exception - we can log a warning and reset it to 1 automatically.

core/src/autogluon/core/utils/utils.py
```diff
@@ -40,40 +40,66 @@ def get_gpu_count():
...
+            self.n_splits = num_groups
+            splitter_cls = LeaveOneGroupOut
```
gradientsky, line 66: log a warning that the feature is experimental

tabular/src/autogluon/tabular/predictor/predictor.py
Innixma: 1. This is very hard to fix, even if we knew there was imbalance it wouldn't necessarily help us that much.

2. Worst case scenario which causes crash is if a class is only present in 1 fold, meaning that a model that uses that fold as validation won't see that class in training. This is what I'd consider an invalid grouping, and probably isn't worth trying to work with (unless we later determine that there are valid use-cases for this being the case)

core/src/autogluon/core/utils/utils.py
Innixma: This would log spam: This is repeated for every bag.

core/src/autogluon/core/utils/utils.py
Innixma: I prefer an exception, n_repeats is simply not valid for this situation and should never be passed in this fashion.

core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
```diff
@@ -194,6 +198,41 @@ def _fit(self,
...
+                             groups):
+        if groups is not None:
+            if self._n_repeats_finished != 0:
+                raise AssertionError('Bagged models cannot call fit with `groups` specified when a full k-fold set has already been fit.')
+            if n_repeats > 1:
+                raise AssertionError('Cannot perform repeated bagging with `groups` specified.')
+            return
+
+        if k_fold_end is None:
+            k_fold_end = k_fold
```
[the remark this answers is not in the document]
Innixma: This occurs elsewhere. At the point where this validate is called, groups can also be present, meaning k_fold itself can still change, so we don't want to change k_fold_end to the old k_fold value.

jwmueller (approved):
LGTM

PR #1241, opened 2021-07-19
Update with RAPIDS 21.06

sarahyurick:
*Issue #, if available:*
N/A

*Description of changes:*
Tested RF, LR, and KNN with RAPIDS 21.06. The RFRapidsModel accuracy issue has been resolved; however, depending on the max_depth parameter, the RFRapidsModel may be slower than the RFModel (see comment in rf_rapids_model.py). 

### DIFF
--- a/tabular/src/autogluon/tabular/models/knn/knn_rapids_model.py
+++ b/tabular/src/autogluon/tabular/models/knn/knn_rapids_model.py
@@ -18,10 +18,10 @@ class KNNRapidsModel(KNNModel):
     RAPIDS KNearestNeighbors model : https://rapids.ai/start.html
 
     NOTE: This code is experimental, it is recommend to not use this unless you are a developer.
-    This was tested on rapids-0.18 via:
+    This was tested on rapids-21.06 via:
 
-    conda create -n rapids-0.18 -c rapidsai -c nvidia -c conda-forge -c defaults rapids-blazing=0.18 python=3.7 cudatoolkit=10.1 -y
-    conda activate rapids-0.18
+    conda create -n rapids-21.06 -c rapidsai -c nvidia -c conda-forge rapids=21.06 python=3.8 cudatoolkit=11.2
+    conda activate rapids-21.06
     pip install --pre autogluon.tabular[all]
     """
     def _get_model_type(self):
--- a/tabular/src/autogluon/tabular/models/rf/rf_rapids_model.py
+++ b/tabular/src/autogluon/tabular/models/rf/rf_rapids_model.py
@@ -10,17 +10,20 @@ logger = logging.getLogger(__name__)
 
 # TODO: Improve memory safety
 # TODO: Respect time limit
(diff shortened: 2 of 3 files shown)


szha:
Job PR-1241-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1241/1/index.html

tabular/src/autogluon/tabular/models/knn/knn_rapids_model.py
```diff
@@ -18,10 +18,10 @@ class KNNRapidsModel(KNNModel):
     RAPIDS KNearestNeighbors model : https://rapids.ai/start.html
 
     NOTE: This code is experimental, it is recommend to not use this unless you are a developer.
-    This was tested on rapids-0.18 via:
+    This was tested on rapids-21.06 via:
 
-    conda create -n rapids-0.18 -c rapidsai -c nvidia -c conda-forge -c defaults rapids-blazing=0.18 python=3.7 cudatoolkit=10.1 -y
-    conda activate rapids-0.18
+    conda create -n rapids-21.06 -c rapidsai -c nvidia -c conda-forge rapids=21.06 python=3.8 cudatoolkit=11.2
```
gradientsky, line 23: Will this work with python 3.7? We haven't fully migrated to 3.8.

sarahyurick: Yes, it will work with both python 3.7 and 3.8

gradientsky: Also, is CUDA 11.2 min requirement?

sarahyurick: CUDA 11.0 is also supported - I used the RAPIDS Release Selector from https://rapids.ai/start.html to generate the command specific to my local setup

tabular/src/autogluon/tabular/models/rf/rf_rapids_model.py
```diff
@@ -10,17 +10,20 @@
 
 # TODO: Improve memory safety
 # TODO: Respect time limit
-# TODO: Significantly less accurate than RFModel with same hyperparameters.
-#  Refer to https://github.com/rapidsai/cuml/issues/2518
+# TODO: Depending on max_depth parameter, RFRapidsModel is slower than RFModel.
```
Innixma, line 13: Thanks for mentioning this detail. If someone from RAPIDS team has a recommendation at present for a max_depth default value that maintains good accuracy without suffering too heavily from this bug, we can update the default max_depth (Currently 99).

Innixma (approved):
Thanks for this! I tested myself and it does appear the accuracy bug has been fixed, which is very nice and makes the RAPIDS RF model much more useable. Assuming the other bugs related to slow training/inference are fixed, we could begin to use RAPIDS RF as the default RF when GPU is available.

PR #1249, opened 2021-07-25
[Vision] Add torch/timm backed image predictor

zhreshold:
*Issue #, if available:*

*Description of changes:*
Add torch/timm backed image predictor, now either mxnet or torch is optional, as long as one of them is available, image predictor can automatically choose the best strategy for fit.

## Switch between mxnet/pytorch
The underneath framework is dispatched by type of model, all models are available as `image_predictor._SUPPORTED_MODELS`, note that depending on the availability of mxnet/torch(timm, torchvision), you will get different model list.

```python
predictor = ImagePredictor()
predictor.fit(train_dataset, hyperparameters={'model': 'efficientnet_b0'})  # to torch backend
predictor.fit(train_dataset, hyperparameters={'model': 'resnest200'})  # to mxnet backend
```

The default presets have been updated accordingly.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -376,6 +376,7 @@ stage("Build Tutorials") {
         conda list
         ${setup_pip_venv}
         ${setup_mxnet_gpu}
+        ${setup_torch_gpu}
         export CUDA_VISIBLE_DEVICES=${VISIBLE_GPU}
         export AG_DOCS=1
         env
@@ -406,6 +407,7 @@ stage("Build Tutorials") {
         conda list
         ${setup_pip_venv}
         ${setup_mxnet_gpu}
+        ${setup_torch_gpu}
         export CUDA_VISIBLE_DEVICES=${VISIBLE_GPU}
         env
         export LD_LIBRARY_PATH=/usr/local/cuda-10.1/lib64
@@ -680,7 +682,7 @@ stage("Build Docs") {
         cd vision/
         python3 -m pip install --upgrade -e .
         cd ..
-        
+
         cd forecasting/
         python3 -m pip install --upgrade -e .
         cd ..
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -30,6 +30,10 @@ setup_mxnet_gpu = """
     export MXNET_CUDNN_AUTOTUNE_DEFAULT=0
 """
 
+setup_torch_gpu = """
+    python3 -m pip install torch==1.7.1+cu101 torchvision==0.8.2+cu101 -f https://download.pytorch.org/whl/torch_stable.html
+"""
+
 cleanup_venv = """
     deactivate
     rm -rf venv
@@ -264,6 +268,7 @@ stage("Unit Test") {
           conda list
           ${setup_pip_venv}
           ${setup_mxnet_gpu}
+          ${setup_torch_gpu}
           export CUDA_VISIBLE_DEVICES=${VISIBLE_GPU}
           env
 
--- a/vision/src/autogluon/vision/predictor/predictor.py
+++ b/vision/src/autogluon/vision/predictor/predictor.py
@@ -4,6 +4,7 @@ import logging
 import os
 import pickle
 import warnings
+from autogluon.core.space import Categorical
 
 import numpy as np
 import pandas as pd
@@ -118,7 +119,7 @@ class ImagePredictor(object):
                 # Recommended for applications that benefit from the best possible model accuracy.
                 best_quality={
                     'hyperparameters': {
-                        'model': Categorical('resnet50_v1b', 'resnet101_v1d', 'resnest200'),
+                        'model': Categorical('coat_lite_small', 'twins_pcpvt_base', 'swin_base_patch4_window7_224'),
                         'lr': Real(1e-5, 1e-2, log=True),
                         'batch_size': Categorical(8, 16, 32, 64, 128),
                         'epochs': 200,
@@ -135,7 +136,7 @@ class ImagePredictor(object):
                 # Recommended for applications that require reasonable inference speed and/or model size.
                 good_quality_fast_inference={
                     'hyperparameters': {
-                        'model': Categorical('resnet50_v1b', 'resnet34_v1b'),
+                        'model': Categorical('resnet50d', 'efficientnet_b1', 'mobilenetv3_large_100'),
                         'lr': Real(1e-4, 1e-2, log=True),
                         'batch_size': Categorical(8, 16, 32, 64, 128),
                         'epochs': 150,
@@ -152,7 +153,7 @@ class ImagePredictor(object):
                 # This is the default preset in AutoGluon, but should generally only be used for quick prototyping.
                 medium_quality_faster_train={
                     'hyperparameters': {
-                        'model': 'resnet50_v1b',
+                        'model': 'resnet50d',
                         'lr': 0.01,
                         'batch_size': 64,
                         'epochs': 50,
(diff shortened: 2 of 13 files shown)


vision/src/autogluon/vision/predictor/predictor.py
```diff
@@ -400,6 +405,30 @@ def fit(self,
             logging.getLogger('gluoncv.auto.tasks.image_classification').propagate = False
             logging.getLogger("ImageClassificationEstimator").propagate = False
             logging.getLogger("ImageClassificationEstimator").setLevel(log_level)
+
+        # TODO: remove this check when pytorch model supports other problems
+        if self._problem_type != MULTICLASS:
```
Innixma, line 410: Does TIMM support binary classification if it is specified as multiclass with 2 classes?

vision/src/autogluon/vision/predictor/predictor.py
```diff
@@ -715,13 +744,36 @@ def _set_valid_labels(data, label):
...
+    except ImportError:
+        _torch = None
+    try:
+        import timm as _timm
+    except ImportError:
+        _timm = None
+    if _timm is not None:
+        cls_models += list(_timm.list_models())
+    elif _torch is None:
+        logger.warning('timm installed but torch is required to enable it.')
```
Innixma, line 774: This does not seem to trigger in the intended way. This never checks if TIMM was installed.

Innixma (commented):
Excited to see this! Perhaps in this PR's comments could you add a minimal example for reviewers to try that showcase a TIMM model?

vision/src/autogluon/vision/predictor/predictor.py
zhreshold: timm-clean(no deps mock pure python package) is added to dependency so I assume timm is installed anyway, since import timm triggers import torch internally, the hack that will work is probably trying to catch the ImportError and parse it.

szha:
Job PR-1249-23 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1249/23/index.html

zhreshold:
@Innixma passed ci, ready for review

szha:
Job PR-1249-24 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1249/24/index.html

vision/src/autogluon/vision/predictor/predictor.py
```diff
@@ -165,7 +169,7 @@ def fit(self,
                 # Comparing with `medium_quality_faster_train` it uses faster model but explores more hyperparameters.
                 medium_quality_faster_inference={
                     'hyperparameters': {
-                        'model': Categorical('resnet18_v1b', 'mobilenetv3_small'),
+                        'model': Categorical('resnet18', 'mobilenetv3_small_100', 'resnet18_v1b'),
                         'lr': Categorical(0.01, 0.005, 0.001),
                         'batch_size': Categorical(64, 128),
                         'epochs': Categorical(50, 100),
```
jwmueller, line 175: Should `epochs` hyperparameter really be searched?

szha:
Job PR-1249-25 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1249/25/index.html

vision/src/autogluon/vision/predictor/predictor.py
Innixma: Agreed, I don't think we should be searching epochs. Also, why search learning rate via Categorical here but in others it is searched via Real?

vision/src/autogluon/vision/predictor/predictor.py
```diff
@@ -118,7 +122,7 @@ def fit(self,
                 # Recommended for applications that benefit from the best possible model accuracy.
                 best_quality={
                     'hyperparameters': {
-                        'model': Categorical('resnet50_v1b', 'resnet101_v1d', 'resnest200'),
+                        'model': Categorical('coat_lite_small', 'twins_pcpvt_base', 'swin_base_patch4_window7_224'),
```
Innixma, line 125: Why these models? Can we do a benchmark on which models actually give best quality? What about resnet101d and resnet101_v1d?

vision/src/autogluon/vision/predictor/predictor.py
```diff
@@ -118,7 +122,7 @@ def fit(self,
...
                         'lr': Real(1e-5, 1e-2, log=True),
                         'batch_size': Categorical(8, 16, 32, 64, 128),
```
Innixma, line 127: Please double check if batch_size 128 causes the full system to crash (and not continue doing trials). I suspect it does.

vision/src/autogluon/vision/predictor/predictor.py
```diff
@@ -152,7 +156,7 @@ def fit(self,
                 # This is the default preset in AutoGluon, but should generally only be used for quick prototyping.
```
Innixma, line 156: This is not the setting used by default. You should align the actual defaults with this setting.

Jenkinsfile
```diff
@@ -30,6 +30,10 @@ setup_mxnet_gpu = """
     export MXNET_CUDNN_AUTOTUNE_DEFAULT=0
 """
 
+setup_torch_gpu = """
+    python3 -m pip install torch==1.7.1+cu101 torchvision==0.8.2+cu101 -f https://download.pytorch.org/whl/torch_stable.html
```
Innixma, line 34: Is this necessary for the default install of AG to use TIMM? How does this differ from pip install torch as in Tabular?

Our CI should probably not be installing a different version of Torch than is installed by users unless we document it in our installation instructions.

vision/src/autogluon/vision/predictor/predictor.py
zhreshold: @Innixma These are the cpp data result for different models, basically they have the same trend of large dataset(imagenet): https://github.com/rwightman/pytorch-image-models/blob/d5ed58d623be27aada78035d2a19e2854f8b6437/results/results-imagenet-real.csv
```csv
model,score_test,score_val,pred_time_test,pred_time_val,fit_time,pred_time_test_marginal  ,pred_time_val_marginal,fit_time_marginal,stack_level,can_infer,fit_order
1     swin_base_patch4_window7_224    0.940399   0.944386       14.661540      10.654264   764.274587                14.661540               10.654264         764.274587            1       True          4
2     twins_pcpvt_base    0.929423   0.937033       10.634345       7.909998   855.020386                10.634345                7.909998         855.020386            1       True          3
3     coat_lite_small    0.924517   0.930694        9.310690       7.005245   517.815660                 9.310690                7.005245         517.815660            1       True          2
4     gluon_resnet101_v1d    0.915675   0.939413        9.036937       6.305846   484.620409                 9.036937                6.305846         484.620409            1       True          1
```

Jenkinsfile
zhreshold: timm-clean is a pure python package with no dependency at all so it's safe. Regarding torch installation, it's okay to use single `pip install torch` but how do you make sure that pip will pick up the right version? AFAIK pip install torch will require the specific cuda version otherwise you have to use conda install with the proper cuda version.


For example, the default version of cuda is 10.2 if you install using `pip install torch`  per instruction here(https://pytorch.org/get-started/locally/)

szha:
Job PR-1249-26 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1249/26/index.html

szha:
Job PR-1249-27 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1249/27/index.html

Innixma (approved):
Looks good! Thanks for all the work in integrating TIMM into GluonCV and AutoGluon!

Issue #1247, opened 2021-07-23
typo in object detection tutorial causes unexpected behavior


giaco5988:
Hi and thank you for all your work,

I think I have found a typo in the [tutorial docs](https://auto.gluon.ai/stable/tutorials/object_detection/beginner.html)

```
time_limit = 60*30  # at most 0.5 hour
detector = ObjectDetector()
hyperparameters = {'epochs': 5, 'batch_size': 8}
hyperparamter_tune_kwargs={'num_trials': 2}
detector.fit(dataset_train, time_limit=time_limit, hyperparameters=hyperparameters, hyperparamter_tune_kwargs=hyperparamter_tune_kwargs)
```
In the last line the `hyperparamter_tune_kwargs` is mispelled, I think this causes `num_trials` not to be updated.

Innixma:
Thanks for this! Very well spotted.

This has been fixed in #1251.

PR #1251, opened 2021-07-25
Vision: Added exception if invalid kwargs specified

Innixma:
*Issue #, if available:*
#1247

*Description of changes:*

- Added error if kwarg is not valid in fit call to ImagePredictor and ObjectDetector. Previously, the invalid kwarg was silently ignored.
- Fixed typo in object detection tutorial causing incorrect hpo behavior.
- Fixed bug in HPO causing crash if all configurations have been tried and another trial is attempted.
- Fixed vision specific bug in HPO that caused bayesopt to crash by attempting to train the same config twice.
- Removed unusable HPO options from vision documentation.
- Fixed typos in various other tutorials that had incorrect HPO options specified.

### DIFF
--- a/vision/src/autogluon/vision/predictor/predictor.py
+++ b/vision/src/autogluon/vision/predictor/predictor.py
@@ -220,7 +220,7 @@ class ImagePredictor(object):
                 num_trials : int, default = 1
                     The limit of HPO trials that can be performed within `time_limit`. The HPO process will be terminated
                     when `num_trials` trials have finished or wall clock `time_limit` is reached, whichever comes first.
-                search_strategy : str, default = 'random'
+                searcher : str, default = 'random'
                     Searcher strategy for HPO, 'random' by default.
                     Options include: ‘random’ (random search), ‘bayesopt’ (Gaussian process Bayesian optimization),
                     ‘grid’ (grid search).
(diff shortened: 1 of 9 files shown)


szha:
Job PR-1251-4 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1251/4/index.html

vision/src/autogluon/vision/predictor/predictor.py
```diff
@@ -409,10 +409,9 @@ def fit(self,
         self._train_classes = train_data.classes
         with warnings.catch_warnings(record=True) as w:
             warnings.simplefilter("always")
-            with MXNetErrorCatcher() as err:
-                self._classifier = task.fit(train_data, tuning_data, 1 - holdout_frac, random_state)
-            if err.exc_value is not None:
-                raise RuntimeError(err.exc_value + err.hint)
+            # TODO: MXNetErrorCatcher was removed because it didn't return traceback
+            #  Re-add once it returns full traceback regardless of which exception was caught
+            self._classifier = task.fit(train_data, tuning_data, 1 - holdout_frac, random_state)
```
Innixma, line 414: @zhreshold This was done because previously the error would not show stack trace making it extremely hard to debug where errors are coming from. If you can add that functionality to the error catcher, feel free to add it back in, but we need stack traces to be visible.

zhreshold (approved):

PR #1253, opened 2021-07-27
add .md jupyter editing link

jwmueller:


szha:
Job PR-1253-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1253/1/index.html

Innixma:
@jwmueller Thoughts on having our docs be ipynb by default and then converted to markdown in the build phase?

jwmueller:
That seems fine to me, but we should first ask Mu why he chose to do everything in markdown to begin with as he probably had good reasons.

Innixma:
@mli Any reason why we use markdown directly instead of using notebooks and converting them to markdown during the doc build?

mli:
the only reason for using markdown is it's easy to diff the changes. also
people can use any markdown editor to change the text or even do minor
changes of the code.

szha:
Job PR-1253-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1253/3/index.html

PR #1266, opened 2021-08-04
Tabular: Refactor AbstractModel

Innixma:
*Issue #, if available:*

*Description of changes:*

- Fixed incorrect implementation of `convert_to_template`
- Fixed incorrect output of `_get_init_args`, renamed to `get_params` to align with sklearn
- Moved `stopping_metric` and `quantile_levels` to ag_args_fit to simplify code logic
- Refactored refit_full logic to minimize chance of defects in future
- Fixed user specified stopping_metric being ignored

### DIFF
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -28,7 +28,6 @@ from ...utils.savers import save_json, save_pkl
 
 logger = logging.getLogger(__name__)
 
-# TODO: Consider removing stopping_metric from init, only use ag_args_fit to specify stopping_metric.
 # TODO: Consider removing quantile_levels from init, only use ag_args_fit to specify quantile_levels.
 
 
@@ -80,8 +79,7 @@ class AbstractModel:
                  problem_type: str = None,
                  eval_metric: Union[str, metrics.Scorer] = None,
                  hyperparameters=None,
-                 quantile_levels=None,
-                 stopping_metric=None):
+                 quantile_levels=None):
 
         if name is None:
             self.name = self.__class__.__name__
@@ -136,11 +134,6 @@ class AbstractModel:
         self._user_params = hyperparameters  # TODO: Delete after initialization?
         if self._user_params is None:
             self._user_params = dict()
-
-        if stopping_metric is not None:
-            if 'stopping_metric' in self._user_params_aux:
-                raise AssertionError('stopping_metric was specified in both hyperparameters ag_args_fit and model init. Please specify only once.')
-        self.stopping_metric = stopping_metric
 
         self.params_trained = dict()
         self._is_initialized = False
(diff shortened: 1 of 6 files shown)


Innixma:
@truebluejason This PR should resolve the issues you were experiencing with `convert_to_template` and `_get_init_args`.

szha:
Job PR-1266-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1266/2/index.html

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -79,9 +76,7 @@ def __init__(self,
                  name: str = None,
                  problem_type: str = None,
                  eval_metric: Union[str, metrics.Scorer] = None,
-                 hyperparameters=None,
-                 quantile_levels=None,
```
jwmueller, line 83: I'm surprised removing this doesn't break quantile regression?

Or perhaps it is longer being properly passed into the model-constructor here: 
https://github.com/awslabs/autogluon/blob/70d64be1419e69e81210d162ffc0827ee7cfc3ea/tabular/src/autogluon/tabular/trainer/model_presets/presets.py#L319-L320

jwmueller: Nevermind was looking at wrong branch, I now see where ag_args is updated in trainer

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -766,37 +767,42 @@ def get_trained_params(self) -> dict:
...
+            name=name,
+            problem_type=problem_type,
+            eval_metric=eval_metric,
+            hyperparameters=hyperparameters,
+        )
+
+        return args
+
     def convert_to_template(self):
         """After calling this function, returned model should be able to be fit as if it was new, as well as deep-copied."""
```
jwmueller, line 791: Docstring might want to clarify model-name remains the same, so may overwrite the old model in the leaderboard?

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -405,28 +394,40 @@ def _initialize(self, X=None, y=None, feature_metadata=None, num_classes=None, *
             if self.num_classes is None:
                 label_cleaner = LabelCleaner.construct(problem_type=self.problem_type, y=y)
                 self.num_classes = label_cleaner.num_classes
+
+        self._init_params_aux()
```
jwmueller, line 398: It may be helpful for the `_init_params_aux()` method to also have docstring that briefly mentions AG_ARGS_FIT and explains what could be found in the  `params_aux` it updates.

jwmueller (approved):
LGTM

core/src/autogluon/core/models/abstract/abstract_model.py
Innixma: added

PR #1274, opened 2021-08-12
Automated Feature Selection

truebluejason:
*Issue #, if available:*

*Description of changes:*
Add the ability to perform stack-level permutation feature importance based feature pruning in AbstractTrainer. The workflow is as follows for now.

1. Fit all base learners once.
2. Select one of the base learners as the proxy model, and repeatedly fit on different feature subsets based on permutation feature importance scores to find the best feature subset. Pruning threshold is determined by permutation feature importance of a synthetic feature by default. Unless specified otherwise, proxy model feature pruning receives maximum of 2x the time it took to fit all the learners in the stack layer and 5% of the remaining layer fit time.
3. Refit all base learners once on the best feature subset. If validation score improves, use the newly fitted model. Otherwise, use the old model.

Optimizations (on top of RFE with permutation feature importance)
1. Proxy Model: We use a LightGBM-based model (bagged or not depends on presets) with the best validation score in the stack layer. If there are >50000 trianing rows, only train and evaluated the proxy model on 50000 random training rows. If there are >60000 rows and the proxy model is bagged, use the unbagged version of the proxy model.
2. Prune Threshold: Pruning threshold is determined by permutation feature importance of a random feature sampled from the standard normal distribution.
3. Feature Removal: Remove up to 5% of all features at a time whose feature importance is below prune threshold. First try to remove bottom 5% of features. If that doesn't improve the validation score, randomly sample 5% of features where the probability of sampling a feature is proportional to its feature importance rank (feature with lowest feature importance has the highest probability of being sampled). We ensure at least 1% of the feature space is pruned per iteration.
4. Permutation Feature Importance Shuffles: Ensure permutation feature importance for each feature is evaluated on at least 10,000 datapoints. If the number of shuffle required is >20 due to small dataset size, fix it to 20 shuffles.
5. Feature Importance Computation: If feature importance statistics from previous proxy model fits are available within the single RFE loop and we do not have time to evaluate all feature importance statistics, use some of them as the current feature importance estimates. When there are too many features in the dataset, prioritize evaluating feature importance for features we have no estimates from then features with low feature importance from previous fit estimates.


szha:
Job PR-1274-12 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1274/12/index.html

szha:
Job PR-1274-13 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1274/13/index.html

core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
```diff
@@ -134,7 +134,11 @@ def _fit(self,
              n_repeat_start=0,
              groups=None,
              **kwargs):
-        use_child_oof = self.params.get('use_child_oof', False)
+        # HACK to be able to change these variables at fit time
+        self.params['use_child_oof'] = kwargs.get('use_child_oof', self.params.get('use_child_oof', False))
+        self.params['save_bag_folds'] = kwargs.get('save_bag_folds', self.params.get('save_bag_folds', True))
```
jwmueller, line 139: why can't these just be passed into fit() thru `params`?

core/src/autogluon/core/models/ensemble/stacker_ensemble_model.py
```diff
@@ -217,6 +217,8 @@ def _add_stack_to_feature_metadata(self):
                 self.feature_metadata = stacker_feature_metadata
             else:
                 # FIXME: This is a hack, stack feature special types should be already present in feature_metadata, not added here
+                existing_features = self.feature_metadata.get_features()
                 existing_stack_features = self.feature_metadata.get_features(required_special_types=[S_STACK])
+                stacker_feature_metadata = stacker_feature_metadata.keep_features([feature for feature in existing_features if feature in type_map_raw])
```
jwmueller, line 222: could you explain why L222 needed for feature-selection, but was not needed before?

tabular/tests/unittests/test_tabular.py
```diff
@@ -447,6 +447,49 @@ def test_tabularHPO():
...
+    seed_val = 99  # random seed
+    subsample_size = None
+    ag_args = {
+        'feature_prune_kwargs': {
+            'stop_threshold': 3,
+            'prune_ratio': 0.05,
+            'prune_threshold': None,
+            'n_train_subsample': 10000,
+            'n_fi_subsample': 5000,
+            'min_fi_samples': 10000,
```
jwmueller, line 463: smaller values might make this test run quicker? is there any advantage to using bigger values here?

tabular/tests/unittests/test_tabular.py
```diff
@@ -447,6 +447,49 @@ def test_tabularHPO():
...
+    verbosity = 2  # how much output to print
+    time_limit = None
+    fast_benchmark = True  # False
+    # If True, run a faster benchmark (subsample training sets, less epochs, etc),
+    # otherwise we run full benchmark with default AutoGluon settings.
+    # performance_value warnings are disabled when fast_benchmark = True.
+
+    #### If fast_benchmark = True, can control model training time here. Only used if fast_benchmark=True ####
+    if fast_benchmark:
+        subsample_size = 1000
```
jwmueller, line 477: the above setting of `n_train_subsample=10000` seems to have no effect if `subsample_size=1000`

tabular/src/autogluon/tabular/learner/default_learner.py
```diff
@@ -207,7 +207,7 @@ def general_data_processing(self, X: DataFrame, X_val: DataFrame, X_unlabeled: D
                 else:
                     y_unlabeled = pd.Series(np.nan, index=X_unlabeled.index)
                     y_super = pd.concat([y, y_unlabeled], ignore_index=True)
-                X_super = self.fit_transform_features(X_super, y_super, problem_type=self.label_cleaner.problem_type_transform)
+                X_super = self.fit_transform_features(X_super, y_super, problem_type=self.label_cleaner.problem_type_transform, eval_metric=self.eval_metric)
```
jwmueller, line 210: why is `eval_metric` needed in this call to `fit_transform_features()`, but not in the call to this same function on L190:  https://github.com/awslabs/autogluon/blob/e70e0e85146bbe11b8bebc2cb782e7415a54ff94/tabular/src/autogluon/tabular/learner/default_learner.py#L190

core/tests/unittests/test_feature_selection.py
```diff
@@ -0,0 +1,89 @@
+from autogluon.core.utils.feature_selection import *
```
jwmueller, line 1: should we consider a more end-to-end test of feature-selection on a tiny toy dataset (verifying that the irrelevant feature is actually removed)? in particular is there any way to implement such a test that it can run in a few seconds?

jwmueller:
Noticed you sometime call it `feature_selection` sometimes `feature_prune` but it may be good to standardize the naming for better clarity.

core/tests/unittests/test_feature_selection.py
```diff
@@ -0,0 +1,89 @@
...
+    X = pd.DataFrame({'a': [1, 2]})
+    feature_metadata = FeatureMetadata.from_df(X)
+    X_noised = add_noise_column(X, **args, feature_metadata=feature_metadata)
+    expected_features = X.columns.tolist() + ['noise_prefix_1', 'noise_prefix_2']
+    assert expected_features == X_noised.columns.tolist()
+    assert expected_features == feature_metadata.get_features()
+
+
+def test_merge_importance_dfs():
+    # test first time
```
jwmueller, line 33: add comment to explain what's being tested here

core/tests/unittests/test_feature_selection.py
```diff
@@ -0,0 +1,89 @@
...
+    importance_df['importance'] = rng.standard_normal(len(features)) if importance is None else importance
+    importance_df['stddev'] = rng.standard_normal(len(features))
+    importance_df['p_value'] = None
+    importance_df['n'] = 5 if n is None else n
+    importance_df.set_index('name', inplace=True)
+    importance_df.index.name = None
+    return importance_df
+
+
+def test_add_noise_column():
```
jwmueller, line 20: add comment to explain what's being tested here

core/tests/unittests/test_feature_selection.py
```diff
@@ -0,0 +1,89 @@
...
+    assert result_df['n'].tolist() == [15, 15, 10, 5, 0]
+    # test subsequent time where some estimates are from previous runs
+    prev_fit_estimates = set(['a'])
+    result_df = merge_importance_dfs(df1, df2, prev_fit_estimates)
+    assert [score if score == score else None for score in result_df['importance'].tolist()] == [-0.1, 0., 0.1, 1., None]
+    assert result_df['n'].tolist() == [5, 15, 10, 5, 0]
+    assert prev_fit_estimates == set()
+
+
+def test_sort_features_by_priority():
```
jwmueller, line 52: add comment to explain what's being tested here

jwmueller:
It'd be nice to demonstrate the use of feature-pruning as something advanced users can try out in our tabular-indepth tutorial.

jwmueller:
After feature-pruning has been utilized, `predictor.fit_summary()` should clarify which features have been pruned and are no longer necessary to be present in future test data.  And always outputting a similar print statement would be good at the end of `fit()` with feature-pruning.

tabular/src/autogluon/tabular/trainer/abstract_trainer.py
```diff
@@ -1015,6 +1023,27 @@ def _train_and_save(self, X, y, model: AbstractModel, X_val=None, y_val=None, st
...
+                original_score = leaderboard[leaderboard['model'] == original_model.name]['score_val'].item()
+                score_str = f"({round(score, 4)} vs {round(original_score, 4)})"
+                if score > original_score:
+                    logger.log(30, f"Pruned model's score is better than original model's score {score_str}. Replacing original model...")
+                    self.delete_models(models_to_delete=original_model.name, dry_run=False)
+                    self._add_model(model=model, stack_name=stack_name, level=level)
+                    model_names_trained.append(model.name)
+                    self._debug_info['proxy_model'][-1]['score_improvement_from_proxy_yes'] += 1
+                else:
+                    logger.log(30, f"Pruned model's score is not better than original model's score {score_str}. Keeping original model...")
```
jwmueller, line 1038: I don't think "Pruned model" is clear enough terminology. This would be better replaced everywhere with something like "Model trained with feature pruning"

tabular/src/autogluon/tabular/trainer/abstract_trainer.py
```diff
@@ -1015,6 +1023,27 @@ def _train_and_save(self, X, y, model: AbstractModel, X_val=None, y_val=None, st
                 logger.exception('Detailed Traceback:')
             del model
         else:
+            if pruned_features:
+                original_model = self.load_model('_'.join(model.name.split('_')[:-1]))
+                leaderboard = self.leaderboard()
+                original_score = leaderboard[leaderboard['model'] == original_model.name]['score_val'].item()
+                score_str = f"({round(score, 4)} vs {round(original_score, 4)})"
+                if score > original_score:
+                    logger.log(30, f"Pruned model's score is better than original model's score {score_str}. Replacing original model...")
```
jwmueller, line 1032: I don't think "Pruned model" is clear enough terminology. This would be better replaced everywhere with something like "Model trained with feature pruning"

jwmueller: should print the list of features that were pruned and are no longer needed for prediction with future test data.

tabular/src/autogluon/tabular/trainer/abstract_trainer.py
```diff
@@ -953,12 +957,16 @@ def _train_and_save(self, X, y, model: AbstractModel, X_val=None, y_val=None, st
                 if time_limit <= 0:
                     logger.log(15, f'Skipping {model.name} due to lack of time remaining.')
                     return model_names_trained
+                if pruned_features and time_limit and model.is_fit() and model.fit_time > time_limit:
+                    logger.log(15, f'Skipping feature pruning for {model.name} due to projected lack of time.')
```
jwmueller, line 961: I'd make this logging statement level 20, and have your message additionally clarify to the user: 
"Consider increasing `time_limit` to train this model with feature pruning"

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -528,6 +528,12 @@ def fit(self, **kwargs):
             logger.warning(f'\tWarning: Model has no time left to train, skipping model... (Time Left = {round(kwargs["time_limit"], 1)}s)')
```
jwmueller, line 528: After model has been trained with feature-pruning, will `AbstractModel.get_info()` show the remaining features needed for prediction, or the full list of features used during training?  Might be good to store both in this object.

jwmueller:
I don't see much code updated regarding the `predict()` functions, but shouldn't only a reduced set of features need to be present for prediction after fit with feature-pruning? Can you point out where this set is specified?

core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
truebluejason: It's because we don't know which model will be used as the proxy model until validation score results are returned. What do you think about me just setting `self.params['use_child_oof'] = False` in FeatureSelector instead of putting this logic in `fit`? Still hacky but not as much.

core/src/autogluon/core/models/ensemble/stacker_ensemble_model.py
truebluejason: I believe it was because AutoGluon auto-adds all base learner outputs features into stacker_feature_metadata, and if I didn't do this the program would crash when FeatureSelector prunes a base learner output feature.

tabular/tests/unittests/test_tabular.py
truebluejason: Good point - no particular reason, I just ensured that pruning logic does trigger and removes some feature in one of the test dataset and the time limit is not too long (~15s). I will see what happens when I lower this number

core/tests/unittests/test_feature_selection.py
gradientsky, line 52: Consider splitting the tests below into a separate test cases each with fixtures to reduce duplication. Rationale: this will create a finer failure picture when multiple things get broken.

core/tests/unittests/test_feature_selection.py
```diff
@@ -0,0 +1,89 @@
...
+    assert add_noise_column(X, **args) is None
+    X = pd.DataFrame({'a': [1, 2]})
+    feature_metadata = FeatureMetadata.from_df(X)
+    X_noised = add_noise_column(X, **args, feature_metadata=feature_metadata)
+    expected_features = X.columns.tolist() + ['noise_prefix_1', 'noise_prefix_2']
+    assert expected_features == X_noised.columns.tolist()
+    assert expected_features == feature_metadata.get_features()
+
+
+def test_merge_importance_dfs():
```
gradientsky, line 32: Consider splitting the tests below into a separate test cases each with fixtures or separate helper methods to reduce duplication. Rationale: this will create a finer failure picture when multiple things get broken.

truebluejason:
We would highlight features that are not needed anymore if all base learners in AutoGluon do not use them, but the majority of the time at least one model uses the unpruned features because its validation accuracy does not improve when using the pruned features so I don't know if it will be really worth the effort.

core/src/autogluon/core/utils/utils.py
```diff
@@ -949,3 +952,14 @@ def get_gpu_free_memory():
...
+
+
+def unevaluated_fi_df_template(features: List[str]) -> pd.DataFrame:
+    importance_df = pd.DataFrame({'name': features})
+    importance_df['importance'] = None
+    importance_df['stddev'] = None
+    importance_df['p_value'] = None
+    importance_df['n'] = 0
+    importance_df.set_index('name', inplace=True)
+    importance_df.index.name = None
```
gradientsky, line 964: Alternatively in one operation:
```python
importance_df = pd.DataFrame({
    'importance': None,
    'stddev': None,
    'p_value': None,
    'n': 0
}, index=features)
```

core/src/autogluon/core/utils/feature_selection.py
```diff
@@ -0,0 +1,497 @@
...
+                if score >= best_info['score'] * (1 + improvement_threshold):
+                    logger.log(30, f"\tFit {index} ({fit_time}s): Current score {score} is better than best score {best_info['score']}. Updating model.")
+                    logger.log(30, f"\tOld # Features: {len(best_info['features'])} / New # Features: {len(candidate_features)}.")
+                    prev_best_model = best_info['model']
+                    best_info = {'model': curr_model, 'features': candidate_features, 'score': score, 'index': index}
+                    if not self.keep_models and prev_best_model.name != self.original_model.name:
+                        prev_best_model.delete_from_disk()
+                    self._debug_info['index_trajectory'].append(True)
+                else:
+                    logger.log(30, f"\tFit {index} ({fit_time}s): Current score {score} is not better than best score {best_info['score']}. Retrying.")
```
jwmueller, line 232: "Fit k" is not very informative output for user, consider more clear print statement, eg. "Fit in kth round of feature pruning".  

For all print statements, I'd assume the user does not know what is happening internally, so make them as informative as possible while still concise.

core/src/autogluon/core/utils/feature_selection.py
```diff
@@ -0,0 +1,497 @@
...
+                fn_args.update(self.get_extra_fn_args(**kwargs))
+                candidate_features, importance_df, success, prune_time = self.compute_next_candidate(fn_args, time_budget_fi, prev_candidate_features,
+                                                                                                     importance_df)
+                if not success:
+                    logger.log(30, f"\tTime is up while computing feature importance or there are no more features to prune. Ending...")
+                    break
+                curr_model, score, fit_time = self.fit_score_model(X, y, X_val, y_val, candidate_features, model_name, **kwargs)
+
+                if score >= best_info['score'] * (1 + improvement_threshold):
+                    logger.log(30, f"\tFit {index} ({fit_time}s): Current score {score} is better than best score {best_info['score']}. Updating model.")
```
jwmueller, line 224: "Fit k" is not very informative output for user, consider more clear print statement, eg. "Fit in kth round of feature pruning".  

For all print statements, I'd assume the user does not know what is happening internally, so make them as informative as possible while still concise.

core/src/autogluon/core/utils/feature_selection.py
```diff
@@ -0,0 +1,497 @@
...
+                        prev_best_model.delete_from_disk()
+                    self._debug_info['index_trajectory'].append(True)
+                else:
+                    logger.log(30, f"\tFit {index} ({fit_time}s): Current score {score} is not better than best score {best_info['score']}. Retrying.")
+                    if not self.keep_models:
+                        curr_model.delete_from_disk()
+                    self._debug_info['index_trajectory'].append(False)
+
+                if max_fits is not None and index >= max_fits:
+                    logger.log(30, f"\tReached maximum number of fits {max_fits}. Ending...")
```
jwmueller, line 238: user may not understand why there are many fits happening here so would clarify as:
"Reached maximum allowed number of fits allowed during feature-pruning"

core/src/autogluon/core/utils/feature_selection.py
```diff
@@ -0,0 +1,497 @@
...
+            Prune up to prune_ratio * number of current features at once whose feature importance scores are below prune_threshold when
+            generating new feature subset candidates. At least one feature is always removed if there are any feature whose importance
+            score is below prune_threshold.
+        stop_threshold : int, default 10
+            If the validation scores of models fit on pruned data do not improve for stop_threshold iterations, end the pruning procedure.
+        improvement_threshold: int, default = 1e-6
+            The newly fitted model's validation score must be >= improvement_threshold * best validation score seen so far for its input
+            feature subset to be considered to be superior to the previous feature subset.
+        max_fits: int, default None
+            If this many models have been fitted during feature pruning, exit the feature pruning loop. Can potentially prevent overfitting.
```
jwmueller, line 174: would clarify: we re-fit the model using the remaining features after each round of feature-pruning

core/src/autogluon/core/utils/feature_selection.py
```diff
@@ -0,0 +1,497 @@
...
+        prune_threshold : float, default None
+            Consider features whose feature importance scores are below this threshold for pruning. If set to None, a synthetic columns
+            whose values come from standard normal distribution will be appended to the original dataset, and any features whose feature
+            importance score is lower than the synthetic feature's score will be considered for pruning.
+        prune_ratio : float, default 0.05
+            Prune up to prune_ratio * number of current features at once whose feature importance scores are below prune_threshold when
+            generating new feature subset candidates. At least one feature is always removed if there are any feature whose importance
+            score is below prune_threshold.
+        stop_threshold : int, default 10
+            If the validation scores of models fit on pruned data do not improve for stop_threshold iterations, end the pruning procedure.
```
jwmueller, line 169: may be clearer to substitute "iterations" -> "feature pruning rounds" since iterations might be misconstrued as referring to permutation shuffles

core/src/autogluon/core/utils/feature_selection.py
```diff
@@ -0,0 +1,497 @@
...
+        if auto_threshold:
+            best_info['features'] = [feature for feature in best_info['features'] if self.noise_prefix not in feature]
+        if not self.keep_models and best_info['model'].name != self.original_model.name:
+            best_info['model'].delete_from_disk()
+        self._debug_info['total_prune_time'] = self.original_time_limit - self.time_limit
+        self._debug_info['total_prune_fit_time'] = self._fit_time_elapsed
+        self._debug_info['total_prune_fi_time'] = self._fi_time_elapsed
+        self._debug_info['kept_ratio'] = len(best_info['features']) / len(original_features)
+        logger.log(30, f"\tSuccessfully ended prune loop after {index} iterations. Best score: {best_info['score']}.")
+        logger.log(30, f"\tFeature Count: {len(original_features)} -> {len(best_info['features'])} ({round(self.original_time_limit - self.time_limit, 2)}s)")
```
jwmueller, line 264: "Count:" -> "Count before/after pruning:"

jwmueller: should `original_features` and reduced features be somehow both stored in the model's `get_info()` for full reproducibility/informativeness?

core/src/autogluon/core/utils/feature_selection.py
```diff
@@ -0,0 +1,497 @@
...
+        n_evaluated_features = max([i for i in range(0, n_features+1) if i * expected_single_feature_time <= time_budget])
+        if n_evaluated_features == 0:
+            prune_time = time.time() - time_start
+            self.time_limit = self.time_limit - prune_time
+            self._fi_time_elapsed = self._fi_time_elapsed + prune_time
+            return features, unevaluated_fi_df_template(features), prune_time
+        evaluated_features = features[:n_evaluated_features]
+        unevaluated_features = features[n_evaluated_features:]
+        time_budget_fi = time_budget - (time.time() - time_start)
+        logger.log(30, f"\tComputing feature importance for {n_evaluated_features}/{n_features} features with {n_shuffle} shuffles.")
```
jwmueller, line 326: FYI logging level 30 is supposed to be reserved for important statements. Regular statements should be logging level 20, and statements only related to internal operations should probably be left at 15 so they only appear for users who've purposefully set the verbosity level to be higher (to avoid overwhelming them)

core/src/autogluon/core/utils/feature_selection.py
```diff
@@ -0,0 +1,497 @@
...
+        features = unevaluated_rows.index.tolist() + prev_fit_evaluated_rows.index.tolist() + curr_fit_evaluated_rows.index.tolist()
+    if auto_threshold:
+        features = list(prioritized) + [feature for feature in features if feature not in prioritized]
+    return features
+
+
+class FeatureSelector:
+    def __init__(self, model: AbstractModel, time_limit: float, seed: int = 0, keep_models: bool = False, raise_exception=False) -> None:
+        # TODO: Make this work with unlabelled data
+        assert time_limit is not None, "Time limit cannot be unspecified."
```
gradientsky, line 110: It is possible that the user is not specifying a time limit. It would be reasonable in this case to pick a good default and log a warning that the default time limit was picked and to change it, a specific parameter needs to be adjusted.

core/src/autogluon/core/utils/feature_selection.py
```diff
@@ -0,0 +1,497 @@
...
+        as well as some features whose importance scores are below threshold if more than prune_ratio * num features
+        features are below threshold. In the latter case, randomly sample without replacement from features whose
+        importance scores are below threshold until removal candidate configuration that has not yet been tried
+        is encountered. Give higher probability to features whose scores are lower than others when sampling.
+        """
+        time_start = time.time()
+        n_remove = max(1, int(prune_ratio * len(importance_df)))
+        above_threshold_rows = importance_df[(importance_df['importance'] > prune_threshold) | (importance_df['importance'].isna())]
+        below_threshold_rows = importance_df[importance_df['importance'] <= prune_threshold].sort_values(by='importance', axis=0, ascending=True)
+        logger.log(30, f"\tNumber of identified features below prune threshold {round(prune_threshold, 4)}: {len(below_threshold_rows)}/{len(importance_df)}")
```
jwmueller, line 371: statement related to internal operations should have lower logging level (probably 15)

core/src/autogluon/core/utils/feature_selection.py
```diff
@@ -0,0 +1,497 @@
...
+        min_fi_samples = kwargs.get('min_fi_samples', 10000)
+        max_fi_samples = kwargs.get('max_fi_samples', 100000)
+        n_total_samples = max(min_fi_samples, min(max_fi_samples, len(X_fi)))
+        fi_time_single = self.compute_expected_fi_time_single(X_fi, self.model_predict_time, n_subsample, n_total_samples)
+        return min(fi_time_single * min(len(X_fi.columns), 50), self.max_time_budget_fi)
+
+    def fit_score_model(self, X: pd.DataFrame, y: pd.Series, X_val: pd.DataFrame, y_val: pd.Series,
+                        features: List[str], model_name: str, **kwargs) -> Tuple[AbstractModel, float, float]:
+        """
+        Fits and scores a model. Updates self.time_limit. Returns the fitted model, its score, and time elapsed.
```
jwmueller, line 418: description could clarify how this relates to feature pruning, eg.
"Fits and scores a model over the remaining feature subset in a particular feature-pruning round"

core/src/autogluon/core/utils/feature_selection.py
```diff
@@ -0,0 +1,497 @@
...
+        return {
+            'weighted': kwargs.get('weighted', True),
+            'min_fi_samples': kwargs.get('min_fi_samples', 10000),
+            'max_fi_samples': kwargs.get('max_fi_samples', 100000)
+        }
+
+    def consider_subsampling(self, X: pd.DataFrame, y: pd.Series, X_val: pd.DataFrame, y_val: pd.DataFrame, n_train_subsample: int,
+                             **kwargs) -> Tuple[pd.DataFrame, pd.Series, pd.DataFrame, pd.Series, bool, dict]:
+        """
+        If using a proxy model and dataset size is larger than n_train_subsample, subsample data to make model training faster.
```
jwmueller, line 458: unclear why this description mentions proxy model?

core/src/autogluon/core/utils/feature_selection.py
```diff
@@ -0,0 +1,497 @@
...
+        curr_fit_evaluated_rows = prev_importance_df[~(prev_importance_df['importance'].isna()) &
+                                                     ~(prev_importance_df.index.isin(using_prev_fit_fi))].sort_values(by='importance')
+        features = unevaluated_rows.index.tolist() + prev_fit_evaluated_rows.index.tolist() + curr_fit_evaluated_rows.index.tolist()
+    if auto_threshold:
+        features = list(prioritized) + [feature for feature in features if feature not in prioritized]
+    return features
+
+
+class FeatureSelector:
+    def __init__(self, model: AbstractModel, time_limit: float, seed: int = 0, keep_models: bool = False, raise_exception=False) -> None:
```
jwmueller, line 108: why not have proxy model as another (optional) argument

jwmueller: also proxy_model should probably be variable name in the code, given it is mentioned numerous time in the comments and in the `from_proxy` args, otherwise is a bit confusing what the proxy model is exactly.

core/src/autogluon/core/utils/feature_selection.py
```diff
@@ -0,0 +1,497 @@
...
+            the number of feature importance datapoints, set n_fi_subsample = number of feature importance datapoints.
+        prune_threshold : float, default None
+            Consider features whose feature importance scores are below this threshold for pruning. If set to None, a synthetic columns
+            whose values come from standard normal distribution will be appended to the original dataset, and any features whose feature
+            importance score is lower than the synthetic feature's score will be considered for pruning.
+        prune_ratio : float, default 0.05
+            Prune up to prune_ratio * number of current features at once whose feature importance scores are below prune_threshold when
+            generating new feature subset candidates. At least one feature is always removed if there are any feature whose importance
+            score is below prune_threshold.
+        stop_threshold : int, default 10
```
jwmueller, line 168: `prune_threshold` is on the scale of importance scores, whereas `stop_threshold` is on the scale of pruning rounds. I'd consider using a different name for the latter to avoid confusion, eg. `stopping_rounds` or `patience`

core/src/autogluon/core/utils/feature_selection.py
```diff
@@ -0,0 +1,497 @@
...
+            If the validation scores of models fit on pruned data do not improve for stop_threshold iterations, end the pruning procedure.
+        improvement_threshold: int, default = 1e-6
+            The newly fitted model's validation score must be >= improvement_threshold * best validation score seen so far for its input
```
jwmueller, line 171: missing (1+improvement_threshold in this description)

core/tests/unittests/test_feature_selection.py
jwmueller: one edge case is dataset with only 1 feature, have you verified this is ok there?

szha:
Job PR-1274-14 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1274/14/index.html

tabular/src/autogluon/tabular/trainer/abstract_trainer.py
```diff
@@ -1263,29 +1297,53 @@ def _train_multi_initial(self, X, y, models: List[AbstractModel], k_fold, n_repe
...
+                                            k_fold_end=k_fold, n_repeats=n_repeats, n_repeat_start=0, time_limit=bagged_time_limit, **fit_args)
 
-            models = self._train_multi_fold(models=models, hyperparameter_tune_kwargs=None, feature_prune=False, k_fold_start=k_fold_start, k_fold_end=k_fold, n_repeats=n_repeats, n_repeat_start=0, time_limit=time_limit, **fit_args)
+        multi_fold_time_elapsed = time.time() - multi_fold_time_start
+        if time_limit is not None:
+            time_limit = time_limit - multi_fold_time_elapsed
 
+        if feature_prune and len(models) > 0:
+            feature_prune_time_start = time.time()
+            candidate_features = self._proxy_model_feature_prune(feature_prune_kwargs, time_limit, multi_fold_time_elapsed, kwargs['level'], X.columns.tolist())
```
gradientsky, line 1323: In case if `multimodal` setup is active, please ensure that the columns marked for other modalities are not removed (i.e. images names might be useless feature, but the image itself is a different story).

tabular/src/autogluon/tabular/trainer/abstract_trainer.py
```diff
@@ -1432,6 +1491,67 @@ def _get_dummy_stacker(self, level: int, model_levels: dict, use_orig_features=T
         dummy_stacker.initialize(num_classes=self.num_classes)
         return dummy_stacker
 
+    def _proxy_model_feature_prune(self, feature_prune_kwargs: dict, time_limit: float, layer_fit_time: float, level: int, features: List[str]) -> List[str]:
+        """
+        Uses the best LightGBM-based base learner of this layer to perform time-aware permutation feature importance based feature pruning.
+        If all LightGBM models fail, use the model that achieved the highest validation accuracy. Feature pruning gets the smaller of the
```
gradientsky, line 1497: This gets tricky in cases where Image or Text model was the best. Computer Vision models, in particular will 'find' all other tabular features as useless. Shall those be excluded from model candidates?

tabular/src/autogluon/tabular/trainer/abstract_trainer.py
```diff
@@ -1432,6 +1491,67 @@ def _get_dummy_stacker(self, level: int, model_levels: dict, use_orig_features=T
...
+        if feature_prune_kwargs[proxy_model.name].get('feature_prune_time_limit', None) is not None:
+            feature_prune_time_limit = min(max(time_limit - layer_fit_time, 0), feature_prune_kwargs[proxy_model.name].pop('feature_prune_time_limit'))
+        elif time_limit is not None:
+            feature_prune_time_limit = min(max(time_limit - layer_fit_time, 0), k * layer_fit_time)
+        else:
+            feature_prune_time_limit = k * layer_fit_time
+        logger.log(30, f"Proxy model feature pruning models for up to {round(feature_prune_time_limit)}s...")
+        raise_exception_on_fail = feature_prune_kwargs[proxy_model.name].pop('raise_exception', False)
+        selector = FeatureSelector(model=proxy_model, time_limit=feature_prune_time_limit, raise_exception=raise_exception_on_fail)
+        candidate_features = selector.select_features(**feature_prune_kwargs[proxy_model.name], **self._model_fit_kwargs[proxy_model.name])
```
Innixma, line 1548: storing self._model_fit_kwargs is not what we want to be doing. We need a better way to do this before merging.

tabular/src/autogluon/tabular/trainer/abstract_trainer.py
```diff
@@ -1342,14 +1400,15 @@ def _train_multi(self, X, y, models: List[AbstractModel], hyperparameter_tune_kw
             n_repeats = self.n_repeats
         if (k_fold == 0) and (n_repeats != 1):
             raise ValueError(f'n_repeats must be 1 when k_fold is 0, values: ({n_repeats}, {k_fold})')
-        if time_limit is None:
+        feature_prune = feature_prune_kwargs is not None
+        if time_limit is None and not feature_prune:
```
Innixma, line 1404: Remove `feature_prune` variable, simplify:

`if time_limit is None and feature_prune_kwargs is None`

tabular/src/autogluon/tabular/trainer/abstract_trainer.py
```diff
@@ -1432,6 +1491,67 @@ def _get_dummy_stacker(self, level: int, model_levels: dict, use_orig_features=T
...
+        """
+        leaderboard = self.leaderboard()
+        leaderboard = leaderboard[~leaderboard['model'].str.contains('WeightedEnsemble')]
+        fit_models = leaderboard[(~leaderboard['score_val'].isna()) & (leaderboard['stack_level'] == level)]
+        if len(fit_models) == 0:
+            return features
+        best_fit_models = fit_models.loc[fit_models['score_val'] == fit_models['score_val'].max()]
+        proxy_model = self.load_model(best_fit_models.loc[best_fit_models['fit_time'].idxmin()]['model'])
+
+        if feature_prune_kwargs[proxy_model.name].pop('proxy_model_class', "LGB") == "LGB":
```
gradientsky, line 1531: nit: `feature_prune_kwargs[proxy_model.name]` used 5 more times below in this method. Consider extracting it into a variable to improve readability.

tabular/src/autogluon/tabular/trainer/abstract_trainer.py
```diff
@@ -377,6 +377,11 @@ def stack_new_level_core(self, X, y, models: Union[List[AbstractModel], dict], X
                     for model_name in model_args_fit if 'hyperparameter_tune_kwargs' in model_args_fit[model_name]
                 }
                 kwargs['hyperparameter_tune_kwargs'] = hyperparameter_tune_kwargs
+                feature_prune_kwargs = {
```
Innixma, line 380: The way this is currently used is strange. I understand we originally did this because we were pruning on a per-model basis, but now it is very odd how it is leveraged. At minimum add code comment on why this is structured the way it is and why we give each model its own feature prune kwargs when we inevitably only train a single proxy model and ignore the other model's feature prune kwargs.

tabular/src/autogluon/tabular/trainer/abstract_trainer.py
```diff
@@ -1432,6 +1491,67 @@ def _get_dummy_stacker(self, level: int, model_levels: dict, use_orig_features=T
...
+        If all LightGBM models fail, use the model that achieved the highest validation accuracy. Feature pruning gets the smaller of the
+        remaining layer time limit and k times (default=2) it took to fit the base learners of this layer as its resource. Note that feature pruning can
+        exit earlier based on arguments in feature_prune_kwargs. The method returns the list of feature names that survived the pruning procedure.
+
+        Parameters
+        ----------
+        feature_prune_kwargs : dict
+            Feature pruning kwarg arguments. Should contain arguments passed to FeatureSelector.select_features. One can optionally attach the following
+            additional kwargs that are consumed at this level: 'proxy_model_class' to tell this method not to prioritize LGB as the proxy model,
+            'feature_prune_time_limit' to manually specify how long we should perform the feature pruning procedure for, 'k' to specify how long we should
```
gradientsky, line 1506: Please clarify if `k` is how many rounds will be performed vs time units.

Innixma (commented):
This is a review of the code excluding `feature_selection.py`. I'll review `feature_selection.py` in a separate review.

core/src/autogluon/core/utils/feature_selection.py
```diff
@@ -0,0 +1,497 @@
...
+        removal_candidate_rows = below_threshold_rows[:n_remove]
+        removal_candidates = tuple(removal_candidate_rows.index)
+        if removal_candidates not in self.attempted_removals:
+            acceptance_candidates = importance_df[~importance_df.index.isin(removal_candidates)].index.tolist()
+            self.attempted_removals.add(removal_candidates)
+            return acceptance_candidates
+
+        sample_weights = [1/i for i in range(1, len(below_threshold_rows)+1)] if weighted else None
+        while time_budget - (time.time() - time_start) > 0:
+            random_state = self.rng.integers(low=0, high=1e5)
```
gradientsky, line 387: Do we want this to be non-deterministic?

core/src/autogluon/core/utils/feature_selection.py
```diff
@@ -0,0 +1,497 @@
...
+        # Try removing features with lowest importance first
+        removal_candidate_rows = below_threshold_rows[:n_remove]
+        removal_candidates = tuple(removal_candidate_rows.index)
+        if removal_candidates not in self.attempted_removals:
+            acceptance_candidates = importance_df[~importance_df.index.isin(removal_candidates)].index.tolist()
+            self.attempted_removals.add(removal_candidates)
+            return acceptance_candidates
+
+        sample_weights = [1/i for i in range(1, len(below_threshold_rows)+1)] if weighted else None
+        while time_budget - (time.time() - time_start) > 0:
```
gradientsky, line 386: Is it a good idea to run another iteration even if there's very little time left?

core/src/autogluon/core/utils/feature_selection.py
truebluejason: This is already done in `abstract_trainer`'s `_proxy_model_feature_prune`

tabular/src/autogluon/tabular/trainer/abstract_trainer.py
```diff
@@ -1432,6 +1491,67 @@ def _get_dummy_stacker(self, level: int, model_levels: dict, use_orig_features=T
...
+        else:
+            feature_prune_time_limit = k * layer_fit_time
+        logger.log(30, f"Proxy model feature pruning models for up to {round(feature_prune_time_limit)}s...")
+        raise_exception_on_fail = feature_prune_kwargs[proxy_model.name].pop('raise_exception', False)
+        selector = FeatureSelector(model=proxy_model, time_limit=feature_prune_time_limit, raise_exception=raise_exception_on_fail)
+        candidate_features = selector.select_features(**feature_prune_kwargs[proxy_model.name], **self._model_fit_kwargs[proxy_model.name])
+
+        # TODO: Remove these
+        self._debug_info['proxy_model'].append(selector._debug_info)
+        self._debug_info['proxy_model'][-1]['layer_fit_time'] = layer_fit_time
```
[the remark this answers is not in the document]
truebluejason: Nope - I will remove them after I finish testing things

core/src/autogluon/core/utils/feature_selection.py
truebluejason: Yeap, because we want to propose a different feature subset to prune every time (in case previous attempts failed to improve the validation score)

core/src/autogluon/core/utils/feature_selection.py
truebluejason: The time budget here is feature importance computation time budget so it should be fine.

core/src/autogluon/core/models/abstract/abstract_model.py
truebluejason: It shows the remaining feature only, since we fit a copy of the original model on the pruned set of features. In my opinion just having the remaining features should be fine, but I can add this in if you feel otherwise.

core/tests/unittests/test_feature_selection.py
truebluejason: We didn't previously feature prune in that edge case (because model fit throws an exception) but I added an if statement + logs to make things more clear. Also because we are pruning features based on a noise column's feature importance + validation performance increase, there are no guarantees that all irrelevant features will be removed (although a good number of them will probably be pruned). One option is displaying a warning if no irrelevant features are removed. I'll get to this if I have time after doing everything else.

jwmueller:
> We would highlight features that are not needed anymore if all base learners in AutoGluon do not use them, but the majority of the time at least one model uses the unpruned features because its validation accuracy does not improve when using the pruned features so I don't know if it will be really worth the effort.

Isn't a major advantage of feature-selection that we don't need to keep all the features around anymore? 
What if you have a dataset with 3 useless features (and assume they're properly pruned during feature selection), are you saying these 3 features still need to be there in future test data?  If so, why?

truebluejason:
It would be a major advantage, but the issue now is that while feature pruning generates a reduced feature set to be applied across the entire stack layer, the layer's base learners only work with the reduced feature set if their validation accuracy increases. So for us to not need the 3 useless features all base learners would need to have better validation accuracy on the reduced feature set and I've almost never seen that happen. We can optionally have a flag that forces all base learners to take the reduced feature set though. I would also need to make sure higher layer models don't use the pruned features from the first layer if I were to get this to reliably work.

szha:
Job PR-1274-15 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1274/15/index.html

szha:
Job PR-1274-16 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1274/16/index.html

Innixma (commented):
More review comments, still need to finish reviewing feature_selector.py tomorrow.

core/src/autogluon/core/utils/utils.py
```diff
@@ -955,11 +957,10 @@ def get_gpu_free_memory():
...
-    importance_df['p_value'] = None
-    importance_df['n'] = 0
-    importance_df.set_index('name', inplace=True)
-    importance_df.index.name = None
+    importance_df = pd.DataFrame({
+        'importance': None,
+        'stddev': None,
+        'p_value': None,
+        'n': 0
+    }, index=features)
```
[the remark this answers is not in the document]
truebluejason: I double checked the dataframe that the original feature importance method creates and the index column doesn't have a name.

tabular/src/autogluon/tabular/trainer/abstract_trainer.py
```diff
@@ -1528,30 +1500,61 @@ def _proxy_model_feature_prune(self, feature_prune_kwargs: dict, time_limit: flo
         best_fit_models = fit_models.loc[fit_models['score_val'] == fit_models['score_val'].max()]
         proxy_model = self.load_model(best_fit_models.loc[best_fit_models['fit_time'].idxmin()]['model'])
 
-        if feature_prune_kwargs[proxy_model.name].pop('proxy_model_class', "LGB") == "LGB":
+        if proxy_model_class == "LGBModel":
```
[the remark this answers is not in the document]
Innixma: The class itself is available in the DAG, see 'type_inner' added in `_add_model`.

```
self.model_graph.add_node(
            model.name,
            fit_time=model.fit_time,
            predict_time=model.predict_time,
            val_score=model.val_score,
            path=model.path,
            type=type(model),  # Outer type, can be BaggedEnsemble, StackEnsemble (Type that is able to load the model)
            type_inner=type_inner,  # Inner type, if Ensemble then it is the type of the inner model (May not be able to load with this type)
            can_infer=model.can_infer(),
            can_fit=model.can_fit(),
            is_valid=model.is_valid(),
            stack_name=stack_name,
            level=level,
            **model._fit_metadata,
        )
```

tabular/src/autogluon/tabular/trainer/abstract_trainer.py
```diff
@@ -1432,6 +1491,67 @@ def _get_dummy_stacker(self, level: int, model_levels: dict, use_orig_features=T
...
+        leaderboard = self.leaderboard()
+        leaderboard = leaderboard[~leaderboard['model'].str.contains('WeightedEnsemble')]
+        fit_models = leaderboard[(~leaderboard['score_val'].isna()) & (leaderboard['stack_level'] == level)]
+        if len(fit_models) == 0:
+            return features
+        best_fit_models = fit_models.loc[fit_models['score_val'] == fit_models['score_val'].max()]
+        proxy_model = self.load_model(best_fit_models.loc[best_fit_models['fit_time'].idxmin()]['model'])
+
+        if feature_prune_kwargs[proxy_model.name].pop('proxy_model_class', "LGB") == "LGB":
+            lgb_models = fit_models[fit_models['model'].str.contains('LightGBM')]
```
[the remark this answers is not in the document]
Innixma: The class itself is available in the DAG, see 'type_inner' added in `_add_model`.

```
self.model_graph.add_node(
            model.name,
            fit_time=model.fit_time,
            predict_time=model.predict_time,
            val_score=model.val_score,
            path=model.path,
            type=type(model),  # Outer type, can be BaggedEnsemble, StackEnsemble (Type that is able to load the model)
            type_inner=type_inner,  # Inner type, if Ensemble then it is the type of the inner model (May not be able to load with this type)
            can_infer=model.can_infer(),
            can_fit=model.can_fit(),
            is_valid=model.is_valid(),
            stack_name=stack_name,
            level=level,
            **model._fit_metadata,
        )
```

core/src/autogluon/core/utils/feature_selection.py
```diff
@@ -0,0 +1,537 @@
...
+        subsample_train = n_train_subsample is not None and len(X) > n_train_subsample
+        subsample_val = max_fi_samples is not None and X_val is not None and len(X_val) > max_fi_samples
+
+        if subsample_train:
+            logger.log(30, f"\tNumber of training samples {len(X)} is greater than {n_train_subsample}. Using {n_train_subsample} samples as training data.")
+            if kwargs.get('num_classes', None) is not None:
+                # stratified sample if classification
+                combined = pd.concat([X, y], axis=1)
+                y_name = combined.columns[-1]
+                combined = combined.groupby(y_name).apply(lambda x: x.sample(frac=n_train_subsample/len(X), random_state=random_state)).reset_index(drop=True)
```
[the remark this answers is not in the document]
Innixma: https://github.com/awslabs/autogluon/blob/master/core/src/autogluon/core/utils/utils.py#L362

core/src/autogluon/core/utils/feature_selection.py
```diff
@@ -0,0 +1,537 @@
...
+        self.model_fit_time = model.fit_time
+        self.model_predict_time = model.predict_time
+        self.original_val_score = model.val_score
+        self.trained_models = []
+        self.importance_dfs = []
+        self.attempted_removals = set()
+        self.noise_prefix = 'AG_normal_noise'
+        self.safety_time_multiplier = 1.1
+        self.max_n_shuffle = 100
+        self.max_time_budget_fi = 300
```
Innixma, line 135: seems like an odd limitation. What if 24 hours of time is given? Surely we could spare more than 5 minutes on FI calculations.

core/src/autogluon/core/utils/feature_selection.py
```diff
@@ -0,0 +1,537 @@
...
+
+                if max_fits is not None and index >= max_fits:
+                    logger.log(30, f"\tReached maximum number of allowed fits, {max_fits}, during feature selection. Ending...")
+                    break
+                if index - best_info['index'] >= stopping_round:
+                    logger.log(30, f"\tScore has not improved for {stopping_round} feature pruning rounds. Ending...")
+                    break
+                if self.time_limit <= self.model_fit_time + prune_time:
+                    logger.log(30, f"\tInsufficient time to finish next pruning round. Ending...")
+                    break
```
Innixma, line 261: Why not move into a separate method? This seems overly nested.

core/src/autogluon/core/utils/feature_selection.py
```diff
@@ -0,0 +1,537 @@
...
+        Updates self.time_limit. Returns the fitted model, its score, and time elapsed. If this is the first time we are fitting
+        a model in the pruning procedure, save time and score statistics.
+        """
+        time_start = time.time()
+        model = self.base_model.convert_to_template()
+        X = X[features]
+        X_val = None if self.is_bagged else X_val[features]
+        model.rename(model_name)
+        if 'time_limit' in kwargs:
+            kwargs['time_limit'] = self.time_limit
```
Innixma, line 445: Why overwrite time_limit? Why pass time_limit in the first place if it is going to be overwritten?

core/src/autogluon/core/utils/feature_selection.py
```diff
@@ -0,0 +1,537 @@
...
+        model.fit(X=X, y=y, X_val=X_val, y_val=y_val, **kwargs)
+        fit_time = time.time() - time_start
+        time_start = time.time()
```
Innixma, line 448: time_fit_end = time.time(), then you don't need 2 separate time.time() calls one after the other.

core/src/autogluon/core/utils/feature_selection.py
```diff
@@ -0,0 +1,537 @@
...
+        score = model.score_with_oof(y) if self.is_bagged else model.score(X=X_val, y=y_val)
+        predict_time = time.time() - time_start
+        self.time_limit = self.time_limit - (fit_time + predict_time)
+        if self.model_fit_time is None:
+            self.model_fit_time = fit_time
+        if self.model_predict_time is None:
+            self.model_predict_time = predict_time
+        if self.original_val_score is None:
+            self.original_val_score = score
```
Innixma, line 457: Why? This is weird considering you originally require the model to be fit in init.

core/src/autogluon/core/utils/feature_selection.py
```diff
@@ -0,0 +1,537 @@
...
+        if self.keep_models:
+            self.trained_models.append(model)
```
Innixma, line 459: This is a really great way to run out of memory.

Innixma: Consider looking at how AbstractTrainer does things.

core/src/autogluon/core/utils/feature_selection.py
```diff
@@ -0,0 +1,537 @@
...
+        self._fit_time_elapsed = self._fit_time_elapsed + (fit_time + predict_time)
+        return model, round(score, 4), round(fit_time + predict_time, 4)
```
Innixma, line 461: why are we rounding score and time? Only round when you print, don't destroy the original values.

core/src/autogluon/core/utils/feature_selection.py
```diff
@@ -0,0 +1,537 @@
...
+
+    def get_extra_fn_args(self, **kwargs) -> dict:
```
Innixma, line 463: kwargs as an argument here seems weird. Why not just do a dict update or merge?

core/src/autogluon/core/utils/feature_selection.py
```diff
@@ -0,0 +1,537 @@
...
+                combined = pd.concat([X, y], axis=1)
+                y_name = combined.columns[-1]
+                combined = combined.groupby(y_name).apply(lambda x: x.sample(frac=n_train_subsample/len(X), random_state=random_state)).reset_index(drop=True)
+                X_train, y_train = combined.drop(columns=[y_name]), combined[y_name]
+            else:
+                X_train = X.sample(n=n_train_subsample, random_state=random_state)
+                y_train = y.loc[X_train.index]
+            if kwargs.pop('replace_bag', True) and self.is_bagged and len(X) >= n_train_subsample + kwargs.get('min_fi_samples', 10000):
+                logger.log(30, f"\tThere is a lot of data and the feature selection model is bagged. Using non-bagged model.")
+                self.is_bagged = False
```
Innixma, line 490: updating object variables in here seems not ideal.

core/src/autogluon/core/utils/feature_selection.py
```diff
@@ -0,0 +1,537 @@
...
+                original_k_fold = kwargs['k_fold']
+                self.base_model = self.original_model.convert_to_template_child()
+                self.base_model.rename(self.original_model.name.replace('_BAG', ''))
```
Innixma, line 493: This is weird to do here, it has no relation to the other logic in this method. This is asking for bugs.

core/src/autogluon/core/utils/feature_selection.py
Innixma: Just return is_bagged=False as an output of this method instead.

core/src/autogluon/core/utils/feature_selection.py
```diff
@@ -0,0 +1,537 @@
...
+            if kwargs.get('num_classes', None) is not None:
+                # stratified sample if classification
+                combined = pd.concat([X, y], axis=1)
+                y_name = combined.columns[-1]
+                combined = combined.groupby(y_name).apply(lambda x: x.sample(frac=n_train_subsample/len(X), random_state=random_state)).reset_index(drop=True)
+                X_train, y_train = combined.drop(columns=[y_name]), combined[y_name]
+            else:
+                X_train = X.sample(n=n_train_subsample, random_state=random_state)
+                y_train = y.loc[X_train.index]
+            if kwargs.pop('replace_bag', True) and self.is_bagged and len(X) >= n_train_subsample + kwargs.get('min_fi_samples', 10000):
```
Innixma, line 488: Why is this based on X if we already subsampled to X_train?

Innixma: Shouldn't there be a choice between subsampling the data and disabling bagging?

core/src/autogluon/core/utils/feature_selection.py
```diff
@@ -0,0 +1,537 @@
...
+                self.is_bagged = False
+                original_k_fold = kwargs['k_fold']
+                self.base_model = self.original_model.convert_to_template_child()
+                self.base_model.rename(self.original_model.name.replace('_BAG', ''))
+                X_val = X[~X.index.isin(X_train.index)]
+                y_val = y.loc[X_val.index]
+                if self.model_fit_time is not None:
+                    self.model_fit_time = self.model_fit_time / original_k_fold
+                if self.model_predict_time is not None:
+                    self.model_predict_time = self.model_predict_time / original_k_fold
```
Innixma, line 499: What if it was None? Is this even a valid situation?

core/src/autogluon/core/utils/feature_selection.py
```diff
@@ -0,0 +1,537 @@
...
+            kwargs['feature_metadata'] = deepcopy(kwargs['feature_metadata']) if 'feature_metadata' in kwargs else None
+            X = add_noise_column(X, self.noise_prefix, self.rng, feature_metadata=kwargs.get('feature_metadata', None))
+            X_val = add_noise_column(X_val, self.noise_prefix, self.rng, feature_metadata=kwargs.get('feature_metadata', None))
+        X_fi, y_fi = (X, y) if self.is_bagged else (X_val, y_val)
+        if self.is_bagged:
+            # HACK: Consider not modifying params field
+            self.base_model.params['use_child_oof'] = False
+            self.base_model.params['save_bag_folds'] = True
+        if self.is_bagged and self.original_model._child_oof and self.model_fit_time is not None:
+            self.model_fit_time = self.model_fit_time * kwargs['k_fold']
```
Innixma, line 525: Why? The fit time of a bagged model is already the combined fit times of its children.

core/src/autogluon/core/utils/feature_selection.py
```diff
@@ -0,0 +1,537 @@
...
+        time_budget_select = time_budget - (time.time() - time_start)
+        candidate_features = self.compute_next_candidate_given_fi(importance_df, prune_threshold, prune_ratio, time_budget_select, weighted)
+
+        # if noise columns exist, they should never be removed
+        if auto_threshold:
+            candidate_features = candidate_features + list(prioritized)
+            importance_df = pd.concat([importance_df, noise_rows])
+
+        feature_selection_time = time.time() - time_start
+        self.time_limit = self.time_limit - feature_selection_time
```
Innixma, line 371: This is dangerous. You may want to instead keep track of self.time_limit as a self.time_start, then you can compare time_elapsed as `time_elapsed = time.time() - self.time_start`, `time_left = self.time_limit - time_elapsed`, that way you never edit time_limit.

core/src/autogluon/core/utils/feature_selection.py
```diff
@@ -0,0 +1,537 @@
...
+        evaluated_features = features[:n_evaluated_features]
+        unevaluated_features = features[n_evaluated_features:]
+        time_budget_fi = time_budget - (time.time() - time_start)
+        logger.log(15, f"\tComputing feature importance for {n_evaluated_features}/{n_features} features with {n_shuffle} shuffles.")
+        fi_kwargs = {'X': X, 'y': y, 'num_shuffle_sets': n_shuffle, 'subsample_size': n_subsample, 'features': evaluated_features,
+                     'time_limit': time_budget_fi, 'silent': True}
+        fi_kwargs.update({'is_oof': True} if self.is_bagged else {})
+        evaluated_df = model.compute_feature_importance(**fi_kwargs)
+        if self.is_bagged:
+            evaluated_df['n'] = evaluated_df['n'] // len(model.models)
```
Innixma, line 349: Add a comment why this is done.

core/src/autogluon/core/utils/feature_selection.py
```diff
@@ -0,0 +1,537 @@
...
+
+        # if we could not compute feature importance for all features and previous feature importance estimates exist, use them
+        importance_df = pd.concat([evaluated_df, unevaluated_fi_df_template(unevaluated_features)])
+        importance_df = merge_importance_dfs(prev_importance_df, importance_df, using_prev_fit_fi)
+
+        # if auto_threshold, threshold is the mean of noise column importance score
+        if auto_threshold:
```
Innixma, line 356: auto_threshold should just be communicated via `prune_threshold='noise'` or something like that. It is strange to pass a prune threshold value just for it to be overwritten.

core/src/autogluon/core/utils/feature_selection.py
```diff
@@ -0,0 +1,537 @@
...
+        candidate_found = False
+        total_prune_time = 0.
+        # mark previous fit's computed feature importances here.
+        if prev_importance_df is not None:
+            fn_args['prev_importance_df'] = prev_importance_df
+            fn_args['using_prev_fit_fi'] = set(prev_importance_df[prev_importance_df['n'] > 0].index.tolist())
+        while self.time_limit > round_time_budget + self.model_fit_time:
+            candidate_features, importance_df, prune_time = self.compute_next_candidate_round(**fn_args)
+            # HACK: Line below is needed to get this working with n-repeated bagged models. Related to feature ordering.
+            candidate_features = [feature for feature in fn_args['X'].columns.tolist() if feature in candidate_features]
```
Innixma, line 304: Why exactly? Is this a bug in AG?

core/src/autogluon/core/utils/feature_selection.py
```diff
@@ -0,0 +1,537 @@
...
+                logger.log(30, f"\tNo time to perform even a single pruning round (remaining: {self.time_limit}, needed: {self.model_fit_time}).")
+                raise TimeLimitExceeded
+
+            # fit self.base_model model once on the subsampled dataset to serve as scoring reference if using subsamples or added a noise column
+            if subsampled or auto_threshold or (self.is_bagged and self.original_model._child_oof):
+                best_info['model'], best_info['score'], _ = self.fit_score_model(X, y, X_val, y_val, candidate_features, f"{self.base_model.name}_1", **kwargs)
+            self._debug_info['index_trajectory'].append(True)
+
+            time_budget_fi = self.compute_time_budget_fi(X_fi=X_fi, n_subsample=n_fi_subsample, **kwargs)
+            logger.log(30, f"\tExpected model fit time: {round(self.model_fit_time, 4)}s, and expected candidate generation time: {round(time_budget_fi, 2)}s.")
```
Innixma, line 210: Why 30? That is warning log level. At most this should be 20. Same for several other logs.

core/src/autogluon/core/utils/feature_selection.py
```diff
@@ -0,0 +1,537 @@
...
+    evaluated_both_rows['p_value'] = None
+    evaluated_both_rows['n'] = n_old + n_new
+    # remove features evaluated in df_new from using_prev_fit_fi if they exist
+    using_prev_fit_fi.difference_update(evaluated_new_rows.index.tolist())
+    result = pd.concat([evaluated_both_rows, evaluated_new_only_rows, evaluated_old_only_rows, evaluated_neither_rows]).sort_values('importance')
+    assert len(result) == len(df_new), "Length of the updated DataFrame must be equal to the inputted DataFrame."
+    return result
+
+
+def sort_features_by_priority(features: List[str], prioritized: Set[str], prev_importance_df: pd.DataFrame, using_prev_fit_fi: Set[str]) -> List[str]:
```
Innixma, line 81: prioritized seems like bloat, I don't see why it belongs in this function logic. Just call this logic without the prioritized features present, then prepend the prioritized outside of the function. Otherwise its needlessly complicated.

core/src/autogluon/core/utils/feature_selection.py
```diff
@@ -0,0 +1,537 @@
...
+    # for features with info on both df_new and whose df_old feature importance came from the current model, return combined statistics
+    evaluated_both_rows = pd.DataFrame()
+    evaluated_both_rows_new = evaluated_new_rows.loc[evaluated_both].sort_index()
+    evaluated_both_rows_old = evaluated_old_rows.loc[evaluated_both].sort_index()
+    mean_old, mean_new = evaluated_both_rows_old['importance'], evaluated_both_rows_new['importance']
+    stddev_old, stddev_new = evaluated_both_rows_old['stddev'], evaluated_both_rows_new['stddev']
+    n_old, n_new = evaluated_both_rows_old['n'], evaluated_both_rows_new['n']
+    evaluated_both_rows['importance'] = (n_old * mean_old + n_new * mean_new) / (n_old + n_new)
+    evaluated_both_rows['stddev'] = (((n_old - 1) * stddev_old ** 2 + (n_new - 1) * stddev_new ** 2) / (n_old + n_new - 1) +
+                                     (n_old * n_new * (mean_old - mean_new) ** 2) / ((n_old + n_new) * (n_old + n_new - 1))) ** 0.5
```
Innixma, line 71: Probably worth adding a comment here walking through the logic so its easier to understand for those not intimately familiar with the statistical method.

Innixma: This may even be worth making into a function and adding a unit test, since its very easy to sneak an error here.

core/src/autogluon/core/utils/feature_selection.py
```diff
@@ -0,0 +1,537 @@
...
+
+    def compute_expected_fi_time_single(self, X_fi: pd.DataFrame, model_predict_time: float, n_subsample: int, n_total_sample: int) -> float:
+        """
+        Return the expected time to compute permutation feature importance for a single feature on n_total_sample datapoints.
+        """
+        n_subsample = min(n_subsample, len(X_fi))
+        n_shuffle = min(np.ceil(n_total_sample / n_subsample).astype(int), self.max_n_shuffle)
+        return self.safety_time_multiplier * model_predict_time * (n_subsample / len(X_fi)) * n_shuffle
+
+    def compute_time_budget_fi(self, X_fi: pd.DataFrame, n_subsample: int, **kwargs):
```
Innixma, line 421: why use kwargs when you only use two kwarg values and you know what they are and their defaults?

core/src/autogluon/core/utils/feature_selection.py
```diff
@@ -0,0 +1,537 @@
...
+                                                      (prev_importance_df.index.isin(using_prev_fit_fi))].sort_values(by='importance')
+        curr_fit_evaluated_rows = prev_importance_df[~(prev_importance_df['importance'].isna()) &
+                                                     ~(prev_importance_df.index.isin(using_prev_fit_fi))].sort_values(by='importance')
+        features = unevaluated_rows.index.tolist() + prev_fit_evaluated_rows.index.tolist() + curr_fit_evaluated_rows.index.tolist()
+    if auto_threshold:
+        features = list(prioritized) + [feature for feature in features if feature not in prioritized]
+    return features
+
+
+class FeatureSelector:
```
Innixma, line 107: What does this do? What are the parameters? This is missing init documentation.

Innixma (commented):
Finished reviewing feature_selector.py

tabular/src/autogluon/tabular/trainer/abstract_trainer.py
```diff
@@ -2229,9 +2226,37 @@ def get_model_fit_kwargs(self, X: pd.DataFrame, X_val: pd.DataFrame, time_limit:
...
+        proxy_model_class can be a string in MODEL_TYPES (ex. GBM) or an AbstractModel object (ex. LGBModel).
+        """
+        proxy_model = None
+        if isinstance(proxy_model_class, str):
+            assert proxy_model_class in MODEL_TYPES, f"proxy_model_class must be one of {MODEL_TYPES.keys()}"
+            proxy_model_class = MODEL_TYPES[proxy_model_class]
+        assert proxy_model_class not in [GreedyWeightedEnsembleModel, SimpleWeightedEnsembleModel], "WeightedEnsemble models cannot be feature pruning proxy models."
+
+        leaderboard = self.leaderboard()
+        candidate_model_rows = leaderboard[(~leaderboard['model'].str.contains('WeightedEnsemble')) & (~leaderboard['score_val'].isna()) & (leaderboard['stack_level'] == level)]
```
Innixma, line 2250: Don't rely on model name conventions

core/src/autogluon/core/utils/feature_selection.py
truebluejason: It's because if I don't override here the model assumes it has the time limit it had for original fit. I don't think I pass model fit time limit anywhere else in feature selection logic.

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -741,6 +741,7 @@ def fit(self,
             'ag_args_ensemble': ag_args_ensemble,
             'ag_args_fit': ag_args_fit,
             'excluded_model_types': excluded_model_types,
+            'feature_prune_kwargs': kwargs.get('feature_prune_kwargs', None)
```
Innixma, line 744: This is currently undocumented in predictor, add documentation.

core/src/autogluon/core/utils/feature_selection.py
truebluejason: It's because we merge it to a different dictionary (fn_args)

core/src/autogluon/core/utils/feature_selection.py
truebluejason: This is because it only triggers when original model uses child_oof and we disable that for the proxy model.

Innixma: ok, add a comment here otherwise requires deep understanding of AG to understand.

core/src/autogluon/core/utils/feature_selection.py
truebluejason: I wrote it this way so I can potentially write extensions of feature pruning logic (ex. bandit based) and be able to use select_features / compute_next_candidate

core/src/autogluon/core/utils/feature_selection.py
truebluejason: This used to cause an issue for certain models when used with repeated bagging (ex. Catboost) related to feature ordering but I will test if it still is an issue after my changes.

szha:
Job PR-1274-18 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1274/18/index.html

core/src/autogluon/core/utils/feature_selection.py
```diff
@@ -400,8 +391,8 @@ def compute_next_candidate_given_fi(self, importance_df: pd.DataFrame, prune_thr
             return acceptance_candidates
 
         sample_weights = [1/i for i in range(1, len(below_threshold_rows)+1)] if weighted else None
-        while time_budget - (time.time() - time_start) > 0:
-            random_state = self.rng.integers(low=0, high=1e5)
+        for _ in range(50):
```
Innixma, line 394: This seems strange to me. Why is this a for loop (or a loop at all)?

truebluejason: This is here to bypass cases where the sampled removal_candidates have already been pruned in previous fits but did not lead to validation set performance increase. This can happen a lot in low dimensional features when only a few features are below threshold. In those cases, we just sample again for up to 50 times and return the inputted features if no untried removal_candidates were found.

szha:
Job PR-1274-19 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1274/19/index.html

core/src/autogluon/core/utils/feature_selection.py
Innixma: why allow those invalid candidates to be chosen in the first place? Couldn't you just prune the list of features that are allowed to be picked instead of having this for loop?

szha:
Job PR-1274-20 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1274/20/index.html

core/src/autogluon/core/utils/feature_selection.py
truebluejason: That would work for pruning single feature at a time but we are sampling multiple features, which would require us to keep a list whose size is combinatorial w.r.t. the prunable feature count.

szha:
Job PR-1274-21 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1274/21/index.html

Innixma (approved):
Fantastic work! Really excited that this functionality is now in AutoGluon.

PR #1275, opened 2021-08-13
Upgrade numpy, graphviz, dill

Innixma:
*Issue #, if available:*

*Description of changes:*

- Upgrade numpy
- Upgrade graphviz
- Upgrade dill
- Fix logging / warning filters, avoided altering global context

### DIFF
--- a/core/src/autogluon/core/utils/miscs.py
+++ b/core/src/autogluon/core/utils/miscs.py
@@ -20,17 +20,13 @@ def in_ipynb():
         return False
 
 
-class warning_filter(object):
+class warning_filter(warnings.catch_warnings):
     def __enter__(self):
+        super().__enter__()
         warnings.filterwarnings("ignore", category=UserWarning)
         warnings.filterwarnings("ignore", category=FutureWarning)
         warnings.filterwarnings("ignore", category=DeprecationWarning)
         return self
-
-    def __exit__(self, *args):
-        warnings.filterwarnings("default", category=UserWarning)
-        warnings.filterwarnings("default", category=FutureWarning)
-        warnings.filterwarnings("default", category=DeprecationWarning)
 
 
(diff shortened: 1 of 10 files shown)


szha:
Job PR-1275-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1275/1/index.html

core/src/autogluon/core/utils/miscs.py
```diff
@@ -20,18 +20,14 @@ def in_ipynb():
...
-class warning_filter(object):
+class warning_filter(warnings.catch_warnings):
     def __enter__(self):
+        super().__enter__()
         warnings.filterwarnings("ignore", category=UserWarning)
         warnings.filterwarnings("ignore", category=FutureWarning)
         warnings.filterwarnings("ignore", category=DeprecationWarning)
         return self
 
-    def __exit__(self, *args):
```
sxjscience, line 30: Should we still keep the `__exit__`?

Innixma: `__exit__` was altering global context. Therefore, if a different module was suppressing warnings, this code would globally reset the warnings so they are no longer suppressed. This led to MXNet spamming warnings about numpy deprecations after this class was used, because it overrode MXNet's custom warnings.

Now, since this inherits from `catch_warnings`, this reset is done automatically and does not reset global context, instead it resets to the global context, which is what we want.

sxjscience: Thanks for the explanation. It looks good.

sxjscience (approved):

szha:
Job PR-1275-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1275/2/index.html

Issue #1279, opened 2021-08-15
ImagePredictor returns bad model due to time limit


Innixma:
If given a time limit and performing HPO, ImagePredictor can return a very poor model because after HPO is finished and the best model is selected, it is retrained.

During this retraining, there might not be enough time left to finish training the model, so it is stopped early. This model will be very bad because of this.

Example Colab Notebook of the issue: https://colab.research.google.com/drive/1bQiruKftkwN43jTov09R8sjoXGdLdbuj?usp=sharing

For an image dataset given the same configs, TabularPredictor gets 68% accuracy, while ImagePredictor gets 39% accuracy.

Innixma:
Fixed in #1282!

PR #1282, opened 2021-08-17
fix final_fit

yinweisu:
*Issue #, if available:*
https://github.com/awslabs/autogluon/issues/1279

*Description of changes:*
This issue was caused by final_fit not being properly passed to ImagePredictor. ImagePredictor will retrieve the best model so far if final_fit is true. Passed final_fit as True to ImagePredictor's train_fn.

Updating the args won't really update what's being used by the train_fn as the scheduler is actually doing a deepcopy of the kwvars and convert that as the args passed to train_fn. Therefore, an update to the kwvars is needed. I updated the args as well as an indication of what the code is doing. 

This is only a quick fix. We should consider refactor base_task instead later on

### DIFF
--- a/core/src/autogluon/core/task/base/base_task.py
+++ b/core/src/autogluon/core/task/base/base_task.py
@@ -63,6 +63,8 @@ class BaseTask(object):
         args.final_fit = True
         if hasattr(args, 'epochs') and hasattr(args, 'final_fit_epochs'):
             args.epochs = args.final_fit_epochs
+        train_fn.args.update({'final_fit':True})
+        train_fn.kwvars.update({'final_fit':True})
         scheduler_final = create_scheduler(train_fn, search_strategy, scheduler_options)
         results = scheduler_final.run_with_config(best_config)
         total_time = time.time() - start_time


yinweisu:
@Innixma Please review the pr.

szha:
Job PR-1282-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1282/1/index.html

core/src/autogluon/core/task/base/base_task.py
```diff
@@ -63,6 +63,8 @@ def run_fit(cls, train_fn, search_strategy, scheduler_options,
         args.final_fit = True
         if hasattr(args, 'epochs') and hasattr(args, 'final_fit_epochs'):
             args.epochs = args.final_fit_epochs
+        train_fn.args.update({'final_fit':True})
+        train_fn.kwvars.update({'final_fit':True})
         scheduler_final = create_scheduler(train_fn, search_strategy, scheduler_options)
```
Innixma, line 68: Are you sure this actually fixes the problem? scheduler_final wouldn't have searched that space so it would have to train it from scratch still.

yinweisu: I tried locally and it fixes the problem. When providing the final_fit, gluoncv won't actually fit anything. It will reload the best models trained so far instead. I know this is confusing :(.

Innixma (approved):
This fixed a long running issue, great stuff!

PR #1286, opened 2021-08-22
Tabular: Added automatic boolean conversion

Innixma:
*Issue #, if available:*

*Description of changes:*

- Automatically convert features with only 2 unique values to boolean
- Drop boolean features for KNN, speeds up KNN by 100x+ without loss of quality on datasets such as CoverType.

TODO:

- [x] Benchmark

### DIFF
--- a/core/src/autogluon/core/features/infer_types.py
+++ b/core/src/autogluon/core/features/infer_types.py
@@ -138,24 +138,35 @@ def check_if_nlp_feature(X: Series) -> bool:
     return True
 
 
-def get_bool_true_val(series: pd.Series, replace_val=None):
+def get_bool_true_val(series: pd.Series):
     """
     From a pandas series, get the replace_val to convert to boolean when calling:
     series_bool = series == replace_val
 
+    Therefore, any value other than `replace_val` will be set to `False` when converting to boolean.
+
     series must have exactly 2 unique values
+
+    We make the assumption that the value chosen as `True` between the two options is mostly arbitrary, with the exception that np.nan will not be considered `True`.
+    When possible, we try to sort the values so that (0, 1) will choose 1 as True, however this decision should ideally not impact downstream models much.
+    Any new unseen values (including nan) at inference time will be mapped to `False` automatically.
+
+    In this code, 0 and 0.0 (int and float) are treated as the same value. Similarly with any other integer and float (such as 1 and 1.0).
+
     """
-    if replace_val is None:
-        try:
-            uniques = series.unique()
-            uniques.sort()
-        except:
-            uniques = series.unique()
-        replace_val = uniques[1]
-        try:
-            is_nan = np.isnan(replace_val)
-        except:
-            is_nan = False
-        if is_nan:
-            replace_val = uniques[0]
+    # This is a safety net in case the unique types are mixed (such as string and int). In this scenario, an exception is raised and therefore we use the unsorted values.
+    try:
+        uniques = series.unique()
+        # Sort the values to avoid relying on row-order when determining which value is mapped to `True`.
+        uniques.sort()
+    except:
+        uniques = series.unique()
+    replace_val = uniques[1]
+    try:
+        # This is to ensure that we don't map np.nan to `True` in the boolean.
+        is_nan = np.isnan(replace_val)
+    except:
+        is_nan = False
+    if is_nan:
+        replace_val = uniques[0]
     return replace_val
--- a/core/src/autogluon/core/features/infer_types.py
+++ b/core/src/autogluon/core/features/infer_types.py
@@ -136,3 +136,26 @@ def check_if_nlp_feature(X: Series) -> bool:
         return False
 
     return True
+
+
+def get_bool_true_val(series: pd.Series, replace_val=None):
+    """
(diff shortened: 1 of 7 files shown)


szha:
Job PR-1286-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1286/1/index.html

Innixma:
Benchmark results show improvement (errors are server-side and can be ignored)

```
                                  framework  > AutoGluon_bestquality_1h_2021_08_22_bool  < AutoGluon_bestquality_1h_2021_08_22_bool  = AutoGluon_bestquality_1h_2021_08_22_bool  % Less Avg. Errors Than AutoGluon_bestquality_1h_2021_08_22_bool  time_train_s  metric_error  time_infer_s  loss_rescaled  time_train_s_rescaled  time_infer_s_rescaled      rank  rank=1_count  rank=2_count  rank=3_count  rank>3_count  error_count
0  AutoGluon_bestquality_1h_2021_08_22_bool                                           0                                           0                                        1032                                           0.000000                  3379.353981  61601.164201     40.616505       0.477670               1.014899               1.165075  1.492233           497           535             0             0            8
1       AutoGluon_bestquality_1h_2021_08_14                                         479                                         495                                          56                                          -0.750196                  3382.340874  62252.525237     39.666117       0.486408               1.016602               1.112381  1.507767           487           551             0             0            2

```

core/src/autogluon/core/features/infer_types.py
```diff
@@ -136,3 +136,26 @@ def check_if_nlp_feature(X: Series) -> bool:
...
+    series_bool = series == replace_val
+
+    series must have exactly 2 unique values
+    """
+    if replace_val is None:
+        try:
+            uniques = series.unique()
+            uniques.sort()
+        except:
+            uniques = series.unique()
```
gradientsky, line 153: Effectively this:
```python
        uniques = series.unique()
        try:
            uniques.sort()
        except:
            pass # log something?
```

core/src/autogluon/core/features/infer_types.py
```diff
@@ -136,3 +136,26 @@ def check_if_nlp_feature(X: Series) -> bool:
...
+    """
+    From a pandas series, get the replace_val to convert to boolean when calling:
+    series_bool = series == replace_val
+
+    series must have exactly 2 unique values
+    """
+    if replace_val is None:
+        try:
+            uniques = series.unique()
+            uniques.sort()
```
gradientsky, line 151: maybe add a comment that this is a check for mixed types present in series

core/src/autogluon/core/features/infer_types.py
```diff
@@ -136,3 +136,26 @@ def check_if_nlp_feature(X: Series) -> bool:
...
+def get_bool_true_val(series: pd.Series, replace_val=None):
+    """
+    From a pandas series, get the replace_val to convert to boolean when calling:
+    series_bool = series == replace_val
+
+    series must have exactly 2 unique values
+    """
+    if replace_val is None:
+        try:
+            uniques = series.unique()
```
gradientsky, line 150: Design decision: so we want to treat `0` and `0.0` values (`int` and `float` types) as the same value?
```python
pd.DataFrame({'a': [0, 0.0]}).a.unique()  # array([0.])
```

Another point, this will auto-cast numeric outputs and `sort()` call won't detect mixed types for numeric types:
```python
pd.DataFrame({'a': [0, 0.0, 1]}).a.unique()  # array([0., 1.]): 1 -> 1.
```
Assuming 0 and 0.0 are treated the same, I would recommend commenting this assumption here

gradientsky: Do we handle common conventions anywhere what's considered as `true`? With the current logic series might have confusing interpretations (I understand it's not really important for prediction quality):
```python
['true', 'false']  # True == ' true'
['true', 'zzzz']   # True == ' zzz'
```

core/src/autogluon/core/features/infer_types.py
Innixma: Yes, I did it how it is currently because .sort() is an inplace operation and I was unsure if calling it (and then erroring) would lead to an altered uniques object.

core/src/autogluon/core/features/infer_types.py
Innixma: added comments, it is for more than checking for mixed types. `.sort()` updates in-place, which allows consistency in which value is chosen as `True`.

core/src/autogluon/core/features/infer_types.py
Innixma: added comments on this. Yes we want to treat 0 and 0.0 as the same.

For now I'd prefer avoiding hard-coded 'true' values for simplicity. If we want to later on, we can borrow from our code that determines positive class when label is binary.

core/src/autogluon/core/features/infer_types.py
Innixma: added comments

gradientsky (approved):

szha:
Job PR-1286-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1286/2/index.html

Issue #1357, opened 2021-10-28
IndexError on Tabular Predictor fit() with hyperparameter tuning and num_bag_folds


lebatchelor:
IndexError: list index out of range is thrown when running Tabular Predictor fit() method with below config:

tabular_predictor = TabularPredictor(
            label=target_variable,
            problem_type='regression',
            eval_metric='mean_absolute_error',
            path=model_output_path,
        ).fit(
            train_data=model_training_dat,
            tuning_data=model_tuning_dat,
            feature_metadata=self.feature_metadata,
            hyperparameters={'CAT': {}, 'XGB': {}, 'GBM': [{}, {'extra_trees': True, 'ag_args': {'name_suffix': 'XT'}}]},
            hyperparameter_tune_kwargs='auto',
            presets='medium_quality_faster_train',
            use_bag_holdout=True,
            num_bag_folds=2,
            num_bag_sets=2,
            keep_only_best=True,
            save_space=True
        )

We have tested it the same but with num_bag_folds removed and it works.
It also works when using excluded_model_types: ['NN', 'FASTAI', 'KNN', 'RF', 'XT'] and removing hyperparameters and hyperparameter_tune_kwargs. 
We have tested on different data also and same behaviour is shown. 

Version 0.3.1
MacOS Big Sur and Ubuntu 18.04

Beginning AutoGluon training ...
AutoGluon will save models to "agModels/cpc_regression_model/20211028_111324/"
AutoGluon Version:  0.3.1
Train Data Rows:    591757
Train Data Columns: 111
Tuning Data Rows:    8098
Tuning Data Columns: 111
Preprocessing data ...
Using Feature Generators to preprocess the data ...
Fitting AutoMLPipelineFeatureGenerator...
	Available Memory:                    6930.83 MB
	Train Data (Original)  Memory Usage: 555.91 MB (8.0% of available memory)
	Warning: Data size prior to feature transformation consumes 8.0% of available memory. Consider increasing memory or subsampling the data to avoid instability.
	Stage 1 Generators:
		Fitting AsTypeFeatureGenerator...
			WARNING: Actual dtype differs from dtype in FeatureMetadata for feature "domain". Actual dtype: int | Expected dtype: category
			WARNING: Actual dtype differs from dtype in FeatureMetadata for feature "day_of_week". Actual dtype: object | Expected dtype: category
[...]

PR #1365, opened 2021-11-03
Tabular: Fixed index error bug in bagging logic

Innixma:
*Issue #, if available:*
#1357

*Description of changes:*

- A bug occurs if doing HPO without time limit, with bagging enabled, and with `num_bag_sets>1` which causes an index error during bagging due to an incorrectly valued `fold_end_n_repeat` value.
- This fix corrects the calculation of `fold_end_n_repeat` and updates the `cv_splitter` if it doesn't have enough `n_repeats` during its init to have enough `kfolds` output.
- Previously, when doing HPO the first fold of the first repeat is trained. Then later the remaining folds and repeats are trained after HPO has finished. Because n_repeats=1 during HPO, the cv_splitter doesn't have information on the kfold splits for n_repeats>1 when performing the remaining folds + repeats in the 2nd part, causing a crash. Additionally, the incorrect calculation of `fold_end_n_repeat` when `k_fold_start != 0` meant that an incorrect number of folds would be trained at each repeat (although this never happened because the previous index error would occur first).
- This bug can only occur in a very specific configuration, which is why we hadn't found it until now.

### DIFF
--- a/core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
+++ b/core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
@@ -363,23 +363,27 @@ class BaggedEnsembleModel(AbstractModel):
         # noinspection PyCallingNonCallable
         fold_fitting_strategy: AbstractFoldFittingStrategy = fold_fitting_strategy(
             self, X, y, sample_weight, time_limit, time_start, models, oof_pred_proba, oof_pred_model_repeats, save_folds=save_folds)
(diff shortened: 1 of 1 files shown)


gradientsky (approved):

szha:
Job PR-1365-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1365/1/index.html

yinweisu (approved):
LGTM!

PR #1369, opened 2021-11-07
Tabular: Cache AG version

Innixma:
*Issue #, if available:*

*Description of changes:*

- Added caching of AG version during save of predictor.
- Adding comparison of cached AG version during load of predictor prior to loading the actual predictor itself.
- This avoids errors when there is a version mismatch that could occur before the actual warning of version incompatibility was triggered, leading to a lot of confusion from users.
- Added error-by-default when versions are incompatible. This should reduce the chances of users missing the warning message.
- Added Python Version and Operating System to the Learner logs at start of fit.

### DIFF
--- a/core/src/autogluon/core/utils/loaders/load_str.py
+++ b/core/src/autogluon/core/utils/loaders/load_str.py
@@ -1,0 +1,45 @@
+import logging
+
+from ...utils import s3_utils
+
+logger = logging.getLogger(__name__)
+
+
+def load(path: str) -> str:
+    """
+    Loads the `data` value from a file saved via `savers.save_str.save(path=path, data=data)`.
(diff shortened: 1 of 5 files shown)


szha:
Job PR-1369-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1369/1/index.html

yinweisu (approved):
LGTM! We may want to add similar check in TextPredictor as well.

Innixma:
> LGTM! We may want to add similar check in TextPredictor as well.

Definitely, this is easiest achieved by having TextPredictor directly re-use the Predictor code by creating an AbstractPredictor class, which is planned potentially for v0.4.

PR #1397, opened 2021-12-04
Remove outdated code

Innixma:
*Issue #, if available:*

*Description of changes:*

This PR is required to enable the removal of MXNet as a dependency and the switch to using Syne Tune, and to greatly clean-up documentation and examples.

- Removed numerous old files
- Removed autogluon.extra (Unused in all situations except NAS)
- Removed autogluon.mxnet (Unused in all situations except outdated docs)
- Removed RL Scheduler/Searcher as it depended on MXNet
- Removed all NAS / ENAS code
- Removed skopt optional dependency (Does not work with sklearn >= 1.0)
- Removed numerous outdated docs
- Removed numerous outdated examples


core/src/autogluon/core/task/base/base_task.py
```diff
@@ -1,172 +1,26 @@
...
-    if isinstance(scheduler, str):
-        scheduler_cls = schedulers[scheduler.lower()]
-    else:
-        assert callable(scheduler)
-        scheduler_cls = scheduler
-        scheduler_options = copy.copy(scheduler_options)
-    return scheduler_cls(train_fn, **scheduler_options)
-
-
-class BaseTask(object):
```
Innixma, line 41: @yinweisu This code is never used in AG, but it is [used in GluonCV](https://github.com/dmlc/gluon-cv/blob/master/gluoncv/auto/tasks/image_classification.py#L21). We may need to move this code to GluonCV and do a release as otherwise we can't get rid of it in AG. There are a variety of other problems with this code being used in GluonCV, since this also uses the schedulers/searchers which we are planning to get rid of in v0.4 in favor of Syne Tune / Ray Tune.

Innixma: Essentially we need to either

1. refactor GluonCV to not depend on AG (and also refactor AG.vision to avoid calling gluoncv.auto) -> My preferred option
2. AG drops support for GluonCV and we use a different vision backend -> Preferred option if removing AG from GluonCV is too difficult / conflicts with existing GluonCV users
3. refactor GluonCV and AG to work similarly to how they do now but with the updated scheduler / searcher API and continue having GluonCV depend on AG. (**Not ideal**)

szha:
Job PR-1397-4 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1397/4/index.html

gradientsky (approved):

PR #1414, opened 2021-12-13
Added Local Searcher

Innixma:
*Issue #, if available:*

*Description of changes:*

- Added LocalSearcher and LocalRandomSearcher which don't depend on ConfigSpace
- These searchers use no multiprocessing or locks, and therefore can be ran in Python debugger
- Greatly simplified HPO logic, removed usage of ag.args and decorators

TODO:

- [x] Fix TextPredictor HPO
- [x] Fix ImagePredictor HPO
- [x] Fix ObjectDetector HPO
- [x] Add unit test for LocalRandomSearcher

### DIFF
--- a/core/src/autogluon/core/searcher/local_random_searcher.py
+++ b/core/src/autogluon/core/searcher/local_random_searcher.py
@@ -1,0 +1,75 @@
+import logging
+import pickle
+
+import numpy as np
+from scipy.stats import randint
+from sklearn.model_selection import ParameterSampler
+
+from .local_searcher import LocalSearcher
+from ..space import Categorical
+
+__all__ = ['LocalRandomSearcher']
+
+logger = logging.getLogger(__name__)
+
+
+class LocalRandomSearcher(LocalSearcher):
+    """Searcher which randomly samples configurations to try next."""
+    MAX_RETRIES = 100
+
+    def __init__(self, *, first_is_default=True, random_seed=0, **kwargs):
+        super().__init__(**kwargs)
+        self._first_is_default = first_is_default
+        # We use an explicit random_state here, in order to better support checkpoint and resume
+        self.random_state = np.random.RandomState(random_seed)
+        self._params_space = self._get_params_space()
+        self._params_default = self._get_params_default()
+
+    def _get_params_default(self) -> dict:
+        params_default = dict()
+        for key, val in self.config.items():
+            if isinstance(val, Categorical):
+                # FIXME: Don't do this, fix the outer code to not require this
+                d = 0
+            else:
+                d = val.default
+            params_default[key] = d
+        return params_default
+
+    def _get_params_space(self) -> dict:
+        param_space = dict()
+        for key, val in self.config.items():
+            sk = val.convert_to_sklearn()
+            if isinstance(sk, list):
+                sk = randint(0, len(sk))
+            param_space[key] = sk
+        return param_space
+
+    # FIXME: Return actual params instead of encoded params that need to be decoded
+    def _sample_config(self) -> dict:
+        params = list(ParameterSampler(self._params_space, n_iter=1, random_state=self.random_state))[0]
+        for key in params:
+            if isinstance(params[key], np.float64):
+                # Fix error in FastAI, can't handle np.float64
+                params[key] = float(params[key])
+        return params
+
+    def get_config(self, **kwargs) -> dict:
+        """Sample a new configuration at random
+
+        Returns
+        -------
+        A new configuration that is valid.
+        """
+        if self._first_is_default and (not self._results):
+            # Try default config first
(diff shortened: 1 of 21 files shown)


szha:
Job PR-1414-5 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1414/5/index.html

szha:
Job PR-1414-7 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1414/7/index.html

szha:
Job PR-1414-8 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1414/8/index.html

tabular/tests/unittests/scheduler/test_seq_scheduler.py
```diff
@@ -49,58 +49,37 @@ def test_has_enough_time_for_trial__enough_time__avg_time_not_allows_trials_by_f
...
-            ag.space.Int(1, 2),
-            ag.space.Categorical(4, 5),
-        ),
-        g=ag.space.Dict(
-            a=ag.Real(0, 10),
-            obj=ag.space.Categorical('auto', 'gluon'),
-        ),
-        h=ag.space.Categorical('test', ag.space.Categorical('auto', 'gluon')),
-        i=ag.space.Categorical('mxnet', 'pytorch'),
     )
```
gradientsky, line 68: I don't think we should remove nested spaces.

Innixma: We don't support them in LocalRandomSearcher. We can add them back if we have need later.

core/src/autogluon/core/searcher/local_searcher.py
```diff
@@ -0,0 +1,135 @@
...
+        self._results = OrderedDict()
+        if reward_attribute is None:
+            reward_attribute = 'accuracy'
+        self._reward_attribute = reward_attribute
+        self._params_default = self._get_params_default()
+
+    def _get_params_default(self) -> dict:
+        params_default = dict()
+        for key, val in self.search_space.items():
+            if isinstance(val, Space):
```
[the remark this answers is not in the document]
Innixma: This is currently for compatibility with the existing logic in GluonCV which passes all parameters to searcher, not just the spaces. This acts as a way to filter those out. We may change in future to instead error.

core/src/autogluon/core/searcher/local_searcher.py
```diff
@@ -0,0 +1,135 @@
...
+        :param search_space: Configuration space to sample from or search in
+        :param reward_attribute: Reward attribute passed to update.
+            Default: 'accuracy'
+
+        """
+        self.search_space = search_space
+        self._results = OrderedDict()
+        if reward_attribute is None:
+            reward_attribute = 'accuracy'
+        self._reward_attribute = reward_attribute
```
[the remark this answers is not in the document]
Innixma: Currently this is to align with API of the old searchers. I'd probably prefer changing this to be some kind of generic `reward` attribute rather than being tied to a particular metric. But will wait to do this until later on as GluonCV may be hard-coded currently to reference this attribute.

core/src/autogluon/core/searcher/local_random_searcher.py
```diff
@@ -0,0 +1,60 @@
...
+        self.random_state = np.random.RandomState(random_seed)
+        self._params_space = self._get_params_space()
+
+    def _get_params_space(self) -> dict:
+        param_space = dict()
+        for key, val in self.search_space.items():
+            if isinstance(val, Space):
+                sk = val.convert_to_sklearn()
+                param_space[key] = sk
+        return param_space
```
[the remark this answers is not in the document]
Innixma: This is currently for compatibility with the existing logic in GluonCV which passes all parameters to searcher, not just the spaces. This acts as a way to filter those out. We may change in future to instead error.

tabular/tests/unittests/scheduler/test_seq_scheduler.py
gradientsky: This is a reduction of existing functionality and I can see valid cases when nested spaces are needed: different architecture blocks with different sub-spaces in each for NAS-like HPO or even in simpler cases where we have `auto` options.
```python
{ 'epochs' ag.space.Categorical('auto', ag.space.Int(1, 100))' }
```

Innixma: I am not removing nested spaces as a class, I am removing the usage of them in this test because the local random searcher does not support nested spaces. This would have to be custom implemented and is not high priority at present.

gradientsky (approved):

PR #1417, opened 2021-12-14
Ray parallel fix

yinweisu:
*Description of changes:*
cap dask and distributed to fix parallel. re-enable parallel


yinweisu:
@Innixma There was some issue for the previous PR because I was using a branch that pushed the disabling code. This one should be good

core/setup.py
```diff
@@ -29,8 +29,8 @@
     'requests',
     'matplotlib',
     'paramiko>=2.4',
-    'dask>=2021.09.1',
-    'distributed>=2021.09.1',
+    'dask>=2021.09.1,<=2021.11.2',
```
Innixma, line 32: Add comment saying that 2021.12.0 causes ray to error, we will want to avoid capping this forever to this version and check if the newer dask versions will work in future.

Innixma: also mention version of ray and maybe a link to the error we saw

szha:
Job PR-1417-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1417/1/index.html

szha:
Job PR-1417-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1417/2/index.html

Innixma (approved):
LGTM!

PR #1427, opened 2021-12-17
Migrate gluoncv/auto/task to autogluon

yinweisu:
*Description of changes:*
Having auto logic in gluoncv that also depends on autogluon is not a good design choice. It prevents autogluon's to modify non-user exposed code. This migration would make the design cleaner by having the auto logic resides in autogluon, and gluoncv will purely be the model provider and trainer

### DIFF
--- a/vision/src/autogluon/vision/_gluoncv/image_classification.py
+++ b/vision/src/autogluon/vision/_gluoncv/image_classification.py
@@ -1,0 +1,460 @@
+"""Auto pipeline for image classification task"""
+# pylint: disable=bad-whitespace,missing-class-docstring,bare-except
+import time
+import os
+import math
+import copy
+import logging
+import pprint
+import json
+import pickle
+from typing import Union, Tuple
+import uuid
+import shutil
+
+import numpy as np
+import pandas as pd
+import autogluon.core as ag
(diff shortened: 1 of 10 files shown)


vision/src/autogluon/vision/_gluoncv/image_classification.py
```diff
@@ -0,0 +1,453 @@
...
+    timm = None
+try:
+    import torch
+except ImportError:
+    torch = None
+try:
+    import mxnet as mx
+    from gluoncv.model_zoo import get_model_list
+except ImportError:
+    mx = None
```
Innixma, line 47: This logic is a bit concerning if we have to load mxnet and torch upfront rather than lazily.

Don't need to address in this PR, but something to think about for later refactoring.

Innixma (approved):
Assuming CI passes, LGTM!

yinweisu:
GluonCV auto deprecation PR: https://github.com/dmlc/gluon-cv/pull/1718

szha:
Job PR-1427-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1427/1/index.html

PR #1439, opened 2021-12-20
[bigfix] removing num_threads setting for lightgbm due to segfaults

gradientsky:
*Description of changes:*
removing num_threads setting for lightgbm due to segfaults


yinweisu:
Is this only on Mac? Removing this will make each parallel bagging uses all cores. Given we disabled parallel bagging, can we only remove this for Mac?

PR #1470, opened 2022-01-04
Fix label cleaner test failure on Windows

yinweisu:
*Description of changes:*
* Try to recover original dtype when inverse_transform
* Assign smaller dtype when possible when doing transform to save memory usage

The fix is now able to pass label cleaner test on windows machine.


szha:
Job PR-1470-5 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1470/5/index.html

Innixma (approved):
LGTM!

Issue #1458, opened 2021-12-26
Included Model Types


jkennedyvz:
Hi, 

I would like a feature to specify included_model_types on .fit(). 

I am currently using a very long excluded_model_types list like:
`excluded_model_types = ['KNN', 'NN', 'XT', 'RF', 'FASTAI']`

willsmithorg:
Try this
```
from autogluon.tabular.configs.hyperparameter_configs import get_hyperparameter_config

custom_hyperparameters = get_hyperparameter_config('default')
```
Now delete observe that every model type has a key in the dict.  You can remove those you don't want, or create a simpler hyperparameter dict containing only the models you want.

Then do:
```
from autogluon.tabular import TabularPredictor
predictor = TabularPredictor(label=label).fit(train_data, hyperparameters=custom_hyperparameters)
```

willsmithorg:
Note for developers : users seem to want an included_model_types parameter to make this functionality more intuitive.

yinweisu:
This feature has been added in the mainline and will be available after 0.8 release. An example usage would be 
`predictor = TabularPredictor(label=label).fit(train_data, included_model_types=["KNN"])`

jkennedyvz:
TYSM @yinweisu and team!

PR #1491, opened 2022-01-15
TabularPredictor config helper

gradientsky:
*Description of changes:*
Added helper methods to enable AG config building with fluent python APIs. The primary benefits are intellisense assistance in IDE with granular help docs on attribute level.

Additional methods added to facilitate the request https://github.com/awslabs/autogluon/issues/1458.

Usage:
```python
config = (ConfigBuilder()
          .presets('best_quality')
          .hyperparameters('light')
          .included_model_types(['GBM', 'KNN'])
          .feature_generator().enable_numeric_features(False).enable_text_ngram_features(False).build()
          .time_limit(60)
          .build())

predictor = TabularPredictor(label=label, path='/Users/ashyrkou/Projects/autogluon/temp/').fit(train_data, **config)

```

### DIFF
--- a/tabular/src/autogluon/tabular/configs/config_helper.py
+++ b/tabular/src/autogluon/tabular/configs/config_helper.py
@@ -174,6 +174,7 @@ class ConfigBuilder:
         Values > 10 may produce diminishing returns, and can even harm overall results due to overfitting.
         To further improve predictions, avoid increasing `num_bag_folds` much beyond 10 and instead increase `num_bag_sets`.
         """
+        assert num_bag_folds >= 0, 'num_bag_folds must be greater or equal than zero'
         self.config['num_bag_folds'] = num_bag_folds
         return self
 
@@ -183,6 +184,7 @@ class ConfigBuilder:
         Defaults to 1 if `time_limit` is not specified, otherwise 20 (always disabled if `num_bag_folds` is not specified).
         Values greater than 1 will result in superior predictive performance, especially on smaller problems and with stacking enabled (reduces overall variance).
         """
+        assert num_bag_sets > 0, 'num_bag_sets must be greater than zero'
         self.config['num_bag_sets'] = num_bag_sets
         return self
 
@@ -192,6 +194,7 @@ class ConfigBuilder:
         Disabled by default (0), but we recommend values between 1-3 to maximize predictive performance.
         To prevent overfitting, `num_bag_folds >= 2` must also be set or else a ValueError will be raised.
         """
+        assert num_stack_levels >= 0, 'num_stack_levels must be greater or equal than zero'
         self.config['num_stack_levels'] = num_stack_levels
         return self
 
@@ -202,6 +205,7 @@ class ConfigBuilder:
         Default value is doubled if `hyperparameter_tune_kwargs` is set, up to a maximum of 0.2.
         Disabled if `num_bag_folds >= 2` unless `use_bag_holdout == True`.
         """
+        assert (holdout_frac >= 0) & (holdout_frac <= 1), 'holdout_frac must be between 0 and 1'
         self.config['holdout_frac'] = holdout_frac
         return self
 
@@ -225,7 +229,13 @@ class ConfigBuilder:
             'random': Performs HPO via random search using local scheduler.
         The 'searcher' key is required when providing a dict.
         """
+        valid_str_values = ['auto', 'random']
+        if isinstance(hyperparameter_tune_kwargs, str):
+            assert hyperparameter_tune_kwargs in valid_str_values, f'{hyperparameter_tune_kwargs} string must be one of {valid_str_values}'
+        elif not isinstance(hyperparameter_tune_kwargs, dict):
+            raise ValueError(f'hyperparameter_tune_kwargs must be either str: {valid_str_values} or dict')
         self.config['hyperparameter_tune_kwargs'] = hyperparameter_tune_kwargs
+
         return self
 
     def ag_args(self, ag_args: dict) -> ConfigBuilder:
--- a/tabular/src/autogluon/tabular/configs/config_helper.py
+++ b/tabular/src/autogluon/tabular/configs/config_helper.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 import copy
 from typing import Union
 
-from Cython.Utils import OrderedSet
 from sklearn.base import BaseEstimator
 
 from autogluon.features import AutoMLPipelineFeatureGenerator
@@ -280,7 +279,7 @@ class ConfigBuilder:
             models = [models]
         for model in models:
             assert model in valid_keys, f'{model} is not one of the valid models {valid_keys}'
-        self.config['excluded_model_types'] = list(OrderedSet(models))
+        self.config['excluded_model_types'] = sorted(list(set(models)))
         return self
 
     def included_model_types(self, models: Union[str, list]) -> ConfigBuilder:
--- a/tabular/src/autogluon/tabular/configs/config_helper.py
+++ b/tabular/src/autogluon/tabular/configs/config_helper.py
@@ -102,7 +102,7 @@ class ConfigBuilder:
     def __init__(self):
         self.config = {}
 
-    def presets(self, presets: Union[str, list]) -> ConfigBuilder:
+    def presets(self, presets: Union[str, list, dict]) -> ConfigBuilder:
         """
         List of preset configurations for various arguments in `fit()`. Can significantly impact predictive accuracy, memory-footprint, and inference latency of trained models, and various other properties of the returned `predictor`.
         It is recommended to specify presets and avoid specifying most other `fit()` arguments or model hyperparameters prior to becoming familiar with AutoGluon.
@@ -111,11 +111,12 @@ class ConfigBuilder:
         If there is an overlap in presets keys, the latter presets will override the earlier ones.
         """
         valid_keys = list(tabular_presets_dict.keys())
-        if not isinstance(presets, list):
+        if isinstance(presets, str):
             presets = [presets]
 
-        unknown_keys = [k for k in presets if k not in valid_keys]
-        assert len(unknown_keys) == 0, f'The following preset are not recognized: {unknown_keys} - use one of the valid presets: {valid_keys}'
+        if isinstance(presets, list):
+            unknown_keys = [k for k in presets if k not in valid_keys]
+            assert len(unknown_keys) == 0, f'The following presets are not recognized: {unknown_keys} - use one of the valid presets: {valid_keys}'
 
         self.config['presets'] = presets
         return self
--- a/tabular/src/autogluon/tabular/configs/config_helper.py
+++ b/tabular/src/autogluon/tabular/configs/config_helper.py
@@ -99,6 +99,7 @@ class FeatureGeneratorBuilder:
 
 
 class ConfigBuilder:
+
     def __init__(self):
         self.config = {}
 
--- a/tabular/src/autogluon/tabular/configs/config_helper.py
+++ b/tabular/src/autogluon/tabular/configs/config_helper.py
@@ -108,6 +108,7 @@ class ConfigBuilder:
         It is recommended to specify presets and avoid specifying most other `fit()` arguments or model hyperparameters prior to becoming familiar with AutoGluon.
         Available Presets: ['best_quality', 'high_quality_fast_inference_only_refit', 'good_quality_faster_inference_only_refit', 'medium_quality_faster_train', 'optimize_for_deployment', 'ignore_text']
         It is recommended to only use one `quality` based preset in a given call to `fit()` as they alter many of the same arguments and are not compatible with each-other.
+        If there is an overlap in presets keys, the latter presets will override the earlier ones.
         """
         valid_keys = list(tabular_presets_dict.keys())
         if not isinstance(presets, list):
@@ -150,18 +151,6 @@ class ConfigBuilder:
         Note: This can increase training time (and inference time) by up to 20x, but can greatly improve predictive performance.
         """
         self.config['auto_stack'] = auto_stack
-        return self
-
-    def use_bag_holdout(self, use_bag_holdout: bool = True) -> ConfigBuilder:
-        """
-        Number of folds used for bagging of models. When `num_bag_folds = k`, training time is roughly increased by a factor of `k` (set = 0 to disable bagging).
-        Disabled by default (0), but we recommend values between 5-10 to maximize predictive performance.
-        Increasing num_bag_folds will result in models with lower bias but that are more prone to overfitting.
-        `num_bag_folds = 1` is an invalid value, and will raise a ValueError.
-        Values > 10 may produce diminishing returns, and can even harm overall results due to overfitting.
-        To further improve predictions, avoid increasing `num_bag_folds` much beyond 10 and instead increase `num_bag_sets`.
-        """
-        self.config['use_bag_holdout'] = use_bag_holdout
         return self
 
     def num_bag_folds(self, num_bag_folds: int) -> ConfigBuilder:
@@ -286,8 +275,8 @@ class ConfigBuilder:
         """
         Subset of model types to train during `fit()`.
         Reference `hyperparameters` documentation for what models correspond to each value.
-        Useful when a particular model type such as 'KNN' or 'custom' is not desired but altering the `hyperparameters` dictionary is difficult or time-consuming.
-            Example: To exclude both 'KNN' and 'custom' models, specify `excluded_model_types=['KNN', 'custom']`.
+        Useful when only the particular models should be trained such as 'KNN' or 'custom', but altering the `hyperparameters` dictionary is difficult or time-consuming.
+            Example: To keep only 'KNN' and 'custom' models, specify `included_model_types=['KNN', 'custom']`.
         """
         valid_keys = [m for m in MODEL_TYPES.keys() if m not in ['ENS_WEIGHTED', 'SIMPLE_ENS_WEIGHTED']]
         if not isinstance(models, list):
--- a/tabular/src/autogluon/tabular/configs/config_helper.py
+++ b/tabular/src/autogluon/tabular/configs/config_helper.py
@@ -136,7 +136,7 @@ class ConfigBuilder:
         if isinstance(hyperparameters, str):
             assert hyperparameters in hyperparameter_config_dict, f'{hyperparameters} is not one of the valid presets {valid_str_values}'
         elif isinstance(hyperparameters, dict):
-            unknown_keys = [k for k in hyperparameters.keys() if k not in valid_keys]
+            unknown_keys = [k for k in hyperparameters.keys() if isinstance(k, str) and (k not in valid_keys)]
             assert len(unknown_keys) == 0, f'The following model types are not recognized: {unknown_keys} - use one of the valid models: {valid_keys}'
         else:
             raise ValueError(f'hyperparameters must be either str: {valid_str_values} or dict with keys of {valid_keys}')
@@ -283,7 +283,7 @@ class ConfigBuilder:
         if not isinstance(models, list):
             models = [models]
 
-        unknown_keys = [k for k in models if k not in valid_keys]
+        unknown_keys = [k for k in models if isinstance(k, str) and (k not in valid_keys)]
         assert len(unknown_keys) == 0, f'The following model types are not recognized: {unknown_keys} - use one of the valid models: {valid_keys}'
 
         models = [m for m in valid_keys if m not in models]
--- a/tabular/src/autogluon/tabular/configs/config_helper.py
+++ b/tabular/src/autogluon/tabular/configs/config_helper.py
@@ -4,17 +4,112 @@ import copy
 from typing import Union
 
 from Cython.Utils import OrderedSet
+from sklearn.base import BaseEstimator
 
+from autogluon.features import AutoMLPipelineFeatureGenerator
 from autogluon.tabular.configs.hyperparameter_configs import hyperparameter_config_dict
 from autogluon.tabular.configs.presets_configs import tabular_presets_dict
 from autogluon.tabular.trainer.model_presets.presets import MODEL_TYPES
+
+
+class FeatureGeneratorBuilder:
+    def __init__(self, parent):
+        self.parent = parent
+        self.config = {}
+
+    def enable_numeric_features(self, value: bool = True) -> FeatureGeneratorBuilder:
+        """
+        Whether to keep features of 'int' and 'float' raw types.
+        These features are passed without alteration to the models.
+        Appends IdentityFeatureGenerator(infer_features_in_args=dict(valid_raw_types=['int', 'float']))) to the generator group.
+        """
+        self.config['enable_numeric_features'] = value
+        return self
+
+    def enable_categorical_features(self, value: bool = True) -> FeatureGeneratorBuilder:
+        """
+        Whether to keep features of 'object' and 'category' raw types.
+        These features are processed into memory optimized 'category' features.
+        Appends CategoryFeatureGenerator() to the generator group.
+        """
+        self.config['enable_categorical_features'] = value
+        return self
+
+    def enable_datetime_features(self, value: bool = True) -> FeatureGeneratorBuilder:
+        """
+        Whether to keep features of 'datetime' raw type and 'object' features identified as 'datetime_as_object' features.
+        These features will be converted to 'int' features representing milliseconds since epoch.
+        Appends DatetimeFeatureGenerator() to the generator group.
+        """
+        self.config['enable_datetime_features'] = value
+        return self
+
+    def enable_text_special_features(self, value: bool = True) -> FeatureGeneratorBuilder:
+        """
+        Whether to use 'object' features identified as 'text' features to generate 'text_special' features such as word count, capital letter ratio, and symbol counts.
+        Appends TextSpecialFeatureGenerator() to the generator group.
+        """
+        self.config['enable_text_special_features'] = value
+        return self
+
+    def enable_text_ngram_features(self, value: bool = True) -> FeatureGeneratorBuilder:
+        """
+        Whether to use 'object' features identified as 'text' features to generate 'text_ngram' features.
+        Appends TextNgramFeatureGenerator(vectorizer=vectorizer) to the generator group.
+        """
+        self.config['enable_text_ngram_features'] = value
+        return self
+
+    def enable_raw_text_features(self, value: bool = True) -> FeatureGeneratorBuilder:
+        """
+        Whether to keep the raw text features.
+        Appends IdentityFeatureGenerator(infer_features_in_args=dict(required_special_types=['text'])) to the generator group.
+        """
+        self.config['enable_raw_text_features'] = value
+        return self
+
+    def enable_vision_features(self, value: bool = True) -> FeatureGeneratorBuilder:
+        """
+        [Experimental]
+        Whether to keep 'object' features identified as 'image_path' special type. Features of this form should have a string path to an image file as their value.
+        Only vision models can leverage these features, and these features will not be treated as categorical.
+        Note: 'image_path' features will not be automatically inferred. These features must be explicitly specified as such in a custom FeatureMetadata object.
+        Note: It is recommended that the string paths use absolute paths rather than relative, as it will likely be more stable.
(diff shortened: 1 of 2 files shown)


szha:
Job PR-1491-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1491/3/index.html

tabular/src/autogluon/tabular/configs/config_helper.py
```diff
@@ -0,0 +1,368 @@
...
+        Reference `hyperparameters` documentation for what models correspond to each value.
+        Useful when a particular model type such as 'KNN' or 'custom' is not desired but altering the `hyperparameters` dictionary is difficult or time-consuming.
+            Example: To exclude both 'KNN' and 'custom' models, specify `excluded_model_types=['KNN', 'custom']`.
+        """
+        valid_keys = [m for m in MODEL_TYPES.keys() if m not in ['ENS_WEIGHTED', 'SIMPLE_ENS_WEIGHTED']]
+        if not isinstance(models, list):
+            models = [models]
+        for model in models:
+            assert model in valid_keys, f'{model} is not one of the valid models {valid_keys}'
+        self.config['excluded_model_types'] = sorted(list(set(models)))
```
yinweisu, line 282: why sort it?

tabular/src/autogluon/tabular/configs/config_helper.py
```diff
@@ -0,0 +1,368 @@
...
+        return self
+
+    def included_model_types(self, models: Union[str, list]) -> ConfigBuilder:
+        """
+        Subset of model types to train during `fit()`.
+        Reference `hyperparameters` documentation for what models correspond to each value.
+        Useful when a particular model type such as 'KNN' or 'custom' is not desired but altering the `hyperparameters` dictionary is difficult or time-consuming.
+            Example: To exclude both 'KNN' and 'custom' models, specify `excluded_model_types=['KNN', 'custom']`.
```
yinweisu, line 290: This is not for `included_model_types`

yinweisu:
A general comment: we might want to add a section in the doc for this, but we can do it in a separate pr

tabular/src/autogluon/tabular/configs/config_helper.py
```diff
@@ -0,0 +1,368 @@
...
+        return self
+
+    def use_bag_holdout(self, use_bag_holdout: bool = True) -> ConfigBuilder:
+        """
+        Number of folds used for bagging of models. When `num_bag_folds = k`, training time is roughly increased by a factor of `k` (set = 0 to disable bagging).
+        Disabled by default (0), but we recommend values between 5-10 to maximize predictive performance.
+        Increasing num_bag_folds will result in models with lower bias but that are more prone to overfitting.
+        `num_bag_folds = 1` is an invalid value, and will raise a ValueError.
+        Values > 10 may produce diminishing returns, and can even harm overall results due to overfitting.
+        To further improve predictions, avoid increasing `num_bag_folds` much beyond 10 and instead increase `num_bag_sets`.
```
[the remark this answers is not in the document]
gradientsky: good catch, it was actually duplicate method, removed

tabular/src/autogluon/tabular/configs/config_helper.py
```diff
@@ -0,0 +1,368 @@
...
+class ConfigBuilder:
+    def __init__(self):
+        self.config = {}
+
+    def presets(self, presets: Union[str, list]) -> ConfigBuilder:
+        """
+        List of preset configurations for various arguments in `fit()`. Can significantly impact predictive accuracy, memory-footprint, and inference latency of trained models, and various other properties of the returned `predictor`.
+        It is recommended to specify presets and avoid specifying most other `fit()` arguments or model hyperparameters prior to becoming familiar with AutoGluon.
+        Available Presets: ['best_quality', 'high_quality_fast_inference_only_refit', 'good_quality_faster_inference_only_refit', 'medium_quality_faster_train', 'optimize_for_deployment', 'ignore_text']
+        It is recommended to only use one `quality` based preset in a given call to `fit()` as they alter many of the same arguments and are not compatible with each-other.
```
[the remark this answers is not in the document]
gradientsky: Added

tabular/src/autogluon/tabular/configs/config_helper.py
```diff
@@ -0,0 +1,368 @@
...
+        return self
+
+    def use_bag_holdout(self, use_bag_holdout: bool = True) -> ConfigBuilder:
+        """
+        If True, a `holdout_frac` portion of the data is held-out from model bagging.
+        This held-out data is only used to score models and determine weighted ensemble weights.
+        Enable this if there is a large gap between score_val and score_test in stack models.
+        Note: If `tuning_data` was specified, `tuning_data` is used as the holdout data.
+        Disabled if not bagging.
+        """
```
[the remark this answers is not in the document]
gradientsky: fixed

tabular/src/autogluon/tabular/configs/config_helper.py
```diff
@@ -0,0 +1,368 @@
...
+    def hyperparameter_tune_kwargs(self, hyperparameter_tune_kwargs: Union[str, dict]) -> ConfigBuilder:
+        """
+        Hyperparameter tuning strategy and kwargs (for example, how many HPO trials to run).
+        If None, then hyperparameter tuning will not be performed.
+        Valid preset values:
+            'auto': Uses the 'random' preset.
+            'random': Performs HPO via random search using local scheduler.
+        The 'searcher' key is required when providing a dict.
+        """
+        valid_str_values = ['auto', 'random']
```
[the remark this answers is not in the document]
gradientsky: The predictor doesn't have a generic list to validate against, like the other enumerations; the checks are done via comparison vs strings.

tabular/src/autogluon/tabular/configs/config_helper.py
gradientsky: This is for the test stability. `set` order is non-deterministic and `['LR', 'NN']` should not be treated differently from `['NN', 'LR']` in this case.

tabular/src/autogluon/tabular/configs/config_helper.py
gradientsky: fixed

yinweisu (approved):
LGTM!

szha:
Job PR-1491-4 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1491/4/index.html

tabular/src/autogluon/tabular/configs/config_helper.py
```diff
@@ -0,0 +1,357 @@
...
+
+from autogluon.features import AutoMLPipelineFeatureGenerator
+from autogluon.tabular.configs.hyperparameter_configs import hyperparameter_config_dict
+from autogluon.tabular.configs.presets_configs import tabular_presets_dict
+from autogluon.tabular.trainer.model_presets.presets import MODEL_TYPES
+
+
+class FeatureGeneratorBuilder:
+    def __init__(self, parent):
+        self.parent = parent
```
Innixma, line 16: Why does this need a parent? A FeatureGenerator is useful outside of a Predictor. What if I want to just build a FeatureGenerator without a Predictor?

tabular/src/autogluon/tabular/configs/config_helper.py
```diff
@@ -0,0 +1,357 @@
...
+        Available Presets: ['best_quality', 'high_quality_fast_inference_only_refit', 'good_quality_faster_inference_only_refit', 'medium_quality_faster_train', 'optimize_for_deployment', 'ignore_text']
+        It is recommended to only use one `quality` based preset in a given call to `fit()` as they alter many of the same arguments and are not compatible with each-other.
+        If there is an overlap in presets keys, the latter presets will override the earlier ones.
+        """
+        valid_keys = list(tabular_presets_dict.keys())
+        if not isinstance(presets, list):
+            presets = [presets]
+
+        unknown_keys = [k for k in presets if k not in valid_keys]
+        assert len(unknown_keys) == 0, f'The following preset are not recognized: {unknown_keys} - use one of the valid presets: {valid_keys}'
```
Innixma, line 118: The following preset are not -> The following presets are not

tabular/src/autogluon/tabular/configs/config_helper.py
```diff
@@ -0,0 +1,357 @@
...
+        It is recommended to specify presets and avoid specifying most other `fit()` arguments or model hyperparameters prior to becoming familiar with AutoGluon.
+        Available Presets: ['best_quality', 'high_quality_fast_inference_only_refit', 'good_quality_faster_inference_only_refit', 'medium_quality_faster_train', 'optimize_for_deployment', 'ignore_text']
+        It is recommended to only use one `quality` based preset in a given call to `fit()` as they alter many of the same arguments and are not compatible with each-other.
+        If there is an overlap in presets keys, the latter presets will override the earlier ones.
+        """
+        valid_keys = list(tabular_presets_dict.keys())
+        if not isinstance(presets, list):
+            presets = [presets]
+
+        unknown_keys = [k for k in presets if k not in valid_keys]
```
Innixma, line 117: technically this isn't identical to what TabularPredictor offers as I believe I coded it so users could specify their own presets as dictionaries of config options. Its probably not a big deal that this behaves differently, but just thought I'd let you know.

tabular/src/autogluon/tabular/configs/config_helper.py
```diff
@@ -0,0 +1,357 @@
...
+        """
+        self.config['text_ngram_params'] = value
+        return self
+
+    def build(self) -> ConfigBuilder:
+        self.parent.config['feature_generator'] = AutoMLPipelineFeatureGenerator(**self.config)
+        return self.parent
+
+
+class ConfigBuilder:
```
Innixma, line 101: re 'intellisense assistance', doesn't this already occur when calling `.fit` in an IDE? I suppose the benefit is that we can split up the documentation into manageable chunks?

tabular/src/autogluon/tabular/configs/config_helper.py
```diff
@@ -0,0 +1,357 @@
...
+
+        self.config['presets'] = presets
+        return self
+
+    def time_limit(self, time_limit: int) -> ConfigBuilder:
+        """
+        Approximately how long `fit()` should run for (wallclock time in seconds).
+        If not specified, `fit()` will run until all models have completed training, but will not repeatedly bag models unless `num_bag_sets` is specified.
+        """
+        assert time_limit > 0, 'time_limit must be greater than zero'
```
Innixma, line 128: What if they specify `time_limit=None`?

tabular/src/autogluon/tabular/configs/config_helper.py
```diff
@@ -0,0 +1,357 @@
...
+            assert hyperparameters in hyperparameter_config_dict, f'{hyperparameters} is not one of the valid presets {valid_str_values}'
+        elif isinstance(hyperparameters, dict):
+            unknown_keys = [k for k in hyperparameters.keys() if k not in valid_keys]
+            assert len(unknown_keys) == 0, f'The following model types are not recognized: {unknown_keys} - use one of the valid models: {valid_keys}'
+        else:
+            raise ValueError(f'hyperparameters must be either str: {valid_str_values} or dict with keys of {valid_keys}')
+        self.config['hyperparameters'] = hyperparameters
+        return self
+
+    def auto_stack(self, auto_stack=True) -> ConfigBuilder:
```
Innixma, line 145: general comment: Should we do type checking in builder or should we entrust that downstream type checks are made?

tabular/src/autogluon/tabular/configs/config_helper.py
```diff
@@ -0,0 +1,357 @@
...
+        """
+        Number of stacking levels to use in stack ensemble. Roughly increases model training time by factor of `num_stack_levels+1` (set = 0 to disable stack ensembling).
+        Disabled by default (0), but we recommend values between 1-3 to maximize predictive performance.
+        To prevent overfitting, `num_bag_folds >= 2` must also be set or else a ValueError will be raised.
+        """
+        assert num_stack_levels >= 0, 'num_stack_levels must be greater or equal than zero'
+        self.config['num_stack_levels'] = num_stack_levels
+        return self
+
+    def holdout_frac(self, holdout_frac: float) -> ConfigBuilder:
```
Innixma, line 189: General comment: Do we allow `None` (the default). Currently it would error in assert.

tabular/src/autogluon/tabular/configs/config_helper.py
```diff
@@ -0,0 +1,357 @@
...
+    def hyperparameter_tune_kwargs(self, hyperparameter_tune_kwargs: Union[str, dict]) -> ConfigBuilder:
+        """
+        Hyperparameter tuning strategy and kwargs (for example, how many HPO trials to run).
+        If None, then hyperparameter tuning will not be performed.
+        Valid preset values:
+            'auto': Uses the 'random' preset.
+            'random': Performs HPO via random search using local scheduler.
+        The 'searcher' key is required when providing a dict.
+        """
+        valid_str_values = ['auto', 'random']
```
Innixma, line 220: Why does this logic live here when it could live in TabularPredictor.fit()?

Innixma, line 220: 'bayesopt' should be a valid value, it is converted to 'random' downstream with a warning to avoid breaking user code based on prior releases.

tabular/src/autogluon/tabular/configs/config_helper.py
```diff
@@ -0,0 +1,357 @@
...
+        valid_keys = [m for m in MODEL_TYPES.keys() if m not in ['ENS_WEIGHTED', 'SIMPLE_ENS_WEIGHTED']]
+        if not isinstance(models, list):
+            models = [models]
+
+        unknown_keys = [k for k in models if k not in valid_keys]
+        assert len(unknown_keys) == 0, f'The following model types are not recognized: {unknown_keys} - use one of the valid models: {valid_keys}'
+
+        models = [m for m in valid_keys if m not in models]
+        self.config['excluded_model_types'] = models
+        return self
```
Innixma, line 290: Why set excluded_model_types? Does this mean someone cannot specify included_model_types in `.fit` directly?

tabular/src/autogluon/tabular/configs/config_helper.py
```diff
@@ -0,0 +1,357 @@
...
+        Reference `hyperparameters` documentation for what models correspond to each value.
+        Useful when only the particular models should be trained such as 'KNN' or 'custom', but altering the `hyperparameters` dictionary is difficult or time-consuming.
+            Example: To keep only 'KNN' and 'custom' models, specify `included_model_types=['KNN', 'custom']`.
+        """
+        valid_keys = [m for m in MODEL_TYPES.keys() if m not in ['ENS_WEIGHTED', 'SIMPLE_ENS_WEIGHTED']]
+        if not isinstance(models, list):
+            models = [models]
+
+        unknown_keys = [k for k in models if k not in valid_keys]
+        assert len(unknown_keys) == 0, f'The following model types are not recognized: {unknown_keys} - use one of the valid models: {valid_keys}'
```
Innixma, line 286: It is perfectly reasonable for someone to specify their own custom model here, yet it would error in this assert statement.

tabular/src/autogluon/tabular/configs/config_helper.py
Innixma, line 101: [Major Question]

Are there examples of ML packages that use builder patterns? I'd like to at least compare our implementation to something widely used. If no ML packages use builder patterns, then that would also be useful information.

tabular/src/autogluon/tabular/configs/config_helper.py
```diff
@@ -0,0 +1,357 @@
...
+        """
+        If True and the problem_type is classification, temperature scaling will be used to calibrate the Predictor's estimated class probabilities
+        (which may improve metrics like log_loss) and will train a scalar parameter on the validation set.
+        If True and the problem_type is quantile regression, conformalization will be used to calibrate the Predictor's estimated quantiles
+        (which may improve the prediction interval coverage, and bagging could futher improve it) and will compute a set of scalar parameters on the validation set.
+        """
+        self.config['calibrate'] = calibrate
+        return self
+
+    def build(self) -> dict:
```
Innixma, line 353: So users can't specify kwargs not specified in this builder? Ex: unlabeled_data. Is this what we want?

tabular/tests/unittests/autogluon/tabular/configs/test_config_helper.py
```diff
@@ -0,0 +1,284 @@
...
+        ConfigBuilder().excluded_model_types('unknown1').build()
+
+    with pytest.raises(AssertionError, match=r"unknown2 is not one of the valid models .*"):
+        ConfigBuilder().excluded_model_types(['unknown2']).build()
+
+
+def test_included_model_types():
+    expected_config = dict(excluded_model_types=['RF', 'XT', 'KNN', 'GBM', 'CAT', 'XGB', 'QNN', 'LR', 'FASTAI', 'TRANSF', 'AG_TEXT_NN', 'AG_IMAGE_NN', 'FASTTEXT', 'VW'])
+    acutal_config = ConfigBuilder().included_model_types('NN').build()
+    assert acutal_config == expected_config
```
Innixma, line 58: nit: acutal -> actual, ditto for all other occurrences.

Innixma (commented):
Added comments, most urgent comment is marked with [Major Question].

tabular/src/autogluon/tabular/configs/config_helper.py
gradientsky: To return control to the main builder after it is done:
```python
config = (ConfigBuilder()  
          .included_model_types(['GBM', 'KNN'])
          .feature_generator()   # <<<< switching to FeatureGeneratorBuilder
                    .enable_numeric_features(False).enable_text_ngram_features(False)
                    .build()  # <<<< Return control back to ConfigBuilder
          .time_limit(60)
          .build())
```

tabular/src/autogluon/tabular/configs/config_helper.py
gradientsky: fixed

tabular/src/autogluon/tabular/configs/config_helper.py
gradientsky: good catch - fixed, now differentiating `dict` vs `str`/`list`.

tabular/src/autogluon/tabular/configs/config_helper.py
gradientsky: It won't give help on kwargs.
Couple other benefits: 1) parameter-level doc strings (there's a lot of good information in `fit` docstring, but you have to dig it to find specific information on a parameter; 2) can do more complex setups/cross-checks on top of existing parameters. Technically things like presets can live in builders.

tabular/src/autogluon/tabular/configs/config_helper.py
gradientsky: Why adding it then? It's `None` by default.

tabular/src/autogluon/tabular/configs/config_helper.py
gradientsky: Correct: we don't have `included_model_types` in `.fit`. This is convenience method built on top of existing API capabilities.

tabular/src/autogluon/tabular/configs/config_helper.py
gradientsky: I think we can delay this check till `.build` call and then also include models from `hyperparameters` block.

tabular/src/autogluon/tabular/configs/config_helper.py
gradientsky: [Method chaining](https://en.wikipedia.org/wiki/Method_chaining) is heavily used in pandas. Builder pattern is used in `.groupby` - it returns an object methods you can do on the group operation. `pyplot` is effectively a massive builder because the actual chart built on save methods or when the chart has to be displayed; though it's not things in a chaining session, more like a context, which is mutated by different calls. 

Effectively the best time to use builders are: 1) complex object hierarchy creation/management (like pyplot/urllib) or 2) convenience: massive number properties.

tabular/src/autogluon/tabular/configs/config_helper.py
gradientsky: I don't think we should put kwargs here just because they can be easily put in the fit call directly.

```python
config = (ConfigBuilder()
          .presets('best_quality')
          .hyperparameters('light')
          .time_limit(60)
          .build())

predictor = TabularPredictor(label=label, path='/Users/ashyrkou/Projects/autogluon/temp/').fit(train_data, **config, **kwargs)
```

Both regular parameters and builders can be mixed if needed:
```python
config = (ConfigBuilder().presets('best_quality').build())
predictor = TabularPredictor(label=label, path='/Users/ashyrkou/Projects/autogluon/temp/').fit(train_data, **config, time_limit=60)
```

tabular/tests/unittests/autogluon/tabular/configs/test_config_helper.py
gradientsky: fixed

szha:
Job PR-1491-5 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1491/5/index.html

szha:
Job PR-1491-6 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1491/6/index.html

szha:
Job PR-1491-7 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1491/7/index.html

tabular/src/autogluon/tabular/configs/config_helper.py
Innixma: I understand, I mean what if someone doesn't want to make a TabularPredictor config and instead just a FeatureGenerator config?

tabular/src/autogluon/tabular/configs/config_helper.py
Innixma: why force users to use the builder API to get access to `included_model_types`?

tabular/src/autogluon/tabular/configs/config_helper.py
Innixma: I'm a bit concerned on the value add here vs the cost of maintaining it. I don't think ML users are familiar with builder patterns, so I'm struggling to see a large portion of users preferring this API to the general TabularPredictor API. We can certainly give it a try, but unsure how much it will be used.

tabular/src/autogluon/tabular/configs/config_helper.py
Innixma: a preset could specify time_limit=5, only way to override would be by specifying time_limit=None here. Additionally, we might change time_limit to not default to None in future.

tabular/src/autogluon/tabular/configs/config_helper.py
gradientsky: Updated the code to handle this case; parent now is optional: if passed, then it will pass control back; if `None`, then it will return the generator back.
```
FeatureGeneratorBuilder().enable_text_special_features(False).enable_vision_features(False).build()
```

tabular/src/autogluon/tabular/configs/config_helper.py
gradientsky: Good point, added `None` handling.

tabular/src/autogluon/tabular/configs/config_helper.py
gradientsky: Python type hints are just 'suggestions' even if specified. It really helps when API prevent you to make mistakes by doing strict checks as early as possible.

tabular/src/autogluon/tabular/configs/config_helper.py
gradientsky: I use assumption that if you want default, you don't use this call.

tabular/src/autogluon/tabular/configs/config_helper.py
Innixma: I'm not referring to type hints, but to your explicit assert statements in the builder.

tabular/src/autogluon/tabular/configs/config_helper.py
Innixma: ok, I suppose this is a 2-way-door for now if we start with not allowing.

tabular/src/autogluon/tabular/configs/config_helper.py
gradientsky: This is actually can be replaced with checking `scheduler_factory._scheduler_presets.keys()`

gradientsky: it is not: it's explicitly checked if the key is in the map keys: `autogluon.core.scheduler.scheduler_factory.get_hyperparameter_tune_kwargs_preset()`. And here are the valid keys:
```python
_scheduler_presets = {
    'auto': {'scheduler': 'local', 'searcher': 'local_random'},
    'local_random': {'scheduler': 'local', 'searcher': 'local_random'},
    'random': {'scheduler': 'local', 'searcher': 'random'},
}
```

tabular/src/autogluon/tabular/configs/config_helper.py
gradientsky: IMHO neither methods should be actually a part of `.fit` API and should be handled via helpers, otherwise you will be forced changing .fit API every time a new kind of filter is wanted.

tabular/src/autogluon/tabular/configs/config_helper.py
Innixma: Good catch, can you fix in this PR by adding a new preset:

```
'bayesopt': {'scheduler': 'local', 'searcher': 'bayesopt'},
```

tabular/src/autogluon/tabular/configs/config_helper.py
gradientsky: I think we should review what should be part of core basic APIs `.fit` and what parts are syntactic sugar on top of it and should not result into adding new things to core APIs. It's much easier to handle parameters configuration and interaction code when it is separated from ML code.

tabular/src/autogluon/tabular/configs/config_helper.py
Innixma: Good points. I think it is fine as is then.

tabular/src/autogluon/tabular/configs/config_helper.py
Innixma: Definitely worth a discussion on how to simplify APIs. Tough part will be that it will be breaking (we could do a deprecation warning though)

szha:
Job PR-1491-9 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1491/9/index.html

gomlfx:
Could you please fix the plot images at
https://auto.gluon.ai/dev/tutorials/nas/rl_searcher.html?highlight=searcher

[<mpl_toolkits.mplot3d.art3d.Line3D at 0x7fc24a6d3760>,
 <mpl_toolkits.mplot3d.art3d.Line3D at 0x7fc2096867f0>]

or how do I fix it?

Innixma:
@gomlfx 

> Could you please fix the plot images at https://auto.gluon.ai/dev/tutorials/nas/rl_searcher.html?highlight=searcher [<mpl_toolkits.mplot3d.art3d.Line3D at 0x7fc24a6d3760>, <mpl_toolkits.mplot3d.art3d.Line3D at 0x7fc2096867f0>] or how do I fix it?

We are removing RL searcher from AG in v0.4. If you are interested in using RL searcher, consider using [syne-tune](https://github.com/awslabs/syne-tune) which was a package that used to live in AutoGluon but is now standalone.

Innixma (approved):
LGTM!

gomlfx:
Why removing RL searcher? I thought that's leading edge ml?

Innixma:
> Why removing RL searcher? I thought that's leading edge ml?

Because maintaining these advanced searchers is a lot of effort and is better done in a dedicated HPO repository like syne-tune. The process of integrating with syne-tune requires investigation effort which the AutoGluon team does not have bandwidth for in the scope of v0.4 release.

PR #1502, opened 2022-01-20
[WIP] Enable calibrate

Innixma:
*Issue #, if available:*

*Description of changes:*

Enable calibrate by default.

TODO:

- [ ] Benchmark

### DIFF
--- a/tabular/src/autogluon/tabular/predictor/predictor.py
+++ b/tabular/src/autogluon/tabular/predictor/predictor.py
@@ -651,7 +651,7 @@ class TabularPredictor:
                 to any amount of labeled data.
             verbosity : int
                 If specified, overrides the existing `predictor.verbosity` value.
-            calibrate: bool, default = False
+            calibrate: bool, default = True
                 If True and the problem_type is classification, temperature scaling will be used to calibrate the Predictor's estimated class probabilities
                 (which may improve metrics like log_loss) and will train a scalar parameter on the validation set.
                 If True and the problem_type is quantile regression, conformalization will be used to calibrate the Predictor's estimated quantiles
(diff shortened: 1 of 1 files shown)


szha:
Job PR-1502-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1502/3/index.html

Innixma:
@yinweisu @gradientsky ready for review, CI should pass.

core/src/autogluon/core/calibrate/temperature_scaling.py
```diff
@@ -24,14 +24,23 @@ def tune_temperature_scaling(y_val_probs: np.ndarray, y_val: np.ndarray, init_va
...
 
+    # This is required to avoid error when passing np.uint16 to torch.tensor. This can occur if >255 classes (Dionis)
+    y_val = y_val.astype('int')
+
     y_val_tensor = torch.tensor(y_val)
     temperature_param = torch.nn.Parameter(torch.ones(1).fill_(init_val))
     logits = torch.tensor(np.log(y_val_probs))
+
+    # TODO: Could alternatively add epsilon to y_val_probs in order to avoid.
+    is_invalid = torch.isinf(logits).any().tolist()
```
gradientsky, line 40: Torch used to be optional dependency. If this is on by default, then we need to move it into required dependencies. This would have some deployment size implications (torch library is big). Do we need torch here or scikit methods can be used instead?

gradientsky (approved):

core/src/autogluon/core/calibrate/temperature_scaling.py
Innixma: I don't have context to say if torch is necessary for this. It probably isn't. Regardless, this is not actually a required dependency. If torch is not installed then calibration is simply skipped. Refer to line 882.

Issue #1418, opened 2021-12-15
Error on feature_importance


lamres:
Hi.
I use the latest version of the library. I trained the model.
When I try to call the feature_importance method like this I get the result and everything is ok (87527 - number of rows).
`my_model.feature_importance(data=testset, subsample_size=87527, num_shuffle_sets=30)`

When I try to reduce the subsample_size like this 
`my_model.feature_importance(data=testset, subsample_size=10000, num_shuffle_sets=30)`
I get this error:
> ValueError: Found input variables with inconsistent numbers of samples: [222191, 10000]

If subsample_size=1000 this error:
> ValueError: Found input variables with inconsistent numbers of samples: [21537, 1000]

What does this error mean? What do these numbers 21537, 222191 mean?

recycletechno:
Hello! The same with me. I fit a model with 'XGB', 'CAT', 'RF', 'XT', 'FASTAI' as hyperparameters. When try to get feature_importance() I receive the same:

`ValueError: Found input variables with inconsistent numbers of samples: [288144, 10000]`

feature_importance() work only if I set subsample_size=None which means no subsample and I can not wait for the result if data is rather big.

Here is the traceback:

Innixma:
Can you try the following as a sanity check:

```
data = data.reset_index(drop=True)
predictor.feature_importance(data, subsample_size=...)
```

I suspect you have duplicate indices in your data, and if so the above code should fix the problem. This is something I plan to fix for v0.4 but want to ensure this is the problem you are facing.

recycletechno:
Yep, removing dups solve the problem. I have duplicate indicies in my data, but its ok for my pipeline - index may contain duplicate entries

recycletechno:
Also, in almost every run I do not see any printouts from feature_importance() even when I set silent=False. I use autogluon in Google Colab, maybe it is about dependencies or packages installed after autogluon. In feature_importance() its very convenient to see expected runtime but I do not see it. Sounds like 'silent=' param doesnt work as intended. Thank you!

PR #1504, opened 2022-01-20
Validate unique indices for various API

yinweisu:
*Issue #, if available:*
https://github.com/awslabs/autogluon/issues/1418

*Description of changes:*
Raise exception if duplicate index found in the data. Previously this check was done only for `fit`, added this check to other APIs

### DIFF
--- a/tabular/src/autogluon/tabular/predictor/predictor.py
+++ b/tabular/src/autogluon/tabular/predictor/predictor.py
@@ -1237,6 +1237,8 @@ class TabularPredictor:
                        f'Predictor not fit prior to pseudolabeling. Fitting now...')
             self.fit(**kwargs)
 
+        self._validate_unique_indices(pseudo_data, 'pseudo_data')
+
         if self.problem_type is MULTICLASS and self.eval_metric.name != 'accuracy':
             logger.warning('AutoGluon has detected the problem type as \'multiclass\' and '
                            f'eval_metric is {self.eval_metric.name}, we recommend using'
@@ -1298,6 +1300,7 @@ class TabularPredictor:
         """
         self._assert_is_fit('predict')
         data = self.__get_dataset(data)
+        self._validate_unique_indices(data, 'data')
         return self._learner.predict(X=data, model=model, as_pandas=as_pandas)
 
     def predict_proba(self, data, model=None, as_pandas=True, as_multiclass=True):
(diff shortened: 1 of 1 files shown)


yinweisu:
platform_tests

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -1861,9 +1870,7 @@ def feature_importance(self, data=None, model=None, features=None, feature_stage
         if (data is None) and (not self._trainer.is_data_saved):
             raise AssertionError(
                 'No data was provided and there is no cached data to load for feature importance calculation. `cache_data=True` must be set in the `TabularPredictor` init `learner_kwargs` argument call to enable this functionality when data is not specified.')
-        if data is not None:
-            # Avoid crash when indices are duplicated
-            data = data.reset_index(drop=True)
+        self._validate_unique_indices(data, 'data')
```
Innixma, line 1873: I forgot I had already added a work-around to fix this via the .reset_index logic.

This is a bit tricky in terms of which we want:

1. Encourage best practice from users and error if they have invalid data
2. Make it work even if they have invalid data
3. Same as 2 but log a warning.

Curious on thoughts.

szha:
Job PR-1504-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1504/2/index.html

tabular/src/autogluon/tabular/predictor/predictor.py
yinweisu: I think raise exception is better if the user simply provides invalid data. Making it to work in our way might not be what the user truly wants. Making it work with a warning is reasonable, but my concern is user would simply neglect the warning because of the amount of logs autogluon give and don't know what we've done.

Innixma: I think that makes sense, let's raise an exception.

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -1861,9 +1863,7 @@ def feature_importance(self, data=None, model=None, features=None, feature_stage
...
```
Innixma, line 1866: data can be None here, only validate if data is not None, otherwise will crash.

```
if data is not None:
    self._validate_unique_indices(data, 'data')
```

Innixma (approved):
LGTM!

Issue #1031, opened 2021-03-24
Confusing error message: HPO on Random Forest


jmrichardson:
Hi,

When tuning RF with the following:

```
    rf_options = {
        'max_depth': ag.space.Int(5, 15),
    }
```

I get the following error:

```
Hyperparameter tuning model: RandomForest_BAG_L1 ...
Warning: Exception caused RandomForest_BAG_L1 to fail during hyperparameter tuning... Skipping this model.
Traceback (most recent call last):
  File "D:\Projects\tactory\autogluon\tabular\trainer\abstract_trainer.py", line 1075, in _train_single_full
    hpo_models, hpo_model_performances, hpo_results = model.hyperparameter_tune(X=X, y=y, k_fold=k_fold, scheduler_options=hyperparameter_tune_kwargs, **model_fit_kwargs)
  File "D:\Projects\tactory\autogluon\core\models\abstract\abstract_model.py", line 683, in hyperparameter_tune
    return self._hyperparameter_tune(scheduler_options=scheduler_options, **kwargs)
  File "D:\Projects\tactory\autogluon\core\models\ensemble\stacker_ensemble_model.py", line 166, in _hyperparameter_tune
    return super()._hyperparameter_tune(X=X, y=y, k_fold=k_fold, scheduler_options=scheduler_options, preprocess_kwargs=preprocess_kwargs, **kwargs)
  File "D:\Projects\tactory\autogluon\core\models\ensemble\bagged_ensemble_model.py", line 640, in _hyperparameter_tune
    hpo_models, hpo_model_performances, hpo_results = self.model_base.hyperparameter_tune(X=X_fold, y=y_fold, X_val=X_val_fold, y_val=y_val_fold, scheduler_options=scheduler_options, **kwargs)
  File "D:\Projects\tactory\autogluon\core\models\abstract\abstract_model.py", line 683, in hyperparameter_tune
    return self._hyperparameter_tune(scheduler_options=scheduler_options, **kwargs)
  File "D:\Projects\tactory\autogluon\tabular\models\rf\rf_model.py", line 166, in _hyperparameter_tune
    return skip_hpo(self, **kwargs)
  File "D:\Projects\tactory\autogluon\core\models\abstract\model_trial.py", line 90, in skip_hpo
    fit_and_save_model(model=model, params=dict(), fit_args=fit_model_args, predict_proba_args=predict_proba_args, y_val=y_val, time_start=time.time(), time_limit=time_limit)
[...]

PR #1511, opened 2022-01-24
Various HPO Cleanup / Fixes

Innixma:
*Issue #, if available:*
#1031

*Description of changes:*

This PR introduces numerous code cleanup / quality of life improvements and major bug fixes to HPO logic.

- **Major Bug Fix**: HPO with bagging did not work correctly. In fact, the hyperparameters found during HPO were completely ignored for all fold models except the first, and instead the default hyperparameters were used. The first fold used the correct hyperparameters, and all other folds used the default hyperparameters. This was detected after the other fixes were made due to more stringent sanity checks on the hyperparameters. It is unknown how long this bug was present, but at minimum since v0.1. This has been fixed by avoiding the copying of the model entirely and instead initializing a new model with the originals parameters after updating based on the search hyperparameters.
- When search space is exhausted, no longer produces a confusing message and instead logs that all possible search spaces have been tried.
- Automatically early stop HPO if 80%+ of trials are failures to avoid hundreds of HPO trials failing in quick succession.
- HPO Searcher code for `self._results` incorrectly pickled configs in a way that the same config could have two different pickles: `{'a': 1, 'b': 2}` vs `{'b': 2, 'a': 1}`. This led to 2 results per config (didn't impact users, but would have impacted anyone extending the searchers). Refactored the pickling process to ensure a config will have a single unique pickle key.
- RandomForest, Linear, KNN, FastText models errored if a user specified a search space, now these models (all models) are able to perform HPO. By default, they have no search space, and I have further added logic to fallback to normal model training if no search space is provided when `hyperparameter_tune` is called. This means RandomForest and KNN will not do HPO by default if HPO is enabled, since their default search space is empty.
- Incremented trial ID's in model names by 1, so the first trial is `\T1` instead of `\T0`. This aligns the indexing with child folds, which start at `F1`.
- Fixed bug in grid searcher caused by returning np.float64 and np.int64 types instead of float and int types. This caused downstream models that did not support np.int64 and np.float64 to crash / spam warnings (XGBoost).
- Trial failures were not being properly tracked, and even if an exception was raised in the trial, the `is_failed` variable would be False. This has been fixed as tracking was necessary for failure based HPO early stop.
- Refactored usage of `_set_cpu_params` in models causing an update to `self.params`. Now this is done in the local variable `params` rather than updating `self.params`.
- Cleaned old HPO search space / parameter specification in LightGBM. The new configs are identical in functionality, while being less cluttered.
- Removed max_reward defaults for ImagePredictor and ObjectDetector that would have caused HPO to stop far earlier than necessary.
- Cleaned ImagePredictor and ObjectDetector HPO logic.
- Fixed CatBoost custom metric support.


szha:
Job PR-1511-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1511/3/index.html

szha:
Job PR-1511-6 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1511/6/index.html

gomlfx:
Nice!

core/src/autogluon/core/searcher/local_searcher.py
```diff
@@ -121,15 +111,102 @@ def get_best_reward(self):
...
+        assert len(config) == len(self._params_order), f'Config length does not match expected params count!\n' \
+                                                       f'Expected: {self._params_order}\n' \
+                                                       f'Actual:   {list(config.keys())}'
+
+        # Note: This code is commented out because it can be computationally and memory expensive if user sends large objects in search space, such as datasets.
+        """
+        for key in self._params_static:
+            assert pickle.dumps(config[key]) == pickle.dumps(self._params_static[key]), \
+                f'Invalid config value for search space parameter "{key}" | Invalid Value: {config[key]} | Expected Value: {self._params_static[key]}'
+        """
```
gradientsky, line 181: Should we even allow to do this? This would be a misuse of functionality. HPO spaces must be light. Large objects might create bottlenecks for distributed settings. I would suggest raising an error.

vision/src/autogluon/vision/_gluoncv/object_detection.py
```diff
@@ -373,7 +357,7 @@ def fit(self, train_data, val_data=None, train_size=0.9, random_state=None, time
         self._logger.info(pprint.pformat(self._fit_summary, indent=2))
 
         if self._cleanup_disk:
-            shutil.rmtree(config['log_dir'], ignore_errors=True)
```
gradientsky, line 376: is it possible that `log_dir` is passed as `/`? This is potentially very destructive operation if absolute paths are allowed.

core/src/autogluon/core/searcher/local_random_searcher.py
```diff
@@ -31,12 +32,24 @@ def _get_params_space(self) -> dict:
...
+    def _get_num_configs(self) -> int:
+        num_unique = 1
+        for key, val in self.search_space.items():
+            if isinstance(val, Space):
+                if isinstance(val, DiscreteSpace):
+                    num_unique *= len(val)
+                else:
+                    num_unique = None
+                    break
+        return num_unique
```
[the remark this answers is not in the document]
Innixma: Currently we would simply set num_unique = None, which only impacts the log message the user receives about the search space being exhausted. So the impact is not very significant. We can revisit if we add nested spaces, likely it would just be a call to the nested space such as len(nested_space).

core/src/autogluon/core/scheduler/seq_scheduler.py
```diff
@@ -147,31 +147,48 @@ def run(self, **kwargs):
...
                 logger.exception('Detailed Traceback:')  # TODO: Avoid logging if verbosity=0
                 break
             trial_end_time = time.time()
-            trial_run_times.append(np.NaN if is_failed else (trial_end_time - trial_start_time))
+
+            trial_count += 1
+            if is_failed:
+                failure_count += 1
+            else:
+                trial_run_times += trial_end_time - trial_start_time
```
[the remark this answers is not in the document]
Innixma: 1. Updated to trials_total_time
2. This is the LocalSequentialScheduler so we don't need to write logic with parallel in mind yet.

core/src/autogluon/core/searcher/local_searcher.py
Innixma: This is something I might plan to address after NLP refactor is merged. One possibility is to disallow static parameters in search spaces. Currently we don't duplicate these parameters in memory after this PR (previously we were pickling and copying these parameters every trial unnecessarily).

vision/src/autogluon/vision/_gluoncv/object_detection.py
Innixma: No because it is suffixed with a UUID str: 

```
log_dir = os.path.join(config.pop('log_dir', os.getcwd()), str(uuid.uuid4())[:8])
```

szha:
Job PR-1511-7 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1511/7/index.html

gradientsky (approved):

Issue #1426, opened 2021-12-17
ImagePredictor trained on GPU does not load on cpu machine


yinweisu:
Detailed error:
```
RuntimeError: Attempting to deserialize object on a CUDA device but torch.cuda.is_available() is False. If you are running on a CPU-only machine, please use torch.load with map_location=torch.device('cpu') to map your storages to the CPU
```

yinweisu:
Should be fixed: https://github.com/dmlc/gluon-cv/pull/1717

PR #1533, opened 2022-02-03
Fix image cpu loading

yinweisu:
*Issue #, if available:*
https://github.com/awslabs/autogluon/issues/1426

*Description of changes:*
Previous changes in GluonCV only solved the estimator loading. Still need to handle the pickle logic. Added a cpu unpickler to handle the special logic. Refer to: https://github.com/pytorch/pytorch/issues/16797

### DIFF
--- a/vision/src/autogluon/vision/predictor/predictor.py
+++ b/vision/src/autogluon/vision/predictor/predictor.py
@@ -22,6 +22,7 @@ from autogluon.core.utils.utils import generate_train_test_split
 
 from ..configs.presets_configs import unpack, _check_gpu_memory_presets
 from ..utils import sanitize_batch_size
+from ..utils.pickle import CPU_Unpickler
 
 __all__ = ['ImagePredictor']
 
@@ -697,7 +698,11 @@ class ImagePredictor(object):
         if os.path.isdir(path):
             path = os.path.join(path, 'image_predictor.ag')
         with open(path, 'rb') as fid:
-            obj = pickle.load(fid)
+            gpu_count = get_gpu_count_all()
+            if gpu_count > 0:
+                obj = pickle.load(fid)
+            else:
+                obj = CPU_Unpickler(fid).load()
         obj._verbosity = verbosity
         return obj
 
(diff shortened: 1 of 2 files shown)


szha:
Job PR-1533-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1533/1/index.html

vision/src/autogluon/vision/utils/pickle.py
```diff
@@ -0,0 +1,12 @@
+import io
+import pickle
+
+
+# https://github.com/pytorch/pytorch/issues/16797
+class CPU_Unpickler(pickle.Unpickler):
```
Innixma, line 6: nit: Perhaps `CPUUnpickler`? Unsure what the proper naming convention is here, but I don't usually see `_` in class names.

vision/src/autogluon/vision/predictor/predictor.py
```diff
@@ -697,7 +698,11 @@ def load(cls, path, verbosity=2):
         if os.path.isdir(path):
             path = os.path.join(path, 'image_predictor.ag')
         with open(path, 'rb') as fid:
-            obj = pickle.load(fid)
+            gpu_count = get_gpu_count_all()
+            if gpu_count > 0:
```
Innixma, line 702: General question: Should we even be pickling an object with a torch cuda tensor in the first place? What alternative serialization methods are there?

Innixma (approved):
LGTM! Added a couple minor comments/questions.

vision/src/autogluon/vision/predictor/predictor.py
yinweisu: This is actually an awesome question and I had a similar concern as well. Currently, ImagePredictor is saving a single pickle file. This behavior differs from what tabular and text are doing: saving to a directory. Ideally, we should save the estimator with torch.save() method, which saves to a file, along with our ImagePredictor pickle file into a directory. However, this is an api breaking change

vision/src/autogluon/vision/utils/pickle.py
yinweisu: The two UU's standing together feels weird, that's why I had the underscore.

Issue #1519, opened 2022-01-28
BUG: LightGBM and XGBoost slow down FastAI and TorchNN on Linux


Innixma:
On Linux OS, LightGBM and XGBoost slow down FastAI and TorchNN training by a large margin on medium quality / when parallel training is disabled. Across all datasets in AutoMLBenchmark, **this averages to be a roughly 6x increase in training time for NN models**.

In LightGBM code:

```
    def _predict_proba(self, X, num_cpus=0, **kwargs):
        ...
        if self.problem_type == REGRESSION:
            return self.model.predict(X, num_threads=num_cpus)

        y_pred_proba = self.model.predict(X, num_threads=num_cpus)
        ...
```

The slow-down is gone if this is changed to:

```
    def _predict_proba(self, X, num_cpus=0, **kwargs):
        ...
        if self.problem_type == REGRESSION:
            return self.model.predict(X)

        y_pred_proba = self.model.predict(X)
        ...
```

In XGBoost code, it is caused in two separate areas:

```
    def _predict_proba(self, X, num_cpus=-1, **kwargs):
        ...
        self.model.set_params(n_jobs=num_cpus)
        ...

    @classmethod
    def load(cls, path: str, reset_paths=True, verbose=True):
        ...
        model.model = model._xgb_model_type()
        model.model.load_model(path + 'xgb.ubj')
        ...
```

If either of these two logics are used, XGBoost will slow down training of NN models.
Note that `model.model.load_model(path + 'xgb.ubj')` modifies `n_jobs` internally. Using the old much slower save/load method of XGBoost side-steps the issue. Another way to side-step the issue is specify `n_jobs=1` in `self.model.set_params(n_jobs=1)` and `model.model = model._xgb_model_type(n_jobs=1)`, however this has the side effect of slowing down inference for XGBoost.

On dataset abalone, Torch NN training speed:

```
Without GBM/XGB: 6.77s
With GBM/XGB   : 18.50s
```

FastAI training speed:

```
Without GBM/XGB: 4.30s
With GBM/XGB   : 14.03s
```

I suspect this also slows down inference speed, but I haven't investigated inference speed differences. I did not observe the slow-down on MacOS.

Innixma:
Resolved in #1535, thanks @yinweisu!

PR #1535, opened 2022-02-04
Set num_threads for torch nn models

yinweisu:
*Issue #, if available:*
https://github.com/awslabs/autogluon/issues/1519

*Description of changes:*
Making sure TorchNN and FASTAI are using the correct number of threads. This would address issues when other models modify the global thread setting.
Only updated this logic for fit because we find out that the overhead of setting the thread is not worth it during inference after some experiments.


Innixma (commented):
Thanks for looking into this! Will plan to benchmark to confirm fix.

Innixma (approved):
LGTM! On average this speeds up medium quality by 58% (Also likely speeds up non-ray based bagging too)! No drop in performance observed.

PR #1536, opened 2022-02-08
Tabular: Exclude ngrams for FastAI

Innixma:
*Issue #, if available:*

*Description of changes:*

Exclude ngrams for FastAI to speedup inference speed by 20x+.

Previously, FastAI is very slow if given many ngrams as input, and accuracy is not good compared to a language model. It is better to not send ngrams to FastAI. This improves inference latency for 1 row of data from 1.5 seconds to 75 ms for NLP datasets.

Additionally added a copy to preprocessing to avoid warning from pandas regarding fragmented DF.


szha:
Job PR-1536-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1536/1/index.html

gradientsky (approved):

PR #1537, opened 2022-02-08
[0.4 Release]Add Pytorch-based TextPredictor & Experimental AutoMMPredictor

zhiqiangdon:
*This PR implements the Pytorch-based multimodal predictors: TextPredictor and AutoMMPredictor (experimental)*

**Pytorch-based TextPredictor**

- Better performance than the MXNet-based TextPredictor. (Average score **0.672** vs. 0.66411 in the multimodal benchmark paper https://arxiv.org/pdf/2111.02705.pdf)
- Faster training and inference. The new version has **~2.88x** speedup in training and **~1.40x** speedup in inference. With g4dn.12x instance, the model can further achieve **2.26x** speedup with 4 GPUs.
- Support multilingual models for zero-shot cross lingual transfer. This includes XLM-R and mdeberta-v3 (https://arxiv.org/abs/2111.09543).
- Support backbones in Huggingface/Transformers, such as deberta-v3, roberta, electra-base and gpt2.
- Half-precision training (less memory and faster training).

**AutoMMPredictor**

- Flexible late-fusion for any combinations of single and multimodal models.
- Support HuggingFace auto models (https://github.com/huggingface/transformers).
- Support TIMM image auto models (https://github.com/rwightman/pytorch-image-models).
- A general pytorch data loader system that can simultaneously load data for multiple models.
- A scalable and easy-to-use configuration design with templates provided. Users can flexibly switch backbones by overriding the checkpoint_name hyper-parameter.
- Support finetuning with CLIP.
- Automatic GPU selection.


text/setup.py
```diff
@@ -22,9 +22,24 @@
...
+    'tqdm',
+    'pandas',
+    'nptyping',
+    'omegaconf',
+    'timm',
+    'pytest',
+    'sentencepiece',
     f'autogluon.core=={version}',
-    'autogluon-contrib-nlp==0.0.1b20210201',
+    # 'autogluon-contrib-nlp==0.0.1b20210201',
```
sxjscience, line 42: Will try to work to upgrade `autogluon-contrib-nlp`.

sxjscience:
@Innixma @jwmueller @gradientsky @yinweisu The logic can be reviewed first while we are fixing the test + add more comments +  fix the tutorials.

text/src/autogluon/text/automm/data/randaug.py
```diff
@@ -0,0 +1,269 @@
+# code in this file is adpated from rpmcruz/autoaugment
```
sxjscience, line 1: Is it possible to remove this file and reuse https://pytorch.org/vision/main/generated/torchvision.transforms.RandAugment.html?

text/src/autogluon/text/automm/configs/environment/default.yaml
```diff
@@ -0,0 +1,12 @@
+env:
+  num_gpus: -1
+  num_nodes: 1
+  batch_size: 128  # this is a desired batch size; pl trainer will accumulate gradients when per step batch is smaller.
+  per_gpu_batch_size: 16  # you can define this according to your GPU memory
```
sxjscience, line 5: Change to 8.

text/src/autogluon/text/automm/configs/model/hf_text.yaml
```diff
@@ -0,0 +1,20 @@
...
+  hf_text:
+#    checkpoint_name: "google/electra-base-discriminator"
+    checkpoint_name: 'microsoft/deberta-v3-small'
+#     checkpoint_name: 'bert-base-uncased'
+#    checkpoint_name: 'xlm-roberta-base'
+#    checkpoint_name: 'microsoft/deberta-base'
+#    checkpoint_name: 'roberta-base'
+#    checkpoint_name: "distilbert-base-uncased"
+#    checkpoint_name: "bert-base-chinese"
+#    checkpoint_name: "gpt2"
```
sxjscience, line 12: remove comments.

text/src/autogluon/text/automm/configs/environment/default.yaml
```diff
@@ -0,0 +1,12 @@
...
+  precision: 16  # training precision
+  num_workers: 2  # pytorch training dataloader workers
+  num_workers_evaluation: 2  # pytorch prediction/test dataloader workers
+  fast_dev_run: False
+  deterministic: False
+  auto_select_gpus: True
+  strategy: "ddp"
```
sxjscience, line 12: Change to "ddp_spawn" so that it will work for jupyter notebooks / colab.

text/src/autogluon/text/automm/data/infer_types.py
```diff
@@ -0,0 +1,301 @@
...
+    ----------
+    data
+        The training data series
+    Returns
+    -------
+    is_image
+        Whether the column is a image column
+    """
+    try:
+        if (data.iloc[0].startswith('/') or data.iloc[0].startswith('.')):
```
sxjscience, line 118: Consider to revise the logic. This essentially detects if the data contain paths but these paths may not be images. We may sample a few entries from the column, use PIL to open them and check the success ratio.

text/setup.py
```diff
@@ -22,9 +22,24 @@
...
+    'torchmetrics',
+    'transformers>=4.16.2',
+    'Pillow',
+    'tqdm',
+    'pandas',
+    'nptyping',
+    'omegaconf',
+    'timm',
+    'pytest',
+    'sentencepiece',
```
Innixma, line 40: [!!!] This is extremely fragile. We need version caps in place.

Why do we depend on so much? We should not depend on timm, Pillow, etc. in text.

text/setup.py
```diff
@@ -22,9 +22,24 @@
...
+    'pytorch_lightning>=1.5.9',
+    'torchmetrics',
+    'transformers>=4.16.2',
+    'Pillow',
+    'tqdm',
+    'pandas',
+    'nptyping',
+    'omegaconf',
+    'timm',
+    'pytest',
```
Innixma, line 39: We should not depend on pytest.

text/setup.py
```diff
@@ -22,9 +22,24 @@
     'scipy',
     'pandas',
     'scikit-learn',
-
+    'fairscale',
+    'scikit-image',
+    'boto3',
+    'pytest',
```
Innixma, line 28: We should not depend on pytest + it is present multiple times in this list.

text/setup.py
```diff
@@ -22,9 +22,24 @@
...
+    'smart_open',
+    'pytorch_lightning>=1.5.9',
+    'torchmetrics',
+    'transformers>=4.16.2',
+    'Pillow',
+    'tqdm',
+    'pandas',
```
Innixma, line 35: Why is pandas in the list twice?

text/setup.py
Innixma: Won't this break MXNet version of code if it is commented out?

text/src/autogluon/text/automm/__init__.py
```diff
@@ -0,0 +1,9 @@
+__version__ = '0.1'
```
Innixma, line 1: Incorrect version, this is not how we specify version.

Instead:

```
try:
except ImportError:
    pass
```

text/src/autogluon/text/automm/configs/model/fusion_mlp_image_text_tabular.yaml
```diff
@@ -0,0 +1,77 @@
+model:
```
Innixma, line 1: general question on configs: What happens if I typo one of the fields (such as `hidden_size_INVALID` instead of `hidden_size`. Does the code error / warn me, or does it silently ignore the value and train anyways?

text/src/autogluon/text/automm/constants.py
```diff
@@ -0,0 +1,70 @@
...
+TEST_ACC = "test_acc"
+MODE = "mode"
+MAX = "max"
+MIN = "min"
+ACCURACY = "accuracy"
+ACC = "acc"
+RMSE = "rmse"
+QUADRATIC_KAPPA = "quadratic_kappa"
+val_metric_name = {BINARY: VAL_ACC, MULTICLASS: VAL_ACC}
+# minmax_mode = {VAL_ACC: MAX, VAL_LOSS: MIN}
```
yinweisu, line 40: remove comments if not needed

Innixma (commented):
Added initial review of data preprocessing / outer AutoML logic / dependencies.

text/src/autogluon/text/automm/models/clip.py
```diff
@@ -0,0 +1,99 @@
...
+        if len(names) > 0:
+            print(f"outer layers are treated as head: {names}")
+        for n in names:
+            assert n not in name_to_id
+            name_to_id[n] = 0
+
+        return name_to_id
+
+
+
```
gradientsky, line 99: nit: extra lines

text/tests/unittests/automm/configs/model/categorical_mlp.yaml
```diff
@@ -0,0 +1,10 @@
+model:
+  names: "categorical_mlp"
+  categorical_mlp:
+    hidden_size: 64
+    activation: "leaky_relu"
+    num_layers: 1
+    drop_rate: 0.1
+    normalization: "layer_norm"
+    data_types:
+      - "categorical"
```
gradientsky, line 10: nit: no new line at the end of the file

text/tests/unittests/automm/datasets.py
```diff
@@ -0,0 +1,374 @@
...
+        #     random_state=np.random.RandomState(123),
+        #     stratify=self._train_df[self.label_columns[0]],
+        # )
+        # _, self._test_df = train_test_split(
+        #     self._test_df,
+        #     test_size=0.5,
+        #     random_state=np.random.RandomState(123),
+        #     stratify=self._test_df[self.label_columns[0]],
+        # )
+        self._train_df.reset_index(drop=True, inplace=True)
```
gradientsky, line 233: dead code - remove

text/tests/unittests/automm/datasets.py
```diff
@@ -0,0 +1,374 @@
...
+        #     test_size=0.5,
+        #     random_state=np.random.RandomState(123),
+        # )
+        # _, self._test_df = train_test_split(
+        #     self._test_df,
+        #     test_size=0.5,
+        #     random_state=np.random.RandomState(123),
+        # )
+        self._train_df.reset_index(drop=True, inplace=True)
+        self._test_df.reset_index(drop=True, inplace=True)
```
gradientsky, line 313: remove dead code

text/tests/unittests/automm/run.py
```diff
@@ -0,0 +1,130 @@
...
+        metric_name = "rmse"
+
+    # train_data, tuning_data = train_test_split(
+    #     dataset.train_df,
+    #     test_size=0.1,
+    #     random_state=np.random.RandomState(123)
+    # )
+    # print("fixed split, no sampling")
+    # train_data = dataset.train_df.head(int(0.9 * len(dataset.train_df)))
+    # tuning_data = dataset.train_df.tail(int(0.1 * len(dataset.train_df)))
```
gradientsky, line 77: remove dead commented-out code

text/setup.py
sxjscience: I've updated `autogluon-contrib-nlp`. It can be resolved via

```python
' autogluon-contrib-nlp==0.0.1b20220208',
```

text/src/autogluon/text/automm/data/collator.py
```diff
@@ -0,0 +1,324 @@
+import math
+import warnings
+import numpy as np
+import torch
+
+__all__ = ["Stack", "Pad", "Tuple", "List", "Dict", "PadImage", "Zip"]
+
+
+def _pad_arrs_to_max_length(arrs, pad_axis, pad_val, round_to=None, max_length=None):
```
gradientsky, line 9: Is this code used to pad multi-dimensional sequences like:
```
[torch.rand([2,3]), torch.rand([2,8]), torch.rand([2,5])]
```
If it's just for one-dimensional sequences, then why not use [`torch.nn.utils.rnn.pad_sequence`](https://pytorch.org/docs/stable/generated/torch.nn.utils.rnn.pad_sequence.html)?

text/src/autogluon/text/automm/data/collator.py
```diff
@@ -0,0 +1,324 @@
...
+
+    if round_to is not None:
+        max_arr_len = round_to * math.ceil(max_arr_len / round_to)
+    elif max_length is not None:
+        if max_length < max_arr_len:
+            raise ValueError(
+                "If max_length is specified, max_length={} must be larger "
+                "than the maximum length {} of the given arrays at axis={}.".format(
+                    max_length, max_arr_len, pad_axis
+                )
```
gradientsky, line 37: Is there any scenario where we would want to trim to the `max_length` then pad instead of erroring-out?

text/src/autogluon/text/automm/data/collator.py
gradientsky: Please add tests for this function

text/src/autogluon/text/automm/data/collator.py
```diff
@@ -0,0 +1,324 @@
...
+    for i, tensor in enumerate(arrs):
+        length = tensor.size(pad_axis)
+        out_tensor[i].narrow(pad_axis, 0, length)[:] = tensor
+
+    original_length = torch.as_tensor(original_length)
+
+    return out_tensor, original_length
+
+
+def _stack_arrs(arrs, use_shared_mem):
```
gradientsky, line 56: Tests

text/src/autogluon/text/automm/data/collator.py
```diff
@@ -0,0 +1,324 @@
...
+            Mutually exclusive with :attr:`max_length`. (default ``None``)
+        max_length (int, optional):
+            If specified, the padded dimension will have length :attr:`max_length`,
+            and it must be larger than the maximum length in the arrays at :attr:`axis`.
+            Mutually exclusive with :attr:`round_to`.  (default ``None``)
+        ret_length (bool, optional): Whether to return the valid length in the output.
+            (default ``False``)
+    """
+
+    def __init__(self, axis=0, pad_val=None, round_to=None, max_length=None, ret_length=False):
```
gradientsky, line 110: nit: `pad_val=0` default

text/src/autogluon/text/automm/data/collator.py
```diff
@@ -0,0 +1,324 @@
...
+
+        ret = torch.zeros(batch_size, 3, max_height, max_width,
+                          device=image_list[0].device,
+                          dtype=image_list[0].dtype)
+
+        for i, image in enumerate(image_list):
+            # assign to the top-left since convolution starts processing from top-left
+            ret[i, :, :image.shape[1], :image.shape[2]] = image
+
+        return ret
```
gradientsky, line 312: 1. Please reuse torch code: [`torchvision.transforms.Pad()`](https://pytorch.org/vision/stable/transforms.html) or pillow version (that one I think is faster).
2. Zero padding in many cases not the optimal choice.

text/src/autogluon/text/automm/data/datamodule.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def predict_dataloader(self):
+        loader = DataLoader(
+            self.predict_dataset,
+            batch_size=self.per_gpu_batch_size,
+            num_workers=self.num_workers,
+            pin_memory=False,
+            collate_fn=self.get_collate_fn(),
+        )
+        return loader
```
gradientsky, line 95: Code duplication for the `DataLoader` construction with only 1st parameter changing - consider extracting a function

text/src/autogluon/text/automm/data/collator.py
```diff
@@ -0,0 +1,324 @@
...
+                "The number of attributes in each data sample "
+                "should contains {} elements".format(len(self._fn))
+            )
+        ret = []
+        for i, ele_fn in enumerate(self._fn):
+            ret.append(ele_fn([ele[i] for ele in data]))
+        return tuple(ret)
+
+
+class List:
```
gradientsky, line 228: It looks like the List is not used anywhere

text/src/autogluon/text/automm/data/infer_types.py
```diff
@@ -0,0 +1,301 @@
+import collections
+import pandas as pd
+import warnings
+from typing import Union, Optional, List, Dict, Tuple
+from pandas import DataFrame, Series
+from ..constants import NULL, CATEGORICAL, NUMERICAL, TEXT, IMAGE_PATH, MULTICLASS, BINARY, REGRESSION
+
+
+def is_categorical_column(data: pd.Series,
```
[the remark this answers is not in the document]
gradientsky: We had a few issues with different detection mechanisms in core vs text. Is it possible to unify the logic?

text/src/autogluon/text/automm/data/infer_types.py
```diff
@@ -0,0 +1,301 @@
...
+        Whether the column is a categorical column
+    """
+    if data.dtype.name == 'category':
+        return True
+    else:
+        if threshold is None:
+            if is_label:
+                threshold = 100
+                oov_ratio_threshold = 0
+                ratio = 0.1
```
gradientsky, line 53: what if these are provided? Do we want to override the ones passed as the method args?

text/src/autogluon/text/automm/data/infer_types.py
```diff
@@ -0,0 +1,301 @@
...
+            else:
+                threshold = 20
+                oov_ratio_threshold = 0
+                ratio = 0.1
```
gradientsky, line 57: ditto

text/src/autogluon/text/automm/data/infer_types.py
```diff
@@ -0,0 +1,301 @@
...
+        key_set = set(data_value_counts.keys())
+        if len(data_value_counts) < threshold:
+            valid_value_counts = valid_data.value_counts(dropna=False)
+            total_valid_num = len(valid_data)
+            oov_num = 0
+            for k, v in zip(valid_value_counts.keys(), valid_value_counts.values):
+                if k not in key_set:
+                    oov_num += v
+            if is_label and oov_num != 0:
+                return False
```
gradientsky, line 69: If we have `[1, 2]` in training label and `[1, 2, 3]` in validation set, why this method would return this is not categorical variable? Granted that it's a bad news if we got into this situation, but it is totally possible if the splits are generated using methods other than stratification by the label. 

If this one is not categorical (assuming we have only a few values on a large dataset), that would make it a regression problem?

text/src/autogluon/text/automm/data/preprocess_dataframe.py
```diff
@@ -0,0 +1,336 @@
...
+)
+from sklearn.base import (
+    TransformerMixin,
+    BaseEstimator,
+)
+from ..constants import CATEGORICAL, NUMERICAL, TEXT, IMAGE_PATH, NULL
+logger = logging.getLogger(__name__)
+
+
+class MultiModalFeaturePreprocessor(TransformerMixin, BaseEstimator):
```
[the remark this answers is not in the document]
sxjscience: I actually tried to use PipelineFeatureGenerator previously but that may not make the code concise. How about consider to refactor the preprocessing logic in a later PR (We may add other features like prompt-based encoding).

text/src/autogluon/text/automm/data/process_text.py
```diff
@@ -0,0 +1,262 @@
...
+            "ignore",
+            "Token indices sequence length is longer than.*result in indexing errors"
+        )
+        for col_text in text:
+            col_tokens = self.tokenizer.encode(
+                col_text,
+                add_special_tokens=False,
+                truncation=False,
+            )
+            tokens.append(np.array(col_tokens, dtype=np.int32))
```
gradientsky, line 177: Is there any good reason why we work with numpy instead of working directly with torch tensors? Seems inefficient and memory consuming to me.

text/src/autogluon/text/automm/data/process_text.py
```diff
@@ -0,0 +1,262 @@
...
+    def get_pretrained_tokenizer(
+            tokenizer_name: str,
+            checkpoint_name: str,
+    ):
+        tokenizer_class = ALL_TOKENIZERS[tokenizer_name]
+        if torch.distributed.is_initialized():
+            if torch.distributed.get_rank() == 0:
+                tokenizer_class.from_pretrained(checkpoint_name)
+            torch.distributed.barrier()
+        return tokenizer_class.from_pretrained(checkpoint_name)
```
gradientsky, line 203: `tokenizer_class.from_pretrained(checkpoint_name)` would be called twice if `.get_rank()` returns 0. Is this expected?

text/src/autogluon/text/automm/data/utils.py
```diff
@@ -0,0 +1,17 @@
+from typing import Union, Optional, List, Tuple
+
+
+def get_default_config_value(
+        config: dict,
+        keys: Tuple[str, ...],
+):
```
gradientsky, line 7: The function name does not match what it is doing. This looks to me more like filter to only allowed values.

text/src/autogluon/text/automm/configs/environment/default.yaml
```diff
@@ -0,0 +1,12 @@
+env:
+  num_gpus: -1
+  num_nodes: 1
+  batch_size: 128  # this is a desired batch size; pl trainer will accumulate gradients when per step batch is smaller.
+  per_gpu_batch_size: 16  # you can define this according to your GPU memory
+  precision: 16  # training precision
```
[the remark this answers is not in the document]
sxjscience: 16 means using fp16. We have 32, 16, bf16 (those in pytorch lightning: https://pytorch-lightning.readthedocs.io/en/latest/advanced/mixed_precision.html).

text/src/autogluon/text/automm/configs/environment/default.yaml
```diff
@@ -0,0 +1,12 @@
...
+  num_workers: 2  # pytorch training dataloader workers
+  num_workers_evaluation: 2  # pytorch prediction/test dataloader workers
+  fast_dev_run: False
+  deterministic: False
```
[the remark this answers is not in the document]
sxjscience: When it is True, it will turn on `torch.cudnn.deterministic` that will force to use some deterministic kernels and throw error if there are only non-deterministic implementations (doc: https://pytorch.org/docs/stable/notes/randomness.html#avoiding-nondeterministic-algorithms). However, there are certain operations that are really difficult to be made deterministic. One example is the backward operation of affineTransform, which require the usage of "atomicAdd()". Thus, we may not turn on this flag.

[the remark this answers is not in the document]
sxjscience: This does not mean that our method won't mostly be deterministic with the fixed random seed.

text/src/autogluon/text/automm/data/preprocess_dataframe.py
```diff
@@ -0,0 +1,336 @@
...
+    def __init__(
+            self,
+            cfg: dict,
+            column_types: collections.OrderedDict,
+            label_column: str,
+            label_generator: Optional[LabelEncoder] = None,
```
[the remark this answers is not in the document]
sxjscience: This is just about preprocessing the data-frame. We preprocess feature / label jointly.

text/src/autogluon/text/automm/data/collator.py
sxjscience: @zhiqiangdon I think we may just remove PadImage since the functionality is supported by `Pad()`.

text/src/autogluon/text/automm/models/utils.py
```diff
@@ -0,0 +1,190 @@
...
+def init_weights(module: nn.Module):
+    if isinstance(module, nn.Embedding):
+        nn.init.xavier_normal_(module.weight)
+    elif isinstance(module, nn.Linear):
+        nn.init.xavier_uniform_(module.weight)
+        if module.bias is not None:
+            nn.init.zeros_(module.bias)
+    elif isinstance(module, nn.LayerNorm):
+        module.bias.data.zero_()
+        module.weight.data.fill_(1.0)
```
gradientsky, line 14: This should account what activations are used - if ReLU is used then `kaiming_normal_` should be used; if Leaky ReLU, then `kaiming_uniform_` is better at preventing exploding/collapsing gradients. Xavier initialization is better for sigmoid activation function. I know transformers heavily prefer sigmoid gates, but if this code is used in other architectures with ReLU, it's not an optimal choice - maybe add documentation when it supposed to be used.

text/src/autogluon/text/automm/data/process_text.py
sxjscience: @zhiqiangdon I think we may simply remove the code related to `torch.distributed.get_rank() == 0`. These are not useful.

text/src/autogluon/text/automm/data/collator.py
zhiqiangdon: PadImage removed.

text/src/autogluon/text/text_prediction/predictor.py
```diff
@@ -0,0 +1,414 @@
...
+                presets=presets,
+                hyperparameters=hyperparameters,
+                column_types=column_types,
+                num_cpus=num_cpus,
+                num_gpus=num_gpus,
+                num_trials=num_trials,
+                plot_results=plot_results,
+                holdout_frac=holdout_frac,
+                save_path=save_path,
+                seed=123,
```
sxjscience, line 260: Change: `seed=seed`

text/src/autogluon/text/automm/predictor.py
```diff
@@ -0,0 +1,924 @@
...
+    #     self.verbosity = verbosity
+    #     set_logger_verbosity(self.verbosity, logger=logger)
+
+    def fit(
+            self,
+            train_data: pd.DataFrame,
+            config: Optional[dict] = None,
+            tuning_data: Optional[pd.DataFrame] = None,
+            save_path: Optional[str] = None,
+            overrides: Optional[Union[str, Dict, List[str]]] = None,
```
sxjscience, line 155: Consider to also use 'hyperparameters'.

text/src/autogluon/text/automm/data/randaug.py
zhiqiangdon: This file is copied from ViLT's implementation, which has removed two operations: color inversion and cutout. They give the reason of doing so in their paper:

"We use all the original policies except two: color inversion, because texts often contain color information as well, and cutout, as it may clear out small but important objects dispersed throughout the whole image."

text/src/autogluon/text/automm/data/infer_types.py
```diff
@@ -0,0 +1,301 @@
...
+    try:
+        if (data.iloc[0].startswith('/') or data.iloc[0].startswith('.')):
+            return True
+        else:
+            return False
+    except:
+        return False
+
+
+def check_if_nlp_feature(X: Series,
```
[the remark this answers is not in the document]
zhiqiangdon: I copied this function from previous TextPredictor's implementation. After reading the code in autogluon/common/features/infer_types.py, I think we may need to redesign our APIs there by referring to automm/data/infer_types. Basically, we may need to have one function for each modality to check whether a column belongs to that modality.

text/src/autogluon/text/automm/data/infer_types.py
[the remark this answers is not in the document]
zhiqiangdon: I agree that valid_data are not necessary.
This function also comes from the previous TextPredictor.
I agree that we can unify `autogluon/common/features/infer_types.py` and `automm/data/infer_types`. We can discuss API design. Some APIs in `autogluon/common/features/infer_types.py` are not intuitive.

text/src/autogluon/text/automm/data/infer_types.py
```diff
@@ -0,0 +1,301 @@
...
+    is_image
+        Whether the column is a image column
+    """
+    try:
+        if (data.iloc[0].startswith('/') or data.iloc[0].startswith('.')):
+            return True
+        else:
+            return False
+    except:
+        return False
```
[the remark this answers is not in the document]
zhiqiangdon: Good point! Will change the detection mechanism.

text/src/autogluon/text/automm/data/infer_types.py
```diff
@@ -0,0 +1,301 @@
...
+
+
+def printable_column_type_string(column_types):
+    ret = 'Column Types:\n'
+    for col_name, col_type in column_types.items():
+        ret += f'   - "{col_name}": {col_type}\n'
+    return ret
+
+
+def infer_problem_type(column_types, label_column, data_df,
```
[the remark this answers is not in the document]
zhiqiangdon: Our benchmark datasets already provide problem_type. This function is mainly used to get the class number (for classification) double-check whether the class number matches binary or multi-class.

I feel that we can use `autogluon.core.utils.utils.infer_problem_type` here, but need to compute the class number after that. Let me try it.

text/setup.py
zhiqiangdon: Pytest removed.

text/src/autogluon/text/automm/__init__.py
zhiqiangdon: Thanks. Revised.

text/src/autogluon/text/automm/configs/model/fusion_mlp_image_text_tabular.yaml
zhiqiangdon: It will throw errors when the required keys are missing. Adding unused keys has no effect.

text/src/autogluon/text/automm/configs/model/categorical_mlp.yaml
```diff
@@ -0,0 +1,10 @@
+model:
+  names: "categorical_mlp"
+  categorical_mlp:
+    hidden_size: 64
+    activation: "leaky_relu"
+    num_layers: 1
+    drop_rate: 0.1
+    normalization: "layer_norm"
+    data_types:
+      - "categorical"
```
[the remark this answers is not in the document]
zhiqiangdon: Fixed.

text/src/autogluon/text/automm/configs/model/hf_text.yaml
```diff
@@ -0,0 +1,20 @@
...
+    data_types:
+      - "text"
+    tokenizer_name: "hf_auto"
+#    merge: True  # Whether we will merge different text columns or treat them independently.
+    max_text_len: 512  # The maximum possible length.
+    insert_sep: True
+    text_segment_num: 2
+    stochastic_chunk: False
```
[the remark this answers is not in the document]
zhiqiangdon: Thanks. All fixed.

text/src/autogluon/text/automm/constants.py
zhiqiangdon: Removed.

text/src/autogluon/text/automm/configs/environment/default.yaml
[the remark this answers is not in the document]
zhiqiangdon: Added the reference link.

text/src/autogluon/text/automm/constants.py
```diff
@@ -0,0 +1,70 @@
...
+VAL_ACC = "val_acc"
+TEST_ACC = "test_acc"
+MODE = "mode"
+MAX = "max"
+MIN = "min"
+ACCURACY = "accuracy"
+ACC = "acc"
+RMSE = "rmse"
+QUADRATIC_KAPPA = "quadratic_kappa"
+val_metric_name = {BINARY: VAL_ACC, MULTICLASS: VAL_ACC}
```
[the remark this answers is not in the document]
zhiqiangdon: This code turns out unused. Removed.

text/src/autogluon/text/automm/configs/optimization/adamw.yaml
```diff
@@ -0,0 +1,15 @@
...
+  optim_type: "adamw"
+  learning_rate: 1.0e-4
+  weight_decay: 0.001
+  lr_choice: "layerwise_decay"
+  lr_decay: 0.8
+  lr_schedule: "cosine_decay"
+  max_epochs: 10
+  max_steps: -1
+  warmup_steps: 0.1
+  end_lr: 0
```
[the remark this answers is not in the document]
zhiqiangdon: Agree. I followed ViLT's implementation https://github.com/dandelin/ViLT/blob/762fd3975c180db6fc88f577cf39549983fa373a/vilt/config.py#L61

I will try to run some experiments to see the comparisons.

text/src/autogluon/text/automm/models/clip.py
zhiqiangdon: Fixed.

text/tests/unittests/automm/configs/model/categorical_mlp.yaml
zhiqiangdon: Fixed.

text/tests/unittests/automm/datasets.py
zhiqiangdon: Removed.

text/tests/unittests/automm/run.py
zhiqiangdon: Removed.

text/tests/unittests/automm/datasets.py
zhiqiangdon: Removed.

text/src/autogluon/text/automm/configs/model/timm_image.yaml
```diff
@@ -0,0 +1,23 @@
...
+  names: "timm_image"
+  timm_image:
+    checkpoint_name: "swin_base_patch4_window7_224"
+#    checkpoint_name: "beit_base_patch16_224"
+#    checkpoint_name: "vit_small_patch16_384"
+#    checkpoint_name: "convmixer_1536_20"
+#    checkpoint_name: "efficientnet_b1"
+#    checkpoint_name: "resnet18"
+#    checkpoint_name: "legacy_seresnet18"
+#    checkpoint_name: "resnest14d"
```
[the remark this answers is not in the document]
zhiqiangdon: Removed.

text/src/autogluon/text/automm/constants.py
```diff
@@ -0,0 +1,70 @@
...
+# Modality keys. may need to update here if new modality keys are added in above.
+ALL_MODALITIES = [IMAGE, TEXT, CATEGORICAL, NUMERICAL]
+
+# Keys to compute metrics
+Y_PRED = "y_pred"
+Y_PRED_PROB = "y_pred_prob"
+Y_TRUE = "y_true"
+
+# Configuration keys
+# CONFIG_DIR = "config_dir"
```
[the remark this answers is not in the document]
zhiqiangdon: Removed.

text/src/autogluon/text/automm/data/collator.py
zhiqiangdon: None will be handled later by `self._pad_val = 0 if pad_val is None else pad_val`. But following your suggestion, I have changed the default to 0 since it makes more sense.

text/src/autogluon/text/automm/data/collator.py
zhiqiangdon: Added.

text/src/autogluon/text/automm/data/datamodule.py
zhiqiangdon: The {train/val/test/predict}_dataloader is registered by Pytorch Lightning. Since we don't know when extactly these methods will be called, we should use it in this way.
Refer to https://pytorch-lightning.readthedocs.io/en/latest/extensions/datamodules.html#lightningdatamodule-api

text/src/autogluon/text/automm/data/collator.py
zhiqiangdon: Just in case we may need it in future, I have added a unit test case for it too. Maybe we can keep it for now.

text/src/autogluon/text/automm/data/infer_types.py
zhiqiangdon: This is a legacy function from previous TextPredictor. We can replace it after aligning our API design between `automm/data/infer_types.py` and `autogluon.common.features.infer_types`.

text/src/autogluon/text/automm/data/infer_types.py
zhiqiangdon: Same here. We can fix it after API alignment.

text/src/autogluon/text/automm/data/process_text.py
zhiqiangdon: Removed.

text/src/autogluon/text/automm/data/utils.py
zhiqiangdon: Name changed to `extract_value_from_config` because we try to extract some default hyper-parameter's value from a config.

text/src/autogluon/text/automm/data/infer_types.py
zhiqiangdon: We can fix the logic after aligning the API design for inferring types.

text/src/autogluon/text/automm/data/process_image.py
```diff
@@ -0,0 +1,175 @@
...
+            max_img_num_per_col = 1
+        self.max_img_num_per_col = max_img_num_per_col
+        print(f"max_img_num_per_col: {max_img_num_per_col}")
+
+        self.train_processor = self.construct_processor(self.train_transform_types)
+        self.val_processor = self.construct_processor(self.val_transform_types)
+
+    def collate_fn(self) -> dict:
+        fn = {}
+        fn.update({f"{self.prefix}_{IMAGE}": Pad(pad_val=0)})
```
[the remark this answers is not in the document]
zhiqiangdon: Have you seen other paddings for image data?

text/src/autogluon/text/automm/data/process_image.py
```diff
@@ -0,0 +1,175 @@
...
+
+    @staticmethod
+    def mean_std(norm_type: str):
+        if norm_type == "inception":
+            return IMAGENET_INCEPTION_MEAN, IMAGENET_INCEPTION_STD
+        elif norm_type == "imagenet":
+            return IMAGENET_DEFAULT_MEAN, IMAGENET_DEFAULT_STD
+        elif norm_type == "clip":
+            return (0.48145466, 0.4578275, 0.40821073),\
+                   (0.26862954, 0.26130258, 0.27577711)
```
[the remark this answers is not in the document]
zhiqiangdon: Sure. Done.

text/src/autogluon/text/automm/data/process_text.py
zhiqiangdon: The main reason is to use NDArray in the type hint to control array shape. We don't find memory issue using the current implementation. We can leave this to the future discussion whether it's necessary to use torch.tensor.

text/src/autogluon/text/automm/models/clip.py
```diff
@@ -0,0 +1,99 @@
...
+    def forward(self, batch):
+        text_token_ids = batch[self.text_token_ids_key]
+        text_valid_length = batch[self.text_valid_length_key]
+        image = batch[self.image_key]
+
+        steps = torch.arange(0, text_token_ids.shape[1]).type_as(text_valid_length)
+        text_masks = (steps.reshape((1, -1)) < text_valid_length.reshape((-1, 1))).type_as(text_token_ids)
+
+        if image.dim() == 5 and image.shape[1] == 1:
+            image = torch.squeeze(image, dim=1)
```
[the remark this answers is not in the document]
zhiqiangdon: Added two lines of comments:

# Image batch has shape (batch_size, image_num, 3, height, width)
# Currently, we only support image_num=1 for CLIP input.

text/src/autogluon/text/automm/data/collator.py
```diff
@@ -0,0 +1,324 @@
...
+        arrs (list)
+        pad_axis (int)
+        pad_val (number)
+        round_to (int, optional). (default: ``None``)
+        max_length (int, optional). (default: ``None``)
+
+    Returns:
+        ret : torch.Tensor
+        original_length : torch.Tensor
+    """
```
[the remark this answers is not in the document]
zhiqiangdon: Transformers don't where the padding is as long as we give it the correct mask. Here we pad the right side since our masks are computed based on the valid length.

text/src/autogluon/text/automm/optimization/lr_scheduler.py
```diff
@@ -0,0 +1,126 @@
+import functools
```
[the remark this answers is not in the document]
zhiqiangdon: Added.

text/src/autogluon/text/text_prediction/predictor.py
zhiqiangdon: Changed.

text/src/autogluon/text/automm/predictor.py
zhiqiangdon: Changed.

szha:
Job PR-1537-15 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1537/15/index.html

text/src/autogluon/text/automm/data/process_image.py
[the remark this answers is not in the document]
gradientsky: Yes, reflection and replication padding perform better (both are [supported in torch](https://pytorch.org/docs/stable/nn.html#padding-layers)): you don't have massive data distribution shift on the edges.
See karpathy's twitter thread on this topic: https://twitter.com/karpathy/status/720622989289644033

szha:
Job PR-1537-16 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1537/16/index.html

jwmueller:
Just a personal suggestion, but I would rename `AutoMMPredictor` to something more meaningful like `MultimodalNetworkPredictor` and in the file-paths/names would also change `automm` to something like `multimodalnet`

text/src/autogluon/text/automm/configs/environment/default.yaml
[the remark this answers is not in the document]
jwmueller: Is there a way to auto-estimate GPU memory usage and auto-reduce batch-size to ensure the training will not crash?
I presume many image+text+tabular datasets will often push the limits of GPU memory.

sxjscience:
@jwmueller Regarding the naming, would you think that AutoMM is a good umbrella name for what we are doing about deep learning + multimodal data tables + AutoML?

text/src/autogluon/text/automm/data/infer_types.py
zhiqiangdon: Done.

text/src/autogluon/text/automm/data/process_image.py
```diff
@@ -0,0 +1,175 @@
...
+            else:
+                raise ValueError("image normalization mean and std are missing")
+        else:
+            print(f"using detected image normalization: {self.mean} and {self.std}")
+        self.normalization = transforms.Normalize(self.mean, self.std)
+        if max_img_num_per_col <= 0:
+            print(f"max_img_num_per_col {max_img_num_per_col} is reset to 1")
+            max_img_num_per_col = 1
+        self.max_img_num_per_col = max_img_num_per_col
+        print(f"max_img_num_per_col: {max_img_num_per_col}")
```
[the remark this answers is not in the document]
zhiqiangdon: Replaced.

text/tests/unittests/automm/configs/model/hf_text.yaml
```diff
@@ -0,0 +1,20 @@
...
+  hf_text:
+#    checkpoint_name: "google/electra-base-discriminator"
+    checkpoint_name: 'microsoft/deberta-v3-small'
+#     checkpoint_name: 'bert-base-uncased'
+#    checkpoint_name: 'xlm-roberta-base'
+#    checkpoint_name: 'microsoft/deberta-base'
+#    checkpoint_name: 'roberta-base'
+#    checkpoint_name: "distilbert-base-uncased"
+#    checkpoint_name: "bert-base-chinese"
+#    checkpoint_name: "gpt2"
```
sxjscience, line 12: remove comments + add newline.

text/tests/unittests/automm/configs/model/numerical_mlp.yaml
```diff
@@ -0,0 +1,11 @@
...
+  names: "numerical_mlp"
+  numerical_mlp:
+    hidden_size: 128
+    activation: "leaky_relu"
+    num_layers: 1
+    drop_rate: 0.1
+    normalization: "layer_norm"
+    data_types:
+      - "numerical"
+    merge: "concat"
```
sxjscience, line 11: add newline

text/tests/unittests/automm/configs/model/timm_image.yaml
```diff
@@ -0,0 +1,23 @@
...
+      - "image"
+    train_transform_types:
+      - "resize_shorter_side"
+      - "center_crop"
+    val_transform_types:
+      - "resize_shorter_side"
+      - "center_crop"
+    image_norm: "imagenet"
+    image_size: 224
+    max_img_num_per_col: 1
```
sxjscience, line 23: remove comments + add new line

text/tests/unittests/automm/run.py
```diff
@@ -0,0 +1,130 @@
+import os
```
sxjscience, line 1: Do we need this file?

text/tests/unittests/automm/configs/model/hf_text.yaml
zhiqiangdon: Are you seeing an old commit? These were already fixed.

text/tests/unittests/automm/configs/model/numerical_mlp.yaml
zhiqiangdon: Already fixed.

text/tests/unittests/automm/configs/model/timm_image.yaml
zhiqiangdon: Already fixed.

sxjscience: Oh, I noticed that these files are in the `unittests` folder.

szha:
Job PR-1537-17 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1537/17/index.html

text/tests/unittests/automm/run.py
zhiqiangdon: removed.

text/tests/unittests/automm/configs/model/timm_image.yaml
zhiqiangdon: Thanks. Fixed this time.

text/tests/unittests/automm/configs/model/numerical_mlp.yaml
zhiqiangdon: Thanks. Fixed this time.

text/tests/unittests/automm/configs/model/hf_text.yaml
zhiqiangdon: Thanks. Fixed this time.

jwmueller:
> @jwmueller Regarding the naming, would you think that AutoMM is a good umbrella name for what we are doing about deep learning + multimodal data tables + AutoML?

Ya I do think AutoMM is a good name as long as there's a paper/blog describing the AutoMM model. Otherwise from the code alone this name seems too vague without being formally introduced in some writeup.

text/src/autogluon/text/automm/models/utils.py
zhiqiangdon: Good suggestion! Will add some introduction in the docstrings.

text/src/autogluon/text/automm/data/collator.py
zhiqiangdon: It can pad multi-dimensional input. Unit tests added.

sxjscience:
@jwmueller We may adopt the code name and write a paper about that later (potentially along with 0.5 release).

text/src/autogluon/text/automm/data/collator.py
zhiqiangdon: We also use padding for images, which have multi-dimensions. torch.nn.utils.rnn.pad_sequence seems not suitable.

szha:
Job PR-1537-18 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1537/18/index.html

docs/tutorials/text_prediction/customization.md
```diff
@@ -15,10 +15,17 @@ warnings.filterwarnings('ignore')
 np.random.seed(123)
 ```
 
+
+```{.python .input}
+!pip install ipywidgets
+!jupyter nbextension enable --py widgetsnbextension
+```
```
sxjscience, line 22: Remove these two lines.

sxjscience:
Have one minor comment. Once fixed, the PR looks good to me.

docs/tutorials/text_prediction/customization.md
zhiqiangdon: Removed.

text/src/autogluon/text/automm/configs/data/default.yaml
```diff
@@ -0,0 +1,21 @@
+data:
+  image:
+  text:
+    auto_max_length: True  # Try to automatically shrink the maximal length based on the statistics of the dataset.
+    auto_max_length_quantile: 0.95  # We will ensure that the new max_length is around the quantile of the lengths of all samples
+    auto_max_length_round_to: 32  # We will ensure that the automatically determined max length will be divisible by round_to
+  categorical:
+    minimum_cat_count: 100  # The minimal number of data per categorical group
+    maximum_num_cat: 20  # The minimal number of data per categorical group
```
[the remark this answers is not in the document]
zhiqiangdon: Thanks. Corrected.

text/src/autogluon/text/automm/configs/environment/default.yaml
```diff
@@ -0,0 +1,12 @@
+env:
+  num_gpus: -1
+  num_nodes: 1
+  batch_size: 128  # this is a desired batch size; pl trainer will accumulate gradients when per step batch is smaller.
+  per_gpu_batch_size: 16  # you can define this according to your GPU memory
+  precision: 16  # training precision
+  num_workers: 2  # pytorch training dataloader workers
+  num_workers_evaluation: 2  # pytorch prediction/test dataloader workers
```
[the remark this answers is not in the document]
zhiqiangdon: This is a because we use "ddp_spawn" as default, which can only use limited cpu workers. Otherwise, the job would get stucked. Refer to: https://pytorch-lightning.readthedocs.io/en/stable/advanced/multi_gpu.html#distributed-data-parallel-spawn

The reason we use "ddp_spawn" instead of "ddp" is to support jupyter notebook user cases. "ddp" seems not working in notebook.

szha:
Job PR-1537-24 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1537/24/index.html

sxjscience (approved):
This PR looks good to me. However, we will need to have follow-up PRs to further improve the tutorial / example / alignment between tabular and AutoMMPredictor.

text/src/autogluon/text/automm/data/process_image.py
```diff
@@ -0,0 +1,175 @@
...
+            all_image_paths: List[List[List[str]]],
+            idx: int,
+            is_training: bool,
+    ) -> dict:
+
+        per_sample_paths = [per_column_paths[idx] for per_column_paths in all_image_paths]
+        return self.process_one_sample(per_sample_paths, is_training)
+
+
+
```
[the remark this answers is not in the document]
zhiqiangdon: Removed.

text/src/autogluon/text/automm/data/process_label.py
```diff
@@ -0,0 +1,34 @@
+from typing import Optional, List, Any, Union
+from nptyping import NDArray
+from ..constants import LABEL
+from .collator import Stack
+
+
+class LabelProcessor:
```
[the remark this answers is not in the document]
zhiqiangdon: Docstrings added. This is where label customization goes since we want to fuse any combinations of models. Currently, we don't do label customization for each model. But in future this will provide us the flexibility to do so. For now, you can think this as the abstract label processing class.

szha:
Job PR-1537-25 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1537/25/index.html

Innixma (approved):
Approving because this is a special circumstance, but in future we will have to have more strict requirements to be merge-able, this form of contribution is not sustainable from a tech debt perspective.

gradientsky:
+1 to @Innixma comment; open PR questions/discussions should be resolved in the follow-up PRs (tracking doc is created)

gradientsky (approved):

PR #1198, opened 2021-06-22
Add forecasting predictor

yx1215:
*Issue #, if available:*

*Description of changes:*
Add forecasting predictor.
#
This PR is used for further discussion about add the forecasting task to Autogluon, which should be parallel to tasks such as tabular prediction and image classifications etc, replacing  #684 

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Output in ```simple_forecasting_example.py```:
```
    model     score  fit_order
0     SFF -0.617567          2
1   MQCNN -0.647109          1
2  DeepAR -0.776917          3

0.3063789898931557
```

Output in ```advanced_forecasting_example.py```:
```
              model     score  fit_order  test_score
0       SFF/trial_4 -0.771542          5   -0.455381
1       SFF/trial_5 -0.800340          6   -0.469165
2     MQCNN/trial_0 -0.859082          1   -0.819518
3     MQCNN/trial_1 -0.859113          2   -0.819403
4    DeepAR/trial_2 -0.882853          3   -0.894841
5    DeepAR/trial_3 -0.899381          4   -0.917463
6  SFF/trial_4_FULL       NaN          7   -0.559157

0.4553322822285291

                    0.5
2020-04-22   509.987183
2020-04-23   617.542175
2020-04-24  1040.601807
2020-04-25  1154.234009
2020-04-26  1073.659058
2020-04-27  1090.976196
2020-04-28   696.362183
2020-04-29  1438.374390
2020-04-30  1149.915405
2020-05-01   468.797913
2020-05-02   889.827393
2020-05-03   938.827820
2020-05-04   557.264038
2020-05-05  1205.893555
2020-05-06   379.958221
2020-05-07   616.763733
2020-05-08   877.277222
2020-05-09  1134.502930
2020-05-10   620.372009
```
Things to improve in the future:
- Ensembling 
- Dynamic features
- More detailed preprocessing for dynamic/static features.(currently the code only do auto inferring of real/cat features and fillna, other things like standardizing might also be useful)
- Making plots for predictions
- Subsampling dataset(select a range from given train data)
- Improve ways to do validation
- New cat-static features in prediction might lead to failure.

PR #1538, opened 2022-02-08
Add forecasting submodule to autogluon

canerturkmen:
*Issue #, if available:*

*Description of changes:*

Replaces #1198 as the PR for adding the `autogluon.forecasting` submodule to autogluon. The PR adds the forecasting submodule, along with packages for Learner, Predictor, Trainer, and Model objects (containing, for now, GluonTS model wrappers). 

On top of #1198, this PR rebases the earlier PR to the latest master; resolves compatibility issues with respect to HPO functionality, fixes the M4 benchmark script (after rerunning examples and benchmarks), fixes some dependency issues, fixes docs, and contains improvements to style. 

**Edit:** In agreement with core team, we will break this PR into sub-PRs, although this PR can continue serving as a reference to the snapshot of the main code and docs for autogluon.forecasting.

~~There remain several todos before the submodule can be merged: _[non-exhaustive]_~~

~~- [ ] Improving tutorials and documentation, particularly around how forecasting datasets are represented for GluonTS and autogluon.forecasting.~~
~~- [ ] Improving the implementation and API. e.g.,~~
    ~~- [ ] removing redundancies with `core` and `common` modules~~
    ~~- [ ] changing Trainers to depend on `AbstractModel` (instead of `GluonTSModel`),~~
    ~~- [ ] simplifying `Predictor` API and `fit` method.~~
~~- [ ] implement default search spaces and nondefault_params for forecasting models to adhere to new `AbstractModel` API for HPO.~~
~~- [ ] improve unit test coverage~~
~~- [ ] improve docstrings (format and coverage)~~
~~- [ ] style improvements (black everything?)~~


canerturkmen:
@huibinshen @dcmaddix for visibility.

canerturkmen:
also tagging @jwmueller @yx1215 @lovvge @Innixma

Innixma:
Thanks for opening this! I will plan to review after v0.4 release.

forecasting/src/autogluon/forecasting/models/gluonts_model/mqcnn/mqcnn_model.py
```diff
@@ -0,0 +1,39 @@
...
+                         freq=freq,
+                         prediction_length=prediction_length,
+                         hyperparameters=hyperparameters,
+                         name=name,
+                         eval_metric=eval_metric,
+                         model=model,
+                         **kwargs)
+
+    def create_model(self):
+        # if "ctx" in self.params:
```
dcmaddix, line 30: remove comments

forecasting/src/autogluon/forecasting/trainer/model_presets/presets.py
```diff
@@ -0,0 +1,147 @@
...
+from ...models.gluonts_model.mqcnn import MQCNNModel
+from ...models.gluonts_model.sff import SimpleFeedForwardModel
+from ...models.gluonts_model.deepar import DeepARModel
+from ...models.gluonts_model.auto_tabular import AutoTabularModel  # TODO: this module is not fully prepared.
+
+logger = logging.getLogger(__name__)
+
+
+MODEL_TYPES = dict(
+    MQCNN=MQCNNModel,
```
dcmaddix, line 17: Is there a way to not hard code these so that we can more easily extend to more models?

forecasting/src/autogluon/forecasting/utils/dataset_utils.py
```diff
@@ -0,0 +1,297 @@
...
+        return pd.DataFrame(data_dic)
+
+    X = reshape_dataframe(X)
+    return X
+
+# TODO: Improve the way to do the split
+def train_test_split_dataframe(data, prediction_length):
+    test_ds = data.copy()
+    train_ds = data.iloc[:, :-prediction_length]
+    if not any(data.iloc[:, -prediction_length:]):
```
dcmaddix, line 186: this is checking the test set right?

forecasting/src/autogluon/forecasting/utils/dataset_utils.py
```diff
@@ -0,0 +1,297 @@
...
+            freq = data.list_data[0]["start"].freq
+        else:
+            raise ValueError("Cannot infer freq if the data is not ListDataset.")
+
+    for entry in data:
+        test_data_lst.append(entry.copy())
+        tmp = {}
+        for k, v in entry.items():
+            if "dynamic" in k.lower():
+                tmp[k] = v[..., : -prediction_length]
```
dcmaddix, line 206: maybe use more descriptive name than `tmp`

dcmaddix (commented):
Thanks Caner! This looks great - only some minor changes for better readability.

forecasting/tests/unittests/test_forecasting.py
```diff
@@ -1,4 +1,169 @@
-import autogluon.forecasting
+import tempfile
 
-def test_forecasting():
-	print("Not Implemented")
+import autogluon.core as ag
+from autogluon.core.dataset import TabularDataset
+
+from autogluon.forecasting.predictor.predictor import ForecastingPredictor
```
jwmueller, line 6: should just be imported from autogluon.forecasting?

forecasting/tests/unittests/test_forecasting.py
```diff
@@ -1,4 +1,169 @@
-import autogluon.forecasting
+import tempfile
```
jwmueller, line 1: add TODO to reduce duplicate code?

examples/forecasting/Forecasting_Task_In_Depth.ipynb
```diff
@@ -0,0 +1,323 @@
+{
```
jwmueller, line 1: Can this be deleted since we now have in-depth tutorial in docs/tutorials/?

examples/forecasting/Forecasting_Task_Quick_Start.ipynb
```diff
@@ -0,0 +1,164 @@
+{
```
jwmueller, line 1: Can this be deleted since we now have quickstart tutorial in docs/tutorials/?

examples/forecasting/benchmark_m4.py
```diff
@@ -0,0 +1,53 @@
+import os
```
jwmueller, line 1: Add comment to explain what is being benchmarked in this file

examples/forecasting/example_custom_model_forecast.py
```diff
@@ -0,0 +1,64 @@
+""" Example script for defining and using custom models in AutoGluon Forecasting """
```
jwmueller, line 1: would rename this file: `example_custom_forecasting_model.py`

forecasting/src/autogluon/forecasting/models/abstract/abstract_model.py
```diff
@@ -0,0 +1,3 @@
+class AbstractModel:
```
jwmueller, line 1: may want to call this `AbstractForecastingModel` in the future to reduce confusion?

forecasting/src/autogluon/forecasting/utils/dataset_utils.py
```diff
@@ -0,0 +1,297 @@
...
+    'rebuild_tabular',
+    'time_series_dataset',
+    'train_test_split_dataframe',
+    'train_test_split_gluonts'
+]
+
+logger = logging.getLogger()
+
+
+class TimeSeriesDataset(ListDataset):
```
huibinshen, line 24: There are no tests for this important class at all, I would add a big todo here.

forecasting/src/autogluon/forecasting/configs/presets_configs.py
```diff
@@ -0,0 +1,40 @@
+forecasting_presets_configs = dict(
```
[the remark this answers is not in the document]
dcmaddix: +1

forecasting/src/autogluon/forecasting/learner/abstract_learner.py
```diff
@@ -0,0 +1,157 @@
...
+
+    def score(self, data, model=None, quantiles=None):
+        trainer = self.load_trainer()
+        return trainer.score(data, model=model, quantiles=quantiles)
+
+    def leaderboard(self, data=None):
+        trainer = self.load_trainer()
+        return trainer.leaderboard(data)
+
+    def save(self):
```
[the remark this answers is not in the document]
yx1215: We need to save the trainer and the learner object separately as the trainer cannot be pickled.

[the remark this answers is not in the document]
yx1215: > I am not sure I understand this save function and why it has to be in this way. Can it be simplified or add some comments on what it does?

We need to save the trainer and the learner object separately as the trainer cannot be pickled.

forecasting/src/autogluon/forecasting/models/gluonts_model/auto_tabular/auto_tabular_model.py
```diff
@@ -0,0 +1,46 @@
...
+            time_limit = max(sum(AbstractGluonTSModel.prev_fitting_time) / len(AbstractGluonTSModel.prev_fitting_time), 300)
+        if "time_limit" not in self.params:
+            self.params["time_limit"] = time_limit
+        self.model = TabularEstimator(freq=self.params["freq"],
+                                      prediction_length=self.params["prediction_length"],
+                                      time_limit=self.params["time_limit"],
+                                      last_k_for_val=self.params["prediction_length"])
+
+    def fit(self, train_data, val_data=None, time_limit=None):
+        if time_limit is None or time_limit > 0:
```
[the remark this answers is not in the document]
yx1215: > Maybe I missed something, this `if` will always be true unless `time_limit` can be negative.

Maybe you can take a look at the code [here](https://github.com/awslabs/autogluon/pull/1538/files/2e62e728d4e1efcc574c7e1e2a1e51b3387d9f6f#diff-3a6597115bab6b198283037deb8e5edef4fc7ce7ab009d6c91ccafe828855554R51), `time_limit` is reduced to `time_left` each time a model is trained and eventually it will be reduced to less than 0 and that's when `time_limit `is used up and the process will stop.

forecasting/src/autogluon/forecasting/utils/dataset_utils.py
yx1215: > this is checking the test set right?

The actual data in the test dataset is the last `prediction_length` data points of each time series. So this is checking whether the last `prediction_length` data points of each time series contain only NAN.

examples/forecasting/advanced_forecasting_example.py
```diff
@@ -0,0 +1,70 @@
...
+                                                                                 'context_length': ag.Int(70, 90,
+                                                                                                          default=prediction_length * 4),
+                                                                                 "num_batches_per_epoch": 10,
+                                                                                 "epochs": 30},
+                                                                         },
+                                                                         time_limit=60
+                                                                         )
+
+
+predictor = None
```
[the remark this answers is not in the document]
yx1215: > Why do we need to initially set it to `None`

Setting it to None and then do the loading so that we can make sure the loading function is working properly.

forecasting/src/autogluon/forecasting/utils/dataset_utils.py
dcmaddix: Yes we may want to check this logic a bit. We should ignore any time series with any `NaN` in the backtest window or at least a large portion since we can't compute metrics on time series with NaNs.
I think it would be helpful to set this to `eval_window_test_ts = data.iloc[:, -prediction_length:]` or we can call it `backtest_ts` but let's think about how to generally extend to multi backtest.

examples/forecasting/advanced_forecasting_example.py
```diff
@@ -0,0 +1,70 @@
...
+prediction_length = 19
+path = "hpo_models"
+
+predictor = ForecastingPredictor(path=path, eval_metric=eval_metric).fit(train_data,
+                                                                         prediction_length,
+                                                                         static_features=static_features,
+                                                                         index_column="name",
+                                                                         target_column="ConfirmedCases",
+                                                                         time_column="Date",
+                                                                         # hyperparameter_tune_kwargs={
```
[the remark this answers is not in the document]
yx1215: > why are these commented?

`hyperparameter_tune_kwargs` can either take a string as input to use default settings or it can take a dictionary to apply customized settings. A problem is that the customized setting for now must contain the keyword `scheduler` and `searcher`, maybe this can be optimized so that if some of the keywords are not provided we will give it a default value.

forecasting/src/autogluon/forecasting/learner/abstract_learner.py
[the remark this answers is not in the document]
huibinshen: I see, thanks! I find it can be easier to consume in the following, but not strongly. Your call here.
```Python
if self.trainer is not None and self.is_trainer_present:
    self.trainer.save()

trainer = self.trainer
self.trainer = None
save_pkl.save(path=self.save_path, object=self)
self.trainer = trainer
```

forecasting/src/autogluon/forecasting/utils/dataset_utils.py
huibinshen: +1, also maybe using more descriptive name for  `k` and `v`.

forecasting/src/autogluon/forecasting/models/gluonts_model/auto_tabular/auto_tabular_model.py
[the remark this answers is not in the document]
yx1215: > I see, thanks. Is this logic defined at line 54 in model_trial.py? If not, please point me to the line. If yes, when `time_left <= 0`, it will raise an exception so `time_left > 0` will always be true right?

I see what you mean and you seem to be correct. The time logic here is indeed a bit confusing and might need some improvements.

examples/forecasting/advanced_forecasting_example.py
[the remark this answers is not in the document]
canerturkmen: I'll keep them for now as it shows there are both alternatives

forecasting/src/autogluon/forecasting/trainer/model_presets/presets.py
canerturkmen: I understand that the intention is that these are "defaults," and will be hardcoded. Otherwise the Predictor always allows keying hyperparameter dictionaries with the types (e.g., `DeepARModel`) that inherit from `AbstractModel` so the models that can be used are not limited to these. For easier extensibility, we can think about a generic adapter around GluonTS `Estimator`s as a next step (currently all models are wrapped one by one).

examples/forecasting/Forecasting_Task_In_Depth.ipynb
canerturkmen: will be removed once docs are revised 👍

examples/forecasting/Forecasting_Task_Quick_Start.ipynb
canerturkmen: will be removed once docs are revised 👍

forecasting/src/autogluon/forecasting/utils/dataset_utils.py
canerturkmen: I'll add a todo for this, but thanks for raising this, I believe it is a bug. 
```
df = pd.DataFrame({
    "a": np.array([np.nan, np.nan, np.nan]),
    "b": np.array([np.nan, np.nan, np.nan]),
})

assert any(df)
```

forecasting/src/autogluon/forecasting/learner/abstract_learner.py
```diff
@@ -0,0 +1,157 @@
...
+class AbstractLearner:
+
+    learner_file_name = 'learner.pkl'
+    learner_info_name = 'info.pkl'
+    learner_info_json_name = 'info.json'
+
+    def __init__(self, path_context: str, eval_metric=None, is_trainer_present=False, random_seed=0):
+        self.path, self.model_context, self.save_path = self.create_contexts(path_context)
+
+        if eval_metric is not None and eval_metric not in ["MASE", "MAPE", "sMAPE", "mean_wQuantileLoss"]:
```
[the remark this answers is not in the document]
canerturkmen: @dcmaddix for now I believe this is since although evaluation is outsourced totally to GluonTS, whether "less/more is better" in the loss is encoded in `utils.metric_utils`. I made this more modular so all evaluation checks refer to this module for available metrics, and new metrics can be added easily there.

forecasting/src/autogluon/forecasting/learner/abstract_learner.py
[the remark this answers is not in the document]
canerturkmen: the logic differs between @huibinshen 's recommendation and the original (`not self.is_trainer_present`) and this version is directly from `tabular`. I'll add a todo and leave as is.

forecasting/src/autogluon/forecasting/models/gluonts_model/auto_tabular/auto_tabular_model.py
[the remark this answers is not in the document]
canerturkmen: I agree there seems to be a larger issue here. The superclass seems to keep track of "previous fitting times" in a class variable, which should not be the case seeing especially it is an abstract class. I will add todos.

canerturkmen:
Thank you everyone for the reviews. I believe all comments have either been addressed or marked clearly as todos. 

As agreed, we will break this PR up into smaller PRs, although this PR can serve as a final snapshot of the autogluon.forecasting codebase.

PR #1541, opened 2022-02-14
Added infer_limit to fit

Innixma:
*Issue #, if available:*

*Description of changes:*

- Added infer_limit to fit, which allows users to specify a maximum per-row inference time that models should have.
- Added infer_limit_batch_size, which controls the batch size to use when calculating per-row inference time.
- Fixed warning message during refit_full for CatBoost.
- Fixed bug in conformalize when dealing with refit_full presets (previously crashed or computed value incorrectly)
- Added new option to refit_full to set refit best to best model automatically (enabled by default)

Note that meaningfully unit testing this functionality would be very challenging, so I am planning to rely on end-to-end tutorials that showcase the functionality as a sanity check that it is working correctly. Eventually we should add a recurring benchmark that tracks how well this functionality is adhering to the user limits and how well it performs in terms of model quality.

Example Code:

```
# Guarantee `best_model` will have <= 50 ms online-inference speed (or raise exception if impossible)
predictor_online = TabularPredictor(label=label).fit(train_data, infer_limit=0.05, infer_limit_batch_size=1)
predictor_online.persist_models()

# will take <=50 ms
predictor_online.predict(X_1_row)

# Guarantee `best_model` will have <= 0.1 ms per-row batch inference speed (aka 10,000 rows / second throughput)
predictor_batch = TabularPredictor(label=label).fit(train_data, infer_limit=0.0001, infer_limit_batch_size=10000)
predictor_batch.persist_models()

# will take <=1 second
predictor_batch.predict(X_10000_row)
```

TODO:

- [x] Add section to in-depth tutorial
- [x] Fix unit test errors


szha:
Job PR-1541-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1541/3/index.html

szha:
Job PR-1541-5 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1541/5/index.html

core/src/autogluon/core/utils/time.py
```diff
@@ -0,0 +1,95 @@
...
+        args = list()
+    if kwargs is None:
+        kwargs = dict()
+    time_start = time.time()
+    f(*args, **kwargs)
+    time_end = time.time()
+    avg_time = time_end - time_start
+    if max_repeats == 1:
+        return avg_time
+    if time_limit is None or avg_time < (time_limit / max_repeats):
```
yinweisu, line 43: Should we instead make `time_limit` to be the time spent for 1 repeat? Otherwise, the user needs to manually calculate the total `time_limit`. Plus, user might not know the default `max_repeats` is 10 and assumed the time_limit was for 1 repeat.

yinweisu (commented):
LGTM in general! Added some comments

sxjscience:
Usually, the environment used for deployment will be different from the environment used for training. Thus, it may not be reliable to measure the inference time in the same instance. Should we measure these values offline and store it in a database?

sxjscience:
The ideal way to implement this feature is to maintain a database (or cost model) to help estimate the inference latencies. This is the strategy adopted in [HAT](https://hat.mit.edu/), [Lorien](https://dl.acm.org/doi/abs/10.1145/3472883.3486973). When we pick a new model, we can quickly estimate the inference latency by querying the database or rely our cost model.

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -855,6 +901,8 @@ def hyperparameter_tune(self, scheduler_options, time_limit=None, **kwargs):
         if 'time_out' not in scheduler_options[1]:
             scheduler_options[1]['time_out'] = time_limit
         kwargs = self.initialize(time_limit=scheduler_options[1]['time_out'], **kwargs)
+        self._register_fit_metadata(**kwargs)
+        self._validate_fit_memory_usage(**kwargs)
```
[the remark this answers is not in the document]
Innixma: This code is not being added in `get_features` (Notice there is a code block separation, this is in `hyperparameter_tune`. We call them in `hyperparameter_tune` so we can raise NotEnoughMemoryError before starting HPO rather than raising on each trial.

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -177,6 +179,14 @@ def is_valid(self) -> bool:
         """
         return self.is_fit()
 
+    def is_initialized(self) -> bool:
```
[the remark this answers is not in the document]
Innixma: This is reasonable, but the other property candidate methods such as `is_valid()` would be API breaking if we switched them to properties. We may consider switching all of them to properties in future, but it will be API breaking.

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -123,6 +124,7 @@ def __init__(self,
 
         self.fit_time = None  # Time taken to fit in seconds (Training data)
         self.predict_time = None  # Time taken to predict in seconds (Validation data)
+        self.predict_1_time = None  # Time taken to predict 1 row of data in seconds (with batch size `predict_1_batch_size` in params_aux)
```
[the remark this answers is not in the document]
Innixma: numbers in variable names are ok I think, @gradientsky feel free to comment. I generally try to avoid overly long variable names when they don't add much to the clarity.

For example, if we did this then we would also have:

```
predict_one_row_batch_size
```

which is starting to get pretty long

core/src/autogluon/core/utils/time.py
Innixma: `time_limit` is to limit the amount of time used by the call to `time_func`. I think this is more important than the user trying to specify a `time_limit` for how long the actual function they want to time takes to execute, that is the reason they are calling `time_func` in the first place. `max_repeats` is largely an implementation detail, it only matters to stabilize the estimate, and generally stabilizing the estimate is only meaningful if the function is fast to execute and thus isn't time consuming to repeatedly estimate. (Plus longer functions are likely less variable in their runtime)

Innixma:
@sxjscience I agree that these options could provide improvements, particularly the ability to calculate inference speed for separate hardware, however this could be considered in future as it is difficult to accurately calculate this due to potential dependency differences, python versions, etc. This may be better done as a pre-training step where the user identifies the speed difference between training and inference instances (say inference is 0.7x as fast), then they simply multiply their target inference speed by 0.7 when sending to the training model as `infer_limit`.

szha:
Job PR-1541-8 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1541/8/index.html

core/src/autogluon/core/models/abstract/abstract_model.py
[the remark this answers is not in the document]
yinweisu: My bad.

yinweisu (approved):
LGTM!

core/src/autogluon/core/models/abstract/abstract_model.py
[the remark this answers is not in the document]
gradientsky: Shall we create an issue to track this?

Thoughts (not for this PR): This is 'internal' APIs - only model developers will be impacted. Saying this, we probably want a migration guide for the change if we decide to go with the properties in future.

For properties `is_` prefix probably not needed: `valid` vs `is_valid`

core/src/autogluon/core/trainer/abstract_trainer.py
```diff
@@ -288,43 +288,126 @@ def train_multi_levels(self, X, y, hyperparameters: dict, X_val=None, y_val=None
...
+
+        base_model_names_copy = base_model_names.copy()
+        for base_model_name in base_model_names_copy:
+            predict_1_time_full = self.get_model_attribute_full(model=base_model_name, attribute=attribute)
+            if predict_1_time_full >= infer_limit_threshold:
+                predict_1_time_full_set_old = predict_1_time_full_set
+                base_model_names.remove(base_model_name)
+                predict_1_time_full_set = self.get_model_attribute_full(model=base_model_names, attribute=attribute)
+                if verbose:
+                    messages_to_log.append(f'\t{round(predict_1_time_full_set_old, 4)}s -> {round(predict_1_time_full_set, 4)}s ({base_model_name})')
```
gradientsky, line 356: General comment: this method is too long: this section looks like a good candidate for extraction + tests + docs whats is happening

core/src/autogluon/core/trainer/abstract_trainer.py
```diff
@@ -288,43 +288,126 @@ def train_multi_levels(self, X, y, hyperparameters: dict, X_val=None, y_val=None
...
+            i += 1
+            predict_1_time_full_set = self.get_model_attribute_full(model=base_model_names, attribute=attribute)
+            if verbose:
+                messages_to_log.append(f'\t{round(predict_1_time_full_set_old, 4)}s -> {round(predict_1_time_full_set, 4)}s ({base_model_to_remove})')
+
+        if messages_to_log:
+            logger.log(20, f'Removing {len(messages_to_log)}/{num_models_og} base models to satisfy inference constraint '
+                           f'(constraint={round(infer_limit_threshold, 4)}s) ...')
+            for msg in messages_to_log:
+                logger.log(20, msg)
```
gradientsky, line 375: ditto; + please add 1-liner to describe intended algorithm here (so if we have to touch the code, there's no need for deep-dive).

core/src/autogluon/core/trainer/abstract_trainer.py
```diff
@@ -1104,6 +1215,29 @@ def _add_model(self, model: AbstractModel, stack_name: str = 'core', level: int
...
                     raise AssertionError(f"Model '{model.name}' depends on model '{base_model_name}', but '{base_model_name}' is not in a lower stack level. ('{model.name}' level: {level}, '{base_model_name}' level: {self.model_graph.nodes[base_model_name]['level']})")
                 self.model_graph.add_edge(base_model_name, model.name)
+        if model.val_score is not None:
+            if model.eval_metric.name != self.eval_metric.name:
+                logger.log(20, f'\tNote: model has different eval_metric than default.')
+            logger.log(20, f'\t{round(model.val_score, 4)}\t = Validation score   ({model.eval_metric.name})')
+        if model.fit_time is not None:
+            logger.log(20, f'\t{round(model.fit_time, 2)}s\t = Training   runtime')
+        if model.predict_time is not None:
+            logger.log(20, f'\t{round(model.predict_time, 2)}s\t = Validation runtime')
```
gradientsky, line 1225: refactor to separate method - this block adds lots of visual noise when reading code

core/src/autogluon/core/trainer/abstract_trainer.py
```diff
@@ -1104,6 +1215,29 @@ def _add_model(self, model: AbstractModel, stack_name: str = 'core', level: int
...
+                # TODO: Consider moving this to a dedicated function and using across package for nicer logging of time
+                time_unit = 'ms'
+                predict_1_time = predict_1_time*1000
+            logger.log(20, f'\t{round(predict_1_time, 3)}{time_unit}\t = Validation runtime (1 row | {predict_1_batch_size} batch size)')
+            predict_1_time_full = self.get_model_attribute_full(model=model.name, attribute='predict_1_time')
+            time_unit = "s"
+            if round(predict_1_time_full, 2) == 0:
+                time_unit = 'ms'
+                predict_1_time_full = predict_1_time_full * 1000
+            logger.log(20, f'\t{round(predict_1_time_full, 3)}{time_unit}\t = Validation runtime (1 row | {predict_1_batch_size} batch size | FULL)')
```
gradientsky, line 1240: logging/units selection code duplication - almost same logging with different args

core/src/autogluon/core/utils/time.py
```diff
@@ -0,0 +1,95 @@
+import logging
+import time
+from typing import Optional
+
+from pandas import DataFrame
+
+logger = logging.getLogger(__name__)
+
+
+def time_func(f, args: list = None, kwargs: dict = None, time_limit: float = 0.2, max_repeats: int = 10) -> float:
```
gradientsky, line 10: We should use python's [`timeit()`](https://docs.python.org/3/library/timeit.html#basic-examples):
```python
import timeit
time = timeit.timeit(lambda: f(x), number=100)
```

tabular/src/autogluon/tabular/learner/default_learner.py
```diff
@@ -61,13 +65,34 @@ def _fit(self, X: DataFrame, X_val: DataFrame = None, X_unlabeled: DataFrame = N
...
         time_preprocessing_start = time.time()
         logger.log(20, 'Preprocessing data ...')
         self._pre_X_rows = len(X)
         if self.groups is not None:
             num_bag_sets = 1
             num_bag_folds = len(X[self.groups].unique())
+        if infer_limit_batch_size is not None:
+            X_og = X
+        else:
+            X_og = None
```
gradientsky, line 78: nit: one-liner
```python
X_og = None if infer_limit_batch_size is None else X
```

tabular/src/autogluon/tabular/learner/default_learner.py
```diff
@@ -61,13 +65,34 @@ def _fit(self, X: DataFrame, X_val: DataFrame = None, X_unlabeled: DataFrame = N
...
         X, y, X_val, y_val, X_unlabeled, holdout_frac, num_bag_folds, groups = self.general_data_processing(X, X_val, X_unlabeled, holdout_frac, num_bag_folds)
+        if infer_limit_batch_size is not None:
+            X_og_1 = sample_df_for_time_func(df=X_og, sample_size=infer_limit_batch_size)
+            infer_limit_batch_size_actual = len(X_og_1)
+            self.preprocess_1_time = time_func(f=self.transform_features, args=[X_og_1]) / infer_limit_batch_size_actual
```
gradientsky, line 83: Consider timeit

core/src/autogluon/core/models/abstract/abstract_model.py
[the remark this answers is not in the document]
Innixma: Probably doesn't need its own issue ATM. If we are to API break AbstractModel, we'd probably want to do a lot of API-break related changes at once. If we do we would of course want to have a migration guide.

core/src/autogluon/core/trainer/abstract_trainer.py
Innixma: What do you mean by 1-liner to describe algorithm? I have detailed method documentation detailing implementation strategy.

core/src/autogluon/core/trainer/abstract_trainer.py
Innixma: Full method is 45 lines, with only ~20 lines of non-boilerplate/logging related logic, not that long imo. Documentation is already present in method doc-string. I agree it could be unit tested, but this is hard without extra work because this links into the DAG metadata. We may consider this when working on refactor of AbstractTrainer into a bunch of mixin classes.

Innixma: If you'd like I can add in-line comments for the two stages of this method to describe what is happening.

Innixma: added two in-line comments to better describe the 2 logical sections

core/src/autogluon/core/trainer/abstract_trainer.py
Innixma: added 1-liner

core/src/autogluon/core/trainer/abstract_trainer.py
Innixma: Agreed, moved to new method

core/src/autogluon/core/utils/time.py
Innixma: Added TODO. I don't want to use timeit without further testing as time_func works in a very particular way to avoid wasting too much time and I'd like to avoid too much fixed time-sink into estimating time calls. Not urgent, so we can consider post v0.4.0

tabular/src/autogluon/tabular/learner/default_learner.py
Innixma: updated

tabular/src/autogluon/tabular/learner/default_learner.py
Innixma: see other comment regarding timeit

gradientsky (approved):

szha:
Job PR-1541-9 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1541/9/index.html

PR #1547, opened 2022-02-19
[PR #1357 patch 1] Fix Tabular tutorials for Pytorch TextPredictor

zhiqiangdon:
*This is to fix small issues when using the new Pytorch TextPredictor in the Tabular tutorials.*

- Removed the legacy "presets" of MXNet TextPredictor in tabular-gpu.md and added new configuration introduction.
- Removed one legacy preset string "medium_quality_faster_train" in autogluon.tabular.configs.hyperparameter_configs.
- Resolved one bug of converting pandas' category column to text data.
- Supported presets for Pytorch TextPredictor.
- Capped the package versions.


szha:
Job PR-1547-5 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1547/5/index.html

sxjscience:
We may try to enable the SwinTransformer + ELECTRA-small model for the PetFinder dataset since it's essentially directly integrating AutoMMPredictor

tabular/src/autogluon/tabular/configs/hyperparameter_configs.py
```diff
@@ -81,7 +81,7 @@
         'CAT': {},
         'XGB': {},
         # 'FASTAI': {},  # FastAI gets killed if the dataset is large (400K rows).
-        'AG_TEXT_NN': ['medium_quality_faster_train'],  # TODO, Support changing the config w.r.t the preset option.
+        'AG_TEXT_NN': {},  # TODO, Support changing the config w.r.t the preset option.
```
sxjscience, line 84: We can consider to support the presets.

docs/tutorials/tabular_prediction/tabular-gpu.md
```diff
@@ -41,19 +41,47 @@ hyperparameters
...
+    "model": "/path/to/model/yaml/file",
+    "data": "/path/to/data/yaml/file",
+    "optimization": "/path/to/optimization/yaml/file",
+    "environment": "/path/to/optimization/yaml/file",
+}
 ```
 
-Available text model presets:
+
+Text model uses the following configurations as default.
```
sxjscience, line 56: We may ask the user to refer to another document page about how to customize the text predictor and multimodal predictor.

text/src/autogluon/text/automm/data/preprocess_dataframe.py
```diff
@@ -175,8 +175,9 @@ def fit(
             elif col_type == CATEGORICAL:
                 if self._config["categorical"]["convert_to_text"]:
                     # Convert categorical column as text column
-                    processed_data = col_value.apply(lambda ele: '' if ele is None else str(ele))
-                    if len(np.unique(processed_data)) == 1:
+                    col_value = col_value.astype("object")
```
sxjscience, line 178: We may mark a TODO here. We need to investigate if this is the optimal choice.

sxjscience: Basically, we can consider to treat it as categorical value if the column has the "category" type in pandas. However, this may require some further investigation so we may mark it with `#TODO(...) ...`

zhiqiangdon: If self._config["categorical"]["convert_to_text"] is True, we should convert categorical data to text since it is what the config requires. Users can set it False if not wanting this conversion.

tabular/src/autogluon/tabular/configs/hyperparameter_configs.py
zhiqiangdon: If we use name "AG_MM_NN" to indicate Text Predictor, there will be conflicts in naming the Pytorch-based Image Predictor and Image+Text Predictor. I think using name "AG_TEXT_NN" may be a better choice.

zhiqiangdon: The previous presets have three choices: low quality, medium quality, and best quality, corresponding to three backbones: electra small, base, and large. These seem not that necessary since we already provide the backbone customization functionality. Users can try any huggingface text backbones. They shouldn't just rely on the naive preset rule.

sxjscience: After a second thought, I agree that keeping AG_TEXT_NN is good. Regarding presets, we should add some documentations in the customization tutorial to recommend some configuration choices to the user (e.g. configurations for fast inference and training). It is non-trivial for beginners to search for the huggingface backbone names and set those keys.

docs/tutorials/tabular_prediction/tabular-gpu.md
```diff
@@ -41,19 +41,47 @@ hyperparameters
 
 ### Text models
```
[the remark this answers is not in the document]
sxjscience: We may add another section to teach user how to use the experimental AutoMMPredictor in tabular.

text/src/autogluon/text/automm/data/preprocess_dataframe.py
sxjscience: We keep categorical values to text if those categorical labels has meanings. We may mark a TODO about how we can do it better.

zhiqiangdon:
Enabling the SwinTransformer + ELECTRA-small model in the tabular tutorial needs to add a new model class in the tabular code and other possible code changes in its trainer/learner. I think we can leave this to next version. For now, we can change the customization tutorial in text to introduce how to configure AutoMMPredictor.

docs/tutorials/tabular_prediction/tabular-gpu.md
zhiqiangdon: Agree. We can change the customization.md tutorial to do this.

docs/tutorials/tabular_prediction/tabular-gpu.md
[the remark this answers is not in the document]
zhiqiangdon: We can add a tutorial introducing how to customize the configurations of AutoMMPredictor. But directly using AutoMMPredictor in tabular requires code changes in tabular's source code. I think we may leave this to the next version.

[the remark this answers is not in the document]
sxjscience: Will it require code changes in tabular? I thought we can just enable it by setting “model” key to the image_text_tabular one.

tabular/src/autogluon/tabular/configs/hyperparameter_configs.py
zhiqiangdon: Agree. We can add some recommendations in the tutorial.

docs/tutorials/tabular_prediction/tabular-gpu.md
```diff
@@ -41,19 +41,47 @@ hyperparameters
...
+Text model's configuration can be set via:
 
 ```{.python}
-hyperparameters['AG_TEXT_NN'] = ['<preset>']
+hyperparameters['AG_TEXT_NN'] = {
+    "model": "/path/to/model/yaml/file",
+    "data": "/path/to/data/yaml/file",
+    "optimization": "/path/to/optimization/yaml/file",
+    "environment": "/path/to/optimization/yaml/file",
+}
```
sxjscience, line 52: I noticed that later you mentioned to override certain keys with 'model.hf_text.checkpoint_name': 'google/electra-small-discriminator'. What’s the difference between these two ways of setting the hyperparameters?

text/setup.py
```diff
@@ -18,10 +18,10 @@
...
 install_requires = [
     # version ranges added in ag.get_dependency_version_ranges()
-    'numpy',
-    'scipy',
-    'pandas',
-    'scikit-learn',
+    'numpy<=1.22',
+    'scipy<1.8.0',
+    'pandas<1.4',
+    'scikit-learn<1.1',
```
sxjscience, line 24: For version cap, the general rule is to make the supported version strictly smaller than the next minor version,e.g. xxx<1.2. For common packages, these are covered in https://github.com/awslabs/autogluon/blob/master/core/src/autogluon/core/_setup_utils.py#L16

docs/tutorials/tabular_prediction/tabular-gpu.md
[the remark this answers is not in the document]
zhiqiangdon: Yes, it requires code changes. It seems that tabular needs a corresponding class created inside here:

https://github.com/awslabs/autogluon/tree/master/tabular/src/autogluon/tabular/models

And some presets:

https://github.com/awslabs/autogluon/blob/master/tabular/src/autogluon/tabular/trainer/model_presets/presets.py

https://github.com/awslabs/autogluon/blob/master/tabular/src/autogluon/tabular/configs/hyperparameter_configs.py

Not sure whether these are all the places requiring changes.

sxjscience:
This PR looks good to me once the version cap is fixed. The general rule is to make the supported version strictly smaller than the next minor version,e.g. xxx<1.2. For common packages, these are covered in https://github.com/awslabs/autogluon/blob/master/core/src/autogluon/core/_setup_utils.py#L16

szha:
Job PR-1547-6 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1547/6/index.html

sxjscience (approved):

tabular/src/autogluon/tabular/configs/hyperparameter_configs.py
yinweisu: I think presets does not conflict with customization. It's just some preselected customization and there shouldn't be a reason we don't support it. It is critical for beginners

yinweisu: I think presets does not conflict with customization. It's just some preselected customization and there shouldn't be a reason we don't support it. It is critical for beginners

docs/tutorials/tabular_prediction/tabular-gpu.md
yinweisu: This is actually an API breaking change. User's old code will no longer work with newer version of AG. We should still support preset along with the custom configuration.

text/setup.py
```diff
@@ -22,19 +22,19 @@
...
-    'transformers>=4.16.2',
     'Pillow',
     'tqdm',
-    'nptyping',
-    'omegaconf',
-    'timm',
-    'sentencepiece',
+    'fairscale<0.4.6',
+    'scikit-image<0.19.3',
+    'boto3<1.21.4',
```
yinweisu, line 29: Why do we need boto3?

text/src/autogluon/text/automm/data/preprocess_dataframe.py
yinweisu: I believe the feature generator is already handling this logic. We should consider align it in the future.

yinweisu (commented):
Added some comments

text/src/autogluon/text/automm/data/preprocess_dataframe.py
```diff
@@ -175,8 +175,9 @@ def fit(
...
+                    processed_data = col_value.apply(lambda ele: '' if pd.isnull(ele) else str(ele))
+                    if len(processed_data.unique()) == 1:
```
Innixma, line 180: Just a note: This is automatically handled in feature generator (removing useless columns)

text/src/autogluon/text/automm/data/preprocess_dataframe.py
```diff
@@ -175,8 +175,9 @@ def fit(
             elif col_type == CATEGORICAL:
                 if self._config["categorical"]["convert_to_text"]:
                     # Convert categorical column as text column
-                    processed_data = col_value.apply(lambda ele: '' if ele is None else str(ele))
-                    if len(np.unique(processed_data)) == 1:
+                    col_value = col_value.astype("object")
+                    processed_data = col_value.apply(lambda ele: '' if pd.isnull(ele) else str(ele))
```
Innixma, line 179: Note: This is automatically handled in feature generator (text NaN -> empty str)

docs/tutorials/tabular_prediction/tabular-gpu.md
```diff
@@ -41,19 +41,47 @@ hyperparameters
...
-ag_text_presets.list_keys()
+from omegaconf import OmegaConf
+from autogluon.text.automm.utils import get_config
+config = get_config({
+    "model": "fusion_mlp_text_tabular",
+    "data": "default",
+    "optimization": "adamw",
+    "environment": "default",
+})
+OmegaConf.to_container(config)
```
Innixma, line 68: This would need to be changed if the default ever changed, consider instead creating a method that returns whatever the default config is (very similar if not the same as the presets logic)

docs/tutorials/tabular_prediction/tabular-gpu.md
```diff
@@ -41,19 +41,47 @@ hyperparameters
 
 ### Text models
 
-Text model preset to use can be set via:
+Text model's configuration can be set via:
```
Innixma, line 44: [!!!] Potential bug:

[Logs](http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1547/6/tutorials/tabular_prediction/tabular-multimodal.html) shows TextPredictor getting 0.3 val score in training logs, but in Tabular leaderboard / val score it only has 0.29 (worse).

Is this because of the weight averaging at end of fit? @sxjscience I believe we talked about this before, but I still recommend we check val score of weight averaged model and use non-weight averaged version if weighted averaging makes val score worse.

text/setup.py
```diff
@@ -22,19 +22,19 @@
...
+    'smart_open<5.2.2',
+    'pytorch_lightning<1.5.11',
+    'torchmetrics<0.7.3',
+    'transformers<4.16.3',
+    'nptyping<1.4.5',
+    'omegaconf<2.1.2',
+    'timm<0.5.5',
+    'sentencepiece<0.1.97',
```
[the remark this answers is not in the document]
sxjscience: +1 on this. @zhiqiangdon I think we need to ensure `pytorch_lightning>=1.5.9`.

docs/tutorials/tabular_prediction/tabular-gpu.md
sxjscience: @zhiqiangdon Need to double check the cause.

sxjscience:
Regarding tutorial, I think we can add an experimental tutorial about the new AutoMMPredictor that is similar to https://auto.gluon.ai/stable/tutorials/tabular_prediction/tabular-multimodal.html . We may also use the PetFinder dataset. (This should be in a follow-up PR as the major focus of this PR is bugfix).

docs/tutorials/tabular_prediction/tabular-gpu.md
Innixma: Could also go further by evaluating multiple weight average checkpoints: top-2, top-3, top-2-weight-by-val-score, top-3-weight-by-val-score, even HPO on weights of top-N if the process to do so is fast.

text/setup.py
sxjscience: I feel that depend on boto3 is fine. It's actually a dependency of autogluon.common. It is used in the test-cases in text.

docs/tutorials/tabular_prediction/tabular-gpu.md
sxjscience: The risk is that it may give us worse performance in the test set (since we are tuning it on the validation set). The purpose of weight averaging is to ensure that the model generalize better. Nevertheless, we can add a flag about this behavior and verify it with benchmark.

text/src/autogluon/text/automm/data/preprocess_dataframe.py
zhiqiangdon: Yes, considering the current time constraint, we may need to align the API after 0.4 release.

text/src/autogluon/text/automm/data/preprocess_dataframe.py
zhiqiangdon: We can change this after API alignment. We'd better keep it since automm currently doesn't use the feature generator.

text/src/autogluon/text/automm/data/preprocess_dataframe.py
zhiqiangdon: Ditto. Will change after API alignment.

text/src/autogluon/text/automm/data/preprocess_dataframe.py
sxjscience: I think we may pass the column types generated via tabular to the text in the integration: https://github.com/awslabs/autogluon/blob/c0e33c76b62db887684fe9b607b34420d6bb9b1e/tabular/src/autogluon/tabular/models/text_prediction/text_prediction_v1_model.py#L144-L150. It will guarantee the consistency.

sxjscience: We can use the `feature_metadata` https://github.com/awslabs/autogluon/blob/c0e33c76b62db887684fe9b607b34420d6bb9b1e/tabular/src/autogluon/tabular/models/text_prediction/text_prediction_v1_model.py#L53

text/setup.py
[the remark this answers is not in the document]
zhiqiangdon: If we use pytorch_lightning>=1.5.9 and pytorch_lightning<1.5.11, why don't we just set pytorch_lightning==1.5.10? Currently, we don't test on different versions, so not sure what's a good range.

docs/tutorials/tabular_prediction/tabular-gpu.md
zhiqiangdon: Do you think adding a AutoMM tutorial is a solution? The previous three presets were only temporary since they provide users very limited flexibility.

docs/tutorials/tabular_prediction/tabular-gpu.md
zhiqiangdon: Agree.

szha:
Job PR-1547-8 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1547/8/index.html

tabular/src/autogluon/tabular/configs/hyperparameter_configs.py
zhiqiangdon: Code updated. Preset is supported now.

docs/tutorials/tabular_prediction/tabular-gpu.md
zhiqiangdon: Will a tutorial in the next PR.

zhiqiangdon: The preset is back now.

text/setup.py
zhiqiangdon: Both minimum and maximum are capped now.

text/setup.py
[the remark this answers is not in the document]
zhiqiangdon: Capped on both sides now.

docs/tutorials/tabular_prediction/tabular-gpu.md
zhiqiangdon: Preset is supported now.

szha:
Job PR-1547-9 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1547/9/index.html

szha:
Job PR-1547-10 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1547/10/index.html

text/src/autogluon/text/text_prediction/text_presets.py
```diff
@@ -0,0 +1,54 @@
...
+    model_type = "fusion_mlp_text_tabular"
+    config = get_preset(model_type)
+    preset = preset.lower()
+    if preset == "default" or preset == "medium_quality_faster_train":
+        overrides = {
+            "model.hf_text.checkpoint_name": "google/electra-base-discriminator",
+        }
+    elif preset == "lower_quality_fast_train":
+        overrides = {
+            "model.hf_text.checkpoint_name": "google/electra-small-discriminator",
```
sxjscience, line 42: For ELECTRA-small, use `lr=4E-4`.

text/src/autogluon/text/text_prediction/text_presets.py
```diff
@@ -0,0 +1,54 @@
...
+        }
+    elif preset == "best_quality":
+        overrides = {
+            "model.hf_text.checkpoint_name": "google/electra-large-discriminator",
```
sxjscience, line 46: For ELECTRA-large, use `batch_size=2`.

text/src/autogluon/text/text_prediction/text_presets.py
```diff
@@ -0,0 +1,54 @@
...
+    """
+    Returns
+    -------
+    A list of naive presets available in AutoGluon-Text.
+    """
+    simple_presets = [
+        "default", "lower_quality_fast_train",
+        "medium_quality_faster_train", "best_quality",
+    ]
+    return simple_presets
```
sxjscience, line 14: Consider to use the registry design later.https://github.com/awslabs/autogluon/blob/master/text/src/autogluon/text/text_prediction/presets.py You may directly copy `registry.py` in https://github.com/sxjscience/autogluon-contrib-nlp/blob/master/src/autogluon_contrib_nlp/utils/registry.py to AutoGluon. The corresponding test-case is in https://github.com/dmlc/gluon-nlp/blob/master/tests/test_utils_registry.py .

text/src/autogluon/text/text_prediction/text_presets.py
zhiqiangdon: Done.

text/src/autogluon/text/text_prediction/text_presets.py
zhiqiangdon: Done.

szha:
Job PR-1547-12 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1547/12/index.html

text/src/autogluon/text/automm/presets.py
```diff
@@ -0,0 +1,66 @@
...
+    """
+    List all available model types.
+    Image/text backbones can be customized for one model type.
+
+    Returns
+    -------
+    A list of model types.
+    """
+    cur_path = os.path.dirname(os.path.abspath(__file__))
+    model_config_dir = os.path.join(cur_path, "configs/model")
```
yinweisu, line 20: Change to `os.path.join(cur_path, "configs", "model")` Otherwise, this won't work with windows

text/src/autogluon/text/automm/presets.py
```diff
@@ -0,0 +1,66 @@
...
+
+    if model_type == "text":
+        config[MODEL] = "hf_text"
+    elif model_type == "image":
+        config[MODEL] = "timm_image"
+    else:
+        config[MODEL] = model_type
+
+        cur_path = os.path.dirname(os.path.abspath(__file__))
+        model_config_dir = os.path.join(cur_path, "configs/model")
```
yinweisu, line 58: ditto

yinweisu (commented):
LGTM in general! Added a tiny comment.

text/src/autogluon/text/automm/presets.py
zhiqiangdon: Done.

text/src/autogluon/text/automm/presets.py
zhiqiangdon: Done.

yinweisu (approved):

szha:
Job PR-1547-14 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1547/14/index.html

Issue #1554, opened 2022-02-23
Trial regression test suite


willsmithorg:
I am testing out a TabularPredictor regression test suite with the following aims:

- runs very fast, potentially fast enough to run after every checkin (max 10 minutes), or at least nightly
- runs for a wide range of commonly use parameters (no params, presets='best' etc)
- both regression and classification problems
- both all-numeric and mixed numeric/categorical features
- tolerant to floating point precision differences between architectures etc.
- easy to identify what's changed
- if a change is detected and accepted by the user, it should be easy to update the tests.  Ideally the failing test should dump out new config so to make it pass.
- ideally, tests can be run by users and are confirmed to pass on multiple common configurations (windows/linux, different python versions etc)

I will attach a trial PR to this issue, so I can get it running on the github automated tests to compare with my own results.

PR #1555, opened 2022-02-23
Tabular regressiontests

willsmithorg:
*Issue #, if available:*
#1554 

*Description of changes:*
- Regression testing of model list and scores in tabular on small synthetic datasets (for speed).
- Tests about 20 different calls to `TabularPredictor()` on both regression and classification tasks, multiple presets etc. 
- When a test fails it dumps out the config change required to make it pass, for ease of updating.


szha:
Job PR-1555-7 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1555/7/index.html

szha:
Job PR-1555-8 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1555/8/index.html

szha:
Job PR-1555-9 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1555/9/index.html

Innixma:
This is an awesome contribution!!

@yinweisu How could we get this to run on a nightly test but not in the CI for PRs?

yinweisu (commented):
Thanks for the contribution!!
Since this test has already been marked, we can modify the `conftest` a little bit to make it not run in default. For example:
```python
def pytest_addoption(parser):
    parser.addoption(
        "--runslow", action="store_true", default=False, help="run slow tests"
    )
    parser.addoption(
        "--runregression", action="store_true", default=False, help="run regression tests"
    )


def pytest_configure(config):
    config.addinivalue_line("markers", "slow: mark test as slow to run")
    config.addinivalue_line("markers", "regression: mark test as regression test")


def pytest_collection_modifyitems(config, items):
    skip_slow = pytest.mark.skip(reason="need --runslow option to run")
    skip_regression = pytest.mark.skip(reason="need --runregression option to run")
    custom_markers = dict(
        slow=skip_slow,
        regression=skip_regression
    )
    if config.getoption("--runslow"):
        # --runslow given in cli: do not skip slow tests
        custom_markers.pop("slow", None)
    if config.getoption("--runregression"):
        # --runregression given in cli: do not skip slow tests
        custom_markers.pop("regression", None)

    for item in items:
        for marker in custom_markers:
            if marker in item.keywords:
                item.add_marker(custom_markers[marker])
```
After this test is merged in, I can work on adding a nightly test.

Innixma:
@willsmithorg I've checked both the test failures and they are valid changes. Can you update the expected configs so tests pass? Thanks!

willsmithorg:
Regression tests updated so everything passes.
If this runs once per night, what is the process to review the results?

Innixma:
> If this runs once per night, what is the process to review the results?

Likely it would be manual at first (manually checking if the tests are passing on the nightly).

Later on we would probably like to have some kind of alarm system to notify maintainers when it fails.
Finally, it can be incorporated into a release readiness checklist, so that the tests must pass prior to making a new release.

We could also add logic to run the tests on a particular PR if we think it is relevant to the PR's changes. @yinweisu I believe we already did this kind of logic for the Windows and MacOS tests right? Can we run a subset of tests via this method?

Innixma (approved):

yinweisu:
Yes. We should be able to use ChatOP for that too.

Innixma:
Merged, thanks again for the contribution!

Issue #991, opened 2021-03-02
Fix TextPredictorModel during refit_full


Innixma:
Calling refit_full on a TextPredictorModel can take overly long since it doesn't use the epochs trained by the original model. This should be fixed for users to use good and high quality presets for faster inference compared to best_quality.

sxjscience:
In fact, I added one test case in https://github.com/awslabs/autogluon/blob/7af68d6d80ff7dfe6e03f9555e85491cdfde9c7a/tabular/tests/unittests/models/test_text_prediction_v1_model.py#L8-L10. May need to investigate what causes the failure.

Innixma:
I looked into code user mentioned failure in, error was actually caused by #994, so the issue isn't TextPredictor. Still, it will be good eventually to handle refit_full correctly to respect time_limit and ensure refit_full version isn't significantly different from non-refit_full version in terms of accuracy.

PR #1567, opened 2022-02-24
Improved refit_full support, refactored refit_full logic

Innixma:
*Issue #, if available:*
#991 

*Description of changes:*
This PR addresses 4 issues:

1. Previously, for models TextPredictorModel and ImagePredictorModel when used in Tabular, if the user called refit_full, the refit models would ignore the best epoch found in the models and instead train all epochs. This could result in much worse model quality and could take many times longer to train than expected.
2. Additionally, models that already are effectively `_FULL` such as RandomForest in bag mode would be trained again as an identical copy, instead of simply copying the original model (same end result, avoids doubling training time).
3. Finally, new models contributed by users had to address the refit_full functionality of their model, which often was very complex (the most complex part of the entire contribution). This shouldn't be necessary for a basic contribution.
4. I have also included minor updates and improvements to logging clarity.

In this PR, I have implemented a new model tag `can_refit_full` which defaults to False. Models which have this tag as False will not use the previous refit_full logic when refit. Instead, they will use a generic fallback logic that is safe from overfitting / hyperparameter drift issues: Simply copy the original bag, but only keep the first fold model. Similarly for non-bag, it will simply copy the original model in its entirety. This will avoid needing to train the model again during refit as it is just a copy, and doesn't have all the complexities of having to train again with no validation data (requires advanced knowledge from user).

This means new model contributions will no longer need to:

1. Support `_fit` call where X_val and y_val are None.
2. Pass final hyperparameter information to refit_full model training (such as best epoch).
3. Know that refit_full exists.

Advanced contributors can still add support for `can_refit_full=True`, which is a superior solution in terms of model quality, but this is generally not critical for initial contributions.

### DIFF
--- a/core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
+++ b/core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
@@ -348,6 +348,8 @@ class BaggedEnsembleModel(AbstractModel):
         model_base.reduce_memory_size(remove_fit=True, remove_info=False, requires_save=True)
         if not self.params.get('save_bag_folds', True):
             model_base.model = None
+        if self.low_memory:
+            self.save_child(model_base)
         self.add_child(model=model_base, add_child_times=True)
         self._set_n_repeat_single()
 
@@ -480,7 +482,7 @@ class BaggedEnsembleModel(AbstractModel):
 
         for model in models:
             # No need to add child times or save child here as this already occurred in the fold_fitting_strategy
-            self.add_child(model=model, add_child_times=False, save_child=False, verbose=False)
+            self.add_child(model=model, add_child_times=False)
         self._bagged_mode = True
 
         if self._oof_pred_proba is None:
@@ -642,7 +644,7 @@ class BaggedEnsembleModel(AbstractModel):
         else:
             return model
 
-    def add_child(self, model: Union[AbstractModel, str], add_child_times=False, save_child=True, verbose=False):
+    def add_child(self, model: Union[AbstractModel, str], add_child_times=False):
         """
         Add a new fit child model to `self.models`
 
@@ -653,10 +655,6 @@ class BaggedEnsembleModel(AbstractModel):
         add_child_times : bool, default = False
             Whether to add child metadata on times to the bag times.
             This includes fit_time, predict_time, and predict_1_time.
-        save_child : bool, default = True
-            Whether to save the child model to disk.
-        verbose : bool, default = False
-            Whether to log the child saving process.
         """
         if self.models is None:
             self.models = []
@@ -666,10 +664,6 @@ class BaggedEnsembleModel(AbstractModel):
         else:
             model_name = model.name
         if self.low_memory:
-            if save_child:
-                if model is None:
-                    model = self.load_child(model=model_name, verbose=False)
-                self.save_child(model, verbose=verbose)
             self.models.append(model_name)
         else:
             if model is None:
@@ -681,6 +675,7 @@ class BaggedEnsembleModel(AbstractModel):
             self._add_child_times_to_bag(model=model)
 
     def save_child(self, model: Union[AbstractModel, str], verbose=False):
+        """Save child model to disk."""
         child = self.load_child(model)
         child.set_contexts(self.path + child.name + os.path.sep)
         child.save(verbose=verbose)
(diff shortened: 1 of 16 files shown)


szha:
Job PR-1567-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1567/1/index.html

core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
```diff
@@ -622,14 +635,52 @@ def get_features(self):
...
+        if isinstance(model, str):
+            model_name = model
+            model = None
+        else:
+            model_name = model.name
+        if self.low_memory:
+            if save_child:
+                if model is None:
+                    model = self.load_child(model=model_name, verbose=False)
+                self.save_child(model, verbose=verbose)
```
gradientsky, line 672: [design/naming] From single responsibility principle it's not a good idea to bring save into this method. Saving should be done separately, outside of `add_child` method. Not to mention this is done only in low_memory + save_child section and not in the other ones.

tabular/src/autogluon/tabular/models/lgb/lgb_model.py
```diff
@@ -340,3 +340,7 @@ def _features(self):
 
     def _ag_params(self) -> set:
         return {'ag.early_stop'}
+
+    def _more_tags(self):
+        # `can_refit_full=True` because num_boost_round is communicated at end of `_fit`
+        return {'can_refit_full': True}
```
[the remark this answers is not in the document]
Innixma: `_more_tags` is the method name in sklearn, this logic is taken from sklearn and how it handles tags for its models.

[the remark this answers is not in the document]
Innixma: We may consider adding a mix-in, but would rather wait until we add many mix-ins when we start breaking down AbstractModel to better support Forecast. This way we can have a better idea how many mix-ins is acceptable / what is worthy of being a mix-in.

It is defined in parent class : But parent class is set to False by default, which is the setting for beginner contributors / those who don't know about refit_full.

szha:
Job PR-1567-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1567/2/index.html

core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
Innixma: saving model and adding model to self.models are linked, because with low_memory setting self.models appends the name, but in low_memory=False, self.models appends the model. We wouldn't want to save the model + add the model to self.models generally speaking.

Do you have a proposal for how it should be done instead?

Innixma: If we passed low_memory to `save_child`, then `save_child` would do nothing when `low_memory=False`. This is probably not what we want.

yinweisu: I feel it's ok to do nothing in `save_child` if `low_memory=False`. Basically, save_child will be the function to save the model if needed and not saving it otherwise. @gradientsky what's your thought?

szha:
Job PR-1567-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1567/3/index.html

core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
gradientsky: > saving model and adding model to self.models are linked, because with low_memory setting self.models appends the name, but in low_memory=False, self.models appends the model. We wouldn't want to save the model + add the model to self.models generally speaking.

They don't need to live in the same method. Separating them reduces cyclomatic complexity of the method and makes unit tests for each of these more concise.

Doing nothing in save_child is ok. The call to `save_child()` can be made after checking `low_memory` and `save_child` outside.

Innixma: I don't necessarily fully agree that it should be separated, but I have done so and it is now revised to address this.

gradientsky (approved):

Issue #1469, opened 2021-12-31
[v0.4] Refactor model valid features to be allow instead of exclude


Innixma:
Currently, most models specify the feature types they don't support and accept all other feature types.

Instead, refactor to have models specify the feature types they support and reject all other feature types.

This simplifies development, particularly for new contributors, as they only need to worry about the list of features they allow instead of understanding everything they must reject. This also makes adding new feature types easier, as existing models won't need to be modified to reject the new features when they are added.

Innixma:
Merged in #1569

PR #1569, opened 2022-02-25
Updated models to primarily use include-lists instead of exclude-lists

Innixma:
Updated models to primarily use include-lists instead of exclude-lists

*Issue #, if available:*
#1469 

*Description of changes:*

- Use include-list for specifying valid model feature types as a best practice.
- No functional change at present, but reduces defects in future when adding new types and makes contributing models simpler (no need to understand all types, only ones that are relevant to the model).


szha:
Job PR-1569-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1569/1/index.html

yinweisu (approved):
LGTM!

PR #1571, opened 2022-02-25
[PR #1537 Patch 4] Separate users' evaluation metric and the internal validation metric

zhiqiangdon:
*This PR is to separate users' provided evaluation metric and the internal validation metric. If a user provides metric `r2`, we will use `rmse` as the internal validation metric since `torchmetrics.R2Score` encounters errors for batch size 1.*

1. Always keep the evaluation metric provided by users. It can be used in `.evaluate()`.
2. Use a separate internal validation metric, which may be different from the provided evaluation metric.
3. Rename the names of evaluation and validation metrics so that their names are both distinguishable and accurate, especially in the context of deep learning.
4. Clean some unused log keys.

### DIFF
--- a/text/src/autogluon/text/automm/predictor.py
+++ b/text/src/autogluon/text/automm/predictor.py
@@ -39,7 +39,7 @@ from .utils import (
     compute_score,
     gather_top_k_ckpts,
     average_checkpoints,
-    infer_eval_metric,
+    infer_validation_metric,
     get_config,
 )
 from .optimization.utils import (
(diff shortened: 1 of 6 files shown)


text/src/autogluon/text/automm/predictor.py
```diff
@@ -356,11 +349,19 @@ def fit(
...
+            )
+        else:
+            validation_metric_name = self._validation_metric_name
+
+        validation_metric, minmax_mode = get_metric(
+            metric_name=validation_metric_name,
+            num_classes=output_shape,
         )
-        custom_metric_func = get_custom_metric_func(self._eval_metric_name)
+        custom_metric_func = get_custom_metric_func(validation_metric_name)
```
sxjscience, line 364: I feel that this function can be combined into `get_metric`. Basically, for each validation_metric_name, it will return the aggregator and the custom_metric_func.

zhiqiangdon: If merged, most metrics need to return None for custom_metric_func, which seems weird.

sxjscience: In the current implementation, the `get_metric` and `get_custom_metric_func` will be highly coupled. If you are going to add a new metric, you will need to change the implementation of both functions. Instead, you may just implement the 'custom_metric_func' feature in `get_metric` so the future contributors will only need to revise the `get_metric` function.

zhiqiangdon: Merged.

sxjscience (approved):

szha:
Job PR-1571-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1571/2/index.html

szha:
Job PR-1571-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1571/3/index.html

PR #1576, opened 2022-02-28
numpy downgrade - py3.7 is required

gradientsky:
*Description of changes:*

numpy downgrade - py3.7 is required
open CVEs present, but fixes in 1.22.x is python 3.8 only: CVE-2021-41496 | CVE-2021-34141


Innixma (approved):
LGTM!

PR #1580, opened 2022-03-02
Fix web format

yinweisu:
*Issue #, if available:*

*Description of changes:*
* Add badges
* Fix badge and table format. Table now has correct alignment 
* Table still has word break problem. This should be caused by d2lbook as sphinx rendering even with mxtheme doesn't have this problem. Upon checking the html source code, a single word is sometimes break into two part. This might be addressed in a separate pr if necessary, where we migrate from d2l to other method, i.e. sphinx + nbnotebook

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1,6 +1,25 @@
 AutoGluon: AutoML for Text, Image, and Tabular Data
 ====================================================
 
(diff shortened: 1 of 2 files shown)


szha:
Job PR-1580-4 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1580/4/index.html

Innixma (approved):
LGTM, thanks for looking into this!!

PR #1583, opened 2022-03-03
Add holiday columns and encoding cyclical continuous features

Kota28:
*Issue #, if available:*

*Description of changes:*
I added holiday columns that judge the date is a holiday or not based on Internationalization services.
Moreover, I added [Encoding cyclical continuous features](https://ianlondon.github.io/blog/encoding-cyclical-features-24hour-time/)


Innixma:
Thanks for the contribution! We are currently in the process of releasing v0.4.0, expected by March 9th, so we will wait to review until after v0.4.0 is released.

features/src/autogluon/features/generators/datetime.py
```diff
@@ -1,7 +1,8 @@
 import logging
-
 import pandas as pd
 from pandas import DataFrame
+from datetime import date
+import holidays
```
Innixma, line 5: unit test error: No module named holidays

features/src/autogluon/features/generators/datetime.py
```diff
@@ -39,42 +39,54 @@ def _transform(self, X: DataFrame) -> DataFrame:
...
+                # Best guess is currently to fill by the mean.
+                fillna_datetime = datetime_series.mean()
+                fillna_map[datetime_feature] = fillna_datetime
         return fillna_map
 
     # TODO: Improve handling of missing datetimes
     def _generate_features_datetime(self, X: DataFrame) -> DataFrame:
-        X_datetime = DataFrame(index=X.index)
+        X_datetime = X.copy()
+        country = locale.getlocale()[0].split("_")[1] #　Obtain the country used to determine if it is a holiday or not.
```
Innixma, line 58: If this gets the locale of the machine, this is dangerous. If someone is training on the cloud, the locale could be different than what they intend. I'd prefer that we don't use this kind of information as it means reproducing results will be impossible between two different regions.

features/src/autogluon/features/generators/datetime.py
```diff
@@ -39,42 +39,54 @@ def _transform(self, X: DataFrame) -> DataFrame:
...
-            (R_DATETIME, None),
-            (None, [S_DATETIME_AS_OBJECT])
-        ])
+        return dict()
 
     def _compute_fillna_map(self, X: DataFrame):
         fillna_map = dict()
         for datetime_feature in self.features_in:
-            datetime_series = pd.to_datetime(X[datetime_feature], errors='coerce')
+            if type(X[datetime_feature].iloc[0]) == pd.Timestamp:
```
Innixma, line 47: Why adding if statements? What if the type is not timestamp?

features/src/autogluon/features/generators/datetime.py
```diff
@@ -12,15 +13,14 @@
...
     Parameters
     ----------
     features : list, optional
         A list of datetime features to parse out of dates.
         For a full list of options see the methods inside pandas.Series.dt at https://pandas.pydata.org/docs/reference/api/pandas.Series.html
     """
-    def __init__(self, 
-            features: list = ['year', 'month', 'day', 'dayofweek'],
+    def __init__(self,
+            features: list = ['year', 'month', 'day', 'dayofweek', 'hour'],
```
Innixma, line 23: Let's not have hour be used by default, it is often just noise.

features/src/autogluon/features/generators/datetime.py
```diff
@@ -39,42 +39,54 @@ def _transform(self, X: DataFrame) -> DataFrame:
...
+                # Most of the pandas Series.dt properties are here, a few are omitted (e.g. is_month_start) if they can be inferred
+                # from other features.
+                for feature in self.features:
+                    X_datetime[datetime_feature + '.' + feature] = getattr(X_datetime[datetime_feature].dt, feature).astype(int)
+
+                for i in range(len(X_datetime)):
+                    X_datetime[datetime_feature + "_holiday"].iloc[i] = int(date(X_datetime[datetime_feature + ".year"].iloc[i], X_datetime[datetime_feature + ".month"].iloc[i], X_datetime[datetime_feature + ".day"].iloc[i]) in holidays_list)
+                X_datetime[datetime_feature] = pd.to_numeric(X_datetime[datetime_feature])
+
+                X_datetime["dow_hour"] = X_datetime[datetime_feature + ".dayofweek"] + X_datetime[datetime_feature + ".hour"] / 24
```
Innixma, line 77: This logic should be toggle-able. All of these things should have parameters that allow them to be enabled/disabled, as they are not always desirable.

Innixma, line 77: This crashes if hour is not in self.features

features/src/autogluon/features/generators/datetime.py
```diff
@@ -12,15 +13,14 @@
...
-
     Parameters
     ----------
     features : list, optional
         A list of datetime features to parse out of dates.
         For a full list of options see the methods inside pandas.Series.dt at https://pandas.pydata.org/docs/reference/api/pandas.Series.html
     """
-    def __init__(self, 
-            features: list = ['year', 'month', 'day', 'dayofweek'],
+    def __init__(self,
```
Innixma, line 22: Need to update / add unit tests or else CI won't pass.

features/src/autogluon/features/generators/datetime.py
```diff
@@ -39,42 +39,54 @@ def _transform(self, X: DataFrame) -> DataFrame:
...
         return X_datetime
+            
 
     def _remove_features_in(self, features: list):
         super()._remove_features_in(features)
         if self._fillna_map:
             for feature in features:
                 if feature in self._fillna_map:
-                    self._fillna_map.pop(feature)
+                    self._fillna_map.pop(feature)
```
Innixma, line 92: nit: add EOF newline

Innixma (commented):
Added initial review

Kota28:
Thank you for your review comments.
I would like to correct code based on your comments.

Thank you for your continued support.

features/src/autogluon/features/generators/datetime.py
Kota28: I added holidays library in setup.py

features/src/autogluon/features/generators/datetime.py
Kota28: In deciding which country should be used, I used environment variables.

features/src/autogluon/features/generators/datetime.py
Kota28: Deleted if statement

features/src/autogluon/features/generators/datetime.py
Kota28: Deleted "hour"

features/src/autogluon/features/generators/datetime.py
Kota28: I set toggles that let users choose whether they use holidays and sincos encoding or not.

features/setup.py
```diff
@@ -21,6 +21,7 @@
     'numpy',
     'pandas',
     'scikit-learn',
+    'holidays'
```
myui, line 24: typo. `,` is missing.

Kota28: Thank you for pointing that out.

features/src/autogluon/features/generators/datetime.py
```diff
@@ -47,17 +46,27 @@ def get_default_infer_features_in_args() -> dict:
...
-            # Best guess is currently to fill by the mean.
-            fillna_datetime = datetime_series.mean()
-            fillna_map[datetime_feature] = fillna_datetime
+                # Best guess is currently to fill by the mean.
+                fillna_datetime = datetime_series.mean()
+                fillna_map[datetime_feature] = fillna_datetime
         return fillna_map
+    
+    def _get_country(self): # Obtain the country used to determine if it is a holiday or not.
+        if os.environ.get('LANG') == None:
```
myui, line 58: How about introducing `AG_COUNTRY`?  Also, you can use default return for getenv.

```
os.getenv('AG_COUNTRY', default='US')
```

LANG can have LANG=C and overriding LANG env is sometimes not preferred.
`os.environ.get('LANG').split("_")[1][0:2]` fails where `LANG=C`.

szha:
Job PR-1583-25 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1583/25/index.html

szha:
Job PR-1583-27 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1583/27/index.html

features/src/autogluon/features/generators/datetime.py
```diff
@@ -67,10 +79,19 @@ def _generate_features_datetime(self, X: DataFrame) -> DataFrame:
...
+            
+            if holiday_encoding == True:
+                for i in range(len(X_datetime)):
+                    X_datetime[datetime_feature + "_holiday"].iloc[i] = int(date(X_datetime[datetime_feature + ".year"].iloc[i], X_datetime[datetime_feature + ".month"].iloc[i], X_datetime[datetime_feature + ".day"].iloc[i]) in holidays_list)
+                X_datetime[datetime_feature] = pd.to_numeric(X_datetime[datetime_feature])
+            
+            if sincos_encoding == True: # add Encoding cyclical continuous features
+    # https://ianlondon.github.io/blog/encoding-cyclical-features-24hour-time/
+                X_datetime[datetime_feature + '.dayofweek'] = np.sin(2 * np.pi * X_datetime[datetime_feature + '.dayofweek'] / 7.0) 
+                X_datetime[datetime_feature + '.dayofweek'] = np.cos(2 * np.pi * X_datetime[datetime_feature + '.dayofweek'] / 7.0)
```
Innixma, line 91: We are just overwriting the value here? Isn't the idea to have 2 features computed from this? They should have different names. Perhaps:

```
X_datetime[datetime_feature + '.dayofweek_sin'] = np.sin(2 * np.pi * X_datetime[datetime_feature + '.dayofweek'] / 7.0) 
X_datetime[datetime_feature + '.dayofweek_cos'] = np.cos(2 * np.pi * X_datetime[datetime_feature + '.dayofweek'] / 7.0)
X_datetime = X_datetime.drop(columns=[datetime_feature + '.dayofweek'])
```

features/src/autogluon/features/generators/datetime.py
```diff
@@ -53,11 +59,17 @@ def _compute_fillna_map(self, X: DataFrame):
...
+    
+    def _get_country(self): # Obtain the country used to determine if it is a holiday or not.
+        return os.getenv('AG_COUNTRY', default='US')
 
     # TODO: Improve handling of missing datetimes
-    def _generate_features_datetime(self, X: DataFrame) -> DataFrame:
-        X_datetime = DataFrame(index=X.index)
+    def _generate_features_datetime(self, X: DataFrame, holiday_encoding=True, sincos_encoding=True) -> DataFrame:
+        X_datetime = X.copy()
+        country = self._get_country()
```
Innixma, line 69: We need to cache the country at fit time so we use the same country at test time regardless of the location of the machine doing inference. For example, we need to set it as: `self._country`. This can be set in the same location self._fillna_map is set.

features/src/autogluon/features/generators/datetime.py
```diff
@@ -53,11 +59,17 @@ def _compute_fillna_map(self, X: DataFrame):
...
         return fillna_map
+    
+    def _get_country(self): # Obtain the country used to determine if it is a holiday or not.
+        return os.getenv('AG_COUNTRY', default='US')
 
     # TODO: Improve handling of missing datetimes
-    def _generate_features_datetime(self, X: DataFrame) -> DataFrame:
-        X_datetime = DataFrame(index=X.index)
+    def _generate_features_datetime(self, X: DataFrame, holiday_encoding=True, sincos_encoding=True) -> DataFrame:
+        X_datetime = X.copy()
```
Innixma, line 68: Why are we copying X? We didn't need to do this before.

features/src/autogluon/features/generators/datetime.py
```diff
@@ -1,7 +1,10 @@
 import logging
-
+import os
+import holidays
```
Innixma, line 3: lazily import holidays within `_generate_features_datetime` and only when `self.holiday_encoding=True`, this way the user won't crash if they don't have holidays installed and they aren't using the holiday's functionality.

features/setup.py
```diff
@@ -21,6 +21,7 @@
     'numpy',
     'pandas',
     'scikit-learn',
+    'holidays',
```
Innixma, line 24: need to version cap. The others don't appear version capped but they are since we use a global version cap logic across shared dependencies: https://github.com/awslabs/autogluon/blob/master/core/src/autogluon/core/_setup_utils.py

Probably `'holidays>=0.13,<0.14'`

features/src/autogluon/features/generators/datetime.py
```diff
@@ -12,21 +15,24 @@
...
+            sincos_encoding = True,
             **kwargs
     ):
         super().__init__(**kwargs)
         self.features = features
+        self.holiday_encoding = holiday_encoding
+        self.sincos_encoding = sincos_encoding
 
-    def _fit_transform(self, X: DataFrame, **kwargs) -> (DataFrame, dict):
+    def _fit_transform(self, X: DataFrame, **kwargs):
```
Innixma, line 35: Why get rid of type hints?

features/tests/features/generators/test_pipeline.py
```diff
@@ -48,15 +50,17 @@ def test_pipeline_feature_generator(generator_helper, data_helper):
...
-            'datetime_as_object',
+            # 'datetime.dayofweek',
+            # 'datetime_as_object',
             'datetime_as_object.year',
             'datetime_as_object.month',
-            'datetime_as_object.day',
-            'datetime_as_object.dayofweek'
+            'datetime_as_object.day'
+            # 'datetime_as_object.day',
+            # 'datetime_as_object.dayofweek'
```
Innixma, line 63: Where is holiday?

features/src/autogluon/features/generators/datetime.py
```diff
@@ -53,11 +59,17 @@ def _compute_fillna_map(self, X: DataFrame):
...
             fillna_map[datetime_feature] = fillna_datetime
         return fillna_map
+    
+    def _get_country(self): # Obtain the country used to determine if it is a holiday or not.
+        return os.getenv('AG_COUNTRY', default='US')
 
     # TODO: Improve handling of missing datetimes
-    def _generate_features_datetime(self, X: DataFrame) -> DataFrame:
-        X_datetime = DataFrame(index=X.index)
+    def _generate_features_datetime(self, X: DataFrame, holiday_encoding=True, sincos_encoding=True) -> DataFrame:
```
Innixma, line 67: This ignores `self.holiday_encoding` and `self.sincos_encoding`, even if I set those values to False that wouldn't change the behavior.

features/src/autogluon/features/generators/datetime.py
```diff
@@ -67,10 +79,19 @@ def _generate_features_datetime(self, X: DataFrame) -> DataFrame:
...
             for feature in self.features:
                 X_datetime[datetime_feature + '.' + feature] = getattr(X_datetime[datetime_feature].dt, feature).astype(int)
-
-            X_datetime[datetime_feature] = pd.to_numeric(X_datetime[datetime_feature])
-
+            
+            if holiday_encoding == True:
+                for i in range(len(X_datetime)):
+                    X_datetime[datetime_feature + "_holiday"].iloc[i] = int(date(X_datetime[datetime_feature + ".year"].iloc[i], X_datetime[datetime_feature + ".month"].iloc[i], X_datetime[datetime_feature + ".day"].iloc[i]) in holidays_list)
+                X_datetime[datetime_feature] = pd.to_numeric(X_datetime[datetime_feature])
```
Innixma, line 86: Move holiday logic to a separate method (ideally staticmethod)

features/src/autogluon/features/generators/datetime.py
```diff
@@ -67,10 +79,19 @@ def _generate_features_datetime(self, X: DataFrame) -> DataFrame:
...
+            
+            if sincos_encoding == True: # add Encoding cyclical continuous features
```
Innixma, line 88: move sincos_encoding to a separate method (ideally staticmethod)

features/src/autogluon/features/generators/datetime.py
```diff
@@ -53,11 +59,17 @@ def _compute_fillna_map(self, X: DataFrame):
...
 
     # TODO: Improve handling of missing datetimes
-    def _generate_features_datetime(self, X: DataFrame) -> DataFrame:
-        X_datetime = DataFrame(index=X.index)
+    def _generate_features_datetime(self, X: DataFrame, holiday_encoding=True, sincos_encoding=True) -> DataFrame:
+        X_datetime = X.copy()
+        country = self._get_country() 
+        holidays_list = holidays.country_holidays(country)
         for datetime_feature in self.features_in:
+            X_datetime[datetime_feature + "_holiday"] = 0
```
Innixma, line 72: Don't do this unless holiday_encoding == True, this shouldn't be necessary until after the `if holiday_encoding == True` check.

features/setup.py
Innixma, line 24: We want to be extremely selective with which packages we install by default in `autogluon.features`. We probably want to move holidays to be an optional dependency not installed by default, with the holidays logic not enabled by default.

features/src/autogluon/features/generators/datetime.py
myui: @Innixma 

Is [_feature_generator_kwargs](https://github.com/awslabs/autogluon/blob/aab0dbef6de2132b1a74c92ecb93201e06f8f38f/tabular/src/autogluon/tabular/predictor/predictor.py#L652) right place for the toggle switch?  If it's not, let us know.
We consider adding constructor arguments to DatetimeFeatureGenerator based on [kwargs](https://github.com/awslabs/autogluon/blob/aab0dbef6de2132b1a74c92ecb93201e06f8f38f/features/src/autogluon/features/generators/auto_ml_pipeline.py#L75).

If `holiday_encoding(=jp)` is specified in `_feature_generator_kwargs`, try to import holiday package. Then, add holiday feature if holiday packages exists something like [try_import_mxboard](https://github.com/awslabs/autogluon/blob/fbdce82b4cdcaa4a1f2513c984e5aed91bbe1b51/core/src/autogluon/core/utils/try_import.py#L20
). Exclude holiday from setup.py.

Similary, enable adding sincos feature if `sincos_encoding(=True)` exists `_feature_generator_kwargs`.

BTW, @Kota28 is an intern student working at our company.

Issue #1358, opened 2021-10-28
[v0.4] Fix CatBoost on GPU


Innixma:
Callbacks and custom metrics are not supported on GPU for CatBoost:

```
_catboost.CatBoostError: User defined loss functions, metrics and callbacks are not supported for GPU
```

CatBoost GitHub Issue: https://github.com/catboost/catboost/issues/1792

- [x] Add a work-around that avoids using callbacks in CatBoost if using GPU. This may mean returning to the logic in v0.3.1 prior to callback usage.

PR #1587, opened 2022-03-05
Fix CatBoost on GPU

Innixma:
*Issue #, if available:*
#1358 

*Description of changes:*
- Fix CatBoost on GPU, previously crashed.
- This is a hack workaround because GPU CatBoost doesn't support callbacks.
- I've added the logic we used in v0.3.1 back in for GPU which trains a dummy model to estimate memory and training time usage.
- This isn't ideal but there is no better solution until CatBoost addresses issues with callbacks in GPU mode.

### DIFF
--- a/tabular/src/autogluon/tabular/models/catboost/catboost_model.py
+++ b/tabular/src/autogluon/tabular/models/catboost/catboost_model.py
@@ -159,21 +159,42 @@ class CatBoostModel(AbstractModel):
 
         logger.log(15, f'\tCatboost model hyperparameters: {params}')
 
+        extra_fit_kwargs = dict()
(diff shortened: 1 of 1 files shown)


yinweisu (approved):
LGTM! A comment in general: shall we create an issue to track this so that we don't forget about it? Or we might just leave the old one open

Innixma:
@yinweisu Good point! Added a new GitHub issue: https://github.com/awslabs/autogluon/issues/1590

szha:
Job PR-1587-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1587/1/index.html

PR #1591, opened 2022-03-05
Use timm instead of timm-clean

Innixma:
*Issue #, if available:*

*Description of changes:*

- timm-clean is not maintained and uses an old version of timm, better to use newer version of timm that aligns with autogluon.text.


yinweisu (approved):
This makes sense. Previously, timm-clean was used because gluoncv doesn't require the user to install torch while timm has torch as a dependency. It's ok for autogluon since we install torch anyway

szha:
Job PR-1591-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1591/3/index.html

PR #1599, opened 2022-03-08
[Tabular] Fix crash in FastAI

Innixma:
*Issue #, if available:*

*Description of changes:*

Fixes crash in FastAI preprocessing when cat_cols_to_drop is not empty since it would be a numpy array and starting in numpy 1.21 checking `if cat_cols_to_drop` will error.

```
  File "/repo/frameworks/AutoGluon/lib/autogluon/tabular/src/autogluon/tabular/models/fastainn/tabular_nn_fastai.py", line 134, in _preprocess
    if cat_cols_to_drop:
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```


gradientsky (approved):

szha:
Job PR-1599-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1599/1/index.html

PR #1603, opened 2022-03-09
Add deprecation warnings for MXNet NN

Innixma:
*Issue #, if available:*

*Description of changes:*

- Add deprecation warnings for MXNet NN
- Instead of raising exception when user specifies 'NN' in hyperparameters, instead warn and treat as 'NN_MXNET'. This avoids breaking scripts which specify 'NN' in custom hyperparameters dictionaries when upgrading to v0.4.0.


gradientsky (approved):

yinweisu (approved):
LGTM

PR #1628, opened 2022-03-25
TimeSeriesDataFrame

huibinshen:
*Issue #, if available:*

This is part of the #1538

*Description of changes:*

Added `TimeSeriesDataFrame` class to represent time-series dataset for autogluon.forecasting.

### DIFF
--- a/forecasting/src/autogluon/forecasting/dataset/ts_dataframe.py
+++ b/forecasting/src/autogluon/forecasting/dataset/ts_dataframe.py
@@ -3,25 +3,39 @@ from __future__ import annotations
 from typing import Any, Tuple
 
 import pandas as pd
+import numpy as np
 from gluonts.dataset.common import ListDataset
 
 
 class TimeSeriesDataFrame(pd.DataFrame):
 
     def __init__(self, data: Any, *args, **kwargs):
-        # infer if data is a known type and attempt to cast to dataframe
-        # for example:
         if isinstance(data, ListDataset):
-            # todo: validate data in ListDataset
-            # validate time series structure
-            # assign indexes (index by item id and timestamp)
             data = self.from_gluonts(data)
-
-        # todo: validate data DataFrame format
+        self.validate(data)
         super().__init__(data=data, *args, **kwargs)
 
-        # todo: infer freq
-        self.freq = None  # infer frequency string
+    @classmethod
+    def sample_dataframe(cls):
+        target = list(range(9))
+        datetime_index = tuple(pd.date_range(pd.Timestamp("01-01-2019"), periods=3, freq='D'))
+        item_ids = (0, 1, 2)
+        multi_index = pd.MultiIndex.from_product([item_ids, datetime_index], names=['item_id', 'timestamp'])
+        ts_df = pd.Series(target, name='target', index=multi_index).to_frame()
+        return TimeSeriesDataFrame(ts_df)
+
+    @classmethod
+    def validate(cls, data):
+        if not isinstance(data, pd.DataFrame):
+            raise ValueError(f"data must be a pd.DataFrame, get {type(data)}")
+        if not isinstance(data.index, pd.MultiIndex):
+            raise ValueError(f"data must have pd.MultiIndex, get {type(data.index)}")
+        if 'target' not in data.columns:
+            raise ValueError(f"data must have a column called target")
+        if not data.index.dtypes.array[0] == np.dtype(np.int64):
+            raise ValueError(f"for item_id, the only NumPy dtype allowed is ‘np.int64’.")
+        if not data.index.dtypes.array[1] == np.dtype('datetime64[ns]'):
+            raise ValueError(f"for timestamp, the only NumPy dtype allowed is ‘datetime64[ns]’.")
 
     @classmethod
     def from_gluonts(
@@ -57,3 +71,4 @@ class TimeSeriesDataFrame(pd.DataFrame):
     ) -> TimeSeriesDataFrame:
         nanosecond_before_end = end - pd.Timedelta(nanoseconds=1)
         return TimeSeriesDataFrame(self.loc[(slice(None), slice(start, nanosecond_before_end)), :])
+
--- a/forecasting/src/autogluon/forecasting/dataset/ts_dataframe.py
+++ b/forecasting/src/autogluon/forecasting/dataset/ts_dataframe.py
@@ -1,74 +1,211 @@
 from __future__ import annotations
 
-from typing import Any, Tuple
+from typing import Any, Tuple, Dict, List
+from dataclasses import dataclass
 
 import pandas as pd
 import numpy as np
-from gluonts.dataset.common import ListDataset
+
+
+@dataclass
+class TimeSeriesListDataset:
+    data_iter: List[Dict[str, Any]]
+    freq: str
 
 
 class TimeSeriesDataFrame(pd.DataFrame):
+    """TimeSeriesDataFrame to represent time-series dataset.
+
+    Parameters
+    ----------
+    data : Any
+        Time-series data to construct a TimeSeriesDataFrame.
+        It currently supports two formats:
+            1. Time-series data in TimeSeriesListDataset format.
+            2. Time-series data pd.DataFrame format with multi-index on item and timestamp.
+            An example can be found using example_dataframe() function.
+    """
 
     def __init__(self, data: Any, *args, **kwargs):
-        if isinstance(data, ListDataset):
-            data = self.from_gluonts(data)
-        self.validate(data)
+        if isinstance(data, TimeSeriesListDataset):
+            data = self.from_list_dataset(data)
+        self._validate_data_frame(data)
         super().__init__(data=data, *args, **kwargs)
 
     @classmethod
-    def sample_dataframe(cls):
+    def example_dataframe(cls):
+        """An example TimeSeriesDataFrame.
+
+        Returns
+        -------
+        ts_df : TimeSeriesDataFrame
+            It returns an example TimeSeriesDataFrame as:
+                                    target
+            item_id timestamp
+            0       2019-01-01       0
+                    2019-01-02       1
+                    2019-01-03       2
+            1       2019-01-01       3
+                    2019-01-02       4
+                    2019-01-03       5
+            2       2019-01-01       6
+                    2019-01-02       7
+                    2019-01-03       8
+        """
+
         target = list(range(9))
-        datetime_index = tuple(pd.date_range(pd.Timestamp("01-01-2019"), periods=3, freq='D'))
+        datetime_index = tuple(
+            pd.date_range(pd.Timestamp("01-01-2019"), periods=3, freq="D")
+        )
         item_ids = (0, 1, 2)
-        multi_index = pd.MultiIndex.from_product([item_ids, datetime_index], names=['item_id', 'timestamp'])
-        ts_df = pd.Series(target, name='target', index=multi_index).to_frame()
-        return TimeSeriesDataFrame(ts_df)
+        multi_index = pd.MultiIndex.from_product(
+            [item_ids, datetime_index], names=["item_id", "timestamp"]
+        )
+        ts_df = TimeSeriesDataFrame(pd.Series(target, name="target", index=multi_index).to_frame())
+        return ts_df
 
     @classmethod
-    def validate(cls, data):
+    def _validate_data_frame(cls, data: pd.DataFrame):
+        """Validate a pd.DataFrame can be converted to TimeSeriesDataFrame
+
+        Parameters:
+        -----------
+        data : pd.DataFrame
+            a data frame in pd.DataFrame format.
+        """
+
         if not isinstance(data, pd.DataFrame):
             raise ValueError(f"data must be a pd.DataFrame, get {type(data)}")
         if not isinstance(data.index, pd.MultiIndex):
             raise ValueError(f"data must have pd.MultiIndex, get {type(data.index)}")
-        if 'target' not in data.columns:
+        if "target" not in data.columns:
             raise ValueError(f"data must have a column called target")
         if not data.index.dtypes.array[0] == np.dtype(np.int64):
-            raise ValueError(f"for item_id, the only NumPy dtype allowed is ‘np.int64’.")
-        if not data.index.dtypes.array[1] == np.dtype('datetime64[ns]'):
-            raise ValueError(f"for timestamp, the only NumPy dtype allowed is ‘datetime64[ns]’.")
+            raise ValueError(
+                f"for item_id, the only NumPy dtype allowed is ‘np.int64’."
+            )
+        if not data.index.dtypes.array[1] == np.dtype("datetime64[ns]"):
+            raise ValueError(
+                f"for timestamp, the only NumPy dtype allowed is ‘datetime64[ns]’."
+            )
 
     @classmethod
-    def from_gluonts(
-            cls, list_dataset: ListDataset
-    ) -> TimeSeriesDataFrame:
+    def _validate_list_dataset(cls, list_dataset):
+        if not isinstance(list_dataset, TimeSeriesListDataset):
+            raise ValueError(f"list_dataset must be a TimeSeriesListDataset, get {type(list_dataset)}")
+        if len(list_dataset.data_iter) == 0:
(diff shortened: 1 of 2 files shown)


huibinshen:
@canerturkmen @Innixma @gradientsky @dcmaddix @jwmueller 
cc @RSNirwan @lostella @jaheba @yx1215

forecasting/src/autogluon/forecasting/dataset/ts_dataframe.py
```diff
@@ -0,0 +1,226 @@
...
+        if not data.index.dtypes.array[1] == np.dtype("datetime64[ns]"):
+            raise ValueError(
+                f"for timestamp, the only NumPy dtype allowed is ‘datetime64[ns]’."
+            )
+
+    @classmethod
+    def _validate_list_dataset(cls, list_dataset):
+        if not isinstance(list_dataset, TimeSeriesListDataset):
+            raise ValueError(
+                f"list_dataset must be a TimeSeriesListDataset, get {type(list_dataset)}"
```
canerturkmen, line 98: nitpick: maybe "got" or "not"?

canerturkmen: similarly for below

forecasting/src/autogluon/forecasting/dataset/ts_dataframe.py
```diff
@@ -0,0 +1,226 @@
+from __future__ import annotations
```
canerturkmen, line 1: Very cool!! Thanks Huibin!

canerturkmen (changes requested):

forecasting/src/autogluon/forecasting/dataset/ts_dataframe.py
```diff
@@ -0,0 +1,226 @@
...
+        Time-series data to construct a TimeSeriesDataFrame.
+        It currently supports two formats:
+            1. Time-series data in TimeSeriesListDataset format.
+            2. Time-series data pd.DataFrame format with multi-index on item and timestamp.
+            An example can be found using example_dataframe() function.
+    """
+
+    def __init__(self, data: Any, *args, **kwargs):
+        if isinstance(data, TimeSeriesListDataset):
+            data = self.from_list_dataset(data)
```
jaheba, line 31: Couldn't we implement `as_dataframe` on `TimeSeriesListDataset` instead?

forecasting/src/autogluon/forecasting/dataset/ts_dataframe.py
```diff
@@ -0,0 +1,226 @@
...
+        self._validate_data_frame(data)
+        super().__init__(data=data, *args, **kwargs)
+
+    @classmethod
+    def example_dataframe(cls):
```
jaheba, line 36: What's the idea behind this? Why not just name it `example()`?

forecasting/src/autogluon/forecasting/dataset/ts_dataframe.py
canerturkmen: I think we never want to instantiate / pass around `TimeSeriesListDatasets`, but this just serves as a Type. So I agree with Jasper's point on `UserList`, but not with implementing `as_dataframe`. `TimeSeriesListDataset` will never be depended on by the main library.

forecasting/src/autogluon/forecasting/dataset/ts_dataframe.py
```diff
@@ -0,0 +1,226 @@
...
+        Parameters:
+        -----------
+        data : pd.DataFrame
+            a data frame in pd.DataFrame format.
+        """
+
+        if not isinstance(data, pd.DataFrame):
+            raise ValueError(f"data must be a pd.DataFrame, get {type(data)}")
+        if not isinstance(data.index, pd.MultiIndex):
+            raise ValueError(f"data must have pd.MultiIndex, get {type(data.index)}")
```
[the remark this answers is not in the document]
huibinshen: Make sense. How about having a method `from_dataframe(df)` where `df` is simply a normal dataframe with columns `item_id` and `timestamp`?

forecasting/src/autogluon/forecasting/dataset/ts_dataframe.py
```diff
@@ -0,0 +1,226 @@
...
+        if "target" not in data.columns:
+            raise ValueError(f"data must have a column called target")
+        if not data.index.dtypes.array[0] == np.dtype(np.int64):
```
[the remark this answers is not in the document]
huibinshen: The numeric index seems to be 64 bits in pandas, another possible choice is `np.uint64`. This is what I found: https://stackoverflow.com/questions/44090944/how-to-change-index-dtype-of-pandas-dataframe-to-int32

forecasting/src/autogluon/forecasting/dataset/ts_dataframe.py
```diff
@@ -0,0 +1,226 @@
...
+
+class TimeSeriesDataFrame(pd.DataFrame):
+    """TimeSeriesDataFrame to represent time-series dataset.
+
+    Parameters
+    ----------
+    data : Any
+        Time-series data to construct a TimeSeriesDataFrame.
+        It currently supports two formats:
+            1. Time-series data in TimeSeriesListDataset format.
```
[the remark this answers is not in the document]
huibinshen: Yes, will add them in the doc-string.

forecasting/src/autogluon/forecasting/dataset/ts_dataframe.py
huibinshen: Noted, will change in the next revision.

forecasting/src/autogluon/forecasting/dataset/ts_dataframe.py
```diff
@@ -0,0 +1,226 @@
...
+from typing import Any, Tuple, Dict, List
+from dataclasses import dataclass
+
+import pandas as pd
+import numpy as np
+
+
+@dataclass
+class TimeSeriesListDataset:
+    data_iter: List[Dict[str, Any]]
```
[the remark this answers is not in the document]
huibinshen: Good point, will change!

forecasting/src/autogluon/forecasting/dataset/ts_dataframe.py
huibinshen: Yes, as Caner told, the core data structure used in the forecasting module will be `TimeSeriesDataFrame`, the users does not need to know `TimeSeriesListDataset`.

forecasting/src/autogluon/forecasting/dataset/ts_dataframe.py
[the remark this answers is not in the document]
canerturkmen: Hmm. Sure. but I think we should also default to it somehow in the constructor. So how about in the constructor we can check if there's a multiindex, and if there isn't try to call `self.from_dataframe`? Or is this too much obfuscation of the API?

forecasting/src/autogluon/forecasting/dataset/ts_dataframe.py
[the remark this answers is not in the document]
canerturkmen: interesting! but if int64 is anyway enforced by pandas, would it make sense still to check `int` so `uint64` is also covered?

forecasting/src/autogluon/forecasting/dataset/ts_dataframe.py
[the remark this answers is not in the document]
huibinshen: Yeah, I was also thinking that. Basically as the way we handle `TimeSeriesListDataset'.

forecasting/src/autogluon/forecasting/dataset/ts_dataframe.py
```diff
@@ -0,0 +1,331 @@
...
+
+    @classmethod
+    def _validate_list_dataset(cls, list_dataset):
+        if not isinstance(list_dataset, TimeSeriesListDataset):
+            raise ValueError(
+                f"list_dataset must be a TimeSeriesListDataset, got {type(list_dataset)}"
+            )
+        if len(list_dataset.data) == 0:
+            raise ValueError(f"list_dataset has no time-series.")
+        for i, ts in enumerate(list_dataset.data):
```
canerturkmen, line 156: as we are now using `UserList`, all references to `list_dataset.data` could be replaced by `list_dataset`. Similarly for `len` checks above and other iterators below.

forecasting/src/autogluon/forecasting/dataset/ts_dataframe.py
```diff
@@ -0,0 +1,331 @@
...
+            raise ValueError(f"`timestamp` columns can not have nan")
+        if not df["item_id"].dtype == "int64":
+            raise ValueError(f"for item_id, the only pandas dtype allowed is ‘int64’.")
+        if not df["timestamp"].dtype == "datetime64[ns]":
+            raise ValueError(
+                f"for timestamp, the only pandas dtype allowed is ‘datetime64[ns]’."
+            )
+
+    @classmethod
+    def _validate_list_dataset(cls, list_dataset):
```
jaheba, line 149: Shouldn't we put this on `ListDataset`?

Then we can do `list_dataset.validate()`.

canerturkmen (commented):
Looks good to me. Over to @Innixma

forecasting/src/autogluon/forecasting/dataset/ts_dataframe.py
huibinshen: Good point, will move it inside ListDataset.

canerturkmen: Actually I disagree with this point, going back to the design, and maybe will walk back some of my earlier discussions. Our original objective with the `ListDataset` object was that someone could bring their GluonTS data set and work out of the box with AG Forecasting. We seem to be headed in the other direction here by making `TimeSeriesListDataset` a concrete object,  also with checks like `isinstance(.., TimeSeriesListDataset`, and encapsulating its validation logic. This would make sense if we wanted this object to live and be passed around the library. We in fact want it to disappear.

We can revisit this in another PR.

huibinshen: Hi Caner, it is clear to me that our core data structure will be `TimeSeriesDataFrame`. Having the validation logic inside `TimeSeriesListDataset` is very far from passing around in our library IMHO: 
1) I guess we will make it explicit that the forecasting predictor will require a `TimeSeriesDataFrame` to work. 
2) We have no plan to support data split or other functions in `TimeSeriesListDataset`.
3) When `TimeSeriesListDataset` disappears, it is actually easier to delete it as the validation is part of the `TimeSeriesListDataset` class.

huibinshen: @canerturkmen @jaheba If we don't want to introduce a new class but to provide easy conversion for ListDataset in gluonts, one thing we could do is:
```
from collections.abc import Iterable

# in the __init__ of TimeSeriesDataFrame
def __init__(self, data: Any, *args, **kwargs):
    if isinstance(data, Iterable):
        data = self.from_list_dataset(data)    
    ...
```
Then we need to write in the documentation what format we are expecting and validate it in `self.from_list_dataset` method. One caveat is that we will assume the `start` timestamp has `freq`.

jaheba: Got me thinking, why do we want to subclass `DataFrame`? Couldn't we create a new class which uses a `DataFrame` internally instead?

canerturkmen: @huibinshen as @jaheba mentioned in the call, I think this problem is solved if we simply check for `DataFrame` first.

@jaheba this was discussed and decided in the design review. We want to carry over the data frame API; with all of its functions to enable ease of use for the customer in a non-obstrusive way and 2/adhere as much as possible to the rest of autogluon which uses simple pandas dataframes. So in this case, inheritance over composition.

I'm happy to take care of the validation + type issue in a future PR.

huibinshen: thanks for the discussion. I'll resolve the conversation for now.

forecasting/src/autogluon/forecasting/dataset/ts_dataframe.py
```diff
@@ -0,0 +1,328 @@
...
+        data_after : TimeSeriesDataFrame
+            The second one after split contains time-series after the cutoff_time (include cutoff_time).
+        """
+
+        nanosecond_before_cutoff = cutoff_time - pd.Timedelta(nanoseconds=1)
+        data_before = self.loc[(slice(None), slice(None, nanosecond_before_cutoff)), :]
+        data_after = self.loc[(slice(None), slice(cutoff_time, None)), :]
+        return TimeSeriesDataFrame(data_before), TimeSeriesDataFrame(data_after)
+
+    def split_by_item(
```
Innixma, line 281: Does it make sense to cutoff based off an int? Is this restrictive? Why not have `cutoff_items: list`? (I could be naive here)

Innixma (approved):
LGTM! I just have minor comments that I'd like addressed

forecasting/src/autogluon/forecasting/dataset/ts_dataframe.py
```diff
@@ -0,0 +1,328 @@
...
+
+    @classmethod
+    def _validate_data_frame(cls, df: pd.DataFrame):
+        if not isinstance(df, pd.DataFrame):
+            raise ValueError(f"data must be a pd.DataFrame, got {type(df)}")
+        if "item_id" not in df.columns:
+            raise ValueError(f"data must have a `item_id` column")
+        if "timestamp" not in df.columns:
+            raise ValueError(f"data must have a `timestamp` column")
+        if df["item_id"].isnull().any():
```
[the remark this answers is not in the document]
huibinshen: Good point, now they are in constants. I did the same for the following comments, will not reply them one-by-one.

forecasting/src/autogluon/forecasting/dataset/ts_dataframe.py
```diff
@@ -0,0 +1,328 @@
...
+            raise ValueError(f"data must have pd.MultiIndex, got {type(data.index)}")
+        if not data.index.dtypes.array[0] == "int64":
+            raise ValueError(f"for item_id, the only pandas dtype allowed is ‘int64’.")
+        if not data.index.dtypes.array[1] == "datetime64[ns]":
+            raise ValueError(
+                f"for timestamp, the only pandas dtype allowed is ‘datetime64[ns]’."
+            )
+        if not data.index.names == ("item_id", "timestamp"):
+            raise ValueError(
+                f"data must have index names as ('item_id', 'timestamp'), got {type(data.index.names)}"
```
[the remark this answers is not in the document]
huibinshen: That's a typo, now it's been fixed.

forecasting/src/autogluon/forecasting/dataset/ts_dataframe.py
huibinshen: I see here needs more discussion actually. @RSNirwan also mentioned we may want to support item_id in categorical type or string. Also, we type of split options we will support for those types. Does it make sense that we take this as a follow up after the current PR has been merged?

szha:
Job PR-1628-5 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1628/5/index.html

canerturkmen (approved):

forecasting/src/autogluon/forecasting/dataset/ts_dataframe.py
Innixma: Why was this marked as resolved? Doesn't sound resolved from the comment. Happy to have a follow-up, but would prefer some kind of todo

huibinshen: I saw you approved the change and considered that as an agreement for taking it as a follow up. For now I put a comment in the design doc so that we don't forget about it. We can also explicitly add a todo in the code if you wish.

Innixma: Ok no worries, so long as it is tracked in design doc its fine

PR #1631, opened 2022-03-28
Marking regression test non-default

yinweisu:
*Issue #, if available:*
https://github.com/awslabs/autogluon/pull/1555

*Description of changes:*
Marking tabular regression tests to be added as non-default. Requires `--runregression` to run


Innixma (approved):
LGTM!

PR #1641, opened 2022-04-01
Parallel HPO

yinweisu:
*Issue #, if available:*

*Description of changes:*
* Created Abstract Class `LocalScheduler`, which `LocalSequentialScheduler` and `LocalParallelScheduler` will implement
* Implement parallel HPO with ray
  * Tasks are scheduled and executed on the run
  * Time limits are arranged so that each trial gets full time that's left
  * Memory are calculated naively with default config.
  * User can specify number of trials running in parallel via `num_trials_parallel`. However, the actual number of trials will be determined by a combination of cpu cores, memory, and user specification.
* Added unit tests for `LocalParallelScheduler`, and moved unit tests for `LocalSequentialScheduler` to better group tests
* Modified `get_cpu_count` to provide either virtual or physical cores. Changed to use `psutil` because it performs as a superset of `multiprocessing`


szha:
Job PR-1641-5 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1641/5/index.html

PR #1656, opened 2022-04-07
Fix saving and loading tokenizers and text processors

zhiqiangdon:
1. Saved pre-trained huggingface tokenizers separately from the text processors.
2. Saved text processors separately from other processors.
3. Also added the corresponding loading methods.
4. Improved implementations and docstrings of `save_pretrained_models` and `convert_checkpoint_name`.

### DIFF
--- a/text/src/autogluon/text/automm/data/process_text.py
+++ b/text/src/autogluon/text/automm/data/process_text.py
@@ -326,6 +326,14 @@ class TextProcessor:
             self,
             path: str,
     ):
+        """
+        Save the tokenizer and text processor.
+
+        Parameters
+        ----------
+        path
+            Saving path.
+        """
         self.tokenizer.save_pretrained(path)
         self.tokenizer = None
         with open(os.path.join(path, "text_processor.pkl"), "wb") as fp:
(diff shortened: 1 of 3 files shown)


text/src/autogluon/text/automm/data/process_text.py
```diff
@@ -320,6 +322,48 @@ def get_trimmed_lengths(
...
+
+        Parameters
+        ----------
+        path
+            Saving path.
+        """
+        self.tokenizer.save_pretrained(path)
+        self.tokenizer = None
+        with open(os.path.join(path, "text_processor.pkl"), "wb") as fp:
+            pickle.dump(self, fp)
```
sxjscience, line 340: need to restore the tokenizer so the logic won't impact the inner states of this object.

```python
tokenizer = self.tokenizer
self.tokenizer = None
# DO_SOMETHING
self.tokenizer = tokenizer

```

text/src/autogluon/text/automm/predictor.py
```diff
@@ -1063,6 +1075,12 @@ def load(
             df_preprocessor = pickle.load(fp)
         with open(os.path.join(path, "data_processors.pkl"), "rb") as fp:
             data_processors = pickle.load(fp)
+        if TEXT in data_processors:
+            # Load text processors separately due to tokenizers.
+            data_processors[TEXT] = load_text_processors(
+                relative_paths=data_processors[TEXT],
+                path=path,
+            )
```
sxjscience, line 1083: We need to also support loading a checkpoint trained via `0.4.0`. We can detect the version and give special treatment on the 0.4.0-version models.

PR #1657, opened 2022-04-07
add forecasting trainer, minor fixes to forecasting models

canerturkmen:
Co-authored-by: yx1215 <yx1215@nyu.edu>

*Issue #, if available:*

*Description of changes:*

- Adds `SimpleAbstractTrainer`, a single abstract class that both core `AbstractTrainer` and `AbstractForecastingTrainer` can inherit from.
- adds `AbstractForecastingTrainer` and `AutoForecastingTrainer`
- minor fixes and improvements to models, notably introducing a convenient factory class for `GenericGluonTSModel` s.


canerturkmen:
@huibinshen @lnnixma @dcmaddix 

also cc @yx1215, I had added your signature to the last commit as well but that seems to have been lost during a squash, apologies for that!

szha:
Job PR-1657-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1657/1/index.html

szha:
Job PR-1657-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1657/2/index.html

forecasting/src/autogluon/forecasting/trainer/abstract_trainer.py
```diff
@@ -0,0 +1,668 @@
...
+            self.hpo_results[model.name] = hpo_results
+            model_names_trained = []
+
+            # add each of the trained HPO configurations to the trained models
+            for model_hpo_name, model_path in hpo_models.items():
+                model_hpo = self.load_model(
+                    model_hpo_name, path=model_path, model_type=type(model)
+                )
+                self._add_model(model_hpo)
+                model_names_trained.append(model_hpo.name)
```
Innixma, line 343: Just a note: Code for HPO is pretty hacky in AbstractTrainer in Tabular at the moment, probably deserves a refactor at some point so it isn't crammed into `_train_single_full` but instead has its own method. If this PR can lead the charge here by making a separate method for this in forecast so its tidier, I think it would be great.

forecasting/src/autogluon/forecasting/trainer/abstract_trainer.py
```diff
@@ -0,0 +1,668 @@
...
+        model_names_trained = []
+        try:
+            if time_limit is not None:
+                if time_limit <= 0:
+                    logging.log(
+                        15, f"Skipping {model.name} due to lack of time remaining."
+                    )
+                    return model_names_trained
+            else:
+                logging.log(20, f"Fitting model: {model.name} ...")
```
Innixma, line 375: logic like this is pretty generic, I wonder if we can have a clever way of reusing this between the different modules.

forecasting/src/autogluon/forecasting/trainer/abstract_trainer.py
```diff
@@ -0,0 +1,668 @@
...
+            model = self._train_single(
+                train_data, model, val_data=val_data, time_limit=time_limit
+            )
+            fit_end_time = time.time()
+
+            val_score = (
+                model.score(val_data) * METRIC_COEFFICIENTS[model.eval_metric]
```
Innixma, line 382: careful with metric coefficients, they can cause a lot of pain and hardship, for example with sorting best models. This is why in tabular all `model.score` values will return the higher_is_better style scores as a rule. We may want to consider doing this for forecasting as well.

forecasting/src/autogluon/forecasting/trainer/abstract_trainer.py
```diff
@@ -0,0 +1,668 @@
...
+        test_score = []
+        if data is not None:
+            logger.log(
+                30, "Additional data provided, testing on the additional data..."
+            )
+            for model_name in model_names:
+                model: AbstractForecastingModel = self.load_model(model_name)
+                test_score.append(
+                    model.score(data) * METRIC_COEFFICIENTS[self.eval_metric]
+                )
```
Innixma, line 487: This kind of trick is going to stop working once we add ensembling, maybe instead move score logic to a separate method. such as `self.score(data, model)`, this way an ensemble model could preprocess the data to be able to properly calculate score via first calling predict on the base models.

forecasting/src/autogluon/forecasting/trainer/abstract_trainer.py
```diff
@@ -0,0 +1,668 @@
...
+        logger.log(30, "Generating leaderboard for all models trained...")
+        model_names = self.get_model_names_all()
+        score_val = []
+        fit_time_marginal = []
+        fit_order = list(range(1, len(model_names) + 1))
+        score_dict = self.get_models_attribute_dict("val_score")
+        fit_time_marginal_dict = self.get_models_attribute_dict("fit_time")
+        for model_name in model_names:
+            score_val.append(score_dict[model_name])
+            fit_time_marginal.append(fit_time_marginal_dict[model_name])
```
Innixma, line 476: fit_time_marginal never used?

forecasting/src/autogluon/forecasting/trainer/abstract_trainer.py
```diff
@@ -0,0 +1,668 @@
...
+        data: TimeSeriesDataFrame,
+        model: Union[str, AbstractForecastingModel],
+        **kwargs,
+    ) -> TimeSeriesDataFrame:
+        if isinstance(model, str):
+            model = self.load_model(model)
+        return model.predict(data, **kwargs)
+
+    # TODO: experimental
+    def refit_single_full(self, train_data=None, val_data=None, models=None):
```
Innixma, line 578: Do we think it is possible to make this in SimpleAbstractTrainer or some kind of mixin? The logic for refit full is pretty generic overall, and I foresee that we could maybe clean up the code to make it do-able as fully generic.

forecasting/src/autogluon/forecasting/trainer/abstract_trainer.py
```diff
@@ -0,0 +1,668 @@
...
+        # for now we assume validation data includes train data (as in GluonTS)
+        refit_full_data = val_data
+
+        if models is None:
+            self.get_model_names_all()
+
+        for model in models:
+            model: AbstractForecastingModel = self.load_model(model)
+            model_name = model.name
+            model_full = model.convert_to_refit_full_template()  # FIXME: not available
```
Innixma, line 597: What is reason for it not being available? Is there complexity in implementation?

forecasting/src/autogluon/forecasting/trainer/abstract_trainer.py
```diff
@@ -0,0 +1,668 @@
...
+                    model_trained
+                ] = self.get_model_attribute(model_name, "val_score")
+            models_trained_full += models_trained
+
+        self.model_full_dict.update(model_full_dict)
+        self.save()
+        return models_trained_full
+
+    # TODO: experimental
+    def refit_full(self, models="all"):
```
Innixma, line 619: Ditto on if this can be made part of SimpleAbstractModel or a mixin

forecasting/src/autogluon/forecasting/trainer/abstract_trainer.py
```diff
@@ -0,0 +1,668 @@
...
+        train_data: TimeSeriesDataFrame,
+        hyperparameters: Optional[Union[str, Dict]] = None,
+        models: Optional[List[AbstractForecastingModel]] = None,
+        val_data: Optional[TimeSeriesDataFrame] = None,
+        hyperparameter_tune: bool = False,
+        time_limit: Optional[float] = None,
+    ) -> List[str]:
+        time_start = time.time()
+        if hyperparameters is not None:
+            self.hyperparameters = hyperparameters
```
Innixma, line 424: Careful: Imagine users being able to call fit multiple times (train 5 models in 1 fit call, then 7 more in another, etc.). Having a `self.hyperparameters` shouldn't be necessary.

forecasting/src/autogluon/forecasting/trainer/abstract_trainer.py
```diff
@@ -0,0 +1,668 @@
...
+        model_names_trained = []
+        for i, model in enumerate(models):
+            if hyperparameter_tune:
+                time_left = time_limit_model_split
+            else:
+                if time_limit is None:
+                    time_left = None
+                else:
+                    time_start_model = time.time()
+                    time_left = time_limit - (time_start_model - time_start)
```
Innixma, line 454: `if time_left <= 0` do we want to keep for looping or stop immediately with a log message? I believe Tabular stops immediately, but this code looks to continue looping.

forecasting/src/autogluon/forecasting/trainer/auto_trainer.py
```diff
@@ -0,0 +1,67 @@
...
+    # todo: including CVSplitter logic
+    def fit(
+        self,
+        train_data: TimeSeriesDataFrame,
+        hyperparameters: Union[str, Dict[Any, Dict]],
+        val_data: Optional[TimeSeriesDataFrame] = None,
+        hyperparameter_tune: bool = False,
+        time_limit: float = None,
+        infer_limit: float = None,  # todo: implement
+    ):
```
Innixma, line 38: can enable additional calls to fit by allowing `train_data=None` to trigger loading via `self.load_data()`, but would recommend first double checking how Tabular does it. (refer to `fit_extra` method in `TabularPredictor`)

forecasting/src/autogluon/forecasting/trainer/auto_trainer.py
```diff
@@ -0,0 +1,67 @@
...
+            eval_metric=eval_metric,
+            prediction_length=self.prediction_length,
+            freq=self.freq,
+            hyperparameters=hyperparameters,
+            hyperparameter_tune=hyperparameter_tune,
+            quantiles=self.quantiles,
+            use_feat_static_cat=self.use_feat_static_cat,
+            use_feat_static_real=self.use_feat_static_real,
+            cardinality=self.cardinality,
+        )
```
Innixma, line 26: How do you ensure models don't have name collisions over multiple calls to `get_preset_models`?

The logic to ensure duplicate names aren't generated is implemented in AbstractModel currently, but it doesn't look like it is implemented here.

Imagine `self.model_info` already contains `model_name`. Must ensure `model_name` isn't used as a name for models in this call to avoid collision.

Ex in Tabular's AutoTrainer:

```
class AutoTrainer(AbstractTrainer):
    def construct_model_templates(self, hyperparameters, **kwargs):
        path = kwargs.pop('path', self.path)
        problem_type = kwargs.pop('problem_type', self.problem_type)
        eval_metric = kwargs.pop('eval_metric', self.eval_metric)
        quantile_levels = kwargs.pop('quantile_levels', self.quantile_levels)
        invalid_model_names = kwargs.pop('invalid_model_names', self._get_banned_model_names())
        silent = kwargs.pop('silent', self.verbosity < 3)
        ag_args_fit = kwargs.pop('ag_args_fit', None)
        if quantile_levels is not None:
            if ag_args_fit is None:
                ag_args_fit = dict()
            ag_args_fit = ag_args_fit.copy()
            ag_args_fit['quantile_levels'] = quantile_levels

        return get_preset_models(path=path,
                                 problem_type=problem_type,
                                 eval_metric=eval_metric,
                                 ag_args_fit=ag_args_fit,
                                 invalid_model_names=invalid_model_names,
                                 silent=silent, **kwargs)
```

Innixma (commented):
Looks very good for an initial submission! Added a variety of comments.

forecasting/src/autogluon/forecasting/models/presets.py
```diff
@@ -0,0 +1,199 @@
...
+                "context_length": ag.Int(5, 25),
+            },
+        },
+        "default": {
+            "SimpleFeedForward": {},
+            "MQCNN": {},
+            "DeepAR": {},
+            # "AutoTabular": {} # AutoTabular model is experimental.
+        },
+        "default_hpo": {
```
huibinshen, line 72: I guess the search space for now is tentative and the final search space definition will come in the future?

forecasting/src/autogluon/forecasting/models/presets.py
```diff
@@ -0,0 +1,199 @@
...
+            get_default_hps(hyperparameters, prediction_length)
+        )
+    else:
+        hp_str = "default" if not hyperparameter_tune else "default_hpo"
+        default_hps = copy.deepcopy(get_default_hps(hp_str, prediction_length))
+
+        if hyperparameters is not None:
+            # filter only default_hps for models with hyperparameters provided
+            default_hps = {
+                model: default_hps.get(model, {}) for model in hyperparameters
```
huibinshen, line 127: Nitpick: would `hp` be a better name when iterating through hyperparameters? A `hp` can be used to create a model.

forecasting/src/autogluon/forecasting/trainer/auto_trainer.py
```diff
@@ -0,0 +1,67 @@
...
+
+    # todo: implement cross-validation / holdout strategy
+    # todo: including CVSplitter logic
+    def fit(
+        self,
+        train_data: TimeSeriesDataFrame,
```
huibinshen, line 32: Will we have a convenient function in this class to convert TimeSeriesDaaFrame to ListDataset?

forecasting/src/autogluon/forecasting/models/presets.py
```diff
@@ -0,0 +1,199 @@
...
+                    max(min(500, 12 * prediction_length), prediction_length),
+                    default=prediction_length,
+                ),
+            },
+        },
+    }
+    return default_model_hps[key]
+
+
+def get_preset_models(
```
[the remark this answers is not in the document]
canerturkmen: I don't think it is duped, tabular also has a function with the same name but quite different signature and content. I would leave it in for now.

forecasting/src/autogluon/forecasting/models/presets.py
```diff
@@ -0,0 +1,199 @@
...
+    freq: str,
+    prediction_length: int,
+    path: str,
+    eval_metric: str,
+    hyperparameters: Union[str, Dict],
+    hyperparameter_tune: bool,
+    use_feat_static_cat: bool,
+    use_feat_static_real: bool,
+    cardinality: int,
```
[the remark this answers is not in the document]
canerturkmen: they can alternatively be specified via hyperparameters. will remove them.

forecasting/src/autogluon/forecasting/trainer/abstract_trainer.py
```diff
@@ -0,0 +1,668 @@
...
+from ..utils.warning_filters import evaluator_warning_filter
+from ..utils.metric_utils import METRIC_COEFFICIENTS, check_get_evaluation_metric
+
+logger = logging.getLogger(__name__)
+
+
+# TODO: This class is meant to be moved to `core`, where it will likely
+# TODO: be renamed `AbstractTrainer` and the current `AbstractTrainer`
+# TODO: will inherit from this class.
+class SimpleAbstractTrainer:
```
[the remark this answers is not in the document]
canerturkmen: yes, I thought it would be better to have it as a different PR. wdyt?

forecasting/src/autogluon/forecasting/trainer/abstract_trainer.py
```diff
@@ -0,0 +1,668 @@
...
+
+    @property
+    def path_utils(self) -> str:
+        return self.path_root + "utils" + os.path.sep
+
+    @property
+    def path_data(self) -> str:
+        return self.path_utils + "data" + os.path.sep
+
+    def set_contexts(self, path_context: str) -> None:
```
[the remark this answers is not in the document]
canerturkmen: Done. We can also address this in a followup PR when adding trainer to core.

forecasting/src/autogluon/forecasting/models/presets.py
canerturkmen: I didn't think these were bad defaults. Should we think about others?

forecasting/src/autogluon/forecasting/models/presets.py
canerturkmen: No these are model names indeed. so the hyperparameters dict maps model names to a nested dict of hyperparameters.

forecasting/src/autogluon/forecasting/trainer/abstract_trainer.py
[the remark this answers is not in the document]
Innixma: Yes, seems good to me

forecasting/src/autogluon/forecasting/trainer/abstract_trainer.py
```diff
@@ -0,0 +1,668 @@
...
+
+    def __init__(self, path: str, low_memory: bool, save_data: bool, *args, **kwargs):
+        self.path = path
+        self.reset_paths = False
+
+        self.low_memory = low_memory
+        self.save_data = save_data
+
+        self.models = {}
+        self.model_info = {}
```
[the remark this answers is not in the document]
canerturkmen: Just checked this. Is this a one-for-one replacement (in terms of metadata) for `self.model_info` in the current core `AbstractTrainer`. If so, I can just replace `self.model_info` in the new `SimpleAbstractTrainer` with `model_graph`?

forecasting/src/autogluon/forecasting/trainer/abstract_trainer.py
canerturkmen: we could probably employ a decorator like the following. I'll propose to push this to a different PR?

```python
import functools
import logging

class TimeLimitExceeded(Exception):
    pass

def time_limited(*decorator_args, **decorator_kwargs):
    
    def decorator(func):
    
        @functools.wraps(func)
        def wrapper(*args, **kwargs):
            time_limit = kwargs.pop("time_limit", None)
            
            if time_limit is not None:
                if time_limit <= 0:
                    logging.log(
                        15, f"Skipping due to lack of time remaining."
                    )
                    raise TimeLimitExceeded()
                else:
                    logging.log(20, f"Fitting model: ...")

            
            return func(*args, **kwargs)
        
        return wrapper
    
    return decorator

@time_limited()
def my_wrapped_function(a, b):
    print(a + b)
    
my_wrapped_function(3, 4, time_limit=-5)
```

forecasting/src/autogluon/forecasting/models/presets.py
huibinshen: I don't know they are good or bad but I know their implication is significant. Want to call out so that we put enough attention to it. Especially given that we don't do multi-layer stacking in the beginning where the model is just trained with default HPs without any HPO.

forecasting/src/autogluon/forecasting/trainer/abstract_trainer.py
canerturkmen: Yes, I completely agree and will fix this. However, note that this will also result in a discrepancy similar to tabular regression. For example, although the eval metric is called `root_mean_squared_error`, the "negative" RMSE will show up in the leaderboard.

forecasting/src/autogluon/forecasting/trainer/abstract_trainer.py
canerturkmen: I revised to `self.score`, and added a todo in self score that reminds us this will break when ensembling is implemented.

forecasting/src/autogluon/forecasting/trainer/abstract_trainer.py
canerturkmen: The problem is that `train_data` and `val_data` conflict with the `X, y` based signatures of core. So whenever there is a function with data arguments we have to override it in forecast. I think this is part of a bigger issue which would be great to tackle to make this and other logic more generic.

forecasting/src/autogluon/forecasting/trainer/abstract_trainer.py
canerturkmen: Not particularly, but we do not currently support `params_trained` and incremental training. I know this is important functionality but I'd like to address this in a future PR.

forecasting/src/autogluon/forecasting/trainer/abstract_trainer.py
canerturkmen: As far as I can tell Tabular keeps looping, but usually doesn't say it's "skipping due to lack of time remaining" because the log level is too low. I think it doesn't hurt to do this.

forecasting/src/autogluon/forecasting/trainer/auto_trainer.py
canerturkmen: Tabular also doesn't accept fitting without data. This is handled by the predictor although the data is cached by the trainer. I'll resolve this for now?

forecasting/src/autogluon/forecasting/trainer/abstract_trainer.py
```diff
@@ -0,0 +1,668 @@
...
+        for model, path in model_paths.items():
+            self.set_model_attribute(model=model, attribute="path", val=path)
+
+    def create_contexts(self, path_context: str) -> Tuple[str, dict]:
+        path = path_context
+        model_paths = self.get_models_attribute_dict(attribute="path")
+        for model, prev_path in model_paths.items():
+            model_local_path = prev_path.split(self.path, 1)[1]
+            new_path = path + model_local_path
+            model_paths[model] = new_path
```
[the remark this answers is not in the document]
canerturkmen: In `load` we provide an option to `reset_paths` which leads me to think we want to be able to keep track of trainers that do not have to have models in their subdirectories. That is why I would leave this for now. Adding a todo.

forecasting/src/autogluon/forecasting/trainer/abstract_trainer.py
Innixma: I checked and Tabular stops immediately, it doesn't loop. I'd probably prefer this as it avoids redundant logging.

forecasting/src/autogluon/forecasting/trainer/abstract_trainer.py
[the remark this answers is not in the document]
Innixma: Yes I believe so

forecasting/src/autogluon/forecasting/trainer/abstract_trainer.py
Innixma: Yes, this is the price necessary for logical consistency.

forecasting/src/autogluon/forecasting/trainer/abstract_trainer.py
Innixma: Ok, sounds good, definitely a bigger change so we can think about it in a future PR.

forecasting/src/autogluon/forecasting/trainer/abstract_trainer.py
Innixma: Ok, we can add in follow-up PR.

szha:
Job PR-1657-5 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1657/5/index.html

forecasting/src/autogluon/forecasting/trainer/auto_trainer.py
Innixma: Sure, can resolve for now. This might also just be an issue with API outside of forecasting, we may want to instead have trainer load the data it cached, but thats a separate discussion.

forecasting/src/autogluon/forecasting/trainer/abstract_trainer.py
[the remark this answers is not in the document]
Innixma: > which leads me to think we want to be able to keep track of trainers that do not have to have models in their subdirectories.

This was indeed the intent when I coded it, however I'm unsure if it is useful to continue enabling this kind of behavior. I suppose we can keep it for now, but it does add a bit of complexity.

forecasting/src/autogluon/forecasting/models/presets.py
```diff
@@ -0,0 +1,198 @@
...
+            hyperparameters=model_hps,
+            **kwargs,
+        )
+
+        # add models while preventing name collisions
+        model = model_type(**model_type_kwargs)
+        if model.name in [m.name for m in models] + invalid_model_names:
+            model = model_type(
+                name=f"{model.name}_{str(uuid.uuid4())[:6]}", **model_type_kwargs
+            )
```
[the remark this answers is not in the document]
canerturkmen: Thanks for pointing this out. I agree with points  2-3 and especially 4. The reason I went for this is since one has to keep track of state of name prefixes, or "infer" it like the code you referenced which I thought was overkill. I'll revise this.

forecasting/src/autogluon/forecasting/trainer/abstract_trainer.py
canerturkmen: That's strange. I'm running `example_simple_tabular.py` specifying 0.5 time limit to the predictor and also changing log level to 20 on https://github.com/awslabs/autogluon/blob/25f98756591c0cdb5344b575222e08bcef13e496/core/src/autogluon/core/trainer/abstract_trainer.py#L1054

The output: 
```
...
Fitting model: LightGBMXT ... Training model for up to 0.38s of the 0.38s of remaining time.
	Warning: Exception caused LightGBMXT to fail during training (ImportError)... Skipping this model.
		`import lightgbm` failed. A quick tip is to install via `pip install lightgbm`.
Fitting model: LightGBM ... Training model for up to 0.37s of the 0.37s of remaining time.
	Warning: Exception caused LightGBM to fail during training (ImportError)... Skipping this model.
		`import lightgbm` failed. A quick tip is to install via `pip install lightgbm`.
Fitting model: RandomForestGini ... Training model for up to 0.36s of the 0.36s of remaining time.
	0.81	 = Validation score   (accuracy)
	0.58s	 = Training   runtime
	0.04s	 = Validation runtime
Skipping RandomForestEntr due to lack of time remaining.
Skipping CatBoost due to lack of time remaining.
Skipping ExtraTreesGini due to lack of time remaining.
Skipping ExtraTreesEntr due to lack of time remaining.
Skipping NeuralNetFastAI due to lack of time remaining.
Skipping XGBoost due to lack of time remaining.
Skipping NeuralNetTorch due to lack of time remaining.
Skipping LightGBMLarge due to lack of time remaining.
```

canerturkmen: Anyway, I'll add a break here.

canerturkmen:
Thank you for the detailed and rigorous review @lnnixma!

szha:
Job PR-1657-6 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1657/6/index.html

forecasting/src/autogluon/forecasting/models/presets.py
```diff
@@ -0,0 +1,209 @@
...
+            **kwargs,
+        )
+
+        # add models while preventing name collisions
+        model = model_type(**model_type_kwargs)
+        num_increment = 2
+        model_type_kwargs.pop("name", None)
+        while (
+            model.name in [m.name for m in models]
+            or model.name in invalid_model_names_set
```
Innixma, line 169: simpler: Just add model.name to invalid_model_names_set when it is added to models. Then no need to check `m.name for m in models`

forecasting/src/autogluon/forecasting/models/presets.py
```diff
@@ -0,0 +1,209 @@
...
+        ):
+            model = model_type(
+                name=f"{model.name}_{num_increment}", **model_type_kwargs
+            )
+            # increment only if model name collided in this round
+            if model.name not in invalid_model_names_set:
+                num_increment += 1
```
Innixma, line 176: Don't you mean `in`? Won't this infinite loop otherwise? Plus, this won't work if the collision is in `models` and not in `invalid_model_names_set`.

Further, simpler to just always `num_increment += 1` at start of while loop and have `num_increment = 1` initially, avoids having to check twice.

canerturkmen: The loop itself doesn't depend on `num_increment` and if there is a collision it just keeps appending `_{num_increment}` so I don't think it would infinite loop in any case. The logic, and why it doesn't add to `invalid_model_names_set` as suggested above is as follows.

`num_increment` is only incremented if the name collision happened this round. So if 
```
invalid_model_names (those from previous calls to fit): ["A", "B"]
new default model names being pushed: ["B", "B"]
the output is ["B_2", "B_3"]
```
and 
```
invalid_model_names (those from previous calls to fit): ["A", "B", "B_2"]
new default model names being pushed: ["B", "B"]
the output is ["B_2_2", "B_2_3"]
```
If we didn't perform the check on L175, then the outputs would be `"B_2_3", "B_2_4"`.
Logic looks OK to me but yes I think it will pay to make it more legible.

gradientsky:
Namespace package must have `src/autogluon/__init__.py` with the following content:
```python
__import__("pkg_resources").declare_namespace(__name__)
```

gradientsky:
Module `setup.py` is missing (see other modules for example)

forecasting/src/autogluon/forecasting/trainer/auto_trainer.py
```diff
@@ -0,0 +1,66 @@
...
+        hyperparameters: Union[str, Dict[Any, Dict]],
+        val_data: Optional[TimeSeriesDataFrame] = None,
+        hyperparameter_tune: bool = False,
+        time_limit: float = None,
+        infer_limit: float = None,  # todo: implement
```
gradientsky, line 36: nit: ditto- `train_data` and `val_data`, consider standardizing parameters order.

forecasting/src/autogluon/forecasting/trainer/abstract_trainer.py
```diff
@@ -0,0 +1,659 @@
...
+            self._add_model(model_hpo)
+            model_names_trained.append(model_hpo.name)
+        return model_names_trained
+
+    def _train_and_save(
+        self,
+        train_data: TimeSeriesDataFrame,
+        model: AbstractForecastingModel,
+        val_data: Optional[TimeSeriesDataFrame] = None,
+        time_limit: Optional[float] = None,
```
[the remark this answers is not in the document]
canerturkmen: Ordering the arguments as such breaks consistency with core where we go `X, y, model`. Similarly for those below, especially with `fit` it would be very counter-intuitive to have `fit(hyperparameters, data)`.

canerturkmen:
@gradientsky I believe both namespace declaration and setup.py are already there (merged before)?

forecasting/src/autogluon/forecasting/models/presets.py
```diff
@@ -0,0 +1,214 @@
...
+        name_stem = model.name
+
+        model_type_kwargs.pop("name", None)
+        increment = 1
+        while model.name in all_assigned_names:
+            # if name collides with a name prohibited by the function argument
+            # then we always append _2
+            if model.name in invalid_model_names:
+                name_stem = model.name
+                model = model_type(name=f"{model.name}_2", **model_type_kwargs)
```
Innixma, line 175: Why? I don't see why we use this logic instead of the generic increment logic in the else statement.

It shouldn't matter if I call this function once with models A B C or three times with model A, then model B, then model C, the exact same model names should be given to A B C.

With how the code currently works, if I called once and their names were all "M", then it would be:

A.name = M
B.name = M_2
C.name = M_3

And if I called three times each with just 1 model, it would instead be:

A.name = M
B.name = M_2
C.name = M_2_2

I propose that the first result should apply always, and the entire while loop be simplified to:

```
while model.name in all_assigned_names:
    increment += 1
    model = model_type(name=f"{name_stem}_{increment}", **model_type_kwargs)
```

forecasting/src/autogluon/forecasting/models/presets.py
Innixma: Thanks for looking into this. It is nuanced but actually pretty important.

Please refer to my new comment discussing what I think can be improved from the current implementation.

In your example, my suggested implementation instead of:

```
invalid_model_names (those from previous calls to fit): ["A", "B", "B_2"]
new default model names being pushed: ["B", "B"]
the output is ["B_2_2", "B_2_3"]
```

would be:

```
invalid_model_names (those from previous calls to fit): ["A", "B", "B_2"]
new default model names being pushed: ["B", "B"]
the output is ["B_3", "B_4"]
```

Innixma (approved):
LGTM! I still have a open comment regarding the name dedupe logic, but once that is fixed I think we are good to merge. (Feel free to merge on your end once you are satisfied with the dedupe logic and have considered my comment).

forecasting/src/autogluon/forecasting/models/presets.py
canerturkmen: I see. referring to your original comment

> I'd like to propose we use the same solution we use in TabularPredictor: appending an integer such as A and A_2. If we had 5 A models, we would then have A, A_2, A_3, A_4, A_5.
> If we have 2 A models and an A_2 model, it would then look like: A, A_2, A_2_2. This will always guarantee uniqueness, will be consistent across different fit calls, and is intuitive to the user.

We could implement this solution without `_2_2` as well. However then, the user could provide a name ending with `_\d+$` and we would disregard if such a name was provided by the user or assigned by the library in a previous round. That is, the user can have their own reasons for appending integers, such as versioning, and we would have to either 1/override these numbers to stay consistent with your ask in 

> my suggested implementation would be:
```
invalid_model_names (those from previous calls to fit): ["A", "B", "B_2"]
new default model names being pushed: ["B", "B"]
the output is ["B_3", "B_4"]
```

or 2/somehow keep track of a global state of all assigned/user-provided names also outside the scope of this function, maybe in a Trainer instance variable. I think the current version is a good tradeoff but I'm eager to hear your thoughts.

canerturkmen: tl; dr. In the above example, the user could be providing `B_2` as their own numbering scheme and we would be just incrementing the number to 3 regardless.

Innixma: I'm not saying that if the user provides B_2 it should become B_3, it should become B_2_2 in this case.

However, in the example, the user explicitly is passing only "B", NOT "B_2":

```
invalid_model_names (those from previous calls to fit): ["A", "B", "B_2"]
new default model names being pushed: ["B", "B"]
the output is ["B_3", "B_4"]
```

If they instead pass "B_2" (and more complex options for sake of clarity):
```
invalid_model_names (those from previous calls to fit): ["A", "B", "B_2"]
new default model names being pushed: ["B", "B_2", "B", "B_2", "B_2_3"]
the output is ["B_3", "B_2_2", "B_4", "B_2_3", "B_2_3_2"]
```

Does this make sense? We never touch the default model name, only append a suffix to it.

This is achieved via:

```
while model.name in all_assigned_names:
    increment += 1
    model = model_type(name=f"{name_stem}_{increment}", **model_type_kwargs)
```

szha:
Job PR-1657-10 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1657/10/index.html

PR #1669, opened 2022-04-16
Use class property to set keys

zhiqiangdon:
Use class property to set keys in model classes. In this way, if we customize the `prefix` key, other keys are automatically updated.


szha:
Job PR-1669-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1669/1/index.html

sxjscience (approved):

Issue #1684, opened 2022-04-30
[BUG]


leandroimail:
- [ x ] I have checked that this bug exists on the latest stable version of AutoGluon
- [ x ] and/or I have checked that this bug exists on the latest mainline of AutoGluon via source installation

**Describe the bug**

autogluon/core/trainer/abstract_trainer.py", line 2436, in _get_feature_prune_proxy_model
    for model_name, type_inner in candidate_models_type_inner.items():

RuntimeError: dictionary changed size during iteration

Code in this line tries to change a dictionary in runtime into a for. Python does not allow to do this. Code bellow:

        for model_name, type_inner in candidate_models_type_inner.items():
            if type_inner in banned_models:
                banned_names.append(model_name)
                candidate_models_type_inner.pop(model_name, None)

When a change a code to make a copy of a dictionary that will be used in for, the issue solve. Example code below:

        for model_name, type_inner in candidate_models_type_inner.copy().items():
            if type_inner in banned_models:
                banned_names.append(model_name)
                candidate_models_type_inner.pop(model_name, None)

**Expected behavior**

Do not trow a RumtimeError

Innixma:
Thanks for the bug report and fix, I've merged the PR!

PR #1685, opened 2022-04-30
fix [BUG] #1684

leandroimail:
*Issue #, if available:*

#1684 

*Description of changes:*

autogluon/core/trainer/abstract_trainer.py", line 2436, in _get_feature_prune_proxy_model
for model_name, type_inner in candidate_models_type_inner.items():

**RuntimeError: dictionary changed size during iteration**


Innixma (approved):
Thanks for catching this bug, the fix looks good to me!

Issue #1635, opened 2022-03-30
[BUG] Catboost tabular does not work with f1 as eval_metric


jhmenke:
- [x] I have checked that this bug exists on the latest stable version of AutoGluon
- [ ] and/or I have checked that this bug exists on the latest mainline of AutoGluon via source installation

**Describe the bug**
When i set "f1" as the eval_metric for a TabularPredictor using _CPU_, Catboost will yield a big stacktrace of exceptions. If i change the eval_metric to, e.g., "log_loss", it works as expected.

**Expected behavior**
No exceptions being thrown and training continuing as usual.

**To Reproduce**
I can create a minimal script including my data if the error cannot be reproduced using toy data. I guess the stack trace should already be enough information.

**Installed Versions**
0.4.0
<details>
INSTALLED VERSIONS
------------------
date              : 2022-03-30
time              : 10:53:50.820393
python            : 3.9.11.final.0
OS                : Linux
OS-release        : 5.15.11-76051511-generic
Version           : #202112220937~1640185481~21.04~b3a2c21-Ubuntu SMP Mon Jan 3 16:5
machine           : x86_64
processor         : x86_64
num_cores         : 12
cpu_ram_mb        : 15746
cuda version      : None
num_gpus          : 0
gpu_ram_mb        : []
avail_disk_size_mb: 246143

autogluon.common  : 0.4.0
autogluon.core    : 0.4.0
autogluon.features: 0.4.0
autogluon.tabular : 0.4.0
boto3             : 1.21.29
catboost          : 1.0.4
dask              : 2021.11.2
distributed       : 2021.11.2
fastai            : 2.5.5
lightgbm          : 3.3.2
matplotlib        : 3.5.1
networkx          : 2.7.1
numpy             : 1.22.3
pandas            : 1.3.5
psutil            : 5.8.0
ray               : 1.7.0
requests          : 2.27.1
scipy             : 1.7.3
sklearn           : 1.0.2
torch             : 1.10.2+cu102
tqdm              : 4.63.1
xgboost           : 1.4.2

</details>

**Additional context**

`TabularPredictor(label=target_col, path=save_dir, problem_type="binary", eval_metric="f1")`

Stack trace (last exceptions is repeatedly many times):

> ...
> Fitting model: CatBoost_BAG_L1 ... Training model for up to 57563.37s of the 86377.63s of remaining time.
[...]

PR #1690, opened 2022-05-02
Fix CatBoost custom metric & F1 support

Innixma:
*Issue #, if available:*
#1635 

*Description of changes:*
- Fix CatBoost custom metric & F1 support

### DIFF
--- a/tabular/src/autogluon/tabular/models/catboost/catboost_utils.py
+++ b/tabular/src/autogluon/tabular/models/catboost/catboost_utils.py
@@ -25,104 +25,48 @@ class CustomMetric:
         raise NotImplementedError
 
 
-class BinaryCustomMetric(CustomMetric):
-    @staticmethod
-    def _get_y_pred_proba(approxes):
-        return np.array(approxes[0])
-
-    @staticmethod
-    def _get_y_pred(y_pred_proba):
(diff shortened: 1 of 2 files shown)


tabular/src/autogluon/tabular/models/catboost/catboost_utils.py
```diff
@@ -25,104 +25,48 @@ def evaluate(self, approxes, target, weight):
...
-    return metric_class(metric=metric, is_higher_better=is_higher_better, needs_pred_proba=needs_pred_proba)
+    elif problem_type == BINARY:
+        metric_map = dict(
+            log_loss='Logloss',
+            accuracy='Accuracy',
+            roc_auc='AUC',
+            f1='Logloss',
+            f1_macro='Logloss',
+            f1_micro='Logloss',
+            f1_weighted='Logloss',
```
gradientsky, line 42: nit: maybe add a comment why we can't use F1

gradientsky (approved):

yinweisu (approved):

tabular/src/autogluon/tabular/models/catboost/catboost_utils.py
Innixma: added

Issue #1700, opened 2022-05-04
[BUG] v0.4 Install instructions torch cpu install overridden


Innixma:
- [x] I have checked that this bug exists on the latest stable version of AutoGluon
- [x] and/or I have checked that this bug exists on the latest mainline of AutoGluon via source installation

**Describe the bug**
Following [CPU install instructions for linux](https://auto.gluon.ai/stable/install.html), GPU torch is installed instead.

**Expected behavior**
CPU version of torch should remain installed, GPU version of torch should not be installed or downloaded.

**Solution**
Update CPU install instructions to:

```
pip3 install "torch>=1.0,<1.11+cpu" -f https://download.pytorch.org/whl/cpu/torch_stable.html
```

**To Reproduce**
Colab Notebook: https://colab.research.google.com/drive/1BQYST3tRAiXxJ8T9D-CfmA1nE-MNRjRe?usp=sharing
Colab Notebook with Fix: https://colab.research.google.com/drive/11ALE_bWKgk0sVaoMgYjdNx6FYMZDDhlk?usp=sharing

**Installed Versions**
0.4.0

sxjscience:
Will we meet issues if we are training on a cpu-only instance with the torch-GPU? I think torch is loading cuda libs dynamically and the GPU version may still be runnable in CPU. 

If not, it may not be a bug (the users can still use AutoGluon) but we can provide improved instructions.

Innixma:
@sxjscience It will still work, but GPU is 800MB vs CPU 200 MB.

sxjscience:
I think this can be closed since the PR has been merged.

PR #1701, opened 2022-05-04
Fix torch CPU install instructions

Innixma:
*Issue #, if available:*
#1700

*Description of changes:*
- Fixes torch CPU install instructions to not be overwritten
- Switch to using `pip` instead of `pip3` for install instructions.

### DIFF
--- a/docs/install-cpu-pip.rst
+++ b/docs/install-cpu-pip.rst
@@ -1,10 +1,10 @@
 .. code-block:: bash
 
-    pip3 install -U pip
-    pip3 install -U setuptools wheel
+    pip install -U pip
+    pip install -U setuptools wheel
 
     # CPU version of pytorch has smaller footprint - see installation instructions in
     # pytorch documentation - https://pytorch.org/get-started/locally/
-    pip3 install torch==1.10.1+cpu -f https://download.pytorch.org/whl/cpu/torch_stable.html
+    pip install "torch>=1.0,<1.11+cpu" -f https://download.pytorch.org/whl/cpu/torch_stable.html
 
-    pip3 install --pre autogluon
+    pip install --pre autogluon
--- a/docs/install-cpu-pip.rst
+++ b/docs/install-cpu-pip.rst
@@ -1,10 +1,10 @@
 .. code-block:: bash
 
-    pip install -U pip
-    pip install -U setuptools wheel
+    pip3 install -U pip
+    pip3 install -U setuptools wheel
 
     # CPU version of pytorch has smaller footprint - see installation instructions in
     # pytorch documentation - https://pytorch.org/get-started/locally/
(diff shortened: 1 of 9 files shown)


yinweisu:
Why change from `pip3` to `pip`?

docs/install-cpu-pip.rst
```diff
@@ -1,10 +1,10 @@
...
 
-    pip3 install -U pip
-    pip3 install -U setuptools wheel
+    pip install -U pip
+    pip install -U setuptools wheel
 
     # CPU version of pytorch has smaller footprint - see installation instructions in
     # pytorch documentation - https://pytorch.org/get-started/locally/
-    pip3 install torch==1.10.1+cpu -f https://download.pytorch.org/whl/cpu/torch_stable.html
+    pip install "torch>=1.0,<1.11+cpu" -f https://download.pytorch.org/whl/cpu/torch_stable.html
```
gradientsky, line 8: `1.0` left boundary looks too wide: I would suggest doing at least `>=1.10`, otherwise we have a risk having environments with `1.0.8`, which are too old and possibly incompatible with other libraries in the bundle (`fastai` for instance)

docs/install-gpu-pip-windows.rst
```diff
@@ -1,6 +1,6 @@
 .. code-block:: bash
 
-    pip3 install -U pip
-    pip3 install -U setuptools wheel
+    pip install -U pip
+    pip install -U setuptools wheel
 
-    pip3 install --pre autogluon
+    pip install --pre autogluon
```
gradientsky, line 6: Sanity check: do we want `--pre` here?

docs/install-cpu-pip.rst
Innixma: I am merely matching our version range we have in setup.py. Changing the range can be a separate conversation for v0.5

docs/install-gpu-pip-windows.rst
Innixma: Yes. --pre is removed only in the `v0.x.y` and `stable` branches, not in master.

Innixma:
> Why change from `pip3` to `pip`?

Was based on a comment from Alex Smola, asking him now for clarification on the exact reason.

szha:
Job PR-1701-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1701/2/index.html

yinweisu:
@Innixma  Thanks for look into this. Yes, I thought `pip` could be problematic, for example some macs are pre-installed with python2. Unless the user added an alias to let `pip` point to `pip3`, the `pip` command will actually install into python2

Innixma:
@gradientsky @yinweisu CI passed, anything else before approvals?

yinweisu (approved):

PR #1689, opened 2022-05-02
Update resource logic in models

Innixma:
*Issue #, if available:*
#1676

*Description of changes:*
Simplified logic added in #1676 and made it more flexible.

- Previous implementation would crash if user specified `'auto'` for resources, fixed in this PR.
- Added `get_minimum_resources` to explicitly define minimum resource requirements within a method.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1702, opened 2022-05-05
Fix repeated bagging

Innixma:
*Issue #, if available:*
#1689 (Introduced first bug, but 2nd and 3rd were already present but rare / inactive)

*Description of changes:*
- Fix repeated bagging failing to fit repeats beyond the first when time limit is specified due to `self.model_base` being None in the loaded model (need to use `self._get_model_base()`).
- Fixed HPO not working for bagged models if the bagged model is loaded from disk (same issue as above).
- Fixed Feature importance erroring if `self.model_best` is None (can happen if no Weighted Ensemble is fit). Now it does same logic as `predict` to find best model if `self.model_best` is None.

### DIFF
--- a/core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
+++ b/core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
@@ -988,10 +988,10 @@ class BaggedEnsembleModel(AbstractModel):
         return memory_size
 
     def validate_fit_resources(self, **kwargs):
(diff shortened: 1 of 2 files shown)


szha:
Job PR-1702-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1702/1/index.html

yinweisu (approved):

gradientsky (approved):

PR #1710, opened 2022-05-06
Make AutoMM dataloader return feature column information

zhiqiangdon:
To do feature column level distillation, we need to let AutoMM dataloader return feature column information.

1. Add feature column names in data processors.
2. Dataframe preprocessor outputs dictionary instead of list.
3. Data processors's return contains feature column information.
4. Use property for return keys in data processors.
5. CLIP uses eos token as the cls token. So avoid using eos as the sep token for CLIP.

### DIFF
--- a/text/src/autogluon/text/automm/data/process_categorical.py
+++ b/text/src/autogluon/text/automm/data/process_categorical.py
@@ -76,6 +76,7 @@ class CategoricalProcessor:
         A dictionary containing the processed categorical features.
         """
         ret = {}
+        # TODO: consider moving this for loop into __init__() since each sample has the same information.
         for i, col_name in enumerate(categorical_features.keys()):
             ret[f"{self.categorical_column_prefix}_{col_name}"] = i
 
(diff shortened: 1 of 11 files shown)


szha:
Job PR-1710-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1710/2/index.html

text/src/autogluon/text/automm/data/process_categorical.py
```diff
@@ -57,16 +75,20 @@ def process_one_sample(
         -------
         A dictionary containing the processed categorical features.
         """
-        return {
-            f"{self.prefix}_{CATEGORICAL}": categorical_features
-        }
+        ret = {}
+        for i, col_name in enumerate(categorical_features.keys()):
```
sxjscience, line 79: You may add a TODO item about the potential performance issue.

zhiqiangdon: Done. We may move the for loop into __init__(), but need to consider the backward compatibility.

bryanyzhu (approved):

sxjscience (approved):

szha:
Job PR-1710-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1710/3/index.html

PR #1711, opened 2022-05-09
Optionally turn on returning feature column information

zhiqiangdon:
We optionally turn on returning feature column information for the case of feature column distillation. By default, we turn this functionality off to reduce latency of the dataloader.

1. Add a requires_column_info flag in data processors.
2. Add a util function to turn this flag on or off.


szha:
Job PR-1711-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1711/1/index.html

sxjscience:
What will be the impact on data loading speed?

PR #1718, opened 2022-05-10
Extract column features from AutoMM's model outputs

zhiqiangdon:
1. Add one util function to extract column features for both image and text.
2. Support extracting column features for models `timm_image`, `hf_text`, and `clip`.

### DIFF
--- a/text/src/autogluon/text/automm/models/clip.py
+++ b/text/src/autogluon/text/automm/models/clip.py
@@ -2,13 +2,17 @@ import torch
 import logging
 from torch import nn
 from transformers import CLIPModel
-from .utils import assign_layer_ids
+from .utils import (
+    assign_layer_ids,
+    init_weights,
+    get_column_features,
+)
 from ..constants import (
     IMAGE, IMAGE_VALID_NUM, TEXT_TOKEN_IDS,
-    TEXT_VALID_LENGTH, LABEL, LOGITS, FEATURES, AUTOMM
+    TEXT_VALID_LENGTH, LABEL, LOGITS, FEATURES,
+    AUTOMM, COLUMN,
 )
 from typing import Optional
-from .utils import init_weights
 
 logger = logging.getLogger(AUTOMM)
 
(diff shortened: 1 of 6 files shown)


szha:
Job PR-1718-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1718/1/index.html

text/src/autogluon/text/automm/models/clip.py
```diff
@@ -98,28 +118,60 @@ def forward(
 
         assert images.dim() == 5
         b, n, c, h, w = images.shape
-        image_features = self.model.get_image_features(
+        vision_outputs = self.model.vision_model(
```
bryanyzhu, line 121: Just a question, why we want to do this naming change, like from image -> vision?

text/src/autogluon/text/automm/models/fusion.py
```diff
@@ -67,6 +67,7 @@ def __init__(
             Basically, it supports adding an auxilliary loss for each individual model.
         """
         super().__init__()
+        logger.debug("initializing MultimodalFusionMLP")
```
bryanyzhu, line 70: Do we want to leave debug logger here?

bryanyzhu (commented):
Nice work. Just left two questions from my side, will approve once I get a better understanding.

text/src/autogluon/text/automm/models/clip.py
zhiqiangdon: Changing from `get_image_features` to `vision_model` is to be consistent with changing from `get_text_features` to `text_model`. See https://github.com/huggingface/transformers/blob/v4.14.1/src/transformers/models/clip/modeling_clip.py#L1006-L1010

text/src/autogluon/text/automm/models/fusion.py
zhiqiangdon: This is to be consistent with the debug loggers in other models' init functions. The debug messages are helpful to track whether code is run as expected in the debug mode. By default, debug message is not printed unless the `verbosity` hyper-parameter is set to 4.

text/src/autogluon/text/automm/models/clip.py
bryanyzhu: I see, thank you for the reference.

bryanyzhu (approved):

sxjscience (approved):

PR #1720, opened 2022-05-10
Add frequency field to time series dataframes

canerturkmen:
*Issue #, if available:*

*Description of changes:*
Add a frequency property to time series data frames. The field is inferred directly via pandas from the multiindex. + minor fixes and additions.

### DIFF
--- a/forecasting/tests/unittests/test_ts_dataset.py
+++ b/forecasting/tests/unittests/test_ts_dataset.py
@@ -36,7 +36,7 @@ SAMPLE_ITERABLE = [
 ]
 
 
-def test_from_iteratble():
+def test_from_iterable():
     ts_df = TimeSeriesDataFrame(SAMPLE_ITERABLE)
     pd.testing.assert_frame_equal(ts_df, SAMPLE_TS_DATAFRAME, check_dtype=True)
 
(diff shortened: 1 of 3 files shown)


canerturkmen:
@huibinshen

szha:
Job PR-1720-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1720/1/index.html

forecasting/tests/unittests/test_ts_dataset.py
```diff
@@ -243,5 +243,47 @@ def test_subsequence(start_timestamp, end_timestamp, item_ids, datetimes, target
...
+@pytest.mark.parametrize(
+    "timestamps, expected_freq",
+    [
+        (["2020-01-01 00:00:00", "2020-01-02 00:00:00", "2020-01-03 00:00:00"], "D"),
+        (["2020-01-01 00:00:00", "2020-01-03 00:00:00", "2020-01-05 00:00:00"], "2D"),
+        (["2020-01-01 00:00:00", "2020-01-01 00:01:00", "2020-01-01 00:02:00"], "T"),
+        (["2020-01-01 00:00:00", "2020-01-01 01:00:00", "2020-01-01 02:00:00"], "H"),
+    ],
+)
+def test_when_dataset_constructed_from_dataframe_without_freq_then_freq_is_inferred(
```
Innixma, line 255: What about a test for an example that is invalid (not consistent frequency)? Such as:

["2020-01-01 00:00:00", "2020-01-02 00:00:00", "2020-01-04 00:00:00"]

Maybe this already exists. If so no action is required.

Innixma (approved):
LGTM, would recommend adding an extra test if it doesn't already exist.

forecasting/tests/unittests/test_ts_dataset.py
canerturkmen: in GluonTS we always assume uniformly sampled (perfectly regular) time series, with an explicit NaN if the time step is missing. We are (somewhat implicitly) assuming this for our data frames as well.

Innixma: This still doesn't explain what happens when it isn't uniformly sampled (should we throw an exception, and is there a test that covers this situation?)

canerturkmen: Added functionality to validate and added test.

PR #1721, opened 2022-05-10
Fix a unit test and improve top k averaging

zhiqiangdon:
1. Fix a unit test by cleaning up saving path.
2. Support resuming top k averaging if encountering failures.
3. Replace `per_gpu_batch_size_evaluation` with `eval_batch_size_ratio`. Users only needs to override `per_gpu_batch_size`. `per_gpu_batch_size_evaluation` would be automatically updated.

### DIFF
--- a/text/src/autogluon/text/automm/configs/environment/default.yaml
+++ b/text/src/autogluon/text/automm/configs/environment/default.yaml
@@ -3,7 +3,7 @@ env:
   num_nodes: 1
   batch_size: 128  # this is a desired batch size; pl trainer will accumulate gradients when per step batch is smaller.
   per_gpu_batch_size: 8  # training per gpu batch size
-  per_gpu_batch_size_evaluation: 64  # evaluation per gpu batch size
+  eval_batch_size_ratio: 4  # per_gpu_batch_size_evaluation = per_gpu_batch_size * eval_batch_size_ratio
   precision: 16  # training precision. Refer to https://pytorch-lightning.readthedocs.io/en/latest/advanced/mixed_precision.html
   num_workers: 2  # pytorch training dataloader workers
   num_workers_evaluation: 2  # pytorch prediction/test dataloader workers
--- a/text/src/autogluon/text/automm/configs/environment/default.yaml
+++ b/text/src/autogluon/text/automm/configs/environment/default.yaml
@@ -4,6 +4,7 @@ env:
   batch_size: 128  # this is a desired batch size; pl trainer will accumulate gradients when per step batch is smaller.
   per_gpu_batch_size: 8  # training per gpu batch size
   eval_batch_size_ratio: 4  # per_gpu_batch_size_evaluation = per_gpu_batch_size * eval_batch_size_ratio
+  per_gpu_batch_size_evaluation:  # This is deprecated. Use eval_batch_size_ratio instead.
   precision: 16  # training precision. Refer to https://pytorch-lightning.readthedocs.io/en/latest/advanced/mixed_precision.html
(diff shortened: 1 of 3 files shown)


szha:
Job PR-1721-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1721/1/index.html

sxjscience (approved):

text/src/autogluon/text/automm/configs/environment/default.yaml
```diff
@@ -3,7 +3,7 @@ env:
   num_nodes: 1
   batch_size: 128  # this is a desired batch size; pl trainer will accumulate gradients when per step batch is smaller.
   per_gpu_batch_size: 8  # training per gpu batch size
-  per_gpu_batch_size_evaluation: 64  # evaluation per gpu batch size
```
sxjscience, line 6: You should keep the old key. Lots of users are using the key. You may consider to add a new key and change the default value of the old key to null.

zhiqiangdon: We already make this change backward compatible here https://github.com/awslabs/autogluon/pull/1721/files#diff-9a3106c0c6088dce9389c316f54a34d170a1080af4121fba0c194359f676d89eR935

It's kind of misleading if we keep the old key but not use it.

sxjscience: What's our general design guideline for adding new keys?

sxjscience: Also, it will be problematic if the users are revising the `per_gpu_batch_size_evaluation` via overrides. They will be adding a new key rather than updating an existing key.

sxjscience: Adding a new key is designed to raise an exception in https://github.com/awslabs/autogluon/blob/c03d7ba762b1982113b797fd9e8ee3559d2c3a43/text/src/autogluon/text/automm/utils.py#L217

sxjscience: Since this PR will go into 0.4.1, we need to make it as backward compatible as possible.

zhiqiangdon: Let's say we still keep `per_gpu_batch_size_evaluation ` but not actually use it. If users override `per_gpu_batch_size_evaluation `, it doesn't behave as they expected. Is this OK?

szha:
Job PR-1721-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1721/3/index.html

text/src/autogluon/text/automm/configs/environment/default.yaml
sxjscience: We may just change the default value to null.

bryanyzhu (approved):

text/src/autogluon/text/automm/configs/environment/default.yaml
zhiqiangdon: Done.

szha:
Job PR-1721-4 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1721/4/index.html

szha:
Job PR-1721-5 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1721/5/index.html

PR #1734, opened 2022-05-19
Improve AutoMM's validation metric computation

zhiqiangdon:
We avoid computing the metric score at each validation step through the following:
1. logging metric object instead of score at each validation step. 
2. using `metric.update()` instead of `metric()`.

In this way, we can solve the previous issues regarding the `r2` (error with batch size 1) and `roc_auc` (wrong score and warning: No positive/negative samples in targets, false negative/positive value should be meaningless.)

Now, it's safe to use the validation metrics `r2`, `roc_auc`, etc.

### DIFF
--- a/text/src/autogluon/text/automm/optimization/lit_module.py
+++ b/text/src/autogluon/text/automm/optimization/lit_module.py
@@ -133,18 +133,20 @@ class LitModule(pl.LightningModule):
             ) * weight
         return loss
 
-    def _compute_metric(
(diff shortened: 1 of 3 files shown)


szha:
Job PR-1734-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1734/1/index.html

sxjscience (approved):

PR #1745, opened 2022-05-22
remove direct dependency link to mxnet from timeseries

canerturkmen:
*Issue #, if available:*

*Description of changes:*

Removed direct mxnet (CPU version) dependency from autogluon forecasting and added a version check in setup.py instead.


forecasting/setup.py
```diff
@@ -26,10 +28,17 @@
...
+    f'autogluon.common=={version}',
 ]
 
+try:
+    from mxnet import __version__ as mxnet_version
+    assert vparse(mxnet_version) < vparse("2.0")
+except (ImportError, AssertionError):
+    raise ImportError("autogluon.forecasting depends on Apache MxNet v1.x. Please install a suitable version of "
+                      "MxNet before continuing with the installation.")
+
```
Innixma, line 41: Does it really support any version `<2.0`? Should we provide a lower bound as well?

forecasting/setup.py
```diff
@@ -26,10 +28,17 @@
     "pandas",
     "psutil>=5.7.3,<5.9",
     "gluonts>=0.8.0",
-    "mxnet<2.0",
```
Innixma, line 29: What would happen if we released forecasting today and had it install when `pip install autogluon` was called? Would the user get an error during install if they didn't have MXNet installed prior? If so, why error instead of installing MXNet automatically? Do any of forecasting's dependencies forcibly install MXNet?

Innixma: For v0.5 release, are we planning to continue having a hard dependency on MXNet? If so we must choose one of the following options:

1. Consider not having forecasting install with `pip install autogluon` and instead request the user to specifically install `pip install autogluon.forecasting`.
2. Install forecasting with autogluon and don't auto-install mxnet, then if the user tries to import forecasting they get an error if they didn't have mxnet installed which tells them to install mxnet. This is what currently happens with autogluon's object detector which only works with MXNet.
3. (Not Recommended) Forecasting installs with autogluon, forcibly installs mxnet when calling `pip install autogluon`.

Open to hearing your thoughts.

canerturkmen: also tagging @gradientsky as this PR grew from a discussion with him.

Yes, if we released like this, installation would fail if mxnet is not installed. Also, I don't think any of the dependencies forcibly install it. The reason is as follows.

`mxnet`, in PyPI, installs the library without CUDA bindings, defaulting to no GPU usage. If we include `mxnet` in dependencies, it reinstalls the library in a CPU-only mode, which was actually the issue uncovered by CI in this PR. CUDA-ready versions go by `mxnet-cu***`.

Configured like this, users are expected to bring their own mxnet, often via a container configuration. This is also the path followed by GluonTS. See their [README](https://github.com/awslabs/gluon-ts).

For the v0.5 we will definitely need mxnet for healthy operation. I'd prefer option 2. I am good with option 1, where we can ask the user to install with 

```pip install autogluon.forecasting mxnet~=1.7```

Another path we might take is to put more time into actually averting mxnet altogether, and depending on only pytorch models in GluonTS. This would shrink the model zoo in the interim.

**Comment by szha** [comment]
Job PR-1745-9 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1745/9/index.html

File: Jenkinsfile [hunk 3]
```diff
@@ -1,7 +1,7 @@
 max_time = 180
 
 setup_mxnet_gpu = """
-    python3 -m pip install mxnet-cu102==1.8.*
+    python3 -m pip install mxnet-cu101==1.7.*
```
gradientsky, line 4: 1) this mxnet version have outstanding CVEs opened against it. We should upgrade to 1.9.x 
2) cu101 might be an issue with containers: our tests showed 10.2 is needed.

canerturkmen: cu101 is the only one I got it to work with the CI. will check again re mxnet version (this is coupled to the GluonTS requirement).

forecasting/src/autogluon/forecasting/__init__.py
```diff
@@ -1,10 +1,21 @@
...
 
+try:
+    from mxnet import __version__ as mxnet_version
+
+    assert parse("2.0") > parse(mxnet_version) >= parse("1.7")
+except (ImportError, AssertionError):
+    raise ImportError(
+        "autogluon.forecasting depends on Apache MxNet v1.7 or greater (below v2.0). "
+        "Please install a suitable version of MxNet in order to use autogluon.forecasting."
+    )
```
Innixma, line 17: Unclear how to install mxnet from this instruction (Ditto for warning in setup.py). Either provide a pip install command or a link to a website with more info.

Innixma (approved):
LGTM, have 1 minor comment. Please also sync with Alex Shirkov on his comment before merging.

forecasting/setup.py
gradientsky: +1 on option 2. These binaries are also pre-installed in DL AMI (which one depends on instance type and if supports GPU).

szha:
Job PR-1745-35 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1745/35/index.html

Jenkinsfile
canerturkmen: After a lot of experimentation, I was able to bump the mxnet in CI to 1.9, and make sure cuda 10.2 with cuDNN support was installed (via conda). This effects `timeseries` unit tests as well as tutorial build jobs. Note that the prod code does not include any constraints on the cuda version, and leaves the cuda/mxnet version up to the user.

gradientsky (approved):

PR #1747, opened 2022-05-23
Update cheatsheet links to GitHub instead of S3

Innixma:
*Issue #, if available:*

*Description of changes:*
- Update cheatsheet links to GitHub instead of S3.
- One minor downside to this is that the PDF link immediately downloads on chrome when using the GitHub link instead of previewing like it does with the S3 link. Thoughts on how to enable preview for the GitHub artifact are welcome.
- Note: We should do the same update to branch `0.4.1` and `0.4.0` but instead link the v0.4.0 GitHub artifacts (instead of stable).

### DIFF
--- a/docs/cheatsheet.rst
+++ b/docs/cheatsheet.rst
@@ -2,11 +2,11 @@ Cheat Sheet
 -----------
 
 .. image:: https://raw.githubusercontent.com/Innixma/autogluon-doc-utils/main/docs/cheatsheets/stable/autogluon-cheat-sheet.jpeg
-   :width: 300
+   :width: 900
 
 Download the PDF version with clickable links `here`_.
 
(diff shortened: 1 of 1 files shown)


szha:
Job PR-1747-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1747/1/index.html

yinweisu (approved):
Found one hack to do it with google doc preview,
for example: https://docs.google.com/viewer?url=https://raw.githubusercontent.com/Innixma/autogluon-doc-utils/main/docs/cheatsheets/stable/autogluon-cheat-sheet.pdf

Still not optimal.

Innixma:
Another option: https://nbviewer.org/github/Innixma/autogluon-doc-utils/blob/main/docs/cheatsheets/v0.4.0/autogluon-cheat-sheet.pdf

PR #1751, opened 2022-05-23
fix preset discrepancy between backends of TextPredictor

Qiaochu-Song:
*Issue #, if available:*

*Description of changes:*
Change the default `convert_to_text` in TextPredictor pytorch backend to False.
This will resolve the discrepancy between TextPredictor pytorch and mxnet backend default presets: 

- PyTorch backend (no categorical MLP when using the current default convert_to_text = True): https://github.com/awslabs/autogluon/blob/0.4.1/text/src/autogluon/text/automm/configs/data/default.yaml#L8 

- Mxnet backend: 
https://github.com/awslabs/autogluon/blob/0.4.1/text/src/autogluon/text/text_prediction/legacy_presets.py#L104

### DIFF
--- a/text/src/autogluon/text/automm/configs/data/default.yaml
+++ b/text/src/autogluon/text/automm/configs/data/default.yaml
@@ -5,7 +5,7 @@ data:
   categorical:
     minimum_cat_count: 100  # The minimum number of occurrences a category must have in the training data to avoid being considered a rare category.
     maximum_num_cat: 20  # The maximum amount of categories that can be considered non-rare.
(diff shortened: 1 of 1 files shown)


sxjscience:
Thanks for the contribution @Qiaochu-Song! We are going to deprecate the MXNet-based implementation (remove it in 0.5) so this is intended. We are developing all the latest features with pytorch.

sxjscience:
You may try out the PyTorch backend if you are using TextPredictor.

Qiaochu-Song:
@sxjscience Got it! Thanks for your reply. I will close this pr.

PR #1753, opened 2022-05-24
Improve AutoMM's binary classification's metrics by supporting customizing positive class

zhiqiangdon:
Previously, AutoMM always used label 1 as the positive label in computing the `roc_auc` and `average_precision` metrics in binary classification. However, 1, corresponding to class name label_encoder.classes_[1], may not always be the semantically positive class's label for specific tasks. For example, users provide string class names `a` and `b`, and wants `a` to be the semantically positive class. This PR makes it possible for users to customize their positive classes.

Related discussions:
https://github.com/scikit-learn/scikit-learn/issues/15303
https://github.com/scikit-learn/scikit-learn/pull/15405

1. Add a hyper-parameter `pos_label` (positive class name) in data config. We use `pos_label` to align with sklearn's APIs, e.g., https://scikit-learn.org/stable/modules/generated/sklearn.metrics.roc_curve.html#sklearn.metrics.roc_curve
2. Add a util to infer the encoded positive label from `pos_label`.
3. Add the `pos_label` choice in functions related to computing binary classification's metrics.
4. Make it backward compatible.

### DIFF
--- a/text/src/autogluon/text/automm/configs/data/default.yaml
+++ b/text/src/autogluon/text/automm/configs/data/default.yaml
@@ -10,4 +10,4 @@ data:
     convert_to_text: False  # Whether to convert the feature to text.
     scaler_with_mean: True  # Whether to normalize with mean.
     scaler_with_std: True  # Whether to normalize with std.
-  positive_class:  # The name of binary classification's positive class. It's used in computing some metrics, e.g., roc_auc. If not provided, then use label_encoder.classes_[1],
+  pos_label:  # The name of binary classification's positive class. It's used in computing some metrics, e.g., roc_auc. If not provided, then use label_encoder.classes_[1],
--- a/text/src/autogluon/text/automm/configs/data/default.yaml
+++ b/text/src/autogluon/text/automm/configs/data/default.yaml
@@ -10,3 +10,4 @@ data:
     convert_to_text: False  # Whether to convert the feature to text.
     scaler_with_mean: True  # Whether to normalize with mean.
     scaler_with_std: True  # Whether to normalize with std.
+  positive_class:  # The name of binary classification's positive class. It's used in computing some metrics, e.g., roc_auc. If not provided, then use label_encoder.classes_[1],
--- a/text/src/autogluon/text/automm/optimization/utils.py
+++ b/text/src/autogluon/text/automm/optimization/utils.py
@@ -46,7 +46,7 @@ def get_metric(
         metric_name: str,
         problem_type: str,
         num_classes: Optional[int] = None,
-        positive_label: Optional[int] = 1,
+        pos_label: Optional[int] = 1,
 ):
     """
     Obtain a torchmerics.Metric from its name.
@@ -60,7 +60,7 @@ def get_metric(
         The type of the problem.
     num_classes
         Number of classes, used in the quadratic_kappa metric for binary classification.
-    positive_label
+    pos_label
         The label (0 or 1) of binary classification's positive class, which is used in some metrics, e.g., AUROC.
 
     Returns
@@ -87,9 +87,9 @@ def get_metric(
         return torchmetrics.CohenKappa(num_classes=num_classes,
                                        weights="quadratic"), MAX, None
     elif metric_name == ROC_AUC:
-        return torchmetrics.AUROC(pos_label=positive_label), MAX, None
+        return torchmetrics.AUROC(pos_label=pos_label), MAX, None
     elif metric_name == AVERAGE_PRECISION:
-        return torchmetrics.AveragePrecision(pos_label=positive_label), MAX, None
+        return torchmetrics.AveragePrecision(pos_label=pos_label), MAX, None
     elif metric_name in [LOG_LOSS, CROSS_ENTROPY]:
         return torchmetrics.MeanMetric(), MIN, \
                functools.partial(F.cross_entropy, reduction="none")
@@ -106,7 +106,7 @@ def get_metric(
         elif problem_type == MULTICLASS:
             return torchmetrics.Accuracy(), MAX, None
         elif problem_type == BINARY:
-            return torchmetrics.AUROC(pos_label=positive_label), MAX, None
+            return torchmetrics.AUROC(pos_label=pos_label), MAX, None
         else:
             raise ValueError(f'The problem_type={problem_type} is currently not supported')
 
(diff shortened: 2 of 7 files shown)


text/src/autogluon/text/automm/configs/data/default.yaml
```diff
@@ -10,3 +10,4 @@ data:
     convert_to_text: False  # Whether to convert the feature to text.
     scaler_with_mean: True  # Whether to normalize with mean.
     scaler_with_std: True  # Whether to normalize with std.
+  positive_class:  # The name of binary classification's positive class. It's used in computing some metrics, e.g., roc_auc. If not provided, then use label_encoder.classes_[1],
```
sxjscience, line 13: How about following sklearn and still call it `pos_label`: https://scikit-learn.org/stable/modules/generated/sklearn.metrics.roc_curve.html#sklearn.metrics.roc_curve

bryanyzhu: +1 on `pos_label`. What about other names, like the function name `try_to_set_positive_label`, do we want to change it to `try_to_set_pos_label`?

Innixma:
Why not simply set the positive class to be 1 in the label transformation stage to internal representation?

Aka if 'Dog' is positive instead of 'Cat' 'Dog' = 1 0, and then tell roc_auc that 0 is positive, instead:

'Cat' 'Dog' = 0 1, don't need to tell roc_auc anything.

This is what is done in Tabular and works fine, no need to worry about specifying arguments to metrics.

Innixma (changes requested):
Refer to prior comment

text/src/autogluon/text/automm/optimization/utils.py
```diff
@@ -59,6 +60,8 @@ def get_metric(
         The type of the problem.
     num_classes
         Number of classes, used in the quadratic_kappa metric for binary classification.
+    positive_label
+        The label (0 or 1) of binary classification's positive class, which is used in some metrics, e.g., AUROC.
```
sxjscience, line 64: We may use `None` by default.

text/src/autogluon/text/automm/configs/data/default.yaml
sxjscience: I get the rationale for naming it `positive_class` now because we can specify it as a string and need to convert it to `pos_label`.

text/src/autogluon/text/automm/predictor.py
```diff
@@ -399,10 +400,16 @@ def fit(
             validation_metric_name = self._validation_metric_name
             eval_metric_name = self._eval_metric_name
 
+        if problem_type == BINARY:
+            try_to_set_positive_label(
+                data_config=config.data,
+                label_encoder=df_preprocessor.label_generator,
+            )
```
sxjscience, line 407: We may just return the pos_label.

text/src/autogluon/text/automm/optimization/utils.py
zhiqiangdon: Done.

text/src/autogluon/text/automm/configs/data/default.yaml
zhiqiangdon: Name changed.

zhiqiangdon:
> Why not simply set the positive class to be 1 in the label transformation stage to internal representation?
> 
> Aka if 'Dog' is positive instead of 'Cat' 'Dog' = 1 0, and then tell roc_auc that 0 is positive, instead:
> 
> 'Cat' 'Dog' = 0 1, don't need to tell roc_auc anything.
> 
> This is what is done in Tabular and works fine, no need to worry about specifying arguments to metrics, and automatically works for custom metrics from the user. I would strongly encourage the strategy I mention instead of what is done in this PR.
> 
> For more details on how to do this, please run a debugger on Tabular when specifying positive label explicitly to see how the code logic works. You can see the main logic here: https://github.com/awslabs/autogluon/blob/master/core/src/autogluon/core/data/label_cleaner.py#L175

Thanks for giving the suggestions!
After looking into the details of label cleaner, I think it may not be fully compatible with sklearn's LabelEncoder's APIs. Changing from the current LabelEncoder to label cleaner would result in other API changes in AutoMM, which can potentially bring new bugs. The current way is simpler and safer to handle this issue. In future, we can discuss the API alignment and maybe refactor `core`'s code to make it easier to support other modules.

szha:
Job PR-1753-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1753/2/index.html

text/src/autogluon/text/automm/utils.py
```diff
@@ -1221,3 +1225,38 @@ def turn_on_off_feature_column_info(
...
+
+    Returns
+    -------
+
+    """
+    if problem_type != BINARY:
+        return None
+
+    pos_label = OmegaConf.select(data_config, "pos_label", default=None)
+    if pos_label:
```
sxjscience, line 1257: Will it be problematic if pos_label is 0!

zhiqiangdon: Thanks. Fixed.

szha:
Job PR-1753-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1753/3/index.html

szha:
Job PR-1753-4 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1753/4/index.html

szha:
Job PR-1753-5 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1753/5/index.html

sxjscience:
We may add another test-case and it looks good overall. @Innixma I think relying on LabelEncoder is sufficient for now. Depending on LabelCleaner may cause overhead when we are going to later revise the label encoding mechanism (e.g, supporting other types of labels like text, image, bounding boxes, entities)

bryanyzhu (approved):

Innixma (approved):
Approving however I don't agree with the design.

Would greatly prefer to use a custom label encoder such as:

```
class PosLabelEncoder(LabelEncoder):
    def __init__(self, pos_label=None, **kwargs):
        super().__init__(**kwargs)
        self.pos_label = pos_label

    def fit(self, y):
        super().fit(y)
        if self.pos_label is not None:
            self.classes_ = np.array([c for c in self.classes_ if c != self.pos_label] + [self.pos_label])
```

PR #1756, opened 2022-05-25
Adding augmentation to text 

lzcemma:
*Description of changes:*

Add text augmentation, including synonym replacement, random swap and random delete for now.


text/src/autogluon/text/automm/data/process_text.py
```diff
@@ -397,4 +455,14 @@ def __call__(
...
+
+    #nlpaug couldn't be serialize, so not saving augmenter for now
+    def __deepcopy__(self, memo):
+        cls = self.__class__
+        result = cls.__new__(cls)
+        memo[id(self)] = result
+        for k, v in self.__dict__.items():
+            if(k!="train_augmenter"):
+                setattr(result, k, deepcopy(v, memo))
+        return result
```
sxjscience, line 468: nit: add newline

text/setup.py
```diff
@@ -41,6 +41,8 @@
     f'autogluon.features=={version}',
     f'autogluon.common=={version}',
     'autogluon-contrib-nlp==0.0.1b20220208',
+    'nlpaug',
```
Innixma, line 44: 1. missing min and max versions
2. How heavy is this package? Is it something we want as a required install?

lzcemma: nlpaug is about 2.2M, and it is used during text augmentation.

sxjscience: https://pypi.org/project/nlpaug/ . The latest version seems to be 1.1.10.

lzcemma: Yes, I made another commit specify the version.

text/src/autogluon/text/automm/data/process_text.py
```diff
@@ -260,13 +314,17 @@ def build_one_token_sequence_from_text(
             "Token indices sequence length is longer than.*result in indexing errors"
         )
         for col_name, col_text in text.items():
+            # augment text if specify in config
+            if (is_training and (len(col_text.split(' ')) >= 5)):
+                 # naive way detect text/categorical/numerical: more than 5 words
```
sxjscience, line 319: Should we put it as part of the config?

text/src/autogluon/text/automm/data/process_text.py
```diff
@@ -169,6 +180,46 @@ def collate_fn(self) -> Dict:
...
+        ----------
+        augment_types
+            A list of text augment types.
+
+        Returns
+        -------
+        A nlpaug sequantial flow.
+        
+        """
+        if(augment_types == None):
```
sxjscience, line 200: Consider change to `if augment_types is None:`.

text/src/autogluon/text/automm/data/process_text.py
lzcemma: I think augmentation should be applied on to text column, not categorical/numerical columns even though they are fed into text backbone. I will put the length for detection into config.

szha:
Job PR-1756-7 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1756/7/index.html

text/setup.py
```diff
@@ -41,6 +41,8 @@
     f'autogluon.features=={version}',
     f'autogluon.common=={version}',
     'autogluon-contrib-nlp==0.0.1b20220208',
+    'nlpaug>=1.1.10',
```
sxjscience, line 44: Let's upper bound it:

```
'nlpaug>=1.1.10,<2.0.0',
```

text/setup.py
```diff
@@ -41,6 +41,8 @@
...
+    'nltk>=3.4.5'
```
sxjscience, line 45: Upper bound: `'nltk>=3.4.5,<4.0.0'`

text/src/autogluon/text/automm/configs/model/fusion_mlp_image_text_tabular.yaml
```diff
@@ -62,6 +62,11 @@ model:
     insert_sep: True
     text_segment_num: 2
     stochastic_chunk: False
+    text_detection_length: 10
+    train_augment_types:
+      # - "synonym_replacement({'aug_p': 0.1})"  # less than 0.1 based on eda paper
```
sxjscience, line 67: Add  the following comment above to indicate that it's an example.
```
# example

```

text/src/autogluon/text/automm/data/process_text.py
```diff
@@ -168,6 +184,63 @@ def collate_fn(self) -> Dict:
         )
 
         return fn
+        
+    @staticmethod
+    def construct_augmenter(
+            augment_types: List[str],
+    ) -> naf.Sometimes:
```
sxjscience, line 191: No need to use static method. I prefer using a function.

text/src/autogluon/text/automm/data/process_text.py
```diff
@@ -129,6 +139,12 @@ def __init__(
 
         self.stochastic_chunk = stochastic_chunk
 
+        #construct augmentor
+        self.train_augment_types = train_augment_types
+        self.text_detection_length = text_detection_length
+        self.train_augmenter = self.construct_augmenter(train_augment_types)
+        print("===================self.train_augmenter",  self.train_augmenter)
```
sxjscience, line 146: remove print. Change to logger.debug

text/src/autogluon/text/automm/data/process_text.py
```diff
@@ -168,6 +184,63 @@ def collate_fn(self) -> Dict:
...
+            nltk.download('averaged_perceptron_tagger')
+
+        for aug_type in augment_types:
+            if '(' in aug_type:
+                trans_mode = aug_type[0:aug_type.find('(')]
+                kwargs = ast.literal_eval(aug_type[aug_type.find('('):])
+            else:
+                trans_mode = aug_type
+                kwargs = {}
+            if(trans_mode == "synonym_replacement"):
```
sxjscience, line 222: `if(trans_mode == "synonym_replacement")` --> `if trans_mode == "synonym_replacement"`

text/src/autogluon/text/automm/data/process_text.py
```diff
@@ -168,6 +184,63 @@ def collate_fn(self) -> Dict:
...
+                    nltk.data.find('corpora/wordnet')
+                except LookupError:
+                    nltk.download('wordnet')
+                try:
+                    nltk.data.find('corpora/omw-1.4')
+                except LookupError:
+                    nltk.download('omw-1.4')
+                    
+                auglist.append(naw.SynonymAug(**kwargs))
+            elif(trans_mode == "random_swap"):
```
sxjscience, line 235: Same here.

text/src/autogluon/text/automm/data/process_text.py
```diff
@@ -168,6 +184,63 @@ def collate_fn(self) -> Dict:
...
+                kwargs['action'] = 'swap'
+                auglist.append(naw.RandomWordAug(**kwargs))
+            elif(trans_mode == "random_delete"):
```
sxjscience, line 238: Same here.

text/src/autogluon/text/automm/data/process_text.py
```diff
@@ -397,4 +484,15 @@ def __call__(
...
+    #nlpaug couldn't be serialize, so not saving augmenter for now
+    def __deepcopy__(self, memo):
+        cls = self.__class__
+        result = cls.__new__(cls)
+        memo[id(self)] = result
+        
+        for k, v in self.__dict__.items():
+            if(k!="train_augmenter"):
+                setattr(result, k, deepcopy(v, memo))
+        return result
```
sxjscience, line 498: new line

text/src/autogluon/text/automm/data/process_text.py
```diff
@@ -397,4 +484,15 @@ def __call__(
...
+
+    #nlpaug couldn't be serialize, so not saving augmenter for now
+    def __deepcopy__(self, memo):
+        cls = self.__class__
+        result = cls.__new__(cls)
+        memo[id(self)] = result
+        
+        for k, v in self.__dict__.items():
+            if(k!="train_augmenter"):
+                setattr(result, k, deepcopy(v, memo))
```
sxjscience, line 497: What will be `result.train_augmenter` if we do this. Should we reconstruct it manually via
```
result.train_augmenter = construct_augmenter(result.train_augment_types)
```

Also, we need to test it.

text/src/autogluon/text/automm/utils.py
```diff
@@ -818,9 +819,9 @@ def load_text_tokenizers(
             per_path = os.path.join(path, per_text_processor.tokenizer)
             per_text_processor.tokenizer = per_text_processor.get_pretrained_tokenizer(
                 tokenizer_name=per_text_processor.tokenizer_name,
-                checkpoint_name=per_path,
-            )
-
+                checkpoint_name=per_path)
+            per_text_processor.train_augmenter = per_text_processor.construct_augmenter(
```
sxjscience, line 823: This line won't be necessary if you have correctly implemented `__deepcopy__`

text/src/autogluon/text/automm/data/process_text.py
```diff
@@ -260,13 +336,24 @@ def build_one_token_sequence_from_text(
             "Token indices sequence length is longer than.*result in indexing errors"
         )
         for col_name, col_text in text.items():
+
+            #if training
+            if(is_training):
+
```
sxjscience, line 342: Empty line.

text/src/autogluon/text/automm/data/process_text.py
lzcemma: deepcopy(naw.SynonymAug()) gives "cannot pickle '_io.BufferedReader' object". So as a work around, I am not saving the train_augmenter, and construct again during load.

sxjscience: I can understand the error. Basically, you can reconstruct `result.train_augmenter` after you have deep-copied the other attributes. Also, for support pickle/unpickle, you can also override `__getstate__` and `__setstate__`.

text/src/autogluon/text/automm/data/process_text.py
```diff
@@ -35,6 +41,59 @@
...
+        if trans_mode == "synonym_replacement": 
+            kwargs['aug_src'] = 'wordnet'
+            try:
+                nltk.data.find('corpora/wordnet')
+            except LookupError:
+                nltk.download('wordnet')
+            try:
+                nltk.data.find('corpora/omw-1.4')
+            except LookupError:
+                nltk.download('omw-1.4')
```
sxjscience, line 84: We may consider to later move these download to a separate `download_nltk_resources` function.

text/src/autogluon/text/automm/data/process_text.py
```diff
@@ -35,6 +41,59 @@
 }
 
 
+def construct_augmenter(
+        augment_types: List[str],
+) -> naf.Sometimes:
```
sxjscience, line 46: Actually you may change the return type to `Optional[naf.Sometimes]` since it can be None.

szha:
Job PR-1756-11 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1756/11/index.html

szha:
Job PR-1756-12 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1756/12/index.html

szha:
Job PR-1756-13 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1756/13/index.html

text/src/autogluon/text/automm/utils.py
```diff
@@ -820,9 +821,9 @@ def load_text_tokenizers(
             per_path = os.path.join(path, per_text_processor.tokenizer)
             per_text_processor.tokenizer = per_text_processor.get_pretrained_tokenizer(
                 tokenizer_name=per_text_processor.tokenizer_name,
-                checkpoint_name=per_path,
-            )
-
+                checkpoint_name=per_path)
+            # per_text_processor.train_augmenter = per_text_processor.construct_augmenter(
+            #     augment_types= per_text_processor.train_augment_types)
```
sxjscience, line 826: Remove comments.

sxjscience (approved):
LGTM

text/src/autogluon/text/automm/data/process_text.py
```diff
@@ -35,6 +41,59 @@
 }
 
 
+def construct_augmenter(
```
zhiqiangdon, line 44: I think making this a static method inside `TextProcessor` is better. There are three reasons:

1. This is mainly used by `TextProcessor`. If it aims to be a general util, then we need to move it to `data/utils.py`.
2. To be be consistent with other processors. Static methods also exist in other processors. Each processor file has one processor class with its class methods and closely related static methods.
3. The name `construct_augmenter` sounds like a general name, but it is to construct text augmenter. So, it is better to put inside `TextProcessor`.

text/src/autogluon/text/automm/data/process_text.py
```diff
@@ -260,13 +331,21 @@ def build_one_token_sequence_from_text(
             "Token indices sequence length is longer than.*result in indexing errors"
         )
         for col_name, col_text in text.items():
+            #if training
+            if(is_training):
+                #if augment
+                if( self.train_augmenter is not None  ):
```
zhiqiangdon, line 337: Need to remove the redundant `()` in if conditions. Double check other places.

text/src/autogluon/text/automm/data/process_text.py
```diff
@@ -35,6 +41,59 @@
...
+        nltk.download('averaged_perceptron_tagger')
+    for aug_type in augment_types:
+        if '(' in aug_type:
+            trans_mode = aug_type[0:aug_type.find('(')]
+            kwargs = ast.literal_eval(aug_type[aug_type.find('('):])
+        else:
+            trans_mode = aug_type
+            kwargs = {}
+        if trans_mode == "synonym_replacement": 
+            kwargs['aug_src'] = 'wordnet'
```
zhiqiangdon, line 76: I recommend to use `"` instead of `'` to be consistent with the existing code.

text/src/autogluon/text/automm/configs/model/fusion_mlp_image_text_tabular.yaml
```diff
@@ -62,6 +62,12 @@ model:
     insert_sep: True
     text_segment_num: 2
     stochastic_chunk: False
+    text_detection_length: 10
+    train_augment_types:
+      # example
```
zhiqiangdon, line 67: Need to add the paper's link.

text/src/autogluon/text/automm/configs/model/fusion_mlp_image_text_tabular.yaml
```diff
@@ -95,6 +101,12 @@ model:
     insert_sep: False
     text_segment_num: 1
     stochastic_chunk: False
+    text_detection_length: 10
+    train_augment_types:
```
zhiqiangdon, line 105: `train_augment_types` is a general name. We need to try distinguish this key from image transformation keys. How about `text_augment_types` or `text_train_augment_types`? Do we need to emphasize `train` in this key?

text/src/autogluon/text/automm/data/process_text.py
sxjscience: I think it's a generic function and can be useful elsewhere. Thus, staticmethod may not be a good design.

text/src/autogluon/text/automm/utils.py
```diff
@@ -820,9 +821,7 @@ def load_text_tokenizers(
             per_path = os.path.join(path, per_text_processor.tokenizer)
             per_text_processor.tokenizer = per_text_processor.get_pretrained_tokenizer(
                 tokenizer_name=per_text_processor.tokenizer_name,
-                checkpoint_name=per_path,
-            )
-
+                checkpoint_name=per_path)
```
zhiqiangdon, line 824: This change is unnecessary and violates the format of `()` in our code.

text/src/autogluon/text/automm/data/process_text.py
sxjscience: It should be the same. We may change it after we have adopted black.

text/src/autogluon/text/automm/configs/model/fusion_mlp_image_text_tabular.yaml
lzcemma: I follow the style under timm_image. Since train_augment_types is listed under hf_text.

text/src/autogluon/text/automm/utils.py
sxjscience: Just say that we may fix all the style issues in one PR once we adopted black.

text/src/autogluon/text/automm/data/process_text.py
zhiqiangdon: A static method doesn't mean it can't be used in other places. It is mainly to organize the methods based on their semantic meanings. Obviously, this method is mainly tied to this class.

text/src/autogluon/text/automm/data/process_text.py
zhiqiangdon: Although they are both correct logically, the existence of both does affect the code quality, which is important to attract potential contributors and grow community.

szha:
Job PR-1756-15 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1756/15/index.html

text/src/autogluon/text/automm/data/process_text.py
```diff
@@ -248,6 +317,8 @@ def build_one_token_sequence_from_text(
         ----------
         text
             The raw text data of one sample.
+        is_training: 
+            Only augment for training data
```
zhiqiangdon, line 321: Need to explain what is `is_training` and then what it is used for.

text/src/autogluon/text/automm/configs/model/fusion_mlp_image_text_tabular.yaml
```diff
@@ -95,6 +101,12 @@ model:
     insert_sep: False
     text_segment_num: 1
     stochastic_chunk: False
+    text_detection_length: 10
```
zhiqiangdon, line 104: Why do we choose 10? Does it mean 10 words?

text/src/autogluon/text/automm/configs/model/fusion_mlp_image_text_tabular.yaml
zhiqiangdon: `train_augment_types` may also be used in multimodal models, e.g., clip. So, we need to distinguish this key from the image transformation keys.

text/src/autogluon/text/automm/data/process_text.py
zhiqiangdon: After offline discussion with @sxjscience, I think it's fine to keep this function here. But we need to change the name to make it specific for text.

text/src/autogluon/text/automm/configs/model/fusion_mlp_image_text_tabular.yaml
zhiqiangdon: For tabular data, can we assume each categorical and numerical value, e.g, 3, is only one word?

zhiqiangdon: I think we need a better key name to make users easier to get what is used for. `text_detection_length` tells nothing about the minimum word number to perform augmentation.

text/src/autogluon/text/automm/configs/model/fusion_mlp_image_text_tabular.yaml
zhiqiangdon: The image augmentation key names also become too general after adding the text augmentation key. I used those keys because only image augmentations were available and there seems no well-known augmentations in other modalities. We can add `image` into the image augmentation keys later. But here we need to add `text` into the text augmentation key.

sxjscience: I think only text augmentations are available in hf_text? We can definitely refactor the config later.

text/src/autogluon/text/automm/data/process_text.py
sxjscience: Let’s go to this type of details after we adopted pylint for lint check, or have adopted black. It’s okay for now.

sxjscience: Basically, paying too much attention to coding style without a specific lint tool/style guidance is not efficient. You can read pep8 to see a reference style guidance: https://peps.python.org/pep-0008/ . From my point of view, we should focus on enabling lint check as soon as possible instead of ensuring the coding style via code review, which is super inefficient.

text/src/autogluon/text/automm/configs/model/fusion_mlp_image_text_tabular.yaml
zhiqiangdon: Both lines 66 and 103 have `train_augment_types`. I guess it's not only under `hf_text`. Anyway, it is easy to make the key name more accurate in this PR. No need to leave the debt to future.

text/src/autogluon/text/automm/data/process_text.py
zhiqiangdon: After we enable pylink, can pylink automatically correct the code already merged? If not, we'd better fix `"` before merging.

text/src/autogluon/text/automm/configs/model/fusion_mlp_image_text_tabular.yaml
sxjscience: Oh, I just got the point. @lzcemma Should we add augmentation to “data”?

text/src/autogluon/text/automm/data/process_text.py
bryanyzhu: I think pylint is able to check both existing and new code, depending on how we use it. I agree with @sxjscience  that we should prioritize lint check before other features. This is very important for such a big project with so many modules and new PRs almost every day.

text/tests/unittests/automm/test_automm_predictor.py
```diff
@@ -601,3 +604,49 @@ def test_mixup():
...
+    with tempfile.TemporaryDirectory() as save_path:
+        predictor.fit(
+            train_data=dataset.train_df,
+            config=config,
+            time_limit=10,
+            save_path=save_path,
+            hyperparameters=hyperparameters,
+        )
+
+    #deep copy data preprocessor
```
bryanyzhu, line 641: Missing space after `#`

bryanyzhu (approved):

text/src/autogluon/text/automm/data/process_text.py
```diff
@@ -260,13 +331,21 @@ def build_one_token_sequence_from_text(
             "Token indices sequence length is longer than.*result in indexing errors"
         )
         for col_name, col_text in text.items():
+            #if training
```
bryanyzhu, line 334: May I ask why we have parentheses () for the if statements? I didn't see such usage in other places at AG. Besides, sometimes there are extra spaces (like `if( self.train_augmenter is not None  )`), sometimes there are missing spaces (like ` if(is_training):` and ` #if augment`). Consistency would make the code nicer to read, thanks.

text/src/autogluon/text/automm/data/process_text.py
bryanyzhu: +1

text/src/autogluon/text/automm/configs/model/fusion_mlp_image_text_tabular.yaml
zhiqiangdon: Moving to `data.text` can avoid the issue, but we would lose the flexibility to customize augmentations for each text backbone if we use multiple text backbones.

sxjscience: After a second thought, I agree with @zhiqiangdon on adopting text_augment_types so that it will be consistent.

text/src/autogluon/text/automm/data/process_text.py
sxjscience: @lzcemma let’s remove that. Also, remove the comments

text/src/autogluon/text/automm/configs/model/fusion_mlp_image_text_tabular.yaml
zhiqiangdon: In the AutoMM design, each data processor is tied to one model/backbone. A data processor aims to provide customized input for its model. For example, different text models may need different tokenizers. The model-agnostic data configs are put in the `config.data`, which is used in creating `df_preprocessor`. Based on this design philosophy, everything put in data processors should be customizable for different models. Since augmentation is used in data processor rather than df_preprocessor, it's better we leave it customizable. Hence putting augmentation config under `hf_text` fits the design philosophy.

sxjscience: You will have to write it down somewhere. No one will be able to understand without reading an article that explains the design.

text/src/autogluon/text/automm/configs/model/fusion_mlp_image_text_tabular.yaml
lzcemma: My understanding is that  when tabular feature are used text, it will also have column name eg: age 3. Column name maybe long. But I put this in config such that it can be specified given datasets.

szha:
Job PR-1756-16 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1756/16/index.html

PR #1758, opened 2022-05-26
[Experimental] Enable multimodal matching in AutoMM

zhiqiangdon:
Matching is an important feature for both unimodal and multimodal learning. This PR enables matching through:

1. Design the matching config.
2. Add training loop `lit_matcher.py` and its util functions.
3. Support using `pytorch_metric_learning`, which covers a wide range of metric learning losses.
4. Support returning probability in `._predict()`.

## Design details
1. Matching is one type of binary classification problem. Users provide a dataframe including several feature columns and want to train a model to predict whether some columns' features match other columns'. Users need to specify the `matcher.match_label` in the config since different tasks probably have various semantic labels. Users can easily customize the columns to match through the `matcher.matches` key. We support adding arbitrary matches to the list. For each match, users can customize the metric learning loss, distance, and miner to compute the match's loss. We support using both supervised and unsupervised training. If `use_label=True`, we use the labels from the label column to compute the matching loss. Otherwise, we assume each sample has one unique label.
2. We support extracting column features at the end of models' (`hf_text`, `timm_image`, and `clip`) forwards. In the training loop, i.e., `MatcherLitModule`, we enumerate all the provided matches and add up their losses. For each match, we gather the column features from the models' outputs. If one side of a match has multiple text columns, then we use their cls features from the text models as their joint features. Each image column has independent features. If one side of the match has multiple features (image+text or multiple image columns) available, then we average the features. Before computing the matching loss, each side of a match becomes one feature vector of the same dimension.
3. We use the package [pytorch metric learning ](https://github.com/KevinMusgrave/pytorch-metric-learning) to compute the match losses. Users can customize the types of loss, distance, and miner to compute each match's loss. For more details, refer to https://kevinmusgrave.github.io/pytorch-metric-learning/.
4. Since the concept of `logits` doesn't exist in the matching pipeline,  the `.predict_step()` in `MatcherLitModule` can only return the probabilities. Accordingly, we also support receiving not only `logits` but also `probability` in `AutoMMPredictor`'s `._predict()`, `.predict()`, `.evaluate()`, and `.predict_proba()`.


szha:
Job PR-1758-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1758/2/index.html

text/src/autogluon/text/automm/predictor.py
```diff
@@ -432,6 +433,15 @@ def fit(
         # save artifacts for the current running, except for model checkpoint, which will be saved in _fit()
         self.save(save_path)
 
+        if hasattr(config, MATCHER):
```
sxjscience, line 436: Raise a warning to say that it is experimental.

text/src/autogluon/text/automm/configs/matcher/default.yaml
```diff
@@ -0,0 +1,34 @@
...
+          type: "cosine_similarity"
+        miner:
+          type: "pair_margin_miner"
+          pos_margin: 0.6
+          neg_margin: 0.4
+        use_label: True
+        pair:
+          - "Previous Image"
+          - "Current Image"
+#      - loss:
```
bryanyzhu, line 19: If this is commented, do we want to remove it? If not, maybe add a TODO saying why we want to leave them here.

bryanyzhu (approved):

text/src/autogluon/text/automm/configs/matcher/default.yaml
zhiqiangdon: TODO added.

text/src/autogluon/text/automm/predictor.py
zhiqiangdon: Done.

sxjscience (approved):
Since it’s an experimental feature, this should be good.

szha:
Job PR-1758-8 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1758/8/index.html

szha:
Job PR-1758-10 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1758/10/index.html

rhkenne:
Thank you for this great feature. 

Do you have any example or code snippet that showcase how to use this feature ? I was looking for an approach to perform multimodal document pairs classification or regression

An example of a multimodal document will be product listing containing title, description & image.

sxjscience:
@rhkenne We are planning to make it non-experimental in 0.6 and will write some codes to refactor that part. Is it possible to share the rough use-case with us. Also ping @zhiqiangdon, @FANGAreNotGnu, @cheungdaven

PR #1764, opened 2022-05-30
Add ForecastingLearner and ForecastingPredictor

canerturkmen:
*Issue #, if available:*

*Description of changes:*

- Add `ForecastingLearner` and `ForecastingPredictor` and tests.
- Fixes to save/load logic of trainers
- Minor fixes to Forecasting
- Move `is_fit` property from tabular abstract learner to core.


canerturkmen:
@huibinshen fyi

forecasting/src/autogluon/forecasting/trainer/abstract_trainer.py
```diff
@@ -132,8 +132,11 @@ def save(self) -> None:
         try:
             save_pkl.save(path=self.path_pkl, object=self)
         except:  # noqa
+            for model in self.models.values():
+                model.save()
```
Innixma, line 136: This feels spooky. I feel like we should just crash if saving the pkl is raising an exception.

What do you think about removing the `self.low_memory=False` logic entirely? I have stopped maintaining `self.low_memory=False` in Tabular for awhile since most users wouldn't use it and it is very clunky to deal with as many models may internally want to cache to disk during fit. This could simplify our code and minimize defects in future.

forecasting/src/autogluon/forecasting/learner.py
```diff
@@ -0,0 +1,139 @@
...
+            hyperparameter_tune=hyperparameter_tune,
+            hyperparameters=hyperparameters,
+            time_limit=time_limit,
+        )
+        self.save_trainer(trainer=self.trainer)
+
+        self._time_fit_training = time.time() - time_start
+        logger.info(
+            f"AutoGluon ForecastingLearner complete, total runtime = {round(self._time_fit_training, 2)}s",
+        )
```
Innixma, line 102: logger.log(20, ...) is what is used for Tabular for these kinds of logs, otherwise .info won't show up by default I think.

forecasting/src/autogluon/forecasting/learner.py
```diff
@@ -0,0 +1,139 @@
...
+
+        learner_info.update(trainer_info)
+        return learner_info
+
+    def refit_full(self, models="all"):
+        # TODO: Implement refitting
+        # return self.load_trainer().refit_full(models=models)
+        raise NotImplementedError(
+            "refitting logic currently not implemented in autogluon.forecasting"
+        )
```
Innixma, line 139: Why do the presets specify `refit_full=True` if it isn't implemented?

forecasting/src/autogluon/forecasting/predictor.py
```diff
@@ -0,0 +1,420 @@
...
+    @property
+    def _trainer(self) -> AbstractForecastingTrainer:
+        return self._learner.load_trainer()  # noqa
+
+    @apply_presets(FORECASTING_PRESETS_CONFIGS)
+    def fit(
+        self,
+        train_data: TimeSeriesDataFrame,
+        prediction_length: int = 1,
+        target_column: str = "target",
```
Innixma, line 105: Wondering why this is specified in fit and not in init? Both are probably valid, but just wondering if there is a specific reason why it should be in fit.

Also: Any reason why this isn't called 'label' to align with the other modules?

forecasting/src/autogluon/forecasting/predictor.py
```diff
@@ -0,0 +1,420 @@
...
+        val_data: Optional[TimeSeriesDataFrame] = None,
```
Innixma, line 106: val_data is called tuning_data in the other modules. Thoughts?

Also: Shall we re-order this to be directly after train_data to align with the other modules? (followed by time_limit, presets, hyperparameters)

forecasting/src/autogluon/forecasting/predictor.py
```diff
@@ -0,0 +1,420 @@
...
+            `presets="best_quality"` to get a high-accuracy predictor, or set `presets="low_quality"` to get a
+            toy predictor that trains quickly but lacks accuracy. Any user-specified arguments in `fit()` will
+            override the values used by presets.
+
+            Available presets are "best_quality", "high_quality", "good_quality", "medium_quality", "low_quality",
+            and "low_quality_hpo". Details for these presets can be found in
+            `autogluon/forecasting/configs/presets_configs.py`
+        hyperparameter_tune_kwargs: Optional, None by default, can be str or dict
+            # TODO
+        hyperparameters: str or dict, default = None
```
Innixma, line 149: Missing what happens if not specified (default)

forecasting/src/autogluon/forecasting/predictor.py
```diff
@@ -0,0 +1,420 @@
...
+            Name of column that contains the target values to forecast (ie. the numeric observations of the
+            time series). This column must contain numeric values, and missing target values
+            should be in a pandas compatible format:
+            https://pandas.pydata.org/pandas-docs/stable/user_guide/missing_data.html
+            By default, `target_column="target"`.
+        val_data: TimeSeriesDataFrame
+            Validation data reserved for model selection or hyperparameter tuning, rather than training individual
+            models. If None, AutoGluon will reserve the most recent portion of `train_data` for validation. Validation
+            scores will by default be computed over the last `prediction_length` time points in the validation data.
+        presets: str, default = None
```
Innixma, line 135: Missing what happens if not specified (default)

forecasting/src/autogluon/forecasting/predictor.py
```diff
@@ -0,0 +1,420 @@
...
+            The forecast horizon, i.e., How many time points into the future forecasters should be trained to predict.
+            For example, if our time series contain daily observations, setting `prediction_length=3` will train
+            models that predict up to 3 days in the future from the most recent observation.
+        target_column: str
+            Name of column that contains the target values to forecast (ie. the numeric observations of the
+            time series). This column must contain numeric values, and missing target values
+            should be in a pandas compatible format:
+            https://pandas.pydata.org/pandas-docs/stable/user_guide/missing_data.html
+            By default, `target_column="target"`.
+        val_data: TimeSeriesDataFrame
```
Innixma, line 131: `val_data: TimeSeriesDataFrame, default = None`

forecasting/src/autogluon/forecasting/predictor.py
```diff
@@ -0,0 +1,420 @@
...
+        Parameters
+        ----------
+        train_data: TimeSeriesDataFrame
+            Training data in the TimeSeriesDataFrame format. See documentation for `autogluon.forecasting.dataset` for
+            further information.
+        prediction_length: int
+            The forecast horizon, i.e., How many time points into the future forecasters should be trained to predict.
+            For example, if our time series contain daily observations, setting `prediction_length=3` will train
+            models that predict up to 3 days in the future from the most recent observation.
+        target_column: str
```
Innixma, line 125: `target_column: str, default = "target"`

forecasting/src/autogluon/forecasting/predictor.py
```diff
@@ -0,0 +1,420 @@
...
+    ) -> "ForecastingPredictor":
+        """Fit models to predict (distributional) forecasts of multiple related time series
+        based on historical observations.
+
+        Parameters
+        ----------
+        train_data: TimeSeriesDataFrame
+            Training data in the TimeSeriesDataFrame format. See documentation for `autogluon.forecasting.dataset` for
+            further information.
+        prediction_length: int
```
Innixma, line 121: `prediction_length: int, default = 1`

forecasting/src/autogluon/forecasting/predictor.py
```diff
@@ -0,0 +1,420 @@
...
+            the predictor will only train the given model types. Stable model options include: 'DeepAR', 'MQCNN', and
+            'SFF' (SimpleFeedForward). See References for more detail on these models.
+
+            Values in the `hyperparameters` dict are themselves dictionaries of hyperparameter settings for each model
+            type. Each hyperparameter can either be a single fixed value or a search space containing many possible
+            values. A search space should only be provided when `hyperparameter_tune_kwargs` is specified (i.e.,
+            hyperparameter-tuning is utilized). Any omitted hyperparameters not specified here will be set to default
+            values which are given in`autogluon/forecasting/trainer/models/presets.py`. Specific hyperparameter
+            choices for each of the recommended models can be found in the references.
+        time_limit: int, default=None
```
Innixma, line 163: `time_limit: int, default = None`

forecasting/src/autogluon/forecasting/predictor.py
```diff
@@ -0,0 +1,420 @@
...
+            )
+            logger.info("============ fit arguments ============")
+            logger.info("fit() called with arguments:")
+            logger.info(f"{pprint.pformat(fit_args)}")
+            logger.info("=======================================")
+
+        # Inform the user extra columns in dataset will not be used.
+        extra_columns = [c for c in train_data.columns.copy() if c != target_column]
+        if len(extra_columns) > 0:
+            logger.warning(f"Provided columns {extra_columns} will not be used.")
```
Innixma, line 210: I don't see a check on if the target column exists at all in the train data. Perhaps this should be the first thing checked?

forecasting/src/autogluon/forecasting/predictor.py
```diff
@@ -0,0 +1,420 @@
...
+            when making distributional forecasts. Can alternatively be provided with the keyword
+            argument `quantiles`.
+
+        References
+        ----------
+            - DeepAR: https://ts.gluon.ai/api/gluonts/gluonts.model.deepar.html
+            - MQCNN: https://ts.gluon.ai/api/gluonts/gluonts.model.seq2seq.html
+            - SFF: https://ts.gluon.ai/api/gluonts/gluonts.model.simple_feedforward.html
+        """
+        start_time = time.time()
```
Innixma, line 180: nit: for consistency with other areas of the code: `time_start`

forecasting/src/autogluon/forecasting/predictor.py
```diff
@@ -0,0 +1,420 @@
...
+        num_trials: Optional[int] = None
+        if isinstance(hyperparameter_tune_kwargs, dict):
+            num_trials = hyperparameter_tune_kwargs.get("num_trials")
+            if time_limit is None and num_trials is None:
+                logger.warning(
+                    "None of time_limit and num_trials are set, defaulting to num_trials=2",
+                )
+                num_trials = 2
+            else:
+                num_trials = hyperparameter_tune_kwargs.get("num_trials", 9999)
```
Innixma, line 267: Note: We may want to reign this in eventually. Running out of disk space becomes a real problem when you save 9999 models over the course of a few hours. This applies to the other modules as well.

forecasting/src/autogluon/forecasting/predictor.py
```diff
@@ -0,0 +1,420 @@
...
+        """Returns the list of model names trained in this `predictor` object."""
+        return self._trainer.get_model_names()
+
+    def predict(
+        self,
+        data: TimeSeriesDataFrame,
+        model: Optional[str] = None,
+        **kwargs,
+    ) -> TimeSeriesDataFrame:
+        """Return quantile and mean forecasts given a dataset to condition on.
```
Innixma, line 307: What if the user only wants the mean?

forecasting/src/autogluon/forecasting/predictor.py
```diff
@@ -0,0 +1,420 @@
...
+    def leaderboard(self, data: Optional[TimeSeriesDataFrame] = None) -> pd.DataFrame:
+        """Return a leaderboard showing the performance of every trained model
+
+        Parameters
+        ----------
+        data: TimeSeriesDataFrame
+            dataset used for additional evaluation. If None, the validation set used during training will
+            be used.
+        """
+        return self._learner.leaderboard(data)
```
Innixma, line 367: Tabular describes each column present in the leaderboard. We may want to do the same here.

forecasting/src/autogluon/forecasting/predictor.py
```diff
@@ -0,0 +1,420 @@
...
+from autogluon.core.utils.decorators import apply_presets
+from autogluon.core.utils.loaders import load_pkl
+from autogluon.core.utils.savers import save_pkl
+
+from .configs import FORECASTING_PRESETS_CONFIGS
+from .dataset import TimeSeriesDataFrame
+from .learner import AbstractLearner, ForecastingLearner
+from .trainer import AbstractForecastingTrainer
+
+logger = logging.getLogger()
```
Innixma, line 20: `logger = logging.getLogger(__name__)  # return autogluon root logger`

This avoids altering the root logger. Note that this should be done for every logger in every file.

Innixma (commented):
Looks quite good, thanks for this! Added some comments.

forecasting/src/autogluon/forecasting/trainer/abstract_trainer.py
canerturkmen: I completely agree re `low_memory`. I'll revise.

forecasting/src/autogluon/forecasting/learner.py
canerturkmen: I believe `logger.info` is pretty much an alias for `logger.log(20, ...`. See [here](https://github.com/python/cpython/blob/f545fc955aeb701ae4e73b07ff2283f823d857b8/Lib/logging/__init__.py#L1504) for example. Do you have any example where they are not equivalent?

If they are equivalent I would prefer `logger.info` as it's more explicit.

forecasting/src/autogluon/forecasting/predictor.py
canerturkmen: Changed and reordered for consistency.

forecasting/src/autogluon/forecasting/predictor.py
canerturkmen: I'll move this and `prediction_length` to init args for both Predictor and Learner. I'll revise the name to `target` to echo Tabular's style however I'd like to refrain from using `label` as in forecasting we use the "past target" to forecast the "future target" so it's not a 1:1 analogy with label.

forecasting/src/autogluon/forecasting/predictor.py
canerturkmen: Hmm, right now they "have" to want a quantile. Right now probabilistic forecasting is the "default" mode of operation for the library. The default validation score is mean quantile loss, and default `quantile_levels` is [0.1,...0.9]. In some sense we made probabilistic forecasting the "first class citizen" and point forecasting a side-effect of the library. I think this makes sense. Most libraries either choose one or the other. We prioritize quantile forecasting but still provide mean forecasts. If we want to design this back, I'm happy to discuss.

szha:
Job PR-1764-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1764/3/index.html

forecasting/src/autogluon/forecasting/learner.py
Innixma: oh you are correct, for some reason I though info was 15, but it is 20. Feel free to keep it as is then

forecasting/src/autogluon/forecasting/predictor.py
Innixma: thanks for explaination! `target` is ok, although we may consider adding an alias `label` which maps to `target` and warns the user to use target instead in case users are coming in from other modules

forecasting/src/autogluon/forecasting/predictor.py
Innixma: Ok, we can keep for now

Innixma (approved):
Overall looks good to me! Feel free to merge once you finish resolving any remaining comments

forecasting/src/autogluon/forecasting/trainer/abstract_trainer.py
canerturkmen: Just FYI I left the logic as is in the `SimpleAbstractTrainer` class which we intend to move later to core. We can revise the logic there when we'd like. Left a todo.

szha:
Job PR-1764-4 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1764/4/index.html

PR #1777, opened 2022-06-03
Support using BCELoss for training regression models in AutoMM + Support MinMaxScaler

Linuxdex:
*Issue #, if available:*
Adding BCEloss into automm regression. It is a special method for dealing with regression problems.
*Description of changes:*
Add BCEloss. It will be active if the problem_type = binary and the eval_metrics belong to regression.
This requires the label to be converted to [0,1].
So the default standardization should be modified. Now users can choose "minmaxscaler", "standardscaler" or nothing to do in the data default.

### DIFF
--- a/text/src/autogluon/text/automm/configs/data/default.yaml
+++ b/text/src/autogluon/text/automm/configs/data/default.yaml
@@ -12,14 +12,15 @@ data:
     scaler_with_std: True  # Whether to normalize with std.
   label:
     numerical_label_preprocessing:  # The mode of label preprocessing for . Support "standardscaler" or "minmaxscaler" or null.
+    predict_heads:  # The projecting head for prediction. It is used for loss functions like BCEWithLogitsLoss which contain projection.
   pos_label:  # The name of binary classification's positive class. It's used in computing some metrics, e.g., roc_auc. If not provided, then use label_encoder.classes_[1],
   mixup:
-    turn_on: False # The total control of mixup.
+    turn_on: False  # The total control of mixup.
     mixup_alpha: 0.8  # Mixup alpha.
     cutmix_alpha: 1.0  # Cutmix alpha.
     cutmix_minmax:  #Cutmix min/max ratio, it will override cutmix alpha if set, a list/tuple with size two.
     mixup_prob: 1.0  # The probability of conducting mixup/cutmix if enable.
     mixup_switch_prob: 0.5  # The probability of switching mixup to cutmix if both enable.
     mixup_mode: "batch"  # Perform mixup/cutmix on "batch" or "pair" or "elem".
-    mixup_off_epoch: 5 # The epoch when the mixup will be turned off.
+    mixup_off_epoch: 5  # The epoch when the mixup will be turned off.
     label_smoothing: 0.1  # Label smoothing.
--- a/text/src/autogluon/text/automm/configs/data/default.yaml
+++ b/text/src/autogluon/text/automm/configs/data/default.yaml
@@ -10,6 +10,8 @@ data:
     convert_to_text: False  # Whether to convert the feature to text.
     scaler_with_mean: True  # Whether to normalize with mean.
     scaler_with_std: True  # Whether to normalize with std.
+  label:
+    numerical_label_preprocessing:  #The mode of label preprocessing for . Support "standardscaler" or "minmaxscaler" or null.
   pos_label:  # The name of binary classification's positive class. It's used in computing some metrics, e.g., roc_auc. If not provided, then use label_encoder.classes_[1],
   mixup:
     turn_on: False # The total control of mixup.
@@ -21,4 +23,3 @@ data:
     mixup_mode: "batch"  # Perform mixup/cutmix on "batch" or "pair" or "elem".
     mixup_off_epoch: 5 # The epoch when the mixup will be turned off.
     label_smoothing: 0.1  # Label smoothing.
-
--- a/text/src/autogluon/text/automm/configs/data/default.yaml
+++ b/text/src/autogluon/text/automm/configs/data/default.yaml
@@ -11,7 +11,7 @@ data:
     scaler_with_mean: True  # Whether to normalize with mean.
     scaler_with_std: True  # Whether to normalize with std.
   label:
-    numerical_label_preprocessing:  #The mode of label preprocessing for . Support "standardscaler" or "minmaxscaler" or null.
+    numerical_label_preprocessing:  # The mode of label preprocessing for . Support "standardscaler" or "minmaxscaler" or null.
   pos_label:  # The name of binary classification's positive class. It's used in computing some metrics, e.g., roc_auc. If not provided, then use label_encoder.classes_[1],
   mixup:
     turn_on: False # The total control of mixup.
--- a/text/src/autogluon/text/automm/configs/data/default.yaml
+++ b/text/src/autogluon/text/automm/configs/data/default.yaml
@@ -12,7 +12,6 @@ data:
     scaler_with_std: True  # Whether to normalize with std.
   label:
     numerical_label_preprocessing:  # The mode of label preprocessing for . Support "standardscaler" or "minmaxscaler" or null.
-    predict_heads:  # The projecting head for prediction. It is used for loss functions like BCEWithLogitsLoss which contain projection.
   pos_label:  # The name of binary classification's positive class. It's used in computing some metrics, e.g., roc_auc. If not provided, then use label_encoder.classes_[1],
   mixup:
     turn_on: False  # The total control of mixup.
--- a/text/src/autogluon/text/automm/configs/data/default.yaml
+++ b/text/src/autogluon/text/automm/configs/data/default.yaml
@@ -17,7 +17,7 @@ data:
     turn_on: False  # The total control of mixup.
     mixup_alpha: 0.8  # Mixup alpha.
     cutmix_alpha: 1.0  # Cutmix alpha.
-    cutmix_minmax:  #Cutmix min/max ratio, it will override cutmix alpha if set, a list/tuple with size two.
+    cutmix_minmax:  # Cutmix min/max ratio, it will override cutmix alpha if set, a list/tuple with size two.
     mixup_prob: 1.0  # The probability of conducting mixup/cutmix if enable.
     mixup_switch_prob: 0.5  # The probability of switching mixup to cutmix if both enable.
     mixup_mode: "batch"  # Perform mixup/cutmix on "batch" or "pair" or "elem".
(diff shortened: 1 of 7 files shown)


text/src/autogluon/text/automm/configs/data/default.yaml
```diff
@@ -10,6 +10,8 @@ data:
     convert_to_text: False  # Whether to convert the feature to text.
     scaler_with_mean: True  # Whether to normalize with mean.
     scaler_with_std: True  # Whether to normalize with std.
+  label:
+    numerical_label_preprocessing:  #The mode of label preprocessing for . Support "standardscaler" or "minmaxscaler" or null.
```
sxjscience, line 14: Change `#The mode` to `# The mode`.

text/src/autogluon/text/automm/optimization/utils.py
```diff
@@ -23,6 +23,7 @@
 def get_loss_func(
         problem_type: str,
         mixup_active: bool,
+        bce_type: bool,
```
sxjscience, line 26: How about to change `bce_type` to `loss_function` and use None by default. The loss function can then be `bce` and will be used if it's regression problem.

Also, we need to revise the docstring below (add mixup_active and loss_function to parameters).

text/src/autogluon/text/automm/predictor.py
```diff
@@ -414,16 +418,17 @@ def fit(
             pos_label=pos_label,
         )
 
-        mixup_active, mixup_fn = get_mixup(model_config=OmegaConf.select(config,'model'),
-                                     mixup_config=OmegaConf.select(config,'data.mixup'),
-                                     num_classes=output_shape,
+        mixup_active, mixup_fn = get_mixup(
+            model_config=OmegaConf.select(config,'model'),
```
sxjscience, line 422: Add space `(config,'model')` should be `(config, 'model')`.

text/src/autogluon/text/automm/predictor.py
```diff
@@ -147,6 +147,8 @@ def __init__(
             eval_metric = eval_metric.name
 
         if eval_metric is not None and eval_metric.lower() in ["rmse", "r2", "pearsonr", "spearmanr"]:
+            if problem_type == BINARY:
+                bce_regression_state = True
```
sxjscience, line 151: Cannot understand the logic.

text/src/autogluon/text/automm/predictor.py
```diff
@@ -1032,6 +1037,13 @@ def _logits_to_prob(logits: torch.Tensor):
         prob = prob.detach().cpu().float().numpy()
         return prob
 
+    @staticmethod
```
sxjscience, line 1040: Instead of using a static method, let's make it a function and put it in utils or other places.

sxjscience:
I think we may later turn on `MinMaxScale` if it's regression problem and the user is using bce loss.

text/src/autogluon/text/automm/predictor.py
Linuxdex: In my original design, this mode is determine to introduce binary into regression. So users treat it as a binary problem (but regression indeed) and the eval metrics may be regression.

szha:
Job PR-1777-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1777/3/index.html

szha:
Job PR-1777-5 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1777/5/index.html

szha:
Job PR-1777-6 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1777/6/index.html

text/src/autogluon/text/automm/configs/optimization/adamw.yaml
```diff
@@ -16,3 +16,6 @@ optimization:
   top_k_average_method: "greedy_soup"  # We support averaging method described in https://arxiv.org/pdf/2203.05482.pdf.
                                        # Currently support "uniform_soup", "greedy_soup", and "best".
   efficient_finetune: null  # Can be 'bit_fit' (only finetune bias), 'norm_fit' (finetune the normalization terms + bias terms), or null
+  loss_function: "BCEWithLogitsLoss" # The replaced loss for regresssion. Can only support loss function in torch.nn.
```
sxjscience, line 19: Use an empty value by default and only specify the value in test-cases.

text/src/autogluon/text/automm/optimization/utils.py
```diff
@@ -477,3 +483,34 @@ def apply_layerwise_lr_decay(
         parameter_group_names[group_name]["params"].append(name)
 
     return list(parameter_group_vars.values())
+
+
+def modify_config_with_loss_func(
```
sxjscience, line 488: Consider to use a name like 
```
def config_update_loss_function(...):
```

text/src/autogluon/text/automm/data/preprocess_dataframe.py
```diff
@@ -435,6 +448,9 @@ def transform_prediction(
         -------
         Predicted labels ready to compute metric scores.
         """
+        if OmegaConf.select(self.config, "label.sigmoid"):
```
sxjscience, line 451: There is no "label.sigmoid" in the config.

text/src/autogluon/text/automm/configs/optimization/adamw.yaml
sxjscience: Actually, we can also pick "auto" by default, which is an indicator that it should be updated by the system.

sxjscience:
We recently enabled black to check the coding style so you will have to rebase the code with the latest master.

szha:
Job PR-1777-22 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1777/22/index.html

sxjscience (approved):
LGTM

text/tests/unittests/automm/test_metrics.py
```diff
@@ -40,3 +40,35 @@ def test_cross_entropy(metric_name, class_num):
...
+        problem_type=problem_type,
+        mixup_active=False,
+        loss_func_replaced=loss_func_replaced,
+    )
+
+    score1 = loss_func(input=preds, target=targets)
+    preds = preds.sigmoid()
+    bceloss = torch.nn.BCELoss()
+    score2 = bceloss(input=preds, target=targets)
+    assert pytest.approx(score1, 1e-6) == score2
```
bryanyzhu, line 74: Maybe need a newline for the end of the file.

bryanyzhu (approved):

text/src/autogluon/text/automm/optimization/utils.py
```diff
@@ -57,7 +59,13 @@ def get_loss_func(
...
             loss_func = nn.CrossEntropyLoss()
     elif problem_type == REGRESSION:
-        loss_func = nn.MSELoss()
+        if loss_func_replaced is not None:
+            if "bcewithlogitsloss" in loss_func_replaced.lower():
+                loss_func = nn.BCEWithLogitsLoss()
+            else:
+                loss_func = nn.MSELoss()
+        else:
+            loss_func = nn.MSELoss()
```
zhiqiangdon, line 68: This implementation restricted using bce to only regression tasks. We can improve this code. Keep the original code and Insert the:
```
if loss_func_replaced is not None and "bcewithlogitsloss" in loss_func_replaced.lower():
    loss_func = nn.BCEWithLogitsLoss()
```
at the end of this function. In this way, users can also use bce loss for classification tasks.

sxjscience: Yeah, how about to revise it in a later PR. I think BCE for classification has not been tested. We can fix the logic later.

text/src/autogluon/text/automm/optimization/utils.py
```diff
@@ -38,6 +39,7 @@
 def get_loss_func(
     problem_type: str,
     mixup_active: bool,
+    loss_func_replaced: Optional[str] = None,
```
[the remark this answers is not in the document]
zhiqiangdon: How about changing `loss_func_replaced`  to `loss_func_name`?

text/src/autogluon/text/automm/optimization/utils.py
```diff
@@ -480,3 +488,30 @@ def apply_layerwise_lr_decay(
         parameter_group_names[group_name]["params"].append(name)
 
     return list(parameter_group_vars.values())
+
+
+def config_update_loss_func(
```
zhiqiangdon, line 497: Could we change `config_update_loss_func` to `update_config_by_rules`? I guess we will have other similar needs to update config based on our pre-defined rules. So, we can make this function name more general and add other similar stuff into this function in future.

text/src/autogluon/text/automm/predictor.py
```diff
@@ -452,7 +458,8 @@ def fit(
         self._df_preprocessor = df_preprocessor
         self._data_processors = data_processors
         self._model = model
-        self.mixup_fn = mixup_fn
+        self._mixup_fn = mixup_fn
+        self._loss_func = loss_func
```
zhiqiangdon, line 462: There is no need to have these two attributes since they are not saved.

zhiqiangdon: Yeah, I just see `self._loss_func` is used below, but `self._mixup_fn` seem not necessary.

text/src/autogluon/text/automm/optimization/utils.py
```diff
@@ -480,3 +492,30 @@ def apply_layerwise_lr_decay(
...
+    config
+        The config of the project. It is a Dictconfig object.
+
+    Returns
+    -------
+    The modified config.
+    """
+    if problem_type == REGRESSION and OmegaConf.select(config, "optimization.loss_function") is not None:
+        loss_func = config.optimization.loss_function
+        if "bceloss" in loss_func.lower() or "bcewithlogitsloss" in loss_func.lower():
```
zhiqiangdon, line 518: Is `bceloss` a valid choice? I see only `bcewithlogitsloss` are allowed in `get_loss_func`.

text/src/autogluon/text/automm/optimization/utils.py
zhiqiangdon: The current implementations in `transform_prediction` and `_compute_metric_score` seem not to support BCE for classification. Considering this limitation, we should raise error if detecting BCE in classification tasks.

zhiqiangdon (approved):
Thanks for adding BCE support for regression!
Since the current BCE is crafted for regression, we need to raise warnings/errors if users specify BCE for classification tasks.

szha:
Job PR-1777-36 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1777/36/index.html

PR #1666, opened 2022-04-14
Add "efficient_finetuning" flag and supports bit_fit and norm_fit

sxjscience:
*Description of changes:*
- add "optimization.efficient_finetune" flag. We are considering to support a bunch of efficient finetuning algorithms. For now, we support
   - "bit_fit": "BitFit: Simple Parameter-efficient Fine-tuning for Transformer-based Masked Language-models" (https://arxiv.org/abs/2106.10199)
   - "norm_fit": This is an extension of the algorithm in "Training BatchNorm and Only BatchNorm: On the Expressive Power of Random Features in CNNs" (https://arxiv.org/abs/2003.00152) and BitFit. We finetune both the parameters in the norm layers as long as the biases. Actually, it may make more sense for modern NN architectures because we are removing the bias terms.
- Fix the bug that the categorical label won't be transformed to the original space.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1780, opened 2022-06-04
[WIP] Add LoRA as efficient finetuning strategy

Raldir:
Added the option to apply LoRA (https://arxiv.org/pdf/2106.09685.pdf) to Huggingface text model as an additional option for efficient fine-tuning, see PR #1666.  

- Added flag for model: `model.hf_text.adaptation=lora`. This currently only supports loRA but can be extended to adapters or similar adaptations to transformers that modify the architecture directly. LoRA is currently set to be inserted only into Query and Value attention weights of a transformer, as done in their paper and shown to be most effective.
-  While loRA is currently only added to hf_text, should be extendable to most other transformer architecture, as soon as the general structure stands. As the adaptation is currently coded into the model class directly, might consider adding a separate adaptation class to be then added to different models.
- For optimizer `optimization.efficient_finetuned` added flag `lora` and `lora_bias`. The first fine-tunes only loRA weights while the second updates both loRA weights and all biases of the transformer. The latter has not been comprehensively explored in the original paper but has been suggested as an additional modification.
- LoRA uses two hyperparameters. Currently, these are set also through the model config. However, this is rather a hack. Since adapters etc. also use hyperparameters, maybe consider adding a separate config file for adaptations?   
- Currently, when saving the model the entire model is stored, which is unnecessary. TODO: Need to change to solely save fine-tuned weights, also applies to bit_fit and norm_fit.


Raldir:
Just to say that experiments on how LoRA compares to bit_fit and norm_fit are ongoing.

text/src/autogluon/text/automm/models/lora_layers.py
```diff
@@ -0,0 +1,373 @@
+
```
sxjscience, line 1: Mention that this file is modified based on https://github.com/microsoft/LoRA/blob/main/loralib/layers.py

text/src/autogluon/text/automm/models/lora_layers.py
```diff
@@ -0,0 +1,373 @@
...
+import math
+from typing import List
+
+
+def identity(x):
+    return x
+
+
+class LoRALayer():
```
sxjscience, line 19: Consider to just use the following (remove parenthesis).

```
class LoRALayer:

```

text/src/autogluon/text/automm/models/huggingface_text.py
```diff
@@ -14,12 +14,24 @@
...
 logger = logging.getLogger(AUTOMM)
 
 
+def replace_layers_with_lora(model, old, lora_r, lora_alpha):
+    for n, module in model.named_children():
+        if len(list(module.children())) > 0:
+            replace_layers_with_lora(module, old, lora_r, lora_alpha)
+
+        if old in n:
+            lora_layer = LoRALinear(module.in_features, module.out_features, r= lora_r, lora_alpha=lora_alpha, merge_weights=False)
```
sxjscience, line 29: We need to detect the type of module. Should we do the following instead? 

```
if isinstance(module, nn.Linear):
   lora_layer = LoRALinear(...)
```

sxjscience:
We recently enabled black to check the coding style so you will have to rebase the code with the latest master.

text/src/autogluon/text/automm/models/huggingface_text.py
```diff
@@ -14,12 +14,24 @@
...
     get_column_features,
 )
-
+from .lora_layers import LoRALinear
 hf_logging.set_verbosity_error()
 
 logger = logging.getLogger(AUTOMM)
 
 
+def replace_layers_with_lora(model, old, lora_r, lora_alpha):
```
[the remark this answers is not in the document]
zhiqiangdon: Also, we need to add docstrings and type hints for arguments. Please refer to other functions for examples.

text/src/autogluon/text/automm/models/huggingface_text.py
Raldir: I agree that relying on the name is brittle. That said, LoRA is not applied to all linear layers, only query and value attention weights are replaced. I was thinking we might be able to maybe make a check on the type of parent module? But again here, different models define their own attention module (e.g. RobertaSelfAttention, DisentangledSelfAttention (Deberta)) so we might have to explicitly state which modules we support LoRA for?

text/src/autogluon/text/automm/optimization/utils.py
```diff
@@ -11,13 +11,14 @@
...
 from ..constants import (
-    BINARY, MULTICLASS, REGRESSION, MAX, MIN, NORM_FIT, BIT_FIT,
+    BINARY, MULTICLASS, REGRESSION, MAX, MIN, NORM_FIT, BIT_FIT, LORA, LORA_BIAS,
     ACC, ACCURACY, RMSE, ROOT_MEAN_SQUARED_ERROR, R2, QUADRATIC_KAPPA,
     ROC_AUC, AVERAGE_PRECISION, LOG_LOSS, CROSS_ENTROPY,
     PEARSONR, SPEARMANR,
 )
 import warnings
 from .soft_target_crossentropy import SoftTargetCrossEntropy
+from ..models.lora_layers import LoRALayer
```
zhiqiangdon, line 21: Looks like `LoRALayer` is unused in this file.

text/src/autogluon/text/automm/models/huggingface_text.py
```diff
@@ -53,12 +69,23 @@ def __init__(
...
         self.model = AutoModel.from_pretrained(checkpoint_name)
+        self.adaptation = adaptation
+        self.lora_r = lora_r
+        self.lora_alpha = lora_alpha
+
+        if self.adaptation == 'lora':
+            print('Applying LoRA adaptation to attention layers...')
+            replace_layers_with_lora(self.model, "query", self.lora_r, self.lora_alpha)
+            replace_layers_with_lora(self.model, "value", self.lora_r, self.lora_alpha)
+
```
zhiqiangdon, line 88: This layer replacement operations can be moved outside of specific model classes. We can move them to the end of `create_model`, i.e., after the model is created. In this way, we can apply lora to any models, not limited to text models.

Raldir: Good idea!

text/src/autogluon/text/automm/configs/model/fusion_mlp_image_text_tabular.yaml
```diff
@@ -62,6 +62,9 @@ model:
     insert_sep: True
     text_segment_num: 2
     stochastic_chunk: False
+    adaptation: null # Select from options: Lora
+    lora_r: 8 #TODO: Hacky way of changing LoRA hyperparameter. Seperate config file for all adaptation approaches (e.g.. adapters, LoRA...)?
+    lora_alpha: 8
```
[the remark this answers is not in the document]
sxjscience: Moving to `optimization` sounds good to me.

text/src/autogluon/text/automm/optimization/utils.py
```diff
@@ -11,13 +11,14 @@
     get_linear_schedule_with_warmup,
 )
 from ..constants import (
-    BINARY, MULTICLASS, REGRESSION, MAX, MIN, NORM_FIT, BIT_FIT,
+    BINARY, MULTICLASS, REGRESSION, MAX, MIN, NORM_FIT, BIT_FIT, LORA, LORA_BIAS,
```
sxjscience, line 14: Considering to add LORA_NORM (LoRA + NormFit).

.gitignore
```diff
@@ -123,6 +123,7 @@ train/*
 tutorials/checkpoint/*
 tutorials/train/*
 examples/image_classification/*.csv
+examples/few_shot/*
```
sxjscience, line 126: Remove this line

sxjscience:
The design looks good to me overall. Two remaining items:
1. Add a test-case about LoRA in this function https://github.com/awslabs/autogluon/blob/cc8362684ed4412dabe5a2114cd35a1cfce1c8d3/text/tests/unittests/automm/test_automm_predictor.py#L65-L93 . You may change some of the BIT_FIT or NORM_FIT to LORA_BIAS or LORA_NORM.
2. Resolve conflicts.

text/src/autogluon/text/automm/utils.py
```diff
@@ -703,6 +708,31 @@ def create_model(
...
+        A PyTorch model.
+    config:
+        A DictConfig object. The optimization config should be accessible by "config.optimization".
+    """
+    if 'lora' in config.optimization.efficient_finetune:
+        model = inject_lora_to_linear_layer(
+            model = model,
+            lora_r = config.optimization.lora.r,
+            lora_alpha = config.optimization.lora.alpha,
+            filter = ['query', 'value'] # Fine-tune only Query And Value Attention, recommended in https://arxiv.org/abs/2106.09685
```
sxjscience, line 730: potentially make `filter` a flag in config.

text/src/autogluon/text/automm/models/utils.py
```diff
@@ -282,39 +283,49 @@ def assign_layer_ids(
...
     )
     return name_to_id, left_names
 
 
 def get_column_features(
-    batch: Dict[str, torch.Tensor],
-    column_name_prefix: str,
-    features: torch.Tensor,
-    valid_lengths: torch.Tensor,
+        batch: Dict[str, torch.Tensor],
```
sxjscience, line 325: It might be related to the black formatting issue (indentation).

Raldir:
Opened PR #1809 for an easier merge with master.

PR #1696, opened 2022-05-03
[HPO] Ray Tune HPO Integration

yinweisu:
*Issue #, if available:*

*Description of changes:*
* Added HPO functionality for AutoMM using ray tune + ray lightning.
  * Supports `random, bayes` searchers
  * Supports `FIFO, ASHA` schedulers
  * PBT/PB2 will be supported once ray lightning supports PT 1.6. Currently there's a bug in PT 1.6 stopping PBT/PB2 to work correctly
  * Extracted `_fit()` so that model specific configs will be generated in it, and `fit()` will generate general info that are model independent
  * Loading of the model has been divided into two parts to load general info, and config specific info
  * Checkpoint averaging is only done to the best trial once HPO is finished.
* Changed HPO logic for tabular and forecasting to use ray tune
  * Supports `random, bayes` searchers
  * Supports `FIFO` scheduler
  * Searchers and schedulers requiring checkpoint in the middle of training or tune report after epochs are not supported
* Added notice for serializable custom model in the tutorial. The example we provided is serializable, but needs to be defined in a separate file.

- [x] Benchmark default preset hpo result
- [x] Benchmark best preset(hpo with stacking) hpo result
- [x] Add `auto` preset for `hyperparameter_tune_kwargs`

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1790, opened 2022-06-08
[HPO] Ray HPO Automm and General

yinweisu:
*Issue #, if available:*

*Description of changes:*

This is a PR that cherry-picked Automm and general changes of PR https://github.com/awslabs/autogluon/pull/1696 to avoid having a giant PR that affects multiple modules at once.

This PR happens to still add quite some lines of code compared to the original one. That's because most of changes are on Automm side and we recently had a black reformat, which results in quite a lot line changes to merge in as well. The general hpo changes should be around 1000 lines.

* Added HPO functionality for AutoMM using ray tune + ray lightning.
  * Supports `random, bayes` searchers
  * Supports `FIFO, ASHA` schedulers
  * PBT/PB2 will be supported once ray lightning supports PT 1.6. Currently there's a bug in PT 1.6 stopping PBT/PB2 to work correctly
  * Extracted `_fit()` so that model specific configs will be generated in it, and `fit()` will generate general info that are model independent
  * Loading of the model has been divided into two parts to load general info, and config specific info
  * Checkpoint averaging is only done to the best trial once HPO is finished.


sxjscience:
The logic looks good to me. I will approve once the CI gets fixed.

szha:
Job PR-1790-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1790/2/index.html

zhiqiangdon (approved):
Thanks for adding HPO support!

sxjscience (approved):
LGTM. I will try it out after it gets merged. This is a critical PR!

core/src/autogluon/core/hpo/__init__.py
```diff
@@ -0,0 +1,10 @@
+from .ray_hpo import (
+    run,
+    cleanup_trials,
+    cleanup_checkpoints,
+    EmptySearchSpace,
+    RayTuneAdapter,
+    TabularRayTuneAdapter,
+    ForecastingRayTuneAdapter,
+    AutommRayTuneAdapter
```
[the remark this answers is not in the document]
yinweisu: It shouldn't as `AutommRayTuneAdapter` does not have any class or static variable that's package dependent

core/src/autogluon/core/hpo/ray_hpo.py
```diff
@@ -0,0 +1,457 @@
...
+    trainable_args: dict,
+    search_space: dict,
+    hyperparameter_tune_kwargs: Union[dict, str],
+    metric: str,
+    mode: str,
+    save_dir: str,
+    ray_tune_adapter: RayTuneAdapter,
+    total_resources: Optional[dict] = dict(),
+    minimum_cpu_per_trial: int = 1,
+    minimum_gpu_per_trial: float = 1.0,
```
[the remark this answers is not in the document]
yinweisu: `minimum_gpu_per_trial ` is only passed to gpu resource calculator

core/src/autogluon/core/hpo/ray_hpo.py
```diff
@@ -0,0 +1,457 @@
...
+from ray_lightning.tune import get_tune_resources
+
+
+logger = logging.getLogger(__name__)
+
+MIN = 'min'
+MAX = 'max'
+
+searcher_presets = SearcherFactory.searcher_presets
+scheduler_presets = SchedulerFactory.scheduler_presets
```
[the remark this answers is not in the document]
yinweisu: I'm not putting HPO in a class because there's no state that needs to be maintained. In ray tune, the `run` function is also not wrapped in a class. Maybe we can have a `constant.py` file to store these constants.

core/src/autogluon/core/hpo/ray_hpo.py
```diff
@@ -0,0 +1,457 @@
...
+    verbose
+        0 = silent, 1 = only status updates, 2 = status and brief trial results, 3 = status and detailed trial results.
+    **kwargs
+        Additional args being passed to tune.run
+    """
+    assert mode in [MIN, MAX], f'mode is {mode}'
+    if isinstance(hyperparameter_tune_kwargs, str):
+        assert hyperparameter_tune_kwargs in ray_tune_adapter.presets, f'{hyperparameter_tune_kwargs} is not a valid option. Options are {ray_tune_adapter.presets.keys()}'
+        hyperparameter_tune_kwargs = ray_tune_adapter.presets.get(hyperparameter_tune_kwargs)
+    num_samples = hyperparameter_tune_kwargs.get('num_trials', hyperparameter_tune_kwargs.get('num_samples', None))
```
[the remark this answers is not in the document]
yinweisu: Well, this is what our current hpo logic works. We supports for example `hyperparameter_tune_kwargs='auto'` for example. I think this is still important for a good user experience.

core/src/autogluon/core/hpo/ray_hpo.py
```diff
@@ -0,0 +1,457 @@
...
+    if resources_per_trial is None:
+        resources_per_trial = ray_tune_adapter.get_resources_per_trial(
+            total_resources=total_resources,
+            num_samples=num_samples,
+            minimum_gpu_per_trial=minimum_gpu_per_trial,
+            minimum_cpu_per_trial=minimum_cpu_per_trial,
+            model_estimate_memory_usage=model_estimate_memory_usage
+        )
+    else:
+        if isinstance(ray_tune_adapter, AutommRayTuneAdapter):
```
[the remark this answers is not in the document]
yinweisu: The adapter cannot be `None`. The logic here is if `resource_per_trial` is not None, means user wants to provide their own `resource_per_trial`, then we don't calculate it ourselves. Otherwise, we do the calculation. However, if it's `AutommRayTuneAdapter`, we do calculation no matter if user provided `resource_per_trial` or not

core/src/autogluon/core/hpo/ray_hpo.py
```diff
@@ -0,0 +1,457 @@
...
+    from ray.tune.sample import Categorical
+    tune_search_space = search_space.copy()
+    for hyperparmaeter, space in search_space.items():
+        if isinstance(space, Space):
+            tune_search_space[hyperparmaeter] = space.convert_to_ray()
+        # This is a hack
+        # nested list are not correctly converted to hyperopt space by ray https://github.com/ray-project/ray/issues/24050
+        # convert list to tuple instead. models need to check for tuple and convert it back to list
+        # TODO: remove the hack once ray releases new version containing the fix
+        # TODO: remove searcher parameter passed to this function once the fix is out
```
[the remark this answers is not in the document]
yinweisu: Good idea. It is closed and I'll need to check if it's being shipped with the recent ray release or not.

core/src/autogluon/core/hpo/ray_hpo.py
```diff
@@ -0,0 +1,457 @@
...
+        self,
+        total_resources: dict,
+        num_samples: int,
+        minimum_cpu_per_trial: int = 1,
+        minimum_gpu_per_trial: float = 1.0,
+        model_estimate_memory_usage: Optional[int] = None,
+        **kwargs,
+        ) -> dict:
+        assert isinstance(minimum_cpu_per_trial, int) and minimum_cpu_per_trial >= 1, 'minimum_cpu_per_trial must be a integer that is larger than 0'
+        assert isinstance(minimum_gpu_per_trial, (int, float)) and minimum_gpu_per_trial > 0, 'minimum_gpu_per_trial must be a integer or float that is larger than 0'
```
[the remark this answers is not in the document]
yinweisu: So you only specify `minimum_gpu_per_trial` if your trial requires gpu and you provided gpu in `total_resource`. Otherwise, this parameter will do nothing

core/src/autogluon/core/hpo/resources_calculator.py
```diff
@@ -0,0 +1,172 @@
...
+    def get_resources_per_trial(self, **kwargs) -> dict:
+        """Calculate resources per trial and return additional info"""
+        raise NotImplementedError
+    
+
+class CpuResourceCalculator(ResourceCalculator):
+    
+    @property
+    def type(self):
+        return 'cpu'
```
gradientsky, line 25: nit: potentially a bad name for property. `type` is a [built-in function](https://docs.python.org/3/library/stdtypes.html#type-objects).

core/src/autogluon/core/hpo/resources_calculator.py
```diff
@@ -0,0 +1,172 @@
...
+        total_num_cpus,
+        num_samples,
+        minimum_cpu_per_trial,
+        **kwargs,
+    ):
+        # TODO: for cpu case, is it better to have more workers or more cpus per worker?
+        cpu_per_job = max(minimum_cpu_per_trial, total_num_cpus // num_samples)
+        num_parallel_jobs = min(num_samples, total_num_cpus // cpu_per_job)
+        num_workers = max(minimum_cpu_per_trial, cpu_per_job - 1)  # 1 cpu for master process
+        cpu_per_worker = 1
```
gradientsky, line 100: ditto: no memory assessments?

core/src/autogluon/core/space.py
```diff
@@ -123,6 +127,14 @@ def convert_to_sklearn(self):
             sampler = uniform(self.lower, self.upper - self.lower)
         return sampler
 
+    def convert_to_ray(self):
```
gradientsky, line 130: Design concerns: `Space` should not be aware of all possible ways to convert it. Conversion strategy should be decoupled from the entity. See [adapter pattern](https://refactoring.guru/design-patterns/adapter/python/example).
I would expect that we will pass `Space` to Ray-enabled sub-code, where it would be wrapped into Ray-specific adapter, which knows how to translate `Space` into ray spaces.

text/src/autogluon/text/automm/constants.py
```diff
@@ -42,6 +42,20 @@
...
+    ROOT_MEAN_SQUARED_ERROR: MIN,
+    R2: MAX,
+    QUADRATIC_KAPPA: MAX,
+    ROC_AUC: MAX,
+    LOG_LOSS: MIN,
+    CROSS_ENTROPY: MIN,
+    PEARSONR: MAX,
+    SPEARMANR: MAX,
+}
+VALID_METRICS = METRIC_MODE_MAP.keys()
```
gradientsky, line 58: Can we use `higher_is_better` instead of manually hard-coding? See to [torchmetrics](https://torchmetrics.readthedocs.io/en/stable/pages/overview.html#metrics-and-hyperparameter-optimization).

text/src/autogluon/text/automm/utils.py
```diff
@@ -125,6 +124,62 @@ def infer_metrics(
...
+    -------
+    mode
+        The min/max mode used in selecting model checkpoints.
+        - min
+             Its means that smaller metric is better.
+        - max
+            It means that larger metric is better.
+    """
+    assert metric_name in METRIC_MODE_MAP, f"{metric_name} is not a supported metric. Options are: {VALID_METRICS}"
+    return METRIC_MODE_MAP.get(metric_name)
```
gradientsky, line 146: use metric's [`higher_is_better`](https://torchmetrics.readthedocs.io/en/stable/pages/overview.html#metrics-and-hyperparameter-optimization) instead.

text/src/autogluon/text/automm/predictor.py
```diff
@@ -452,78 +435,153 @@ def fit(
...
+            "hyperparameters"
+        ] = sampled_hyperparameters  # The original hyperparameters is the search space, replace it with the hyperparameters sampled
+        _fit_args["save_path"] = tune.get_trial_dir()  # We want to save each trial to a separate directory
+        logger.debug(f"hpo trial save_path: {_fit_args['save_path']}")
+        if checkpoint_dir is not None:
+            _fit_args["resume"] = True
+            _fit_args["ckpt_path"] = os.path.join(checkpoint_dir, RAY_TUNE_CHECKPOINT)
+        self._fit(**_fit_args)
+
     def _setup_distillation(
```
gradientsky, line 585: this method is too long. Can we split it into smaller logical units of work so it is easier to read and maintain?

core/src/autogluon/core/hpo/ray_hpo.py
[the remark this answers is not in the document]
yinweisu: I think you are right. We can only keep `num_trials`.

core/src/autogluon/core/hpo/ray_hpo.py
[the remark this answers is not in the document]
yinweisu: Good point. I'll pass `resource_per_trial` to each adapter's `get_resources_per_trial()` and let them decide what to do. For example, `TabularAdapter` could simply do nothing. This way we don't need to change the code if there are more adapters in the future

core/src/autogluon/core/hpo/resources_calculator.py
```diff
@@ -0,0 +1,172 @@
...
+        total_num_gpus,
+        num_samples,
+        minimum_cpu_per_trial,
+        minimum_gpu_per_trial,
+        **kwargs,
+    ):
+        gpu_per_job = max(minimum_gpu_per_trial, total_num_gpus // num_samples)
+        num_parallel_jobs = total_num_gpus // gpu_per_job  # num_parallel_jobs purely based on gpu
+        cpu_per_job = max(minimum_cpu_per_trial, total_num_cpus // num_parallel_jobs)
+        num_parallel_jobs = min(num_parallel_jobs, total_num_cpus // cpu_per_job)  # update num_parallel_jobs in case cpu is not enough
```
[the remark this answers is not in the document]
yinweisu: This becomes tricky here. Normally if a job is running on gpu, the mem usage is not that high. I guess what we want is `model_estimate_gpu_memory_usage`, and gpu ram are actually split into chunks(if we have multiple gpus) and how to properly calculate it is unclear.

core/src/autogluon/core/hpo/resources_calculator.py
yinweisu: automm doesn't currently has memory estimation, but maybe it's still a good idea to have the logic here in case

core/src/autogluon/core/space.py
yinweisu: Good point!

yinweisu: I guess this is not a strict adapter pattern because we don't want to implement the adapter for ray tune search space interface. This is more like a converter, but I can make it to be a standalone object instead of embedded in `Space` so that it's decoupled

text/src/autogluon/text/automm/constants.py
yinweisu: I think this is out of scope in this PR, and the change could result in other changes in AutoMM. We can discuss it and improve it in a separate PR

sxjscience: Yeah, we can refine this part in a separate PR. We can actually mark it as a TODO.

text/src/autogluon/text/automm/predictor.py
yinweisu: I think you refer to `_hyperparameter_tune()`This method is long because of many function signatures, for example `run()` and `_top_k_avetage()`, moving other things in to function will only short it by a little. For better readability, I've added comment on each logical operation.

core/src/autogluon/core/hpo/ray_hpo.py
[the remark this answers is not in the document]
yinweisu: https://github.com/awslabs/autogluon/issues/1798

text/src/autogluon/text/automm/constants.py
yinweisu: Created an issue to track: https://github.com/awslabs/autogluon/issues/1799

gradientsky (approved):

szha:
Job PR-1790-5 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1790/5/index.html

PR #1795, opened 2022-06-09
add early stopping for GluonTS model during training

huibinshen:
*Issue #, if available:*

Enable training early stopping for GluonTS model.

Local tests passed.

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/gluonts/callback.py
+++ b/timeseries/src/autogluon/timeseries/models/gluonts/callback.py
@@ -37,3 +37,26 @@ class TimeLimitCallback(Callback):
                 logger.warning("Time limit exceed during training, stop training.")
                 return False
         return True
+
+
+class EarlyStoppingCallback(Callback):
+    """GluonTS callback to early stop the training if the validation loss
+    is not improved for `patience' round. For the GluonTS models used in autogluon,
+    the loss is always minimized."""
+
+    def __init__(self, patience=10):
+        self.patience = patience
+        self.best_round = 0
+        self.best_loss = float('inf')
+
+    def on_validation_epoch_end(self, epoch_no, epoch_loss, **kwargs):
+        if epoch_loss < self.best_loss:
+            self.best_loss = epoch_loss
+            self.best_round = epoch_no
+            return True
+        else:
+            contniue = (epoch_no - self.best_round) < self.patience
+            if not contniue:
+                logger.warning(f"Early stopping triggered, stop training. Best epoch {self.best_round}")
+            return contniue
+
--- a/timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
+++ b/timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
@@ -22,7 +22,7 @@ from ...dataset.ts_dataframe import TIMESTAMP, ITEMID
 from ...utils.metric_utils import METRIC_COEFFICIENTS
 from ...utils.warning_filters import evaluator_warning_filter, serialize_warning_filter
 from ..abstract import AbstractTimeSeriesModel
-from .callback import TimeLimitCallback
+from .callback import TimeLimitCallback, EarlyStoppingCallback
 
 logger = logging.getLogger(__name__)
 
@@ -151,8 +151,8 @@ class AbstractGluonTSModel(AbstractTimeSeriesModel):
                     "during initialization. Please provide a `freq` string to `fit`."
                 )
 
-        if "callback" in kwargs:
-            self.callbacks.append(kwargs["callback"])
+        if "callbacks" in kwargs:
+            self.callbacks += kwargs["callbacks"]
 
     def _get_model_params(self) -> dict:
         """Gets params that are passed to the inner model."""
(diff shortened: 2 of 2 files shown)


szha:
Job PR-1795-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1795/1/index.html

gradientsky:
please add unit tests for early stopping logic.

canerturkmen (commented):
I checked the GluonTS source. I remember we used to have early stopping by default but now I couldn't find it. In any case, why do we want to enable it for autogluon (as GluonTS has moved away and to learning rate scheduling by default) and if so, why do we want to enable it by default without giving the user the choice to do it or even adjust the patience?

Maybe for now should we just keep the callback, not add it to models, and add a unit test?

timeseries/src/autogluon/timeseries/models/gluonts/callback.py
```diff
@@ -37,3 +37,26 @@ def on_epoch_end(
...
+        self.best_round = 0
+        self.best_loss = float('inf')
+
+    def on_validation_epoch_end(self, epoch_no, epoch_loss, **kwargs):
+        if epoch_loss < self.best_loss:
+            self.best_loss = epoch_loss
+            self.best_round = epoch_no
+            return True
+        else:
+            contniue = (epoch_no - self.best_round) < self.patience
```
Innixma, line 58: nit: typo on continue

timeseries/src/autogluon/timeseries/models/gluonts/callback.py
```diff
@@ -37,3 +37,26 @@ def on_epoch_end(
...
                 return False
         return True
+
+
+class EarlyStoppingCallback(Callback):
+    """GluonTS callback to early stop the training if the validation loss
+    is not improved for `patience' round. For the GluonTS models used in autogluon,
+    the loss is always minimized."""
+
+    def __init__(self, patience=10):
```
Innixma, line 47: We need to be careful here, this is not a good patience value to use in all situations. This is a complex topic and I put a lot of effort into this in Tabular.

For example, I have Early stopping logic coded here: https://github.com/awslabs/autogluon/blob/master/core/src/autogluon/core/utils/early_stopping.py

In particular, AdaptiveES will be much better than naive patience based ES. But regardless, we should err on the side of caution and use high patience values rather than low ones, otherwise we risk crippling the models in situations where time isn't much of a concern.

huibinshen: Thanks for the comments!

I am not very sure about 10 is not good for *all* situations but I agree that we need some data points to set this default value. Do you have some results to share about the adaptive early stopping?

I also implemented the adaptive early stopping logic for GluonTS model, there was quite some code duplicates and I am thinking of talking the `_update_patience` out of the class `AdaptiveES` so I could reuse it. Maybe you have better suggestion.

I also remove the default usage of early stopping for now. After we have more results for it, we could set the default option later based on that. Let me know what do you think of this option.

timeseries/src/autogluon/timeseries/models/gluonts/callback.py
huibinshen: Changed, thanks!

huibinshen:
> I checked the GluonTS source. I remember we used to have early stopping by default but now I couldn't find it. In any case, why do we want to enable it for autogluon (as GluonTS has moved away and to learning rate scheduling by default) and if so, why do we want to enable it by default without giving the user the choice to do it or even adjust the patience?
> 
> Maybe for now should we just keep the callback, not add it to models, and add a unit test?
Thanks Caner for the comments!

I asked Lorenzo and Jasper about it, they all think that it can be worth to use it (looking back, they would have not removed it). But I think the default choice worth a study so I don't use it for now. After we have more data points, we could change the default option then. Let me know what do you think of this option.

szha:
Job PR-1795-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1795/2/index.html

timeseries/src/autogluon/timeseries/models/gluonts/callback.py
Innixma: Sorry for confusion, when I said "all situations" I meant that in some situations it might not be good, not that it is always not good.

Regarding your implementation, to avoid code dupe why not directly use the AdaptiveES code? You can import it and use it in your callback to avoid code dupe. I do this for individual models, for example, having it as `self.es`: https://github.com/awslabs/autogluon/blob/master/tabular/src/autogluon/tabular/models/xgboost/callbacks.py#L27-L30

We probably shouldn't remove default usage of early stopping, just err on the side of caution and use a larger value. What value is best will be understood after we start doing benchmarking. Maybe 10 is fine but it depends on the model.

canerturkmen: perhaps even 5 is fine if we have it on by default. if anything it's a little too cautious considering that we work with 50-100 epochs.

huibinshen: Re the code duplicates, I followed your suggestions. For the default patience, I put it to 20 for now. We may change it later with more data points.

szha:
Job PR-1795-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1795/3/index.html

Innixma:
@canerturkmen I'll leave this PR to you, feel free to approve and merge if you wish.

huibinshen:
Closing this and will create another one.

Issue #1762, opened 2022-05-28
[BUG] Segmentation fault while importing TextPredictor


cjj490168650:
- [X] I have checked that this bug exists on the latest stable version of AutoGluon
- [ ] and/or I have checked that this bug exists on the latest mainline of AutoGluon via source installation

**Describe the bug**

```
>>> from autogluon.text import TextPredictor
/home/cjj/anaconda3/envs/pytorch/lib/python3.9/site-packages/torchvision/io/image.py:11: UserWarning: Failed to load image Python extension: libc10_cuda.so: cannot open shared object file: No such file or directory
  warn(f"Failed to load image Python extension: {e}")
Segmentation fault (core dumped)
```

**Expected behavior**

import TextPredictor

**To Reproduce**

```python
from autogluon.text import TextPredictor
```

**Screenshots**

![QQ图片20220528200354](https://user-images.githubusercontent.com/45029740/170824684-b4396233-6298-4e57-954d-137802573218.png)

**Installed Versions**

<details>

```
/home/cjj/anaconda3/envs/pytorch/lib/python3.9/site-packages/torchvision/io/image.py:11: UserWarning: Failed to load image Python extension: libc10_cuda.so: cannot open shared object file: No such file or directory
  warn(f"Failed to load image Python extension: {e}")
/home/cjj/anaconda3/envs/pytorch/lib/python3.9/site-packages/gluoncv/__init__.py:40: UserWarning: Both `mxnet==1.9.1` and `torch==1.10.2+cpu` are installed. You might encounter increased GPU memory footprint if both framework are used at the same time.
  warnings.warn(f'Both `mxnet=={mx.__version__}` and `torch=={torch.__version__}` are installed. '

INSTALLED VERSIONS
------------------
date                 : 2022-05-28
time                 : 08:05:29.672541
python               : 3.9.12.final.0
OS                   : Linux
OS-release           : 4.15.0-76-generic
Version              : #86-Ubuntu SMP Fri Jan 17 17:24:28 UTC 2020
machine              : x86_64
processor            : x86_64
num_cores            : 40
cpu_ram_mb           : 257841
cuda version         : 11.515.43.04
num_gpus             : 10
gpu_ram_mb           : [2886, 12195, 12195, 6636, 3766, 3766, 3766, 1598, 6930, 4332]
avail_disk_size_mb   : 1487138

autogluon.common     : 0.4.2b20220528
autogluon.core       : 0.4.2b20220528
[...]

PR #1802, opened 2022-06-10
Cap protobuf version in text/automm

sxjscience:
*Issue #, if available:*

*Description of changes:*

Cap protobuf version to help solve https://github.com/awslabs/autogluon/issues/1762

### DIFF
--- a/text/setup.py
+++ b/text/setup.py
@@ -37,6 +37,7 @@ install_requires = [
     'nptyping>=1.4.4,<1.5.0',
     'omegaconf>=2.1.1,<2.2.0',
     'sentencepiece>=0.1.95,<0.2.0',
+    'protobuf<=3.18.1',  # https://github.com/awslabs/autogluon/issues/1762
     f'autogluon.core=={version}',
     f'autogluon.features=={version}',
     f'autogluon.common=={version}',


zhiqiangdon (approved):

text/setup.py
```diff
@@ -37,6 +37,7 @@
     'nptyping>=1.4.4,<1.5.0',
     'omegaconf>=2.1.1,<2.2.0',
     'sentencepiece>=0.1.95,<0.2.0',
+    'protobuf<=3.18.1',  # https://github.com/awslabs/autogluon/issues/1762
```
yinweisu, line 40: So there are two versions range of protobuf exists in autogluon now. We probably want to just have 1. Otherwise, which version will be installed is unclear.

szha:
Job PR-1802-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1802/1/index.html

text/setup.py
Innixma: Agreed, it needs to be specified in the core setup_utils and both text and core must use that method of specifying verison

Innixma: Agreed, it needs to be specified in the core setup_utils and both text and core must use that method of specifying verison

Innixma (changes requested):
Refer to comments

text/setup.py
sxjscience: Will the dependency resolver in pypi be able to handle that automatically? https://pip.pypa.io/en/stable/topics/dependency-resolution/ . For example, we may not ask core to depend on protobuf /limit the version but rely on the dependency resolver of pypi to figure that out automatically?

Disclaimer: I'm not the expert on that. But constraining the version in core due to version constraint in the sub-package does not sound right to me.

sxjscience: Ideally, the version caps in sub-packages such as text/image/timeseries should not impact the version caps in core. This helps these sub-packages be loosely coupled with each other.

Innixma: @sxjscience What you describe makes sense, but it comes at the cost of increased complexity for detecting issues, as now the presence of AutoGluon-Text impacts the version of Protobuf in core, and thus it is harder to test that both versions (with and without text installed) work correctly. I'm open to either option, but would generally recommend having one golden version of package we are confident works across everything than potentially many different versions depending on which subset of modules are installed by the user.

sxjscience: My point is that core should not depend on unnecessary packages. For example, users should be able to use core without installing protobuf.

yinweisu: So ray also has the protobuf issue, and that's why core might depend on protobuf. Even though the new ray release has the protobuf fix, but it should be addressed in another PR.

Innixma: Ok, I'll leave the details of this to @sxjscience and @yinweisu , I am ok with either approach.

Innixma (approved):

text/setup.py
sxjscience: If I’m going to revise core, should I revise the version cap in ray @weisy?

yinweisu: Actually, I think maybe separate them out and let pypi to resolve the conflict is better because each package might have a different protobuf version range. It would be non-trivial to figure it out manually by ourselves every time. 

In the meanwhile, I'll try to update ray with new release to avoid the protobuf issue on their side

yinweisu (approved):

PR #1826, opened 2022-06-15
Fix time series dataframe freq inference

canerturkmen:
*Issue #, if available:*

*Description of changes:*
- Adds the option of inferring data frame frequency from the first element's index. This handles edge cases where pandas cannot tell the inferred frequency of multiple concatenated date time indices -- an edge case we discovered during benchmarking.
- Temporarily disables "uniform time series index" check, since pandas appears to sometimes get confused by timedeltas of e.g., months (some months have 30 days and others 31, so not uniform). This led the benchmarks to spuriously fail.
- Moves inner logic of `from_***` constructors out of the constructor methods. This is because we used the constructor methods also in `__init__`, which led to an (inadvertent) recursion.

cc: @huibinshen 

### DIFF
--- a/timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
+++ b/timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import itertools
-from typing import Any, Tuple, Type, Optional
+from typing import Any, Tuple, Type
 from collections.abc import Iterable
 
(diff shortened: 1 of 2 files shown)


szha:
Job PR-1826-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1826/1/index.html

Innixma (approved):
LGTM!

PR #1833, opened 2022-06-16
Improve get_oof_pred_proba for refit models

Innixma:
*Issue #, if available:*

*Description of changes:*
- Previously, if `best_model` was not able to predict (aka not saved), calling `get_oof_pred_proba` would error even if the oof_pred_proba was technically available.
  - This PR fixes this.
- Refit models can now be called in `get_oof_pred_proba` without crashing. The `oof_pred_proba` of their parent (original) model will be used as a proxy.
- Code cleanup to maintain parent <-> refit relationship in DAG instead of in a separate dictionary.

### DIFF
--- a/core/src/autogluon/core/trainer/abstract_trainer.py
+++ b/core/src/autogluon/core/trainer/abstract_trainer.py
@@ -767,11 +767,11 @@ class AbstractTrainer:
                 if len(models_trained) == 1:
                     model_full_dict[model_name] = models_trained[0]
                 for model_trained in models_trained:
-                    self._add_model_attr(model_trained,
-                                         refit_full=True,
-                                         refit_full_parent=model_name,
-                                         refit_full_parent_val_score=self.get_model_attribute(model_name, 'val_score'),
-                                         )
+                    self._update_model_attr(model_trained,
+                                            refit_full=True,
+                                            refit_full_parent=model_name,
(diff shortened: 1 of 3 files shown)


szha:
Job PR-1833-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1833/1/index.html

szha:
Job PR-1833-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1833/2/index.html

Innixma:
(Note: CI passed for this logic, the error is due to issues outside of this code change

core/src/autogluon/core/trainer/abstract_trainer.py
```diff
@@ -1209,6 +1217,12 @@ def _add_model(self, model: AbstractModel, stack_name: str = 'core', level: int
             del model
         return True
 
+    def _add_model_attr(self, model: str, **attributes):
```
yinweisu, line 1220: This function name gives people impression that it only add attributes. However, it actually update the attribute. Maybe rename to `_update_model_attr`?

yinweisu (approved):
LGTM! Added one tiny comment

core/src/autogluon/core/trainer/abstract_trainer.py
Innixma: Good catch, updated!

PR #1836, opened 2022-06-17
[WIP] Support PT 1.11

sxjscience:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/multimodal/setup.py
+++ b/multimodal/setup.py
@@ -27,7 +27,7 @@ install_requires = [
     'boto3',
     'setuptools<=59.5.0',
     'timm<0.6.0',
-    'torch>=1.0,<1.11',
+    'torch>=1.0,<1.12',
     'fairscale>=0.4.5,<0.5.0',
     'scikit-image>=0.19.1,<0.20.0',
     'smart_open>=5.2.1,<5.3.0',


multimodal/setup.py
```diff
@@ -27,13 +27,13 @@
...
     'timm<0.6.0',
-    'torch>=1.0,<1.11',
+    'torch>=1.0,<1.12',
     'fairscale>=0.4.5,<0.5.0',
     'scikit-image>=0.19.1,<0.20.0',
     'smart_open>=5.2.1,<5.3.0',
     'pytorch_lightning>=1.5.10,<1.7.0',
     'torchmetrics>=0.7.2,<0.8.0',
-    'transformers>=4.16.2,<4.17.0',
+    'transformers>=4.18.0,<4.21.0',
```
sxjscience, line 36: Need to upgrade transformers to support Torch 1.11 due to the availability of https://github.com/huggingface/transformers/pull/16043 starting from 4.18.0

multimodal/setup.py
```diff
@@ -25,15 +25,14 @@
     'Pillow',
     'tqdm',
     'boto3',
-    'setuptools<=59.5.0',
```
sxjscience, line 28: @ddelange Trying to upgrade to PT 1.11 and remove the setuptools pin in this PR.

andreas12345 (approved):

multimodal/setup.py
```diff
@@ -25,15 +25,14 @@
...
     'timm<0.6.0',
-    'torch>=1.0,<1.11',
+    'torch>=1.0,<1.12',
```
Innixma, line 29: You need to update this in Tabular: https://github.com/awslabs/autogluon/blob/master/tabular/setup.py#L42

sxjscience: Let me try it out after our CI is fixed.

andreas12345: I just tested autogluon successfully with torch 1.13 (nightly build). So I wonder, is there a need to be so conservative about torch versions, or can we just require 1.x? Have there be any cases of changes being necessary to support torch minor version upgrades?

sxjscience: I just checked the version requirement of pytorch_lightning: https://github.com/Lightning-AI/lightning/blob/554056c21ae191b7bd2ea669d9ac84917dacdc6d/requirements/base.txt#L2 . They are also strict for torch and we may follow what they are doing.

ddelange: > I just checked the version requirement of pytorch_lightning: https://github.com/Lightning-AI/lightning/blob/554056c21ae191b7bd2ea669d9ac84917dacdc6d/requirements/base.txt#L2 . They are also strict for torch and we may follow what they are doing.

makes sense to me: torch does not guarantee strict semver2 as they sometimes ship breaking changes with a minor version release (also the case e.g. with 1.11)

gradientsky:
Installation instructions need to be updated (still instructing `<1.11`)

gradientsky (approved):

Innixma (approved):
LGTM!

szha:
Job PR-1836-9 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1836/9/index.html

PR #1837, opened 2022-06-17
[Ray] Ray manager

yinweisu:
*Issue #, if available:*

*Description of changes:*
* Create a singleton to manage ray cluster resource. Also, it will only init the ray cluster if config has changed to avoid needless shutdown

### DIFF
--- a/core/src/autogluon/core/ray/manager.py
+++ b/core/src/autogluon/core/ray/manager.py
@@ -1,0 +1,20 @@
+from autogluon.core.utils.try_import import try_import_ray
+
+class RayManager:
+    """
+    A singleton maintaining status of ray cluster resources
+    """
+    
+    ray = None
+    current_config = None
+    
+    @staticmethod
+    def init_ray(**config):
+        """Init ray only if the configuration changes"""
+        if RayManager.ray is None:
+            RayManager.ray = try_import_ray()
+        if RayManager.current_config != config:
+            if RayManager.ray.is_initialized():
+                RayManager.ray.shutdown()
+            RayManager.ray.init(**config)
+            RayManager.current_config = config
--- a/core/src/autogluon/core/hpo/ray_hpo.py
+++ b/core/src/autogluon/core/hpo/ray_hpo.py
@@ -5,7 +5,6 @@ import shutil
 
 from ..utils.try_import import try_import_ray
 try_import_ray()  # try import ray before importing the remaining contents so we can give proper error messages
-import ray
 
 from abc import ABC, abstractmethod
 from typing import Optional, Callable, Union, List
@@ -16,6 +15,7 @@ from .constants import (
     SEARCHER_PRESETS,
     SCHEDULER_PRESETS
 )
+from ..ray.manager import RayManager
 from ..ray.resources_calculator import ResourceCalculatorFactory, ResourceCalculator
 from .scheduler_factory import SchedulerFactory
 from .searcher_factory import SearcherFactory
@@ -212,10 +212,8 @@ def run(
     scheduler = _get_scheduler(hyperparameter_tune_kwargs, supported_schedulers=ray_tune_adapter.get_supported_schedulers())
     search_space = _convert_search_space(search_space)
 
-    if ray.is_initialized():
-        # shutdown to reinitialize resources because different model might require different total resources
-        ray.shutdown()
-    ray.init(log_to_driver=False, **total_resources)
+    
+    RayManager.init_ray(log_to_driver=False, **total_resources)
 
     resources_per_trial = hyperparameter_tune_kwargs.get('resources_per_trial', None)
     resources_per_trial = ray_tune_adapter.get_resources_per_trial(
--- a/core/src/autogluon/core/models/ensemble/fold_fitting_strategy.py
+++ b/core/src/autogluon/core/models/ensemble/fold_fitting_strategy.py
@@ -13,6 +13,7 @@ from pandas import DataFrame, Series
 
 from autogluon.common.utils.pandas_utils import get_approximate_df_mem_usage
 
+from ...ray.manager import RayManager
 from ...ray.resources_calculator import ResourceCalculatorFactory
 from ...utils.exceptions import TimeLimitExceeded, NotEnoughMemoryError, NotEnoughCudaMemoryError
 from ...utils import get_cpu_count, get_gpu_count_all
@@ -404,11 +405,10 @@ class ParallelLocalFoldFittingStrategy(LocalFoldFittingStrategy):
         return num_cpus
 
     def schedule_fold_model_fit(self, fold_ctx):
-        if not self.ray.is_initialized():
-            if self.num_gpus > 0:
-                self.ray.init(log_to_driver=False, num_gpus=self.num_gpus)
-            else:
-                self.ray.init(log_to_driver=False)
+        ray_init_args = dict(log_to_driver=False, num_cpus=self.num_cpus)
+        if self.num_gpus > 0:
+            ray_init_args['num_gpus'] = self.num_gpus
+        RayManager.init_ray(**ray_init_args)
         self.jobs.append(fold_ctx)
 
     def after_all_folds_scheduled(self):


core/src/autogluon/core/ray/manager.py
```diff
@@ -0,0 +1,20 @@
+from autogluon.core.utils.try_import import try_import_ray
+
+class RayManager:
```
sxjscience, line 3: Why will we need this singleton?

sxjscience: Actually, why not just rely on `ray.init()`?

sxjscience:
What's the use-case for initializing a ray cluster with changed configuration? Usually we call `ray.init()` only once in the source code.

yinweisu:
Sometimes we need to assign the resources used by ray. This can happen in tabular, where some models are trained faster with only all physical cores while some are trained faster by using virtual cores. In such a scenario, we will need to re-init the ray cluster

core/src/autogluon/core/ray/manager.py
yinweisu: Multiple places could use ray. For example, tabular bagging and hpo are both using ray. Having a singleton here to manage ray easier.

sxjscience: I feel that this is not how a Singleton pattern is usually implemented: https://www.geeksforgeeks.org/singleton-pattern-in-python-a-complete-guide/amp/ . It is used when we need to protect certain resources. However, what’s stored here is the module “ray”, which itself can be viewed as a singleton.

sxjscience: Also, does ray provide methods for accessing the config, e.g., something like ray._config?

core/src/autogluon/core/ray/manager.py
```diff
@@ -0,0 +1,20 @@
...
+    current_config = None
+    
+    @staticmethod
+    def init_ray(**config):
+        """Init ray only if the configuration changes"""
+        if RayManager.ray is None:
+            RayManager.ray = try_import_ray()
+        if RayManager.current_config != config:
+            if RayManager.ray.is_initialized():
+                RayManager.ray.shutdown()
```
sxjscience, line 18: Shutdown is a very time consuming process and we may avoid shutting down the cluster silently.

sxjscience: Usually, the expectation is to call init once and shutdown once.

sxjscience:
> Sometimes we need to assign the resources used by ray. This can happen in tabular, where some models are trained faster with only all physical cores while some are trained faster by using virtual cores. In such a scenario, we will need to re-init the ray cluster

Does it have to be optimized this way? Should we actually assign different resource requirements to different jobs and ask ray to schedule that?

core/src/autogluon/core/ray/manager.py
yinweisu: So you can't really implement singleton in python because there's no private constructor...
The singleton is mainly here to maintain the config not ray module.

core/src/autogluon/core/ray/manager.py
yinweisu: We have to shutdown if we want to change the total resource

yinweisu:
> > Sometimes we need to assign the resources used by ray. This can happen in tabular, where some models are trained faster with only all physical cores while some are trained faster by using virtual cores. In such a scenario, we will need to re-init the ray cluster
> 
> Does it have to be optimized this way? Should we actually assign different resource requirements to different jobs and ask ray to schedule that?

We cannot. Ray sees the total resource and will continue to assign jobs if there are more resources.

core/src/autogluon/core/ray/manager.py
sxjscience: If that’s the case, we may try to check if there are other ways to access the ray config and return that with a function. This singleton implementation looks unnecessary to me because it wraps a module.

core/src/autogluon/core/ray/manager.py
yinweisu: A local shutdown is not too bad. In the case of a remote cluster, we are more than likely not needing this logic anymore

sxjscience:
> > > Sometimes we need to assign the resources used by ray. This can happen in tabular, where some models are trained faster with only all physical cores while some are trained faster by using virtual cores. In such a scenario, we will need to re-init the ray cluster
> > 
> > 
> > Does it have to be optimized this way? Should we actually assign different resource requirements to different jobs and ask ray to schedule that?
> 
> We cannot. Ray sees the total resource and will continue to assign jobs if there are more resources.

If that’s the case, we should actually revise the scheduler in ray rather than calling init and shutdown. This may be problematic in the distributed training setting. Also, what’s the expected speed boost?

core/src/autogluon/core/ray/manager.py
yinweisu: The reason I wrap a module is so that we can do lazy import.

yinweisu:
> > > > Sometimes we need to assign the resources used by ray. This can happen in tabular, where some models are trained faster with only all physical cores while some are trained faster by using virtual cores. In such a scenario, we will need to re-init the ray cluster
> > > 
> > > 
> > > Does it have to be optimized this way? Should we actually assign different resource requirements to different jobs and ask ray to schedule that?
> > 
> > 
> > We cannot. Ray sees the total resource and will continue to assign jobs if there are more resources.
> 
> If that’s the case, we should actually revise the scheduler in ray rather than calling init and shutdown. This may be problematic in the distributed training setting. Also, what’s the expected speed boost?

Then what logic should you have in the scheduler? Whether to use only half the resource(number of physical cores) for scheduling? What if there's no hyperthreading for this cpu.

core/src/autogluon/core/ray/manager.py
sxjscience: How about doing that with a function? Ray should have stored the configuration somewhere.

core/src/autogluon/core/ray/manager.py
sxjscience: If that’s the case, I recommend make it part of tabular rather than core. Functions in core should be concise and reusable.

sxjscience:
> > > > > Sometimes we need to assign the resources used by ray. This can happen in tabular, where some models are trained faster with only all physical cores while some are trained faster by using virtual cores. In such a scenario, we will need to re-init the ray cluster
> > > > 
> > > > 
> > > > Does it have to be optimized this way? Should we actually assign different resource requirements to different jobs and ask ray to schedule that?
> > > 
> > > 
> > > We cannot. Ray sees the total resource and will continue to assign jobs if there are more resources.
> > 
> > 
> > If that’s the case, we should actually revise the scheduler in ray rather than calling init and shutdown. This may be problematic in the distributed training setting. Also, what’s the expected speed boost?
> 
> Then what logic should you have in the scheduler? Whether to use only half the resource(number of physical cores) for scheduling? What if there's no hyperthreading for this cpu.

This might be something handled by the scheduler. You can ask the process to provide additional information on the priority/resource requirement. Also, if such logic is only used in Tabular, you may not put it to core.

core/src/autogluon/core/ray/manager.py
yinweisu: So this is not tabular only. All this is doing is if the user wants to init with a different config, do it. Otherwise, it will do nothing. Why it's not reusable?

yinweisu:
> > > > > > Sometimes we need to assign the resources used by ray. This can happen in tabular, where some models are trained faster with only all physical cores while some are trained faster by using virtual cores. In such a scenario, we will need to re-init the ray cluster
> > > > > 
> > > > > 
> > > > > Does it have to be optimized this way? Should we actually assign different resource requirements to different jobs and ask ray to schedule that?
> > > > 
> > > > 
> > > > We cannot. Ray sees the total resource and will continue to assign jobs if there are more resources.
> > > 
> > > 
> > > If that’s the case, we should actually revise the scheduler in ray rather than calling init and shutdown. This may be problematic in the distributed training setting. Also, what’s the expected speed boost?
> > 
> > 
> > Then what logic should you have in the scheduler? Whether to use only half the resource(number of physical cores) for scheduling? What if there's no hyperthreading for this cpu.
> 
> This might be something handled by the scheduler. You can ask the process to provide additional information on the priority/resource requirement. Also, if such logic is only used in Tabular, you may not put it to core.

We are already providing resource requirement of a job to ray, and we are providing the total resource. The thing is ray will keep schedule job if there's resources. We cannot simply double the job's resource requirement so that we allocate half amount of job. This way each job is getting double resources, which is incorrect

core/src/autogluon/core/ray/manager.py
yinweisu: Yea, if ray have that info, then a function makes sense

PR #1842, opened 2022-06-19
Removing defaults channel

gradientsky:
*Description of changes:*
Removed defaults channel since it's a bad practice

### DIFF
--- a/docs/build_gpu.yml
+++ b/docs/build_gpu.yml
@@ -1,7 +1,6 @@
 name: autogluon_py3
 channels:
 - conda-forge
-- defaults
 dependencies:
 - cudatoolkit=10.2
 - cudnn=8.2
--- a/docs/build_gpu.yml
+++ b/docs/build_gpu.yml
@@ -2,9 +2,8 @@ name: autogluon_py3
 channels:
 - conda-forge
 dependencies:
-- openmp
-- cudatoolkit=10.2
-- cudnn=8.2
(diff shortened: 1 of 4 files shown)


docs/build_gpu.yml
```diff
@@ -1,8 +1,8 @@
 name: autogluon_py3
 channels:
 - conda-forge
-- defaults
 dependencies:
+- openmp
```
Innixma, line 5: why add openmp?

sxjscience:
Will doc-failure be related to https://github.com/awslabs/autogluon/blob/70af6f34d47eb85779fdcefaf1b94c4a7d16f6da/docs/build_contrib_gpu.yml#L13 ? I previously tried d2l-book in another project and I used the following version:

```python
python -m pip install -U git+https://github.com/d2l-ai/d2l-book.git@6209db8923aca9306a02bf6b65defca1ae375ec6
```

sxjscience:
Another reason is sphinx. I'm not sure what happened but we can also try to add sphinx in https://github.com/awslabs/autogluon/blob/master/docs/build_contrib_gpu.yml as dependencies rather than inside pip.

szha:
Job PR-1842-12 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1842/12/index.html

sxjscience (approved):
Thanks!!!

PR #1850, opened 2022-06-20
fix sequential args

yinweisu:
*Issue #, if available:*

*Description of changes:*
* Fix init args when parallel bagging fall back to sequential


Innixma (approved):
LGTM!

PR #1854, opened 2022-06-21
Infer soft label loss function if it's not given.

FANGAreNotGnu:
*Description of changes:*
Removing cross entropy as the default soft label loss function, since it will cause error if problem type is regression. By default, we infer the soft label loss function if not specified by user.
Also fixed some typos and enable the regression task (stsb) in the example for distillation with AutoMMPredictor.


szha:
Job PR-1854-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1854/1/index.html

PR #1859, opened 2022-06-21
Fix sktime prediction discrepancy in timeseries

canerturkmen:
*Issue #, if available:*

*Description of changes:*

Fixes sktime discrepancy when partially matching indices when unmatched lengths are passed as training and test data.

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/sktime/abstract_sktime.py
+++ b/timeseries/src/autogluon/timeseries/models/sktime/abstract_sktime.py
@@ -85,10 +85,12 @@ class AbstractSktimeModel(AbstractTimeSeriesModel):
         """
         return pd.DataFrame(
             data=data.values,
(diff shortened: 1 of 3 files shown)


timeseries/src/autogluon/timeseries/models/sktime/abstract_sktime.py
```diff
@@ -85,10 +85,12 @@ def _to_skt_data_frame(self, data: TimeSeriesDataFrame) -> pd.DataFrame:
         """
         return pd.DataFrame(
             data=data.values,
-            index=pd.MultiIndex.from_arrays([
-                data.index.get_level_values(0),
-                data.index.get_level_values(1).to_period(freq=data.freq),  # noqa
-            ]),
+            index=pd.MultiIndex.from_arrays(
```
canerturkmen, line 88: simple change on styling form black. please ignore.

PR #1838, opened 2022-06-18
TimeSeries WeightedEnsemble

Innixma:
*Issue #, if available:*
#1808

*Description of changes:*
- Add WeightedEnsemble to TimeSeries

TODO:
- Remove Print Statements
- Benchmark
- Integrate with new sktime model logic
- Fix unit tests

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1863, opened 2022-06-22
Add TimeSeries quickstart guide

canerturkmen:
*Issue #, if available:*

Depends on weighted ensemble PR: #1838 

Adds FAQ and quickstart tutorial for time series forecasting.

*Description of changes:*


Innixma (approved):
LGTM! We probably will want to refine this as a fast-follow since some of it looks a bit weird (for example, calling `data.head()` showing all values being 0), but it is good enough for the v0.5 release.

PR #1864, opened 2022-06-22
[Bugfix]The wrong load of data_processors in standalone models with old version.

Linuxdex:
*Issue #, if available:*
When using stand-alone model, the downloading is still needed when using the data_processors created by "autogluon.text.automm".
*Description of changes:*
Use the customUnpickler in data_processors.

### DIFF
--- a/multimodal/src/autogluon/multimodal/utils.py
+++ b/multimodal/src/autogluon/multimodal/utils.py
@@ -1466,19 +1466,7 @@ class CustomUnpickler(pickle.Unpickler):
 
     def find_class(self, module, name):
         renamed_module = module
-        if module == "autogluon.text.automm.data.preprocess_dataframe":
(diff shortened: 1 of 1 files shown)


multimodal/src/autogluon/multimodal/utils.py
```diff
@@ -1468,6 +1468,16 @@ def find_class(self, module, name):
         renamed_module = module
         if module == "autogluon.text.automm.data.preprocess_dataframe":
             renamed_module = "autogluon.multimodal.data.preprocess_dataframe"
+        elif module == "autogluon.text.automm.data.process_categorical":
```
sxjscience, line 1471: This can be simplified as if module.startswith(“autogluon.text.automm”). We can then replace it with autogluon.multimodal

Issue #1890, opened 2022-06-24
[BUG]predict_proba() and predict() doesn't work for Torch tensor inference with batch size larger than 1


ChengeLi:
- [x] I have checked that this bug exists on the latest stable version of AutoGluon
- [ ] and/or I have checked that this bug exists on the latest mainline of AutoGluon via source installation

**Describe the bug**
When using torch tensor (not pandas dataframe) as data format during inference, we found that the predict_proba() function will have errors when the batch size is larger than 1. The list of prediction is a flattened list instead of list of lists, and [this assignment](https://github.com/awslabs/autogluon/blob/master/vision/src/autogluon/vision/predictor/predictor.py#L546) will have errors. 

`ValueError: Columns must be same length as key`

**Expected behavior**
It should return a list of confidence scores with different class as different columns. 

**To Reproduce**

```
from PIL import Image
import torchvision.transforms as transforms
import math 
import torch


crop_pct = 0.99
img_size = 224
preprocess = transforms.Compose([
                                transforms.Resize(int(math.floor(img_size / crop_pct))),
                                transforms.CenterCrop(img_size),
                                transforms.ToTensor(),
                                transforms.Normalize(
                                    mean=[0.485, 0.456, 0.406], 
                                    std=[0.229, 0.224, 0.225]),
                                ]
                                 )
device = torch.device("cuda" if torch.cuda.is_available() else "cpu")

def predict_using_torch_tensor(model, img_files):
    batch_decoded = preprocess(Image.open(img_files[0])).unsqueeze(0)
    for img_file in img_files[1:]:
        current_img_tensor = preprocess(Image.open(img_file)).unsqueeze(0)
        batch_decoded = torch.cat((batch_decoded, current_img_tensor), 0)
    print(batch_decoded.shape)
    batch_decoded = batch_decoded.to(device)
    pred_proba = model.predict_proba(batch_decoded)
#     pred_proba = model.predict(batch_decoded)
    return pred_proba


### one-by-one inference is working
for ind in range(10):
    pred_proba = predict_using_torch_tensor(my_model, [image_list[ind]])
    print('{}'.format(np.array(pred_proba)))

### batch inference is not working
pred_proba = predict_using_torch_tensor(my_model, image_list[:10])
print('{}'.format(np.array(pred_proba)))

```

**Screenshots**
[...]

PR #1896, opened 2022-06-25
Fix predict_proba bug in image predictor

bryanyzhu:
*Issue #, if available:*
#1890. Problem is caused by using GluonCV image classifier [`predict_proba` function](https://github.com/dmlc/gluon-cv/blob/master/gluoncv/auto/estimators/image_classification/image_classification.py#L601). In this function, the [returned predicted probabilities](https://github.com/dmlc/gluon-cv/blob/master/gluoncv/auto/estimators/image_classification/image_classification.py#L631) are flattened into `BC x 1`. However, in most of the cases, the output should be in `B x C` format. B indicates batch size and C indicates number of classes. This error only happens when input data type is torch.Tensor. When input data type is str, list of strs, dataframe, etc, existing code works fine. So I added an `if` statement.

*Description of changes:*
Add an `if` statement: if input data type is torch.Tensor and we want to inference on multiple images, we reshape the predicted probabilities to `B x C` format, instead of a flattened list. 

### DIFF
--- a/multimodal/tests/unittests/test_predictor.py
+++ b/multimodal/tests/unittests/test_predictor.py
@@ -700,7 +700,7 @@ def test_hpo(searcher, scheduler):
     hyperparameters = {
         "optimization.learning_rate": tune.uniform(0.0001, 0.01),
         "optimization.max_epochs": 1,
-        "model.names": ["numerical_mlp", "categorical_mlp", "fusion_mlp"],
+        "model.names": ["numerical_mlp"],
         "env.num_workers": 0,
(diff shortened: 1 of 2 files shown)


szha:
Job PR-1896-6 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1896/6/index.html

sxjscience (approved):
LGTM

zhiqiangdon (approved):
LGTM

PR #1909, opened 2022-06-28
Disable multilayer stacking in binary

Innixma:
*Issue #, if available:*

*Description of changes:*
- Disable multilayer stacking in binary when `auto_stack=True`
- This avoids the stack info leakage problem that can commonly occur in binary classification.
- While this makes results worse on some datasets, it greatly improves results on others.
- In general, inference speed is improved significantly for binary problems.

Benchmark: 172 Binary datasets with >=10,000 rows

```
AutoGluon_bq_4h64c_2022_06_25_nn VS all
                              framework   winrate   >   <    =  % Less Avg. Errors  Avg Inf Speed Diff  time_train_s  time_infer_s  loss_rescaled  time_train_s_rescaled  time_infer_s_rescaled      rank  rank=1_count  rank=2_count  rank=3_count  rank>3_count  error_count
0  AutoGluon_bq_4h64c_2022_06_26_binary  0.531977  88  77    7            1.754443            2.354029   8079.973256      0.008257       0.459302               1.000447               1.048294  1.468023            97            84             0             0            2
1      AutoGluon_bq_4h64c_2022_06_25_nn  0.500000   0   0  174            0.000000                 NaN  11301.577907      0.021823       0.523256               1.670676               3.402322  1.531977            79            95             0             0            9

```

Results: 3.4x inference speedup, 53.2% win-rate.

### DIFF
--- a/tabular/src/autogluon/tabular/learner/abstract_learner.py
+++ b/tabular/src/autogluon/tabular/learner/abstract_learner.py
@@ -681,9 +681,20 @@ class AbstractTabularLearner(AbstractLearner):
             X = X.loc[y.index]
         return X, y
 
+    def infer_problem_type(self, y: Series, silent=False):
+        problem_type = self._infer_problem_type(y, silent=silent)
+        if self.quantile_levels is not None:
+            if problem_type == REGRESSION:
+                problem_type = QUANTILE
+            else:
+                raise ValueError("autogluon infers this to be classification problem for which quantile_levels "
+                                 "cannot be specified. If it is truly a quantile regression problem, "
+                                 "please specify:problem_type='quantile'")
+        return problem_type
+
     @staticmethod
-    def infer_problem_type(y: Series):
-        return infer_problem_type(y=y)
+    def _infer_problem_type(y: Series, silent=False):
+        return infer_problem_type(y=y, silent=silent)
 
     # Loads models in memory so that they don't have to be loaded during predictions
     def persist_trainer(self, low_memory=False, models='all', with_ancestors=False, max_memory=None) -> list:
--- a/tabular/src/autogluon/tabular/learner/abstract_learner.py
+++ b/tabular/src/autogluon/tabular/learner/abstract_learner.py
@@ -683,13 +683,16 @@ class AbstractTabularLearner(AbstractLearner):
 
     def infer_problem_type(self, y: Series, silent=False):
         problem_type = self._infer_problem_type(y, silent=silent)
-        if self.quantile_levels is not None:
+        if problem_type == QUANTILE:
+            if self.quantile_levels is None:
+                raise AssertionError(f'problem_type is inferred to be {QUANTILE}, yet quantile_levels is not specified.')
+        elif self.quantile_levels is not None:
             if problem_type == REGRESSION:
                 problem_type = QUANTILE
             else:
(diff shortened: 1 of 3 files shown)


tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -3132,14 +3140,24 @@ def _set_feature_generator(self, feature_generator='auto', feature_metadata=None
...
             if num_bag_folds is None:
                 num_bag_folds = min(8, max(5, math.floor(num_train_rows / 100)))
+            # TODO: Leverage use_bag_holdout when data is large to enable multi-layer stacking
+            #  if num_train_rows >= 100000 and num_val_rows is None and use_bag_holdout is None:
+            #      use_bag_holdout = True
             if num_stack_levels is None:
-                num_stack_levels = min(1, max(0, math.floor(num_train_rows / 750)))
+                if problem_type == BINARY:
+                    # Disable multi-layer stacking to avoid stack info leakage
+                    num_stack_levels = 0
```
sxjscience, line 3155: Just ask, should we use `num_stack_levels=0` or `num_stack_levels=1`?

tabular/src/autogluon/tabular/learner/abstract_learner.py
```diff
@@ -681,9 +681,20 @@ def _remove_nan_label_rows(X, y):
...
 
+    def infer_problem_type(self, y: Series, silent=False):
+        problem_type = self._infer_problem_type(y, silent=silent)
+        if self.quantile_levels is not None:
+            if problem_type == REGRESSION:
+                problem_type = QUANTILE
+            else:
+                raise ValueError("autogluon infers this to be classification problem for which quantile_levels "
+                                 "cannot be specified. If it is truly a quantile regression problem, "
+                                 "please specify:problem_type='quantile'")
```
gradientsky, line 692: this would raise an error if the `problem_type == QUANTILE`

tabular/src/autogluon/tabular/predictor/predictor.py
Innixma: 0 = one-layer stacking (L1 models, L2 Weighted), 1 = multi-layer stack with L1 and L2 models (and L3 Weighted), 2 = L1,L2,L3 models + L4 Weighted, etc.

I know that it is a bit confusing, but I think this works, since 0 == a weighted combination of base models with no complex stacker models.

I also want to avoid changing cause it will be a breaking change.

tabular/src/autogluon/tabular/learner/abstract_learner.py
Innixma: It is impossible for it to be quantile because `self._infer_problem_type` never returns QUANTILE. The logic is unchanged from before, just moved around.

tabular/src/autogluon/tabular/predictor/predictor.py
sxjscience: Yeah, that makes sense.

tabular/src/autogluon/tabular/learner/abstract_learner.py
Innixma: addressed

gradientsky (approved):

PR #1919, opened 2022-07-01
Cleanup HPO artifacts

Innixma:
*Issue #, if available:*

*Description of changes:*
- Cleanup HPO artifacts
- Previously, train and val datasets were pickled to disk and not removed for every model that underwent HPO. This led to >20 copies of the train and val data cached to disk. With this PR, the dataset caches are deleted directly after HPO for a given model, ensuring only 1 copy is saved to disk at a time, and 0 copies are saved after training is completed.
- Previously when HPO was conducted with bagging enabled, the HPO artifact of the first fold model of each trial would be duplicated on disk and not deleted. Now the duplicate is deleted after HPO finishes for a given model.

### DIFF
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -996,7 +996,21 @@ class AbstractModel:
         scheduler.run()
         scheduler.join_jobs()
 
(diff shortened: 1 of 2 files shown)


szha:
Job PR-1919-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1919/1/index.html

yinweisu (approved):

PR #1920, opened 2022-07-01
Added caching y_pred_proba_val

Innixma:
*Issue #, if available:*

*Description of changes:*
- Added caching y_pred_proba_val in TabularPredictor.
- Previously, if using a holdout/validation set, each model's predictions of the val data would not be cached. This mean't at minimum that the model would have to predict on the val data (1) to compute val score, (2) during weighted ensembling, (3) during each higher multi-layer stacking call. In the typical case, this meant that we were predicting on val data twice. This caused issues for models that require a great deal of time to predict, causing us to go over time limit in some rare cases. This was a huge problem if `use_bag_holdout=True` and `num_stack_levels` was high, as every model trained prior would need to predict on the val data again for every new stack layer trained.
- This new logic guarantees that each model only needs to predict on the validation data once, not multiple times.

### DIFF
--- a/core/src/autogluon/core/trainer/abstract_trainer.py
+++ b/core/src/autogluon/core/trainer/abstract_trainer.py
@@ -9,6 +9,8 @@ import networkx as nx
 import numpy as np
 import pandas as pd
 import psutil
+import shutil
+from pathlib import Path
 
 from autogluon.common.features.feature_metadata import FeatureMetadata
 
@@ -109,6 +111,11 @@ class AbstractTrainer:
     @property
     def path_utils(self) -> str:
         return self.path_root + 'utils' + os.path.sep
+
+    @property
+    def _path_attr(self) -> str:
+        """Path to cached model graph attributes"""
+        return f'{self.path_utils}attr{os.path.sep}'
 
     @property
     def path_data(self) -> str:
@@ -472,7 +479,7 @@ class AbstractTrainer:
         logger.log(20, f'Fitting {len(models)} L{level} models ...')
         X_init = self.get_inputs_to_stacker(X, base_models=base_model_names, fit=True)
         if X_val is not None:
-            X_val = self.get_inputs_to_stacker(X_val, base_models=base_model_names, fit=False)
+            X_val = self.get_inputs_to_stacker(X_val, base_models=base_model_names, fit=False, use_val_cache=True)
         compute_score = not refit_full
         if refit_full and X_val is not None:
             X_init = pd.concat([X_init, X_val])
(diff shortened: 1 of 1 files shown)


szha:
Job PR-1920-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1920/1/index.html

core/src/autogluon/core/trainer/abstract_trainer.py
```diff
@@ -591,13 +605,17 @@ def get_model_pred_proba_dict(self, X, models, model_pred_proba_dict=None, model
...
             model_set = set()
             for model in models:
-                if model in model_set:
+                if model in model_set or model in model_pred_proba_dict:
                     continue
                 min_model_set = set(self.get_minimum_model_set(model))
                 model_set = model_set.union(min_model_set)
             model_set = model_set.difference(set(model_pred_proba_dict.keys()))
+            if use_val_cache:
+                model_set, model_pred_proba_dict = self._update_pred_proba_dict_with_val_cache(model_set=model_set, model_pred_proba_dict=model_pred_proba_dict)
```
yinweisu, line 618: Can you explain why we are calling `_update_pred_proba_dict_with_val_cache` both here and line 609?

Innixma: We don't know the full DAG of models we need to predict with yet at 609. 609 is used to filter the head of the DAG so we don't need to consider models that are redundant because the head has been pruned. In the case of a many-level stack, this reduces complexity from O(n) to O(1).

In case the heads aren't cached but some DAG members are, this is where 618 fetches these cases.

Essentially we call twice for the purposes of compute optimization.

yinweisu: makes sense

yinweisu (approved):
LGTM!

gidler:
Sorry I didn't bring this up before merge: The `if` blocks in lines 669-676 of [abstract_trainer.py](https://github.com/awslabs/autogluon/pull/1920/files#diff-22d0cf6acd4136b7a005b11d83b8110687c9d40cc2bbd78c2d72c2f697ab68c3) look overspecified to me. On line 669 we check if `y_pred_prob` is an instance of bool and if its truthy. Then on the next line we check if `y_pred_prob` is truthy again. I don't see how the `else` statement on 675 could ever be used since we already know `y_pred_prob` is truthy. I just wanted to make sure there wasn't a bug here. If `y_pred_prob` is `False`, it wouldn't get sent to `None` on line 676 which might have been the intention.

Innixma:
> Sorry I didn't bring this up before merge: The `if` blocks in lines 669-676 of [abstract_trainer.py](https://github.com/awslabs/autogluon/pull/1920/files#diff-22d0cf6acd4136b7a005b11d83b8110687c9d40cc2bbd78c2d72c2f697ab68c3) look overspecified to me. On line 669 we check if `y_pred_prob` is an instance of bool and if its truthy. Then on the next line we check if `y_pred_prob` is truthy again. I don't see how the `else` statement on 675 could ever be used since we already know `y_pred_prob` is truthy. I just wanted to make sure there wasn't a bug here. If `y_pred_prob` is `False`, it wouldn't get set to `None` on line 676 which might have been the intention.

Great catch! This is indeed a bug. While `False` is never a value set here at present, it would indeed crash if it was `False` for whatever reason. I fixed in #1924.

PR #1923, opened 2022-07-06
Add kwargs support to metrics

Innixma:
*Issue #, if available:*

*Description of changes:*
- Add kwargs support to metrics
- Now can specify values at call time such as `pos_label` for F1 metric.

### DIFF
--- a/core/src/autogluon/core/metrics/__init__.py
+++ b/core/src/autogluon/core/metrics/__init__.py
@@ -15,13 +15,78 @@ from . import quantile_metrics
 
 
 class Scorer(object, metaclass=ABCMeta):
-    def __init__(self, name, score_func, optimum, sign, kwargs):
+    """
+    Scorer wraps an external or custom metric function to align it with AutoGluon's metric logic and API.
+
+    Parameters
+    ----------
+    name : str
+        Name of the metric. Used in logs and as a key in dictionaries when multiple metric outputs are computed.
+    score_func :
+        Scoring metric function that will be called internally to score.
+        Required to be a callable with the first two arguments corresponding to y_true, y_pred that returns a float indicating the metric value.
+    optimum : float
+        The highest/best value the metric can return. For example, optimal=1 for accuracy, optimal=0 for mean_squared_error.
+        This is used to calculate regret / error. For example, a score of 1 for accuracy would have an error of 0.
+    sign : int
+        Valid values are 1 and -1.
+        The sign of the metric to ensure greater_is_better.
+        For score metrics such as accuracy and r2, the sign should be 1.
+        For error metrics such as log_loss and mean_squared_error, the sign should be -1.
+    kwargs : dict, optional
+        kwargs to pass to score_func when called.
+        For example, kwargs = {"beta": 2} when using sklearn.metrics.fbeta_score where beta is a required argument.
+    """
+    def __init__(self, name: str, score_func, optimum: float, sign: int, kwargs: dict = None):
         self.name = name
+        if kwargs is None:
+            kwargs = dict()
         self._kwargs = kwargs
         self._score_func = score_func
         self._optimum = optimum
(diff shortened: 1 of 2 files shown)


gradientsky (approved):

core/src/autogluon/core/metrics/__init__.py
```diff
@@ -167,9 +175,12 @@ def __call__(self, y_true, y_pred, sample_weight=None):
...
             return self._sign * self._score_func(y_true, y_pred,
                                                  sample_weight=sample_weight,
-                                                 **self._kwargs)
+                                                 **self._kwargs,
+                                                 **kwargs)
         else:
-            return self._sign * self._score_func(y_true, y_pred, **self._kwargs)
+            return self._sign * self._score_func(y_true, y_pred,
+                                                 **self._kwargs,
+                                                 **kwargs)
```
sxjscience, line 183: This might be problematic. For example, the keys in `kwargs` may duplicate the keys in `self._kwargs`. Thus, we may consider to merge these two dictionary and then feed into the `_score_func`. For example:

```python
kwargs.update(self._kwargs)
self._score_func(y_true, y_pred, **self._kwargs)
```

core/src/autogluon/core/metrics/__init__.py
```diff
@@ -374,7 +396,13 @@ def make_scorer(name, score_func, optimum=1, greater_is_better=True,
         cls = _QuantileScorer
     else:
         cls = _PredictScorer
-    return cls(name, score_func, optimum, sign, kwargs)
+    return cls(
+        name=name,
+        score_func=score_func,
+        optimum=optimum,
+        sign=sign,
+        kwargs=kwargs,
```
sxjscience, line 404: Should we use `sign=sign, **kwargs` ?

Innixma: I actually updated the code just now so this works better and is fully generic: For example we don't want to fail when the metric needs `optimum` as a kwarg.

core/src/autogluon/core/metrics/__init__.py
Innixma: It is unclear if this is what we want. It seems strange the user would specify a specific value in construction and then pass a different one at scoring time. We could definitely make it work, but it might be better to have it crash as it currently does. Any thoughts on if there is a valid use-case?

core/src/autogluon/core/metrics/__init__.py
Innixma: You can refer to the new unit test to see how it works

sxjscience: I see, this is the cached kwargs from the Scorer.

core/src/autogluon/core/metrics/__init__.py
sxjscience: I think users may want to overwrite some predefined kwargs.

Innixma: The code you provided does the opposite, it overrides the user kwargs with the predefined kwargs:

```
kwargs.update(self._kwargs)
self._score_func(y_true, y_pred, **kwargs)
```

Did you instead mean:

```
k = copy.deepcopy(self._kwargs)
k.update(kwargs)
self._score_func(y_true, y_pred, **k)
```

sxjscience: Yeah, the 2nd version should be better.

szha:
Job PR-1923-1 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1923/1/index.html

core/src/autogluon/core/metrics/__init__.py
Innixma: updated the PR to add support for the 2nd version + simplify the code and add documentation.

PR #1924, opened 2022-07-06
Fix y_pred_proba lookup when cached_y_pred_proba_val == False

Innixma:
*Issue #, if available:*
#1920 

*Description of changes:*
- Fix bug in #1920 causing a crash if cached_y_pred_proba_val == False (Never the case, but could happen in future).


gidler (approved):
LGTM

yinweisu (approved):
LGTM!

PR #1933, opened 2022-07-10
Support gradient checkpointing for finetuning billion-scale models on a single GPU + Loose the version bound of torch to be <=1.12

sxjscience:
*Description of changes:*
After the change, we are able to finetune the [mT5-XL](https://huggingface.co/google/mt5-xl) model that has **1.7B** parameters on a single NVIDIA G4 GPU (In AutoMM, we only use the T5-encoder like in [Sentence-T5](https://aclanthology.org/2022.findings-acl.146.pdf) so it's 1.7B.). Thus, users can now use billion-scale models like mT5-XL on **a single g4.2x instance** with AutoGluon Mulimodal.

To run the feature, use the example script: https://gist.github.com/sxjscience/b011460574c7e41c47a42ba86d026cbc

```
Auto select gpus: [0]
GPU available: True, used: True
TPU available: False, using: 0 TPU cores
IPU available: False, using: 0 IPUs
HPU available: False, using: 0 HPUs
LOCAL_RANK: 0 - CUDA_VISIBLE_DEVICES: [0,1,2,3]

  | Name              | Type                         | Params
-------------------------------------------------------------------
0 | model             | HFAutoModelForTextPrediction | 1.7 B 
1 | validation_metric | AUROC                        | 0     
2 | loss_func         | CrossEntropyLoss             | 0     
-------------------------------------------------------------------
3.2 M     Trainable params
1.7 B     Non-trainable params
1.7 B     Total params
6,692.434 Total estimated model params size (MB)

```

In addition, for training the model on multiple GPUs, we can just use the `fsdp` strategy supported in pytorch lightning:
https://gist.github.com/sxjscience/a86fc9c8e49b10af9495bc30322c3c26

- [x] Support gradient checkpointing: https://pytorch.org/docs/stable/checkpoint.html. If it is turned on, T5 in Huggingface seems to have some issues related to https://discuss.pytorch.org/t/checkpoint-with-no-grad-requiring-inputs-problem/19117/9. We fixed the issue by adding a DummyLayer.
- [x] Fix LoRA. The previous version has problems when used together gradient checkpointing. Also, we revised patterns to support regex.
- [x] Add new "pooling_mode" flag in `model.hf_text`. It now supports "cls" or "mean"
- [x] Turn on `track_grad_norm` in Trainer
- [x] New problem_type "classification"
- [x] Performance verification (since we are just using encoder, it may have worse performance in some benchmarking datasets, but the result is reasonable).

@Raldir @FANGAreNotGnu @Linuxdex @bryanyzhu @zhiqiangdon @gradientsky


docs/install-cpu-source.rst
```diff
@@ -5,7 +5,7 @@
 
     # CPU version of pytorch has smaller footprint - see installation instructions in
     # pytorch documentation - https://pytorch.org/get-started/locally/
-    pip3 install "torch>=1.0,<1.12+cpu" -f https://download.pytorch.org/whl/cpu/torch_stable.html
+    pip3 install "torch>=1.0,<1.13+cpu" -f https://download.pytorch.org/whl/cpu/torch_stable.html
```
sxjscience, line 8: @Innixma @yinweisu I revised our installation guide so that users can install torch 1.12

Innixma: Are we still claiming support for torch 1.0? Should we tighten bounds?

sxjscience: How about just recommending the latest torch in our tutorial? I think we should work for torch>=1.9.

wenming:
really nice, let's prioritize reviewing this.

docs/install-cpu-source.rst
sxjscience: I checked the requirement from torchmetrics, it is currently asking for torch 1.11 in the stable version. Thus, we may also recommend using torch 1.11 in our website for now and upgrade to torch 1.12 once that is stable. Meanwhile, I think we can turn on torch 1.12 in our unittests.

multimodal/src/autogluon/multimodal/data/process_text.py
```diff
@@ -243,13 +244,23 @@ def build_one_token_sequence(
             max_length = self.max_len - (len(text_tokens) + 1)
         else:
             max_length = self.max_len - 2
+        if self.eos_only:
+            # For EOS-only, the tokens will be combined as
+            # [Field1 Tokens] [EOS] [Field2 Tokens] [EOS] [Field3 Tokens] [EOS]
+            # Otherwise, the tokens will be combined as
+            # [CLS] [Field1 Tokens] [SEP] [Field2 Tokens] [SEP] [Field3 Tokens] [EOS]
+            max_length += 1
```
zhiqiangdon, line 252: Why do we need to increase the max length in the `eos_only` mode?

sxjscience: The number of special tokens will be reduced by 1, so the max length we can allocate for the text field tokens will be increased by 1.

multimodal/tests/unittests/test_backward_compatibility.py
```diff
@@ -6,7 +6,7 @@
 
 from autogluon.multimodal import MultiModalPredictor, AutoMMPredictor
 from autogluon.multimodal.utils import download
-from datasets import AmazonReviewSentimentCrossLingualDataset
+from unittest_datasets import AmazonReviewSentimentCrossLingualDataset
```
zhiqiangdon, line 9: Does the name `datasets` cause confusion? Since it is can only be used in in this unit tests folder, I feel the previous name should be fine.

multimodal/src/autogluon/multimodal/configs/optimization/adamw.yaml
```diff
@@ -18,13 +18,17 @@ optimization:
...
   lora:
     r: 8
     alpha: 8
     filter: # Default fine-tune only query and value attention weights, recommended in https://arxiv.org/abs/2106.09685
       - "query"
       - "value"
+      - "^q$"  # The filter can also be a regex. We will use re.match(filter, name) to determine if LoRA should be applied.
+      - "^v$"
+      - "^k$"
+      - "^o$"
```
zhiqiangdon, line 31: Any examples how to use them?

multimodal/src/autogluon/multimodal/models/huggingface_text.py
```diff
@@ -28,6 +28,15 @@
 logger = logging.getLogger(AUTOMM)
 
 
+class DummyLayer(nn.Module):
```
zhiqiangdon, line 31: Better to move `DummyLayer` into utils.py. Need to keep this file clean.

multimodal/src/autogluon/multimodal/models/huggingface_text.py
```diff
@@ -28,6 +28,15 @@
...
+    def __init__(self):
+        super().__init__()
+        self.dummy_bias = nn.Parameter(torch.ones(1, dtype=torch.float32))
```
zhiqiangdon, line 34: Do we need to specify `dtype=torch.float32`? I think lightning can automatically handle this.

multimodal/tests/unittests/test_backward_compatibility.py
sxjscience: Yes, it conflicts with the datasets packagez

multimodal/tests/unittests/test_predictor_advanced.py
```diff
@@ -0,0 +1,62 @@
...
+                      "optimization.efficient_finetune": efficient_finetuning,
+                      "optimization.lr_decay": 1.0,
+                      "optimization.learning_rate": 1e-03,
+                      "optimization.max_epochs": 1,
+                      "env.per_gpu_batch_size": 2,
+                      "env.num_workers": 0,
+                      "env.num_workers_evaluation": 0,
+                      "env.num_gpus": 1,
+                  })
+    predictions = predictor.predict(test_data)
```
zhiqiangdon, line 59: Do we need to test `verify_predictor_save_load()`?

multimodal/src/autogluon/multimodal/models/huggingface_text.py
```diff
@@ -61,18 +72,37 @@ def __init__(
...
+            # Remove the decoder in T5
+            del self.model.decoder
+        else:
+            self.is_t5 = False
+
+        self.gradient_checkpointing = gradient_checkpointing
+        if gradient_checkpointing:
+            self.model.gradient_checkpointing_enable()
+            if self.is_t5:
+                self.dummy_layer = DummyLayer()
```
zhiqiangdon, line 97: Is this gradient checkpointing issue unique to the `t5` model? If it also exist in other models, the dummy layer shouldn't condition only on `t5`.

multimodal/src/autogluon/multimodal/models/huggingface_text.py
sxjscience: We can consider to refactor it later. This is  a issue I met when finetuning T5 models but haven’t seen such issue in other backbones. We may remove the hack later.

multimodal/src/autogluon/multimodal/models/huggingface_text.py
```diff
@@ -61,18 +72,37 @@ def __init__(
...
         self.checkpoint_name = checkpoint_name
         self.num_classes = num_classes
+
         self.model = AutoModel.from_pretrained(checkpoint_name)
+        if isinstance(self.model, T5PreTrainedModel):
+            self.is_t5 = True
+            # Remove the decoder in T5
+            del self.model.decoder
+        else:
+            self.is_t5 = False
```
zhiqiangdon, line 91: Could we wrap this into a util function to detect special models? I guess `t5` is just one of them we need to deal with specially. The return of this util function is the model type.

multimodal/src/autogluon/multimodal/data/process_text.py
zhiqiangdon: Got it.

multimodal/src/autogluon/multimodal/models/huggingface_text.py
sxjscience: Lightning will set the dtype/device of the input to forward. For parameters, their dtype will be converted if we call functions like mod.half().

multimodal/src/autogluon/multimodal/models/huggingface_text.py
sxjscience: Yes, but we may do it later. Encoder-decoder models like  BART should also be specially treated.

sxjscience: In fact, we may need to figure out the way for supporting applying the decoder in model training.

multimodal/src/autogluon/multimodal/configs/optimization/adamw.yaml
sxjscience: Basically some layers may be “xxx.q.yyy”. So it’s matching the “q” element here.

sxjscience:
I'm still running some benchmarks and will share results afterwards.

bryanyzhu (commented):
Great feature. Left a few minor comments.

multimodal/src/autogluon/multimodal/configs/environment/default.yaml
```diff
@@ -11,4 +11,5 @@ env:
   fast_dev_run: False
   deterministic: False
   auto_select_gpus: True
+  track_grad_norm: 2  # Whether to check gradient norm. We can set it to 2 to check for gradient norm.
```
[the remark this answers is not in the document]
sxjscience: Yes, this is a newly added feature. It will track the gradient norms in Tensorboard. It helped me debug the LoRA + gradient checkpointing scenario.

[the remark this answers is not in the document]
sxjscience: track_grad_norm: https://pytorch-lightning.readthedocs.io/en/stable/common/debugging.html#inspect-gradient-norms

multimodal/src/autogluon/multimodal/models/huggingface_text.py
zhiqiangdon: Maybe we need a config `uses_decoder`.

zhiqiangdon: I find that people will tend to follow previous design to add new stuff. By following this design, people will add more if-else patterns here.

multimodal/src/autogluon/multimodal/models/huggingface_text.py
zhiqiangdon: We don't when Pytorch will fix this. Moving it to utils is still better. This may not be a good example for future developers to follow.

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -1746,10 +1751,15 @@ def save(self, path: str, standalone: Optional[bool] = False):
             model_path = os.path.join(self._save_path, "model.ckpt")
             if os.path.isfile(model_path):
                 shutil.copy(model_path, path)
+            else:
+                # FIXME(?) Fix the saving logic
+                RuntimeError(f"Cannot find the model checkpoint in '{model_path}'. Have you removed the folder that "
```
sxjscience, line 1756: @zhiqiangdon

multimodal/src/autogluon/multimodal/data/infer_types.py
```diff
@@ -51,6 +62,7 @@ def is_categorical_column(
     else:
         if threshold is None:
             if is_label:
+                # TODO(?) The following logic will be problematic if the task is few-shot learning.
```
sxjscience, line 65: @Linuxdex @Raldir Will be related to few-shot learning. We should fix the logic in a future PR.

multimodal/src/autogluon/multimodal/models/huggingface_text.py
sxjscience: If I do not add the DummyLayer, finetuning mT5-XL on MRPC with the implementation can only get 0.70 ACC. As a reference, adding the dummy layer can get 0.872 ACC.

```
# Without Dummy
{'acc': 0.7009803921568627, 'f1': 0.8195266272189349, 'time_cost': 4389.08921289444}` 

# With Dummy
{"acc": 0.8725490196078431, "f1": 0.9047619047619048, "time_cost": 8712.896693944931}
```

Also, we need to wrap dummy layer after the embedding layer so it is not generally applicable.

multimodal/tests/unittests/test_predictor_advanced.py
sxjscience: Tested.

sxjscience:
I think the performance is reasonable.

bryanyzhu (approved):

zhiqiangdon (approved):

PR #1936, opened 2022-07-11
Improve column type detection by supporting fallback type

zhiqiangdon:
1. Get data types based on model types.
2. Filter auto-detected columns types to make sure all types are allowed. Replace the invalid types with a default fallback type.
3. Use the CLIP large@336 in zero-shot learning by default.

### DIFF
--- a/multimodal/src/autogluon/multimodal/data/infer_types.py
+++ b/multimodal/src/autogluon/multimodal/data/infer_types.py
@@ -404,11 +404,7 @@ def infer_problem_type_output_shape(
             )
 
 
-def set_fallback_column_type(
-    column_types: Dict,
-    allowable_column_types: List[str],
-    fallback_column_type: str
-) -> Dict:
+def set_fallback_column_type(column_types: Dict, allowable_column_types: List[str], fallback_column_type: str) -> Dict:
     """
     Filter the auto-detected column types to make sure that all column types are allowable.
     Use the fallback type to replace those out of the allowable_column_types.
--- a/multimodal/src/autogluon/multimodal/data/infer_types.py
+++ b/multimodal/src/autogluon/multimodal/data/infer_types.py
@@ -4,7 +4,7 @@ import pandas as pd
 import warnings
 import PIL
 from typing import Union, Optional, List, Dict, Tuple
-from ..constants import NULL, CATEGORICAL, NUMERICAL, TEXT, IMAGE_PATH, MULTICLASS, BINARY, REGRESSION, AUTOMM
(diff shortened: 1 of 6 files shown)


multimodal/src/autogluon/multimodal/data/infer_types.py
```diff
@@ -250,12 +256,19 @@ def infer_column_types(
         elif is_numerical_column(data[col_name], valid_data[col_name]):  # Infer numerical column
             column_types[col_name] = NUMERICAL
         elif is_imagepath_column(data[col_name], col_name):  # Infer image-path column
-            column_types[col_name] = IMAGE_PATH
+            column_types[col_name] = IMAGE
```
bryanyzhu, line 259: What is the difference between IMAGE and IMAGE_PATH?

multimodal/src/autogluon/multimodal/data/infer_types.py
```diff
@@ -389,3 +402,28 @@ def infer_problem_type_output_shape(
                 f"The label column '{label_column}' has type"
                 f" '{column_types[label_column]}', which is not supported yet."
             )
+
+
+def set_fallback_column_type(column_types: Dict, allowable_column_types: List[str], fallback_column_type: str) -> Dict:
```
bryanyzhu, line 407: How to use this function? In particular, how to set `fallback_column_type` based on user's input?

multimodal/src/autogluon/multimodal/data/infer_types.py
zhiqiangdon: They are the same currently, meaning image type. Since we use `image` in the `data_types` and we want to it in the filtering, it's better to change to `image` here. We still keep the `image_path` in the dataframe preprocessor for backward compatibility.

multimodal/src/autogluon/multimodal/data/infer_types.py
zhiqiangdon: See the new util function `infer_dtypes_by_model_names`. The `allowable_column_types ` can be obtained from config. The `fallback_column_type ` is some prior knowledge we can set based on user cases. Here we have the case of `image` and `text`. Since `image` is easy to detect, we can change other detected types all to `text`. We can add more rules in `infer_dtypes_by_model_names`.

szha:
Job PR-1936-2 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1936/2/index.html

sxjscience (approved):

bryanyzhu (approved):

szha:
Job PR-1936-3 is done. 
Docs are uploaded to http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-1936/3/index.html

Issue #1937, opened 2022-07-12
[BUG] Ray is no longer an optional dependency


Innixma:
- [ ] I have checked that this bug exists on the latest stable version of AutoGluon
- [x] and/or I have checked that this bug exists on the latest mainline of AutoGluon via source installation

**Describe the bug**
AutoGluon v0.5.1 pre-release imports ray on import due to a recent commit (#1915). This crashes MacOS on import because ray is not a required dependency, and MacOS does not install ray by default.

**Expected behavior**
AutoGluon should be able to run successfully without ray installed.

**To Reproduce**
On MacOS, install from source, and then:

```
import autogluon.tabular
```

Error:

<details>

```
Traceback (most recent call last):
  File "/Users/neerick/workspace/virtual/autogluon/lib/python3.8/site-packages/IPython/core/interactiveshell.py", line 3457, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-2-bd2a278b3f61>", line 1, in <module>
    runfile('/Users/neerick/workspace/code/autogluon-scratch/scripts/run_adult.py', wdir='/Users/neerick/workspace/code/autogluon-scratch/scripts')
  File "/Users/neerick/Library/Application Support/JetBrains/Toolbox/apps/PyCharm-P/ch-0/213.6777.50/PyCharm.app/Contents/plugins/python/helpers/pydev/_pydev_bundle/pydev_umd.py", line 198, in runfile
    pydev_imports.execfile(filename, global_vars, local_vars)  # execute the script
  File "/Users/neerick/Library/Application Support/JetBrains/Toolbox/apps/PyCharm-P/ch-0/213.6777.50/PyCharm.app/Contents/plugins/python/helpers/pydev/_pydev_imps/_pydev_execfile.py", line 18, in execfile
    exec(compile(contents+"\n", file, 'exec'), glob, loc)
  File "/Users/neerick/workspace/code/autogluon-scratch/scripts/run_adult.py", line 1, in <module>
    from autogluon.tabular import TabularPredictor, TabularDataset
  File "/Users/neerick/Library/Application Support/JetBrains/Toolbox/apps/PyCharm-P/ch-0/213.6777.50/PyCharm.app/Contents/plugins/python/helpers/pydev/_pydev_bundle/pydev_import_hook.py", line 21, in do_import
    module = self._system_import(name, *args, **kwargs)
[...]

PR #1938, opened 2022-07-12
Fix optional ray import

yinweisu:
*Issue #, if available:*
https://github.com/awslabs/autogluon/issues/1937

*Description of changes:*
* Removed import of modules inside hpo init
* Renamed multiple files to separate out ray dependent components. Now ray will only be directly imported within those files.


Innixma (commented):
Thanks for the quick fix! Will wait for unit tests to pass before merging.

Innixma (approved):
LGTM!

PR #1939, opened 2022-07-12
Improve RAPIDS support (require GPU, fix usage of GPU in parallel)

Innixma:
*Issue #, if available:*
#1845 

*Description of changes:*
- Improve RAPIDS support (require GPU, fix usage of GPU in parallel)
- Clean up RAPIDS code by adding RapidsModelMixin

### DIFF
--- a/tabular/src/autogluon/tabular/models/_utils/rapids_utils.py
+++ b/tabular/src/autogluon/tabular/models/_utils/rapids_utils.py
@@ -1,0 +1,28 @@
+from typing import Dict
+
+from autogluon.core.utils import get_gpu_count_torch
+
+
+class RapidsModelMixin:
+    """Mixin class for methods re-used across RAPIDS models"""
+    # FIXME: Efficient OOF doesn't work in RAPIDS
+    @classmethod
+    def _get_default_ag_args_ensemble(cls, **kwargs) -> dict:
+        default_ag_args_ensemble = super()._get_default_ag_args_ensemble(**kwargs)
+        extra_ag_args_ensemble = {'use_child_oof': False}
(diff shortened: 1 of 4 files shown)


tabular/src/autogluon/tabular/models/_utils/rapids_utils.py
```diff
@@ -0,0 +1,28 @@
+from typing import Dict
+
+from autogluon.core.utils import get_gpu_count_torch
+
+
+class RapidsModelMixin:
```
yinweisu, line 6: Is this model supposed to inherit `AbstractModel`? Seen couple super() calls, but this model is not inheriting any class.

Innixma: I don't believe Mixin's are supposed to inherit. @gradientsky thoughts?

gradientsky: Yes, they not supposed to be inherit `AbstractModel`, it's an additional functionality mixed-into the main class via multiple inheritance. One thing: I see some versions implement `ABC`: https://docs.python.org/3/library/abc.html#abc.ABCMeta. Not sure if that's really necessary though.

gradientsky (approved):

tabular/src/autogluon/tabular/models/_utils/rapids_utils.py
Innixma: I don't think ABC is necessary, I tested with a dummy model using inheritance and it worked fine (tbh I'm unsure what ABC does and when it is necessary).

PR #1945, opened 2022-07-12
Fix nightly platform test

yinweisu:
*Issue #, if available:*

*Description of changes:*
* Github action started to install libomp by default, previous script didn't remove the existing installation. So the libomp 11.1 wasn't correctly installed and caused segfault.
* Set fail fast to be false, so that different OS won't be affected by failure status of each other.

Tested on the fork:
https://github.com/yinweisu/autogluon/actions/runs/2659133804


Innixma (approved):
LGTM, thanks for the fix!

PR #1951, opened 2022-07-13
Fix RayExecutor Interface

yinweisu:
*Issue #, if available:*

*Description of changes:*
* Fixed the inconsistency of `register_resources` interface.
* Updated ray hpo interface to use GPU calculator when the trial requires gpu not when the machine has gpu.

### DIFF
--- a/core/src/autogluon/core/hpo/executors.py
+++ b/core/src/autogluon/core/hpo/executors.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import copy
 import logging
 import os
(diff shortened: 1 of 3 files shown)


core/src/autogluon/core/hpo/executors.py
```diff
@@ -60,17 +61,29 @@ def initialize(
...
-    def register_resources(self, resources: dict):
+    def register_resources(self, initialized_model: "AbstractModel"):
         """
         Register total resources used for the experiment
         
         Parameters
         ----------
-        resources
-            Total resources available for the experiment
+        model
```
gradientsky, line 70: `initialized_model`, not `model`

core/src/autogluon/core/hpo/executors.py
```diff
@@ -60,17 +61,29 @@ def initialize(
         """
         raise NotImplementedError
     
-    @abstractmethod
-    def register_resources(self, resources: dict):
+    def register_resources(self, initialized_model: "AbstractModel"):
```
gradientsky, line 64: I suggest using [PEP 563: Postponed Evaluation of Annotations](https://docs.python.org/3.7/whatsnew/3.7.html#pep-563-postponed-evaluation-of-annotations) instead of string.

gradientsky (approved):

PR #1952, opened 2022-07-13
Unit Test for pinball_loss method for quantile regression

gidler:
*Description of changes:*

- Unit tests to check AutoGluon's `pinball_loss` method which is used in quantile regression

- I added test cases for the various exceptions the method can raise, and I performed the operation by hand and verified the code produced the same result. I also added cases for optional kwargs.

- I'd appreciate feedback on if these changes match our expectations for UTs because I have more tests coming.


Innixma (approved):
Looks great, thanks for the contribution!

PR #1953, opened 2022-07-13
Update configs and add distiller customization docs

zhiqiangdon:
1. Move `track_grad_norm` from environment into optimization and change its default to -1.
2. Use clip base in zero shot unit tests.
3. Update the customization doc by adding distiller configs.
4. Move one util from optimization/utils.py into multimodal/utils.py (a semantically better location for this function).


bryanyzhu (approved):
LGTM

sxjscience (approved):
LGTM

PR #1975, opened 2022-07-21
Support timm zero shot learning

zhiqiangdon:
1. Use timm to create head to cover the case that, if the provided `num_classes` is the same as the pre-trained `num_classes`, we can reuse the pre-trained head.
2. Move the timm head outside since we need to get the pooled features for late fusion sometimes.
3. Output logits only if the `num_classes` > 0.
4. Add a unit test for timm zero shot learning.

### DIFF
--- a/multimodal/src/autogluon/multimodal/models/timm_image.py
+++ b/multimodal/src/autogluon/multimodal/models/timm_image.py
@@ -65,7 +65,12 @@ class TimmAutoModelForImagePrediction(nn.Module):
         self.model = create_model(checkpoint_name, pretrained=pretrained, num_classes=num_classes)
         self.num_classes = self.model.num_classes
         self.out_features = self.model.num_features
-        self.head = self.model.head  # move the head outside
+        if hasattr(self.model, "head"):
+            self.head = self.model.head  # move the head outside
+        elif hasattr(self.model, "last_linear"):
+            self.head = self.model.last_linear
+        else:
+            raise ValueError(f"Model {checkpoint_name} doesn't have head. Need to check its TIMM implementation.")
         self.model.reset_classifier(0)  # remove the internal head
 
         self.mix_choice = mix_choice
--- a/multimodal/src/autogluon/multimodal/models/timm_image.py
+++ b/multimodal/src/autogluon/multimodal/models/timm_image.py
@@ -33,7 +33,7 @@ class TimmAutoModelForImagePrediction(nn.Module):
         self,
         prefix: str,
         checkpoint_name: str,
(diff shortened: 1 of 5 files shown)


multimodal/src/autogluon/multimodal/models/timm_image.py
```diff
@@ -114,7 +115,9 @@ def forward(
         image_valid_num = batch[self.image_valid_num_key]
         ret = {COLUMN_FEATURES: {FEATURES: {}, MASKS: {}}}
         if self.mix_choice == "all_images":  # mix inputs
-            mixed_images = images.sum(dim=1) / image_valid_num[:, None, None, None]  # mixed shape: (b, 3, h, w)
+            mixed_images = (
```
bryanyzhu, line 118: I don't understand this part. Can you explain a bit? Thank you

multimodal/tests/unittests/test_zero_shot.py
```diff
@@ -98,3 +103,36 @@ def test_clip_zero_shot():
     # invalid API usage 2: predicting probability with only one dictionary as input.
     with pytest.raises(AssertionError):
         prob = predictor.predict_proba({"image": [cat_image_name], "text": [cat_text]})
+
+
+def test_timm_zero_shot():
+    cat_image_name, dog_image_name = download_sample_images()
+
+    predictor = MultiModalPredictor(
+        hyperparameters={
```
bryanyzhu, line 112: Can we test some CNN models as well? (maybe at least a ResNet50) CNN has some structural differences than transformers, that may break current code logic.

bryanyzhu (changes requested):
LGTM, left some comments. It would be great to add some unittest for CNN models in TIMM.

multimodal/src/autogluon/multimodal/models/timm_image.py
zhiqiangdon: Each row in a dataframe may have multiple images. `all_images` means mixing the images into one before the model forward. If no valid images exist in a row, the `image_valid_num` is 0. To avoid the divide-by-zero error, we clamp the min value to 1e-6.

multimodal/tests/unittests/test_zero_shot.py
zhiqiangdon: Thanks. Added CNN tests.

multimodal/src/autogluon/multimodal/models/timm_image.py
bryanyzhu: Thanks

bryanyzhu (approved):

Issue #1794, opened 2022-06-09
[BUG] Error when calculation feature importance. (Quantile model)


easyautoml:
When model is builded with problem_type='quantile', I got bug when calculation feature_important.

`predictor.feature_importance(model='RandomForestMSE', data=train_data, feature_stage='transformed_model')`

**Error :**
![image](https://user-images.githubusercontent.com/100541539/172775913-df945da4-3ce3-4bd2-9f4c-2396656748a1.png)

I found that eval_metric function at bellow file required quantile_levels as paramecter, but only y and y_pred_cur given. 
`autogluon\core\src\autogluon\core\utils\utils.py`

**Installed Versions**
autogluon 0.4.0

easyautoml:
The problem is bellow code. 

https://github.com/awslabs/autogluon/blob/3f21e7a3240754bc88ac3e42a7e5c7aee5b2b495/core/src/autogluon/core/utils/utils.py#L712

https://github.com/awslabs/autogluon/blob/3f21e7a3240754bc88ac3e42a7e5c7aee5b2b495/core/src/autogluon/core/utils/utils.py#L765

We need give eval_metric quantile_levels as parameter.

leloykun:
@Innixma @easyautoml, I added a fix here: https://github.com/awslabs/autogluon/pull/1977

Innixma:
Thanks for the fix @leloykun! Closing with merge of #1977.

PR #1977, opened 2022-07-21
Fix calculation of feature importances on Quantile Models

leloykun:
*Issue #, if available:* https://github.com/awslabs/autogluon/issues/1794

*Description of changes:*
- Just passes `quantile_levels` to `compute_permutation_feature_importance` like how we do it in `AbstractTrainer.score`


Innixma (approved):
LGTM! Thanks for the contribution. Will wait for CI to pass and then merge

Issue #1969, opened 2022-07-19
[Feature Request] API tweaks


mli:
A few minor API tweaks, which I believe to improve usability. I'd like to add it to the 0.6 milestone, but not necessary to address all by the release. I'll probably add more in future

## Tabular

- [ ] `evaluate` accepts `auxiliary_metrics=True/False` while `leaderboard` has a `extra_metrics` to accept a list of metrics. it's good to unify them, e.g. both support these two arguments, or just keep one

- [x] `silent=False` in default for `evaluate` and `leaderboard`. i suggest to change it to true, because the two common cases are 1/ `res = predict.evaluate(...)` or run `predict.evaluate(...)` in jupyter notebook to print the results. Now we need to always set `silent=True` to avoid double printing. 

- [x] metrics are adopted from scikit-learn, but removed the `_score` and `_loss` postfix for most metrics for simplicity. there are still some of them such as `pac_score`, we could rename it to `pac`.  

- [x] when using `hyperparameter_tune_kwargs` in `fit`, we must schedule `scheduler` and `searcher`. we can set default values `local` and `auto`.

- [ ] `fit` accepts `**kwargs`,  it's better to explicitly write down what kwargs it accepts 

In general, our methods have a long list of arguments. Comparing to others such as scikit-learn or pytorch, our API doc is harder to read. Maybe we should simplify the API doc, or organize them into sections. 

## MultiModal  

One main issue I found is multimodal's API is different than tabular in several places, e.g. argument names and orders. They can be interchangeably used in several tasks, as they all accept dataframe as inputs. I imagine that users may just replace `TabularPredictor` to `MultiModalPredictor` during prototyping. (you can use the `multimodal` option in tabular, but using MultiModalPredictor alone often runs faster.) So making their API similar to each others, at least for basic usages, will benefit users.

- [ ] documents call dataset from other modules, such as `vision.ImageDataset`. We can create alias in the multimodal as well. 

  - [x] when you installed both pytorch and mxnet, gluoncv will warn you about the memory usage. i feel we should fix its issue instead of warning, e.g. only import pytorch even when mxnet exists

- [ ] `default_metric` is auc for binary classification, with a warning that it will be slow. while not use accuracy in default for all classification ? ( tabular uses accuracy.) 

- [ ] `evaluate` has a `metrics`, which is insistent with tabular.  

[...]

PR #1980, opened 2022-07-22
Improve AutoMM's save_path logic

zhiqiangdon:
*Issue #1969 *

1. Support generating random paths in continuous training if `save_path` is not specified.
2. Avoid overwriting if old paths are detected.
3. Add unit tests.

### DIFF
--- a/multimodal/src/autogluon/multimodal/utils.py
+++ b/multimodal/src/autogluon/multimodal/utils.py
@@ -85,6 +85,7 @@ from .constants import (
     FEATURES,
     MASKS,
     S3_PREFIX,
+    LAST_CHECKPOINT,
 )
 from .presets import get_automm_presets, get_basic_automm_config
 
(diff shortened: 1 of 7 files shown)


multimodal/src/autogluon/multimodal/utils.py
```diff
@@ -1976,3 +1977,37 @@ def update_config_by_rules(
                 UserWarning,
             )
     return config
+
+
+def process_save_path(path, resume: Optional[bool] = False, raise_if_exist: Optional[bool] = True):
+    """
+    Convert the provided path to an absolute path and check whether it is valid.
+    If a path exists, either raise error or return None.
+    A None path can be identified by the `setup_outputdir` to generate a random path.
```
bryanyzhu, line 1986: What does the random path look like?

multimodal/tests/unittests/test_save_path.py
```diff
@@ -0,0 +1,69 @@
...
+        problem_type=dataset.problem_type,
+        eval_metric=dataset.metric,
+    )
+    with pytest.raises(ValueError):
+        predictor.fit(train_data=dataset.train_df, save_path=save_path)
+
+    shutil.rmtree(abs_path)
+
+
+def test_continuous_training_save_path():
```
bryanyzhu, line 35: Can we have multiple functions to test different use cases of continuous training? Like (1) one predictor, but continue to train on new data (2) load predictor and resume training. And also, is it possible to load optimizer states beyond just model weights? Thanks.

multimodal/src/autogluon/multimodal/utils.py
zhiqiangdon: Something like f"AutogluonModels/ag-{timestamp}"

multimodal/tests/unittests/test_save_path.py
zhiqiangdon: These two cases are covered in this test (Lines 62 and 66). Continuous training is different from resuming training. Resuming training means the previous `fit()` doesn't finish and we resume from `last.ckpt`. Continuous training means the previous `fit()` finish successfully and a model checkpoint `model.ckpt` is saved. Thus, continuous training only loads the model weights without optimizer states.

multimodal/src/autogluon/multimodal/utils.py
```diff
@@ -1976,3 +1977,37 @@ def update_config_by_rules(
...
+        Whether to raise error if the path exists.
+
+    Returns
+    -------
+    A complete and verified path or None.
+    """
+    path = os.path.abspath(os.path.expanduser(path))
+    if resume:
+        last_ckpt_path = os.path.join(path, LAST_CHECKPOINT)
+        assert os.path.isfile(last_ckpt_path), f"The last checkpoint {last_ckpt_path} doesn't exist."
```
sxjscience, line 2004: Consider to add more details in the Assertion message:

> Trying to resume training from '{path}'. However, it does not contain the last checkpoint file: '{last_ckpt_path}'. Are you using the correct path?

zhiqiangdon: Thanks. Updated.

bryanyzhu (approved):

sxjscience (approved):

PR #1985, opened 2022-07-26
Linting code in the `timeseries` module with `black` and `isort`

shchur:
This commit ensures that code style (`black`) and imports order (`isort`) are consistent in the `autogluon.timeseries` package.

I decided to extend the existing script in `.github/workflow_scripts/lint_check.sh` instead of creating a separate action for the timeseries module to reduce the amount of duplicate code.

- Run `black` and `isort` on the contents of the `timeseries` module
- Add `isort` config to `pyproject.toml`
- Extend the Github action `.github/workflow_scripts/lint_check.sh` to run `black` and `isort` on the `timeseries` module
- Add `isort>=5.10` and `black~=22.0,>=22.3` dependencies to `timeseries/setup.py`, `docs/build.yml` and `.github/workflow_scripts/env_setup.sh`

To do:
- [x] Check if the `isort` dependency needs to be added in other places in the codebase


.github/workflow_scripts/lint_check.sh
```diff
@@ -6,3 +6,5 @@ source $(dirname "$0")/env_setup.sh
 
 setup_build_env
 black --check --diff multimodal/src/autogluon/multimodal
+black --check --diff timeseries/
```
canerturkmen, line 9: let's please have time series lint check as a separate CI step, running in parallel with `multimodal` lint check?

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
```diff
@@ -85,9 +85,7 @@ def __init__(self, data: Any, *args, **kwargs):
         elif isinstance(data, Iterable):
             data = self._construct_pandas_frame_from_iterable_dataset(data)
         else:
-            raise ValueError(
```
canerturkmen, line 88: I think this is where black hurts the most. My black version is 22.1 and it doesn't mind this line for example (this is true for _most_ of the changes this PR introduces, but this might just be that I use `blackd` and it doesn't respect the line length config). Still, in my experience, black is also not backward-compatible (maybe they don't even claim to be). 

How are we going to handle this? Should we fix a black version among us (and also in our dedicated env_setup) for use?

timeseries/setup.py
```diff
@@ -47,14 +45,14 @@
     )
 
 extras_require = {
-    "tests": ["pytest", "flake8~=4.0", "flaky~=3.7", "pytest-timeout~=2.1"],
+    "tests": ["pytest", "flake8~=4.0", "flaky~=3.7", "pytest-timeout~=2.1", "isort>=5.10"],
```
canerturkmen, line 48: do we need to add this to requirements? Isn't this only necessary at development time?

.github/workflow_scripts/lint_check.sh
shchur: I think this option will result in a noticeable amount of redundant code in [`.github/workflows/continuous_integration.yml`](https://github.com/awslabs/autogluon/blob/master/.github/workflows/continuous_integration.yml) and may potentially hurt future extensibility:
- We will need to add `lint_check_timeseries` and `lint_check_multimodal` as a prerequisite (`needs: ...`) to many other actions. 
- If other modules decide to add linting they will again need to extend `needs` list for every existing action as well as add a separate action to the list. However, with the current solution where all linting happens in a single action people will only need to add a single line of code.
- The actual execution time of `black ...` and `isort ...` is negligible compared to setting up new environments, most of the time is spent creating a test environment - we can remove this overhead and not create new actions.

canerturkmen: For the first two points, we seem to have paid this price for other cases such as unit tests and tutorials I think it's OK to pay it here. Also, @yinweisu your comments would be welcome.

I also agree with the final point, but as the overall setup cost is negligible I would prefer to have it as a standalone item in the graph just for usability's sake (and to not have lint fails of some modules masking others if we keep scaling in this manner).

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
shchur: I haven't discovered backwards compatibility issues with black yet, but fixing the version sounds reasonable.

We can also postpone this decision until the first time that we encounter some backwards-incompatible behavior from a future version of black. The change over here definitely seems related to the `line-length` parameter in the config, can you try running a local version of `black` to see if it causes problems?

timeseries/setup.py
shchur: I don't think we need it here but we need to mention it somewhere so that people can automatically lint the code they develop. I followed the logic in [`text/setup.py`](https://github.com/awslabs/autogluon/blob/master/text/setup.py) and [`multimodal/setup.py`](https://github.com/awslabs/autogluon/blob/master/multimodal/setup.py) where `black` is mentioned as a requirement in the `tests` scope. Is there some other place where we could move this dependency?

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
canerturkmen: OK sounds good to me. will try.

timeseries/setup.py
canerturkmen: I see. But our tests don't depend on `black` or `isort` and here we don't have "black" anyway? I would either move these to `CONTRIBUTION.md` or put black also here?

.github/workflow_scripts/lint_check.sh
yinweisu: @shchur Thanks for the contribution! I think the duplicate code is not too much of a concern as @canerturkmen has discussed. However, we do pay an overhead of spinning up aws batch machines and downloading image, which is not cheap compared to lint check. Maybe we can start with having lint checks all together, and we can always change that later easily if we want.

canerturkmen: thanks @yinweisu . resolving this then.

timeseries/setup.py
canerturkmen: Could we maybe add black to test requirements also to trigger CI again?

canerturkmen (approved):

PR #1986, opened 2022-07-26
adding static features to time series data set API

canerturkmen:
This is a request for comments / discussion on updating the time series API to handle static features (item metadata in forecasting). In this PR, we propose to do this via several important changes to the `TimeSeriesDataFrame`.

1. We remove the `split_by_item` API as generic item indices do not necessarily have to follow a total ordering. In other words, for example, if each item id represents a distinct customer, then there are no use cases to really slice "all users up to user A."
2. We introduce the `static_features` property and metadata, which allows users to specify item metadata along with the time series data encapsulated in a `TimeSeriesDataFrame`. Embedding static features into the object has the advantage of separating data from logic, as well as avoiding introducing new arguments to all `fit` and `predict` functions in the time series module just for handling static features. However, it poses some difficulty in ensuring consistency in the time series data frame's data during data manipulation, which is what most of this PR deals with. 


timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
```diff
@@ -112,11 +156,11 @@ def freq(self):
         return freq
 
     def iter_items(self) -> Iterable[Any]:
-        return iter(self.index.levels[0])
+        return iter(self._item_index)
```
shchur, line 159: This also fixes a bug that we currently have on `master`, where `iter_items` is broken for subsets of a `TimeSeriesDataFrame` selected via `loc` :+1:

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
```diff
@@ -89,11 +104,40 @@ def __init__(self, data: Any, *args, **kwargs):
...
+    def static_features(self):
+        return self._static_features
+
+    @static_features.setter
+    def static_features(self, value: Optional[pd.DataFrame]):
+        # if the current item index is not a multiindex, then we are dealing with a single
+        # item slice. this should only happen when the user explicitly requests only a
+        # single item or during `slice_by_timestep`. In this case we do not set static features
+        if not isinstance(self.index, pd.MultiIndex):
+            return
```
shchur, line 129: Maybe `return None` here to be more explicit?

timeseries/tests/unittests/test_ts_dataset.py
```diff
@@ -20,13 +27,36 @@
...
+        [item_ids, datetime_index], names=["item_id", "timestamp"]
+    )
+    if static_features is not None:
+        return TimeSeriesDataFrame(
+            pd.Series(target, name="target", index=multi_inds).to_frame(),
+            static_features=static_features,
+        )
+    return TimeSeriesDataFrame(
+        pd.Series(target, name="target", index=multi_inds).to_frame(),
+    )
```
shchur, line 41: Wouldn't this be equivalent to the following code (no if condition)?
```python
def _build_ts_dataframe(item_ids, datetime_index, target, static_features=None):
    multi_inds = pd.MultiIndex.from_product(
        [item_ids, datetime_index], names=["item_id", "timestamp"]
    )
    return TimeSeriesDataFrame(
        pd.Series(target, name="target", index=multi_inds).to_frame(),
        static_features=static_features,
    )
```

timeseries/tests/unittests/test_ts_dataset.py
```diff
@@ -251,7 +235,9 @@ def test_subsequence(start_timestamp, end_timestamp, item_ids, datetimes, target
         (["2020-01-01 00:00:00", "2020-01-01 01:00:00", "2020-01-01 02:00:00"], "H"),
     ],
 )
-def test_when_dataset_constructed_from_dataframe_without_freq_then_freq_is_inferred(timestamps, expected_freq):
```
shchur, line 254: It seems that you have `black` running with a different value of `line-length` (not 119 as in `pyproject.toml`), which generates these redundant formatting changes

shchur (commented):
This looks good to me overall (except some minor changes requested in the comments)

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
canerturkmen: I would think this is semantically a bit cleaner. See https://stackoverflow.com/questions/15300550/return-return-none-and-no-return-at-all 
So to my understanding is that we `return None` when we have an `Optional` return type and we're explicitly returning a `None` somebody else may do something with. And `return` just means stop evaluating the function. What do you think?

timeseries/tests/unittests/test_ts_dataset.py
canerturkmen: 👍 blacking this file again.

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
shchur: Somehow my read of that StackOverflow answer was that `return None` would be more appropriate here, but this is a really minor nitpick from my side, so I guess either option is fine.

shchur (approved):

PR #1989, opened 2022-07-27
[Security] Add bandit to check for security vulnerabilities in multimodal package.

sxjscience:
*Issue #, if available:*

*Description of changes:*

Add [bandit](https://github.com/PyCQA/bandit) to check for security errors.

### DIFF
--- a/multimodal/src/autogluon/multimodal/utils.py
+++ b/multimodal/src/autogluon/multimodal/utils.py
@@ -1806,7 +1806,7 @@ def download(
         with open(filename, mode="rb") as f:
             # Disable bandit check because we are using sha1sum for evaluating the checksums.
             # It is not used for hosting credentials.
-            d = hashlib.new('sha1', usedforsecurity=False)  # nosec(sxjscience)
+            d = hashlib.new("sha1", usedforsecurity=False)  # nosec(sxjscience)
             for buf in iter(functools.partial(f.read, 1024 * 100), b""):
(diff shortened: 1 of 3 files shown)


multimodal/src/autogluon/multimodal/utils.py
```diff
@@ -1804,7 +1804,7 @@ def sha1sum(filename: str):
             The sha1sum
         """
         with open(filename, mode="rb") as f:
-            d = hashlib.sha1()
+            d = hashlib.new('sha1', usedforsecurity=False)  # Mark as insecure due to bandit check
```
yinweisu, line 1807: Is this the correct solution? This might prevent bandit to complain, but might just turned off the bandit check. Just want to double check

sxjscience: Actually I decide to mark it as `nosec` because we are using sha1sum as checksums instead of using it as credentials.

sxjscience: Basically, it's not protecting any sensitive contents. We can safely also use md5.

sxjscience:
@yinweisu We can leave enabling adding bandit to the other folders as future work.

sxjscience:
@yinweisu @gradientsky Would you take another look?

yinweisu (approved):
LGTM!

gradientsky (approved):

PR #1991, opened 2022-07-27
Fix Platform Tests

yinweisu:
*Issue #, if available:*

*Description of changes:*
* https://github.com/awslabs/autogluon/pull/1987/files removed ray lightning as dependency of core temporarily causing core tests to fail. Ray lightning was added to CI env setup script but not used by platform tests. Now platform tests uses CI scripts as well.
* Added a way for test scripts to take in additional test args
* Platform tests change won't be reflected until merged in


sxjscience (approved):
LGTM. Thanks!

PR #1993, opened 2022-07-28
Fix timeseries failing silently on irregularly sampled time series

canerturkmen:
*Issue #, if available:*

*Description of changes:*

AG-TS failed silently on irregularly sampled time series that also can't be inferred as business day-frequency by pandas. As we depend on pandas for frequency inference, this will continue to be the case. However, we can also add a `fit` argument to the predictor called `ignore_time_indexes` to enable customers using irregular timestamps as is (as if they are irregularly sampled), but this would somewhat break API semantics during predict (where a "dummy" time index will be output). 

## TODO

- [x] add `ignore_time_indexes` logic to `TimeSeriesPredictor` and corresponding view logic to data frames
- [x] add unit tests.

### DIFF
--- a/timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
+++ b/timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
@@ -468,7 +468,7 @@ class TimeSeriesDataFrame(pd.DataFrame):
         except Exception as err:  # noqa
             raise IOError(f"Could not load pickled data set due to error: {str(err)}")
 
-    def get_reindexed_view(self, freq: str = "s") -> TimeSeriesDataFrame:
+    def get_reindexed_view(self, freq: str = "S") -> TimeSeriesDataFrame:
         """Returns a new TimeSeriesDataFrame object with the same underlying data and
         static features as the current data frame, except the time index is replaced by
         a new "dummy" time series index with the given frequency. This is useful when
--- a/timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
+++ b/timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
@@ -136,20 +136,17 @@ class TimeSeriesDataFrame(pd.DataFrame):
 
     @property
     def freq(self):
-        ts_index = self.index.levels[1]  # noqa
-        freq = (
-            ts_index.freq
-            or ts_index.inferred_freq
-            or self.loc[0].index.freq  # fall back to freq of first item
-            or self.loc[0].index.inferred_freq
-        )
-        if freq is None:
-            raise ValueError("Frequency not provided and cannot be inferred")
-        if isinstance(freq, str):
-            return freq
-        elif isinstance(freq, pd._libs.tslibs.BaseOffset):
-            return freq.freqstr
-        return freq
+        def get_freq(series):
+            return series.index.freq or series.index.inferred_freq
+        freq_for_each_series = [get_freq(self.loc[idx]) for idx in self.iter_items()]
+        freq = freq_for_each_series[0]    
+        if len(set(freq_for_each_series)) > 1 or freq is None:
+            raise ValueError(
+                "Frequency not provided and cannot be inferred. This is often due to the "
+                "time index of the data being irregularly sampled. Please ensure that the "
+                "data set used has a uniform time index."
+            )
+        return freq.freqstr if isinstance(freq, pd._libs.tslibs.BaseOffset) else freq
 
     def iter_items(self) -> Iterable[Any]:
         return iter(self._item_index)
--- a/timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
+++ b/timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
@@ -12,6 +12,8 @@ from pandas.core.internals import ArrayManager, BlockManager
 ITEMID = "item_id"
 TIMESTAMP = "timestamp"
 
+IRREGULAR_TIME_INDEX_FREQSTR = "IRREG"
+
 
 class TimeSeriesDataFrame(pd.DataFrame):
     """``TimeSeriesDataFrame`` s represent a collection of time series, where each row
@@ -106,6 +108,10 @@ class TimeSeriesDataFrame(pd.DataFrame):
         if static_features is not None:
             self.static_features = static_features
 
+        # internal value for cached frequency values that are inferred. corresponds to either a
+        # pandas-compatible frequency string, the value IRREGULAR_TIME_INDEX_FREQSTR that signals
+        # the time series have irregular timestamps (in which case tsdf.freq returns None), or None
+        # if inference was not yet performed.
         self._cached_freq: Optional[str] = None
 
     @property
@@ -140,7 +146,9 @@ class TimeSeriesDataFrame(pd.DataFrame):
 
     @property
     def freq(self):
-        if self._cached_freq:
+        if self._cached_freq is not None and self._cached_freq == IRREGULAR_TIME_INDEX_FREQSTR:
+            return None  # irregularly sampled time series
+        elif self._cached_freq:
             return self._cached_freq
 
         def get_freq(series):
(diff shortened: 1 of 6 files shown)


timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
```diff
@@ -96,15 +96,20 @@ def _constructor(self) -> Type[TimeSeriesDataFrame]:
 
     @property
     def freq(self):
-        ts_index = self.index.levels[1]  # noqa
+        ts_index = self.index.get_level_values(1)  # noqa
```
shchur, line 99: Do we ever even use `ts_index` to infer the frequency here? 

I thought that `self.index` is always a `pd.MultiIndex` where the first level is the item ID and the second level are the timestamps for each individual time series. Therefore, `ts_index = self.index.get_level_values(1)` will always contain the concatenated timestamps for all individual time series (most likely not even sorted chronologically) where `freq` and `inferred_freq` don't make that much sense.

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
```diff
@@ -96,15 +96,20 @@ def _constructor(self) -> Type[TimeSeriesDataFrame]:
...
+        first_item = self.index.get_level_values(0)[0]
         freq = (
             ts_index.freq
             or ts_index.inferred_freq
-            or self.loc[0].index.freq  # fall back to freq of first item
-            or self.loc[0].index.inferred_freq
+            or self.loc[first_item].index.freq  # fall back to freq of first item
+            or self.loc[first_item].index.inferred_freq
```
shchur, line 105: Wouldn't it make more sense to look at `freq` / `inferred_freq` of each individual time series to ensure that 
1. The frequencies are consistent across series in the dataset
2. It doesn't happen that other time series (not the first one) have an irregular index?

Something like the following snippet
```python
def get_freq(series):
    return series.index.freq or series.index.inferred_freq

freq_for_series = [get_freq(self.loc[idx]) for idx in self.iter_items()]
if len(set(freq_for_series)) > 1 or freq_for_series[0] is None:
    raise ValueError("...")
```

shchur (changes requested):

Innixma (commented):
Code looks reasonable to me, but will defer to @shchur for making the final call on merging, as I am less familiar with the particular bug.

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
Innixma: If it is cheap to do, this will be valuable and I agree it makes sense to check per time series.

canerturkmen: It's actually not very cheap. For one of the data sets I'm looking at, this takes about ~5sec for 10K items with each about 30 time steps. Do we want to pay this or maybe just look at top 100 items as a middle ground solution?

canerturkmen: also note this will be called every time freq is accessed. so we might want to cache this value as well.

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
```diff
@@ -136,19 +140,21 @@ def static_features(self, value: Optional[pd.DataFrame]):
...
-            or self.loc[0].index.inferred_freq
-        )
-        if freq is None:
-            raise ValueError("Frequency not provided and cannot be inferred")
-        if isinstance(freq, str):
-            return freq
-        elif isinstance(freq, pd._libs.tslibs.BaseOffset):
-            return freq.freqstr
+        if self._cached_freq:
+            return self._cached_freq
```
shchur, line 144: Do I understand correctly that if we have an irregularly-sampled TS, we will set `self._cached_freq = None` and therefore rerun lines 143-157 each time we access `df.freq`?

If this is the case, should we use some other token to denote irregularly-sampled TS (like `"None"` or `nan`), or is the overhead negligible?

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
```diff
@@ -136,19 +140,21 @@ def static_features(self, value: Optional[pd.DataFrame]):
...
-            or self.loc[0].index.freq  # fall back to freq of first item
-            or self.loc[0].index.inferred_freq
-        )
-        if freq is None:
-            raise ValueError("Frequency not provided and cannot be inferred")
-        if isinstance(freq, str):
-            return freq
-        elif isinstance(freq, pd._libs.tslibs.BaseOffset):
-            return freq.freqstr
+        if self._cached_freq:
```
shchur, line 143: Do I understand correctly that if we have an irregularly-sampled TS, we will set `self._cached_freq = None` and therefore rerun lines 143-157 each time we access `df.freq`?

If this is the case, should we use some other token to denote irregularly-sampled TS (like `"None"` or `nan`), or is the overhead negligible?

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
```diff
@@ -406,7 +417,9 @@ def slice_by_timestep(self, time_step_slice: slice) -> TimeSeriesDataFrame:
             idx = pd.MultiIndex.from_product([(ix,), data_slice.index], names=[ITEMID, TIMESTAMP])
             data_slice.set_index(idx, inplace=True)
             slices.append(data_slice)
-        return self.__class__(pd.concat(slices), static_features=self.static_features)
+        slice_df = self.__class__(pd.concat(slices), static_features=self.static_features)
+        slice_df._cached_freq = self._cached_freq
```
shchur, line 421: We might need to recompute the `_cached_freq` here - for example, a user could specify `time_step_slice=slice(None, None, 2)` which would upsample the time series

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -219,6 +245,8 @@ def fit(
         if hyperparameters is None:
             hyperparameters = "default"
 
+        train_data, tuning_data = self._check_and_prepare_data_frames(train_data, tuning_data)
```
shchur, line 248: Nitpick: Maybe the code would be a bit cleaner if
-  `_check_and_prepare_data_frame` always took one argument (`Union[TimeSeriesDataFrame, None]`)
- we would call `_check_and_prepare_data_frame` separately for train & tuning data here

This will remove some nesting in `_check_and_prepare_data_frame` and allow us to replace `(train_data,) = ...` with `train_data = ...` in a few places

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
shchur: Maybe `nan` also makes more sense to distinguish between the cases where `_cached_freq` wasn't set vs. where it's impossible to determine

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
canerturkmen: Hmm, right. But I think I will rather block this kind of upsampling here as it may cause other problems we don't know. What do you think?

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
canerturkmen: Hmm. I could use a token `"NA"`, but then every consumer of this class has to also check `NA or None`, perhaps through a new instance method of `TimeSeriesDataFrame`. That said, you are right I don't think the overhead is negligible.

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
shchur: I thought the whole point of passing a `slice(...)` object here instead of something like `start_idx=None, end_idx=None` here was to allow upsampling 😅 

I'm okay with disabling subsampling. Should we change the API in this case to only allow start_idx / end_idx (in a separate PR)?

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
canerturkmen: `_cached_freq` is a string type so nan doesn't really work imo.

shchur: What about
```python
if self._cached_freq == "NA":  # or self._cached_freq == np.nan
    return None
elif self._cached_freq is not None:
    return self._cached_freq
```

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
shchur: I would take care of this in #2020, if we decide to go down this route

shchur (approved):

PR #1996, opened 2022-07-29
[Cloud] Image cloud predictor

yinweisu:
*Issue #, if available:*

*Description of changes:*
* Added ImageCloudPredictor. API and usage are as follow:
  * Training: user needs to pass an argument `image_path`, which is a local path or s3 path to the images, to `fit()`.
  * General workflow is to archive the images, upload to s3 and setup the channel to sagemaker. Within the container, we uncompress the archive and do fit.
```python
"""
image_path: str, default = None
  A local path or s3 path to the images. If you provided this parameter, the image path inside your train/tune data MUST be relative.
  If local path, path needs to be either a compressed file containing the images or a folder containing the images.
  If it's a folder, we will zip it for you and upload it to the s3.
  If s3 path, the path needs to be a path to a compressed file containing the images
"""
```
  * Realtime inference:
  * General workflow is to detect images from dataframe and converting them into numpy arrays. Pass numpy arrays as payload to the endpoint. In the endpoint, we decrypt the array, save them as files, and pass file paths to predictor.
```python
def predict_real_time(self, test_data, test_data_image_column=None, accept='application/x-parquet'):
        """
        Predict with the deployed SageMaker endpoint. A deployed SageMaker endpoint is required.
        This is intended to provide a low latency inference.
        If you want to inference on a large dataset, use `predict()` instead.

        Parameters
        ----------
        test_data: Union(str, pandas.DataFrame)
            The test data to be inferenced.
            Can be a pandas.DataFrame, a numpy.ndarray, a local path or a s3 path to a csv file containing paths of test images.
            Or a local path to a single image file.
            Or a list of local paths to image files.
        test_data_image_column: Optional(str)
            If provided a pandas.DataFrame as the test_data, you must specify the column name corresponding to image paths
            Images has to live in the same directory specified by the column.
        accept: str, default = application/x-parquet
            Type of accept output content.
            Valid options are application/x-parquet, text/csv, application/json
        """
```
  * Batch inference: user needs to pass `test_data` as a path to a directory containing the images to `predict()`
  * Upload image folders directly to s3, and use transformers with `content_type = 'application/x-image'`, which is natively supported by batch transform.
```python
"""
test_data: str
  The test data to be inferenced. Can be a local path or a s3 path to a directory containing the images.
"""
```
* Added unit tests for ImageCloudPredictor
* Updated the way to detect large payload size when doing realtime inference. Previously, we calculate the payload size ourselves. However, that calculation was inaccurate. Now, we catch the error and warn user instead.


cloud/src/autogluon/cloud/predictor/predictor.py
```diff
@@ -238,12 +244,40 @@ def _upload_fit_artifact(
...
+        if images_input is not None:
+            inputs['images'] = images_input
+
+        return inputs
 
     def fit(
         self,
         predictor_init_args,
         predictor_fit_args,
+        image_path=None,
```
bryanyzhu, line 280: Just want to ask if `image_path` is a must to provide. If it is a must, do we want to set a default value here?

cloud/src/autogluon/cloud/utils/ag_sagemaker.py
```diff
@@ -81,8 +81,18 @@ def __init__(self, *args, **kwargs):
...
+            *args,
+            serializer=NumpySerializer(),
+            deserializer=PandasDeserializer(),
+            **kwargs
+        )
+
+
 # Predictor documentation: https://sagemaker.readthedocs.io/en/stable/api/inference/predictors.html
-# SageMaker can only take in csv format for batch transformation because files need to be easily splitable to  be batch processed.
+# SageMaker can only take in csv format for batch transformation because files need to be easily splitable to be batch processed.
```
bryanyzhu, line 95: I saw the default input and output format is x-parquet, but here only takes in csv format. Are they different use scenarios? 

Also, is x-parquet easy to check, like you can open it in sublime or browser, or you need some app to read it? Thanks.

bryanyzhu (commented):
Thanks for the PR

cloud/src/autogluon/cloud/utils/ag_sagemaker.py
yinweisu: So for the image prediction, the batch inference input format is actually x-image. When you are referring to the default input and output format as x-parquet, I believe that's realtime-inference and training.

I don't think you can open an editor and view x-parquet directly.

cloud/src/autogluon/cloud/predictor/predictor.py
yinweisu: In the `ImagePredictor.fit()`, `image_path` is actually required. Here is the `fit()` for general `CloudPredictor`.
You can also imagine when we implement multimodality, this `image_path` could be optional

cloud/src/autogluon/cloud/predictor/predictor.py
```diff
@@ -897,3 +951,125 @@ def _get_local_predictor_cls(self):
...
+        **kwargs
+    ):
+        super().fit(
+            predictor_init_args,
+            predictor_fit_args,
+            image_path=image_path,
+            **kwargs
+        )
+
+    def predict_real_time(self, test_data, test_data_image_column=None, accept='application/x-parquet'):
```
gradientsky, line 987: Do we accept only parquet? SM API [seems support])(https://sagemaker.readthedocs.io/en/stable/api/inference/predictors.html#sagemaker.predictor.Predictor.accept) multiple accept types.

cloud/src/autogluon/cloud/predictor/predictor.py
```diff
@@ -897,3 +951,125 @@ def _get_local_predictor_cls(self):
...
+    ):
+        """
+        test_data: str
+            The test data to be inferenced. Can be a local path or a s3 path to a directory containing the images.
+        kwargs:
+            Refer to `CloudPredictor.predict()`
+        """
+        split_type = None
+        content_type = 'application/x-image'
+        transformer_kwargs['strategy'] = 'SingleRecord'
```
gradientsky, line 1068: This will mutate passed argument. I suggest making a copy and perform modification on the copy.

cloud/src/autogluon/cloud/predictor/predictor.py
```diff
@@ -691,10 +736,11 @@ def predict(
             logger.warning('Providing a custom predictor_cls could break the deployment. Please refer to `AutoGluonBatchPredictor` for how to provide a custom predictor')
             predictor_cls = user_predictor_cls
 
-        split_type = kwargs.pop('split_type', None)
         content_type = kwargs.pop('content_type', None)
-        if not split_type:
```
gradientsky, line 696: Can we just use `Line` as a default `split_type` vs passing `None` and then converting into default?

cloud/src/autogluon/cloud/predictor/predictor.py
```diff
@@ -897,3 +951,125 @@ def _get_local_predictor_cls(self):
...
+        transformer_kwargs=dict(),
+        **kwargs,
+    ):
+        """
+        test_data: str
+            The test data to be inferenced. Can be a local path or a s3 path to a directory containing the images.
+        kwargs:
+            Refer to `CloudPredictor.predict()`
+        """
+        split_type = None
```
gradientsky, line 1066: why pass `None` instead of rely on default (see comment on super-class)?

cloud/src/autogluon/cloud/scripts/image_serve.py
```diff
@@ -0,0 +1,62 @@
...
+
+import pandas as pd
+import numpy as np
+
+
+def model_fn(model_dir):
+    """loads model from previously saved artifact"""
+    model = ImagePredictor.load(model_dir)
+
+    return model
```
gradientsky, line 15: Is there `.persist()` call for image predictors?

cloud/src/autogluon/cloud/predictor/predictor.py
```diff
@@ -897,3 +951,125 @@ def _get_local_predictor_cls(self):
...
+            images = []
+            for image in test_data:
+                images.append(cv2.imread(image))
+            test_data = np.array(images, dtype=object)
+        if isinstance(test_data, pd.DataFrame):
+            assert test_data_image_column is not None, 'Please specify an image column name'
+            assert test_data_image_column in test_data, 'Please specify a valid image column name'
+
+            # Convert test data to be numpy array for network transfer
+            test_data = test_data[test_data_image_column].apply(lambda path: cv2.imread(path)).to_numpy()
```
Innixma, line 1034: `.apply` is a very slow operation, consider using list comprehension + .zip() which is faster.

https://stackoverflow.com/questions/52607864/dataframe-list-comprehension-zip-loop-through-chosen-df-columns-efficien

cloud/src/autogluon/cloud/predictor/predictor.py
```diff
@@ -897,3 +951,125 @@ def _get_local_predictor_cls(self):
         from autogluon.text import TextPredictor
         predictor_cls = TextPredictor
         return predictor_cls
+
+
+class ImageCloudPredictor(CloudPredictor):
```
Innixma, line 956: More broad question but do we want to call this ImageSMCloudPredictor and differentiate between logic that is SM specific and generic? Doesn't need to be in this PR, but something to think about in-case we want to add support for solutions other than SM.

cloud/src/autogluon/cloud/predictor/predictor.py
```diff
@@ -897,3 +951,125 @@ def _get_local_predictor_cls(self):
...
+        test_data,
+        transformer_kwargs=dict(),
+        **kwargs,
+    ):
+        """
+        test_data: str
+            The test data to be inferenced. Can be a local path or a s3 path to a directory containing the images.
+        kwargs:
+            Refer to `CloudPredictor.predict()`
+        """
```
Innixma, line 1065: transformer_kwargs not documented?

cloud/src/autogluon/cloud/predictor/predictor.py
```diff
@@ -897,3 +951,125 @@ def _get_local_predictor_cls(self):
...
+            )
+            logger.log(20, 'Data uploaded successfully')
+        else:
+            test_input = test_data
+        return test_input
+
+    def predict(
+        self,
+        test_data,
+        transformer_kwargs=dict(),
```
Innixma, line 1057: Do not use dict() as default, it is mutable!! (aka your default can be mutated and not be empty anymore in future calls...) Instead default=None and then set to dict() inside of method.

cloud/src/autogluon/cloud/predictor/predictor.py
Innixma: +1

cloud/src/autogluon/cloud/predictor/predictor.py
yinweisu: We support more than that. This is just the default

cloud/src/autogluon/cloud/predictor/predictor.py
```diff
@@ -535,33 +569,43 @@ def attach_endpoint(self, endpoint):
...
 
+    def _predict_realtime(self, test_data, accept):
+        try:
+            return self.endpoint.predict(test_data, initial_args={'Accept': accept})
+        except ClientError as e:
+            if not e.response['Error']['Code'] == '413':  # Error code for pay load too large
+                raise e
+            else:
+                logger.warning('The invocation of endpoint failed with Error Code 413. This is likely due to pay load size being too large.')
+                logger.warning('SageMaker endpoint could only take maximum 5MB. Please consider reduce test data size or use `predict()` instead.')
```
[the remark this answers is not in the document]
yinweisu: The stack trace given by sagemaker is not useful unless you google the error code, but I can change it to raise the exception still.

cloud/src/autogluon/cloud/scripts/image_serve.py
yinweisu: I don't think so. Why?

cloud/src/autogluon/cloud/scripts/image_serve.py
```diff
@@ -0,0 +1,62 @@
...
+
+def transform_fn(model, request_body, input_content_type, output_content_type="application/json"):
+
+    if input_content_type == "application/x-npy":
+        buf = BytesIO(request_body)
+        data = np.load(buf, allow_pickle=True)
+        image_paths = []
+        for i, image in enumerate(data):
+            im = Image.fromarray(image)
+            im_name = f'{i}.jpeg'
```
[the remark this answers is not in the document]
yinweisu: No reason, I just picked one randomly. I can change to png if it's better.

cloud/src/autogluon/cloud/predictor/predictor.py
yinweisu: Discussed offline. Will resolve

cloud/src/autogluon/cloud/predictor/predictor.py
yinweisu: ditto

cloud/src/autogluon/cloud/scripts/image_serve.py
```diff
@@ -0,0 +1,62 @@
...
+    else:
+        raise Exception(
+            f'{input_content_type} input content type not supported.'
+        )
+
+    if model._problem_type != REGRESSION:
+        pred_proba = model.predict_proba(image_paths)
+        pred = get_pred_from_proba_df(pred_proba, problem_type=model._problem_type)
+        pred_proba.columns = [str(c) + '_proba' for c in pred_proba.columns]
+        pred.name = pred.name + '_pred' if pred.name is not None else 'pred'
```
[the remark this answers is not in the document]
yinweisu: Converted to str to avoid this situation

cloud/setup.py
```diff
@@ -20,6 +20,8 @@
     # version ranges added in ag.get_dependency_version_ranges()
     'autogluon.common<0.6'
     'boto3',
+    'numpy',
+    'opencv-python'
```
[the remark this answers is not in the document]
yinweisu: added a version range

cloud/src/autogluon/cloud/predictor/predictor.py
yinweisu: I think maybe letting the user to specify a backend is better? This way they don't need to mess around with all different kinds of CloudPredictor class. Open to discussion too

cloud/src/autogluon/cloud/predictor/predictor.py
```diff
@@ -897,3 +951,125 @@ def _get_local_predictor_cls(self):
...
+        predictor_fit_args,
+        image_path,
+        **kwargs
+    ):
+        super().fit(
+            predictor_init_args,
+            predictor_fit_args,
+            image_path=image_path,
+            **kwargs
+        )
```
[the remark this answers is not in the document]
yinweisu: Enforced keyword args

cloud/src/autogluon/cloud/predictor/predictor.py
yinweisu: I removed transformer_kwargs as a standalone arg as I don't think regular user should worry about it. User can still pass it through kwargs, and the documentation is available in the base class

cloud/src/autogluon/cloud/scripts/image_serve.py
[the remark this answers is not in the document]
Innixma: @bryanyzhu @sxjscience I'm assuming png is better to avoid lossy compression, but wanted to hear your thoughts.

cloud/src/autogluon/cloud/predictor/predictor.py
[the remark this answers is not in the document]
Innixma: Ok, it is up to you

cloud/src/autogluon/cloud/predictor/predictor.py
Innixma: Ok, that sounds fine for now.

cloud/src/autogluon/cloud/scripts/image_serve.py
gradientsky: `TabularPredictor` pre-loads models only on persist call this improves latency during inference times. I was not sure if `ImagePredictor` has a similar method or is it always in-memory.

cloud/src/autogluon/cloud/scripts/image_serve.py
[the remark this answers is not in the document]
gradientsky: Many datasets have a mix of `jpg`, `jpeg`, `png`; could be also `tif` and `tiff`.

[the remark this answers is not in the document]
yinweisu: @gradientsky I think you misunderstood. We are discussing the format we save to disk in the sagemaker container after taking in the payload

cloud/src/autogluon/cloud/scripts/image_serve.py
yinweisu: I see. For `TabularPredictor`, should we call `.persist()` after loading then?

Innixma (approved):
LGTM, nice work!

gradientsky (approved):

bryanyzhu (approved):

PR #2001, opened 2022-08-01
Improve sktime models API

shchur:
This PR fixes several problems with `sktime` models (`ThetaModel`, `TBATSModel`, `AutoETSModel`, `ARIMAModel`, `AutoARIMAModel`) in the timeseries module:
- Suppress `RuntimeWarning`s during prediction that flooded the model output in some cases.
- Add docstrings for `sktime` models. These describe the possible hyperparameters that can be passed to the models.
- A warning is made if user provides a hyperparameter that isn't recognized by the model.
- New method `AbstractSktimeModel._get_skt_forecaster_init_args` 
  - allows us to decouple the API exposed to AG users (via `hyperparameters`) from the API of `sktime` that sometimes uses inconsistent / cryptic parameter names.
  - gives us flexibility to change the hyperparameters in some cases to make the model more robust (e.g., turn off seasonality if we know that the model will fail with the current config).
- Improve seasonality support
  - All `sktime` models now exposes a `Optional[int]` hyperparameter `seasonal_period`. When `seasonal_period=None` (default), the seasonal period will be inferred automatically from the data.
  - User can now turn off seasonality and override the seasonal period for different models (before seasonality couldn't be turned off, `sp` was always inferred automatically)
  - `ARIMA` now also supports seasonality.
- Replace all occurrences of `*skt_*` with `*sktime_*` for better consistency.

Discussion points
1. Is decoupling the AG API from `sktime` API worth it or can it have some potential side effects?

To do
- [x] Add tests

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/sktime/abstract_sktime.py
+++ b/timeseries/src/autogluon/timeseries/models/sktime/abstract_sktime.py
@@ -70,14 +70,9 @@ class AbstractSktimeModel(AbstractTimeSeriesModel):
         self.sktime_forecaster: Optional[BaseForecaster] = None
         self._fit_index: Optional[pd.Index] = None
 
-    def _get_sktime_forecaster_init_args(
-        self, model_params: Dict[str, Any], min_length: int, inferred_period: int = 1
-    ):
-        """Get arguments that will be passed to the sktime forecaster at initialization.
-
-        model_params may be modified in-place based on the length of training series and inferred seasonality.
-        """
-        return model_params
+    def _get_sktime_forecaster_init_args(self, min_length: int, inferred_period: int = 1):
+        """Get arguments that will be passed to the sktime forecaster at initialization."""
+        return self._get_model_params().copy()
 
     def _get_sktime_forecaster(self, sktime_forecaster_init_args: dict) -> BaseForecaster:
         """Create an sktime forecaster object for the model with given args."""
@@ -135,7 +130,7 @@ class AbstractSktimeModel(AbstractTimeSeriesModel):
         min_length = train_data.index.get_level_values(0).value_counts(sort=False).min()
         inferred_period = get_seasonality(train_data.freq)
         sktime_forecaster_init_args = self._get_sktime_forecaster_init_args(
-            model_params=self._get_model_params().copy(), min_length=min_length, inferred_period=inferred_period
+            min_length=min_length, inferred_period=inferred_period
         )
         self.sktime_forecaster = self._get_sktime_forecaster(sktime_forecaster_init_args)
 
--- a/timeseries/src/autogluon/timeseries/models/sktime/abstract_sktime.py
+++ b/timeseries/src/autogluon/timeseries/models/sktime/abstract_sktime.py
@@ -70,15 +70,24 @@ class AbstractSktimeModel(AbstractTimeSeriesModel):
         self.skt_forecaster: Optional[BaseForecaster] = None
         self._fit_index: Optional[pd.Index] = None
 
-    def _get_skt_forecaster(self, sp: Optional[int] = None) -> BaseForecaster:
-        """Return the sktime forecaster object for the model"""
-        params = self._get_model_params().copy()
-        if sp is not None and sp > 1:
-            params["sp"] = sp
-            params["seasonal"] = "add"
+    def _get_skt_forecaster_args(self, args: dict, min_length: int, inferred_period: int = 1):
+        """Get arguments that will be passed to the sktime forecaster.
+
+        These may be modified based on the length of training series and inferred seasonality.
+        """
+        return args
+
+    def _get_skt_forecaster(self, args: dict) -> BaseForecaster:
+        """Create an sktime forecaster object for the model with given args."""
+        unused_args = [k for k in args.keys() if k not in self.sktime_allowed_init_args]
+        if len(unused_args) > 0:
+            warnings.warn(
+                f"{self.name} ignores following arguments: {unused_args}. "
+                f"See `{self.name}.sktime_allowed_init_args` for the list of allowed arguments."
+            )
 
         return self.sktime_forecaster_class(
-            **{k: v for k, v in params.items() if k in self.sktime_allowed_init_args}  # noqa
+            **{k: v for k, v in args.items() if k in self.sktime_allowed_init_args}  # noqa
         )
 
     def _to_skt_data_frame(self, data: TimeSeriesDataFrame) -> pd.DataFrame:
@@ -122,8 +131,12 @@ class AbstractSktimeModel(AbstractTimeSeriesModel):
         self._check_fit_params()
 
         min_length = min(len(train_data.loc[i]) for i in train_data.iter_items())
-        period = get_seasonality(train_data.freq)
-        self.skt_forecaster = self._get_skt_forecaster(sp=period if min_length > 2 * period else None)
+        inferred_period = get_seasonality(train_data.freq)
+        initial_args = self._get_model_params().copy()
+        skt_forecaster_args = self._get_skt_forecaster_args(
+            args=initial_args, min_length=min_length, inferred_period=inferred_period
+        )
+        self.skt_forecaster = self._get_skt_forecaster(skt_forecaster_args)
 
         with warnings.catch_warnings():
             warnings.simplefilter("ignore", category=UserWarning)
@@ -149,11 +162,16 @@ class AbstractSktimeModel(AbstractTimeSeriesModel):
             )
             self._fit(data)
 
-        mean_predictions = self.skt_forecaster.predict(fh=self._fh())
+        with warnings.catch_warnings():
+            # Models in statsmodels may run into numerical issues for some datasets - ignore these
+            warnings.simplefilter("ignore", category=RuntimeWarning)
+
+            mean_predictions = self.skt_forecaster.predict(fh=self._fh())
+            quantile_predictions = self.skt_forecaster.predict_quantiles(
+                fh=self._fh(), alpha=quantile_levels or self.quantile_levels
+            )
+
         mean_predictions.columns = ["mean"]
-        quantile_predictions = self.skt_forecaster.predict_quantiles(
-            fh=self._fh(), alpha=quantile_levels or self.quantile_levels
-        )
         quantile_predictions.columns = [str(q) for q in quantile_predictions.columns.levels[1]]  # noqa
 
         predictions = pd.concat([mean_predictions, quantile_predictions], axis=1)
--- a/timeseries/src/autogluon/timeseries/models/sktime/abstract_sktime.py
+++ b/timeseries/src/autogluon/timeseries/models/sktime/abstract_sktime.py
@@ -16,7 +16,7 @@ from ...utils.seasonality import get_seasonality
 from ..abstract import AbstractTimeSeriesModel
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


timeseries/src/autogluon/timeseries/models/sktime/models.py
```diff
@@ -7,11 +9,94 @@
...
+        if deseasonalize and (min_length < 2 * seasonal_period or seasonal_period <= 1):
+            error_message = (
+                f"{self.name} with deseasonalize = {deseasonalize} requires training series of length "
+                f"at least 2 * seasonal_period and seasonal_period > 1 "
+                f"(received min_length = {min_length} and seasonal_period = {seasonal_period})."
+            )
+            if fail_if_misoconfigured:
+                raise ValueError(error_message)
+            else:
+                warnings.warn(error_message + "\nSetting seasonality to None.")
```
shchur, line 62: Is `logging.info` better suited for this type of information or warning is fine?

canerturkmen: I'd go with 

```
logger = logging.getLoger(__name__)

...

logger.warn(..)
```

timeseries/src/autogluon/timeseries/models/sktime/models.py
```diff
@@ -59,8 +194,79 @@ class AutoETSModel(AbstractSktimeModel):
...
+        seasonality = args.pop("seasonality", "add")
+        if seasonality not in ["add", "mul", None]:
+            raise ValueError(
+                f"Invalid seasonality {seasonality} for model {self.name} (must be one of 'add', 'mul', None)"
+            )
+
+        seasonal_period = args.pop("seasonal_period", None)
+        # Handle the case when seasonal_period is intentionally set to None
+        if seasonal_period is None:
+            seasonal_period = inferred_period
```
canerturkmen, line 217: what happens if `seasonality is None` but inferred period is not 1?

canerturkmen (commented):
Looks great. Thanks! Mostly minor comments, also looking forward to the tests.

timeseries/src/autogluon/timeseries/models/sktime/models.py
```diff
@@ -59,8 +194,79 @@ class AutoETSModel(AbstractSktimeModel):
...
+                args["initialization_method"] = "estimated"
+
+        seasonality = args.pop("seasonality", "add")
+        if seasonality not in ["add", "mul", None]:
+            raise ValueError(
+                f"Invalid seasonality {seasonality} for model {self.name} (must be one of 'add', 'mul', None)"
+            )
+
+        seasonal_period = args.pop("seasonal_period", None)
+        # Handle the case when seasonal_period is intentionally set to None
```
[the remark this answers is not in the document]
shchur: This is an explanation for what the code below does to remove the temptation to rewrite it as 
```python
        seasonal_period = args.pop("seasonal_period", inferred_period)
```
because this option won't handle the case when the user explicitly sets `seasonal_period=None` to perform inference automatically.

timeseries/src/autogluon/timeseries/models/sktime/models.py
shchur: If we set `seasonal_period` but `seasonality=None`, seasonality will be turned off. This is also true for other sktime models (except TBATS, where setting `seasonal_period != 1` is enough to turn on seasonality).

timeseries/src/autogluon/timeseries/models/sktime/models.py
```diff
@@ -7,11 +9,94 @@
...
+
 
 class TBATSModel(AbstractSktimeModel):
+    """TBATS forecaster with multiple seasonalities.
+
+    See `AbstractSktimeModel` for common parameters.
+
+    Other Parameters
+    ----------------
+    use_box_cox: bool or None, default = None
```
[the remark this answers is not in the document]
shchur: I will add a TODO to double-check this later

timeseries/src/autogluon/timeseries/models/sktime/abstract_sktime.py
```diff
@@ -67,21 +67,32 @@ def __init__(
...
         self._fit_index: Optional[pd.Index] = None
 
-    def _get_skt_forecaster(self, sp: Optional[int] = None) -> BaseForecaster:
-        """Return the sktime forecaster object for the model"""
-        params = self._get_model_params().copy()
-        if sp is not None and sp > 1:
-            params["sp"] = sp
-            params["seasonal"] = "add"
+    def _get_sktime_forecaster_init_args(
+        self, model_params: Dict[str, Any], min_length: int, inferred_period: int = 1
```
shchur, line 74: Is `model_params` an okay name here? The logic here is that we received them by calling `self._get_get_model_params().copy()` @canerturkmen

canerturkmen: cool

timeseries/src/autogluon/timeseries/models/sktime/models.py
```diff
@@ -7,11 +9,94 @@
...
+        example, 4 for quarterly data with an annual cycle, or 7 for daily data with a
+        weekly cycle.
+        When set to None, seasonal_period will be inferred from the frequency of the
+        training data. Can also be specified manually by providing an integer > 1.
+    fail_if_misoconfigured: bool, default = False
+        If True, the model will raise an exception and fail when given an invalid
+        configuration (e.g., selected seasonality is incompatible with seasonal_period).
+        If False, the model will instead raise a warning and try to adjust the
+        configuration (e.g., turn off seasonality).
+        Setting this parameter to True is useful during HPO to avoid training multiple
```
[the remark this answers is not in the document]
canerturkmen: Let's leave it in for now, just making sure to set it to default `False` in all presets and function calls?

timeseries/src/autogluon/timeseries/models/sktime/models.py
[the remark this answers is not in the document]
canerturkmen: I see. But the comment is a little confusing as the code below handles the case when `seasonal_period` is also _not_ intentionally (just, by default) set to None.

timeseries/src/autogluon/timeseries/models/sktime/abstract_sktime.py
```diff
@@ -67,30 +67,32 @@ def __init__(
...
 
-    def _get_skt_forecaster_args(self, args: dict, min_length: int, inferred_period: int = 1):
-        """Get arguments that will be passed to the sktime forecaster.
+    def _get_sktime_forecaster_init_args(
+        self, model_params: Dict[str, Any], min_length: int, inferred_period: int = 1
+    ):
+        """Get arguments that will be passed to the sktime forecaster at initialization.
 
-        These may be modified based on the length of training series and inferred seasonality.
+        model_params may be modified in-place based on the length of training series and inferred seasonality.
```
canerturkmen, line 78: oh, I missed this before. I believe this method should guarantee not to modify _in-place_. I also see in other places we `pop` from this dictionary. 

More generally, I would treat all arguments as `const` (in other languages) unless there's a strong reason not to do so. what do you think?

canerturkmen: you'll see in other places in the code we very often do `copy.deepcopy` of hyperparameter dicts to avoid this.

shchur: Agreed, the new version fixes this. Now we call `self._get_model_params().copy()` inside `_get_sktime_forecaster_init_args` (instead of passing a copy of `model_params` to this function). No more in-place modifications :)

timeseries/tests/unittests/models/test_sktime.py
```diff
@@ -177,32 +127,78 @@ def test_when_predict_called_with_test_data_then_predictor_inference_correct(
```
shchur, line 128: Added new tests for `sktime` models

timeseries/src/autogluon/timeseries/models/sktime/abstract_sktime.py
```diff
@@ -67,21 +67,27 @@ def __init__(
...
-    def _get_skt_forecaster(self, sp: Optional[int] = None) -> BaseForecaster:
-        """Return the sktime forecaster object for the model"""
-        params = self._get_model_params().copy()
-        if sp is not None and sp > 1:
-            params["sp"] = sp
-            params["seasonal"] = "add"
+    def _get_sktime_forecaster_init_args(self, min_length: int, inferred_period: int = 1):
+        """Get arguments that will be passed to the sktime forecaster at initialization."""
+        return self._get_model_params().copy()
+
```
shchur, line 76: We now call `self._get_model_params().copy()` inside `_get_sktime_forecaster_init_args`, so no more in-place modifications.

canerturkmen: Just nit that this results in a shallow copy, but should be good enough for us 💯

timeseries/tests/unittests/models/test_sktime.py
```diff
@@ -126,32 +127,37 @@ def test_when_predict_called_with_test_data_then_predictor_inference_correct(
...
+        ("B", 100, 5, False),
+        ("M", 100, 12, False),
+        ("H", 5, 1, True),
+        ("2H", 5, 1, True),
+        ("B", 5, 1, True),
+        ("M", 5, 1, True),
     ],
 )
-def test_when_fit_called_with_then_seasonality_period_set_correctly(
+def test_seasonal_period_is_inferred_correctly_when_set_to_none(
```
canerturkmen, line 144: we tried (although probably not always succeeded) to follow the [GWT](https://en.wikipedia.org/wiki/Given-When-Then) convention when naming test cases, sometimes omitting the G.

timeseries/tests/unittests/models/test_sktime.py
```diff
@@ -177,32 +127,78 @@ def test_when_predict_called_with_test_data_then_predictor_inference_correct(
...
+        hyperparameters=hyperparameters,
+    )
+
+    train_data = get_data_frame_with_item_index(
+        ["A", "B", "C"],
+        data_length=ts_length,
+        freq=freqstr,
+    )
+
+    with caplog.at_level(logging.WARNING):
```
canerturkmen, line 169: cool!

timeseries/tests/unittests/models/test_sktime.py
```diff
@@ -177,32 +127,78 @@ def test_when_predict_called_with_test_data_then_predictor_inference_correct(
...
+        model.fit(train_data=train_data)
```
canerturkmen, line 170: perhaps make explicit that `fail_if_misconfigured=False`?

timeseries/tests/unittests/models/test_sktime.py
```diff
@@ -213,4 +209,46 @@ def test_when_fit_called_with_then_seasonality_period_set_correctly(
...
+
+    train_data = get_data_frame_with_item_index(
+        ["A", "B", "C"],
+        data_length=40,
+        freq="H",
+    )
+
+    with pytest.raises(ValueError):
+        model.fit(train_data=train_data)
+        raise Exception("Model should have failed because train_data too short and fail_if_misconfigured = True")
```
canerturkmen, line 237: very small nit: a little more correct semantically:
```
pytest.fail("model should have ...")
```

canerturkmen (approved):
LGTM. Thanks!

PR #2003, opened 2022-08-01
Enable isort in AutoMM

zhiqiangdon:
*Issue #, if available:*

1. Enable isort in AutoMM.
2. To be consistent with timeseries, black all files in AutoMM.


yinweisu (approved):
LGTM! Thanks!
Should we also update the README to include instruction on isort?

pyproject.toml
```diff
@@ -17,6 +17,24 @@ known_third_party = [
...
+    "scikit-image",
+    "smart_open",
+    "pytorch_lightning",
+    "torchmetrics",
+    "transformers",
+    "nptyping",
+    "omegaconf",
+    "pytorch-metric-learning",
+    "nlpaug",
+    "nltk",
```
sxjscience, line 37: @yinweisu Is it safe to add these new dependencies?

sxjscience (approved):
LGTM! One question about additional dependencies to pyproject

bryanyzhu (approved):
LGTM, thanks.

pyproject.toml
yinweisu: Not too familiar with it, but seems it's only used by isort. So should be fine

FANGAreNotGnu:
LGTM! Thanks!

PR #2013, opened 2022-08-04
Multi-window backtesting for TimeSeriesPredictor

shchur:
Currently, if no `tuning_data` is passed to `autogluon.timeseries.TimeSeriesPredictor.fit`, we use the last `prediction_length` timesteps of `train_data` as hold-out validation set. This PR enables new train / val splitting strategies for timeseries datasets, such as multi-window backtesting.

- New abstract class `autogluon.timeseries.splitter.AbstractTimeSeriesSplitter` that encapsulates the train / val split logic in its `split` method.
- Two currently supported strategies:
  - `LastWindowSplitter` - reserve last `prediction_length` steps (default, same as before)
  - `SlidingWindowSplitter` - reserve multiple windows of length `prediction_length` (potentially overlapping) for validation
- `autogluon.timeseries.TimeSeriesPredictor` now accepts an optional argument `splitter`.
- `tuning_data.index` is now different from `train_data.index` (a suffix `_[{start}:{end}]` is appended to show how the original series was split). We do this because a single item in `train_data` may now correspond to multiple items in `tuning_data`.  This shouldn't affect users since `tuning_data` is not exposed in the public API.

----
Updates:
- `autogluon.timeseries.dataset.ts_dataframe.TimeSeriesDataFrame` now makes sure that the provided `item_id` index either has all string or all integer types.

### DIFF
--- a/timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
+++ b/timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
@@ -208,11 +208,13 @@ class TimeSeriesDataFrame(pd.DataFrame):
         if df[TIMESTAMP].isnull().any():
             raise ValueError(f"`{TIMESTAMP}` column can not have nan")
         if not df[TIMESTAMP].dtype == "datetime64[ns]":
-            raise ValueError(f"for {TIMESTAMP}, the only pandas dtype allowed is ‘datetime64[ns]’.")
-
-        # TODO: check if time series are irregularly sampled. this check was removed as
-        # TODO: pandas is inconsistent in identifying freq when period-end timestamps
-        # TODO: are provided.
+            raise ValueError(f"for {TIMESTAMP}, the only pandas dtype allowed is `datetime64[ns]`.")
+        item_id_column = df[ITEMID]
+        # workaround for pd.api.types.is_string_dtype issue https://github.com/pandas-dev/pandas/issues/15585
+        item_id_is_string = (item_id_column == item_id_column.astype(str)).all()
+        item_id_is_int = pd.api.types.is_integer_dtype(item_id_column)
+        if not (item_id_is_string or item_id_is_int):
+            raise ValueError(f"all entries in column `{ITEMID}` must be of integer or string dtype")
 
     @classmethod
     def _validate_multi_index_data_frame(cls, data: pd.DataFrame):
@@ -229,9 +231,15 @@ class TimeSeriesDataFrame(pd.DataFrame):
         if not isinstance(data.index, pd.MultiIndex):
             raise ValueError(f"data must have pd.MultiIndex, got {type(data.index)}")
         if not data.index.dtypes.array[1] == "datetime64[ns]":
-            raise ValueError(f"for {TIMESTAMP}, the only pandas dtype allowed is ‘datetime64[ns]’.")
+            raise ValueError(f"for {TIMESTAMP}, the only pandas dtype allowed is `datetime64[ns]`.")
         if not data.index.names == (f"{ITEMID}", f"{TIMESTAMP}"):
             raise ValueError(f"data must have index names as ('{ITEMID}', '{TIMESTAMP}'), got {data.index.names}")
+        item_id_index = data.index.get_level_values(level=ITEMID)
+        # workaround for pd.api.types.is_string_dtype issue https://github.com/pandas-dev/pandas/issues/15585
+        item_id_is_string = (item_id_index == item_id_index.astype(str)).all()
+        item_id_is_int = pd.api.types.is_integer_dtype(item_id_index)
+        if not (item_id_is_string or item_id_is_int):
+            raise ValueError(f"all entries in index `{ITEMID}` must be of integer or string dtype")
 
     @classmethod
     def _construct_pandas_frame_from_iterable_dataset(cls, iterable_dataset: Iterable) -> pd.DataFrame:
--- a/timeseries/src/autogluon/timeseries/splitter.py
+++ b/timeseries/src/autogluon/timeseries/splitter.py
@@ -70,11 +70,11 @@ def append_suffix_to_item_id(ts_dataframe: TimeSeriesDataFrame, suffix: str) -> 
 class MultiWindowSplitter(AbstractTimeSeriesSplitter):
     """Slide window from the end of each series to generate validation series.
 
-    The first valdation series contains the entire series (i.e. the last `prediction_length` elements are used for
-    computing the validation score). The end of each following validation series is moved `prediction_length - overlap`
-    steps to the left.
+    The first valdation series contains the entire series (i.e. the last ``prediction_length`` elements are used for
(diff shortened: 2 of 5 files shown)


timeseries/src/autogluon/timeseries/splitter.py
```diff
@@ -0,0 +1,192 @@
...
+    ) -> Tuple[TimeSeriesDataFrame, TimeSeriesDataFrame]:
+        """Split each series in the input dataset into one train and potentially multiple validation series.
+
+        The `item_id` of each validation series is a string of format `f"{idx}_[{start}:{end}]". Here `idx` is
+        the `item_id` of the input series, and `start` and `end` correspond to the slice parameters used to generate
+        the validation series.
+
+        For example, the series `val_data.loc["0_[None:-10]"]` in the validation dataset can equivalently be
+        obtained as `ts_dataframe.loc[0][None:-10]`.
+
```
shchur, line 25: One problem with this approach is that we cannot determine if the original `item_id` was an `int` or a `str` anymore after obtaining the new ID as `f"{idx}_[{start}:{end}]`. 

This may be slightly problematic if, for example, we at some point want to retrieve the respective local model trained on `train_data` and score it on a sequence from `tuning_data` with the new index.

The only alternatives I could think of are
1. Converting the `item_id` index of `train_data` to `str` as well
2. Adding an additional index level that denotes the split (in addition to the `item_id` and `timestamp` levels)

Both of these, however, look really dangerous and can have side effects.

timeseries/src/autogluon/timeseries/splitter.py
```diff
@@ -0,0 +1,192 @@
...
+
+def append_suffix_to_item_id(ts_dataframe: TimeSeriesDataFrame, suffix: str) -> TimeSeriesDataFrame:
+    """Append a suffix to each item_id in a TimeSeriesDataFrame."""
+
+    def add_suffix(multiindex_element):
+        item_id, timestamp = multiindex_element
+        return (f"{item_id}{suffix}", timestamp)
+
+    result = ts_dataframe.copy(deep=False)
```
shchur, line 62: Something bad may happen if a model tries to modify the data in-place. Should we 
1. hope that this never happens, or
2. be extra safe, do a deep copy and waste memory?

timeseries/src/autogluon/timeseries/splitter.py
```diff
@@ -0,0 +1,192 @@
...
+        for window_idx in range(1, self.num_windows):
+            ts_dataframe = ts_dataframe.slice_by_timestep(slice(None, -step_size))
+            total_offset = step_size * window_idx
+            next_val_dataframe = append_suffix_to_item_id(ts_dataframe, f"_[None:{-total_offset}]")
+            validation_dataframes.append(next_val_dataframe)
+
+        train_data = ts_dataframe.slice_by_timestep(slice(None, -prediction_length))
+
+        return train_data, pd.concat(validation_dataframes)
+
```
shchur, line 183: At some point after benchmarking the current version I will implement an `AutomaticSplitter` that determines `num_windows` and `overlap` based on some heuristic.

timeseries/src/autogluon/timeseries/splitter.py
```diff
@@ -0,0 +1,192 @@
...
+            )
+        step_size = prediction_length - self.overlap
+
+        length_per_series = ts_dataframe.index.get_level_values(0).value_counts(sort=False)
+        num_total_validation_steps = prediction_length + step_size * (self.num_windows - 1)
+        num_too_short_series = (length_per_series <= num_total_validation_steps).sum()
+        if num_too_short_series > 0:
+            logger.warning(f"{num_too_short_series} are too short and won't appear in the training set")
+        if num_too_short_series == ts_dataframe.num_items:
+            raise ValueError(f"{self.name} produced an empty training set since all sequences are too short")
```
shchur, line 170: Does it make sense to completely remove the series that are too short to produce `num_windows` many prediction intervals (as current version of the code does), or maybe we should make sure that, e.g., at least 50% of original series's length is preserved for training?

canerturkmen: I'm OK with this logic, but I can't see where the too short time series disappear. Does `slice_by_timestep` handle this gracefully?

timeseries/tests/unittests/test_splitter.py
```diff
@@ -0,0 +1,104 @@
...
+    num_total_validation_steps = num_windows * (prediction_length - overlap) + overlap
+
+    # Training series are as long as expected
+    for item_id in train_data.iter_items():
+        new_length = len(train_data.loc[item_id])
+        expected_length = original_lengths.loc[item_id] - num_total_validation_steps
+        assert expected_length == new_length
+
+    # Validation series are as long as implied by their item_id
+    for new_item_id in val_data.iter_items():
```
canerturkmen, line 66: we can follow one assert per test here?

canerturkmen (commented):
looks great in general! just a few minor comments.

timeseries/src/autogluon/timeseries/splitter.py
shchur: I guess this shouldn't be a problem anymore with the new implementation for local models that we discussed.

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
```diff
@@ -229,9 +231,15 @@ def _validate_multi_index_data_frame(cls, data: pd.DataFrame):
...
-            raise ValueError(f"for {TIMESTAMP}, the only pandas dtype allowed is ‘datetime64[ns]’.")
+            raise ValueError(f"for {TIMESTAMP}, the only pandas dtype allowed is `datetime64[ns]`.")
         if not data.index.names == (f"{ITEMID}", f"{TIMESTAMP}"):
             raise ValueError(f"data must have index names as ('{ITEMID}', '{TIMESTAMP}'), got {data.index.names}")
+        item_id_index = data.index.get_level_values(level=ITEMID)
+        # workaround for pd.api.types.is_string_dtype issue https://github.com/pandas-dev/pandas/issues/15585
+        item_id_is_string = (item_id_index == item_id_index.astype(str)).all()
+        item_id_is_int = pd.api.types.is_integer_dtype(item_id_index)
+        if not (item_id_is_string or item_id_is_int):
+            raise ValueError(f"all entries in index `{ITEMID}` must be of integer or string dtype")
```
shchur, line 242: We now always ensure that the `ITEMID` columns of the index either contains all integers or all strings. This will avoid potential problems, such as collision of two time series with `item_id`s `0` (int) and `"0"` (str).

shchur:
Thanks @canerturkmen! I've addressed the comments and additionally added a check to `TimeSeriesDataFrame` that ensures that the `item_id` index contains only integer or only string values. This should prevent us from getting duplicate `item_id`s in the validation set.

canerturkmen (approved):
looks great! thanks!

PR #2018, opened 2022-08-06
Fix single label error

zhiqiangdon:
Training encounters error if the label column has only 1 label class. Since training with 1 label is meaningless, we can raise error if detecting so.

### DIFF
--- a/multimodal/src/autogluon/multimodal/data/infer_types.py
+++ b/multimodal/src/autogluon/multimodal/data/infer_types.py
@@ -345,9 +345,9 @@ def infer_label_column_type_by_problem_type(
             check_missing_values(data=data, column_name=col_name, split="training")
         if valid_data is not None:
             check_missing_values(data=valid_data, column_name=col_name, split="validation")
-        if col_name in column_types and column_types[col_name] == NULL:
-            warnings.warn(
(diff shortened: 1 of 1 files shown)


multimodal/src/autogluon/multimodal/data/infer_types.py
```diff
@@ -345,9 +345,9 @@ def infer_label_column_type_by_problem_type(
             check_missing_values(data=data, column_name=col_name, split="training")
         if valid_data is not None:
             check_missing_values(data=valid_data, column_name=col_name, split="validation")
-        if col_name in column_types and column_types[col_name] == NULL:
```
yinweisu, line 348: Is this check happening somewhere else? Wonder why removing it

zhiqiangdon: We assert `col_name in column_types` in Line 342.

yinweisu (approved):
LGTM! Thanks!

FANGAreNotGnu:
While training sentence embedding, we may want to train with all sentence pairs as positive data and sample negatives in batch or from memory bank. We may need some modification when we have the codes for that. Currently LGTM, thanks!

FANGAreNotGnu (approved):

PR #2020, opened 2022-08-08
Speed up TimeSeriesDataFrame.slice_by_timestep

shchur:
Several changes to `autogluon.timeseries.dataset.ts_dataframe.TimeSeriesDataFrame`
- Speed up `slice_by_timestep` - use a Boolean mask instead of slicing each series individually in a for loop
  - Train / val split for the `m4_quarterly` dataset (24000 time series) is now ~1000x faster (used to be 178 seconds, now 0.15 seconds).
- Add method `num_timesteps_per_item` that returns the length of each individual time series in the dataset.
- Slightly cleaner code for `_item_index`.
- Refer to index levels using their names `ITEMID` and `TIMESTAMP` (instead of `0` and `1`).

### DIFF
--- a/timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
+++ b/timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
@@ -120,7 +120,7 @@ class TimeSeriesDataFrame(pd.DataFrame):
 
     @property
     def _item_index(self) -> pd.Index:
-        return self.index.get_level_values(0).unique()
+        return self.index.unique(level=ITEMID)
 
     @property
     def static_features(self):
@@ -172,6 +172,9 @@ class TimeSeriesDataFrame(pd.DataFrame):
     @property
     def num_items(self):
         return len(self._item_index)
+
+    def num_timesteps_per_item(self) -> pd.Series:
+        return self.groupby(level=ITEMID).size()
 
     @classmethod
     def _validate_iterable(cls, data: Iterable):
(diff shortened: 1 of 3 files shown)


timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
```diff
@@ -110,7 +111,7 @@ def _constructor(self) -> Type[TimeSeriesDataFrame]:
 
     @property
     def _item_index(self) -> pd.Index:
-        return self.index.get_level_values(0).unique()
+        return self.index.unique(level=ITEMID)
```
canerturkmen, line 114: cool!

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
```diff
@@ -400,13 +404,15 @@ def slice_by_timestep(self, time_step_slice: slice) -> TimeSeriesDataFrame:
...
-        return self.__class__(pd.concat(slices), static_features=self.static_features)
+        num_timesteps_per_item = self.num_timesteps_per_item()
+        # Create a boolean index that selects the correct slice in each timeseries
+        boolean_indicators = []
+        for length in num_timesteps_per_item:
+            indicator = np.zeros(length, dtype=bool)
+            indicator[time_step_slice] = True
+            boolean_indicators.append(indicator)
+        index = np.concatenate(boolean_indicators)
+        return self.__class__(self.loc[index].copy(), static_features=self.static_features)
```
canerturkmen, line 415: nit: it's very surprising `loc` works here and it does feel a little wobbly. I'd leave it in for now, but it felt a little too much of overloading on pandas part.

canerturkmen (approved):
very cool! thanks for this. happy to merge after rebasing / CI.

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
shchur: Changed this to `self[index].copy()`, I agree that this makes more sense

Issue #2022, opened 2022-08-08
Passing a non string path can cause unexpected behavior


geoalgo:
### Describe the issue linked to the documentation

`TabularPredictor` takes an input `path` that must be of type `str`.

https://github.com/awslabs/autogluon/blob/1b4529045ac1ba8efc4963295fc24106cba4949e/tabular/src/autogluon/tabular/predictor/predictor.py#L85

However, it also accepts other types than string, for instance the following correctly runs:

```python
import tempfile
import numpy as np
import pandas as pd
from autogluon.tabular import TabularPredictor

df = pd.DataFrame(np.random.uniform(size=(10, 2)), columns=["x", "y"])
path = tempfile.TemporaryDirectory()
model = TabularPredictor(
    label="y",
    path=path,
    problem_type="regression",
).fit(df, time_limit=10)
```

but it writes into the local folder `./TemporaryDirectory '/var/folders/9k/n2mtln3x1y579hv8xklj54_d6_m_l_/T/tmp4bls6w0r'>/)` which is not what is expected in this case.

### Suggest a potential alternative/fix

To avoid such type of issues, would you be OK for you to either:

1. add the assert `assert isinstance(path, str)` in setup_outputdir?
2. add the assert `assert isinstance(path, (str, pathlib.Path))` in setup_outputdir and update the doc to mention that either type is supported?

(the second option is probably the best given that it is likely that many people are passing paths from `pathlib` even though the type should be `str` according to the doc)

If you think that one of the two option makes sense, I am happy to open a small PR, if not feel free to close the issue.

Innixma:
Hey @geoalgo, good catch! Yes I think option 2 would be best. Happy to review a PR if you'd like to send one.

Innixma:
Resolved in #2026, thanks @geoalgo for the contribution!

PR #2026, opened 2022-08-09
Assert that path have the right types and test path logic

geoalgo:
*Issue #, if available:* https://github.com/awslabs/autogluon/issues/2022

*Description of changes:* 
* assert that types of path are `str` or `pathlib.Path` to avoid unwanted conversion from weird types to string
* add unit-tests to check the path logic
* update doc to mention that pathlib.Path are supported


Innixma (approved):
Code looks great, thanks for the contribution!

PR #2030, opened 2022-08-10
[Cloud] MultiModalCloud Predictor + Multimodal Support for TabularCloudPredictor

yinweisu:
*Issue #, if available:*

*Description of changes:*
* This PR adds MultiModalCloudPredictor and adds multimodal support for TabularCloudPredictor
  * Training is straight forward as we can directly upload a tarball containing images and provide that to the training container similar to ImageCloudPredictor.
  * Main idea behind inference is converting images to bytes, then to base64 encoding, then to utf-8 to get string representation and embed into `pandas.DataFrame`. In the inference container we revert this process to recover the images and save to disk.
  * Some observation on the overhead on a g4dn.12x.large machine but with only 1 gpu enabled(Multimodal Predictor has trouble with multi-gpu training) fitted on petfinder dataset:
    * Inference on 1 row: 4.5 seconds vs 4.63 seconds
    * Inference on 10 rows: 4.77 seconds vs 5.93 seconds
    * Inference on 100 rows: 6.36 seconds vs 16.79 seconds
    * It is observed that as the number of rows increases, the overhead becomes more significant. Therefore, added a warning to batch inference to let the user know.
* Also moved unit tests and predictors into separate files for better readability.

### MultiModalCloudPredictor
* Supports text only modality.
* Supports image only modality.
  * Similar to ImageCloudPredictor, user needs to provide `image_path`, which is a s3 path to image tarball, a local path to the image tarball, or a local path to the image folder, as an argument to `fit()` call
  * Similar to ImageCloudPredictor, user needs to provide `test_data_image_column` if they provided a `pandas.DataFrame` as the test data. Otherwise, they can provide a local path to a single image file or a list of local paths to images files.
  * One difference here is that the user needs to provide `image_modality_only` to `predict()` call for batch inference if they want to directly provide a folder containing images. Otherwise, they must provide `pandas.DataFrame` or a path to csv files and specify `test_data_image_column`
* Supports for tabular+text modality.
* Supports for tabular+text+image modality. 
  * User needs to provide `image_path`, which is a s3 path to image tarball, a local path to the image tarball, or a local path to the image folder, as an argument to `fit()` call
  * For realtime inference and batch inference, test data can be a `pandas.DataFrame`, a local path or a s3 path to a csv file. And the user needs to provide `test_data_image_column`
* Added unit tests

### TabularCloudPredictor with Multimodality
* Supports tabular+text modality.
* Supports tabular+text+image modality.
  * User needs to provide `image_path`, which is a s3 path to image tarball, a local path to the image tarball, or a local path to the image folder, as an argument to `fit()` call
  * For realtime inference and batch inference, test data can be a `pandas.DataFrame`, a local path or a s3 path to a csv file. And the user needs to provide `test_data_image_column`
* Added unit tests


cloud/src/autogluon/cloud/predictor/multimodal_cloud_predictor.py
```diff
@@ -0,0 +1,143 @@
...
+                You need to specify `test_data_image_column`, and make sure the image column contains relative path to the image.
+            When predicting with only images:
+                User has to specify `image_modality_only` for this mode.
+                Can be a local path or a s3 path to a directory containing the images.
+        test_data_image_column: Optional(str)
+            If test_data involves image modality, you must specify the column name corresponding to image paths.
+            Images have to live in the same directory specified by the column.
+        image_modality_only: Bool
+            Whether the test_data only contains image modality.
+            If so, test data can be a local path or a s3 path to a directory containing the images.
```
zhiqiangdon, line 119: Could we infer `image_modality_only` instead of letting users specify it? If we detect a local folder or a s3 bucket of images, we may say it is image only. If it's a s3 bucket, we may need to download the files anyway.

cloud/src/autogluon/cloud/scripts/multimodal_serve.py
```diff
@@ -0,0 +1,150 @@
...
+    elif input_content_type == "application/jsonl":
+        buf = StringIO(request_body)
+        data = pd.read_json(buf, orient='records', lines=True)
+
+    elif input_content_type == "application/x-npy":
+        buf = BytesIO(request_body)
+        data = np.load(buf, allow_pickle=True)
+        image_paths = []
+        for i, bytes in enumerate(data):
+            im = Image.open(BytesIO(base64.b64decode(bytes, validate=True)))
```
gradientsky, line 78: consider using `b85encode` instead - it is more efficient encoding

cloud/src/autogluon/cloud/scripts/script_manager.py
```diff
@@ -19,22 +19,24 @@ class ScriptManager:
...
-        assert predictor_type in ['tabular', 'text', 'image']
+        assert predictor_type in ['tabular', 'text', 'image', 'multimodal']
         # tabular, text and image share the same training script
         return TRAIN_SCRIPT_PATH
 
     @classmethod
     def get_serve_script(cls, predictor_type, framework_version):
-        assert predictor_type in ['tabular', 'text', 'image']
+        assert predictor_type in ['tabular', 'text', 'image', 'multimodal']
         if predictor_type == 'text' and LooseVersion(framework_version) >= LooseVersion('0.4'):
```
gradientsky, line 40: Sanity check: is `0.4` correct version here? Should it be 0.5?

cloud/tests/unittests/test_multimodal_cloud_predictor.py
```diff
@@ -0,0 +1,155 @@
...
+from utils import (
+    _prepare_data,
+    _test_functionality
+)
+
+
+# @pytest.mark.cloud
+# def test_multimodal_text_only():
+#     train_data = 'text_train.csv'
+#     tune_data = 'text_tune.csv'
```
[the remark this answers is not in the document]
yinweisu: Oops, this is actually a mistake. Commented this test while addressing Zhiqiang's comment

cloud/src/autogluon/cloud/scripts/script_manager.py
yinweisu: Text switched backend to multimodal starting 0.4. 0.5 is where we updated the namespace

zhiqiangdon (approved):
Thanks!

PR #2031, opened 2022-08-11
Optimize LightGBM DART retraining

Innixma:
*Issue #, if available:*

*Description of changes:*
- Optimize LightGBM DART retraining by not calculating stopping metric scores on val data on each iteration, since we aren't using early stopping and are training to a fixed number of iterations.


yinweisu (approved):
LGTM!

PR #1809, opened 2022-06-13
Add LoRA as efficient finetuning strategy

Raldir:
Added the option to apply LoRA (https://arxiv.org/pdf/2106.09685.pdf) to Huggingface text model as an additional option for efficient fine-tuning, see PR #1666. New PR based on #1780 for easier merge. Code is largely based on https://github.com/microsoft/LoRA/blob/main/loralib/layers.py

- Added flag to optimizer `optimization.efficient_finetuned`: `lora`, `lora_bias` and `lora_norm`. The first fine-tunes only loRA weights, the second updates both LoRA weights and all biases of the transformer (LoRA + bit_fit), and the third updates LoRA, all biases and normalization terms (LoRA + norm_fit).
- LoRA uses three hyperparameters. These are also set through optimization config. While Lora `r` and `alpha` are quite robust and likely do not need much tuning, the `filter` parameter restricts which linear layers LoRA should be injected into. The original paper indicates that Query and Value attention weights work best so this is set as the default for now. Depending on use we might need to consider a more sophisticated filter algorithm, e.g. using regular expressions.   
- Currently, when saving the model the entire model is stored, which is unnecessary. TODO: Need to change to solely save fine-tuned weights, also applies to bit_fit and norm_fit.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2032, opened 2022-08-11
Adding T-Few to automm for few-shot learning

Raldir:
Implementation of T-Few ([Liu et al, (2022)](https://arxiv.org/pdf/2205.05638.pdf)) in automm for few-shot classification. T-Few is a recently proposed state-of-the-art few-shot learner for textual data. The model is based on T0, a T5+LM based model that has been trained on multiple tasks using input and target prompts based on the respective dataset and task. Adding T-Few to AutoGluon (AG) is a first step in solving problems in the few-shot regime.

**Usage**:

T-Few can be accessed either through specifying `model.names: ["t_few"]` and using a T0 backbone (e.g. `bigscience/T0_3B`, or directly for few_shot scenarios using a new few_shot preset. Also added to throw a suggestion to use the few_shot preset, if training size is very small.

```python
predictor = MultiModalPredictor(label=label,
                                path=save_path,
                                eval_metric="accuracy",
                                problem_type='classification')
predictor.fit(train_data,
              tuning_data=val_data,
              hyperparameters={
                    "data.templates.preset_templates" : list(dataset.dataset_stash)
              }, # sourceprompt name of dataset
              presets="few_shot_text_classification")
```
Note that T-Few works best when providing expressive input and target prompts. A large collection of templates from [sourceprompt](https://github.com/bigscience-workshop/promptsource) can be accessed, by specifying `preset_templates`, or custom_templates can be written. Specify templates in `config.data.templates`:
```yaml
templates:
  turn_on: True
  num_templates: 10 # The number of templates to sample from uniformly.
  template_length: 2048 # Truncation of jinja template variables
  preset_templates: ["super_glue", "rte"] # Select templates from a dataset
                                          # template collection in the form
                                          # (Dataset, Subset). For full list see data/templates. 
  custom_templates: # Specify your own template in jinja format as well as answer choices for the model to select from.
      # 1:
      #   template: "{{premise}} \n\nQuestion: Does this imply that '{{hypothesis}}'? Yes, no, or maybe? ||| {{answer_choices[label]}}"
      #   answer_choices: "Yes ||| Maybe ||| No"
      # 2:
      #   template: "{{premise}} \n\nQuestion: Are we justified in saying that '{{hypothesis}}'? Yes, no, or maybe? ||| {{answer_choices[label]}}"
      #   answer_choices: "Yes ||| Maybe ||| No"
```
**Contribution**

Major

* Adding T-Few as a prompt-based encoder-decoder architecture for solving classification tasks in a few-shot setting. 
    * Adding T-Few’s prompt based loss terms into AG. 
    * Adding a prompt-engine to process multiple text inputs into a single one, by generalizing SEP as a means of generalizing merging of text, and enable specification of target prompts. 
    * Defining execution pipeline for template-based few-shot learning. 
* Adding (IA)^3 as another parameter-efficient fine-tuning strategy  (see PR #1666 and #1809). Comparable performance to LoRA while updating roughly as many parameters as Bitfit. Allows mixed-task batches. Added variations ia3_bias and ia3_norm.

Minor

* Generalized lora_layers.py to adaptation_layers.py. Future file for other adaptation techniques, such as Hulsby or MAM adapters.
* Added a preset called ‘few_shot_text_classification’, that specifies all relevant hyperparameters.
* Modularized the hardcoded automm_custom_layer specification (introduced in PR #1933) to fine-tune non-pretrained layers, by setting layer with id=0 as being trainable (e.g. `head`, `adapter`, `fusion_mlp`). T_Few is set to start with id=1, as all layers are pre-trained. 
* Modified apply_layerwise_lr_decay. Handling efficient_finetune is now part of (`get_trainable_params_efficient_finetune()`), as the conditionals can now be simply added as a trainable parameters specification.
* Allowing greater customization of where to insert adaptations, such as LoRA or (IA)^3 by adding a hyperparameter `module_filter`. Default is set to consider all (as it was before) but for T-Few only some specific layers are considered. 
* Refactored `DummyLayer()` to models.utils.
* Added Adafactor as a scheduling option, following T-Few. 

Todos:
- T-few backbone is a strong zero-shot system (T0). Enable zero_shot_text_classification with T0 as pipeline.
- Some tasks T-FEW was used for, such as copa, where the task is rather question answering than classification, is not supported with the current integration, which focuses on classification only.
- Entire model is loaded/saved (see issue mentioned above PRs) which leads to a large memory footprint when used with large models such as T-Few. Following this point, repo of T-Few provides pre-trained weights for (IA)^3, which could then be included as well.


multimodal/src/autogluon/multimodal/models/t_few.py
```diff
@@ -0,0 +1,291 @@
...
+    TEXT_VALID_LENGTH,
+)
+from .utils import assign_layer_ids, get_column_features
+
+hf_logging.set_verbosity_error()
+
+logger = logging.getLogger(AUTOMM)
+
+
+class DummyLayer(nn.Module):
```
zhiqiangdon, line 32: Consider moving `DummyLayer` to utils since it is used in both hf_text.py and t_few.py.

multimodal/src/autogluon/multimodal/models/t_few.py
```diff
@@ -0,0 +1,291 @@
...
+    def __init__(self):
+        super().__init__()
+        self.dummy_bias = nn.Parameter(torch.zeros(1, dtype=torch.float32))
+
+    def forward(self, x):
+        return x + self.dummy_bias - self.dummy_bias
+
+
+class TFewModel(nn.Module):
```
zhiqiangdon, line 41: Is it better to inherit from `HFAutoModelForTextPrediction`?

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -437,6 +439,15 @@ def fit(
             provided_problem_type=self._problem_type,
         )
 
+        # Determine data scarcity mode, i.e. a few-shot scenario
+        scarcity_mode = infer_scarcity_mode_by_data_size(
+            df_train=train_data, scarcity_threshold=50
```
bryanyzhu, line 444: Is there a standard for scarcity_threshold? Can users adjust this value after pip install autogluon?

multimodal/src/autogluon/multimodal/models/t_few.py
```diff
@@ -0,0 +1,291 @@
...
+        """
+        super().__init__()
+        logger.debug(f"initializing {checkpoint_name}")
+
+        self.checkpoint_name = checkpoint_name
+        self.num_classes = num_classes
+
+        self.model = AutoModelForSeq2SeqLM.from_pretrained(checkpoint_name)
+        self.tokenizer = AutoTokenizer.from_pretrained(checkpoint_name)
+        self.eos_token = self.tokenizer.eos_token  # TODO: Do not hardcode
```
FANGAreNotGnu, line 95: Is this still a todo?

multimodal/src/autogluon/multimodal/models/t_few.py
```diff
@@ -0,0 +1,291 @@
...
+
+        inputs_embeds = self.model.encoder.embed_tokens(text_token_ids)
+
+        if self.gradient_checkpointing:
+            # FIXME(?) This is a hack! We added a DummyLayer to ensure that the
+            #  gradient checkpointing will assign output layer as require_grad=True
+            #  Reference: https://discuss.pytorch.org/t/checkpoint-with-no-grad-requiring-inputs-problem/19117/9
+            inputs_embeds = self.dummy_layer(inputs_embeds)
+
+        # Foward input through the encoder
```
FANGAreNotGnu, line 198: typo *Forward

multimodal/src/autogluon/multimodal/optimization/utils.py
```diff
@@ -469,12 +473,61 @@ def apply_two_stages_lr(
...
+    """
+    if not trainable_param_names:
+        trainable_param_names = []
+
+    if efficient_finetune == BIT_FIT:
+        trainable_param_names.append(".*bias*.")
+    elif efficient_finetune == NORM_FIT:
+        trainable_param_names.append(".*bias*.")
+        trainable_param_names += norm_param_names
+    elif efficient_finetune in [LORA, IA3]:
```
FANGAreNotGnu, line 505: Do we need to add new efficient finetune options in the tutorial here https://auto.gluon.ai/stable/tutorials/multimodal/customization.html?highlight=efficient_finetune?

multimodal/src/autogluon/multimodal/configs/optimization/adamw.yaml
```diff
@@ -20,16 +21,21 @@ optimization:
...
+  trainable_param_names: # If efficient_finetune is not None, specifies which layers should be trained.
+    - ".*head*."  # regex (re.match(trainable_param_name, name)
+    - ".*fusion_mlp*."
+    - ".*adapter*."
   lora:
-    r: 8
-    alpha: 8
-    filter: # Default fine-tune only query and value attention weights, recommended in https://arxiv.org/abs/2106.09685
+    adaptation_to_modules: # Specify which module (if any) to adapt(e.g. ".*EncDecAttention|.*DenseReluDense"). Default consider all modules in a model (i.e. empty filter).
+    adaptation_to_layers: # Specify which layer to adapt. Default fine-tune only query and value attention weights, recommended in https://arxiv.org/abs/2106.09685
```
lzcemma, line 30: Can we have a reference to all the potential choices with explanation here and only leave default uncommented?

multimodal/src/autogluon/multimodal/predictor.py
Raldir: The few-shot experiment with the largest number of training samples in the t-few paper was 50 and in other work ( mostly in NLP) this seems kind of the limit to be considered few-shot. The check itself does not affect the pipeline, it simply throws a warning that the user might want to consider a few-shot preset, given the data scarcity. Given the current state of the feature, I would not change the mode automatically to few-shot.

multimodal/src/autogluon/multimodal/configs/optimization/adamw.yaml
Raldir: Difficult to list them all as they depend on the specific architecture selected from the zoo of models available on Huggingface.

multimodal/src/autogluon/multimodal/models/t_few.py
sxjscience: We may do it later? This is an encoder-decoder model.

multimodal/src/autogluon/multimodal/predictor.py
sxjscience: I agree. This requires some benchmarking.

multimodal/src/autogluon/multimodal/data/templates.py
```diff
@@ -0,0 +1,785 @@
...
+    "wo": "Wolof",
+    "xh": "Xhosa",
+    "yi": "Yiddish",
+    "yo": "Yoruba",
+    "za": "Zhuang, Chuang",
+    "zu": "Zulu",
+}
+
+
+def download_sourceprompt_templates():
```
sxjscience, line 247: Consider to reuse https://github.com/awslabs/autogluon/blob/efea4c2822d532c6724da54c0d2d87fe39fabfea/multimodal/src/autogluon/multimodal/utils.py#L1706

multimodal/src/autogluon/multimodal/models/t_few.py
sxjscience: However, what the class is doing is very similar to HFAutoModelForTextPrediction. The major  difference is it’s an encoder-decoder model. Thus, it also makes sense to inherit from HFAutoModelForTextPrediction but focus on encoder-decoder models.

multimodal/src/autogluon/multimodal/models/utils.py
```diff
@@ -392,8 +406,50 @@ def get_column_features(
...
+    Parameters
+    ----------
+    model
+        A PyTorch model.
+    filter
+        Apply (IA)^3 only to linear layers filtered by name.
+        If None, (IA)^3 is applied to all linear Layers in module.
+    module_filter
+        Apply (IA)^3 only to modules filtered by name (e.g. ".*EncDecAttention|.*DenseReluDense")
+        If None, (IA)^3 is considered for all modules
```
sxjscience, line 427: nit: add a newline to separate “Returns” section.

multimodal/src/autogluon/multimodal/models/t_few.py
Raldir: Yeah, I am a bit divided on this one too. While it might be fine to inherit from HFAutoModelForTextPrediction given the current functionality, it is an encoder-decoder and as such might be deployed in different tasks/cases, so maybe instead of reverting the inheritance down the road, maybe just keep as is for now?

multimodal/tests/unittests/test_data_augmentation.py
```diff
@@ -51,6 +51,7 @@ def test_mixup():
     hyperparameters = {
         "optimization.max_epochs": 1,
         "optimization.top_k_average_method": BEST,
+        "model.t_few.checkpoint_name": "t5-small",
```
sxjscience, line 54: Is the previous test still runnable without this line?

multimodal/src/autogluon/multimodal/models/t_few.py
sxjscience: May be we use a name like `HFAutoModelForConditionalGeneration` then? I think it is implementing a more advanced feature than just t-few and can be reused.

sxjscience: In addition, we can extend the class to support decoder-only models.

multimodal/src/autogluon/multimodal/configs/optimization/adamw.yaml
```diff
@@ -20,16 +21,21 @@ optimization:
   top_k_average_method: "greedy_soup"  # We support averaging method described in https://arxiv.org/pdf/2203.05482.pdf.
                                        # Currently support "uniform_soup", "greedy_soup", and "best".
   efficient_finetune:  # Can be 'bit_fit' (only finetune bias), 'norm_fit' (finetune the normalization terms + bias terms), lora (LoRA Adaptations only), lora_bias (LoRA Adaptation + bit_fit), lora_norm (LoRA Adaptation + norm_fit), or null
+  trainable_param_names: # If efficient_finetune is not None, specifies which layers should be trained.
```
zhiqiangdon, line 24: 1.  The layers (`head`, `fusion_mlp`, and `adapter`) depends on our model definitions. It may not suitable for users to customize them.
2. The name `trainable_param_names ` looks over-general. It is used only in the efficient finetuning mode, but its name seems indicating more than this, which can cause confusions.

Therefore, I suggest we may need to remove this flag from config.

zhiqiangdon: Actually, if we want to automatically detect such layers, we can check if layer id == 0. Layers with id 0 are usually what we create. You could refer to the `.get_layer_ids()` in each model. This way of detecting trainable layers are more general than matching the names (`head`, `fusion_mlp`, and `adapter`).

multimodal/tests/unittests/test_data_augmentation.py
Raldir: The default t_few model is the 3B one, so the CI cannot handle it. Is that addressing your point?

multimodal/src/autogluon/multimodal/configs/optimization/adamw.yaml
Raldir: I see, agreed. Is there a reason that we assume these layers should always be fine-tuned? In T-Few the LM-head is frozen and in some experiments I've run, freezing the head performs much better (also the head itself adds around 70M parameters in T0-3B). Maybe I can add a binary flag to the config, whether id 0 should be fine-tuned?

zhiqiangdon: If you check `HFAutoModelForTextPrediction` and `MultimodalFusionMLP` definitions, you can find that these layers (`head`, `fusion_mlp`, and `adapter`) are created by us instead of from the pretrained model. Since these layers are randomly initialized, we need to fine-tune them. 

You can use the layer id ==0 to determine which layers to fine-tune. If you don't want to fine-tune some layers, just make their ids >=1.

Raldir: Ah, makes total sense. Sounds good.

sxjscience:
Looks good to me overall! One remaining discussion point is on the name of “TFewModel”, but we can also merge this in first and figure out the name later. I’ll approve after the CI passes.

sxjscience (approved):
LGTM! We can create follow-up PRs to improve certain design choices.

bryanyzhu (approved):
Thanks for the PR, great feature!

FANGAreNotGnu (approved):
LGTM, thanks for the awesome work!

zhiqiangdon (approved):
LGTM. Thanks! 
@Raldir feels free to address the comments of our offline discussion either in this PR or the following.

PR #2033, opened 2022-08-11
Bump sktime to 0.13.1 and roll back presets for time series

shchur:
This PR contains two changes:
1. It bumps the `sktime` dependency to `~=0.13,>=0.13.1`, which will allow us to also bump the `numpy` version.
2. It rolls back the presets to the version that we had in `0.5.2` release of AG. I discovered some problems with high disk/memory usage of the current version of the presets (last modified in #2001 but not included in any release yet). I want to fix these problems before we update the presets to avoid impacting the user experience in a negative way.

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/presets.py
+++ b/timeseries/src/autogluon/timeseries/models/presets.py
@@ -78,7 +78,7 @@ def get_default_hps(key, prediction_length):
             "ARIMA": {
                 "maxiter": 50,
                 "order": (1, 1, 1),
-                "seasonal_order": (1, 0, 0),
+                "seasonal_order": (0, 0, 0),
                 "suppress_warnings": True,
             },
             "SimpleFeedForward": {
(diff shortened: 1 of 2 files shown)


timeseries/src/autogluon/timeseries/models/presets.py
```diff
@@ -116,9 +116,9 @@ def get_default_hps(key, prediction_length):
                 "fail_if_misconfigured": True,
             },
             "ARIMA": {
-                "maxiter": 50,
-                "order": ag.Categorical((1, 1, 1), (2, 0, 1)),
-                "seasonal_order": ag.Categorical((1, 0, 0), (1, 0, 1), (1, 1, 1)),
+                "maxiter": ag.Categorical(50),
```
canerturkmen, line 119: it looks like we're not leaving any hyperparameter space to explore. can we at least have (1, 1, 0), (1, 0, 0) in `order`?

canerturkmen (approved):
minor comment. lgtm. thanks!

timeseries/src/autogluon/timeseries/models/presets.py
shchur: I plan to soon add a reasonable search space for HPO with many more configurations, but that requires taking care of some other issues first. 

My idea was to keep the config exactly as-is to ensure that presets are not changing at all between 0.5.1, 0.5.2 and 0.5.3 (if it gets released soon), and then again for 0.5.4.

canerturkmen:
@gradientsky FYI, this PR also resolves AG-TS blocking upgrading numpy version.

PR #2038, opened 2022-08-15
Fix MultiWindowSplitter for timeseries and use faster presets

shchur:
Minor fixes for problems revealed by the latest benchmarking run for timeseries:
- Keep `_cached_freq` during splitting with `MultiWindowSplitter`
- Fix bug in `get_seasonality`
- Add test for the `validation_splitter` argument in `TimeSeriesPredictor`
- Use presets that don't lead to OOM issues on large datasets with ARIMA (band-aid fix until we implement better local models)

### DIFF
--- a/timeseries/src/autogluon/timeseries/splitter.py
+++ b/timeseries/src/autogluon/timeseries/splitter.py
@@ -181,8 +181,10 @@ class MultiWindowSplitter(AbstractTimeSeriesSplitter):
             validation_dataframes.append(next_val_dataframe)
 
         train_data = ts_dataframe.slice_by_timestep(slice(None, -prediction_length))
+        val_data = pd.concat(validation_dataframes)
+        val_data._cached_freq = train_data._cached_freq
 
-        return train_data, pd.concat(validation_dataframes)
+        return train_data, val_data
 
 
 class LastWindowSplitter(MultiWindowSplitter):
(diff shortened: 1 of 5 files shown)


timeseries/src/autogluon/timeseries/splitter.py
```diff
@@ -181,8 +181,10 @@ def _split(
             validation_dataframes.append(next_val_dataframe)
 
         train_data = ts_dataframe.slice_by_timestep(slice(None, -prediction_length))
+        val_data = pd.concat(validation_dataframes)
```
canerturkmen, line 184: cool. am I understanding correctly that `slice` preserves `_cached_freq` but `concat` doesn't? And how did this problem surface? didn't freq inference work after concatenation?

timeseries/src/autogluon/timeseries/splitter.py
```diff
@@ -181,8 +181,10 @@ def _split(
...
+        val_data._cached_freq = train_data._cached_freq
```
canerturkmen, line 185: anything to test here? can we add a unit test in case this problem reappears?

timeseries/src/autogluon/timeseries/splitter.py
shchur: Yes, `_cached_freq` isn't preserved by concatenation. 

The source of the problem are datasets where some time series are really short. After we "slice off" the last `prediction_length` time steps, freq inference stops working (some series only have 2-3 time steps), so `freq` returns `None` and this led to problems downstream (`get_seasonality` didn't handle the case where `freq is None`).

canerturkmen (approved):
thanks !

PR #2040, opened 2022-08-15
Refactor time series ensemble

canerturkmen:
*Issue #, if available:*

*Description of changes:*
This PR removes duplicated code from the timeseries module for the weighted ensemble, and refactors the ensemble APIs to be simpler and safer. It also adds an option to the predictor (public API change) to disable the ensemble (currently the ensemble is still enabled by default and for all presets), and automatically disables ensemble fitting if only one model was fitted. We also make minor changes to the evaluator for simpler handling of coefficients and `higher_is_better`. 

TODO:

- [x] run the quick benchmark to check for regressions in performance

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/ensemble/greedy_ensemble.py
+++ b/timeseries/src/autogluon/timeseries/models/ensemble/greedy_ensemble.py
@@ -8,6 +8,8 @@ import autogluon.core as ag
 from autogluon.core.models.greedy_ensemble.ensemble_selection import AbstractWeightedEnsemble, EnsembleSelection
 from autogluon.timeseries import TimeSeriesDataFrame
 from autogluon.timeseries.models.abstract import AbstractTimeSeriesModel
+from autogluon.timeseries.evaluator import TimeSeriesEvaluator
+
 
 logger = logging.getLogger(__name__)
 
@@ -16,7 +18,7 @@ class TimeSeriesEnsembleSelection(EnsembleSelection):
     def __init__(
         self,
         ensemble_size: int,
-        metric,
+        metric: TimeSeriesEvaluator,
         problem_type: str = ag.constants.QUANTILE,
         sorted_initialization: bool = False,
         bagging: bool = False,
--- a/timeseries/src/autogluon/timeseries/models/ensemble/greedy_ensemble.py
+++ b/timeseries/src/autogluon/timeseries/models/ensemble/greedy_ensemble.py
@@ -48,8 +48,7 @@ class TimeSeriesEnsembleSelection(EnsembleSelection):
     def _calculate_regret(self, y_true, y_pred_proba, metric, dummy_pred=None, sample_weight=None):  # noqa
         dummy_pred = copy.deepcopy(self.dummy_pred if dummy_pred is None else dummy_pred)
         dummy_pred[list(dummy_pred.columns)] = y_pred_proba
-        sign = 1 if metric.higher_is_better else -1
-        return sign * metric(y_true, dummy_pred)
+        return metric(y_true, dummy_pred) * metric.coefficient
 
 
 class SimpleTimeSeriesWeightedEnsemble(AbstractWeightedEnsemble):
--- a/timeseries/src/autogluon/timeseries/models/ensemble/greedy_ensemble.py
+++ b/timeseries/src/autogluon/timeseries/models/ensemble/greedy_ensemble.py
@@ -44,6 +44,7 @@ class TimeSeriesEnsembleSelection(EnsembleSelection):
             time_limit=time_limit,
             sample_weight=sample_weight,
         )
+        self.dummy_pred = None
 
     def _calculate_regret(self, y_true, y_pred_proba, metric, dummy_pred=None, sample_weight=None):  # noqa
         dummy_pred = copy.deepcopy(self.dummy_pred if dummy_pred is None else dummy_pred)
--- a/timeseries/src/autogluon/timeseries/models/ensemble/greedy_ensemble.py
+++ b/timeseries/src/autogluon/timeseries/models/ensemble/greedy_ensemble.py
@@ -1,11 +1,14 @@
 import copy
 import logging
-import time
 from typing import List
 
 import numpy as np
 
-from autogluon.core.models.greedy_ensemble.ensemble_selection import AbstractWeightedEnsemble, EnsembleSelection
+import autogluon.core as ag
+from autogluon.core.models.greedy_ensemble.ensemble_selection import (
+    AbstractWeightedEnsemble,
+    EnsembleSelection,
+)
 from autogluon.timeseries import TimeSeriesDataFrame
 from autogluon.timeseries.models.abstract import AbstractTimeSeriesModel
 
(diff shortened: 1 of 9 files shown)


timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -572,10 +570,28 @@ def _train_multi(
...
+            if time_limit is not None:
+                time_left_for_ensemble = time_limit - (time.time() - time_start)
+
+            if time_left_for_ensemble is not None and time_left_for_ensemble <= 0:
+                logger.info(
+                    "Not fitting ensemble due to lack of time remaining. "
+                    "Time left: {time_left_for_ensemble:.2f} seconds"
+                )
+            elif len(models_available_for_ensemble) <= 1:
+                logger.info(f"Not fitting ensemble as only {len(models_available_for_ensemble)} models were trained.")
```
shchur, line 585: Maybe have a separate error message for "all models have failed" (`len(models_available_for_ensble) == 0`) and "only 1 model was trained" (`len(models_available_for_ensble) == 1`)?

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -607,59 +625,56 @@ def fit_ensemble(self, val_data, model_names):
...
+        ensemble.fit(
+            predictions=predictions,
+            labels=val_data,
+            time_limit=time_limit,
+        )
+
+        # Ensure we don't have name collisions
+        ensemble_name = "WeightedEnsemble"
+        increment = 1
+        while ensemble_name in self._get_banned_model_names():
```
shchur, line 643: Could you explain the purpose of this code block?

timeseries/src/autogluon/timeseries/models/ensemble/greedy_ensemble.py
```diff
@@ -1,226 +1,107 @@
...
-    ) -> None:
+    def _fit(self, *args, **kwargs) -> None:
         raise NotImplementedError
 
     @property
     def weights(self):
-        return self.model.weights_
+        return self.weighted_ensemble.weights_
+
+    def predict(self, data: Dict[str, TimeSeriesDataFrame], **kwargs):
```
shchur, line 100: The signature of `predict` is different from the signature of `AbstractTimeSeriesModel`. Should we call this function differently (e.g., `predict_ensemble` or `combine_predictions`)?

shchur: Also, could we add a pointer here (or somewhere else) that explains the format of the predictions made by each individual model. As far as I understand, this `TimeSeriesDataFrame` is quite different from the usual dataset format.

timeseries/tests/unittests/test_predictor.py
```diff
@@ -396,3 +395,47 @@ def test_when_passing_magic_string_as_validation_splitter_then_correct_splitter_
 ):
     predictor = TimeSeriesPredictor(validation_splitter=splitter_string)
     assert isinstance(predictor.validation_splitter, expected_splitter_class)
+
+
+def test_given_enable_ensemble_false_when_predictor_called_then_ensemble_is_fitted(temp_model_path):
```
shchur, line 400: enable_ensemble_true?

timeseries/src/autogluon/timeseries/models/ensemble/greedy_ensemble.py
canerturkmen: for the first issue, I made the abstract model method signature more generic for now. I don't think it's a perfect solution but it's also not part of the public API.

timeseries/src/autogluon/timeseries/models/ensemble/greedy_ensemble.py
```diff
@@ -1,226 +1,107 @@
 import copy
 import logging
-import time
-from typing import List
+from typing import Dict, List
 
 import numpy as np
 
+import autogluon.core as ag
```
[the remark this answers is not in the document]
canerturkmen: Is there a performance overhead? (not afaik)

This way QUANTILE namespacing is repeated which is the objective here, to keep reminding ourselves in the code where we're getting the value from.

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -572,10 +570,35 @@ def _train_multi(
...
+                        "no models were successfully trained."
+                        if not models_available_for_ensemble
+                        else "only 1 model was trained."
+                    )
+                )
+            else:
+                try:
+                    model_names_trained.append(
+                        self.fit_ensemble(
+                            val_data=val_data, model_names=models_available_for_ensemble, time_limit=time_limit
```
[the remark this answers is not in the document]
canerturkmen: thanks! 👍

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -572,10 +570,35 @@ def _train_multi(
                 )
 
         if self.enable_ensemble:
-            try:
-                model_names_trained.append(self.fit_ensemble(val_data=val_data, model_names=model_names_trained))
-            except Exception as e:  # noqa
-                logger.error(f"\tEnsemble training failed with error \n{traceback.format_exc()}.")
+            models_available_for_ensemble = self.get_model_names()
```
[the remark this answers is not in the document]
canerturkmen: added logic for "model level" in the model graph so for now we only ensemble models that do not have parents.

timeseries/src/autogluon/timeseries/models/ensemble/greedy_ensemble.py
```diff
@@ -1,226 +1,107 @@
...
+            problem_type=problem_type,
+            sorted_initialization=sorted_initialization,
+            bagging=bagging,
+            tie_breaker=tie_breaker,
+            random_state=random_state,
+            **kwargs,
+        )
+
+    def _fit(self, predictions, labels, time_limit=None, sample_weight=None):
+        self.dummy_pred = copy.deepcopy(predictions[0])
```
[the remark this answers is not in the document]
canerturkmen: This is because `dummy_pred` is introduced only in this class, and the superclass (`EnsembleSelection`) doesn't know to provide `dummy_pred` to this class' `_calculate_regret`. Regarding the memory issue, I'm adding 
```
self.dummy_pred = None
```
to make sure after fitting the instance variable is cleared 👍

Innixma (approved):
LGTM!

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -333,7 +334,7 @@ def save(self) -> None:
 
     def _add_model(
         self,
-        model: Union[AbstractTimeSeriesModel, TimeSeriesEnsembleWrapper],
+        model: AbstractTimeSeriesModel,
         base_models=None,
     ):
```
shchur, line 339: Can we add a docstring (and maybe a type annotation) here to explain the meaning of `base_models`? Is this a list of names of models that `model` depends on?

shchur: Also say that `base_models` should only be provided if `model` is an ensemble model (as checked by the assert).

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -569,10 +598,37 @@ def _train_multi(
...
+            models_available_for_ensemble = self.get_model_names(level=0)
+
+            time_left_for_ensemble = None
+            if time_limit is not None:
+                time_left_for_ensemble = time_limit - (time.time() - time_start)
+
+            if time_left_for_ensemble is not None and time_left_for_ensemble <= 0:
+                logger.info(
+                    "Not fitting ensemble due to lack of time remaining. "
+                    "Time left: {time_left_for_ensemble:.2f} seconds"
```
shchur, line 610: Should be an f-string here?

shchur (approved):

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
canerturkmen: thanks! 👍

PR #2044, opened 2022-08-17
clean up flake and pep8 warnings in timeseries

canerturkmen:
*Issue #, if available:*

*Description of changes:*

Clean up warnings that appear in `test_check_style` run. 

### DIFF
--- a/timeseries/tests/unittests/models/test_sktime.py
+++ b/timeseries/tests/unittests/models/test_sktime.py
@@ -5,21 +5,10 @@ import pandas as pd
 import pytest
 
 from autogluon.timeseries import TimeSeriesDataFrame
-
-try:
-    from sktime.forecasting.arima import ARIMA, AutoARIMA
-    from sktime.forecasting.ets import AutoETS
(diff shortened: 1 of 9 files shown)


timeseries/tests/unittests/models/test_sktime.py
```diff
@@ -5,21 +5,10 @@
 import pytest
 
 from autogluon.timeseries import TimeSeriesDataFrame
-
```
shchur, line 8: To clarify, we're removing this check here since we will anyway import the Sktime models in `timeseries.models.sktime.models` and we would get a failure if that import didn't work, right?

shchur (approved):

timeseries/tests/unittests/models/test_sktime.py
canerturkmen: Hmm, you're right this seems to have been the intention; but now we're treating sktime as an optional dependency in the `setup.py` but a hard dependency in the code. We should fix this in migrating to statsmodels, adding it as a core dependency?

PR #2051, opened 2022-08-19
Fix AutoMM CPU inference

zhiqiangdon:
1. Fix AutoMM CPU inference. Users may customize `config.env.num_gpus` when using GPUs in training. Later, they would encounter an error if they do CPU inference.
2. Make saved `num_gpus`, `precision`, and `strategy` consistent with those used in training.
3. Add the `column_types` property to support the cloud predictor.

### DIFF
--- a/multimodal/src/autogluon/multimodal/utils.py
+++ b/multimodal/src/autogluon/multimodal/utils.py
@@ -2060,6 +2060,11 @@ def compute_num_gpus(config_num_gpus: Union[int, float, List], strategy: str):
         num_gpus = detected_num_gpus
     else:
         num_gpus = min(config_num_gpus, detected_num_gpus)
+        warnings.warn(
+            f"Using the detected GPU number {detected_num_gpus}, "
+            f"smaller than the GPU number {config_num_gpus} in the config.",
+            UserWarning,
+        )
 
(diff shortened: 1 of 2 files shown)


multimodal/src/autogluon/multimodal/utils.py
```diff
@@ -2033,3 +2034,41 @@ def process_save_path(path, resume: Optional[bool] = False, raise_if_exist: Opti
...
+    """
+    config_num_gpus = (
+        math.floor(config_num_gpus) if isinstance(config_num_gpus, (int, float)) else len(config_num_gpus)
+    )
+    detected_num_gpus = torch.cuda.device_count()
+
+    if config_num_gpus < 0:  # In case config_num_gpus is -1, meaning using all gpus.
+        num_gpus = detected_num_gpus
+    else:
+        num_gpus = min(config_num_gpus, detected_num_gpus)
```
yinweisu, line 2062: Maybe give a warning here saying detected less gpu than what's in the config?

yinweisu (approved):
LGTM! Thanks for the fix! Had one tiny comment

multimodal/src/autogluon/multimodal/utils.py
zhiqiangdon: Sure. Added a warning.

sxjscience (approved):
LGTM

PR #2053, opened 2022-08-19
[Cloud] Remove feature meatadata from tabular cloud predictor

yinweisu:
*Issue #, if available:*

*Description of changes:*
* This PR removes FeatureMetadata as a required input to train tabular with image modality as this would force Tabular to be a dependency of Cloud module.
  * Added `image_column` parameter to `fit()` api instead.
* Updated training script to accommodate 0.5.2 container seg fault issue.
* Updated multimodal serving script to include `'image'` as valid `image_path` too.
* Updated serving script to force `as_pandas = True` while doing inference to avoid bugs.
* Fixed an issue in previous tabular multimodality unit tests where the hyperparameter is not correctly set.
  * The updated tests revealed two issues from multimodal module:
    1. Tabular predictor in multimodal setting trained on gpu machine cannot be loaded on cpu machine. Discussed with @zhiqiangdon  and the issue has been addressed in https://github.com/awslabs/autogluon/pull/2051
       1. Try and catched the error and prompt the user on potential solution for now until 0.6 container release
    2. Tabular predictor in multimodal setting CANNOT do batch inference because Text predictor trained under Tabular predictor cannot be saved standalone; therefore, requiring loading from hugging face. However, the batch inference container cannot connect to hugging face correctly. This leads to unable to load the trained model in batch inference.
       1. Added a warning to warn user tabular doesn't support multimodal batch inference yet for now. And synced with @zhiqiangdon  on solution.
       2. Skipped batch inference test for tabular multimodal settings temporarily until 0.6 container release
  

### DIFF
--- a/cloud/src/autogluon/cloud/scripts/train.py
+++ b/cloud/src/autogluon/cloud/scripts/train.py
@@ -99,11 +99,11 @@ if __name__ == "__main__":
 
     train_file = get_input_path(args.training_dir)
     training_data = TabularDataset(train_file)
-    if predictor_type == 'tabular' and 'image_column_name' in config:
+    if predictor_type == 'tabular' and 'image_column' in config:
         feature_metadata = predictor_fit_args.get('feature_metadata', None)
         if feature_metadata is None:
             feature_metadata = FeatureMetadata.from_df(training_data)
-        feature_metadata = feature_metadata.add_special_types({config['image_column_name']: ['image_path']})
+        feature_metadata = feature_metadata.add_special_types({config['image_column']: ['image_path']})
         predictor_fit_args['feature_metadata'] = feature_metadata
 
     tuning_data = None
--- a/cloud/src/autogluon/cloud/scripts/train.py
+++ b/cloud/src/autogluon/cloud/scripts/train.py
@@ -7,7 +7,7 @@ from autogluon.tabular import TabularPredictor, TabularDataset, FeatureMetadata
 from autogluon.vision import ImagePredictor
 from autogluon.text import TextPredictor
 
-# Not following import style because importing autogluon.tet before other module would cause segfault
+# Not following import style because importing autogluon.text before other module would cause segfault
 # https://github.com/awslabs/autogluon/issues/2042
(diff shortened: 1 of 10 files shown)


cloud/src/autogluon/cloud/scripts/train.py
```diff
@@ -3,11 +3,13 @@
...
 import yaml
 
-import autogluon.text
-
 from autogluon.tabular import TabularPredictor, TabularDataset, FeatureMetadata
-from autogluon.text import TextPredictor
 from autogluon.vision import ImagePredictor
+from autogluon.text import TextPredictor
+
+# Not following import style because importing autogluon.tet before other module would cause segfault
```
Innixma, line 10: nit: autogluon.tet typo

cloud/src/autogluon/cloud/scripts/train.py
```diff
@@ -3,11 +3,13 @@
...
+# https://github.com/awslabs/autogluon/issues/2042
+import autogluon.text
```
Innixma, line 12: redundant text import? Don't you import TextPredictor right above this line?

Innixma (approved):
LGTM!

cloud/src/autogluon/cloud/scripts/train.py
yinweisu: So the import of text module itself is for determining how to correctly import MultimodalPredictor as we moved it from text to multimodal module in 0.5 release. The import of TextPredictor sort of just align with other module's import. We can of course only import autogluon.text and use autogluon.text.TextPredictor. Not sure what's the convention

zhiqiangdon (approved):
LGTM. Thanks!

cloud/src/autogluon/cloud/scripts/train.py
Innixma: I'm just confused why we are doing:

```
from autogluon.text import TextPredictor
import autogluon.text
```

Instead of just:

```
from autogluon.text import TextPredictor
```

Issue #2047, opened 2022-08-18
Check Python Version on predictor load


Innixma:
### Describe the issue linked to the documentation

When loading a fitted predictor with a different python version, can raise exceptions that are hard to understand such as 

```
AttributeError: 'DiGraph' object has no attribute '_adj'
```

This is due to differences in packages between python versions.

### Suggest a potential alternative/fix

We should store the python version used to fit AutoGluon as part of the metadata, and then warn the user if they try to load with a different python version.

PR #2054, opened 2022-08-19
Add metadata.json file to check python and package versions

Innixma:
*Issue #, if available:*
#2047

*Description of changes:*
- Added a variety of checks and logs that detect differences in environment between fit time and inference time.
- Raise exception by default if Python minor version differs during predictor load.

Example output (created by manually editing the metadata.json file contents to force mismatches):

```
predictor = TabularPredictor.load(path=path, check_packages=True)
```

```
Found 15 mismatches between original and current metadata:
	WARNING: AutoGluon version mismatch (original=0.5.8, current=0.3.2b20211216)
	WARNING: AutoGluon Python version mismatch (original=3.10, current=3.8)
	WARNING: System mismatch (original=Windows, current=Darwin)
	WARNING: Package version diff 'regex'	(original=2025.11.10, current=2021.11.10)
	WARNING: Package version diff 'attrs'	(original=21.1, current=21.2.0)
	WARNING: Package version diff 'notebook'	(original=12.4.6, current=6.4.6)
	WARNING: Package version diff 'threadpoolctl'	(original=3.0.1, current=3.0.0)
	WARNING: Package version diff 'fastai'	(original=4.1.2, current=2.7.6)
	WARNING: Package version diff 'torch'	(original=5.0.0, current=1.12.0)
	INFO: New package 'dask==2021.11.2'
	INFO: New package 'distributed==2021.11.2'
	INFO: New package 'certifi==2021.10.8'
	INFO: New package 'pytz==2021.3'
	INFO: New package 'setuptools==59.5.0'
	INFO: New package 'cryptography==36.0.0'
  File "/Users/neerick/workspace/virtual/autogluon/lib/python3.8/site-packages/IPython/core/interactiveshell.py", line 3457, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-6-231f1544ec1b>", line 1, in <module>
    a = TabularPredictor.load(path=predictor.path, check_packages=True)
  File "/Users/neerick/workspace/code/autogluon/tabular/src/autogluon/tabular/predictor/predictor.py", line 2954, in load
    raise AssertionError(
AssertionError: Predictor was created on Python version 3.10 but is being loaded with Python version 3.8. Please ensure the versions match to avoid instability. While it is NOT recommended, this error can be bypassed by specifying `require_py_version_match=False`.

```

### DIFF
--- a/common/src/autogluon/common/utils/utils.py
+++ b/common/src/autogluon/common/utils/utils.py
@@ -3,6 +3,9 @@ from datetime import datetime
 
 import logging
 import os
+import platform
+import sys
+from typing import Dict, Any
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


common/src/autogluon/common/utils/utils.py
```diff
@@ -47,3 +50,62 @@ def setup_outputdir(path, warn_if_exist=True, create_dir=True, path_suffix=None)
...
+def compare_autogluon_metadata(*, original: dict, current: dict, check_packages=True) -> list:
+    logs = []
+    og = original
+    cu = current
+    if og['version'] != cu['version']:
+        logs.append((30, f"WARNING: AutoGluon version mismatch (original={og['version']}, current={cu['version']})"))
+    if og['py_version'] != cu['py_version']:
+        logs.append((30, f"WARNING: AutoGluon Python version mismatch (original={og['py_version']}, current={cu['py_version']})"))
+    elif og['py_version_micro'] != cu['py_version_micro']:
+        logs.append((30, f"INFO: AutoGluon Python micro version mismatch (original={og['py_version_micro']}, current={cu['py_version_micro']})"))
```
gradientsky, line 91: I think the warning have sense for 3.7 vs 3.8 (specifically because of serialization issues). Not sure if micro version should be a warning. Do we have any example of incompatibility for these?

common/src/autogluon/common/utils/utils.py
```diff
@@ -47,3 +50,62 @@ def setup_outputdir(path, warn_if_exist=True, create_dir=True, path_suffix=None)
...
+        og_pac = og['packages']
+        cu_pac = cu['packages']
+        for k in og_pac.keys():
+            if k not in cu_pac:
+                logs.append((30, f"WARNING: Missing package '{k}=={og_pac[k]}'"))
+            elif og_pac[k] != cu_pac[k]:
+                logs.append((30, f"WARNING: Package version diff '{k}'\t(original={og_pac[k]}, current={cu_pac[k]})"))
+        for k in cu_pac.keys():
+            if k not in og_pac:
+                logs.append((30, f"INFO: New package '{k}=={cu_pac[k]}'"))
```
gradientsky, line 104: This will produce lots of warnings between models trained on GPU and deployed on CPU. I don't think it's a good idea.

common/src/autogluon/common/utils/utils.py
Innixma: I don't know, currently it is just an info log and not a warning.

common/src/autogluon/common/utils/utils.py
Innixma: This is not enabled by default, it only shows these warnings when check_packages=True (False by default)

Innixma: This is mainly for debugging, if it fails user can load with check_packages=True to see the differences.

common/src/autogluon/common/utils/utils.py
Innixma: If you want I can remove it so it is never logged

gradientsky (approved):

PR #2058, opened 2022-08-22
Improve AutoMM loading checkpoints offline

zhiqiangdon:
1. Change `standalone=True` by default to support the cloud predictors.
2. If `standalone=True`, save only the pretrained model config instead of the model weights. This can significantly reduce the disk usages.

### DIFF
--- a/multimodal/src/autogluon/multimodal/utils.py
+++ b/multimodal/src/autogluon/multimodal/utils.py
@@ -705,6 +705,7 @@ def create_model(
                 prefix=model_name,
                 checkpoint_name=model_config.checkpoint_name,
                 num_classes=num_classes,
+                pretrained=pretrained,
             )
         elif model_name.lower().startswith(TIMM_IMAGE):
             model = TimmAutoModelForImagePrediction(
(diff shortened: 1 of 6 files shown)


multimodal/src/autogluon/multimodal/utils.py
```diff
@@ -901,25 +902,24 @@ def save_pretrained_models(
         model = model.model
     for per_model in model:
         if per_model.prefix.lower().startswith((CLIP, HF_TEXT)):
-            per_model.model.save_pretrained(os.path.join(path, per_model.prefix))
+            per_model.config.save_pretrained(os.path.join(path, per_model.prefix))
```
sxjscience, line 905: For huggingface models, you may also need to save the tokenizer.

zhiqiangdon: The text tokenizers are saved by `save_text_tokenizers()`.

yinweisu (approved):
LGTM! Thanks!

bryanyzhu (approved):
LGTM, thanks. In particular, I like the way to get config and model at the same time. We really need the `config` throughout the process. For some libraries like MMlab series, you can't get config from the checkpoint itself, so it is better to pass the config.

sxjscience (approved):
LGTM

PR #2061, opened 2022-08-22
Fix the gpu inference for object detection

zhiqiangdon:
Fix the gpu inference for object detection.

### DIFF
--- a/multimodal/src/autogluon/multimodal/models/mmdet_image.py
+++ b/multimodal/src/autogluon/multimodal/models/mmdet_image.py
@@ -8,6 +8,8 @@ from torch import nn
 
 try:
     import mmcv
+    from mmcv.ops import RoIPool
+    from mmcv.parallel import scatter
     from mmcv.runner import load_checkpoint
 except ImportError:
     mmcv = None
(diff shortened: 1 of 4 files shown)


multimodal/src/autogluon/multimodal/models/mmdet_image.py
```diff
@@ -121,6 +123,15 @@ def forward(
         data = batch[self.image_key]
         data["img_metas"] = [img_metas.data[0] for img_metas in data["img_metas"]]
         data["img"] = [img.data[0] for img in data["img"]]
+
+        device = next(self.model.parameters()).device  # model device
```
bryanyzhu, line 127: Will this interfere with pytorch-lightening? Also, does this work with DDP on multi-cards or multi-nodes?

bryanyzhu (commented):
Thanks for the fix.

FANGAreNotGnu (approved):
LGTM, thanks!

multimodal/src/autogluon/multimodal/models/mmdet_image.py
zhiqiangdon: I guess it should work with ddp. Each gpu runs this code and gets its device id.

bryanyzhu (approved):
LGTM, thank you.

PR #2067, opened 2022-08-24
Implement wrappers for local models from statsmodels

shchur:
This PR introduces local models (ETS, ARIMA) from the `statsmodels` package to AutoGluonTS. This takes care of a few problems that we currently have when using `sktime`
- Correctly handles datasets with multiple time series of different lengths (`sktime` used to have problems with estimating `cutoff` for such datasets and sometimes produced wrong results).
- High disk usage - models now take a few KB instead of many MB/GB when saved to disk.

Other changes:
- Training and prediction are parallelized over CPU cores with `joblib`
- `statsmodels` models are used as default local models; `sktime` models received a `Sktime` prefix and were removed from the default presets.

Comparison of `statsmodels` and `sktime` versions of ETS and ARIMA on the `artificial_seasonal` dataset (parallelizing over 16 CPU cores)
```
ETSModel (statsmodels):
 - fit time: 10.6 s
 - predict time: 0.5 s
 - mean_wQuantileLoss: -0.15596
 - pickled size: 64 KB

SktimeAutoETSModel (sktime):
 - fit time: 30.6 s
 - predict time: 8.7 s
 - mean_wQuantileLoss: -0.15596
 - pickled size: 4421 KB

ARIMAModel (statsmodels):
 - fit time: 5.4 s
 - predict time: 0.8 s
 - mean_wQuantileLoss: -0.15570
 - pickled size: 25 KB

SktimeARIMAModel (sktime):
 - fit time: 56.8 s
 - predict time: 15.7 s
 - mean_wQuantileLoss: -0.15600
 - pickled size: 468814 KB
```
Note that the original dataset only takes 228 KB on disk, but the fitted ARIMA model from `sktime`/`pmdarima` takes 468 MB of disk space, fit and predict time are both improved ~10x. We get closer to linear scaling with # of cores when using bigger datasets.

image
image

Use attached file [compare_sktime_vs_statsmodels.py.zip](https://github.com/awslabs/autogluon/files/9416815/compare_sktime_vs_statsmodels.py.zip) to reproduce the results.

TODO:
- [x] Add docstrings, make sure hyperparameter names are consistent & defaults are reasonable
- [x] Parallelize training / prediction across cores
- [x] Add tests

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/presets.py
+++ b/timeseries/src/autogluon/timeseries/models/presets.py
@@ -16,8 +16,7 @@ from .gluonts import (
     TransformerModel,
 )
 from .sktime import ARIMAModel, AutoARIMAModel, AutoETSModel
-from .statsmodels import StatsmodelsETSModel, StatsmodelsARIMAModel
-
+from .statsmodels import StatsmodelsARIMAModel, StatsmodelsETSModel
 
 logger = logging.getLogger(__name__)
 
@@ -48,6 +47,8 @@ DEFAULT_MODEL_PRIORITY = dict(
     AutoARIMA=20,
     ARIMA=50,
     AutoETS=60,
+    StatsmodelsARIMA=50,
+    StatsmodelsETS=60,
 )
 DEFAULT_CUSTOM_MODEL_PRIORITY = 0
 MINIMUM_CONTEXT_LENGTH = 10
--- a/timeseries/src/autogluon/timeseries/models/presets.py
+++ b/timeseries/src/autogluon/timeseries/models/presets.py
@@ -15,8 +15,8 @@ from .gluonts import (
     SimpleFeedForwardModel,
     TransformerModel,
 )
-from .sktime import ARIMAModel, AutoARIMAModel, AutoETSModel
-from .statsmodels import StatsmodelsARIMAModel, StatsmodelsETSModel
+from .sktime import SktimeARIMAModel, SktimeAutoARIMAModel, SktimeAutoETSModel
+from .statsmodels import ARIMAModel, ETSModel
 
 logger = logging.getLogger(__name__)
 
@@ -29,11 +29,11 @@ MODEL_TYPES = dict(
     AutoTabular=AutoTabularModel,
     Prophet=ProphetModel,
     Transformer=TransformerModel,
+    SktimeARIMA=SktimeARIMAModel,
+    SktimeAutoARIMA=SktimeAutoARIMAModel,
+    SktimeAutoETS=SktimeAutoETSModel,
+    ETS=ETSModel,
     ARIMA=ARIMAModel,
-    AutoARIMA=AutoARIMAModel,
-    AutoETS=AutoETSModel,
-    StatsmodelsETS=StatsmodelsETSModel,
-    StatsmodelsARIMA=StatsmodelsARIMAModel,
 )
 DEFAULT_MODEL_NAMES = {v: k for k, v in MODEL_TYPES.items()}
 DEFAULT_MODEL_PRIORITY = dict(
@@ -44,11 +44,11 @@ DEFAULT_MODEL_PRIORITY = dict(
     DeepAR=50,
     Prophet=10,
     AutoTabular=10,
-    AutoARIMA=20,
+    SktimeAutoARIMA=20,
+    SktimeARIMA=50,
+    SktimeAutoETS=60,
     ARIMA=50,
-    AutoETS=60,
-    StatsmodelsARIMA=50,
-    StatsmodelsETS=60,
+    ETS=60,
 )
 DEFAULT_CUSTOM_MODEL_PRIORITY = 0
 MINIMUM_CONTEXT_LENGTH = 10
@@ -66,21 +66,20 @@ def get_default_hps(key, prediction_length):
             },
             "Transformer": {"epochs": 10, "num_batches_per_epoch": 10, "context_length": 5},
             "DeepAR": {"epochs": 10, "num_batches_per_epoch": 10, "context_length": 5},
-            "AutoETS": {"maxiter": 20, "seasonal": None},
+            "ETS": {"maxiter": 20, "seasonal": None},
             "ARIMA": {
                 "maxiter": 10,
                 "order": (1, 0, 0),
                 "seasonal_order": (0, 0, 0),
-                "suppress_warnings": True,
             },
         },
         "default": {
-            "StatsmodelsETS": {
+            "ETS": {
                 "maxiter": 200,
                 "trend": "add",
                 "seasonal": "add",
             },
-            "StatsmodelsARIMA": {
+            "ARIMA": {
                 "maxiter": 50,
                 "order": (1, 1, 1),
                 "seasonal_order": (0, 0, 0),
@@ -110,15 +109,15 @@ def get_default_hps(key, prediction_length):
                 "model_dim": ag.Categorical(8, 16, 32),
                 "context_length": context_length,
             },
-            "StatsmodelsETS": {
+            "ETS": {
+                "maxiter": 200,
                 "error": ag.Categorical("add", "mul"),
                 "trend": ag.Categorical("add", "mul", None),
                 "seasonal": ag.Categorical("add", None),
-                "maxiter": 200,
             },
-            "StatsmodelsARIMA": {
+            "ARIMA": {
                 "maxiter": 50,
-                "order": ag.Categorical((2, 0, 1), (2, 1, 1)),
+                "order": ag.Categorical((2, 0, 1), (2, 1, 1), (1, 1, 1)),
                 "seasonal_order": ag.Categorical((0, 0, 0), (1, 0, 1)),
             },
         },
--- a/timeseries/src/autogluon/timeseries/models/presets.py
+++ b/timeseries/src/autogluon/timeseries/models/presets.py
@@ -82,7 +82,7 @@ def get_default_hps(key, prediction_length):
             },
             "StatsmodelsARIMA": {
                 "maxiter": 50,
-                "order": (2, 0, 0),
+                "order": (1, 1, 1),
                 "seasonal_order": (0, 0, 0),
             },
             "SimpleFeedForward": {
--- a/timeseries/src/autogluon/timeseries/models/presets.py
+++ b/timeseries/src/autogluon/timeseries/models/presets.py
@@ -16,6 +16,8 @@ from .gluonts import (
     TransformerModel,
 )
 from .sktime import ARIMAModel, AutoARIMAModel, AutoETSModel
+from .statsmodels import StatsmodelsETSModel, StatsmodelsARIMAModel
+
 
 logger = logging.getLogger(__name__)
 
@@ -31,6 +33,8 @@ MODEL_TYPES = dict(
     ARIMA=ARIMAModel,
     AutoARIMA=AutoARIMAModel,
     AutoETS=AutoETSModel,
+    StatsmodelsETS=StatsmodelsETSModel,
+    StatsmodelsARIMA=StatsmodelsARIMAModel,
 )
 DEFAULT_MODEL_NAMES = {v: k for k, v in MODEL_TYPES.items()}
 DEFAULT_MODEL_PRIORITY = dict(
@@ -70,18 +74,15 @@ def get_default_hps(key, prediction_length):
             },
         },
         "default": {
-            "AutoETS": {
+            "StatsmodelsETS": {
                 "maxiter": 200,
                 "trend": "add",
                 "seasonal": "add",
-                "auto": False,
-                "initialization_method": "estimated",
             },
-            "ARIMA": {
+            "StatsmodelsARIMA": {
                 "maxiter": 50,
-                "order": (1, 0, 0),
+                "order": (2, 0, 0),
                 "seasonal_order": (0, 0, 0),
-                "suppress_warnings": True,
             },
             "SimpleFeedForward": {
                 "context_length": context_length,
@@ -108,21 +109,16 @@ def get_default_hps(key, prediction_length):
                 "model_dim": ag.Categorical(8, 16, 32),
                 "context_length": context_length,
             },
-            "AutoETS": {
+            "StatsmodelsETS": {
                 "error": ag.Categorical("add", "mul"),
-                "trend": ag.Categorical("add", "mul"),
-                "seasonal": ag.Categorical("add", "mul", None),
-                "auto": False,
-                "initialization_method": "estimated",
+                "trend": ag.Categorical("add", "mul", None),
+                "seasonal": ag.Categorical("add", None),
                 "maxiter": 200,
-                "fail_if_misconfigured": True,
             },
-            "ARIMA": {
-                "maxiter": ag.Categorical(50),
-                "order": (1, 0, 0),
-                "seasonal_order": (0, 0, 0),
-                "suppress_warnings": True,
-                "fail_if_misconfigured": True,
+            "StatsmodelsARIMA": {
+                "maxiter": 50,
+                "order": ag.Categorical((2, 0, 1), (2, 1, 1)),
+                "seasonal_order": ag.Categorical((0, 0, 0), (1, 0, 1)),
             },
         },
     }
--- a/timeseries/src/autogluon/timeseries/models/statsmodels/abstract_statsmodels.py
+++ b/timeseries/src/autogluon/timeseries/models/statsmodels/abstract_statsmodels.py
@@ -158,7 +158,6 @@ class AbstractStatsmodelsModel(AbstractTimeSeriesModel):
         model_init_args, model_fit_args = self._update_model_init_and_fit_args(
             default_model_init_args=default_model_init_args, default_model_fit_args=default_model_fit_args
         )
-        print("model_init_args", model_init_args)
 
         # Fit models in parallel
         fit_fn = partial(self._fit_local_model, model_fit_args=model_fit_args, model_init_args=model_init_args)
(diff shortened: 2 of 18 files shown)


timeseries/src/autogluon/timeseries/models/presets.py
```diff
@@ -16,6 +16,7 @@
     TransformerModel,
 )
 from .sktime import ARIMAModel, AutoARIMAModel, AutoETSModel
```
canerturkmen, line 18: if all goes well, can we use this PR to also add guards for `sktime` and dropping it from the hard requirements?

timeseries/src/autogluon/timeseries/models/presets.py
```diff
@@ -31,6 +32,8 @@
     ARIMA=ARIMAModel,
     AutoARIMA=AutoARIMAModel,
     AutoETS=AutoETSModel,
+    StatsmodelsETS=StatsmodelsETSModel,
```
canerturkmen, line 35: How about we remove `Statsmodels` from both class names and aliases here, and remove `ARIMAModel` from sktime entirely?

timeseries/src/autogluon/timeseries/models/statsmodels/abstract_statsmodels.py
```diff
@@ -0,0 +1,143 @@
+import logging
+import re
+from dataclasses import dataclass
+from typing import Any, Dict, List, Optional, Tuple
+
+import pandas as pd
+
+from ...dataset.ts_dataframe import ITEMID, TIMESTAMP, TimeSeriesDataFrame
```
canerturkmen, line 8: should we maybe use this as an opportunity to stop doing relative imports after one level? for example see: ﻿﻿https://google.github.io/styleguide/pyguide.html#224-decision

timeseries/src/autogluon/timeseries/models/statsmodels/abstract_statsmodels.py
```diff
@@ -0,0 +1,143 @@
...
+from ...utils.hashing import hash_ts_dataframe_items
+from ...utils.warning_filters import statsmodels_warning_filter
+from ..abstract import AbstractTimeSeriesModel
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass
```
canerturkmen, line 16: cool!

timeseries/src/autogluon/timeseries/models/statsmodels/abstract_statsmodels.py
```diff
@@ -0,0 +1,143 @@
...
+    def _fit(self, train_data: TimeSeriesDataFrame, time_limit: int = None, **kwargs):
+        train_hash = hash_ts_dataframe_items(train_data)
+        items_to_fit = [item_id for item_id, ts_hash in train_hash.iteritems() if ts_hash not in self._fitted_models]
+        default_model_init_args, default_model_fit_args = self._get_default_model_init_and_fit_args()
+        # TODO: Parallelize fitting
+        with statsmodels_warning_filter():
+            for item_id in items_to_fit:
+                self._fitted_models[train_hash.loc[item_id]] = self._fit_local_model(
+                    timeseries=train_data.loc[item_id],
+                    default_model_init_args=default_model_init_args,
```
canerturkmen, line 92: shouldn't these parameters be called `model_init_args` etc.?

timeseries/src/autogluon/timeseries/models/statsmodels/models.py
```diff
@@ -0,0 +1,116 @@
...
+        "trend",
+        "seasonal",
+        "seasonal_period",
+    ]
+    statsmodels_allowed_fit_args = [
+        "maxiter",
+    ]
+
+    def _fit_local_model(
+        self, timeseries: TimeSeriesDataFrame, default_model_init_args: dict, default_model_fit_args: dict
```
canerturkmen, line 31: kwarg names do not match superclass.

canerturkmen (commented):
Great! My review is WIP, some comments below.

timeseries/src/autogluon/timeseries/models/statsmodels/abstract_statsmodels.py
shchur: Totally agree, will change the imports here and will keep an eye on this when working on future PRs to clean up existing code

timeseries/src/autogluon/timeseries/models/presets.py
shchur: I agree that dropping `Statsmodels` in the model names would be nice, but I'm 100% sure about how we want to handle the changes in presets between different releases.

How about we keep the `Statsmodels` suffix for now, but soon make another PR where we completely remove the `sktime` dependency and change SM model names to `ETS`, `ARIMA`, etc.?

timeseries/src/autogluon/timeseries/models/presets.py
shchur: How about doing this in a separate PR? I think this PR will already be quite large after all the TODOs are completed

timeseries/src/autogluon/timeseries/models/statsmodels/abstract_statsmodels.py
shchur: The logic here is that `default_model_init_args` and `default_model_fit_args` are obtained by splitting the `self._get_model_params()`. These 2 dicts contain the (hyper)parameters exposed to the user, and these might be different from the `model_init_args` that are passed to `statsmodels` inside the `_fit_local_model` function (e.g., we always expose a `seasonal_period` parameter to the user, but internally each model uses a different name).

Do you think this naming `default_model_init_args` and `model_init_args` is reasonable to distinguish the two sets of parameters (exposed vs. internal), or would something else make more sense?

timeseries/src/autogluon/timeseries/models/statsmodels/abstract_statsmodels.py
```diff
@@ -0,0 +1,175 @@
...
+        """Fit a single local model to the time series."""
+        raise NotImplementedError
+
+    def _fit(self, train_data: TimeSeriesDataFrame, time_limit: int = None, **kwargs):
+        # Select timeseries that don't have a fitted local model
+        train_hash = hash_ts_dataframe_items(train_data)
+        items_to_fit = [item_id for item_id, ts_hash in train_hash.iteritems() if ts_hash not in self._fitted_models]
+        timeseries_to_fit = (train_data.loc[item_id][self.target] for item_id in items_to_fit)
+        default_model_init_args, default_model_fit_args = self._get_default_model_init_and_fit_args()
+        # TODO: Should we rather set n_jobs in __init__?
```
canerturkmen, line 113: if this is coming through the hyperparameter system, i.e., we want to expose it to the user, I believe it will have to stay here? 

that said,  instance variables should not be initialized outside of `__init__`. For example if someone pickled the class and unpickled it and tried to use it to predict, then `self.n_jobs` would throw an `AttributeError`. In fact, we do this in AG as well. You could initialize it to `-1` in `__init__` and override it here.

if this case is not covered by a test, we should probably add it for all models (i.e., pickle unpickle and predict with the same item index).

timeseries/src/autogluon/timeseries/models/statsmodels/abstract_statsmodels.py
```diff
@@ -0,0 +1,175 @@
...
+
+    def _fit_local_model(
+        self, timeseries: pd.Series, default_model_init_args: dict, default_model_fit_args: dict
+    ) -> FittedLocalModel:
+        """Fit a single local model to the time series."""
+        raise NotImplementedError
+
+    def _fit(self, train_data: TimeSeriesDataFrame, time_limit: int = None, **kwargs):
+        # Select timeseries that don't have a fitted local model
+        train_hash = hash_ts_dataframe_items(train_data)
```
canerturkmen, line 109: Nit: By our design `_fit` never refits a certain hash. Wouldn't any consumer of this class want to refit a certain hash? Is `hash -> FittedLocalModel` always deterministic for all concrete implementations of this model? If a user wants to refit a hash, do they have some mechanism available to them ? (`_fitted_models` is protected, so subclasses can override this design if they want, but not other dependants).

timeseries/src/autogluon/timeseries/models/statsmodels/abstract_statsmodels.py
```diff
@@ -0,0 +1,175 @@
...
+        # Make sure that we fitted a local model to each time series in data
+        data_hash = hash_ts_dataframe_items(data)
+        items_to_fit = [item_id for item_id, ts_hash in data_hash.iteritems() if ts_hash not in self._fitted_models]
+        if len(items_to_fit) > 0:
+            logger.info(f"{self.name} received {len(items_to_fit)} items not seen during training, re-running fit")
+            self._fit(train_data=data)
+
+        # Make predictions in parallel
+        predict_fn = partial(self._predict_with_local_model, quantile_levels=quantile_levels)
+        timeseries_with_models = (
```
canerturkmen, line 140: ditto for generators+joblib

timeseries/src/autogluon/timeseries/models/statsmodels/abstract_statsmodels.py
```diff
@@ -0,0 +1,175 @@
...
+        raise NotImplementedError
+
+    def _get_predictions_from_initialized_model(self, initialized_model, cutoff, quantile_levels):
+        start = cutoff + pd.tseries.frequencies.to_offset(self.freq)
+        end = cutoff + self.prediction_length * pd.tseries.frequencies.to_offset(self.freq)
+        predictions = initialized_model.get_prediction(start=start, end=end)
+        results = [predictions.predicted_mean.rename("mean")]
+        for q in quantile_levels:
+            if q < 0.5:
+                coverage = 2 * q
```
canerturkmen, line 165: does this work? doesn't `q=0.5` result in `alpha/coverage=1` ?

timeseries/src/autogluon/timeseries/models/statsmodels/models.py
```diff
@@ -0,0 +1,174 @@
...
+    def _fit_local_model(
+        self, timeseries: pd.Series, default_model_init_args: dict, default_model_fit_args: dict
+    ) -> FittedLocalModel:
+        model_init_args = default_model_init_args.copy()
+        # Infer seasonal_periods if seasonal_periods is not given / is set to None
+        seasonal_periods = model_init_args.pop("seasonal_periods", None)
+        if seasonal_periods is None:
+            seasonal_periods = get_seasonality(self.freq)
+        # Disable seasonality if the model cannot be fit with seasonality
+        if len(timeseries) < 2 * seasonal_periods or seasonal_periods == 1:
```
canerturkmen, line 69: nit: this computation seems to be repeated for each thread. is there a way to already compute these once per `_fit` call?

timeseries/src/autogluon/timeseries/models/statsmodels/abstract_statsmodels.py
```diff
@@ -0,0 +1,175 @@
...
+        result_df = pd.concat({item_id: pred for item_id, pred in zip(data.iter_items(), predictions)})
+        result_df.index.rename([ITEMID, TIMESTAMP], inplace=True)
+        return TimeSeriesDataFrame(result_df)
+
+    def _predict_with_local_model(
+        self, timeseries: pd.Series, fitted_model: FittedLocalModel, quantile_levels: List[float]
+    ) -> pd.DataFrame:
+        raise NotImplementedError
+
+    def _get_predictions_from_initialized_model(self, initialized_model, cutoff, quantile_levels):
```
canerturkmen, line 158: nit: is the word `initialized` redundant here? should it be something like `sm_model` or `internal_model` instead?

canerturkmen: also, could we add type hints perhaps?

timeseries/src/autogluon/timeseries/models/statsmodels/models.py
```diff
@@ -0,0 +1,174 @@
...
+        "maxiter",
+        "n_jobs",
+    ]
+
+    def _fit_local_model(
+        self, timeseries: TimeSeriesDataFrame, default_model_init_args: dict, default_model_fit_args: dict
+    ) -> FittedLocalModel:
+        model_init_args = default_model_init_args.copy()
+        # Set trend to constant if trend = True
+        model_init_args.setdefault("enforce_stationarity", True)
```
canerturkmen, line 142: cool!

canerturkmen (commented):
great! looks like we're mostly there.

timeseries/src/autogluon/timeseries/models/statsmodels/abstract_statsmodels.py
```diff
@@ -0,0 +1,175 @@
...
+        self, timeseries: pd.Series, default_model_init_args: dict, default_model_fit_args: dict
+    ) -> FittedLocalModel:
+        """Fit a single local model to the time series."""
+        raise NotImplementedError
+
+    def _fit(self, train_data: TimeSeriesDataFrame, time_limit: int = None, **kwargs):
+        # Select timeseries that don't have a fitted local model
+        train_hash = hash_ts_dataframe_items(train_data)
+        items_to_fit = [item_id for item_id, ts_hash in train_hash.iteritems() if ts_hash not in self._fitted_models]
+        timeseries_to_fit = (train_data.loc[item_id][self.target] for item_id in items_to_fit)
```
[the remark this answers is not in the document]
shchur: According to my understanding and based on the experiments that I ran, joblib always uses `n_jobs` x `ram_for_a_single_process` memory. Even for the largest datasets in our benchmark, `ram_for_a_single_process` doesn't go above 300-400MB. This seems to be very manageable since we default to using 50% of the available cores. E.g., even with only 4GB of RAM available to AGTS, we will only run into problems if a machine has 20 cores, which is very unrealistic.

As for your original question, using lists or generators doesn't seem to impact the performance (the generator is slightly faster).

timeseries/src/autogluon/timeseries/models/statsmodels/abstract_statsmodels.py
shchur: I think `initialized` is slightly more appropriate here since we initialize the internal intermediate results inside the model that are necessary for prediction (by calling `fit` with `maxiter=0`), but this is not a strong opinion, we can change the name if you think it improves the readability.

Will add the type hints.

timeseries/src/autogluon/timeseries/models/statsmodels/abstract_statsmodels.py
shchur: `alpha=1` computes that median, so that's what we need

timeseries/src/autogluon/timeseries/models/statsmodels/abstract_statsmodels.py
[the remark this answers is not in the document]
shchur: The extra memory for each job doesn't change if we train multiple local models, e.g., training `ETS` or `ETS + ARIMA` takes the same amount of RAM in total.

timeseries/src/autogluon/timeseries/models/statsmodels/abstract_statsmodels.py
```diff
@@ -0,0 +1,205 @@
...
+        possible values.
+    """
+
+    statsmodels_allowed_init_args: List[str] = []
+    statsmodels_allowed_fit_args: List[str] = []
+    quantile_method_name: str = "pred_int"
+
+    def __init__(
+        self,
+        freq: Optional[str] = None,
```
shchur, line 65: Can we expect that `freq` is always not `None`? We need `freq` when making predictions and computing the `start` and `end` dates for the forecast.

timeseries/src/autogluon/timeseries/models/statsmodels/abstract_statsmodels.py
canerturkmen: Up to you. I think one of the main difficulties we have in naming is that the word `model` is overloaded. It refers to both our model, and the SM model. I thought `initialized` was redundant as you can't really get predictions from a model that's not initialized. Up to you, no strong opinions as well.

timeseries/src/autogluon/timeseries/models/statsmodels/abstract_statsmodels.py
canerturkmen: the hope is that we always have freq available in the context through `TimeSeriesDataFrame.freq`

timeseries/src/autogluon/timeseries/models/statsmodels/abstract_statsmodels.py
canerturkmen: I think it's OK to leave it like this.

timeseries/src/autogluon/timeseries/models/statsmodels/abstract_statsmodels.py
```diff
@@ -78,9 +78,11 @@ def __init__(
             hyperparameters=hyperparameters,
             **kwargs,
         )
+        self.n_jobs: int = self._get_model_params().get("n_jobs", -1)
         self._fitted_models: Dict[str, FittedLocalModel] = {}
 
     def _get_default_model_init_and_fit_args(self) -> Tuple[dict, dict]:
+        """Separate arguments that will be passed to __init__ and fit methods of the underlying SM model."""
         default_model_init_args = {}
```
canerturkmen, line 99: So we are keeping "default" in naming?

canerturkmen (approved):
LGTM! Just a couple of minor points to discuss.

timeseries/src/autogluon/timeseries/models/statsmodels/abstract_statsmodels.py
shchur: I agree, changes this to `sm_model` and the method name to `_get_predictions_from_statsmodels_model`

timeseries/src/autogluon/timeseries/models/statsmodels/abstract_statsmodels.py
shchur: Thanks, fixed now

shchur:
Thanks a lot for the feedback, @canerturkmen! All points have been addressed now.

canerturkmen (approved):
looking forward to try this!

PR #2071, opened 2022-08-25
Add predictor clone support

Innixma:
*Issue #, if available:*

*Description of changes:*
- Add predictor clone support to make a perfect replica of a predictor in a new location on disk
- Add predictor optimized clone support to make an optimized version of the original predictor in a new location on disk
- Add tutorial on deployment optimization + usage of clone
- Add unit tests for clone
- Add get_size_disk method
- Add get_size_disk_per_file method

Refer to tutorial for example usage: http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-2071/0daed4a/tutorials/tabular_prediction/tabular-deployment.html

### DIFF
--- a/tabular/src/autogluon/tabular/predictor/predictor.py
+++ b/tabular/src/autogluon/tabular/predictor/predictor.py
@@ -14,6 +14,7 @@ import pandas as pd
 
 from autogluon.common.loaders import load_json
 from autogluon.common.savers import save_json
+from autogluon.common.utils.file_utils import get_directory_size, get_directory_size_per_file
 from autogluon.common.utils.log_utils import set_logger_verbosity
 from autogluon.common.utils.pandas_utils import get_approximate_df_mem_usage
 from autogluon.common.utils.utils import setup_outputdir, get_autogluon_metadata, compare_autogluon_metadata
@@ -2566,16 +2567,36 @@ class TabularPredictor:
         """
         Returns the combined size of all files under the `predictor.path` directory in bytes.
         """
-        start_path = self.path
-        total_size = 0
-        for dirpath, dirnames, filenames in os.walk(start_path):
-            for f in filenames:
-                fp = os.path.join(dirpath, f)
-                # skip if it is symbolic link
-                if not os.path.islink(fp):
-                    file_size = os.path.getsize(fp)
-                    total_size += file_size
-        return total_size
+        return get_directory_size(self.path)
+
+    def get_size_disk_per_file(self,
+                               *,
+                               sort_by: str = "size",
+                               include_path_in_name: bool = False) -> pd.Series:
+        """
+        Returns the size of each file under the `predictor.path` directory in bytes.
+
+        Parameters
+        ----------
+        sort_by : str, default = "size"
+            If None, output files will be ordered based on order of search in os.walk(path).
+            If "size", output files will be ordered in descending order of file size.
+            If "name", output files will be ordered by name in ascending alphabetical order.
+        include_path_in_name : bool, default = False
+            If True, includes the full path of the file including the input `path` as part of the index in the output pd.Series.
+            If False, removes the `path` prefix of the file path in the index of the output pd.Series.
+
+            For example, for a file located at `foo/bar/model.pkl`, with path='foo/'
+                If True, index will be `foo/bar/model.pkl`
+                If False, index will be `bar/model.pkl`
+
+        Returns
+        -------
+        pd.Series with index file path and value file size in bytes.
+        """
+        return get_directory_size_per_file(self.path,
+                                           sort_by=sort_by,
+                                           include_path_in_name=include_path_in_name)
 
     # TODO: v0.1 add documentation for arguments
     def get_model_names(self, stack_name=None, level=None, can_infer: bool = None, models: list = None) -> list:
@@ -3353,10 +3374,7 @@ class TabularPredictor:
         path_clone = shutil.copytree(src=self.path, dst=path, dirs_exist_ok=dirs_exist_ok)
         logger.log(30, f"Cloned {self.__class__.__name__} located in '{self.path}' to '{path_clone}'.\n"
                        f"\tTo load the cloned predictor: predictor_clone = {self.__class__.__name__}.load(path=\"{path_clone}\")")
-        if return_clone:
-            return self.__class__.load(path=path_clone)
-        else:
-            return path_clone
+        return self.__class__.load(path=path_clone) if return_clone else path_clone
 
     def clone_for_deployment(self,
                              path: str,
@@ -3370,6 +3388,13 @@ class TabularPredictor:
         This is ideal for use-cases where saving a snapshot of the predictor is desired before performing
         more advanced operations (such as fit_extra and refit_full).
 
+        Identical to performing the following operations in order:
+
+        predictor_clone = predictor.clone(path=path, return_clone=True, dirs_exist_ok=dirs_exist_ok)
+        predictor_clone.delete_models(models_to_keep=model, dry_run=False)
+        predictor_clone.set_model_best(model=model, save_trainer=True)
+        predictor_clone.save_space()
+
         Parameters
         ----------
         path : str
@@ -3377,12 +3402,14 @@ class TabularPredictor:
         model : str, default = 'best'
             The model to use in the optimized predictor clone.
             All other unrelated models will be deleted to save disk space.
+            Refer to the `models_to_keep` argument of `predictor.delete_models` for available options.
+            Internally calls `predictor_clone.delete_models(models_to_keep=model, dry_run=False)`
         return_clone : bool, default = False
             If True, returns the loaded cloned TabularPredictor object.
             If False, returns the local path to the cloned TabularPredictor object.
         dirs_exist_ok : bool, default = False
             If True, will clone the predictor even if the path directory already exists, potentially overwriting unrelated files.
-            If False, will raise an exception if the path directory already exists and avoid performing the copy.
+            If False, will raise an exception if the path directory already exists and avoids performing the copy.
 
         Returns
         -------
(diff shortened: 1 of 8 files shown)


tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -3299,6 +3319,93 @@ def explain_classification_errors(self, data, model = None, print_rules: bool =
...
+        assert path != self.path, f"Cannot clone into the same directory as the original predictor! (path='{path}')"
+        path_clone = shutil.copytree(src=self.path, dst=path, dirs_exist_ok=dirs_exist_ok)
+        logger.log(30, f"Cloned {self.__class__.__name__} located in '{self.path}' to '{path_clone}'.\n"
+                       f"\tTo load the cloned predictor: predictor_clone = {self.__class__.__name__}.load(path=\"{path_clone}\")")
+        if return_clone:
+            return self.__class__.load(path=path_clone)
+        else:
+            return path_clone
+
+    def clone_for_deployment(self,
```
sxjscience, line 3361: Can we merge this method in `.clone()`?

Innixma: This is possible, I don't think there is much objective difference between having it in 1 vs 2 methods.

The main motivation for having it separate is so that when a user is typing `predictor.c` they see both `clone` and `clone_for_deployment` which may inspire them to look up what the differences between the two are. Also, deployment optimized may be multiple steps in future, and it isn't easy to default all of them without it being clunky in 1 method. Happy to change to one method if we think it is better though. We can always deprecate the clone_for_deployment method and add in new args to clone.

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -2557,6 +2562,21 @@ def delete_models(self, models_to_keep=None, models_to_delete=None, allow_delete
...
+        start_path = self.path
+        total_size = 0
+        for dirpath, dirnames, filenames in os.walk(start_path):
+            for f in filenames:
+                fp = os.path.join(dirpath, f)
+                # skip if it is symbolic link
+                if not os.path.islink(fp):
+                    file_size = os.path.getsize(fp)
+                    total_size += file_size
+        return total_size
```
yinweisu, line 2578: Feel this might be a useful util function in general. Shall we move it into common with an extra parameter `path`, and then call it within in the `predictor.get_size_disk()`?

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -3299,6 +3319,93 @@ def explain_classification_errors(self, data, model = None, print_rules: bool =
...
+        This is ideal for use-cases where saving a snapshot of the predictor is desired before performing
+        more advanced operations (such as fit_extra and refit_full).
+
+        Parameters
+        ----------
+        path : str
+            Directory path the cloned predictor will be saved to.
+        model : str, default = 'best'
+            The model to use in the optimized predictor clone.
+            All other unrelated models will be deleted to save disk space.
```
yinweisu, line 3379: Maybe document other options besides `best`?

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -3299,6 +3319,93 @@ def explain_classification_errors(self, data, model = None, print_rules: bool =
...
+        if model == 'best':
+            model = predictor_clone.get_model_best()
+            logger.log(30, f"Clone: Keeping minimum set of models required to predict with best model '{model}'...")
+        else:
+            logger.log(30, f"Clone: Keeping minimum set of models required to predict with model '{model}'...")
+        predictor_clone.delete_models(models_to_keep=model, dry_run=False)
+        if isinstance(model, str) and model in predictor_clone.get_model_names(can_infer=True):
+            predictor_clone.set_model_best(model=model, save_trainer=True)
+        logger.log(30, f"Clone: Removing artifacts unnecessary for prediction. "
+                       f"NOTE: Clone can no longer fit new models, and most functionality except for predict and predict_proba will no longer work")
```
yinweisu, line 3402: I think we should add this note to docstring of this function in general to remind user.

tabular/tests/unittests/test_tabular.py
```diff
@@ -138,6 +139,36 @@ def test_advanced_functionality():
     assert len(leaderboard) == len(leaderboard_loaded)
     assert predictor_loaded.get_model_names_persisted() == []  # Assert that models were not still persisted after loading predictor
 
+    assert predictor.get_size_disk() > 0  # Assert that .get_size_disk() produces a >0 result and doesn't crash
+    # Test cloning logic
```
yinweisu, line 143: Just a comment in general: we might want to split `test_advanced_functionality` into smaller functions. Currently it's quite long and kind of hard to read

yinweisu (approved):
LGTM! Had some tiny comments

tabular/src/autogluon/tabular/predictor/predictor.py
gidler: This snippet looks pretty similar to code from a stack overflow answer. I just wanted to surface this to double check about creative commons licensing issues.

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -3299,6 +3319,93 @@ def explain_classification_errors(self, data, model = None, print_rules: bool =
...
+
+        """
+        assert path != self.path, f"Cannot clone into the same directory as the original predictor! (path='{path}')"
+        path_clone = shutil.copytree(src=self.path, dst=path, dirs_exist_ok=dirs_exist_ok)
+        logger.log(30, f"Cloned {self.__class__.__name__} located in '{self.path}' to '{path_clone}'.\n"
+                       f"\tTo load the cloned predictor: predictor_clone = {self.__class__.__name__}.load(path=\"{path_clone}\")")
+        if return_clone:
+            return self.__class__.load(path=path_clone)
+        else:
+            return path_clone
```
gidler, line 3359: TIOLI: I like the ternary operator here; avoids having branched return statements.
`return self.__class__.load(path=path_clone) if return_clone else path_clone`

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -3299,6 +3319,93 @@ def explain_classification_errors(self, data, model = None, print_rules: bool =
...
+            If True, returns the loaded cloned TabularPredictor object.
+            If False, returns the local path to the cloned TabularPredictor object.
+        dirs_exist_ok : bool, default = False
+            If True, will clone the predictor even if the path directory already exists, potentially overwriting unrelated files.
+            If False, will raise an exception if the path directory already exists and avoid performing the copy.
+
+        Returns
+        -------
+        If return_clone == True, returns the loaded cloned TabularPredictor object.
+        If return_clone == False, returns the local path to the cloned TabularPredictor object.
```
gidler, line 3349: Having the return type change based on this parameter seems like a code smell to me. But I don't have a good solution/suggestion here, so I'm just calling it out to see if anyone else had thoughts.

Innixma: Could you explain why it is a code smell? What situation would this lead to suboptimal functionality / usage?

tabular/tests/unittests/test_tabular.py
Innixma: I moved the disk size check to a new function, however moving other things is somewhat tricky: We want to ensure that the combination of all the advanced things work as expected, so testing them individually isn't enough.

tabular/src/autogluon/tabular/predictor/predictor.py
Innixma: good point, added

tabular/src/autogluon/tabular/predictor/predictor.py
Innixma: added

tabular/src/autogluon/tabular/predictor/predictor.py
Innixma: moved to common, also created a new `get_size_disk_per_file` method + added to common, you can refer to the tutorial update to see how the new method works.

Re @gidler comment: The logic has been updated.

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -3299,6 +3319,93 @@ def explain_classification_errors(self, data, model = None, print_rules: bool =
...
+        predictor_clone.delete_models(models_to_keep=model, dry_run=False)
+        if isinstance(model, str) and model in predictor_clone.get_model_names(can_infer=True):
+            predictor_clone.set_model_best(model=model, save_trainer=True)
+        logger.log(30, f"Clone: Removing artifacts unnecessary for prediction. "
+                       f"NOTE: Clone can no longer fit new models, and most functionality except for predict and predict_proba will no longer work")
+        predictor_clone.save_space()
+        if return_clone:
+            return predictor_clone
+        else:
+            return predictor_clone.path
```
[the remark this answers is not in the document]
Innixma: updated, thanks!

tabular/src/autogluon/tabular/predictor/predictor.py
Innixma: updated, thanks!

yinweisu (approved):
LGTM!

tabular/src/autogluon/tabular/predictor/predictor.py
gidler: If we were using a statically typed language, we would need to return either the cloned object or a path. In some languages you could do something fancy with the return type like `-> Union[TabularPredictor, PathStr]`, but often this kind of return type handling reveals a bad interface. Obviously since we're using Python, we don't have the limitations a statically typed language, but I think that perspective is still valid. My concern is if a user has logic that expects a TabularPredictor, then they decide to change this parameter and suddenly they are working with string, they would be surprised even if the documentation spells it out clearly.

An alternative solution is to have clone always return the path string, and users have to call `load()` separately if they want to bring the cloned TabularPredictor into memory. That seems like a good separation of concerns despite being a bit less convenient.

I just wanted to surface this as an outside observer. If you are happy with this interface, I don't oppose it.

Innixma: I understand where you are coming from, and it is definitely a reasonable perspective. I think that so long as the separation of concerns is only broken if they opt-in to use the convenience argument, then it is fine. I foresee that if this argument didn't exist, the two lines of code would be very common, so I like to simplify this down to 1 line of code when I anticipate it will be the primary use-case. You could argue our `predict_proba` having `as_pandas` argument is the same situation. There is no objectively right answer here, but I don't see the harm.

PR #2072, opened 2022-08-25
Handle timeseries model failures during the prediction phase

shchur:
Currently, if some model raises an exception during prediction (inside `AbstractTimeSeriesTrainer.predict`) the whole program crashes. This PR allows us to handle such model failures gracefully.

Changes:
- Add try/catch block in `AbstractTimeSeriesTrainer.predict` + instructions about what other models are available to the user if the best model failed.
- Re-normalize the weights in `SimpleTimeSeriesWeightedEnsemble.weight_pred_proba` in case some input models failed during prediction.

### DIFF
--- a/timeseries/tests/unittests/models/test_ensemble.py
+++ b/timeseries/tests/unittests/models/test_ensemble.py
@@ -1,0 +1,21 @@
+import pytest
+
+from autogluon.timeseries.models import AutoETSModel
+from autogluon.timeseries.models.ensemble.greedy_ensemble import TimeSeriesEnsembleWrapper
+
+from ..common import DUMMY_TS_DATAFRAME
+
+
+def test_when_some_base_models_fail_during_prediction_then_ensemble_can_still_predict():
+    ets = AutoETSModel(prediction_length=1)
+    ets.fit(train_data=DUMMY_TS_DATAFRAME, hyperparameters={"maxiter": 1, "seasonal": None})
+    ensemble = TimeSeriesEnsembleWrapper(weights={"ARIMA": 0.5, "AutoETS": 0.5}, name="WeightedEnsemble")
+    ets_preds = ets.predict(DUMMY_TS_DATAFRAME)
+    ensemble_preds = ensemble.predict(data={"ARIMA": None, "AutoETS": ets_preds})
+    assert (ets_preds.values == ensemble_preds.values).all()
+
+
+def test_when_all_base_models_fail_during_prediction_then_ensemble_raises_runtime_error():
+    ensemble = TimeSeriesEnsembleWrapper(weights={"ARIMA": 0.5, "AutoETS": 0.5}, name="WeightedEnsemble")
+    with pytest.raises(RuntimeError):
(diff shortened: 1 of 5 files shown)


timeseries/tests/unittests/models/test_ensemble.py
```diff
@@ -0,0 +1,21 @@
...
+
+from ..common import DUMMY_TS_DATAFRAME
+
+
+def test_when_some_base_models_fail_during_prediction_then_ensemble_can_still_predict():
+    ets = AutoETSModel(prediction_length=1)
+    ets.fit(train_data=DUMMY_TS_DATAFRAME, hyperparameters={"maxiter": 1, "seasonal": None})
+    ensemble = TimeSeriesEnsembleWrapper(weights={"ARIMA": 0.5, "AutoETS": 0.5}, name="WeightedEnsemble")
+    ets_preds = ets.predict(DUMMY_TS_DATAFRAME)
+    ensemble_preds = ensemble.predict(data={"ARIMA": None, "AutoETS": ets_preds})
```
canerturkmen, line 14: Nit: I don't follow why one of the models is failing. Should we maybe make it explicit by mocking one of the models' predict?

canerturkmen (approved):
LGTM!

timeseries/tests/unittests/models/test_ensemble.py
shchur: When we call `model.predict`, the model either returns a `TimeSeriesDataFrame` with predictions or raises an exception. It's the job of the `AbstractTimeSeriesTrainer.predict(model)` to return `None` if the model has failed. 

Here I just provided `None` instead of the ARIMA predictions to mimic the failure of ARIMA that was caught by the trainer.

I tried what you describe at first (setting a `side_effect = RuntimeError` with `mock`), but ran into the following problem: We need to instantiate a Trainer (or Predictor) to implement the logic of model fails => predict return None. However, when fitting an actual Predictor to `DUMMY_TS_DATAFRAME`, the ensemble always put 100% of the weight on a single model, which means I cannot test this case of "only one model fails while other work fine".

To summarize, this test here checks whether `TimeSeriesEnsembleWrapper` handles model failures correctly, but I couldn't find a way to do the same with `TimeSeriesPredictor` because of weird behavior of `fit_ensemble` that I couldn't control.

canerturkmen: I see. all OK. Then, on another note we seem to be silently changing the signature of `Learner` and `Predictor` to also optionally return None. Should we prevent that at least at the level of the Predictor to keep raising an error if we can't predict?

shchur: I guess that makes sense. So inside `Learner` we will check if the output of `trainer.predict` is `None`, and if so, raise an exception?

shchur: Done

canerturkmen: seems like we need an update to a test as well.

Issue #2042, opened 2022-08-16
Segmentation Fault when Importing AutoGluon


yinweisu:
We have noticed that a specific order of import AutoGluon modules will result in seg fault. For example,
```python
import autogluon.text  # same applies to autogluon.multimodal
from autogluon.vision import ImagePredictor
```
and
```python
import autogluon.text. # same applies to autogluon.multimodal
import autogluon.timeseries
```
will result in
```bash
free(): invalid size
Aborted (core dumped)
```

This is because of some underlying issue with mxnet when it is used along with pytorch. There is not too much AutoGluon team can do about this, but we are actively working toward getting rid of mxnet. 

To get around with it, simply reverse the order you import. For example,
```python
from autogluon.vision import ImagePredictor
import autogluon.text
```
and
```python
import autogluon.timeseries
import autogluon.text
```

gradientsky:
Another option to do `import mxnet` before other imports.

sxjscience:
Will we just remove the dependency on MXNet in 0.6?

canerturkmen:
I was able to reproduce this issue before but cannot anymore, with `torch==1.12.1` and `mxnet==1.9.0`. Could you perhaps check again @yinweisu ?

@sxjscience for timeseries we hope to make it optional and disabled by default.

yinweisu:
Created a new env and wasn't able to reproduce it either.
```bash
(temp) ubuntu@ip-172-31-11-12:~/yinweisu/autogluon$ pip3 freeze | grep mxnet
mxnet-cu110==1.9.1
(temp) ubuntu@ip-172-31-11-12:~/yinweisu/autogluon$ pip3 freeze | grep torch
pytorch-lightning==1.7.7
pytorch-metric-learning==1.3.2
torch==1.12.1
torchmetrics==0.8.2
torchtext==0.13.1
torchvision==0.13.1
(temp) ubuntu@ip-172-31-11-12:~/yinweisu/autogluon$ python3
Python 3.8.13 | packaged by conda-forge | (default, Mar 25 2022, 06:04:18) 
[GCC 10.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import autogluon.text
>>> from autogluon.vision import ImagePredictor
[...]

PR #2074, opened 2022-08-25
Add GluonTs PyTorch models to timeseries

canerturkmen:
*Issue #, if available:*

#2042 

*Description of changes:*

Introduces PyTorch-lightning based models to autogluon.timeseries, which will gradually phase out mxnet based models. 

Currently, PyTorch based DeepAR performance appear to be on par with MXNet DeepAR with a slight tendency to overfit. We will explore this further in more benchmark runs. Meanwhile,  we also get a 1.5x speedup in training and 1.3x speedup in inference.

TODO
- [x] Rename classes to make pytorch default
- [x] Reorganize files.
- [x] Switch SFF and DeepAR to default to pytorch
- [x] Extend tests to `test_models` and cover new callback with unit tests.
- [x] Benchmark on the full regression test data sets
- [x] Check regularization parameters to prevent overfitting
- [x] Suppress log messages from PyTorch-lightning

In a later PR:
- remove direct dependency on mxnet
- Adjust presets and default presets
- Clarify #2042 issue, which currently is not reproduced with `torch==1.12.1` and `mxnet==1.9.0`.
- Revise documentation including model zoo docs and tutorials


timeseries/src/autogluon/timeseries/models/gluonts/pytorch/models.py
```diff
@@ -0,0 +1,106 @@
...
+            logger.warning(
+                "GluonTS logging is turned on during training. Note that losses reported by GluonTS "
+                "may not correspond to those specified via `eval_metric`."
+            )
+
+        self._check_fit_params()
+        # TODO: reintroduce early stopping callbacks
+
+        # update auxiliary parameters
+        self._deferred_init_params_aux(dataset=train_data, callbacks=[PLTimeLimitCallback(time_limit)], **kwargs)
```
shchur, line 78: Maybe something to consider later: return parameters from `_deferred_init_params_aux` and pass them as an argument to `_get_estimator`? Currently we are modifying the state of the object, not sure if this the most elegant solution.

timeseries/src/autogluon/timeseries/models/gluonts/pytorch/models.py
```diff
@@ -0,0 +1,106 @@
...
+    gluonts_estimator_class: Type[GluonTSPyTorchLightningEstimator] = DeepAREstimator
+
+    def _get_estimator_init_args(self) -> Dict[str, Any]:
+        """Get GluonTS specific constructor arguments for estimator objects, an alias to
+        `self._get_model_params` for better readability."""
+        init_kwargs = self._get_model_params()
+
+        # GluonTS does not handle context_length=1 well, and sets the context to only prediction_length
+        # we set it to a minimum of 10 here.
+        init_kwargs["context_length"] = max(10, init_kwargs.get("context_length", self.prediction_length))
```
shchur, line 104: Isn't this already taken care of inside `presets.py`?

timeseries/tests/unittests/models/test_models.py
```diff
@@ -18,7 +18,7 @@
 from autogluon.timeseries.models.gluonts import GenericGluonTSModel
 
 from ..common import DUMMY_TS_DATAFRAME, dict_equal_primitive, get_data_frame_with_item_index
-from .test_gluonts import TESTABLE_MODELS as GLUONTS_TESTABLE_MODELS
+from .test_gluonts import TESTABLE_MX_MODELS as GLUONTS_TESTABLE_MODELS
```
shchur, line 21: Why do we only test mxnet models here instead of all models (`TESTABLE_MODELS`)?

timeseries/src/autogluon/timeseries/models/gluonts/__init__.py
```diff
@@ -8,10 +8,12 @@
     SimpleFeedForwardModel,
     TransformerModel,
 )
+from .pytorch.models import DeepARPyTorchModel
 
 __all__ = [
     "AutoTabularModel",
     "DeepARModel",
+    "DeepARPyTorchModel",
```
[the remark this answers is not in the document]
shchur: Probably in the presets we eventually want `DeepAR` to correspond to the PyTorch version, and the MXNet models should be renamed to something like `MxDeepAR`?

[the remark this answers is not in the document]
canerturkmen: Thanks!

I'm rather for changing `AbstractGluonTSPytorchModel` to `PyTorch`. this is more consistent with GluonTS's own conventions and our own conventions where we usually keep caps (`AbstractGluon*TS*Model`)

For MxDeepAR I completely agree, but could not decide whether we should just do this in the aliases in presets or also in the class names. Class names are not in the public API, so maybe it should be fine to change altogether. What do you think?

timeseries/src/autogluon/timeseries/models/gluonts/pytorch/models.py
canerturkmen: Hmm, but it doesn't have to do that. I'll leave it in just as a safeguard.

timeseries/tests/unittests/models/test_models.py
canerturkmen: pytorch tests coming soon.

timeseries/src/autogluon/timeseries/models/gluonts/__init__.py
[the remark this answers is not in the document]
shchur: I think renaming the classes is the way to go since our [documentation of the model zoo](https://auto.gluon.ai/dev/tutorials/timeseries/forecasting-model-zoo.html) relies on the class names, not aliases in presets.

timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
```diff
@@ -36,15 +36,27 @@ def __init__(
...
         self.target_field_name = target_field_name
+        self.float_dtype = float_dtype
 
     @property
     def freq(self):
-        return self.time_series_df.freq
+        # FIXME: GluonTS expects a frequency string, but only supports a limited number of such strings
+        # for feature generation. If the frequency string doesn't match or is not provided, it raises an exception.
+        # Here we bypass this by issuing a default "yearly" frequency, tricking it into not producing
+        # any lags or features.
```
shchur, line 52: Nice!

timeseries/src/autogluon/timeseries/models/gluonts/mx/models.py
```diff
@@ -434,7 +447,7 @@ def _get_estimator(self):
             return self.gluonts_estimator_class.from_hyperparameters(**self._get_estimator_init_args())
 
 
-class GenericGluonTSModelFactory(AbstractTimeSeriesModelFactory):
+class GenericGluonTSMXNetModelFactory(AbstractTimeSeriesModelFactory):
```
shchur, line 450: Do we actually ever use this class in the code? Or is this just functionality for advanced users to wrap arbitrary GluonTS models in AGTS?

timeseries/src/autogluon/timeseries/models/gluonts/mx/models.py
```diff
@@ -455,7 +468,7 @@ def train(self, train_data, validation_data=None, **kwargs):
         return self.predictor
 
 
-class ProphetModel(AbstractGluonTSModel):
+class ProphetModel(AbstractGluonTSMXNetModel):
```
shchur, line 471: I guess Prophet doesn't really depend on MXNet, but I'm fine with keeping it in this file.

On a related note, should we even enable this model / expose it in the presets given that it's not tested properly?

timeseries/src/autogluon/timeseries/models/gluonts/torch/models.py
```diff
@@ -0,0 +1,181 @@
...
+        # update auxiliary parameters
+        callbacks = [Timer(timedelta(seconds=time_limit))] if time_limit is not None else []
+        self._deferred_init_params_aux(dataset=train_data, callbacks=callbacks, **kwargs)
+
+        estimator = self._get_estimator()
+        with warning_filter(), disable_root_logger():
+            self.gts_predictor = estimator.train(
+                self._to_gluonts_dataset(train_data),
+                validation_data=self._to_gluonts_dataset(val_data),
+            )
```
shchur, line 116: It seems that the difference between `AbstractGluonTSPyTorchModel._fit` and `AbstractGluonTSMXNetModel._fit` is only related to how we set up the callbacks. Maybe we can move the callback setup to a separate method and have a single `_fit` method in `AbstractGluonTSModel`?

timeseries/src/autogluon/timeseries/models/gluonts/torch/models.py
```diff
@@ -0,0 +1,181 @@
...
+
+    def _get_estimator_init_args(self) -> Dict[str, Any]:
+        init_kwargs = super()._get_estimator_init_args()
+
+        # FIXME: PyTorch StudentT does not implement quantile functions
+        if "distr_output" in init_kwargs:
+            warnings.warn(
+                f"distr_output {init_kwargs['distr_output']} specified for SimpleFeedForward, however training"
+                "will default to the Gaussian distribution.",
+                category=UserWarning,
```
shchur, line 140: No need to specify this, `UserWarning` is the default https://docs.python.org/3/library/warnings.html#warnings.warn

timeseries/src/autogluon/timeseries/models/gluonts/torch/models.py
```diff
@@ -0,0 +1,181 @@
...
+            )
+        init_kwargs.update(dict(distr_output=NormalOutput()))
+        return init_kwargs
+
+    def _to_gluonts_dataset(self, time_series_df: Optional[TimeSeriesDataFrame]) -> Optional[GluonTSDataset]:
+        return (
+            SimpleGluonTSDataset(time_series_df, target_field_name=self.target, float_dtype=np.float32)
+            if time_series_df is not None
+            else None
+        )
```
shchur, line 150: Can we set `float_dtype` as a class variable for `AbstractGluonTSModel` and share the `_to_gluonts_dataset` method between MXNet & Pytorch? I will need to modify this method when introducing the features, so having a single method in `AbstractGluonTSModel` will lead to less code duplication in my PR.

timeseries/src/autogluon/timeseries/models/presets.py
```diff
@@ -23,13 +25,15 @@
...
+    MQRNN=MQRNNMXNetModel,
     SimpleFeedForward=SimpleFeedForwardModel,
+    SimpleFeedForwardMXNet=SimpleFeedForwardMXNetModel,
     DeepAR=DeepARModel,
+    DeepARMXNet=DeepARMXNetModel,
     Prophet=ProphetModel,
-    Transformer=TransformerModel,
-    TemporalFusionTransformer=TemporalFusionTransformerModel,
+    Transformer=TransformerMXNetModel,
+    TemporalFusionTransformer=TemporalFusionTransformerMXNetModel,
```
shchur, line 36: Maybe we should add the MX suffix to all MXNet models so that users understand that enabling these requires installing MXNet? Otherwise it's not very clear right now

timeseries/src/autogluon/timeseries/models/gluonts/mx/models.py
```diff
@@ -19,40 +22,50 @@
...
     from gluonts.mx.model.seq2seq import MQCNNEstimator, MQRNNEstimator
 
-from .abstract_gluonts import AbstractGluonTSModel
+from .callback import GluonTSEarlyStoppingCallback, TimeLimitCallback
 
 logger = logging.getLogger(__name__)
+gts_logger = logging.getLogger(gluonts.__name__)
+
+
+class AbstractGluonTSMXNetModel(AbstractGluonTSModel):
```
shchur, line 31: Should we also add some warning / check ensuring that the user has MXNet installed?

shchur (commented):
Looks good, thank you!

timeseries/src/autogluon/timeseries/models/presets.py
canerturkmen: This follows other conventions in Tabular (NeuralNetMXNet) and I also think it makes more sense as MXNet really qualified Model, not that it's a Transformer.

timeseries/src/autogluon/timeseries/models/gluonts/mx/models.py
canerturkmen: will do in a follow up PR

timeseries/src/autogluon/timeseries/models/gluonts/torch/models.py
canerturkmen: so we have 
```
class GluonTSDataset:
    float_dtype: Type = np.float64

class Float32GluonTSDataset(GluonTSDataset):
    float_dtype: Type = np.float32
```
?

timeseries/src/autogluon/timeseries/models/gluonts/mx/models.py
canerturkmen: That was the intention, but we're not really promoting it anywhere.

timeseries/src/autogluon/timeseries/models/gluonts/mx/models.py
canerturkmen: 👍 commenting it out from presets.

timeseries/src/autogluon/timeseries/models/gluonts/torch/models.py
shchur: I meant more like
```python
class AbstractGluonTSModel:
    float_dtype: Type

    def _to_gluonts_dataset(self, time_series_df: Optional[TimeSeriesDataFrame]) -> Optional[GluonTSDataset]:
        return (
            SimpleGluonTSDataset(time_series_df, target_field_name=self.target, float_dtype=self.float_dtype)
            if time_series_df is not None
            else None
        )

class AbstractGluonTSPyTorchModel(AbstractGluonTSModel):
    float_dtype = np.float32

class AbstractGluonTSMXNetModel(AbstractGluonTSModel):
    float_dtype = np.float64
```

canerturkmen: I see, yes, very cool

timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
```diff
@@ -47,14 +51,26 @@ def __init__(
         self.feat_static_cat = feat_static_cat
         self.feat_static_real = feat_static_real
 
+        self.int_dtype = int_dtype
+        self.float_dtype = float_dtype
         if self.feat_static_cat is not None:
-            self.feat_static_cat = self.feat_static_cat.astype(np.int64)
+            self.feat_static_cat = self.feat_static_cat.astype(self.int_dtype)
```
shchur, line 57: As far as I know, torch always uses int64 (and so did MXNet), so this might be redundant

timeseries/setup.py
```diff
@@ -29,6 +29,8 @@
     "pandas",
     "psutil>=5.7.3,<5.9",
     "gluonts~=0.11.0",
+    "torch>=1.9,<1.13",
```
shchur, line 32: Are we sure we don't want to set this higher? Or is this inherited from other AG modules?

timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
```diff
@@ -64,13 +80,17 @@ def __iter__(self) -> Iterator[Dict[str, Any]]:
...
             time_series = {
                 FieldName.ITEM_ID: item_id,
-                FieldName.TARGET: df[self.target_field_name].to_numpy(),
+                FieldName.TARGET: df[self.target_field_name].to_numpy(dtype=self.float_dtype),
                 FieldName.START: pd.Period(df.index[0], freq=self.freq),
             }
             if self.feat_static_cat is not None:
-                time_series[FieldName.FEAT_STATIC_CAT] = self.feat_static_cat.loc[item_id].to_numpy()
+                time_series[FieldName.FEAT_STATIC_CAT] = self.feat_static_cat.loc[item_id].to_numpy(
+                    dtype=self.int_dtype
```
shchur, line 88: We already did conversion to the correct float/int dtype in `__init__`, this seems redundant?

timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
```diff
@@ -235,23 +260,24 @@ def _fit(
             )
 
         self._check_fit_params()
-
-        callbacks = [TimeLimitCallback(time_limit)]
-
-        early_stopping_patience = self._get_model_params().get("early_stopping_patience", None)
-        if early_stopping_patience:
-            callbacks.append(GluonTSEarlyStoppingCallback(early_stopping_patience))
+        # TODO: reintroduce early stopping callbacks
```
shchur, line 263: Isn't this already resolved?

timeseries/src/autogluon/timeseries/models/gluonts/torch/models.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+class DeepARModel(AbstractGluonTSPyTorchModel):
+    gluonts_estimator_class: Type[GluonTSPyTorchLightningEstimator] = DeepAREstimator
+
+
+class SimpleFeedForwardModel(AbstractGluonTSPyTorchModel):
+    gluonts_estimator_class: Type[GluonTSPyTorchLightningEstimator] = SimpleFeedForwardEstimator
+    float_dtype: Type = np.float32
+    int_dtype: Type = np.int32
```
shchur, line 121: Shouldn't torch use `int64`? That's what `nn.Embedding` usually expects, even if all the weights are of `float32` dtype.

timeseries/src/autogluon/timeseries/models/gluonts/torch/models.py
```diff
@@ -0,0 +1,167 @@
...
+                )
+                for ix, quantile in enumerate(quantile_levels):
+                    item_forecast_dict[str(quantile)] = q_transformed[ix]
+            else:
+                for quantile in quantile_levels:
+                    item_forecast_dict[str(quantile)] = forecast.quantile(str(quantile))
+
+            df = pd.DataFrame(item_forecast_dict)
+            df[ITEMID] = forecast.item_id
+            df[TIMESTAMP] = pd.date_range(
```
shchur, line 160: Can you add a TODO to replace this with https://github.com/awslabs/autogluon/pull/2258/files#diff-0dcbec1000b6cda6a58a3e1c4cb6b5e8639942b0442f716e7f74d965215af35dR4?

timeseries/src/autogluon/timeseries/models/gluonts/torch/models.py
```diff
@@ -0,0 +1,167 @@
...
+        with torch_warning_filter():
+            model = load_pkl.load(path=path + cls.model_file_name, verbose=verbose)
+            if reset_paths:
+                model.set_contexts(path)
+            model.gts_predictor = GluonTSPyTorchPredictor.deserialize(Path(path) / cls.gluonts_model_path)
+        return model
+
+
+class DeepARModel(AbstractGluonTSPyTorchModel):
+    gluonts_estimator_class: Type[GluonTSPyTorchLightningEstimator] = DeepAREstimator
```
shchur, line 115: Could add these TODOs here?
- TODO: Add docstring for the hyperparameters
- TODO: Enable static features

shchur (approved):
LGTM! Thank you!

timeseries/setup.py
canerturkmen: The version cap is across all modules, 1.9 is from multimodal.

timeseries/src/autogluon/timeseries/models/gluonts/torch/models.py
canerturkmen: will check, thanks!

PR #2080, opened 2022-08-29
[Multimodal] Fix presets by constraining modal.names

sxjscience:
*Description of changes:*

Follow-up revision PR of https://github.com/awslabs/autogluon/pull/2032 . Now "t-few" is enabled in the default config

https://github.com/awslabs/autogluon/blob/e3fda1fc9d84ded0c612bc13c7472194503089f0/multimodal/src/autogluon/multimodal/configs/model/fusion_mlp_image_text_tabular.yaml#L2-L9

We need to explicitly set "model.names" in the presets.

### DIFF
--- a/multimodal/src/autogluon/multimodal/configs/model/fusion_mlp_image_text_tabular.yaml
+++ b/multimodal/src/autogluon/multimodal/configs/model/fusion_mlp_image_text_tabular.yaml
@@ -1,11 +1,10 @@
 model:
   names:
-    # Can attach the list of models to use, e.g.
-    # - "categorical_mlp"
-    # - "numerical_mlp"
-    # - "hf_text"
-    # - "timm_image"
-    # - "fusion_mlp"
+    - "categorical_mlp"
+    - "numerical_mlp"
+    - "hf_text"
+    - "timm_image"
+    - "fusion_mlp"
   categorical_mlp:
     hidden_size: 64
     activation: "leaky_relu"
--- a/multimodal/src/autogluon/multimodal/configs/model/fusion_mlp_image_text_tabular.yaml
+++ b/multimodal/src/autogluon/multimodal/configs/model/fusion_mlp_image_text_tabular.yaml
@@ -4,9 +4,7 @@ model:
     - "numerical_mlp"
     - "hf_text"
     - "timm_image"
-    - "clip"
     - "fusion_mlp"
-    - "t_few"
   categorical_mlp:
     hidden_size: 64
     activation: "leaky_relu"
--- a/multimodal/src/autogluon/multimodal/configs/model/fusion_mlp_image_text_tabular.yaml
+++ b/multimodal/src/autogluon/multimodal/configs/model/fusion_mlp_image_text_tabular.yaml
@@ -1,10 +1,11 @@
 model:
   names:
-    - "categorical_mlp"
-    - "numerical_mlp"
-    - "hf_text"
-    - "timm_image"
-    - "fusion_mlp"
+    # Can attach the list of models to use, e.g.
+    # - "categorical_mlp"
+    # - "numerical_mlp"
+    # - "hf_text"
+    # - "timm_image"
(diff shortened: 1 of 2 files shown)


multimodal/src/autogluon/multimodal/presets.py
```diff
@@ -9,6 +9,7 @@
 @automm_presets.register()
 def default():
     return {
+        "model.names": ["categorical_mlp", "numerical_mlp", "timm_image", "hf_text", "fusion_mlp"],
```
zhiqiangdon, line 12: Previously, `clip` was used in the `default`.

multimodal/src/autogluon/multimodal/presets.py
```diff
@@ -17,6 +18,7 @@ def default():
 @automm_presets.register()
 def medium_quality_faster_train():
     return {
+        "model.names": ["categorical_mlp", "numerical_mlp", "timm_image", "hf_text", "fusion_mlp"],
```
zhiqiangdon, line 21: Previously, clip was used in the `medium_quality_faster_train `.

multimodal/src/autogluon/multimodal/presets.py
```diff
@@ -26,6 +28,7 @@ def medium_quality_faster_train():
 @automm_presets.register()
 def high_quality():
     return {
+        "model.names": ["categorical_mlp", "numerical_mlp", "timm_image", "hf_text", "fusion_mlp"],
```
zhiqiangdon, line 31: Previously, clip was used in the `high_quality `.

multimodal/src/autogluon/multimodal/presets.py
```diff
@@ -34,6 +37,7 @@ def high_quality():
 @automm_presets.register()
 def best_quality():
     return {
+        "model.names": ["categorical_mlp", "numerical_mlp", "timm_image", "hf_text", "fusion_mlp"],
```
zhiqiangdon, line 40: Previously, clip was used in the best_quality.

multimodal/src/autogluon/multimodal/presets.py
```diff
@@ -43,6 +47,7 @@ def best_quality():
 @automm_presets.register()
 def multilingual():
     return {
+        "model.names": ["categorical_mlp", "numerical_mlp", "timm_image", "hf_text", "fusion_mlp"],
```
zhiqiangdon, line 50: Do we need `timm_image` in the multilingual setting?

multimodal/src/autogluon/multimodal/configs/model/fusion_mlp_image_text_tabular.yaml
```diff
@@ -4,9 +4,7 @@ model:
     - "numerical_mlp"
     - "hf_text"
     - "timm_image"
-    - "clip"
     - "fusion_mlp"
-    - "t_few"
```
zhiqiangdon, line 9: Could we empty the `model.names` here?

multimodal/src/autogluon/multimodal/presets.py
sxjscience: I added "clip" back. But I think we can remove it in 0.6. Currently, there is no clear advantage of fusing "clip" together with "timm" + "huggingface".

zhiqiangdon: I remember that adding clip helped improves the performances in image+text+tabular benchmarks. Will double check this.

multimodal/src/autogluon/multimodal/configs/model/fusion_mlp_image_text_tabular.yaml
zhiqiangdon: Maybe we can only remove `t_few` to pass the tests. I will dig into this later.

zhiqiangdon (approved):
Thanks!

PR #2085, opened 2022-08-30
Added TCGA example/ Wrapped FT_Transformer for TabularPredictor

BingzhaoZhu:
This is BZ's warmup project as an intern. The changes include:

1. Added an example of using AG for classification tasks with the TCGA datset.
2. Wrapped FT_Transformer as a standalone model to use for TabularPredictor.
```TabularPredictor().fit(train_data, hyperparameters={'FT_TRANSFORMER':{}})```
Now the above code directly calls FT_Transformer, to train on tables with numeric and categorical fields. 

### CODE REVIEW

File: tabular/src/autogluon/tabular/models/automm/ft_transformer.py [hunk 1]
```diff
@@ -0,0 +1,59 @@
+"""Wrapper of the MultiModalPredictor."""
+import logging
+from .automm_model import MultiModalPredictorModel
+
+logger = logging.getLogger(__name__)
+
+
+class FTTransformerModel(MultiModalPredictorModel):
```
sxjscience, line 8: There is no need to wrap this class. We can reuse our AutoMMPredictor wrapper but specify the flags.

sxjscience:
There is no need to write a separate FT_Transformer predictor. We may reuse our current integration: https://github.com/awslabs/autogluon/blob/master/tabular/src/autogluon/tabular/models/automm/automm_model.py

examples/automm/TCGA_cancer_survival/readme.md
```diff
@@ -0,0 +1,43 @@
+# Predict Vital Status from Patients with Cancer
```
sxjscience, line 1: Use `README.md` instead of `readme.md`.

tabular/src/autogluon/tabular/trainer/model_presets/presets.py
```diff
@@ -78,6 +78,7 @@
     AG_TEXT_NN=TextPredictorModel,
     AG_IMAGE_NN=ImagePredictorModel,
     AG_AUTOMM=MultiModalPredictorModel,
+    FT_TRANSFORMER=FTTransformerModel,
```
sxjscience, line 81: Is there a way for us to just reuse MultiModalPredictorModel but change the defaults to use FTTransformer?

BingzhaoZhu: I think only the name is changed here. In [ft_transformer.py](https://github.com/awslabs/autogluon/pull/2085/files/9c1af3c527019794f0d42d2d147e316f10084b70#diff-e624d1064c00f87634e12d90d62d4d5c398d34cb5129f49adae2b22d2d0fdd5d), FTTransformer is implemented by changing the default params of MultiModalPredictorModel.

tabular/src/autogluon/tabular/models/automm/ft_transformer.py
Innixma: I don't see much harm with doing it this way unless you are concerned with future backwards compatibility. This is easier to access and understand from the user's POV

examples/automm/TCGA_cancer_survival/example_cancer_survival.py
```diff
@@ -0,0 +1,118 @@
...
+    parser.add_argument('--path', default='./dataset')
+    parser.add_argument('--test_size', default=0.3)
+    parser.add_argument('--shuffle', default=True)
+    parser.add_argument('--seed', type=int, default=123)
+    parser.add_argument('--task', choices=['TCGA_HNSC', 'adult'],
+                        default='adult')
+    parser.add_argument('--mode', choices=['FT_Transformer', 'all_models'],
+                        default='all_models')
+    parser.add_argument('--num_gpus', type=int, default=None)
+    parser.add_argument('--num_workers', type=int, default=12)
```
Innixma, line 41: Is it unable to infer the optimal value? Why default 12? Have you tried None/-1?

tabular/src/autogluon/tabular/models/automm/ft_transformer.py
Innixma, line 8: Add # TODO: Add unit tests

Innixma (commented):
LGTM! Had a couple comments.

tabular/src/autogluon/tabular/models/automm/ft_transformer.py
sxjscience: Yes, +1 on a `#TODO` item about unit-test.

sxjscience: Sounds good. I'm resolving this.

sxjscience (approved):
LGTM. A small comment on adding a TODO item in `ft_transformer.py` about adding unit-test.

examples/automm/TCGA_cancer_survival/example_cancer_survival.py
BingzhaoZhu: There doesn't seems to be an easy of of setting the optimal value here. 'num_workers' seems to be relevant to [many factors](https://discuss.pytorch.org/t/guidelines-for-assigning-num-workers-to-dataloader/813/3). Here, 0 means single thread and -1 is not allowed. Should we just go with with the default setting with [MultiModal default parameter](https://github.com/awslabs/autogluon/blob/06fbc64cd1df53618b571c9a50835385d7b28c58/multimodal/src/autogluon/multimodal/configs/environment/default.yaml#:~:text=env%3A,%22ddp_spawn%22)? In that case num_gpus=-1 and num_workers=2.

sxjscience: Default value makes sense.

BingzhaoZhu:
Hi, I have added the "TODO" comment as suggested, and changed the default config of FT_Transformer in a recent commit.

Innixma (approved):

BingzhaoZhu:
> LGTM, thanks for the contribution!

Great. Thanks a lot!

jturner116:
Would this make more sense in Tabular examples than in AutoMM? I looked for it first in Tabular examples, but I'm just a goon. 

Thanks :)

PR #2086, opened 2022-08-30
[Cloud] Cloud CI

yinweisu:
*Issue #, if available:*

*Description of changes:*
This PR adds CI for Cloud module
* To keep the workflow secure, cloud predictor would require `cloud safe to test` tag in the PR for pull requests initiated from a fork unless the author has write permission to our repo. Adding label can only be done by maintainers, hence making sure no random code changes can trigger our cloud CI. Label will be removed by bot after each trigger and further commits require adding labels again.
* Cloud CI will only be triggered if code changes under cloud module is detected.
* Moved multimodal predictor text module only test to text test in general to speed up CI
* Moved multimodal predictor image module only test to image test in general to speed up CI
* Updated regular CI flow so that cloud module changes won't trigger it
* Changes of this PR will only take effect after merging in. Tests have been done here: https://github.com/yinweisu/autogluon/actions/runs/2960364517


.github/workflows/cloud_continuous_integration.yml
```diff
@@ -0,0 +1,123 @@
...
+        if: ${{ github.event_name == 'push' }}
+        uses: actions/checkout@v2
+      - name: Checkout repository(Pull Request Target)
+        if: ${{ github.event_name == 'pull_request_target' }}
+        uses: actions/checkout@v2
+        with:
+          ref: "refs/pull/${{ github.event.number }}/merge"
+      - name: Test General Cloud
+        uses: ./.github/actions/test-cloud
+        with:
```
Innixma, line 58: What is 'general'? Not fully getting where this maps to.

cloud/tests/README.md
```diff
@@ -0,0 +1,4 @@
+# Welcome to Contributing to AutoGLuon-Cloud!
+We do not run CI on your PRs directly for security reason.
```
Innixma, line 2: grammar: 'reasons'

cloud/tests/conftest.py
```diff
@@ -0,0 +1,91 @@
...
+import pytest
+
+
+class CloudTestHelper:
+
+    @staticmethod
+    def prepare_data(*args):
+        s3 = boto3.client('s3')
+        for arg in args:
+            s3.download_file('autogluon-cloud', arg, os.path.basename(arg))
```
Innixma, line 12: would be nice to have some documentation here, it isn't obvious how this works without looking at the code that uses it, especially since ordered args are being used instead of keyword args.

Innixma, line 12: if possible use keyword arguments here so it is easier to understand

cloud/tests/unittests/multimodal/test_multimodal.py
```diff
@@ -0,0 +1,74 @@
+import tempfile
+import zipfile
+
+from autogluon.cloud import MultiModalCloudPredictor
+
+
+def test_multimodal_tabular_text(test_helper):
+    train_data = 'tabular_text_train.csv'
+    test_data = 'tabular_text_test.csv'
```
Innixma, line 9: So we don't have a directory structure, its all flat? Better to have a directory structure IMO. Not a dealbreaker but just mentioning in case these files get crowded in s3.

Innixma (approved):
LGTM, nicely done! I especially like the separation and auto-detection of when to run the different CIs.

.github/workflows/cloud_continuous_integration.yml
yinweisu: This more refers to logic that's not specific to any predictor. For example, the data preparation. Please let me know if you think there's some better naming

cloud/tests/conftest.py
yinweisu: Here we basically get args as name of the file being downloaded from s3 bucket. I think it makes more sense to use *args directly as there's no variable name involved.

cloud/tests/unittests/multimodal/test_multimodal.py
yinweisu: Yea I think you are right. I will improve this in a follow up PR

cloud/tests/conftest.py
Innixma: I don't mean that, I mean for the `s3.download_file(...)` call

Innixma: regardless, this is a nit comment, so feel free to ignore

PR #2092, opened 2022-08-31
Fix item order in GluonTS models predictions

shchur:
Currently `AbstractGluonTSModel.predict` automatically sorts the predictions according to the `item_id` (even if the original `item_id`s are not sorted). This leads to problems downstream inside `TimeSeriesEnsembleSelection`, where all predictions are expected to have the exact same index. This PR fixes this problem and adds safeguards that should prevent similar problems in the future.

Side note: This was the source of the performance regressions that we saw compared to the 0.5.1 release of AGTS.

*Description of changes:*
- Ensure that the order of `item_id`s in predictions in the same as in `data` for `AbstractGluonTSModel`
- Add sanity check to `TimeSeriesEnsembleSelection` to ensure that the index of all predictions has the same order
- Modified the tests to catch incorrect item ordering in `predict` in the future

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/ensemble/greedy_ensemble.py
+++ b/timeseries/src/autogluon/timeseries/models/ensemble/greedy_ensemble.py
@@ -38,10 +38,8 @@ class TimeSeriesEnsembleSelection(EnsembleSelection):
 
     def _fit(self, predictions, labels, time_limit=None, sample_weight=None):
         self.dummy_pred = copy.deepcopy(predictions[0])
-        # This should never happen; sanity check to make sure that predict is implemented correctly for all models
-        # TODO: Maybe automatically sort the index to align the index for all predictions?
-        if not all(self.dummy_pred.index.equals(pred.index) for pred in predictions):
-            raise RuntimeError("Predictions for all models should have the exact same index")
+        # This should never happen; sanity check to make sure that all predictions have the same index
+        assert all(self.dummy_pred.index.equals(pred.index) for pred in predictions)
         super()._fit(
             predictions=[d.values for d in predictions],
             labels=labels,
--- a/timeseries/src/autogluon/timeseries/models/ensemble/greedy_ensemble.py
+++ b/timeseries/src/autogluon/timeseries/models/ensemble/greedy_ensemble.py
@@ -38,6 +38,10 @@ class TimeSeriesEnsembleSelection(EnsembleSelection):
 
     def _fit(self, predictions, labels, time_limit=None, sample_weight=None):
         self.dummy_pred = copy.deepcopy(predictions[0])
+        # This should never happen; sanity check to make sure that predict is implemented correctly for all models
+        # TODO: Maybe automatically sort the index to align the index for all predictions?
+        if not all(self.dummy_pred.index.equals(pred.index) for pred in predictions):
+            raise RuntimeError("Predictions for all models should have the exact same index")
         super()._fit(
             predictions=[d.values for d in predictions],
             labels=labels,
--- a/timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
+++ b/timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
@@ -48,10 +48,10 @@ class SimpleGluonTSDataset(GluonTSDataset):
         return self.time_series_df.freq
 
     def __len__(self):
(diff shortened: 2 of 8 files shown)


timeseries/tests/unittests/common.py
```diff
@@ -51,7 +51,7 @@ def get_data_frame_with_item_index(
     )
 
 
-DUMMY_TS_DATAFRAME = get_data_frame_with_item_index(["A", "B", "C", "D"])
+DUMMY_TS_DATAFRAME = get_data_frame_with_item_index(["10", "A", "2", "1"])
```
shchur, line 54: Change the items to not be sorted alphabetically, this allows us to catch bugs related to changed item order.

timeseries/tests/unittests/models/test_models.py
```diff
@@ -272,7 +273,7 @@ def test_when_predict_called_with_test_data_then_predictor_inference_correct(
     assert isinstance(predictions, TimeSeriesDataFrame)
     assert len(predictions) == test_data.num_items * prediction_length
 
-    predicted_item_index = predictions.index.levels[0]
```
shchur, line 275: `index.levels[0]` returned sorted item IDs, `unique(level=0)` doesn't sort them.

timeseries/src/autogluon/timeseries/models/ensemble/greedy_ensemble.py
```diff
@@ -38,6 +38,10 @@ def __init__(
 
     def _fit(self, predictions, labels, time_limit=None, sample_weight=None):
         self.dummy_pred = copy.deepcopy(predictions[0])
+        # This should never happen; sanity check to make sure that predict is implemented correctly for all models
```
canerturkmen, line 41: if it's a sanity check (the user cannot provide anything to break this), then we can replace with an `assert`. If the user can break this, we can replace with a `ValueError`. In any case I think we're fine without the TODO?

timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
```diff
@@ -247,7 +247,8 @@ def predict(self, data: TimeSeriesDataFrame, quantile_levels: List[float] = None
                         inplace=True,
                     )
 
-        return df
+        # Make sure the item_ids are sorted in the same order as in data
+        return df.loc[list(data.iter_items())]
```
canerturkmen, line 251: If this is the reason for the regression, didn't we also have predictions sorted incorrectly in v0.5.1? Which commit specifically led to the regression?

canerturkmen (approved):
LGTM!

timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
shchur: The source of the problem is the new dataset splitter: we converted item IDs to strings (from ints), so the sorted order is `["1_[None:None]", "100_[None:None]", "101_[None:None]", ..., "999_[None:None]"]` instead of `[1, 2, 3, ..., 999]` before.

In any case, I think that preserving the item order during prediction is what the users would expect from the model anyway.

timeseries/src/autogluon/timeseries/models/ensemble/greedy_ensemble.py
shchur: Regarding the TODO: This assumption can break if we introduce new models that don't respect the item order or break the old ones.

The options are:
1. Keep an assert to catch this behavior at runtime & force us to fix the item order
2. Sort the items here to handle even incorrect item order

Probably 1 is the better option, but I wasn't sure

shchur: Should be fixed now

timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
canerturkmen: all good. would you like to wait to rebase this on #2093 ? just so you can check offline `df.loc[data.item_ids]` doesn't break?

shchur: Sure, will do that once you merge #2093 into master.

shchur:
@canerturkmen the latest update also uses `df.item_ids` inside the tests replacing old code `df.index.levels[0]`.

PR #2093, opened 2022-08-31
Add deprecation decorator, deprecate TimeSeriesDataFrame.iter_items

canerturkmen:
*Issue #, if available:*

*Description of changes:*

This PR 
- Adds a deprecation decorator to ag.core
- deprecates `TimeSeriesDataFrame.iter_items` as it shadows the superclass's `iteritems` method.

### DIFF
--- a/core/src/autogluon/core/utils/deprecated.py
+++ b/core/src/autogluon/core/utils/deprecated.py
@@ -8,12 +8,15 @@ logger = logging.getLogger(__name__)
 
 def deprecated(message="", *args, **kwargs):
     def deprecated_decorator(func):
-        warning_message = f"Calling deprecated function {func.__name__}. This function is deprecated " \
-                           "and will be removed in a future release. " + message
+        warning_message = (
+            f"Calling deprecated function {func.__name__}. This function is deprecated "
+            "and will be removed in a future release. " 
+            + message
+        )
         @functools.wraps(func)
         def wrapped_func(*args, **kwargs):
             warnings.warn(warning_message)
-            logger.warn(warning_message)
+            logger.warning(warning_message)
             return func(*args, **kwargs)
         return wrapped_func
     return deprecated_decorator
--- a/core/src/autogluon/core/utils/deprecated.py
+++ b/core/src/autogluon/core/utils/deprecated.py
@@ -1,0 +1,19 @@
+import logging
+import warnings
+import functools
+
+
+logger = logging.getLogger(__name__)
+
+
+def deprecated(message="", *args, **kwargs):
+    def deprecated_decorator(func):
+        warning_message = f"Calling deprecated function {func.__name__}. This function is deprecated " \
+                           "and will be removed in a future release. " + message
+        @functools.wraps(func)
+        def wrapped_func(*args, **kwargs):
+            warnings.warn(warning_message)
+            logger.warn(warning_message)
+            return func(*args, **kwargs)
+        return wrapped_func
+    return deprecated_decorator
--- a/timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
+++ b/timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
@@ -9,6 +9,8 @@ import numpy as np
 import pandas as pd
 from pandas.core.internals import ArrayManager, BlockManager
 
+from autogluon.core.utils.deprecated import deprecated
+
 ITEMID = "item_id"
 TIMESTAMP = "timestamp"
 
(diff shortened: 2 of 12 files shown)


core/src/autogluon/core/utils/deprecated.py
```diff
@@ -0,0 +1,19 @@
...
+import warnings
+import functools
+
+
+logger = logging.getLogger(__name__)
+
+
+def deprecated(message="", *args, **kwargs):
+    def deprecated_decorator(func):
+        warning_message = f"Calling deprecated function {func.__name__}. This function is deprecated " \
```
shchur, line 11: Nit: Should we use parenthesis for multi-line strings? For example (or we can also move `+ message` to the `logger` call
```
default_warning_ = (
    f"Calling deprecated function {func.__name__}. This function is deprecated "
    f"and will be removed in a future release. " 
    + message
)
```

core/src/autogluon/core/utils/deprecated.py
```diff
@@ -0,0 +1,19 @@
...
+                           "and will be removed in a future release. " + message
+        @functools.wraps(func)
+        def wrapped_func(*args, **kwargs):
+            warnings.warn(warning_message)
+            logger.warn(warning_message)
```
shchur, line 16: `logger.warn` is deprecated, we should use `logger.warning`

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
```diff
@@ -166,9 +168,13 @@ def get_freq(series):
         self._cached_freq = freq
         return freq
 
+    @deprecated("Please use `TimeSeriesDataFrame.iter_item_ids()` instead.")
     def iter_items(self) -> Iterable[Any]:
         return iter(self._item_index)
 
+    def iter_item_ids(self) -> Iterable[Any]:
```
shchur, line 175: Should we maybe expose `TimeSeriesDataFrame._item_index` as a public attribute `TimeSeriesDataFrame.item_ids`?
```python
@property
def item_ids(self):
    return self.index.unique(level=ITEMID)
```

There are many places in the code where we're doing something inefficient to get this information (e.g., `list(ts_dataframe.iter_items())` or `ts_dataframe.index.get_level_values(0).unique()`), and I don't really see downsides of exposing this attribute. 
- With the implementation above it will be impossible to mutate the underlying dataframe by messing with the `item_ids`.
- We are not really saving memory by using the iterator, since the array is anyway instantiated inside `iter` (and the size of the array is tiny compared to the size of the entire TimeSeriesDataFrame, so I don't think it's a concern).

We can expose both the iterator and the index with `.item_ids` and `.iter_item_ids()`.

shchur:
Another thing to potentially discuss related to this PR: Should we do a review of the API of the `TimeSeriesDataFrame` to make sure that we are consistent with the naming (e.g., we always use `item` to refer to a single time series, `timestep` to refer to the value of a timeseries at a specific time, etc.), and also describe this terminology in the documentation / tutorials?

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
canerturkmen: I agree 100%.

canerturkmen:
> Another thing to potentially discuss related to this PR: Should we do a review of the API of the `TimeSeriesDataFrame` to make sure that we are consistent with the naming (e.g., we always use `item` to refer to a single time series, `timestep` to refer to the value of a timeseries at a specific time, etc.), and also describe this terminology in the documentation / tutorials?

I looked things over pretty quick and I think we look good overall.

Innixma (commented):
Added initial review, once comments are addressed LGTM!

core/src/autogluon/core/utils/deprecated.py
```diff
@@ -0,0 +1,22 @@
...
+    def deprecated_decorator(func):
+        warning_message = (
+            f"Calling deprecated function {func.__name__}. This function is deprecated "
+            "and will be removed in a future release. " 
+            + message
+        )
+        @functools.wraps(func)
+        def wrapped_func(*args, **kwargs):
+            warnings.warn(warning_message)
+            logger.warning(warning_message)
```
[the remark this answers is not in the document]
canerturkmen: The initial intention was to make sure we warn, as the first statement is not guaranteed to log the message, and not all of our execution environments are really guaranteed to capture stderr (which is where warnings outputs stuff). Thinking again, I'll drop logging as (from the docs)

> [warnings.warn()](https://docs.python.org/3/library/warnings.html#warnings.warn) in library code if the issue is avoidable and the client application should be modified to eliminate the warning
> 
> [logging.warning()](https://docs.python.org/3/library/logging.html#logging.warning) if there is nothing the client application can do about the situation, but the event should still be noted

core/src/autogluon/core/utils/deprecated.py
```diff
@@ -0,0 +1,22 @@
+import logging
+import warnings
+import functools
+
+
+logger = logging.getLogger(__name__)
+
+
+def deprecated(message="", *args, **kwargs):
+    def deprecated_decorator(func):
```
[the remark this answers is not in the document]
canerturkmen: Having dropped logging, it is now up to the client application's warning settings whether to warn once or always.

shchur (approved):
LGTM!

Issue #2069, opened 2022-08-25
[BUG] Ray-based Tuning Optimization


Innixma:
- [x] I have checked that this bug exists on the latest stable version of AutoGluon
- [x] and/or I have checked that this bug exists on the latest mainline of AutoGluon via source installation

NOTE: Despite all these issues, ray based tuning is still much better than custom tuning *IF* enough time is specified by the user. The majority of these issues are mostly relevant when small amounts of time are given, such that we can't do many HPO trials. For example, with bagging disabled on 64 CPUs, fastai HPO can train 200 trials with ray in the same time custom HPO can train 5 trials, and ray gets a significantly better test score too due to the sheer amount of trials trained.

**Describe the bug**

- [x] (Fixed in https://github.com/awslabs/autogluon/pull/2133)1. #2068

- [x] (Mostly fixed in #2121) 2. Ray based Tuning kills trials that haven't finished training instead of early stopping them. Custom Tuning early stops trials instead of killing them.

> This situation is tricky because it can sometimes be good to kill them to avoid wasting resource bagging them in the bagging mode. If we were to implement a solution to issue 5 then making things early stop instead of kill would no longer be a problem.

- [x] (Fixed in #2094) 3. Ray based Tuning with FASTAI logs many warnings. Custom Tuning does not.
> NaN or Inf found in input tensor.
NaN or Inf found in input tensor.
NaN or Inf found in input tensor.

- [x] (Fixed in #2094) 4. Ray based Tuning with FASTAI does not tune `layers` hyperparameter (or it doesn't report it properly if it does). Custom Tuning does.

> Example warning log:
> 2022-08-25 01:52:37,376	INFO logger.py:727 -- Removed the following hyperparameter values when logging to tensorboard: {'layers': (50, 25)}
2022-08-25 01:52:37,403	INFO logger.py:727 -- Removed the following hyperparameter values when logging to tensorboard: {'layers': (500,)}
2022-08-25 01:52:37,430	INFO logger.py:727 -- Removed the following hyperparameter values when logging to tensorboard: {'layers': (200, 100)}

- [x] (Fixed in #2121) 5. Ray based Tuning can underestimate training time required when using bagging. Custom Tuning does not.

[...]

PR #2094, opened 2022-08-31
Fix ray tune layer warning

yinweisu:
*Issue #, if available:*
https://github.com/awslabs/autogluon/issues/2069

*Description of changes:*
This PR addresses point 4 in the above issue. To be notice, I cannot reproduce point 3 now for some reason even though I was able to last week.
* Handle layer as tuple passed to fastai. `nn_torch` doesn't have layers as as nested list so it doesn't have this issue
* Disable tensorboard logging for tabular hpo tune until ray tune handles hyperopt tune issue correctly
* Added `super().__init__()`, which was missing previously

### DIFF
--- a/core/src/autogluon/core/hpo/executors.py
+++ b/core/src/autogluon/core/hpo/executors.py
@@ -162,6 +162,7 @@ class RayHpoExecutor(HpoExecutor):
     }
     
     def __init__(self):
+        super().__init__()
         self.resources = None
         self.hyperparameter_tune_kwargs = None
         self.analysis = None
@@ -239,6 +240,9 @@ class RayHpoExecutor(HpoExecutor):
             run,
             RayTuneAdapterFactory
         )
+        # Disable tensorboard logging to avoid layer warning
+        # TODO: remove this when ray tune fix ray tune pass tuple to hyperopt issue
+        os.environ['TUNE_DISABLE_AUTO_CALLBACK_LOGGERS'] = '1'
         analysis = run(
             trainable=model_trial,
             trainable_args=train_fn_kwargs,
@@ -255,6 +259,7 @@ class RayHpoExecutor(HpoExecutor):
             time_budget_s=self.time_limit,
             verbose=0,
         )
+        os.environ.pop('TUNE_DISABLE_AUTO_CALLBACK_LOGGERS', None)
         self.analysis = analysis
         
     def report(self, reporter, **kwargs):
@@ -291,6 +296,7 @@ class CustomHpoExecutor(HpoExecutor):
     """Implementation of HpoExecutor Interface, where our custom logic is used as the backend"""
     
     def __init__(self):
+        super().__init__()
         self.scheduler_options = None
         self.scheduler = None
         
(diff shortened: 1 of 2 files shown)


core/src/autogluon/core/hpo/executors.py
```diff
@@ -162,6 +162,7 @@ class RayHpoExecutor(HpoExecutor):
     }
     
     def __init__(self):
+        super().__init__()
```
Innixma, line 165: What is the impact of this?

core/src/autogluon/core/hpo/executors.py
```diff
@@ -239,6 +240,9 @@ def execute(
             run,
             RayTuneAdapterFactory
         )
+        # Disable tensorboard logging to avoid layer warning
```
Innixma, line 243: Does this speed up training?

core/src/autogluon/core/hpo/executors.py
```diff
@@ -291,6 +296,7 @@ class CustomHpoExecutor(HpoExecutor):
     """Implementation of HpoExecutor Interface, where our custom logic is used as the backend"""
     
     def __init__(self):
+        super().__init__()
```
Innixma, line 299: What is the impact of this?

tabular/src/autogluon/tabular/models/fastainn/tabular_nn_fastai.py
```diff
@@ -216,6 +216,8 @@ def _fit(self,
         # TODO: calculate max emb concat layer size and use 1st layer as that value and 2nd in between number of classes and the value
         if params.get('layers', None) is not None:
             layers = params['layers']
+            if isinstance(layers, tuple):
+                layers = list(layers)
```
Innixma, line 220: Is this a bandaid to a bigger problem? Will every model need to keep this in mind and code logic to handle? Is there a way to generically fix this?

core/src/autogluon/core/hpo/executors.py
yinweisu: It's more like a bug fix. Previously, forgot to call `super().__init__()` will result in some property not being initialized in some case. It didn't cause problem as I set those property later. But added this just to be safe

core/src/autogluon/core/hpo/executors.py
yinweisu: It's likely but I didn't do a benchmark

tabular/src/autogluon/tabular/models/fastainn/tabular_nn_fastai.py
yinweisu: Yes, if a model needs a list as input and is using hyperopt as the search algorithm, this is needed. It's hard to find a way to generically handle it, as we don't know if the input is intended to be a list but got converted to tuple by hyperopt, or it's intended to be a tuple originally.

core/src/autogluon/core/hpo/executors.py
yinweisu: ditto

tabular/src/autogluon/tabular/models/fastainn/tabular_nn_fastai.py
Innixma: ah I see, that makes sense regarding the ambiguity.

Innixma (approved):
LGTM!

PR #2099, opened 2022-09-02
Make sure that time series splitter doesn't trim training series shorter than prediction_length + 1

shchur:
Some models (e.g. DeepAR) expect training time series to have at least `prediction_length + 1` time steps. This constraint was violated by the old implementation of the `MultiWindowSplitter`, which led to GluonTS models failing. 

*Description of changes:*
- `MultiWindowSplitter` now ensures that no training time series gets cut shorter than `prediction_length + 1` (but if the original series already was shorter, it stays that way).
- Removed the `overlap` parameter from `MultiWindowSplitter` as it wasn't used, but unnecessarily complicates the logic.
- Improved exception message & tests.

### DIFF
--- a/timeseries/src/autogluon/timeseries/splitter.py
+++ b/timeseries/src/autogluon/timeseries/splitter.py
@@ -3,7 +3,7 @@ from typing import Tuple
 
 import pandas as pd
 
-from .dataset.ts_dataframe import TimeSeriesDataFrame, ITEMID, TIMESTAMP
+from .dataset.ts_dataframe import ITEMID, TIMESTAMP, TimeSeriesDataFrame
 
 logger = logging.getLogger(__name__)
 
--- a/timeseries/src/autogluon/timeseries/splitter.py
+++ b/timeseries/src/autogluon/timeseries/splitter.py
@@ -3,7 +3,7 @@ from typing import Tuple
 
 import pandas as pd
 
-from .dataset import TimeSeriesDataFrame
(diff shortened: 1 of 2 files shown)


timeseries/src/autogluon/timeseries/splitter.py
```diff
@@ -135,54 +129,70 @@ class MultiWindowSplitter(AbstractTimeSeriesSplitter):
...
+
+            validation_dataframes.append(append_suffix_to_item_id(ts_dataframe, suffix))
+            ts_dataframe = ts_dataframe.slice_by_timestep(slice(None, -prediction_length))
+
+        train_dataframes.append(ts_dataframe)
+
+        train_data = pd.concat(train_dataframes).loc[original_item_order]
+        train_data._cached_freq = original_freq
         val_data = pd.concat(validation_dataframes)
-        val_data._cached_freq = train_data._cached_freq
```
canerturkmen, line 185: out of curiosity, do the freqs change, so slice does not preserve cached freqs? and why change this line specifically?

canerturkmen (approved):
LGTM! Ready to merge when you are.

timeseries/src/autogluon/timeseries/splitter.py
shchur: `_cached_freq` isn't preserved during `concat`. We could keep it as-is, but this makes it more transparent that we just use the `freq` of the original dataset as `_cached_freq` for both `train_data` and `val_data`.

PR #2105, opened 2022-09-05
Clean up the TimeSeriesDataFrame public API

shchur:
*Description of changes:*

- `TimeSeriesDataFrame.slice_by_timestep` now accepts `(start_index: Optional[int] = None, end_index: Optional[int] = None)` as input instead of a `slice` object.  
  - Passing a `slice` object is still supported; a deprecation warning is made in this case.
- `TimeSeriesDataFrame.subsequence` is renamed to `TimeSeriesDataFrame.slice_by_time`, old method is deprecated

### DIFF
--- a/timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
+++ b/timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
@@ -507,29 +507,31 @@ class TimeSeriesDataFrame(pd.DataFrame):
         result._cached_freq = self._cached_freq
         return result
 
-    def subsequence(self, start_time: pd.Timestamp, end_time: pd.Timestamp) -> TimeSeriesDataFrame:
+    def subsequence(self, start: pd.Timestamp, end: pd.Timestamp) -> TimeSeriesDataFrame:
         """Select a subsequence from each time series between start (inclusive) and end (exclusive) timestamps.
+
         Parameters
         ----------
-        start_time: pd.Timestamp
-            Start time (inclusive) of the slice for each time series.
-        end_time: pd.Timestamp
-            End time (exclusive) of the slice for each time series.
+        start: pd.Timestamp
+            The start time (inclusive) of a time range that will be used for subsequence.
+        end: pd.Timestamp
+            The end time (exclusive) of a time range that will be used for subsequence.
+
         Returns
         -------
         ts_df: TimeSeriesDataFrame
-            A new time series dataframe containing entries of the original time series between start and end timestamps.
+            A new data frame in ``TimeSeriesDataFrame`` format contains time-series in a time range
+            defined between start and end time.
         """
-        if end_time < start_time:
-            raise ValueError(f"end_time {end_time} is earlier than start_time {start_time}")
 
-        nanosecond_before_end_time = end_time - pd.Timedelta(nanoseconds=1)
-        result = TimeSeriesDataFrame(
-            self.loc[(slice(None), slice(start_time, nanosecond_before_end_time)), :],
+        if end < start:
+            raise ValueError(f"end time {end} is earlier than stat time {start}")
+
+        nanosecond_before_end = end - pd.Timedelta(nanoseconds=1)
+        return TimeSeriesDataFrame(
+            self.loc[(slice(None), slice(start, nanosecond_before_end)), :],
             static_features=self.static_features,
         )
-        result._cached_freq = self._cached_freq
-        return result
 
     @classmethod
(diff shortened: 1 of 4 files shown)


timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
```diff
@@ -181,7 +181,8 @@ def iter_items(self) -> Iterable[Any]:
     def num_items(self):
         return len(self.item_ids)
 
-    def num_timesteps_per_item(self) -> pd.Series:
+    def length_per_item(self) -> pd.Series:
```
canerturkmen, line 184: How is this making `length` consistent with other naming? Do we have any other APIs where we refer to "length" of timeseries? I think `num_timesteps` is more explicit---`length` in the context of time may refer to absolute time (2 days, etc.) or python `__len__`.

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
```diff
@@ -399,6 +401,123 @@ def split_by_time(self, cutoff_time: pd.Timestamp) -> Tuple[TimeSeriesDataFrame,
...
+        for length in length_per_item:
+            indicator = np.zeros(length, dtype=bool)
+            indicator[start_index:end_index] = True
+            boolean_indicators.append(indicator)
+        index = np.concatenate(boolean_indicators)
+        result = TimeSeriesDataFrame(self[index].copy(), static_features=self.static_features)
+        result._cached_freq = self._cached_freq
+        return result
+
+    def slice_items_by_time(self, start_time: pd.Timestamp, end_time: pd.Timestamp) -> TimeSeriesDataFrame:
```
canerturkmen, line 493: `items` appearing in the function name leads me to think that I can provide a subset of items that I want to slice. Why not `slice_by_time` alone?

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
```diff
@@ -399,6 +401,123 @@ def split_by_time(self, cutoff_time: pd.Timestamp) -> Tuple[TimeSeriesDataFrame,
         after._cached_freq = self._cached_freq
         return before, after
 
+    def slice_items_by_index(self, start_index: Union[int, None], end_index: Union[int, None]) -> TimeSeriesDataFrame:
```
canerturkmen, line 404: again 1/ I think the word `index` is overloaded and increases ambiguity over timestep 2/ I don't see why we specify `items` in the name. I don't have a strong opinion on 1 though.

canerturkmen (commented):
Looks good overall. Some comments, and I will look in more detail in the implementations as well. Thanks!

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
shchur: So how about we name the new methods `slice_by_index` and `slice_by_time`? I don't think `slice_by_index` is that ambiguous since the function accepts two integers, so it shouldn't be confused with a `pd.Index` object.

I like this option more than keeping the old name `slice_by_timestep` + adding some logic to handle both `slice` and `(int, int)` objects passed to the function.

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
shchur: My idea was that if we remove `timestep` from the slice method, then we can also remove it here since `length` sounds a bit more intuitive, coincides with python `__len__` for each item, and is more in line with how we usually use this information in other methods
```python
for item_id, length in df.length_per_item().iteritems():
    ...
```
vs.
```python
for item_id, length in df.num_timesteps_per_item().iteritems():
    ...
```
But I'm also fine with keeping it as `num_timesteps_per_item`, not really a strong preference.

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
canerturkmen: I still think it is confusing. If we iterate over each individual item, the _index_ of those items are time.

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
canerturkmen: `num_timesteps_per_item` then?

shchur:
Ready to be merged once the tests pass

canerturkmen (approved):
LGTM. thanks!

PR #2106, opened 2022-09-05
Add TemporalFusionTransformer time series model

shchur:
Thanks to Lorenzo's fix from https://github.com/awslabs/gluon-ts/issues/2264  we can add TFT to AutoGluon (the model should work both on CPU and GPU).

*Description of changes:*
- Add `TemporalFusionTransformerModel` to `timeseries.models`, wrapping `TemporalFusionTransformerEstimator` from GluonTS.

The model will be added to the default presets later after we complete benchmarking.

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/gluonts/models.py
+++ b/timeseries/src/autogluon/timeseries/models/gluonts/models.py
@@ -15,6 +15,7 @@ with warning_filter():
     from gluonts.model.seq2seq import MQCNNEstimator, MQRNNEstimator
     from gluonts.model.simple_feedforward import SimpleFeedForwardEstimator
     from gluonts.model.transformer import TransformerEstimator
+    from gluonts.model.tft import TemporalFusionTransformerEstimator
     from gluonts.mx.context import get_mxnet_context
     from gluonts.nursery.autogluon_tabular import TabularEstimator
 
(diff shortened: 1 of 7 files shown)


timeseries/src/autogluon/timeseries/models/gluonts/models.py
```diff
@@ -15,6 +15,7 @@
     from gluonts.model.seq2seq import MQCNNEstimator, MQRNNEstimator
     from gluonts.model.simple_feedforward import SimpleFeedForwardEstimator
     from gluonts.model.transformer import TransformerEstimator
+    from gluonts.model.tft import TemporalFusionTransformerEstimator
```
canerturkmen, line 18: very cool. is the fix for TFTs backported to v0.9?

timeseries/src/autogluon/timeseries/models/gluonts/models.py
```diff
@@ -190,6 +191,51 @@ class SimpleFeedForwardModel(AbstractGluonTSModel):
...
+
+    def _get_estimator(self) -> GluonTSEstimator:
+        """Return the GluonTS Estimator object for the model"""
+        hyperparameters = self._get_estimator_init_args()
+        # Turning off hybridization prevents MXNet errors when training on GPU
+        hyperparameters["hybridize"] = False
+        # TFT cannot handle arbitrary quantiles, this is a workaround
+        hyperparameters["num_outputs"] = 9
+        supported_quantiles = set([0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9])
+        if not set(self.quantile_levels).issubset(supported_quantiles):
```
canerturkmen, line 230: cool. so if `self.quantile_levels=[0.1, 0.5, 0.9]`, the current GluonTS logic makes sure these are returned? Are the default tests covering this?

timeseries/tests/unittests/models/test_gluonts.py
```diff
@@ -13,6 +13,7 @@
     MQCNNModel,
     ProphetModel,
     SimpleFeedForwardModel,
+    TemporalFusionTransformerModel,
```
canerturkmen, line 16: will we also add to TESTABLE_MODELS?

canerturkmen (commented):
Thanks!! Added very minor comments on testing, otherwise looks great.

timeseries/src/autogluon/timeseries/models/gluonts/models.py
canerturkmen: ah sorry I was asking about this: "Thanks to Lorenzo's fix from https://github.com/awslabs/gluonts/issues/2264 we can add TFT to AutoGluon (the model should work both on CPU and GPU)." but I guess that's `hybridize=False`.

timeseries/src/autogluon/timeseries/models/gluonts/models.py
```diff
@@ -190,6 +191,59 @@ class SimpleFeedForwardModel(AbstractGluonTSModel):
...
+        if not set(self.quantile_levels).issubset(self.supported_quantiles):
+            raise ValueError(
+                f"{self.name} requires that quantile_levels are a subset of "
+                f"{self.supported_quantiles} (received quantile_levels = {self.quantile_levels})"
+            )
+        with warning_filter():
+            return self.gluonts_estimator_class.from_hyperparameters(**hyperparameters)
+
+    def predict(self, data: TimeSeriesDataFrame, quantile_levels: List[float] = None, **kwargs) -> TimeSeriesDataFrame:
+        if quantile_levels is not None and not set(self.quantile_levels).issubset(self.supported_quantiles):
```
canerturkmen, line 239: is this relying on the fact that `self.quantile_levels` is set by default to [0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9]? if so, should we make it explicit in the init of this class with something like?

```
def __init__(self, *args, **kwargs):
    super().__init__(*args, **kwargs)
    self.quantile_levels = self.supported_quantiles  # list not set
```

shchur: This was a bug that should be fixed now.

- We make sure that `self.quantile_levels` is a subset of `self.supported_quantiles` when initializing the model in `_get_estimator`
- We make sure that if `quantile_levels` are provided as an argument to `predict` (and are not equal to `None`), these are still a subset of `self.supported_quantiles`

The updated code should correctly take care of the second check.

canerturkmen (approved):
Thanks!

PR #2109, opened 2022-09-07
How can i use child models to infer test data?

lujiaming-12138:
In the TabularPredictor fitting process, the autogluon will try to create some ensemble model such as 'LightGBMXT_BAG_L1' , and the model will try to 'Fitting 8 child models (S1F1 - S1F8) | Fitting with ParallelLocalFoldFittingStrategy'. So my question is how to use the child models to predict the new data? I can try to load_child in the _learn but i can not use them to predict. 

Actually, i can load_child('S1F1'), but when i call load_child('S1F1').predict() i get some error information which i can not use 
'S1F1' to infer test data. Is there anything about the save_bag_folds=False? if i set save_bag_folds=True, can i use S1F1 to infer?


Innixma:
Why is this a pull request? This reads like a question. Please create a fresh GitHub issue or discussion for your question.

PR #2120, opened 2022-09-08
Update documentation for TimeSeriesPredictor

shchur:
*Description of changes:*
This PR fixes several factually incorrect statements & improves wording in the documentation of the `TimeSeriesPredictor` API.

### DIFF
--- a/timeseries/src/autogluon/timeseries/predictor.py
+++ b/timeseries/src/autogluon/timeseries/predictor.py
@@ -23,54 +23,62 @@ logger = logging.getLogger(__name__)
 
 
 class TimeSeriesPredictor:
-    """AutoGluon ``TimeSeriesPredictor`` predicts future values of multiple related time-series by fitting
-    global time series models.
+    """AutoGluon ``TimeSeriesPredictor`` predicts future values of multiple related time series.
 
-    ``TimeSeriesPredictor`` provides probabilistic (distributional) forecasts for univariate time series, where the
-    time series model is essentially a mapping from the past of the time series to its future of length (i.e., forecast
-    horizon) defined by the user. Models are trained to give both forecast "means" (i.e., conditional expectations of
-    future values of a time series given its past), and quantiles of forecast distributions.
+    ``TimeSeriesPredictor`` provides probabilistic (distributional) multi-step-ahead forecasts for univariate time
+    series. The forecast includes both the mean (i.e., conditional expectation of future values given the past), as
+    well as the quantiles of the forecast distribution, indicating the range of possible future outcomes.
 
-    ``TimeSeriesPredictor`` models are learned "globally" from a collection of time series; i.e., a set of time
-    series model parameters are shared across all time series to be predicted, in contrast to
-    classical "local" approaches such as ARIMA.
+    ``TimeSeriesPredictor`` fits both "global" deep learning models that are shared across all time series (e.g.,
+    DeepAR, Temporal Fusion Transformer), as well as "local" statistical models that are fit to each individual time
+    series (e.g., ARIMA, ETS).
 
-    ``TimeSeriesPredictor`` fits a variety of neural network-based forecasting models as well as Bayesian models such
-    as Prophet. It expects input data sets and outputs predictions in the
+    ``TimeSeriesPredictor`` expects input data and makes predictions in the
     :class:`~autogluon.timeseries.TimeSeriesDataFrame` format.
+
 
     Parameters
     ----------
     target : str, default = "target"
-        Name of column that contains the target values to forecast (i.e., numeric observations of the
-        time series). This column must contain numeric values, and missing target values
-        should be in a pandas compatible format:
+        Name of column that contains the target values to forecast (i.e., numeric observations of the time series).
+        This column must contain numeric values, and missing target values should be in a pandas compatible format:
         https://pandas.pydata.org/pandas-docs/stable/user_guide/missing_data.html
-    eval_metric : str, default = None
+        # TODO: Do we even handle missing values? If so, should we make this more prominent?
+    prediction_length : int, default = 1
+        The forecast horizon, i.e., How many time steps into the future the forecasters should be trained to predict.
+        For example, if time series contain daily observations, setting ``prediction_length=3`` will train
+        models that predict up to 3 days in the future from the most recent observation.
+    eval_metric : str, default = "mean_wQuantileLoss"
         Metric by which predictions will be ultimately evaluated on future test data. AutoGluon tunes hyperparameters
         in order to improve this metric on validation data, and ranks models (on validation data) according to this
-        metric. Available options include: "MASE", "MAPE", "sMAPE", "mean_wQuantileLoss".
+        metric. Available options:
 
-        If ``eval_metric is None``, it is set by default to "mean_wQuantileLoss".
+        # TODO: mean_wQuantileLoss is not discussed in GluonTS docs
+        - ``"mean_wQuantileLoss"`` - mean weighted quantile loss
+        - ``"MAPE"`` - mean absolute percentage error
+        - ``"sMAPE"`` - symmetric mean absolute percentage error
+        - ``"MASE"`` - mean absolute scaled error
+
+        # TODO: Evaluator is not part of the API, do we expect people to dig into the code? Maybe include description
+        # of metrics in the tutorial?
         For more information about these options, see ``autogluon.timeseries.TimeSeriesEvaluator`` and GluonTS
-        docs at https://ts.gluon.ai/api/gluonts/gluonts.evaluation.metrics.html
-    path : str, default = None
-        Path to directory where models and intermediate outputs should be saved. If unspecified, a timestamped folder
+        docs at https://ts.gluon.ai/stable/api/gluonts/gluonts.evaluation.metrics.html
+    quantile_levels : List[float] or None, default = None
+        List of increasing decimals that specifies which quantiles should be estimated when making distributional
+        forecasts. If None, defaults to ``[0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9]``.
+        Can alternatively be provided with the keyword argument ``quantiles``.
+        # TODO: Add link to explanation of distributional forecasts?
+    path : str or None, default = None
+        Path to the directory where models and intermediate outputs will be saved. If None, a timestamped folder
         ``AutogluonModels/ag-[TIMESTAMP]`` will be created in the working directory to store all models.
     verbosity : int, default = 2
         Verbosity levels range from 0 to 4 and control how much information is printed to stdout. Higher levels
         correspond to more detailed print statements, and ``verbosity=0`` suppresses output including warnings.
         If using ``logging``, you can alternatively control amount of information printed via ``logger.setLevel(L)``,
-        where ``L`` ranges from 0 to 50 (Note: higher values of ``L`` correspond to fewer print statements,
-        opposite of verbosity levels).
-    prediction_length : int, default = 1
-        The forecast horizon, i.e., How many time points into the future forecasters should be trained to predict.
-        For example, if time series contain daily observations, setting ``prediction_length=3`` will train
-        models that predict up to 3 days in the future from the most recent observation.
-    quantile_levels : List[float], default = None
-        List of increasing decimals that specifies which quantiles should be estimated
-        when making distributional forecasts. Can alternatively be provided with the keyword
-        argument ``quantiles``. If ``None``, defaults to ``[0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9]``.
+        where ``L`` ranges from 0 to 50 (Note: higher values of ``L`` correspond to fewer print statements, opposite
+        of verbosity levels).
+
+    # TODO: Why not move Other Parameters to Parameters? At least `ignore_time_index` and `validation_splitter` seem relevant
 
     Other Parameters
     ----------------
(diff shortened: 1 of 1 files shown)


timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -23,54 +23,63 @@
...
     ----------
     target : str, default = "target"
-        Name of column that contains the target values to forecast (i.e., numeric observations of the
-        time series). This column must contain numeric values, and missing target values
-        should be in a pandas compatible format:
+        Name of column that contains the target values to forecast (i.e., numeric observations of the time series).
+        This column must contain numeric values, and missing target values should be in a pandas compatible format:
         https://pandas.pydata.org/pandas-docs/stable/user_guide/missing_data.html
-    eval_metric : str, default = None
+        # TODO: Do we even handle missing values? If so, should we make this more prominent?
```
canerturkmen, line 46: we don't, and we can remove this for now.

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -23,54 +23,63 @@
...
+        For example, if time series contain daily observations, setting ``prediction_length = 3`` will train
+        models that predict up to 3 days into the future from the most recent observation.
+    eval_metric : str, default = "mean_wQuantileLoss"
         Metric by which predictions will be ultimately evaluated on future test data. AutoGluon tunes hyperparameters
         in order to improve this metric on validation data, and ranks models (on validation data) according to this
-        metric. Available options include: "MASE", "MAPE", "sMAPE", "mean_wQuantileLoss".
+        metric. Available options:
+
+        # TODO: Should we add MSE and RMSE here?
+        # TODO: mean_wQuantileLoss is not discussed in GluonTS docs
```
canerturkmen, line 57: we can put a brief explainer that this is the mean of normalized quantile losses for specified `quantile_levels`, and temporarily have a link to Amazon Forecast if you'd like: https://docs.aws.amazon.com/forecast/latest/dg/metrics.html#metrics-wQL .

As a side note, we should not have TODOs left in the docstrings and should move them to the code if we have any left.

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -23,54 +23,63 @@
...
+        The forecast horizon, i.e., How many time steps into the future the models should be trained to predict.
+        For example, if time series contain daily observations, setting ``prediction_length = 3`` will train
+        models that predict up to 3 days into the future from the most recent observation.
+    eval_metric : str, default = "mean_wQuantileLoss"
         Metric by which predictions will be ultimately evaluated on future test data. AutoGluon tunes hyperparameters
         in order to improve this metric on validation data, and ranks models (on validation data) according to this
-        metric. Available options include: "MASE", "MAPE", "sMAPE", "mean_wQuantileLoss".
+        metric. Available options:
+
+        # TODO: Should we add MSE and RMSE here?
```
canerturkmen, line 56: If they're available, sure!

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -23,54 +23,63 @@
...
         For more information about these options, see ``autogluon.timeseries.TimeSeriesEvaluator`` and GluonTS
-        docs at https://ts.gluon.ai/api/gluonts/gluonts.evaluation.metrics.html
-    path : str, default = None
-        Path to directory where models and intermediate outputs should be saved. If unspecified, a timestamped folder
+        docs at https://ts.gluon.ai/stable/api/gluonts/gluonts.evaluation.metrics.html
+    quantile_levels : List[float] or None, default = None
+        List of increasing decimals that specifies which quantiles should be estimated when making distributional
+        forecasts. If None, defaults to ``[0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9]``.
+        Can alternatively be provided with the keyword argument ``quantiles``.
+        # TODO: Add link to explanation of distributional forecasts?
```
canerturkmen, line 71: I would not prioritize this. Up to you.

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -23,54 +23,63 @@
...
-        For example, if time series contain daily observations, setting ``prediction_length=3`` will train
-        models that predict up to 3 days in the future from the most recent observation.
-    quantile_levels : List[float], default = None
-        List of increasing decimals that specifies which quantiles should be estimated
-        when making distributional forecasts. Can alternatively be provided with the keyword
-        argument ``quantiles``. If ``None``, defaults to ``[0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9]``.
+        where ``L`` ranges from 0 to 50 (Note: higher values of ``L`` correspond to fewer print statements, opposite
+        of verbosity levels).
+
+    # TODO: Why not move Other Parameters to Parameters? At least `ignore_time_index` and `validation_splitter` seem relevant
```
canerturkmen, line 82: I think convention is to have `**kwargs` in Other Parameters. If you mean we can move ignore_time_index and validation_splitter to named keyword arguments, I agree. (Needless nit: as long as their default values are also specified in the constructor definition).

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -86,37 +95,38 @@ class inheriting from ``AbstractTrainer`` which controls training of multiple mo
...
 
 
+    # TODO: Why is only target listed under Attributes and not, e.g., eval_metric? Do we even need to mention anything here?
     Attributes
     ----------
     target : str
-        Name of column in training/validation data that contains the target time-series value to be predicted. If
-        not specified explicitly during :meth:`~autogluon.timeseries.TimeSeriesPredictor.fit`, this will default to
-        ``"target"``.
+        Name of column in training/validation data that contains the target time series value to be predicted.
```
canerturkmen, line 117: I see now that this should be "if not specified in the constructor," but why are we removing the explanation? This is an attribute (property) and does not really "default" to anything?

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -209,39 +218,48 @@ def fit(
             used to compute the validation scores. Note that only the last ``prediction_length`` time steps of each
             time series are used for computing the validation score.
 
-            If ``None``, AutoGluon will split :attr:`train_data` into training and tuning subsets using
+            If None, AutoGluon will split :attr:`train_data` into training and tuning subsets using
```
canerturkmen, line 221: out of curiosity, why remove the backticks? we are referring to the keyword None after all.

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -209,39 +218,48 @@ def fit(
...
             ``self.validation_splitter``. If ``tuning_data`` is provided, ``self.validation_splitter`` will be ignored.
             See the description of ``validation_splitter`` in the docstring for
             :class:`~autogluon.timeseries.TimeSeriesPredictor` for more details.
+
+            # TODO: Mention that this is a good default that users should not override?
```
canerturkmen, line 226: Up to you on both TODOs. We can add links when the tutorials are there.

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -209,39 +218,48 @@ def fit(
...
-            run until all models have completed training.
-        presets : str, default = None
+            Approximately how long :meth:`~autogluon.timeseries.TimeSeriesPredictor.fit` will run (wall-clock time in
+            seconds). If not specified, :meth:`~autogluon.timeseries.TimeSeriesPredictor.fit` will run until all models
+            have completed training.
+        presets : str, default = "medium_quality"
             Optional preset configurations for various arguments in
             :meth:`~autogluon.timeseries.TimeSeriesPredictor.fit`.
 
+            # TODO: Better explanation of how to choose presets
```
canerturkmen, line 236: we can give an itemized list of what each preset will do as in Tabular.

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -209,39 +218,48 @@ def fit(
...
             Can significantly impact predictive accuracy, memory footprint, inference latency of trained models,
             and various other properties of the returned predictor. It is recommended to specify presets and avoid
             specifying most other :meth:`~autogluon.timeseries.TimeSeriesPredictor.fit` arguments or model
             hyperparameters prior to becoming familiar with AutoGluon. For example, set ``presets="best_quality"``
             to get a high-accuracy predictor, or set ``presets="low_quality"`` to get a toy predictor that
             trains quickly but lacks accuracy.
+            # TODO: This looks like a legacy sentence, remove it? Are we referring to the `hyperparameters` kwarg here?
```
canerturkmen, line 243: We can remove if we add a list.

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -209,39 +218,48 @@ def fit(
...
             Any user-specified arguments in :meth:`~autogluon.timeseries.TimeSeriesPredictor.fit` will
             override the values used by presets.
 
+            # TODO: Should we even mention low_quality_hpo?
```
canerturkmen, line 247: I think we could even remove the preset low quality HPO :)

canerturkmen: and I'm happy to just do this as a breaking change as I doubt if any users depend on it

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -209,39 +218,48 @@ def fit(
...
             Available presets are "best_quality", "high_quality", "good_quality", "medium_quality", "low_quality",
             and "low_quality_hpo". Details for these presets can be found in
             ``autogluon/timeseries/configs/presets_configs.py``. If not provided, user-provided values for other
             arguments (specifically, ``hyperparameters`` and ``hyperparameter_tune_kwargs`` will be used (defaulting
             to their default values specified below).
         hyperparameters : str or dict, default = "default"
+            # TODO: Rework this section, similar to TabularPredictor
```
canerturkmen, line 254: 👍

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -250,15 +268,16 @@ def fit(
             values which are given in``autogluon/timeseries/trainer/models/presets.py``. Specific hyperparameter
             choices for each of the recommended models can be found in the references.
         hyperparameter_tune_kwargs : str or dict, default = None
-            # TODO
+            # TODO ?????
```
canerturkmen, line 271: Can wait until we overhaul HPO and Ray? But I'm OK to remove the TODO and just add a generic explanation with values it can take for now.

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -390,15 +409,15 @@ def predict(
...
+        """Return quantile and mean forecasts for the given dataset, starting from the end of each time series.
 
         Parameters
         ----------
         data : TimeSeriesDataFrame
             Time series data to forecast with.
-        model : str, default=None
-            Name of the model that you would like to use for forecasting. If None, it will by default use the
-            best model from trainer.
+        model : str or None, default = None
```
canerturkmen, line 418: If we're being nitpicky with this, the style guide says to use:
```
model : str, optional
```
https://numpydoc.readthedocs.io/en/latest/format.html#parameters

canerturkmen (commented):
Looks super. Thanks! I'm OK to merge when we can address the comments but I will read in detail once we have presets and hyperparams.

timeseries/src/autogluon/timeseries/predictor.py
shchur: To rephrase my question a bit: A `TimeSeriesPredictor` object has many other attributes (e.g., `eval_metric`, `quantile_levels`, `path`) that behave exactly like `target` - they are specified in the constructor and have some default values. Is there some reason why only `target` is mentioned under `Attributes`, but the others are not?

I feel like we can safely remove this, as we don't expect users to use these attributes in their code.

timeseries/src/autogluon/timeseries/predictor.py
shchur: My bad, fixed this.

timeseries/src/autogluon/timeseries/predictor.py
shchur: How about we postpone the updated docstrings for `presets`, `hyperparameters` and `hyperparameter_tune_kwargs` to another PR (after we incorporate PyTorch models / TFT)?

canerturkmen: works for me 👍

timeseries/src/autogluon/timeseries/predictor.py
canerturkmen: Ah yes, fair. OK to remove.

canerturkmen (approved):

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -213,45 +206,40 @@ def fit(
...
             override the values used by presets.
 
-            # TODO: Should we even mention low_quality_hpo?
-            Available presets are "best_quality", "high_quality", "good_quality", "medium_quality", "low_quality",
-            and "low_quality_hpo". Details for these presets can be found in
-            ``autogluon/timeseries/configs/presets_configs.py``. If not provided, user-provided values for other
-            arguments (specifically, ``hyperparameters`` and ``hyperparameter_tune_kwargs`` will be used (defaulting
-            to their default values specified below).
+            Available presets are "best_quality", "high_quality", "good_quality", "medium_quality", and "low_quality".
+            Details for these presets can be found in ``autogluon/timeseries/configs/presets_configs.py``. If not
```
canerturkmen, line 231: Is this the case? If not provided, doesn't it "default" to medium_quality?

shchur: As far as I understand, currently the default is `presets = None, hyperparameters = None`. In this case `hyperparameters` are set to `"default"`, so, technically, the presets are not used and exactly what's described in this docstring happens. This, however, is equivalent to setting `presets = "medium_quality"`. I think this is the same logic as in tabular.

Does this make sense, or do you think we can improve the phrasing here?

shchur: The correct signature here would be `presets : str, optional`

canerturkmen: I see. I agree with `presets : str, optional`. Thanks!

shchur:
Ready to be merged once the tests pass

PR #2123, opened 2022-09-11
Simplify XGBoost HPO search space to be stronger with fewer trials

Innixma:
*Issue #, if available:*

*Description of changes:*
Tested on adult dataset and saw that 50 trials with old config 0/50 were able to get better test score than default. 10/50 trials were able to beat default with new config. Until we do more testing with HPO, it is better to have simple search spaces as it avoids wasting most trials due to a poor hyperparameter value being present in a trial. This change aligns XGBoost search space with LightGBM search space in terms of complexity.

(Side note: @yinweisu this old search space is maybe why you saw parallel HPO was worse for trees, because it didn't use the default config during HPO and that was stronger than the vast majority of random configs.)

### DIFF
--- a/tabular/src/autogluon/tabular/models/xgboost/hyperparameters/searchspaces.py
+++ b/tabular/src/autogluon/tabular/models/xgboost/hyperparameters/searchspaces.py
@@ -25,6 +25,7 @@ def get_base_searchspace():
         'max_depth': Int(lower=3, upper=10, default=6),
         'min_child_weight': Int(lower=1, upper=5, default=1),
         'colsample_bytree': Real(lower=0.5, upper=1.0, default=1.0),
+        # Below lines are commented out as they made search worse. Refine ranges before considering reintroducing these hyperparameters.
         # 'gamma': Real(lower=0, upper=5, default=0),
         # 'subsample': Real(lower=0.5, upper=1.0, default=1.0),
         # 'reg_alpha': Real(lower=0.0, upper=10.0, default=0.0),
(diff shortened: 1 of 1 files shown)


tabular/src/autogluon/tabular/models/xgboost/hyperparameters/searchspaces.py
```diff
@@ -24,11 +24,11 @@ def get_base_searchspace():
...
         'min_child_weight': Int(lower=1, upper=5, default=1),
-        'gamma': Real(lower=0, upper=5, default=0.01),
-        'subsample': Real(lower=0.5, upper=1.0, default=1.0),
         'colsample_bytree': Real(lower=0.5, upper=1.0, default=1.0),
-        'reg_alpha': Real(lower=0.0, upper=10.0, default=0.0),
-        'reg_lambda': Real(lower=0.0, upper=10.0, default=1.0),
+        # 'gamma': Real(lower=0, upper=5, default=0),
+        # 'subsample': Real(lower=0.5, upper=1.0, default=1.0),
+        # 'reg_alpha': Real(lower=0.0, upper=10.0, default=0.0),
+        # 'reg_lambda': Real(lower=0.0, upper=10.0, default=1.0),
```
gidler, line 31: I support these changes based on the performance data. I just want to double check that we prefer to leave these commented out rather than deleted (in which case we could still recover them from git history). If we leave them commented out, I like the idea of adding a inline comment about why they need to still be in the code while not being used. TIOLI

gidler (approved):
One suggestion, but changes LGTM

gradientsky (approved):

tabular/src/autogluon/tabular/models/xgboost/hyperparameters/searchspaces.py
yinweisu: +1

yinweisu (approved):

yinweisu:
I'll work investigate on default parameter for tune search and send a PR this week.

yinweisu:
CI for Push events passed. The failure on Pull Request event looks unrelevent

tabular/src/autogluon/tabular/models/xgboost/hyperparameters/searchspaces.py
Innixma: Good point, added inline comment about why they are disabled.

PR #2124, opened 2022-09-12
Cleanup Temperature Scaling code

Innixma:
*Issue #, if available:*

*Description of changes:*
- Moved calibration method from predictor to trainer
- Removed temperature_scalar attribute from AbstractModel, moved to params_aux so it can propagate correctly to refit_full models.
- No functional change except that now refit model will inherit the temperature scaling term from the parent (this is good, previously it was not inherited which caused inconsistency between settings which applied calibration prior to refitting and after refitting). No impact to presets, only if user manually called refit_full.


yinweisu (approved):
LGTM!

PR #2125, opened 2022-09-12
Optimize data processor logic

zhiqiangdon:
1. Create data processors after models. In this way, we can directly use the pre-trained config from the model creation without creating the pre-trained model/config again. This is useful for `hf_text`, `timm_image`, and `mmdet_image` models since we need to use some default hyper-parameters in creating the data processors.
2. Refactor `init_data_processors` into `create_data_processor` and `create_fusion_data_processors` to improve modular design. `create_data_processor` can be used more flexibly.
3. Fix the `select_model` by constraining the numerical_transformer and categorical_transformer. Otherwise, it may encounter errors for numerical_mlp and categorical_mlp. We may need to refactor ft_transformer to remove this hack.
4. Sort model names before creating models. Otherwise, different model name orders may result in different fusion models.
 


multimodal/src/autogluon/multimodal/models/mmocr_text_detection.py
```diff
@@ -38,6 +38,8 @@ class MMOCRAutoModelForTextDetection(nn.Module):
     Refer to https://github.com/open-mmlab/mmocr
     """
 
+    data_types = [IMAGE]
```
bryanyzhu, line 41: Data types for other OCR use cases could be [IMAGE, TEXT] too, will that be a problem in the future?

bryanyzhu (commented):
LGTM, just one minor comment about future compatibility, thank you.

multimodal/src/autogluon/multimodal/models/mmocr_text_detection.py
zhiqiangdon: Good point! I have removed this class attribute. Users can still customize the data_types for each model.

multimodal/src/autogluon/multimodal/data/process_label.py
```diff
@@ -15,15 +16,15 @@ class LabelProcessor:
 
     def __init__(
         self,
-        prefix: str,
+        model: nn.Module,
```
sxjscience, line 19: Ideally, this file should not depend on `torch` since it's only about data processing.

We should allow users to use classes in this file without torch.

zhiqiangdon: In AutoMM, data processors are created for models. In data processors, we may need to use model configs (e.g., timm, hugginface, mmdet). This logic seem reasonable. Besides, `torch` is a basic package in AutoMM, which is as basic as `numpy`. 

`process_text` and `process_image` are conditioned on the third-party packages like `timm`, `transformers`, `mmcv`, etc. We may move them into the model definition. In this way, the processor classes are clearer.

FANGAreNotGnu (approved):
LGTM, with this change we can fix the current limitation that eval_batch_size_ratio has to be 1 for MMDet, thanks!

bryanyzhu (approved):
Thanks for addressing the comments, LGTM.

multimodal/src/autogluon/multimodal/data/process_label.py
sxjscience: Would it make sense to support both prefix and model as arguments to the class initializer?

```python
def __init__(
   self,
   prefix: str,
   model: nn.Module, 
   ...
)
```

I'm raising this concern because of the deployment scenario. Ideally, the user should be able to create these processors without knowing the model, and creating the model will incur additional overhead.

sxjscience: If we consider the deployment scenario, we may ideally ensure the processors do not depend on the models. This is largely a two-way door though. We can refactor it in a later stage. Just need to bear it in mind.

sxjscience: We can convert model to a deployable format like ONNX / torchscript. However, if we add another layer of dependency, the logic will be quite complicated.

zhiqiangdon: Thanks for the reminding. 

1. Only the processor creation depends on model. I guess we probably only need to load the processors in deployment. 
2. The current image and text processor require the packages timm, huggingface, mmcv, which I think are more complicated than the plain torch package. By using the model as input, we can further remove timm, huggingface, mmcv from processors.

sxjscience (approved):
LGTM overall. We can revisit the concern about deployment once we have concrete scenarios.

PR #2127, opened 2022-09-12
Update timeseries quickstart tutorial

shchur:
*Description of changes:*
Clarify a few things, improve layout and slightly change structure of the quickstart tutorial for timeseries forecasting.

For convenience, [a PDF of the rendered tutorial](https://github.com/awslabs/autogluon/files/9549540/ts_quickstart.pdf) (this doesn't include the output of the code cells, need to wait for the build to finish to see that).


docs/tutorials/timeseries/forecasting-quickstart.md
```diff
@@ -1,208 +1,253 @@
...
 
+In this tutorial we work with COVID case data as of April 2020.
+Our goal is to forecast the cumulative number of confirmed COVID cases for each country in the dataset given the past observations.
 
+We load the dataset from an [AWS S3 bucket](https://aws.amazon.com/s3/) as a `pandas.DataFrame`
 ```{.python .input}
 df = pd.read_csv(
     "https://autogluon.s3-us-west-2.amazonaws.com/datasets/CovidTimeSeries/train.csv",
-    parse_dates=["Date"],
+    parse_dates=["Date"],  # make sure that pandas parses the dates
```
Innixma, line 50: Is this really necessary? Shouldn't we be able to convert to the correct type automatically later on without having the user do so during data load? Why not load as a TimeSeriesDataFrame or TabularDataset directly?

docs/tutorials/timeseries/forecasting-quickstart.md
```diff
@@ -1,208 +1,253 @@
...
 )
+df.head()
+```
+Each row of the data frame contains a single observation (timestep) of a single time series represented by
 
-train_data = TimeSeriesDataFrame.from_data_frame(
+- unique ID of the time series — in our case, name of the country (`"name"`)
```
Innixma, line 56: Are we guaranteed to have an ID? What if there is only 1 time series? What happens if they don't have an ID? These questions should be answered in the quick start.

Innixma (commented):
LGTM, had a few comments, will have @canerturkmen provide approval for merge.

docs/tutorials/timeseries/forecasting-quickstart.md
shchur: Currently the API provides options to construct a `TimeSeriesDataFrame` from a pandas DataFrame or from a list of dictionaries. I agree that it would make sense to automatically load from disk, similar to a `TabularDataset`. Will discuss this with Caner and see if it's feasible to do this by 0.6.0.

canerturkmen (approved):
All looks good to me! Thanks! Feel free to merge with or without the additional points mentioned.

PR #2132, opened 2022-09-14
add shift to eda

jsharpna:
*Description of changes:*

I have added shift to eda via `XShiftDetector` analysis and `XShiftSummary` visualization.


eda/src/autogluon/eda/analysis/shift.py
```diff
@@ -0,0 +1,356 @@
...
+            sample_label = self.sample_label
+            assert sample_label in data.columns, "sample_label needs to be a column of data"
+            assert self.split, "sample_label requires the split parameter"
+            data = data.copy() # makes a copy
+        else:
+            assert len(data) == 2, "Data needs to be tuple/list of (source, target) if sample_label is None"
+            data = self._make_source_target_label(data, self.sample_label) # makes a copy
+        if data.index.has_duplicates:
+            data.index = pd.RangeIndex(data.shape[0])
+        train = data.sample(frac=self.split)
```
jsharpna, line 88: does this shuffle?

jsharpna: it does shuffle https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.sample.html

sxjscience: We may still need to keep a seed to ensure that it will be reproducible.

gradientsky:
Notebooks should not be checked-in with the rest of the code

eda/src/autogluon/eda/analysis/shift.py
```diff
@@ -0,0 +1,347 @@
...
+        Training/test split proportion for classifier 2 sample test
+    classifier_kwargs : dict, default = {}
+        The kwargs passed to the classifier, a member of classifier_class
+    """
+    def __init__(self,
+                 classifier_class,
+                 sample_label='xshift_label',
+                 eval_metric=balanced_accuracy,
+                 split=0.5,
+                 compute_fi = True,
```
gradientsky, line 44: no docs for `compute_fi` above

eda/src/autogluon/eda/analysis/shift.py
```diff
@@ -0,0 +1,347 @@
...
+        if isinstance(data, pd.DataFrame):
+            sample_label = self.sample_label
+            assert sample_label in data.columns, "sample_label needs to be a column of data"
+            assert self.split, "sample_label requires the split parameter"
+            data = data.copy() # makes a copy
+        else:
+            assert len(data) == 2, "Data needs to be tuple/list of (source, target) if sample_label is None"
+            data = self._make_source_target_label(data, self.sample_label) # makes a copy
+        if data.index.has_duplicates:
+            data.index = pd.RangeIndex(data.shape[0])
```
gradientsky, line 87: maybe just always do `data = data.reset_index(drop=True)`

eda/src/autogluon/eda/analysis/shift.py
```diff
@@ -0,0 +1,347 @@
...
+        self.has_fi = None
+        self.compute_fi = compute_fi
+
+    @staticmethod
+    def _make_source_target_label(data, sample_label):
+        """Turn a source, target pair into a single dataframe with label column"""
+        source, target = data[0].copy(), data[1].copy()
+        source.loc[:,sample_label] = 0
+        target.loc[:,sample_label] = 1
+        data = pd.concat((source, target))
```
gradientsky, line 65: `data = pd.concat((source, target), ignore_index=True)` to avoid overlapping/duplicate indices.

eda/src/autogluon/eda/analysis/shift.py
```diff
@@ -0,0 +1,347 @@
...
+            sample_label = self.sample_label
+            assert sample_label in data.columns, "sample_label needs to be a column of data"
+            assert self.split, "sample_label requires the split parameter"
+            data = data.copy() # makes a copy
+        else:
+            assert len(data) == 2, "Data needs to be tuple/list of (source, target) if sample_label is None"
+            data = self._make_source_target_label(data, self.sample_label) # makes a copy
+        if data.index.has_duplicates:
+            data.index = pd.RangeIndex(data.shape[0])
+        train = data.sample(frac=self.split)
```
gradientsky, line 88: Let's fix the seed (`random_state=0`) so it is reproducible; otherwise it might be confusing to the users to get different results on repeated runs.
Open question: does it make sense to do stratification by label?

eda/src/autogluon/eda/analysis/shift.py
```diff
@@ -0,0 +1,347 @@
...
+        yhat = self.classifier.predict(test)
+        self.test_stat = self.eval_metric(test[self.sample_label], yhat)
+        self.has_fi = (getattr(self.classifier, "feature_importance", None) is not None)
+        if self.has_fi and self.compute_fi:
+            self._test = test  # for feature importance
+        self._is_fit = True
+
+    @post_fit
+    def _pvalue_half_permutation(self,
+                                 num_permutations=1000):
```
gradientsky, line 100: do we need `1000` rounds here? What is the number of measurements we need to get the statistically-significant results?

eda/src/autogluon/eda/analysis/shift.py
```diff
@@ -0,0 +1,347 @@
...
+        """The half permutation method for computing p-values.
+        See Section 9.1 of https://arxiv.org/pdf/1602.02210.pdf
```
gradientsky, line 102: It's section 9.2

eda/src/autogluon/eda/analysis/shift.py
```diff
@@ -0,0 +1,347 @@
...
+                self._test[self.sample_label],
+                perm_yhat
+            )
+            perm_stats.append(perm_test_stat)
+        pval = (self.test_stat <= np.array(perm_stats)).mean()
+        return pval
+
+    @post_fit
+    def pvalue(self,
+               num_permutations: int=1000):
```
gradientsky, line 118: ditto: do we need `1000` rounds to get the conclusion?

eda/src/autogluon/eda/analysis/shift.py
```diff
@@ -0,0 +1,347 @@
...
+
+        Returns
+        -------
+        One of ['detected', 'not detected']
+        """
+        # default teststat_thresh by metric
+        if self.p_value < self.pvalue_thresh:
+            return 'detected'
+        else:
+            return 'not_detected'
```
gradientsky, line 284: `True/False` and let the rendering layer to decide what to display?

eda/src/autogluon/eda/analysis/shift.py
```diff
@@ -0,0 +1,347 @@
...
+        -------
+        pd.DataFrame of feature importances
+        """
+        assert self.has_fi, "Classifier class does not have feature_importance method"
+        assert self.compute_fi, "Set compute_fi to True to compute feature importances"
+        fi_scores = self.classifier.feature_importance(self._test)
+        return fi_scores
+
+
+class C2STShiftDetector:
```
gradientsky, line 147: This class is redundant: data preprocessing logic can be inside `Classifier2ST` class while `decision()/results()` can be in Analysis/Visualization classes.

eda/src/autogluon/eda/analysis/shift.py
```diff
@@ -0,0 +1,347 @@
...
+        if 'label' in args:
+            label = args['label']
+        else:
+            label = None
+        tst = C2STShiftDetector(classifier_class=self.classifier_class,
+                                label=label,
+                                compute_fi=self.compute_fi,
+                                classifier_kwargs=self.classifier_kwargs)
+        assert 'train_data' in args, 'train_data required as arg'
+        assert 'test_data' in args, 'test_data required as arg'
```
gradientsky, line 342: I don't think we need to hard-stop if an analysis don't have enough inputs. Warning might suffice and do nothing if inputs are not provided. I already thought that we might have generic method: something like `can_fit(...) -> bool` as a pre-check.

eda/src/autogluon/eda/analysis/shift.py
jsharpna: Thanks, this is a holdout from when I outputted anomalies and needed to map back to the original index.

eda/src/autogluon/eda/analysis/shift.py
jsharpna: It could make sense to stratify by label, that would ensure that there is enough of both source and target.  Especially helpful when there are few target samples (a reasonable scenario).

eda/src/autogluon/eda/analysis/shift.py
jsharpna: nice

eda/src/autogluon/eda/analysis/shift.py
jsharpna: well, we can't resolve a very low pvalue otherwise, think of 1000 permutations at 0.01 level to mean 10 sample above the threshold.

eda/src/autogluon/eda/analysis/shift.py
jsharpna: yeah, this is also a holdout from before - if you think that it is important I can do this

eda/src/autogluon/eda/analysis/shift.py
jsharpna: let's talk about this to see if we want to add `can_fit`

eda/src/autogluon/eda/analysis/shift.py
```diff
@@ -0,0 +1,347 @@
...
+        Binary classification metric to use for the classifier 2 sample test
+    split : float, default = 0.5
+        Training/test split proportion for classifier 2 sample test
+    classifier_kwargs : dict, default = {}
+        The kwargs passed to the classifier, a member of classifier_class
+    """
+    def __init__(self,
+                 classifier_class,
+                 sample_label='xshift_label',
+                 eval_metric=balanced_accuracy,
```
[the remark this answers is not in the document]
gradientsky: Why not use `.predict_proba()`? IMO there is a big difference between very confident incorrect prediction (0.99999999 score for 0 while the label was 1) vs the one near threshold (0.50000001 score for 0 while label was 1) - this would be reflected differently in the `roc_auc` curve in ranking.

eda/notebooks/helpers.py
```diff
@@ -0,0 +1,34 @@
...
+    p_married = (0.5 + data_married.mean() - p_nonmarr) / data_married.mean()
+    train_p = data_married * p_married + (1 - data_married) * p_nonmarr
+    train_ind = np.random.binomial(1,train_p) == 1
+    train_cs = data[train_ind]
+    test_cs = data[~train_ind]
+    if val:
+        train_cs, val_cs = model_selection.train_test_split(train_cs)
+        return train_cs, val_cs, test_cs
+    else:
+        return train_cs, test_cs
```
gradientsky, line 34: Notebooks should not be checked-in

eda/src/autogluon/eda/analysis/shift.py
gradientsky: it is additional layer to maintain if we can remove it, we should

eda/tests/unittests/test_shift.py
```diff
@@ -0,0 +1,75 @@
+import pandas as pd
+import numpy as np
+import autogluon.eda.analysis as eda
+import autogluon.eda.auto as auto
+import autogluon.eda.visualization as viz
+
+s3_url = 'https://autogluon.s3.us-west-2.amazonaws.com/datasets/AdultIncomeBinaryClassification/'
```
gradientsky, line 7: nit: coding standards - naming - `S3_URL`

eda/src/autogluon/eda/analysis/shift.py
```diff
@@ -0,0 +1,256 @@
...
+        return self.all_keys_must_be_present(args, ['train_data', 'test_data'])
+
+    def _fit(self, state: AnalysisState, args: AnalysisState, **fit_kwargs):
+        """ Fit method.  `args` can contain
+        - 'train_data': pd.DataFrame, required
+        - 'test_data': pd.DataFrame, required
+        - 'label': str, optional
+            The Y variable that is to be predicted (if it appears in the train/test data then it will be removed)
+        """
+        self.can_handle(state, args)
```
gradientsky, line 89: This is not needed - `fit()` all will do this automatically and if the result is `True`, it'll pass control over to `_fit()`,

eda/src/autogluon/eda/analysis/shift.py
```diff
@@ -0,0 +1,256 @@
...
+    """
+    def __init__(self,
+                 classifier_class,
+                 sample_label='xshift_label',
+                 eval_metric=roc_auc,
+                 split=0.5,
+                 compute_fi=True,
+                 classifier_kwargs={},
+                 ):
+        classifier_kwargs.update({'label': sample_label, 'eval_metric': eval_metric})
```
gradientsky, line 154: We should not mutate incoming parameters - please do updates to a `deepcopy`

eda/src/autogluon/eda/analysis/shift.py
```diff
@@ -0,0 +1,256 @@
...
+            assert sample_label in data.columns, "sample_label needs to be a column of data"
+            assert self.split, "sample_label requires the split parameter"
+            data = data.copy() # makes a copy
+        else:
+            assert len(data) == 2, "Data needs to be tuple/list of (source, target) if sample_label is None"
+            data = self._make_source_target_label(data, self.sample_label) # makes a copy
+        if data.index.has_duplicates:
+            data = data.reset_index(drop=True)
+        train, test, y_train, y_test = generate_train_test_split(data.drop(columns=[self.sample_label]),
+                                                                 data[self.sample_label],
```
gradientsky, line 196: Out-of-the box `test_size` for the split is `0.1`. Since the only thing we are looking for is overfitting, I suggest increasing it to at least `0.3`.

gradientsky (approved):

PR #2135, opened 2022-09-15
Improve averaging checkpoints and creating model

zhiqiangdon:
1. Decouple weight sharing in checkpoint averaging. Convert weights to float64 to improve averaging robustness.
2. Refactor previous `create_model` into `create_model` and `create_fusion_model` to improve modularization. `create_model` is only in charge of creating a single model, which will be used in creating models in the siamese matcher.

### DIFF
--- a/multimodal/src/autogluon/multimodal/predictor.py
+++ b/multimodal/src/autogluon/multimodal/predictor.py
@@ -86,7 +86,7 @@ from .utils import (
     compute_num_gpus,
     compute_score,
     create_fusion_data_processors,
-    create_model,
+    create_fusion_model,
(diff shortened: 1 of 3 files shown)


sxjscience (approved):
LGTM. Thanks for fixing this. We need to rerun our experiments with larger backbones since the average state becomes float64. @Raldir

bryanyzhu (approved):
Thank you for the fix, LGTM.

suzhoum (approved):
The fix and refactor look great! Thanks!

FANGAreNotGnu (approved):
LGTM, thanks!

PR #2140, opened 2022-09-16
Fix data shuffling for 1-gpu training

zhiqiangdon:
Make sure data is shuffled automatically in the single-gpu training.


FANGAreNotGnu:
LGTM! Nice catch, thanks!

FANGAreNotGnu (approved):

Raldir:
Just to leave it here since it took me a while to figure it out: If wanting to NOT shuffle the data in  multi-gpu environment (ddp), setting the above flag is not enough. It is required to set `replace_sampler_ddp=False` as an argument to the Trainer.

PR #2141, opened 2022-09-17
Fix warning and refactor losses

zhiqiangdon:
1. Fix a gpu num warning. Raise the warning under the correct condition.
2. Add a new constant `HF_MODELS` to gather all the huggingface models.
3. Move all the customized loss definitions into one file. We will only need to add new losses into this file rather than keeping adding new loss files in future.


FANGAreNotGnu (commented):
We will have losses for different tasks, and putting them altogether might be confusing, e.g. iou_loss (only for bbox prediction) and rkd_loss (only for distillation) in one file.

zhiqiangdon:
> We will have losses for different tasks, and putting them altogether might be confusing, e.g. iou_loss (only for bbox prediction) and rkd_loss (only for distillation) in one file.

Given that AutoMM is for multimodal and multitask, having different types of losses is natural. 

1. Putting losses together is easier to use. We/users only need import from a fixed path `multimodal.optimization.losses`. No need to remember specific names of loss files.
2. Organizing losses together gives a cleaner design. A single loss seems not important enough to stand side-by-side with training loop files. Adding many loss files in the `optimization` module sacrifices the code readability, which is not good way to scale up. 
3. Some repos, like [`sentencetransformer`](https://github.com/UKPLab/sentence-transformers/tree/master/sentence_transformers/losses), have a specific losses folder containing all the loss files. If necessary, we may consider adding a `losses` folder in `optimization` and put the separate loss files there. I am not sure whether we need to do so for now.

bryanyzhu (approved):
LGTM!

sxjscience (approved):
LGTM

suzhoum (approved):
LGTM!

PR #2142, opened 2022-09-17
Fix feature_extraction pipeline for text

sxjscience:
*Issue #, if available:*

*Description of changes:*

Fix the feature_extraction pipeline. There are two bugs:

- [x] The `feature_extraction` presets has not been registered.
- [x] If the data only contains a single row,  the previous implementation will infer the column as a categorical column. This PR fixed it by chaging the fallback_type to be the allowable type.

### DIFF
--- a/multimodal/tests/unittests/test_pipeline_feature_extraction.py
+++ b/multimodal/tests/unittests/test_pipeline_feature_extraction.py
@@ -23,6 +23,6 @@ def test_sentence_transformer_embedding(model_name):
     outputs_case2_from_df = predictor.extract_embedding(pd.DataFrame(case2))
     outputs_case2_from_dict = predictor.extract_embedding(case2)
     npt.assert_allclose(outputs_case2_from_df["sentence"],
-                        outputs_case2_from_dict["sentence"])
+                        outputs_case2_from_dict["sentence"], 1E-3, 1E-3)
     npt.assert_allclose(outputs_case2_from_df["sentence"][:1],
-                        outputs_case1_from_df["sentence"])
+                        outputs_case1_from_df["sentence"], 1E-3, 1E-3)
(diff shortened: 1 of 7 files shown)


FANGAreNotGnu (approved):
LGTM with a minor comments: shall we add a warning if we infer a column as categorical? It could be hard to debug if there is a wrong infer since this is very implicit.

multimodal/tests/unittests/test_pipeline_feature_extraction.py
```diff
@@ -0,0 +1,26 @@
...
+
+    outputs_case1_from_df = predictor.extract_embedding(pd.DataFrame(case1))
+    outputs_case1_from_dict = predictor.extract_embedding(case1)
+
+    npt.assert_allclose(outputs_case1_from_dict["sentence"], outputs_case1_from_df["sentence"])
+
+    outputs_case2_from_df = predictor.extract_embedding(pd.DataFrame(case2))
+    outputs_case2_from_dict = predictor.extract_embedding(case2)
+    npt.assert_allclose(outputs_case2_from_df["sentence"], outputs_case2_from_dict["sentence"], 1e-3, 1e-3)
+    npt.assert_allclose(outputs_case2_from_df["sentence"][:1], outputs_case1_from_df["sentence"], 1e-3, 1e-3)
```
bryanyzhu, line 26: Do we want to make the threshold smaller like 1e-5 or 1e-6?

bryanyzhu (approved):
LGTM!

multimodal/tests/unittests/test_pipeline_feature_extraction.py
sxjscience: I found that 1E-5 / 1E-6 cannot pass and have to pick 1E-3.

Issue #2134, opened 2022-09-14
Have a property in TabularPredictor to track the original feature columns


yinweisu:
TabularPredictor will drop columns sometime, and this would bring trouble to model deployment when user omits the header info to minimize payload size. In such case, we would need to attach original column names to the inference data but would have trouble when TabularPredictor drop features during training. Therefore, we need to store the original feature info somewhere

Some notes:
* A property inside TabularPredictor instead of metadata saved along with CloudPredictor would allow users to deploy local trained TabularPredictor to cloud as well.
* We would want to keep the columns being stored in original order instead of sorted

PR #2144, opened 2022-09-19
Add Original Features

yinweisu:
*Issue #, if available:*
https://github.com/awslabs/autogluon/issues/2134

*Description of changes:*
* Add original training data features

### DIFF
--- a/tabular/src/autogluon/tabular/learner/abstract_learner.py
+++ b/tabular/src/autogluon/tabular/learner/abstract_learner.py
@@ -7,6 +7,7 @@ from collections.abc import Iterable
 import numpy as np
 import pandas as pd
 from pandas import DataFrame, Series
+from typing import List
 from sklearn.metrics import classification_report
 
 from autogluon.core.constants import BINARY, MULTICLASS, REGRESSION, QUANTILE, AUTO_WEIGHT, BALANCE_WEIGHT
@@ -76,7 +77,7 @@ class AbstractTabularLearner(AbstractLearner):
             raise ValueError('groups must be a string indicating the name of the column that contains the split groups. If you have a vector of split groups, first add these as an extra column to your data.')
 
     @property
-    def original_features(self):
+    def original_features(self) -> List[str]:
         """Original features user passed in before autogluon doing any processing"""
         return self._original_features
 
--- a/tabular/src/autogluon/tabular/learner/abstract_learner.py
+++ b/tabular/src/autogluon/tabular/learner/abstract_learner.py
@@ -61,6 +61,7 @@ class AbstractTabularLearner(AbstractLearner):
         self.label_cleaner: LabelCleaner = None
         self.feature_generator: PipelineFeatureGenerator = feature_generator
 
+        self._original_features = None
         self._pre_X_rows = None
         self._post_X_rows = None
         self._positive_class = positive_class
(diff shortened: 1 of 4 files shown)


tabular/src/autogluon/tabular/learner/default_learner.py
```diff
@@ -72,6 +72,8 @@ def _fit(self, X: DataFrame, X_val: DataFrame = None, X_unlabeled: DataFrame = N
         time_preprocessing_start = time.time()
         logger.log(20, 'Preprocessing data ...')
         self._pre_X_rows = len(X)
+        self._original_features = list(X)
+        self._original_features.remove(self.label)
```
Innixma, line 76: Not quite correct, this could contain additional features still such as sample weight and `self.groups` which wouldn't be present at test time. A better option is to record original features in the feature generator object, since this is the stage at which all of those extra columns have been removed and only the true features are present.

Innixma: Another option: Add self._original_features right before the call to fit feature generator, this will work identically to current implementation, but should be after the other columns are removed and only the real features are present. This might be a bit better than making the logic part of FeatureGenerator.

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -246,6 +246,10 @@ def quantile_levels(self):
     @property
     def eval_metric(self):
         return self._learner.eval_metric
+    
+    @property
+    def original_features(self):
+        return self._learner.original_features
```
Innixma, line 252: Add documentation:

 """Original features user passed in to fit before processing"""

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -246,6 +246,10 @@ def quantile_levels(self):
     @property
     def eval_metric(self):
         return self._learner.eval_metric
+    
+    @property
+    def original_features(self):
```
Innixma, line 251: Add exception if not fit.

Innixma, line 251: Add return type hint

tabular/src/autogluon/tabular/learner/abstract_learner.py
```diff
@@ -74,6 +75,11 @@ def __init__(self, path_context: str, label: str, feature_generator: PipelineFea
         if groups is not None and not isinstance(groups, str):
             raise ValueError('groups must be a string indicating the name of the column that contains the split groups. If you have a vector of split groups, first add these as an extra column to your data.')
 
+    @property
+    def original_features(self):
```
Innixma, line 79: Add return type hint

tabular/src/autogluon/tabular/learner/default_learner.py
```diff
@@ -213,6 +213,8 @@ def general_data_processing(self, X: DataFrame, X_val: DataFrame, X_unlabeled: D
             y_val = None
             w_val = None
 
+        self._original_features = list(X)
+        self._original_features.remove(self.label)
```
Innixma: Unnecessary, self.label is already removed from X at this point.

tabular/src/autogluon/tabular/learner/default_learner.py
```diff
@@ -213,6 +213,8 @@ def general_data_processing(self, X: DataFrame, X_val: DataFrame, X_unlabeled: D
             y_val = None
             w_val = None
 
+        self._original_features = list(X)
```
Innixma, line 216: Might be better to make more easily to understand via `list(X.columns)`

Innixma (approved):
Approving assuming comments are resolved.

PR #2161, opened 2022-09-22
Modularize realtime inference

zhiqiangdon:
This PR is to modularize some logic shared by the realtime inference and dataloader. This modularization can also benefit the ONNIX export by @FANGAreNotGnu.

1. Add util `apply_df_preprocessor` shared by the realtime inference and dataloader.
2. Add util `apply_data_processor` shared by the realtime inference and dataloader.
3. Add util `infer_batch` for realtime inference.

### DIFF
--- a/multimodal/src/autogluon/multimodal/utils.py
+++ b/multimodal/src/autogluon/multimodal/utils.py
@@ -29,6 +29,7 @@ from scipy.special import softmax
 from sklearn.metrics import f1_score
 from sklearn.preprocessing import LabelEncoder
 from torch import nn
+from torch.nn.modules.loss import _Loss
 
 from autogluon.core.metrics import get_metric
 from autogluon.core.utils.loaders import load_pd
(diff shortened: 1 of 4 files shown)


bryanyzhu (approved):
Great!

multimodal/src/autogluon/multimodal/utils.py
```diff
@@ -2781,3 +2782,40 @@ def move_to_device(obj: Union[torch.Tensor, nn.Module, Dict, List], device: torc
...
+    batch_size = len(batch[next(iter(batch))])
+    if 1 < torch.cuda.device_count() <= batch_size:
+        model = nn.DataParallel(model)
+    model.to(device).eval()
+    batch = move_to_device(batch, device=device)
+    with torch.autocast(device_type=device_type, dtype=precision):
+        with torch.no_grad():
+            output = model(batch)[model.prefix]
+
+    if isinstance(loss_func, nn.BCEWithLogitsLoss):
```
sxjscience, line 2818: Should we put it under `torch.no_grad()?`

multimodal/src/autogluon/multimodal/utils.py
```diff
@@ -2781,3 +2782,40 @@ def move_to_device(obj: Union[torch.Tensor, nn.Module, Dict, List], device: torc
...
+    precision
+        The desired precision used in inference.
+    loss_func
+        The loss function used in training the model.
+
+    Returns
+    -------
+    Model output.
+    """
+    precision = infer_precision(num_gpus=1, precision=precision, as_torch=True)
```
sxjscience, line 2806: When reading the code, I think we should set `num_gpus` to `toch.cuda.device_count()`.

multimodal/src/autogluon/multimodal/utils.py
zhiqiangdon: Done.

multimodal/src/autogluon/multimodal/utils.py
zhiqiangdon: Good suggestion! Updated.

multimodal/src/autogluon/multimodal/utils.py
zhiqiangdon: BTW, I am thinking about a form way to handle this kind of post-processing like `nn.BCEWithLogitsLoss` . The current way is kind of a hack and have several repeated code snippets, which is not easy to maintain.

suzhoum (approved):
LGTM! Thanks for answering my questions!

PR #2164, opened 2022-09-25
Add the model post-processing function

zhiqiangdon:
Add `model_postprocess_fn` to process the model output for inference. This is to handle the cases where the training loss function includes some operations that are also required to transform predictions in inference. For example, if we use the `nn.BCEWithLogitsLoss` on a regression task, we need to use sigmoid to transform the logits in inference.

### DIFF
--- a/multimodal/src/autogluon/multimodal/models/utils.py
+++ b/multimodal/src/autogluon/multimodal/models/utils.py
@@ -3,8 +3,10 @@ from typing import Dict, List, Optional, Tuple
 
 import torch
 from torch import nn
+from torch.nn.modules.loss import _Loss
 from transformers import AutoConfig, AutoModel
 
+from ..constants import LOGITS, REGRESSION
 from .adaptation_layers import IA3Linear, LoRALinear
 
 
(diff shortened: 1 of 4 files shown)


multimodal/src/autogluon/multimodal/models/utils.py
```diff
@@ -543,3 +545,44 @@ def get_hf_config_and_model(checkpoint_name: str, pretrained: Optional[bool] = T
         model = AutoModel.from_config(config)
 
     return config, model
+
+
+def custom_sigmoid(output: Dict):
```
sxjscience, line 550: How about to change the name to `apply_sigmoid`?

zhiqiangdon: Thanks! Updated.

sxjscience:
Noticed a conflict about ` multimodal/src/autogluon/multimodal/models/utils.py ` . The approval is conditioned on 1) conflict has been resolved, 2) PR has passed the CI.

bryanyzhu (approved):
Looks good, thank you. Maybe you want to merge the utils refactor PR first.

suzhoum (approved):
LGTM!

PR #2167, opened 2022-09-26
Deprecate high, good and low quality presets for time series

shchur:
*Description of changes:*
- Update `default` and `default_hpo` hyperparameter configurations
- Remove `high_quality` and `good_quality` presets for time series (these were rather meaningless, only changing the # of random search runs in HPO)
- Rename `low_quality` -> `fast_training` that now only contains local models

### DIFF
--- a/timeseries/src/autogluon/timeseries/configs/presets_configs.py
+++ b/timeseries/src/autogluon/timeseries/configs/presets_configs.py
@@ -13,9 +13,9 @@ TIMESERIES_PRESETS_CONFIGS = dict(
         },
     },
     medium_quality={"hyperparameters": "default"},
-    local_models={"hyperparameters": "local_models"},
+    fast_training={"hyperparameters": "local_only"},
     # TODO: Remove deprecated presets below
     high_quality={"hyperparameters": "default"},
     good_quality={"hyperparameters": "default"},
-    low_quality={"hyperparameters": "local_models"},
+    low_quality={"hyperparameters": "local_only"},
 )
--- a/timeseries/src/autogluon/timeseries/configs/presets_configs.py
+++ b/timeseries/src/autogluon/timeseries/configs/presets_configs.py
@@ -12,8 +12,10 @@ TIMESERIES_PRESETS_CONFIGS = dict(
             "num_trials": 20,
         },
     },
+    medium_quality={"hyperparameters": "default"},
+    local_models={"hyperparameters": "local_models"},
+    # TODO: Remove deprecated presets below
     high_quality={"hyperparameters": "default"},
     good_quality={"hyperparameters": "default"},
-    medium_quality={"hyperparameters": "default"},
-    low_quality={"hyperparameters": "toy"},
+    low_quality={"hyperparameters": "local_models"},
 )
--- a/timeseries/src/autogluon/timeseries/configs/presets_configs.py
+++ b/timeseries/src/autogluon/timeseries/configs/presets_configs.py
@@ -12,22 +12,8 @@ TIMESERIES_PRESETS_CONFIGS = dict(
             "num_trials": 20,
         },
     },
-    high_quality={
-        "hyperparameters": "default_hpo",
-        "hyperparameter_tune_kwargs": {
-            "scheduler": "local",
-            "searcher": "random",
-            "num_trials": 10,
-        },
-    },
-    good_quality={
-        "hyperparameters": "default_hpo",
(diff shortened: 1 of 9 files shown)


timeseries/src/autogluon/timeseries/models/presets.py
```diff
@@ -62,66 +62,72 @@ def get_default_hps(key, prediction_length):
...
-            "Transformer": {
+            "DeepAR": {
                 "context_length": context_length,
             },
-            "DeepAR": {
+            "TemporalFusionTransformer": {
                 "context_length": context_length,
             },
         },
         "default_hpo": {
```
huibinshen, line 102: The search space should be somewhat stable, just want to double click on it so we don't change it soon.

timeseries/src/autogluon/timeseries/configs/presets_configs.py
```diff
@@ -12,22 +12,8 @@
...
-    },
-    good_quality={
-        "hyperparameters": "default_hpo",
-        "hyperparameter_tune_kwargs": {
-            "scheduler": "local",
-            "searcher": "random",
-            "num_trials": 2,
-        },
-    },
+    high_quality={"hyperparameters": "default"},
```
canerturkmen, line 15: Should we just drop these as the Predictor already seems to be handling the fallback?

canerturkmen (approved):
LGTM! Minor comments only.

timeseries/src/autogluon/timeseries/models/presets.py
```diff
@@ -62,66 +62,72 @@ def get_default_hps(key, prediction_length):
...
                 "context_length": 5,
             },
-            "Transformer": {"epochs": 10, "num_batches_per_epoch": 10, "context_length": 5},
-            "DeepAR": {"epochs": 10, "num_batches_per_epoch": 10, "context_length": 5},
-            "ETS": {"maxiter": 20, "seasonal": None},
-            "ARIMA": {
-                "maxiter": 10,
-                "order": (1, 0, 0),
-                "seasonal_order": (0, 0, 0),
+            "Transformer": {
```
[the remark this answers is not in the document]
shchur: Actually not sure about these settings. It feels like the only purpose to have these deep learning models in `low_quality` is for the quickstart tutorial, but in reality if users want "decent predictions very quickly", we should only use local models here (maybe + some more baselines like SeasonalNaive in the future). I don't think any DL models can achieve comparable results in `"low_quality"` given these toy configurations. What do you think?

timeseries/src/autogluon/timeseries/configs/presets_configs.py
shchur: We notify the user inside the predictor, but I think the `@apply_presets` decorator will fail if we remove the presets completely
https://github.com/awslabs/autogluon/blob/bd1615584863e01b8123fb5d515e16cc5aca5639/timeseries/src/autogluon/timeseries/predictor.py#L183

timeseries/src/autogluon/timeseries/models/presets.py
```diff
@@ -151,18 +156,23 @@ def get_preset_models(
...
-        if hyperparameters is not None:
-            # filter only default_hps for models with hyperparameters provided
-            default_hps = {model: default_hps.get(model, {}) for model in hyperparameters}
-            for model in hyperparameters:
-                default_hps[model].update(hyperparameters[model])
-        hyperparameters = copy.deepcopy(default_hps)
+        raise ValueError(
+            f"hyperparameters must be a dict, a string or None (received {type(hyperparameters)}). "
+            "Please see the documentation for TimeSeriesPredictor.fit"
+        )
```
shchur, line 175: Updated logic for hyperparameter handling (that fixes a bug)
- Now: If `hyperparameters` is a dict, we take the default presets from `default`, and override them based on `hyperparameters`.
- Before: `hyperparameters` is a dict and `hyperparameter_tune == True`, we take the default presets from `default_hpo`, and override them based on `hyperparameters`. This behavior resulted in the HPO search space potentially including parameters not provided by the user.

timeseries/src/autogluon/timeseries/models/presets.py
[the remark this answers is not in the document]
shchur: Sounds good, I will update the PR with these changes and ping you again once it's ready to be reviewed

canerturkmen (approved):
LGTM! Thanks!!

Issue #2036, opened 2022-08-14
Normalize unicode data preprocess feature for tabular data


MountPOTATO:
Hi, I’m using autogluon’s AutoMMPredictor to take part in a Kaggle contest https://www.kaggle.com/competitions/feedback-prize-effectiveness , a multi-class classification task for text tabular data (to classify argumentative elements in student writing as “effective”, “adequate”, or “ineffective” to be specific). 

During the contest I found some trick for text data preprocess and wonder if a similar feature like this could be added to autogluon.

In some high-ranked solution’s code like https://www.kaggle.com/code/quclongphan/deberta-v3-training, I found some normalize Unicode method applied in the data preprocessing of the tabular text data in order to resolve the encoding problems and normalize the abnormal characters:

```python
from text_unidecode import unidecode 
from typing import Dict, List, Tuple
import codecs

# codecs error handlers
def replace_encoding_with_utf8(error: UnicodeError) -> Tuple[bytes, int]:
    return error.object[error.start : error.end].encode("utf-8"), error.end

def replace_decoding_with_cp1252(error: UnicodeError) -> Tuple[str, int]:
    return error.object[error.start : error.end].decode("cp1252"), error.end

# Register the encoding and decoding error handlers for `utf-8` and `cp1252`.
codecs.register_error("replace_encoding_with_utf8", replace_encoding_with_utf8)
codecs.register_error("replace_decoding_with_cp1252", replace_decoding_with_cp1252)

# core function, resolve the encoding problems and normalize the abnormal characters.
def resolve_encodings_and_normalize(text: str) -> str:
    text = (
        text.encode("raw_unicode_escape")
        .decode("utf-8", errors="replace_decoding_with_cp1252")
        .encode("cp1252", errors="replace_encoding_with_utf8")
        .decode("utf-8", errors="replace_decoding_with_cp1252")
    )
    text = unidecode(text)
    return text
```

This code is applied on the text content in the tabular data frame. In this code, the text goes through a encode-decode process with ‘utf-8’ and ‘cp1252’(using the python encode() and decode() function). Meanwhile, it uses codecs package to register the encode() and decode()’s error handlers). Then the unidecode() is applied to do an ASCII transliterations to Unicode text, as well as fixing the string with abnormal characters (the input string may contains non-ASCII characters, and unidecode’s output string are safe to be encoded to ASCII). There seems to be improvement after applying this to the task.

[...]

PR #2168, opened 2022-09-26
Add normalize_text option

suzhoum:
*Issue #2036 

*Description of changes:*
This PR adds a text normalization option to multimodal `TextProcessor` to incorporate the suggestion proposed in the issue

### DIFF
--- a/docs/tutorials/multimodal/customization.md
+++ b/docs/tutorials/multimodal/customization.md
@@ -501,6 +501,16 @@ predictor.fit(hyperparameters={"data.image.missing_value_strategy": "skip"})
 predictor.fit(hyperparameters={"data.image.missing_value_strategy": "zero"})
 ```
 
+### data.text.normalize_text
+Whether to normalize text with encoding problems. If True, TextProcessor will run through a series of encoding and decoding for text normalization. Please refer to the [Example](https://github.com/awslabs/autogluon/tree/master/examples/automm/kaggle_feedback_prize) of Kaggle competition for applying text normalization.
+
+```
+# default used by AutoMM
+predictor.fit(hyperparameters={"data.text.normalize_text": False})
+# turn on text normalization
+predictor.fit(hyperparameters={"data.text.normalize_text": True})
+```
+
 ### data.categorical.convert_to_text
 
 Whether to treat categorical data as text. If True, no categorical models, e.g., `"categorical_mlp"` and `"categorical_transformer"`, would be used.
--- a/multimodal/setup.py
+++ b/multimodal/setup.py
@@ -36,6 +36,7 @@ install_requires = [
     "scikit-image>=0.19.1,<0.20.0",
     "smart_open>=5.2.1,<5.3.0",
     "pytorch_lightning>=1.6.0,<1.7.0",
+    "text-unidecode<=1.3",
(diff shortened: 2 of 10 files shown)


multimodal/setup.py
```diff
@@ -28,6 +28,7 @@
     "tqdm",
     "boto3",
     "requests",
+    "text-unidecode",
```
sxjscience, line 31: Check the current latest version of text-unidecode in pypi and add a version cap of this package.

multimodal/src/autogluon/multimodal/data/utils.py
```diff
@@ -212,3 +214,29 @@ def apply_data_processor(modality_features: dict, data_processors: dict, idx: in
...
+
+    def replace_decoding_with_cp1252(error: UnicodeError) -> Tuple[str, int]:
+        return error.object[error.start : error.end].decode("cp1252"), error.end
+
+    codecs.register_error("replace_encoding_with_utf8", replace_encoding_with_utf8)
+    codecs.register_error("replace_decoding_with_cp1252", replace_decoding_with_cp1252)
+
+
+def normalize_text(text: str) -> str:
+    """Resolve the encoding problems and normalize the abnormal characters."""
```
bryanyzhu, line 233: Maybe add a short example in the docstring here to show what text normalization does and how can it help downstream usage. But if this is a common term in NLP which requires no further explanation, it is also fine.

suzhoum: Good idea! Do you think it would be fine if I add a [link](https://github.com/awslabs/autogluon/tree/master/examples/automm/kaggle_feedback_prize#15-a-few-examples-of-normalized-texts) to the example I added before

bryanyzhu: Sure, that would be good

docs/tutorials/multimodal/customization.md
```diff
@@ -501,6 +501,17 @@ predictor.fit(hyperparameters={"data.image.missing_value_strategy": "skip"})
 predictor.fit(hyperparameters={"data.image.missing_value_strategy": "zero"})
 ```
 
+### data.text.normalize_text
+Whether to normalize text with encoding problems. If True, text will go through a series of encoding and
+decoding for normalization, which can slow down data preprocessing.
```
sxjscience, line 506: You may point to the example so that the users can know more details.

sxjscience (approved):
LGTM.

bryanyzhu (approved):
Great, thanks

multimodal/setup.py
```diff
@@ -36,6 +36,7 @@
     "scikit-image>=0.19.1,<0.20.0",
     "smart_open>=5.2.1,<5.3.0",
     "pytorch_lightning>=1.6.0,<1.7.0",
+    "text-unidecode<=1.3",
```
zhiqiangdon, line 39: Do we need to add a minimum version constraint?

zhiqiangdon (approved):
LGTM! Thanks! We may try this on our text+tabular benchmarks.

FANGAreNotGnu (approved):
LGTM, thanks!

PR #2175, opened 2022-09-28
[Cloud] Cloud Batch Inference Optimization

yinweisu:
*Issue #, if available:*

*Description of changes:*
This PR enable CloudPredictor to skip repacking step when doing batch inference and deployment. This could save a lot of time when the model artifact is large and the network speed is slow
* Adding inference script into model artifacts to avoid repack it with model afterwards.
* Instead of using `MxnetModel` implementation, now we inherit from `Model`. Also updated estimator to inherit from base `Estimator` instead of `Framework`

### DIFF
--- a/cloud/src/autogluon/cloud/scripts/train.py
+++ b/cloud/src/autogluon/cloud/scripts/train.py
@@ -60,6 +60,7 @@ if __name__ == "__main__":
         "--images_dir", type=str, required=False, default=get_env_if_present("SM_CHANNEL_IMAGES")
     )
     parser.add_argument("--ag_config", type=str, default=get_env_if_present("SM_CHANNEL_CONFIG"))
+    parser.add_argument("--serving_script", type=str, default=get_env_if_present("SM_CHANNEL_SERVING"))
 
     args, _ = parser.parse_known_args()
 
@@ -132,3 +133,9 @@ if __name__ == "__main__":
         if config.get("leaderboard", False):
             lb = predictor.leaderboard(silent=False)
             lb.to_csv(f"{args.output_data_dir}/leaderboard.csv")
+
+    print('Saving serving script')
+    serving_script_saving_path = os.path.join(save_path, 'code')
+    os.mkdir(serving_script_saving_path)
+    serving_script_path = get_input_path(args.serving_script)
+    shutil.move(serving_script_path, os.path.join(serving_script_saving_path, os.path.basename(serving_script_path)))
--- a/cloud/src/autogluon/cloud/utils/ag_sagemaker.py
+++ b/cloud/src/autogluon/cloud/utils/ag_sagemaker.py
@@ -12,7 +12,7 @@ from .deserializers import PandasDeserializer
 from .sagemaker_utils import retrieve_latest_framework_version
 
 
-# Framework documentation: https://sagemaker.readthedocs.io/en/stable/api/training/estimators.html#sagemaker.estimator.Framework
+# Estimator documentation: https://sagemaker.readthedocs.io/en/stable/api/training/estimators.html#estimators
 class AutoGluonSagemakerEstimator(Estimator):
     def __init__(
         self,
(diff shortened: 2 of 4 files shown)


cloud/src/autogluon/cloud/scripts/train.py
```diff
@@ -132,3 +133,9 @@ def get_env_if_present(name):
         if config.get("leaderboard", False):
             lb = predictor.leaderboard(silent=False)
             lb.to_csv(f"{args.output_data_dir}/leaderboard.csv")
+
+    print('Saving serving script')
+    serving_script_saving_path = os.path.join(save_path, 'code')
+    os.mkdir(serving_script_saving_path)
+    serving_script_path = get_input_path(args.serving_script)
+    shutil.move(serving_script_path, os.path.join(serving_script_saving_path, os.path.basename(serving_script_path)))
```
gradientsky, line 141: do we want to move or copy?

cloud/src/autogluon/cloud/utils/ag_sagemaker.py
```diff
@@ -163,3 +157,95 @@ def transformer(
...
+    def prepare_container_def(
+        self,
+        instance_type=None,
+        accelerator_type=None,
+        serverless_inference_config=None,
+    ):  # pylint: disable=unused-argument
+        deploy_env = copy.deepcopy(self.env)
+        deploy_env.update(self._script_mode_env_vars())
+        deploy_env[SCRIPT_PARAM_NAME.upper()] = os.path.basename(deploy_env[SCRIPT_PARAM_NAME.upper()])
+        deploy_env[DIR_PARAM_NAME.upper()] = '/opt/ml/model/code'
```
gradientsky, line 202: nit: this can be extracted into a separate method, then the rest of the code can be moved into abstract class and reuse - this would make re-pack vs non-repack difference easier to read:
```python
class AbstractAutoGluonInferenceModel(AutoGluonSagemakerInferenceModel, ABC):

    def prepare_container_def(
            self,
            instance_type=None,
            accelerator_type=None,
            serverless_inference_config=None,
    ):  # pylint: disable=unused-argument
        deploy_env = copy.deepcopy(self.env)
        deploy_env.update(self._script_mode_env_vars())
        self._prepare_container_def(deploy_env)

        return sagemaker.container_def(
            self.image_uri,
            self.model_data,
            deploy_env,
            image_config=self.image_config,
        )

    @abstractmethod
    def _prepare_container_def(self, deploy_env):
        raise NotImplementedError


class AutoGluonNonRepackInferenceModel(AbstractAutoGluonInferenceModel):

    def _prepare_container_def(self, deploy_env):
        deploy_env[SCRIPT_PARAM_NAME.upper()] = os.path.basename(deploy_env[SCRIPT_PARAM_NAME.upper()])
        deploy_env[DIR_PARAM_NAME.upper()] = '/opt/ml/model/code'


class AutoGluonRepackInferenceModel(AbstractAutoGluonInferenceModel):

    def _prepare_container_def(self, deploy_env):
        deploy_key_prefix = fw_utils.model_code_key_prefix(
            self.key_prefix, self.name, self.image_uri
        )
        self._upload_code(deploy_key_prefix, repack=True)

```

cloud/src/autogluon/cloud/scripts/train.py
yinweisu: Move is fine. We don't use this script while training. Just save it as part of the model artifacts so that we don't need to repack the inference code later.

cloud/src/autogluon/cloud/utils/ag_sagemaker.py
yinweisu: Actually your provided code snippet doesn't work. The thing is `self._upload_code` will affect the result of `self._script_mode_env_vars()` hence affecting the `deploy_env` being updated. If we move the `deploy_env.update(self._script_mode_env_vars())` into `_prepare_container_def `, there's very little thing the two method shares so a `ABC` is likely not needed.

gradientsky: This is a side effect, which should be avoided. Can we make the side effect updating vars explicit in this method?

gradientsky (approved):

PR #2179, opened 2022-09-30
Add AutoMM matcher

zhiqiangdon:
1. Add the MultiModalMatcher class.
2. Support the siamese network, i.e., query and response share the same encoder.
3. Refactor the the matcher config and training loop logic.
4. Support inputting `corpus`, from which we can index the query/response values.
5. Can reproduce the ID change detection result with MultiModalMatcher.
6. Add unit tests.

```
matcher = MultiModalMatcher(
        query=query,  # query columns
        response=response,  # response columns
        label=label,  # label column (optional)
        match_label=match_label, # for binary labels, which label represent being match
)
matcher.fit(
      train_data=train_data,
      corpus=corpus, # query and response corpus (optional)
)
```


multimodal/src/autogluon/multimodal/configs/matcher/default.yaml
```diff
@@ -1,21 +1,15 @@
 matcher:
-  match_label: 0
+  loss:
+    type: "contrastive_loss"
+    pos_margin: 0.8
+    neg_margin: 0.2
+  distance:
+    type: "cosine_similarity"
+  miner:
```
sxjscience, line 8: We can later add some documentations about “miner”. May be via a paper/tech blog.

multimodal/src/autogluon/multimodal/data/datamodule.py
```diff
@@ -71,13 +74,15 @@ def __init__(
         self.val_data = val_data
         self.test_data = test_data
         self.predict_data = predict_data
+        self.corpus = corpus
```
sxjscience, line 77: After reading the source code, I’m not sure why do we need it.

sxjscience: At least, corpus is not a good name.

multimodal/src/autogluon/multimodal/configs/matcher/default.yaml
zhiqiangdon: Agree. We will introduce the usage in the `Customize AutoMM` tutorial. We can refer to here: https://kevinmusgrave.github.io/pytorch-metric-learning/miners/#miners

multimodal/src/autogluon/multimodal/data/datamodule.py
zhiqiangdon: For the text-to-text matching, one document may appear in several samples. The corpus is to store the id-to-doc mapping and the train_df is only needs to contain the doc indexes. This can reduce the memory usages since we don't need to replicate one long document several times in different samples. Storing only the indexes in the train_df is more memory-efficient. Actually, the ir_dataset uses this kind of format, e.g., https://ir-datasets.com/antique.html

multimodal/src/autogluon/multimodal/matcher.py
```diff
@@ -0,0 +1,1715 @@
...
+    tensor_to_ndarray,
+    try_to_infer_pos_label,
+)
+
+logger = logging.getLogger(AUTOMM)
+
+
+class MultiModalMatcher:
+    """
+    MultiModalMatcher is a deep learning "model zoo" of model zoos. It can automatically build deep learning models that
```
bryanyzhu, line 105: It seems here is a duplicate paragraph from `MultimodalPredictor`. We should change the description to better fit what does matcher do.

multimodal/src/autogluon/multimodal/matcher.py
```diff
@@ -0,0 +1,1715 @@
...
+        eval_metric: Optional[str] = None,
+        path: Optional[str] = None,
+        verbosity: Optional[int] = 3,
+        warn_if_exist: Optional[bool] = True,
+        enable_progress_bar: Optional[bool] = None,
+    ):
+        """
+        Parameters
+        ----------
+        label
```
cheungdaven, line 131: parameters description is not updated.

multimodal/src/autogluon/multimodal/matcher.py
```diff
@@ -0,0 +1,1715 @@
+from __future__ import annotations
```
cheungdaven, line 1: if everything about matcher is moved here, should we delete the matcher related code in the predictor.py ?

multimodal/src/autogluon/multimodal/utils/matcher.py
```diff
@@ -0,0 +1,280 @@
...
+    Returns
+    -------
+    Whether to share the same fusion part.
+    """
+    query_model_names = [n for n in query_model_names if not n.lower().startswith(FUSION)]
+    response_model_names = [n for n in response_model_names if not n.lower().startswith(FUSION)]
+    return sorted(query_model_names) == sorted(response_model_names)
+
+
+def create_siamese_model(
```
cheungdaven, line 222: it can be kind of difficult to distinguish at the first glance create_siamese_model and build_siamese_network

multimodal/src/autogluon/multimodal/data/preprocess_dataframe.py
```diff
@@ -359,16 +377,27 @@ def transform_image(
...
-        for col_name in self._image_path_names:
-            processed_data = df[col_name].apply(lambda ele: ele.split(";")).tolist()
-            image_paths[col_name] = processed_data
-        return image_paths
+        image_features = {}
+        image_types = {}
+        for col_name in self._image_feature_names:
+            col_value = df[col_name]
+            col_type = self._column_types[col_name]
+            if col_type == IMAGE_PATH or IMAGE:
```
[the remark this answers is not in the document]
zhiqiangdon: In the `elif` below (line 387), we need to check if `col_type == image_index`. Changing to `startswith(IMAGE)` would make the `elif` ignored.

multimodal/src/autogluon/multimodal/data/infer_types.py
```diff
@@ -12,7 +12,8 @@
     BINARY,
     CATEGORICAL,
     CLASSIFICATION,
-    IMAGE,
+    IMAGE_PATH,
```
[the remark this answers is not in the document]
zhiqiangdon: `IMAGE` is used in the `data_types` in the model configs to indicate one model requires image data. `IMAGE_PATH` is to be specific that the dataframe contains image paths. There may be other kinds of formats for image data, e.g., images encoded in base64 or image indexes. By default, we assume image paths if not specified otherwise.

multimodal/src/autogluon/multimodal/configs/matcher/default.yaml
```diff
@@ -1,21 +1,15 @@
...
+    pos_margin: 0.6
+    neg_margin: 0.4
   matches:
-      - loss:
-          type: "contrastive_loss"
-          pos_margin: 0.8
-          neg_margin: 0.2
-          weight: 1
-        distance:
-          type: "cosine_similarity"
```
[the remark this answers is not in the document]
zhiqiangdon: Right. We will support using more distances. `cosine` is just the default.

multimodal/src/autogluon/multimodal/data/infer_types.py
```diff
@@ -239,13 +240,86 @@ def is_text_column(X: pd.Series) -> bool:
     return True
 
 
+def is_index_column(data: pd.Series, col_name: str, corpus: Dict[str, Dict]) -> bool:
```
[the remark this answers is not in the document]
zhiqiangdon: You are right. How about `idx`? Or do you have other names in mind?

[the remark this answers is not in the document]
cheungdaven: identifier?

multimodal/src/autogluon/multimodal/data/datamodule.py
zhiqiangdon: `matcher_collection ` seems not to give the hint of data. How about `metadata`, `database`, or `corpus_metadata`?

multimodal/src/autogluon/multimodal/data/infer_types.py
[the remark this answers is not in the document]
zhiqiangdon: Sounds good.

multimodal/src/autogluon/multimodal/data/datamodule.py
zhiqiangdon: Changed the name to `id_mappings`, which is more straightforward to understand.

multimodal/src/autogluon/multimodal/matcher.py
zhiqiangdon: Thanks. Updated the docstrings.

multimodal/src/autogluon/multimodal/data/infer_types.py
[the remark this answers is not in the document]
zhiqiangdon: Updated to use `identifier `.

multimodal/src/autogluon/multimodal/matcher.py
zhiqiangdon: Thanks. Docstrings updated.

multimodal/src/autogluon/multimodal/matcher.py
zhiqiangdon: Good point. Removed the matcher related in predictor.py.

multimodal/src/autogluon/multimodal/utils/matcher.py
zhiqiangdon: Changed `build_siamese_network ` to `make_siamese` be distinguishable.

multimodal/src/autogluon/multimodal/data/infer_types.py
```diff
@@ -239,13 +240,87 @@ def is_text_column(X: pd.Series) -> bool:
...
+    for index in data:
+        try:
+            per_value = id_mappings[col_name][index]
+        except:
+            failure_count += 1
+
+    if failure_count == 0:
+        return True
+    elif 0 < failure_count < sample_num:
+        raise warnings.warn(
```
FANGAreNotGnu, line 276: seems that raise warnings will cause "exceptions must derive from BaseException": https://github.com/sonus21/error-tracker/issues/18

multimodal/src/autogluon/multimodal/data/infer_types.py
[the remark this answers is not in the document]
FANGAreNotGnu: Currently do we support other type of image input expect for image_path?

multimodal/src/autogluon/multimodal/data/infer_types.py
```diff
@@ -371,6 +453,9 @@ def infer_label_column_type_by_problem_type(
     -------
     Column types with the label columns' types inferred from the problem type.
     """
+    if label_columns is None:
```
FANGAreNotGnu, line 456: Shall we check here if this is for matcher then return column_types otherwise raise an error?

multimodal/src/autogluon/multimodal/data/preprocess_dataframe.py
```diff
@@ -120,7 +122,14 @@ def column_types(self):
...
     def image_path_names(self):
-        return self._image_path_names
+        if hasattr(self, "_image_path_names"):
+            return self._image_path_names
+        else:
+            return [col_name for col_name in self._image_feature_names if self._column_types[col_name] == IMAGE_PATH]
+
+    @property
+    def image_feature_names(self):
+        return self._image_feature_names if hasattr(self, "_image_feature_names") else self._image_path_names
```
FANGAreNotGnu, line 132: There is self._image_feature_names = [] in __init__. Seems it's not necessary to add the else statement.

multimodal/src/autogluon/multimodal/data/preprocess_dataframe.py
```diff
@@ -303,7 +317,7 @@ def fit(self, X: Optional[pd.DataFrame] = None, y: Optional[pd.Series] = None):
     def transform_text(
         self,
         df: pd.DataFrame,
-    ) -> Dict[str, List[str]]:
+    ) -> Tuple[Dict[str, List[str]], Dict[str, str]]:
```
FANGAreNotGnu, line 320: We may want to update the description of the returns accordingly (in all transform_xxx)

multimodal/src/autogluon/multimodal/data/utils.py
```diff
@@ -188,34 +192,36 @@ def apply_df_preprocessor(data: pd.DataFrame, df_preprocessor: MultiModalFeature
...
-def apply_data_processor(modality_features: dict, data_processors: dict, idx: int, is_training: bool):
+def apply_data_processor(per_sample_features: dict, data_processors: dict, is_training: bool):
     """
     Process one sample's features.
 
     Parameters
     ----------
-    modality_features
-        Features of different modalities got from `apply_df_preprocessor`.
+    per_sample_features
```
FANGAreNotGnu, line 221: This looks better. Thanks for the refactor!

multimodal/src/autogluon/multimodal/data/infer_types.py
zhiqiangdon: Good catch! Thanks.

multimodal/src/autogluon/multimodal/data/infer_types.py
zhiqiangdon: If `label_columns` is None in supervised training, it would throw errors in other places, e.g., computing the loss. If we add a condition to check if a matching problem, then we would need check more problem types in future to determine whether to throw errors, which is kind of a burden. I think no condition here is more general.

multimodal/src/autogluon/multimodal/data/preprocess_dataframe.py
zhiqiangdon: Thanks for pointing it out. Updated.

multimodal/src/autogluon/multimodal/data/preprocess_dataframe.py
zhiqiangdon: Good suggestion. Updated the return docstrings.

multimodal/src/autogluon/multimodal/data/dataset.py
```diff
@@ -35,6 +36,9 @@ def __init__(
             A list of multimodal feature preprocessors generating model-agnostic features.
         processors
             Data processors customizing data for each modality per model.
+        id_mappings
+             Id-to-content mappings. The contents can be text, image, etc.
+             This is used when the dataframe contains the query/response indexes instead of their contents.
```
sxjscience, line 41: I think str—> dict is fine for now. But we may consider to change it to str —> dataframe. Because we can just store all data in Dataframe-like objects.

sxjscience (approved):
LGTM!

multimodal/src/autogluon/multimodal/data/dataset.py
sxjscience: Also, we may need to revise the name, like calling it `linked_data`.

The idea is that the columns in the train_data / val_data are actually linked to another dataframe.

sxjscience: Or "linked_train_data".

bryanyzhu (approved):
Looks great, I'm ok to merge.

PR #2180, opened 2022-09-30
Add ThetaModel timeseries models

shchur:
*Description of changes:*
- Add a wrapper for [ThetaModel](https://www.statsmodels.org/stable/generated/statsmodels.tsa.forecasting.theta.ThetaModel.html) from statsmodels to AGTS


timeseries/src/autogluon/timeseries/models/statsmodels/models.py
```diff
@@ -220,3 +221,147 @@ def _predict_with_local_model(
...
+        target = timeseries.values
+        forecast = {}
+        if len(target) >= seasonal_period:
+            indices = [len(target) - seasonal_period + k % seasonal_period for k in range(self.prediction_length)]
+            forecast["mean"] = target[indices].reshape(-1)
+        else:
+            forecast["mean"] = np.full(shape=[self.prediction_length], fill_value=target.mean())
+
+        for q in quantile_levels:
+            forecast[str(q)] = forecast["mean"]
```
canerturkmen, line 365: I see the intention here but I think this is semantically incorrect and could fail silently. Shouldn't we instead set to `NaN`?

timeseries/src/autogluon/timeseries/models/statsmodels/models.py
```diff
@@ -220,3 +221,147 @@ def _predict_with_local_model(
...
+            else:
+                coverage = 2 * (1 - q)
+                column_index = 1
+            quantile_pred = sm_model.prediction_intervals(steps=self.prediction_length, alpha=coverage)
+            # Select lower bound of the confidence interval if q < 0.5, upper bound otherwise
+            results.append(quantile_pred.iloc[:, column_index].rename(str(q)))
+        return pd.concat(results, axis=1)
+
+
+class SeasonalNaiveModel(AbstractStatsmodelsModel):
```
canerturkmen, line 316: I have a concern about putting this under the statsmodels namespace just to be able to use the `AbstractStatsmodelsModel` type. I think we can redesign now if necessary to have `AbstractLocalModel` `AbstractStatsmodelsModel < AbstractLocalModel` and not mix this one in the Statsmodels namespace. This is both semantically confusing (i.e., what if the test for this model fails due to a dependency quirk in the statsmodels import statements?) and also makes dependency management quite difficult (e.g., what if we want to lazily import statsmodels, will we also not import SeasonalNaive?).

canerturkmen (commented):
All looks good to me except how we seem to have introduced SeasonalNaive in the sm namespace although it doesn't necessarily have a dependency on sm as far as I can understand. I'm OK to merge this without SeasonalNaive and move the redesign to a different PR though.

timeseries/src/autogluon/timeseries/models/presets.py
```diff
@@ -52,6 +54,8 @@
     SktimeAutoETS=60,
     ARIMA=50,
     ETS=60,
+    SeasonalNaive=65,
+    Theta=65,
```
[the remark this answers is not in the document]
shchur: Theta is faster than ETS, but I agree that we can keep both at the same priority level.

timeseries/src/autogluon/timeseries/models/presets.py
```diff
@@ -123,6 +131,13 @@ def get_default_hps(key, prediction_length):
                 "order": ag.Categorical((2, 0, 1), (2, 1, 1), (1, 1, 1)),
                 "seasonal_order": ag.Categorical((0, 0, 0), (1, 0, 1)),
             },
+            "SeasonalNaive": {
+                "seasonal_period": ag.Categorical(None, 1),
```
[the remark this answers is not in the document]
shchur: - None - infer `seasonal_period` from `freq`
- 1 - `seasonal_period = 1` (= use last observed value as prediction for the entire forecast horizon)

Agree that a comment makes sense here (or maybe even another class name for the case where `seasonal_period = 1` like `Naive`?)

timeseries/src/autogluon/timeseries/models/statsmodels/models.py
shchur: Makes sense, I also had some other ideas about optimizing local models that we can discuss in the other PR (though, probably not highest priority right now)

timeseries/src/autogluon/timeseries/models/statsmodels/models.py
shchur: Using NaNs here can later cause problems in `TimeSeriesEvaluator` / `WeightedEnsemble`. If we decide to use NaNs, we will need to introduce more safeguards in those classes / logic that handles columns with NaNs as missing.

We should in general discuss how we should handle models that cannot produce probabilistic forecasts. Should we include them at all? Should we let the users decide if they are fine receiving point forecasts only? This is also very relevant for the Tabular forecasting model, we can discuss it in that PR

timeseries/src/autogluon/timeseries/models/presets.py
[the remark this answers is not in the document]
canerturkmen: ah my bad. thanks!

timeseries/src/autogluon/timeseries/models/statsmodels/models.py
canerturkmen: does it cause problems in the sense that it raises an error? or do the NaNs propagate in a meaningful way?

canerturkmen (approved):
LGTM once tests pass. Thanks!!!

PR #2181, opened 2022-10-01
Fix realtime inference for mmcv models

zhiqiangdon:
1. Move the `per_gpu_batch_size ` from image processor initialization to `get_collate_fn()`. In this way, batch size no longer needs to bound with image processor, which allows dynamic batch size in inference. 
2. Add `MMCV_MODELS` constant. We could put all the mmcv models here.
3. Improve the precision context manager.
4. Consider batch size in determining whether to use the realtime inference.

### DIFF
--- a/multimodal/src/autogluon/multimodal/utils/environment.py
+++ b/multimodal/src/autogluon/multimodal/utils/environment.py
@@ -1,12 +1,13 @@
+import contextlib
 import logging
 import math
 import sys
 import warnings
 from typing import Dict, List, Optional, Tuple, Union
-import contextlib
 
 import torch
 from torch import nn
+
 try:
     from mmcv.parallel import DataContainer
 except:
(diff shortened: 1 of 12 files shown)


multimodal/tests/unittests/test_object_detection.py
```diff
@@ -35,7 +35,7 @@ def test_mmdet_object_detection_inference(checkpoint_name):
         pipeline="object_detection",
     )
 
-    pred = predictor.predict({"image": [mmdet_image_name] * 10}, realtime=False)  # test batch inference
+    pred = predictor.predict({"image": [mmdet_image_name] * 10})  # test batch inference
```
bryanyzhu, line 38: Is the default for realtime still False or True?

multimodal/src/autogluon/multimodal/utils/environment.py
```diff
@@ -7,6 +8,11 @@
 import torch
 from torch import nn
 
+try:
+    from mmcv.parallel import DataContainer
+except:
+    pass
```
bryanyzhu, line 14: Should we do `pass` or set mmcv to `None`?

multimodal/tests/unittests/test_object_detection.py
zhiqiangdon: The default is None. We infer whether True or Flase.

multimodal/src/autogluon/multimodal/utils/environment.py
zhiqiangdon: We don't use `mmcv` in this file. I think it doesn't matter setting None or not.

bryanyzhu (approved):
LGTM, thanks for the quick fix.

sxjscience (approved):
LGTM

PR #2183, opened 2022-10-03
Add AutoMM named entity recognition

cheungdaven:
*Issue #, if available:*

*Description of changes:*

1. Add support of the named entity recognition task in AutoMM.
2. Add unit test.

```
label_col = "entity_annotations" # Specify which column is the NER annotation.
predictor = MultiModalPredictor(problem_type="ner", label=label_col)
predictor.fit(
    train_data=train_data
)
```
We will use bert-base-cased by default, users can specify the checkpoint using the hyperparameters argument "model.ner.checkpoint_name".


multimodal/setup.py
```diff
@@ -28,6 +28,8 @@
     "tqdm",
     "boto3",
     "requests",
+    "evaluate",
```
sxjscience, line 31: We may use "evaluate<=0.2.2"

cheungdaven: Done in this [commit](https://github.com/awslabs/autogluon/pull/2183/commits/ec2fc468bf390caee74616c10600a347cbcde3d4).

cheungdaven: It seems that the error is because of the evaluate lib [this line](https://github.com/huggingface/evaluate/blob/81f841b3adebd2a78b3831388e74555acce61fa2/src/evaluate/evaluator/base.py#L20).

multimodal/src/autogluon/multimodal/data/labelencoder_ner.py
```diff
@@ -0,0 +1,175 @@
...
+        logger.debug(f"Unique entity groups in the data: {entity_groups}")
+
+    def extract_ner_annotations(self, y: pd.Series):
+        """
+        Validate the JSON annotations with predefined JSON schema and convert it into list.
+
+        Parameters
+        ----------
+        y
+            The raw json annotations.
```
sxjscience, line 46: nit: may need an empty line.

multimodal/src/autogluon/multimodal/constants.py
```diff
@@ -52,6 +52,7 @@
 MIN = "min"
 ACCURACY = "accuracy"
 ACC = "acc"
+NER_METRIC = "overall_accuracy"
```
zhiqiangdon, line 55: Is it possible that ner uses other metrics besides "overall_accuracy"? Is it better to define 
`OVERALL_ACCURACY = "overall_accuracy"`?

multimodal/src/autogluon/multimodal/data/labelencoder_ner.py
cheungdaven: done in the least commit.

multimodal/src/autogluon/multimodal/configs/model/fusion_mlp_image_text_tabular.yaml
```diff
@@ -69,6 +69,20 @@ model:
...
+    checkpoint_name: "bert-base-cased"
+    max_text_len: 512
+    gradient_checkpointing: False
+    data_types:
+      - "text"
+    tokenizer_name: "hf_auto"
+    insert_sep: False
+    text_segment_num: 2
+    stochastic_chunk: False
+    num_classes: 1
```
zhiqiangdon, line 82: Is this `num_classes ` necessary? I don't see it being used.

multimodal/src/autogluon/multimodal/constants.py
cheungdaven: yes, there are overall_f1, overall_precision, overall_recall, and for each entity type, we will also output the specific evaluation results. Users can directly use the entity group name to access the results. If no specific evaluation metric is given, the default output would be like:
```
 {'LOC': {'precision': 0.9313022700119474, 'recall': 0.9307462686567164, 'f1': 0.9310241863242759, 'number': 1675}, 
     'MISC': {'precision': 0.7767379679144385, 'recall': 0.8114525139664804, 'f1': 0.7937158469945355, 'number': 716}, 
     'ORG': {'precision': 0.8750719631548647, 'recall': 0.9074626865671642, 'f1': 0.8909730363423213, 'number': 1675}, 
     'PER': {'precision': 0.9567047504509922, 'recall': 0.958433734939759, 'f1': 0.957568462232922, 'number': 1660}, 
     'overall_precision': 0.9019237375472346, 'overall_recall': 0.9170450576318547, 'overall_f1': 0.9094215448562522, 'overall_accuracy': 0.9842387467323067}
```

multimodal/src/autogluon/multimodal/configs/model/fusion_mlp_image_text_tabular.yaml
cheungdaven: yeah, it is not necessary. I will remove it.

multimodal/src/autogluon/multimodal/data/labelencoder_ner.py
```diff
@@ -0,0 +1,175 @@
...
+            try:
+                jsonschema.validate(json_ner_annotations, schema)  # verify the json schema
+            except jsonschema.ValidationError as e:
+                # Raise an error if the provided json annotations do not match the predefined schema
+                raise ValueError(f"The provided json annotations are invalid: {e.message}")
+            sentence_annotations = []
+            for annot in json_ner_annotations:
+                all_entity_groups.append(annot[ENTITY_GROUP])
+                if self.entity_map is not None:
+                    if annot[ENTITY_GROUP] in self.entity_map:
```
zhiqiangdon, line 83: `if annot[ENTITY_GROUP] not in self.entity_map`, then no action is taken. Is this expected?

multimodal/src/autogluon/multimodal/data/labelencoder_ner.py
```diff
@@ -0,0 +1,175 @@
...
+    def __init__(self, entity_map: Optional[dict] = None):
+        self.entity_map = entity_map
+
+    def fit(self, y: pd.Series, x: pd.Series, config: DictConfig):
+        """
+        Extract the annotations, check the unique entity groups, and build entity to index mappings.
+        """
+        model_config = config.model.ner
+        self.ner_special_tags = OmegaConf.to_object(model_config.special_tags)
+        self.prefix = config.model.names[0]
```
zhiqiangdon, line 32: These three lines seem more suitable in __init__().

multimodal/src/autogluon/multimodal/data/labelencoder_ner.py
cheungdaven: yes, it is expected. For example, if there is a new entity group (which is not seen in the training stage) in a validation/test datasets, the model will ignore it. Without this if statement, an error will be raised in such cases.

multimodal/src/autogluon/multimodal/data/preprocess_dataframe.py
```diff
@@ -53,7 +53,7 @@ def __init__(
             self._label_generator = label_generator
 
         # Scaler used for numerical labels
-        numerical_label_preprocessing = OmegaConf.select(config, "label.numerical_label_preprocessing")
+        numerical_label_preprocessing = OmegaConf.select(self._config.data, "label.numerical_label_preprocessing")
```
zhiqiangdon, line 56: If we pass the model config to the ner label encoder initialization, there would be no need to change the config in this file.

multimodal/src/autogluon/multimodal/data/preprocess_dataframe.py
```diff
@@ -534,6 +551,13 @@ def transform_prediction(
             y_pred = np.squeeze(y_pred)
             # Convert nan to 0
             y_pred = np.nan_to_num(y_pred)
+        elif self.label_type == NER_ANNOTATION:
+            y_pred = self._label_generator.inverse_transform(y_pred)
+            if inverse_categorical:
+                # Return annotations and offsets
+                y_pred = y_pred[1]
+            else:
+                y_pred = y_pred[0]
```
zhiqiangdon, line 560: The `inverse_categorical` is for classification tasks. It seems not meaningful for the ner task (whether returning the offsets or not).

cheungdaven: NER is essentially a classification task. In the prediction stage, we need to tell a user which group an entity belongs

zhiqiangdon: Yeah, ner is a classification task. But I feel the `inverse_categorical ` may not applicable to ner. Here are some examples of what the label encoder does for `inverse_transform`. I think you have done the similar stuff in the ner `inverse_transform`.

zhiqiangdon: What the code does here is whether to return the offsets or not, which is unrelated to `inverse_categorical `.

cheungdaven: Oh, yeah. We need to get inversed label for both evaluation metric calculation and prediction output. So you are right it is used differently here. I used it to avoid adding another flag and changing the code in evaluate

multimodal/src/autogluon/multimodal/data/labelencoder_ner.py
sxjscience: +1, config should be put in init.

multimodal/src/autogluon/multimodal/data/utils.py
```diff
@@ -258,3 +259,83 @@ def normalize_txt(text: str) -> str:
...
+
+    Returns
+    -------
+    Token-level labels or word-level lavels
+    """
+    col_tokens, token_to_word_mappings, word_offsets = tokenize_ner_text(ner_text, tokenizer)
+    num_words = len(set(token_to_word_mappings)) - 1
+    word_label = [1] * num_words
+    for idx, word_offset in enumerate(word_offsets[:num_words, :]):
+        for annot in ner_annotations:
```
sxjscience, line 287: Potentially optimize this function via binary search.

sxjscience: May add a TODO. You can do it in a separate PR.

multimodal/src/autogluon/multimodal/models/ner.py
```diff
@@ -0,0 +1,265 @@
...
+    WORD_OFFSETS,
+)
+from .utils import DummyLayer, assign_layer_ids, get_column_features, get_hf_config_and_model, init_weights
+
+hf_logging.set_verbosity_error()
+
+logger = logging.getLogger(AUTOMM)
+
+
+class HFAutoModelForNER(nn.Module):
```
sxjscience, line 33: Can we inherit https://github.com/awslabs/autogluon/blob/c1e1eb5fe6eabac739dcf5019f1e25c5cd4e1979/multimodal/src/autogluon/multimodal/models/huggingface_text.py#L28 ?

multimodal/src/autogluon/multimodal/models/ner.py
```diff
@@ -0,0 +1,265 @@
...
+            )
+
+        sequence_output = outputs.last_hidden_state
+        batch_size, max_len, feat_dim = sequence_output.shape
+        valid_output = torch.zeros(batch_size, max_len, feat_dim, dtype=torch.float32)
+
+        pooled_features = outputs.last_hidden_state[:, 0, :]
+
+        logits = self.classifier(sequence_output)
+
```
sxjscience, line 196: May be we can just add a flag to HFAutoModelForTextPrediction to indicate that it will be used for entity-recognition problems? For example, HFAutoModelForTextPrediction(is_ner=True)

cheungdaven: If we have more than two columns or other modalities, it might be easier to deal with in a standalone script.

multimodal/src/autogluon/multimodal/data/utils.py
cheungdaven: Sure, I will add a TODO here.

multimodal/src/autogluon/multimodal/data/labelencoder_ner.py
cheungdaven: it is done in the latest commit

multimodal/src/autogluon/multimodal/constants.py
zhiqiangdon: Since `NER_METRIC ` have several choices, e.g., overall_accuracy, overall_f1, overall_precision, or overall_recall, I would suggest no using it to refer to only `overall_accuracy`. Maybe it's better to define `OVERALL_ACCURACY = "overall_accuracy"` and replace `NER_METRIC` with `OVERALL_ACCURACY` in the code. In this way, we could easily use other metrics, e.g., `overall_f1`.

multimodal/src/autogluon/multimodal/data/preprocess_dataframe.py
```diff
@@ -459,6 +461,17 @@ def transform_label(
             y = self._label_scaler.transform(np.expand_dims(y, axis=-1))[:, 0].astype(np.float32)
         elif self.label_type == ROIS:
             y = y_df  # Do nothing. TODO: Shall we transform this?
+        elif self.label_type == NER_ANNOTATION:
+            text_column_index = 0  # Currently, we only support one text column.
+            x_df = df[self._text_feature_names[text_column_index]]
+            y = self._label_generator.transform(y_df)
+            x = self.transform_text(df)
```
zhiqiangdon, line 468: Calling `transform_text` inside `transform_label` looks kind of hacky since they are supposed to independent. How about adding a `transform_multimodal()` class function to handle the cases where we need to transform more than one modalities at the same time?

multimodal/src/autogluon/multimodal/constants.py
cheungdaven: Sounds good, I will add that.

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -856,14 +861,17 @@ def _fit(
...
+            self._output_shape += len(OmegaConf.to_object(config.model.ner.special_tags))
+
         config = update_config_by_rules(
             problem_type=self._problem_type,
             config=config,
         )
 
         if self._df_preprocessor is None:
             df_preprocessor = init_df_preprocessor(
-                config=config.data,
```
zhiqiangdon, line 866: I guess this change can be reversed. Using `config=config.data` should be fine.

multimodal/src/autogluon/multimodal/utils/data.py
```diff
@@ -54,10 +66,16 @@ def init_df_preprocessor(
     -------
     Initialized dataframe preprocessor.
     """
+    if label_column is not None and column_types[label_column] == NER_ANNOTATION:
+        label_generator = NerLabelEncoder(config)
+    else:
+        label_generator = None
+
     df_preprocessor = MultiModalFeaturePreprocessor(
-        config=config,
```
zhiqiangdon, line 58: I guess this change is unnecessary.

multimodal/src/autogluon/multimodal/predictor.py
cheungdaven: NerLabelEncoder is initialized init_df_preprocessor, it needs config.model. 

```   
 if label_column is not None and column_types[label_column] == NER_ANNOTATION:
        label_generator = NerLabelEncoder(config)
    else: 
        label_generator = None

    df_preprocessor = MultiModalFeaturePreprocessor(
        config=config.data,
        column_types=column_types,
        label_column=label_column,
        label_generator=label_generator,
    )
```

multimodal/src/autogluon/multimodal/constants.py
```diff
@@ -173,3 +175,12 @@
...
 SOURCEPROMPT_URL = "https://automl-mm-bench.s3.amazonaws.com/few_shot/templates.zip"
 SOURCEPROMPT_SHA1 = "c25cdf3730ff96ab4859b72e18d46ff117b62bd6"
+
+# ner
+ENTITY_GROUP = "entity_group"
+START_OFFSET = "start"
+END_OFFSET = "end"
+TOKEN_WORD_MAPPING = "token_word_mapping"
+WORD_OFFSETS = "word_offsets"
+NER = "ner"
```
zhiqiangdon, line 185: If using `NER` as a model name, better to put it together with other model names in this file. Also, don't forget to put `NER` in `HF_MODELS`.

multimodal/src/autogluon/multimodal/utils/inference.py
```diff
@@ -59,6 +74,25 @@ def extract_from_output(outputs: List[Dict], ret_type: str, as_ndarray: Optional
         return [ele[TEXT] for ele in outputs]  # single image
     elif ret_type == SCORE:
         return [ele[SCORE] for ele in outputs]
+    elif ret_type == NER:
```
zhiqiangdon, line 77: It seems that `NER` is used as both the model key and return type. How about using different names for them?

cheungdaven: what about NER_RET?

multimodal/src/autogluon/multimodal/predictor.py
zhiqiangdon: Got it. Thanks.

multimodal/src/autogluon/multimodal/utils/inference.py
zhiqiangdon: `NER_RET ` is fine. It would be better if being more specific about return contents.

cheungdaven: It is done in the latest commit.

multimodal/src/autogluon/multimodal/constants.py
cheungdaven: done in this [commit](https://github.com/awslabs/autogluon/pull/2183/commits/07f37bb1e29bb5e37a807c5498052e166455b4d7)

multimodal/src/autogluon/multimodal/data/preprocess_dataframe.py
cheungdaven: I added a TODO here and I will do the refactor in the next PR this week.

multimodal/src/autogluon/multimodal/optimization/lit_ner.py
```diff
@@ -0,0 +1,344 @@
+import logging
```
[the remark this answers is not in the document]
cheungdaven: done in this [commit](https://github.com/awslabs/autogluon/pull/2183/commits/7181f37aca173042221104ecb710e9079d6c368d).

multimodal/src/autogluon/multimodal/data/preprocess_dataframe.py
```diff
@@ -492,14 +507,17 @@ def transform_label_for_metric(
         elif self.label_type == NUMERICAL:
             # need to compute the metric on the raw numerical values (no normalization)
             y = pd.to_numeric(y_df).to_numpy()
+        elif self.label_type == NER_ANNOTATION:
+            x_df = df[self._text_feature_names[0]]
+            y = self._label_generator.transform_label_for_metric(y_df, x_df, tokenizer)
```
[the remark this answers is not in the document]
cheungdaven: labelencoder is passed when initialize the dataframe_preprocessor, code can be found [here](https://github.com/awslabs/autogluon/blob/7181f37aca173042221104ecb710e9079d6c368d/multimodal/src/autogluon/multimodal/utils/data.py#L70)

multimodal/tests/unittests/test_ner.py
```diff
@@ -0,0 +1,50 @@
...
+        time_limit=30,
+        hyperparameters={"model.ner.checkpoint_name": checkpoint_name},
+    )
+
+    scores = predictor.evaluate(train_data)
+
+    test_predict = train_data.drop(label_col, axis=1)
+    test_predict = test_predict.head(2)
+    predictions = predictor.predict(test_predict)
+    embeddings = predictor.extract_embedding(test_predict)
```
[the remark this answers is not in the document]
cheungdaven: Good suggestion, I will think about that.

multimodal/src/autogluon/multimodal/utils/metric.py
```diff
@@ -127,6 +132,10 @@ def compute_score(
     -------
     Computed score.
     """
+    if metric_name == OVERALL_ACCURACY:
+        metric = evaluate.load("seqeval")
+        return metric.compute(references=metric_data[Y_TRUE], predictions=metric_data[Y_PRED])
```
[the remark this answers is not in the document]
cheungdaven: get_metric used here is from the autogluon core, which does not support ner evaluate yet.

multimodal/setup.py
```diff
@@ -28,6 +28,9 @@
     "tqdm",
     "boto3",
     "requests",
+    "jsonschema",
+    "seqeval",
```
sxjscience, line 32: nit: We need to upper bound both packages.

sxjscience (approved):
LGTM. We can start to benchmark after this PR. The overall implementation is correct and we can move fast.

bryanyzhu (approved):
Looks good, thanks.

zhiqiangdon (approved):
Awesome new feature! Thanks!

multimodal/src/autogluon/multimodal/data/utils.py
```diff
@@ -303,3 +304,84 @@ def normalize_txt(text: str) -> str:
...
+        Whether it is for evaluation or not, default: False
+
+    Returns
+    -------
+    Token-level labels or word-level lavels
+    """
+    col_tokens, token_to_word_mappings, word_offsets = tokenize_ner_text(ner_text, tokenizer)
+    num_words = len(set(token_to_word_mappings)) - 1
+    word_label = [1] * num_words
+    # TODO: Potentially optimize word label generation via binary search
```
FANGAreNotGnu, line 331: Or sort both by offset and use a cursor?

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -250,6 +253,9 @@ def __init__(
             self._problem_type = None
             self._pipeline = ZERO_SHOT_IMAGE_CLASSIFICATION
 
+        if problem_type is not None and problem_type.lower() == NER:
```
FANGAreNotGnu, line 256: Based on our previous discussions, I think we shall use pipeline for these advanced pipelines, problem_type is for primitive tasks like classification/regression. We could consider this is a later pr.

multimodal/src/autogluon/multimodal/models/ner_text.py
```diff
@@ -0,0 +1,206 @@
...
+
+    @property
+    def text_token_word_mapping_key(self):
+        return f"{self.prefix}_{TOKEN_WORD_MAPPING}"
+
+    @property
+    def text_word_offsets_key(self):
+        return f"{self.prefix}_{WORD_OFFSETS}"
+
+    def forward(
```
FANGAreNotGnu, line 85: In a later pr we may refactor the forward code in HFAutoModelForTextPrediction a bit to reduce duplicate codes.

FANGAreNotGnu (approved):

FANGAreNotGnu (approved):
LGTM, there are a few TODOs we may consider in a later pr. Thanks for the new features!

PR #2188, opened 2022-10-05
Add Tabular forecasting model for timeseries

shchur:
*Description of changes:*
- Introduce `TabularModel` for forecasting that internally uses `autogluon.tabular.TabularPredictor` to forecast time series. Based on the GluonTS version (https://github.com/awslabs/gluonts/tree/dev/src/gluonts/nursery/autogluon_tabular) with some improvements.


timeseries/src/autogluon/timeseries/models/tabular/tabular_model.py
```diff
@@ -0,0 +1,207 @@
...
+                label=self.target,
+                problem_type=ag.constants.QUANTILE,
+                quantile_levels=self.quantile_levels,
+            )
+        else:
+            self.tabular_predictor = TabularPredictor(
+                label=self.target,
+                problem_type=ag.constants.REGRESSION,
+                eval_metric=TS_METRIC_TO_TABULAR_METRIC[eval_metric],
+            )
```
shchur, line 78: As far as I understand, `problem_type=ag.constants.QUANTILE` is still somewhat experimental in AG-T, so some changes might happen to the API there. Training with `problem_type=ag.constants.REGRESSION` is much faster and more models are available for this task. However, the regression predictor cannot produce quantiles.

Here are the possible solutions that I see:
1. Always use `problem_type=QUANTILE`. 
    - Pros: Simpler code, less hassle for the user.
    - Cons: much longer training times and potentially worse performance for users that care about any metric other than `mean_wQuantileLoss`.
2. (Current code) Train with `problem_type=QUANTILE` if `eval_metric="mean_wQuantileLoss`, or `problem_type=REGRESSION` otherwise. 
    - Pros: Faster training, better accuracy of point forecast
    - Cons: not aligned with treating probabilistic forecasting as a first-class citizen
3. Fit two `TabularPredictor`s under the hood, one for 
    - Pros: Good point forecasts with quantiles available
    - Cons: "mean" and quantile forecasts may be incompatible, longer training times

timeseries/src/autogluon/timeseries/models/tabular/tabular_model.py
```diff
@@ -0,0 +1,207 @@
...
+    "sMAPE": "mean_absolute_percentage_error",
+    "MASE": "mean_absolute_error",
+    "RMSE": "root_mean_squared_error",
+    "MSE": "mean_squared_error",
+}
+
+logger = logging.getLogger(__name__)
+
+
+class TabularModel(AbstractTimeSeriesModel):
```
shchur, line 28: Name is subject to change

timeseries/src/autogluon/timeseries/models/tabular/tabular_model.py
shchur: I did some profiling, we should be able to significantly speed up training / prediction if we vectorize `fit` and `predict` for the `BaseForestQuantileRegressor` in AG-T https://github.com/awslabs/autogluon/blob/bd1615584863e01b8123fb5d515e16cc5aca5639/tabular/src/autogluon/tabular/models/rf/rf_quantile.py#L347

Right now the training time for the quantile models is extremely long even for medium-sized datasets (e.g., `tourism_monthly`: 366 series with 60-300 timesteps each).

shchur: I went with Option 1 for now and disabled the slow models with
```
excluded_model_types=["XT", "RF"]
```

timeseries/src/autogluon/timeseries/models/tabular/tabular_model.py
```diff
@@ -0,0 +1,198 @@
...
+                feature.__class__.__name__: feature(series.index[selected_slice]) for feature in self._time_features
+            }
+            lag_columns = {f"lag_{idx}": series.shift(idx).values.ravel()[selected_slice] for idx in self._lag_indices}
+            columns = {**time_feature_columns, **lag_columns, "target": series.values.ravel()[selected_slice]}
+            df = pd.DataFrame(columns, index=series.index[selected_slice])
+            dataframe_per_item.append(df.iloc[selected_slice])
+
+        return pd.concat(dataframe_per_item, ignore_index=True)
+
+    def _generate_features_from_freq(self, freq: str):
```
canerturkmen, line 94: nit: this is not really generating the features right? A somewhat more cohesive way is to fold this under `_initialize` (see `AbstractTimeSeriesModel`)

```
def _initialize(..):
    super()._initialize()
    self._lag_indices = ..
```

timeseries/src/autogluon/timeseries/models/__init__.py
```diff
@@ -1,5 +1,4 @@
 from .gluonts import (
-    AutoTabularModel,
```
canerturkmen, line 2: do we have benchmark results for this model?

timeseries/src/autogluon/timeseries/models/tabular/tabular_model.py
```diff
@@ -0,0 +1,198 @@
...
+
+        # TODO: Other presets for TabularPredictor?
+        # TODO: Speed up prediction for Quantile XT/RF models
+        # TODO: Add catboost with MultiQuantile loss (after catboost v1.1)
+        # TODO: Other tabular models in quantile? (LightGBM?)
+        self.tabular_predictor.fit(
+            train_data=train_df,
+            tuning_data=val_df,
+            time_limit=time_limit,
+            excluded_model_types=["XT", "RF"],
```
canerturkmen, line 139: shouldn't we be allowlisting models instead of excluding them? what if Tabular introduced a new non-quantile compliant model? Can we do this through `hyperparameters`?

timeseries/src/autogluon/timeseries/models/tabular/tabular_model.py
```diff
@@ -0,0 +1,198 @@
...
+        last_observed_timestamp = data.slice_by_timestep(-1, None).index.get_level_values(TIMESTAMP)
+        offset = pd.tseries.frequencies.get_offset(data.freq)
+        item_ids = data.item_ids
+        nan_array = np.full(len(item_ids), fill_value=np.nan)
+
+        predictions_list = []
+        full_df = data
+
+        # Autoregressively generate predictions
+        for idx in range(1, self.prediction_length + 1):
```
canerturkmen, line 158: this part looks like it depends on a lot of moving parts and indexing logic. can we add a test maybe mocking out `_get_features_dataframe` to make sure the features align with the value to be forecasted?

canerturkmen (commented):
Looking good!! Mostly minor comments.

timeseries/tests/unittests/models/test_models.py
```diff
@@ -37,7 +41,7 @@ def trained_models():
             path=temp_model_path + os.path.sep,
             freq="H",
             prediction_length=prediction_length,
-            hyperparameters={"epochs": 1},
+            hyperparameters={"epochs": 1, "maxiter": 1},
```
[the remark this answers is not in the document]
shchur: `maxiter` comes from Statsmodels, I added this here to speed up the tests. Probably worth changing later when we refactor the local models / add SeasonalNaive?

timeseries/tests/unittests/models/test_models.py
```diff
@@ -170,14 +174,15 @@ def test_when_fit_called_then_models_train_and_returned_predictor_inference_has_
...
             "epochs": 1,
+            "maxiter": 1,
         },
     )
     # TFT cannot handle arbitrary quantiles
     if model.name == "TemporalFusionTransformer":
         return
 
     model.fit(train_data=DUMMY_TS_DATAFRAME)
-    predictions = model.predict(DUMMY_TS_DATAFRAME, quantile_levels=quantile_levels)
```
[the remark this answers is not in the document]
shchur: Tabular model can only predict the pre-defined quantile levels given at initialization time. Currently, we just [raise an exception](https://github.com/awslabs/autogluon/pull/2188/files#diff-d365f09219e551149f41672670e7de64bd0a5884dbbc8dcc245ea241fa703273R144) if any `quantile_levels` are passed to the predict function.

We could add a more sophisticated check that checks if the required quantiles are a subset of `self.quantile_levels`, but there is the annoying aspect of keeping track whether the `0.5` quantile [needs to be dropped](https://github.com/awslabs/autogluon/pull/2188/files#diff-d365f09219e551149f41672670e7de64bd0a5884dbbc8dcc245ea241fa703273R51). Which do you think is the better option?
1. Ban non-default quantile_levels passed to predict
2. Allow non-default quantile_levels passed to predict if they are a subset of `self.quantile_levels`?

timeseries/src/autogluon/timeseries/models/tabular/tabular_model.py
```diff
@@ -0,0 +1,198 @@
+import logging
+import re
+from typing import Any, Dict, List, Optional, Tuple
+
+import numpy as np
+import pandas as pd
+from gluonts.time_feature import TimeFeature, get_lags_for_frequency, time_features_from_frequency_str
```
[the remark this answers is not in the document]
shchur: Do you mean copy-pasting GluonTS code for `get_lags_for_frequency` and `time_features_from_frequency_str` to AGTS? I think it's fine to rely on their implementation, but I would move the patched versions to `autogluon/timeseries/utils`.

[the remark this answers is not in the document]
canerturkmen: no let's not move them. note that they are monkeypatching (we're modifying GluonTS code in place) only required for DeepAR and we want to get rid of this as soon as possible. I can take care of this later. Let's just mark this with `TODO: drop gluonts dependency`?

timeseries/src/autogluon/timeseries/models/tabular/tabular_model.py
canerturkmen: My point is that we should be allowlisting the models to use, not denylisting them. The scenario is, what if Tabular introduced `XTWORSE` algorithm that's even worse than XT? This would immediately lead to all of AutoGluonTabularModel breaking.

timeseries/src/autogluon/timeseries/models/tabular/tabular_model.py
canerturkmen: Probably both, but I hope the latter is already tested.

timeseries/src/autogluon/timeseries/models/tabular/tabular_model.py
shchur: Got it!

timeseries/src/autogluon/timeseries/models/tabular/tabular_model.py
shchur: This depends on `train_data.freq` that we only obtain during `_fit`, so doing this inside `_initialize` is too early. I will move this code inside `_fit`.

timeseries/tests/unittests/models/test_models.py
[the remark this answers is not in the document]
canerturkmen: I'm more for option 2, and in the worst case keeping this test case but skipping for Tabular?

canerturkmen (approved):
All looks good to me. Feel free to merge after addressing the test issue and failing tests. We do not necessarily need benchmarks at this stage imho.

PR #2189, opened 2022-10-06
Fix hyperparameter overloading in HPO for time series models

shchur:
*Description of changes:*
- Simplifies the logic for constructing the hyperparameter dict
- Fixes a bug where the HPO space was incorrectly bigger than the user-specified HPO space
- Enable training with fixed configs (some models have HP search spaces while others have fixed HPs)

For example the code below now only does HPO on the `num_cells` parameters.
```python
import autogluon.core as ag

TimeSeriesPredictor().fit(
    data,
    hyperparameters={
        "DeepAR": {"num_cells": ag.space.Categorical(10, 20)},
    },
    hyperparameter_tune_kwargs="random",
```
Before the same code used to do HPO on all hyperparameters of DeepAR specified in the `default_hpo` dict.

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/presets.py
+++ b/timeseries/src/autogluon/timeseries/models/presets.py
@@ -151,7 +151,7 @@ def get_preset_models(
     else:
         raise ValueError(
             f"hyperparameters must be a dict, a string or None (received {type(hyperparameters)}). "
-            "Please see the documentation for TimeSeriesPredictor.fit"
+            f"Please see the documentation for TimeSeriesPredictor.fit"
         )
 
     if hyperparameter_tune:
--- a/timeseries/src/autogluon/timeseries/models/presets.py
+++ b/timeseries/src/autogluon/timeseries/models/presets.py
@@ -135,18 +135,24 @@ def get_preset_models(
     will create models according to presets.
     """
     models = []
-    if isinstance(hyperparameters, str):
+    if hyperparameters is None:
+        hp_string = "default_hpo" if hyperparameter_tune else "default"
+        hyperparameters = copy.deepcopy(get_default_hps(hp_string, prediction_length))
+    elif isinstance(hyperparameters, str):
         hyperparameters = copy.deepcopy(get_default_hps(hyperparameters, prediction_length))
+    elif isinstance(hyperparameters, dict):
+        default_hps = copy.deepcopy(get_default_hps("default", prediction_length))
+        updated_hyperparameters = {}
+        # Only train models from `hyperparameters`, overload default HPs if provided
+        for model, hps in hyperparameters.items():
+            updated_hyperparameters[model] = default_hps.get(model, {})
+            updated_hyperparameters[model].update(hps)
+        hyperparameters = copy.deepcopy(updated_hyperparameters)
     else:
-        hp_str = "default" if not hyperparameter_tune else "default_hpo"
-        default_hps = copy.deepcopy(get_default_hps(hp_str, prediction_length))
-
-        if hyperparameters is not None:
-            # filter only default_hps for models with hyperparameters provided
-            default_hps = {model: default_hps.get(model, {}) for model in hyperparameters}
-            for model in hyperparameters:
-                default_hps[model].update(hyperparameters[model])
-        hyperparameters = copy.deepcopy(default_hps)
+        raise ValueError(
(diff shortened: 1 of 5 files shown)


timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -568,6 +568,7 @@ def _train_multi(
         if time_limit is not None and len(models) > 0:
             time_limit_model_split /= len(models)
 
+        # TODO: Allow HPO only for some models?
```
shchur, line 571: @canerturkmen, do you think we should we allow users to train some models with HPO and other models without HPO? Currently this is not supported but seems like a rather straightforward fix.

timeseries/tests/unittests/test_predictor.py
```diff
@@ -179,8 +179,7 @@ def test_given_hyperparameters_when_predictor_called_and_loaded_back_then_all_mo
 @pytest.mark.parametrize(
     "hyperparameters",
     [
-        {"ETS": {"maxiter": 1}, "SimpleFeedForward": {"epochs": 1}},
-        {"ETS": {"maxiter": 1}, "SimpleFeedForward": {"epochs": ag.Int(1, 3)}},
+        {"ETS": {"maxiter": ag.space.Categorical(1)}, "SimpleFeedForward": {"epochs": ag.space.Int(1, 3)}},
```
shchur, line 182: This test used to rely on the undocumented behavior where the HPs will be overloaded based on the `default_hpo` dict (i.e., in this a search space would be constructed for ETS).

We can either update the test (as done here), or change the logic such that some models can be trained with HPO while other models are trained with fixed configs ([as described above](https://github.com/awslabs/autogluon/pull/2189/files#r989051212))

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
canerturkmen: I agree it sounds like we should allow this (few people would do HPO for ETS, etc.)

timeseries/tests/unittests/test_predictor.py
canerturkmen: If you think we can have such mixed configs we should add test cases for those as well..

canerturkmen (approved):
LGTM! Thanks!

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -580,13 +580,18 @@ def _train_multi(
...
                 logger.info(fit_log_message)
 
-                with tqdm.external_write_mode():
-                    model_names_trained += self.tune_model_hyperparameters(
-                        model,
-                        time_limit=time_left,
-                        train_data=train_data,
-                        val_data=val_data,
-                        hyperparameter_tune_kwargs=hyperparameter_tune_kwargs,
+                if contains_searchspace(model._user_params):
```
canerturkmen, line 583: judging by `_user_params` we should not be accessing it from the outside. If there isn't a public method, should we create one?

timeseries/src/autogluon/timeseries/models/presets.py
```diff
@@ -198,30 +204,26 @@ def get_preset_models(
...
-            hp_value = model_hps[hp]
-            if isinstance(hp_value, ag.space.Space):
-                raise ValueError(
-                    f"Hyperparameter tuning not specified, so hyperparameters must have fixed values. For "
-                    f"{model}, hyperparameter {hp} currently given as search space: {hp_value}."
-                )
+def verify_contains_at_least_one_searchspace(hyperparameters: Dict[str, Dict[str, Any]]):
+    if not any(contains_searchspace(model_hps) for model_hps in hyperparameters.values()):
+        raise ValueError(
+            f"Hyperparameter tuning specified, but not a single model contains a hyperparameter search space. "
```
canerturkmen, line 217: nit: "but no model contains a hyperparameter search space"?

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
shchur: It seems that there is no public method for accessing the model parameters before initialization, so I created one

canerturkmen (approved):
LGTM! Thanks!

PR #2190, opened 2022-10-06
[Cloud] Polish IAM Policy

yinweisu:
*Issue #, if available:*

*Description of changes:*
* Polish the CloudPredictor SageMaker backend policy
* Updated the way IAM role work with CloudPredictor.
  * The previous method doesn't work properly as the role being passed into CloudPredictor is only passed to the SageMaker container. However, steps like uploading data, and start the training job actually happens locally and requires the user to authenticate themselves within the terminal. Now we fetch the current execution role instead.
  * Added a util function `CloudPredictor.generate_trust_relationship_and_iam_policy_file()` to generate necessary trust relationship and iam policy in json format. We then prompt user to review the files and create an IAM role with those files.


.github/workflows/cloud_continuous_integration.yml
```diff
@@ -7,7 +7,7 @@ on:
       - '.github/workflows/cloud_continuous_integration.yml'
       - '.github/workflow_scripts/test_cloud.sh'
   pull_request_target:
-    types: [labeled, synchronize]
```
gradientsky, line 10: Why is this needed?

cloud/src/autogluon/cloud/utils/iam.py
```diff
@@ -0,0 +1,295 @@
...
+                "arn:aws:s3:::*SageMaker*",
+                "arn:aws:s3:::*Sagemaker*",
+                "arn:aws:s3:::*sagemaker*"
+            ]
+        },
+        {
+            "Sid": "S3Bucket",
+            "Effect": "Allow",
+            "Action": [
+                "s3:CreateBucket",
```
gradientsky, line 126: Do we really need `CreateBucket`? What is it for?

cloud/src/autogluon/cloud/utils/iam.py
```diff
@@ -0,0 +1,295 @@
...
+                "s3:GetBucketLocation",
+                "s3:ListBucket",
+                "s3:ListAllMyBuckets",
```
gradientsky, line 129: Except for convenience, do we need these?

cloud/src/autogluon/cloud/utils/iam.py
```diff
@@ -0,0 +1,295 @@
...
+                "cloudwatch:GetMetricData",
+                "cloudwatch:GetMetricStatistics",
+                "cloudwatch:ListMetrics",
+                "cloudwatch:PutMetricAlarm",
+                "cloudwatch:PutMetricData",
+                "ec2:CreateNetworkInterface",
+                "ec2:CreateNetworkInterfacePermission",
+                "ec2:CreateVpcEndpoint",
+                "ec2:DeleteNetworkInterface",
+                "ec2:DeleteNetworkInterfacePermission",
```
gradientsky, line 207: Do we need network modification permissions?

cloud/src/autogluon/cloud/utils/iam.py
```diff
@@ -0,0 +1,295 @@
...
+                "ec2:DescribeSubnets",
+                "ec2:DescribeVpcEndpoints",
+                "ec2:DescribeVpcs",
+                "ecr:BatchCheckLayerAvailability",
+                "ecr:BatchGetImage",
+                "ecr:CreateRepository",
+                "ecr:Describe*",
+                "ecr:GetAuthorizationToken",
+                "ecr:GetDownloadUrlForLayer",
+                "ecr:StartImageScan",
```
gradientsky, line 221: Are we building/uploading images? Why scan is required?

cloud/src/autogluon/cloud/utils/iam.py
```diff
@@ -0,0 +1,295 @@
...
+            "Action": [
+                "s3:CreateBucket",
+                "s3:GetBucketLocation",
+                "s3:ListBucket",
+                "s3:ListAllMyBuckets",
+                "s3:GetBucketCors",
+                "s3:PutBucketCors"
+            ],
+            "Resource": [
+                "arn:aws:s3:::*"
```
gradientsky, line 134: this is very broad given the permissions

cloud/src/autogluon/cloud/utils/iam.py
```diff
@@ -0,0 +1,295 @@
...
+                "logs:GetLogEvents",
+                "logs:ListLogDeliveries",
+                "logs:PutLogEvents",
+                "logs:PutResourcePolicy",
+                "logs:UpdateLogDelivery",
+                "sns:ListTopics",
+                "tag:GetResources"
+            ],
+            "Resource": [
+                "*"
```
gradientsky, line 245: could be too broad given the permissions like `Create*`

.github/workflows/cloud_continuous_integration.yml
yinweisu: I noticed the cloud CI won't be triggered when the PR get opened and will only be triggered with following commits

cloud/src/autogluon/cloud/utils/iam.py
yinweisu: sagemaker create buckets to store some intermediate artifacts

cloud/src/autogluon/cloud/utils/iam.py
yinweisu: Technically, the training/deploy etc api call to api support vpc settings

cloud/src/autogluon/cloud/utils/iam.py
yinweisu: This one probably can be removed

.github/workflows/cloud_continuous_integration.yml
```diff
@@ -7,7 +7,7 @@ on:
...
+    types: [labeled, synchronize, opened]
```
tonyhoo, line 10: do we need this change for IAM policy refinement?

cloud/src/autogluon/cloud/predictor/cloud_predictor.py
```diff
@@ -86,13 +95,19 @@ def __init__(
...
-        self.role_arn = role_arn
-        if not self.role_arn:
-            self.role_arn = setup_sagemaker_role_and_policy(
-                role_name='ag_cloud_predictor_role',
-                trust_relationship=SAGEMAKER_TRUST_REPLATIONSHIP,
-                policies=SAGEMAKER_POLICIES
+        try:
+            self.role_arn = sagemaker.get_execution_role()
+        except ClientError as e:
+            logger.warning(
```
tonyhoo, line 101: Do we assume the user will run this with some form of Admin role. If not, can we provide the minimal set of policies required to the get the role setup?

cloud/src/autogluon/cloud/predictor/cloud_predictor.py
```diff
@@ -127,6 +142,62 @@ def endpoint_name(self):
...
         return None
 
+    @staticmethod
+    def generate_trust_relationship_and_iam_policy_file(
+        account_id: str,
+        cloud_output_bucket: str,
+        output_path: Optional[str] = None
+    ):
+        """
+        Generate required trust relationship and IAM policy file in json format for CloudPredictor wit SageMaker backend.
```
tonyhoo, line 152: [tpyo] `wit` → `with`

cloud/src/autogluon/cloud/utils/iam.py
```diff
@@ -0,0 +1,295 @@
...
+                "sagemaker:InvokeEndpoint"
+            ],
+            "Resource": [
+                f"arn:aws:sagemaker:*:{POLICY_ACCOUNT_PLACE_HOLDER}:artifact/*",
+                f"arn:aws:sagemaker:*:{POLICY_ACCOUNT_PLACE_HOLDER}:transform-job/*",
+                f"arn:aws:sagemaker:*:{POLICY_ACCOUNT_PLACE_HOLDER}:endpoint/*",
+                f"arn:aws:sagemaker:*:{POLICY_ACCOUNT_PLACE_HOLDER}:training-job/*",
+                f"arn:aws:sagemaker:*:{POLICY_ACCOUNT_PLACE_HOLDER}:model/*",
+                f"arn:aws:sagemaker:*:{POLICY_ACCOUNT_PLACE_HOLDER}:endpoint-config/*"
+            ]
```
tonyhoo, line 58: [minor] for security sensitive customers, can we make sure all the SM resources created with prefix `ag` so that we can further constraint the resource access required? Some customers might be reluctant to allow us to access other resources which might not be related to AG at all

cloud/src/autogluon/cloud/utils/iam.py
tonyhoo: Just curious what if the bucket already exists, what do we do? S3 bucket names are global so it is highly likely

cloud/src/autogluon/cloud/utils/iam.py
tonyhoo: This might not be friendly to security sensitive customers. Let's narrow it down if can be avoided

cloud/src/autogluon/cloud/utils/iam.py
```diff
@@ -0,0 +1,295 @@
...
+                "s3:PutBucketCors"
+            ],
+            "Resource": [
+                "arn:aws:s3:::*"
+            ]
+        },
+        {
+            "Sid": "ListEvents",
+            "Effect": "Allow",
+            "Action": [
```
tonyhoo, line 140: If we can prefix the resource with something like `ag`, we can avoid all list operations or limit them with constraints

cloud/src/autogluon/cloud/utils/iam.py
```diff
@@ -0,0 +1,295 @@
...
+                "sagemaker:ListTrainingJobs",
+                "sagemaker:ListModels"
+            ],
+            "Resource": [
+                "*"
+            ]
+        },
+        {
+            "Effect": "Allow",
+            "Action": [
```
tonyhoo, line 154: Not sure what are these for?

cloud/src/autogluon/cloud/utils/iam.py
tonyhoo: What is the VPC being used for? Do we need VPC for training and inference endpoint? We can just enable network isolation to block the internet access if needed: https://docs.amazonaws.cn/en_us/sagemaker/latest/dg/mkt-algo-model-internet-free.html

cloud/src/autogluon/cloud/utils/iam.py
yinweisu: Then it's a sagemaker problem I guess. But sagemaker will create a bucket with a format like `sagemaker-region-account_id`

cloud/src/autogluon/cloud/utils/iam.py
yinweisu: There's some api call requiring this. I remember removing it will end up with error

cloud/src/autogluon/cloud/utils/iam.py
yinweisu: so sagemaker api allow user to pass vpc stuffs. After discussion with Alex, we agree that we shouldn't put this in the policy we provided to the user. User can add it themselves if they need so

.github/workflows/cloud_continuous_integration.yml
yinweisu: No. It's just a quick fix for the CI. Didn't want to bother to create a specific PR just for this

cloud/src/autogluon/cloud/predictor/cloud_predictor.py
yinweisu: We are providing the util function for that

cloud/src/autogluon/cloud/predictor/cloud_predictor.py
```diff
@@ -54,7 +68,6 @@ class CloudPredictor(ABC):
     def __init__(
         self,
         cloud_output_path,
```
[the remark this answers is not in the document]
yinweisu: We likely won't because of political issue. When it comes to we are going to support a different backend. We could probably turn the current CloudPredictor, which is not user facing as user will use `TabularCloudPredictor` for example, into SageMakerCloudPredictor. And create a new interface CloudPredictor that has backend parameter inside the init function. It will just wrap SageMakerCloudPredictor

cloud/src/autogluon/cloud/utils/iam.py
```diff
@@ -0,0 +1,295 @@
...
+    "Statement": [
+        {
+            "Sid": "SageMaker",
+            "Effect": "Allow",
+            "Action": [
+                "sagemaker:DescribeEndpoint",
+                "sagemaker:DescribeEndpointConfig",
+                "sagemaker:DescribeModel",
+                "sagemaker:DescribeTrainingJob",
+                "sagemaker:DescribeTransformJob",
```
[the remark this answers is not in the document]
yinweisu: We do

cloud/src/autogluon/cloud/utils/iam.py
yinweisu: I think this is a good idea. This would require me look into how to add the prefix assuming it's doable. We can have a follow up PR that modify both the source code and further scope down the policy

cloud/src/autogluon/cloud/utils/iam.py
yinweisu: ditto

cloud/src/autogluon/cloud/utils/iam.py
gradientsky: @tonyhoo VPC/subnets/security groups should be created outside of any SM activity, they should just be passed to the call as the existing resources.

cloud/src/autogluon/cloud/utils/iam.py
yinweisu: Updated to only include logs with permission like `Create`, should be less of a concern now

cloud/src/autogluon/cloud/utils/iam.py
tonyhoo: I know some users might have VPC set up which enhance the data security. However, wondering if AG users need it as the container won't actively polling the data outside of the container env. Am I right? If so, we can set the network_isolation flag to True to prevent any internet access for security

cloud/src/autogluon/cloud/utils/iam.py
gradientsky: +1 on follow-up PR to add prefixes

tonyhoo (approved):

PR #2028, opened 2022-08-10
Upgrade PyTorch Lightning to 1.7.x

sxjscience:
*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2191, opened 2022-10-06
Remove ONNX Install on Macos

yinweisu:
*Issue #, if available:*

*Description of changes:*
* https://github.com/awslabs/autogluon/actions/runs/3195498402/jobs/5216230879
macos doesn't have the required onnx package; therefore, add platform check to avoid installing it on macos

@FANGAreNotGnu Feel free to update this PR to add necessary warning or try import onnx functionality in case user tries to export onnx on a mac but onnx is not installed

### DIFF
--- a/multimodal/setup.py
+++ b/multimodal/setup.py
@@ -61,7 +61,7 @@ extras_require = {
         "black~=22.0,>=22.3",
         "isort>=5.10",
         "datasets>=2.3.2,<=2.3.2",
-        "onnxruntime-gpu>=1.12.1,<=1.12.1",
+        "onnxruntime-gpu>=1.12.1,<=1.12.1;platform_system!='Darwin'",
     ]
 }
 


FANGAreNotGnu (approved):
LGTM, thanks! I'll update it to use cpu version on macos later.

multimodal/setup.py
```diff
@@ -61,7 +61,7 @@
         "black~=22.0,>=22.3",
         "isort>=5.10",
         "datasets>=2.3.2,<=2.3.2",
-        "onnxruntime-gpu>=1.12.1,<=1.12.1",
+        "onnxruntime-gpu>=1.12.1,<=1.12.1;platform_system!='Darwin'",
```
zhiqiangdon, line 64: The master branch job failed due to timeout. I wonder whether it's due to the installation of `datasets` and `onnxruntime-gpu`.

yinweisu: https://github.com/awslabs/autogluon/pull/2028 The master branch seems to start fail since this PR

zhiqiangdon: I see. It's lightning upgrade. Do you think we need to fall back to the previous?

PR #2192, opened 2022-10-07
Temporarily skip AutoMM HPO tests (lightning 1.7)

zhiqiangdon:
Lightning 1.7 seems not fully compatible with ray tune, which causes the AutoMM HPO tests to get stuck. Temporarily skip the AutoMM HPO tests. Need to investigate how to make ray tune or ray lightning work with lightning 1.7.  

### DIFF
--- a/multimodal/setup.py
+++ b/multimodal/setup.py
@@ -35,7 +35,7 @@ install_requires = [
     "fairscale>=0.4.5,<=0.4.6",
     "scikit-image>=0.19.1,<0.20.0",
     "smart_open>=5.2.1,<5.3.0",
-    "pytorch_lightning>=1.6.0,<1.7.0",
+    "pytorch_lightning>=1.6.0,<1.8.0",
     "text-unidecode<=1.3",
     "torchmetrics>=0.7.2,<0.8.0",
     "transformers>=4.18.0,<4.21.0",
--- a/multimodal/setup.py
+++ b/multimodal/setup.py
@@ -35,7 +35,7 @@ install_requires = [
     "fairscale>=0.4.5,<=0.4.6",
     "scikit-image>=0.19.1,<0.20.0",
     "smart_open>=5.2.1,<5.3.0",
-    "pytorch_lightning>=1.6.0,<1.8.0",
+    "pytorch_lightning>=1.6.0,<1.7.0",
     "text-unidecode<=1.3",
     "torchmetrics>=0.7.2,<0.8.0",
     "transformers>=4.18.0,<4.21.0",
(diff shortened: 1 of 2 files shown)


yinweisu (approved):
LGTM! Let's wait for the CI to pass before merging in. Thanks!

sxjscience:
@yinweisu @zhiqiangdon Let's skip the HPO test first instead of downgrading lightning. We should fix our the HPO part of AutoMM.

sxjscience:
AutoMM + HPO is largely still experimental and require additional work. We may also wait for https://github.com/ray-project/ray_lightning/pull/196 . Meanwhile, let's find the reason for HPO test to stuck.

yinweisu:
> @yinweisu @zhiqiangdon Let's skip the HPO test first instead of downgrading lightning. We should fix our the HPO part of AutoMM.

That's another option. We are not sure why PL 1.7 will hang with ray tune yet. My question is: is there a specific reason that we want to upgrade to PL 1.7? If it's an issue between ray tune and PL 1.7, it's likely that we won't be able to resolve it in a short time and we will lose hpo capability.

yinweisu:
> AutoMM + HPO is largely still experimental and require additional work. We may also wait for [ray-project/ray_lightning#196](https://github.com/ray-project/ray_lightning/pull/196) . Meanwhile, let's find the reason for HPO test to stuck.

Sure. I'll try to upgrade ray to 2.0 next week to see if it helps

sxjscience:
@yinweisu Upgrading lightning will unlock the official support of torch 1.12. We may need to find the root cause of the hanging and enable some tests again.

Does the hanging happen very often? Also, which test hanged?

yinweisu:
> @yinweisu Upgrading lightning will unlock the official support of torch 1.12. We may need to find the root cause of the hanging and enable some tests again.
> 
> Does the hanging happen very often? Also, which test hanged?

It seems to happen quite often and @zhiqiangdon find that hpo tests are the tests that got stuck. It could be tricky to find the cause as there's no log at all. I'll try upgrading ray to 2.0 as the first step because that's something I plan to do anyway

zhiqiangdon:
> @yinweisu Upgrading lightning will unlock the official support of torch 1.12. We may need to find the root cause of the hanging and enable some tests again.
> 
> Does the hanging happen very often? Also, which test hanged?

https://github.com/awslabs/autogluon/blob/master/multimodal/tests/unittests/test_hpo.py#L15
I saw it got stuck at the 4th hpo test.

sxjscience:
> > @yinweisu Upgrading lightning will unlock the official support of torch 1.12. We may need to find the root cause of the hanging and enable some tests again.
> > Does the hanging happen very often? Also, which test hanged?
> 
> It seems to happen quite often and @zhiqiangdon find that hpo tests are the tests that got stuck. It could be tricky to find the cause as there's no log at all. I'll try upgrading ray to 2.0 as the first step because that's something I plan to do anyway

Yes, let's upgrade to ray 2.0. We should upgrade it anyway.

PR #2193, opened 2022-10-07
Try Removing Protobuf as A Direct Dependency

yinweisu:
*Issue #, if available:*

*Description of changes:*
* Had a hard dependency of protobuf because our dependencies didn't handle protobuf well when it had a major version upgrade. Try removing it now


gradientsky (approved):

gradientsky:
Linking the original issue why the pin was in place: #1762

PR #2196, opened 2022-10-07
Replace models in detection test to a smaller one.

FANGAreNotGnu:
Changed cascade_rcnn_x101 to mask_rcnn_r50 in detection test to avoid the occasional error if the downloading model size is too large.

Also changed mim install timeout value from default (15s) to 120s.

TODO: Fix missing config contents in centernet (test_pipeline.image_scale) and ssd (mean, std) in next pr.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2197, opened 2022-10-08
Enabling DeepSpeed to support parameter-efficient finetuning of very large models on single GPU (11B+ parameters)

Raldir:
*Description of changes:*

- Enabling Deepspeed in AutoGluon, by adapting the associated [pytorch-lightning strategy](https://pytorch-lightning.readthedocs.io/en/stable/advanced/model_parallel.html#deepspeed-zero-stage-3). Setting the strategy to `deepspeed_stage_3_offload` now allows to fine-tune models such as T0pp (11B parameters) with parameter-efficient finetuning methods, such as IA^3, on a single A10G GPU (23GB memory), by offloading optimizer states to the CPU. Particularly suited to few-shot learning, where training is generally very quick and the effect of offloading is relatively minor (see PR #2032). Some hacky workarounds are currently in place and indicated in code, since pytorch-lightning has some limitations in its use.
- Now intermediate checkpoints only store weights that are updated during training, i.e. `requires_grad=True`. Added a flag `standalone` to `fit()` to indicate whether the final checkpoint should consist of the entire model's weights or only trained weights. In the case of T0pp and fine-tuning with (IA)^3, instead of needing 40GB per checkpoint only 4.2MB are needed. 
- Fixed gradient_checkpointing for T-Few, works correctly now.
- Adding seed  argument to `_predict()` to ensure reproducibility.
- Fixed some style/design issues regarding T-Few.

TODOs:
- Add zero-shot mode for T0pp.
- Add Deepspeed inference mode with custom kernels and quantization for improved inference speed and memory consumption, see [here](https://www.deepspeed.ai/tutorials/inference-tutorial/).


multimodal/src/autogluon/multimodal/constants.py
```diff
@@ -137,6 +137,12 @@
 IA3_BIAS = "ia3_bias"
 IA3_NORM = "ia3_norm"
 
+# DeepSpeed constants (for now gather/reduce size are constants for single GPU)
+DEEPSPEED_OFFLOADING = "deepspeed_stage_3_offload"
+DEEPSPEED_MIN_PL_VERSION = "1.7.7"
+DEEPSPEED_ALLGATHER_SIZE = 1e9
+DEEPSPEED_ALLREDUCE_SIZE = 1e9
```
sxjscience, line 144: We can put the `ALLGATHER_SIZE` and `ALLREDUCE_SIZE` as configurable options in https://github.com/awslabs/autogluon/blob/master/multimodal/src/autogluon/multimodal/configs/environment/default.yaml .

multimodal/src/autogluon/multimodal/optimization/deepspeed.py
```diff
@@ -0,0 +1,976 @@
+# Slightly adapted file of pytorch_lightning.strategies.deepspeed to adjust for additional memory requirements when running deepspeed inference.
```
sxjscience, line 1: We can also summarize about what we have changed.

multimodal/src/autogluon/multimodal/optimization/lit_module.py
```diff
@@ -41,7 +41,7 @@ def __init__(
         custom_metric_func: Callable = None,
         test_metric: Optional[torchmetrics.Metric] = None,
         efficient_finetune: Optional[str] = None,
-        trainable_param_names: Optional[List[str]] = None,
+        trainable_param_names: Optional[List] = [],
```
sxjscience, line 44: We should still use `None` as the default value. Also, should it be `List[str]`?

sxjscience: We can cast None to `[]` later .

multimodal/src/autogluon/multimodal/utils/checkpoint.py
```diff
@@ -53,6 +61,70 @@ def average_checkpoints(
     return avg_state_dict
 
 
+class AutoMMModelCheckpointIO(pl.plugins.CheckpointIO):
+    """
+    Class that customizes how checkpoints are saved. Saves either the entire the entire model or only parameters that have been explicitly updated during training. The latter reduces memory footprint substentially when training very large models with parameter-efficient finetuning methods.
```
sxjscience, line 66: Nit: "the entire the entire" --> "the entire".

multimodal/src/autogluon/multimodal/utils/checkpoint.py
```diff
@@ -53,6 +61,70 @@ def average_checkpoints(
...
+    Class is based on pl.plugins.TorchCheckpointIO.
+    """
+
+    def __init__(self, trainable_param_names, model_name_to_id):
```
sxjscience, line 70: Need to add docstring of `trainable_param_names` and `model_name_to_id`.

multimodal/src/autogluon/multimodal/optimization/utils.py
```diff
@@ -535,7 +535,7 @@ def apply_layerwise_lr_decay(
     lr_decay: float,
     weight_decay: float,
     efficient_finetune: Optional[str] = None,
-    trainable_param_names: Optional[List[str]] = None,
+    trainable_param_names: Optional[List] = [],
```
sxjscience, line 538: Similar to the previous line. We should use `None` as default.

sxjscience:
I noticed that there seems to an installation timeout issue associated mmcv-full. 

Error message:

```
 [2022-10-12T00:17:40.619Z] + mim install mmcv-full --default-timeout 120
[2022-10-12T00:17:42.038Z] Looking in links: https://download.openmmlab.com/mmcv/dist/cu102/torch1.12.0/index.html
[2022-10-12T00:17:42.878Z] Collecting mmcv-full
[2022-10-12T00:17:43.107Z]   Downloading https://download.openmmlab.com/mmcv/dist/cu102/torch1.12.0/mmcv_full-1.6.2-cp38-cp38-manylinux1_x86_64.whl (33.4 MB)
[2022-10-12T00:19:00.281Z]      ━━━━━━━━━━                               8.8/33.4 MB 262.1 kB/s eta 0:01:34
```

Will it be related to the timeout? @bryanyzhu 
https://github.com/awslabs/autogluon/blob/f8c0ede595093c13011fe6c50b1c7ffa8b9c1575/.github/workflow_scripts/env_setup.sh#L61

multimodal/src/autogluon/multimodal/data/process_text.py
```diff
@@ -357,8 +357,8 @@ def build_one_token_sequence_from_text(
             if col_name == CHOICES_IDS:
                 answer_ids = self.tokenizer(
                     col_text,
-                    return_tensors="pt",
```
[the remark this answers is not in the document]
Raldir: Thanks! :) Other point on the list was the `trainable_param_names` argument, which was an artefact then but now has proper functionality.

sxjscience (approved):
LGTM!! Thanks! Let’s add a tutorial afterwards.

.github/workflow_scripts/env_setup.sh
```diff
@@ -58,7 +58,7 @@ function install_multimodal {
     # launch different process for each test to make sure memory is released
     python3 -m pip install --upgrade pytest-xdist
     python3 -m pip install --upgrade -e multimodal/[tests]
-    mim install mmcv-full --default-timeout 120
```
zhiqiangdon, line 61: @FANGAreNotGnu added this flag in this [PR](https://github.com/awslabs/autogluon/pull/2196) as suggested by @bryanyzhu . I guess it is to handle the timeout issue when installing mmcv. Is there a reason that we remove it?

sxjscience: It is causing CI to fail.

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -1771,6 +1875,8 @@ def predict(
         elif self._pipeline == OCR_TEXT_RECOGNITION:
             ret_type = [TEXT, SCORE]
 
+        pl.seed_everything(seed, workers=True)
```
zhiqiangdon, line 1878: Why do we need seeding in prediction?

.github/workflow_scripts/env_setup.sh
sxjscience: May be @Raldir we can try --timeout 1000.

sxjscience: 120 can still be too small I think. May be we can also revise it in the next PR.

multimodal/src/autogluon/multimodal/predictor.py
sxjscience: I think it’s a good practice to seed before running models, even in the inference script. This avoids potential unreproducibility issues. An example is https://github.com/NVIDIA/FasterTransformer/blob/bc077a9078b99244c69cfe0dc44af86fd974bc71/examples/pytorch/bert/run_glue.py#L290

Raldir: Specifically T-Few selects a template randomly (both for training and inference as per default in the paper), so not seeding before running it results in reproducibility issues.

.github/workflow_scripts/env_setup.sh
Raldir: Ok, will revise to 1000

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -1346,6 +1412,37 @@ def _default_predict(
...
+                lr=self._config.optimization.learning_rate,
+                lr_decay=self._config.optimization.lr_decay,
+                end_lr=self._config.optimization.end_lr,
+                lr_mult=self._config.optimization.lr_mult,
+                weight_decay=self._config.optimization.weight_decay,
+                warmup_steps=self._config.optimization.warmup_steps,
+            )
+        else:
+            optimization_kwargs = {}
+            trainable_param_names = []
```
zhiqiangdon, line 1444: Do we still need this deepspeed strategy in inference? Isn't it only necessary for training?

zhiqiangdon: Consider adding a TODO to improve the design. Using `optimization_kargs` and `trainable_param_names` in inference is kind of confusing since no training and optimization are not expected in inference.

multimodal/src/autogluon/multimodal/predictor.py
zhiqiangdon: Here we only seed in `.predict()`. It can't be used by other APIs, e.g., `.predict_proba()`.

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -1560,6 +1662,7 @@ def _predict(
         data: Union[pd.DataFrame, dict, list],
         requires_label: bool,
         realtime: Optional[bool] = None,
+        seed=123,
```
zhiqiangdon, line 1665: The seed seems unused in `_predict()`?

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -2034,9 +2147,11 @@ def save(self, path: str, standalone: Optional[bool] = True):
                 fp,
                 ensure_ascii=True,
             )
+
         # In case that users save to a path, which is not the original save_path.
         if os.path.abspath(path) != os.path.abspath(self._save_path):
             model_path = os.path.join(self._save_path, "model.ckpt")
+            print(model_path, path)
```
zhiqiangdon, line 2154: Remove `print()` or use `logger.debug()`.

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -1321,7 +1373,21 @@ def _top_k_average(
...
+                checkpoint = {
+                    "state_dict": {
+                        name.partition("module.")[2]: param
+                        for name, param in strategy.model._zero3_consolidated_16bit_state_dict().items()
+                    }
+                }
+            else:
+                checkpoint = {
+                    "state_dict": {"model." + name: param for name, param in self._model.state_dict().items()}
+                }
```
zhiqiangdon, line 1389: I guess this is the same as `checkpoint = {"state_dict": avg_state_dict}` in `if not standalone`, right?

zhiqiangdon (approved):
Thanks for the awesome feature! Left several minor comments.

multimodal/src/autogluon/multimodal/predictor.py
sxjscience: @Raldir I think we can change it.

multimodal/src/autogluon/multimodal/predictor.py
Raldir: In the situation that not all parameters are updated, `avg_state_dict` does not contain all model weights, only the updated ones. If `standalone==True`, we want to save entire the model's weights, hence going through the model's entire `state_dict` in that case.

multimodal/src/autogluon/multimodal/predictor.py
Raldir: Ah right, good point!

multimodal/src/autogluon/multimodal/predictor.py
Raldir: Good point, added a TODO now. Unfortunately, specifying `optmization_kwargs` is currently a requirement in pytorch-lightning integration of deepspeed but should be removed as soon as possible. Calling deepspeed still needed in inference to run large models to reduce computation to `largest_layer_memory + 2 * params`.

PR #2203, opened 2022-10-12
Enable Ray Tune backend for time series models

shchur:
*Description of changes:*
- Use Ray Tune as default HPO backend for GluonTS models. We don't use Ray for other models (e.g., local models from Statsmodels) since 
    1. their searchspace is just a grid
    2. they don't benefit from being trained in parallel
    3. they are deterministic, so always training `num_trials` models with repeated hyperparameters (as done by Ray) doesn't make sense.
- Update HPO results reporting logic in `abstract_trainer.py`
- Update `searcher` mapping in `autogluon.core`

*To do:*
- [x] Discuss TODOs marked in code
- [x] Add documentation for `hyperparameter_tune_kwargs` in `TimeSeriesPredictor`
- [x] Add more tests?
- [x] Fix Ray warnings

### DIFF
--- a/core/src/autogluon/core/hpo/executors.py
+++ b/core/src/autogluon/core/hpo/executors.py
@@ -158,6 +158,7 @@ class RayHpoExecutor(HpoExecutor):
     custom_to_ray_searcher_preset_map = {
         'local_random': 'random',
         'random': 'random',
+        # TODO: Should this be changed to 'auto': 'bayes'?
         'auto': 'random',
     }
     
--- a/core/src/autogluon/core/hpo/executors.py
+++ b/core/src/autogluon/core/hpo/executors.py
@@ -158,8 +158,7 @@ class RayHpoExecutor(HpoExecutor):
     custom_to_ray_searcher_preset_map = {
         'local_random': 'random',
         'random': 'random',
-        # TODO: Should this be changed to 'auto': 'bayes'?
-        'auto': 'random',
+        'auto': 'bayes',
     }
     
     def __init__(self):
--- a/timeseries/src/autogluon/timeseries/predictor.py
+++ b/timeseries/src/autogluon/timeseries/predictor.py
@@ -57,8 +57,7 @@ class TimeSeriesPredictor:
         in order to improve this metric on validation data, and ranks models (on validation data) according to this
         metric. Available options:
 
-        - ``"mean_wQuantileLoss"``: mean weighted quantile loss, defined as average of quantile losses for the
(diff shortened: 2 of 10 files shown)


core/src/autogluon/core/hpo/executors.py
```diff
@@ -158,6 +158,7 @@ class RayHpoExecutor(HpoExecutor):
     custom_to_ray_searcher_preset_map = {
         'local_random': 'random',
         'random': 'random',
+        # TODO: Should this be changed to 'auto': 'bayes'?
         'auto': 'random',
     }
```
shchur, line 163: Currently it's impossible to simultaneously train 
- some models with `backend=custom`, `searcher=random`
- other models with `backend=ray`, `searcher=bayes`

Mapping `auto` to `bayes` backend here would enable us to do that with
```
hyperparameter_tune_kwargs={
    "scheduler": "local",
    "searcher": "auto",
    "num_trials": 10,
}
""

File: timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py [hunk 2]
```diff
@@ -427,18 +427,19 @@ def tune_model_hyperparameters(
         ):
             default_num_trials = 10
 
+        tuning_start_time = time.time()
         with disable_tqdm():
-            hpo_models, hpo_results = model.hyperparameter_tune(
+            hpo_models, _ = model.hyperparameter_tune(
```
shchur, line 432: `hpo_models` contains all the information that we need to report in the same format regardless of the backend, so we ignore `hpo_results` (which is backend-dependent)

core/src/autogluon/core/hpo/executors.py
canerturkmen: @yinweisu

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -347,7 +347,7 @@ def fit(
         self.save()
         return self
 
-    # TODO: to be changed after ray tune integration
+    # TODO: Is this method safe to delete?
```
canerturkmen, line 350: I believe so.

canerturkmen (commented):
All looks to me for now. I'll defer to @yinweisu for the discussion on adjustments to core.

yinweisu:
Have synced with @shchur offline. We can add mappings in core to address the presets. Regarding the warnings, remove `checkpoint_dir` inside `model_trial` can avoid the checkpoint warning. Will need a minimal example to checkout other warnings

PR #2205, opened 2022-10-12
[WIP] Bump gluonts dependency version to 0.11

canerturkmen:
*Issue #, if available:*

*Description of changes:*

Bump GluonTS dependency version to the latest v0.11. GluonTS v0.10 and v0.11 bring a lot of benefits (i.e., PyTorch models) in addition to breaking changes, most importantly changing the timestamp data type to `pd.Period`. We handle these and other side effects in this PR in a non-breaking manner.

TODO:
- [ ] Ensure no performance regressions.


timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
```diff
@@ -261,54 +260,49 @@ def _predict_gluonts_forecasts(self, data: TimeSeriesDataFrame, **kwargs) -> Lis
 
         return list(self.gts_predictor.predict(**predictor_kwargs))
 
+    @staticmethod
+    def _sample_to_quantile_forecast(forecast: np.ndarray, quantile_levels: List[float]) -> QuantileForecast:
+        forecast_arrays = []
```
shchur, line 265: Is `forecast` a 1D array of sampled values at a certain timestamp? That is, we call `_sample_to_quantile_forecast` `data.num_items * prediction_length` times?

Maybe it makes sense to add a short annotation explaining the expected shapes of arrays?

shchur (approved):
Thanks, LGTM!

I would also try fitting a predictor with all GluonTS models enabled with `hyperparameters` to make sure that nothing broke internally

PR #2215, opened 2022-10-14
AutoMM MMDet

FANGAreNotGnu:
1. Add MMDetLitModule including data pack/unpack for mmcv formats.
2. Enable evaluation in COCO format. Note that we also support evaluation for VOC and custom dataset, but output metrics are COCO style.
3. Enable training and finetuning, especially when num_classes changes, e.g. finetune a COCO-pretrained model on VOC.
4. For VOC dataset, we support input in both VOC style and COCO style, but it's recommended to use COCO style.
5. Multi-gpu is not tested. Will work on it in next pr.

Finetuning result of yolov3_mobilenetv2_320_300e on VOC dataset:
```
    map = 0.3856 (MMDetection: 0.351)
    map50 = 0.7082 (MMDetection: 0.639)
```

Evaluation example code:
```
    predictor = MultiModalPredictor(
        label="rois_label",
        hyperparameters={
            "model.mmdet_image.checkpoint_name": <checkpoint_name>,
            "env.num_gpus": 1,
        },
        pipeline="object_detection",
    )
    predictor.evaluate(<test_data_path>)
```

Finetuning example code:
```
    predictor = MultiModalPredictor(
        label="rois_label",
        hyperparameters={
            "model.mmdet_image.checkpoint_name": <model_pretrained_on_COCO>,
            "env.num_gpus": 1,
        },
        pipeline="object_detection",
        output_shape=20,
    )
    predictor.fit(<train_data_path>)
```


examples/automm/object_detection/detection_train.py
```diff
@@ -0,0 +1,97 @@
...
+
+    detection_train(
+        train_path=args.train_path,
+        test_path=args.test_path,
+        checkpoint_name=args.checkpoint_name,
+        num_classes=args.num_classes,
+        lr=args.lr,
+        wd=args.wd,
+        epochs=args.epochs,
+    )
```
sxjscience, line 97: nit: newline.

.github/workflow_scripts/env_setup.sh
```diff
@@ -62,7 +62,7 @@ function install_multimodal {
     # launch different process for each test to make sure memory is released
     python3 -m pip install --upgrade pytest-xdist
     python3 -m pip install --upgrade -e multimodal/[tests]
-    mim install mmcv-full --timeout 1000
+    mim install mmcv-full --timeout 60
```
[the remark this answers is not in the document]
FANGAreNotGnu: It seems that setting this to 1000 caused mmcv import error in tutorial build.

examples/automm/object_detection/detection_train.py
```diff
@@ -0,0 +1,100 @@
+"""
+The example to train an object detection model in AutoMM.
+
+An example to finetune an MMDetection model on COCO:
+    python detection_train.py \
+        --train_path coco17/annotations/instances_train2017.json
```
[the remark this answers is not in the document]
FANGAreNotGnu: So it will be like build a custom coco dataset like [this (using code)](https://www.immersivelimit.com/tutorials/create-coco-annotations-from-scratch) or [this (using DataTorch GUI)](https://anujsyal.com/building-computer-vision-datasets-in-coco-format)?

.github/workflow_scripts/env_setup.sh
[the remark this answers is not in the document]
FANGAreNotGnu: I think that is because the server mim using is not stable enough. When there is a connection issue, setting this number too low will cause a timeout error, but too high will cause the import error while building tutorial.

multimodal/src/autogluon/multimodal/data/preprocess_dataframe.py
```diff
@@ -398,12 +398,16 @@ def transform_image(
...
         ), "You will need to first call preprocessor.fit_x() before calling preprocessor.transform_image."
+
         image_features = {}
         image_types = {}
         for col_name in self._image_feature_names:
             col_value = df[col_name]
             col_type = self._column_types[col_name]
-            if col_type == IMAGE_PATH or IMAGE:
+
+            if col_name == ROIS:
```
zhiqiangdon, line 408: Should be `col_type == ROIS`?

multimodal/src/autogluon/multimodal/data/preprocess_dataframe.py
```diff
@@ -281,7 +281,7 @@ def _fit_x(self, X: pd.DataFrame):
                     generator = self._feature_generators[col_name]
                     generator.fit(np.expand_dims(processed_data.to_numpy(), axis=-1))
                     self._numerical_feature_names.append(col_name)
-            elif col_type.startswith(IMAGE):
+            elif col_type.startswith(IMAGE) or col_type == ROIS:
```
zhiqiangdon, line 284: Maybe mark a TODO that makes `ROIS` an independent modality and uses the `transform_multimodal()` in which we can call `transform_image()` and `transform_rois()`.

multimodal/src/autogluon/multimodal/data/process_image.py
```diff
@@ -164,9 +164,12 @@ def __init__(
...
                 assert mmocr is not None, "Please install MMOCR by: pip install mmocr."
             cfg = model.model.cfg
-            cfg.data.test.pipeline = replace_ImageToTensor(cfg.data.test.pipeline)
-            self.val_processor = Compose(cfg.data.test.pipeline)
-            self.train_processor = Compose(cfg.data.test.pipeline)
+            try:  # yolov3
+                training_pipeline = cfg.data.train.dataset.pipeline
+            except:  # faster_rcnn
+                training_pipeline = cfg.data.train.pipeline
+            self.val_processor = Compose(replace_ImageToTensor(cfg.data.val.pipeline))
```
zhiqiangdon, line 171: What's the difference between `cfg.data.test.pipeline` and `cfg.data.val.pipeline`? The `val_processor` is used in both validation and testing, e.g., when calling `.predict()`, `.evaluate()`.

multimodal/src/autogluon/multimodal/data/process_image.py
```diff
@@ -395,14 +398,21 @@ def process_one_sample(
...
         zero_images = []
+        mm_data = dict(img_prefix=None, bbox_fields=[])
         ret = {}
         column_start = 0
-        for per_col_name, per_col_image_paths in image_paths.items():
-            for img_path in per_col_image_paths[: self.max_img_num_per_col]:
-                if self.prefix == MMDET_IMAGE or self.prefix.lower().startswith(MMOCR):
-                    data = dict(img_info=dict(filename=img_path), img_prefix=None)
-                    images.append(self.val_processor(data))
+        if self.prefix == MMDET_IMAGE or self.prefix.lower().startswith(MMOCR):
```
zhiqiangdon, line 404: Consider simplifying the condition: ` if self.prefix.lower().startswith(MMCV_MODELS):`.

multimodal/src/autogluon/multimodal/optimization/lit_mmdet.py
```diff
@@ -0,0 +1,332 @@
...
+        self.validation_metric_name = f"val_{validation_metric_name}"
+        if isinstance(validation_metric, BaseAggregator) and custom_metric_func is None:
+            raise ValueError(
+                f"validation_metric {validation_metric} is an aggregation metric,"
+                f"which must be used with a customized metric function."
+            )
+        self.custom_metric_func = custom_metric_func
+        self.id2label = dict(zip(range(100), range(100)))  # TODO: replace with real id2label
+
+    def forward(self, x):
```
zhiqiangdon, line 67: What's this `forward` used for?

multimodal/src/autogluon/multimodal/optimization/lit_mmdet.py
```diff
@@ -0,0 +1,332 @@
...
+
+        data = batch["mmdet_image_image"]
+        data["img_metas"] = [img_metas.data[0] for img_metas in data["img_metas"]]
+        data["img"] = [img.data[0] for img in data["img"]]
+        if next(self.model.parameters()).is_cuda:
+            # scatter to specified GPU
+            data = scatter(data, [self.device])[0]
+        else:
+            for m in self.model.modules():
+                assert not isinstance(m, RoIPool), "CPU inference with RoIPool is not supported currently."
```
zhiqiangdon, line 94: What's the benefit of moving this code from `mmdet_image` to the lightning module?

multimodal/src/autogluon/multimodal/data/preprocess_dataframe.py
FANGAreNotGnu: The column_type here for ROIS column is IMAGE. This hack will be solved in a later PR by using transform_multimodal().

multimodal/src/autogluon/multimodal/data/process_image.py
FANGAreNotGnu: In MMDet, the same pipeline is used for val and test in detection on COCO and VOC. So there is no difference.

multimodal/src/autogluon/multimodal/data/process_image.py
FANGAreNotGnu: they both start with mm, this works for now but may cause conflict in the future? or shall we rename MMDET_IMAGE and MMOCR to start with same prefix like "mmlab_detection", "mmlab_ocr"

multimodal/src/autogluon/multimodal/data/process_label.py
```diff
@@ -43,7 +44,10 @@ def collate_fn(self, label_column_names: Optional[List] = None) -> Dict:
         -------
         A dictionary containing one model's collator function for labels.
         """
-        fn = {self.label_key: Stack()}
+        if self.prefix == "mmdet_image":
```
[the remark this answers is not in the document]
FANGAreNotGnu: Changed this to MMDET, but not sure if MMOCR will also be using List to collate label.

examples/automm/object_detection/detection_train.py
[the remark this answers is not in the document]
bryanyzhu: I think it will be more like the first one using code. Let's see how users respond. If we need GUI later, let's decide at that time which GUI to use.

multimodal/src/autogluon/multimodal/optimization/lit_mmdet.py
FANGAreNotGnu: I think it's cleaner to do batch unpack and .to() in lightning module. But we have to leave a forward() with batch unpack and scatter in mmdet_image. We can discuss if we shall move all the unpack and .to() to mmdet_image. We can do this in next pr since there are a few changes in implementing multi gpu support.

.github/workflow_scripts/env_setup.sh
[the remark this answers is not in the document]
bryanyzhu: +1 to consensus. I prefer setting it to a relative lower number is better. If we get timeout error, we just run the code/CI again.

multimodal/src/autogluon/multimodal/data/preprocess_dataframe.py
zhiqiangdon: In `infer_column_types `, the rois column is detected as type `ROIS`. In lines 211 and 284, the column type is also `ROIS`. It's confusing from where the column type is changed to `IMAGE`.

multimodal/src/autogluon/multimodal/data/process_image.py
bryanyzhu: I like "mmlab_detection", "mmlab_ocr", we can discuss later.

multimodal/src/autogluon/multimodal/optimization/lit_mmdet.py
```diff
@@ -0,0 +1,317 @@
...
+        self.model = model
+        self.validation_metric = validation_metric
+        self.validation_metric_name = f"val_{validation_metric_name}"
+        if isinstance(validation_metric, BaseAggregator) and custom_metric_func is None:
+            raise ValueError(
+                f"validation_metric {validation_metric} is an aggregation metric,"
+                f"which must be used with a customized metric function."
+            )
+        self.custom_metric_func = custom_metric_func
+        self.id2label = dict(zip(range(100), range(100)))  # TODO: replace with real id2label
```
cheungdaven, line 65: why 100?

multimodal/src/autogluon/multimodal/optimization/lit_mmdet.py
```diff
@@ -0,0 +1,317 @@
...
+                )
+            )
+
+        # use MeanAveragePrecision, example code: https://github.com/Lightning-AI/metrics/blob/master/examples/detection_map.py
+        self.validation_metric.update(preds, target)
+
+        return pred_results
+
+    def compute_loss(self, img, img_metas, gt_bboxes, gt_labels, *args, **kwargs):
+        """
```
cheungdaven, line 156: comments  can be clearer

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -161,6 +162,7 @@ def __init__(
         hyperparameters: Optional[dict] = None,
         path: Optional[str] = None,
         verbosity: Optional[int] = 3,
+        output_shape: Optional[int] = None,
```
cheungdaven, line 165: do we really need to expose *output_shape* to users?

multimodal/src/autogluon/multimodal/optimization/lit_mmdet.py
FANGAreNotGnu: Removed.

multimodal/src/autogluon/multimodal/optimization/lit_mmdet.py
```diff
@@ -0,0 +1,317 @@
...
+        return res
+
+    def predict_step(self, batch, batch_idx, dataloader_idx=0):
+        pred = self._predict_step(batch, batch_idx)
+        if "mmdet_image_label" in batch:
+            return {"bbox": pred, "label": batch["mmdet_image_label"]}
+        else:
+            return {"bbox": pred}
+
+    def configure_optimizers(self):
```
zhiqiangdon, line 242: Is this `configure_optimizers ` the same as the one in `lit_module`? If so, consider reusing it by inheritance or util.

examples/automm/object_detection/detection_train.py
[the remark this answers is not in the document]
FANGAreNotGnu: Got it, I can add this tutorial first after multi gpu. Or do you think this has higher priority?

multimodal/src/autogluon/multimodal/optimization/lit_mmdet.py
FANGAreNotGnu: yes for now, but later we may want to freeze backbone and only train bboxhead, or using different lr for them.

multimodal/src/autogluon/multimodal/optimization/lit_mmdet.py
FANGAreNotGnu: Done.

multimodal/src/autogluon/multimodal/optimization/lit_mmdet.py
FANGAreNotGnu: Replaced this hardcode with a id2label in self.model.

multimodal/src/autogluon/multimodal/data/infer_types.py
```diff
@@ -380,13 +373,18 @@ def infer_column_types(
...
+            and is_training
+        ):
             column_types[col_name] = NULL
             continue
+        # TODO: valid check for collections
 
         if is_rois_column(data[col_name]):
             column_types[col_name] = ROIS
-        if is_identifier_column(data[col_name], col_name=col_name, id_mappings=id_mappings):
+        elif is_identifier_column(data[col_name], col_name=col_name, id_mappings=id_mappings):
```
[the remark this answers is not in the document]
FANGAreNotGnu: No but I think logically seems clearer

multimodal/src/autogluon/multimodal/constants.py
```diff
@@ -50,6 +50,7 @@
 SCORE = "score"
 
 # Metric
+MAP = "map"
```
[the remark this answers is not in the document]
FANGAreNotGnu: Added a MEAN_AVERAGE_PRECISION constant

multimodal/src/autogluon/multimodal/data/infer_types.py
```diff
@@ -428,10 +426,11 @@ def check_missing_values(
 def infer_label_column_type_by_problem_type(
     column_types: Dict,
     label_columns: Union[str, List[str]],
-    problem_type: str,
+    problem_type: Optional[str],
```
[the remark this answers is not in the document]
FANGAreNotGnu: changed to infer_label_column_type_by_problem_type_and_pipeline

multimodal/src/autogluon/multimodal/data/infer_types.py
```diff
@@ -484,6 +485,10 @@ def infer_label_column_type_by_problem_type(
         elif problem_type == NER:
             column_types[col_name] = NER_ANNOTATION
 
+        if problem_type is None:
```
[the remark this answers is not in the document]
FANGAreNotGnu: changed to infer_label_column_type_by_problem_type_and_pipeline

multimodal/src/autogluon/multimodal/data/preprocess_dataframe.py
FANGAreNotGnu: Added a TODO comment here. Per our offline discussion, I'll bypass/overwrite the type fallback in while detection model predicts. This change will be completed in next PR.

.github/workflow_scripts/env_setup.sh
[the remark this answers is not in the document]
sxjscience: What’s the error message? Previously the CI becomes flaky so we changed it to 1000.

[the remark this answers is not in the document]
sxjscience: How about setting it to 120? May be 1000 is too large. We can adjust the number again when we see flaky CI.

multimodal/src/autogluon/multimodal/constants.py
sxjscience, line 53: Not sure which one is better but we can consider use “mAP”.

multimodal/src/autogluon/multimodal/optimization/lit_mmdet.py
```diff
@@ -0,0 +1,321 @@
...
+        self.id2label = self.model.id2label
+
+    def _predict_step(self, batch, batch_idx=0):
+        from mmcv.ops import RoIPool
+        from mmcv.parallel import scatter
+
+        data = batch["mmdet_image_image"]
+        data["img_metas"] = [img_metas.data[0] for img_metas in data["img_metas"]]
+        data["img"] = [img.data[0] for img in data["img"]]
+        if next(self.model.parameters()).is_cuda:
```
sxjscience, line 74: Why do we need this?

multimodal/src/autogluon/multimodal/optimization/lit_mmdet.py
```diff
@@ -0,0 +1,321 @@
...
+            for m in self.model.modules():
+                assert not isinstance(m, RoIPool), "CPU inference with RoIPool is not supported currently."
+        pred_results = self.model.model(return_loss=False, rescale=True, **data)
+        return pred_results
+
+    def _training_step(self, batch, batch_idx=0):
+        batch = unpack_datacontainers(batch)
+
+        # TODO: test mmcv scatter
+        img_metas = batch["mmdet_image_image"]["img_metas"][0]
```
sxjscience, line 87: Regarding the name here. Will “mmdet_image” be better than “mmdet_image_image”

examples/automm/object_detection/detection_train.py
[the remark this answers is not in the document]
bryanyzhu: Not high priority, as long as it is done before release 0.6.

multimodal/src/autogluon/multimodal/constants.py
bryanyzhu: Yes, mAP is more accurate.

multimodal/src/autogluon/multimodal/optimization/lit_mmdet.py
bryanyzhu: `mmdet_image_image` seems duplicated, why we need two image here?

multimodal/src/autogluon/multimodal/optimization/lit_mmdet.py
FANGAreNotGnu: The output of collate function is a DataContainer object(s), we need to unpack and send to gpu here.

multimodal/src/autogluon/multimodal/optimization/lit_mmdet.py
FANGAreNotGnu: The first "image" is in the model name "mmdet_image", and second "image" is appended by image_processor. We could consider changing the model name from mmdet_image to mmdet.

FANGAreNotGnu: mmdet may not be a good naming as well since it duplicates with our constant MMDET. Also there are instance/semantic/panoptic segmentation in mmdet repo we may want to implement in another model named like mmdet_segmentation.

FANGAreNotGnu: I'll remove the hardcode to make it look better.

multimodal/src/autogluon/multimodal/data/process_image.py
FANGAreNotGnu: Changed the constant MMCV_MODELS to MMLAB_MODELS and updated the logic here

.github/workflow_scripts/env_setup.sh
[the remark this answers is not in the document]
FANGAreNotGnu: Reached an agreement to set this to 60 in offline meeting.

multimodal/src/autogluon/multimodal/optimization/lit_mmdet.py
sxjscience: Okay, will data be on CPU if we do not do this? I think this seems like a problem associated with our collate function / the way of integrating the collate function with lightning.

bryanyzhu (approved):
LGTM, thank you. I have a minor question regarding naming of `output_shape`, but other than that, all looks great.

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -477,17 +485,20 @@ def fit(
                 stratify=stratify,
                 random_state=np.random.RandomState(seed),
             )
+            train_data = train_data.reset_index(drop=True)
+            tuning_data = tuning_data.reset_index(drop=True)
```
zhiqiangdon, line 489: 1. How about resetting index only for object detection? It seems that only object detection requires this.
2. Consider investigating the reason why we need this, which is necessary to understand our system and avoid fusion potential bugs.

zhiqiangdon (approved):
LGTM. Thanks! 

1. Left a comment regarding resetting dataframe indexes. Maybe we can reset the index only for object detection for now and continue looking into the reasons behind it. 
2. We may need to refactor the column name check in `transform_image()` ASAP since it can affect other tasks.
3. Reconsider if it's necessary to double the ROIS columns in dataframe. Though necessary for now, it seems some redundant to do so. Maybe we can refer to MMDET's training code to see how they handle transforming rois with images and computing metric in the original scale.

sxjscience (approved):
LGTM and we can merge this PR for now. But we should check our collate function and potentially fix it in another PR.

multimodal/src/autogluon/multimodal/optimization/lit_mmdet.py
FANGAreNotGnu: Yes it will be on CPU. I'll check if we can move the unpack codes to collate function. Then it will be better integrated with lightning.

multimodal/src/autogluon/multimodal/predictor.py
FANGAreNotGnu: Done, also marked a todo to investigate why we need this and remove it.

PR #2219, opened 2022-10-17
Fix timm

zhiqiangdon:
*Issue #, if available:*

*Description of changes:*

1. Make timm caps consistent between core and multimodal. Temporarily keep the timm cap in core since the vision module also needs to install timm. We can remove it after refactoring the vision module.
2. Set the default num_classes=0 and compute logits only when `num_classes>0`. This is to handle some special heads in timm, e.g., https://github.com/rwightman/pytorch-image-models/blob/8a1ba67a5019cac4e0e6c430a846e5653f636c3f/timm/models/layers/classifier.py#L38

### DIFF
--- a/core/src/autogluon/core/_setup_utils.py
+++ b/core/src/autogluon/core/_setup_utils.py
@@ -24,7 +24,6 @@ DEPENDENT_PACKAGES = {
     'gluoncv': '>=0.10.5,<0.10.6',
     'tqdm': '>=4.38.0',
     'Pillow': '>=9.0.1,<9.1.0',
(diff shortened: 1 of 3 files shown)


FANGAreNotGnu (approved):
LGTM

bryanyzhu (approved):
Thanks for the fix, LGTM.

suzhoum (approved):
LGTM!

yinweisu (approved):
LGTM!

sxjscience (approved):
LGTM

PR #2225, opened 2022-10-19
[Tabular] support random forest model prediction with onnxrunntime

liangfu:
*Description of changes:*
This PR brings random forest model prediction with onnxruntime. skl2onnx is used for converting sklearn random forest model to onnx.

In order to evaluate prediction time improvements, we have performed benchmark based on 10 datasets. The timing results are shown in the following table:

```
     name                task_name  test_score  time_fit  time_predict eval_metric  test_error
RF-native                     wilt    0.990679  0.735087      0.110379     roc_auc    0.009321
RF-native                 credit-g    0.865238  0.798425      0.122325     roc_auc    0.134762
RF-native                  jasmine    0.869732  1.046427      0.174290     roc_auc    0.130268
RF-native                 madeline    0.848690  1.176366      0.112416     roc_auc    0.151310
RF-native               eucalyptus   -0.702908  2.404443      0.118938    log_loss    0.702908
RF-native              qsar-biodeg    0.950992  0.761442      0.112315     roc_auc    0.049008
RF-native                      pc4    0.915365  0.755036      0.111635     roc_auc    0.084635
RF-native                      kc1    0.843925  0.757150      0.111022     roc_auc    0.156075
RF-native                  segment   -0.166335  2.091211      0.111191    log_loss    0.166335
RF-native  Internet-Advertisements    0.987936  5.353352      1.467787     roc_auc    0.012064
  RF-onnx                     wilt    0.990721  2.255235      0.006050     roc_auc    0.009279
  RF-onnx                 credit-g    0.864762  3.569216      0.017000     roc_auc    0.135238
  RF-onnx                  jasmine    0.869664  8.856808      0.070683     roc_auc    0.130336
  RF-onnx                 madeline    0.848710  6.630268      0.007998     roc_auc    0.151290
  RF-onnx               eucalyptus   -0.702908  6.946166      0.013174    log_loss    0.702908
  RF-onnx              qsar-biodeg    0.951389  2.333043      0.006884     roc_auc    0.048611
  RF-onnx                      pc4    0.915365  1.984404      0.006151     roc_auc    0.084635
  RF-onnx                      kc1    0.843837  4.133331      0.005890     roc_auc    0.156163
  RF-onnx                  segment   -0.166314  7.007544      0.006683    log_loss    0.166314
  RF-onnx  Internet-Advertisements    0.987820  9.794042      1.357964     roc_auc    0.012180
```

Geometric mean of the speed up can be calculated with

```
>>> gmean(native['time_predict'].to_numpy() / onnx['time_predict'].to_numpy())
9.283319349283909
```

This indicate that we are about 9x faster than the native implementation, which is available in the scikit-learn package.

image

Note that it takes a few seconds to convert the random forest model to onnx format. This could be about 10x the time for fitting the model. This is currently slow because of the `add_node` overhead in skl2onnx package.

### DIFF
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -867,7 +867,12 @@ class AbstractModel:
             transform_func=transform_func, transform_func_kwargs=transform_func_kwargs, silent=silent, **kwargs
         )
 
-    # TODO: Pick up from here, rename to save, implement _get_compiler(), etc.
+    # TODO: PoC, refactor prior to v0.6 release
+    #  Add Documentation
+    #  Define when this should be called
+    #  Is it called during fit? Is it called after fit?
+    #  Can it be called as a post-fit operation on an already fit predictor on a per-model basis?
+    #  Does this call overwrite the existing model or make a new one?
     def compile(self, path: str = None, verbose=True) -> str:
         if path is None:
             path = self.path
@@ -886,8 +891,25 @@ class AbstractModel:
         self.model = _model
         return path
 
-    def _get_compiler(self):
+    def _default_compiler(self):
         return None
+
+    def _get_compiler(self):
+        compiler = self.params_aux.get('compiler', None)
+        compilers = self._valid_compilers()
+        compiler_names = {c.name: c for c in compilers}
+        if compiler is not None and compiler not in compiler_names:
+            raise AssertionError(f'Unknown compiler: {compiler}. Valid compilers: {compiler_names}')
+        if compiler is None:
+            return self._default_compiler()
+        compiler_cls = compiler_names[compiler]
+        compiler_fallback_to_native = self.params_aux.get('compiler_fallback_to_native', True)
+        if not compiler_cls.can_compile():
+            if not compiler_fallback_to_native:
+                raise AssertionError(f'Specified compiler ({compiler}) is unable to compile'
+                                     ' (potentially lacking dependencies) and "compiler_fallback_to_native==False"')
+            compiler_cls = self._default_compiler()
+        return compiler_cls
 
     def get_trained_params(self) -> dict:
         """
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -866,6 +866,25 @@ class AbstractModel:
             transform_func=transform_func, transform_func_kwargs=transform_func_kwargs, silent=silent, **kwargs
         )
 
+    # TODO: Pick up from here, rename to save, implement _get_compiler(), etc.
+    def compile(self, path: str = None, verbose=True) -> str:
+        if path is None:
+            path = self.path
+        file_path = path + self.model_file_name
+
+        save_in_pkl = True
+        if self.model is not None:
+            self._compiler = self._get_compiler()
+            if self._compiler is not None:
+                self.model = self._compiler.compile(obj=self, path=path)
+                save_in_pkl = self._compiler.save_in_pkl
+        _model = self.model
+        if not save_in_pkl:
+            self.model = None
+        save_pkl.save(path=file_path, object=self, verbose=verbose)
+        self.model = _model
+        return path
+
     def get_trained_params(self) -> dict:
         """
         Returns the hyperparameters of the trained model.
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -250,6 +250,7 @@ class AbstractModel:
             get_features_kwargs_extra=None,  # If not None, applies an additional feature filter to the result of get_feature_kwargs. This should be reserved for users and be None by default. | Currently undocumented in task.
             predict_1_batch_size=None,  # If not None, calculates `self.predict_1_time` at end of fit call by predicting on this many rows of data.
             temperature_scalar=None,  # Temperature scaling parameter that is set post-fit if calibrate=True during TabularPredictor.fit() on the model with the best validation score and eval_metric="log_loss".
+            compiler='native', # The compiler backend that is used for prediction. This defaults to 'native' backend for all models. A list of supported compilers can be found via calling _valid_compilers.
         )
         return default_auxiliary_params
 
@@ -884,6 +885,9 @@ class AbstractModel:
         save_pkl.save(path=file_path, object=self, verbose=verbose)
         self.model = _model
         return path
+
+    def _get_compiler(self):
+        return None
 
     def get_trained_params(self) -> dict:
         """
--- a/tabular/setup.py
+++ b/tabular/setup.py
@@ -54,7 +54,7 @@ extras_require = {
         'vowpalwabbit>=8.10,<8.11'
     ],
     'skl2onnx': [
-        'skl2onnx>=1.7.0,<1.13.0'
+        'skl2onnx>=1.12.0,<1.13.0'
     ]
 }
 
--- a/tabular/setup.py
+++ b/tabular/setup.py
@@ -52,6 +52,9 @@ extras_require = {
     ],
     'vowpalwabbit': [
         'vowpalwabbit>=8.10,<8.11'
+    ],
+    'skl2onnx': [
+        'skl2onnx>=1.7.0,<1.13.0'
     ]
 }
 
@@ -64,7 +67,7 @@ extras_require['all'] = all_requires
 
 
 test_requires = []
-for test_package in ['imodels', 'vowpalwabbit']:
+for test_package in ['imodels', 'vowpalwabbit', 'skl2onnx']:
     test_requires += extras_require[test_package]
 extras_require['tests'] = test_requires
 install_requires = ag.get_dependency_version_ranges(install_requires)
--- a/tabular/src/autogluon/tabular/models/rf/rf_model.py
+++ b/tabular/src/autogluon/tabular/models/rf/rf_model.py
@@ -230,6 +230,7 @@ class RFModel(AbstractModel):
             model.estimators_ = None
         self.model = model
         self.params_trained['n_estimators'] = self.model.n_estimators
+        self.compile()
 
     # TODO: Remove this after simplifying _predict_proba to reduce code duplication. This is only present for SOFTCLASS support.
     def _predict_proba(self, X, **kwargs):
--- a/tabular/src/autogluon/tabular/models/rf/rf_model.py
+++ b/tabular/src/autogluon/tabular/models/rf/rf_model.py
@@ -19,7 +19,6 @@ from autogluon.features.generators import LabelEncoderFeatureGenerator
 logger = logging.getLogger(__name__)
 
 
-# FIXME: DOESNT WORK ON REGRESSION PREDICTIONS
 class RFOnnxPredictor:
     def __init__(self, obj, model):
         self.num_classes = obj.num_classes
@@ -28,7 +27,7 @@ class RFOnnxPredictor:
     def predict(self, X):
         input_name = self.model.get_inputs()[0].name
         label_name = self.model.get_outputs()[0].name
-        return self.model.run([label_name], {input_name: X})[0]
+        return self.model.run([label_name], {input_name: X})[0].squeeze()
 
     def predict_proba(self, X):
         input_name = self.model.get_inputs()[0].name
@@ -54,21 +53,17 @@ class RFNativeCompiler:
         if isinstance(obj.model, (RandomForestClassifier, RandomForestRegressor,
                                   ExtraTreesClassifier, ExtraTreesRegressor)):
             with open(path + 'model_native.pkl', 'wb') as fp:
-                fp.write(pickle.dumps(obj))
+                fp.write(pickle.dumps(obj.model))
+            return obj.model
         else:
-            pkl = None
-            with open(path + 'model_native.pkl', 'rb') as fp:
-                pkl = fp.read()
-            return pickle.loads(pkl).model
-        return obj.model
+            return RFNativeCompiler.load(obj=obj, path=path)
 
     @staticmethod
     def load(obj, path: str):
         pkl = None
         with open(path + 'model_native.pkl', 'rb') as fp:
             pkl = fp.read()
-        model = pickle.loads(pkl)
-        return model.model
+        return pickle.loads(pkl)
 
 
 class InferenceSessionWrapper:
@@ -78,8 +73,8 @@ class InferenceSessionWrapper:
     """
     def __init__(self, onnx_bytes):
         import onnxruntime as rt
-        self.sess = rt.InferenceSession(onnx_bytes.SerializeToString())
-        self.onnx_bytes = onnx_bytes
+        self.sess = rt.InferenceSession(onnx_bytes.SerializeToString(),
+                                        providers=['CPUExecutionProvider'])
 
     def run(self, *args):
         return self.sess.run(*args)
@@ -91,13 +86,11 @@ class InferenceSessionWrapper:
         return self.sess.get_outputs(*args)
 
     def __getstate__(self):
-        return {'onnx_bytes': self.onnx_bytes}
+        # No need to duplicate the model parameters here.
+        return {}
 
     def __setstate__(self, values):
-        import onnxruntime as rt
-        self.onnx_bytes = values['onnx_bytes']
-        self.sess = rt.InferenceSession(onnx_bytes.SerializeToString())
-
+        pass
 
 class RFOnnxCompiler:
     name = 'onnx'
@@ -115,7 +108,6 @@ class RFOnnxCompiler:
     def compile(obj, path: str):
         if isinstance(obj.model, RFOnnxPredictor):
             return obj.model
-        print('compiling')
         # Convert into ONNX format
         from skl2onnx import convert_sklearn
         from skl2onnx.common.data_types import FloatTensorType
@@ -490,7 +482,11 @@ class RFModel(AbstractModel):
         if compiler is None:
             return RFNativeCompiler
         compiler_cls = compiler_names[compiler]
+        compiler_fallback = self.params_aux.get('compiler_fallback_to_native', True)
         if not compiler_cls.can_compile():
+            if not compiler_fallback:
+                raise AssertionError(f'Specified compiler ({compiler}) is unable to '
+                                     'compile (potentially lacking dependencies) and "compiler_fallback==False"')
             compiler_cls = RFNativeCompiler
         return compiler_cls
 
--- a/tabular/src/autogluon/tabular/models/rf/rf_model.py
+++ b/tabular/src/autogluon/tabular/models/rf/rf_model.py
@@ -230,6 +230,7 @@ class RFModel(AbstractModel):
             model.estimators_ = None
         self.model = model
         self.params_trained['n_estimators'] = self.model.n_estimators
+        # FIXME: v0.6 Move to generic compile call, ensure compile time is included in fit time
         self.compile()
 
     # TODO: Remove this after simplifying _predict_proba to reduce code duplication. This is only present for SOFTCLASS support.
--- a/tabular/src/autogluon/tabular/models/rf/rf_model.py
+++ b/tabular/src/autogluon/tabular/models/rf/rf_model.py
@@ -37,6 +37,66 @@ class RFOnnxPredictor:
         return pred_proba
 
 
+class RFNativeCompiler:
+    name = 'native'
+    save_in_pkl = True
+
+    @staticmethod
+    def can_compile():
+        return True
+
+    @staticmethod
+    def compile(obj, path: str):
+        from sklearn.ensemble import RandomForestClassifier, RandomForestRegressor
+        from sklearn.ensemble import ExtraTreesRegressor, ExtraTreesClassifier
+        if isinstance(obj.model, (RandomForestClassifier, RandomForestRegressor,
+                                  ExtraTreesClassifier, ExtraTreesRegressor)):
+            with open(path + 'model_native.pkl', 'wb') as fp:
+                fp.write(pickle.dumps(obj))
+        else:
+            pkl = None
+            with open(path + 'model_native.pkl', 'rb') as fp:
+                pkl = fp.read()
+            return pickle.loads(pkl).model
+        return obj.model
+
+    @staticmethod
+    def load(obj, path: str):
+        pkl = None
+        with open(path + 'model_native.pkl', 'rb') as fp:
+            pkl = fp.read()
+        model = pickle.loads(pkl)
+        return model.model
+
+
+class InferenceSessionWrapper:
+    """
+    Wrap around InferenceSession in onnxruntime, since it cannot be pickled.
+    See https://github.com/microsoft/onnxruntime/issues/10097
+    """
+    def __init__(self, onnx_bytes):
+        import onnxruntime as rt
+        self.sess = rt.InferenceSession(onnx_bytes.SerializeToString())
+        self.onnx_bytes = onnx_bytes
+
+    def run(self, *args):
+        return self.sess.run(*args)
+
+    def get_inputs(self, *args):
+        return self.sess.get_inputs(*args)
+
+    def get_outputs(self, *args):
+        return self.sess.get_outputs(*args)
+
+    def __getstate__(self):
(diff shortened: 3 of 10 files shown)


tabular/src/autogluon/tabular/models/rf/rf_model.py
```diff
@@ -18,6 +19,121 @@
 logger = logging.getLogger(__name__)
 
 
+# FIXME: DOESNT WORK ON REGRESSION PREDICTIONS
```
Innixma, line 22: Can we make it work for regression?

tabular/src/autogluon/tabular/models/rf/rf_model.py
```diff
@@ -358,3 +477,39 @@ def _more_tags(self):
...
+        compiler = self.params_aux.get('compiler', None)
+        compilers = self._valid_compilers()
+        compiler_names = {c.name: c for c in compilers}
+        if compiler is not None and compiler not in compiler_names:
+            raise AssertionError(f'Unknown compiler: {compiler}. Valid compilers: {compiler_names}')
+        if compiler is None:
+            return RFNativeCompiler
+        compiler_cls = compiler_names[compiler]
+        if not compiler_cls.can_compile():
+            compiler_cls = RFNativeCompiler
```
Innixma, line 494: Add logic here so we can test to ensure that the compiler actually works in unit tests:

```
compiler_fallback = self.params_aux.get('compiler_fallback_to_native', True)
if not compiler_cls.can_compile():
    if not compiler_fallback:
         raise AssertionError(f'Specified compiler ({compiler}) is unable to compile (potentially lacking dependencies) and "compiler_fallback==False"')
    compiler_cls = RFNativeCompiler
```

Then change unit tests to specify `compiler_fallback_to_native=False`

tabular/src/autogluon/tabular/models/rf/rf_model.py
```diff
@@ -18,6 +19,121 @@
...
+        os.makedirs(os.path.dirname(path), exist_ok=True)
+        if isinstance(obj.model, (RandomForestClassifier, RandomForestRegressor,
+                                  ExtraTreesClassifier, ExtraTreesRegressor)):
+            with open(path + 'model_native.pkl', 'wb') as fp:
+                fp.write(pickle.dumps(obj))
+        else:
+            pkl = None
+            with open(path + 'model_native.pkl', 'rb') as fp:
+                pkl = fp.read()
+            return pickle.loads(pkl).model
```
Innixma, line 62: replace with `return RFNativeCompiler.load(obj=obj, path=path)`

tabular/src/autogluon/tabular/models/rf/rf_model.py
```diff
@@ -18,6 +19,121 @@
...
+        return obj.model
```
Innixma, line 63: Move to the end of the `if` statement right before `else`

tabular/src/autogluon/tabular/models/rf/rf_model.py
```diff
@@ -18,6 +19,121 @@
...
+
+    @staticmethod
+    def compile(obj, path: str):
+        from sklearn.ensemble import RandomForestClassifier, RandomForestRegressor
+        from sklearn.ensemble import ExtraTreesRegressor, ExtraTreesClassifier
+        os.makedirs(os.path.dirname(path), exist_ok=True)
+        if isinstance(obj.model, (RandomForestClassifier, RandomForestRegressor,
+                                  ExtraTreesClassifier, ExtraTreesRegressor)):
+            with open(path + 'model_native.pkl', 'wb') as fp:
+                fp.write(pickle.dumps(obj))
```
Innixma, line 57: Why are we pickling obj? Shouldn't we be pickling `obj.model`?

tabular/src/autogluon/tabular/models/rf/rf_model.py
```diff
@@ -18,6 +19,121 @@
...
+            return pickle.loads(pkl).model
+        return obj.model
+
+    @staticmethod
+    def load(obj, path: str):
+        pkl = None
+        with open(path + 'model_native.pkl', 'rb') as fp:
+            pkl = fp.read()
+        model = pickle.loads(pkl)
+        return model.model
```
Innixma, line 71: Why are we returning `model.model`?? (See comment on how we should be pickling `obj.model`).

tabular/src/autogluon/tabular/models/rf/rf_model.py
```diff
@@ -18,6 +19,121 @@
...
+
+class InferenceSessionWrapper:
+    """
+    Wrap around InferenceSession in onnxruntime, since it cannot be pickled.
+    See https://github.com/microsoft/onnxruntime/issues/10097
+    """
+    def __init__(self, onnx_bytes):
+        import onnxruntime as rt
+        self.sess = rt.InferenceSession(onnx_bytes.SerializeToString())
+        self.onnx_bytes = onnx_bytes
```
Innixma, line 82: Is this doubling our memory usage by keeping this in memory after creating the session? If it is, we probably don't want to store the variable.

tabular/tests/conftest.py
```diff
@@ -235,6 +235,46 @@ def fit_dataset(train_data, model, label, fit_args, sample_size=None):
...
+        assert len(model_names) == expected_model_count
+        predictor.info()
+        results = predictor.leaderboard(test_data, extra_info=False, extra_metrics=extra_metrics)
+        with open(save_path + "results.csv", "w") as fp:
+            fp.write(results.to_csv(index=False))
+
+        assert os.path.realpath(save_path) == os.path.realpath(predictor.path)
+        if delete_directory:
+            shutil.rmtree(save_path, ignore_errors=True)  # Delete AutoGluon output directory to ensure runs' information has been removed.
+        return predictor
```
Innixma, line 275: Why??

This looks like a copy paste. I don't understand why this is needed.

tabular/tests/unittests/models/test_rf.py
```diff
@@ -25,3 +25,14 @@ def test_rf_regression(fit_helper):
     )
     dataset_name = 'ames'
     fit_helper.fit_and_validate_dataset(dataset_name=dataset_name, fit_args=fit_args)
+
+
+def test_rf_compile_binary(compile_helper):
```
Innixma, line 30: This isn't actually compiling to ONNX since skl2onnx isn't installed. It is simply failing to import onnx and using native. Need to add the onnx dependency to tabular's setup.py `tests` extra dependencies. https://github.com/awslabs/autogluon/blob/master/tabular/setup.py#L67-L69

tabular/tests/unittests/models/test_xt.py
```diff
@@ -24,3 +24,14 @@ def test_xt_regression(fit_helper):
     )
     dataset_name = 'ames'
     fit_helper.fit_and_validate_dataset(dataset_name=dataset_name, fit_args=fit_args)
+
+
+def test_xt_compile_binary(compile_helper):
```
Innixma, line 29: ditto, this isn't actually compiling to ONNX due to missing dependency.

tabular/src/autogluon/tabular/models/rf/rf_model.py
```diff
@@ -18,6 +19,121 @@
...
+            import skl2onnx
+            return True
+        except ImportError:
+            return False
+
+    @staticmethod
+    def compile(obj, path: str):
+        if isinstance(obj.model, RFOnnxPredictor):
+            return obj.model
+        print('compiling')
```
Innixma, line 118: Remove print statement

tabular/src/autogluon/tabular/models/rf/rf_model.py
liangfu: Yes, it works for regression tasks now. I found that we need to, somehow, squeeze the onnx output.

tabular/src/autogluon/tabular/models/rf/rf_model.py
liangfu: As suggested, this is now modified into
```python
        if isinstance(obj.model, (RandomForestClassifier, RandomForestRegressor,
                                  ExtraTreesClassifier, ExtraTreesRegressor)):
            with open(path + 'model_native.pkl', 'wb') as fp:
                fp.write(pickle.dumps(obj.model))
            return obj.model
        else:
            return RFNativeCompiler.load(obj=obj, path=path)
```

tabular/src/autogluon/tabular/models/rf/rf_model.py
liangfu: As suggested, this is now modified into
```python
        if isinstance(obj.model, (RandomForestClassifier, RandomForestRegressor,
                                  ExtraTreesClassifier, ExtraTreesRegressor)):
            with open(path + 'model_native.pkl', 'wb') as fp:
                fp.write(pickle.dumps(obj.model))
            return obj.model
        else:
            return RFNativeCompiler.load(obj=obj, path=path)
```

tabular/src/autogluon/tabular/models/rf/rf_model.py
liangfu: Now we only pickle `obj.model` into 'model_native.pkl`.

tabular/src/autogluon/tabular/models/rf/rf_model.py
liangfu: As commented above, this is now revised to pickle `obj.model` into 'model_native.pkl`, instead of the object itself.

tabular/src/autogluon/tabular/models/rf/rf_model.py
liangfu: I tested in my local env that we actually don't need to duplicate the onnx parameters in the inference session class. This extra copy has been removed.

tabular/src/autogluon/tabular/models/rf/rf_model.py
liangfu: Removed.

tabular/tests/unittests/models/test_rf.py
liangfu: Added `skl2onnx` into extra dependencies.

NOTE: When installing skl2onnx, it would also introduce `onnxruntime-gpu` package. We are now explicitly taking CPU provider in the onnx inference session.

tabular/tests/unittests/models/test_xt.py
liangfu: Added skl2onx into extra dependency list. Tested with a clean venv.

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -866,6 +867,28 @@ def _compute_permutation_importance(self, X, y, features: list, eval_metric=None
...
+            self._compiler = self._get_compiler()
+            if self._compiler is not None:
+                self.model = self._compiler.compile(obj=self, path=path)
+                save_in_pkl = self._compiler.save_in_pkl
+        _model = self.model
+        if not save_in_pkl:
+            self.model = None
+        save_pkl.save(path=file_path, object=self, verbose=verbose)
+        self.model = _model
+        return path
```
Innixma, line 887: This should be marked as 

```
# TODO: PoC, refactor prior to v0.6 release
#  Add Documentation
#  Define when this should be called
#  Is it called during fit? Is it called after fit?
#  Can it be called as a post-fit operation on an already fit predictor on a per-model basis?
#  Does this call overwrite the existing model or make a new one?
```

tabular/src/autogluon/tabular/models/rf/rf_model.py
```diff
@@ -358,3 +469,43 @@ def _more_tags(self):
...
+        if compiler is not None and compiler not in compiler_names:
+            raise AssertionError(f'Unknown compiler: {compiler}. Valid compilers: {compiler_names}')
+        if compiler is None:
+            return RFNativeCompiler
+        compiler_cls = compiler_names[compiler]
+        compiler_fallback = self.params_aux.get('compiler_fallback_to_native', True)
+        if not compiler_cls.can_compile():
+            if not compiler_fallback:
+                raise AssertionError(f'Specified compiler ({compiler}) is unable to '
+                                     'compile (potentially lacking dependencies) and "compiler_fallback==False"')
```
Innixma, line 489: `compiler_fallback_to_native`

tabular/src/autogluon/tabular/models/rf/rf_model.py
```diff
@@ -358,3 +469,43 @@ def _more_tags(self):
...
+    def _get_compiler(self):
+        compiler = self.params_aux.get('compiler', None)
+        compilers = self._valid_compilers()
+        compiler_names = {c.name: c for c in compilers}
+        if compiler is not None and compiler not in compiler_names:
+            raise AssertionError(f'Unknown compiler: {compiler}. Valid compilers: {compiler_names}')
+        if compiler is None:
+            return RFNativeCompiler
+        compiler_cls = compiler_names[compiler]
+        compiler_fallback = self.params_aux.get('compiler_fallback_to_native', True)
```
Innixma, line 485: Can renamed to `compiler_fallback_to_native` for the variable to avoid any confusion

tabular/src/autogluon/tabular/models/rf/rf_model.py
```diff
@@ -358,3 +469,43 @@ def _more_tags(self):
         else:
             tags['valid_oof'] = True
         return tags
+
+    def _valid_compilers(self):
+        return [RFNativeCompiler, RFOnnxCompiler]
+
+    def _get_compiler(self):
```
Innixma, line 476: Move to AbstractModel. The native compiler should be based on a new `_default_compiler(self):` method, default returns None in AbstractModel

tabular/src/autogluon/tabular/models/rf/rf_model.py
```diff
@@ -224,6 +334,7 @@ def _fit(self,
             model.estimators_ = None
         self.model = model
         self.params_trained['n_estimators'] = self.model.n_estimators
+        self.compile(path=self.path)
```
Innixma, line 337: Are we compiling twice?? Once at end of fit and once during save? Why?

Innixma (commented):
Added second review, once addressed should be close to mergable. Follow-up cleanup of `.compile` can be in a separate PR.

tabular/setup.py
```diff
@@ -52,6 +52,9 @@
     ],
     'vowpalwabbit': [
         'vowpalwabbit>=8.10,<8.11'
+    ],
+    'skl2onnx': [
+        'skl2onnx>=1.7.0,<1.13.0'
```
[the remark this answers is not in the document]
liangfu: I'm not quite sure about the lower bound. I've narrowed down the range to `skl2onnx>=1.12.0,<1.13.0`.

tabular/src/autogluon/tabular/models/rf/rf_model.py
```diff
@@ -18,6 +19,113 @@
...
+        with open(path + "model.onnx", "wb") as f:
+            f.write(onx.SerializeToString())
+        return RFOnnxCompiler.load(obj=obj, path=path)
+
+    @staticmethod
+    def load(obj, path: str):
+        import onnx
+        onnx_bytes = onnx.load(path + "model.onnx")
+        model = InferenceSessionWrapper(onnx_bytes)
+        return RFOnnxPredictor(obj=obj, model=model)
```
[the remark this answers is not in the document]
liangfu: Make sense for me. These are separated into `compilers/native.py` and `compilers/onnx.py` source files. Only the supported compilers are imported into `rf_model.py`.

core/src/autogluon/core/models/abstract/abstract_model.py
liangfu: Added these lines on top of the function.

tabular/src/autogluon/tabular/models/rf/rf_model.py
liangfu: Renamed to `compiler_fallback_to_native` for clarity.

tabular/src/autogluon/tabular/models/rf/rf_model.py
liangfu: Make sense. This has been renamed to `compiler_fallback_to_native` to avoid confusion.

tabular/src/autogluon/tabular/models/rf/rf_model.py
liangfu: Following code has been added to the AbstractModel:
```python
def _default_compiler(self):
  return None
```
In RFModel, this returns RFNativeCompiler class.

tabular/src/autogluon/tabular/models/rf/rf_model.py
liangfu: Removed this entry here. 

NOTE: Actually the second trigger to the compile function would simply return the precompiled predictor.

tabular/tests/unittests/models/test_xt.py
```diff
@@ -24,3 +24,36 @@ def test_xt_regression(fit_helper):
...
+
+def test_xt_compile_regression(fit_helper):
+    fit_args = dict(
+        hyperparameters={XTModel: [
+            {}, # defaults to native compiler
+            {'ag_args_fit': {'compiler': "onnx", 'compiler_fallback_to_native': False}},
+        ]},
+    )
+    dataset_name = 'ames'
+    fit_helper.fit_and_validate_dataset(dataset_name=dataset_name, fit_args=fit_args, expected_model_count=3)
```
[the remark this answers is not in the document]
liangfu: Added `_compile_onnx` as suffix to the functions here.

tabular/tests/unittests/models/test_rf.py
```diff
@@ -25,3 +25,36 @@ def test_rf_regression(fit_helper):
...
+
+def test_rf_compile_regression(fit_helper):
+    fit_args = dict(
+        hyperparameters={RFModel: [
+            {}, # defaults to native compiler
+            {'ag_args_fit': {'compiler': "onnx", 'compiler_fallback_to_native': False}},
+        ]},
+    )
+    dataset_name = 'ames'
+    fit_helper.fit_and_validate_dataset(dataset_name=dataset_name, fit_args=fit_args, expected_model_count=3)
```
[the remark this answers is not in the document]
liangfu: Added _compile_onnx as suffix to the functions here.

liangfu:
Updated. Feel free to leave more comments.

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -866,6 +867,50 @@ def _compute_permutation_importance(self, X, y, features: list, eval_metric=None
             transform_func=transform_func, transform_func_kwargs=transform_func_kwargs, silent=silent, **kwargs
         )
 
+    # TODO: PoC, refactor prior to v0.6 release
+    #  Add Documentation
+    #  Define when this should be called
+    #  Is it called during fit? Is it called after fit?
+    #  Can it be called as a post-fit operation on an already fit predictor on a per-model basis?
+    #  Does this call overwrite the existing model or make a new one?
+    def compile(self, path: str = None, verbose=True) -> str:
```
Innixma, line 876: Let's add back the `compile` call at the end of RF fit for now, with a `# FIXME: v0.6 Move to generic compile call, ensure compile time is included in fit time`. This way the compile time is incorporated into the fit time instead of being hidden.

Innixma (approved):
Looks awesome!

Here is a benchmark I ran on my laptop:

```python3
import pandas as pd

from autogluon.tabular import TabularPredictor, TabularDataset
from autogluon.core.utils.infer_utils import get_model_true_infer_speed_per_row_batch


def get_model_true_infer_speed_per_row_batch_bulk(data, predictor, batch_sizes: list = None, repeats=1) -> (pd.DataFrame, pd.DataFrame):
    if batch_sizes is None:
        batch_sizes = [
            1,
            10,
            100,
            1000,
            10000,
            100000,
        ]
    infer_dfs = dict()
    infer_transform_dfs = dict()
    for batch_size in batch_sizes:
        infer_df, time_per_row_transform = get_model_true_infer_speed_per_row_batch(data=data, predictor=predictor, batch_size=batch_size, repeats=repeats)
        infer_dfs[batch_size] = infer_df
        infer_transform_dfs[batch_size] = time_per_row_transform
    for key in infer_dfs.keys():
        infer_dfs[key] = infer_dfs[key].reset_index()
        infer_dfs[key]['batch_size'] = key

    infer_df_full_transform = pd.Series(infer_transform_dfs, name='pred_time_test').to_frame().rename_axis('batch_size')
    infer_df_full_transform['pred_time_test_marginal'] = infer_df_full_transform['pred_time_test']
    infer_df_full_transform['pred_time_test_with_transform'] = infer_df_full_transform['pred_time_test']
    infer_df_full_transform = infer_df_full_transform.reset_index()

    infer_df_full = pd.concat([infer_dfs[key] for key in infer_dfs.keys()])

    return infer_df_full, infer_df_full_transform


if __name__ == '__main__':
    subsample_size = 10000  # subsample subset of data for faster demo, try setting this to much larger values
    if subsample_size is not None and subsample_size < len(train_data):
    train_data.head()
    label = 'class'

    hyperparameters = {
        'RF': [
            {},  # defaults to native compiler
            {
                'ag_args_fit': {'compiler': "onnx", 'compiler_fallback_to_native': False},
                'ag_args': {'name_suffix': '_ONNX'},
            },
        ],
        'XT': [
            {},  # defaults to native compiler
            {
                'ag_args_fit': {'compiler': "onnx", 'compiler_fallback_to_native': False},
                'ag_args': {'name_suffix': '_ONNX'},
            },
        ],
        # 'CAT': {},
        # 'GBM': {},
        # 'XGB': {},
        # 'FASTAI': {},
        # 'NN_TORCH': {},
        # 'KNN': {},
    }

    predictor = TabularPredictor(
        label=label,
    )

    predictor.fit(
        train_data,
        # presets='high',
        fit_weighted_ensemble=False,
    )

    test_data = TabularDataset('https://autogluon.s3.amazonaws.com/datasets/Inc/test.csv')

    predictor.persist_models('all')
    leaderboard = predictor.leaderboard(test_data)

    repeats = 2
    batch_sizes = [
        1,
        10,
        100,
        1000,
        10000,
        100000,
        1000000,
    ]

    infer_df_full, infer_df_full_transform = get_model_true_infer_speed_per_row_batch_bulk(
        data=test_data,
        predictor=predictor,
        batch_sizes=batch_sizes,
        repeats=repeats,
    )

    infer_df_full['rows_per_second'] = 1 / infer_df_full['pred_time_test']
    infer_df_full_transform['rows_per_second'] = 1 / infer_df_full_transform['pred_time_test']

    fig, ax = plt.subplots()
    fig.set_size_inches(12, 12)
    fig.set_dpi(300)

    plt.xscale('log')
    plt.yscale('log')

    models = list(infer_df_full['model'].unique())
    batch_sizes = list(infer_df_full['batch_size'].unique())
    for model in models:
        infer_df_model = infer_df_full[infer_df_full['model'] == model]
        ax.plot(infer_df_model['batch_size'].values, infer_df_model['rows_per_second'].values, label=model)
    ax.plot(infer_df_full_transform['batch_size'].values, infer_df_full_transform['rows_per_second'].values, label='transform_features')

    ax.set(xlabel='batch_size', ylabel='rows_per_second',
           title='Rows per second inference throughput by data batch_size (AdultIncome)')
    ax.grid()
    ax.legend()
    fig.savefig('infer_speed_adult_rf_onnx.png', dpi=300)
    plt.show()


```
infer_speed_adult

infer_speed_adult_rf_onnx

We can see that ONNX compile makes the model much faster with low batch sizes, and importantly shifts the bottleneck from the models to transform_features (preprocessing). There is a bit of a slowdown for large batch sizes, but not so large as to be overly concerning at the moment (still able to process 100K rows/s on a laptop)

ONNX compile also makes RF and XT go from the slowest model to nearly the fastest model in low batch size scenarios.

The new bottleneck models for low batch sizes are the neural networks (NeuralNetFastAI and NeuralNetTorch).

core/src/autogluon/core/models/abstract/abstract_model.py
liangfu: Added compile() call at the end of _fit in RFModel, with the comment above the call site of compile function.

liangfu:
The benchmark result looks very promising. I think it could guide us on which model to optimize first. Maybe with another chart demonstrating accuracy vs throughput metric, we can see what models are faster and more accurate. This could do this for both online processing (with batch size=1) and batch processing (with a large batch size).

PR #2238, opened 2022-10-21
[WIP] Use static features in GluonTS forecasting models

shchur:
*Description of changes:*
- Pass static features to GluonTS forecasting models (if supported). Currently, this only applies to DeepAR

*To do:*
- [ ] Add tests
- [ ] Add checking for validity of static features (matching columns, etc)
- [ ] Logging messages?


timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
```diff
@@ -85,6 +105,10 @@ class AbstractGluonTSModel(AbstractTimeSeriesModel):
 
     gluonts_model_path = "gluon_ts"
     gluonts_estimator_class: Type[GluonTSEstimator] = None
+    use_feat_static_cat: bool = False
+    use_feat_static_real: bool = False
+    use_feat_dynamic_real: bool = False
+    feat_static_cat_cardinality: List[int] = []
```
shchur, line 111: We will infer these from `train_data` for all models (i.e., set to True if some of the static features available), but they will only be passed down to the model for `DeepAR` and `MQCNN`

timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
```diff
@@ -32,16 +33,29 @@ class SimpleGluonTSDataset(GluonTSDataset):
     GluonTS Dataset protocol via lazy iterations.
     """
 
+    time_series_df: TimeSeriesDataFrame
```
canerturkmen, line 36: should these be defined in the `__init__` as instance variables instead? 

```
class A:
    myvar: int = 5

a = A()
a.myvar = 4

assert a.myvar == 4
assert a.__class__.myvar == 5
```

```
class A:
    def __init__(self):
        self.myvar: int = 5

a = A()

try:
    assert a.__class__.myvar == 5
except AttributeError:
    pass
```

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -196,6 +200,7 @@ def fit(
         presets: Optional[str] = None,
         hyperparameters: Dict[Union[str, Type], Any] = None,
         hyperparameter_tune_kwargs: Optional[Union[str, Dict]] = None,
+        static_feature_metadata: Optional[FeatureMetadata] = None,
```
canerturkmen, line 203: maybe also docstrings for this?

timeseries/src/autogluon/timeseries/learner.py
```diff
@@ -106,6 +136,50 @@ def _fit(
...
+    def _preprocess_static_features(
+        self,
+        train_data: TimeSeriesDataFrame,
+        static_feature_metadata: FeatureMetadata,
+        val_data: Optional[TimeSeriesDataFrame] = None,
+    ) -> Tuple[TimeSeriesDataFrame, Optional[TimeSeriesDataFrame]]:
+        self.feature_pipeline = ContinuousAndCategoricalFeatureGenerator(feature_metadata=static_feature_metadata)
+
+        # Avoid modifying train_data outside of predictor
+        train_data = train_data.copy(deep=False)
```
canerturkmen, line 148: is this working as intended? should we maybe cover it with tests in `TimeSeriesDataFrame`?

canerturkmen (commented):
looks very good overall! added some minor comments.

timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
```diff
@@ -85,6 +105,10 @@ class AbstractGluonTSModel(AbstractTimeSeriesModel):
 
     gluonts_model_path = "gluon_ts"
     gluonts_estimator_class: Type[GluonTSEstimator] = None
+    use_feat_static_cat: bool = False
```
canerturkmen, line 108: probably also instance variables here?

timeseries/tests/unittests/models/test_gluonts.py
```diff
@@ -203,3 +210,52 @@ def test_when_tft_quantiles_are_deciles_then_forecast_contains_correct_quantiles
...
+            assert (feat_static_real.dtypes == "float").all()
+
+
+@pytest.mark.parametrize("model_class", MODELS_WITH_STATIC_FEATURES)
+def test_when_static_features_present_then_model_attributes_set_correctly(model_class, df_with_static):
+    model = model_class(hyperparameters={"epochs": 1, "num_batches_per_epoch": 1})
+    model.fit(train_data=df_with_static)
+    assert model.use_feat_static_cat
+    assert model.use_feat_static_real
+    assert len(model.feat_static_cat_cardinality) == 1
```
canerturkmen, line 245: should we assert the content of cardinality as well?

canerturkmen (approved):
LGTM! Some minor comments.

timeseries/tests/unittests/models/test_gluonts.py
shchur: The static features are generated randomly at test time, we don't know it in advance. Should we make the generator deterministic?

canerturkmen: I think we might. at least randomly with fixed cardinality for both cardinality = 1 and many?

shchur: How about now?

PR #2242, opened 2022-10-24
Add Scorer error calculation plus code cleanup

Innixma:
*Issue #, if available:*

*Description of changes:*
- Add scorer error calculation via `Scorer.error`
- Majorly cleaned up scorer API, improved documentation
- Improved custom metric tutorial
- Added unit tests for scorer.optimum, we now confirm via unit tests that the optimum value for all metrics is set correctly, along with the `greater_is_better` flag. (No bugs in metrics found via this PR, so it appears our metrics are correctly implemented already)
- Fixed NeuralNetTorch using incorrect sign in HPO reporter for lower_is_better native metrics like log_loss and RMSE. This caused Bayesian Optimization and Hyperband to search poorly for NeuralNetTorch. Also fixed this issue for MXNet version of autogluon.text and Tabular's NeuralNetMXNet. (Fixed by correcting `Scorer.greater_is_better`)
- Fixed theoretical bug where `optimum` is a negative value and `greater_is_better=False` when calculating `Scorer.error`, now this situation works correctly.
- Removed old unused code such as `Scorer.sklearn_scorer`
- Added alias `Scorer.score` of `Scorer` callable to complement `Scorer.error`.
- Added `Scorer.optimum` property
- Added type hints
- Added unit test to ensure all expected metrics + aliases are available (and no unexpected ones are present)
- Added unit tests to ensure all metrics work without crashing and satisfy sanity checks of correctness regarding score consistency.

### DIFF
--- a/core/src/autogluon/core/metrics/__init__.py
+++ b/core/src/autogluon/core/metrics/__init__.py
@@ -8,7 +8,6 @@ import sklearn.metrics
 
 from . import classification_metrics
 from ..constants import BINARY, MULTICLASS, REGRESSION, QUANTILE, SOFTCLASS
-from ..utils.miscs import warning_filter
 from .classification_metrics import *
 from . import quantile_metrics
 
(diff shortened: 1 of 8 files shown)


canerturkmen (commented):
Doesn't look like timeseries is affected as our evaluation logic is independent. We can switch to using the Scorer object in the future, although we do not expose it in any way to the user so I will not prioritize it for now. Just over to @shchuro  to make sure our use of the HPO logic from core doesn't impact timeseries.

shchur:
@canerturkmen I agree, this PR should have no effect on the HPO for time series.

gidler (approved):

core/tests/unittests/metrics/test_metrics.py
```diff
@@ -0,0 +1,156 @@
...
+BINARY_METRICS = list(METRICS[BINARY].keys())
+MULTICLASS_METRICS = list(METRICS[MULTICLASS].keys())
+REGRESSION_METRICS = list(METRICS[REGRESSION].keys())
+
+
+def test_metric_exists():
+    """
+    Ensure all expected metrics are present and no unexpected metrics are present
+    """
+    expected_metrics_and_aliases = {
```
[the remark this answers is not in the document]
Innixma: It is because those metrics are present in more than 1 problem type. It is due to how I created the list to begin with by for looping through METRICS. Good catch, I'll make it so I check each problem type individually, this will actually be more comprehensive.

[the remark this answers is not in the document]
Innixma: updated to check each problem type individually

core/src/autogluon/core/metrics/__init__.py
```diff
@@ -95,36 +125,46 @@ def add_alias(self, alias):
...
         """
+        return True
+
+    @property
+    def greater_is_better_internal(self) -> bool:
+        """
+        Return whether the inner self.score_func returns metric values in greater_is_better format.
+
+        We use the stored `sign` variable to decide the property.
+        Users should not need to access this property during normal usage.
```
gradientsky, line 145: Since it is not supposed to be used by the users, should we make it private (starts with underscore)?

Innixma: It isn't quite not supposed to be used by users as it is that it would be a very odd situation where they would use this. The only time I use it is when logging metric values, I'll add a `-` in front of the metric name if `greater_is_better != greater_is_better_internal`. I think it is ok to stay as is.

Innixma: Other reasons to use it is when interacting with non-AG implementations of the metric such as models that don't support Scorer, so that you can convert the values to a consistent representation between the metrics.

gradientsky (approved):

PR #2256, opened 2022-10-26
[eda] Added ApplyFeatureGenerator

gradientsky:
*Issue #, if available:*

*Description of changes:*
Applies `FeatureGenerator` and shadows outer scope with transformed data. This is needed for analysis like correlation since it requires numeric features. If `category_to_numbers` is `True` then categories will be translated into `cat.code`
```python
    analysis = BaseAnalysis(train_data=df_train, test_data=df_test, label=target_label, children=[
        ApplyFeatureGenerator(category_to_numbers=True, children=[
            # SomeAnalysis()  # This analysis sees transformed features
        ]),
        # SomeAnalysis()  # This analysis sees original features
    ])
```

### DIFF
--- a/eda/src/autogluon/eda/analysis/transform.py
+++ b/eda/src/autogluon/eda/analysis/transform.py
@@ -50,7 +50,7 @@ class ApplyFeatureGenerator(AbstractAnalysis, StateCheckMixin):
 
     def __init__(self,
                  parent: Optional[AbstractAnalysis] = None,
-                 children: List[AbstractAnalysis] = [],
+                 children: Optional[List[AbstractAnalysis]] = None,
                  state: Optional[AnalysisState] = None,
                  category_to_numbers: bool = False,
                  feature_generator: Optional[AbstractFeatureGenerator] = None,
--- a/eda/src/autogluon/eda/analysis/transform.py
+++ b/eda/src/autogluon/eda/analysis/transform.py
@@ -1,0 +1,92 @@
+import logging
+from typing import List, Optional
+
+import pandas as pd
+
+from autogluon.features import AbstractFeatureGenerator, AutoMLPipelineFeatureGenerator
+from .base import AbstractAnalysis
+from ..state import AnalysisState
+from ..state import StateCheckMixin
+
+logger = logging.getLogger(__name__)
+
+
+class ApplyFeatureGenerator(AbstractAnalysis, StateCheckMixin):
+    """
+    This wrapper provides transformed features to all `children` shadowing outer datasets with the updated one after application of FeatureGenerator.
+
+    Parameters
+    ----------
+    category_to_numbers: bool, default = False
+        if `True', then transform `category` variables into their codes. This is useful when wrapped analyses expect numeric values
+    feature_generator: Optional[AbstractFeatureGenerator], default = None
+        feature generator to use for the transformation. If `None` is provided then `AutoMLPipelineFeatureGenerator` is applied.
+    parent: Optional[AbstractAnalysis], default = None
+        parent Analysis
+    children: List[AbstractAnalysis], default []
+        wrapped analyses; these will receive sampled `args` during `fit` call
+    kwargs
+
+    See also :func:`autogluon.features.AbstractFeatureGenerator`
+
+    Examples
+    --------
+    >>> from autogluon.eda.analysis.base import BaseAnalysis, Namespace
+    >>> import pandas as pd
+    >>> import numpy as np
+    >>> df_train = pd.DataFrame(...)
+    >>> df_test = pd.DataFrame(...)
+    >>>
+    >>> analysis = BaseAnalysis(train_data=df_train, test_data=df_test, label='D', children=[
+    >>>     Namespace(namespace='feature_generator_numbers', children=[
+    >>>         ApplyFeatureGenerator(category_to_numbers=True, children=[
+    >>>             # SomeAnalysis()  # This analysis will be called with transformed `train_data` and `test_data`
+    >>>         ])
+    >>>     ]),
+    >>>     # SomeAnalysis()  # This analysis will be called with the original features
+    >>> ])
+
+    """
+
+    def __init__(self,
+                 parent: Optional[AbstractAnalysis] = None,
+                 children: List[AbstractAnalysis] = [],
+                 state: Optional[AnalysisState] = None,
+                 category_to_numbers: bool = False,
+                 feature_generator: Optional[AbstractFeatureGenerator] = None,
+                 **kwargs) -> None:
+        super().__init__(parent, children, state, **kwargs)
+        self.category_to_numbers = category_to_numbers
+        if feature_generator is None:
+            feature_generator = AutoMLPipelineFeatureGenerator(
+                enable_numeric_features=True,
+                enable_categorical_features=True,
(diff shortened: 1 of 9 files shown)


eda/tests/unittests/analysis/test_transform.py
```diff
@@ -0,0 +1,54 @@
...
+    assert list(df_test.dtypes.apply(str).values) == ['object', 'int64', 'object', 'int64']
+
+    assert list(state.no_wrapper.args.train_data[df_train.columns].dtypes.apply(str).values) == ['object', 'int64', 'object', 'int64']
+    assert list(state.no_wrapper.args.test_data[df_test.columns].dtypes.apply(str).values) == ['object', 'int64', 'object', 'int64']
+
+    assert list(state.feature_generator_default.args.train_data[df_train.columns].dtypes.apply(str).values) == ['category', 'int64', 'category', 'int64']
+    assert list(state.feature_generator_default.args.test_data[df_test.columns].dtypes.apply(str).values) == ['category', 'int64', 'category', 'int64']
+
+    assert list(state.feature_generator_numbers.args.train_data[df_train.columns].dtypes.apply(str).values) == ['int8', 'int64', 'int8', 'int64']
+    assert list(state.feature_generator_numbers.args.test_data[df_test.columns].dtypes.apply(str).values) == ['int8', 'int64', 'int8', 'int64']
```
gidler, line 54: TIOLI: [the official recomendation](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.values.html) is now to use `Series.array` or `Series.to_numpy()` instead of `Series.values`

eda/src/autogluon/eda/state.py
```diff
@@ -53,7 +58,7 @@ def at_least_one_key_must_be_present(self, state: AnalysisState, *keys):
         for k in keys:
             if k in state:
                 return True
-        logger.warning(f'{self.__class__.__name__}: at least one of the following keys must be present: {keys}')
+        self.logger.warning(f'{self.__class__.__name__}: at least one of the following keys must be present: {keys}')
         return False
 
     def all_keys_must_be_present(self, state: AnalysisState, *keys):
```
gidler, line 64: While we're here, we could add return types to this function

eda/src/autogluon/eda/state.py
```diff
@@ -35,6 +38,8 @@ def __dict__(self):
 
 
 class StateCheckMixin:
+    logger = logging.getLogger(__name__)
+
     def at_least_one_key_must_be_present(self, state: AnalysisState, *keys):
```
gidler, line 43: While we're here, we could add return types to these functions

eda/src/autogluon/eda/analysis/transform.py
```diff
@@ -0,0 +1,92 @@
...
+    >>>         ])
+    >>>     ]),
+    >>>     # SomeAnalysis()  # This analysis will be called with the original features
+    >>> ])
+
+    """
+
+    def __init__(self,
+                 parent: Optional[AbstractAnalysis] = None,
+                 children: List[AbstractAnalysis] = [],
```
gidler, line 53: Could this be a case where the "default mutable arguments" behavior that is often a gotcha in Python (see `Important warning` [here](https://towardsdatascience.com/python-pitfall-mutable-default-arguments-9385e8265422)) could cause unexpected results? I'm having a hard time seeing all the ways `children` might be used as it moves through the inheritance tree, but I thought it was worth surfacing in case you can tell if it might be dangerous to users.

eda/src/autogluon/eda/analysis/transform.py
```diff
@@ -0,0 +1,92 @@
...
+logger = logging.getLogger(__name__)
+
+
+class ApplyFeatureGenerator(AbstractAnalysis, StateCheckMixin):
+    """
+    This wrapper provides transformed features to all `children` shadowing outer datasets with the updated one after application of FeatureGenerator.
+
+    Parameters
+    ----------
+    category_to_numbers: bool, default = False
```
tonyhoo, line 20: Is that possible for us to infer this automatically based on wrapped analyses? Not necessary to incorporate in this PR but something worthwhile to simplify in future

eda/src/autogluon/eda/analysis/transform.py
tonyhoo: +1, we can init it to `None`. Also wondering if we should add some bot code reviewer (e.g. this [one](https://docs.aws.amazon.com/codeguru/latest/reviewer-ug/create-code-reviews.html)) comments here to capture these

eda/src/autogluon/eda/analysis/transform.py
```diff
@@ -0,0 +1,92 @@
...
+                 children: List[AbstractAnalysis] = [],
+                 state: Optional[AnalysisState] = None,
+                 category_to_numbers: bool = False,
+                 feature_generator: Optional[AbstractFeatureGenerator] = None,
+                 **kwargs) -> None:
+        super().__init__(parent, children, state, **kwargs)
+        self.category_to_numbers = category_to_numbers
+        if feature_generator is None:
+            feature_generator = AutoMLPipelineFeatureGenerator(
+                enable_numeric_features=True,
```
tonyhoo, line 62: Wondering if we can just use defaults instead of providing overrides which might change over time

eda/src/autogluon/eda/analysis/transform.py
```diff
@@ -0,0 +1,92 @@
...
+                enable_raw_text_features=False,
+                enable_vision_features=False,
+            )
+        self.feature_generator = feature_generator
+
+    def can_handle(self, state: AnalysisState, args: AnalysisState) -> bool:
+        return self.all_keys_must_be_present(args, 'train_data', 'label')
+
+    def _fit(self, state: AnalysisState, args: AnalysisState, **fit_kwargs) -> None:
+        x = args.train_data.drop(columns=args.label)
```
tonyhoo, line 76: Not sure if `train_data` can be mutated in the same analysis. If not, checking on previous results is available might save some duplicate efforts

gradientsky: this line doesn't mutate the original data

eda/src/autogluon/eda/analysis/transform.py
gradientsky: Feature generators are stateful, we don't want to initialize an object in defaults.

eda/src/autogluon/eda/analysis/transform.py
gradientsky: This is option because some users would prefer one thing (keep `category` type), while others the other (get category codes).

eda/tests/unittests/analysis/test_transform.py
gradientsky: Updated. Thanks for the reference link.

eda/src/autogluon/eda/state.py
gradientsky: done

eda/src/autogluon/eda/analysis/transform.py
gradientsky: fixed

tonyhoo (approved):

gidler (approved):

PR #2257, opened 2022-10-26
[Warning] Remove a few warnings / output messages

sxjscience:
*Issue #, if available:*

*Description of changes:*

- [x] Previously, we will raise a warning if the user is using CPU for inference. This PR disables this warning.
- [x] There was a deprecation warning like the following. This has been fixed since we now require `pytorch_lightning>=1.7.4,<1.8.0`
```bash
/home/ci/opt/venv/lib/python3.8/site-packages/pytorch_lightning/trainer/connectors/accelerator_connector.py:447: LightningDeprecationWarning: Setting Trainer(gpus=1) is deprecated in v1.7 and will be removed in v2.0. Please use Trainer(accelerator='gpu', devices=1) instead.
  rank_zero_deprecation(
```  
- [x] In the first stage of `._predict()`, we added `pl.seed_everything(...)`, which will print a info message called `Global seed set to ...`. The PR filters this print message.


gradientsky (approved):

cheungdaven (approved):
LGTM!

PR #2260, opened 2022-10-26
[Tabular] Introduce compile_models function to tabular predictor

liangfu:
*Description of changes:*

Following https://github.com/awslabs/autogluon/pull/2225, this PR takes a further step to introduce `compile_models` function to TabularPredictor. This helps

* Decouple `compile_models` from `fit` and `save` function from tabular predictor.
* Remove compiler option from hyperparameters, since hyperparameters are primarily designed for model fitting. Compiler options are more suitable for post-fit process.
* Add support to compile WeightedEnsembleModel.

Also, it is required to persist model before model compilation. Therefore, the convention for model compilation would be
```python
predictor = TabularPredictor(label=label, path=save_path).fit(train_data)

# Compile and persist the best model for efficient prediction without loading model in every predict() call.
# NOTE: leave `with_ancestors` argument to be the default value True, in case the best model is an ensemble model.
compiler_configs = {RFModel: {'compiler': 'onnx'}}
perdictor.compile_models(models='best', compiler_configs=compiler_configs)
predictor.persist_models(models='best')

# Perform prediction just as before
y_pred = predictor.predict(test_data)
```

To compare performance with/without zipmap
image

### DIFF
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -908,7 +908,6 @@ class AbstractModel:
         NOTE:
         - The model is assumed to be fitted before compilation.
         - If save_in_pkl attribute of the compiler is False, self.model would be set to None.
-          The compiled model can be loaded via load_model method.
 
         Parameters
         ----------
@@ -942,16 +941,52 @@ class AbstractModel:
                 self.model = None
             self.save(path=self.path, verbose=False)
 
-    def _get_input_types(self, batch_size=None):
+    # FIXME: This won't work for all models, and self._features is not
+    # a trustworthy variable for final input shape
+    def _get_input_types(self, batch_size=None) -> list:
+        """
+        Get input types as a list of tuples, containining shape and dtype.
+        This can be useful for building the input_types argument for
+        model compilation. This method can be overloaded in derived classes,
+        in order to satisfy class-specific requirements.
+
+        Parameters
+        ----------
+        batch_size : int, default=None
+            The batch size for all returned input types.
+
+        Returns
+        -------
+        List of (shape: Tuple[int], dtype: Any)
+        shape: Tuple[int]
+            A tuple that describes input
+        dtype: Any, default=np.float32
+            The element type in numpy dtype.
+        """
         return [((batch_size, len(self._features)), np.float32)]
 
     def _default_compiler(self):
+        """The default compiler for the underlining model."""
         return None
 
-    def _valid_compilers(self):
+    def _valid_compilers(self) -> list:
+        """A list of supported compilers for the underlining model."""
         return []
 
-    def _get_compiler(self, compiler='native', compiler_fallback_to_native=True):
+    def _get_compiler(self, compiler='native', compiler_fallback_to_native=False):
+        """
+        Verify whether the dependencies of the compiler class can be satisfied,
+        and return the specified compiler from _valid_compilers.
+
+        Parameters
+        ----------
+        compiler : str, default='native'
+            The specific compiler for model compilation.
+        compiler_fallback_to_native : bool, default=False
+            If this is True, the method would return native compiler when
+            dependencies of the specified compiler is not installed. The fallback
+            strategy won't be used by default.
+        """
         compilers = self._valid_compilers()
         compiler_names = {c.name: c for c in compilers}
         if compiler is not None and compiler not in compiler_names:
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -881,18 +881,27 @@ class AbstractModel:
         assert self.is_fit(), "The model must be fit before calling the compile method."
         if compiler_configs is None:
             compiler_configs = {}
-        compiler = compiler_configs.get("compiler", "native")
-        batch_size = compiler_configs.get("batch_size", None)
 
-        save_in_pkl = True
+        # Get model specific compiler options
+        # Model type can be described with either model type, or model name as string
+        if type(self) in compiler_configs:
+            configs = compiler_configs[type(self)]
+        elif self.name in compiler_configs:
+            configs = compiler_configs[self.name]
+        else:
+            configs = compiler_configs
+        compiler = configs.get("compiler", "native")
+        batch_size = configs.get("batch_size", None)
+
         if self.model is not None:
             self._compiler = self._get_compiler(compiler=compiler)
             if self._compiler is not None:
                 input_types = [(tuple([batch_size]+list(shape)[1:]), dtype)
                                for shape, dtype in self._input_types_post_process]
-                self.model = self._compiler.compile(model=self.model, input_types=input_types)
-                save_in_pkl = self._compiler.save_in_pkl
-        return self.model, save_in_pkl
+                self._compiler.compile(model=self.model, path=self.path, input_types=input_types)
+                if not self._compiler.save_in_pkl:
+                    self.model = None
+                self.save(path=self.path, verbose=False)
 
     def _default_compiler(self):
         return None
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -880,16 +880,8 @@ class AbstractModel:
         """
         if not self.is_fit():
             return False
-        # Get model specific compiler options
-        # Model type can be described with either model type, or model name as string
-        if type(self) in compiler_configs:
-            configs = compiler_configs[type(self)]
-        elif self.name in compiler_configs:
-            configs = compiler_configs[self.name]
-        else:
-            configs = compiler_configs
-        compiler = configs.get("compiler", "native")
-        compiler_fallback_to_native = configs.get('compiler_fallback_to_native', False)
+        compiler = compiler_configs.get("compiler", "native")
+        compiler_fallback_to_native = compiler_configs.get('compiler_fallback_to_native', False)
 
         compilers = self._valid_compilers()
         compiler_names = {c.name: c for c in compilers}
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -154,6 +154,8 @@ class AbstractModel:
         self._is_fit_metadata_registered = False
         self._fit_metadata = dict()
 
+        self._compiler = None
+
     def _init_params(self):
         """Initializes model hyperparameters"""
         hyperparameters = self._user_params
@@ -771,7 +773,16 @@ class AbstractModel:
         if path is None:
             path = self.path
         file_path = path + self.model_file_name
+        _model = self.model
+        if self.model is not None:
+            if self._compiler is None:
+                self._compiler = self._get_compiler()
+                if self._compiler is not None and not self._compiler.save_in_pkl:
+                    self._compiler.save(model=self.model, path=path)
+            if self._compiler is not None and not self._compiler.save_in_pkl:
+                self.model = None  # Don't save model in pkl
         save_pkl.save(path=file_path, object=self, verbose=verbose)
+        self.model = _model
         return path
 
     @classmethod
@@ -801,6 +812,8 @@ class AbstractModel:
         model = load_pkl.load(path=file_path, verbose=verbose)
         if reset_paths:
             model.set_contexts(path)
+        if model._compiler is not None and not model._compiler.save_in_pkl:
+            model.model = model._compiler.load(path=path)
         return model
 
     def compute_feature_importance(self,
@@ -919,10 +932,7 @@ class AbstractModel:
                                             compiler_fallback_to_native=compiler_fallback_to_native)
         if self._compiler is not None:
             input_types = self._get_input_types(batch_size=batch_size)
-            self._compiler.compile(model=self.model, path=self.path, input_types=input_types)
-            if not self._compiler.save_in_pkl:
-                self.model = None
-            self.save(path=self.path, verbose=False)
+            self.model = self._compiler.compile(model=self.model, path=self.path, input_types=input_types)
 
     # FIXME: This won't work for all models, and self._features is not
     # a trustworthy variable for final input shape
@@ -956,14 +966,14 @@ class AbstractModel:
         """A list of supported compilers for the underlining model."""
         return []
 
-    def _get_compiler(self, compiler='native', compiler_fallback_to_native=False):
+    def _get_compiler(self, compiler: str = None, compiler_fallback_to_native=False):
         """
         Verify whether the dependencies of the compiler class can be satisfied,
         and return the specified compiler from _valid_compilers.
 
         Parameters
         ----------
-        compiler : str, default='native'
+        compiler : str, default=None
             The specific compiler for model compilation.
         compiler_fallback_to_native : bool, default=False
             If this is True, the method would return native compiler when
@@ -983,6 +993,13 @@ class AbstractModel:
                                      ' (potentially lacking dependencies) and "compiler_fallback_to_native==False"')
             compiler_cls = self._default_compiler()
         return compiler_cls
+
+    def get_compiler_name(self) -> str:
+        assert self.is_fit(), "The model must be fit before calling the get_compiler_name method."
+        if self._compiler is not None:
+            return self._compiler.name
+        else:
+            return 'native'
 
     def get_trained_params(self) -> dict:
         """
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -127,7 +127,6 @@ class AbstractModel:
         self._features_internal = None  # Internal features, safe to use internally via the `_features` property
         self._feature_metadata = None  # Internal feature metadata, safe to use internally
         self._is_features_in_same_as_ex = None  # Whether self.features == self._features_internal
-        self._input_types_post_process = None # A list of tuples containing [(shape, dtype),] for compiler optimization
 
         self.fit_time = None  # Time taken to fit in seconds (Training data)
         self.predict_time = None  # Time taken to predict in seconds (Validation data)
@@ -307,7 +306,6 @@ class AbstractModel:
             X = self._preprocess_nonadaptive(X, **kwargs)
         if preprocess_stateful:
             X = self._preprocess(X, **kwargs)
-        self._input_types_post_process = [(x.shape, x.dtype) for x in X] if isinstance(X, list) else [(X.shape, X.dtype),]
         return X
 
     # TODO: Remove kwargs?
@@ -930,12 +928,14 @@ class AbstractModel:
         self._compiler = self._get_compiler(compiler=compiler,
                                             compiler_fallback_to_native=compiler_fallback_to_native)
         if self._compiler is not None:
-            input_types = [(tuple([batch_size]+list(shape)[1:]), dtype)
-                           for shape, dtype in self._input_types_post_process]
+            input_types = self._get_input_types(batch_size=batch_size)
             self._compiler.compile(model=self.model, path=self.path, input_types=input_types)
             if not self._compiler.save_in_pkl:
                 self.model = None
             self.save(path=self.path, verbose=False)
+
+    def _get_input_types(self, batch_size=None):
+        return [((batch_size, len(self._features)), np.float32)]
 
     def _default_compiler(self):
         return None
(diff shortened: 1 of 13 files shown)


core/src/autogluon/core/trainer/abstract_trainer.py
```diff
@@ -1134,6 +1134,15 @@ def save(self):
         if self.low_memory:
             self.models = models
 
+    def compile_models(self, **kwargs):
+        models_to_compile = list(self.models.keys())
```
Innixma, line 1138: Don't use self.models. Instead

```
if models == 'all':
    models = self.get_model_names(can_infer=True)
```

core/src/autogluon/core/trainer/abstract_trainer.py
```diff
@@ -1134,6 +1134,15 @@ def save(self):
...
+        if len(models_to_compile) == 0:
+            raise RuntimeError("No valid persisted model to be compiled. "
+                               "Make sure make a call to `persist_models` before trying to compile the models.")
```
Innixma, line 1141: The models should not need to be persisted for this.

core/src/autogluon/core/trainer/abstract_trainer.py
```diff
@@ -1134,6 +1134,15 @@ def save(self):
         if self.low_memory:
             self.models = models
 
+    def compile_models(self, **kwargs):
```
Innixma, line 1137: Missing documentation

core/src/autogluon/core/trainer/abstract_trainer.py
```diff
@@ -1134,6 +1134,15 @@ def save(self):
...
+        models_to_compile = list(self.models.keys())
+        if len(models_to_compile) == 0:
+            raise RuntimeError("No valid persisted model to be compiled. "
+                               "Make sure make a call to `persist_models` before trying to compile the models.")
+        for model_name in models_to_compile:
+            self.models[model_name].compile(compiler_configs=kwargs)
```
Innixma, line 1143:
```
for model_name in models:
    model = self.load_model(model_name)
    # Check if already compiled or if can't compile due to missing dependencies or if model hasn't implemented compiling
    if model.can_compile(kwargs):
         model.compile(compiler_configs=kwargs)
         model.save()
```

core/src/autogluon/core/trainer/abstract_trainer.py
Innixma, line 1137: Add `models` argument so user can specify the models they care about. Default to 'all'.

Add `# TODO: v0.6 Support 'best'`

core/src/autogluon/core/trainer/abstract_trainer.py
Innixma, line 1143: This will probably crash with bagged models. Need to add logic that compiles for bagged_ensemble_model by calling all children's compile methods.

core/src/autogluon/core/models/ensemble/weighted_ensemble_model.py
```diff
@@ -72,3 +72,8 @@ def _set_default_params(self):
         for param, val in default_params.items():
             self._set_default_param_value(param, val)
         super()._set_default_params()
+
+    def compile(self, path: str = None, verbose=True, compiler_configs={}) -> str:
```
Innixma, line 76: Don't use `{}` as default (mutable), use `None` and then `if compiler_configs is None: compiler_configs = {}`

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -1950,6 +1950,29 @@ def feature_importance(self, data=None, model=None, features=None, feature_stage
             fi_df[low_str] = pd.Series(ci_low_dict)
         return fi_df
 
+    def compile_models(self, compiler_configs={}):
```
Innixma, line 1953: Don't use {} as default (mutable), use None and then if compiler_configs is None: compiler_configs = {}

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -1950,6 +1950,29 @@ def feature_importance(self, data=None, model=None, features=None, feature_stage
...
+        This can be helpful to reduce prediction latency and improve throughput.
+
+        Note that this is currently an experimental feature, the supported alternative compiler can be ['native', 'onnx'].
+
+        Parameters
+        ----------
+        compiler_configs : dict, default = {}
+            compiler : str, default = 'native'
+                The compiler that is used for model compilation.
+            batch_size : int, default = None
```
Innixma, line 1965: unused?

tabular/src/autogluon/tabular/models/rf/compilers/onnx.py
```diff
@@ -0,0 +1,80 @@
...
+    @staticmethod
+    def can_compile():
+        try:
+            import skl2onnx
+            return True
+        except ImportError:
+            return False
+
+    @staticmethod
+    def compile(obj, path: str, batch_size=None):
```
Innixma, line 62: batch_size unused?

Innixma: batch_size not available in abstract model, will cause crashes?

tabular/tests/conftest.py
```diff
@@ -130,7 +130,9 @@ def fit_and_validate_dataset(dataset_name,
                                  refit_full=True,
                                  delete_directory=True,
                                  extra_metrics=None,
-                                 expected_model_count=2):
+                                 expected_model_count=2,
+                                 compile_models=False,
+                                 compiler_configs={}):
```
Innixma, line 135: Don't use {} as default (mutable), use None and then if compiler_configs is None: compiler_configs = {}

Innixma (commented):
Added initial review

core/src/autogluon/core/trainer/abstract_trainer.py
liangfu: Done. 

In addition to this, an argument `model_names` has been added with default value to be 'all'.

core/src/autogluon/core/trainer/abstract_trainer.py
liangfu: Deleted, persist_model() is no longer required before compile_model().

tabular/src/autogluon/tabular/predictor/predictor.py
liangfu: Yes, this is currently unused for onnx backend. As this is going to be a backend that would be used for other backends, the batch_size argument is very much useful to backends that doesn't support dynamic shape. Examples could be:
* https://onnxruntime.ai/docs/tutorials/mobile/helpers/make-dynamic-shape-fixed.html

tabular/src/autogluon/tabular/predictor/predictor.py
liangfu: Done.

core/src/autogluon/core/models/ensemble/weighted_ensemble_model.py
liangfu: Done.

core/src/autogluon/core/trainer/abstract_trainer.py
liangfu: Done.

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -306,6 +306,7 @@ def preprocess(self, X, preprocess_nonadaptive=True, preprocess_stateful=True, *
             X = self._preprocess_nonadaptive(X, **kwargs)
         if preprocess_stateful:
             X = self._preprocess(X, **kwargs)
+        self._input_types_post_process = [(x.shape, x.dtype) for x in X] if isinstance(X, list) else [(X.shape, X.dtype),]
```
liangfu, line 310: Remove this and only take this for compilation?

liangfu: Removed and replaced with self._features attribute.

tabular/tests/conftest.py
liangfu: Done.

core/src/autogluon/core/trainer/abstract_trainer.py
liangfu: Currently we are supporting stacker_ensemble_model, but not bagged_ensemble_model. I still didn't get the idea how to get base_model in bagged_ensemble_model.

core/src/autogluon/core/trainer/abstract_trainer.py
liangfu: Updated. Now we takes `models` argument, and both `all` and `best` are supported.

core/src/autogluon/core/trainer/abstract_trainer.py
liangfu: 1) Added `can_compile` method, 2) added the comment after loading the model.

tabular/src/autogluon/tabular/models/rf/compilers/onnx.py
liangfu: > batch_size unused?

batch_size is not used for onnx backend, but would be used for other backends in the future.

> batch_size not available in abstract model, will cause crashes?

Will not crash. batch_size is now available in abstract model. `batch_size` is essentially controlling the compiler to optimize towards either _online_ or _batch_ prediction.

core/src/autogluon/core/trainer/abstract_trainer.py
```diff
@@ -1134,6 +1134,44 @@ def save(self):
         if self.low_memory:
             self.models = models
 
+    def compile_models(self, model_names='all', compiler_configs=None):
```
Innixma, line 1137: Add `with_ancestors` argument, refer to persist_models for implementation

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -867,43 +867,98 @@ def _compute_permutation_importance(self, X, y, features: list, eval_metric=None
...
+            e.g. {"RandomForest": {"compiler": "onnx"}}
+        """
+        if not self.is_fit():
+            return False
+        # Get model specific compiler options
+        # Model type can be described with either model type, or model name as string
+        if type(self) in compiler_configs:
+            configs = compiler_configs[type(self)]
+        elif self.name in compiler_configs:
+            configs = compiler_configs[self.name]
```
Innixma, line 888: Why are we allowing passing of other model compiler configs? This should be handled upstream, and `compiler_configs` should simply be `{"compiler": "onnx"}` in this context.

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -867,43 +867,98 @@ def _compute_permutation_importance(self, X, y, features: list, eval_metric=None
...
+        assert self.is_fit(), "The model must be fit before calling the compile method."
+        if compiler_configs is None:
+            compiler_configs = {}
+
+        # Get model specific compiler options
+        # Model type can be described with either model type, or model name as string
+        if type(self) in compiler_configs:
+            configs = compiler_configs[type(self)]
+        elif self.name in compiler_configs:
+            configs = compiler_configs[self.name]
```
Innixma, line 929: ditto

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -867,43 +867,98 @@ def _compute_permutation_importance(self, X, y, features: list, eval_metric=None
...
+                                            compiler_fallback_to_native=compiler_fallback_to_native)
+        if self._compiler is not None:
+            input_types = self._get_input_types(batch_size=batch_size)
+            self._compiler.compile(model=self.model, path=self.path, input_types=input_types)
+            if not self._compiler.save_in_pkl:
+                self.model = None
+            self.save(path=self.path, verbose=False)
+
+    def _get_input_types(self, batch_size=None):
+        return [((batch_size, len(self._features)), np.float32)]
```
Innixma, line 946: This is a false assumption. I guess we can wait to fix it until it becomes a problem, but `self._features` is not the truth of the input shape.

Innixma: Need to add documentation to this method

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -867,43 +867,98 @@ def _compute_permutation_importance(self, X, y, features: list, eval_metric=None
...
 
     def _default_compiler(self):
         return None
 
-    def _get_compiler(self):
-        compiler = self.params_aux.get('compiler', None)
+    def _valid_compilers(self):
+        return []
```
Innixma, line 952: documentation

Innixma: add return types

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -867,43 +867,98 @@ def _compute_permutation_importance(self, X, y, features: list, eval_metric=None
...
+
+    def _get_compiler(self, compiler='native', compiler_fallback_to_native=True):
```
Innixma, line 954: documentation

core/src/autogluon/core/models/ensemble/stacker_ensemble_model.py
```diff
@@ -229,3 +229,10 @@ def _add_stack_to_feature_metadata(self):
                 stacker_feature_metadata = stacker_feature_metadata.keep_features([feature for feature in existing_features if feature in type_map_raw])
                 if set(stacker_feature_metadata.get_features()) != set(existing_stack_features):
                     self.feature_metadata = self.feature_metadata.add_special_types(stacker_feature_metadata.get_type_map_special())
+
+    def compile(self, compiler_configs=None) -> str:
+        if compiler_configs is None:
+            compiler_configs = {}
+        for model_name in self.base_model_names:
+            model = self.load_base_model(model_name)
+            model.compile(compiler_configs=compiler_configs)
```
Innixma, line 238: Not correct. Must be fixed before merging. Do not edit stacker_ensemble_model. Edit BaggedEnsembleModel.

for `model_name in self.models`. Follow predict_proba logic for bagged ensemble model.

tabular/src/autogluon/tabular/models/rf/compilers/onnx.py
```diff
@@ -59,22 +60,53 @@ def can_compile():
...
+        if isinstance(model, (RandomForestClassifier, ExtraTreesClassifier)):
+            options = {id(model): {'zipmap': False}}
+
+        # Convert the model to onnx
+        onnx_model = convert_sklearn(model, initial_types=initial_type, options=options)
+        predictor =  RFOnnxPredictor(model=onnx_model)
+        RFOnnxCompiler.save(onnx_model, path)
+
+    @staticmethod
+    def save(model, path: str):
```
Innixma, line 102: add type hints for output, ditto elsewhere

Innixma (changes requested):

core/src/autogluon/core/models/abstract/abstract_model.py
Innixma: Please add `# FIXME: This won't work for all models, and self._features is not a trustworthy variable for final input shape`

core/src/autogluon/core/trainer/abstract_trainer.py
Innixma: refer to latest review for clarification

core/src/autogluon/core/models/abstract/abstract_model.py
liangfu: This is kind of a model-specific compiler configration. For ensemble models, we could use a specific backend for a specific model. For instance, an ensemable model with RF and TORCH_NN, I think we could configure the compiler to optimize RF with onnx, while optimize TORCH_NN model with tvm, in order to maximize potential performance benefits. Of course, we can still use the top-level compiler config if model-specific compiler is not specified.

core/src/autogluon/core/models/ensemble/stacker_ensemble_model.py
liangfu: I still cannot follow predict_proba logic for bagged ensemble model, because:
1. BaggedEnsemableModel doesn’t have the attribute base_model_paths_dict, therefore, won’t be able to load the base models for compilation.
2. The predict_proba method in BaggedEnsembleModel only takes outputs of the base models. The outputs of the base models are calculated in StackerEnsembleModel.preprocess, where it takes base_model_paths_dict attribute to load the base model.

Here are some details:
```python
def compile(self, compiler_configs=None):
    if compiler_configs is None:
        compiler_configs = {}
    # Following predict_proba logic for bagged ensemble model
    for model in self.models:
        # model is a GreedyWeightedEnsembleModel
        model = self.load_child(model)
        # model is still a GreedyWeightedEnsembleModel
        # NOTE: GreedyWeightedEnsembleModel would only weight outputs of base models
       
        # We need to compile the base model here, in order to compile it.
        model.compile(compiler_configs=compiler_configs)
```
```python
def predict_proba(self, X, normalize=None, **kwargs):
    model = self.load_child(self.models[0])
    # The input X is a pd.DataFrame
    # For a WeightedEnsembleModel, this will call into preprocess method in StackerEnsembleModel.
    # Such preprocess method takes base_model_paths_dict to load base model
    X = self.preprocess(X, model=model, **kwargs)
    # The output X is a list of np array
   
    pred_proba = model.predict_proba(X=X, preprocess_nonadaptive=False, normalize=normalize)
    for model in self.models[1:]:
        model = self.load_child(model)
        pred_proba += model.predict_proba(X=X, preprocess_nonadaptive=False, normalize=normalize)
    pred_proba = pred_proba / len(self.models)
```
In summary, We need to load the base model in BaggedEnsembleModel, in order to compile it. However, it is the StackerEnsembleModel that keeps the base_model_paths_dict attribute, which is required to load the base model.

liangfu: As per offline discussion, we can have access to base models by introducing `with_ancestors` argument in `compile_models` interface. Such that the base models are loaded by abstract_trainer and get compiled there.

core/src/autogluon/core/trainer/abstract_trainer.py
liangfu: Added the `with_ancestors` argument.

core/src/autogluon/core/models/abstract/abstract_model.py
liangfu: > This is a false assumption. 

Yes, this may not be true for all models. I think we can override this method when it's not applicable.

Added document and the comment to keep further notice.

core/src/autogluon/core/models/abstract/abstract_model.py
liangfu: Added document and return type hint.

core/src/autogluon/core/models/abstract/abstract_model.py
liangfu: Documentation added.

core/src/autogluon/core/models/ensemble/stacker_ensemble_model.py
liangfu: Added `with_ancestors` argument in abstract_trainer, in order to compile base models.

tabular/src/autogluon/tabular/models/rf/compilers/onnx.py
liangfu: Added documents and output type hints.

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -867,43 +867,133 @@ def _compute_permutation_importance(self, X, y, features: list, eval_metric=None
...
-    #  Add Documentation
-    #  Define when this should be called
-    #  Is it called during fit? Is it called after fit?
-    #  Can it be called as a post-fit operation on an already fit predictor on a per-model basis?
-    #  Does this call overwrite the existing model or make a new one?
-    def compile(self, path: str = None, verbose=True) -> str:
-        if path is None:
-            path = self.path
-        file_path = path + self.model_file_name
+    def can_compile(self, compiler_configs=None):
```
tonyhoo, line 870: Do we need `compiler_config` to be passed by the caller each time?

core/src/autogluon/core/models/abstract/abstract_model.py
Innixma: ```a model-specific compiler configration```

This shouldn't be available at the AbstractModel level, we only need the configuration for the model itself in this context, not others.

core/src/autogluon/core/models/abstract/abstract_model.py
liangfu: For now, I think the answer is yes. It makes the interface a little bit difficult to configure, since users may not aware the underlining compilers that are supported.

In the long term, I think we may introduce the `presets` argument to the `compile_models` method.

To support edge devices, we could support presets like
* rpi
* jetson-tx1

For cloud deployment, presets could be something like
* aws-lambda # for online prediction, batch_size=1
* sagemaker-endpoint # for batch prediction, batch_size=1024

core/src/autogluon/core/models/abstract/abstract_model.py
liangfu: Model-specific config logic has been moved outside of abstract_model.

Innixma (approved):
Looks great! Thanks for the contribution!

PR #2263, opened 2022-10-27
Tabular Resource Allocation Logic Rework

yinweisu:
*Issue #, if available:*
https://github.com/awslabs/autogluon/issues/2254

*Description of changes:*
This PR started to only address https://github.com/awslabs/autogluon/issues/2254. But evolved into a resource allocation logic rework to correctly handle different user specified resources. Ideally, user don't need to play around with these values, but we don't eliminate the possibilities
* Added `num_cpus` and `num_gpus` as a parameter to the `fit()` call to control the total resources given to AG.
* `AbstractModel._preprocess_fit_resources()` now handles the total resource calculation based on 1. user specified total resources 2. system resources 3. model default resources. Hyperparameter tune will also call this method to handle the total resources. This method should be the only entry point to get the total resources.
* When hpo, `executor.register_resources()` now handles user custom resources per trial / resources per fold. It will calculate resources_per_trial based on total resource being passed to it and user specified values
* `FoldFittingStrategy` will use total resource being passed to it and user specified values to determine resources per fold.
* Added minimum resource when gpu is available. Conceptually, we can have different minimum requirement based on the resource available if a model can be trained both on cpu and gpu. For example, a model might require minimum of 4 cpus when there's no gpu because cpu is slower. Or 1 cpu and 1 gpu or gpu is available. Previously, the minimum resource gpu required by NN_Torch and NN_FASTAI is 0; therefore, not using gpu even when it's available.
* Fixed `time_limit = None` crash issue
* Added many unit tests.

Screenshot of 8 fastai trials running in parallel on 4 gpus. Each taking 0.5 gpu.
image


Innixma (approved):
Looks great! Feel free to merge once CI passes.

Question: Is GPU used during bagging in the same way? (FastAI using 0.5 GPU for each fold in 8 folds?) Was bagging working correctly before this PR?

tabular/tests/unittests/resource_allocation/test_total_resource_allocation.py
```diff
@@ -0,0 +1,219 @@
...
+        'num_cpus': 1,
+        'num_gpus': 0,
+    }
+    resources = bagged_model._preprocess_fit_resources(total_resources=total_resources)
+    assert resources == total_resources
+    
+    # Given total resources more than what the system has
+    total_resources = {
+        'num_cpus': 99,
+        'num_gpus': 99,
```
Innixma, line 45: Make 99999, it is totally possible for a machine to have over 99 cpus.

tabular/tests/unittests/resource_allocation/test_total_resource_allocation.py
```diff
@@ -0,0 +1,219 @@
...
+    
+    
+def test_bagged_model_without_total_resources_but_with_ensemble_resources():
+    model_base = DummyModel()
+    bagged_model = DummyBaggedModel(
+        model_base,
+        hyperparameters={
+            'ag_args_fit': {
+                'num_cpus': 99,
+                'num_gpus': 99,
```
Innixma, line 115: 99999

Innixma: Ditto elsewhere

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -475,16 +476,48 @@ def _init_misc(self, **kwargs):
         else:
             self.normalize_pred_probas = False
 
-    def _preprocess_fit_resources(self, silent=False, **kwargs):
+    def _preprocess_fit_resources(self, silent=False, total_resources=None, **kwargs):
```
Innixma, line 479: Since we made this more complicated, add docstring

core/src/autogluon/core/models/ensemble/fold_fitting_strategy.py
```diff
@@ -249,7 +284,21 @@ def _fit(self, model_base, time_start_fold, time_limit_fold, fold_ctx, kwargs):
             X_fold = pd.concat([X_fold, self.X_pseudo], axis=0, ignore_index=True)
             y_fold = pd.concat([y_fold, self.y_pseudo], axis=0, ignore_index=True)
 
-        fold_model.fit(X=X_fold, y=y_fold, X_val=X_val_fold, y_val=y_val_fold, time_limit=time_limit_fold, **kwargs_fold)
+        num_cpus = self.num_cpus
+        num_gpus = self.num_gpus
+        if self.user_resources_per_job is not None:
+            num_cpus = min(self.num_cpus, self.user_resources_per_job.get('num_cpus', math.inf))
+            num_gpus = min(self.num_gpus, self.user_resources_per_job.get('num_gpus', math.inf))
```
Innixma, line 291: Why math.inf?

core/src/autogluon/core/models/ensemble/fold_fitting_strategy.py
```diff
@@ -524,24 +560,41 @@ def _get_fold_time_limit(self):
             time_limit_fold = None
         return time_limit_fold
 
-    def _get_resource_suggestions(self, num_jobs, user_specified_num_folds_parallel):
+    def _get_resource_suggestions(
```
Innixma, line 563: Lets add documentation here

core/src/autogluon/core/ray/resources_calculator.py
```diff
@@ -19,25 +19,39 @@ def calc_type(self):
...
+            Number of gpus specified by the user.
+            This corresponds to the number of gpus used by the hpo experiment(shared by all trials when training a model)
+            or the bagging of the model(shared by all bags)
+        """
         if user_specified_num_gpus is not None:
-            total_gpu = get_gpu_count_all() if num_gpus_available is None else min(get_gpu_count_all(), num_gpus_available)
-            num_gpus = min(user_specified_num_gpus, total_gpu)
+            num_gpus = min(user_specified_num_gpus, get_gpu_count_all())
+            if num_gpus < user_specified_num_gpus:
+                logger.warning(f'Specified num_gpus = {user_specified_num_gpus}, but only {num_gpus} are available')
```
Innixma, line 37: Should this only be a warning? Would it make sense to raise an exception instead?

core/src/autogluon/core/ray/resources_calculator.py
```diff
@@ -19,25 +19,39 @@ def calc_type(self):
...
+    def get_total_cpu_count(
+        model_default_num_cpus: int,
+        user_specified_num_cpus: Optional[int] = None
+    ):
         if user_specified_num_cpus is not None:
-            total_cpu = get_cpu_count() if num_cpus_available is None else min(get_cpu_count(), num_cpus_available)
-            num_cpus = min(user_specified_num_cpus, total_cpu)
+            num_cpus = min(user_specified_num_cpus, get_cpu_count())
+            if num_cpus < user_specified_num_cpus:
+                logger.warning(f'Specified num_cpus = {user_specified_num_cpus}, but only {num_cpus} are available')
```
Innixma, line 52: ditto

core/src/autogluon/core/hpo/executors.py
```diff
@@ -77,18 +87,121 @@ def register_resources(self, initialized_model: AbstractModel):
...
+                assert user_specified_trial_num_gpus <= num_gpus, \
+                    f"Detected trial level gpu requirement = {user_specified_trial_num_gpus} > total gpu granted to AG predictor = {num_gpus}"
+                assert user_specified_trial_num_gpus >= minimum_model_num_gpus, \
+                    f"The trial requires minimum gpu {minimum_model_num_gpus}, but you only specified {user_specified_trial_num_gpus}"
+
+                # Custom backend should set its total resource to be resources_per_trial
+                self.hyperparameter_tune_kwargs['resources_per_trial'] = {
+                    'num_cpus': user_specified_trial_num_cpus,
+                    'num_gpus': user_specified_trial_num_gpus
+                }
```
Innixma, line 124: Unsure where self.hyperparameter_tune_kwargs is defined, but just be careful to not be editing outer context. Maybe this is ok, depends on where it is coming from.

core/src/autogluon/core/hpo/executors.py
```diff
@@ -68,7 +71,14 @@ def initialize(
         """
         raise NotImplementedError
     
-    def register_resources(self, initialized_model: AbstractModel):
+    def register_resources(
```
Innixma, line 74: Add to the doc-string now that this is much more complicated

tabular/tests/unittests/resource_allocation/test_total_resource_allocation.py
```diff
@@ -0,0 +1,228 @@
...
+    # Given total resources more than what the system has
+    total_resources = {
+        'num_cpus': 99999,
+        'num_gpus': 99999,
+    }
+    resources = bagged_model._preprocess_fit_resources(total_resources=total_resources)
+    assert resources == {'num_cpus': get_cpu_count(), 'num_gpus': get_gpu_count_all()}
+    
+
+def test_bagged_model_with_total_resources_and_ensemble_resources():
```
Innixma, line 51: I'm assuming these tests fail if a GPU isn't present.

Have you considered overwriting `get_gpu_count_all()` for this unit test so it has no ties to the actual compute of the instance running the test? Not sure exactly the right way to do this, but Alex Shirkov would probably know.

Ditto for `get_cpu_count()`

Innixma (commented):
Added initial review after looking through all files except unit tests, as I first want to check if unit tests can be ran without needing a GPU on the system.

core/src/autogluon/core/models/ensemble/fold_fitting_strategy.py
yinweisu: Just to make sure that it will use `self.num_cpus` or `self.num_gpus` if user didn't specify resources per job

core/src/autogluon/core/hpo/executors.py
yinweisu: Good point. Will double check

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -538,6 +540,10 @@ def fit(self,
             If True, a WeightedEnsembleModel will be fit in each stack layer.
             A weighted ensemble will often be stronger than an individual model while being very fast to train.
             It is recommended to keep this value set to True to maximize predictive quality.
+        num_cpus: int, default = auto
+            The total amount of cpus you want AutoGluon predictor to use.
+        num_gpus: int, default = auto
+            The total amount of gpus you want AutoGluon predictor to use.
```
Innixma, line 546: Explain auto,

`"auto"` instead of `auto`

Tell the user that they generally shouldn't need to set this value.

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -946,6 +958,10 @@ def fit_extra(self,
...
```
Innixma, line 964: ditto

tabular/tests/unittests/resource_allocation/test_total_resource_allocation.py
```diff
@@ -0,0 +1,232 @@
...
+class DummyModel(DummyBaseModel):
+    pass
+
+    
+class DummyBaggedModel(BaggedEnsembleModel):
+    pass
+
+
+get_cpu_count.mock = 16
+get_gpu_count_all.mock = 2
```
Innixma, line 33: 1. Will this alter other tests?
2. Assign to variables: MOCK_CPU_COUNT = 16, MOCK_GPU_COUNT = 2, can re-use in tests

tabular/tests/unittests/resource_allocation/test_total_resource_allocation.py
```diff
@@ -0,0 +1,232 @@
...
+
+class DummyModel(DummyBaseModel):
+    pass
+
+    
+class DummyBaggedModel(BaggedEnsembleModel):
+    pass
+
+
+get_cpu_count.mock = 16
```
Innixma, line 32: Add test that mock works

core/src/autogluon/core/hpo/executors.py
```diff
@@ -68,27 +71,148 @@ def initialize(
...
         ----------
         initialized_model
-            The model that will be performed HPO. This model MUST be initialized
+            The model that will be performed HPO. This model MUST be initialized.
+        num_cpus
+            Total number of cpus available for the experiment.
+        num_gpus
+            Total number of gpus available for the experiment.
+        kfold
+            Number of folds if bagging. Used to check if an individual trial is a bagged model.
```
Innixma, line 99: missing kwargs in docstring

core/src/autogluon/core/hpo/executors.py
```diff
@@ -68,27 +71,148 @@ def initialize(
...
         Parameters
         ----------
         initialized_model
-            The model that will be performed HPO. This model MUST be initialized
+            The model that will be performed HPO. This model MUST be initialized.
+        num_cpus
+            Total number of cpus available for the experiment.
+        num_gpus
+            Total number of gpus available for the experiment.
+        kfold
```
Innixma, line 98: typo: k_fold

tabular/tests/unittests/resource_allocation/test_total_resource_allocation.py
yinweisu: I've added a context manager within conftest and fixtures to pass in the context manager and mock variables to tests

PR #2266, opened 2022-10-28
Use regression problem type inside AutoGluonTabular forecasting model

shchur:
*Description of changes:*
- This PR significantly speeds up training and prediction time for `AutoGluonTabularModel`. Internally, we now treat forecasting as a regression problem, and estimate the quantiles by fitting a zero-mean normal distribution to the residuals.
- Besides being much faster, the new model achieves significantly improved predictive accuracy of point forecasts on some datasets (e.g., `walmart`).

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/autogluon_tabular/tabular_model.py
+++ b/timeseries/src/autogluon/timeseries/models/autogluon_tabular/tabular_model.py
@@ -4,6 +4,7 @@ from typing import Any, Callable, Dict, List, Optional, Tuple
 
 import numpy as np
 import pandas as pd
+import scipy.stats
 
 # TODO: Drop GluonTS dependency
 from gluonts.time_feature import get_lags_for_frequency, time_features_from_frequency_str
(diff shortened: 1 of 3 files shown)


timeseries/src/autogluon/timeseries/models/presets.py
```diff
@@ -40,20 +40,20 @@
...
-    MQRNN=40,
-    SimpleFeedForward=50,
-    Transformer=40,
-    TemporalFusionTransformer=45,
-    DeepAR=50,
+    MQCNN=30,
+    MQRNN=30,
+    SimpleFeedForward=45,
+    Transformer=30,
+    TemporalFusionTransformer=20,
```
canerturkmen, line 47: why is TFT lower than Transformer?

timeseries/src/autogluon/timeseries/models/autogluon_tabular/tabular_model.py
```diff
@@ -183,29 +191,21 @@ def extend_single_time_series(group):
...
 
+        predictions = predictions.rename("mean").to_frame()
         preds_index = data_extended.slice_by_timestep(-self.prediction_length, None).index
         predictions.set_index(preds_index, inplace=True)
 
-        predictions["mean"] = predictions[0.5]
-        predictions = predictions[["mean"] + output_quantiles]
-        predictions.rename(str, axis=1, inplace=True)
+        for q in quantile_levels:
+            predictions[str(q)] = predictions["mean"] + self.residuals_std * scipy.stats.norm.ppf(q)
```
canerturkmen, line 208: this is very cool! this also means that with very minimal effort, thanks to `scipy` we can generalize the residual distributions, which probably will bring a huge boost to performance.

canerturkmen (approved):
LGTM. Ready to merge. In the tests: the AGT model has also been flaking a lot for me in HPO. We can maybe disable that or just disable Ray for it.

timeseries/src/autogluon/timeseries/models/presets.py
shchur: TFT takes much longer to train, so I lowered it priority. I would anyway double-check the presets later, the purpose of the changes here was to "squeeze in" the AG-T model between the faster & slower methods.

Issue #1706, opened 2022-05-05
Replace string concatenation of path to os.path.join


yinweisu:
Currently, we use string concatenation and os.path.sep many places instead of doing os.path.join. This could be error prone.

PR #2288, opened 2022-11-02
Replace string concatenation of path to os.path.join

gradientsky:
*Issue #, if available:*
#1706 

*Description of changes:*

### DIFF
--- a/cloud/src/autogluon/cloud/predictor/cloud_predictor.py
+++ b/cloud/src/autogluon/cloud/predictor/cloud_predictor.py
@@ -221,7 +221,7 @@ class CloudPredictor(ABC):
         if path is None:
             utcnow = datetime.utcnow()
             timestamp = utcnow.strftime("%Y%m%d_%H%M%S")
-            path = f'AutogluonCloudPredictor/ag-{timestamp}{os.path.sep}'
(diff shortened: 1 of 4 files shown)


cloud/src/autogluon/cloud/predictor/cloud_predictor.py
```diff
@@ -221,7 +221,7 @@ def _setup_local_output_path(self, path):
         if path is None:
             utcnow = datetime.utcnow()
             timestamp = utcnow.strftime("%Y%m%d_%H%M%S")
-            path = f'AutogluonCloudPredictor/ag-{timestamp}{os.path.sep}'
```
tonyhoo, line 224: wondering how you capture this, is there any IDE hint?

tonyhoo (approved):

gidler (approved):

cloud/src/autogluon/cloud/predictor/cloud_predictor.py
yinweisu: Wondering it too

gradientsky: I was just using `os.path.sep` and going over these

PR #2291, opened 2022-11-02
[timeseries] Increase the number of samples used by DeepAR at prediction time

canerturkmen:
*Issue #, if available:*

set DeepAR to 250 samples (2.2x slowdown in benchmarks)

*Description of changes:*

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/gluonts/torch/models.py
+++ b/timeseries/src/autogluon/timeseries/models/gluonts/torch/models.py
@@ -162,6 +162,7 @@ class DeepARModel(AbstractGluonTSPyTorchModel):
     """
 
     gluonts_estimator_class: Type[GluonTSPyTorchLightningEstimator] = DeepAREstimator
+    default_num_samples: int = 250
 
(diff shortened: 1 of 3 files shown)


timeseries/src/autogluon/timeseries/models/gluonts/torch/models.py
```diff
@@ -162,6 +162,7 @@ class DeepARModel(AbstractGluonTSPyTorchModel):
     """
 
     gluonts_estimator_class: Type[GluonTSPyTorchLightningEstimator] = DeepAREstimator
+    default_num_samples: int = 250
```
shchur, line 165: Just to double-check, the increased number here leads to a lower variance of the forecasts, right?

shchur (approved):
LGTM!

timeseries/src/autogluon/timeseries/models/gluonts/torch/models.py
canerturkmen: by almost a perfect factor of 2.5 👍

PR #2293, opened 2022-11-02
[timeseries] Make WeightedEnsemble robust to base model failures

shchur:
*Description of changes:*
- Currently, if one of the base models for the ensemble fails, the ensemble can `predict` but cannot `score`. This PR fixes this problem and adds tests covering such cases.

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/ensemble/greedy_ensemble.py
+++ b/timeseries/src/autogluon/timeseries/models/ensemble/greedy_ensemble.py
@@ -119,7 +119,7 @@ class TimeSeriesEnsembleWrapper(AbstractTimeSeriesModel):
         if len(failed_models) > 0:
             logger.warning(
                 f"Following models failed during prediction: {failed_models}. "
-                f"{self.name} will ignore these models and re-normalize the weights to sum up to 1 when predicting."
+                f"{self.name} will set the weight of these models to zero and re-normalize the weights when predicting."
             )
 
         return self.weighted_ensemble.predict([data[k] for k in self.model_names])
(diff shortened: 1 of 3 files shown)


Innixma:
The premise here sounds spooky to me. Why are our base models failing at inference time? This is a big deal. If they didn't fail during training / validation, they shouldn't fail during test. If they do, that is a bug right?

shchur:
For local time series models like ARIMA, we need to fit a separate model to each time series in the data set. Usually the test set contains time series that we didn't see during training, so we need to refit all local models to them. ARIMA sometimes fails during fitting, so with this PR we are guaranteed to give a prediction to the user even if the ensemble relies on models that didn't fail during training but failed during prediction.

shchur:
There is no way to know in advance if the model will fail at test time but luckily, these failures only happen to ARIMA on a small number of benchmark datasets. However, ARIMA is very popular and works well on other datasets, so removing it from our model zoo isn't the best option.

canerturkmen (approved):
LGTM in principle. I think this is a good and graceful way to handle this. 

To Nick's question, yes the reason this happens is that ARIMA is effectively refit at prediction time and the reason we include it in the ensemble is we "hope" that the performance on the prediction set will be the same (thus will continue to hold to same vote in the ensemble) as in the training set. The model can validly fail to fit during predict time: this is not a bug. But it can be an argument that ARIMA should not be included in the ensemble at all. I believe there are more benefits to including it for now, but this consideration will definitely come up again in stack ensembling.

I'm approving for now but let's keep discussing.

shchur:
Another solution would be to catch these failures inside ARIMA and fall back to some simple model like Naive. This would also better handle cases when ARIMA only fails for a single time series but works fine for others. However, this would be a major change to the code that I would rather do in the next release.

Innixma:
Ok, I understand the reasoning and will let this PR move forward. If we haven't already, I'd recommend making it very clear to the user when this happens and what we are doing to make it not crash the process and instead alter the weights of the weighted ensemble (and what those weights alter to perhaps).

PR #2294, opened 2022-11-02
[timeseries] Use mini-batch prediction for tabular forecasting model

shchur:
*Description of changes:*
- Remove memory (and disk) heavy models RF, XT
- Mostly use the RMSE metric as it doesn't have a failure mode for datasets with target close to 0 (as MAPE)
- Limit the training set size
- Use batched prediction to limit memory usage

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/autogluon_tabular/tabular_model.py
+++ b/timeseries/src/autogluon/timeseries/models/autogluon_tabular/tabular_model.py
@@ -1,5 +1,6 @@
 import logging
 import re
+import time
 from typing import Any, Callable, Dict, List, Optional, Tuple
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


timeseries/src/autogluon/timeseries/models/autogluon_tabular/tabular_model.py
```diff
@@ -27,25 +28,28 @@ class AutoGluonTabularModel(AbstractTimeSeriesModel):
 
     Other Parameters
     ----------------
+    max_train_size : int, default = 1_000_000
```
canerturkmen, line 31: is this parameter handled anywhere?

timeseries/src/autogluon/timeseries/models/autogluon_tabular/tabular_model.py
```diff
@@ -27,25 +28,28 @@ class AutoGluonTabularModel(AbstractTimeSeriesModel):
...
+        Maximum number of rows in the training and validation sets. If the number of rows in train or validation data
+        exceeds ``max_train_size``, then ``max_train_size`` many rows are subsampled from the dataframe.
     tabular_hyperparmeters : Dict[Dict[str, Any]], optional
         Hyperparameters dictionary passed to `TabularPredictor.fit`. Contains the names of models that should be fit.
         Defaults to ``AutoGluonTabularModel.default_tabular_hyperparameters``.
         Note that the selected models must support ``problem_type="quantile"``.
```
canerturkmen: is this current? (models must support quantile)

canerturkmen (commented):
LGTM in code, but we appear to have a related test fail.

canerturkmen (approved):
LGTM

PR #2295, opened 2022-11-02
Remove pytrec_eval and rewrite retrieval metrics

cheungdaven:
*Description of changes:*

We removed the dependency on pytrec_eval and rewrite the information retrieval metrics including: precision, recall, ndcg, map, and mrr, with given cutoff values.

### DIFF
--- a/multimodal/src/autogluon/multimodal/utils/metric.py
+++ b/multimodal/src/autogluon/multimodal/utils/metric.py
@@ -1,9 +1,11 @@
 import logging
+import math
+import operator
 import warnings
 from typing import Dict, List, Optional, Tuple, Union
 
 import evaluate
-import pytrec_eval
+import numpy as np
 from sklearn.metrics import f1_score
 
 from autogluon.core.metrics import get_metric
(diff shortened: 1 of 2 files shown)


multimodal/src/autogluon/multimodal/utils/metric.py
```diff
@@ -178,6 +180,171 @@ def compute_score(
...
+            # compute the discounted cumulative gain
+            dcg = np.sum([1 / math.log(rank + 2, 2) for rank in hit_rank])
+
+        # compute the ideal discounted cumulative gain
+        idcg = np.sum([1 / math.log(i + 2, 2) for i in range(at_k)])
+        # compute the normalized discounted cumulative gain
+        ndcg = dcg / idcg
+        mAP /= at_k
+
+        return precision, recall, mrr, mAP, ndcg
```
sxjscience, line 345: Shall we use dcg_score in sklearn? https://scikit-learn.org/stable/modules/generated/sklearn.metrics.dcg_score.html

cheungdaven: dcg_score in sklearn is for classification problems. it requires the input to be "(n_samples, n_labels)" where n_labels is the number of labels, while in ranking tasks, there are no labels and the relevant score are continuous values like cosine similarity.  Also, sklearn requires y_true and y_pred to have the exact same shape, while in our task, we support the same input type as pytrec_eval to ensure minimum code refactor effect in other parts.

multimodal/src/autogluon/multimodal/utils/metric.py
```diff
@@ -178,6 +180,171 @@ def compute_score(
         return metric._sign * metric(metric_data[Y_TRUE], metric_data[Y_PRED])
 
 
+class RankingMetrics:
+    def __init__(
+        self,
+        pred: Dict[str, Dict],
+        target: Dict[str, Dict],
+        is_higher_better=True,
```
FANGAreNotGnu, line 188: To be consistent with our previous code, here we can call it "min_max_mode". I think "is_higher_better" is easier to understand though.

FANGAreNotGnu (approved):
LGTM with a minor comment, thanks!

multimodal/src/autogluon/multimodal/utils/metric.py
cheungdaven: min_max_mode is used for metrics, here is_higher_better is used to indicate how we should rank the list. They are essentially different

sxjscience (approved):
LGTM

zhiqiangdon (commented):
How could we test the implementations?

cheungdaven:
> How could we test the implementations?

I compared the results with pytrec_eval locally. 

For online test, it will be tested in the test cases where you use the "compute_ranking_score" function you added.

zhiqiangdon (approved):
LGTM! Thanks for the implementations!

PR #2304, opened 2022-11-04
[timeseries] Update presets for TimeSeriesPredictor

shchur:
*Description of changes:*
- Change model priority to prefer more accurate / faster-to-train models
- Add `high_quality` presents offering a middle ground between `medium_quality` and `best_quality`
- Change default `batch_size` for GluonTS model to 64, as this showed better results in the experiments for most models
- Change presets to lower the training time for all configurations


timeseries/src/autogluon/timeseries/models/presets.py
```diff
@@ -124,46 +144,37 @@ def get_default_hps(key, prediction_length):
...
+            "best_quality": {
+                "TemporalFusionTransformerMXNet": {"context_length": context_length},
+                "TransformerMXNet": {"context_length": context_length},
             },
         }
         for k in default_model_hps:
             default_model_hps[k] = dict(**default_model_hps[k], **mxnet_default_updates.get(k, {}))
 
+    # For backwards compatibility
+    default_model_hps["default"] = default_model_hps["medium_quality"]
```
canerturkmen, line 175: cool

canerturkmen (approved):
LGTM!

PR #2305, opened 2022-11-04
[timeseries] Move statsmodels models to inherit from AbstractLocalModel

shchur:
When running the benchmarks, we noticed that the local models from based on `timeseries.models.statsmodels.AbstractStatsmodelsModel` had extremely long training / prediction times (>2 hours) on datasets with a large number of items (>20K items). This is a critical issue, since some of these datasets (e.g., `m4_yearly`, `m4_quarterly`) are well-established benchmarks that people will likely use to test the library. This PR addresses this problem and reduces the train/test time on these datasets to 1-2 minutes.

We planned to perform the move `AbstractStatsmodelsModel` -> `AbstractLocalModel` after the 0.6 release, but after discussion with @canerturkmen we decided to do it now, as extremely long training times can be critical for the users.

*Description of changes:*
- Move `ARIMA`, `ETS` and `Theta` models to inherit from `timeseries.models.local.AbstractLocalModel`. The change was really straightforward, since both `AbstractLocalModel` and `AbstractStatsmodelsModel` follow the same high-level logic, except that `AbstractLocalModel` caches the  predictions (not learned models) and doesn't access class attributes during prediction.
- The new versions of  `ARIMA`, `ETS` and `Theta` have been tested against the old versions. Both versions produce identical results. We will re-run the benchmark to make sure there are no regressions.
- The tests from `timeseries/tests/unittests/models/test_local.py` exactly replicate the logic from `timeseries/tests/unittests/models/test_statsmodels.py`, so no new tests had to be written.


timeseries/src/autogluon/timeseries/models/local/statsmodels.py
```diff
@@ -54,74 +82,71 @@ class ETSModel(AbstractStatsmodelsModel):
...
+        seasonal = local_model_args.setdefault("seasonal", "add")
         if seasonal is not None and seasonal_period <= 1:
             logger.warning(
                 f"{self.name} with seasonal = {seasonal} requires seasonal_period > 1 "
                 f"(received seasonal_period = {seasonal_period}). Disabling seasonality."
             )
-            sm_model_init_args["seasonal"] = None
-            sm_model_init_args["seasonal_periods"] = 1
+            local_model_args["seasonal"] = None
+            local_model_args["seasonal_periods"] = 1
```
canerturkmen, line 108: `seasonal_period` and `seasonal_periods` look a bit dangerous. Should we at least mark it with a TODO to rename here?

timeseries/src/autogluon/timeseries/models/local/statsmodels.py
```diff
@@ -54,74 +82,71 @@ class ETSModel(AbstractStatsmodelsModel):
...
+        freq: str,
+        prediction_length: int,
+        quantile_levels: List[float],
+        local_model_args: dict,
+        **kwargs,
+    ) -> pd.DataFrame:
+        forecast_timestamps = get_forecast_horizon_index_single_time_series(
+            past_timestamps=time_series.index, freq=freq, prediction_length=prediction_length
+        )
+        from statsmodels.tsa.exponential_smoothing.ets import ETSModel
```
canerturkmen, line 126: any particular reason we move the imports in?

timeseries/src/autogluon/timeseries/models/local/statsmodels.py
```diff
@@ -261,55 +287,49 @@ class ThetaModel(AbstractStatsmodelsModel):
...
-        # so we delegate training to `_predict_with_local_model`
-        if sm_model_init_args["deseasonalize"] and len(timeseries) < 2 * sm_model_init_args["period"]:
-            sm_model_init_args = sm_model_init_args.copy()
-            sm_model_init_args["deseasonalize"] = False
-        return FittedLocalModel(model_name=self.name, sm_model_init_args=sm_model_init_args, parameters=None)
 
+    def _update_local_model_args(
+        self, local_model_args: Dict[str, Any], data: TimeSeriesDataFrame, **kwargs
+    ) -> Dict[str, Any]:
+        local_model_args.setdefault("deseasonalize", True)
```
canerturkmen, line 301: have we decided dropping the copy operations is safe, or does the abstract model handle these?

canerturkmen (approved):
LGTM, only minor comments.

canerturkmen:
also @Innixma FYI, this is a 100x speedup on some critical large benchmark datasets for some models, important for us to get in.

timeseries/src/autogluon/timeseries/models/statsmodels/__init__.py
```diff
@@ -1,15 +0,0 @@
-import warnings
-
-import statsmodels.tools.sm_exceptions
-
-from .models import ARIMAModel, ETSModel, ThetaModel
-
-warnings.simplefilter("ignore", statsmodels.tools.sm_exceptions.ModelWarning)
```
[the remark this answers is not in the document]
shchur: Good point! Moved the filter to `models/local/statsmodels.py`.

timeseries/src/autogluon/timeseries/models/local/statsmodels.py
shchur: Every method in `statsmodels` has a different name for this. We chose to always expose `seasonal_period` to the user in our API, and internally map this to the model's parameter (`seasonal_periods`, `period` or an entry in `seasonal_order` tuple). Not sure if we can change this. The name `seasonal_period` is also used in our 0.5.2 release and in Darts, so I'm not included to change this.

timeseries/src/autogluon/timeseries/models/local/statsmodels.py
shchur: Will make our life easier should we move to soft dependency on statsmodels? But can also be placed outside. What do you think?

timeseries/src/autogluon/timeseries/models/local/statsmodels.py
shchur: Yes, `_predict_with_local_model` always receives a copy of the `_local_model_args` dict. 

We're fine with modifying the dict inside `_update_local_model_args`, this only happens once and doesn't affect the `self._get_model_params()`.

PR #2308, opened 2022-11-04
Disable parallel bagging with AutoMM

Innixma:
*Issue #, if available:*

*Description of changes:*
- Disable parallel bagging with AutoMM
- It appears that parallel bagging with AutoMM hangs the process. Reverting to using sequential bagging which does work correctly.


sxjscience (approved):
LGTM. Let me investigate later after 0.6.

PR #2309, opened 2022-11-04
Fix AutoMM Tabular altering outer context

Innixma:
*Issue #, if available:*

*Description of changes:*
- Fixed a major bug causing the outer context to be altered when Tabular AutoMM model concatenates the label to the features. This caused future models to either crash or have the label in their training and validation data features.

### DIFF
--- a/tabular/src/autogluon/tabular/models/automm/automm_model.py
+++ b/tabular/src/autogluon/tabular/models/automm/automm_model.py
@@ -1,9 +1,11 @@
 """Wrapper of the MultiModalPredictor."""
-from typing import Dict, Optional
+
+import copy
 import logging
 import os
-import pandas as pd
+from typing import Dict, Optional
 
+import pandas as pd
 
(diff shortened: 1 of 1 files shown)


tabular/src/autogluon/tabular/models/automm/automm_model.py
```diff
@@ -127,8 +129,11 @@ def _fit(self,
             logger.log(15, "sample_weight not yet supported for MultiModalPredictorModel, "
                            "this model will ignore them in training.")
 
+        # Need to deep copy to avoid altering outer context
+        X_train = copy.deepcopy(X_train)
```
sxjscience, line 133: We can use `X_train.copy()`. But the current logic looks good.

sxjscience (approved):
LGTM

tabular/src/autogluon/tabular/models/automm/automm_model.py
Innixma: good point, updated

PR #2314, opened 2022-11-05
[timeseries] Save lightning_logs to model directory when training PyTorch models & clean up after training

shchur:
*Description of changes:*
- We save the lightning logs `self.path / "lightning_logs"`. This folder gets deleted up after training. Now training PyTorch models doesn't produce a `lightning_logs` folder in the directory from where the script was run. 
- Perform all hyperparameter updates for all models in `_get_estimator_init_args` (some models did this inside `_get_estimator`).

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
+++ b/timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
@@ -1,6 +1,7 @@
 import logging
 import re
 from pathlib import Path
+from tempfile import TemporaryDirectory
 from typing import Any, Callable, Dict, Iterator, List, Optional, Type
 
 import gluonts
@@ -226,7 +227,7 @@ class AbstractGluonTSModel(AbstractTimeSeriesModel):
         `self._get_model_params` for better readability."""
         return self._get_model_params()
 
-    def _get_estimator(self) -> GluonTSEstimator:
+    def _get_estimator(self, **kwargs) -> GluonTSEstimator:
         """Return the GluonTS Estimator object for the model"""
         with warning_filter():
             return self.gluonts_estimator_class.from_hyperparameters(**self._get_estimator_init_args())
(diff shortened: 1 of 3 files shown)


timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
```diff
@@ -310,12 +311,13 @@ def _fit(
...
         )
 
-        estimator = self._get_estimator()
-        with warning_filter(), disable_root_logger(), gluonts.core.settings.let(gluonts.env.env, use_tqdm=False):
-            self.gts_predictor = estimator.train(
-                self._to_gluonts_dataset(train_data),
-                validation_data=self._to_gluonts_dataset(val_data),
-            )
+        with TemporaryDirectory() as temp_dir:
+            estimator = self._get_estimator(root_dir=temp_dir)
```
canerturkmen, line 315: I understand the tempdir is creeping all the way up here (for all gluonts models instead of just torch) because the train call is here. Is there another way around this?

timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
```diff
@@ -226,7 +227,7 @@ def _get_estimator_init_args(self) -> Dict[str, Any]:
         `self._get_model_params` for better readability."""
         return self._get_model_params()
 
-    def _get_estimator(self) -> GluonTSEstimator:
```
canerturkmen, line 229: Isn't this too big a change for the task we're solving? We already have a method for getting init_args for estimators, why should we need to pass other args into this method, and especially for all models?

timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
```diff
@@ -1,6 +1,7 @@
 import logging
 import re
 from pathlib import Path
+from tempfile import TemporaryDirectory
```
canerturkmen, line 4: On second thought I think we had troubles with temp directories before, as possibly not all platforms treat them equally. Can we instead bury lightning logs in the model path? Are they ever large enough on disk that cleaning them up is a concern from disk size perspective?

timeseries/src/autogluon/timeseries/models/gluonts/mx/models.py
```diff
@@ -112,12 +112,11 @@ class AbstractGluonTSSeq2SeqModel(AbstractGluonTSMXNetModel):
 
     gluonts_estimator_class: Type[GluonTSEstimator] = None
 
-    def _get_estimator(self):
+    def _get_estimator_init_args(self):
+        init_kwargs = super()._get_estimator_init_args()
```
canerturkmen, line 116: thanks for the cleanup! 🪣

timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
shchur: Thanks for the comments, I agree will all of them. Now the `lightning_logs` are saved to the model directory and are deleted after the training is finished (we don't use the checkpoint, GluonTS gives us the trained model's weights). Does this look better to you?

canerturkmen (approved):
Very cool, LGTM!

PR #2316, opened 2022-11-06
Add AutoMM backend to ag.vision

bryanyzhu:
*Issue #, if available:*
NA

*Description of changes:*
1. Given deprecation of ag.vision in v0.7, we would like to give users the option to try MultiModalPredictor in ag.vision first. Simply call `predictor = ImagePredictor(backend='automm')` will do the trick.
2. Add fit_summary() to be consistent with tabular and timeseries predictor
3. Add list_supported_models() to let users know which checkpoint name to use in the model pool


bryanyzhu:
This PR is working at this moment, @sxjscience @zhiqiangdon @FANGAreNotGnu , let me know if you think this solution is ok. Then I will work on adding backend for object detector in ag.vision.

sxjscience (dismissed):
LGTM

vision/src/autogluon/vision/predictor/predictor.py
```diff
@@ -55,20 +59,28 @@ class ImagePredictor(object):
...
+        logger.warning(
             f"AutoGluon ImagePredictor will be deprecated in v0.7. "
             f"Please use AutoGluon MultiModalPredictor instead for more functionalities and better support. "
-            f"Visit https://auto.gluon.ai/stable/tutorials/multimodal/index.html for more details!",
-            DeprecationWarning,
+            f"Visit https://auto.gluon.ai/stable/tutorials/multimodal/index.html for more details! "
         )
+
+        if backend == 'automm':
+            self._classifier = MultiModalPredictor(label=label, problem_type=problem_type, eval_metric=eval_metric,
```
FANGAreNotGnu, line 71: Here checkpoint_name is not specified, also we shall set `pipeline` to `OBJECT_DETECTION` (or `problem_type` if we want to remove `pipeline`)

vision/src/autogluon/vision/predictor/predictor.py
```diff
@@ -541,6 +561,11 @@ def predict_proba(self, data, as_pandas=True):
         """
         if self._classifier is None:
             raise RuntimeError('Classifier is not initialized, try `fit` first.')
+        
+        if self._backend == 'automm':
+            prob = self._classifier.predict_proba(data=data, as_pandas=as_pandas)
```
FANGAreNotGnu, line 566: `predict_proba` is not implemented for OBJECT_DETECTION. What is expected to return for this function in detection task?

bryanyzhu: We simply return a runtime error message saying that AutoMM doesn't support this functionality yet. Actually for image classification, we can't support every feature at this moment either.

vision/src/autogluon/vision/predictor/predictor.py
bryanyzhu: checkpoint_name will be provided when calling `fit`. Here is just initializing the predictor. 

For object detection, we are not using this predictor.py file. There is another file in ag.vision called detector.py (https://github.com/awslabs/autogluon/blob/master/vision/src/autogluon/vision/detector/detector.py).

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -1457,6 +1462,7 @@ def _top_k_average(
                             strict=strict_loading,
                         )
                         best_score = self.evaluate(val_df, [validation_metric_name])[validation_metric_name]
+                        self.best_score = best_score
```
zhiqiangdon, line 1465: The self.best_score here is within the greedy soup, using which would cause error for uniform_soup or best.

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -1369,11 +1371,14 @@ def _fit(
                 ".* in the `DataLoader` init to improve performance.*",
             )
             warnings.filterwarnings("ignore", "Checkpoint directory .* exists and is not empty.")
+            training_start = time.time()
```
zhiqiangdon, line 1374: This only counts the lightning trainer's fitting time. Should we calculate predictor's fitting time?

multimodal/src/autogluon/multimodal/predictor.py
bryanyzhu: @zhiqiangdon Then where I can find the best validation score? For example, if using `best`, where is the val_accuracy located? Thank you.

multimodal/src/autogluon/multimodal/predictor.py
bryanyzhu: Sure, we can do that. So how about I put the `training_start = time.time()` in the very beginning of fit()? like here https://github.com/awslabs/autogluon/blob/master/multimodal/src/autogluon/multimodal/predictor.py#L537

And then put the end_time at here? 
https://github.com/awslabs/autogluon/blob/master/multimodal/src/autogluon/multimodal/predictor.py#L729

multimodal/src/autogluon/multimodal/predictor.py
zhiqiangdon: I think we can add
```self.best_score = self.evaluate(val_df, [validation_metric_name])[validation_metric_name]```
at line 1508 (https://github.com/awslabs/autogluon/pull/2316/files#diff-a73844176f6eae3b4538467b3565e3bdbf4d88997c3c670173bbb44e1a55b7a3R1508).

Comment (zhiqiangdon) [reply multimodal/src/autogluon/multimodal/predictor.py:1465]: Also, self.best_score hasn't considered matcher yet. Calling the fit_summary can cause errors for matcher.

File: multimodal/src/autogluon/multimodal/predictor.py (hunk 4 above)
Comment (zhiqiangdon) [reply multimodal/src/autogluon/multimodal/predictor.py:1374]: Sure.

Comment (bryanyzhu) [reply multimodal/src/autogluon/multimodal/predictor.py]: Done

File: multimodal/src/autogluon/multimodal/predictor.py (hunk 3 above)
Comment (bryanyzhu) [reply multimodal/src/autogluon/multimodal/predictor.py]: As discussed, add another line at 1508, self.best_score = self.evaluate(val_df, [validation_metric_name])[validation_metric_name]. 

But for the matcher issue, how should I address it? Any particular metric to get and where? Thank you.

File: vision/src/autogluon/vision/predictor/predictor.py (hunk 1 above)
Comment (FANGAreNotGnu) [reply vision/src/autogluon/vision/predictor/predictor.py:71]: I see, but if we use automm as backend, in `MultiModalPredictor` the pipeline flag is used, and the `MMDetAutoModelForObjectDetection` (in mmdet_image.py) is initialized in `MultiModalPredictor.__init__`.

Comment (FANGAreNotGnu) [reply vision/src/autogluon/vision/predictor/predictor.py:71]: I might have a misunderstanding. Is this only for timm and not for mmdet in 0.6?

File: multimodal/src/autogluon/multimodal/predictor.py (hunk 3 above)
Comment (zhiqiangdon) [reply multimodal/src/autogluon/multimodal/predictor.py]: I think you need to add a similar line in matcher's top_k_average to compute best score and add a property in predictor to check which best_score (predictor's or matcher's) to use.

Comment (bryanyzhu) [reply multimodal/src/autogluon/multimodal/predictor.py]: I see. I think I will just do this for image classification in this PR, and leaves the rest for each feature owner.

File: multimodal/src/autogluon/multimodal/predictor.py [hunk 5]
```diff
@@ -2832,6 +2839,47 @@ def positive_class(self):
...
+            TODO: Higher levels correspond to more detailed print statements
+        show_plot : bool, default = False
+            If True, shows the model summary plot in browser when verbosity > 1.
+
+        Returns
+        -------
+        Dict containing various detailed information.
+        We do not recommend directly printing this dict as it may be very large.
+        """
+        results = {"val_accuracy": self.best_score, "training_time": self.elapsed_time}
```
sxjscience, line 2860: We may change "val_accuracy" to "val_score".

sxjscience: Or we can use

```
f"val_{self._validation_metric_name}"
```

sxjscience (approved):
LGTM

zhiqiangdon (approved):
LGTM!

PR #2321, opened 2022-11-07
Support fallback layer ids

zhiqiangdon:
*Issue #, if available:*

*Description of changes:*

The current assigning layer ids logic can't handle all kinds of backbones. This PR is to support using the uniform layer ids for these backbones.


sxjscience (approved):
LGTM.

PR #2323, opened 2022-11-07
[timeseries] Enable GPU and transform caching for GluonTS models

shchur:
*Description of changes:*
- Cache the transformation results & use GPU acceleration in PyTorch models if `torch.cuda.is_available()`. This leads to much faster training. For example, training time for DeepAR (PyTorch) on the `m4_yearly` dataset on `p3.8xlarge`:
  - CPU, `cache_data=False`: 748s
  - CPU, `cache_data=True`: 340s
  - GPU, `cache_data=True`: 60s
- This doesn't lead to problems with other models. I re-ran the benchmark with this setting, all models complete training without crashing.


canerturkmen (approved):
LGTM

PR #2330, opened 2022-11-08
Resource Manager

yinweisu:
*Issue #, if available:*

*Description of changes:*
* Add resource manager to report sys info for easier mocking in the unit tests.
* Updated unit test mocking method

### DIFF
--- a/core/src/autogluon/core/utils/utils.py
+++ b/core/src/autogluon/core/utils/utils.py
@@ -33,14 +33,20 @@ class ResourceManager():
         return multiprocessing.cpu_count()
     
     @staticmethod
+    def get_cpu_count_psutil(logical=True):
+        return psutil.cpu_count(logical=logical)
+    
+    @staticmethod
     def get_gpu_count_all():
         num_gpus = ResourceManager._get_gpu_count_cuda()
         if num_gpus == 0:
+            # Get num gpus from mxnet first because of https://github.com/awslabs/autogluon/issues/2042
+            # TODO: stop using mxnet to determine num gpus once mxnet is removed from AG
             num_gpus = ResourceManager.get_gpu_count_mxnet()
             if num_gpus == 0:
                 num_gpus = ResourceManager.get_gpu_count_torch()
         return num_gpus
-    
+
     @staticmethod
     def get_gpu_count_mxnet():
         # TODO: Remove this once AG get rid off mxnet
--- a/core/src/autogluon/core/utils/utils.py
+++ b/core/src/autogluon/core/utils/utils.py
@@ -21,86 +21,89 @@ from .miscs import warning_filter
 from ..constants import BINARY, REGRESSION, MULTICLASS, SOFTCLASS, QUANTILE
 from ..metrics import accuracy, root_mean_squared_error, pinball_loss, Scorer
 
+
 logger = logging.getLogger(__name__)
 
 
-
-# TODO: RespirceManager to replace these two hacks
-class _get_cpu_count:
-    def __init__(self):
-        self.mock = None
-
-    def __call__(self):
-        if self.mock is not None:
-            return self.mock
+class ResourceManager():
+    """Manager that fetches system related info"""
+    
+    @staticmethod
+    def get_cpu_count():
         return multiprocessing.cpu_count()
     
-
-class _get_gpu_count_all:
-    """
-    Attempts to get number of GPUs available for use via multiple means.
-    """
-    # FIXME: update to use only torch for TIMM or find a better GPU detection strategy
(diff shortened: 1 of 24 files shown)


core/src/autogluon/core/utils/utils.py
```diff
@@ -21,86 +21,89 @@
...
-        num_gpus = _get_gpu_count_cuda()
+    @staticmethod
+    def get_gpu_count_all():
+        num_gpus = ResourceManager._get_gpu_count_cuda()
         if num_gpus == 0:
-            num_gpus = get_gpu_count_mxnet()
+            num_gpus = ResourceManager.get_gpu_count_mxnet()
             if num_gpus == 0:
-                num_gpus = get_gpu_count_torch()
+                num_gpus = ResourceManager.get_gpu_count_torch()
```
Innixma, line 41: Tiny edit: We should actually be calling `get_gpu_count_torch` before we call `get_gpu_count_mxnet` now that Torch is an expected dependency and MXNet is being deprecated. This lowers the risk of MXNet import occurring when it isn't actually being used, which could cause issues.

gradientsky (approved):

core/src/autogluon/core/utils/utils.py
yinweisu: Yea, I was thinking the same. Let me update it

gradientsky: @Innixma be aware of this thing https://github.com/awslabs/autogluon/issues/2042

core/src/autogluon/core/utils/utils.py
```diff
@@ -21,86 +21,89 @@
...
-        self.mock = None
 
-    def __call__(self):
-        if self.mock is not None:
-            return self.mock
+class ResourceManager():
+    """Manager that fetches system related info"""
+    
+    @staticmethod
+    def get_cpu_count():
```
Innixma, line 32: In several areas of the code we actually use psutil to get num cpus, lets try moving these usages to `def get_cpu_count_psutil(logical=True)` in ResourceManager so it is easier to track.

Example in LightGBM:

```
num_cpus = psutil.cpu_count(logical=False)
```

core/src/autogluon/core/utils/utils.py
Innixma: Oh good point @gradientsky . Then we can keep the import order as is for now, but add a code comment about this specific issue so we are aware of it when viewing the code.

core/src/autogluon/core/utils/utils.py
```diff
@@ -21,86 +21,95 @@
...
-    # FIXME: get_gpu_count by itself doesn't always work for Windows
-    def __init__(self):
-        self.mock = None
-
-    def __call__(self):
-        if self.mock is not None:
-            return self.mock
-        num_gpus = _get_gpu_count_cuda()
+    @staticmethod
+    def get_cpu_count_psutil(logical=True):
```
tonyhoo, line 36: we don't necessarily need to bind a different impl for the same function. If we see one is more trustworthy than the other, we should swap without impacting the caller

core/src/autogluon/core/utils/utils.py
```diff
@@ -21,86 +21,95 @@
...
+        return psutil.cpu_count(logical=logical)
+    
+    @staticmethod
+    def get_gpu_count_all():
```
tonyhoo, line 40: curious why we suffix the func name with `all`? it would be better to make it consistent with cpu func

yinweisu: I think this function means get gpu count from all possible sources.

core/src/autogluon/core/utils/utils.py
yinweisu: @Innixma Do you remember why we have two ways retrieving num_cpus. Is it because psutil has more overhead but offer the logical option?

Innixma: Keep it for now, but this is a good question and we should figure out if one is strictly superior to the other

core/src/autogluon/core/utils/utils.py
Innixma: For GPU there is not a single source of truth as to how many GPUs there are, as it depends on the package if it can actually detect and leverage the GPUs. We might consider removing the `_all` part of the name, but maybe after v0.6 release.

core/src/autogluon/core/utils/utils.py
Innixma: Ideally if we can just use multiprocessing that would be best since it is native python

PR #2335, opened 2022-11-08
[timeseries] Update log messages

shchur:
*Description of changes:*
- Raise a human-readable exception when calling `leaderboard` if all models failed during training
- Remove redundant log messages

### DIFF
--- a/timeseries/src/autogluon/timeseries/predictor.py
+++ b/timeseries/src/autogluon/timeseries/predictor.py
@@ -424,7 +424,7 @@ class TimeSeriesPredictor:
                 f"Provided tuning data set with {len(tuning_data)} rows, {tuning_data.num_items} items. "
                 f"Average time series length is {len(tuning_data) / tuning_data.num_items:.1f}."
             )
-        logger.info(f"Training artifacts will be saved to: {Path(self.path).resolve()}")
+        logger.info(f"AutoGluon will save models to {Path(self.path).resolve()}")
         logger.info("=====================================================")
 
         if random_seed is not None:
--- a/timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
+++ b/timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
@@ -767,6 +767,9 @@ class AbstractTimeSeriesTrainer(SimpleAbstractTrainer):
                     model_info[model_name]["score_test"] = float("nan")
                     model_info[model_name]["pred_time_test"] = float("nan")
 
+        if len(model_info) == 0:
+            raise ValueError("Trainer has no fit models that can predict.")
+
         df = pd.DataFrame(model_info.values())
 
         sort_column = "score_test" if "score_test" in df.columns else "score_val"
--- a/timeseries/src/autogluon/timeseries/learner.py
+++ b/timeseries/src/autogluon/timeseries/learner.py
@@ -87,7 +87,6 @@ class TimeSeriesLearner(AbstractLearner):
             "Beginning AutoGluon training with TimeSeriesLearner "
             + (f"Time limit = {time_limit}" if time_limit else "")
         )
-        logger.info(f"AutoGluon will save models to {self.path}")
 
         logger.info(f"AutoGluon will gauge predictive performance using evaluation metric: '{self.eval_metric}'")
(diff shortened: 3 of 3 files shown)


timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -767,6 +767,9 @@ def leaderboard(self, data: Optional[TimeSeriesDataFrame] = None) -> pd.DataFram
                     model_info[model_name]["score_test"] = float("nan")
                     model_info[model_name]["pred_time_test"] = float("nan")
 
+        if len(model_info) == 0:
```
canerturkmen, line 770: Let's imagine a scenario where someone is looping through different predictors fit to the same data, and collecting their leaderboards. When asking for this job's leaderboard, why should an exception (to say, something went wrong in calculating the leaderboard -- and especially a ValueError to say something went wrong because the customer provided a wrong input argument to this function) be thrown? Isn't an empty leaderboard the semantically correct response?

I would rather just raise a UserWarning and return the empty dataframe.

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -424,7 +424,7 @@ def fit(
                 f"Provided tuning data set with {len(tuning_data)} rows, {tuning_data.num_items} items. "
                 f"Average time series length is {len(tuning_data) / tuning_data.num_items:.1f}."
             )
-        logger.info(f"Training artifacts will be saved to: {Path(self.path).resolve()}")
```
canerturkmen, line 427: Isn't Autogluon not only saving models, but also saving other training artifacts such as predictor and learner states?

canerturkmen (commented):
Let's discuss whether we should proceed with these changes.

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
shchur: I agree, this is a much more elegant solution. Currently, if no models were trained, user receives following output:
```
  File "/local/home/shchuro/workspace/autogluon/shchuro/train.py", line 43, in <module>
    predictor.leaderboard(d_test)
  File "/local/home/shchuro/workspace/autogluon/timeseries/src/autogluon/timeseries/predictor.py", line 621, in leaderboard
    leaderboard = self._learner.leaderboard(data)
  File "/local/home/shchuro/workspace/autogluon/timeseries/src/autogluon/timeseries/learner.py", line 367, in leaderboard
    return self.load_trainer().leaderboard(data)
  File "/local/home/shchuro/workspace/autogluon/timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py", line 773, in leaderboard
    df.sort_values(by=[sort_column, "model"], ascending=[False, False], inplace=True)
  File "/home/shchuro/miniconda3/envs/ag/lib/python3.9/site-packages/pandas/util/_decorators.py", line 311, in wrapper
    return func(*args, **kwargs)
  File "/home/shchuro/miniconda3/envs/ag/lib/python3.9/site-packages/pandas/core/frame.py", line 6304, in sort_values
    keys = [self._get_label_or_level_values(x, axis=axis) for x in by]
  File "/home/shchuro/miniconda3/envs/ag/lib/python3.9/site-packages/pandas/core/frame.py", line 6304, in <listcomp>
    keys = [self._get_label_or_level_values(x, axis=axis) for x in by]
  File "/home/shchuro/miniconda3/envs/ag/lib/python3.9/site-packages/pandas/core/generic.py", line 1840, in _get_label_or_level_values
    raise KeyError(key)
KeyError: 'score_val'
```
I will update the PR to follow your suggestion!

shchur:
These are minor changes that can be postponed to the next minor version

timeseries/src/autogluon/timeseries/predictor.py
shchur: My goal here was to 
1. Remove a duplicated print statement (the path is printed twice)
2. Make the output of `TimeSeriesPredictor` similar to the output of `TabularPredictor`:
```
AutoGluon will save models to "AutogluonModels/ag-20221108_130039/"
```
How about `AutoGluon will save models and other training artifacts to {...}`?

PR #2337, opened 2022-11-08
Merge pipelines to problem_type

sxjscience:
*Issue #, if available:*

*Description of changes:*

- [x] Deprecate `pipeline` and merge the keys into `problem_type`.
- [x] Fix use-cases in documents


multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -254,7 +304,19 @@ def __init__(
...
         sample_data_path
             This is used for automatically inference num_classes, classes, or label.
+
         """
+        # Parse problem_type to pipeline
+        if problem_type in PROBLEM_TYPES_SUPPORT_INFERENCE:
+            if pipeline is None:
+                pipeline = problem_type
+            else:
+                assert pipeline == problem_type, (
```
zhiqiangdon, line 314: I think this causes some inconsistency between the `fit()` and `init()` APIs. The presets need to be passed to `problem_type` in `init()`, but to `presets` in `fit()`.

sxjscience: We should submit another PR to support `presets` in `__init__()`. The current logic of using presets as pipeline keys in matching should be refactored.

Currently, I'm keeping the `pipeline` design. `pipeline` essentially means that we can call `predictor.predict()` and `predictor.evaluate()` without calling `predictor.fit()`.

Problem type means the predictor support `predictor.fit()`.

zhiqiangdon: I think the current pipeline has the same semantic meaning as `presets`. I guess simply replacing `pipeline` with `presets` should work. No need to keep both `presets` and `pipeline`.

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -153,6 +159,38 @@
...
+@dataclass
+class ProblemProperty:
+    name: str  # Name of the problem
+    support_fit: bool = True  # Whether the problem type support `.fit()`
+    inference_ready: bool = False  # Support `.predict()` and `.evaluate()` without calling `.fit()`
+    is_matching: bool = False  # Whether the problem belongs to the matching category
+    experimental: bool = False  # Indicate whether the problem is experimental
+
+
+problem_property_dict = OrderedDict(
```
sxjscience, line 171: @zhiqiangdon @bryanyzhu I added this problem_property_dict to record certain properties, like whether it supports `.fit()` or support `.predict()` without calling `.fit()`.

sxjscience: Also @FANGAreNotGnu @cheungdaven

bryanyzhu (approved):
LGTM, thank you!

multimodal/src/autogluon/multimodal/data/process_text.py
```diff
@@ -154,10 +154,12 @@ def __init__(
             self.max_len = self.tokenizer.model_max_length
         else:
             if max_len < self.tokenizer.model_max_length:
-                warnings.warn(
-                    f"provided max length: {max_len} "
-                    f"is smaller than {model.checkpoint_name}'s default: {self.tokenizer.model_max_length}"
-                )
+                # TODO, Consider to fix the logic
+                if self.tokenizer.model_max_length < 10**6:
```
tonyhoo, line 158: Wondering why we introduce another check on the constant here?

multimodal/src/autogluon/multimodal/matcher.py
```diff
@@ -129,12 +128,14 @@ def __init__(
...
             Column names of response data. If no label column is provided,
             query and response columns form positive pairs.
-        negative
-            Column names of negative data. Query and negative make up negative pairs.
         label
-            Name of the label column. Label and negative shouldn't be used simultaneously.
+            Name of the label column.
         match_label
-            For binary labels, it is the label indicating the query and response should match.
+            The label class that indicates the <query, response> pair is counted as "match".
```
tonyhoo, line 134: Thanks for polishing this

multimodal/src/autogluon/multimodal/matcher.py
```diff
@@ -595,7 +585,7 @@ def _fit(
         hpo_mode: bool = False,
         **hpo_kwargs,
     ):
-        if presets == None:
+        if presets is None:
```
tonyhoo, line 588: +1

multimodal/src/autogluon/multimodal/presets.py
```diff
@@ -10,7 +10,7 @@
 @automm_presets.register()
 def default():
     return {
-        "model.names": ["categorical_mlp", "numerical_mlp", "timm_image", "hf_text", "clip", "fusion_mlp"],
+        "model.names": ["categorical_mlp", "numerical_mlp", "timm_image", "hf_text", "fusion_mlp"],
```
tonyhoo, line 13: wondering if it is intended to remove `clip` from all presets?

multimodal/src/autogluon/multimodal/utils/metric.py
```diff
@@ -43,9 +43,9 @@
 
 def infer_metrics(
     problem_type: Optional[str] = None,
-    pipeline: Optional[str] = None,
     eval_metric_name: Optional[str] = None,
     validation_metric_name: Optional[str] = None,
+    pipeline: Optional[str] = None,
```
tonyhoo, line 48: shall we remove this here as well?

multimodal/src/autogluon/multimodal/data/process_text.py
zhiqiangdon, line 158: Why do we need to add this if condition?

sxjscience: In some models (like T5 / deberta-v3), the model_max_length is really large so I added the condition here to avoid popping up too many warning messages.

multimodal/src/autogluon/multimodal/presets.py
sxjscience: This is about removing clip in our default presets. We need to have a separate PR for improving our presets logic.

multimodal/src/autogluon/multimodal/data/process_text.py
sxjscience: In some models (like T5 / deberta-v3), the model_max_length is really large so I added the condition here to avoid popping up too many warning messages.

This is mainly to guard against warnings.

multimodal/src/autogluon/multimodal/utils/metric.py
sxjscience: This is only used in `matcher`. I tried but failed to remove it in this PR so we may do it in a follow-up PR.

multimodal/src/autogluon/multimodal/utils/metric.py
```diff
@@ -82,6 +84,9 @@ def infer_metrics(
                 validation_metric_name = HIT_RATE
             else:
                 validation_metric_name = eval_metric_name
+            if validation_metric_name == NDCG:
+                # TODO(?) Currently NDCG is not supported. We may pick a better replacement of NDCG for validation in the future.
+                validation_metric_name = HIT_RATE
```
zhiqiangdon, line 89: I think this can be removed.

sxjscience: Let's do it in a follow-up PR. The logic should be correct. Basically, validation_metric_name can never be NDCG because it is not supported.

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -191,22 +239,50 @@ def __init__(
...
+            This is used when the problem_type is one of the matching problem types, and when the labels are binary.
+            For example, the label column can contain ["match", "not match"]. And match_label can be "match".
+            It is similar as the "pos_label" in F1-score: https://scikit-learn.org/stable/modules/generated/sklearn.metrics.f1_score.html
+            Internally, we will set match_label to self.class_labels[1] by default.
         pipeline
-            This defines inference tasks like FeatureExtraction, ZeroShotClassification, etc.
-            TODO: add more pipelines (ref: https://huggingface.co/docs/transformers/main_classes/pipelines)
+            Pipeline has been deprecated and merged in problem_type.
+        presets
+            The presets for loading model parameters / training the model
```
zhiqiangdon, line 285: `presets` seems not to appear in the arguments.

suzhoum (approved):
LGTM!

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -546,6 +649,15 @@ def fit(
         -------
         An "MultiModalPredictor" object (itself).
         """
+        self._fit_called = True
```
zhiqiangdon, line 652: Not sure if a good idea to put `self._fit_called = True` in the fit beginning.

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -744,6 +853,16 @@ def fit(
         self.elapsed_time = (training_end - training_start) / 60.0
         return self
 
+    def _verify_inference_ready(self):
+        if not self._fit_called and not self._model_loaded:
```
zhiqiangdon, line 857: I guess the criteria of ready for inference may need to be reconsidered. If a the fit corrupted and we load the last checkpoint, is it ready for inference?

zhiqiangdon: The `self._fit_called` is set True at the beginning of `fit()`, but the predictor is not actually fit yet. It may cause some wrong check here in some special cases I guess.

sxjscience: Previously there is no check of `inference_ready` so we are better by providing some error messages at least.

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -153,6 +160,48 @@
 logger = logging.getLogger(AUTOMM)
 
 
+@dataclass
+class ProblemProperty:
+    name: str  # Name of the problem
+    support_fit: bool = True  # Whether the problem type support `.fit()`
+    inference_ready: bool = False  # Support `.predict()` and `.evaluate()` without calling `.fit()`
```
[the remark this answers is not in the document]
zhiqiangdon: Moreover, I don't see that we update the `inference_ready` after `fit()` is called.

multimodal/src/autogluon/multimodal/predictor.py
zhiqiangdon: Besides, I feel this logic is not straightforward as this is affected by several flags. May need to simply it.

cheungdaven (approved):
LGTM. There are a few TODOs, but we can merge this one first and fix later.

zhiqiangdon (approved):
Thanks for refactor! Left some comments that can be addressed in next PR.

multimodal/src/autogluon/multimodal/predictor.py
sxjscience: Basically, if the scenario can be detected by `_verify_inference_ready`, we will raise the RuntimeError. Otherwise, the behavior is the same as the previous implementation.

sxjscience: We can mark it as a lower-priority TODO and fix it in a follow-up PR.

multimodal/src/autogluon/multimodal/predictor.py
FANGAreNotGnu: Shall we later move this to constant?

PR #2340, opened 2022-11-08
Fix crash when tuning_data label dtype is invalid

Innixma:
*Issue #, if available:*

*Description of changes:*
- Fixed bug caused by tuning_data having a different dtype in the label column than train_data, which caused a crash due to holdout_frac being incorrectly specified. Now we correctly discard the tuning_data and split the train_data.
- Improved warning message when this scenario occurs so it is easier to understand.

Example updated output in this scenario:

```
No path specified. Models will be saved in: "AutogluonModels/ag-20221108_223519/"
WARNING: train_data and tuning_data have mismatched label column dtypes! train_label_type=object, tuning_data_type=int64.
	You should ensure the dtypes match to avoid bugs or instability.
	AutoGluon will attempt to convert the dtypes to align.
AutoGluon will save models to "AutogluonModels/ag-20221108_223519/"
AutoGluon Version:  0.5.3b20221101
Python Version:     3.8.10
Operating System:   Darwin
Platform Machine:   x86_64
Platform Version:   Darwin Kernel Version 21.6.0: Thu Sep 29 20:12:57 PDT 2022; root:xnu-8020.240.7~1/RELEASE_X86_64
Train Data Rows:    605
Train Data Columns: 4
Tuning Data Rows:    145
Tuning Data Columns: 4
Label Column: class
Preprocessing data ...
Warning: Some classes in the training set have fewer than 10 examples. AutoGluon will only keep 3 out of 4 classes for training and will not try to predict the rare classes. To keep more classes, increase the number of datapoints from these rare classes in the training data or reduce label_count_threshold.
Fraction of data from classes with at least 10 examples that will be kept for training models: 0.9950413223140496
Train Data Class Count: 3
############################################################################################################
WARNING: All X_val data contained low frequency classes, ignoring X_val and generating from subset of X
	Your input validation data or training data labels might be corrupted, please manually inspect them for correctness!
	Training Classes: ['0', '1', '2', 'N']
	Tuning   Classes: [0, 1, 2]
	Training Class Dtype: object
	Tuning   Class Dtype: int64
	Classes missing from Training Data: [0, 1, 2]
############################################################################################################
```

### DIFF
--- a/tabular/src/autogluon/tabular/learner/default_learner.py
+++ b/tabular/src/autogluon/tabular/learner/default_learner.py
@@ -172,6 +172,7 @@ class DefaultLearner(AbstractTabularLearner):
             logger.warning(f"Warning: Ignoring {len(missinglabel_inds)} (out of {len(X)}) training examples for which the label value in column '{self.label}' is missing")
             X = X.drop(missinglabel_inds, axis=0)
 
+        holdout_frac_og = holdout_frac
         if X_val is not None and self.label in X_val.columns:
             holdout_frac = 1
 
@@ -202,10 +203,14 @@ class DefaultLearner(AbstractTabularLearner):
         if X_val is not None and self.label in X_val.columns:
             X_val = self.cleaner.transform(X_val)
             if len(X_val) == 0:
-                logger.warning('All X_val data contained low frequency classes, ignoring X_val and generating from subset of X')
+                logger.warning('############################################################################################################\n'
+                               'WARNING: All X_val data contained low frequency classes, ignoring X_val and generating from subset of X\n'
+                               '\tYour input validation data or training data labels might be corrupted, please manually inspect them for correctness!\n'
+                               '############################################################################################################')
                 X_val = None
                 y_val = None
                 w_val = None
+                holdout_frac = holdout_frac_og
             else:
                 X_val, y_val = self.extract_label(X_val)
                 y_val = self.label_cleaner.transform(y_val)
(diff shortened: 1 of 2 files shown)


gradientsky (approved):

tabular/src/autogluon/tabular/learner/default_learner.py
```diff
@@ -200,12 +201,25 @@ def general_data_processing(self, X: DataFrame, X_val: DataFrame, X_unlabeled: D
...
+                logger.warning('############################################################################################################\n'
+                               'WARNING: All X_val data contained low frequency classes, ignoring X_val and generating from subset of X\n'
+                               '\tYour input validation data or training data labels might be corrupted, please manually inspect them for correctness!')
+                if self.problem_type in [BINARY, MULTICLASS]:
+                    train_classes = sorted(list(y_uncleaned.unique()))
+                    val_classes = sorted(list(X_val_og[self.label].unique()))
+                    logger.warning(f'\tTraining Classes: {train_classes}')
+                    logger.warning(f'\tTuning   Classes: {val_classes}')
+                    missing_classes = [c for c in val_classes if c not in train_classes]
+                    logger.warning(f'\tClasses missing from Training Data: {missing_classes}')
```
gradientsky, line 216: maybe worth adding types displayed:
```
	Training Classes: ['0', '1', '2', 'N']; type: object
	Tuning   Classes: [0, 1, 2]: type: int
```

gradientsky (approved):

tabular/src/autogluon/tabular/learner/default_learner.py
gradientsky: Another thought: it could be a problem if tuning data has different distribution of labels even if there is an overlap. I.e. training has labels [0, 1, 2] in roughly the same proportions, but tuning data only [0, 1] in 70/30 ratio - that potentially could to be a problem.

But.. we are closing EDA boundaries here.

Innixma: Nice! Added

PR #2343, opened 2022-11-08
Automatically check the precision of text models

cheungdaven:
*Description of changes:*

Checking the precision of pretrained text models (i.e., HFAutoModelForTextPrediction and NER) automatically.

### DIFF
--- a/multimodal/src/autogluon/multimodal/utils/environment.py
+++ b/multimodal/src/autogluon/multimodal/utils/environment.py
@@ -109,6 +109,9 @@ def infer_precision(
     -------
     The inferred precision.
     """
+    if not isinstance(model, list):
+        model = [model]
+
     precision_mapping = {
         16: torch.float16,
         "bf16": torch.bfloat16,
@@ -136,12 +139,14 @@ def infer_precision(
             )
             precision = 32
 
-        elif isinstance(model, HFAutoModelForTextPrediction):
-            if model.config.torch_dtype is not None:
-                precision = inverse_precision_mapping[model.config.torch_dtype]
-            else:
-                torch_dtype = get_state_dict_dtype(model.state_dict())
-                precision = inverse_precision_mapping[torch_dtype]
+        else:
+            for m in model:
+                if isinstance(m, HFAutoModelForTextPrediction):
+                    if m.config.torch_dtype is not None:
+                        precision = inverse_precision_mapping[m.config.torch_dtype]
+                    else:
+                        torch_dtype = get_state_dict_dtype(m.state_dict())
+                        precision = inverse_precision_mapping[torch_dtype]
 
     if as_torch:
         if precision in precision_mapping:
--- a/multimodal/src/autogluon/multimodal/utils/environment.py
+++ b/multimodal/src/autogluon/multimodal/utils/environment.py
@@ -139,14 +139,21 @@ def infer_precision(
             )
             precision = 32
 
-        else:
-            for m in model:
-                if isinstance(m, HFAutoModelForTextPrediction):
-                    if m.config.torch_dtype is not None:
-                        precision = inverse_precision_mapping[m.config.torch_dtype]
-                    else:
-                        torch_dtype = get_state_dict_dtype(m.state_dict())
-                        precision = inverse_precision_mapping[torch_dtype]
+        checkpoint_precision = None
+        for m in model:
+            if isinstance(m, HFAutoModelForTextPrediction):
+                if m.config.torch_dtype is not None:
+                    checkpoint_precision = inverse_precision_mapping[m.config.torch_dtype]
+                else:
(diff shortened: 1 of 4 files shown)


multimodal/tests/unittests/test_matcher.py
```diff
@@ -273,8 +273,8 @@ def test_text_semantic_search():
     for per_query_hits, per_query_hits_2, per_query_hit_gt in zip(hits, hits_2, hits_gt):
         for per_hit, per_hit_2, per_hit_gt in zip(per_query_hits, per_query_hits_2, per_query_hit_gt):
             assert per_hit["response_id"] == per_hit_2["response_id"] == per_hit_gt["response_id"]
-            npt.assert_almost_equal(per_hit["score"], per_hit_2["score"])
-            npt.assert_almost_equal(per_hit["score"], per_hit_gt["score"])
+            npt.assert_allclose(per_hit["score"], per_hit_2["score"], 1e-3, 1e-3)
+            npt.assert_allclose(per_hit["score"], per_hit_gt["score"], 1e-3, 1e-3)
```
zhiqiangdon, line 277: We may need to find the reason why the precision is required for now. I have a concern that we have missed something.

multimodal/src/autogluon/multimodal/utils/environment.py
```diff
@@ -110,13 +139,16 @@ def infer_precision(
             )
             precision = 32
 
+        else:
```
zhiqiangdon, line 142: Our config precision `bf16` can still overwrite the precision in checkpoint?

multimodal/src/autogluon/multimodal/utils/environment.py
```diff
@@ -70,8 +72,22 @@ def compute_num_gpus(config_num_gpus: Union[int, float, List], strategy: str):
...
+    """
+    Returns the first found floating dtype in `state_dict` if there is one, otherwise returns the first dtype.
+    """
+    for t in state_dict.values():
+
+        if t.is_floating_point():
+            return t.dtype
+
+    # if no floating dtype was found return whatever the first dtype is
+    else:
```
zhiqiangdon, line 85: The `else` doesn't align with `if`.

multimodal/src/autogluon/multimodal/utils/environment.py
```diff
@@ -70,8 +72,22 @@ def compute_num_gpus(config_num_gpus: Union[int, float, List], strategy: str):
...
 
 
+def get_state_dict_dtype(state_dict):
+    """
+    Returns the first found floating dtype in `state_dict` if there is one, otherwise returns the first dtype.
+    """
+    for t in state_dict.values():
+
+        if t.is_floating_point():
+            return t.dtype
```
zhiqiangdon, line 82: Do we need this `if`? Can we always return `next(state_dict.values()).dtype`?

multimodal/src/autogluon/multimodal/utils/environment.py
cheungdaven: It is supposed to be like this. If the for loop does not return anything, it will go to this else statement.

multimodal/src/autogluon/multimodal/utils/environment.py
cheungdaven: The config default is 16 not bf16.

multimodal/src/autogluon/multimodal/utils/environment.py
cheungdaven: t can be int

multimodal/src/autogluon/multimodal/utils/environment.py
zhiqiangdon: You are right. What if users specify `bf16` in config and the checkpoint precision is not `bf16`?

multimodal/src/autogluon/multimodal/utils/environment.py
cheungdaven: This is from the huggingface [sourcecode](https://github.com/huggingface/transformers/blob/67b37891335262bbbb501b0b53418aafd43a04d7/src/transformers/modeling_utils.py#L218-L228).

multimodal/src/autogluon/multimodal/utils/environment.py
cheungdaven: User argument should be treated with the highest priority. I do not think it's a good idea to override them.

sxjscience:
We can reuse https://discuss.huggingface.co/t/model-pre-training-precision-database-fp16-fp32-bf16/5671.

sxjscience:
The current logic is not correct i think. It will always detect float32?

cheungdaven:
> The current logic is not correct i think. It will always detect float32?

Yeah, it’s incorrect unless the torch_dtype is specified.

PR #2349, opened 2022-11-09
AutoMM Detection Tutorials Updates

FANGAreNotGnu:
- update `pipeline` in detection tutorials/examples to `problem_type`
- add tutorial and example about loading a trained predictor and change `num_gpus`
- add FPS for models in tutorials


docs/tutorials/multimodal/object_detection/detection_load_predictor.md
```diff
@@ -0,0 +1,67 @@
...
+ Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets= 10 ] = 0.505
+ Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.515
+ Average Recall     (AR) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.258
+ Average Recall     (AR) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.415
+ Average Recall     (AR) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.556
+```
+
+After the predictor is loaded, you can do the evaluation, prediction,
+or finetune it on another dataset. But note the loaded predictor is saved at the same path
+and the save operation in finetuning will overwrite the previous checkpoint.
```
bryanyzhu, line 60: I think we need to add one sentence here, if users do want to continue training, what will they do to avoid this overwriting.

bryanyzhu (approved):
Except that one minor comment, LGTM. Thank you.

docs/tutorials/multimodal/object_detection/detection_load_predictor.md
FANGAreNotGnu: We don't have this logic (use a new `self._save_path` in predictor). A simple solution is add a function `change_save_path`, but I think it's better to add a parameter to `MultiModalPredictor.load`

bryanyzhu: Ah, I see. Ok, then let's discuss later on whether we want to have a function change_save_path, and also whether we want to add save_path to init. Thanks.

sxjscience (approved):
LGTM!!

docs/tutorials/multimodal/object_detection/detection_eval_ddetr_coco.md
```diff
@@ -13,7 +13,8 @@ from autogluon.multimodal import MultiModalPredictor
 ```
 
 We select the two-stage Deformable DETR with ResNet50 as backbone with bounding box finetune,
-for other Deformable DETR models see [MMDetection's Deformable DETR models](https://github.com/open-mmlab/mmdetection/tree/master/configs/deformable_detr). 
+this model has **15.7 frames per second (FPS)** on single A10e GPU with `batch_size=1`.
+For details about model selection, see :ref:`sec_automm_detection_select_models`.
```
yongxinw, line 17: Is the :ref: working as expected? When I render, I wasn't able to see a link (was still seeing :ref:sec_automm_detection_select_models). Or maybe I was not understanding this correctly.

yongxinw: otherwise, LGTM

PR #2351, opened 2022-11-10
Add presets in AutoMM for vision

bryanyzhu:
*Issue #, if available:*
NA

*Description of changes:*
Add new presets for image classification, no HPO involved for now
1. best_quality_image_classification
2. high_quality_fast_inference_image_classification
3. medium_quality_faster_inference_image_classification

### DIFF
--- a/multimodal/src/autogluon/multimodal/presets.py
+++ b/multimodal/src/autogluon/multimodal/presets.py
@@ -64,7 +64,7 @@ def best_quality():
 
 
 @automm_presets.register()
-def best_quality_image_classification():
+def high_quality_image_classification():
     return {
         "model.names": ["timm_image"],
         "model.timm_image.checkpoint_name": "swin_base_patch4_window7_224",
--- a/multimodal/src/autogluon/multimodal/presets.py
+++ b/multimodal/src/autogluon/multimodal/presets.py
@@ -27,6 +27,24 @@ def medium_quality_faster_train():
 
 
 @automm_presets.register()
+def medium_quality_faster_inference_image_classification():
+    return {
+        "model.names": ["timm_image"],
+        "model.timm_image.checkpoint_name": "mobilenetv3_large_100",
+        "optimization.learning_rate": 1e-3,
+    }
+
+
+@automm_presets.register()
+def high_quality_fast_inference_image_classification():
+    return {
+        "model.names": ["timm_image"],
+        "model.timm_image.checkpoint_name": "resnet50",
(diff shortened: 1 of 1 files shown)


sxjscience (approved):
LGTM!

multimodal/src/autogluon/multimodal/presets.py
```diff
@@ -45,6 +63,14 @@ def best_quality():
     }
 
 
+@automm_presets.register()
+def high_quality_image_classification():
```
FANGAreNotGnu, line 67: Should this be `higher_quality`? Or change `high_quality_fast_inference_image_classification` to be good quality?

FANGAreNotGnu (approved):
LGTM with a minor comment.

multimodal/src/autogluon/multimodal/presets.py
bryanyzhu: This is an extra preset because previous benchmarking is done on swin-base. Ideally, we just want to have 3 per problem type.

multimodal/src/autogluon/multimodal/presets.py
```diff
@@ -26,6 +26,24 @@ def medium_quality_faster_train():
     }
 
 
+@automm_presets.register()
+def medium_quality_faster_inference_image_classification():
+    return {
+        "model.names": ["timm_image"],
+        "model.timm_image.checkpoint_name": "mobilenetv3_large_100",
```
suzhoum, line 33: Just curious, is `model.names` required when `timm_image.checkpoint_name` is specified?

suzhoum (approved):
LGTM!

multimodal/src/autogluon/multimodal/presets.py
bryanyzhu: I don't think it is required, correct me if I'm wrong @sxjscience @zhiqiangdon 

I see all other presets have `model.names` written out even though it is clear from the context, so I just follow.

zhiqiangdon: Better to keep the `model.names=["timm_image"]` to make the context more clear.

bryanyzhu: +1, so let's keep the `model.names`.

suzhoum: Thanks for the insights!

zhiqiangdon (approved):
LGTM!

PR #2353, opened 2022-11-10
Add Presets in AutoMM for Detection

FANGAreNotGnu:
- move val_metric from direct input of `MultiModalPredictor.__init__` to `hyperparameters`
- add presets and fix val_metric in tuto and example
- bypass the `best_score` calculation in `_top_k_average` for detection


multimodal/src/autogluon/multimodal/presets.py
```diff
@@ -102,9 +102,96 @@ def zero_shot_image_classification():
     }
 
 
+@automm_presets.register()
+def medium_quality_faster_train_object_detection():
```
bryanyzhu, line 106: I think the preset name we agreed on is fast_inference and faster_inference.

bryanyzhu (approved):
LGTM with a minor comment.

sxjscience (approved):
LGTM

multimodal/src/autogluon/multimodal/presets.py
FANGAreNotGnu: Just changed the naming. But in "faster" preset, I used validation loss as `val_metric`, which largely increased the training speed on small dataset (from 1474 seconds to 103 seconds on watercolor), and it didn't affect the inference speed. Comparing `faster` and `fast`, the inference fps difference is 85 vs 72, but the training time is 103 seconds v.s. 1645 seconds.

bryanyzhu: I see. I think we can have additional presets if that preset is likely to be used. Because each problem type has its own characteristic. Detection might care about both training and inference time, but matching maybe care more about accuracy.

multimodal/src/autogluon/multimodal/configs/optimization/adamw.yaml
```diff
@@ -17,6 +17,7 @@ optimization:
   gradient_clip_algorithm: "norm"
   track_grad_norm: -1 # Whether to check gradient norm. We can set it to 2 to check for gradient norm.
   log_every_n_steps: 10
+  val_metric: null
```
zhiqiangdon, line 20: I suggest to use `validation_metric` to make it more distinguishable from the already existing `eval_metric`.

FANGAreNotGnu: Oh sure, I'll change this in next pr

PR #2356, opened 2022-11-10
[Multimodal] Split Multimodal Tests

yinweisu:
*Issue #, if available:*

*Description of changes:*
* Naive split of multimodal tests into two parts based on runtime: https://gist.github.com/yinweisu/41d3ac66e3501d634c87e4d216896cd7
* Split into two parts: predictor and others so that the runtime would be capped by `test_predictor.py`

**Note: I've manually cancelled `pull_request_target` run as this would run the old workflow but with new test structure -> error.**


tonyhoo (approved):

zhiqiangdon (approved):
Thanks for the acceleration!

yinweisu:
According to the CI, both multimodal tests finished in 32m

PR #2360, opened 2022-11-11
Index AutoMM Detection Tutorial and Fix save_path problem

FANGAreNotGnu:
- Index AutoMM detection tutorials
- Fix save_path problem and refactor
- Fix typo and path error in example and tutorial


bryanyzhu (approved):
LGTM, thanks. Please add installation guide on detection main page, AG installation main page, and in the code when people try to init problem_type = object_detection''. I think the installation is 

```
mim install mmcv-full
pip install mmdet
```

multimodal/src/autogluon/multimodal/utils/save.py
```diff
@@ -112,3 +114,37 @@ def process_save_path(path, resume: Optional[bool] = False, raise_if_exist: Opti
...
+    hyperparameter_tune_kwargs: Optional[dict] = None,
+    warn_if_exist: Optional[bool] = True,
+    num_gpus: Optional[int] = None,
+):
+    save_path = None
+    raise_if_exist_for_proposed_path = True
+    if num_gpus and num_gpus > 1:
+        raise_if_exist_for_proposed_path = (
+            False  # TODO: Here we allow overwrite previous save_path if we are using multi gpu
+        )
```
zhiqiangdon, line 132: If users give a `proposed_save_path`, which already exists, and use multi gpus, this will create a new folder. If users try to use their specified path, they would still find the old results there. I think this behavior can cause user confusions.

zhiqiangdon: For now, adding a TODO item is fine.

yongxinw (approved):
The save_path part looks good to me!

PR #2361, opened 2022-11-11
add preset for ner

cheungdaven:
*Issue #, if available:*

*Description of changes:*


zhiqiangdon (approved):
LGTM

multimodal/src/autogluon/multimodal/presets.py
```diff
@@ -386,6 +386,31 @@ def medium_quality_faster_inference_image_text_similarity():
 
 
 @automm_presets.register()
+def best_quality_ner():
+    return {
+        "model.names": ["ner"],
+        "model.ner.checkpoint_name": "google/flan-t5-base",
+        "env.precision": 32,
```
sxjscience, line 393: How about use `"bf16"`?

multimodal/src/autogluon/multimodal/presets.py
```diff
@@ -386,6 +386,31 @@ def medium_quality_faster_inference_image_text_similarity():
...
+        "model.ner.checkpoint_name": "google/flan-t5-small",
+        "env.precision": 32,
+    }
+
+
+@automm_presets.register()
+def high_quality_fast_inference_ner():
+    return {
+        "model.names": ["ner"],
+        "model.ner.checkpoint_name": "bert-base-cased",
```
sxjscience, line 410: What's the performance of architectures like `deberta-v3-small`?

cheungdaven: In general, deberta-v3 works better than bert-base, but
using ddp_spawn + multiple gpus will cause an error.
using ddp + multiple gpus does not lead to good performance.

multimodal/src/autogluon/multimodal/presets.py
cheungdaven: changed to "bf16"

multimodal/src/autogluon/multimodal/presets.py
sxjscience: What type of error? Shall we fix it?

cheungdaven: cuda error, cannot find some functions from cuda.   It may be an error from pytorch lightning.

bryanyzhu (approved):
LGTM

FANGAreNotGnu (approved):
LGTM！

PR #2367, opened 2022-11-11
Add missing import sys in setup.py

Innixma:
*Issue #, if available:*

*Description of changes:*
- sys import missing from setup.py, causing pre-release to fail (https://pypi.org/project/autogluon/#history). Also will crash any local install attempts.

Original bug from #2225 that added `if sys.platform == 'darwin'` logic.

Log showing error in pre-release build:

https://github.com/awslabs/autogluon/actions/runs/3392499510/jobs/5638772315#step:5:1061

```
  File "setup.py", line 61, in <module>
    ] if sys.platform == 'darwin' else [
NameError: name 'sys' is not defined
Error: Process completed with exit code 1.
```

### DIFF
--- a/tabular/setup.py
+++ b/tabular/setup.py
@@ -35,12 +35,12 @@ extras_require = {
         'lightgbm>=3.3,<3.4',
     ],
     'catboost': [
-        'catboost>=1.0,<1.2',
+        'catboost>=1.0,<1.1',
     ],
     # FIXME: Debug why xgboost 1.6 has 4x+ slower inference on multiclass datasets compared to 1.4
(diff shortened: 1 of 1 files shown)


tonyhoo (approved):

Innixma:
Question to team: @yinweisu @gradientsky : How is our CI passing with this bug present? Shouldn't it crash on tabular unit tests and website runs?

liangfu:
CI was passing, since we were running
```bash
python3 -m pip install -e tabular/
```
instead of running the setup.py file directly.

Issue #2368, opened 2022-11-11
[BUG] Different behavior of setup.py


yinweisu:
In our CI, we install our package through `python3 -m pip install -e ./module`
This setup, for some reason, has packages like `sys` and `os` by default even when we didn't import it.
For example, put this line at the top of setup.py won't fail the installation process
```python
assert isinstance(sys.platform, str)
```

However, `sys` is not there when we build the package for distribution:
```bash
$ python setup.py sdist bdist_wheel
Traceback (most recent call last):
  File "setup.py", line 4, in <module>
    assert isinstance(sys.platform, str)
NameError: name 'sys' is not defined
```

This difference in behavior would allow bug regarding installation slip away in the CI but fail the actual distribution. Will need to investigate

Innixma:
@yinweisu Please double check status of this and mark as resolved if fixed.

yinweisu:
Close as we are testing wheels building in our CI now.

PR #2369, opened 2022-11-12
[CI] Add building test to CI

yinweisu:
*Issue #, if available:*
https://github.com/awslabs/autogluon/issues/2368

*Description of changes:*
* Added build step in `install` tests
* `platform_test` won't take effect before merging. Need to merge then have a follow-up PR to check if it's working...

### DIFF
--- a/.github/workflow_scripts/env_setup.sh
+++ b/.github/workflow_scripts/env_setup.sh
@@ -102,3 +102,12 @@ function install_all {
     install_eda
     install_autogluon
 }
+
+function build_all {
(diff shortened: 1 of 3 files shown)


tonyhoo (approved):

yinweisu:
Install tests have passed. Shouldn't affect other modules. Will merge now

tonyhoo:
BTW, I think we should create an issue to inspect our release process and if we are following the best practice (e.g. `pip wheel` versus `python setup.py`)

PR #2234, opened 2022-10-20
[M1] Removing deprecated eval_metric from fit - causing segfaults on M1 w/ xgboost 1.4.2

gradientsky:
*Description of changes:*
Removing deprecated eval_metric from fit - causing segfaults on M1 w/ xgboost 1.4.2

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2373, opened 2022-11-13
[v0.6] Updated dependency versions

Innixma:
*Description of changes:*
Updated dependency versions.

TODO: 

- [x] Tested autogluon.tabular locally on Ubuntu Python 3.8 and MacOS Python 3.8
- [x] Checked that min versions of dependencies remain compatible with Python 3.7
- [x] Get approval from autogluon.multimodal lead (scikit-learn, scipy, numpy, pandas)
- [x] Get approval from autogluon.timeseries lead (scikit-learn, scipy, numpy, pandas)
- [x] Benchmark on AutoMLBenchmark : Results look good, no major change, no stability issues.
- [x] Checked XGBoost upgrade is stable
- [x] Checked CatBoost upgrade is stable

Old:

```
'numpy': '>=1.21,<1.23',
'pandas': '>=1.2.5,!=1.4.0,<1.5',
'scikit-learn': '>=1.0.0,<1.1',
'scipy': '>=1.5.4,<1.8.0',
'catboost>=1.0,<1.1',
'xgboost>=1.4,<1.5',
```

New:

```
'numpy': '>=1.21,<1.24',
'pandas': '>=1.2.5,!=1.4.0,<1.6',
'scikit-learn': '>=1.0.0,<1.2',
'scipy': '>=1.5.4,<1.10.0',
'catboost>=1.0,<1.2',
'xgboost>=1.6,<1.8',
```

Beyond the version updates, I have added logic to address deprecation warnings, including:
- Removed the usage of Pandas `.append` which is deprecated, now using `.concat`.
- Removed usage of Pandas `.iteritems` which is deprecated.
- Refactored XGBoost custom metric logic to adhere to the API change in XGBoost 1.6+. I have tested this locally with a variety of metrics and problem types.
- Removed `use_label_encoder` parameter setting in XGBoost, it is now deprecated. No functional change.

Additional changes:
- Fixed bug introduced in #2234 that caused XGBoost to ignore the user-specified eval metric. Now it properly uses the eval metric as before by specifying it in `params`.
- Added version range check when importing XGBoost to ensure user isn't using an old version of XGBoost. This is because callbacks are incompatible with the older versions and will cause the model to never early stop, making the result very poor. Now we error if XGBoost isn't installed with expected version range.
- Changed objective for multiclass XGBoost to `objective': 'multi:softprob'` to satisfy custom metric API requirements. This has no functional change in model training or inference.
- Added documentation to `generate_train_test_split`
- Added new function `generate_train_test_split_combined` that will simplify tutorials where we need to split a dataset into train and test from ~6 LoC to 1 LoC. Currently unused within the core AutoGluon logic.

### DIFF
--- a/core/src/autogluon/core/utils/utils.py
+++ b/core/src/autogluon/core/utils/utils.py
@@ -360,12 +360,115 @@ def get_pred_from_proba(y_pred_proba, problem_type=BINARY):
     return y_pred
 
 
+def extract_label(data: DataFrame, label: str) -> (DataFrame, Series):
+    """
+    Extract the label column from a dataset and return X, y.
+
+    Parameters
+    ----------
+    data : DataFrame
+        The data containing features and the label column.
+    label : str
+        The label column name.
+
+    Returns
+    -------
+    X, y : (DataFrame, Series)
+        X is the data with the label column dropped.
+        y is the label column as a pd.Series.
+    """
+    if label not in list(data.columns):
+        raise ValueError(f"Provided DataFrame does not contain label column: {label}")
+    y = data[label].copy()
+    X = data.drop(label, axis=1)
+    return X, y
+
+
+def generate_train_test_split_combined(data: DataFrame,
+                                       label: str,
+                                       problem_type: str,
+                                       test_size: float = 0.1,
+                                       random_state: int = 0,
+                                       min_cls_count_train: int = 1) -> (DataFrame, DataFrame):
+    """
+    Generate a train test split from a DataFrame that contains the label column.
+
+    Parameters
+    ----------
+    data : DataFrame
+        DataFrame containing the features plus the label column to split into train and test sets.
+    label : str
+        The label column name.
+        Used for stratification and to ensure all classes in multiclass classification are preserved in train data.
+    problem_type : str
+        The problem_type the label is used for. Determines if stratification is used.
+        Options: ["binary", "multiclass", "regression", "softclass", "quantile"]
+    test_size : float, default = 0.1
+        The proportion of data to use for the test set.
+        The remaining (1 - test_size) of data will be used for the training set.
+    random_state : int, default = 0
+        Random seed to use during the split.
+    min_cls_count_train : int, default = 1
+        The minimum number of instances of each class that must occur in the training set (for classification).
+        If not satisfied by the original split, instances of unsatisfied classes are
+        taken from test and put into train until satisfied.
+        Raises an exception if impossible to satisfy.
+
+    Returns
+    -------
+    train_data, test_data : (DataFrame, DataFrame)
+        The train_data and test_data after performing the split. Includes the label column.
+    """
+    X, y = extract_label(data=data, label=label)
+    train_data, test_data, y_train, y_test = generate_train_test_split(
+        X=X,
+        y=y,
+        problem_type=problem_type,
+        test_size=test_size,
+        random_state=random_state,
+        min_cls_count_train=min_cls_count_train)
+    train_data[label] = y_train
+    test_data[label] = y_test
+    return train_data, test_data
+
+
 def generate_train_test_split(X: DataFrame,
                               y: Series,
                               problem_type: str,
                               test_size: float = 0.1,
-                              random_state=0,
-                              min_cls_count_train=1) -> (DataFrame, DataFrame, Series, Series):
+                              random_state: int = 0,
+                              min_cls_count_train: int = 1) -> (DataFrame, DataFrame, Series, Series):
+    """
+    Generate a train test split from input X, y.
+    If you have a combined X, y DataFrame, refer to `generate_train_test_split_combined` instead.
+
+    Parameters
+    ----------
+    X : DataFrame
+        pd.DataFrame containing the features minus the label column to split into train and test sets.
+    y : Series
+        pd.Series containing the label with matching indices to X.
+        Used for stratification and to ensure all classes in multiclass classification are preserved in train data.
+    problem_type : str
+        The problem_type the label is used for. Determines if stratification is used.
+        Options: ["binary", "multiclass", "regression", "softclass", "quantile"]
+    test_size : float, default = 0.1
+        The proportion of data to use for the test set.
+        The remaining (1 - test_size) of data will be used for the training set.
+    random_state : int, default = 0
+        Random seed to use during the split.
+    min_cls_count_train : int, default = 1
+        The minimum number of instances of each class that must occur in the training set (for classification).
+        If not satisfied by the original split, instances of unsatisfied classes are
+        taken from test and put into train until satisfied.
+        Raises an exception if impossible to satisfy.
+
+    Returns
+    -------
+    X_train, X_test, y_train, y_test : (DataFrame, DataFrame, Series, Series)
+        The train_data and test_data after performing the split, separated into X and y.
+
+    """
     if (test_size <= 0.0) or (test_size >= 1.0):
         raise ValueError("fraction of data to hold-out must be specified between 0 and 1")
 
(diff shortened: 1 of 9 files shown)


sxjscience (approved):
Dependency upgrade LGTM.

Innixma:
@canerturkmen Looks like a lot of FutureWarnings are being logged for using deprecated iteritems in TimeSeries. Do you think it is do-able to switch to a non-deprecated option prior to v0.6 release, or would you prefer we avoid the dependency upgrade and upgrade post-v0.6?

Innixma:
Benchmark result shows no notice-able change in performance or stability: all is good:

```
AutoGluon_bq_1h8c_2022_11_12_v06 VS all
                              framework   winrate   >   <   =  % Less Avg. Errors  Avg Inf Speed Diff  time_train_s  time_infer_s  loss_rescaled  time_train_s_rescaled  time_infer_s_rescaled      rank  rank=1_count  rank=2_count  rank=3_count  rank>3_count  error_count
0  AutoGluon_bq_1h8c_2022_11_12_v06_upg  0.547872  29  20  45            0.188582           -0.114332   3818.831868      0.026271       0.189848              44.920428              91.135303  3.142857            12            34             7            42            8
1      AutoGluon_bq_1h8c_2022_11_12_v06  0.500000   0   0  95            0.000000                 NaN   3793.946154      0.023840       0.197768              44.386338              82.380195  3.225275            10            34            11            40            8
2      AutoGluon_hq_1h8c_2022_11_12_v06  0.327957  30  62   1           -3.066443           25.737635   3255.721978      0.001433       0.393019              31.356587               4.960608  4.412088             3            11            20            67            2
3  AutoGluon_hq_1h8c_2022_11_12_v06_upg  0.301075  27  64   2           -3.122772           26.038704   3265.828571      0.001453       0.398810              31.825460               5.080363  4.434066             6            12            16            67            2
4      AutoGluon_gq_1h8c_2022_11_12_v06  0.287234  27  67   0           -3.955741           36.341472   3180.319780      0.001250       0.483376              31.419267               3.236277  4.868132             6             5            11            80            1
5  AutoGluon_gq_1h8c_2022_11_12_v06_upg  0.287234  27  67   0           -4.256028           34.577324   3205.010989      0.001266       0.479140              31.439799               3.786754  4.917582             3             6            16            77            1
6      AutoGluon_mq_1h8c_2022_11_12_v06  0.231579  22  73   0           -7.688752           70.329755    839.820879      0.000901       0.640749               1.010131               3.007559  5.494505             1            22             2            78            0
7  AutoGluon_mq_1h8c_2022_11_12_v06_upg  0.231579  22  73   0           -7.690055           63.865318    855.434066      0.000982       0.641012               1.058098               4.136992  5.505495             3            20             1            79            0

```

(`error_count` can be ignored, those were caused by something separate I was testing and are unrelated to this comparison)

One thing to note it is does appear inference speed became slower with the upgrade.

Honing in on the source of the slowdown, it is in multiclass classification datasets, slower by ~4x:

```
AutoGluon_mq_1h8c_2022_11_12_v06 VS all (Multiclass Datasets)
                              framework   winrate  >  <   =  % Less Avg. Errors  Avg Inf Speed Diff  time_train_s  time_infer_s  loss_rescaled  time_train_s_rescaled  time_infer_s_rescaled      rank  rank=1_count  rank=2_count  rank=3_count  rank>3_count  error_count
0      AutoGluon_mq_1h8c_2022_11_12_v06  0.500000  0  0  30            0.000000                 NaN   1459.156667      0.001017       0.033333               1.002219               1.002564  1.483333             2            28             0             0            0
1  AutoGluon_mq_1h8c_2022_11_12_v06_upg  0.483333  1  2  27           -0.017129           -2.722947   1505.253333      0.001169       0.100000               1.072392               3.725511  1.516667             1            29             0             0            0

```

No slowdown in binary or regression:
```
AutoGluon_mq_1h8c_2022_11_12_v06 VS all (Binary and Regression Datasets)
                              framework   winrate  >  <   =  % Less Avg. Errors  Avg Inf Speed Diff  time_train_s  time_infer_s  loss_rescaled  time_train_s_rescaled  time_infer_s_rescaled      rank  rank=1_count  rank=2_count  rank=3_count  rank>3_count  error_count
0  AutoGluon_mq_1h8c_2022_11_12_v06_upg  0.513699  3  1  69            0.006104           -0.026253    671.369863      0.000934       0.013699               1.043424               1.047632  1.486301             3            70             0             0            0
1      AutoGluon_mq_1h8c_2022_11_12_v06  0.500000  0  0  73            0.000000                 NaN    667.679452      0.000888       0.041096               1.007905               1.021380  1.513699             1            72             0             0            0

```

It turns out there is a multiclass dataset `shuttle` where XGBoost is 100x slower to infer with XGBoost=1.7 than XGBoost=1.4. This is something I noticed awhile ago that XGBoost is generally slower on multiclass in the newer releases, which is why I didn't upgrade the version until now. There are several other datasets where XGBoost is noticeably slower in multiclass as well. I think we are ok to upgrade, but should keep in mind that XGBoost has something strange going on slowing it down.

Innixma:
I fixed the XGBoost slowdown by adding a tiny min_delta to its early stopping callback. This callback delta logic was changed in XGBoost 1.5 release and probably what caused the difference in speed on our end.

canerturkmen:
I'm testing a quick change removing the usages of iteritems. If there is any other deprecated code from numpy or pandas we are using, we may not be able to suppress those by the release but I'm OK to switch out just iteritems for now and make this change.

tonyhoo (approved):

core/src/autogluon/core/utils/utils.py
```diff
@@ -360,12 +360,115 @@ def get_pred_from_proba(y_pred_proba, problem_type=BINARY):
...
+    Parameters
+    ----------
+    data : DataFrame
+        DataFrame containing the features plus the label column to split into train and test sets.
+    label : str
+        The label column name.
+        Used for stratification and to ensure all classes in multiclass classification are preserved in train data.
+    problem_type : str
+        The problem_type the label is used for. Determines if stratification is used.
+        Options: ["binary", "multiclass", "regression", "softclass", "quantile"]
```
gradientsky, line 405: nit: we don't enforce this list anywhere. Add defensive assert here to ensure that we guard against typos?

tabular/src/autogluon/tabular/models/xgboost/xgboost_utils.py
```diff
@@ -30,66 +29,27 @@ def convert_ag_metric_to_xgbm(ag_metric_name, problem_type):
...
                 y_hat = y_hat.argmax(axis=1)
-                res = metric(y_true, y_hat)
-                return metric.name, -1 * res if is_higher_better else res
+                return sign * metric(y_true, y_hat)
         elif problem_type == BINARY:
-            def function_template(y_hat, data):
-                y_true = data.get_label()
-                y_hat = _sigmoid(y_hat)
+            def custom_metric(y_true, y_hat):
                 y_hat = np.round(y_hat)
```
gradientsky, line 46: Not sure if this is a big deal, but this might have some float point issues:
```python
In [5]: {i: np.around(i) for i in np.arange(0, 1, 0.1)}
Out[5]:
{0.0: 0.0,
 0.1: 0.0,
 0.2: 0.0,
 0.30000000000000004: 0.0,
 0.4: 0.0,
 0.5: 0.0,  <<< should be 1.0
 0.6000000000000001: 1.0,
 0.7000000000000001: 1.0,
 0.8: 1.0,
 0.9: 1.0}
```

there's a way to get it done and it is actually faster:
```python
In [15]: %%timeit
    ...: {i: int(i>=0.5) for i in np.arange(0, 1, 0.1)}
    ...:
    ...:
4.89 µs ± 8.45 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```
vs 
```python
In [17]: %%timeit
    ...: {i: np.around(i) for i in np.arange(0, 1, 0.1)}
    ...:
    ...:
36.2 µs ± 818 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

canerturkmen (approved):

tabular/src/autogluon/tabular/models/xgboost/xgboost_utils.py
Innixma: Thanks for taking a look! I don't want to mess with this code any more prior to release, but you can send a PR for post-release follow-up

gradientsky (approved):

core/src/autogluon/core/utils/utils.py
Innixma: will add to follow-up PR

PR #2376, opened 2022-11-13
Fix syntax error in type annotation

daikikatsuragawa:
Signed-off-by: Daiki Katsuragawa <50144563+daikikatsuragawa@users.noreply.github.com>

When I executed the following command ...

```sh
mypy --ignore-missing-imports --warn-unreachable core/
```

The following error was detected.

```sh
core/src/autogluon/core/utils/utils.py:368: error: Syntax error in type annotation  [syntax]
```

When a function returns more than one value, the values are represented as tuples.


daikikatsuragawa:
For example, I get the following error with the following task.

https://github.com/awslabs/autogluon/actions/runs/3455985811/jobs/5768417443

```sh
Error: 🚨 Insuffient Permissions! daikikatsuragawa does not have write permissions
```

What should I do?

core/src/autogluon/core/utils/utils.py
```diff
@@ -365,7 +365,7 @@ def generate_train_test_split(X: DataFrame,
                               problem_type: str,
                               test_size: float = 0.1,
                               random_state=0,
-                              min_cls_count_train=1) -> (DataFrame, DataFrame, Series, Series):
+                              min_cls_count_train=1) -> Tuple(DataFrame, DataFrame, Series, Series):
```
yinweisu, line 368: I believe it should be `Tuple[]`
Ditto for the rest

yinweisu (changes requested):
Thanks for the PR. We probably won't merge this in right now as we are in code freeze for 0.6 release

core/src/autogluon/core/utils/utils.py
daikikatsuragawa: @yinweisu 

Thank you for pointing that out. I agree with your comment. I have fix it!

daikikatsuragawa:
[Merge branch 'master' into master](https://github.com/awslabs/autogluon/pull/2376/commits/bb85eae62786f78a4d0f941044acdf10caefac46)

This is a response to conflicts.

Innixma (approved):

PR #2389, opened 2022-11-14
Fix FTTransformer errors, add max_features

Innixma:
*Issue #, if available:*

*Description of changes:*
- Fix FTTransformer errors, add max_features
- path was not defined, caused error when fitting, now using self.path
- Now warn when no GPU
- Now skip fitting if >300 features, to avoid OOM
- Created https://github.com/awslabs/autogluon/issues/2390 to track hack

### DIFF
--- a/tabular/src/autogluon/tabular/models/automm/ft_transformer.py
+++ b/tabular/src/autogluon/tabular/models/automm/ft_transformer.py
@@ -2,9 +2,13 @@
 import logging
 from typing import Dict
 
+from autogluon.common.features.types import R_INT, R_FLOAT, R_CATEGORY, \
+    S_TEXT_NGRAM, S_TEXT_SPECIAL
+
 from .automm_model import MultiModalPredictorModel
 
 logger = logging.getLogger(__name__)
+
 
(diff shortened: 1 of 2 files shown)


sxjscience (approved):
LGTM

tabular/src/autogluon/tabular/models/automm/ft_transformer.py
```diff
@@ -59,6 +78,7 @@ def _set_default_params(self):
             "optimization.warmup_steps": 0.0,
             "optimization.patience": 20,
             "optimization.top_k": 3,
+            '_max_features': 300,  # FIXME: This is a hack, move to AG_ARGS_FIT for v0.7
```
gradientsky, line 81: Please add an issue tracker for this so it's not lost.

gradientsky (approved):

tabular/src/autogluon/tabular/models/automm/ft_transformer.py
gradientsky: https://github.com/awslabs/autogluon/issues/2390

PR #2394, opened 2022-11-15
Remove dummy layer

cheungdaven:
*Issue #, if available:*

*Description of changes:*

1. overwrite the max length for standalone checkpoints.
2. remove dummy layer to enable gradient_checkpointing 


multimodal/src/autogluon/multimodal/data/process_ner.py
```diff
@@ -124,7 +124,9 @@ def process_ner(
         A dictionary containing one NER sample's features and/or labels.
         """
         ret = {}
-
+        # overwrite model_max_length for standalone checkpoints if it's not specified.
+        if self.max_len is not None and self.tokenizer.model_max_length > 10**6:
+            self.tokenizer.model_max_length = self.max_len
```
zhiqiangdon, line 129: Why not doing this in `init`?

cheungdaven: It's done in init. But for checkpoints that are saved using the standalone mode, the tokenizer will be overwritten in load.py and the model_max_length is gone so here we rewrite it.

zhiqiangdon (approved):
LGTM.

multimodal/src/autogluon/multimodal/data/process_ner.py
zhiqiangdon: It seems that `save_text_tokenizers` missed the max_len attribute.

sxjscience: Shall we save max_len instead?

sxjscience: Or we can rely on our saved yaml config

sxjscience (approved):
LGTM

multimodal/src/autogluon/multimodal/data/process_ner.py
cheungdaven: I will add a TODO and revise it later.

PR #2398, opened 2022-11-15
[timeseries] Fix leaderboard bug if training data contains static features

shchur:
*Description of changes:*
- Fixes a bug where the `.leaderboard()` method would fail if (1) training data contained static features and (2) no data was given at prediction time.
- Add a test to prevent this bug in the future.


canerturkmen (approved):
LGTM!

Innixma (approved):
LGTM!

PR #2405, opened 2022-11-15
[timeseries] make core and tabular dependencies explicit

canerturkmen:
*Issue #, if available:*

*Description of changes:*

Optional dependencies were missing from core and tabular dependency listings in time series. This can lead to errors (currently caught, but leads to failing models) in edge cases: if the user does not `pip install autogluon` or `full_install` but instead tries to do a bare install of time series. 


Innixma (approved):
LGTM!

PR #2413, opened 2022-11-16
Fix cpu inference bug

sxjscience:
*Issue #, if available:*

*Description of changes:*

Fix the bug that `dp` is not supported in CPU-only instance


gradientsky (approved):

tonyhoo (approved):

bryanyzhu (approved):
Thanks for the fix.

PR #2420, opened 2022-11-16
[timeseries] Fix pandas groupby bug + GluonTS index bug

shchur:
*Description of changes:*
- Fix where pandas groupby would sort item ids even when `sort=False`(related to https://github.com/pandas-dev/pandas/issues/17537)
- Fix incorrect forecast index generated by GluonTS and Sktime models in some cases.
  - For example, if data has timestamps `"2022-01-01 12:00"`, `"2022-01-02 12:00"` (frequency is daily), GluonTS will produce forecasts with timestamps `"2022-01-01 00:00"`, `"2022-01-02 00:00"`, which results in an AssertionError inside the `TimeSeriesEvaluator` [here](https://github.com/awslabs/autogluon/blob/master/timeseries/src/autogluon/timeseries/evaluator.py#L203)
- Add tests covering these cases

### DIFF
--- a/timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
+++ b/timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
@@ -109,7 +109,6 @@ class TimeSeriesDataFrame(pd.DataFrame):
         super().__init__(data=data, *args, **kwargs)
         self._static_features: Optional[pd.DataFrame] = None
         if static_features is not None:
-            # TODO: These two checks should go inside the setter
             if isinstance(static_features, pd.Series):
                 static_features = static_features.to_frame()
             if not isinstance(static_features, pd.DataFrame):
@@ -156,10 +155,6 @@ class TimeSeriesDataFrame(pd.DataFrame):
         # subset to ensure consistency
         if value is not None and len(set(value.index) - set(self.item_ids)) > 0:
             value = value.loc[self.item_ids].copy()
-
-        if value is not None and value.index.name != ITEMID:
-            value = value.copy()
-            value.index.rename(ITEMID, inplace=True)
 
         self._static_features = value
 
--- a/timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
+++ b/timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
@@ -109,6 +109,7 @@ class TimeSeriesDataFrame(pd.DataFrame):
         super().__init__(data=data, *args, **kwargs)
         self._static_features: Optional[pd.DataFrame] = None
         if static_features is not None:
+            # TODO: These two checks should go inside the setter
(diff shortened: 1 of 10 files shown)


timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
```diff
@@ -156,6 +157,10 @@ def static_features(self, value: Optional[pd.DataFrame]):
         if value is not None and len(set(value.index) - set(self.item_ids)) > 0:
             value = value.loc[self.item_ids].copy()
 
+        if value is not None and value.index.name != ITEMID:
+            value = value.copy()
+            value.index.rename(ITEMID, inplace=True)
+
```
shchur, line 163: Fixes splitter bug

timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
```diff
@@ -361,7 +362,10 @@ def predict(
                     )
 
         # Make sure the item_ids are sorted in the same order as in data
-        return df.loc[data.item_ids]
+        df = df.loc[data.item_ids]
+        # Set correct timestamp index
+        df.index = get_forecast_horizon_index_ts_dataframe(data, self.prediction_length)
+        return df
```
shchur, line 368: A very ad-hoc fix to the GluonTS bug. Should we look for something cleaner?

canerturkmen: will it work with ignore index?

Innixma (approved):
LGTM!

gradientsky (approved):

timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
```diff
@@ -76,7 +77,7 @@ def __iter__(self) -> Iterator[Dict[str, Any]]:
             df = self.target_df.loc[item_id]
             time_series = {
                 FieldName.ITEM_ID: item_id,
-                FieldName.TARGET: df.squeeze().to_numpy(dtype=self.float_dtype),
```
canerturkmen, line 79: Is this a performance related change?

canerturkmen (approved):
LGTM

timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
shchur: Current code may lead to a bug. If `df` has length 1, then `df.squeeze()` will be a `float` instead of `pd.Series`. There is no clean way to convert a DataFrame to a Series (`squeeze` is the only function recommended for this purpose), so we rather do the flattening into 1D after converting to numpy.

PR #2421, opened 2022-11-16
Print out save_path

bryanyzhu:
*Issue #, if available:*
NA

*Description of changes:*
Print out save_path during fit(), so that users are well informed where are all the outputs are saved


sxjscience (approved):
LGTM

sxjscience:
Just noticed that we should also add here: https://github.com/awslabs/autogluon/blob/bc9094fd83f85ea9ba6e7990c8255aef427105c0/multimodal/src/autogluon/multimodal/matcher.py#L471

tonyhoo (approved):

gradientsky (approved):

FANGAreNotGnu (approved):
LGTM

zhiqiangdon (approved):
LGTM. There is a logger.debug in `setup_save_path`. We can consider making it a logger.info  later.

PR #2433, opened 2022-11-17
bump evaluate to `0.3.0`

lvwerra:
There is an issue with `evaluate` before 0.3.0 concerning the Hugging Face hub integration. I suggest upgrading to `0.3.0` which should not have any breaking changes and be functionally equivalent since this version might break in a few months.


cheungdaven (approved):
LGTM

tonyhoo (approved):
Thanks for your contributions

gidler (approved):

PR #2442, opened 2022-11-18
[Post 0.6][Tabular] Speed up feature transform in tabular NN model

liangfu:
*Description of changes:*

This PR simplifies the feature transform function in tabular NN model.
1. avoid using string-based processing
2. leverage `numpy.isin(b, a, invert=True)` function instead of running `np.array([item in b for item in a])`

Benchmark results are shown below
image

image


liangfu:
cc @tonyhoo @Innixma

Innixma (approved):
Looks awesome, I see an even bigger speed up of around 40% using m6i.16xlarge in my testing!

PR #2444, opened 2022-11-18
Website Revamp PoC

gidler:
Do Not Merge

This PR exists to let people review and make comments on the changes that are on my `new` branch to generate this website: https://auto.gluon.ai/new/index.html


docs/api/autogluon.timeseries.rst
```diff
@@ -0,0 +1,11 @@
...
+====================
+
+
+.. currentmodule:: autogluon.timeseries
+
+.. autosummary::
+    :nosignatures:
+    :toctree: _autogen
+
+    TimeSeriesPredictor
```
shchur, line 11: Could we also add a reference for `TimeSeriesDataFrame` here? Or do we plan to make another pass over the content after the PoC is complete?
```rst
.. autosummary::
    :nosignatures:
    :toctree: _autogen

    TimeSeriesDataFrame
```

gidler: For NeurIPS, we were thinking just the APIs for the Predictors would be adequate. We can add `TimeSeriesDataFrame` (in this case we would probably add `TabularDataset` too) after NeurIPS but I can ask bring up adding it before

shchur: I think including `TimeSeriesDataFrame` can be very helpful for the users since there is no standard way of representing time series data (unlike pandas DataFrames with 1 row = 1 sample for Tabular). Would it be possible to add it before NeurIPS? I will be happy to help with this addition or other parts of this PR

PR #2453, opened 2022-11-21
[timeseries] Gracefully handle inconsistencies in static_features provided by user

shchur:
*Description of changes:*
- Automatically convert the `static_features` to a `pd.DataFrame` if a `pd.Series` is provided
- Always rename the index of `static_features` to `"item_id"`. If we don't rename the index, we will run into an exception inside the splitter [here](https://github.com/awslabs/autogluon/blob/master/timeseries/src/autogluon/timeseries/splitter.py#L63).

### DIFF
--- a/timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
+++ b/timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
@@ -109,12 +109,6 @@ class TimeSeriesDataFrame(pd.DataFrame):
         super().__init__(data=data, *args, **kwargs)
         self._static_features: Optional[pd.DataFrame] = None
         if static_features is not None:
-            if isinstance(static_features, pd.Series):
-                static_features = static_features.to_frame()
-            if not isinstance(static_features, pd.DataFrame):
-                raise ValueError(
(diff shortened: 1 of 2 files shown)


timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
```diff
@@ -148,13 +142,22 @@ def static_features(self, value: Optional[pd.DataFrame]):
...
+                raise ValueError(f"static_features must be a pandas DataFrame (received object of type {type(value)})")
+            missing_item_ids = self.item_ids.difference(value.index)
+            if len(missing_item_ids) > 0:
+                raise ValueError(
+                    f"Following item_ids are missing from the index of static_features: {missing_item_ids.to_list()}"
+                )
+            # if provided static features are a strict superset of the item index, we take a subset to ensure consistency
+            if len(value.index.difference(self.item_ids)) > 0:
+                value = value.loc[self.item_ids].copy()
+            if value.index.name != ITEMID:
```
canerturkmen, line 158: why force the data to a convention we use instead of just replacing the splitter to take the first level of the index ?  especially because we seem to be taking yet another deep copy below just for this purpose.

canerturkmen (approved):

PR #2458, opened 2022-11-22
Fix AutoMM warnings about object detection

zhiqiangdon:
*Issue #, if available:*
If users don't install mmcv, mmdet, or mmocr, they would observe too many warnings.

*Description of changes:*
Check whether mmcv, mmdet, and mmocr are installed only if necessary.


bryanyzhu (approved):
Thanks for the quick fix.

FANGAreNotGnu (approved):
LGTM, I'll replace`"ocr"` with ocr constants later. Thanks for the refactor! This is better than move import into function

zhiqiangdon:
> LGTM, I'll replace`"ocr"` with ocr constants later. Thanks for the refactor! This is better than move import into function

Done. Thanks for the suggestion!

PR #2461, opened 2022-11-22
[Bugfix] Fix matcher cpu inference bug

sxjscience:
*Issue #, if available:*

*Description of changes:*

Previously I only fixed the cpu inference bug (cpu cannot use "dp") in `predictor.py`. However, I just noticed that we should also fix `matcher.py`.


zhiqiangdon (approved):
Thanks for the fix!

FANGAreNotGnu (approved):
LGTM!

PR #2469, opened 2022-11-23
Update timeseries quick start notebook

shchur:
*Description of changes:*
- Update the time series quickstart notebook (`timeseries_quick_start.ipynb`) to correctly collapse sections
- Add notebook in markdown format

### DIFF
--- a/docs/tutorials/timeseries/timeseries_quick_start.ipynb
+++ b/docs/tutorials/timeseries/timeseries_quick_start.ipynb
@@ -36,21 +36,23 @@
    "cell_type": "code",
    "execution_count": null,
    "metadata": {
-    "ExecuteTime": {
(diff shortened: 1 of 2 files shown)


shchur:
It seems that we can also store the notebooks in markdown format. I've added the timeseries notebook as an example. 

Markdown files seem to be much easier to edit, especially if we want to use the MyST tags to collapse inputs/outputs (no need to modify raw JSON), and work better with git. However, we also need to sync the `ipynb` files to git somehow to be able to open them in Colab / Sagemaker studio. This can either be done manually with `jupytext timeseries_quick_start.md --to ipynb`, or with some automated scripts.

Let me know if you think that we don't want to deal with markdown notebooks, I will then remove the markdown version of the tutorial from the PR.

PR #2477, opened 2022-11-25
[Merge After 0.6.1] Update MMLab Processor

FANGAreNotGnu:
- Add MMLabProcessor as an abstract base class. MMDetProcessor/MMOcrProcessor inherit from it and process data for detection/ocr.
- Remove mmlab model related code in ImageProcessor.


multimodal/src/autogluon/multimodal/data/preprocess_dataframe.py
```diff
@@ -379,6 +389,48 @@ def transform_text(
 
         return text_features, text_types
 
+    def transform_rois(
+        self,
+        df: pd.DataFrame,
+    ) -> Tuple[Dict[str, List[List[str]]], Dict[str, str]]:
+        """
+        Preprocess image data by collecting their paths together. If one sample has multiple images
```
bryanyzhu, line 397: Can I ask what is the use case here, i.e., why one sample has multiple images (maybe RoI)?

bryanyzhu (commented):
Thanks for the quick refactor, Will this PR be compatible for future MMseg and MMocr image processor?

FANGAreNotGnu:
> Thanks for the quick refactor, Will this PR be compatible for future MMseg and MMocr image processor?

They should write the processor based on this. For segmentation I think it might be another processor similar to this one. That's the reason I name it rois processor instead of mmlab processor.

multimodal/src/autogluon/multimodal/data/preprocess_dataframe.py
FANGAreNotGnu: Updated the descriptions, but still keep the support for multi images/rois columns. There is no current use cases, and we have limited the number of image columns to 1 in config for now. A possible future example could be we are working on videos and need multiple frames as input per entry.

multimodal/src/autogluon/multimodal/configs/model/fusion_mlp_image_text_tabular.yaml
```diff
@@ -124,16 +124,7 @@ model:
     checkpoint_name: "yolov3_mobilenetv2_320_300e_coco"
     data_types:
       - "image"
```
zhiqiangdon, line 126: `image` -> `rois`?

multimodal/src/autogluon/multimodal/utils/data.py
```diff
@@ -222,6 +231,17 @@ def create_fusion_data_processors(
...
+            # create a multimodal processor for NER.
+            data_processors[ROIS].append(
+                create_data_processor(
+                    data_type=ROIS,
+                    config=config,
+                    model=per_model,
+                )
+            )
+            if data_types is not None and IMAGE in data_types:
+                data_types.remove(IMAGE)
```
zhiqiangdon, line 244: This `elif` seems unnecessary. Consider changing the mmdet_image's data_types in config and reusing the below logic to create a data processor.

multimodal/src/autogluon/multimodal/data/process_rois.py
```diff
@@ -0,0 +1,216 @@
+import logging
```
zhiqiangdon, line 1: It seems that the corresponding code in `process_image.py` is not removed yet.

multimodal/src/autogluon/multimodal/data/process_rois.py
```diff
@@ -0,0 +1,216 @@
...
+
+from ..constants import AUTOMM, COLUMN, IMAGE, IMAGE_VALID_NUM, MMDET_IMAGE
+from .collator import Pad, Stack
+from .utils import is_rois_input
+
+logger = logging.getLogger(AUTOMM)
+ImageFile.LOAD_TRUNCATED_IMAGES = True
+
+
+class RoisProcessor:
```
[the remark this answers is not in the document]
bryanyzhu: One unified mmlab processor would be great. But we can discuss to see if it is feasible.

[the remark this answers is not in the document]
FANGAreNotGnu: Currently RoisProcessor is for both det and ocr. If we want to support mmseg as well, we shall rename it to mmlab processor.

multimodal/src/autogluon/multimodal/data/process_rois.py
FANGAreNotGnu: Yes, but if we remove them we can't load old predictor checkpoints.

multimodal/src/autogluon/multimodal/configs/model/fusion_mlp_image_text_tabular.yaml
FANGAreNotGnu: Image is the only modality for input so I think we shall keep datatype=image. We followed ner's implementation (where datatype is still text), using the problem type to decide which processor to use.

bryanyzhu: I think `image` is better because the input data is still image for most mmlab applications.

multimodal/src/autogluon/multimodal/data/process_rois.py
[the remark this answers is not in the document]
zhiqiangdon: The name of `RoisProcessor` sounds not covering the ocr tasks.

multimodal/src/autogluon/multimodal/data/process_rois.py
zhiqiangdon: We can create a new rois processor during loading and replace the saved one. The refactor seem not done with the existence and usage of old code.

multimodal/src/autogluon/multimodal/utils/data.py
zhiqiangdon: For now, following ner's implementation is fine. We can later refactor the creations of ner and detection processors since the current implementations are hacky.

zhiqiangdon: Besides, do we need `requires_label = False` here? I guess we don't need a separate label processor.

FANGAreNotGnu: Sure I'll remove the label processor in next pr, as well as the hacky implementation.

FANGAreNotGnu: We could change data_type to something like `mmlab_image` to support mmocr and mmseg and also remove this elif.

multimodal/src/autogluon/multimodal/data/process_rois.py
FANGAreNotGnu: Nice suggestion! I'll try.

FANGAreNotGnu: I have finished removing the code from process_image. However I find that to reduce backward compatibility codes we'd better write a processor for all mmlab models to avoid future refactors that change the processor's API.

multimodal/src/autogluon/multimodal/utils/data.py
zhiqiangdon: I think we can still use `image` and check `problem_type` to determine which processor. (image processor or mmlab processor). `mmlab_image` may sound not intuitive for users.

multimodal/src/autogluon/multimodal/data/process_mmlab/process_mmdet.py
```diff
@@ -0,0 +1,100 @@
...
+
+from ...constants import AUTOMM
+from ..utils import is_rois_input
+from .process_mmlab_base import MMLabProcessor
+
+logger = logging.getLogger(AUTOMM)
+ImageFile.LOAD_TRUNCATED_IMAGES = True
+
+
+class MMDetProcessor(MMLabProcessor):
```
zhiqiangdon, line 24: It seems that `MMDetProcessor ` has the same implementation as `MMOcrProcessor`. Why do we need to define both?

multimodal/src/autogluon/multimodal/utils/data.py
```diff
@@ -153,6 +156,12 @@ def create_data_processor(
             model=model,
             max_len=model_config.max_text_len,
         )
+    elif data_type == ROIS:
```
zhiqiangdon, line 159: In this case, both ocr and detection use `MMDetProcessor `. Need to consider the design.

multimodal/src/autogluon/multimodal/utils/data.py
```diff
@@ -193,6 +202,7 @@ def create_fusion_data_processors(
         NUMERICAL: [],
         LABEL: [],
         NER: [],
+        ROIS: [],
```
zhiqiangdon, line 205: One thing to think about : are rois and ner are proper data types?

multimodal/src/autogluon/multimodal/data/process_mmlab/process_mmdet.py
FANGAreNotGnu: Currently our implementation for MMOCR only includes simple inference, and it temporarily reuses the logic from MMDet. Later they should be different

multimodal/src/autogluon/multimodal/utils/data.py
FANGAreNotGnu: We may also need to add the problem_type here while we work on OCR. The current OCR implementation needs a refactor and unit tests are blocked.

multimodal/src/autogluon/multimodal/utils/data.py
FANGAreNotGnu: ROIS should be data types, and we could enable it only in certain problem_types in the future if there is a conflict.

zhiqiangdon (approved):
Thanks for the refactor! We can optimize the data processor creations and distinguish mmdet and mmocr processors later.

bryanyzhu (approved):
Great, thanks for the refactor. Let's merge it after 0.6.1 release.

PR #2490, opened 2022-11-28
Support image bytearray in AutoMM

suzhoum:
*Issue #, if available:*

*Description of changes:*
This PR adds support for image bytearrays in AutoMM.
The DataFrame column can be either `bytearray` or `list[bytearray]`.

Benchmarking on the original `image_path` workflow with [dataset](https://automl-mm-bench.s3.amazonaws.com/petfinder_for_tutorial.zip) on SageMaker endpoint

test data:
100 x bytearray-encoded images 

data preprocess:
0.5s to decode and save to disk to generate `image_path` to pass in `MultiModalPredictor.fit()`

predict_proba:
0.1s

With the support of bytearray, we can save the time spent on data preprocessing, which majority of the time is on saving the image to disk.

### DIFF
--- a/multimodal/src/autogluon/multimodal/data/infer_types.py
+++ b/multimodal/src/autogluon/multimodal/data/infer_types.py
@@ -229,6 +229,8 @@ def is_imagebytearray_column(
         One column of a multimodal pd.DataFrame for training.
     col_name
         Name of column.
+    sample_n
+        Number of sample images to open for sanity check.
 
     Returns
     -------
@@ -259,7 +261,7 @@ def is_imagebytearray_column(
         if failure_ratio > 0:
             warnings.warn(
                 f"Among {sample_num} sampled images in column '{col_name}', "
-                f"{failure_ratio:.0%} images can't be open. "
+                f"{failure_count} images can't be open. "
                 "You may need to thoroughly check your data to see the percentage of missing images, "
                 "and estimate the potential influence. By default, we skip the samples with missing images. "
                 "You can also set hyperparameter 'data.image.missing_value_strategy' to be 'zero', "
--- a/multimodal/src/autogluon/multimodal/data/infer_types.py
+++ b/multimodal/src/autogluon/multimodal/data/infer_types.py
@@ -2,6 +2,7 @@ import collections
 import json
 import logging
 import warnings
+from io import BytesIO
 from typing import Dict, List, Optional, Tuple, Union
 
 import pandas as pd
@@ -15,6 +16,7 @@ from ..constants import (
     ENTITY_GROUP,
     IDENTIFIER,
     IMAGE,
+    IMAGE_BYTEARRAY,
     IMAGE_PATH,
     MULTICLASS,
     NAMED_ENTITY_RECOGNITION,
@@ -84,20 +86,24 @@ def is_categorical_column(
                 ratio = 0.1
         threshold = min(int(len(data) * ratio), threshold)
         data_value_counts = data.value_counts(dropna=False)
-        key_set = set(data_value_counts.keys())
-        if len(data_value_counts) < threshold:
-            valid_value_counts = valid_data.value_counts(dropna=False)
-            total_valid_num = len(valid_data)
-            oov_num = 0
-            for k, v in zip(valid_value_counts.keys(), valid_value_counts.values):
-                if k not in key_set:
-                    oov_num += v
-            if is_label and oov_num != 0:
+        try:
+            key_set = set(data_value_counts.keys())
+            if len(data_value_counts) < threshold:
+                valid_value_counts = valid_data.value_counts(dropna=False)
+                total_valid_num = len(valid_data)
+                oov_num = 0
+                for k, v in zip(valid_value_counts.keys(), valid_value_counts.values):
+                    if k not in key_set:
+                        oov_num += v
+                if is_label and oov_num != 0:
+                    return False
+                if oov_num / total_valid_num > oov_ratio_threshold:
+                    return False
+                return True
+            else:
                 return False
-            if oov_num / total_valid_num > oov_ratio_threshold:
-                return False
-            return True
-        return False
+        except:
+            return False
 
 
 def is_rois_column(data: pd.Series) -> bool:
@@ -208,6 +214,61 @@ def is_imagepath_column(
         return False
 
 
+def is_imagebytearray_column(
+    data: pd.Series,
+    col_name: str,
+) -> bool:
+    """
+    Identify if a column contains valid image bytearray.
+
+    Parameters
+    ----------
+    data
+        One column of a multimodal pd.DataFrame for training.
+    col_name
+        Name of column.
+
+    Returns
+    -------
+    Whether the column is an image-bytearray column.
+    """
+    sample_num = min(len(data), 500)
+    data = data.sample(n=sample_num, random_state=0)
+    data = data.tolist()
+    failure_count = 0
+    for image_bytearrays in data:
+        success = False
+        if not isinstance(image_bytearrays, list):
+            image_bytearrays = [image_bytearrays]
+
+        for img_bytearray in image_bytearrays:
+            try:
+                with PIL.Image.open(BytesIO(img_bytearray)) as img:
+                    pass
+                success = True
+                break
+            except:
+                pass
+            if not success:
+                failure_count += 1
+    failure_ratio = failure_count / sample_num
+    # Tolerate high failure rate in case that many image files may be corrupted.
+    if failure_ratio <= 0.9:
+        if failure_ratio > 0:
+            warnings.warn(
+                f"Among {sample_num} sampled images in column '{col_name}', "
+                f"{failure_ratio:.0%} images can't be open. "
+                "You may need to thoroughly check your data to see the percentage of missing images, "
+                "and estimate the potential influence. By default, we skip the samples with missing images. "
+                "You can also set hyperparameter 'data.image.missing_value_strategy' to be 'zero', "
+                "which uses a zero image to replace any missing image.",
+                UserWarning,
+            )
+        return True
+    else:
+        return False
+
+
 def is_text_column(data: pd.Series) -> bool:
     """
     Identify if a column is one text column.
@@ -224,19 +285,21 @@ def is_text_column(data: pd.Series) -> bool:
     if len(data) > 5000:
         # Sample to speed-up type inference
         data = data.sample(n=5000, random_state=0)
-    data_unique = data.unique()
-    num_unique = len(data_unique)
-    num_rows = len(data)
-    unique_ratio = num_unique / num_rows
-    if unique_ratio <= 0.01:
-        return False
     try:
-        avg_words = pd.Series(data_unique).str.split().str.len().mean()
-    except AttributeError:
+        data_unique = data.unique()
+        num_unique = len(data_unique)
+        num_rows = len(data)
+        unique_ratio = num_unique / num_rows
+        if unique_ratio <= 0.01:
+            return False
+        try:
+            avg_words = pd.Series(data_unique).str.split().str.len().mean()
+        except AttributeError:
+            return False
+        if avg_words < 3:
+            return False
+    except:
         return False
-    if avg_words < 3:
-        return False
-
     return True
 
 
@@ -311,6 +374,8 @@ def infer_id_mappings_types(id_mappings: Union[Dict[str, Dict], Dict[str, pd.Ser
             )
         if is_imagepath_column(per_id_mappings, col_name=per_name):
             id_mappings_types[per_name] = IMAGE_PATH
+        if is_imagebytearray_column(per_id_mappings, col_name=per_name):
+            id_mappings_types[per_name] = IMAGE_BYTEARRAY
         elif is_text_column(per_id_mappings):
             id_mappings_types[per_name] = TEXT
         else:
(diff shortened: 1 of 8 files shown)


multimodal/tests/unittests/predictor/test_predictor.py
```diff
@@ -621,3 +632,39 @@ def test_modifying_duplicate_model_names():
     for per_modality_processors in teacher_predictor._data_processors.values():
         for per_processor in per_modality_processors:
             assert per_processor.prefix in teacher_predictor._config.model.names
+
+
+def test_image_bytearray():
+    dataset_1 = ALL_DATASETS["petfinder"]
```
bryanyzhu, line 638: Can we add a vision-only dataset for unittest, like shopee? In this way, we don't need to specify other models names and hyperparameters.

suzhoum: Good point, just updated.

cheungdaven (approved):
LGTM!

multimodal/tests/unittests/predictor/test_predictor.py
```diff
@@ -621,3 +624,35 @@ def test_modifying_duplicate_model_names():
...
+        label="label",
+    )
+    predictor_2 = MultiModalPredictor(
+        label="label",
+    )
+    model_names = ["timm_image"]
+    hyperparameters = {
+        "optimization.max_epochs": 2,
+        "model.names": model_names,
+        "model.timm_image.checkpoint_name": "swin_tiny_patch4_window7_224",
```
bryanyzhu, line 643: Sorry for not mentioning before, since this is a new input format, we want to test it as comprehensive as possible. So can you (1) add a CNN model like `mobilenetv3_large100` for testing (2) test `.predict()` and `.predict_proba()` functions as well in the unit test? Thank you.

bryanyzhu (approved):
LGTM with a minor suggestion for unit test.

multimodal/tests/unittests/predictor/test_predictor.py
suzhoum: Updated
(1) use CNN model https://github.com/autogluon/autogluon/blob/909b7c3217121fa6a77ad6f23c5b235a77a51d96/multimodal/tests/unittests/predictor/test_predictor.py#L125
(2) test `.predict()` and `.predict_proba()`: https://github.com/autogluon/autogluon/blob/909b7c3217121fa6a77ad6f23c5b235a77a51d96/multimodal/tests/unittests/predictor/test_predictor.py#L663-L664

multimodal/src/autogluon/multimodal/data/infer_types.py
```diff
@@ -208,6 +214,61 @@ def is_imagepath_column(
...
+    data
+        One column of a multimodal pd.DataFrame for training.
+    col_name
+        Name of column.
+
+    Returns
+    -------
+    Whether the column is an image-bytearray column.
+    """
+    sample_num = min(len(data), 500)
```
sxjscience, line 235: Can we make the `500` as a configurable variable?

suzhoum: Sure, updated!

multimodal/src/autogluon/multimodal/data/process_image.py
```diff
@@ -443,7 +448,11 @@ def process_one_sample(
                         )
                         is_zero_img = False
                         try:
-                            with PIL.Image.open(img_path) as img:
+                            if feature_modalities.get(per_col_name) == IMAGE_BYTEARRAY:
+                                image_feature = BytesIO(img_feature)
```
tonyhoo, line 452: The assumption here seems to be bytesarray will follow the `RBG` channel standard. I think we should default to `RBG` while allowing user to pass different channel sequence

tonyhoo (commented):
Thanks for the change. Some follow up actions in general:
* We should identify any doc/tutorial to be updated with this feature as it allows a new input type and we should make it visible via doc to the end user
* Can we build docker images for both GPU and CPU to see the inference impact on Sagemaker endpoint? If already done, ignore this point

multimodal/src/autogluon/multimodal/data/infer_types.py
```diff
@@ -84,20 +86,24 @@ def is_categorical_column(
...
+                if is_label and oov_num != 0:
+                    return False
+                if oov_num / total_valid_num > oov_ratio_threshold:
+                    return False
+                return True
+            else:
                 return False
-            return True
-        return False
+        except:
```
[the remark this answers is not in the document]
suzhoum: Bytearray can't be hashed, so that `set(data_value_counts.keys())` will error out when the column contains bytearray. This function is checking if column contains categorical values, so in this case we just want to return `False`.

multimodal/src/autogluon/multimodal/data/infer_types.py
```diff
@@ -224,19 +287,21 @@ def is_text_column(data: pd.Series) -> bool:
...
+        unique_ratio = num_unique / num_rows
+        if unique_ratio <= 0.01:
+            return False
+        try:
+            avg_words = pd.Series(data_unique).str.split().str.len().mean()
+        except AttributeError:
+            return False
+        if avg_words < 3:
+            return False
+    except:
```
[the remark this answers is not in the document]
suzhoum: Similar as above, we can't call `data.unique()` on bytearray as the function is checking text modality. The expected behavior is to return `False`.

multimodal/src/autogluon/multimodal/data/infer_types.py
```diff
@@ -208,6 +215,62 @@ def is_imagepath_column(
...
+    data = data.tolist()
+    failure_count = 0
+    for image_bytearrays in data:
+        success = False
+        if not isinstance(image_bytearrays, list):
+            image_bytearrays = [image_bytearrays]
+
+        for img_bytearray in image_bytearrays:
+            try:
+                with PIL.Image.open(BytesIO(img_bytearray)) as img:
```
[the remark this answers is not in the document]
suzhoum: The benchmarking was actually done on a Sagemaker inference endpoint so it would have not gone through this `infer_type` procedure. This is only done during the initial phase of training, which time should be negligible compared to the training time.

sxjscience (approved):
LGTM

suzhoum:
> Thanks for the change. Some follow up actions in general:
> 
> * We should identify any doc/tutorial to be updated with this feature as it allows a new input type and we should make it visible via doc to the end user
> * Can we build docker images for both GPU and CPU to see the inference impact on Sagemaker endpoint? If already done, ignore this point

Thanks for the suggestions! We can mention this feature in our tutorials in a follow up PR. The benchmark was done on a GPU instance and I can build a CPU endpoint as well.

multimodal/src/autogluon/multimodal/data/process_image.py
suzhoum: We can definitely make this configurable. I might be wrong but it seems that most pretrained models we support are trained on RBG images, @bryanyzhu do you have any suggestions?

bryanyzhu: In my understanding, most pretrained image models are trained on RGB images in `RGB` channel standard. Only opencv uses `RBG` channel as default.

multimodal/src/autogluon/multimodal/data/infer_types.py
```diff
@@ -208,6 +215,64 @@ def is_imagepath_column(
...
```
zhiqiangdon, line 250: What if the image bytes are encoded by base64?

multimodal/src/autogluon/multimodal/data/infer_types.py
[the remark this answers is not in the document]
zhiqiangdon: Instead of adding this try-except, how about using `is_imagebytearray_column ` ahead of `is_categorical_column `?

multimodal/src/autogluon/multimodal/data/infer_types.py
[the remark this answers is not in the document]
zhiqiangdon: In below,  `is_text_column ` is checked after `is_imagebytearray_column `. Do we still need this try-except?

multimodal/src/autogluon/multimodal/data/infer_types.py
suzhoum: Can we enforce that the type passed into the predictor to be the original `bytearray`? In the cloud serve scripts, I see that currently we would: 1. decode the encoded bytearray in a predefined way, and 2. save to disk to get the path to pass into the predictor. With the new feature, we can save step 2 by assuming no encoding on bytearray. If we are to expect encoded bytearrays in the dataframe, we would probably need to accept a callback decoding function provided by the user. What do you think?

multimodal/src/autogluon/multimodal/utils/misc.py
```diff
@@ -45,7 +45,27 @@ def tensor_to_ndarray(tensor: torch.Tensor):
...
+        b = bytearray(f)
+    return b
+
+
+def path_to_bytearray_expander(path, base_folder):
+    path_l = path.split(";")
+    return [_read_byte(os.path.abspath(os.path.join(base_folder, path))) for path in path_l]
+
+
+def shopee_dataset(
```
zhiqiangdon, line 65: Besides unit tests, where do we use this dataset? It doesn't fit the context of this file.

multimodal/src/autogluon/multimodal/data/process_image.py
```diff
@@ -421,19 +425,20 @@ def process_one_sample(
...
-            for per_col_name, per_col_content in image_paths.items():
+            for per_col_name, per_col_content in image_features.items():
                 if is_rois_input(per_col_content):
                     rois = np.array(per_col_content)
                     # TODO: add gt masks
                     mm_data["ann_info"] = dict(bboxes=rois[:, :4], labels=rois[:, 4], masks=[])
                 else:
-                    with PIL.Image.open(per_col_content[0]) as img:
-                        mm_data["img_info"] = dict(filename=per_col_content[0], height=img.height, width=img.width)
+                    if feature_modalities.get(per_col_name) == IMAGE_PATH:
```
zhiqiangdon, line 434: What if detection uses image bytearray?

multimodal/src/autogluon/multimodal/data/infer_types.py
zhiqiangdon: base64 is used in encoding images in some cases. I think we'd better support users to provide base64 bytes in dataframe.

multimodal/src/autogluon/multimodal/data/infer_types.py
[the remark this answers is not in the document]
suzhoum: That should work, but I'm curious on the time taken on 500 failed attempts to open images versus other checks if we move `is_imagebytearray_column` near the top of the logic flow. Will it take more time in general, or if this is negligible?

multimodal/src/autogluon/multimodal/data/process_image.py
suzhoum: Thanks for the info, that's great to know! I've added this to be configurable.

multimodal/src/autogluon/multimodal/utils/misc.py
suzhoum: It is used in a few tutorials at the moment. Do you think creating a `datasets.py` under the `./utils` will help?

multimodal/src/autogluon/multimodal/data/process_image.py
suzhoum: It is currently not supported in this PR. `mmdet` doesn't directly support load image from bytearray, so it is a bit more work to add the support. I can add in the next PR.

multimodal/src/autogluon/multimodal/data/infer_types.py
```diff
@@ -84,20 +86,24 @@ def is_categorical_column(
                 ratio = 0.1
         threshold = min(int(len(data) * ratio), threshold)
         data_value_counts = data.value_counts(dropna=False)
```
gidler, line 88: I believe this call also fails if `data` is a bytearray because `value_counts` requires hashing the entries (AFAIK).

FANGAreNotGnu (approved):
LGTM

multimodal/src/autogluon/multimodal/data/infer_types.py
gidler: `base64` is fundamentally a different format than `bytearray`, since it is a text encoding rather than a binary encoding. We could consider adding a `IMAGE_BASE64` column type, or we could try to make a more generic `IMAGE_BINARY` type that also infers the encoding. But I think this is another feature request separate from the current issue.

multimodal/src/autogluon/multimodal/data/infer_types.py
suzhoum: It didn't error out during test, but I agree it would be safer to include this inside the `try` as well. Thanks for pointing it out!

tonyhoo (approved):

multimodal/src/autogluon/multimodal/data/infer_types.py
zhiqiangdon: Since base64 requires additional decoding, we may define another type `IMAGE_BASE64`. We can leave this to another PR.

multimodal/src/autogluon/multimodal/data/infer_types.py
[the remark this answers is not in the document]
zhiqiangdon: Make sense.

multimodal/src/autogluon/multimodal/utils/misc.py
zhiqiangdon: We usually put the data preparation logic inside the tutorials, e.g.,  https://auto.gluon.ai/stable/tutorials/multimodal/multimodal_prediction/beginner_multimodal.html.

PR #2491, opened 2022-11-28
[CI] Moved short_sha from env variable to the running command

tonyhoo:
*Issue #, if available:*
Fix the CI [failure](https://github.com/autogluon/autogluon/actions/runs/3562103876/jobs/5993588145) at copy_doc step

*Description of changes:*
Moved `short_sha` from env variable to the running command

### DIFF
--- a/.github/actions/setup-env-vars/action.yml
+++ b/.github/actions/setup-env-vars/action.yml
@@ -10,9 +10,10 @@ runs:
         SHA: ${{ github.sha }}
         branch: ${{ github.ref }}
         git_repo: ${{ github.repository }}
-        short_sha: $(git rev-parse --short "${{ github.sha }}")
         pr_number: ""
       run: |
+        
+        short_sha=$(git rev-parse --short "$SHA")
         echo "BRANCH=$branch" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


.github/actions/setup-env-vars/action.yml
```diff
@@ -10,9 +10,10 @@ runs:
         SHA: ${{ github.sha }}
         branch: ${{ github.ref }}
         git_repo: ${{ github.repository }}
-        short_sha: $(git rev-parse --short "${{ github.sha }}")
         pr_number: ""
       run: |
+        
+        short_sha=$(git rev-parse --short "$SHA")
```
suzhoum, line 16: Just curious, what's causing it to fail now?

suzhoum (approved):

.github/actions/setup-env-vars/action.yml
tonyhoo: The reason is that `short_sha` depends on git command to shrink the hash. However, it doesn't work in the same way when it is defined under the `env`. Hence, moving it to runnable command should resolve the issue

gradientsky (approved):

PR #2492, opened 2022-11-28
Mmdet Fix Image Identifier

FANGAreNotGnu:
Previously we used the number in image file name as identifier for image path. But the assumption that image name contains a unique number is not valid. Here we changed to use the image file name (without extension) as the identifier.


multimodal/src/autogluon/multimodal/utils/object_detection.py
```diff
@@ -514,13 +514,10 @@ def VOCName2Idx(name):
     return VOC_NAME_TO_IDX[name]
 
 
-def get_image_name_num(path):
-    if "/potholes" in path:
-        start_idx = path.rfind("/potholes") + 9
-    else:
-        start_idx = path.rfind("/") + 1
+def get_image_filename(path):
```
bryanyzhu, line 517: One comment is, for functions inside utils.py, it is also better to include a docstring (and add an example would be the best). Because these functions might be reused in a lot of scenarios, users need to know what the function is doing.

bryanyzhu (approved):
Thanks for the PR. Please add a docstring before merging.

multimodal/src/autogluon/multimodal/utils/object_detection.py
FANGAreNotGnu: Done for function modified here. I'll use another PR to go through all the object utils.

yongxinw (approved):
LGTM!

multimodal/src/autogluon/multimodal/utils/object_detection.py
```diff
@@ -514,36 +514,70 @@ def VOCName2Idx(name):
...
+        The path of image.
+
+    Returns
+    -------
+    The file name of image.
+    """
+    start_idx = path.rfind("/") + 1
     end_idx = path.rindex(".")
-    return int(path[start_idx:end_idx])
+    return path[start_idx:end_idx]
```
gradientsky, line 532: This will not work on Windows since it's using `\` as a path separators. I suggest just using `pathlib` + replace backslashes to linux format:

```python
from pathlib import Path

def get_image_filename(path: str):
    return Path(s.replace('\\', '/')).stem
```

FANGAreNotGnu: Done.  Thanks!

gradientsky (approved):

PR #2500, opened 2022-11-30
[NER] add an NER visualizer

cheungdaven:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/multimodal/src/autogluon/multimodal/utils/misc.py
+++ b/multimodal/src/autogluon/multimodal/utils/misc.py
@@ -1,5 +1,6 @@
 import logging
 import os
+from random import randint
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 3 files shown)


multimodal/src/autogluon/multimodal/utils/misc.py
```diff
@@ -74,3 +75,114 @@ def path_expander(path, base_folder):
...
+    def _repr_html_(self):
+        entities = []
+        new_sent = ""
+        last = 0
+        for key, value in self.spans.items():
+            entity_group = value[-1]
+            if entity_group.startswith("B-") or entity_group.startswith("I-"):
+                entity_group = entity_group[2:]
+            if entity_group not in self.colors:
+                self.colors.update({entity_group: "#%06X" % randint(0, 0xFFFFFF)})
```
sxjscience, line 161: We can potentially add a “seed” option in initializer and draw random integers from the random state: https://numpy.org/doc/stable/reference/random/legacy.html?highlight=numpy%20random%20randomstate#numpy.random.RandomState

this will ensure that the colors picked for each NERVisualizer instance will be the same

multimodal/src/autogluon/multimodal/utils/misc.py
```diff
@@ -74,3 +75,114 @@ def path_expander(path, base_folder):
...
+    sentence
+        The input sentence.
+    prediction
+        The NER prediction for the sentence.
+
+    Returns
+    -------
+    An NER html visualizer.
+    """
+    visualizer = NERVisualizer(prediction, sentence)
```
zhiqiangdon, line 187: A little curious about the advantages of defining class `NERVisualizer `. Is the visualization effect the same if we call a function with the same functionality as the above `_repr_html_`?

zhiqiangdon (approved):
Visualization looks great!

multimodal/src/autogluon/multimodal/utils/misc.py
suzhoum: Can we add a docstring to the `_repr_html_` so that developers would know about this interface?

suzhoum (approved):
LGTM!

bryanyzhu (approved):
Looks great.

PR #2509, opened 2022-12-03
Error on non-finite values in label column

gidler:
*Issue #, if available:* [2443](https://github.com/autogluon/autogluon/issues/2443)

*Description of changes:*
After discussion with a customer, we decided that AutoGluon should crash if any un-representable value (`NaN`, `Inf`, `NINF`) is present in the label column, because prediction behavior is not well specified for these rows. Infinite values don't make sense for regression problems but could be treated as additional classes or ignored for classification problems. `NaN` values could be ignored for regression problems but could be treated as a separate class or ignored in regression problems. 

This set of commit changes the behavior of AutoGluon to always raise an Error if these values are present in the label column of `train_data` or `tune_data` so that the user has to explicitly handle these values ahead of time. *This could cause AutoGluon runs that executed successful before this change to fail*.

Previously Behavior:
- If a `NaN` value was passed in the label column of `train_data`, the record was dropped and a warning was logged.
- If an infinite value was passed in a label column of `train_data`, it would get passed on to predictors in the ensemble. Some handled it, some failed.
- If `NaN` values or infinite values were in the label column of the tuning_data, the data would get passed on as-is to predictors in the ensemble.

New Behavior:
- If a `NaN`, `Inf`, or `NINF` value is present in the label column for `train_data` or `tune_data`,  a `ValueError` is raised stating "Label column cannot contain non-finite values (NaN, Inf, Ninf)".

I also added unit testing for the "infer_probem_type" utility function and moved constants to a separate file.

### DIFF
--- a/core/src/autogluon/core/constants.py
+++ b/core/src/autogluon/core/constants.py
@@ -25,6 +25,12 @@ OBJECTIVES_TO_NORMALIZE = ['log_loss', 'pac_score', 'soft_log_loss']  # do not l
 AUTO_WEIGHT = 'auto_weight'
 BALANCE_WEIGHT = 'balance_weight'
 
+# Constants used to infer problem type
+MULTICLASS_UPPER_LIMIT = 1000  # assume regression if dtype is numeric and unique label count is above this limit
+LARGE_DATA_THRESHOLD = 1000
+REGRESS_THRESHOLD_LARGE_DATA = 0.05
+REGRESS_THRESHOLD_SMALL_DATA = 0.1
+
 # TODO: Add docs to dedicated page, or should it live in AbstractModel?
 # TODO: How to reference correct version of docs?
 # TODO: Add error in AG_ARGS if unknown key present
--- a/core/tests/unittests/utils/test_utils.py
+++ b/core/tests/unittests/utils/test_utils.py
@@ -1,0 +1,80 @@
+import unittest
+
+import numpy as np
+import pandas as pd
+
+from autogluon.core.utils import infer_problem_type
+from autogluon.core.constants import BINARY, MULTICLASS, MULTICLASS_UPPER_LIMIT, REGRESSION
+
+
+class TestInferProblemType(unittest.TestCase):
+
+    def test_infer_problem_type_empty(self):
+        with self.assertRaises(ValueError):
+            infer_problem_type(pd.Series([], dtype="float"))
+
+    def test_infer_problem_type_nan(self):
+        with self.assertRaises(ValueError):
+            infer_problem_type(pd.Series([np.nan]))
+
+    def test_infer_problem_type_inf(self):
+        with self.assertRaises(ValueError):
+            infer_problem_type(pd.Series([np.inf]))
+
+    def test_infer_problem_type_ninf(self):
+        with self.assertRaises(ValueError):
+            infer_problem_type(pd.Series([np.NINF]))
+
+    def test_infer_problem_type_binary(self):
+        inferred_problem_type = infer_problem_type(pd.Series([-1, -1, 99, -1, -1, 99]))
+        assert inferred_problem_type == BINARY
+
+    def test_infer_problem_type_binary_with_nan(self):
+        inferred_problem_type = infer_problem_type(pd.Series([-1, -1, 99, -1, -1, 99, np.nan]))
+        assert inferred_problem_type == BINARY
+
+    def test_infer_problem_type_str(self):
+        inferred_problem_type = infer_problem_type(pd.Series(["a", "b", "c"], dtype=str))
+        assert inferred_problem_type == MULTICLASS
+
+    def test_infer_problem_type_category(self):
+        inferred_problem_type = infer_problem_type(pd.Series(["a", "b", "c"], dtype='category'))
+        assert inferred_problem_type == MULTICLASS
+
+    def test_infer_problem_type_object(self):
+        inferred_problem_type = infer_problem_type(pd.Series(["a", "b", "c"], dtype='object'))
+        assert inferred_problem_type == MULTICLASS
+
+    def test_infer_problem_type_multiclass_with_nan(self):
+        inferred_problem_type = infer_problem_type(pd.Series(["a", "b", "c", np.nan]))
+        assert inferred_problem_type == MULTICLASS
+
+    def test_infer_problem_type_big_float_data_regression(self):
+        big_float_regression_series = pd.Series(np.repeat(np.linspace(0.0, 1.0, MULTICLASS_UPPER_LIMIT+1), 2))
+        inferred_problem_type = infer_problem_type(big_float_regression_series)
+        assert inferred_problem_type == REGRESSION
+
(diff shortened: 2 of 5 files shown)


core/src/autogluon/core/constants.py
```diff
@@ -25,6 +25,12 @@
 AUTO_WEIGHT = 'auto_weight'
 BALANCE_WEIGHT = 'balance_weight'
 
+# Constants used to infer problem type
+MULTICLASS_LIMIT = 1000  # assume regression if dtype is numeric and unique label count is above this limit
```
tonyhoo, line 29: [minor] probably name it more explicitly to tell if it is an upper_bound limit or lower_bound limit

tabular/src/autogluon/tabular/learner/default_learner.py
```diff
@@ -166,14 +166,18 @@ def general_data_processing(self, X: DataFrame, X_val: DataFrame, X_unlabeled: D
...
 
-        # Remove all examples with missing labels from this dataset:
-        missinglabel_inds = list(X[X[self.label].isna()].index)
-        if len(missinglabel_inds) > 0:
-            logger.warning(f"Warning: Ignoring {len(missinglabel_inds)} (out of {len(X)}) training examples for which the label value in column '{self.label}' is missing")
-            X = X.drop(missinglabel_inds, axis=0)
+        if pd.api.types.is_float_dtype(X[self.label]):
+            has_invalid_labels = np.any(~np.isfinite(X[self.label]))
+            if has_invalid_labels:
+                raise ValueError("Label column cannot contain non-finite values (NaN, Inf, Ninf)")
```
tonyhoo, line 172: [minor] for the convenience of the user to debug the issue in the label. can we add one line here to indicate the first index of the dataset which has non supported values?

tabular/src/autogluon/tabular/learner/default_learner.py
```diff
@@ -166,14 +166,18 @@ def general_data_processing(self, X: DataFrame, X_val: DataFrame, X_unlabeled: D
...
 
         holdout_frac_og = holdout_frac
         if X_val is not None and self.label in X_val.columns:
+            if pd.api.types.is_float_dtype(X[self.label]):
+                has_invalid_tuning_labels = np.any(~np.isfinite(X_val[self.label]))
+                if has_invalid_tuning_labels:
+                    raise ValueError("Tuning data label column cannot contain non-finite values (NaN, Inf, Ninf)")
```
tonyhoo, line 179: ditto

core/tests/unittests/utils/test_utils.py
```diff
@@ -0,0 +1,68 @@
...
+
+class TestInferProblemType(unittest.TestCase):
+
+    def test_infer_problem_type_empty(self):
+        with self.assertRaises(ValueError):
+            infer_problem_type(pd.Series([], dtype="float"))
+
+    def test_infer_problem_type_non_finite(self):
+        with self.assertRaises(ValueError):
+            infer_problem_type(pd.Series([np.nan, np.inf, np.NINF]))
```
gradientsky, line 18: Can we split this one into 3 different cases (since each one is special)?

tabular/src/autogluon/tabular/learner/default_learner.py
```diff
@@ -166,14 +166,20 @@ def general_data_processing(self, X: DataFrame, X_val: DataFrame, X_unlabeled: D
...
+            if np.any(invalid_labels):
+                first_invalid_label_idx = invalid_labels.argmax()
+                raise ValueError(f"Label column cannot contain non-finite values (NaN, Inf, Ninf). First invalid label at idx: {first_invalid_label_idx}")
 
         holdout_frac_og = holdout_frac
         if X_val is not None and self.label in X_val.columns:
+            if pd.api.types.is_float_dtype(X[self.label]):
+                invalid_tuning_labels = ~np.isfinite(X_val[self.label])
+                if np.any(invalid_tuning_labels):
+                    first_invalid_label_idx = invalid_tuning_labels.argmax()
```
gradientsky, line 180: would argmax correctly work on np.NINF?

core/tests/unittests/utils/test_utils.py
gidler: Updated in new revision

tabular/src/autogluon/tabular/learner/default_learner.py
gidler: Yes, it does work because `np.isfinite()` returns numpy array of binary values where all finite values are `True` and all others (`nan`, `inf`, `ninf`) will be `False`. I negate that array and if `any` of the entries are `True`, I return `argmax()` which will return the index of the first `True` entry, which could be any sort of non-finite value.

I spent a while trying to file the most pythonic way to find the index of the first `True` value in a numpy array of binary values, and `argmax()` was the best I could come up with. If you know of a more elegant way to accomplish this, I would be happy to change it because `argmax()` does seem a bit out of place here.

gidler: I update the unit test to demonstrate it works for `np.inf` and `np.NINF`

core/tests/unittests/utils/test_utils.py
```diff
@@ -0,0 +1,76 @@
+import unittest
+
+import numpy as np
+import pandas as pd
+
+from autogluon.core.utils import infer_problem_type
+from autogluon.core.constants import BINARY, MULTICLASS, MULTICLASS_UPPER_LIMIT, REGRESSION
+
+
+class TestInferProblemType(unittest.TestCase):
```
Innixma, line 10: Missing a test with mix of NaN and valid values

tabular/tests/unittests/test_tabular.py
```diff
@@ -974,3 +974,19 @@ def test_tabular_bagstack_use_bag_holdout():
     ###################################################################
     run_tabular_benchmarks(fast_benchmark=fast_benchmark, subsample_size=subsample_size, perf_threshold=perf_threshold,
                            seed_val=seed_val, fit_args=fit_args, run_distill=True, crash_in_oof=True)
+
+
+def test_tabular_raise_on_nonfinite_labels():
```
Innixma, line 979: What about `["a", "b", np.nan]`? Or is this not possible?

Innixma (approved):
LGTM, had 2 minor comments. Very happy this change is occurring, going to make things a lot less confusing in the edge cases.

tabular/tests/unittests/test_tabular.py
gidler: This was a great question. I wasn't sure if/how `pandas` and `numpy` would handle that. I found a better way to handle this case as well and pushed an update with a new Unit Test to show it correct fails on this case as well.

core/tests/unittests/utils/test_utils.py
gidler: Added

Innixma (changes requested):
Blocking since it appears a unit test in tabular is failing.

Innixma (approved):

Issue #2498, opened 2022-11-30
[BUG] TimeSeriesPredictor.info() raises an exception and then exits


fantauzzi:
- [ X] I have checked that this bug exists on the latest stable version of AutoGluon
- [ ] and/or I have checked that this bug exists on the latest mainline of AutoGluon via source installation

**Describe the bug**
`TimeSeriesPredictor.info()` raises an exception `AttributeError: 'str' object has no attribute 'name'`

```
Traceback (most recent call last):
  File "/home/fanta/.pyenv/versions/python3.9.14/lib/python3.9/site-packages/autogluon/core/models/abstract/abstract_model.py", line 1436, in load_info
    return load_pkl.load(path=load_path)
  File "/home/fanta/.pyenv/versions/python3.9.14/lib/python3.9/site-packages/autogluon/common/loaders/load_pkl.py", line 42, in load
    with compression_fn_map[compression_fn]['open'](validated_path, 'rb', **compression_fn_kwargs) as fin:
FileNotFoundError: [Errno 2] No such file or directory: 'autogluon/models/Naive/info.pkl'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/fanta/workspace/autogluon_example/main.py", line 50, in <module>
    main()
  File "/home/fanta/workspace/autogluon_example/main.py", line 46, in main
    info = predictor.info() # AttributeError: 'str' object has no attribute 'name'
  File "/home/fanta/.pyenv/versions/python3.9.14/lib/python3.9/site-packages/autogluon/timeseries/predictor.py", line 599, in info
    return self._learner.get_info(include_model_info=True)
  File "/home/fanta/.pyenv/versions/python3.9.14/lib/python3.9/site-packages/autogluon/timeseries/learner.py", line 393, in get_info
    trainer_info = trainer.get_info(include_model_info=include_model_info)
  File "/home/fanta/.pyenv/versions/python3.9.14/lib/python3.9/site-packages/autogluon/timeseries/trainer/abstract_trainer.py", line 262, in get_info
    info["model_info"] = self.get_models_info()
  File "/home/fanta/.pyenv/versions/python3.9.14/lib/python3.9/site-packages/autogluon/timeseries/trainer/abstract_trainer.py", line 217, in get_models_info
    model_info_dict[model] = model_type.load_info(path=model_path)
[...]

PR #2510, opened 2022-12-05
[timeseries] Fix info and fit_summary for TimeSeriesPredictor

shchur:
*Issue #, if available:* #2497 and #2498

*Description of changes:*
- Change the implementation of `AbstractTimeSeriesModel.get_info()` and `TimeSeriesPredictor.fit_summary()`.
- Add tests covering these functions.

### DIFF
--- a/timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
+++ b/timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
@@ -735,6 +735,7 @@ class AbstractTimeSeriesTrainer(SimpleAbstractTrainer):
         )
 
         self._add_model(model=simple_ensemble, base_models=simple_ensemble.model_names)
+        self.save_model(model=simple_ensemble)
         return simple_ensemble.name
 
     def leaderboard(self, data: Optional[TimeSeriesDataFrame] = None) -> pd.DataFrame:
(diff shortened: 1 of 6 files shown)


timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -735,6 +735,7 @@ def fit_ensemble(
         )
 
         self._add_model(model=simple_ensemble, base_models=simple_ensemble.model_names)
+        self.save_model(model=simple_ensemble)
```
Innixma, line 738: What was happening before this change?

Innixma (approved):

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
shchur: The `WeightedEnsemble` model is saved [inside the model graph](https://github.com/autogluon/autogluon/blob/master/timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py#L175) (i.e., as an attribute of the trainer saved to `trainer.pkl`). This means that we could predict with the ensemble after loading the predictor from disk. 

The new line `self.save_model(model=simple_ensemble)` saves the metadata of the model (val_score, fit_time, etc.) to disk so that it can be retrieved when calling `info()` and `get_summary()`.

PR #2512, opened 2022-12-05
Fix image properties in df preprocessor

zhiqiangdon:
*Issue #, if available:*
`_image_bytearray_names` attribute doesn't exist in df preprocessor. So, `hasattr(self, "_image_bytearray_names")` is always False.  `hasattr(self, "_image_path_names")` is for backward compatibility since we used `_image_path_names` in the initial AutoMM versions. The image_bytearray is new and doesn't have backward compatible issues.

*Description of changes:*
Removed the `hasattr(self, "_image_bytearray_names")` checks.


suzhoum (approved):
Thanks for fixing it!

bryanyzhu (approved):
LGTM.

PR #2522, opened 2022-12-06
[timeseries] Pass known_covariates to component models of the WeightedEnsemble

shchur:
*Description of changes:*
- Fix a bug where prediction with `WeightedEnsemble` fails if component models use `known_covariates` + add a test case for this
- Make sure that static feature preprocessing doesn't introduce NaNs to the `static_features` dataframe


canerturkmen (approved):
LGTM

PR #2523, opened 2022-12-06
[timeseries] Ensure that timestamps are sorted when creating a TimeSeriesDataFrame

shchur:
*Description of changes:*
- Ensures that the timestamps are always sorted chronologically inside a `TimeSeriesDataFrame`

### DIFF
--- a/timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
+++ b/timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
@@ -102,8 +102,10 @@ class TimeSeriesDataFrame(pd.DataFrame):
                 self._validate_multi_index_data_frame(data)
             else:
                 data = self._construct_pandas_frame_from_data_frame(data)
+            data = self._sort_timestamps(data)
         elif isinstance(data, Iterable):
             data = self._construct_pandas_frame_from_iterable_dataset(data)
+            data = self._sort_timestamps(data)
         else:
             raise ValueError("Data input type not recognized, must be DataFrame or iterable.")
         super().__init__(data=data, *args, **kwargs)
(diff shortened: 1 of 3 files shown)


timeseries/tests/unittests/models/test_sktime.py
```diff
@@ -64,9 +64,6 @@ def test_when_sktime_converts_from_dataframe_then_data_not_duplicated_and_index_
         )
     )
 
-    # data is not copied
-    assert df.values.base is sktime_df.values.base
-
```
shchur, line 69: Creating a `TimeSeriesDataFrame` now results in copying the data because the sort is always performed.

We could try to only do sorting if the timestamps are not already sorted, but I think that it would be more cumbersome & less efficient (since we need to check if the timestamps are sorted for each individual time series).

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
```diff
@@ -102,6 +102,9 @@ def __init__(self, data: Any, static_features: Optional[pd.DataFrame] = None, *a
                 self._validate_multi_index_data_frame(data)
             else:
                 data = self._construct_pandas_frame_from_data_frame(data)
+            # Make sure that timestamps are sorted but item_id order is preserved
+            item_ids = data.index.unique(level=ITEMID)
+            data = data.sort_values(by=TIMESTAMP).loc[item_ids]
```
shchur, line 107: We could add a flag to the `TimeSeriesDataFrame` constructor and set it to `False` when creating the `TimeSeriesDataFrame` internally (e.g., inside `slice_by_timestep`) to avoid redundant sorting every time we slice inside the evaluator (this happens a lot when we're fitting the ensemble).

gradientsky (approved):

PR #2531, opened 2022-12-07
[Tabular] support predicting with model fitted on v0.5.1

liangfu:
*Issue #, if available:*

Models that are fitted on v0.5.1 doesn't have attributes like `_compiler` and `compile_time`.

*Description of changes:*

This PR checks whether the model itself has the compile-related attribute, before accessing its value.


Innixma (approved):
LGTM!

PR #2532, opened 2022-12-07
[Tabular] Avoid copying identical dataframes

liangfu:
*Issue #, if available:*

*Description of changes:*

Avoid copying DataFrame when they are identical. This brings about 10% speedup of single-batch prediction on tabular models.


Innixma (approved):
Looks great! I am seeing a speedup of 9ms -> 7ms on transform_features on my laptop on AdultIncome with this change for single-row batch. I also see similar speedup percentages for larger batches!

liangfu:
Just verified that test_scores are exactly the same across multiple datasets and tasks

image

image

, and we are saving 11.7% of the prediction time on average.

image

gradientsky (approved):

Innixma (approved):
Benchmark results look good!

PR #2536, opened 2022-12-08
[Tabular] Resource Allocation Fix

yinweisu:
*Issue #, if available:*
https://github.com/autogluon/autogluon/issues/2446

*Description of changes:*
* Fixed the case when user didn't specify total resources, but specified lower level resources. Previously would pass default value based on the model to the lower value as the total resources, which is incorrect. Now will check if user passed lower level requirements, and use those.
* Added a special case when user provide both `num_resource` and `ag_args_fit` when not doing bagging nor hpo. Might think about better way to handle it in the future...
* Updated unit tests and added two more tests to cover cases when no resource requirement is specified, we should be using the default value based on the model.
* Added end-to-end tests...

### DIFF
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -496,11 +496,38 @@ class AbstractModel:
         system_num_cpus = ResourceManager.get_cpu_count()
         system_num_gpus = ResourceManager.get_gpu_count_all()
         default_num_cpus, default_num_gpus = self._get_default_resources()
+        # This could be resource requirement for bagged model or individual model
+        user_specified_lower_level_num_cpus = self._user_params_aux.get('num_cpus', None)
+        user_specified_lower_level_num_gpus = self._user_params_aux.get('num_gpus', None)
+        if user_specified_lower_level_num_cpus is not None:
+            assert user_specified_lower_level_num_cpus <= system_num_cpus, f'Specified num_cpus per {self.__class__.__name__} is more than the total: {system_num_cpus}'
+        if user_specified_lower_level_num_gpus is not None:
+            assert user_specified_lower_level_num_gpus <= system_num_cpus, f'Specified num_gpus per {self.__class__.__name__} is more than the total: {system_num_cpus}'
         k_fold = kwargs.get('k_fold', None)
         if k_fold is not None and k_fold > 0:
             # bagged model's default resources should be resources * k_fold if the amount is available
             default_num_cpus = min(default_num_cpus * k_fold, system_num_cpus)
             default_num_gpus = min(default_num_gpus * k_fold, system_num_gpus)
+            # initialize the model base to check if there's user specified resources requirements
+            user_specified_model_level_num_cpus = self.model_base._user_params_aux.get('num_cpus', None)
+            user_specified_model_level_num_gpus = self.model_base._user_params_aux.get('num_gpus', None)
+            if user_specified_model_level_num_cpus is not None:
+                assert user_specified_model_level_num_cpus <= system_num_cpus, f'Specified num_cpus per {self.model_base.__class__.__name__} is more than the total: {system_num_cpus}'
+            if user_specified_model_level_num_gpus is not None:
+                assert user_specified_model_level_num_gpus <= system_num_gpus, f'Specified num_gpus per {self.model_base.__class__.__name__} is more than the total: {system_num_gpus}'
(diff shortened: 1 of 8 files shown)


tabular/tests/unittests/resource_allocation/test_total_resource_allocation.py
```diff
@@ -190,7 +210,7 @@ def test_nonbagged_model_with_total_resources_and_model_resources(mock_system_re
             hyperparameters={
                 'ag_args_fit': {
                     'num_cpus': 1,
-                    'num_gpus': 1
+                    'num_gpus': 0
```
Innixma, line 213: Why is this hard-coded? Wouldn't it be way easier to just have this be in the arguments of the test?

tabular/tests/unittests/resource_allocation/test_total_resource_allocation.py
```diff
@@ -137,6 +144,19 @@ def test_bagged_model_without_total_resources_but_with_ensemble_resources(mock_s
...
+def test_bagged_model_without_total_resources_and_without_model_resources(mock_system_resources_ctx_mgr, mock_num_cpus, mock_num_gpus):
+    with mock_system_resources_ctx_mgr(num_cpus=mock_num_cpus, num_gpus=mock_num_gpus):
+        model_base = DummyModel()
+        bagged_model = DummyBaggedModel(
+            model_base,
+            hyperparameters={}
+        )
+        k_fold = 2
+        resources = bagged_model._preprocess_fit_resources(k_fold=k_fold)
+        resources.pop('k_fold')
```
Innixma, line 159: Why is this hard-coded? Wouldn't it be way easier to just have this be in the arguments of the test?

tabular/tests/unittests/resource_allocation/test_total_resource_allocation.py
```diff
@@ -137,6 +144,19 @@ def test_bagged_model_without_total_resources_but_with_ensemble_resources(mock_s
...
         resources.pop('k_fold')
+        assert resources == {'num_cpus': 1, 'num_gpus': 0}
+        
+        
+def test_bagged_model_without_total_resources_and_without_model_resources(mock_system_resources_ctx_mgr, mock_num_cpus, mock_num_gpus):
+    with mock_system_resources_ctx_mgr(num_cpus=mock_num_cpus, num_gpus=mock_num_gpus):
+        model_base = DummyModel()
+        bagged_model = DummyBaggedModel(
+            model_base,
+            hyperparameters={}
```
Innixma, line 155: Why is this hard-coded? Wouldn't it be way easier to just have this be in the arguments of the test?

This way we can specify ag_args_fit_ensemble for example.

Example for ag_args_fit_ensemble:
```
hyperparameters={
    'ag_args_fit': {'num_cpus': 4, 'num_gpus': 1}
}
```

yinweisu: Here we are intentionally testing when not providing any user specified resources, so the hyperparameter is empty dict. I should just remove hyperparameters

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -496,11 +496,37 @@ def _preprocess_fit_resources(self, silent=False, total_resources=None, **kwargs
...
+                assert user_specified_model_level_num_gpus <= system_num_gpus, f'Specified num_gpus per {self.model_base.__class__.__name__} is more than the total: {system_num_gpus}'
+            if user_specified_lower_level_num_cpus is not None:
+                if user_specified_model_level_num_cpus is not None:
+                    user_specified_lower_level_num_cpus = min(user_specified_lower_level_num_cpus * k_fold, user_specified_model_level_num_cpus)
+            else:
+                if user_specified_model_level_num_cpus is not None:
+                    user_specified_lower_level_num_cpus = min(user_specified_model_level_num_cpus * k_fold, system_num_cpus)
+            if user_specified_lower_level_num_gpus is not None:
+                if user_specified_model_level_num_gpus is not None:
+                    user_specified_lower_level_num_gpus = min(user_specified_lower_level_num_gpus * k_fold, user_specified_model_level_num_gpus)
```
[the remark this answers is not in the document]
Innixma: Please don't mark as resolved without responding with an answer. This is a major bug if it is indeed a bug.

Another question would be: Why didn't the unit tests catch this bug?

gradientsky (approved):

Innixma (approved):

PR #2537, opened 2022-12-08
[CI] Fix build all docs

yinweisu:
*Issue #, if available:*

*Description of changes:*
* The cause is that `d2lbook` depends on `awscli`, which got updated yesterday and have a conflict dependency with `tox` version we depend on in `eda` . We were previously installing all AG modules in the `build_all_docs` step, though I doubt it's needed. Removed for now to see if it solves the issue


gidler (approved):

suzhoum (approved):
LGTM!

gradientsky (approved):

PR #2538, opened 2022-12-08
Fix timezone datetime edgecase

Innixma:
*Issue #, if available:*

*Description of changes:*
- Addresses complicated preprocessing situation where timezones (including multiple timezones and multiple timezone + no-timezone rows) are present in a single datetime column, along with NaN's of multiple formats. Previously this crashed.


features/src/autogluon/features/generators/datetime.py
```diff
@@ -61,16 +65,29 @@ def _generate_features_datetime(self, X: DataFrame) -> DataFrame:
...
+                try:
+                    X_datetime[datetime_feature + '.' + feature] = getattr(X_datetime[datetime_feature].dt, feature).astype(int)
+                except AttributeError:
+                    # Strange datetime object, try removing timezone info
+                    X_datetime[datetime_feature + '.' + feature] = getattr(self._remove_timezones(X_datetime[datetime_feature]).dt, feature).astype(int)
+
+            try:
+                X_datetime[datetime_feature] = pd.to_numeric(X_datetime[datetime_feature])
+            except TypeError:
+                # Strange datetime object, try removing timezone info
```
gradientsky, line 89: Stripping known time zones would change the timestamp for those values. The method below can handle it correctly.

I think we can explicitly handle edge cases via separating good rows vs bad ones; using `utc=True` allows to normalize all time zones (though the cases without tz would be considered UTC):
```python
def normalize_timeseries(series):
    series = pd.to_datetime(series, utc=True)
    broken_idx = series[(series == 'NaT') | series.isna() | series.isnull()].index
    bad_rows = series.iloc[broken_idx]
    good_rows = series[~series.isin(bad_rows)].astype(int)
    series.iloc[good_rows.index] = good_rows
    series[broken_idx] = int(good_rows.mean())
    return series
```

features/tests/features/generators/test_datetime.py
```diff
@@ -92,3 +92,44 @@ def test_datetime_feature_generator(generator_helper, data_helper):
...
+        'datetime_as_object.dayofweek'
+    ]}
+
+    expected_output_data_feat_datetime = [
+        1533140820000000000,
+        1600055337034500096,
+        1600055337034500096,
+        1628528828659000000,
+        1628528895541000000,
+        1610022803938000000
```
gradientsky, line 124: This is not correct because time the known zones were stripped.
```
0    1533140820000000000
1    1600067037034500096
2    1600067037034500096
3    1628543228659000000
4    1628543295541000000
5    1610040803938000000
```

features/src/autogluon/features/generators/datetime.py
Innixma:
```
series = pd.to_datetime(series, utc=True)
```
Crashes.

```

../../../src/autogluon/features/generators/datetime.py:131: in _remove_timezones
    return normalize_timeseries(datetime_as_object)
../../../src/autogluon/features/generators/datetime.py:122: in normalize_timeseries
    series = pd.to_datetime(series, utc=True)
../../../../../../virtual/autogluon38/lib/python3.8/site-packages/pandas/core/tools/datetimes.py:1068: in to_datetime
    values = convert_listlike(arg._values, format)
../../../../../../virtual/autogluon38/lib/python3.8/site-packages/pandas/core/tools/datetimes.py:438: in _convert_listlike_datetimes
    result, tz_parsed = objects_to_datetime64ns(
../../../../../../virtual/autogluon38/lib/python3.8/site-packages/pandas/core/arrays/datetimes.py:2177: in objects_to_datetime64ns
    result, tz_parsed = tslib.array_to_datetime(
pandas/_libs/tslib.pyx:427: in pandas._libs.tslib.array_to_datetime
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   ValueError: mixed datetimes and integers in passed array
```

gradientsky (approved):

PR #2541, opened 2022-12-08
[Merge After 0.6.1] Adding support for str/list input for classification

yongxinw:
*Issue #, if available:*

*Description of changes:* Adding support for str/list input for classification. Unit Test passed.

### DIFF
--- a/multimodal/src/autogluon/multimodal/utils/data.py
+++ b/multimodal/src/autogluon/multimodal/utils/data.py
@@ -1,9 +1,11 @@
 import logging
 import os
+import random
 import warnings
 from typing import Dict, List, Optional, Tuple, Union
 
 import pandas as pd
+import PIL
 from omegaconf import DictConfig, OmegaConf
 from sklearn.preprocessing import LabelEncoder
 from torch import nn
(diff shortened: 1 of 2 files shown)


FANGAreNotGnu (approved):
LGTM with a minor question. If a list contains invalid image_path, will it cause error in processor?

bryanyzhu (approved):
LGTM, let's merge it after 0.6.1.

multimodal/src/autogluon/multimodal/utils/data.py
```diff
@@ -427,12 +429,21 @@ def data_to_df(
     elif isinstance(data, dict):
         data = pd.DataFrame(data)
     elif isinstance(data, list):
-        if header is None:
-            data = pd.DataFrame(data)
+        assert len(data) > 0, f"Expected data to have length > 0, but got {data} of len {len(data)}"
+        if contains_valid_images(data):
+            data_dict = {"image": data}
```
zhiqiangdon, line 434: Is it necessary to check if a list of image paths and use header `image`? Whether image list or not doesn't affect creating the dataframe. pandas creates a dataframe with header `0`. I don't think header `image` is required in later data processing and model forward.

multimodal/src/autogluon/multimodal/utils/data.py
```diff
@@ -515,3 +526,38 @@ def infer_dtypes_by_model_names(model_config: DictConfig):
         fallback_dtype = list(allowable_dtypes)[0]
 
     return allowable_dtypes, fallback_dtype
+
+
+def contains_valid_images(data: Union[str, list], sample_n: Optional[int] = 50) -> bool:
```
zhiqiangdon, line 531: How about reusing `is_imagepath_column()` in `infer_types.py`?

multimodal/src/autogluon/multimodal/utils/data.py
```diff
@@ -427,12 +429,21 @@ def data_to_df(
...
         else:
-            data = pd.DataFrame({header: data})
+            if header is None:
+                data = pd.DataFrame(data)
+            else:
+                data = pd.DataFrame({header: data})
     elif isinstance(data, str):
-        data = load_pd.load(data)
+        if contains_valid_images(data):
+            data_dict = {"image": [data]}
```
zhiqiangdon, line 443: Similarly, header `image` seems unnecessary.

multimodal/src/autogluon/multimodal/utils/data.py
yongxinw: let me double check. I think I saw at some point the code actually uses the header `image`

multimodal/src/autogluon/multimodal/utils/data.py
yongxinw: the is_imagepath_column checks on a df.Series. This test was for array of images. That being said, let me check if the `image` header is needed. If not, I can convert the image array to pd.Series and re-use `is_imagepath_column()`.

Issue #2521, opened 2022-12-06
AutoGluon fails installation on Super Computer [BUG]


asmgx:
I am trying to install AutoGluon on my university computer (Linux)

I followed the instruction given here https://auto.gluon.ai/stable/index.html

```
pip3 install -U pip
pip3 install -U setuptools wheel

# CPU version of pytorch has smaller footprint - see installation instructions in
# pytorch documentation - https://pytorch.org/get-started/locally/
pip3 install torch==1.12+cpu torchvision==0.13.0+cpu torchtext==0.13.0 -f https://download.pytorch.org/whl/cpu/torch_stable.html

pip3 install autogluon
```

all steps ended successfully except for the last one **pip3 install autogluon**

I am getting this error

```

user99@tinaroo1:~/autogluon> pip3 install autogluon
Keyring is skipped due to an exception: 'EntryPoints' object has no attribute 'get'
Defaulting to user installation because normal site-packages is not writeable
Collecting autogluon
  Using cached autogluon-0.6.0-py3-none-any.whl (9.8 kB)
Collecting autogluon.multimodal==0.6.0
  Using cached autogluon.multimodal-0.6.0-py3-none-any.whl (279 kB)
Collecting autogluon.timeseries[all]==0.6.0
  Using cached autogluon.timeseries-0.6.0-py3-none-any.whl (101 kB)
Collecting autogluon.features==0.6.0
  Using cached autogluon.features-0.6.0-py3-none-any.whl (59 kB)
Collecting autogluon.core[all]==0.6.0
  Using cached autogluon.core-0.6.0-py3-none-any.whl (224 kB)
Collecting autogluon.vision==0.6.0
  Using cached autogluon.vision-0.6.0-py3-none-any.whl (49 kB)
Collecting autogluon.tabular[all]==0.6.0
  Using cached autogluon.tabular-0.6.0-py3-none-any.whl (285 kB)
Collecting autogluon.text==0.6.0
  Using cached autogluon.text-0.6.0-py3-none-any.whl (62 kB)
Requirement already satisfied: numpy<1.24,>=1.21 in /home/user99/.local/lib/python3.7/site-packages (from autogluon.core[all]==0.6.0->autogluon) (1.21.5)
Requirement already satisfied: scipy<1.10.0,>=1.5.4 in /home/user99/.local/lib/python3.7/site-packages (from autogluon.core[all]==0.6.0->autogluon) (1.7.1)
Collecting autogluon.common==0.6.0
[...]

PR #2548, opened 2022-12-09
Remove pycocotools

bryanyzhu:
*Issue #, if available:*
AutoGluon installation issue. #2494 #2521 
pycocotools can't be easily built in some linux and windows distribution.

*Description of changes:*
Remove pycocotools from setup.py. Given this is only used for object detection model evaluation, it is better to lazy import rather than effecting AG installation. 


FANGAreNotGnu:
We have lazy install function, but need one additional line, see: https://github.com/autogluon/autogluon/pull/2549

sxjscience (approved):
LGTM

FANGAreNotGnu (approved):
LGTM, thanks!

PR #2558, opened 2022-12-12
Fix int32/int64 difference between windows and other platforms; fix inputs mutation issue

gradientsky:
*Description of changes:*
Fix int32/int64 difference between windows and other platforms; fix inputs mutation issue


yinweisu (approved):
LGTM. Object detection failed with some weird error. Likely random...

PR #2564, opened 2022-12-13
Export TIMM image models to onnx

liangfu:
*Issue #, if available:*

*Description of changes:*
1. This PR exports the TIMM image models to onnx, and it's been tested on onnxruntime for checking correctness.
2. Downgrades onnx to 1.12.x (since https://github.com/autogluon/autogluon/pull/2567), due to conflict version range of protobuf against tensorboardx:
  a. onnx 1.13.x requires protobuf >=3.12.2,<=3.20.1, (onnx 1.12.x requires protobuf >=3.12.2,<=3.20.1)
  b. tensorboardx 2.5.1 requires protobuf >=3.8.0,<=3.20.1

### DIFF
--- a/multimodal/src/autogluon/multimodal/models/timm_image.py
+++ b/multimodal/src/autogluon/multimodal/models/timm_image.py
@@ -1,5 +1,5 @@
 import logging
-from typing import Optional
+from typing import List, Optional
 
 import torch
 from timm import create_model
@@ -78,6 +78,10 @@ class TimmAutoModelForImagePrediction(nn.Module):
         return f"{self.prefix}_{LABEL}"
 
     @property
+    def input_keys(self):
+        return [self.image_key, self.image_valid_num_key]
+
+    @property
     def image_column_prefix(self):
         return f"{self.image_key}_{COLUMN}"
 
@@ -87,21 +91,35 @@ class TimmAutoModelForImagePrediction(nn.Module):
 
     def forward(
         self,
-        batch: dict,
+        images: torch.FloatTensor,
+        image_valid_num: torch.Tensor,
+        image_column_names: Optional[List[str]] = None,
+        image_column_indices: Optional[List[torch.Tensor]] = None,
     ):
         """
         Parameters
         ----------
-        batch
-            A dictionary containing the input mini-batch data.
-            We need to use the keys with the model prefix to index required data.
+        images
+            A tensor in [N, C, H, W] layout to represent the images.
+        image_valid_num
+            A tensor that describes valid number of input images.
+        image_column_names
+            A list of strings that indicates names of the image columns.
+        image_column_indices
+            A list of tensors that indicates start and stop indices of the image columns.
 
         Returns
         -------
             A dictionary with logits and features.
         """
-        images = batch[self.image_key]
-        image_valid_num = batch[self.image_valid_num_key]
+        batch = {
+            self.image_key: images,
+            self.image_valid_num_key: image_valid_num,
+        }
+        if image_column_names:
+            assert len(image_column_names) == len(image_column_indices), "invalid image column inputs"
+            batch.update(**dict(zip(image_column_names, image_column_indices)))
+
         ret = {COLUMN_FEATURES: {FEATURES: {}, MASKS: {}}}
         if self.mix_choice == "all_images":  # mix inputs
             mixed_images = (
--- a/multimodal/src/autogluon/multimodal/models/utils.py
+++ b/multimodal/src/autogluon/multimodal/models/utils.py
@@ -684,3 +684,21 @@ def update_mmdet_config(key, value, config):
             for subsubconfig in subconfig:
                 if isinstance(subsubconfig, dict):
                     update_mmdet_config(key, value, subsubconfig)
+
+
+def run_model(model: nn.Module, batch: dict):
+    from .timm_image import TimmAutoModelForImagePrediction
+
+    if isinstance(model, TimmAutoModelForImagePrediction):
+        input_vec = [batch[k] for k in model.input_keys]
+        column_names, column_values = [], []
+        for k in batch.keys():
+            if k.startswith(model.image_column_prefix):
+                column_names.append(k)
+                column_values.append(batch[k])
+        input_vec.append(column_names)
+        input_vec.append(column_values)
+        output = model(*tuple(input_vec))
+    else:
+        output = model(batch)
+    return output
--- a/multimodal/src/autogluon/multimodal/predictor.py
+++ b/multimodal/src/autogluon/multimodal/predictor.py
@@ -12,7 +12,7 @@ import shutil
 import sys
 import time
 import warnings
-from collections import OrderedDict
+from collections import OrderedDict, namedtuple
 from dataclasses import dataclass
 from datetime import timedelta
 from typing import Dict, List, Optional, Union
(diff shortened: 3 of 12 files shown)


liangfu:
cc @FANGAreNotGnu @zhiqiangdon @tonyhoo

multimodal/src/autogluon/multimodal/models/utils.py
```diff
@@ -684,3 +684,19 @@ def update_mmdet_config(key, value, config):
             for subsubconfig in subconfig:
                 if isinstance(subsubconfig, dict):
                     update_mmdet_config(key, value, subsubconfig)
+
+
+def run_model(model: nn.Module, batch: dict):
+    if type(model).__module__.endswith("timm_image"):
```
tonyhoo, line 690: Wondering if all backbones in TIMM will be ONNX compatible or just the ones we use as of now

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -3147,3 +3155,28 @@ def __init__(self, **kwargs):
...
+
+
+class MultiModalOnnxPredictor(MultiModalPredictor):
+    def __init__(self, model):
+        import onnxruntime as ort
+
+        self.sess = ort.InferenceSession(model.SerializeToString(), providers=["CPUExecutionProvider"])
+
+    def predict(self, data: Union[pd.DataFrame, dict, list, str]):
+        raise NotImplementedError()
```
tonyhoo, line 3167: Wondering why we left this unimplemented?

liangfu: This depends on problem_type / postprocessing, which isn't available within the converted onnx model.

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -2771,10 +2771,19 @@ def export_onnx(
...
+        self._model.to(device).eval()
+        batch = move_to_device(batch, device=device)
+
+        if type(self._model).__module__.endswith("timm_image"):
+            input_vec = tuple([batch[k] for k in self._model.input_keys])
+        else:
+            input_vec = batch
         torch.onnx.export(
             self._model.eval(),
-            batch,
```
zhiqiangdon, line 2776: Can we still use the dict format as model input? It seems that @FANGAreNotGnu made it work when export hf_text models.

liangfu: The reason we are planning to remove away from dict input comes with two folds: 1) there are [limitations](https://pytorch.org/docs/stable/onnx.html#limitations) in exporting onnx, 2) there are lots of benefits if we keep model inputs / outputs clear, see https://github.com/huggingface/transformers/issues/16059

zhiqiangdon: The dict input design is to flexibly support various modalities, backbones, and fusions in the AutoMM. I think changing the design may cause issues for some cases. For example, it seems difficult to list arguments of a fusion model since it may fuse any combinations of image, text, and tabular backbones.

multimodal/src/autogluon/multimodal/utils/inference.py
```diff
@@ -146,6 +147,22 @@ def infer_batch(
     return output[model.prefix]
 
 
+def run_model(model: nn.Module, batch: dict):
```
zhiqiangdon, line 150: Duplicate function? I seem this function also exists in `models/utils.py`.

multimodal/src/autogluon/multimodal/predictor.py
zhiqiangdon: For most functions in AutoMM, we have used the type hints. The dict design can make each model more modular. A batch may contain inputs for several backbones and a model only needs to extract necessary information from it. This is especially useful in various fusion settings.

multimodal/src/autogluon/multimodal/models/utils.py
zhiqiangdon, line 690: Users may customize the model instance name to `timm_image_123` or `timm_image_abc`.

liangfu: This captures name of the module, which is indicated by `timm_image.py` file. As long as we don't change the name of the py file, we are completely compatible with current design. This has been tested with multiple test cases with `timm_image_abc` as part of batch keys.

multimodal/src/autogluon/multimodal/utils/inference.py
liangfu: Nice catch! Yes, it is the one in `utils.py` that works. Import the funtion here would cause circular dependencies in fusion models.

liangfu: removed the duplicated function.

multimodal/src/autogluon/multimodal/models/timm_image.py
```diff
@@ -87,21 +91,35 @@ def image_feature_dim(self):
 
     def forward(
         self,
-        batch: dict,
+        images: torch.FloatTensor,
+        image_valid_num: torch.Tensor,
+        image_column_names: Optional[List[str]] = None,
+        image_column_values: Optional[List[torch.Tensor]] = None,
```
zhiqiangdon, line 97: How about changing the name `image_column_values `  to `image_column_indexes`? It stores the start and end indexes of each column's images in `images`.

multimodal/src/autogluon/multimodal/models/utils.py
zhiqiangdon: What do you think if using `if isinstance(model, TimmAutoModelForImagePrediction)`?

multimodal/src/autogluon/multimodal/models/timm_image.py
liangfu: Renamed to `image_column_indices`.

multimodal/src/autogluon/multimodal/models/utils.py
liangfu: Changed to `if isinstance(model, TimmAutoModelForImagePrediction)`

multimodal/src/autogluon/multimodal/predictor.py
sxjscience: If the dict does not work, does it make sense for us to switch to dataclasses or namedtuple @liangfu @zhiqiangdon ?

sxjscience: I think using namedtuple is feasible. Similar to https://github.com/huggingface/transformers/blob/45da7cec5aa1b1bf1031af9caa9085e95e262e11/examples/research_projects/bertabs/run_summarization.py#L28

zhiqiangdon: We can try namedtuple, but not sure whether it is compatible with onnx.

zhiqiangdon (approved):
LGTM. Thanks for fixing onnx export! Remember resolving the conflicts with the master branch.

multimodal/src/autogluon/multimodal/predictor.py
liangfu: namedtuple is now experimentally working as inputs for tracing, but not yet ready to be part of `collate` function. will try integrate it into `collate` function in a follow-up PR.

theolundqvist:
Hi, how would I go ahead and use this feature. The output from the onnx model is not the same as when running it through autogluon. In my case (binary classification using timm) the onnx model outputs one 768 dimensional vector and one 2 dimensional vector for each image in the batch. The 2d vector is always expit(logits) = 0.5+-0.03, while the same images through autogluon receive a predicted score of 0.13-0.95, something is wrong with the compiled model.

Please add documentation or tests showing that it works.

Should I open a new issue?

PR #2565, opened 2022-12-13
Fix COCO dataset conversion CLI tool for AutoMM detection

FANGAreNotGnu:
- Remove image without bounding boxes from annotation file to speed up mAP calculation.
- Fix a bug in image path extraction

### DIFF
--- a/multimodal/src/autogluon/multimodal/cli/voc2coco.py
+++ b/multimodal/src/autogluon/multimodal/cli/voc2coco.py
@@ -29,7 +29,7 @@ from tqdm import tqdm
 from autogluon.multimodal.utils.object_detection import dump_voc_classes, dump_voc_xml_files, process_voc_annotations
 
 DEFAULT_EXT = ".jpg"
-MIN_AREA = 4  # TODO: put in arg?
+MIN_AREA = 4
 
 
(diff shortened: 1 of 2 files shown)


multimodal/src/autogluon/multimodal/cli/voc2coco.py
```diff
@@ -98,8 +98,8 @@ def get_image_info(annotation_root, extract_num_from_imgid=True):
     else:
         filename = os.path.basename(path)
     img_name = os.path.basename(filename)
-    if not filename[-4:] in [".jpg", ".png"]:
-        filename = filename + ".jpg"
+    if not img_name[-4:] in [".jpg", ".png"]:
```
bryanyzhu, line 101: Just a quick question, how do we know the postfix is definitely .jpg?

FANGAreNotGnu: Added default extension to arguments. This line was a hardcode fix for a dataset's bad annotation file (no extensions for image file names in json annotation file).

bryanyzhu (approved):
LGTM

yongxinw (approved):

PR #2569, opened 2022-12-14
Remove pipeline from matcher initialization API

zhiqiangdon:
*Issue #, if available:*
Need to align some details between matcher and predictor before fully merging them.

*Description of changes:*

1. Remove `pipeline` from the matcher initialization API.
2. Remove `pipeline` from `infer_metrics()`. We still need to know whether it's a matching task. But no longer require the data modalities in matching.
3. For multiclass and regression problem types, we will use spearman correlation as suggested in papers https://arxiv.org/pdf/1908.10084.pdf, https://arxiv.org/pdf/2010.08240.pdf, and https://arxiv.org/pdf/2004.09813.pdf. This PR adds a metric placeholder for such problem types. Will have following PRs to support the metric in matching.


sxjscience:
Is it a temporary solution? We should later unify the design of `problem_type`.

zhiqiangdon:
> Is it a temporary solution? We should later unify the design of `problem_type`.

According to our general design discussions, predictor may need the `presets` argument to better support various presets. This can also avoid the conflicts between some new problem types and traditional problem types.

zhiqiangdon:
> Is it a temporary solution? We should later unify the design of `problem_type`.

PR updated. Now matcher uses only `problem_type` to specify `image_similarity`, `text_similarity`, and `image_text_similarity`.

bryanyzhu (approved):
LGTM

cheungdaven (approved):
lgtm

multimodal/src/autogluon/multimodal/matcher.py
```diff
@@ -434,7 +438,7 @@ def fit(
         if self._validation_metric_name is None or self._eval_metric_name is None:
             validation_metric_name, eval_metric_name = infer_metrics(
                 problem_type=problem_type,
-                pipeline=self._pipeline,
+                is_matching=is_matching(self._pipeline),
```
sxjscience, line 441: We can consider to use a function name other than `is_matching`. Because `is_matching` is used as an argument.

multimodal/src/autogluon/multimodal/optimization/utils.py
```diff
@@ -278,7 +282,10 @@ def get_metric(
     elif metric_name == PEARSONR:
         return torchmetrics.PearsonCorrCoef(), None
     elif metric_name == SPEARMANR:
-        return torchmetrics.SpearmanCorrCoef(), None
+        if is_matching:  # TODO: add support for matching.
+            raise ValueError("spearman relation is supported for matching yet.")
```
sxjscience, line 286: Typo: is not supported

multimodal/src/autogluon/multimodal/utils/matcher.py
```diff
@@ -501,3 +502,7 @@ def convert_data_for_ranking(
     response_data = pd.DataFrame({response_column: data[response_column].unique().tolist()})
 
     return data_with_label, query_data, response_data, label_column
+
+
+def is_matching(pipeline: str):
```
sxjscience, line 507: Shall we use https://github.com/autogluon/autogluon/blob/b7a144aefcd90b4ac6f110fb9d20f73c6de1af26/multimodal/src/autogluon/multimodal/predictor.py#L176 ?

```python
problem_property_dict[pipeline].is_matching
```

zhiqiangdon: There is circular import issue if importing `problem_property_dict ` from `predictor.py`. We may refactor this later.

multimodal/src/autogluon/multimodal/matcher.py
zhiqiangdon: Changed.

multimodal/src/autogluon/multimodal/optimization/utils.py
zhiqiangdon: Thanks. Corrected.

multimodal/src/autogluon/multimodal/matcher.py
```diff
@@ -192,8 +191,8 @@ def __init__(
         self._data_format = PAIR  # TODO: Support Triplet
         self._match_label = match_label
         self._label_column = label
-        self._problem_type = problem_type.lower() if problem_type is not None else None
-        self._pipeline = pipeline.lower() if pipeline is not None else None
+        self._problem_type = None  # always infer problem type for matching.
+        self._pipeline = problem_type.lower() if problem_type is not None else None
```
sxjscience, line 195: In the next PR, we can remove `pipeline`.

sxjscience:
We can refactor the implementation further after https://github.com/autogluon/autogluon/pull/2578

PR #2573, opened 2022-12-14
[Bug] Fix missing Predictor API docs on Website

gidler:
*Issue #, if available:*
https://github.com/autogluon/autogluon/issues/2571

*Description of changes:*

The `install_all` bash call in `build_all_docs` was removed in [this commit](https://github.com/autogluon/autogluon/commit/dfa96254773017b1bd51038f62a9b21c3b20dafb) because of a package dependency conflict between `tox` for testing in `eda` module and dependencies of the `d2lbook` tool.

Unfortunately `install_all` call was required to install modules that are used during autogen of docs, resulting in the website API documentation problem described in [the issue](https://github.com/autogluon/autogluon/issues/2571).

The fix here is to create a new `install_all_no_tests` call that avoids installing tests including the `tox` dependency when `eda/[tests]` is installed. In order to do this, I cleaned up the bash scripts we use to install packages so that it is less pedantic and less code is repeated.

### DIFF
--- a/.github/workflow_scripts/env_setup.sh
+++ b/.github/workflow_scripts/env_setup.sh
@@ -24,87 +24,44 @@ function setup_torch_gpu {
     python3 -m pip install torch==1.12.0+cu113 torchvision==0.13.0+cu113 --extra-index-url https://download.pytorch.org/whl/cu113
 }
 
-function install_common {
-    python3 -m pip install --upgrade -e common/[tests]
-}
-
-function install_core {
-    install_common
-    python3 -m pip install --upgrade -e core/
-}
-
-function install_core_all {
-    install_common
-    python3 -m pip install --upgrade -e core/[all]
-}
-
-function install_core_all_tests {
-    install_common
-    python3 -m pip install --upgrade -e core/[all,tests]
-}
-
-function install_features {
-    python3 -m pip install --upgrade -e features/
-}
-
-function install_eda {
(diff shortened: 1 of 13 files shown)


gidler:
Once this PR passes CI and is approved, we need to also merge the same changes into `stable`

gidler:
Verified the Predictor API docs are back on CI built version: http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-2573/a78915e/api/autogluon.predictor.html

.github/workflow_scripts/env_setup.sh
```diff
@@ -24,87 +24,44 @@ function setup_torch_gpu {
...
+        python3 -m pip install --upgrade -e $1
+        shift
+    done
 }
 
 function install_multimodal {
     # launch different process for each test to make sure memory is released
     python3 -m pip install --upgrade pytest-xdist
-    python3 -m pip install --upgrade -e multimodal/[tests]
+    install_local_packages "multimodal/$1"
```
Innixma, line 37: outside of the scope of this PR, but its a bit concerning we are doing special pip installs afterwards in our tests for multimodal... This isn't replicating the user experience. Especially using uncapped versions in the install.

Innixma (approved):
LGTM!

yinweisu (approved):
LGTM! Thanks for the fix and cleaning up the code!

PR #2574, opened 2022-12-14
[Cloud] Unify input of image modality in fit and inference apis

yinweisu:
*Issue #, if available:*

*Description of changes:*
* As discussed with @tonyhoo, we decide to unify the api for fit and inference when image modality is involved. Now we only ask for `image_column` for both cases, and we require user to have the `image_path` inside the csv file to be an absolute path in both cases.
* We've removed support for files in s3 as it would be extremely complicated when image modality is involved. This also makes sense as tabular/multimodal predictor doesn't support images living in s3 as well.
* Updated unit tests accordingly

### DIFF
--- a/cloud/src/autogluon/cloud/predictor/cloud_predictor.py
+++ b/cloud/src/autogluon/cloud/predictor/cloud_predictor.py
@@ -47,7 +47,6 @@ from ..utils.sagemaker_utils import (
 )
 from ..utils.utils import (
     convert_image_path_to_encoded_bytes_in_dataframe,
-    is_compressed_file,
     is_image_file,
     rename_file_with_uuid,
     unzip_file,
--- a/cloud/src/autogluon/cloud/predictor/cloud_predictor.py
+++ b/cloud/src/autogluon/cloud/predictor/cloud_predictor.py
@@ -337,8 +337,6 @@ class CloudPredictor(ABC):
                 common_tune_data_path = os.path.commonpath(tune_data[image_column].tolist())
             common_path = os.path.commonpath([common_train_data_path, common_tune_data_path])
             common_path_head = os.path.split(common_path)[0]  # we keep the base dir to match zipping behavior
-            print(common_path)
-            print(common_path_head)
             train_data[image_column] = [os.path.relpath(path, common_path_head) for path in train_data[image_column]]
             if tune_data is not None:
                 tune_data[image_column] = [os.path.relpath(path, common_path_head) for path in tune_data[image_column]]
--- a/cloud/src/autogluon/cloud/predictor/cloud_predictor.py
+++ b/cloud/src/autogluon/cloud/predictor/cloud_predictor.py
@@ -367,7 +367,9 @@ class CloudPredictor(ABC):
             path=serving_script, bucket=cloud_bucket, key_prefix=util_key_prefix
         )
 
-        images_input = self._upload_fit_image_artifact(images=common_path, bucket=cloud_bucket, key_prefix=util_key_prefix)
+        images_input = self._upload_fit_image_artifact(
+            images=common_path, bucket=cloud_bucket, key_prefix=util_key_prefix
+        )
         inputs = dict(train=train_input, config=config_input, serving=serving_input)
         if tune_input is not None:
             inputs["tune"] = tune_input
@@ -795,9 +797,7 @@ class CloudPredictor(ABC):
         if isinstance(test_data, pd.DataFrame):
             test_data = self._prepare_data(test_data, "test", output_type="csv")
         logger.log(20, "Uploading data...")
-        test_input = self.sagemaker_session.upload_data(
-            path=test_data, bucket=bucket, key_prefix=key_prefix + "/data"
-        )
+        test_input = self.sagemaker_session.upload_data(path=test_data, bucket=bucket, key_prefix=key_prefix + "/data")
         logger.log(20, "Data uploaded successfully")
 
         return test_input
@@ -888,8 +888,7 @@ class CloudPredictor(ABC):
             if isinstance(test_data, str):
                 test_data = load_pd.load(test_data)
             test_data = convert_image_path_to_encoded_bytes_in_dataframe(
-                dataframe=test_data,
-                image_column=test_data_image_column
+                dataframe=test_data, image_column=test_data_image_column
             )
         test_input = self._upload_batch_predict_data(test_data, cloud_bucket, cloud_key_prefix)
 
--- a/cloud/src/autogluon/cloud/predictor/cloud_predictor.py
+++ b/cloud/src/autogluon/cloud/predictor/cloud_predictor.py
@@ -291,22 +291,30 @@ class CloudPredictor(ABC):
         converter = FormatConverterFactory.get_converter(output_type)
         return converter.convert(data, path, filename)
 
-    def _upload_fit_image_artifact(self, images, bucket, key_prefix):
-        if images is not None:
-            image_zip_filename = images
-            assert os.path.isdir(images), "Please provide a folder containing the images"
-            image_zip_filename = os.path.basename(os.path.normpath(images))
+    def _upload_fit_image_artifact(self, image_dir_path, bucket, key_prefix):
+        upload_image_path = None
+        if image_dir_path is not None:
+            image_zip_filename = image_dir_path
+            assert os.path.isdir(image_dir_path), "Please provide a folder containing the images"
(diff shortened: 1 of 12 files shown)


cloud/src/autogluon/cloud/predictor/cloud_predictor.py
```diff
@@ -322,36 +314,59 @@ def _upload_fit_artifact(
...
+            if isinstance(train_data, str):
+                train_data = load_pd.load(train_data)
+            else:
+                train_data = copy.deepcopy(train_data)
+            if tune_data is not None:
+                if isinstance(tune_data, str):
+                    tune_data = load_pd.load(tune_data)
+                else:
+                    tune_data = copy.deepcopy(tune_data)
+            common_train_data_path = os.path.commonpath(train_data[image_column].tolist())
```
tonyhoo, line 334: We should check if common_path is the same root path such as `/`. Otherwise, we probably will zip the whole volume and trigger `out of disk space` issue

cloud/src/autogluon/cloud/predictor/cloud_predictor.py
```diff
@@ -777,26 +781,21 @@ def predict_real_time(self, test_data, accept="application/x-parquet"):
...
-                        "Are you sure you want to do batch inference on a single image? You might want to try `deploy()` and `predict_realtime()` instead"
-                    )
-                else:
-                    test_data = load_pd.load(test_data)
+        # If a directory of images, upload directly
+        if isinstance(test_data, str) and not os.path.isdir(test_data):
+            # either a file to a dataframe, or a file to an image
+            if is_image_file(test_data):
+                logger.warning(
+                    "Are you sure you want to do batch inference on a single image? You might want to try `deploy()` and `predict_realtime()` instead"
```
tonyhoo, line 789: [minor] `predict_real_time()`

cloud/src/autogluon/cloud/predictor/cloud_predictor.py
yinweisu: synced offline. Mark as resolved

tonyhoo (approved):

cloud/src/autogluon/cloud/predictor/cloud_predictor.py
```diff
@@ -292,71 +291,99 @@ def _prepare_data(self, data, filename, output_type="csv"):
...
+                bucket=bucket,
+                key_prefix=key_prefix,
+            )
+            logger.log(20, "Images uploaded successfully")
+        return upload_image_path
+
+    def _find_common_path_and_replace_image_column(self, data, image_column):
+        common_path = os.path.commonpath(data[image_column].tolist())
+        common_path_head = os.path.split(common_path)[0]  # we keep the base dir to match zipping behavior
+        data[image_column] = [os.path.relpath(path, common_path_head) for path in data[image_column]]
```
gidler, line 315: `data[image_column] = data[image_column].apply(lambda path: os.path.relpath(path, common_path_head))`

TIOLI: This is probably more efficient because it avoids going to a python list and back to pandas Series

cloud/src/autogluon/cloud/predictor/cloud_predictor.py
```diff
@@ -292,71 +291,99 @@ def _prepare_data(self, data, filename, output_type="csv"):
...
+                    tune_data = load_pd.load(tune_data)
+                else:
+                    tune_data = copy.deepcopy(tune_data)
+            train_data, common_train_data_path = self._find_common_path_and_replace_image_column(
+                data=train_data, image_column=image_column
             )
-            logger.log(20, "Train data uploaded successfully")
+            if tune_data is not None:
+                tune_data, common_tune_data_path = self._find_common_path_and_replace_image_column(
+                    data=train_data, image_column=image_column
```
gidler, line 348: *Bug*: Shouldn't this be `data=tune_data`?

cloud/src/autogluon/cloud/predictor/multimodal_cloud_predictor.py
```diff
@@ -112,31 +107,21 @@ def predict(
...
             The test data to be inferenced.
-            Can be a pandas.DataFrame, a local path or a s3 path to a csv file.
+            Can be a pandas.DataFrame, a local path to a csv file.
             When predicting multimodality with image modality:
                 You need to specify `test_data_image_column`, and make sure the image column contains relative path to the image.
             When predicting with only images:
-                Can be a local path or a s3 path to a directory containing the images.
-                or a local path or a s3 path to a single image.
+                Can be a local path to a directory containing the images.
+                or a local path to a single image.
```
gidler, line 115: Can be a local path to a directory containing the images or a local path to a single image.

cloud/src/autogluon/cloud/scripts/train.py
```diff
@@ -104,9 +109,19 @@ def get_env_if_present(name):
...
+        shutil.unpack_archive(train_image_compressed_file, train_images_dir)
+        image_column = config["image_column"]
+        training_data[image_column] = [os.path.join(train_images_dir, path) for path in training_data[image_column]]
+
+    if args.tune_images:
+        tune_image_compressed_file = get_input_path(args.tune_images)
+        tune_images_dir = "tune_images"
+        shutil.unpack_archive(tune_image_compressed_file, tune_images_dir)
+        image_column = config["image_column"]
+        tuning_data[image_column] = [os.path.join(tune_images_dir, path) for path in tune_images_dir[image_column]]
```
gidler, line 124: Similar comment to above

cloud/src/autogluon/cloud/scripts/train.py
```diff
@@ -104,9 +109,19 @@ def get_env_if_present(name):
...
 
-    if args.images_dir:
-        image_compressed_file = get_input_path(args.images_dir)
-        shutil.unpack_archive(image_compressed_file)
+    if args.train_images:
+        train_image_compressed_file = get_input_path(args.train_images)
+        train_images_dir = "train_images"
+        shutil.unpack_archive(train_image_compressed_file, train_images_dir)
+        image_column = config["image_column"]
+        training_data[image_column] = [os.path.join(train_images_dir, path) for path in training_data[image_column]]
```
gidler, line 117: TIOLI: this is probably more efficient since it avoids going to Python list and back:

`training_data[image_column] = training_data[image_column].apply(lambda path: os.path.join(train_images_dir, path))`

gidler (changes requested):
Sorry for all the comments, there's a lot of good changes here so there's lots to comment on. The one comment with *Bug* is high importance, everything else is your discretion.

One random callout (not just here but all over our codebase), we should almost always use `{*entries*}` instead of `dict(*entries*)`. My understanding is its almost always more efficient, and I think it looks more Pythonic.

cloud/src/autogluon/cloud/predictor/cloud_predictor.py
yinweisu: Yes, this is indeed incorrect. Thanks for catching that!

cloud/src/autogluon/cloud/utils/utils.py
```diff
@@ -12,17 +13,40 @@
...
+        image_root_path = /home/user/example_images/
+        path = example_images/test/a.jpg
+        would return /home/user/example_images/test/a.jpg
+    This is needed to make sure the fit and predict api of CloudPredictor follows the same logic.
+    """
+    image_root_path = os.path.abspath(image_root_path)
+    path = path.split(os.path.sep)
+    while path[0] == "":
+        # Avoid cases like /foo/test
+        path.pop(0)
```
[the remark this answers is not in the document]
yinweisu: This function actually should be removed. It's a leftover function when we try another method.

cloud/src/autogluon/cloud/scripts/train.py
```diff
@@ -42,7 +42,12 @@ def get_env_if_present(name):
     parser.add_argument("--n_gpus", type=str, default=get_env_if_present("SM_NUM_GPUS"))
     parser.add_argument("--training_dir", type=str, default=get_env_if_present("SM_CHANNEL_TRAIN"))
```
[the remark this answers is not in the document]
yinweisu: good call. Shouldn't break backward compatibility

cloud/src/autogluon/cloud/scripts/train.py
yinweisu: `train_images_dir` should be `tune_images_dir` ;). Easy to make mistakes when copy paste haha

gidler (approved):
LGTM

PR #2576, opened 2022-12-15
[WIP] Update with latest torchmetrics mAP

FANGAreNotGnu:

### DIFF
--- a/multimodal/setup.py
+++ b/multimodal/setup.py
@@ -41,7 +41,7 @@ install_requires = [
     "smart_open>=5.2.1,<5.3.0",
     "pytorch_lightning>=1.7.4,<1.8.0",
     "text-unidecode<=1.3",
-    "torchmetrics>=0.8.0,<0.9.0",
+    "torchmetrics>=0.11.0,<=0.11.0",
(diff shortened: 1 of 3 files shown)


multimodal/setup.py
```diff
@@ -41,7 +41,7 @@
     "smart_open>=5.2.1,<5.3.0",
     "pytorch_lightning>=1.7.4,<1.8.0",
     "text-unidecode<=1.3",
-    "torchmetrics>=0.8.0,<0.9.0",
+    "torchmetrics>=0.11.0,<=0.11.0",
```
zhiqiangdon, line 44: I think this upgrade has some API breakages. Need to investigate what metrics are affected. For example, `roc_auc` no longer supports `pos_label` in the new releases.

FANGAreNotGnu: Sure, also 0.11.0 didn't include PR that increase mAP evaluation speed. I'll copy the latest mAP implementation here instead of using the latest version.

bryanyzhu:
Hi, will there be any update for this PR? Thanks.

FANGAreNotGnu:
Close and use a newer version of mAP implementation instead,

PR #2578, opened 2022-12-16
[Refactor][Bug fix] Fix fit_summary + move problem_type outside predictor.py

sxjscience:
*Issue #, if available:*

*Description of changes:*

- [x] Fix fit_summary in MultiModalPredictor. Previously, calling `predictor.fit_summary()` won't work after you have loaded the predictor from folder.
- [x] Move problem_type registry outside of predictor.py. In addition, register problem type to ease future refactoring.
- [x] Revise tutorial and fix typo in multimodal NER tutorial filename.


multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -774,8 +743,11 @@ def fit(
             label_column=self._label_column,
             column_types=column_types,
             data=train_data,
-            provided_problem_type=self._problem_type,
+            provided_problem_type=self._problem_type.name,
         )
+        if problem_type is not None:
+            problem_type = problem_type_reg.get(problem_type)
+            self._problem_type = problem_type  # In case problem type isn't provided in __init__().
```
zhiqiangdon, line 750: This logic seems not to work for matching problems. For example, if `self._problem_type.name` is `image_similarity`, the inferred problem types from `infer_problem_type_output_shape ` is a traditional type, e.g., binary. Then `problem_type = problem_type_reg.get(problem_type)` would return the `binary` problem type, which changes the original `image_similarity`.

zhiqiangdon: I guess that few shot image classification will also have a similar issue.

multimodal/src/autogluon/multimodal/problem_types.py
```diff
@@ -0,0 +1,221 @@
...
+
+
+@dataclass
+class ProblemType:
+    """Problem type class. Stores the name of the problem
+    and its related properties. Some of the properties will be useful for the label / feature inference logic."""
+
+    name: str  # Name of the problem
+    support_fit: bool = True  # Whether the problem type support `.fit()`
+    inference_ready: bool = False  # Support `.predict()` and `.evaluate()` without calling `.fit()`
```
zhiqiangdon, line 43: I think inference_ready is better to be a state of predictor.

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -398,7 +371,7 @@ def __init__(
             set_logger_verbosity(verbosity, logger=logger)
 
         self._label_column = label
-        self._problem_type = problem_type if problem_type is not None else None
+        self._problem_type = problem_type
```
zhiqiangdon, line 374: Should be `problem_type.lower() if problem_type is not None else None`?

multimodal/src/autogluon/multimodal/problem_types.py
sxjscience: This indicates whether the problem supports zero-shot inference.

multimodal/src/autogluon/multimodal/predictor.py
sxjscience: It has been called previously.

docs/tutorials/multimodal/multimodal-faq.md
```diff
@@ -0,0 +1,30 @@
...
+
+predictor = MultiModalPredictor.load(SAVE_PATH)
+```
+
+
+### Do I need to preprocess my text or image data before using AutoGluon Multimodal?
+
+Usually you do not need to preprocess the text / image data. AutoGluon Multimodal has built-in 
+support of text / image preprocessing. However, this won't block you from appending custom preprocessing logic before 
+feeding in the dataframe to AutoGluon Multimodal.
```
bryanyzhu, line 26: `However, this won't block you from appending custom preprocessing logic before 
feeding in the dataframe to AutoGluon Multimodal.` Just want to confirm, if user already preprocess the image (e.g., minus 255, divided by customized mean, or data augmentation), will AutoGluon do another round of preprocessing and data augmentation?

bryanyzhu (commented):
Thanks for the refactor, looks good to me with a question.

docs/tutorials/multimodal/multimodal-faq.md
sxjscience: we will do another round of normalization. As in https://github.com/autogluon/autogluon/blob/3f0e36d9ee14dcd11f71d819fdf8be0b40b837da/multimodal/src/autogluon/multimodal/data/process_image.py#L313

multimodal/src/autogluon/multimodal/problem_types.py
zhiqiangdon: How about using name `support_zero_shot`? I think we may need `inference_ready ` in predictor in later refactoring.

multimodal/src/autogluon/multimodal/registry.py
```diff
@@ -1,6 +1,9 @@
 import json
+from collections import OrderedDict
 from json import JSONDecodeError
 from typing import List
+from typing import OrderedDict as t_OrderedDict
+from typing import Union
```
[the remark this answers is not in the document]
sxjscience: t_OrderedDict is used for type hint. Will remove Union.

multimodal/src/autogluon/multimodal/problem_types.py
sxjscience: Sounds good. support_zero_shot is reasonable.

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -434,42 +413,34 @@ def __init__(
...
-                    problem_type=problem_type,
-                    hyperparameters=hyperparameters,
-                    eval_metric=eval_metric,
-                    path=path,
-                    verbosity=verbosity,
-                    warn_if_exist=warn_if_exist,
-                    enable_progress_bar=enable_progress_bar,
-                )
-                return
+        if self.problem_property and self.problem_property.is_matching:
```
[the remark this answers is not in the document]
zhiqiangdon: Never mind. I find it in the property.

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -832,7 +801,6 @@ def fit(
             time_limit = timedelta(seconds=time_limit)
 
         # set attributes for saving and prediction
-        self._problem_type = problem_type  # In case problem type isn't provided in __init__().
```
zhiqiangdon, line 835: Wondering why do we need to move this line to above?

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -1152,7 +1119,7 @@ def _fit(
         )
 
         config = update_config_by_rules(
-            problem_type=self._problem_type,
+            problem_type=self.problem_type,
```
zhiqiangdon, line 1122: I see all `self._problem_type` is replaced by `self.problem_type`. Mind to tell the reasons?

multimodal/src/autogluon/multimodal/predictor.py
sxjscience: Set self._problem_type right after we got the inferred problem_type. This will avoid potential misuse of self._problem_type.

docs/tutorials/multimodal/multimodal-faq.md
sxjscience: I can add a flag to disable normalization in a later PR. I think this is a good point.

multimodal/src/autogluon/multimodal/predictor.py
sxjscience: I can also change all self.problem_type to self._problem_type so that it will be consistent.

sxjscience: One advantage of `self.problem_type` over `self._problem_type` is that you can never use `self.problem_type = XXX` which accidentally overwrite the value.

multimodal/src/autogluon/multimodal/problem_types.py
```diff
@@ -0,0 +1,221 @@
...
+
+# Matching: text <--> text, image <--> image, text <--> image
+PROBLEM_TYPES_REG.register(
+    TEXT_SIMILARITY,
+    ProblemTypeProperty(
+        name=TEXT_SIMILARITY,
+        support_zero_shot=True,
+        is_matching=True,
+        supported_modality_type={TEXT},
+        supported_label_type={CATEGORICAL, NUMERICAL},
```
cheungdaven, line 130: For matching problems, labels can be absent. Will it support this case?

multimodal/src/autogluon/multimodal/problem_types.py
```diff
@@ -0,0 +1,221 @@
...
+    CATEGORICAL,
+    CLASSIFICATION,
+    FEATURE_EXTRACTION,
+    FEW_SHOT_TEXT_CLASSIFICATION,
+    IMAGE,
+    IMAGE_BYTEARRAY,
+    IMAGE_SIMILARITY,
+    IMAGE_TEXT_SIMILARITY,
+    MULTICLASS,
+    NAMED_ENTITY_RECOGNITION,
```
cheungdaven, line 17: I do not recall I used this constant: "NAMED_ENTITY_RECOGNITION". Shall we use NER only?

sxjscience: Both `NER` and `NAMED_ENTITY_RECOGNITION` will be mapped to `NER`.

sxjscience: You can view `NAMED_ENTITY_RECOGNITION` as the alias of `NER` problem type.

multimodal/src/autogluon/multimodal/problem_types.py
sxjscience: For those case, we can treat the label type as BINARY. But we need to submit a follow-up PR to refactor the code-base to better use these flags.

zhiqiangdon: The logic of no label is different from `binary` label. We'd better distinguish between them.

sxjscience: Actually there are multiple concepts: modality type, column type and problem type. This is the first PR to split the problem type into a separate file. Need to submit follow-up PRs.

sxjscience: Matching itself also needs major refactor so I haven’t touched it extensively

zhiqiangdon (approved):
LGTM!

bryanyzhu (approved):
LGTM

PR #2582, opened 2022-12-19
Fix Tabular Ensemble Folding

yinweisu:
*Issue #, if available:*
https://github.com/autogluon/autogluon/issues/2581

*Description of changes:*
* `model_base` is not available when doing repeated bagging. Use `_get_child_aux_val` to retrieve params instead.
* Added check of if `is_initialized()` check in `BaggedEnsembleModel._get_child_aux_val()`. @Innixma Let me know if it's missing intentionally.


Innixma (approved):
LGTM! Thanks for finding and fixing the issue!

PR #2584, opened 2022-12-20
[timeseries] Check if the timestamps are sorted inside the TimeSeriesPredictor

shchur:
*Description of changes:*
- Current solution, where we always sort the timestamps when creating a `TimeSeriesDataFrame` introduces a significant overhead when dealing with large datasets. With this PR, we delegate the task of ensuring that the timestamps are sorted chronologically to the user with a check inside the `TimeSeriesPredictor`.


canerturkmen (approved):
LGTM!

PR #2593, opened 2022-12-21
[timeseries] Speed up prediction for GluonTS models

shchur:
*Description of changes:*
- Significantly speed up prediction for GluonTS models. The two main sources are:
  - Use of `pd.Series` instead of `TimeSeriesDataFrame` inside `SimpleGluonTSDataset` to avoid copying the static features.
  - Speed up `get_forecast_horizon_index_ts_dataframe` with a more efficient `groupby` operation
- Clean up redundant logic in `AbstractGluonTSModel. predict`.

### Benchmarking results on the M5 dataset
(only training for 1 epoch / 1 batch, so the training time/performance are meaningless here)

**With this PR**

5K items / 7M rows
```
Training timeseries model DeepAR.
        -0.9107       = Validation score (-mean_wQuantileLoss)
        5.99    s     = Training runtime
        47.57   s     = Validation (prediction) runtime
```

30K items / 47M rows
```
Training timeseries model DeepAR.
        -0.7019       = Validation score (-mean_wQuantileLoss)
        11.92   s     = Training runtime
        298.31  s     = Validation (prediction) runtime
```

**Current `master` branch**
5K items / 7M rows
```
Training timeseries model DeepAR.
        -0.8331       = Validation score (-mean_wQuantileLoss)
        6.61    s     = Training runtime
        81.16   s     = Validation (prediction) runtime
```

30K items / 47M rows
```
Training timeseries model DeepAR.
        -0.8672       = Validation score (-mean_wQuantileLoss)
        15.64   s     = Training runtime
        1107.97 s     = Validation (prediction) runtime
```

<details><summary>Code to reproduce the results</summary>

```python
import pandas as pd


prediction_length = 28
raw_data = pd.read_parquet("../m5/data/target.parquet")
static = pd.read_parquet("../m5/data/static.parquet")

raw_data["item_id"] = raw_data["item_id"].astype("str")
raw_data["timestamp"] = pd.to_datetime(raw_data["timestamp"])
static["item_id"] = static["item_id"].astype("str")
static.set_index("item_id", inplace=True)

print(f"Loaded dataset with {len(raw_data)} rows and {raw_data['item_id'].nunique()} items.")
df = TimeSeriesDataFrame(raw_data, static_features=static)

predictor = TimeSeriesPredictor(prediction_length=prediction_length, target="demand")
predictor.fit(
    train_data=df,
    hyperparameters={"DeepAR": {"epochs": 1, "num_batches_per_epoch": 1}},
)
```
</details>

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
+++ b/timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
@@ -47,7 +47,10 @@ class SimpleGluonTSDataset(GluonTSDataset):
     ):
         assert target_df is not None
         assert target_df.freq, "Initializing GluonTS data sets without freq is not allowed"
-        self.target_df = target_df
+        # Convert TimeSeriesDataFrame to pd.Series for faster processing
+        self.target_series = target_df.squeeze()
+        self.item_ids = target_df.item_ids
+        self.freq_ = target_df.freq
         self.feat_static_cat = feat_static_cat
         self.feat_static_real = feat_static_real
         self.feat_dynamic_real = feat_dynamic_real
@@ -61,24 +64,23 @@ class SimpleGluonTSDataset(GluonTSDataset):
         # for feature generation. If the frequency string doesn't match or is not provided, it raises an exception.
         # Here we bypass this by issuing a default "yearly" frequency, tricking it into not producing
         # any lags or features.
-        freq_ = self.target_df.freq
-        pd_offset = to_offset(freq_)
+        pd_offset = to_offset(self.freq_)
 
         # normalize freq str to handle peculiarities such as W-SUN
         offset_base_alias = pd_offset.name.split("-")[0]
 
-        return "A" if offset_base_alias is None or offset_base_alias not in GLUONTS_SUPPORTED_OFFSETS else freq_
+        return "A" if offset_base_alias is None or offset_base_alias not in GLUONTS_SUPPORTED_OFFSETS else self.freq_
 
     def __len__(self):
-        return len(self.target_df.item_ids)  # noqa
+        return len(self.item_ids)  # noqa
 
     def __iter__(self) -> Iterator[Dict[str, Any]]:
-        for item_id in self.target_df.item_ids:  # noqa
-            df = self.target_df.loc[item_id]
+        for item_id in self.item_ids:  # noqa
+            ts = self.target_series.loc[item_id]
             time_series = {
                 FieldName.ITEM_ID: item_id,
-                FieldName.TARGET: df.to_numpy(dtype=self.float_dtype).ravel(),
-                FieldName.START: pd.Period(df.index[0], freq=self.freq),
+                FieldName.TARGET: ts.to_numpy(dtype=self.float_dtype).ravel(),
+                FieldName.START: pd.Period(ts.index[0], freq=self.freq),
             }
             if self.feat_static_cat is not None:
                 time_series[FieldName.FEAT_STATIC_CAT] = self.feat_static_cat.loc[item_id].to_numpy(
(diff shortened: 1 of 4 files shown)


timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
```diff
@@ -350,22 +351,9 @@ def predict(
...
-                    df.set_index(
-                        df.index.set_levels([input_index_type(i) for i in df.index.levels[0]], level=0),
-                        inplace=True,
-                    )
-
-        # Make sure the item_ids are sorted in the same order as in data
-        df = df.loc[data.item_ids]
-        # GluonTS uses pd.Period internally, which may lead to loss of precision (e.g., "2020-01-01 12:00" becomes
-        # "2020-01-01 00:00"). We manually set the index to avoid such problems.
-        df.index = get_forecast_horizon_index_ts_dataframe(data, self.prediction_length)
```
shchur, line 368: We now handle this problem inside `_gluonts_forecasts_to_data_frame`. This, however, requires us to pass `data` to the method to avoid the loss of timestamp precision.

timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
```diff
@@ -380,7 +368,7 @@ def _predict_gluonts_forecasts(
 
     @staticmethod
     def _sample_to_quantile_forecast(forecast: SampleForecast, quantile_levels: List[float]) -> QuantileForecast:
-        forecast_arrays = []
+        forecast_arrays = [forecast.mean]
```
tonyhoo, line 371: Why we need to add `mean` forecasts while the function only specify `quantile`?

tonyhoo (approved):

timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
shchur: A `gluonts.model.forecast.QuantileForecast` object can store both the quantiles and the mean forecast (see `forecat_keys` [here](https://ts.gluon.ai/dev/api/gluonts/gluonts.model.forecast.html#gluonts.model.forecast.QuantileForecast)), so we save the mean here as well since we need to report it in output `TimeSeriesDataFrame`.

canerturkmen (commented):
Looks great, and very impressive speedups!! Some minor comments.

timeseries/src/autogluon/timeseries/models/gluonts/torch/models.py
```diff
@@ -234,13 +235,19 @@ def _get_estimator_init_args(self) -> Dict[str, Any]:
         return init_kwargs
 
     def _gluonts_forecasts_to_data_frame(
```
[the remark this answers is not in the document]
shchur: Resolved this by adding the method `_distribution_to_quantile_forecast`.

timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
```diff
@@ -389,21 +377,22 @@ def _sample_to_quantile_forecast(forecast: SampleForecast, quantile_levels: List
...
         if isinstance(forecast.start_date, pd.Timestamp):  # GluonTS version is <0.10
             forecast_init_args.update({"freq": forecast.freq})
         return QuantileForecast(**forecast_init_args)
 
     def _gluonts_forecasts_to_data_frame(
-        self, forecasts: List[Forecast], quantile_levels: List[float]
+        self,
+        forecasts: List[Forecast],
+        quantile_levels: List[float],
+        data: TimeSeriesDataFrame,
```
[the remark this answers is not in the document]
shchur: Added the `forecast_index` parameter, but I don't think we should make it optional. The logic for the two cases is very different and would always force us to keep the two possibilities in mind when working with this method.

canerturkmen (approved):
LGTM! Thanks!

PR #2596, opened 2022-12-21
[CI] parallelize unit tests in ci

liangfu:
*Issue #, if available:*

n/a

*Description of changes:*

1. install pytest-xdist package in autogluon-common[test]
2. parallelize [tabular, multimodal-predictor, multimodal-others] with two concurrent workers.

task | before | after
--|--|--
tabular | 35m | 32m
multimodal-predictor | 41m | 29m
multimodal-others | 39m | n/a

### DIFF
--- a/.github/workflow_scripts/test_tabular.sh
+++ b/.github/workflow_scripts/test_tabular.sh
@@ -17,7 +17,7 @@ install_vision
 cd tabular/
 if [ -n "$ADDITIONAL_TEST_ARGS" ]
 then
-    python3 -m pytest -n=2 --junitxml=results.xml --runslow "$ADDITIONAL_TEST_ARGS" tests
+    python3 -m pytest -n=2 -x --junitxml=results.xml --runslow "$ADDITIONAL_TEST_ARGS" tests
 else
-    python3 -m pytest -n=2 --junitxml=results.xml --runslow tests
+    python3 -m pytest -n=2 -x --junitxml=results.xml --runslow tests
 fi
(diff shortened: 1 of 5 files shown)


.github/workflow_scripts/test_tabular.sh
```diff
@@ -17,7 +17,7 @@ install_vision
 cd tabular/
 if [ -n "$ADDITIONAL_TEST_ARGS" ]
 then
-    python3 -m pytest --junitxml=results.xml --runslow "$ADDITIONAL_TEST_ARGS" tests
+    python3 -m pytest -n=2 --junitxml=results.xml --runslow "$ADDITIONAL_TEST_ARGS" tests
```
tonyhoo, line 20: Can we just do `-n auto`

liangfu: I think that might be a bit aggressive, since we might unexpectedly get OOMs. Shall we start with two, and move to more aggressive settings later?

yinweisu (commented):
This might cause some trouble as tabular bagging would try utilizing all the cores. Not sure if having two tests competing each other would work nice

.github/workflow_scripts/test_tabular.sh
tonyhoo: Fine with starting with 2 but would be good to controlled it in a centralized place so only minimal change required going forward. One idea is to define a new function at https://github.com/autogluon/autogluon/blob/9f33fcb7c0baebd8c9eedc1a44ebfaeb52b9518a/.github/workflow_scripts/env_setup.sh#L10 to set up an env variable `AG_TEST_PARALLELIZATION`

liangfu: That's a good point.

PR #2597, opened 2022-12-22
[eda] Distribution Fit

gradientsky:
*Description of changes:*

# Distribution Fit

[Open In Colab](https://colab.research.google.com/drive/1_Pd0hwdUg_BUk-wzgtW9uhzCqvFlZQE2?usp=sharing)

This component attempts to fit various distributions for further plotting via `FeatureInteractionVisualization`. The data specified in `columns` must be numeric to be considered for fitting (categorical variables are not supported). Only the distributions with statistical significance above `pvalue_min` threshold will be included in the results.

### Parameters

```
Parameters
----------
columns: Union[str, List[str]]
    colums to be included into analysis. Can be passed as a string or a list of strings.
pvalue_min: float = 0.01,
    min pvalue to consider including distribution fit in the results.
keep_top_n: Optional[int] = None,
    how many distributions exceeding `pvalue_min` to include in the results. I.e. if `keep_top_n=3`,
    but 10 distributions satisfied `pvalue_min`, only top 3 will be included.
    If not specified and `distributions_to_fit` is not provided, then only top 3 will be included in the results.
distributions_to_fit: Optional[Union[str, List[str]]] = None,
    list of distributions to fit. See `DistributionFit.AVAILABLE_DISTRIBUTIONS` for the list of supported values.
    If not specified, then all supported distributions will be attempted to fit.
```

Note: this analysis is an augmentation for `FeatureInteraction` and should be used in pair to be visualized via `FeatureInteractionVisualization`.

### Example

```python
import autogluon.eda.analysis as eda
import autogluon.eda.visualization as viz

auto.analyze(
    train_data=df_train,
    anlz_facets=[
        eda.dataset.RawTypesAnalysis(),
        eda.interaction.DistributionFit(columns=['Fare', 'Age'], distributions_to_fit=['lognorm', 'beta', 'gamma', 'fisk']),
        eda.interaction.FeatureInteraction(key='age-density', x='Age'),
        
    ],
    viz_facets=[
        viz.interaction.FeatureInteractionVisualization(key='age-density', headers=True),
    ]
)
```

image

#### Getting fitted distribution parameters
```python
state = auto.analyze(
    train_data=df_train, return_state=True,
    anlz_facets=[eda.interaction.DistributionFit(columns=['Age'], distributions_to_fit=['lognorm'])],
)
state.distributions_fit.train_data.Age.lognorm.param  # -> s, loc, scale
```

```
(0.12851501265611742, -82.47791093444627, 111.23853062837834)
```

Each distribution has it's own [documentation](https://docs.scipy.org/doc/scipy/reference/stats.html) for the parameters. See scipy documentation for [lognorm](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.lognorm.html)

image


eda/src/autogluon/eda/analysis/interaction.py
```diff
@@ -210,3 +208,242 @@ def _fit(self, state: AnalysisState, args: AnalysisState, **fit_kwargs):
...
+
+    See Also
+    --------
+    :py:class:`~autogluon.eda.analysis.interaction.FeatureInteraction`
+    :py:class:`~autogluon.eda.visualization.interaction.FeatureInteractionVisualization`
+    `scipy <https://docs.scipy.org/doc/scipy/reference/stats.html>`_ documentation for each distribution details
+
+    """
+
+    AVAILABLE_DISTRIBUTIONS = [
```
tonyhoo, line 286: Does this cover what scipy support or a subset? If it is a full list, can we just import from scipy if possible?

eda/src/autogluon/eda/analysis/interaction.py
```diff
@@ -210,3 +208,242 @@ def _fit(self, state: AnalysisState, args: AnalysisState, **fit_kwargs):
...
+        results = {}
+        if not is_numeric_dtype(series):
+            self.logger.warning(f"{series.name}: distribution cannot be fit; only numeric columns are supported")
+            return None
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore")
+            for i in self.distributions_to_fit:
+                dist = getattr(stats, i)
+                param = dist.fit(series)
+                statistic, pvalue = stats.kstest(series, i, args=param)
```
tonyhoo, line 433: I assume this will only work for continuous dist so worth to check if only continuous are supported

tonyhoo (approved):

eda/src/autogluon/eda/analysis/interaction.py
gradientsky: We check if the type is numeric; categorical types are not supported.

eda/src/autogluon/eda/analysis/interaction.py
gradientsky: Good idea - updated

PR #2601, opened 2022-12-23
[timeseries] [WIP] Speed up TimeSeriesEvaluator

shchur:
*Description of changes:*
- Greatly speed up the `TimeSeriesEvaluator` by avoiding splitting `data` into `data_past` and `data_future` at each metric evaluation.

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
+++ b/timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
@@ -327,13 +327,16 @@ class AbstractTimeSeriesModel(AbstractModel):
             time steps of each time series.
         """
         metric = self.eval_metric if metric is None else metric
+        data_past = data.slice_by_timestep(None, -self.prediction_length)
+        data_future = data.slice_by_timestep(-self.prediction_length, None)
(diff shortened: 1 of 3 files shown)


timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
```diff
@@ -327,13 +327,16 @@ def score(self, data: TimeSeriesDataFrame, metric: str = None, **kwargs) -> floa
...
+        data_future = data.slice_by_timestep(-self.prediction_length, None)
         evaluator = TimeSeriesEvaluator(
             eval_metric=metric,
             prediction_length=self.prediction_length,
             target_column=self.target,
+            data_past=data_past,
         )
         predictions = self.predict_for_scoring(data)
-        metric_value = evaluator(data, predictions)
+        metric_value = evaluator(data_future=data_future, predictions=predictions)
```
shchur, line 339: Probably we could introduce some utility function (to TSDF?) that splits the test data into:
- `data_past` - target + covariates before the forecast horizon
- `data_future` - target column for the forecast horizon
- `known_covariates_future` - known covariates for the forecast horizon

This would allow us to get rid of `predict_for_scoring` and unify the API a bit.

PR #2602, opened 2022-12-23
[timeseries] Speed up TimeSeriesEnsembleSelection.fit

shchur:
*Description of changes:*
- Avoid the costly `TimeSeriesDataFrame.slice_by_timestep` operation when computing the metrics while fitting an ensemble.

### Runtime comparison of `fit_ensemble()` with 2 base models
**M4-hourly (350K rows, 414 items)**
- Current `master` branch
```
Fitting simple weighted ensemble.
        33.31   s     = Training runtime
```
- This PR
```
Fitting simple weighted ensemble.
        4.59    s     = Training runtime
```

**M4-daily (10M rows, 4227 items)**
- Current `master` branch
```
Fitting simple weighted ensemble.
        1120.93 s     = Training runtime
```
- This PR
```
Fitting simple weighted ensemble.
        18.20   s     = Training runtime
```

### DIFF
--- a/timeseries/src/autogluon/timeseries/evaluator.py
+++ b/timeseries/src/autogluon/timeseries/evaluator.py
@@ -198,14 +198,16 @@ class TimeSeriesEvaluator:
     def save_past_metrics(self, data_past: TimeSeriesDataFrame):
         self._past_naive_1_error = in_sample_naive_1_error(y_past=data_past[self.target_column])
 
-    def score_with_saved_past_data(self, data_future: TimeSeriesDataFrame, predictions: TimeSeriesDataFrame) -> float:
+    def score_with_saved_past_metrics(
+        self, data_future: TimeSeriesDataFrame, predictions: TimeSeriesDataFrame
+    ) -> float:
         """Compute the metric assuming that the historic metrics have already been computed.
 
         This method should be preferred to TimeSeriesEvaluator.__call__ if the metrics are computed multiple times, as
         it doesn't require splitting the test data into past/future portions each time (e.g., when fitting ensembles).
         """
         assert (predictions.num_timesteps_per_item() == self.prediction_length).all()
-        assert self._past_naive_1_error is not None, "Call save_past_metrics before score_with_saved_past_data"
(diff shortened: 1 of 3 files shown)


timeseries/src/autogluon/timeseries/evaluator.py
```diff
@@ -195,11 +195,17 @@ def check_get_evaluation_metric(
             return TimeSeriesEvaluator.DEFAULT_METRIC
         return metric
 
-    def __call__(self, data: TimeSeriesDataFrame, predictions: TimeSeriesDataFrame) -> float:
+    def cache_historic_metrics(self, data_past: TimeSeriesDataFrame):
+        self.historic_naive_1_error = in_sample_naive_1_error(y_history=data_past[self.target_column])
```
canerturkmen, line 199: A couple of notes about caching.

- First, should this method and the instance variable be private?
- Strictly, this is not a cache. A cache is often something that a user doesn't explicitly interact with. How about just `save_past_metrics` ?

canerturkmen (approved):
Looks great! Some comments regarding naming and style.

timeseries/src/autogluon/timeseries/evaluator.py
shchur: - Agree, accepted all renaming suggestions & made `_past_naive_1_error` a private attribute.
- We call `cache_historic_metrics ` (now renamed to `save_past_metrics`) inside `TimeSeriesEnsembleSelection._fit`, so this one should be public.

timeseries/src/autogluon/timeseries/evaluator.py
```diff
@@ -195,11 +195,17 @@ def check_get_evaluation_metric(
             return TimeSeriesEvaluator.DEFAULT_METRIC
         return metric
 
-    def __call__(self, data: TimeSeriesDataFrame, predictions: TimeSeriesDataFrame) -> float:
+    def cache_historic_metrics(self, data_past: TimeSeriesDataFrame):
```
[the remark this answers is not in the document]
shchur: Both accepted 👍

canerturkmen (approved):
LGTM!

PR #2605, opened 2022-12-27
[timeseries] Fix inconsistent item_id handling by GluonTS

shchur:
*Description of changes:*
- This PR fixes a bug introduced by the refactoring of `AbstractGluonTSModel` introduced by #2593 that caused, e.g., `TemporalFusionTransformerMXNet` model to fail. 

Background: Some GluonTS models always convert `item_id` to a string, while others keep the original dtype (e.g., int). With this PR, we ensure that we always convert `item_id` to a string internally, but return the correct dtype to the user.


canerturkmen (approved):
LGTM!

PR #2606, opened 2022-12-27
[timeseries] Automatically convert timestamps to datetime & handle DataFrame inputs

shchur:
*Description of changes:*

This PR contains several potential improvements to the data loading process in `autogluon.timeseries`:
1. Automatically convert the `timestamp` columns to the `datetime64` format when constructing a `TimeSeriesDataFrame` from `pd.DataFrame` without multiindex.
2. Load a `TimeSeriesDataFrame` from path (local or remote), similar to `autogluon.tabular.TabularDataset`.
3. Accept `pd.DataFrame` as input to `TimeSeriesPredictor` and automatically convert to `TimeSeriesDataFrame` under the hood.

These changes make it possible to significantly shorten the code for training AGTS on a dataset. 
- Current minimal example
```python
import pandas as pd

raw_df = pd.read_csv(
    "https://autogluon.s3.amazonaws.com/datasets/timeseries/dummy/dummy.csv", 
    parse_dates=["timestamp"],
)
data = TimeSeriesDataFrame(raw_df)
predictor = TimeSeriesPredictor().fit(data)
```
- Using Option 2 from this PR
```python

data = TimeSeriesDataFrame("https://autogluon.s3.amazonaws.com/datasets/timeseries/dummy/dummy.csv")
predictor = TimeSeriesPredictor().fit(data)
```
- Using Option 3 from this PR
```python
import pandas as pd
from autogluon.timeseries import TimeSeriesPredictor

data = pd.read_csv("https://autogluon.s3.amazonaws.com/datasets/timeseries/dummy/dummy.csv")
predictor = TimeSeriesPredictor().fit(data)
```

### DIFF
--- a/timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
+++ b/timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
@@ -32,15 +32,7 @@ class TimeSeriesDataFrame(pd.DataFrame):
     data : Any
         Time-series data to construct a ``TimeSeriesDataFrame``. The class currently supports four input formats.
 
-        1. Time-series data in Iterable format. For example::
-
-                iterable_dataset = [
-                    {"target": [0, 1, 2], "start": pd.Timestamp("01-01-2019", freq='D')},
-                    {"target": [3, 4, 5], "start": pd.Timestamp("01-01-2019", freq='D')},
-                    {"target": [6, 7, 8], "start": pd.Timestamp("01-01-2019", freq='D')}
-                ]
-
-        2. Time-series data in a pandas DataFrame format without multi-index. For example::
+        1. Time-series data in a pandas DataFrame format without multi-index. For example::
 
                    item_id  timestamp  target
                 0        0 2019-01-01       0
@@ -53,7 +45,7 @@ class TimeSeriesDataFrame(pd.DataFrame):
                 7        2 2019-01-02       7
                 8        2 2019-01-03       8
 
-        3. Time-series data in pandas DataFrame format with multi-index on item_id and timestamp. For example::
+        2. Time-series data in pandas DataFrame format with multi-index on item_id and timestamp. For example::
 
                                         target
                 item_id timestamp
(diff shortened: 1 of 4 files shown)


timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
```diff
@@ -29,18 +30,9 @@ class TimeSeriesDataFrame(pd.DataFrame):
...
-        1. Time-series data in Iterable format. For example::
-
-                iterable_dataset = [
-                    {"target": [0, 1, 2], "start": pd.Timestamp("01-01-2019", freq='D')},
-                    {"target": [3, 4, 5], "start": pd.Timestamp("01-01-2019", freq='D')},
-                    {"target": [6, 7, 8], "start": pd.Timestamp("01-01-2019", freq='D')}
-                ]
-
-        2. Time-series data in a pandas DataFrame format without multi-index. For example::
+        1. Time-series data in a pandas DataFrame format without multi-index. For example::
```
shchur, line 35: Changed the order here to put the most commonly used option on top

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -205,6 +205,11 @@ def _check_and_prepare_data_frame(self, df: TimeSeriesDataFrame) -> TimeSeriesDa
         """
         if df is None:
             return df
+        if not isinstance(df, TimeSeriesDataFrame):
```
canerturkmen, line 208: should we document this behavior, and if not -- should we include it? i.e., type coercion has the risk that a user may provide a perfectly good data frame, something may implicitly get converted wrong, and they have no way to inspect what's wrong.

shchur: We say in documentation that we always expect `TimeSeriesDataFrame` objects as input (i.e., `data`, `train_data`, `tuning_data`, `known_covariates`) but don't perform any checks. 

This means, for example, if a user provides a `pd.DataFrame` without multiindex (that can be seamlessly converted to a TSDF with `TimeSeriesDataFrame(df)`), they will receive a cryptic exception like 
```
KeyError: 'Requested level (timestamp) does not match index name (None)'
```

With the proposed change, we either do the conversion automatically or tell the user 
- that the data must be in TSDF format
- why it cannot be converted to a TSDF automatically (as in the screenshot below)

Screenshot 2022-12-30 at 13 37 23

shchur: Do you have some scenario in mind where the type conversion may be problematic?

shchur: Documented the type coercion behavior, as we discussed.

canerturkmen (approved):
LGTM!

PR #2613, opened 2022-12-29
Suppport realtime inference for matcher

zhiqiangdon:
*Issue #, if available:*

*Description of changes:*

1. Add `predict()`, `realtime_predict()`, and `process_batch()` util functions to support both predictor and matcher.
2. Remove `_predict()`, `_realtime_predict()`, and `_process_batch` from `predictor.py`.
3. Add realtime inference unit tests for matcher.


sxjscience (approved):
LGTM

multimodal/src/autogluon/multimodal/utils/matcher.py
```diff
@@ -501,3 +502,39 @@ def convert_data_for_ranking(
...
+
+    Returns
+    -------
+    Probabilities.
+    """
+    if logits is not None:
+        prob = F.softmax(logits.float(), dim=1)[:, 1]
+    else:
+        cosine_similarity = F.cosine_similarity(embeddings1, embeddings2)
+        prob = 0.5 * (cosine_similarity + 1)
```
cheungdaven, line 535: Is it necessary to convert similarity into probability in this way? I guess users care more about the real matching similarity in most cases.

zhiqiangdon: This function is to compute the probability, normalized to range (0, 1), which is required in computing some validation metrics, e.g., roc_auc. To compute the cosine similarity, users can call another util function `compute_semantic_similarity`.

bryanyzhu (approved):
LGTM

PR #2620, opened 2023-01-02
Support presets in AutoMM predictor initialization

zhiqiangdon:
*Issue #, if available:*

*Description of changes:*
1. Let users use `presets` in predictor initialization. The presets is to control model quality: `best_quality`, `high_quality_fast_inference`, and `medium_quality_faster_inference`.
2. Fix ner presets. Use the registered ner presets rather than the default.
3. Add unit tests.

### DIFF
--- a/multimodal/src/autogluon/multimodal/presets.py
+++ b/multimodal/src/autogluon/multimodal/presets.py
@@ -543,3 +543,26 @@ def get_automm_presets(presets: str):
         )
 
     return overrides
+
+
+def get_preset_str(problem_type: str, presets: str):
+    """
+    Concatenate problem type and presets to get a registered preset string.
+
+    Parameters
+    ----------
+    problem_type
+        Problem type.
+    presets
+        Presets regarding model quality, e.g., best_quality, high_quality_fast_inference, and medium_quality_faster_inference.
+
+    Returns
+    -------
+    A registered preset string.
+    """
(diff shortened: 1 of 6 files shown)


multimodal/src/autogluon/multimodal/presets.py
```diff
@@ -465,19 +471,36 @@ def best_quality_ner():
...
+            "timm_image",
+            "ner_text",
+            "fusion_ner",
+        ],
         "model.ner_text.checkpoint_name": "microsoft/deberta-v3-base",
     }
 
 
+@automm_presets.register()
+def ner():
```
sxjscience, line 500: Shall we call it `default_ner`?

sxjscience:
LGTM in general. Minor comment on the naming.

We can consider to register `default_{problem_type}` for all the problem types, rather than using `{problem_type}`.

multimodal/src/autogluon/multimodal/presets.py
zhiqiangdon: Currently, all specific problem types, e.g., `image_text_similarity` and `object_detection`, have registered presets. `default` is for the general problem types `binary`, `multiclass`, and `regression`.

bryanyzhu (approved):
Looks good. One side question, where should we introduce these presets to users, in the apis documentation or automm customization tutorial page?

PR #2625, opened 2023-01-03
[CI]Fix torch install on platform tests

yinweisu:
*Issue #, if available:*

*Description of changes:*
* Security patches applied in https://github.com/autogluon/autogluon/pull/2588 is linux specific. For our platform tests, we install default torch.
* Platform tests are running in github hosted machine and doesn't contain credentials. Should be fine

### DIFF
--- a/.github/workflow_scripts/env_setup.sh
+++ b/.github/workflow_scripts/env_setup.sh
@@ -41,7 +41,7 @@ function setup_torch_gpu_non_linux {
 }
 
 function setup_torch_cpu_non_linux {
-    pip3 install torch==1.12.1+cpu torchvision==0.13.1+cpu torchtext==0.13.1 -f https://download.pytorch.org/whl/cpu/torch_stable.html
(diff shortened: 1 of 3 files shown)


.github/workflow_scripts/env_setup.sh
```diff
@@ -36,6 +36,14 @@ function setup_torch_cpu {
     python3 -m pip install --no-cache-dir -U ${TORCH_URL} ${TORCHVISION_URL}
 }
 
+function setup_torch_gpu_non_linux {
+    pip3 install torch==1.12.1+cu113 torchvision==0.13.1+cu113 torchtext==0.13.1 --extra-index-url https://download.pytorch.org/whl/cu113
+}
+
+function setup_torch_cpu_non_linux {
+    pip3 install torch==1.12.1 torchvision==0.13.1 torchtext==0.13.1
```
gradientsky, line 44: These has to be patched versions

gradientsky (changes requested):

PR #2628, opened 2023-01-03
Fix AutoMM usage of text in TabularPredictor

Innixma:
*Issue #, if available:*

*Description of changes:*
- Fixed case where AutoMM is specified as a model in TabularPredictor, but the global preprocessor does not keep raw text features in output, leading AutoMM to not use text features.
- Now this situation is fixed for all future models so long as they specify `handles_text=True` in `_class_tags`.

### DIFF
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -22,7 +22,7 @@ from autogluon.common.utils.log_utils import DuplicateFilter
 from autogluon.common.utils.resource_utils import ResourceManager
 
 from .model_trial import model_trial, skip_hpo
-from ._tags import _DEFAULT_TAGS
+from ._tags import _DEFAULT_CLASS_TAGS, _DEFAULT_TAGS
 from ... import metrics, Space
 from ...constants import AG_ARGS_FIT, BINARY, REGRESSION, QUANTILE, REFIT_FULL_SUFFIX, OBJECTIVES_TO_NORMALIZE
 from ...data.label_cleaner import LabelCleaner, LabelCleanerMulticlassToBinary
@@ -1606,8 +1606,14 @@ class AbstractModel:
     def _features(self):
         return self._features_internal
 
-    def _get_tags(self):
-        collected_tags = {}
+    def _get_tags(self) -> dict:
+        """
+        Tags are key-value pairs assigned to an object.
+        These can be accessed after initializing an object.
+        Tags are used for identifying if an object supports certain functionality.
+        """
+        # first get class tags, which are overwritten by any object tags
+        collected_tags = self._get_class_tags()
         for base_class in reversed(inspect.getmro(self.__class__)):
             if hasattr(base_class, '_more_tags'):
                 # need the if because mixins might not have _more_tags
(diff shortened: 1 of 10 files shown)


core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -1606,8 +1606,14 @@ def _ag_params(self) -> set:
     def _features(self):
         return self._features_internal
 
-    def _get_tags(self):
-        collected_tags = {}
+    def _get_tags(self) -> dict:
```
gradientsky, line 1609: discussion point: maybe call these `capabilities` vs `tags`? Tags seems too generic and not pointing to the main purpose of this labeling: what functionality is supported.

tabular/src/autogluon/tabular/models/automm/ft_transformer.py
```diff
@@ -98,3 +98,7 @@ def _get_default_ag_args_ensemble(cls, **kwargs) -> dict:
         }
         default_ag_args_ensemble.update(extra_ag_args_ensemble)
         return default_ag_args_ensemble
+
+    @classmethod
+    def _class_tags(cls):
+        return {'handles_text': False}
```
gradientsky, line 104: Maybe move these constants to a separate constant class where they can be fully documented + reduces likelihood of typos.

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -3494,6 +3488,38 @@ def clone_for_deployment(self,
...
         return predictor_clone if return_clone else predictor_clone.path
 
+    @staticmethod
+    def _check_if_hyperparameters_handle_text(hyperparameters: dict) -> bool:
+        """Check if hyperparameters contain a model that supports raw text features as input"""
+        models_in_hyperparameters = set()
+        is_advanced_hyperparameter_type = False
+        for key in hyperparameters:
+            if isinstance(key, int) or key == 'default':
+                is_advanced_hyperparameter_type = True
```
gradientsky, line 3498: add `break` to skip the rest of the cycle?

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -3494,6 +3488,38 @@ def clone_for_deployment(self,
...
+        if is_advanced_hyperparameter_type:
+            for key in hyperparameters:
+                for m in hyperparameters[key]:
+                    models_in_hyperparameters.add(m)
+        else:
+            for key in hyperparameters:
+                models_in_hyperparameters.add(key)
+        models_in_hyperparameters_raw_text_compatible = []
```
gradientsky, line 3506:
```python
for key, m in hyperparameters.items():
    models_in_hyperparameters.add(m if is_advanced_hyperparameter_type else key)
```

core/src/autogluon/core/models/abstract/abstract_model.py
Innixma: this is the API used by sklearn for this logic, so I'd prefer to be consistent with them.

tabular/src/autogluon/tabular/predictor/predictor.py
Innixma: added

tabular/src/autogluon/tabular/predictor/predictor.py
Innixma:
```
x = {1: {'GBM': {}}}
a = set()
is_advanced_hyperparameter_type = True
for key, m in x.items():
    a.add(m if is_advanced_hyperparameter_type else key)
  File "/Users/neerick/workspace/virtual/autogluon38/lib/python3.8/site-packages/IPython/core/interactiveshell.py", line 3433, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-8-2d35d45650af>", line 2, in <module>
    a.add(m if is_advanced_hyperparameter_type else key)
TypeError: unhashable type: 'dict
```

The proposed code is not the same and will error.

tabular/src/autogluon/tabular/models/automm/ft_transformer.py
Innixma: scikit-learn just uses strings: https://github.com/scikit-learn/scikit-learn/blob/ac665ad80be586a90c7d353ed37f1e64a4f547a1/sklearn/neighbors/_classification.py#L712-L713

I think we can also just use strings unless it becomes a problem.

gradientsky (approved):

Issue #2629, opened 2023-01-03
[BUG] AutoMM does not early stop upon reaching a perfect validation score


Innixma:
- [X] I have checked that this bug exists on the latest stable version of AutoGluon
- [X] and/or I have checked that this bug exists on the latest mainline of AutoGluon via source installation

**Describe the bug**
When a perfect validation score is reached, training continues despite no chance of obtaining a better result.

**Expected behavior**
The training should immediately stop for the model, with the epoch that reached a perfect validation score being used as the final model.

An easy way to achieve this is to use the metrics defined in `autogluon.core.metrics`, for example:

```python3
from autogluon.core.metrics import roc_auc
error = roc_auc.convert_score_to_error(1)
if error <= 0:
    early_stop()
```

**To Reproduce**

```python3

if __name__ == '__main__':
    from autogluon.tabular import TabularPredictor, TabularDataset
    path_prefix = 'https://autogluon.s3.amazonaws.com/datasets/AdultIncomeBinaryClassification/'
    path_train = path_prefix + 'train_data.csv'
    path_test = path_prefix + 'test_data.csv'

    label = 'class'
    sample = 10  # Number of rows to use to train / infer
    train_data = TabularDataset(path_train)

    if sample is not None and (sample < len(train_data)):
        train_data = train_data.sample(n=sample, random_state=0).reset_index(drop=True)

    test_data = TabularDataset(path_test)
    fit_kwargs = dict(
        train_data=train_data,
        hyperparameters={
            'AG_AUTOMM': {},
        },
    )
    predictor = TabularPredictor(
        label=label,
        eval_metric='roc_auc',
        verbosity=2,
    )
    predictor.fit(**fit_kwargs)

    leaderboard = predictor.leaderboard(test_data)

```

**Screenshots**
```
  | Name              | Type                | Params
----------------------------------------------------------
0 | model             | MultimodalFusionMLP | 109 M 
1 | validation_metric | AUROC               | 0     
2 | loss_func         | CrossEntropyLoss    | 0     
----------------------------------------------------------
109 M     Trainable params
0         Non-trainable params
109 M     Total params
439.137   Total estimated model params size (MB)
[...]

Issue #2555, opened 2022-12-12
[Improvement] Improve AutoMM Early Stopping Criteria


sxjscience:
Currently, the early stopping callback only uses `patience` and `mode`.

https://github.com/autogluon/autogluon/blob/8116f275f30e2c0f360808382b2905cf41796eee/multimodal/src/autogluon/multimodal/predictor.py#L1397-L1401

This is not optimal because some models may:

- Saturate the performance in the early stage of training 

   In the following, the model has reached `val_roc_auc` to be `1.0` but still keeps training.

   ```python
   Epoch 3, global step 55263: 'val_roc_auc' reached 1.00000 (best 1.00000), saving model to '/home/ec2-user/SageMaker/t
   ```

One option is to use the `stopping_threshold` flag in the EarlyStopping callback: https://pytorch-lightning.readthedocs.io/en/stable/common/early_stopping.html

Innixma:
@sxjscience @zhiqiangdon Please update status or mark as resolved if addressed.

sxjscience:
Resolved via https://github.com/autogluon/autogluon/pull/2633

PR #2633, opened 2023-01-04
AutoMM early stopping with a threshold

zhiqiangdon:
*Issue #, if available:*
#2555 #2629 

*Description of changes:*
Early stopping if reaching a perfect score. Can't use exactly 1 or 0 due to the precision issues. Tested it with the example in #2629.


multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -1261,6 +1261,7 @@ def _fit(
             monitor=task.validation_metric_name,
             patience=config.optimization.patience,
             mode=minmax_mode,
+            stopping_threshold=1 - 1e-6 if minmax_mode == MAX else 1e-6,
```
sxjscience, line 1264: If it's RMSE, the loss functions can actually go very small so 1E-6 may not be a good number.

We can extend METRIC_MODE_MAP https://github.com/autogluon/autogluon/blob/1386d0ecc5069c5837c048970afd64d086745b68/multimodal/src/autogluon/multimodal/constants.py#L88 and write a METRIC_STOPPING_THRESHOLD_MAP to discuss the stopping threshold case-by-case.

zhiqiangdon:
```
from sklearn.metrics import mean_squared_error
mean_squared_error([1], [1-1e-6], squared=False) ## gives value 1.0000000000287557e-06
mean_squared_error([1], [1-1e-6], squared=True) ## gives value 1.0000000000575112e-12
```
If the rmse metric reaches 1e-6, the mse training loss would be around 1e-12, which probably can't provide enough gradient.

sxjscience: I'm not sure whether 1E-6 for validation RMSE can be ignored for gradient. How about for us to not add the stopping threshold for RMSE for now since we haven't seen examples of RMSE hitting a perfect value. Usually, we should always allow the model to improve during the training process.

For ROC-AUC, will stopping_threshold=1.0 work?

In addition, we need to separately discuss the stopping threshold for DIRECT_LOSS, LOG_LOSS, CROSS_ENTROPY and other loss functions. So we may adopt the same logic as in https://github.com/autogluon/autogluon/blob/1386d0ecc5069c5837c048970afd64d086745b68/multimodal/src/autogluon/multimodal/constants.py#L88-L106

We can just create a `METRIC_STOPPING_THRESHOLD` variable.

sxjscience: Okay, just noticed that you have tested for stopping_threshold = 1 and it won’t work for ROC_AUC. So 1 - 1E-6 might be good for ROC_AUC. Or 1 - 1E-9 to play safer?

zhiqiangdon: I find that 1e-7 works, but 1e-8 or smaller doesn't stop early. I choose 1e-6 to be safe.

sxjscience: May be 1E-7 in case some methods are really strong?

Innixma, line 1264: Are we assuming all positive-is-better metrics have an optimal score of 1? I guess in general this is true, but it is a pretty flimsy assumption especially if a user can provide a custom metric. Ditto for assuming optimal score of 0 for negative-is-better metrics.

These situations are handled in my `autogluon.core.metrics` logic, where each metric has a defined optimum value.

Innixma: Beyond the above point, the fact we need an `if` here seems like unnecessary code bloat and increases risk of bugs. It would be better if the optimum value / minmax_mode was simply provided by the metric itself, rather than hard-coded into this line.

Innixma: for example:

`stopping_threshold = metric.optimum - metric.sign * epsilon`
which translates to:
`stopping_threshold = 1 - (1 * 1e-6)`

I think this would be a lot cleaner.

Both of these things are provided in `autogluon.core.metrics`.

```python3
from autogluon.core.metrics import roc_auc

epsilon = 1e-6
stopping_threshold = roc_auc.optimum - roc_auc._sign * epsilon

print(stopping_threshold)
```
```
0.999999
```

zhiqiangdon: Changed it to 1e-7.

zhiqiangdon: Good suggestion. Thanks.

sxjscience (approved):
LGTM

PR #2635, opened 2023-01-04
Improve save_path in AutoMM

zhiqiangdon:
*Issue #, if available:*

1. For object detection, if model_loaded=True, every time calling `.predict()`  would create a new save folder and change `_save_path`.
2. Creating saving folder in `init()` complicates the logic of save_path.
3. Predictor and matcher use different save_path logic.

*Description of changes:*

1. Remove the creation of saving folder from `init()`. Create saving folder only when it's used.
2. Remove the `_model_loaded` flag. For saving the prediction results of object detection, we can directly save to the path `_save_path`, whether the model is loaded or fitted.
3. Let matcher use the same save_path logic as predictor.

### DIFF
--- a/multimodal/src/autogluon/multimodal/utils/save.py
+++ b/multimodal/src/autogluon/multimodal/utils/save.py
@@ -128,7 +128,6 @@ def setup_save_path(
     proposed_save_path: Optional[str] = None,
     warn_if_exist: Optional[bool] = True,
     raise_if_exist: Optional[bool] = False,
-    model_loaded: Optional[bool] = None,
     fit_called: Optional[bool] = None,
 ):
     # TODO: remove redundant folders in DDP mode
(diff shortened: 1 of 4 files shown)


FANGAreNotGnu (approved):
LGTM, thanks for the fix!

sxjscience (approved):
LGTM

multimodal/src/autogluon/multimodal/utils/save.py
```diff
@@ -139,10 +138,10 @@ def setup_save_path(
     elif proposed_save_path is not None:  # TODO: distinguish DDP and existed predictor
         save_path = process_save_path(path=proposed_save_path, raise_if_exist=(raise_if_exist and rank == 0))
     elif old_save_path is not None:
-        if model_loaded or fit_called:
+        if fit_called:
             save_path = process_save_path(path=old_save_path, raise_if_exist=False)
         else:
-            save_path = os.path.abspath(os.path.expanduser(old_save_path))
+            save_path = process_save_path(path=old_save_path, raise_if_exist=(raise_if_exist and rank == 0))
```
sxjscience, line 144: We may need a multi-GPU machine in the CI to test for multi-GPU training next time.

zhiqiangdon: Agree. Maybe select partial tests to avoid slow down CI.

bryanyzhu (approved):
Awesome, thanks for the fix.

PR #2644, opened 2023-01-05
Add YOLOX support

FANGAreNotGnu:
- Add support for YOLOX (yolox-s and yolox-l)
- Update finetuning example to include yolox
- Update code loading customized detection configs: store locally in multimodal/configs/pretrain/detection instead of download from s3.
- Add `max_tuning_num` in `fit()` to control the max number of data to use for validation.
- Add `clean_old_ckpts` flag to control if we shall clean the old ckpts.

YoloX-L performance on pothole dataset:
training time: 1337 seconds
mAP: 0.449
mAP50: 0.736

Previous best (VFNet-R50):
training time: ~18000 seconds
mAP: 0.450
mAP50: 0.738

TODO:
- Test the YOLOX pretrained on objects365
- More pretrain tunings.
- Test YOLOX-tiny/s/m/x
- update tutorial to use YOLOX


multimodal/src/autogluon/multimodal/configs/pretrain/detection/default_runtime.py
```diff
@@ -0,0 +1,28 @@
+checkpoint_config = dict(interval=1)
```
bryanyzhu, line 1: I have a quick question. If users just do `pip install autogluon` and wants to change the config, what should they do? Do they download the config from somewhere first, and then change hyperparameters manually? Thanks.

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -677,7 +680,11 @@ def fit(
             self._detection_anno_train = train_data
             train_data = from_coco_or_voc(train_data, "train")
             if tuning_data is not None:
+                self.detection_anno_train = tuning_data
                 tuning_data = from_coco_or_voc(tuning_data, "val")
+                if max_tuning_num is not None:
```
bryanyzhu, line 685: Will this work for classification or other problem types as well? Because this argument might solve the problem we had before. That is validation takes too much time, which leads to model fitting doesn't honor user specified time limit.

multimodal/src/autogluon/multimodal/configs/pretrain/detection/default_runtime.py
FANGAreNotGnu: We only retrieve the model structure and list of transforms from the config. Those parts are usually not changed and thus currently we do not support changing the config nor loading a customized config.

multimodal/src/autogluon/multimodal/predictor.py
FANGAreNotGnu: Sure, I'll do it in another PR (combine this with [these line](https://github.com/autogluon/autogluon/blob/fedabac33b8595d07564074411da20251a1c5048/multimodal/src/autogluon/multimodal/predictor.py#L712-L727)).

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -202,6 +202,7 @@ def __init__(
         enable_progress_bar: Optional[bool] = None,
         init_scratch: Optional[bool] = False,
         sample_data_path: Optional[str] = None,
+        clean_old_ckpts: Optional[bool] = True,
```
zhiqiangdon, line 205: What's the case of not cleaning the old checkpoints?

FANGAreNotGnu: While pretraining on a very large dataset (e.g. objects365), validation with original split may crash the process. So we may want to save the checkpoints and evaluate them on valid dataset after the training.

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -528,6 +530,7 @@ def fit(
         presets: Optional[str] = None,
         config: Optional[dict] = None,
         tuning_data: Optional[Union[pd.DataFrame, str]] = None,
+        max_tuning_num: Optional[int] = None,
```
zhiqiangdon, line 533: `max_tuning_num` may also sound like the number of validations to conduct.

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -677,7 +680,11 @@ def fit(
             self._detection_anno_train = train_data
             train_data = from_coco_or_voc(train_data, "train")
             if tuning_data is not None:
+                self.detection_anno_train = tuning_data
                 tuning_data = from_coco_or_voc(tuning_data, "val")
+                if max_tuning_num is not None:
+                    if len(tuning_data) > max_tuning_num:
+                        tuning_data = tuning_data.sample(n=max_tuning_num, replace=False).reset_index(drop=True)
```
zhiqiangdon, line 687: Need to set `random_state`?

multimodal/src/autogluon/multimodal/predictor.py
zhiqiangdon: Is the crash some kind of bug?

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -402,6 +403,7 @@ def __init__(
         self._sample_data_path = sample_data_path
         self._fit_called = False  # While using ddp, after fit called, we can only use single gpu.
         self._matcher = None
+        self._clean_old_ckpts = clean_old_ckpts
```
zhiqiangdon, line 406: Is it better to make clean_old_ckpts an argument in `fit()`? Whether to clean or not is a behavior of one training trial.

zhiqiangdon: Besides, `clean_old_ckpts` make users wonder what are old checkpoints. How about just calling `clean_ckpts`?

multimodal/src/autogluon/multimodal/predictor.py
FANGAreNotGnu: Not really, there may be unavoidable OOM errors on RAM or GRAM.

multimodal/src/autogluon/multimodal/predictor.py
FANGAreNotGnu: True, how about max_num_tuning_data

multimodal/src/autogluon/multimodal/predictor.py
FANGAreNotGnu: Moved to fit and renamed to clean_ckpts.

multimodal/src/autogluon/multimodal/predictor.py
FANGAreNotGnu: Done.

multimodal/src/autogluon/multimodal/predictor.py
zhiqiangdon: If OOM errors happen, do we need to turn the validation off in training?

zhiqiangdon: If the errors are random, we can resume training and still do validation.

FANGAreNotGnu: It's very possible to have OOM error if the validation set is very large. We can sample a small portion of validation data to give us some rough information about the training progress. But we don't delete the checkpoints and will do the full evaluation on validation set after training.

FANGAreNotGnu: Nick bring some concerns today tho that we may still need early stopping, and Alex also mentioned that we might want more advanced sampling options strategy than uniform. We could work on this in next PR.

bryanyzhu (approved):
LGTM, thanks for bringing yolox in.

multimodal/src/autogluon/multimodal/predictor.py
zhiqiangdon: Validation generally uses less memory than training. What's the cause of OOM?

zhiqiangdon (approved):
LGTM!

PR #2647, opened 2023-01-06
Fix null images

zhiqiangdon:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/multimodal/src/autogluon/multimodal/data/preprocess_dataframe.py
+++ b/multimodal/src/autogluon/multimodal/data/preprocess_dataframe.py
@@ -499,7 +499,7 @@ class MultiModalFeaturePreprocessor(TransformerMixin, BaseEstimator):
             if col_type == ROIS:
                 processed_data = df[col_name].tolist()
             elif col_type in [IMAGE_PATH, IMAGE]:
-                processed_data = col_value.apply(lambda ele: ele.split(";") if ele else ["null"]).tolist()
+                processed_data = col_value.apply(lambda ele: str(ele).split(";")).tolist()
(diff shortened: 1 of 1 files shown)


multimodal/src/autogluon/multimodal/data/preprocess_dataframe.py
```diff
@@ -499,7 +499,7 @@ def transform_image(
             if col_type == ROIS:
                 processed_data = df[col_name].tolist()
             elif col_type in [IMAGE_PATH, IMAGE]:
-                processed_data = col_value.apply(lambda ele: ele.split(";")).tolist()
+                processed_data = col_value.apply(lambda ele: str(ele).split(";")).tolist()
```
tonyhoo, line 502: This will convert the None value to ["None"], will that introduce issue during training as the image can't be loaded or we will just ignore invalid path later one

sxjscience: It will be guarded by https://github.com/autogluon/autogluon/blob/70dd543682a6ed0bb1173274e5de63071a71336e/multimodal/src/autogluon/multimodal/data/process_image.py#L359-L365

zhiqiangdon: We would try open the images and skip them if failures happen.

sxjscience (approved):
LGTM as a simple fix. However, we will need to handle/test missing values for all column types (including ROI / entity). Also CC @FANGAreNotGnu @cheungdaven

tonyhoo (approved):

PR #2648, opened 2023-01-06
[eda] shortcut for covariate shift detection

gradientsky:
*Description of changes:*

### `auto.covariate_shift_detection()` shortcut for distributions shift detection

This is a shortcut for analyze and visualize steps encapsulated in a single call.

```
Parameters
----------
train_data: Optional[DataFrame]
    training dataset
test_data: Optional[DataFrame]
    test dataset
label: : Optional[str]
    target variable
state: Union[None, dict, AnalysisState], default = None
    pass prior state if necessary; the object will be updated during `anlz_facets` `fit` call.
sample: Union[None, int, float], default = None
    sample size; if `int`, then row number is used;
    `float` must be between 0.0 and 1.0 and represents fraction of dataset to sample;
    `None` means no sampling
    See also :func:`autogluon.eda.analysis.dataset.Sampler`
path: Optional[str], default = None,
    path for models saving
return_state: bool, default = False
    return state if `True`
verbosity: int, default = 0
    Verbosity levels range from 0 to 4 and control how much information is printed.
    where `L` ranges from 0 to 50 (Note: higher values of `L` correspond to fewer print statements, opposite of verbosity levels).
fit_args
    kwargs to pass into `TabularPredictor` fit
```

**Example**:

```python

 # use defaults
auto.covariate_shift_detection(train_data=df_train, test_data=df_test, label=target_col)

# customize classifier and verbosity level
auto.covariate_shift_detection(train_data=..., test_data=..., label=..., verbosity=2, hyperparameters = {'GBM': {}})
```

image

### DIFF
--- a/eda/src/autogluon/eda/analysis/shift.py
+++ b/eda/src/autogluon/eda/analysis/shift.py
@@ -124,9 +124,8 @@ class XShiftDetector(AbstractAnalysis, StateCheckMixin):
         else:
             fi_scores = None
         pvalue = self.C2ST.pvalue(num_permutations=self.num_permutations)
-        det_status = pvalue <= self.pvalue_thresh
         state.xshift_results = {
-            "detection_status": det_status,
+            "detection_status": bool(pvalue <= self.pvalue_thresh),  # numpy.bool_ -> bool
             "test_statistic": self.C2ST.test_stat,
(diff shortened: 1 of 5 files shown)


gidler (approved):
LGTM

PR #2651, opened 2023-01-06
[tabular] add tabular regression tests into CI

liangfu:
*Issue #, if available:*

Some failures are not captured since we didn't add tabular regression as part of our CI, e.g. #2650.

*Description of changes:*

We're currently have `--runslow` argument in pytest. Therefore, marking regression as slow would simply add regression tests into regular test cases.

### DIFF
--- a/tabular/tests/regressiontests/test_tabular_regression.py
+++ b/tabular/tests/regressiontests/test_tabular_regression.py
@@ -390,7 +390,7 @@ def make_dataset(request, seed):
 def myfloor(x, base=.01):
   return round(base * math.floor(float(x)/base),8)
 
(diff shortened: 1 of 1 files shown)


Innixma:
This code was by an outside contributor in March 2022. While it worked when it was merged, we haven't tested it yet. I think it is better to wait until we refactor benchmarking to revisit enabling/replacing these tests.

Innixma (changes requested):
Blocking for now, we want to revisit once we refactor benchmarking

liangfu:
Regression tests are generally computational expensive to run in each commit. Deprioritizing this for now.

PR #2659, opened 2023-01-06
[NER] Change Validation loss to F1

cheungdaven:
*Description of changes:*

1. Change validation loss to F1 loss.

### DIFF
--- a/docs/tutorials/multimodal/text_prediction/ner.md
+++ b/docs/tutorials/multimodal/text_prediction/ner.md
@@ -45,6 +45,7 @@ model_path = f"./tmp/{uuid.uuid4().hex}-automm_ner"
 predictor = MultiModalPredictor(problem_type="ner", label=label_col, path=model_path)
 predictor.fit(
     train_data=train_data,
+    hyperparameters={'model.ner_text.checkpoint_name':'bert-base-cased'},
     time_limit=300, #second
 )
 ```
@@ -64,8 +65,6 @@ sentence = "Game of Thrones is an American fantasy drama television series creat
 predictions = predictor.predict({'text_snippet': [sentence]})
 print('Predicted entities:', predictions[0])
 
-for entity in predictions[0]:
-    print(f"Word '{sentence[entity['start']:entity['end']]}' belongs to group: {entity['entity_group']}")
 ```
 
 ## Visualization
--- a/multimodal/src/autogluon/multimodal/optimization/utils.py
+++ b/multimodal/src/autogluon/multimodal/optimization/utils.py
@@ -44,6 +44,7 @@ from ..constants import (
     MULTI_NEGATIVES_SOFTMAX_LOSS,
     MULTICLASS,
     NER,
+    NER_TOKEN_F1,
     NORM_FIT,
     OBJECT_DETECTION,
     OVERALL_ACCURACY,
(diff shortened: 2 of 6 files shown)


multimodal/src/autogluon/multimodal/optimization/utils.py
```diff
@@ -263,6 +264,8 @@ def get_metric(
     metric_name = metric_name.lower()
     if metric_name in [ACC, ACCURACY, OVERALL_ACCURACY]:
         return torchmetrics.Accuracy(), None
+    elif metric_name == OVERALL_F1:
```
sxjscience, line 267: Change to `NER_TOKEN_F1`?

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -1421,6 +1422,9 @@ def _top_k_average(
         strict_loading=True,
         standalone=True,
     ):
+        if self._problem_type is not None and self._problem_type == NER:
+            validation_metric_name = OVERALL_F1  # seqeval only support overall_f1
```
sxjscience, line 1426: There is no need to add the additional "if" here. If the problem is NER, the proper validation_metric and eval_metric should be set here: https://github.com/autogluon/autogluon/blob/10b39aceb35e943f7c62f095f0005b5b42aa8aa4/multimodal/src/autogluon/multimodal/utils/metric.py#L104-L105

cheungdaven: > There is no need to add the additional "if" here. If the problem is NER, the proper validation_metric and eval_metric should be set here:
> 
> https://github.com/autogluon/autogluon/blob/10b39aceb35e943f7c62f095f0005b5b42aa8aa4/multimodal/src/autogluon/multimodal/utils/metric.py#L104-L105

in line 1475, best_score = self.evaluate(val_df, [validation_metric_name])[validation_metric_name], the evaluate function is called and validation_metric_name is used. However, in the evaluate function, seqeval is used and only overall_f1, overall_accuracy, overall_precision, overall_recall, and tag name are valid evaluation metric. NER_TOKEN_F1 is not supported, that is why I used OVERALL_F1 as validation_metric_name and eval_metric_name. To separate them, we have to do this trick here.

sxjscience (approved):
LGTM

docs/tutorials/multimodal/text_prediction/ner.md
```diff
@@ -45,6 +45,7 @@ model_path = f"./tmp/{uuid.uuid4().hex}-automm_ner"
 predictor = MultiModalPredictor(problem_type="ner", label=label_col, path=model_path)
 predictor.fit(
     train_data=train_data,
+    hyperparameters={'model.ner_text.checkpoint_name':'bert-base-cased'},
```
tonyhoo, line 48: Can we make this default? Not neccessary for 0.6.2 release, but better to refactor and make it simpler for users going forward (e.g. 0.7 release)

multimodal/src/autogluon/multimodal/predictor.py
sxjscience: Updated that we added a `#FIXME` in the source code. So I'm resolving this and we will revisit later.

docs/tutorials/multimodal/text_prediction/ner.md
cheungdaven: It uses deberta base by default. I set it to bert-base-cased as it gives better demonstration results given such a short training time (300s) just for tutorial purpose.

tonyhoo (approved):

PR #2663, opened 2023-01-09
allow psutil < 6

jfcg:
fixes autogluon build on MacOS 12.6 and Arm64

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of Apache-2.0 license.

### DIFF
--- a/timeseries/setup.py
+++ b/timeseries/setup.py
@@ -27,7 +27,6 @@ install_requires = [
     "numpy",
     "scipy",
     "pandas",
-    "psutil",
     "statsmodels~=0.13.0",
     "gluonts~=0.11.0",
     "torch>=1.9,<1.13",
--- a/timeseries/setup.py
+++ b/timeseries/setup.py
@@ -27,7 +27,7 @@ install_requires = [
     "numpy",
     "scipy",
     "pandas",
-    "psutil>=5.7.3,<6",
+    "psutil",
     "statsmodels~=0.13.0",
     "gluonts~=0.11.0",
     "torch>=1.9,<1.13",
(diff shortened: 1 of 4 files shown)


timeseries/setup.py
```diff
@@ -27,7 +27,7 @@
     "numpy",
     "scipy",
     "pandas",
-    "psutil>=5.7.3,<5.9",
+    "psutil>=5.7.3,<6",
```
Innixma, line 30: `psutil>=5.7.3,<5.10` will be safer. We generally avoid capping at major versions as it is less stable than minor version capping.

Innixma (commented):
Thanks for the contribution! Once my comment is addressed, happy to merge. Please also rebase the PR as we fixed a bug in the CI.

Innixma:
I updated the PR to reflect my comment, will see if CI passes. I also fixed version range for networkx in timeseries which also had a similar problem.

jfcg:
> I updated the PR to reflect my comment, will see if CI passes. I also fixed version range for networkx in timeseries which also had a similar problem.

Thank you.

timeseries/setup.py
shchur: Actually, it seems that we don't use `psutil` anywhere in `timeseries`, so it's fine to remove this dependency completely from `timeseries/setup.py`.

shchur (approved):
LGTM!

jfcg:
Hi @tonyhoo, this PR seems to have made it to `v0.6.2` but not mentioned in the release notes, fyi.

Thanks..

PR #2665, opened 2023-01-09
[timeseries] Refactor handling of static/dynamic covariates with CovariateMetadata

shchur:
*Description of changes:*
- Preprocessing of static/dynamic covariates inside learner is moved to a separate class `TimeSeriesFeatureGenerator` (similar to [AbstractFeatureGenerator](https://github.com/autogluon/autogluon/blob/master/features/src/autogluon/features/generators/abstract.py#L22) in `autogluon.features`).
- All models are given a `CovariateMetadata` object that maps different types of covariates (static/known/past) to columns of the dataset.


canerturkmen (approved):
LGTM! Left some comments. Also, a friendly reminder to wait for the release prior to merging.

PR #2676, opened 2023-01-11
[NER] Add support for Chinese Named Entities 

cheungdaven:
*Description of changes:*
1. Add support for Chinese NER.
2. Add unit test for Chinese NER.
3. Automatically generate BIO tags when tags are not in BIO format.


multimodal/src/autogluon/multimodal/data/labelencoder_ner.py
```diff
@@ -35,6 +36,7 @@ def fit(self, y: pd.Series, x: pd.Series):
         _, entity_groups = self.extract_ner_annotations(y)
         self.unique_entity_groups = self.ner_special_tags + entity_groups
         self.entity_map = {entity: index for index, entity in enumerate(self.unique_entity_groups)}
+        self.config.entity_map = self.entity_map
```
sxjscience, line 39: Do we need to put entity_map under `self.config`? Or we just need to keep `self.entity_map`?

cheungdaven: need this entity map in dataprocessor, that's why it is put in config. special tags such as "O" is also in config.

sxjscience: Shall we extend the`NerProcessor` to include the `config` keyword then?

sxjscience:
```python
class NerProcessor:
    """
    Prepare NER data for the model specified by "prefix".
    """
    def __init__(
        self,
        model: nn.Module,
        max_len: Optional[int] = None,
        entity_map: Optional[dict] = None,
        config: Optional[DictConfig] = None,
    ):
```

cheungdaven: already add self.config in nerprocessor. entity_map can be accessed by self.config.entity_map.

multimodal/src/autogluon/multimodal/data/utils.py
```diff
@@ -336,13 +337,26 @@ def process_ner_annotations(ner_annotations, ner_text, tokenizer, is_eval=False)
     num_words = len(set(token_to_word_mappings)) - 1
     word_label = [1] * num_words
     # TODO: Potentially optimize word label generation via binary search
-    for idx, word_offset in enumerate(word_offsets[:num_words, :]):
-        for annot in ner_annotations:
-            custom_offset = annot[0]
-            custom_label = annot[1]
+    b_prefix = "B-"
```
sxjscience, line 340: Need to add "entity_map" in the docstring:

```python
def process_ner_annotations(ner_annotations, ner_text, entity_map, tokenizer, is_eval=False):
    """
    Generate token-level/word-level labels with given text and NER annotations.

    Parameters
    ----------
    ner_annotations
        The NER annotations.
    ner_text
        The corresponding raw text.
    entity_map
        The entity map between token label to word label.
    tokenizer
        The tokenizer to be used.
    is_eval
        Whether it is for evaluation or not, default: False

    Returns
    -------
    Token-level/word-level labels and text features.
    """
```

multimodal/tests/unittests/others/test_ner_chinese.py
```diff
@@ -0,0 +1,107 @@
...
+    train_df, dev_df = get_data()
+    predictor = MultiModalPredictor(
+        problem_type="ner",
+        label="entity_annotations",
+    )
+    predictor.fit(
+        train_data=train_df,
+        tuning_data=dev_df,
+        hyperparameters={
+            "model.ner_text.checkpoint_name": "microsoft/mdeberta-v3-base",
```
sxjscience, line 99: Shall we use a smaller model?

sxjscience (approved):
LGTM!

PR #2680, opened 2023-01-12
[timeseries] Add support for past_covariates to Predictor & models

shchur:
*Description of changes:*
- Fix several bugs related to `past_covariates` handling
- Include documentation for `past_covariates`
- Add support for `past_covariates` and `known_covariates` to `AutoGluonTabular` model
- Add support for `past_covariates` to `TemporalFusionTransformerMXNet` model
- Add tests

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/autogluon_tabular/tabular_model.py
+++ b/timeseries/src/autogluon/timeseries/models/autogluon_tabular/tabular_model.py
@@ -81,8 +81,8 @@ class AutoGluonTabularModel(AbstractTimeSeriesModel):
             **kwargs,
         )
         self._target_lag_indices: np.array = None
-        self._known_covariates_lag_indices = np.arange(0, 10)
-        self._past_covariates_lag_indices = np.arange(1, 10)
+        self._known_covariates_lag_indices: np.array = None
+        self._past_covariates_lag_indices: np.array = None
         self._time_features: List[Callable] = None
         self._available_features: pd.Index = None
         self.residuals_std = 0.0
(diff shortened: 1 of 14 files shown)


timeseries/tests/unittests/models/test_autogluon_tabular.py
```diff
@@ -46,15 +49,36 @@ def test_when_predict_is_called_then_get_features_dataframe_receives_correct_inp
...
 
 
-def test_when_static_features_present_then_shape_is_correct(temp_model_path):
-    data = DATAFRAME_WITH_STATIC.copy()
-    model = AutoGluonTabularModel(path=temp_model_path)
-    # Initialize model._lag_indices and model._time_features from freq
+@pytest.mark.parametrize("known_covariates_names", [["known_1", "known_2"], []])
+@pytest.mark.parametrize("past_covariates_names", [["past_1", "past_2", "past_3"], []])
+@pytest.mark.parametrize("static_features_names", [["cat_1"], []])
+def test_when_covariates_and_features_present_then_shape_is_correct(
```
canerturkmen, line 55: nit: which shape is correct?

timeseries/src/autogluon/timeseries/models/autogluon_tabular/tabular_model.py
```diff
@@ -106,25 +109,62 @@ def _get_features_dataframe(
...
+            df: pd.DataFrame, num_hidden: np.ndarray, lag_indices: np.ndarray, num_columns: int = 1
+        ) -> pd.DataFrame:
+            if num_columns > 1:
+                lag_indices = np.repeat(lag_indices, num_columns)
+            mask = num_hidden[:, None] >= lag_indices[None]  # shape [len(num_hidden), len(lag_indices) * num_columns]
+            df[mask] = np.nan
+            return df
+
+        def get_lag_features_and_target(time_series):
             # Starting from the end of the time series, mask the values as if the last `prediction_length` steps weren't observed
```
canerturkmen, line 126: is this "docstring" outdated? can we add docstrings to all of the nested functions? I suspect this method got a little too complicated and could use a refactor.

timeseries/src/autogluon/timeseries/models/autogluon_tabular/tabular_model.py
```diff
@@ -51,10 +52,10 @@ class AutoGluonTabularModel(AbstractTimeSeriesModel):
     PREDICTION_BATCH_SIZE = 100_000
 
     TIMESERIES_METRIC_TO_TABULAR_METRIC = {
-        "MASE": "root_mean_squared_error",
+        "MASE": "mean_absolute_error",
```
tonyhoo, line 55: Any reason why we change the mapping?

timeseries/src/autogluon/timeseries/models/autogluon_tabular/tabular_model.py
```diff
@@ -105,26 +108,132 @@ def _get_features_dataframe(
...
             # We don't hide any past values for the first `remainder` values, otherwise the features will be all empty
-            remainder = len(group) - num_windows * self.prediction_length
+            remainder = len(time_series) - num_windows * self.prediction_length
             num_hidden = np.concatenate([np.zeros(remainder), np.tile(np.arange(self.prediction_length), num_windows)])
-            mask = num_hidden[:, None] >= self._lag_indices[None]  # shape [num_timesteps, num_lags]
-            features[mask] = np.nan
+            target_lags = apply_mask(target_lags, num_hidden=num_hidden, lag_indices=self._target_lag_indices)
+            feature_dfs = [target_lags]
+
+            if self.metadata.past_covariates_real:
```
tonyhoo, line 206: wondering if we can support `past_covariates_cat` in the same way here

tonyhoo (commented):
Some minor comments. Can you also run some smoke tests on one large dataset (e.g. M5 sell_price as past_covariates_real) and see if we can handle it gracefully in reasonable amount of time for medium_quality and best_quality presets

timeseries/src/autogluon/timeseries/models/autogluon_tabular/tabular_model.py
shchur: That would also require some other logic throughout the code (e.g., `AbstractTimeSeriesModel.predict_for_scoring`) & require us to infer the covariate type (categorical vs. real). For this reason, I would rather postpone the addition of categorical covariates to v0.8.

timeseries/src/autogluon/timeseries/models/autogluon_tabular/tabular_model.py
shchur: The new mapping consistently leads to better MASE/wQL loss. Training with the RMSE/MSE objective estimates the mean, while MASE is minimized by the median (assuming the data is scaled), so MAE makes more sense here.

shchur:
@tonyhoo regarding the smoke test on M5: There are two models affected by the introduction of `past_covariates`: 
- `AutoGluonTabular`: The feature matrix construction times goes up from 15mins -> 23mins with this PR. This is already quite long and should definitely be improved in the future, but I couldn't find a simple way to optimize this. Parallelizing with Multiprocessing/Joblib leads to minor speedups (~5 minutes), but results in a slowdown for smaller datasets.
- `TemporalFusionTransformerMXNet`: This PR doesn't affect the runtime as the model already generates dummy `past_covariates` if none are present in the dataset.

The overall runtimes of these two models on M5 stay within 1-2 hours (`p3.2xlarge`):
- AutoGluonTabular
```
        -1.0496       = Validation score (-MASE)
        4356.52 s     = Training runtime
        1471.54 s     = Validation (prediction) runtime
```
- TemporalFusionTransformerMXNet
```
        -0.8972       = Validation score (-MASE)
        6523.87 s     = Training runtime
        196.98  s     = Validation (prediction) runtime
```

We will get significant speedups for TFT once we switch to the PyTorch version that comes out in the new GluonTS release in a few days.

tonyhoo (approved):

PR #2682, opened 2023-01-12
Support dump and load timm and hf_text models on MultiModalPredictor

suzhoum:
*Issue #, if available:*

*Description of changes:*
1. Added APIs to dump `timm_image` and `hf_text` models
2. Added support to load model from path for `timm_image`

Usage:

To dump models from fine-tuned MultiModalPredictor:
```
  MultiModalPredictor.dump_timm_image(PATH_TO_MODEL)
  MultiModalPredictor.dump_hf_text(PATH_TO_MODEL)
```

To load models from saved model path:
```
hyperparameters = {
     "model.hf_text.checkpoint_name": PATH_TO_SAVED_HF_TEXT_MODEL,
     "model.timm_image.checkpoint_name": PATH_TO_SAVED_TIMM_IMAGE_MODEL,
} 
```


multimodal/src/autogluon/multimodal/models/timm_image.py
```diff
@@ -48,13 +50,32 @@ def __init__(
         super().__init__()
         # In TIMM, if num_classes==0, then create_model would automatically set self.model.head = nn.Identity()
         logger.debug(f"initializing {checkpoint_name}")
-        self.checkpoint_name = checkpoint_name
+        if os.path.exists(checkpoint_name):
+            checkpoint_path = f"{checkpoint_name}/pytorch_model.bin"
+            try:
+                with open(f"{checkpoint_name}/config.json") as f:
+                    self.config = json.load(f)
```
zhiqiangdon, line 57: This `self.config` will be overwritten later in line 76?

multimodal/src/autogluon/multimodal/utils/model.py
```diff
@@ -523,3 +524,37 @@ def modify_duplicate_model_names(
...
+            continue
+        filtered_cfg[k] = v
+    return filtered_cfg
+
+
+def save_timm_config(
+    model: TimmAutoModelForImagePrediction,
+    config_path: str,
+):
+    hf_config = {}
```
zhiqiangdon, line 545: The name `hf_config` looks more like a config for huggingface models.

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -2812,6 +2817,79 @@ def load(
...
+            Path to directory where models and configs should be saved.
+        """
+        model = None
+        if isinstance(self._model, MultimodalFusionMLP) and isinstance(
+            self._model.model, torch.nn.modules.container.ModuleList
+        ):
+            for per_model in self._model.model:
+                if isinstance(per_model, TimmAutoModelForImagePrediction):
+                    model = per_model
+                    break
```
zhiqiangdon, line 2841: Saving a timm image backbone from a fusion model seems not useful since it can't work individually. We can consider only deal with single timm model for now.

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -2812,6 +2817,79 @@ def load(
...
+            Path to directory where models and configs should be saved.
+        """
+        model = None
+        if isinstance(self._model, MultimodalFusionMLP) and isinstance(
+            self._model.model, torch.nn.modules.container.ModuleList
+        ):
+            for per_model in self._model.model:
+                if isinstance(per_model, HFAutoModelForTextPrediction):
+                    model = per_model
+                    break
```
zhiqiangdon, line 2875: Ditto. We can consider only the single huggingface text model for now.

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -2812,6 +2817,79 @@ def load(
...
+
+        model = self._model if model is None else model
+        if isinstance(model, HFAutoModelForTextPrediction) and model.model is not None:
+            os.makedirs(path)
+            model.model.save_pretrained(path)
+            logger.info(f"Model saved to {path}.")
+            if TEXT in self._data_processors.keys():
```
zhiqiangdon, line 2882: For hf_text model, we can assert the data processors have one text processor.

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -2812,6 +2817,79 @@ def load(
 
         return predictor
 
+    def dump_timm_image(
```
zhiqiangdon, line 2820: There would be too many APIs if each model has a dump function. How about using only one API `dump_model()`? Inside the function, we can check the model type is timm image or hf text.

multimodal/src/autogluon/multimodal/utils/model.py
```diff
@@ -523,3 +524,37 @@ def modify_duplicate_model_names(
...
+    for k, v in cfg.items():
+        if remove_source and k in {"url", "file", "hf_hub_id", "hf_hub_id", "hf_hub_filename", "source"}:
+            continue
+        if remove_null and v is None and k not in keep_null:
+            continue
+        filtered_cfg[k] = v
+    return filtered_cfg
+
+
+def save_timm_config(
```
zhiqiangdon, line 541: docstrings.

multimodal/src/autogluon/multimodal/models/timm_image.py
suzhoum: Good catch, I believe the `self.config` on line 76 should be moved.

multimodal/src/autogluon/multimodal/utils/model.py
suzhoum: Updated the name.

multimodal/src/autogluon/multimodal/predictor.py
suzhoum: We considered the scenario for fusion model because there might be a use case that a user wants to save specifically timm_image or hf_text from the fusion model trained to use on other downstream tasks which only use a single model. Do you feel that would be useful?

multimodal/src/autogluon/multimodal/predictor.py
suzhoum: Do you mean there is no need to check the prefix of the data processor to get the tokenizer?

multimodal/src/autogluon/multimodal/predictor.py
sxjscience: For example, people trained on the mixture of image, text, tabular data, and like to extract the image part from the fusion model. The API “predictor.dump_timm_image” should still work in this case.

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -2812,6 +2817,79 @@ def load(
...
+            self._model.model, torch.nn.modules.container.ModuleList
+        ):
+            for per_model in self._model.model:
+                if isinstance(per_model, HFAutoModelForTextPrediction):
+                    model = per_model
+                    break
+
+        model = self._model if model is None else model
+        if isinstance(model, HFAutoModelForTextPrediction) and model.model is not None:
+            os.makedirs(path)
```
sxjscience, line 2879: Try to use “os.makedirs(path, exist_ok=True)”.

suzhoum: I originally had this flag set, but after discussing with Zhiqiang offline, he suggested that we might not want saving to an existing (or non-empty) directory to avoid potential override of the model.

multimodal/src/autogluon/multimodal/predictor.py
zhiqiangdon: This logic has the limitation that, If multiple timm image models are available, it would only dump the first one.

multimodal/src/autogluon/multimodal/predictor.py
zhiqiangdon: Since we consider dumping a hf model from a fusion model, this logic is OK.

zhiqiangdon: Do we need raise error if there is no hf model available?

multimodal/src/autogluon/multimodal/predictor.py
suzhoum: I've updated the logic to support saving multiple models under timm_image and hf_text.

multimodal/src/autogluon/multimodal/utils/model.py
```diff
@@ -523,3 +524,47 @@ def modify_duplicate_model_names(
 
 def list_timm_models(pretrained=True):
     return timm.list_models(pretrained=pretrained)
+
+
+def _filter_timm_pretrained_cfg(cfg, remove_source=False, remove_null=True):
```
zhiqiangdon, line 529: Since `_filter_timm_pretrained_cfg ` and `save_timm_config ` are about config, is it better to put them into `utils/config.py`?

multimodal/tests/unittests/predictor/test_predictor.py
```diff
@@ -679,3 +681,102 @@ def test_image_bytearray():
...
+        time_limit=5,
+        seed=42,
+    )
+    predictor.dump_hf_text(path=model_dump_path)
+    predictor.dump_timm_image(path=model_dump_path)
+    hf_text_dir = f"{model_dump_path}/hf_text"
+    timm_image_dir = f"{model_dump_path}/timm_image"
+    assert os.path.exists(hf_text_dir) and (len(os.listdir(hf_text_dir)) > 2) == True
+    assert os.path.exists(timm_image_dir) and (len(os.listdir(timm_image_dir)) == 2) == True
+    print("done")
```
zhiqiangdon, line 782: Remove `print`?

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -2824,6 +2829,86 @@ def load(
...
+        Save HuggingFace Text model weights, config and tokenizers to local directory.
+        Model weights are saved in file `pytorch_model.bin`;
+        Configs are saved in file `config.json`
+
+        Parameters
+        ----------
+        path : str
+            Path to directory where models and configs should be saved.
+        """
+        models = {}
```
zhiqiangdon, line 2883: `models` are list and dict in `dump_timm_image ` and `dump_hf_text `, respectively. Can we use list for both of them?

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -2824,6 +2829,86 @@ def load(
...
+        Model weights are saved in file `pytorch_model.bin`;
+        Configs are saved in file `config.json`
+
+        Parameters
+        ----------
+        path : str
+            Path to directory where models and configs should be saved.
+        """
+        models = []
+        if isinstance(self._model, MultimodalFusionMLP) and isinstance(
```
zhiqiangdon, line 2847: A fusion model may also be `MultimodalFusionTransformer`.

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -2824,6 +2829,86 @@ def load(
...
+        Model weights are saved in file `pytorch_model.bin`;
+        Configs are saved in file `config.json`
+
+        Parameters
+        ----------
+        path : str
+            Path to directory where models and configs should be saved.
+        """
+        models = []
+        if isinstance(self._model, (MultimodalFusionMLP, MultimodalFusionTransformer)) and isinstance(
```
sxjscience, line 2847: @zhiqiangdon We can later add a `BaseMultimodalFusionModel` class, and ensure that `MultimodalFusionMLP` and `MultimodalFusionTransformer` inherit from this class.

sxjscience: @suzhoum May be we can add a TODO item here

multimodal/tests/unittests/predictor/test_predictor.py
```diff
@@ -679,3 +681,104 @@ def test_image_bytearray():
     npt.assert_array_equal(
         [prediction_prob_1, prediction_prob_2, prediction_prob_3, prediction_prob_4], [prediction_prob_1] * 4
     )
+
+
+def test_dump_timm_image():
```
sxjscience, line 686: Can we separate these tests to another file? The reason is that `test_predictor.py` is growing to be too huge and too slow. 

We can consider to add it under the following name

```shell
unittests/predictor/test_predictor_dump_third_party.py
```

sxjscience (approved):
Two minor comments!! Overall LGTM!!!! Thanks!!!

multimodal/tests/unittests/predictor/test_predictor.py
suzhoum: Good idea. Done!

multimodal/src/autogluon/multimodal/predictor.py
zhiqiangdon: Sure. We can add a base class.

PR #2691, opened 2023-01-13
Remove dask and distributed dependencies

Innixma:
*Issue #, if available:*

*Description of changes:*
- Remove dask and distributed dependencies
- This is to check if they are actually needed

### DIFF
--- a/multimodal/setup.py
+++ b/multimodal/setup.py
@@ -54,6 +54,7 @@ install_requires = [
     "openmim>0.1.5,<=0.2.1",
     "defusedxml>=0.7.1,<=0.7.1",
     "albumentations>=1.1.0,<=1.2.0",
+    "jinja2>=3.0.3,<3.2",
     "tensorboard",
 ]
 
--- a/multimodal/setup.py
+++ b/multimodal/setup.py
@@ -54,7 +54,7 @@ install_requires = [
     "openmim>0.1.5,<=0.2.1",
     "defusedxml>=0.7.1,<=0.7.1",
     "albumentations>=1.1.0,<=1.2.0",
-    "jinja2>=3.0.3,<3.2",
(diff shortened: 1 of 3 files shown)


multimodal/setup.py
```diff
@@ -55,6 +55,7 @@
     "openmim>0.1.5,<=0.2.1",
     "defusedxml>=0.7.1,<=0.7.1",
     "albumentations>=1.1.0,<=1.2.0",
+    "jinja2>=3.0.3,<3.2",
```
zhiqiangdon, line 58: Where do we use `jinja2`?

Innixma: https://github.com/autogluon/autogluon/blob/master/multimodal/src/autogluon/multimodal/data/templates.py#L18

Innixma: You can refer to the CI error prior to me adding it in this PR: https://github.com/autogluon/autogluon/actions/runs/3914783191/jobs/6692268378#step:4:994

sxjscience: This is for supporting prompt source. I’m okay with removing it for now.

sxjscience (approved):
LGTM

multimodal/setup.py
Innixma: @sxjscience if you'd like to remove jinja usage, you'll need to open a separate follow-up PR. I believe jinja was a dependency of dask/distributed, which is why we didn't error in the past, but with dask/distributed removed, now we need to explicitly add it as a dependency to AutoMM

tensor-alchemist:
@Innixma Is dask and ray support completely removed. I am working with more than 10 M unique time series and I have a memory constraints. What is the best way I can achieve distributed learning with AutoGluon?

PR #2700, opened 2023-01-16
[object detection] Model Save/Load and Objects365 Pretraining

FANGAreNotGnu:
- Add save/load logic for mmdet's ckpt and config
- Refactor mmdet_image
- Add yolox_l_objects365
- Add a test for new save/load logic


multimodal/src/autogluon/multimodal/models/mmdet_image.py
```diff
@@ -41,7 +42,7 @@ def __init__(
         self,
         prefix: str,
         checkpoint_name: str,
-        num_classes: Optional[int] = None,
+        config_file: Optional[str] = None,
```
suzhoum, line 45: For custom models, is it possible to not use `config_file` when loading, since it will be an extra hyperparameter that the user needs to add? For `timm_image` and `hf_text`, currently, we can reuse the `checkpoint_name` parameter and substitute name with checkpoint directory (which contains both weights and config).

examples/automm/object_detection/finetune_coco_format.py
```diff
@@ -139,6 +139,45 @@ def tutorial_script_for_finetune_yolox_pothole_in_coco_format():
     predictor.evaluate(test_path)
 
 
+def tutorial_script_for_finetune_yoloxo365_pothole_in_coco_format():
```
suzhoum, line 142: Just curious, where is this function being used?

multimodal/src/autogluon/multimodal/models/mmdet_image.py
FANGAreNotGnu: Sure, I changed the saved checkpoint name and config name to "model.pth" and "config.py". And if input `checkpoint_name` is a dir, it should include both checkpoint ("model.pth") and config ("config.py"). But I leave the `config_file` here in case we will use a checkpoint with different configs (e.g. yolox with mosaic v.s. yolox without mosaic).

multimodal/src/autogluon/multimodal/models/mmdet_image.py
```diff
@@ -54,119 +55,172 @@ def __init__(
...
-
-        self.prefix = prefix
-
         self.name_to_id = self.get_layer_ids()
         self.head_layer_names = [n for n, layer_id in self.name_to_id.items() if layer_id == 0]
 
-    def _load_checkpoint_and_config(self, checkpoint_name=None):
+    def save_weights_and_configs(self, save_dir: str = None, save_name: str = None):
+        if not save_dir:
+            save_dir = "./"
```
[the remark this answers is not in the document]
FANGAreNotGnu: Done

suzhoum (approved):
LGTM! Thanks!

sxjscience (approved):
LGTM

multimodal/src/autogluon/multimodal/models/mmdet_image.py
```diff
@@ -54,119 +55,171 @@ def __init__(
...
+        self._save_weights(save_path=weights_save_path)
+        self._save_configs(save_path=configs_save_path)
+
+        return save_path
+
+    def _save_weights(self, save_path=None):
+        if not save_path:
+            save_path = f"./{self.checkpoint_name}_autogluon.pth"
+
+        torch.save({"state_dict": self.model.state_dict(), "meta": {"CLASSES": self.model.CLASSES}}, save_path)
```
zhiqiangdon, line 137: Is it better to put `CLASSES ` in the config file? The hf_text and timm_image checkpoints contain only the model weights.

multimodal/src/autogluon/multimodal/utils/model.py
```diff
@@ -286,7 +286,7 @@ def create_model(
         model = MMDetAutoModelForObjectDetection(
             prefix=model_name,
             checkpoint_name=model_config.checkpoint_name,
-            num_classes=num_classes,
```
zhiqiangdon, line 289: How do users customize `num_classes ` if we remove this?

multimodal/src/autogluon/multimodal/models/mmdet_image.py
FANGAreNotGnu: We sometimes still load from MMDetection's model zoo and the `CLASSES` are in its checkpoints. To have a consistent loading operation, we'd better keep `CLASSES` in checkpoints.

zhiqiangdon (approved):
LGTM

PR #2701, opened 2023-01-16
[AutoMM] Refactor inferring image types and fix using invalid images

zhiqiangdon:
*Issue #, if available:*

1. `is_imagepath_column` and `is_imagebytearray_column`  has much shared code.
2. If there are invalid image paths in dataframe, error may happen when using the validation data, which is due to that we only used one sample in inferring the image type.

*Description of changes:*

1. Refactor `is_imagepath_column` and `is_imagebytearray_column` into `is_image_column`.
2. Remove the `sample_n=1` when inferring image types from validation data.
3. Add invalid image tests.


sxjscience (approved):
LGTM

Issue #2697, opened 2023-01-14
slice_by_timestep error[BUG]


amir78pgd:
- [ ] I have checked that this bug exists on the latest stable version of AutoGluon
- [ ] and/or I have checked that this bug exists on the latest mainline of AutoGluon via source installation

**Describe the bug**
slice_by_timestep method of TimeSeriesPredictor fails with the following error

TypeError: n needs to be an int or a list/set/tuple of ints
**Expected behavior**
This was working fine in autogluon version 0.6.1

**To Reproduce**
A minimal script to reproduce the issue. Links to Colab notebooks or similar tools are encouraged.  
If the code is too long, feel free to put it in a public gist and link it in the issue: https://gist.github.com.  
In short, we are going to copy-paste your code to run it and we expect to get the same result as you.

**Screenshots**

**Installed Versions**
Which version of AutoGluon are you are using?  
If you are using 0.4.0 and newer, please run the following code snippet:
<details>

```python
# Replace this code with the output of the following:
from autogluon.core.utils import show_versions
show_versions()
```

</details>

**Additional context**

iamgyang:
Hello, I am getting the same bug. Thanks for pointing it out.

shchur:
To fix this, you need to install a more recent version of pandas >=1.4 (see #2689).

Sorry about this, we will fix the requirements in the next release.

shchur:
This is now fixed with #2703

Issue #2689, opened 2023-01-13
n needs to be an int or a list/set/tuple of ints (error)


nebiyebulan:
i got this error when i want to run my code in collab to make it run faster.
`train_data = TimeSeriesDataFrame.from_data_frame(
            df.reset_index(),
            id_column='Brand',
            timestamp_column='Date'
        )
test_data = train_data.copy()
train_data = train_data.slice_by_timestep(slice(None, -prediction_length))
train_data.head()`

<img width="609" alt="autogluon_error" src="https://user-images.githubusercontent.com/62153842/212332704-13847bf8-2eca-4413-9767-260c8f6ce523.png">
<img width="204" alt="Annotation 2023-01-13 163929" src="https://user-images.githubusercontent.com/62153842/212333195-753951b6-32aa-4660-8542-79a71a2e55e5.png">

nebiyebulan:
ıt's worked.Thank you.

PR #2703, opened 2023-01-16
[timeseries] Bump pandas dependency to >=1.4.0

shchur:
*Issue #, if available:* #2689, #2697

*Description of changes:*
- Ensure that the pandas version is >= 1.4.0 for [`GroupBy.nth`](https://pandas.pydata.org/docs/reference/api/pandas.core.groupby.GroupBy.nth.html) to work with `slice` objects. This is required for https://github.com/autogluon/autogluon/blob/master/timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py#L572

### DIFF
--- a/timeseries/setup.py
+++ b/timeseries/setup.py
@@ -26,7 +26,7 @@ install_requires = [
     "joblib~=1.1",
     "numpy",
     "scipy",
-    "pandas",
+    "pandas>=1.4.0",
     "statsmodels~=0.13.0",
     "gluonts~=0.11.0",
     "torch>=1.9,<1.13",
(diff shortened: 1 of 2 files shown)


canerturkmen:
Let's also wait for @Innixma

canerturkmen (approved):

timeseries/setup.py
```diff
@@ -26,7 +26,7 @@
     "joblib~=1.1",
     "numpy",
     "scipy",
-    "pandas",
+    "pandas>=1.4.0",
```
Innixma, line 29: Please consider updating `core/_setup_utils.py` instead.

```
'pandas': '>=1.2.5,!=1.4.0,<1.6',
```
is the current range.

Change to:

```
'pandas': '>1.4.0,<1.6',
```

Innixma: once this is done and the current LoC change is reverted, I approve, feel free to merge without syncing with me

PR #2708, opened 2023-01-17
Lower-bound Pillow

sxjscience:
*Issue #, if available:*

https://github.com/autogluon/autogluon/issues/2707

*Description of changes:*
The PIL.Image.Resampling was added in Pillow 9.1.0. So added the lower-bound. Reference: https://stackoverflow.com/questions/71738218/module-pil-has-not-attribute-resampling

### DIFF
--- a/multimodal/setup.py
+++ b/multimodal/setup.py
@@ -24,7 +24,7 @@ install_requires = [
     "scipy",
     "pandas",
     "scikit-learn",
-    "Pillow",
+    "Pillow>=9.1.0",
     "tqdm",
     "boto3",
     "requests",


multimodal/setup.py
```diff
@@ -24,7 +24,7 @@
     "scipy",
     "pandas",
     "scikit-learn",
-    "Pillow",
+    "Pillow>=9.1.0",
```
Innixma, line 27: This will simply override the `_setup_utils` range: https://github.com/autogluon/autogluon/blob/master/core/src/autogluon/core/_setup_utils.py#L27

We should have a consistent range. I believe lower bound 9.3.0 was set due to a security issue with earlier versions.

Innixma (changes requested):
We should figure out why it crashes instead of downgrading the version range.

multimodal/setup.py
sxjscience: Okay, there is no bug then. The Colab issue can be resolved by restarting the runtime.

PR #2716, opened 2023-01-17
Cleanup lite flag

Innixma:
*Issue #, if available:*

*Description of changes:*
- Clean up lite flag code
- Now we don't repeatedly call `get_autogluon_metadata` during autogluon import just to get the lite status
- Now developers don't need to re-run setup when pulling the lite commit to avoid crashing. We now wrap the `__lite__` import in a try/except.

### DIFF
--- a/common/src/autogluon/common/utils/utils.py
+++ b/common/src/autogluon/common/utils/utils.py
@@ -7,7 +7,15 @@ import platform
 import sys
 from typing import Dict, Any
 
+from ..version import __version__
+try:
+    from ..version import __lite__
+except ImportError:
+    __lite__ = False
+
 logger = logging.getLogger(__name__)
+
+LITE_MODE: bool = __lite__ is not None and __lite__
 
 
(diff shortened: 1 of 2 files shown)


common/src/autogluon/common/utils/utils.py
```diff
@@ -7,8 +7,16 @@
 import sys
 from typing import Dict, Any
 
+from ..version import __version__
```
gradientsky, line 10: This might create import issues with IDEs since the file is auto-generated. Should we do try-except with this import too?

Innixma: We already input `__version__` in the root init file of each module, so it won't make a difference.

Ex: `common/__init__.py`:

```python3
from autogluon.common.utils.log_utils import _add_stream_handler 

from .dataset import TabularDataset
from .space import Space, Categorical, Real, Int, Bool
from . import metrics
from . import constants


_add_stream_handler()

```

yzhliu (approved):
overall good to me.

yinweisu (approved):
LGTM!

PR #2719, opened 2023-01-18
[eda] Updates for quick-fit to include samples display

gradientsky:
*Description of changes:*

### Minor components changes
- Added `PropertyRendererComponent` to enable simplified properties rendering stored in `state`:
  - The following component apply `transform_fn` to an object stored in `state.model_evaluation.highest_error` then render it (if the object present):
```python
PropertyRendererComponent("model_evaluation.highest_error", transform_fn=(lambda df: df.head(10))),
```
- `MarkdownSectionComponent` now supports `condition_fn` argument to enable conditional rendering:
```python
MarkdownSectionComponent(
    condition_fn=(lambda state: is_key_present_in_state(state, "model_evaluation.undecided")),
    markdown="Text to render",
),
```

### Updates for `quick-fit()`
- samples with the highest prediction error - candidates for inspection
- smaples with the least distance from the other class - candidates for labeling

image

### DIFF
--- a/eda/src/autogluon/eda/analysis/model.py
+++ b/eda/src/autogluon/eda/analysis/model.py
@@ -208,6 +208,8 @@ class AutoGluonModelEvaluator(AbstractAnalysis):
         y_true = val_data[label]
         y_pred = predictor.predict(val_data)
 
+        highest_error = None
+        undecided = None
         if predictor.problem_type in [BINARY, MULTICLASS]:
             y_proba = predictor.predict_proba(val_data)
             highest_error = y_proba[y_true != y_pred].max(axis=1)
@@ -230,23 +232,23 @@ class AutoGluonModelEvaluator(AbstractAnalysis):
                 .join(highest_error, how="inner")
                 .sort_values(by="error", ascending=False)
             )
-            undecided = None
 
         importance = predictor.feature_importance(val_data.reset_index(drop=True), silent=True)
         leaderboard = predictor.leaderboard(val_data, silent=True)
 
-        labels = y_true.unique()
+        labels = predictor.class_labels
         s = {
             "problem_type": predictor.problem_type,
             "y_true": y_true,
             "y_pred": y_pred,
-            "highest_error": highest_error,
             "importance": importance,
             "leaderboard": leaderboard,
             "labels": labels,
         }
         if undecided is not None:
             s["undecided"] = undecided
+        if highest_error is not None:
+            s["highest_error"] = highest_error
 
         if problem_type in [BINARY, MULTICLASS]:
             cm = confusion_matrix(y_true, y_pred, normalize=self.normalize, labels=labels)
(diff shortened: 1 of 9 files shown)


eda/src/autogluon/eda/visualization/layouts.py
```diff
@@ -78,14 +86,70 @@ class MarkdownSectionComponent(AbstractVisualization, JupyterMixin):
...
+        self.property = property
+        self.transform_fn = transform_fn
+
+    def can_handle(self, state: AnalysisState) -> bool:
+        return is_key_present_in_state(state, self.property)
+
+    def _render(self, state: AnalysisState) -> None:
+        obj = state
+        for p in self.property.split("."):
+            obj = obj[p]
```
gidler, line 152: Is there a risk that properties could have the same name and so this overwrites earlier properties?

eda/src/autogluon/eda/visualization/layouts.py
```diff
@@ -7,7 +7,15 @@
 from .base import AbstractVisualization
 from .jupyter import JupyterMixin
 
-__all__ = ["MarkdownSectionComponent", "SimpleVerticalLinearLayout", "SimpleHorizontalLayout", "TabLayout"]
+__all__ = [
+    "MarkdownSectionComponent",
+    "SimpleVerticalLinearLayout",
+    "SimpleHorizontalLayout",
+    "TabLayout",
+    "PropertyRendererComponent",
```
gidler, line 15: If you do a new revision, you could make these alphabetical

eda/src/autogluon/eda/analysis/model.py
```diff
@@ -199,6 +207,31 @@ def _fit(self, state: AnalysisState, args: AnalysisState, **fit_kwargs):
...
+        elif problem_type == REGRESSION:
+            highest_error = np.abs(y_pred - y_true).sort_values(ascending=False)
+            highest_error.name = "error"
+            highest_error = (
+                val_data.join(y_pred, rsuffix="_pred")
+                .join(highest_error, how="inner")
+                .sort_values(by="error", ascending=False)
+            )
+            undecided = None
+
```
gidler, line 234: Should we have an `else` here for if problem types isn't `BINARY, MULTICLASS, or REGRESSION`. Could it be softclass or quantile? If so, should we handle that or throw?

eda/src/autogluon/eda/visualization/layouts.py
gradientsky: You can think about state as a nested maps; so no duplicate keys; writing into existing key will overwrite earlier version.

eda/src/autogluon/eda/visualization/layouts.py
gradientsky: I'm relying on `isort` and `black` for formatting. If it is passing, I assume it's PEP-compliant.

gidler (approved):

PR #2723, opened 2023-01-18
Refactor evaluate_coco out of predictor

zhiqiangdon:
*Issue #, if available:*

*Description of changes:*
Since `evaluate_coco` is only for object detection, we move it into `utils/object_detection`.

### DIFF
--- a/multimodal/src/autogluon/multimodal/predictor.py
+++ b/multimodal/src/autogluon/multimodal/predictor.py
@@ -121,13 +121,12 @@ from .utils import (
     assign_feature_column_names,
     average_checkpoints,
     check_if_packages_installed,
-    cocoeval,
-    compute_inference_batch_size,
     compute_num_gpus,
     compute_score,
     create_fusion_data_processors,
     create_fusion_model,
     data_to_df,
+    evaluate_coco,
     extract_from_output,
     filter_search_space,
     from_coco_or_voc,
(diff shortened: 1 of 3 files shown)


multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -1927,8 +1859,13 @@ def evaluate(
...
                 )
-            return self.evaluate_coco(
-                anno_file_or_df=data, metrics=metrics, return_pred=return_pred, seed=seed, eval_tool=eval_tool
+            return evaluate_coco(
+                predictor=self,
+                anno_file_or_df=data,
+                metrics=metrics,
+                return_pred=return_pred,
+                seed=seed,
+                eval_tool=eval_tool,
```
Innixma, line 1868: Bigger refactoring related question to think about: Should we have separate predictor classes?

```
AbstractMultimodalPredictor
MultimodalPredictor
MulitmodalDetector
MultimodalMatcher
```

This would reduce the if/elif/elif/else logic being repeated in the code.

FANGAreNotGnu:
LGTM

FANGAreNotGnu (approved):
LGTM

multimodal/src/autogluon/multimodal/predictor.py
sxjscience: Yeah, this is actually work-in-progress. We may need to come up with better names. The user will still access all the features via `MultimodalPredictor(problem_type=xxx)`.

sxjscience (approved):
LGTM

PR #2724, opened 2023-01-18
[CI] Add branch check

yinweisu:
*Issue #, if available:*

*Description of changes:*
* Tested [here](https://github.com/yinweisu/hello-github-actions/actions/runs/3952970490/jobs/6768697769)

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -13,7 +13,18 @@ defaults:
     shell: bash
 
 jobs:
+  branch_check:
+    runs-on: ubuntu-latest
+    if: ${{ github.event_name == 'push' }}
+    steps:
+      - name: Fail on restricted branch
+        # https://docs.github.com/en/actions/learn-github-actions/expressions#example-matching-an-array-of-strings
+        if: contains(fromJSON('["cloud"]'), github.ref_name)
+        run: |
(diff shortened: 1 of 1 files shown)


Innixma (approved):
LGTM!

gidler (approved):

.github/workflows/continuous_integration.yml
```diff
@@ -13,7 +13,18 @@ defaults:
     shell: bash
 
 jobs:
+  branch_check:
+    runs-on: ubuntu-latest
+    if: ${{ github.event_name == 'push' }}
+    steps:
+      - name: Fail on restricted branch
+        # https://docs.github.com/en/actions/learn-github-actions/expressions#example-matching-an-array-of-strings
+        if: contains(fromJSON('["cloud"]'), github.ref_name)
```
tonyhoo, line 22: Just to be clear, this will block any branch name containing `cloud`

yinweisu: I don't think so. According to the [documentation](https://docs.github.com/en/actions/learn-github-actions/expressions#contains), if the search is an array, it will only return true if item is inside search. If search is a string, it returns true if item is a substring

tonyhoo: Thanks for clarification

tonyhoo (approved):
LGTM! Thanks for taking this change

PR #2727, opened 2023-01-19
Add predictor.predict_multi and predictor.predict_proba_multi

Innixma:
*Issue #, if available:*

*Description of changes:*
- Add predictor.predict_dict and predictor.predict_proba_dict
- This is an optimized way to get predictions for a list of models, faster than looping through predictor.predict or predictor.predict_proba
- Also serves as a clean way to vend per-model out-of-fold predictions, validation predictions, and test predictions for the purposes of more advanced logic (Meta-learning / Zero-shot HPO)
- Fixed inconsistency with predictions, now will round pred_proba ties to 0 instead of 1 always when converting proba to pred, previously if object was pandas DF we would round to 0 and if it was a pandas Series we would round to 1.
- Fixed bug in save_pkl where it would crash if saving file to working directory root.

### DIFF
--- a/core/src/autogluon/core/trainer/abstract_trainer.py
+++ b/core/src/autogluon/core/trainer/abstract_trainer.py
@@ -873,6 +873,54 @@ class AbstractTrainer:
         else:
             return model_pred_proba_dict
 
+    def get_model_pred_dict(self,
+                            X: pd.DataFrame,
+                            models: List[str],
+                            record_pred_time: bool = False,
+                            **kwargs):
+        """
+        Optimally computes predictions for each model in `models`.
+        Will compute each necessary model only once and store predictions in a `model_pred_dict` dictionary.
+        Note: Mutates model_pred_proba_dict and model_pred_time_dict input if present to minimize memory usage.
+
+        Acts as a wrapper to `self.get_model_pred_proba_dict`, converting the output to predictions.
+
+        Parameters
+        ----------
+        X : pd.DataFrame
+            Input data to predict on.
+        models : List[str]
+            The list of models to predict with.
+            Note that if models have dependency models, their dependencies will also be predicted with and included in the output.
+        record_pred_time : bool, default = False
+            Whether to store marginal inference times of each model as an extra output `model_pred_time_dict`.
+        **kwargs : dict, optional
+            Refer to `self.get_model_pred_proba_dict` for documentation of remaining arguments.
+            This method shares identical arguments.
+
+        Returns
+        -------
+        If `record_pred_time==True`, outputs tuple of dicts (model_pred_dict, model_pred_time_dict), else output only model_pred_dict
+        """
+        model_pred_proba_dict = self.get_model_pred_proba_dict(X=X,
+                                                               models=models,
+                                                               record_pred_time=record_pred_time,
+                                                               **kwargs)
+        if record_pred_time:
+            model_pred_proba_dict, model_pred_time_dict = model_pred_proba_dict
+        else:
+            model_pred_time_dict = None
+
+        model_pred_dict = {}
+        for m in model_pred_proba_dict:
+            # Convert pred_proba to pred
+            model_pred_dict[m] = get_pred_from_proba(y_pred_proba=model_pred_proba_dict[m], problem_type=self.problem_type)
+
+        if record_pred_time:
+            return model_pred_dict, model_pred_time_dict
+        else:
+            return model_pred_dict
+
     def get_model_oof(self, model: str) -> np.ndarray:
         """Gets the out of fold prediction probabilities for a bagged ensemble model"""
         model_type = self.get_model_attribute(model=model, attribute='type')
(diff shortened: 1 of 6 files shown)


sxjscience:
Shall we just extend `.predict()` with more flags?

Innixma:
> Shall we just extend `.predict()` with more flags?

That is possible, but it would add a significant amount of complexity to the doc of `.predict()`. I think I'd rather keep it separate.

Innixma:
Another thought: A different name could be `predict_multi` instead of `predict_dict`, but unsure what the best name would be

core/src/autogluon/core/utils/utils.py
```diff
@@ -260,7 +260,10 @@ def get_pred_from_proba_df(y_pred_proba, problem_type=BINARY):
 
 def get_pred_from_proba(y_pred_proba, problem_type=BINARY):
     if problem_type == BINARY:
-        y_pred = [1 if pred >= 0.5 else 0 for pred in y_pred_proba]
+        # Using > instead of >= to align with Pandas `.idxmax` logic which picks the left-most column during ties.
+        # If this is not done, then predictions can be inconsistent when converting in binary classification from multiclass-form pred_proba and
+        # binary-form pred_proba when the pred_proba is 0.5 for positive and negative classes.
+        y_pred = [1 if pred > 0.5 else 0 for pred in y_pred_proba]
```
gradientsky, line 266: Shall we move `0.5` into a property or method method call? Rationale: at some point we'll add optimization of the threshold so having either method or named property would be a natural place to change instead of looking for magic value here.

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -1491,7 +1491,7 @@ def evaluate_predictions(self, y_true, y_pred, sample_weight=None, silent=False,
         return self._learner.evaluate_predictions(y_true=y_true, y_pred=y_pred, sample_weight=sample_weight, silent=silent,
                                                   auxiliary_metrics=auxiliary_metrics, detailed_report=detailed_report)
 
-    def leaderboard(self, data=None, extra_info=False, extra_metrics=None, only_pareto_frontier=False, skip_score=False, silent=False):
+    def leaderboard(self, data=None, extra_info=False, extra_metrics=None, only_pareto_frontier=False, skip_score=False, silent=False) -> pd.DataFrame:
```
gradientsky, line 1494: nit: add parameter types too

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -1619,6 +1619,107 @@ def leaderboard(self, data=None, extra_info=False, extra_metrics=None, only_pare
...
+        Dictionary with model names as keys and model prediction probabilities as values.
+        """
+        return self._learner.predict_proba_dict(X=data,
+                                                models=models,
+                                                as_pandas=as_pandas,
+                                                as_multiclass=as_multiclass,
+                                                transform_features=transform_features,
+                                                inverse_transform=inverse_transform)
+
+    def predict_dict(self, data=None, models=None, as_pandas=True, transform_features=True, inverse_transform=True) -> dict:
```
gradientsky, line 1676: nit: `... -> Dict[str, DataFrame]` + add parameter types

core/src/autogluon/core/trainer/abstract_trainer.py
```diff
@@ -871,6 +871,54 @@ def get_model_pred_proba_dict(self,
         else:
             return model_pred_proba_dict
 
+    def get_model_pred_dict(self,
+                            X: pd.DataFrame,
+                            models: List[str],
+                            record_pred_time: bool = False,
+                            **kwargs):
```
gradientsky, line 878: add return type hint

core/src/autogluon/core/utils/utils.py
Innixma: Yes, but this should be a separate PR as it is a major change. I've had this in mind for awhile

core/src/autogluon/core/trainer/abstract_trainer.py
Innixma: will wait for future PR, the return type is not trivial as it can be a tuple

tabular/src/autogluon/tabular/predictor/predictor.py
Innixma: added, note type hint for data will wait for future PR due to multiple types support

tabular/src/autogluon/tabular/predictor/predictor.py
Innixma: added, note type hint for data and output will wait for future PR due to multiple types support

Innixma:
Note: renamed to `predict_multi` and `predict_proba_multi` to better communicate the intended usage.

gradientsky (approved):

PR #2728, opened 2023-01-19
More split of multimodal tests

yinweisu:
*Issue #, if available:*

*Description of changes:*
* Tests split according to run time


multimodal/tests/unittests/others_2/test_hpo.py
```diff
@@ -8,9 +8,9 @@
 from autogluon.multimodal import MultiModalPredictor
 
 from ..predictor.test_predictor import verify_predictor_save_load
-from .test_matcher import verify_matcher_save_load
-from .unittest_datasets import IDChangeDetectionDataset, PetFinderDataset
-from .utils import get_home_dir
+from ..others.test_matcher import verify_matcher_save_load
```
Innixma, line 10: are we sure this is a good practice? Why not have utils in a separate folder altogether from `other` `other_2` etc.?

I'd imagine ideally tests shouldn't be importing other tests.

Innixma (approved):
Approving to accelerate debugging

multimodal/tests/unittests/others_2/test_hpo.py
yinweisu: Yea, this probably should live in some util files or being a function in conftest @zhiqiangdon .
Though I think it's better to address it in a separate PR.

With that being said, I should move utils and unittest_datasets into a util folder indeed

zhiqiangdon: Good point. I can refactor this later.

zhiqiangdon (approved):
LGTM

sxjscience:
@zhiqiangdon We can consider to submit a follow-up PR to restructure the test-cases based on their functionality.

yinweisu:
Doc building failure looks unrelated and the unit tests has passed. Will merge

Issue #2695, opened 2023-01-13
[v0.7] Remove autogluon.text


Innixma:
Remove autogluon.text in v0.7, as autogluon.mulitmodal fully replaces it.

Innixma:
Resolved in #2847

PR #2730, opened 2023-01-19
Remove usage of autogluon.text in autogluon.tabular

Innixma:
*Issue #, if available:*
#2695 

*Description of changes:*
- Remove usage of autogluon.text in autogluon.tabular
- Instead directly use MultiModalPredictor
- Identical functionality since TextPredictor simply wrapped MultiModalPredictor


sxjscience (approved):
LGTM!

Issue #2694, opened 2023-01-13
[v0.7] Remove autogluon.vision


Innixma:
Remove autogluon.vision module in v0.7, as it is fully replaced by autogluon.multimodal.

Additionally:

- [x] Remove gluoncv dependency. It is only used in `autogluon.vision`. This should also remove `autocfg` dependency.
- [x] Update tutorials to indicate the change

bryanyzhu:
PR 2790 and 2819 have been merged, the tasks in this issue are completed.

Innixma:
Awesome! Marking as resolved.

PR #2731, opened 2023-01-19
Remove usage of autogluon.vision in autogluon.tabular

Innixma:
*Issue #, if available:*
#2694 

*Description of changes:*
- Remove usage of autogluon.vision in autogluon.tabular, switch to using autogluon.multimodal
- Not identical from prior functionality since previously Tabular was using the gluoncv backend of autogluon.vision.
- Note that the special null image handling logic has been preserved, can be revisited if it makes sense to remove.


yinweisu (approved):
LGTM!

PR #2740, opened 2023-01-20
[timeseries] Obtain quantiles for the tabular model using the empirical distribution

shchur:
*Description of changes:*
- Obtain the quantiles using the empirical distribution of the residuals (instead of fitting a 0-mean normal distribution to the residuals)


canerturkmen (approved):
LGTM!

PR #2742, opened 2023-01-20
[Tabular] Fix ImagePredictor

yinweisu:
*Issue #, if available:*
ImagePredictor was not properly trained in multimodal preset after the transition to MultiModalPredictor.
This issue wasn't caught by our unit test because it doesn't crash AG training. An example for the failure could be found in our [documentation site](http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-2731/579b5dd/tutorials/tabular_prediction/tabular-multimodal.html) 

*Description of changes:*
* Filtered the dataframe to contain only image features prior to call `_fit`


Innixma:
Updated the PR to fix the original reason for the bug.

Also added specific feature type passing from tabular to AutoMM instead of having AutoMM infer feature types.

Innixma (approved):

PR #2745, opened 2023-01-23
Add model failed to train check

yinweisu:
*Issue #, if available:*
Currently unit test will not fail as long as there's at least one model being trained. We would like to capture any failures in our unit test instead

*Description of changes:*
* Added a variable recording failed models and check it during unit tests


Innixma (approved):
LGTM, nice add!

PR #2749, opened 2023-01-24
[AutoMM] Refactor presets

zhiqiangdon:
*Issue #, if available:*

1. Many redundant hyperparameters exist in presets, increasing the maintainance difficulty.
3. It's difficult to add HPO presets based on the current preset design.

*Description of changes:*

1. Reorganize the presets by problem type. No need to repeat the shared hyperparameters. The differences among presets of the same problem type are more clear. 
2. Use `best_quality`, `high_quality`, and `medium_quality` as presets.


multimodal/src/autogluon/multimodal/presets.py
```diff
@@ -1,114 +1,83 @@
...
+                {
+                    "model.hf_text.checkpoint_name": "google/electra-base-discriminator",
+                    "model.timm_image.checkpoint_name": "swin_base_patch4_window7_224",
+                }
+            )
+    elif presets == MEDIUM_QUALITY_FASTER_INFERENCE:
+        hyperparameters.update(
+            {
+                "model.hf_text.checkpoint_name": "google/electra-small-discriminator",
+                "model.timm_image.checkpoint_name": "swin_small_patch4_window7_224",
```
bryanyzhu, line 50: For the quick build setting of image classification, we recommend to use `mobilenetv3_large_100 ` before, we might want to stay consistent.

bryanyzhu (commented):
LGTM with a comment.

multimodal/src/autogluon/multimodal/presets.py
```diff
@@ -130,375 +99,288 @@ def few_shot_text_classification():
...
-        "model.hf_text.checkpoint_name": "cross-encoder/ms-marco-MiniLM-L-12-v2",
-        "env.eval_batch_size_ratio": 1,
-    }
+    return hyperparameters, hyperparameter_tune_kwargs
 
 
 @automm_presets.register()
-def zero_shot_image_classification():
-    return {
+def zero_shot_image_classification(presets: str = DEFAULT, hpo: bool = False):
```
sxjscience, line 108: Will this be too complicated? Can we just register presets a few strings?

The current design is that you need to provide a `presets` flag in `automm_presets.register()`.

sxjscience: Also, consider to add docstrings to explain the design. For example, I can understand the current presets are registered per problem_type. 

For example, for `image_classification`, we can have its own presets, like `DEFAULT`, `BEST_QUALITY`, etc.

multimodal/src/autogluon/multimodal/presets.py
```diff
@@ -1,114 +1,83 @@
...
 from .registry import Registry
 
 automm_presets = Registry("automm_presets")
 matcher_presets = Registry("matcher_presets")
 
 
 @automm_presets.register()
-def high_quality_fast_inference():
-    return {
+def default(presets: str = DEFAULT, hpo: bool = False):
```
sxjscience, line 23: Why do we need `hpo` flag?

zhiqiangdon: This is to support the HPO presets, i.e., returning not only hyperparameters but also hyperparameter_tune_kwargs.

multimodal/src/autogluon/multimodal/presets.py
zhiqiangdon: The string design is not easy to maintain and scale to more problem types and HPO presets. There are much redundancy in the strings (one problem type name may repeat three times) and hyperparameters (different presets of the same problem type may share common ones).

multimodal/src/autogluon/multimodal/presets.py
zhiqiangdon: Changed to `mobilenetv3_large_100`.

multimodal/src/autogluon/multimodal/presets.py
zhiqiangdon: Docstrings added.

FANGAreNotGnu (commented):
LGTM, will update detection's preset by 0.7 release.

multimodal/src/autogluon/multimodal/matcher.py
```diff
@@ -614,39 +619,33 @@ def _fit(
...
+        config = get_config(
+            problem_type=self._pipeline,
+            presets=presets,
+            config=config,
+            overrides=hyperparameters,
+            extra=["matcher"],
+        )
+
+        if self._query_config is None:
+            query_config = copy.deepcopy(config)
```
cheungdaven, line 632: avoid deepcopy? https://github.com/autogluon/autogluon/pull/2729

cheungdaven (commented):
LGTM! Thanks for the refactor.

multimodal/src/autogluon/multimodal/matcher.py
zhiqiangdon: This is to make sure the query_config and response_config are completely detached. copy.copy may not work since there are objects, e.g., list, in the config values. Since this is in training, one time deepcopy doesn't affect much.

bryanyzhu (approved):
LGTM, thanks for the refactor!

sxjscience (approved):
LGTM!

Issue #2741, opened 2023-01-20
[BUG] Tensor size mismatch in multimodalpredictor


hugolytics:
- [X] I have checked that this bug exists on the latest stable version of AutoGluon
- [X] and/or I have checked that this bug exists on the latest mainline of AutoGluon via source installation

**Describe the bug**
when i try to train the multimodal predictor to use a pretrained transformer, i get an error about tensor size mismatch.

**Expected behavior**
that it trains (just like on pytorch-widedeep)

**To Reproduce**
```
from autogluon.multimodal import MultiModalPredictor
from autogluon.tabular import TabularDataset,

train_data = TabularDataset("data.parquet")

predictor = MultiModalPredictor(
    problem_type = 'regression',
    label="los"
    ).fit(train_data,
          hyperparameters={
        "model.hf_text.checkpoint_name": "CLTL/MedRoBERTa.nl",
        "model.hf_text.max_text_len": -1,
        "model.hf_text.insert_sep": True,
        })
```

** traceback **

```
INFO:pytorch_lightning.utilities.seed:Global seed set to 123
No path specified. Models will be saved in: "AutogluonModels/ag-20230120_163940/"

Downloading: 100%
674/674 [00:00<00:00, 17.3kB/s]
Downloading: 100%
504M/504M [00:16<00:00, 27.3MB/s]
Downloading: 100%
893k/893k [00:00<00:00, 2.48MB/s]
Downloading: 100%
539k/539k [00:00<00:00, 709kB/s]

Downloading /usr/local/lib/python3.8/dist-packages/autogluon/multimodal/data/templates.zip from https://automl-mm-bench.s3.amazonaws.com/few_shot/templates.zip...

INFO:pytorch_lightning.trainer.connectors.accelerator_connector:Auto select gpus: [0]
INFO:pytorch_lightning.utilities.rank_zero:Using 16bit native Automatic Mixed Precision (AMP)
INFO:pytorch_lightning.utilities.rank_zero:GPU available: True (cuda), used: True
INFO:pytorch_lightning.utilities.rank_zero:TPU available: False, using: 0 TPU cores
INFO:pytorch_lightning.utilities.rank_zero:IPU available: False, using: 0 IPUs
INFO:pytorch_lightning.utilities.rank_zero:HPU available: False, using: 0 HPUs
INFO:pytorch_lightning.accelerators.cuda:LOCAL_RANK: 0 - CUDA_VISIBLE_DEVICES: [0]
[...]

PR #2757, opened 2023-01-25
[AutoMM] Fix token max length

zhiqiangdon:
*Issue #, if available:*
#2741 

*Description of changes:*
Fix computing the max length. Checkpoint `CLTL/MedRoBERTa.nl`'s max length in tokenizer is incorrect. Use the `max_position_embeddings` instead.


sxjscience (approved):
Thanks for the quick fix!!!

PR #2758, opened 2023-01-25
[timeseries] Add wrappers for Statsforecast models

shchur:
*Description of changes:*
- Add `AutoETS`, `AutoARIMA` and `DynamicOptimizedTheta` models from `StatsForecast`.

*To Do:*
- [x] Add tests
- [x] Benchmark & add models to presets
- [x] Expose the `n_jobs` parameter


timeseries/setup.py
```diff
@@ -32,6 +32,7 @@
     "torch>=1.9,<1.14",
     "pytorch-lightning>=1.7.4,<1.9.0",
     "networkx",
+    "statsforecast==1.4.0",
```
tonyhoo, line 35: Can we make `1.4.0` the lower bound?

timeseries/src/autogluon/timeseries/models/local/statsforecast.py
```diff
@@ -0,0 +1,252 @@
...
+        raise NotImplementedError
+
+    def _fit(self, train_data, time_limit=None, verbosity=2, **kwargs) -> None:
+        """Prepare hyperparameters that will be passed to the underlying model.
+
+        As for all local models, actual fitting + predictions are delegated to the ``predict`` method.
+        """
+        # TODO: Find a way to ensure that SF models respect time_limit
+        # Fitting usually takes >= 20 seconds
+        if time_limit is not None and time_limit < 20:
```
tonyhoo, line 39: The `fit` time should be dependent on instance type as well. If `time_limit` is not implemented and we expect the local model fit shall be quick compared with other models, shall we send warning message instead?

timeseries/src/autogluon/timeseries/models/local/statsforecast.py
```diff
@@ -0,0 +1,252 @@
...
+            raise TimeLimitExceeded
+        super()._fit(train_data=train_data, time_limit=time_limit, verbosity=verbosity, **kwargs)
```
tonyhoo, line 41: One quick idea to achieve `time_limit` is to take advantage of `timeout` from `ThreadPoolExecutor` or `ProcessPoolExecutor`. Details can be found [here](https://docs.python.org/3/library/concurrent.futures.html)

timeseries/src/autogluon/timeseries/models/presets.py
```diff
@@ -75,6 +81,9 @@
     TemporalFusionTransformerMXNet=50,
     SimpleFeedForward=40,
     TransformerMXNet=30,
+    AutoARIMA=30,
+    AutoETS=70,
+    DynamicOptimizedTheta=60,
```
tonyhoo, line 86: How do we determine the priorities of these stats models? How will they added to existing ARIMA and ETA impl? Some accuracy metrics comparison might be helpful to give insights with ensemble results

tonyhoo (commented):
Are we planning to add them to any popular presets such as `medium_quality` or `high_quality`?

timeseries/setup.py
shchur: This way we are protecting ourselves from potentially breaking changes / regressions caused by newer versions of the dependencies. Both things already happened to us a few times (caused by minor releases of sktime & GluonTS), so I would rather be extra cautious here.

timeseries/src/autogluon/timeseries/models/local/statsforecast.py
shchur: I agree but, if I understand correctly, this would require monkey-patching StatsForecast. Is it fine if we leave it as-is for now and add a `TODO` comment?

shchur:
After benchmarking `AutoETS`, `AutoARIMA` and `DynamicOptimizedTheta` models on 28 datasets on `m5.4xlarge`:
- Fit time (i.e., prediction time for the validation set) 
  - ≥25s for all datasets, median time is 45s (these are approximately the same for all models)
  - For 90% of the datasets these times are slower than the respective Statsmodels models (ETS, ARIMA, Theta) by 30s in the median case
  - The runtime numbers are similar on my `p3.8xlarge` cloud desktop
  - 99% of the time is spent inside `StatsForecast.forecast`, so our wrapper does not introduce any noticeable overhead
- Performance comparison (MASE) on the test set
  - `AutoETS` > `ETS` winrate = 71%
  - `AutoARIMA` > `ARIMA` winrate = 62%
  - `DynamicOptimizedTheta` > `Theta` winrate = 57%
 
Based on these findings, I suggest to
- Add `AutoETS` to the `medium_quality` preset
- Add `AutoETS` and `AutoARIMA` to the `high_quality` & `best_quality` presets + remove HPO for `ARIMA`, `ETS` in these presets.

I will do another round of benchmarking after adding TFT & updating the SFF model from the new GluonTS release. We can update the presets based on these results.

timeseries/src/autogluon/timeseries/models/local/statsforecast.py
shchur: Updates this logic based on the discussion below.

timeseries/src/autogluon/timeseries/models/presets.py
shchur: I set the priorities inversely proportional to the average time it takes to fit these models (slower model -> lower priority).

shchur: Also added models to the presets based on the discussion below.

timeseries/src/autogluon/timeseries/models/local/statsforecast.py
tonyhoo: Let's add it in TODO and follow up with that

tonyhoo (approved):

PR #2760, opened 2023-01-25
Fix image type fallback

suzhoum:
*Issue #, if available:*

*Description of changes:*
1. Fallback to original column type when `infer_type` on image was not successful during inference;
2. Update default `missing_value_strategy` to `zero`


zhiqiangdon (approved):
LGTM! Thanks!

Issue #2756, opened 2023-01-24
[BUG- AutoMM] Training is much slower in 0.6.x compared with 0.5.2 for image classification


tonyhoo:
- [x] I have checked that this bug exists on the latest stable version of AutoGluon
- [x] and/or I have checked that this bug exists on the latest mainline of AutoGluon via source installation

**Describe the bug**
Training is much slower in 0.6.x compared with 0.5.2 for image classification 

**Expected behavior**
User should expect similar training time when `time_limit` is set across versions using the same HP and dataset
**To Reproduce**
* `pip install autogluon==0.6.2`
* Run following codes with different verions:

```python
if __name__ == "__main__":

    from autogluon.multimodal import MultiModalPredictor
    from autogluon import core
    import time

    import warnings
    warnings.filterwarnings('ignore')
    from autogluon.vision import ImageDataset
    train_dataset, _, test_dataset = ImageDataset.from_folders("https://autogluon.s3.amazonaws.com/datasets/shopee-iet.zip")
    print(train_dataset)

    version = core.version.__version__
    predictor = MultiModalPredictor(label="label", path=f"./automm_imgcls/{version}/{time.time()}")
    start = time.time()
    predictor.fit(
        train_data=train_dataset,
        time_limit=60, # seconds
    ) # you can trust the default config, e.g., we use a `swin_base_patch4_window7_224` model
    end = time.time()
    print(f"version {version} takes {end-start} seconds")
```
**Screenshots**
|       | Time (sec, p3.8xl) |
|-------|------------|
| 0.5.2 | 149        |
| 0.6.1 | 263        |
| 0.6.2|  285        |

**Installed Versions**
Which version of AutoGluon are you are using?  
* 0.6.x
* 0.5.2

```python
# Replace this code with the output of the following:
from autogluon.core.utils import show_versions
show_versions()
```

</details>

**Additional context**

sxjscience:
@zhiqiangdon May be we can also consider to disable greedysoup by default so the total training time can be reduced?

tonyhoo:
@zhiqiangdon @sxjscience  do we know what the accuracy impact will be for greedysoup?

tonyhoo:
@liangfu and @suzhoum  found the removing `env.num_workers` recovered the training throughput back to 0.5.2. Thanks for the work. Until fix is patched and released (tentatively in 0.7). A stop gap solution is to explicitly set `env.num_workers=0` in the hyperparameters. See below example:

```python
[...]

PR #2416, opened 2022-11-16
Set num_workers for image related presets and clean matcher config

zhiqiangdon:
*Issue #, if available:*

*Description of changes:*

1. num_workers=0 makes training too slow for image related training. Set num_workers=2 in image related presets.
2. clean matcher config.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #2766, opened 2023-01-26
[multimodal] undo changes to num_workers

liangfu:
*Issue #, if available:*

#2756

*Description of changes:*

The root cause of the issue is that https://github.com/autogluon/autogluon/commit/93fcbb955eb628b6896b8d277386a4c9a9345384 changed num_workers=2

Here is a visualization before and after the change #2416 that could affect the overall evaluation time of beginner_image_cls.ipynb .

image

By removing ‘num_workers=2’, we can reduce predictor.fit from 206 seconds to 128 seconds. Training throughput increased from 3.92s/it to 0.68s/it (aka 1.46it/s).

Code to reproduce the results

```python
import pandas as pd
import time
import os
import shutil

def main():
    download_dir = './ag_automm_tutorial_imgcls'
    train_data_byte, test_data_byte = shopee_dataset(download_dir, is_bytearray=True)
    model_path = f"./tmp/automm_shopee"
    if os.path.exists(model_path):
        shutil.rmtree(model_path)
    predictor = MultiModalPredictor(label="label", path=model_path)
    tic = time.time()
    predictor.fit(
        train_data=train_data_byte, # you can use train_data_byte as well
        time_limit=30, # seconds
    ) # you can trust the default config, e.g., we use a `swin_base_patch4_window7_224` model
    print(f"elapsed (fit): {(time.time()-tic)*1000:.1f} ms")

if __name__=="__main__":
    main()
```

### DIFF
--- a/multimodal/src/autogluon/multimodal/configs/environment/default.yaml
+++ b/multimodal/src/autogluon/multimodal/configs/environment/default.yaml
@@ -11,6 +11,6 @@ env:
   fast_dev_run: False
   deterministic: False
   auto_select_gpus: True
-  strategy: "ddp_spawn"
+  strategy: "ddp"
   deepspeed_allgather_size: 1e9 # DeepSpeed allgather size (number of elements gathered at a time). Limits memory required for allgather on large models.
   deepspeed_allreduce_size: 1e9 # DeepSpeed allreduce size (number of elements reduced at a time). Limits memory required for allreduce on large models.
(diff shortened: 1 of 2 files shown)


tonyhoo (approved):
LGTM! Great findings in short period of time. Thanks for the investigation and fix

sxjscience (approved):
Removing the `num_workers` flag is sensible. It also avoids certain multi-processing issues.

zhiqiangdon (commented):
I think 0.5.x also used `env.num_workers=2`. Could you check it? I have some doubts on the cause.

zhiqiangdon (commented):
Per the offline discussion with @liangfu, the root cause of issue needs more investigations. I will take a look at this.

liangfu:
Canceling this PR for now, in favor of #2776

PR #2770, opened 2023-01-26
[eda] use LightGBM in quick_fit if available

gradientsky:
*Description of changes:*
By default quick fit is using `RF` model if no hyperparameters are provided. Switched default to use Extra Trees LightGBM if LightGBM is available.

### DIFF
--- a/eda/src/autogluon/eda/auto/simple.py
+++ b/eda/src/autogluon/eda/auto/simple.py
@@ -29,6 +29,7 @@ from ..analysis.dataset import (
 )
 from ..analysis.interaction import FeatureDistanceAnalysis
 from ..state import is_key_present_in_state
+from ..utils.defaults import QuickFitDefaults
 from ..visualization import (
     ConfusionMatrix,
     CorrelationVisualization,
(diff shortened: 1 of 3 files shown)


eda/src/autogluon/eda/auto/simple.py
```diff
@@ -628,23 +628,41 @@ def covariate_shift_detection(
...
-            ],
-        }
+
+        fit_args["fit_weighted_ensemble"] = False
+        if _is_lightgbm_available():
+            fit_args["hyperparameters"] = {
+                "GBM": [
+                    {"extra_trees": True, "ag_args": {"name_suffix": "XT"}},
+                ]
+            }
```
yinweisu, line 650: Maybe saving these dicts in a dedicated file, i.e. constants, for easier maintenance?

yinweisu (approved):
LGTM! One minor comment

eda/src/autogluon/eda/auto/simple.py
gradientsky: Good call - updated

Innixma (approved):
LGTM!

PR #2781, opened 2023-01-30
[timeseries] Implement missing value imputation for TimeSeriesDataFrame

shchur:
*Description of changes:*
- Two new methods for imputing the missing values. These should be called before passing the data to the `TimeSeriesPredictor`.
  - `to_regular_index(freq)` - fills gaps in an irregularly-sampled time series with NaNs
  - `fill_missing_values(method)`  - drop leading NaNs & replace other NaNs (middle/trailing) using the chosen method (forward fill or interpolation)

### DIFF
--- a/timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
+++ b/timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
@@ -634,3 +634,131 @@ class TimeSeriesDataFrame(pd.DataFrame):
         df_view._cached_freq = freq
 
         return df_view
+
+    def to_regular_index(self, freq: str) -> "TimeSeriesDataFrame":
+        """Fill the gaps in an irregularly-sampled time series with NaNs.
+
+        Parameters
+        ----------
+        freq: str
+            Frequency string of the new time series data index.
+
+        Examples
+        --------
+        >>> print(ts_dataframe)
+                            target
+        item_id timestamp
+        0       2019-01-01     NaN
+                2019-01-03     1.0
+                2019-01-06     2.0
+                2019-01-07     NaN
+        1       2019-02-04     3.0
+                2019-02-07     4.0
+
+        >>> print(ts_dataframe.to_regular_index(freq="D"))
+                            target
+        item_id timestamp
+        0       2019-01-01     NaN
+                2019-01-02     NaN
+                2019-01-03     1.0
+                2019-01-04     NaN
+                2019-01-05     NaN
+                2019-01-06     2.0
+                2019-01-07     NaN
+        1       2019-02-04     3.0
+                2019-02-05     NaN
+                2019-02-06     NaN
+                2019-02-07     4.0
+
+        """
+        if self.freq is not None:
+            if self.freq != freq:
+                raise ValueError(
+                    f"TimeSeriesDataFrame already has a regular index with freq '{self.freq}' "
+                    f"that cannot be converted to the given freq '{freq}'"
+                )
+            else:
+                return self
+
+        filled_series = []
+        for item_id, time_series in self.groupby(level=ITEMID, sort=False):
+            time_series = time_series.droplevel(ITEMID)
+            timestamps = time_series.index
+            if not timestamps.is_monotonic_increasing:
+                raise ValueError(
+                    "Please make sure that the timestamps are sorted in increasing order for all time series."
+                )
+            start = timestamps[0]
+            end = timestamps[-1]
+            filled_timestamps = pd.date_range(start=start, end=end, freq=freq, name=TIMESTAMP)
+            if not timestamps.isin(filled_timestamps).all():
+                raise ValueError(
+                    f"Irregularly-sampled timestamps in this TimeSeriesDataFrame are not compatible "
+                    f"with the given frequency '{freq}'"
+                )
+            filled_series.append(pd.concat({item_id: time_series.reindex(filled_timestamps)}, names=[ITEMID]))
+
+        return TimeSeriesDataFrame(pd.concat(filled_series), static_features=self.static_features)
+
+    def fill_missing_values(self, method: str = "ffill") -> "TimeSeriesDataFrame":
+        """Fill missing values represented by NaN.
+
+        Missing values at the beginning of each time series are dropped. Remaining missing values (in the middle and
+        at the end) are filled using the chosen method.
+
+
+        Parameters
+        ----------
+        method : {"ffill", "interpolate"}, default = "ffill"
+            Method used to impute missing values.
+            "ffill" - propagate last valid observation forward.
+            "interpolate" - fill NaN values using linear interpolation.
+
+        Examples
+        --------
+        >>> print(ts_dataframe)
+                            target
+        item_id timestamp
+        0       2019-01-01     NaN
+                2019-01-02     NaN
+                2019-01-03     1.0
+                2019-01-04     NaN
+                2019-01-05     NaN
+                2019-01-06     2.0
+                2019-01-07     NaN
+        1       2019-02-04     3.0
+                2019-02-05     NaN
+                2019-02-06     NaN
+                2019-02-07     4.0
+
+        >>> print(ts_dataframe.fill_missing_values(method="ffill"))
+                            target
+        item_id timestamp
+        0       2019-01-03     1.0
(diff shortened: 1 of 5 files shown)


timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
```diff
@@ -634,3 +634,131 @@ def get_reindexed_view(self, freq: str = "S") -> TimeSeriesDataFrame:
...
+                2019-01-07     NaN
+        1       2019-02-04     3.0
+                2019-02-05     NaN
+                2019-02-06     NaN
+                2019-02-07     4.0
+
+        """
+        if self.freq is not None:
+            if self.freq != freq:
+                raise ValueError(
```
tonyhoo, line 676: Wondering shall we allow users to do this. In practice, it is possible users want to train various predictors on different frequencies for experimenting purpose

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
```diff
@@ -634,3 +634,131 @@ def get_reindexed_view(self, freq: str = "S") -> TimeSeriesDataFrame:
...
+        for item_id, time_series in self.groupby(level=ITEMID, sort=False):
+            time_series = time_series.droplevel(ITEMID)
+            timestamps = time_series.index
+            if not timestamps.is_monotonic_increasing:
+                raise ValueError(
+                    "Please make sure that the timestamps are sorted in increasing order for all time series."
+                )
+            start = timestamps[0]
+            end = timestamps[-1]
+            filled_timestamps = pd.date_range(start=start, end=end, freq=freq, name=TIMESTAMP)
```
tonyhoo, line 693: For simplicity, can we use `resample` [function](https://pandas.pydata.org/docs/reference/api/pandas.Series.resample.html)?

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
```diff
@@ -634,3 +634,131 @@ def get_reindexed_view(self, freq: str = "S") -> TimeSeriesDataFrame:
...
+            if not timestamps.isin(filled_timestamps).all():
+                raise ValueError(
+                    f"Irregularly-sampled timestamps in this TimeSeriesDataFrame are not compatible "
+                    f"with the given frequency '{freq}'"
+                )
+            filled_series.append(pd.concat({item_id: time_series.reindex(filled_timestamps)}, names=[ITEMID]))
+
+        return TimeSeriesDataFrame(pd.concat(filled_series), static_features=self.static_features)
```
tonyhoo, line 701: shall we resample `covariates` as well?

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
```diff
@@ -634,3 +634,131 @@ def get_reindexed_view(self, freq: str = "S") -> TimeSeriesDataFrame:
...
+
+        Missing values at the beginning of each time series are dropped. Remaining missing values (in the middle and
+        at the end) are filled using the chosen method.
+
+
+        Parameters
+        ----------
+        method : {"ffill", "interpolate"}, default = "ffill"
+            Method used to impute missing values.
+            "ffill" - propagate last valid observation forward.
```
tonyhoo, line 714: We should support all the filling [methods](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.fillna.html#pandas.DataFrame.fillna) available in pandas.

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
```diff
@@ -634,3 +634,131 @@ def get_reindexed_view(self, freq: str = "S") -> TimeSeriesDataFrame:
...
+            )
+
+        if method == "ffill":
+            fill_method = lambda ts: ts.fillna(method="ffill")
+        elif method == "interpolate":
+            fill_method = lambda ts: ts.interpolate()
+        else:
+            raise ValueError(f"Invalid fill method. Expecting one of {'ffill', 'interpolate'}. Got {method}")
+
+        filled_df = pd.DataFrame(self).groupby(level=ITEMID, sort=False, group_keys=False).apply(fill_method)
```
tonyhoo, line 761: We can just call `fillna` on the DataFrame

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
shchur: This is really cool and feels a lot more elegant, thanks!

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
shchur: I would suggest that we ask users to do these transformations outside of AG (should be pretty easy in pandas) or, alternatively, provide another method for this purpose if such request comes from the users. Does this sounds reasonable?

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
shchur: All columns are already resampled (including past/known covariates). I've added another test to make sure that this works as intended.

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
shchur: This might cause problems where leading NaNs in one time series are filled by the last value of the previous time series, e.g.,
```python
        item_id timestamp
        0       2019-01-01     1000.0
        1       2019-02-03     NaN
                2019-02-04     2.0
```
will be filled as
```python
        item_id timestamp
        0       2019-01-01     1000.0
        1       2019-02-03     1000.0
                2019-02-04     2.0
```

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
shchur: I see two potential problems with `bfill`/`backfill`:
- it doesn't fill the trailing NaNs, and these are the ones that are actually important to ensure that all models can generate predictions over the forecast horizon. E.g, after we bfill `[1, 1, NaN, NaN]`, we again get `[1, 1, NaN, NaN]` and this cannot be processed by `TimeSeriesPredictor`. We cannot just drop the trailing NaNs as easily as we can drop leading NaNs with `ffill`.
- it introduces information leakage from the test/val set, which might affect model selection.

Do you think there is a strong potential use case for `bfill` that we should support?

tonyhoo: Information leakage is a critical point, agree that we should drop `bfill` for now. Curious for interpolate, will linear interpolation cause leakage as well by any chance?

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
tonyhoo: I meant `fillna` called on grouped results instead of the whole dataframe. Since the lambda function already uses fillna on individual, shall be good

tonyhoo (approved):

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
shchur: Nice! I wasn't aware of this functionality

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
shchur: Good point regarding `interpolation`. I've checked how other libraries ([sktime](https://www.sktime.org/en/v0.16.0/api_reference/auto_generated/sktime.transformations.series.impute.Imputer.html), [darts](https://unit8co.github.io/darts/_modules/darts/utils/missing_values.html#fill_missing_values)) handle this and updated the functionality to be more in line with them. 
- Do not change the index (never drop the leading NaNs)
- By default, use `ffill` to fill gaps + trailing NaNs, then use `bfill` to fill the leading NaNs.
- Added options `constant` and `bfill`.
- Added warnings for `bfill` and `interpolate` that these may lead to data leakage.

tonyhoo (approved):

PR #2789, opened 2023-01-31
[AutoMM] Filter out unused model configs

zhiqiangdon:
*Issue #, if available:*
Some unused model configs still exist in the saved config, which can confuse users.

*Description of changes:*
Filter out the unused model configs and only keep the used ones.

### DIFF
--- a/multimodal/src/autogluon/multimodal/utils/model.py
+++ b/multimodal/src/autogluon/multimodal/utils/model.py
@@ -142,9 +142,7 @@ def select_model(
     # clean up unused model configs
     model_keys = list(config.model.keys())
     for model_name in model_keys:
-        if model_name == "names":
-            pass
-        if model_name not in selected_model_names:
+        if model_name not in selected_model_names + ["names"]:
             delattr(config.model, model_name)
 
     return config
--- a/multimodal/src/autogluon/multimodal/utils/model.py
+++ b/multimodal/src/autogluon/multimodal/utils/model.py
@@ -141,8 +141,9 @@ def select_model(
 
     # clean up unused model configs
     model_keys = list(config.model.keys())
-    model_keys.remove("names")
     for model_name in model_keys:
+        if model_name == "names":
+            pass
         if model_name not in selected_model_names:
             delattr(config.model, model_name)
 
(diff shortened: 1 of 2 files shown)


multimodal/src/autogluon/multimodal/utils/model.py
```diff
@@ -139,6 +139,13 @@ def select_model(
     for model_name in selected_model_names:
         logger.debug(f"model dtypes: {getattr(config.model, model_name).data_types}")
 
+    # clean up unused model configs
+    model_keys = list(config.model.keys())
+    model_keys.remove("names")
```
sxjscience, line 144: Should we still keep the `names` for clarity?

zhiqiangdon: `names` is necessary for customizing the models to use. For example, https://github.com/autogluon/autogluon/blob/master/multimodal/src/autogluon/multimodal/presets.py#L152

sxjscience: Yeah, I misread the logic.

Is the code 
```python
model_keys = list(config.model.keys())
model_keys.remove("names")
for model_name in model_keys:
    if model_name not in selected_model_names:
        delattr(config.model, model_name)
```

same as the following?

```python
for model_name in model_keys:
    if model_name not in selected_model_names + ["names"]:
        delattr(config.model, model_name)


```

Also, a side note is that I thought it may be unsafe to use `delattr` for a DictConfig object. But I checked the test-case: https://github.com/omry/omegaconf/blob/d02a0c10cbba8b01c92e1faffa995ebf2da4abe2/tests/test_basic_ops_dict.py#L88, it's a valid operator for DictConfig.

zhiqiangdon: If we don't remove key `names`, we need to add the if-pass logic inside the following for loop. They basically have the same effect.

sxjscience: For me, I just got confused when seeing `.remove("name")` as the purpose of the function is to clean up the keys.

zhiqiangdon: > 

I see. `model_keys` is a temporary variable for one-time usage and will not affect other logics.

zhiqiangdon: Updated by putting `names` inside the for-loop.

sxjscience (approved):
Thanks for incorporating the suggestion. LGTM!

PR #2790, opened 2023-01-31
Deprecate ag.vision in AG v0.7

bryanyzhu:
*Issue #, if available:*
#2694 #2695 

*Description of changes:*
Deprecate ag.vision in AG v0.7

- [x] modify readme
- [x] modify index page
- [x] modify tutorial page
- [x] Also remove tutorials, modify index page for ag.text

Will test backward compatibility once HPO preset PR is merged. 
Will open another PR to remove ag.vision from CI


.github/workflows/pypi_release.yml
```diff
@@ -30,7 +30,7 @@ jobs:
         TWINE_PASSWORD: ${{ secrets. PYPI_PASSWORD }}
         RELEASE: 1
       run: |
-        for v in common core features tabular multimodal vision timeseries autogluon
+        for v in common core features tabular multimodal timeseries autogluon
```
Innixma, line 33: Is the intention to not release autogluon.vision 0.7? I remember we talked about two options so I'd just like to double check: 

1. Don't release autogluon.vision 0.7.
2. Release autogluon.vision 0.7, but don't have it as part of the default autogluon install.

Based on the code changes here, this will not release autogluon.vision 0.7.

Innixma: Note: I am fully supportive of either option

sxjscience: I'm supportive in just not releasing vision 0.7.

bryanyzhu: Given `text` is already removed from pypi, I think `vision` can be removed too. If user wants to stick with `ag.vision`, they can `pip install autogluon.vision` from any previous version. So maybe it is easier just not release vision in v0.7.

sxjscience (approved):
LGTM

FANGAreNotGnu (approved):
LGTM, thanks!

zhiqiangdon (approved):
LGTM

yinweisu:
Do we want to remove the vision source code too?

bryanyzhu:
> Do we want to remove the vision source code too?

At this moment, I see the source code for `ag.text` also exists. If we agree to remove the entire folder, we can do this in another PR.

Innixma:
Yes we should remove the source code for text and vision, but this should be a separate PR.

Innixma (approved):
LGTM!

PR #2791, opened 2023-01-31
[CI] Enable HF model caching

yinweisu:
*Issue #, if available:*

*Description of changes:*
* Have setup an AWS Batch job running chronically to sync up our models stored in S3 with HF. An EFS was used as a cache so we don't re-download the model each time.
  * List of models will be fetched dynamically from our repo with `hf_model_list.yaml`
  * All models in the list will be downloaded and synced to our S3, no matter which subfolder it belongs to
* CI job will use `setup_hf_model_mirror.py` to download the corresponding models to `~/.cache/huggingface/hub` for the specific subfolder to test.
* Updated the structure of the model list.
* A follow-up PR to enforce model list updating will be sent

### DIFF
--- a/.github/workflow_scripts/setup_hf_model_mirror.py
+++ b/.github/workflow_scripts/setup_hf_model_mirror.py
@@ -6,8 +6,8 @@ import yaml
 
 parser = argparse.ArgumentParser(formatter_class=argparse.ArgumentDefaultsHelpFormatter)
 
-parser.add_argument("--model_list_file", help="file containing list of models to download.", type=str)
-parser.add_argument("--sub_folder", help="which subfolder to download models for")
+parser.add_argument("--model_list_file", help="file containing list of models to download.", type=str, required=True)
+parser.add_argument("--sub_folder", help="which subfolder to download models for", required=True)
 parser.add_argument("--cache_dir", help="place to cache the downloaded models", default="~/.cache/huggingface/hub")
 
 args = parser.parse_args()
--- a/.github/workflow_scripts/setup_hf_model_mirror.py
+++ b/.github/workflow_scripts/setup_hf_model_mirror.py
@@ -1,0 +1,34 @@
+import argparse
+import os
+import subprocess
+import yaml
+
+
+parser = argparse.ArgumentParser(formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+
+parser.add_argument("--model_list_file", help="file containing list of models to download.", type=str)
+parser.add_argument("--sub_folder", help="which subfolder to download models for")
+parser.add_argument("--cache_dir", help="place to cache the downloaded models", default="~/.cache/hugging_face/hub")
+
+args = parser.parse_args()
+
+model_list_file = args.model_list_file
+sub_folder = args.sub_folder
+cache_dir = args.cache_dir
+cache_dir = os.path.expanduser(cache_dir)
+
+with open(model_list_file, "r") as fp:
+    models = yaml.safe_load(fp)
+    sub_folder_models = models[sub_folder]
+
+os.makedirs(cache_dir)
+
+bucket = "s3://autogluon-hf-model-mirror"
+model_prefix = "models"
+for model in sub_folder_models:
+    model_name = "--".join(model.split('/'))
+    model_name = "--".join([model_prefix, model_name])
+    model_path = os.path.join(cache_dir, model_name)
+    os.makedirs(model_path)
+    s3_path = bucket + "/" + model_name
+    subprocess.run(["aws", "s3", "cp", s3_path, model_path, '--recursive'])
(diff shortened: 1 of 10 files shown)


CI/hf_mirror/download_hf_models.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+df -h  # debug purpose to make sure efs is attached correclty
```
tonyhoo, line 3: [minor] We can fail the script if the volume is not attached correctly with some error message which can be more indicative for users

CI/hf_mirror/download_hf_models.sh
```diff
@@ -0,0 +1,6 @@
...
+git clone https://github.com/autogluon/autogluon.git
```
tonyhoo, line 4: Why we need autogluon repo to be present?

CI/hf_mirror/download_hf_models.py
```diff
@@ -0,0 +1,27 @@
+import argparse
+import yaml
+
+from huggingface_hub import snapshot_download
+from huggingface_hub.utils import RepositoryNotFoundError
+
+
+parser = argparse.ArgumentParser(formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+
+parser.add_argument('--model_list_file', help='file containing list of models to download.', type=str)
```
tonyhoo, line 10: Is this argument optional? If not, let's remove the `--` prefix or set `required=True`

CI/hf_mirror/download_hf_models.sh
```diff
@@ -0,0 +1,6 @@
...
+python3 download_hf_models.py --model_list_file autogluon/multimodal/tests/hf_model_list.yaml
```
tonyhoo, line 5: Wondering shall we even allow users to pass a separate file list to the python function as it might expose us to security risks. If there is no such use case, we shall go with the hard coded solution

.github/workflow_scripts/setup_hf_model_mirror.py
```diff
@@ -0,0 +1,34 @@
...
+import os
+import subprocess
+import yaml
+
+
+parser = argparse.ArgumentParser(formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+
+parser.add_argument("--model_list_file", help="file containing list of models to download.", type=str)
+parser.add_argument("--sub_folder", help="which subfolder to download models for")
+parser.add_argument("--cache_dir", help="place to cache the downloaded models", default="~/.cache/huggingface/hub")
```
tonyhoo, line 11: Remove prefix `--` for required field

CI/hf_mirror/deploy.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+aws ecr get-login-password --region us-east-1 | docker login --username AWS --password-stdin 369469875935.dkr.ecr.us-east-1.amazonaws.com
+docker build -t autogluon-hf-mirror .
+docker tag autogluon-hf-mirror:latest 369469875935.dkr.ecr.us-east-1.amazonaws.com/autogluon-hf-mirror:latest
+docker push 369469875935.dkr.ecr.us-east-1.amazonaws.com/autogluon-hf-mirror:latest
```
suzhoum, line 6: Minor: can we make `369469875935.dkr.ecr.us-east-1.amazonaws.com` a constant?

suzhoum (approved):
Thanks for the nice workaround!

CI/hf_mirror/download_hf_models.sh
yinweisu: We need to fetch the model list dynamically to know which models to download

CI/hf_mirror/download_hf_models.sh
yinweisu: Yea I get your point. Though this script is actually inside the docker container hence cannot be modified unless re-pushed. Therefore, we will always pass `autogluon/multimodal/tests/hf_model_list.yaml` and user cannot modify it

CI/hf_mirror/download_hf_models.sh
yinweisu: good point

zhiqiangdon (approved):
LGTM. Do we also need to automatically sync our s3 with the hf model hub?

yinweisu:
> LGTM. Do we also need to automatically sync our s3 with the hf model hub?

Yes, we are doing that on a nightly basis

.github/workflow_scripts/setup_hf_model_mirror.py
yinweisu: Added `required=True` instead

tonyhoo (approved):

PR #2793, opened 2023-01-31
add  fusion base class

suzhoum:
*Issue #, if available:*

*Description of changes:*


sxjscience:
Thanks for refactoring the code base by adding an abstract class!!

Is it possible for adopting the design like in https://github.com/deepset-ai/haystack/tree/8002cf92d6a54c60ac8180f85d1c42b1105f547c/haystack/nodes/ranker . We can use the following folder structure:

```shell
multimodal/models/fusion/__init__.py
multimodal/models/fusion/base.py
multimodal/models/fusion/fusion_mlp.py
multimodal/models/fusion/fusion_ner.py
multimodal/models/fusion/fusion_transformer.py
```

sxjscience:
@zhiqiangdon We can consider to refactor the other folders by using `base.py` for storing the base class, and separately store important modules as independent files.

zhiqiangdon:
> @zhiqiangdon We can consider to refactor the other folders by using `base.py` for storing the base class, and separately store important modules as independent files.

Sure. Data processors may also need a base class.

sxjscience (approved):
LGTM!

multimodal/src/autogluon/multimodal/models/fusion/base.py
```diff
@@ -0,0 +1,67 @@
...
+
+logger = logging.getLogger(AUTOMM)
+
+
+class BaseMultimodalFusionModel(nn.Module):
+    """
+    A base class to fuse different models' features (single-modal and multimodal).
+    """
+
+    def __init__(self):
```
zhiqiangdon, line 15: Do we need to add the shared arguments, e.g., prefix and models,  in `init()`?

multimodal/src/autogluon/multimodal/models/fusion/base.py
```diff
@@ -0,0 +1,67 @@
+import logging
+
+from torch import nn
+
+from ...constants import AUTOMM, LABEL
+
+logger = logging.getLogger(AUTOMM)
+
+
+class BaseMultimodalFusionModel(nn.Module):
```
sxjscience, line 10: We can also optionally make it an abstract class.

multimodal/src/autogluon/multimodal/models/fusion/base.py
```diff
@@ -0,0 +1,80 @@
...
+        self,
+        prefix: str,
+        models: list,
+        loss_weight: Optional[float] = None,
+    ):
+        super().__init__()
+
+        self.prefix = prefix
+        self.loss_weight = loss_weight
+        self.model = nn.ModuleList(models)
```
sxjscience, line 27: This line will force the models to be pytorch models.

@yongxinw Will it be problematic if we are going to use SVM as a fusion model?

suzhoum: I can move `self.model` out of the base class, or it can also be overwritten in the child class.

zhiqiangdon: `self.model` is for the single models. The fusion part is `self._fusion_xxx`.

zhiqiangdon (approved):
LGTM!

multimodal/src/autogluon/multimodal/models/fusion/base.py
zhiqiangdon: Which one is better: `AbstractModel ` or `BaseModel`?

suzhoum: Abstract might sound better. I can update it.

zhiqiangdon: Also, consider aligning the file and class names, i.e., using the same key word.

PR #2794, opened 2023-02-01
Fix TIMM integration to support adaptive image size

sxjscience:
*Issue #, if available:*

Solve https://github.com/autogluon/autogluon/issues/2585

*Description of changes:*

- [x] Support specifying different image_size for specific TIMM backbones like ResNet / RegNet / MobileNetV3 / EfficientNet, while disable the behavior for backbones like SwinTransformer.


sxjscience:
Lint check had failed since black has been upgraded to `23.1.0`. Thus, limit the black version to be `<23.0`.

sxjscience:
Try to use the new PR https://github.com/autogluon/autogluon/pull/2810 but the HF model list check still failed @yinweisu .

multimodal/tests/unittests/others_2/test_backward_compatibility.py
```diff
@@ -43,3 +44,32 @@ def test_load_old_checkpoint(cls):
         time_limit=10,
         hyperparameters={"optimization.top_k_average_method": "uniform_soup", "env.num_gpus": 1},
     )
+
+
+def test_v0_6_2_checkpoint_timm_image():
```
bryanyzhu, line 49: Can I ask what this unit test is for? Thank you

multimodal/src/autogluon/multimodal/models/timm_image.py
```diff
@@ -13,6 +13,10 @@
 logger = logging.getLogger(AUTOMM)
 
 
+# Stores the
+SUPPORT_VARIABLE_INPUT_SIZE_TIMM_CLASSES = {"convnext", "efficientnet", "mobilenetv3", "regnet", "resnet"}
```
bryanyzhu, line 17: Just curious, does this list need to manually update when there are more models support variable input size?

bryanyzhu (approved):
Thanks for the fix.

multimodal/tests/unittests/others_2/test_backward_compatibility.py
sxjscience: Current checkpoints also have “timm_image.image_size”. However, before this PR, we will simply ignore this flag. This PR correctly implements the “timm_image.image_size” and also changed the default value from 224 to “null”. We need to test the backward compatibility (i.e. we are able to load the checkpoints saved with autogluon.multimodal<=0.6.2).

multimodal/src/autogluon/multimodal/models/timm_image.py
sxjscience: Yes. I implemented two logics: 1) if the test_input_size is different from input_size, this means that the network allows variable input size; 2) stored some commonly used backbones that supports variable input size.

PR #2815, opened 2023-02-02
Fix torch version range, support 1.13

Innixma:
*Issue #, if available:*

*Description of changes:*
- Fix torch version range, support 1.13
- Increase lower bound to 1.9


sxjscience (approved):
LGTM

PR #2816, opened 2023-02-02
[timeseries] Add ujson dependency to silence the GluonTS warning

shchur:
*Description of changes:*
- The new GluonTS version seems to overcome our method for silencing the JSON warning, so it's probably easier to just add `ujson` as a dependency. The library is actively developed and actually isn't use in our code, so it should be safe.


canerturkmen (approved):
LGTM!

PR #2818, opened 2023-02-02
[Multimodal] Update version bounds

sxjscience:
*Issue #, if available:*

https://github.com/autogluon/autogluon/issues/612

*Description of changes:*

Update the version bounds in multimodal dependencies.


sxjscience:
For version bounds, we can adopt the guideline to always use:

```python
# Packages that have low--up bounds.
PACKAGE_NAME>=LOWER_BOUND,<UPPER_BOUND

# Packages that must be upgraded if it's lower version
PACKAGE_NAME>=LOWER_BOUND

# Most common packages
PACKAGE_NAME
```

multimodal/setup.py
```diff
@@ -28,35 +28,33 @@
...
     f"autogluon.common=={version}",
-    "pytorch-metric-learning>=1.3.0,<1.4.0",
+    "pytorch-metric-learning>=1.3.0,<2.0",
     "nlpaug>=1.1.10,<1.2.0",
     "nltk>=3.4.5,<4.0.0",
-    "openmim>0.1.5,<=0.2.1",
-    "defusedxml>=0.7.1,<=0.7.1",
-    "albumentations>=1.1.0,<=1.2.0",
-    "jinja2>=3.0.3,<3.1",
+    "openmim>0.1.5,<0.4.0",
```
sxjscience, line 53: @bryanyzhu I updated the version bound associated with openmim.

multimodal/setup.py
```diff
@@ -28,35 +28,33 @@
...
-    "omegaconf>=2.1.1,<2.2.0",
+    "transformers>=4.23.0,<4.27.0",
+    "nptyping>=1.4.4,<2.5.0",
+    "omegaconf>=2.1.1,<2.3.0",
     "sentencepiece>=0.1.95,<0.2.0",
     f"autogluon.core[raytune]=={version}",
     f"autogluon.features=={version}",
     f"autogluon.common=={version}",
-    "pytorch-metric-learning>=1.3.0,<1.4.0",
+    "pytorch-metric-learning>=1.3.0,<2.0",
```
sxjscience, line 50: @zhiqiangdon I noticed that pytorch-metric-learning released version 2.0. Do we need to consider to update the version?

Innixma:
@sxjscience Regarding this: how about the following:

```python
"numpy",  # version range defined in `_setup_utils.py`
"pandas",  # version range defined in `_setup_utils.py`
"torchvision>=foo,<bar",

...

```

multimodal/setup.py
```diff
@@ -28,35 +28,33 @@
...
     "nlpaug>=1.1.10,<1.2.0",
     "nltk>=3.4.5,<4.0.0",
-    "openmim>0.1.5,<=0.2.1",
-    "defusedxml>=0.7.1,<=0.7.1",
-    "albumentations>=1.1.0,<=1.2.0",
-    "jinja2>=3.0.3,<3.1",
+    "openmim>0.1.5,<0.4.0",
+    "defusedxml>=0.7.1,<0.7.2",
+    "jinja2>=3.0.3,<3.2",
```
Innixma, line 55: jinja2 needs to be `<3.1` or it will error, at least it did in my PR before I dropped to `<3.1`.

Innixma: The error may have been from the `requirements_docs` dependencies though, so maybe `<3.2` is ok for the package depencenies, lets keep it `<3.2` and see if it passes

multimodal/setup.py
bryanyzhu: Cool, this should be fine.

multimodal/setup.py
h-vetinari: I [checked](https://github.com/conda-forge/autogluon.multimodal-feedstock/pull/15/commits/8164e39502a87c0ea265145340d70e4d0e6f5553) and `<2` would be enough to get things working. It's not ideal and should be lifted to allow v2, but it's less urgent than lifting the cap from `<1.4`.

multimodal/setup.py
```diff
@@ -20,43 +20,41 @@
...
+    "Pillow",  # version range defined in `core/_setup_utils.py`
+    "tqdm",  # version range defined in `core/_setup_utils.py`
     "boto3",
     "requests",
-    "jsonschema<=4.8.0",
-    "seqeval<=1.2.2",
-    "evaluate<=0.3.0",
-    "accelerate>=0.9,<0.16",
+    "jsonschema<4.18",
+    "seqeval<1.2.3",
```
h-vetinari, line 32: Unless you expect patch versions to be breaking, I'd suggest
```suggestion
    "seqeval<1.3",
```

multimodal/setup.py
```diff
@@ -20,43 +20,41 @@
...
     "timm<0.7.0",
     "torch>=1.9,<1.14",
     "torchvision<0.15.0",
     "fairscale>=0.4.5,<0.4.14",
     "scikit-image>=0.19.1,<0.20.0",
-    "smart_open>=5.2.1,<5.3.0",
     "pytorch_lightning>=1.8.0,<1.10.0",
-    "text-unidecode<=1.3",
+    "text-unidecode<1.4",
     "torchmetrics>=0.8.0,<0.9.0",
```
h-vetinari, line 42: Latest version for this is 0.11.1, could the cap be lifted?

multimodal/setup.py
```diff
@@ -20,43 +20,41 @@
...
-    "seqeval<=1.2.2",
-    "evaluate<=0.3.0",
-    "accelerate>=0.9,<0.16",
+    "jsonschema<4.18",
+    "seqeval<1.2.3",
+    "evaluate<0.4.0",
+    "accelerate>=0.9,<0.17",
     "timm<0.7.0",
     "torch>=1.9,<1.14",
     "torchvision<0.15.0",
```
h-vetinari, line 37: Side note: what happened to the torchtext dependency? 🤔

sxjscience: We managed to remove it by upgrading lightning... See https://github.com/autogluon/autogluon/pull/2739

multimodal/setup.py
sxjscience: @zhiqiangdon Can we list the cap of torchmetrics?

multimodal/setup.py
sxjscience: Thanks. Revised.

multimodal/setup.py
sxjscience: Okay, let's see if the CI + document build can pass.

multimodal/src/autogluon/multimodal/data/process_text.py
```diff
@@ -235,7 +235,7 @@ def collate_fn(self, text_column_names: Optional[List] = None) -> Dict:
 
     def build_one_token_sequence(
         self,
-        text_tokens: Dict[str, NDArray[(Any,), np.int32]],
+        text_tokens: Dict[str, NDArray],
```
sxjscience, line 238: @zhiqiangdon @cheungdaven I changed the typehints to be just `NDArray` due to the API change of the latest nptyping.

multimodal/setup.py
sxjscience: Thanks!

multimodal/setup.py
zhiqiangdon: We can't loose the cap of torchmetrics for now. We need to make some changes in our data preprocessing to loos the cap.

multimodal/setup.py
zhiqiangdon: pytorch-metric-learning seems okay to upgrade.

multimodal/src/autogluon/multimodal/data/process_text.py
zhiqiangdon: `NDArray ` should be fine, but the shape and dtype info are no longer available.

zhiqiangdon (approved):
LGTM!

PR #2824, opened 2023-02-02
[Post 0.7][multimodal] Add 3 Bigdetection Models

FANGAreNotGnu:
Bigdetection: https://github.com/amazon-science/bigdetection


bryanyzhu (commented):
Great, do we evaluate these models in AutoGluon on the validation set of BigDetection dataset? Just want to see if we can get the numbers as in the paper.

FANGAreNotGnu:
Cancel the PR due to there are some new SOTAs available as our best performance model.

PR #2827, opened 2023-02-03
[Post0.7] Integrating FusionModel + SVM with Learner into pipeline.

yongxinw:
*Issue #, if available:*

*Description of changes:*
[WIP] Integrating FusionModel + SVM into pipeline.


multimodal/src/autogluon/multimodal/learners/learner.py
```diff
@@ -0,0 +1,15 @@
+class MultiModalLearner:
```
sxjscience, line 1: Rename the file as `base.py` and use the following:


```python

class BaseLearner(ABC):
    @abstractmethod
    def fit():
        pass

    @abstractmethod
    def predict():
        pass

    @abstractmethod
    def evaluate():
        pass

    @abstractmethod
    def predict_proba():
        pass

    @abstractmethod
    def extract_embedding():
        pass

    @abstractmethod
    def save():
        pass
    
    @abstractmethod
    def load():
        pass

```

multimodal/src/autogluon/multimodal/learners/svm_learner.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+from sklearn.pipeline import make_pipeline
+from sklearn.preprocessing import StandardScaler
+from sklearn.svm import SVC
+
+from typing import List, Optional, Union
+
+
+class FusionSVMClassificationLearner(MultiModalLearner):
```
tonyhoo, line 24: Curious how we integrate with this learner from `MultiModalPredictor` and how we differentiate high shot versus low shot use case?

multimodal/src/autogluon/multimodal/learners/learner.py
yongxinw: okay got it! I actually thought about using ABC but didn't eventually push it through lol. Thanks for the suggestion!

multimodal/src/autogluon/multimodal/learners/svm_learner.py
yongxinw: I am also thinking about this. I thought about using problem_type but it may not be suitable because the few-shot mode is more like another solution to a problem (for example, classification). So using problem_type alone may not be sufficient. I was thinking that a flag or config where user can explicitly declare that they want to use this svm few-shot mode would be nice.

yongxinw: Maybe we can integrate using presets and configs. We can add svm branches in the predictor's `__init__()`, `fit()`, `predict()`, `evaluate()` functions. When user specifically wants to use svm, we will run the corresponding svm branches.

multimodal/src/autogluon/multimodal/learners/learner.py
zhiqiangdon: @suzhoum uses `abstract` as the keyword in fusion model. Do we need to align the keyword (`abstract` or `base`)? See https://github.com/autogluon/autogluon/blob/master/multimodal/src/autogluon/multimodal/models/fusion/base.py#L12.

PR #2828, opened 2023-02-03
[multimodal] Update torchmetrics map

FANGAreNotGnu:
Use latest implementation if `torchmetrics.__version__ > 0.11.0`


bryanyzhu (approved):
Code looks good. One question, after switching to using torchmetrics 0.11.0, will performance on several datasets/tutorials change?

sxjscience:
Shall we just upgrade torchmetrics (make it >=0.11.0?) I think @zhiqiangdon mentioned that there is another issue about `pos_label`

FANGAreNotGnu:
> Shall we just upgrade torchmetrics (make it >=0.11.0?) I think @zhiqiangdon mentioned that there is another issue about `pos_label`

I think 0.11.0 is not ok for map since the bug is fixed in 0.11.1. I update this so that we don't need to force the version >=0.11.1.

FANGAreNotGnu:
> Code looks good. One question, after switching to using torchmetrics 0.11.0, will performance on several datasets/tutorials change?

I tested the two versions of map implementation on two datasets and they performs similarly. I haven't done a speed analysis on heavy task but the latest one is at least not slower than the previous one.

FANGAreNotGnu:
Tried on COCO17 val split. Both implementations output the same result for `checkpoint_name = yolov3_mobilenetv2_320_300e_coco` and in about the same time (0.11.1 takes 340 seconds and previous one takes 352 seconds). @bryanyzhu @sxjscience

sxjscience (approved):
LGTM

sxjscience:
Do we need to directly upgrade torchmetrics so that the logic can be more clean.

PR #2835, opened 2023-02-03
[multimodal] Log warning for multi-gpu realtime prediction

liangfu:
*Issue #, if available:*

I think we don't have a DictCollator in case of realtime prediction, therefore DataParallel here isn't actually functional.

*Description of changes:*

Log warning for multi-gpu realtime prediction

### DIFF
--- a/multimodal/src/autogluon/multimodal/utils/inference.py
+++ b/multimodal/src/autogluon/multimodal/utils/inference.py
@@ -153,7 +153,7 @@ def infer_batch(
     device = torch.device(device_type)
     batch_size = len(batch[next(iter(batch))])
     if 1 < num_gpus <= batch_size:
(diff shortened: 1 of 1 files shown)


sxjscience:
Thanks for the fix!

Our plan should be to remove the usage of data-parallel in our model inference, and just adopt multiprocessing for inference.

sxjscience (approved):
LGTM

sxjscience:
Just to clarify my previous comment.

Our current `.evaluate()` and `.predict()` will use `DataParallel` for scaling up the workloads to multiple GPUs. This is sub-optimal due to the overhead in DataParallel. We can try to refactor the code to directly use multiprocessing for multi-GPU inference.

PR #2836, opened 2023-02-04
[MultiModal] Fusion model inference acceleration with TensorRT

liangfu:
*Issue #, if available:*

*Description of changes:*

This PR brings support to accelerate fusion models with tensorrt.

For a fusion model trained with petfinder dataset, TensorrtExecutionProvider can boost inference speed up to 2.7x faster, comparing to GPU-based realtime prediction with PyTorch.

```
"petfinder",
["numerical_mlp", "categorical_mlp", "timm_image", "hf_text", "fusion_mlp"],
"google/electra-small-discriminator",
"mobilenetv3_small_100"
```

image

### DIFF
--- a/multimodal/src/autogluon/multimodal/models/clip.py
+++ b/multimodal/src/autogluon/multimodal/models/clip.py
@@ -169,6 +169,7 @@ class CLIPForImageText(nn.Module):
                 return_dict=True,
             )
             text_features = self.model.text_projection(text_outputs.pooler_output)  # (b, num_features)
+            last_hidden_state = text_outputs[0]
 
             # normalized features
             text_features = text_features / text_features.norm(dim=-1, keepdim=True)
@@ -177,7 +178,7 @@ class CLIPForImageText(nn.Module):
             text_column_features, text_column_feature_masks = get_column_features(
                 batch=batch,
                 column_name_prefix=self.text_column_prefix,
-                features=self.model.text_projection(text_outputs.last_hidden_state),
+                features=self.model.text_projection(last_hidden_state),
                 valid_lengths=text_valid_length,
                 cls_feature=text_features,
             )
--- a/multimodal/src/autogluon/multimodal/models/clip.py
+++ b/multimodal/src/autogluon/multimodal/models/clip.py
@@ -169,7 +169,6 @@ class CLIPForImageText(nn.Module):
                 return_dict=True,
             )
             text_features = self.model.text_projection(text_outputs.pooler_output)  # (b, num_features)
-            last_hidden_state = text_outputs[0]
 
             # normalized features
             text_features = text_features / text_features.norm(dim=-1, keepdim=True)
@@ -178,7 +177,7 @@ class CLIPForImageText(nn.Module):
             text_column_features, text_column_feature_masks = get_column_features(
                 batch=batch,
                 column_name_prefix=self.text_column_prefix,
-                features=self.model.text_projection(last_hidden_state),
+                features=self.model.text_projection(text_outputs.last_hidden_state),
                 valid_lengths=text_valid_length,
                 cls_feature=text_features,
             )
--- a/multimodal/src/autogluon/multimodal/models/fusion/fusion_mlp.py
+++ b/multimodal/src/autogluon/multimodal/models/fusion/fusion_mlp.py
@@ -120,19 +120,27 @@ class MultimodalFusionMLP(AbstractMultimodalFusionModel):
         self.head_layer_names = [n for n, layer_id in self.name_to_id.items() if layer_id == 0]
 
     @property
+    def input_keys(self):
+        input_keys = []
+        for m in self.model:
+            assert hasattr(m, "input_keys"), f"invalid model {type(m)}, which doesn't have a 'input_keys' attribute"
+            input_keys += m.input_keys
+        return input_keys
+
+    @property
     def label_key(self):
         return f"{self.prefix}_{LABEL}"
 
     def forward(
         self,
-        batch: dict,
+        *args,
     ):
         """
 
         Parameters
         ----------
-        batch
-            A dictionary containing the input mini-batch data. The fusion model doesn't need to
+        *args
+            A list of torch.Tensor(s) containing the input mini-batch data. The fusion model doesn't need to
             directly access the mini-batch data since it aims to fuse the individual models'
             output features.
 
@@ -143,20 +151,24 @@ class MultimodalFusionMLP(AbstractMultimodalFusionModel):
         including the fusion model's.
         """
         multimodal_features = []
-        output = {}
+        multimodal_logits = []
+        offset = 0
         for per_model, per_adapter in zip(self.model, self.adapter):
+            per_model_args = args[offset : offset + len(per_model.input_keys)]
+            batch = dict(zip(per_model.input_keys, per_model_args))
             per_output = run_model(per_model, batch)
             multimodal_features.append(
                 per_adapter(per_output[per_model.prefix][FEATURES].to(per_adapter.weight.dtype))
             )
-            if self.loss_weight is not None:
-                per_output[per_model.prefix].update(
-                    {WEIGHT: torch.tensor(self.loss_weight).to(multimodal_features[0])}
-                )
-                output.update(per_output)
+            multimodal_logits.append(per_output[per_model.prefix][LOGITS])
+            offset += len(per_model.input_keys)
 
         features = self.fusion_mlp(torch.cat(multimodal_features, dim=1))
         logits = self.head(features)
+
+        return features, logits, multimodal_logits
+
+    def get_output_dict(self, features: torch.Tensor, logits: torch.Tensor, multimodal_logits: List[torch.Tensor]):
         fusion_output = {
             self.prefix: {
                 LOGITS: logits,
@@ -164,6 +176,12 @@ class MultimodalFusionMLP(AbstractMultimodalFusionModel):
             }
         }
         if self.loss_weight is not None:
+            output = {}
+            for per_model, per_logits in zip(self.model, multimodal_logits):
+                per_output = {per_model.prefix: {}}
+                per_output[per_model.prefix][WEIGHT] = torch.tensor(self.loss_weight).to(per_logits.dtype)
+                per_output[per_model.prefix][LOGITS] = per_logits
+            output.update(per_output)
             fusion_output[self.prefix].update({WEIGHT: torch.tensor(1.0).to(logits)})
             output.update(fusion_output)
             return output
--- a/multimodal/src/autogluon/multimodal/models/huggingface_text.py
+++ b/multimodal/src/autogluon/multimodal/models/huggingface_text.py
@@ -199,11 +199,10 @@ class HFAutoModelForTextPrediction(nn.Module):
                     input_ids=text_token_ids,
                     attention_mask=text_masks,
                 )
-        last_hidden_state = outputs[0]
         if self.pooling_mode == "cls":
-            pooled_features = last_hidden_state[:, 0, :]
+            pooled_features = outputs.last_hidden_state[:, 0, :]
         elif self.pooling_mode == "mean":
-            pooled_features = (last_hidden_state * text_masks.unsqueeze(-1)).sum(1)
+            pooled_features = (outputs.last_hidden_state * text_masks.unsqueeze(-1)).sum(1)
             sum_mask = text_masks.unsqueeze(-1).sum(1)
             sum_mask = torch.clamp(sum_mask, min=1e-9)
             pooled_features = pooled_features / sum_mask
@@ -211,6 +210,7 @@ class HFAutoModelForTextPrediction(nn.Module):
             raise NotImplementedError(f"Pooling mode={self.pooling_mode} is not supported.")
 
         logits = self.head(pooled_features)
+        last_hidden_state = outputs.last_hidden_state
 
         batch = {
             self.text_token_ids_key: text_token_ids,
--- a/multimodal/src/autogluon/multimodal/models/huggingface_text.py
+++ b/multimodal/src/autogluon/multimodal/models/huggingface_text.py
@@ -1,5 +1,5 @@
 import logging
-from typing import List, Optional, Tuple
+from typing import Dict, List, Optional, Tuple
 
 import torch
 from torch import nn
@@ -158,7 +158,7 @@ class HFAutoModelForTextPrediction(nn.Module):
             Indices of input sequence segments.
         text_valid_length : torch.Tensor
             Valid length of the input text sequence.
-        text_column_names : list of torch.Tensor, optional
+        text_column_names : list of str, optional
             Names of the text columns.
         text_column_indices : list of torch.Tensor, optional
             Start and stop indices of the text columns.
@@ -199,10 +199,11 @@ class HFAutoModelForTextPrediction(nn.Module):
                     input_ids=text_token_ids,
                     attention_mask=text_masks,
                 )
+        last_hidden_state = outputs[0]
         if self.pooling_mode == "cls":
-            pooled_features = outputs.last_hidden_state[:, 0, :]
+            pooled_features = last_hidden_state[:, 0, :]
         elif self.pooling_mode == "mean":
-            pooled_features = (outputs.last_hidden_state * text_masks.unsqueeze(-1)).sum(1)
+            pooled_features = (last_hidden_state * text_masks.unsqueeze(-1)).sum(1)
             sum_mask = text_masks.unsqueeze(-1).sum(1)
             sum_mask = torch.clamp(sum_mask, min=1e-9)
             pooled_features = pooled_features / sum_mask
@@ -210,7 +211,6 @@ class HFAutoModelForTextPrediction(nn.Module):
             raise NotImplementedError(f"Pooling mode={self.pooling_mode} is not supported.")
 
         logits = self.head(pooled_features)
-        last_hidden_state = outputs.last_hidden_state
 
         batch = {
             self.text_token_ids_key: text_token_ids,
@@ -219,7 +219,8 @@ class HFAutoModelForTextPrediction(nn.Module):
         }
         if text_column_names:
             assert len(text_column_names) == len(text_column_indices), "invalid text column inputs"
-            batch.update(**dict(zip(text_column_names, text_column_indices)))
+            for idx, name in enumerate(text_column_names):
+                batch[name] = text_column_indices[idx]
         column_features, column_feature_masks = get_column_features(
             batch=batch,
             column_name_prefix=self.text_column_prefix,
@@ -237,8 +238,8 @@ class HFAutoModelForTextPrediction(nn.Module):
         self,
         pooled_features: torch.Tensor,
         logits: torch.Tensor,
-        column_features: Optional[List[torch.Tensor]] = None,
-        column_feature_masks: Optional[List[torch.Tensor]] = None,
+        column_features: Optional[Dict[str, torch.Tensor]] = None,
+        column_feature_masks: Optional[Dict[str, torch.Tensor]] = None,
     ):
         ret = {COLUMN_FEATURES: {FEATURES: {}, MASKS: {}}}
         if column_features != None:
--- a/multimodal/src/autogluon/multimodal/models/ner_text.py
+++ b/multimodal/src/autogluon/multimodal/models/ner_text.py
@@ -1,5 +1,5 @@
 import logging
-from typing import List, Optional, Tuple
+from typing import Dict, List, Optional, Tuple
 
 import torch
 import torch.nn.functional as F
@@ -134,14 +134,14 @@ class HFAutoModelForNER(HFAutoModelForTextPrediction):
             Mapping the named entities to task specific labels.
         word_offsets : torch.Tensor
             Locations of the named entities.
-        text_column_names : list of torch.Tensor, optional
+        text_column_names : list of str, optional
             Names of the text columns.
         text_column_indices : list of torch.Tensor, optional
             Start and stop indices of the text columns.
 
         Returns
         -------
-            A tuple that contains (sequence_output, logits, logits_label, token_word_mapping, word_offsets, column_features, column_feature_masks)
+            A tuple that contains (last_hidden_state, logits, logits_label, token_word_mapping, word_offsets, column_features, column_feature_masks)
         """
         if self.disable_seg_ids:
             text_segment_ids = None
@@ -164,13 +164,13 @@ class HFAutoModelForNER(HFAutoModelForTextPrediction):
                 attention_mask=text_masks,
             )
 
-        sequence_output = outputs.last_hidden_state
-        batch_size, max_len, feat_dim = sequence_output.shape
+        last_hidden_state = outputs[0]
+        batch_size, max_len, feat_dim = last_hidden_state.shape
         valid_output = torch.zeros(batch_size, max_len, feat_dim, dtype=torch.float32)
 
-        pooled_features = outputs.last_hidden_state[:, 0, :]
+        pooled_features = last_hidden_state[:, 0, :]
 
-        logits = self.head(sequence_output)
+        logits = self.head(last_hidden_state)
 
         logits_label = torch.argmax(F.log_softmax(logits, dim=-1), dim=-1)
 
@@ -185,16 +185,16 @@ class HFAutoModelForNER(HFAutoModelForTextPrediction):
         column_features, column_feature_masks = get_column_features(
             batch=batch,
             column_name_prefix=self.text_column_prefix,
-            features=outputs.last_hidden_state,
+            features=last_hidden_state,
             valid_lengths=text_valid_length,
             cls_feature=pooled_features,
         )
 
         if column_features == {} or column_feature_masks == {}:
-            return sequence_output, logits, logits_label, token_word_mapping, word_offsets
+            return last_hidden_state, logits, logits_label, token_word_mapping, word_offsets
         else:
             return (
-                sequence_output,
+                last_hidden_state,
                 logits,
                 logits_label,
                 token_word_mapping,
@@ -205,13 +205,13 @@ class HFAutoModelForNER(HFAutoModelForTextPrediction):
 
     def get_output_dict(
         self,
-        sequence_output: torch.Tensor,
+        last_hidden_state: torch.Tensor,
         logits: torch.Tensor,
         logits_label: torch.Tensor,
         token_word_mapping: torch.Tensor,
         word_offsets: torch.Tensor,
-        column_features: Optional[List[torch.Tensor]] = None,
-        column_feature_masks: Optional[List[torch.Tensor]] = None,
+        column_features: Optional[Dict[str, torch.Tensor]] = None,
+        column_feature_masks: Optional[Dict[str, torch.Tensor]] = None,
     ):
         ret = {COLUMN_FEATURES: {FEATURES: {}, MASKS: {}}}
         if column_features != None:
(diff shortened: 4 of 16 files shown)


tonyhoo:
Thanks for the change @liangfu. Do you know why preprocessing logics are generally slower in the TRT execution env?

multimodal/src/autogluon/multimodal/utils/export.py
```diff
@@ -73,82 +76,71 @@ def dump_model(self, save_path: Optional[str] = None):
...
+        path
+            The export path of onnx model.
         batch_size
             The batch_size of export model's input.
             Normally the batch_size is a dynamic axis, so we could use a small value for faster export.
         verbose
             verbose flag in torch.onnx.export.
         opset_version
             opset_version flag in torch.onnx.export.
+        truncate_long_and_double: bool, default False
```
tonyhoo, line 103: Do we need to do special handing as well for `bf16`?

multimodal/src/autogluon/multimodal/utils/export.py
```diff
@@ -73,82 +76,71 @@ def dump_model(self, save_path: Optional[str] = None):
...
-                data=data,
-                valid_input=valid_input,
-                onnx_tracing=True,
-                batch_size=batch_size,
-            )
-
-        if not onnx_path:
-            onnx_path = os.path.join(self.path, default_onnx_path)
-
-        device_type = "cuda" if torch.cuda.is_available() else "cpu"
```
tonyhoo, line 129: Wondering why previously we use GPU for `onnx_export`

tonyhoo (commented):
Can you add some accuracy metrics as well just to make sure accuracy doesn't compromise post TRT transformation

multimodal/src/autogluon/multimodal/utils/export.py
liangfu: It's handled in the TensorrtExecutionProvider configuration.
```python
"TensorrtExecutionProvider",
{
    "device_id": 0,
    "trt_max_workspace_size": 2147483648,
    "trt_fp16_enable": True,
},
```

liangfu: truncate_long_and_double here just to convert double -> float, int64 -> int32, since TRT wouldn't support int64 and double data types at all.

multimodal/src/autogluon/multimodal/utils/export.py
liangfu: We actually used cpu instead, see deleted comment below.

To quote
```
# Perform tracing on cpu, since we're facing an error when tracing with cuda device:
#     ERROR: Tensor-valued Constant nodes differed in value across invocations.
#     This often indicates that the tracer has encountered untraceable code.
#     Comparison exception:   The values for attribute 'shape' do not match: torch.Size([]) != torch.Size([384]).
#     from https://github.com/rwightman/pytorch-image-models/blob/3aa31f537d5fbf6be8f1aaf5a36f6bbb4a55a726/timm/models/swin_transformer.py#L112
device = "cpu"
num_gpus = 0
```

multimodal/tests/unittests/others/test_deployment_tensorrt.py
```diff
@@ -0,0 +1,93 @@
...
+    test_df = dataset.test_df.head(batch_size)
+    tail_df = dataset.test_df.tail(batch_size)
+
+    # Prediction with default predictor
+    y_pred = predictor.predict(test_df)
+
+    trt_module = predictor.export_tensorrt(path=model_path, data=tail_df, batch_size=batch_size)
+
+    # To use the TensorRT module for prediction, simply replace the _model in the predictor
+    predictor._model = trt_module
```
zhiqiangdon, line 84: This design seems different from onnx, which uses `MultiModalOnnxPredictor`. Do we need to unify them?

multimodal/src/autogluon/multimodal/models/clip.py
```diff
@@ -169,6 +169,7 @@ def forward(
                 return_dict=True,
             )
             text_features = self.model.text_projection(text_outputs.pooler_output)  # (b, num_features)
+            last_hidden_state = text_outputs[0]
```
zhiqiangdon, line 172: Does `text_outputs.last_hidden_state` not work for TensorRT?

liangfu: This has to be changed, since we are adding the `torchscript=True` argument in 
```
AutoConfig.from_pretrained(checkpoint_name, torchscript=True)
```

Basically, `text_outputs[0]` (with `torchscript=True`) equals to `text_ouputs.last_hidden_state` (with `torchscript=False`).

multimodal/tests/unittests/others/test_deployment_tensorrt.py
liangfu: Yes, I totally agree. I think `simply replace the _model in the predictor` would be better for integration, comparing to the way we were integrating with MultiModalPredictor.

But shall we put the unifying effort in a follow-up PR? It might be too overwhelm to deprecate too many old implementation in a single PR.

liangfu:
> Thanks for the change @liangfu. Do you know why preprocessing logics are generally slower in the TRT execution env?

Good question, I'm also interested to see why this is happening. Therefore, I did a breakdown of the call stack

image

Seems like data_processor is a bit slower when using TRT.

multimodal/src/autogluon/multimodal/models/fusion/fusion_mlp.py
```diff
@@ -119,20 +119,29 @@ def __init__(
         self.name_to_id = self.get_layer_ids()
         self.head_layer_names = [n for n, layer_id in self.name_to_id.items() if layer_id == 0]
 
+    @property
+    def input_keys(self):
+        module_list = self.model
```
zhiqiangdon, line 124: This line seems unused?

multimodal/src/autogluon/multimodal/models/huggingface_text.py
```diff
@@ -219,7 +219,8 @@ def forward(
         }
         if text_column_names:
             assert len(text_column_names) == len(text_column_indices), "invalid text column inputs"
-            batch.update(**dict(zip(text_column_names, text_column_indices)))
+            for idx, name in enumerate(text_column_names):
+                batch[name] = text_column_indices[idx]
```
zhiqiangdon, line 223: Is this change necessary?

liangfu: Yes, tracing can't unpack dict.

multimodal/src/autogluon/multimodal/models/ner_text.py
```diff
@@ -164,11 +164,11 @@ def forward(
                 attention_mask=text_masks,
             )
 
-        sequence_output = outputs.last_hidden_state
+        sequence_output = outputs[0]  # last_hidden_state
```
zhiqiangdon, line 167: How about using `last_hidden_state = outputs[0]` to be consistent with other models?

liangfu: Agree. changed sequence_output -> last_hidden_state.

multimodal/src/autogluon/multimodal/models/fusion/fusion_mlp.py
liangfu: Nice catch. This line has been removed.

multimodal/src/autogluon/multimodal/models/huggingface_text.py
liangfu: Yes, they have the same effect.

multimodal/src/autogluon/multimodal/models/t_few.py
```diff
@@ -234,25 +245,51 @@ def forward(
...
+            return features, logits, target_template_logits, lm_target, column_features, column_feature_masks
+
+    def get_output_dict(
+        self,
+        features: torch.Tensor,
+        logits: torch.Tensor,
+        target_template_logits: torch.Tensor,
+        lm_target: torch.Tensor,
+        column_features: Optional[List[torch.Tensor]] = None,
+        column_feature_masks: Optional[List[torch.Tensor]] = None,
```
zhiqiangdon, line 280: These types should be `Optional[Dict[str, torch.Tensor]]`. Also need to correct the related arguments in other models' `get_output_dict `.

multimodal/src/autogluon/multimodal/models/utils.py
```diff
@@ -558,10 +558,10 @@ def get_hf_config_and_model(
     -------
     A Huggingface config and model.
     """
-    config = AutoConfig.from_pretrained(checkpoint_name)
+    config = AutoConfig.from_pretrained(checkpoint_name, torchscript=True)
 
     if pretrained:
-        model = AutoModel.from_pretrained(checkpoint_name, low_cpu_mem_usage=low_cpu_mem_usage)
+        model = AutoModel.from_pretrained(checkpoint_name, low_cpu_mem_usage=low_cpu_mem_usage, torchscript=True)
```
zhiqiangdon, line 564: Setting `torchscript=True` make all the tied weights in model decoupled, which can cause problems if finetuning the model. See https://huggingface.co/transformers/v3.0.2/torchscript.html#torchscript-flag-and-tied-weights.

multimodal/src/autogluon/multimodal/models/utils.py
```diff
@@ -686,28 +686,45 @@ def update_mmdet_config(key, value, config):
...
+    supported_models = (
+        TimmAutoModelForImagePrediction,
+        HFAutoModelForTextPrediction,
+        MultimodalFusionMLP,
+        TFewModel,
+        OnnxModule,
+    )
     pure_model = model.module if isinstance(model, nn.DataParallel) else model
+    if isinstance(pure_model, OnnxModule):
+        for k in batch:
```
zhiqiangdon, line 706: What possible types can `batch[k].dtype` be when code runs here? Can we say it is always in `[torch.float32, torch.int32, torch.int64]`?

multimodal/tests/unittests/others/test_deployment_tensorrt.py
zhiqiangdon: Follow-up PRs sound good.

multimodal/src/autogluon/multimodal/models/utils.py
liangfu: This is a special handling for converting the data type in token_ids. I'm not sure why the token_ids are in int32, but the inputs are required to be int64.

multimodal/src/autogluon/multimodal/models/utils.py
liangfu: Here is a list of things to assess the potential impact of this change.
1. see how many models are impacted by tied weights, especially hf_text and derived models.
2. evaulate the potential accuracy loss when decouple the layers that have tied weights
3. try to work around the torchscript argument, and try to avoid setting it to True, in order to avoid the issue
4. (less likely but possible) if the impact of tied weight is high, and impact of TRT is relatively slow, while we don't have workaround to avoid this issue, we could potentially abandon the idea of introducing the onnx/trt exporting feature.

multimodal/src/autogluon/multimodal/models/t_few.py
liangfu: Nice catch. BTW, i'm just curious why we didn't catch such mistake in unittests.

multimodal/tests/unittests/others/test_deployment_tensorrt.py
liangfu: Will be updated in PR https://github.com/autogluon/autogluon/pull/2983

multimodal/src/autogluon/multimodal/utils/onnx.py
```diff
@@ -7,99 +7,70 @@
...
-        The predictor's pipeline.
-    config
-        The predictor's config.
+def onnx_get_dynamic_axes(input_keys: List[str]):
+    dynamic_axes = {}
+    for k in input_keys:
+        if "token_ids" in k or "segment_ids" in k:
+            dynamic_axes[k] = {0: "batch_size", 1: "seq_length"}
+        elif "valid_length" in k or k.startswith("numerical") or k.startswith("timm_image"):
+            dynamic_axes[k] = {0: "batch_size"}
```
zhiqiangdon, line 25: The shape of `images` in `timm_image`'s input should be `(b, n, c, h, w)`, where both b and n may be dynamic.

multimodal/src/autogluon/multimodal/utils/export.py
```diff
@@ -160,23 +152,85 @@ def export_onnx(
...
+            onnx_path = os.path.join(path, "model.onnx")
+        dirname = os.path.dirname(os.path.abspath(onnx_path))
+        if not os.path.exists(dirname):
+            os.makedirs(dirname)
 
-        with precision_context, torch.no_grad():
-            traced_model = torch.jit.trace(self._model, input_vec)
-        torch.jit.save(traced_model, "traced_model.pt")
-        traced_model = torch.jit.load("traced_model.pt")
+        input_keys = self._model.input_keys
```
zhiqiangdon, line 175: Consider putting this line at the beginning of this function. I see some places above uses `self._model.input_keys`. Using one variable can make code more organized.

multimodal/src/autogluon/multimodal/utils/export.py
```diff
@@ -160,23 +152,85 @@ def export_onnx(
...
-        input_vec = InputBatch(**batch)
+        keys_to_drop = list(set(batch.keys()).difference(set(self._model.input_keys)))
+        for k in keys_to_drop:
+            batch.pop(k)
+        # Truncate input data types for TensorRT (only support: bool, int32, half, float)
+        if truncate_long_and_double:
+            for k in batch:
+                if batch[k].dtype == torch.int64:
+                    batch[k] = batch[k].to(torch.int32)
+        input_vec = [batch[k] for k in self._model.input_keys]
```
zhiqiangdon, line 163: Could we move all the logic to process/convert `batch` data, including the code to support mixed precision on floating point inputs, to util function `get_processed_batch_for_deployment `?  I think these logic better fits that context.

multimodal/src/autogluon/multimodal/utils/export.py
```diff
@@ -73,82 +76,71 @@ def dump_model(self, save_path: Optional[str] = None):
 
     def export_onnx(
         self,
-        data: Optional[pd.DataFrame] = None,
```
zhiqiangdon, line 76: We may need to improve this example and add more examples about `timm_image`, `hf_text`, and `fusion_mlp`. Since `data` is required, I guess this example no longer works. You can mark this add a TODO item. I think tutorials may also be necessary to teach users.

multimodal/src/autogluon/multimodal/models/utils.py
zhiqiangdon: Since onnx/trt is mainly for inference, how about still using the previous logic that scripts the model immediately before exporting onnx? Setting `torchscript=True` here can impact training, which may require tied weights, especially for the generative models to add.

multimodal/src/autogluon/multimodal/utils/onnx.py
```diff
@@ -7,99 +7,70 @@
 
 logger = logging.getLogger(__name__)
 
+try:
+    import tensorrt  # Unused but required by TensorrtExecutionProvider
+except:
+    logger.warning(
+        "Failed to import tensorrt package. "
+        "onnxruntime would fallback to CUDAExecutionProvider instead of using TensorrtExecutionProvider."
+    )
```
zhiqiangdon, line 16: Is it easy to install `tensorrt` for users? If not, consider making this a lazy import to avoid unnecessary warnings for users who don't need it.

multimodal/src/autogluon/multimodal/models/utils.py
```diff
@@ -686,28 +686,45 @@ def update_mmdet_config(key, value, config):
...
-    supported_models = (TimmAutoModelForImagePrediction, HFAutoModelForTextPrediction)
+    supported_models = (
+        TimmAutoModelForImagePrediction,
+        HFAutoModelForTextPrediction,
+        MultimodalFusionMLP,
+        TFewModel,
+        OnnxModule,
+    )
     pure_model = model.module if isinstance(model, nn.DataParallel) else model
+    if isinstance(pure_model, OnnxModule):
```
zhiqiangdon, line 705: In which case `pure_model ` can be a `OnnxModule ` object? I don't see where `run_model` is called with a `OnnxModule ` object as input.

multimodal/src/autogluon/multimodal/utils/onnx.py
```diff
@@ -7,99 +7,70 @@
...
-            },
-            "timm_image_image_valid_num": {},
-        }
-        default_onnx_path = "./model.onnx"
-        default_batch = None
-    else:
-        raise ValueError(f"ONNX export is not supported in current pipeline {pipeline}")
 
-    return valid_input, dynamic_axes, default_onnx_path, default_batch
+class OnnxModule(object):
```
zhiqiangdon, line 30: Consider adding some docstrings to explain this class. Tensorrt seems only one provider in onnx runtime. But it's unclear how the three providers work. Does onnxruntime check `TensorrtExecutionProvider `, `CUDAExecutionProvider `, and `CPUExecutionProvider ` sequentially to see their availability? Better to use some docstrings to describe the mechanism.

multimodal/src/autogluon/multimodal/models/utils.py
liangfu: For predicting with the TRT for a fusion model,
```python
onnx_module = predictor.export_tensorrt(data=X)
predictor._model = onnx_module
predictor.predict(X)  # pure_model would be OnnxModule here
```

multimodal/src/autogluon/multimodal/utils/onnx.py
liangfu: > Does onnxruntime check TensorrtExecutionProvider , CUDAExecutionProvider , and CPUExecutionProvider  sequentially to see their availability?

Yes, the execution providers are sequentially checked.

> Better to use some docstrings to describe the mechanism.

Sure.

multimodal/src/autogluon/multimodal/utils/onnx.py
liangfu: > Is it easy to install tensorrt for users?

Yes, it's just `pip install tensorrt`.

liangfu: tried lazy import, but didn't work.

multimodal/src/autogluon/multimodal/utils/export.py
```diff
@@ -160,23 +152,85 @@ def export_onnx(
...
+
+        truncate_long_and_double = False
+        onnx_path = self.export_onnx(
+            data=data, path=path, batch_size=batch_size, truncate_long_and_double=truncate_long_and_double
+        )
+
+        logger.info("Loading ONNX file from path {}...".format(onnx_path))
+        onnx_model = onnx.load(onnx_path)
+
+        trt_module = OnnxModule(onnx_model)
```
zhiqiangdon, line 228: The returned module may not use tensorrt. Maybe moving the warning of importing tensorrt inside `OnnxModule ` is better?

liangfu: I tried to move importing tensorrt inside OnnxModule, but that didn't work. (onnxruntime.InferenceSession doesn't compile to use TensorrtExecutionProvider)

multimodal/src/autogluon/multimodal/utils/onnx.py
liangfu: Added docstring for describing the mechanism.

multimodal/src/autogluon/multimodal/models/t_few.py
liangfu: Corrected related arguments in T-Few model as well as other models' get_output_dict function.

multimodal/src/autogluon/multimodal/models/utils.py
liangfu: > how about still using the previous logic that scripts the model immediately before exporting onnx?

Yes, this is a better solution to satisfy both worlds. I will definitely give it a try.

multimodal/src/autogluon/multimodal/utils/onnx.py
liangfu: I would suggest avoid adding image size into dynamic dimensions, because
1. adding an extra dynamic dimension would increase the complexity of model compilation
2. the compiled model with too many dynamic dimensions could be suboptimal in terms of performance

The question becomes do we really need to support dynamic shape in image data? If yes, what is the lower bound and upper bound of the image size.

multimodal/src/autogluon/multimodal/utils/onnx.py
zhiqiangdon: It's kind of weird why lazy import doesn't work. It's fine to keep it here for now. Maybe later dive into the reason.

multimodal/src/autogluon/multimodal/utils/export.py
liangfu: We might need to consider clean up these examples, since they are not tested / verified in contineous-integration. See https://github.com/autogluon/autogluon/issues/2848 

> I think tutorials may also be necessary to teach users.

I'm planning to add a tutorial for using TRT, but the tutorial would come after this PR.

multimodal/src/autogluon/multimodal/models/utils.py
liangfu: torchscript=True argument has been removed, and seems everything is still working without the argument.

multimodal/src/autogluon/multimodal/utils/export.py
liangfu: moved to get_processed_batch_for_deployment function as suggested.

multimodal/src/autogluon/multimodal/utils/export.py
liangfu: placed to the top location, and the local input_keys variable is reused.

multimodal/src/autogluon/multimodal/utils/export.py
liangfu: A tutorial will be added in https://github.com/autogluon/autogluon/pull/2987

multimodal/src/autogluon/multimodal/utils/export.py
```diff
@@ -73,118 +76,136 @@ def dump_model(self, save_path: Optional[str] = None):
...
-                data=data,
-                valid_input=valid_input,
-                onnx_tracing=True,
-                batch_size=batch_size,
-            )
+        # Infer default onnx path, and create parent directory if needed
+        if not path:
+            onnx_path = os.path.join(self.path, "model.onnx")
+        else:
+            onnx_path = os.path.join(path, "model.onnx")
```
zhiqiangdon, line 137: A better way may be

```
if not path:
    path = self.path
onnx_path = os.path.join(path, "model.onnx")
```

multimodal/src/autogluon/multimodal/models/clip.py
zhiqiangdon: Since we don't use `torchscript=True`, do we still need `last_hidden_state = text_outputs[0]`?

multimodal/src/autogluon/multimodal/utils/onnx.py
```diff
@@ -7,99 +7,100 @@
...
+        Returns
+        -------
+        onnx_outputs : list of torch.Tensor
+            A list of torch.Tensor that are outputs of the model.
+        """
+        import torch
+
+        input_dict = {k: args[i].cpu().numpy() for i, k in enumerate(self.input_names)}
+        onnx_outputs = self.sess.run(self.output_names, input_dict)
+        onnx_outputs = onnx_outputs[:3]
```
zhiqiangdon, line 95: Why do we need only the first 3 model outputs? Don't we need them all?

multimodal/tests/unittests/others/test_deployment_tensorrt.py
```diff
@@ -0,0 +1,93 @@
...
+    batch_size = 2
+
+    # Subsample test data for efficient testing, and use a different subset of the dataset for compilation.
+    test_df = dataset.test_df.head(batch_size)
+    tail_df = dataset.test_df.tail(batch_size)
+
+    # Prediction with default predictor
+    y_pred = predictor.predict(test_df)
+
+    trt_module = predictor.export_tensorrt(path=model_path, data=tail_df, batch_size=batch_size)
```
zhiqiangdon, line 81: Currently, `export_onnx` and `export_tensorrt` either returns a path or a module. When a user calls `export_something()`, do users expect return something? Or users just expect the model is saved to disk? If users want to use the saved model, they probably want to load it?

multimodal/src/autogluon/multimodal/utils/export.py
```diff
@@ -219,9 +240,37 @@ def get_processed_batch_for_deployment(
...
+
+        # Move model data to the specified device
+        for key in input_keys:
+            inp = batch[key]
+            # support mixed precision on floating point inputs, and leave integer inputs (for language models) untouched.
+            if inp.dtype.is_floating_point:
+                batch[key] = inp.to(device, dtype=dtype)
+            else:
+                batch[key] = inp.to(device)
+        self._model.to(device)
```
zhiqiangdon, line 262: Do we need `self._model.to(device)` here?

multimodal/src/autogluon/multimodal/utils/onnx.py
liangfu: In a pretrained setting, n should be a fixed number, since input DataFrame should have same number of image columns, isn't it?

multimodal/src/autogluon/multimodal/utils/onnx.py
liangfu: This comes with the undetermined size of output dict for a fusion_mlp model, which contains a tuple of `(features, logits, multimodal_logits)`. The variable `multimodal_logits` contains the logit output from all modalities.

> Don't we need them all?

Good question. In short, `multimodal_logits` are only used for computing loss, not used for inference.

Specifically, `get_output_dict()` would take outputs from `forward()`, but in onnxruntime outputs are flatten, which means the list of tensors in `multimodal_logits` would be merged into other outputs (e.g. features, logits). We kind of lost the information about where did the extra tensors come from.

multimodal/tests/unittests/others/test_deployment_tensorrt.py
liangfu: These are excellent questions.

There are several different usecases:
U1: Some users might expect an ONNX file to be exported, so that they can use the ONNX file where ever they want.
U2: Some users might expect faster inference time with ONNX. They actually don't care much about the details on where is the ONNX file.

In terms of U1, I think the expected output could be either the location of the ONNX file, or the ONNX model itself. This is the way how `export_onnx` is defined.

In terms of U2, we should be able to generate a drop-in replacement of torch.nn.Module, so that integration with existing inference flow would be easy. This is the way how `export_tensorrt` is defined.

But I would agree, we should have a better name for these public APIs.

multimodal/src/autogluon/multimodal/models/clip.py
liangfu: undid this changed.

multimodal/src/autogluon/multimodal/utils/export.py
liangfu: Updated as suggested.

zhiqiangdon (approved):
LGTM! Thanks for supporting Tensorrt!

PR #2837, opened 2023-02-04
[multimodal] Refine detection evaluation output

FANGAreNotGnu:
Add constants for detection evaluation metrics and parsing the output for pycocotools output. Naming follows torchmetric.

### DIFF
--- a/multimodal/src/autogluon/multimodal/constants.py
+++ b/multimodal/src/autogluon/multimodal/constants.py
@@ -76,7 +76,7 @@ MAR_10 = "mar_10"
 MAR_100 = "mar_100"
 MAR_SMALL = "mar_small"
 MAR_MEDIUM = "mar_medium"
-MAR_LARGE = "mar_large"
+MAR_LARGE = "mar_large" 
 
 # Metric
 MEAN_AVERAGE_PRECISION = "mean_average_precision"
--- a/multimodal/src/autogluon/multimodal/constants.py
+++ b/multimodal/src/autogluon/multimodal/constants.py
@@ -64,8 +64,21 @@ ROIS = "rois"
 SCORE = "score"
 LOGIT_SCALE = "logit_scale"
 
-# Metric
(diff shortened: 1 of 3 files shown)


multimodal/src/autogluon/multimodal/constants.py
```diff
@@ -64,8 +64,21 @@
 SCORE = "score"
 LOGIT_SCALE = "logit_scale"
 
-# Metric
+# Metric for Object Detection
 MAP = "map"
```
bryanyzhu, line 68: One question, MAP standards for mean average precision, will it be confusing to other people when they see MEAN_AVERAGE_PRECISION in line 82?

FANGAreNotGnu: Good point, edited the code to make MEAN_AVERAGE_PRECISION also work (=MAP).

multimodal/src/autogluon/multimodal/constants.py
```diff
@@ -64,9 +64,37 @@
...
+MAP_SMALL = "map_small"
+MAP_MEDIUM = "map_medium"
+MAP_LARGE = "map_large"
+MAR_1 = "mar_1"
+MAR_10 = "mar_10"
+MAR_100 = "mar_100"
+MAR_SMALL = "mar_small"
+MAR_MEDIUM = "mar_medium"
+MAR_LARGE = "mar_large"
+DETECTION_METRICS = [
```
zhiqiangdon, line 81: Do we need tests for object detection metrics?

zhiqiangdon (approved):
LGTM!

multimodal/src/autogluon/multimodal/constants.py
FANGAreNotGnu: Oh sorry for noticing this after the merge. I'll add the test to it in an existing one in next pr.

PR #2850, opened 2023-02-06
FusionModel + SVM as  `FewShotSVMPredictor`

yongxinw:
*Issue #, if available:*

*Description of changes:*
Integrating FusionModel + SVM as a utility `FewShotSVMPredictor` class. The way to use this predictor is the same as in `MultiModalPredictor`. See below:

```
predictor = FewShotSVMPredictor(
                    label="label",
                    problem_type="feature_extraction",
                    eval_metric="acc"
                )
predictor.fit(train_df)
preds = predictor.predict(data=test_df)
result = predictor.evaluate(test_df, metrics=["acc", "macro_f1"])
```


.gitignore
```diff
@@ -136,7 +136,7 @@ tabular/src/autogluon/tabular/version.py
 text/src/autogluon/text/version.py
 multimodal/src/autogluon/multimodal/version.py
 eda/src/autogluon/eda/version.py
-
+experiments/
```
tonyhoo, line 139: Is this folder newly introduced by this PR?

multimodal/src/autogluon/multimodal/utils/few_shot_learning.py
```diff
@@ -0,0 +1,116 @@
...
+    try_to_infer_pos_label,
+)
+
+
+class FewShotSVMPredictor:
+    def __init__(
+        self,
+        label: Optional[str] = None,
+        hyperparameters: dict = None,
+        problem_type: str = None,
```
sxjscience, line 24: We can just emit `problem_type`.

multimodal/src/autogluon/multimodal/utils/few_shot_learning.py
```diff
@@ -0,0 +1,116 @@
...
+        assert self._fit_called, ".fit() is not invoked. Please consider calling .fit() before .predict()"
+        features = self.extract_embedding(data)
+
+        preds = self.clf.predict(features)
+        if as_pandas:
+            preds = self._automm_predictor._as_pandas(data=data, to_be_converted=preds)
+        return preds
+
+    def extract_embedding(self, data: pd.DataFrame):
+        features = self._automm_predictor.extract_embedding(data)
```
tonyhoo, line 54: Do we need to fit/finetune the predictor before we can extract embeddings?

multimodal/src/autogluon/multimodal/utils/few_shot_learning.py
```diff
@@ -0,0 +1,116 @@
...
+
+        if return_pred:
+            return results, self._automm_predictor._as_pandas(data=data, to_be_converted=preds)
+        return results
+
+    def save(self):
+        pass
+
+    def load(self):
+        pass
```
tonyhoo, line 116: I think it would be better to support both as users shall be able to save and load it via different runtimes (e.g AWS sagemaker training and inference)

sxjscience: We can call the `.save()` and `.load()` from the MultiModalPredictor it used.

.gitignore
yongxinw: oh I was running some experiments with the code and thus had this folder within the repo. Should I remove it entirely?

multimodal/src/autogluon/multimodal/utils/few_shot_learning.py
yongxinw: will do!

multimodal/src/autogluon/multimodal/utils/few_shot_learning.py
yongxinw: right. I actually saved the svm model but not the MultiModalPredictor because I think since we are not making any changes to the MultiModalPredictor, users could just load it from `__init__`.

.gitignore
tonyhoo: If it is not part of repo and bother other users, we can remove it for now

multimodal/src/autogluon/multimodal/utils/few_shot_learning.py
yongxinw: I think we could, but since this is targeting few shot scenarios, if the users finetune the large model on the fewshot data and then apply svm, the results might not be good. That being said tho, do we want to allow users to finetune the `MultiModalPredictor` inside the `FewShotSVMPredictor` if they have a larger dataset? I feel like if they have a larger dataset they could directly use `MultiModalPredictor` for finetuning and fewshot svm solution might be a suboptimal solution for them.

.gitignore
yongxinw: Sure!

multimodal/src/autogluon/multimodal/utils/few_shot_learning.py
tonyhoo: Got it. Thanks for clarification

yongxinw:
Just verified for image classification. This implementation also works.

zhiqiangdon (commented):
1. Do we want to expose `FewShotSVMPredictor ` to users, i.e., letting `from autogluon.multimodal import FewShotSVMPredictor `?

zhiqiangdon (commented):
PR #2827 seems to have a similar functionality. What is the consideration of having 2 PRs of doing the same thing?

multimodal/src/autogluon/multimodal/utils/few_shot_learning.py
```diff
@@ -0,0 +1,180 @@
...
+            will be created in the working directory to store all models.
+            Note: To call `fit()` twice and save all results of each fit,
+            you must specify different `path` locations or don't specify `path` at all.
+            Otherwise files from first `fit()` will be overwritten by second `fit()`.
+
+        """
+        self._automm_predictor = MultiModalPredictor(
+            label=label,
+            hyperparameters=hyperparameters,
+            problem_type=FEATURE_EXTRACTION,
```
zhiqiangdon, line 61: `FEATURE_EXTRACTION ` is for text only now. Do we also need to handle image data?

multimodal/src/autogluon/multimodal/utils/few_shot_learning.py
```diff
@@ -0,0 +1,180 @@
...
+            warn_if_exist=False,
+            fit_called=self._fit_called,
+        )
+        params = self.clf.get_params()
+        logger.info(f"Saving into {os.path.join(self._save_path, 'svm_model.pkl')}")
+        with open(os.path.join(self._save_path, "svm_model.pkl"), "wb") as fp:
+            pickle.dump(params, fp)
+        return self._save_path
+
+    def load(self, path: Optional[str] = None):
```
zhiqiangdon, line 169: Based on the current implementation, users need to init a predictor first and then call predictor.load(path), which is different from MultiModalPredictor's load().

sxjscience: Try to make it a classmethod:

```
@classmethod
def load(cls, ...)
```

yongxinw: Right. Lemme see how I can fix that.

yongxinw: This is fixed but the CI has an issue with memory. Might need to push and merge tomorrow

yongxinw:
> 1. Do we want to expose `FewShotSVMPredictor ` to users, i.e., letting `from autogluon.multimodal import FewShotSVMPredictor `?

I think we'll leave it as a util function for now. We can design our learner after 0.7 and expose that to the users.

sxjscience (approved):
LGTM overall. We can refactor the source code later

PR #2851, opened 2023-02-07
Fix time_elapsed being 0

yinweisu:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/tabular/src/autogluon/tabular/models/rf/rf_model.py
+++ b/tabular/src/autogluon/tabular/models/rf/rf_model.py
@@ -199,7 +199,7 @@ class RFModel(AbstractModel):
                     model = model_cls(**params)
             model = model.fit(X, y, sample_weight=sample_weight)
             if (i == 0) and (len(n_estimator_increments) > 1):
-                time_elapsed = time.time() - time_train_start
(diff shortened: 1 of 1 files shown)


tabular/src/autogluon/tabular/models/rf/rf_model.py
```diff
@@ -199,7 +199,7 @@ def _fit(self,
                     model = model_cls(**params)
             model = model.fit(X, y, sample_weight=sample_weight)
             if (i == 0) and (len(n_estimator_increments) > 1):
-                time_elapsed = time.time() - time_train_start
+                time_elapsed = min(time.time() - time_train_start, 0.001)  # avoid it being too small and being truncated to 0
```
liangfu, line 202: min(0, 0.001) would still go to 0, consider max() instead ?

liangfu: or, a better solution could be `xxx/time_elapsed` -> `xxx/(time_elapsed+1e-6)`

yinweisu: oh... what I was doing...

liangfu (approved):
LGTM

PR #2854, opened 2023-02-07
[eda]  API docs

gradientsky:
*Description of changes:*
- Added API doc pages
- link fixes
- docstrings fixes

### DIFF
--- a/eda/src/autogluon/eda/analysis/interaction.py
+++ b/eda/src/autogluon/eda/analysis/interaction.py
@@ -70,7 +70,6 @@ class Correlation(AbstractAnalysis):
 
     See Also
     --------
-    `phik <https://github.com/KaveIO/PhiK>`_ documentation
     :py:class:`~autogluon.eda.analysis.transform.ApplyFeatureGenerator`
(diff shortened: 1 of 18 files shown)


eda/src/autogluon/eda/analysis/interaction.py
```diff
@@ -70,7 +70,6 @@ class Correlation(AbstractAnalysis):
 
     See Also
     --------
-    `phik <https://github.com/KaveIO/PhiK>`_ documentation
```
gradientsky, line 73: Sphinx doesn't like links in See Also section - throws errors, can't compile.

eda/src/autogluon/eda/analysis/model.py
```diff
@@ -59,37 +59,9 @@ class AutoGluonModelQuickFit(AbstractAnalysis):
         wrapped analyses; these will receive sampled `args` during `fit` call
     kwargs
 
-    Examples
```
gradientsky, line 62: There's example above, this pars should not been added here.

PR #2858, opened 2023-02-07
Update version cap of lightning in timeseries

sxjscience:
*Issue #, if available:*

https://github.com/autogluon/autogluon/issues/2830

*Description of changes:*

- [x] Update version cap of lightning in timeseries to be the same as multimodal

### DIFF
--- a/timeseries/setup.py
+++ b/timeseries/setup.py
@@ -30,7 +30,7 @@ install_requires = [
     "statsmodels~=0.13.0",
     "gluonts~=0.12.0",
     "torch>=1.9,<1.14",
-    "pytorch-lightning>=1.7.4,<1.9.0",
+    "pytorch-lightning>=1.7.4,<1.10.0",
     "networkx",  # version range defined in `core/_setup_utils.py`
     "statsforecast==1.4.0",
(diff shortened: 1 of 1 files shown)


timeseries/setup.py
```diff
@@ -30,7 +30,7 @@
     "statsmodels~=0.13.0",
     "gluonts~=0.12.0",
     "torch>=1.9,<1.14",
-    "pytorch-lightning>=1.7.4,<1.9.0",
+    "pytorch-lightning>=1.7.4,<1.10.0",
```
Innixma, line 33: would recommend the following range:

```python3
"pytorch-lightning>=1.8,<1.10",
```

This way we align exactly with multimodal, unless we want for some reason timeseries exclusively to support 1.7.4, but only if installed without `autogluon.multimodal`. @canerturkmen please advise on what you'd prefer.

canerturkmen:
Hey folks, thanks for raising this. I'll run regression tests today on this change isolated to see if we're doing OK.

canerturkmen:
We appear to have no regressions with `pytorch-lightning==1.9`. Thank you @sxjscience !

canerturkmen (approved):

PR #2860, opened 2023-02-07
Focal loss

yongxinw:
*Issue #, if available:*

*Description of changes:*


multimodal/tests/unittests/others/test_classification.py
```diff
@@ -20,3 +22,38 @@ def test_classification_str_list_input():
     predictions_str = predictor.predict(image_path)
     predictions_list1 = predictor.predict([image_path])
     predictions_list10 = predictor.predict([image_path] * 10)
+
+
+@pytest.mark.parametrize(
+    "checkpoint_name",
+    [
+        "swin_base_patch4_window7_224",
```
zhiqiangdon, line 30: Consider using a smaller backbone, e.g., `swin_tiny_patch4_window7_224`.

multimodal/tests/hf_model_list.yaml
```diff
@@ -24,6 +24,10 @@ others:
 - sentence-transformers/all-mpnet-base-v2
 - sentence-transformers/msmarco-MiniLM-L-12-v3
 - xlm-roberta-base
+- swin_base_patch4_window7_224
+- yolov3_mobilenetv2_320_300e_coco
+- faster_rcnn_r50_fpn_2x_coco
+- detr_r50_8x2_150e_coco
```
zhiqiangdon, line 30: This file is only huggingface models. No need to add these models here.

multimodal/src/autogluon/multimodal/optimization/losses.py
```diff
@@ -272,3 +272,71 @@ def forward(self, features_a, features_b, logit_scale, rank=0, world_size=1):
 
         total_loss = (F.cross_entropy(logits_per_a, labels) + F.cross_entropy(logits_per_b, labels)) / 2
         return total_loss
+
+
+class FocalLoss(nn.Module):
```
zhiqiangdon, line 277: Can we use the torchvision's implementation?https://pytorch.org/vision/main/generated/torchvision.ops.sigmoid_focal_loss.html

zhiqiangdon: We have torchvision installed.

yongxinw: oh that would be much nicer

multimodal/tests/hf_model_list.yaml
yongxinw: oh right. I will delete them.

multimodal/tests/unittests/others/test_classification.py
yongxinw: ok!

multimodal/src/autogluon/multimodal/optimization/losses.py
```diff
@@ -272,3 +272,71 @@ def forward(self, features_a, features_b, logit_scale, rank=0, world_size=1):
...
+                "mean", "sum"
+        eps
+            epsilon for numerical stability
+        """
+        super(FocalLoss, self).__init__()
+        self.alpha = alpha
+        self.gamma = gamma
+        self.reduction = reduction
+        self.eps = eps
+        self.NLLLoss = nn.NLLLoss(weight=alpha, reduction="none")
```
zhiqiangdon, line 311: `self.nll_loss` should be better than `self.NLLLoss`.

multimodal/src/autogluon/multimodal/optimization/losses.py
yongxinw: is the torchvision implementation for binary classification? I saw in their source code
```
p = torch.sigmoid(inputs)
    ce_loss = F.binary_cross_entropy_with_logits(inputs, targets, reduction="none")
    p_t = p * targets + (1 - p) * (1 - targets)
    loss = ce_loss * ((1 - p_t) ** gamma)

    if alpha >= 0:
        alpha_t = alpha * targets + (1 - alpha) * (1 - targets)
        loss = alpha_t * loss
```

multimodal/src/autogluon/multimodal/configs/optimization/adamw.yaml
```diff
@@ -38,3 +38,7 @@ optimization:
     "auto" # The replaced loss for regression. Can only support loss function in torch.nn.
     # example
     # "BCEWithLogitsLoss" or "nn.BCEWithLogitsloss"
+  focal_loss:
+    alpha: null
+    gamma: 2.0
+    reduction: "mean"
```
zhiqiangdon, line 44: \newline is missing at the end of file.

multimodal/src/autogluon/multimodal/optimization/utils.py
```diff
@@ -99,7 +97,11 @@ def get_loss_func(
         if mixup_active:
             loss_func = SoftTargetCrossEntropy()
         else:
-            loss_func = nn.CrossEntropyLoss()
+            if loss_func_name is not None and loss_func_name.lower() == "focal_loss":
+                alpha, gamma, reduction = get_focal_loss_params(loss_config)
+                loss_func = FocalLoss(alpha=alpha, gamma=gamma, reduction=reduction)
```
zhiqiangdon, line 102: `get_focal_loss_params() ` seems unnecessary. A more concise way may be like this:

```
loss_func = FocalLoss(
    alpha=OmegaConf.select(config, "focal_loss.alpha"),
    gamma=OmegaConf.select(config, "focal_loss.gamma"),
    reduction=OmegaConf.select(config, "focal_loss.reduction"),
)
```

multimodal/src/autogluon/multimodal/optimization/utils.py
```diff
@@ -78,6 +75,7 @@ def get_loss_func(
     problem_type: str,
     mixup_active: bool,
     loss_func_name: Optional[str] = None,
+    loss_config: Optional[DictConfig] = None,
```
zhiqiangdon, line 78: `loss_config` is not a good name since it contains more than name says. How about just calling it `config` and only assigning it the `predictor._config.optimization`? Also, add a docstring to explain `config`, i.e., the optimization config.

multimodal/src/autogluon/multimodal/optimization/utils.py
yongxinw: oh so if the `focal_loss` is written in the config yaml file, it will be found by OmegaConf.select no matter what the user inputs?

I added the sanity checks just in case users delete the `focal_loss` attribute in the config yaml.

zhiqiangdon: I think OmegaConf.select can handle the missing keys.

multimodal/src/autogluon/multimodal/optimization/losses.py
zhiqiangdon: I find this post had a discussion on the binary/multiclass cases. 
https://github.com/pytorch/vision/issues/3250

yongxinw: oh okay. I feel like based on the discussion we'll still need to modify the code manually?

multimodal/src/autogluon/multimodal/optimization/utils.py
```diff
@@ -99,7 +101,25 @@ def get_loss_func(
...
+                    gamma = OmegaConf.select(config, "focal_loss.gamma")
+                else:
+                    gamma = 2.0
+
+                if OmegaConf.select(config, "focal_loss.reduction"):
+                    reduction = OmegaConf.select(config, "focal_loss.reduction")
+                else:
+                    reduction = "mean"
+
+                loss_func = FocalLoss(alpha=alpha, gamma=gamma, reduction=reduction)
```
zhiqiangdon, line 120: How about changing to this? I think it's more concise.

```
loss_func = FocalLoss(
    alpha=OmegaConf.select(config, "focal_loss.alpha"),
    gamma=OmegaConf.select(config, "focal_loss.gamma", default=2.0),
    reduction=OmegaConf.select(config, "focal_loss.reduction", default="mean"),
)
```

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -1081,11 +1078,11 @@ def _fit(
                 "The per_gpu_batch_size should be >1 and even for reasonable operation",
                 UserWarning,
             )
-
         loss_func = get_loss_func(
             problem_type=self._problem_type,
             mixup_active=mixup_active,
             loss_func_name=OmegaConf.select(config, "optimization.loss_function"),
+            config=OmegaConf.select(config, "optimization"),
```
zhiqiangdon, line 1085: We can directly use `config.optimization` since we can assume the `optimization` key always exists in `config`. The `select()` function is mainly for backward compatibility.

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -2634,6 +2632,7 @@ def load(
             problem_type=predictor._problem_type,
             mixup_active=False,
             loss_func_name=OmegaConf.select(predictor._config, "optimization.loss_function"),
+            config=OmegaConf.select(predictor._config, "optimization"),
```
zhiqiangdon, line 2635: Ditto. `predictor._config.optimization` should be okay.

multimodal/src/autogluon/multimodal/optimization/utils.py
yongxinw: Oh that's better

multimodal/src/autogluon/multimodal/optimization/losses.py
zhiqiangdon: Yes, we can keep this implementation here.

multimodal/src/autogluon/multimodal/optimization/utils.py
yongxinw: the issue is that for alpha, if it is not none, then we have to cast it with `torch.tensor(...)`. Otherwise, we don't cast it and assign `alpha = None`. I can fit this in one line but it would be a little long.

yongxinw: like this: 

```
loss_func = FocalLoss(
    alpha=torch.tensor(OmegaConf.select(config, "focal_loss.alpha")) if OmegaConf.select(config, "focal_loss.alpha") is not None else None,
...
)
```
```

Comment (yongxinw) [reply multimodal/src/autogluon/multimodal/optimization/utils.py:120]: Can I abbreviate the `OmegaConf.select` here to like `config.focal_loss.alpha`?

File: multimodal/src/autogluon/multimodal/optimization/losses.py [hunk 11]
```diff
@@ -272,3 +272,71 @@ def forward(self, features_a, features_b, logit_scale, rank=0, world_size=1):
...
+            # (N, C, d1, d2, ..., dK) --> (N * d1 * ... * dK, C)
+            num_class = input.shape[1]
+            input = input.permute(0, *range(2, input.ndim), 1).reshape(-1, num_class)
+            # (N, d1, d2, ..., dK) --> (N * d1 * ... * dK,)
+            target = target.view(-1)
+
+        pt = F.softmax(input, dim=-1) + self.eps
+
+        # -alpha_t * log(pt) term
+        log_p = torch.log(pt)
```
zhiqiangdon, line 326: Can we use `log_softmax`? It should be more stable doing `log` and `softmax` separately

multimodal/src/autogluon/multimodal/optimization/utils.py
zhiqiangdon: We need to use `select` here since this is for backward compatibility, i.e., you have a new key added.

zhiqiangdon: I think it's better to put `torch.tensor(alpha)` into the forward function of `FocalLoss`.

multimodal/src/autogluon/multimodal/optimization/losses.py
```diff
@@ -272,3 +272,71 @@ def forward(self, features_a, features_b, logit_scale, rank=0, world_size=1):
...
+        gamma
+            the focal parameter for calculating weights on easy/hard samples
+        reduction
+            the reduction to apply to the final loss output. Default: "mean". Options:
+                "mean", "sum"
+        eps
+            epsilon for numerical stability
+        """
+        super(FocalLoss, self).__init__()
+        self.alpha = alpha
```
zhiqiangdon, line 307: I don't see `self.alpha` is used.

multimodal/src/autogluon/multimodal/optimization/utils.py
yongxinw: ok that makes sense.

multimodal/src/autogluon/multimodal/optimization/losses.py
yongxinw: oh I can actually delete `self.alpha`

multimodal/src/autogluon/multimodal/optimization/losses.py
```diff
@@ -272,3 +272,73 @@ def forward(self, features_a, features_b, logit_scale, rank=0, world_size=1):
...
+        pt = F.softmax(input, dim=-1)
+
+        # -alpha_t * log(pt) term
+        log_p = torch.log_softmax(input, dim=-1)
+        ce = self.nll_loss(log_p, target)
+
+        # (1 - pt)^gamma term
+        all_rows = torch.arange(input.shape[0])
+        pt = pt[all_rows, target]
+        focal_term = (1 - pt) ** self.gamma
```
zhiqiangdon, line 334: Is the `exp()` operation missing? See https://github.com/AdeelH/pytorch-multi-class-focal-loss/blob/master/focal_loss.py#L82

yongxinw: oh the author used a log_pt = log_softmax(input), and then computes pt = log_pt.exp(). I was directly computing pt = softmax(input), trying to avoid numerical stability issues during the exp after log_softmax if any.

zhiqiangdon (approved):
LGTM!

Issue #2831, opened 2023-02-03
Update autogluon.timeseries dependency version ranges


Innixma:
Current autogluon.timeseries dependencies:

```python3
install_requires = [
    # version ranges added in ag.get_dependency_version_ranges()
    "joblib~=1.1",
    "numpy",  # version range defined in `core/_setup_utils.py`
    "scipy",  # version range defined in `core/_setup_utils.py`
    "pandas",  # version range defined in `core/_setup_utils.py`
    "statsmodels~=0.13.0",
    "gluonts~=0.12.0",
    "torch>=1.9,<1.14",
    "pytorch-lightning>=1.7.4,<1.9.0",
    "networkx",  # version range defined in `core/_setup_utils.py`
    "statsforecast==1.4.0",
    "tqdm",  # version range defined in `core/_setup_utils.py`
    "ujson",  # needed to silence GluonTS warning
    f"autogluon.core[raytune]=={version}",
    f"autogluon.common=={version}",
    f"autogluon.tabular[catboost,lightgbm,xgboost]=={version}",
]
```

Filtering to relevant timeseries specific dependencies:

```python3
install_requires = [
    "joblib~=1.1",
    "statsmodels~=0.13.0",
    "gluonts~=0.12.0",
    "pytorch-lightning>=1.7.4,<1.9.0",
    "statsforecast==1.4.0",
    "ujson",  # needed to silence GluonTS warning
]
```

The following issues are present and should be fixed before v0.7 release:

- [x] joblib lacks lower and upper version range cap
- [x] statsmodels lacks lower and upper version range cap
- [x] gluonts lacks lower and upper version range cap
- [x] #2830 
- [x] statsforecast is hard pinned to a specific micro release
- [x] ujson lacks lower and upper version range cap

Additionally for the optional dependencies:

```python3
"pmdarima~=1.8.2",
"tbats~=1.1",
```

- [x] pmdarima lacks lower and upper version range cap
- [x] tbats lacks lower and upper version range cap

Innixma:
Note: I misunderstood what `~=` meant, and thought it meant `!=`. Thanks @canerturkmen for making me aware.

Innixma:
Dependencies have been updated in #2875

PR #2875, opened 2023-02-09
[timeseries] Fix dependency version ranges

canerturkmen:
*Issue #, if available:*

#2831 

*Description of changes:*

Fixing dependency version ranges in line with #2831, getting rid of the `~=` notation. Note that we install `ujson` just to silence GluonTS warnings, but we cap the version to <6 nevertheless. Changes were tested with regression tests.


tonyhoo (approved):

Innixma (approved):
LGTM! Thanks for the updates!

PR #2890, opened 2023-02-10
Tabular: Switch to use multimodal fusion

Innixma:
*Issue #, if available:*

*Description of changes:*
- Switch Tabular's `'multimodal'` hyperparameter preset to use 1 AutoMM model rather than 2 models (one for text+tabular and one for vision).
- This leverages the benefits of AutoMM's multimodal fusion accuracy enhancements when text, tabular, and images are present in a single dataset.

### DIFF
--- a/tabular/tests/unittests/models/test_automm.py
+++ b/tabular/tests/unittests/models/test_automm.py
@@ -3,7 +3,6 @@ import pytest
 
 @pytest.mark.gpu
 def test_automm_sts(fit_helper):
-    pytest.skip("Temporary skip the unittest")
     fit_args = dict(
         hyperparameters={'AG_AUTOMM': {}},
     )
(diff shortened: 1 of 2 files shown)


sxjscience (approved):
LGTM!

tabular/tests/unittests/models/test_automm.py
```diff
@@ -3,7 +3,6 @@
 
 @pytest.mark.gpu
 def test_automm_sts(fit_helper):
-    pytest.skip("Temporary skip the unittest")
```
tonyhoo, line 6: Why we skip it?

Innixma: We skipped this test previously because AutoMM was unstable in unit tests and we weren't actually using AutoMM in any Tabular config. It is no longer unstable and now with this switch we are directly using AutoMM in the multimodal tabular config, so we should enable the test. (Technically it should be have been enabled earlier, but wasn't)

tonyhoo (approved):

PR #2909, opened 2023-02-14
Accelerate FastAI Preprocessing + Fix TabularNN time_limit

Innixma:
*Issue #, if available:*

*Description of changes:*
- Accelerate FastAI Preprocessing
- For datasets with many columns (>1000), this speeds up FastAI preprocessing by **>300**x when using `pandas==1.5.3`
- While the speedup is most large using `pandas==1.5.3` due to pandas having major slowdowns in several functions, it is also major using older pandas versions that are not as slow (instead of 300x faster, it is 30x faster).
- This speedup can result in training time going from 1000 seconds -> 25 seconds, and inference time from 70 seconds -> 0.13 seconds. (robert dataset)
- This PR additionally fixes another major bug: TabularNN time estimates were off by a factor of 128x, causing many situations where TabularNN will skip training due to thinking it will run out of time when time_limit is specified. This bug was introduced in #2395 and effects releases v0.6.1 and v0.6.2. The reason is because `len(train_dataloader)` switched from being `train_rows/batch_size` to `train_rows`.

TODO:

- [x] Benchmark
- [ ] Create minimal reproducible example using only pandas and numpy, submit performance regression issue to pandas

### DIFF
--- a/tabular/src/autogluon/tabular/models/fastainn/tabular_nn_fastai.py
+++ b/tabular/src/autogluon/tabular/models/fastainn/tabular_nn_fastai.py
@@ -76,6 +76,7 @@ class NNFastAiTabularModel(AbstractModel):
         self.cat_columns = None
         self.cont_columns = None
         self.columns_fills = None
+        self._columns_fills_names = None
         self.procs = None
         self.y_scaler = None
         self._cont_normalization = None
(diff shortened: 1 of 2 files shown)


gradientsky (approved):

liangfu (approved):
Looks awesome! 

FastAI model was the slowest model at inference time, and now it's not.

image

Innixma:
Benchmark result looks good! Pareto dominates mainline (2022_02_12 is mainline, 2022_02_14 is this PR): 

autogluon-benchmark_–_run_check_inference_speed_py__autogluon-utils_

Issue #2903, opened 2023-02-11
multimodal v0.7 conda package can't support osx because 'from ray import tune'


giswqs:
- [X] I have checked that this bug exists on the latest stable version of AutoGluon
- [X] and/or I have checked that this bug exists on the latest mainline of AutoGluon via source installation

**Describe the bug**
The `ray` conda-forge package currently does not support osx. See https://github.com/conda-forge/ray-packages-feedstock/pull/80

`autogluon.multimodal` v0.7.0 introduces `from ray import tune` in [presets.py](https://github.com/autogluon/autogluon/blob/master/multimodal/src/autogluon/multimodal/presets.py#L3), which is causing issues for the `autogluon.multimodal` conda-forge osx build. See [here](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=656907&view=logs&j=a8dee1cf-d17f-5cb0-b53f-4a93dfdbfb0c&t=d9dbf137-9a7f-5666-c60c-9e1b2fc31445&l=1054). 

https://github.com/conda-forge/autogluon.multimodal-feedstock/pull/17

**Expected behavior**
We should try to make `import autogluon.multimodal` work across platforms.

**To Reproduce**
`import autogluon.multimodal` 

**Screenshots**
![image](https://user-images.githubusercontent.com/5016453/218237252-e7415ffa-1239-494d-947a-360e46cc98a5.png)

**Installed Versions**
Which version of AutoGluon are you are using?  
The latest commit

Innixma:
@zhiqiangdon can we instead use the logic from `autogluon.core.space` to define the ranges, and only import ray when actually starting HPO?

giswqs:
Currently, the `ray` dependency is included in `autogluon.core`. The conda-forge package will skip `ray` installation on osx. In v0.7, the `ray` dependency is kind of hardcoded in the source code because of `from ray import tune`, which will fail the conda-forge osx build becuase `ray` can no longer be skipped.

https://github.com/conda-forge/autogluon.core-feedstock/blob/main/recipe/meta.yaml#L48
![image](https://user-images.githubusercontent.com/5016453/218238314-ffa6f273-4923-4b84-9b68-5b9f6db1f0a5.png)

zhiqiangdon:
Yes, we can do lazy import.

Innixma:
> Yes, we can do lazy import.

Great! Please also confirm that running without HPO will work when ray isn't installed

[...]

PR #2916, opened 2023-02-14
[AutoMM] Lazy import ray

zhiqiangdon:
*Issue #, if available:*
#2903 

*Description of changes:*
Lazy import ray.

### DIFF
--- a/multimodal/src/autogluon/multimodal/presets.py
+++ b/multimodal/src/autogluon/multimodal/presets.py
@@ -85,6 +85,7 @@ def default(presets: str = DEFAULT):
     if presets in [HIGH_QUALITY, DEFAULT]:
         if use_hpo:
             from ray import tune
+
             hyperparameters.update(
                 {
                     "model.hf_text.checkpoint_name": tune.choice(
@@ -117,6 +118,7 @@ def default(presets: str = DEFAULT):
     elif presets == MEDIUM_QUALITY:
         if use_hpo:
             from ray import tune
+
             hyperparameters.update(
                 {
                     "model.hf_text.checkpoint_name": tune.choice(
@@ -147,6 +149,7 @@ def default(presets: str = DEFAULT):
         hyperparameters.update({"env.per_gpu_batch_size": 1})
         if use_hpo:
             from ray import tune
+
             hyperparameters.update(
                 {
                     "model.hf_text.checkpoint_name": tune.choice(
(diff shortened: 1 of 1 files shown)


sxjscience (approved):
LGTM

multimodal/src/autogluon/multimodal/presets.py
```diff
@@ -1,7 +1,5 @@
 from typing import List, Optional
 
-from ray import tune
```
Innixma, line 3: Did you test uninstalling ray and running AutoMM? This should work so long as you aren't using HPO.

multimodal/src/autogluon/multimodal/presets.py
```diff
@@ -74,11 +78,14 @@ def default(presets: str = DEFAULT):
 
     presets, use_hpo = parse_presets_str(presets)
     if use_hpo:
+        default_tunable_hyperparameters, default_hyperparameter_tune_kwargs = get_default_hpo_setup()
         hyperparameters.update(default_tunable_hyperparameters)
         hyperparameter_tune_kwargs.update(default_hyperparameter_tune_kwargs)
 
     if presets in [HIGH_QUALITY, DEFAULT]:
         if use_hpo:
+            from ray import tune
```
Innixma, line 87: consider using `try_import_ray` for a more informative import error message.

```
from autogluon.core.utils.try_import import try_import_ray  # top of file


def import_ray_tune():
    try_import_ray()
    from ray import tune
    return tune


def my_func():
    tune = import_ray_tune()
    ...

```

Innixma (approved):
LGTM, but added a few comments

multimodal/src/autogluon/multimodal/presets.py
zhiqiangdon: Tested. It works without doing HPO.

multimodal/src/autogluon/multimodal/presets.py
zhiqiangdon: If using `try_import_ray ()`, the following error message shows.

`ImportError: ray is required to train folds in parallel. A quick tip is to install via `pip install ray==2.2.0`, or use sequential fold fitting by passing `sequential_local` to `ag_args_ensemble` when calling tabular.fitFor example: `predictor.fit(..., ag_args_ensemble={'fold_fitting_strategy': 'sequential_local'})``

which is more customized for tabular predictor and may bring confusions for AutoMM users.

Innixma: good point, then it is ok as is

Innixma:
LGTM, feel free to merge whenever ready

PR #2943, opened 2023-02-17
Remove platform tests timeout

yinweisu:
*Issue #, if available:*

*Description of changes:*


Innixma:
@yinweisu what if a test hangs? Will the test never end?

yinweisu:
> @yinweisu what if a test hangs? Will the test never end?

I think it has a 6 hrs timeout by default. I removed this because I noticed that sometimes tabular test would go beyond 1 hr causing it fail. Maybe we can change it to 2 hrs instead of removing it

PR #2948, opened 2023-02-18
[Post 0.7] [multimodal] Support xywh output for object detection

FANGAreNotGnu:


multimodal/tests/unittests/others/test_object_detection.py
```diff
@@ -91,6 +91,28 @@ def test_mmdet_object_detection_inference_list_str_dict(checkpoint_name):
...
+        "yolov3_mobilenetv2_320_300e_coco",
+    ],
+)
+def test_mmdet_object_detection_inference_xywh_output(checkpoint_name):
+    mmdet_image_name = download_sample_images()
+
+    predictor = MultiModalPredictor(
+        hyperparameters={
+            "model.mmdet_image.checkpoint_name": checkpoint_name,
+            "model.mmdet_image.output_bbox_format": "xywh",
```
liangfu, line 106: check functional equalvance of the two output formats, otherwise the test case would simply pass even without this hyperparameter.

liangfu (changes requested):

multimodal/tests/unittests/others/test_object_detection.py
FANGAreNotGnu: Good point, just updated the test

liangfu (approved):
LGTM! Thanks for the update.

PR #2970, opened 2023-02-23
[timeseries] Provide custom seasonal_period for the MASE metric

shchur:
In many time series benchmarks (e.g., M4 competition), the MASE error is computed using a `seasonal_period` that depends on the data frequency (e.g., 24 for hourly data). Currently, AGTS always assumes `seasonal_period=1` when computing MASE. This leads to the ensemble optimizing the wrong objective, and therefore achieving suboptimal results on external benchmarks (e.g., measured by GluonTS, AF or the official M4 competition code).

Training AGTS with the correct `seasonal_period` can significantly boost the performance. For example, test set MASE on M4 daily
- using `eval_metric_seasonal_period=1` (current master branch): 1.35
- using `eval_metric_seasonal_period=24` (this PR): 1.06

*Description of changes:*
- This PR provides an option to change the `seasonal_period` used for computing `MASE` by specifying the `eval_meric_seasonal_period: int` argument when creating the `TimeSeriesPredictor`. 
  - This information is propagated to learner/trainer/individual models, since `eval_meric_seasonal_period` must always be known when creating a `TimeSeriesPredictor`. The only alternative I see is to refactor the `TimeSeriesEvaluator` class completely to act similar to the `Scorer` class in Tabular, but that would be an even bigger (potentially breaking) change to the code.
- This is intended as functionality for advanced users who want to get competitive results on benchmarks.

### DIFF
--- a/timeseries/src/autogluon/timeseries/evaluator.py
+++ b/timeseries/src/autogluon/timeseries/evaluator.py
@@ -15,14 +15,10 @@ from autogluon.timeseries.utils.warning_filters import evaluator_warning_filter
 logger = logging.getLogger(__name__)
 
 
-def in_sample_naive_1_error(*, y_past: pd.Series) -> pd.Series:
-    """Compute the error of naive forecast (predict previous value) for each time series."""
-    diff = y_past.diff()
-    # We ignore the differences between the last value of prev item and the first value of the next item
-    length_per_item = y_past.groupby(level=ITEMID, sort=False).size()
-    first_index_for_each_item = length_per_item.cumsum().values[:-1]
-    diff.iloc[first_index_for_each_item] = np.nan
-    return diff.abs().groupby(level=ITEMID, sort=False).mean()
+def in_sample_seasonal_naive_error(*, y_past: pd.Series, seasonal_period: int = 1) -> pd.Series:
+    """Compute seasonal naive forecast error (predict value from seasonal_period steps ago) for each time series."""
+    seasonal_diffs = y_past.groupby(level=ITEMID, sort=False).diff(seasonal_period).abs()
+    return seasonal_diffs.groupby(level=ITEMID, sort=False).mean().fillna(1.0)
 
 
 def mse_per_item(*, y_true: pd.Series, y_pred: pd.Series) -> pd.Series:
@@ -67,7 +63,7 @@ class TimeSeriesEvaluator:
 
     Parameters
     ----------
-    eval_metric: str
+    eval_metric : str
         Name of the metric to be computed. Available metrics are
 
         * ``MASE``: mean absolute scaled error. See https://en.wikipedia.org/wiki/Mean_absolute_scaled_error
@@ -78,10 +74,13 @@ class TimeSeriesEvaluator:
         * ``MSE``: mean squared error
         * ``RMSE``: root mean squared error
 
-    prediction_length: int
+    prediction_length : int
         Length of the forecast horizon
-    target_column: str
+    target_column : str, default = "target"
         Name of the target column to be forecasting.
+    eval_metric_seasonal_period : int
+        Seasonal period used to compute the mean absolute scaled error (MASE) evaluation metric. This parameter is only
+        used if ``eval_metric="MASE"`. See https://en.wikipedia.org/wiki/Mean_absolute_scaled_error for more details.
 
     Class Attributes
     ----------------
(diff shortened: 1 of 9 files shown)


timeseries/src/autogluon/timeseries/evaluator.py
```diff
@@ -130,7 +136,7 @@ def _rmse(self, y_true: pd.Series, predictions: TimeSeriesDataFrame) -> float:
     def _mase(self, y_true: pd.Series, predictions: TimeSeriesDataFrame) -> float:
         y_pred = self._get_median_forecast(predictions)
         mae = mae_per_item(y_true=y_true, y_pred=y_pred)
-        return self._safemean(mae / self._past_naive_1_error)
+        return self._safemean(mae / self._past_naive_error)
```
tonyhoo, line 139: If `self._past_naive_error` is 0 or `None`, I guess some error gonna be raised? Can we handle this gracefully here? Also, wondering if there is any case `self._safemean` gonna return `NaN`. If so, would that be expected?

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -61,6 +61,9 @@ class TimeSeriesPredictor:
         - ``"RMSE"``: root mean squared error
 
         For more information about these metrics, see https://docs.aws.amazon.com/forecast/latest/dg/metrics.html.
+    eval_metric_seasonal_period : int, default = 1
```
tonyhoo, line 64: Can we create some logic to calculate the seasonal_error based on the freq? Some example to follow in GluonTS:https://github.com/awslabs/gluonts/blob/176c533dedcbfdf5616021abba277b93304405c8/src/gluonts/evaluation/metrics.py#LL21C16

timeseries/src/autogluon/timeseries/evaluator.py
shchur: - If `self._past_naive_error` equals to 0 for **some** time series, the scaled loss for them will equal `inf` and will be filtered out by `_safemean` (i.e., they won't affect the total loss computation). This happens, e.g., for constant time series. GluonTS will just return `NaN` loss for the entire dataset in such case. I see following options:
  1. Raise an exception like `The provided dataset contains constant time series. The chosen evaluation metric MASE is undefined in this case. Please choose a different eval_metric or remove the constant time series from the dataset`.
  2. Replace 0 -> 1 in `self._past_naive_error` (do not scale the loss for the constant series)
  3. Keep as-is
  
  Do you have any preferences here?
- The case with `self._past_naive_error is None` should never happen, but if it does, an exception will be raised in https://github.com/autogluon/autogluon/pull/2970/files#diff-1711922b22b494da6f1714235d9ac888b754298c639be656bedd0ca35c623007R213
- `self._safemean` can only return `NaN` if all time series have `NaN` loss (e.g., all time series are constant). This seems like a rare/pathological edge case. Do you think we should handle this somehow?

tonyhoo: See [discussion](https://github.com/awslabs/gluonts/issues/1135) on the same topic from GluonTS. I think we shall follow the same standard and implement MASE which is truthful to its [definition](https://otexts.com/fpp3/accuracy.html)

timeseries/src/autogluon/timeseries/predictor.py
tonyhoo: Agree that different framework implement the seasonality differently. We use GluonTS models internally so wondering if we are overriding this value in any way,  otherwise would it introduce discrepancy when we evaluate the model performance

timeseries/src/autogluon/timeseries/evaluator.py
shchur: I just double-checked and compared our implementation (this PR) to GluonTS, my previous statement was incorrect. As far as I can tell our code & GluonTS produce identical results:
- If `_past_naive_error` equals to 0 for **some** time series, GluonTS ignores them and reports average MASE for time series where `_past_naive_error > 0`.
- If `_past_naive_error` equals to 0 for **all** time series, GluonTS returns NaN.
Do you think there are any other edge cases that we should check?

timeseries/src/autogluon/timeseries/predictor.py
shchur: Caner ran a comparison a while back and found that for local models that we implement/wrap ourselves (ETS, Theta, SeasonalNaive) setting `seasonal_period=7` on `freq="D"` led to better MAPE/wQL on some benchmark datasets compared to the default GluonTS seasonality mapping of `seasonal_period=1`. 

Also, as far as I know, none of the GluonTS models that we use (DeepAR, TFT, SFF) depend on seasonality, and we don't rely on the metrics computed by GluonTS in any way.

Back to your original question
- I would suggest to set `eval_metric_seasonal_period=None` by default and doing the inference automatically based on `freq`. This is the right thing to do and hopefully this won't affect the users negatively.
- I would prefer to keep [our freq->seasonal_period mapping](https://github.com/autogluon/autogluon/blob/master/timeseries/src/autogluon/timeseries/utils/seasonality.py#L5) as-is instead of changing to the GluonTS version.  

Does this sound reasonable to you?

timeseries/src/autogluon/timeseries/evaluator.py
tonyhoo: Fine with it as long as it is a truthful impl of MASE. BTW, we shall fix the `ZeroDivisionError` issue when `self._past_naive_error` is 0 here

shchur: I just double-checked: if `self._past_naive_error` contains 0 values, pandas doesn't raise any exceptions or warnings and just stores the result as `NaN`. These `NaN`s are then discarded by the `self._safemean` method, so the overall result is identical to the output of GluonTS / compatible with the textbook definition of MASE.

I think `ZeroDivisionError` can only [occur if the underlying `dtype` is `object`](https://stackoverflow.com/questions/72953622/pandas-dataframe-conditional-formatting-still-leads-to-division-by-zero-error/72953731#72953731), but this should never happen as `TimeSeriesDataFrame` can only contain numeric dtypes.

timeseries/src/autogluon/timeseries/predictor.py
shchur: Done 👍

tonyhoo (approved):
LGTM!

PR #2973, opened 2023-02-24
[timeseries] Speed up featurization for AutoGluonTabularModel

shchur:
*Description of changes:*
- Speed up the `AutoGluonTabularModel._get_features_dataframe` method using the following 3 improvements:
  1. Parallelize the construction of the feature matrix across individual time series with `joblib`.
  2. Initialize the full matrix as a numpy array and fill its values (instead of concatenating pd.DataFrames)
  3. During prediction, the old code used to construct the full matrix and then select the last `prediction_length` rows; now we only construct the last `prediction_length` rows.

**Checking correctness:**
- I checked that both old & new `_get_features_dataframe` methods produce identical results on synthetic datasets generated by taking the cartesian product of these configs:
```python
"freq", ["30S", "T", "A", "Y", "M", "H"]
"use_past_covariates", [True, False]
"use_known_covariates", [True, False]
"use_static_features_continuous", [True, False]
"use_static_features_categorical", [True, False]
"prediction_length", [1, 5]
"max_len", [13, 100_000]
```
**Performance comparison**
Comparing the runtime for the `AutoGluonTabular` model (using only the `LightGBM` model under the hood).
| Dataset     | Fit time old | Fit time new | Pred time old (s) | Pred time new (s)| Number of rows | Number of items |
|-------------|--------------|--------------|---------------|---------------|----------------|-----------------|
| M3 other    | 16.3         | 15.7         | 1.7           | 0.8           | 12K            | 174             |
| M4 hourly   | 49.3         | 34.5         | 10.8          | 2.2           | 353K           | 414             |
| Electricity | 261.9        | 242.7        | 38.2          | 12.7          | 6755K          | 321             |
| M4 daily    | 246.4        | 96.6         | 116.4         | 33.9          | 9964K          | 4227            |
|M5 (with `sell_price` past covariate & 5 static features)              | 2524.5      | 433.9      | 1829.2       | 673.7        | 46796K        | 30490         |


timeseries/src/autogluon/timeseries/models/autogluon_tabular/tabular_model.py
```diff
@@ -97,64 +98,40 @@ def __init__(
     def _get_features_dataframe(
         self,
         data: TimeSeriesDataFrame,
-        last_k_values: Optional[int] = None,
+        max_rows_per_item: int = 100_000,
```
tonyhoo, line 101: wondering any reason why we choose 100k? It is less than 1 month of min level data

timeseries/src/autogluon/timeseries/models/autogluon_tabular/tabular_model.py
```diff
@@ -163,78 +140,130 @@ def apply_mask(
...
+            prediction_length: int,
+            max_rows_per_item: int,
+            mask: bool,
+            name: str,
+        ):
+            """Generate lag features for all time series in the dataset.
+
+            See the docstring of get_lags for the description of the parameters.
+            """
+            lags_per_item = Parallel(n_jobs=-1)(
```
tonyhoo, line 213: [non blocker] we should also allow users to override the number of workers for the processing job. Something similar we exposed via AutoMM at [here](https://auto.gluon.ai/stable/tutorials/multimodal/advanced_topics/customization.html#env-num-workers). If not overridden , we can use -1

timeseries/src/autogluon/timeseries/models/autogluon_tabular/tabular_model.py
```diff
@@ -163,78 +140,130 @@ def apply_mask(
...
+                for ts in all_series
+            )
+            features = np.concatenate(lags_per_item)
+            return pd.DataFrame(features, columns=[f"{name}_lag_{idx}" for idx in lag_indices])
+
+        df = pd.DataFrame(data)
+        all_series = [ts for _, ts in df.droplevel(TIMESTAMP).groupby(level=ITEMID, sort=False)]
+
+        feature_dfs = []
+        for column_name in df.columns:
```
tonyhoo, line 230: [minor] we can simplify this logic into something like:
```python
column_name_to_mask_indices_mapping = {
  self.target: (True, self._target_lag_indices),
  self.metadata.past_covariates_real: (True, self._past_covariates_lag_indices),
  self.metadata.known_covariates_real:  (False, self._known_covariates_lag_indices)
}

mask, lag_indices = column_name_to_mask_indices_mapping. get(column_name, (None, None))
if mask is None: raise ValueError(f"Unexpected column {column_name} is not among target or covariates.")
```

timeseries/src/autogluon/timeseries/models/autogluon_tabular/tabular_model.py
shchur: Totally agree! This is something that is currently not handled uniformly in the entire timeseries module - there are quite a few places where the number of used CPUs / GPUs is either chosen automatically or hardcoded. Fixing this is on the roadmap for v0.8.

How about we keep this as-is for now and get back to it when aligning all the resource allocation logic with Tabular/AutoMM as a separate PR?

timeseries/src/autogluon/timeseries/models/autogluon_tabular/tabular_model.py
shchur: Not sure, I think this will be a bit more complicated since `self.metadata.past_covariates_real` and `self.metadata.known_covariates_real` are both lists, so we will need to have 3 loops - two when constructing `column_name_to_mask_indices_mapping` and one when iterating over the columns of the `df`.  

The current code takes more space then the suggestion, but I feel like the current version might be slightly easier to read because of the double for-loop in the suggestion.

timeseries/src/autogluon/timeseries/models/autogluon_tabular/tabular_model.py
tonyhoo: Yes, let's create an issue and track it for 0.8 release

tonyhoo (approved):

PR #2977, opened 2023-02-27
[timeseries] Parallelize TimeSeriesDataFrame construction from iterable dataset

shchur:
*Description of changes:*
- Parallelize the construction of `TimeSeriesDataFrame` from GluonTS-style iterable dataset. Instead of processing the items sequentially with a single core, parallelize the work across items with joblib.
  - This significantly reduces the processing times for large datasets (e.g., for `electricity` we go from 337s -> 30s).
  - This speeds up our benchmarking & positively affects users who store their datasets in GluonTS format.

### DIFF
--- a/timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
+++ b/timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
@@ -260,7 +260,7 @@ class TimeSeriesDataFrame(pd.DataFrame):
 
     @classmethod
     def _construct_pandas_frame_from_iterable_dataset(cls, iterable_dataset: Iterable) -> pd.DataFrame:
-        def load_single_df(item_id: int, ts: dict) -> pd.DataFrame:
+        def load_single_item(item_id: int, ts: dict) -> pd.DataFrame:
             start_timestamp = ts["start"]
             freq = start_timestamp.freq
             if isinstance(start_timestamp, pd.Period):
@@ -272,7 +272,7 @@ class TimeSeriesDataFrame(pd.DataFrame):
 
         cls._validate_iterable(iterable_dataset)
         all_ts = Parallel(n_jobs=-1)(
-            delayed(load_single_df)(item_id, ts) for item_id, ts in enumerate(iterable_dataset)
+            delayed(load_single_item)(item_id, ts) for item_id, ts in enumerate(iterable_dataset)
         )
         return pd.concat(all_ts)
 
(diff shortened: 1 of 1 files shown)


timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
```diff
@@ -259,19 +260,20 @@ def _validate_multi_index_data_frame(cls, data: pd.DataFrame):
...
             target = ts["target"]
             datetime_index = tuple(pd.date_range(start_timestamp, periods=len(target), freq=freq))
-            idx = pd.MultiIndex.from_product([(i,), datetime_index], names=[ITEMID, TIMESTAMP])
-            ts_df = pd.Series(target, name="target", index=idx).to_frame()
-            all_ts.append(ts_df)
+            idx = pd.MultiIndex.from_product([(item_id,), datetime_index], names=[ITEMID, TIMESTAMP])
+            return pd.Series(target, name="target", index=idx).to_frame()
+
+        cls._validate_iterable(iterable_dataset)
+        all_ts = Parallel(n_jobs=-1)(
```
tonyhoo, line 274: We should allow users to override this from HPs such as `env.num_workers`

shchur: This code is only executed outside of the `TimeSeriesPredictor` (when user constructs a `TimeSeriesDataframe` with `TimeSeriesDataframe.from_iterable_dataset`), so this method has no access to HPs.

How about we add an optional kwarg `num_cpus: int = -1`  to `TimeSeriesDataframe.from_iterable_dataset` that controls the number of jobs here?

tonyhoo: yes, adding to `from_iterable_dataset` sounds good to me

shchur: @tonyhoo Done 👍

tonyhoo (approved):
LGTM! Thanks for the improvement

jmakov:
Shouldn't this be implemented using ray.io which would allow you to scale to the cluster?

shchur:
@jmakov while we could make this one operation more scalable, it wouldn't really affect the overall scalability of the library because of other bottlenecks (e.g., the fact that all code relies on in-memory pd.DataFrames). With the current design we can handle datasets with <100M rows which covers the majority of use cases. If we decide to add support for even larger datasets, it will require a substantial re-design of the library, which is currently not planned.

jmakov:
Thanks for the clarification. I see, one would need to refactor to sth like ray dataset which supports streaming to nodes. Not trivial.

PR #2981, opened 2023-02-27
Move interpretable logic to InterpretableTabularPredictor

Innixma:
*Issue #, if available:*

*Description of changes:*

- Moved interpretable logic to its own predictor class
- Removed non-interpretable models from the trained models in interpretable setting
- Removed weighted ensemble in interpretable setting
- Disabled calibration in interpretable setting
- Added experimental warnings to interpretable setting and tutorial

- [x] DO NOT MERGE until new website stable is live, this code represents a breaking change in a tutorial


tabular/src/autogluon/tabular/predictor/__init__.py
```diff
@@ -1 +1,2 @@
+from .interpretable_predictor import InterpretableTabularPredictor
 from .predictor import TabularPredictor
```
gradientsky, line 2: nit: empty line at the EOF missing

tabular/src/autogluon/tabular/predictor/interpretable_predictor.py
```diff
@@ -0,0 +1,98 @@
...
+class InterpretableTabularPredictor(TabularPredictor):
+    def fit(self,
+            train_data,
+            tuning_data=None,
+            time_limit=None,
+            *,
+            presets='interpretable',
+            fit_weighted_ensemble=False,
+            calibrate=False,
+            **kwargs):
```
gradientsky, line 21: nit: data types; add docstrings

tabular/src/autogluon/tabular/predictor/interpretable_predictor.py
```diff
@@ -0,0 +1,98 @@
...
+        '''
+        d = self.fit_summary(verbosity=verbosity)
+        summaries = pd.DataFrame.from_dict(d)
+
+        complexities = []
+        info = self.info()
+        for i in range(summaries.shape[0]):
+            model_name = summaries.index.values[i]
+            complexities.append(info['model_info'][model_name].get('complexity', np.nan))
+        summaries.insert(2, 'complexity', complexities)
```
gradientsky, line 47: why insert at specific position (2)? magic number needs an explanation

tabular/src/autogluon/tabular/predictor/interpretable_predictor.py
```diff
@@ -0,0 +1,98 @@
...
+        """
+        if model_name is None:
+            summaries = self.interpretable_models_summary()
+            summaries_filtered = summaries[summaries.complexity <= complexity_threshold]
+            if summaries_filtered.shape[0] == 0:
+                summaries_filtered = summaries
+            model_name = summaries_filtered.index.values[0]  # best model is at top
+        agmodel = self._trainer.load_model(model_name)
+        imodel = agmodel.model
+        print(imodel)
```
gradientsky, line 71: Use logging; print won't get into a file if filehandler is used

tabular/src/autogluon/tabular/predictor/__init__.py
Innixma: added

tabular/src/autogluon/tabular/predictor/interpretable_predictor.py
Innixma: updated

tabular/src/autogluon/tabular/predictor/interpretable_predictor.py
Innixma: keeping as print, we use print for leaderboard, and this is similar

tabular/src/autogluon/tabular/predictor/interpretable_predictor.py
Innixma: keeping to align with TabularPredictor, we can add type hints to TabularPredictor later on, but the type hints are already in TabularPredictor's docstring, which this method inherits.

PR #2991, opened 2023-03-02
[eda] various fixes and quality of life improvements

gradientsky:
*Description of changes:*
- fixed Sampler error when dataset is less than sample length of the sample size
- Model evaluator for regression now can handle test_data; plots will have both val_data and test_data plotted in results charts

image

- added verbosity support for ApplyFeatureGenerator
- dataset_overview is now not plotting category vs category charts since they don't add value
- covariate shift detection shows only top 5 feature importance items to reduce clutter and actionability

### DIFF
--- a/eda/tests/unittests/analysis/test_model.py
+++ b/eda/tests/unittests/analysis/test_model.py
@@ -45,6 +45,37 @@ def test_AutoGluonModelEvaluator_regression():
     _assert_importance_is_present(state)
     assert state.model_evaluation.confusion_matrix is None
     assert state.model_evaluation.confusion_matrix_normalized is None
+    assert state.model_evaluation.y_true_val is None
+    assert state.model_evaluation.y_pred_val is None
+
+
+def test_AutoGluonModelEvaluator_regression__with_test_data():
+    df_train = pd.read_csv(os.path.join(RESOURCE_PATH, "houses", "train_data.csv")).sample(100, random_state=0)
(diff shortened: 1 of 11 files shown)


eda/tests/unittests/analysis/test_model.py
```diff
@@ -45,6 +45,37 @@ def test_AutoGluonModelEvaluator_regression():
...
     assert state.model_evaluation.confusion_matrix is None
     assert state.model_evaluation.confusion_matrix_normalized is None
+    assert state.model_evaluation.y_true_val is None
+    assert state.model_evaluation.y_pred_val is None
+
+
+def test_AutoGluonModelEvaluator_regression__with_test_data():
+    df_train = pd.read_csv(os.path.join(RESOURCE_PATH, "houses", "train_data.csv")).sample(100, random_state=0)
+    df_val = pd.read_csv(os.path.join(RESOURCE_PATH, "houses", "test_data.csv"))[:50]
+    df_test = pd.read_csv(os.path.join(RESOURCE_PATH, "houses", "test_data.csv"))[50:101]
```
gidler, line 55: Not an issue, just caught my eye, did you intend `df_test` to have 51 entries while `df_val` has 50 and `df_train` has 100?

gidler (approved):
LGTM

eda/tests/unittests/analysis/test_model.py
gradientsky: This is by design to quickly identify what dataset was used in the call.

Issue #3003, opened 2023-03-06
[BUG] AutoMM Continuous fit corrupted validation score


Innixma:
- [x] I have checked that this bug exists on the latest stable version of AutoGluon
- [x] and/or I have checked that this bug exists on the latest mainline of AutoGluon via source installation

**Describe the bug**

Calling fit multiple times without specifying identical `tuning_data` and with shared `train_data` rows will lead to a corrupted validation score.

**Expected behavior**

- [ ] A warning should be printed if tuning data changes between fit calls.
- [ ] #3004 If the same `train_data` is passed in multiple fit calls, it should be split into train and val using the same seed so `tuning_data` is identical each time.
- [ ] #3004 splitting logic should be identical to TabularPredictor (including stratification when applicable)

**To Reproduce**

```python3

from autogluon.core.utils.loaders import load_pd

train_data = load_pd.load('https://autogluon-text.s3-accelerate.amazonaws.com/glue/sst/train.parquet')
test_data = load_pd.load('https://autogluon-text.s3-accelerate.amazonaws.com/glue/sst/dev.parquet')
subsample_size = 500  # subsample data for faster demo, try setting this to larger values
train_data = train_data.sample(n=subsample_size, random_state=0)

label = 'label'
eval_metric = 'roc_auc'

from autogluon.multimodal import MultiModalPredictor
predictor = MultiModalPredictor(label=label, eval_metric=eval_metric)
predictor.fit(
    train_data,
    seed=0,
    hyperparameters={
        "optimization.max_epochs": 4,
    }
)

predictor.fit(
    train_data,
    seed=0,
    hyperparameters={
        "optimization.max_epochs": 4,
    }
)
```

You can check the train and val splits by adding the following print statements directly before `trainer.fit` in `MultiModalPredictor._fit`:

```python3
        with warnings.catch_warnings():
            warnings.filterwarnings(
                "ignore",
                ".*does not have many workers which may be a bottleneck. "
                "Consider increasing the value of the `num_workers` argument` "
                ".* in the `DataLoader` init to improve performance.*",
            )
            warnings.filterwarnings("ignore", "Checkpoint directory .* exists and is not empty.")
            print(f'TRAIN DATA: (len: {len(train_dm.train_data)})')
            print(train_dm.train_data.head(5))
            print(f'VAL DATA: (len: {len(train_dm.validate_data)})')
            print(train_dm.validate_data.head(5))
            trainer.fit(
                task,
                datamodule=train_dm,
[...]

PR #3004, opened 2023-03-06
AutoMM: Add consistent seed, stratification, alignment with tabular for train-val split

Innixma:
*Issue #, if available:*

#3003

*Description of changes:*

- Previously AutoMM would split the same data differently even with the same seed when calling fit multiple times. Now it will use the same split if called with same data and same seed.
- Changed default seed to 0 to align with Tabular
- Changed splitting logic to be identical to Tabular, including adding stratification and other guardrails such as ensuring at least 1 sample per class exists in train data after split. Stratification should help improve AutoMM's test scores, especially when train_data has a small amount of rows.

### Tabular & MultiModal Identical Result

Now the following code produces identical results for Tabular and AutoMM (previously did not due to (1) seed difference, (2) data split logic difference:

```python3
from autogluon.core.utils.loaders import load_pd

train_data = load_pd.load('https://autogluon-text.s3-accelerate.amazonaws.com/glue/sst/train.parquet')
test_data = load_pd.load('https://autogluon-text.s3-accelerate.amazonaws.com/glue/sst/dev.parquet')
subsample_size = 1000  # subsample data for faster demo, try setting this to larger values

label = 'label'
eval_metric = 'accuracy'

predictor = TabularPredictor(label=label, eval_metric=eval_metric)
predictor.fit(
    train_data,
    hyperparameters={'AG_AUTOMM': {
        "optimization.max_epochs": 2,
    }},
    fit_weighted_ensemble=False,
)

test_score = predictor.evaluate(test_data)
print(test_score)

predictor = MultiModalPredictor(label=label, eval_metric=eval_metric)
predictor.fit(
    train_data,
    hyperparameters={
        "optimization.max_epochs": 2,
    }
)

test_score = predictor.evaluate(test_data)
print(test_score)

```

Output:

```
# TabularPredictor
{'accuracy': 0.8279816513761468, 'balanced_accuracy': 0.8292708596446914, 'mcc': 0.6639300307917014, 'roc_auc': 0.9188057800791447, 'f1': 0.8179611650485437, 'precision': 0.8868421052631579, 'recall': 0.759009009009009}
# MultiModalPredictor
{'accuracy': 0.8279816513761468}
```

### MultiModalPredictor Continuous Training Identical Split

Now AutoMM uses the same train/val split when passed the same training data across multiple fit calls with the same seed. This avoids corrupting the validation score across multiple fit calls (#3003)

You can check the difference by running this code (Note: You'll need to add print statements to the train/val data rows or enter debugger to see that they are misaligned in mainline, and aligned in this PR):

```python3
from autogluon.core.utils.loaders import load_pd

train_data = load_pd.load('https://autogluon-text.s3-accelerate.amazonaws.com/glue/sst/train.parquet')
test_data = load_pd.load('https://autogluon-text.s3-accelerate.amazonaws.com/glue/sst/dev.parquet')
subsample_size = 1000  # subsample data for faster demo, try setting this to larger values

label = 'label'
eval_metric = 'accuracy'

predictor = MultiModalPredictor(label=label, eval_metric=eval_metric)

predictor.fit(
    train_data,
    hyperparameters={
        "optimization.max_epochs": 2,
    }
)

predictor.fit(
    train_data,
    hyperparameters={
        "optimization.max_epochs": 2,
    }
)
```

### DIFF
--- a/multimodal/src/autogluon/multimodal/predictor.py
+++ b/multimodal/src/autogluon/multimodal/predictor.py
@@ -731,9 +731,9 @@ class MultiModalPredictor(ExportMixin):
         )
 
         if tuning_data is None:
-            train_data, tuning_data = self._split_train_tuning(data=train_data,
-                                                               holdout_frac=holdout_frac,
-                                                               random_state=seed)
+            train_data, tuning_data = self._split_train_tuning(
+                data=train_data, holdout_frac=holdout_frac, random_state=seed
+            )
 
             # FIXME: Because output_shape is paired with problem_type detection,
             #  need to do this again in-case output_shape changed when splitting data
@@ -854,10 +854,9 @@ class MultiModalPredictor(ExportMixin):
 
         return self
 
-    def _split_train_tuning(self,
-                            data: pd.DataFrame,
-                            holdout_frac: float = None,
-                            random_state: int = 0) -> (pd.DataFrame, pd.DataFrame):
+    def _split_train_tuning(
+        self, data: pd.DataFrame, holdout_frac: float = None, random_state: int = 0
+    ) -> (pd.DataFrame, pd.DataFrame):
         """
         Splits `data` into `train_data` and `tuning_data`.
         If the problem_type is one of ['binary', 'multiclass']:
@@ -883,8 +882,7 @@ class MultiModalPredictor(ExportMixin):
         Tuple of (train_data, tuning_data) of the split `data`
         """
         if holdout_frac is None:
-            holdout_frac = default_holdout_frac(num_train_rows=len(data),
-                                                hyperparameter_tune=False)
+            holdout_frac = default_holdout_frac(num_train_rows=len(data), hyperparameter_tune=False)
 
         # TODO: Hack since the recognized problem types are only binary, multiclass, and regression
         #  Problem types used for purpose of stratification, so regression = no stratification
--- a/multimodal/src/autogluon/multimodal/predictor.py
+++ b/multimodal/src/autogluon/multimodal/predictor.py
@@ -838,7 +838,7 @@ class MultiModalPredictor(ExportMixin):
 
     # FIXME: Avoid having separate logic for inferring features and label column that is combined together
     def _infer_column_types(
-            self, train_data: pd.DataFrame, tuning_data: pd.DataFrame = None, column_types: dict = None
+        self, train_data: pd.DataFrame, tuning_data: pd.DataFrame = None, column_types: dict = None
     ) -> dict:
         column_types = infer_column_types(
             data=train_data,
@@ -859,7 +859,9 @@ class MultiModalPredictor(ExportMixin):
     #  don't combine output_shape and problem_type detection, make them separate
     #  Use autogluon.core.utils.utils.infer_problem_type
     def _infer_problem_type(self, train_data: pd.DataFrame, column_types: dict = None) -> str:
-        column_types_label = self._infer_column_types(train_data=train_data[[self._label_column]], column_types=column_types)
+        column_types_label = self._infer_column_types(
+            train_data=train_data[[self._label_column]], column_types=column_types
+        )
 
         problem_type, _ = infer_problem_type_output_shape(
             label_column=self._label_column,
--- a/multimodal/src/autogluon/multimodal/predictor.py
+++ b/multimodal/src/autogluon/multimodal/predictor.py
@@ -688,6 +688,16 @@ class MultiModalPredictor(ExportMixin):
 
         pl.seed_everything(seed, workers=True)
 
+        if self._presets is not None:
+            # FIXME: Silently ignoring user input, there should be a warning
+            presets = self._presets
+        else:
+            self._presets = presets
+
+        if self._config is not None:  # continuous training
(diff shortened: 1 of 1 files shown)


multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -833,6 +838,52 @@ def fit(
...
+
+        Parameters
+        ----------
+        data : pd.DataFrame
+            The data to be split
+        holdout_frac : float, default = None
+            The ratio of data to use as validation.
+            If 0.2, 20% of the data will be used for validation, and 80% for training.
+            If None, the ratio is automatically determined,
+            ranging from 0.2 for small row count to 0.01 for large row count.
```
gradientsky, line 859: Are we changing train-test split ratios logic to match tabular here?

If we are changing, this might not be the right thing to do for text models. Generally you’ll need more text rows for validation than in tabular models. 

Specifically, cases like when you have 100k rows, giving 0.01 of that for validation might not enough information for stable validation. On top of that, generally we don't train multiple multi-modal models unlike the defaults in tabular top presets.

Innixma: It is not changed. It was already matching Tabular. Also, 100k will not have 0.01, it will have 0.025 (2500).

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -530,7 +529,7 @@ def fit(
         column_types: Optional[dict] = None,
         holdout_frac: Optional[float] = None,
         teacher_predictor: Union[str, MultiModalPredictor] = None,
-        seed: Optional[int] = 123,
+        seed: Optional[int] = 0,
```
zhiqiangdon, line 532: We may need to compare the performance on AutoMM benchmark datasets before and after changing the seed. If there is no performance drop, we can make this change.

Innixma: If we have a performance drop, that just means we were overfit on the seed. It shouldn't be a blocker for merging.

Innixma: But yes, I would recommend a benchmark run on this PR, since there are other impactful changes, such as using stratification during splits.

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -718,14 +709,28 @@ def fit(
         if self._config is not None:  # continuous training
             config = self._config
 
-        problem_type, output_shape = infer_problem_type_output_shape(
+        # FIXME: Align logic with Tabular,
+        #  don't combine output_shape and problem_type detection, make them separate
```
zhiqiangdon, line 713: Agree. We need to separate the logic. The data split inconsistency between two fit() is due to that we tie them together and don't infer problem type before splitting data.

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -687,15 +687,6 @@ def fit(
             fit_called=fit_called,
         )
 
-        train_data, tuning_data = split_train_tuning_data(
```
zhiqiangdon, line 690: Moving `split_train_tuning_data ` after `infer_column_types ` would cause issues since `infer_column_types ` requires to use `tuning_data` in training. Otherwise, columns with one unique value can't be ignored. See https://github.com/autogluon/autogluon/blob/master/multimodal/src/autogluon/multimodal/data/infer_types.py#L504-L528

zhiqiangdon (changes requested):
`infer_column_types ` requires `tuning_data` in training.

multimodal/src/autogluon/multimodal/predictor.py
Innixma: Updated to resolve this. Note that it is a bit awkward to resolve this given how logic is being coupled together in the current functions, so I've added TODOs and FIXMEs where I think it can be streamlined.

Innixma:
> `infer_column_types ` requires `tuning_data` in training.

fixed, although I will note the code as previously written made it very easy to do this incorrectly. We should look to refactor the code so such bugs are harder to accidentally introduce.

multimodal/src/autogluon/multimodal/predictor.py
zhiqiangdon: @tonyhoo Do we need to test this PR on the vision and text benchmarks?

zhiqiangdon (approved):
LGTM. We can refactor the logic of inferring problem type and output shape later.

multimodal/src/autogluon/multimodal/predictor.py
tonyhoo: Can we use the tutorial generated from this PR as an indicator on the performance impact? If there are some concerns raised from there, I am fine rerun the NLP and CV benchmark

Innixma:
@zhiqiangdon rebased to resolve merge conflict. Please feel free to merge when you are ready.

Innixma:
> > @zhiqiangdon rebased to resolve merge conflict. Please feel free to merge when you are ready.
> 
> CI has one unit test error.

Should be fixed in latest commit.

PR #3007, opened 2023-03-06
Update tabular tutorial pip installs

gidler:
*Description of changes:*
The `pip install` cell for the tabular tutorials should be `autogluon.tabular[all]` for all the notebooks except `tabular-multimodal` which should be `pip install autogluon` since AutoMM is needed as well.


Innixma (approved):
LGTM!

PR #3014, opened 2023-03-07
[eda] added `explain_rows` method to `autogluon.eda.auto` - Kernel SHAP visualization

gradientsky:
*Description of changes:*
- added `explain_rows` method to `autogluon.eda.auto`; the methods performs Kernel SHAP values analysis and visualization
- `quick_fit`: fixes for highest_error and undecided rows calculations

## Examples

```python
import pandas as pd
import numpy as np

# Load data
df_train = pd.read_csv('https://autogluon.s3.amazonaws.com/datasets/titanic/train.csv')
df_test = pd.read_csv('https://autogluon.s3.amazonaws.com/datasets/titanic/test.csv')
label='Survived'

# Fit model
state = auto.quick_fit(
    train_data=df_train,
    label=label,
    save_model_to_state=True,
    return_state=True,
    render_analysis=False,  # Don't render analysis
)

# Explain the row with highest error
auto.explain_rows(
    train_data=df_train,
    model=state.model,
    backend='shap',  # default | shap/fastshap
    plot='force',  # default | force/waterfall
    rows=state.model_evaluation.highest_error[:1],
)
```

image

```python
# Explain the row predicted incorrectly, but closest to the decision boundary as waterfall plot
auto.explain_rows(
    train_data=df_train,
    model=state.model,
    display_rows=True,
    plot='waterfall',
    rows=state.model_evaluation.undecided[:1],
)
```

image

## Using primitives

```python
s = auto.analyze(
    train_data=df_train, model=state.model, 
    return_state=True, 
    anlz_facets=[
        # Backend using `shap` package
        eda.explain.ShapAnalysis(state.model_evaluation.highest_error[:2]),
        # Backend using `fastshap` package
        # eda.explain.FastShapAnalysis(state.model_evaluation.highest_error[:2]),
    ],
    viz_facets=[
        viz.explain.ExplainForcePlot(),  # Force layout
        viz.explain.ExplainWaterfallPlot(),  # Waterfall layout
    ]
)
```

### DIFF
--- a/eda/setup.py
+++ b/eda/setup.py
@@ -30,6 +30,7 @@ install_requires = [
     'seaborn>=0.12.0,<0.13',
     'ipywidgets>=7.7.1,<9.0',  # min versions guidance: 7.7.1 collab/kaggle
     'shap>=0.41,<0.42',
+    'fastshap>=0.3,<0.4',
     f'autogluon.core=={version}',
     f'autogluon.common=={version}',
     f'autogluon.features=={version}',
--- a/eda/setup.py
+++ b/eda/setup.py
@@ -30,7 +30,6 @@ install_requires = [
     'seaborn>=0.12.0,<0.13',
     'ipywidgets>=7.7.1,<9.0',  # min versions guidance: 7.7.1 collab/kaggle
     'shap>=0.41,<0.42',
-    'fastshap>=0.3,<0.4',
     f'autogluon.core=={version}',
     f'autogluon.common=={version}',
     f'autogluon.features=={version}',
--- a/eda/setup.py
+++ b/eda/setup.py
@@ -29,6 +29,7 @@ install_requires = [
     'phik>=0.12.2,<0.13',
     'seaborn>=0.12.0,<0.13',
     'ipywidgets>=7.7.1,<9.0',  # min versions guidance: 7.7.1 collab/kaggle
+    'shap>=0.41,<0.42',
     f'autogluon.core=={version}',
     f'autogluon.common=={version}',
     f'autogluon.features=={version}',
--- a/eda/src/autogluon/eda/analysis/explain.py
+++ b/eda/src/autogluon/eda/analysis/explain.py
@@ -13,7 +13,7 @@ __all__ = ["ShapAnalysis"]
 logger = logging.getLogger(__name__)
 
 
-class _ShapAutogluonWrapper:
+class _ShapAutoGluonWrapper:
     def __init__(self, predictor, feature_names, target_class=None):
         self.ag_model = predictor
         self.feature_names = feature_names
@@ -54,6 +54,8 @@ class ShapAnalysis(AbstractAnalysis):
         wrapped analyses; these will receive sampled `args` during `fit` call
     state: AnalysisState
         state to be updated by this fit function
+    random_state: int, default = 0
+        random state for sampling
     kwargs
 
     Examples
@@ -89,11 +91,13 @@ class ShapAnalysis(AbstractAnalysis):
         parent: Optional[AbstractAnalysis] = None,
         children: Optional[List[AbstractAnalysis]] = None,
         state: Optional[AnalysisState] = None,
+        random_state: int = 0,
         **kwargs,
     ) -> None:
         super().__init__(parent, children, state, **kwargs)
         self.rows = rows
         self.baseline_sample = baseline_sample
+        self.random_state = random_state
 
     def can_handle(self, state: AnalysisState, args: AnalysisState) -> bool:
         return self.all_keys_must_be_present(args, "model", "train_data")
@@ -104,15 +108,15 @@ class ShapAnalysis(AbstractAnalysis):
         else:
             _baseline_sample = len(args.train_data)
 
-        baseline = args.train_data.sample(_baseline_sample, random_state=0)
+        baseline = args.train_data.sample(_baseline_sample, random_state=self.random_state)
         shap_data = []
         for _, row in self.rows.iterrows():
             _row = pd.DataFrame([row])
             if args.model.problem_type == REGRESSION:
-                predicted_class = 0
+                predicted_class = None
             else:
                 predicted_class = args.model.predict(_row).iloc[0]
-            ag_wrapper = _ShapAutogluonWrapper(args.model, args.train_data.columns, predicted_class)
+            ag_wrapper = _ShapAutoGluonWrapper(args.model, args.train_data.columns, predicted_class)
             explainer = shap.KernelExplainer(ag_wrapper.predict_proba, baseline)
             ke_shap_values = explainer.shap_values(_row[args.train_data.columns], silent=True)
             shap_data.append(
--- a/eda/src/autogluon/eda/analysis/explain.py
+++ b/eda/src/autogluon/eda/analysis/explain.py
@@ -37,6 +37,51 @@ class _ShapAutogluonWrapper:
 
 
 class ShapAnalysis(AbstractAnalysis):
+    """
+    Perform Shapley values calculation using `shap` package for the given rows.
+
+    Parameters
+    ----------
+    rows: pd.DataFrame,
+        rows to explain
+    baseline_sample: int, default = 100
+        The background dataset size to use for integrating out features. To determine the impact
+        of a feature, that feature is set to "missing" and the change in the model output
+        is observed.
+    parent: Optional[AbstractAnalysis], default = None
+        parent Analysis
+    children: List[AbstractAnalysis], default = []
+        wrapped analyses; these will receive sampled `args` during `fit` call
+    state: AnalysisState
+        state to be updated by this fit function
+    kwargs
+
+    Examples
+    --------
+    >>> import autogluon.eda.analysis as eda
+    >>> import autogluon.eda.visualization as viz
+    >>> import autogluon.eda.auto as auto
+    >>>
+    >>> auto.analyze(
+    >>>     train_data=..., model=...,
+    >>>     anlz_facets=[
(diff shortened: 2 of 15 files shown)


eda/setup.py
```diff
@@ -29,6 +29,7 @@
     'phik>=0.12.2,<0.13',
     'seaborn>=0.12.0,<0.13',
     'ipywidgets>=7.7.1,<9.0',  # min versions guidance: 7.7.1 collab/kaggle
+    'shap>=0.41,<0.42',
```
Innixma, line 32: @gradientsky FYI, fastshap may be of interest to test out and compare performance as suggested here: https://github.com/autogluon/autogluon/issues/2222#issuecomment-1455126633

fastshap claims to be much faster than shap:

https://raw.githubusercontent.com/AnotherSamWilson/fastshap/master/benchmarks/iris_benchmark_time.png

gradientsky: It doesn't have row explanations and not compatible with shap visualization tools. I tried to plug it, but the values it generates are not matching `shap` outputs.

gradientsky: Did a code rework. The code now split into backend analysis and rendering parts. Analysis supports both `shap` and `fastshap` libraries (two different backends). Using `shap` as a default one for `auto` functionality because it is faster.
Visualizations also split into separate primitives; compatible with both of the backends.

gradientsky:
Blocked by https://github.com/AnotherSamWilson/fastshap/issues/8

eda/setup.py
```diff
@@ -29,6 +29,8 @@
...
+    'fastshap>=0.3,<0.4',
```
Innixma, line 33: Is fastshap good enough to have as a required dependency? What is the relative advantages of fastshap over shap?

gradientsky: For the purposes of explaining a few rows here, `fastshap` is slower than `shap`. I sent an update to remove fastshap backend completely (performance + support concern).

eda/src/autogluon/eda/analysis/explain.py
```diff
@@ -0,0 +1,127 @@
...
+        if not isinstance(X, pd.DataFrame):
+            X = pd.DataFrame(X, columns=self.feature_names)
+        if self.ag_model.problem_type == REGRESSION:
+            preds = self.ag_model.predict(X)
+        else:
+            preds = self.ag_model.predict_proba(X)
+        if self.ag_model.problem_type == REGRESSION or self.target_class is None:
+            return preds
+        else:
+            return preds[self.target_class]
```
Innixma, line 36: I see a similar hack here that I see in the general AutoGluon Tabular code of using `predict_proba` for regression. Long term do you think this is the right thing to do long-term? is it used to simplify the code logic or just to align with how Tabular does things?

eda/src/autogluon/eda/analysis/explain.py
```diff
@@ -0,0 +1,127 @@
...
+from autogluon.core.constants import REGRESSION
+from autogluon.eda import AnalysisState
+from autogluon.eda.analysis.base import AbstractAnalysis
+
+__all__ = ["ShapAnalysis"]
+
+logger = logging.getLogger(__name__)
+
+
+class _ShapAutogluonWrapper:
```
Innixma, line 16: `_ShapAutogluonWrapper` or `_ShapAutoGluonWrapper`?

eda/src/autogluon/eda/analysis/explain.py
```diff
@@ -0,0 +1,127 @@
...
+    Perform Shapley values calculation using `shap` package for the given rows.
+
+    Parameters
+    ----------
+    rows: pd.DataFrame,
+        rows to explain
+    baseline_sample: int, default = 100
+        The background dataset size to use for integrating out features. To determine the impact
+        of a feature, that feature is set to "missing" and the change in the model output
+        is observed.
```
Innixma, line 50: any guidance on what magnitudes represent what noise level?

Is 100 enough? 1000? Will 10000 differ significantly from 100?

Why choose 100 when I could choose 10000? Is it purely to save compute time?

Will it still work if I set it to 1?

eda/src/autogluon/eda/analysis/explain.py
```diff
@@ -0,0 +1,127 @@
...
+    """
+    Perform Shapley values calculation using `shap` package for the given rows.
+
+    Parameters
+    ----------
+    rows: pd.DataFrame,
+        rows to explain
+    baseline_sample: int, default = 100
+        The background dataset size to use for integrating out features. To determine the impact
+        of a feature, that feature is set to "missing" and the change in the model output
```
Innixma, line 49: "missing": Do we have a consistent definition for what setting to "missing" means?

eda/src/autogluon/eda/analysis/explain.py
```diff
@@ -0,0 +1,127 @@
...
+    def can_handle(self, state: AnalysisState, args: AnalysisState) -> bool:
+        return self.all_keys_must_be_present(args, "model", "train_data")
+
+    def _fit(self, state: AnalysisState, args: AnalysisState, **fit_kwargs) -> None:
+        if self.baseline_sample <= len(args.train_data):
+            _baseline_sample = self.baseline_sample
+        else:
+            _baseline_sample = len(args.train_data)
+
+        baseline = args.train_data.sample(_baseline_sample, random_state=0)
```
Innixma, line 107: move random_state=0 to an init arg?

eda/src/autogluon/eda/analysis/explain.py
```diff
@@ -0,0 +1,127 @@
...
+        shap_data = []
+        for _, row in self.rows.iterrows():
+            _row = pd.DataFrame([row])
+            if args.model.problem_type == REGRESSION:
+                predicted_class = 0
```
Innixma, line 112: Why not `None`?

eda/src/autogluon/eda/analysis/model.py
```diff
@@ -228,13 +228,17 @@ def _predict(self, problem_type, label, predictor, val_data):
         undecided = None
         if predictor.problem_type in [BINARY, MULTICLASS]:
             y_proba = predictor.predict_proba(val_data)
-            highest_error = y_proba[y_true_val != y_pred_val].max(axis=1)
+
+            misclassified = y_proba[y_true_val != y_pred_val]
+            expected_value = misclassified.join(y_true_val).apply(lambda row: row.loc[row[label]], axis=1)
+            predicted_value = misclassified.max(axis=1)
```
Innixma, line 234: Long-term we may need to revisit this, as `.max` might not actually be identical to the predicted value. For example, if we start maximizing `f1` score, we would set a threshold that isn't `0.5`. Probably not necessary to address in this PR, but maybe worth adding a TODO.

Innixma: I'd imagine we eventually make a method like `pred = predictor.proba_to_pred(proba, metric='f1')` which contains an inner dictionary of metric -> threshold such as `{'f1': 0.63}` that influences the `pred` that is returned

eda/src/autogluon/eda/auto/simple.py
```diff
@@ -924,3 +940,75 @@ def missing_values_analysis(
...
+        if `True` then display the row before the explanation chart
+    plot: Optional[str], default = 'force'
+        type of plot to visualize the Shapley values. Supported keys:
+        - `force` - Visualize the given SHAP values with an additive force layout
+        - `waterfall` - Visualize the given SHAP values with a waterfall layout
+        - `None` - do not use any visualization
+    baseline_sample: int, default = 100
+        The background dataset size to use for integrating out features. To determine the impact
+        of a feature, that feature is set to "missing" and the change in the model output
+        is observed.
```
Innixma, line 981: Very random thought completely tangential to the PR: I wish IDEs/code had logic where we could specify a docstring for a parameter one time in some file, then have it auto-populate via a variable reference in the source-code docs so we just have to write

```
baseline_sample: int, default = 100
     {baseline_sample_docstring}
```

and the IDE magically converts it to text unless we click to force it to show source, so we avoid having to copy/paste the same doc-string many times across many usages.

That is my random thought of the day. Carry on.

Innixma (approved):
LGTM! Had some minor comments

eda/src/autogluon/eda/analysis/explain.py
gradientsky: There should be enough rows to cancel-out significant variance in the base value 
image

1 row should work, but you will have garbage output. More rows -> longer it takes to get the values.

eda/src/autogluon/eda/analysis/explain.py
gradientsky: This is how the `shap` describes the process.

eda/src/autogluon/eda/analysis/explain.py
gradientsky: fixed

eda/src/autogluon/eda/analysis/explain.py
gradientsky: Good call; updated

eda/src/autogluon/eda/analysis/explain.py
Innixma: > "There should be enough rows to cancel-out significant variance in the base value"

Do we assume the user knows what "enough rows" is? Would it be better to give guidance in the doc-string?

eda/src/autogluon/eda/analysis/model.py
Innixma: Nevermind this specific comment, I see `predictor.predict` called above, so it should be fine.

PR #3020, opened 2023-03-08
Remove .doctrees build artifacts before s3 upload

gidler:
*Description of changes:*

I was cleaning up our `dev/` website S3 resources and observed that the S3 folder is ~70MBs, and ~28MBs of it is a directory called `.doctrees`. This directory is full of binary files names `*.doctree`. I determined that these files are created and used during the Sphinx build process but are not required for serving or viewing the generated webpage. See this [StackOverflow answer](https://stackoverflow.com/questions/33904042/is-doctrees-folder-required-for-displaying-html-docs-with-sphinx) and the related [Sphinx Docs](https://www.sphinx-doc.org/en/master/man/sphinx-build.html#cmdoption-sphinx-build-d).

In this change, I delete these files from our CI build steps before doing `aws s3 cp` so that we don't waste S3 space, which will become more important when we move to GitHub Pages.

### DIFF
--- a/.github/workflow_scripts/build_all_docs.sh
+++ b/.github/workflow_scripts/build_all_docs.sh
@@ -61,6 +61,8 @@ LOCAL_DOC_PATH=_build/html
 cd docs
 sphinx-build -D nb_execution_mode=off -b html . $LOCAL_DOC_PATH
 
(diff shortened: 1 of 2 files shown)


yinweisu (approved):
LGTM! Let's wait for the CI to check if the website still works after the change

PR #3022, opened 2023-03-09
[AutoMM] Support customizing image data augmentation

zhiqiangdon:
*Issue #, if available:*

*Description of changes:*

1. Support customization with a list of supported strings.
2. Support customization with a list of callable objects.
3. Support customization for both predictor and matcher.
4. Introduce image augmentation in the AutoMM customization doc.
5. Merge test_data_process_image.py into test_process_image.py.


multimodal/src/autogluon/multimodal/utils/config.py
```diff
@@ -761,3 +774,20 @@ def filter_hyperparameters(
...
+
+    advanced_hyperparameters = dict()
+    for k, v in hyperparameters.items():
+        if k in [
+            "model.timm_image.train_transforms",
+            "model.timm_image.val_transforms",
+        ]:
+            if not all([isinstance(trans, str) for trans in hyperparameters[k]]):
+                advanced_hyperparameters[k] = copy.deepcopy(v)
+                hyperparameters[k] = []
```
liangfu, line 791: After split, the hyperparameter dict itself doesn't aware whether it has been splitted or not. Basically, some advanced hyperparameters are modified (not deleted), and copied to another variable, where the naive hyperparameters don't have access to.

Since hyperparameter space has been divided into subspaces with the dot delimitor, how about moving advanced hyperparameters into 'advanced.xxx' subspace ?? Potentially, with OmegaConf, we can easily access to these advanced hyperparameters with `config.advanced`.

liangfu (approved):
LGTM, and thanks for broadening what users can customize through hyperparameter inputs.

PR #3025, opened 2023-03-10
[NER] Fix Negative Index Issue

cheungdaven:
Avoid negative index when padding


zhiqiangdon (approved):
LGTM

PR #3033, opened 2023-03-12
Tabular: Refactor params_aux & memory checks

### LINKED ISSUE #3031: Tabular: Clarify memory requirements when memory is insufficient to fit
In the below example, it is unclear what the user should do:

```
Fitting model: LightGBM ...
        Warning: Not enough memory to safely train model, roughly requires: 9.425 GB, but only 9.642 GB is available...
        Not enough memory to train LightGBM... Skipping this model.
```

For example,

1. The user may want to try to train the model anyways, even if it could result in OOM. The way to do so is not communicated.
2. The user may want to know how much memory they would need available to avoid skipping the model. This is not communicated.
3. The user may want to know why the model requires X GB of memory. This is not communicated.

Note: The code that logs this information is here https://github.com/autogluon/autogluon/blob/0.7.0/core/src/autogluon/core/models/abstract/abstract_model.py#L1445-L1454

Innixma:
*Issue #, if available:*

resolves #3031 

*Description of changes:*

# 1: Refactor params_aux

- Previously it was very confusing how to specify aux hyperparameters of a model.
- Now, we simply prefix the hyperparameter with `ag.` to indicate it as an aux hyperparameter, which is much easier.
- Prior API functionality remains in-tact, this is purely an extra way to pass the arguments.
- Added unit tests for this logic.
- I've added a TODO to expand upon this long term for improved ease of use and extensibility by enabling aux hyperparameters to be part of HPO search spaces. This should be made available by v1.0 release, but is too complicated to make sense including in this PR.

Mainline:

```
hyperparameters = {
    'GBM': {'ag_args_fit': {'max_memory_usage_ratio': 2.5}},
}
```

This PR:

```
hyperparameters = {
    'GBM': {'ag.max_memory_usage_ratio': 2.5}
}
```

# 2: Refactor memory checks

- Previously, it was very unclear what users should do if they run into memory check errors.
- This has been greatly expanded on, and now the logs provide ample information.
- Additionally, the logic has been refactored and standardized to remove code duplication.

Mainline:

```
Fitting model: LightGBM ...
	Warning: Not enough memory to safely train model, roughly requires: 10.0 GB, but only 6.517 GB is available...
	Not enough memory to train LightGBM... Skipping this model.
```

This PR:

```
Fitting model: LightGBM ...
	Warning: Not enough memory to safely train model. Estimated to require 10.0 GB out of 8.273 GB available memory (134.312%)... (90.0% of avail memory is the max safe size)
	To force training the model, specify the model hyperparameter "ag.max_memory_usage_ratio" to a larger value (currently 1.0, set to >=1.39 to avoid the error)
		To set the same value for all models, do the following when calling predictor.fit: `predictor.fit(..., ag_args_fit={"ag.max_memory_usage_ratio": VALUE})`
		Setting "ag.max_memory_usage_ratio" to values above 1 may result in out-of-memory errors. You may consider using a machine with more memory as a safer alternative.
	Not enough memory to train LightGBM... Skipping this model.
```

### DIFF
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -8,7 +8,7 @@ import pickle
 from autogluon.core.utils import try_import
 import sys
 import time
-from typing import Dict, Optional, Union
+from typing import Any, Dict, Optional, Union
 
 import numpy as np
 import pandas as pd
@@ -24,7 +24,7 @@ from autogluon.common.utils.resource_utils import ResourceManager
 from .model_trial import model_trial, skip_hpo
 from ._tags import _DEFAULT_CLASS_TAGS, _DEFAULT_TAGS
 from ... import metrics, Space
-from ...constants import AG_ARGS_FIT, BINARY, REGRESSION, QUANTILE, REFIT_FULL_SUFFIX, OBJECTIVES_TO_NORMALIZE
+from ...constants import AG_ARG_PREFIX, AG_ARGS_FIT, BINARY, REGRESSION, QUANTILE, REFIT_FULL_SUFFIX, OBJECTIVES_TO_NORMALIZE
 from ...data.label_cleaner import LabelCleaner, LabelCleanerMulticlassToBinary
 from ...hpo.exceptions import EmptySearchSpace
 from ...hpo.constants import RAY_BACKEND, CUSTOM_BACKEND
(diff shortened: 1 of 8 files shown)


core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -137,28 +137,78 @@ def __init__(self,
...
         self._compiler = None
 
+    @classmethod
+    def _init_user_params(cls, params: Optional[Dict[str, Any]], ag_args_fit: str = AG_ARGS_FIT, ag_arg_prefix: str = AG_ARG_PREFIX) -> (Dict[str, Any], Dict[str, Any]):
+        """
+        Given the user-specified hyperparameters, split into `params` and `params_aux`.
+
+        Parameters
+        ----------
+        params : Optional[Dict[str, Any]]
```
yinweisu, line 158: Shouldn't we set the default to None if this value is optional?

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -137,28 +137,78 @@ def __init__(self,
...
+        ag_args_fit : str, default = "ag_args_fit"
+            The params key to look for that contains params_aux.
+            If the key is present, the value is used for params_aux and popped from params.
+            If no such key is found, then initialize params_aux as an empty dictionary.
+        ag_arg_prefix : str, default = "ag."
+            The key prefix to look for that indicates a parameter is intended for params_aux.
+            If None, this logic is skipped.
+            If a key starts with this prefix, it is popped from params and added to params_aux with the prefix removed.
+            For example:
+                input:  params={'ag.foo': 2, 'abc': 7}, params_aux={'bar': 3}, and ag_arg_prefix='.ag',
```
yinweisu, line 169: Where do we validate the ag_args? Just wonder if adding the prefix logic would affect the validation

Innixma: We validate them after this logic. ag_args is validated during `.fit`. This logic is done during `.init`

core/src/autogluon/core/models/abstract/abstract_model.py
Innixma: that is reasonable, added

yinweisu (approved):
LGTM

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -137,28 +137,78 @@ def __init__(self,
...
+        params = copy.deepcopy(params) if params is not None else dict()
+        assert isinstance(params, dict), f"Invalid dtype of params! Expected dict, but got {type(params)}"
+        params_aux = params.pop(ag_args_fit, dict())
+        if params_aux is None:
+            params_aux = dict()
+        assert isinstance(params_aux, dict), f"Invalid dtype of params_aux! Expected dict, but got {type(params_aux)}"
+        if ag_arg_prefix is not None:
+            param_aux_keys = list(params_aux.keys())
+            for k in param_aux_keys:
+                if isinstance(k, str) and k.startswith(ag_arg_prefix):
```
gradientsky, line 192: should we do a warning if the key is not string? I don't think we ever use non-string keys -> must be coding error.

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -137,28 +137,78 @@ def __init__(self,
...
+                    k_no_prefix = k[len(ag_arg_prefix):]
```
gradientsky, line 193: do we expect just one level of nesting? Would it bring value supporting multi-level nesting?

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -137,28 +137,78 @@ def __init__(self,
...
+            param_keys = list(params.keys())
+            for k in param_keys:
+                if isinstance(k, str) and k.startswith(ag_arg_prefix):
+                    k_no_prefix = k[len(ag_arg_prefix):]
+                    if k_no_prefix in params_aux:
+                        logger.warning(f'Warning: {cls.__name__} hyperparameter "{k}" is present '
+                                       f'in both `ag_args_fit` and `hyperparameters`. '
+                                       f'Will use `hyperparameters` value.')
+                    params_aux[k_no_prefix] = params[k]
+                    params.pop(k)
```
gradientsky, line 209: this is effectively a repeating code block as above, just on different nesting level -> do we want to generalize this function to multi-level nested construct?

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -1435,28 +1485,89 @@ def get_minimum_resources(self, is_gpu_available=False) -> Dict[str, int]:
...
+            if min_error_memory_ratio >= 1:
+                log_user_guideline += f'\n\t\tSetting "ag.max_memory_usage_ratio" to values above 1 may result in out-of-memory errors. ' \
+                                      f'You may consider using a machine with more memory as a safer alternative.'
+            logger.warning(f'\tWarning: Not enough memory to safely train model. {log_user_guideline}')
             raise NotEnoughMemoryError
-        elif ratio > (0.6 * max_memory_usage_ratio):
-            logger.warning('\tWarning: Potentially not enough memory to safely train model, roughly requires: %s GB, but only %s GB is available...' % (round(approx_mem_size_req / 1e9, 3), round(available_mem / 1e9, 3)))
+        elif min_warning_memory_ratio > max_memory_usage_warning_ratio:
+            log_user_guideline += f'\n\tTo avoid this warning, specify the model hyperparameter "ag.max_memory_usage_ratio" to a larger value ' \
+                                  f'(currently {max_memory_usage_ratio}, set to >={round(min_warning_memory_ratio + 0.05, 2)} to avoid the warning)' \
```
gradientsky, line 1564: ditto

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -1435,28 +1485,89 @@ def get_minimum_resources(self, is_gpu_available=False) -> Dict[str, int]:
...
+
+        log_ag_args_fit_example = '`predictor.fit(..., ag_args_fit={"ag.max_memory_usage_ratio": VALUE})`'
+        log_ag_args_fit_example = f'\n\t\tTo set the same value for all models, do the following when calling predictor.fit: {log_ag_args_fit_example}'
+
+        log_user_guideline = f'Estimated to require {round(approx_mem_size_req / 1e9, 3)} GB ' \
+                             f'out of {round(available_mem / 1e9, 3)} GB available memory ({round(min_error_memory_ratio*100, 3)}%)... ' \
+                             f'({round(max_memory_usage_error_ratio*100, 3)}% of avail memory is the max safe size)'
+        if min_error_memory_ratio > max_memory_usage_error_ratio:
+            log_user_guideline += f'\n\tTo force training the model, specify the model hyperparameter "ag.max_memory_usage_ratio" to a larger value ' \
+                                  f'(currently {max_memory_usage_ratio}, set to >={round(min_error_memory_ratio + 0.05, 2)} to avoid the error)' \
```
gradientsky, line 1555: `round` is not required - formatter supports that out-of-the box:
```python
f'(currently {max_memory_usage_ratio}, set to >={min_warning_memory_ratio + 0.05:.2f} to avoid the warning)'
```

core/src/autogluon/core/models/abstract/abstract_model.py
Innixma: Technically there is nothing stopping a model from using an integer as a key for a hyperparameter (although I would actively be angry if they did). Maybe the scenario is so absurd that we should disallow it or log a warning until proven otherwise.

core/src/autogluon/core/models/abstract/abstract_model.py
Innixma: We could add in future, but there is nothing implemented that would take advantage of it currently.

core/src/autogluon/core/models/abstract/abstract_model.py
Innixma: Given that there are meaningful differences between the code blocks and very different log statements, I think it is ok to keep as is. If you'd like to propose a code refactor of the logic, I'd be happy to review.

core/src/autogluon/core/models/abstract/abstract_model.py
Innixma: Added:

```
Warning: Specified LGBModel hyperparameter key is not of type str: 5 (type=<class 'int'>). There might be a bug in your configuration.
Warning: Specified LGBModel hyperparameter key is not of type str: 7.3 (type=<class 'float'>). There might be a bug in your configuration.
Fitting 1 L1 models ...
Fitting model: LightGBM ...
[LightGBM] [Warning] Unknown parameter: 7.3
[LightGBM] [Warning] Unknown parameter: 5
	-9.7463	 = Validation score   (-root_mean_squared_error)
	2.13s	 = Training   runtime
	0.02s	 = Validation runtime
```

core/src/autogluon/core/models/abstract/abstract_model.py
Innixma: good catch, updated

core/src/autogluon/core/models/abstract/abstract_model.py
Innixma: good catch, updated

gradientsky (approved):

PR #3052, opened 2023-03-16
Support sample_weight in RMSE

jjaeyeon:
*Description of changes:*
- Supporting weighted RMSE for evaluation metric - if user specifies kwargs, it will use scikit learn's RMSE function, or else it will use the existing accelerated RMSE computation
- [tested] test-run with the example dataset

### DIFF
--- a/core/src/autogluon/core/metrics/__init__.py
+++ b/core/src/autogluon/core/metrics/__init__.py
@@ -454,8 +454,11 @@ pearsonr = make_scorer('pearsonr',
                        greater_is_better=True)
 
 
-def rmse_func(y_true, y_pred):
-    return np.sqrt(((y_true - y_pred) ** 2).mean())
+def rmse_func(y_true, y_pred, **kwargs):
+    if kwargs:
+        return sklearn.metrics.mean_squared_error(y_true, y_pred, squared=False, **kwargs)
+    else:
+        return np.sqrt(((y_true - y_pred) ** 2).mean())
 
 
 root_mean_squared_error = make_scorer('root_mean_squared_error',
--- a/core/tests/unittests/metrics/test_metrics.py
+++ b/core/tests/unittests/metrics/test_metrics.py
@@ -2,6 +2,7 @@ from math import isclose
 
 import numpy as np
 import pytest
+import sklearn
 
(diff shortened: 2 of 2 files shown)


core/src/autogluon/core/metrics/__init__.py
```diff
@@ -454,8 +454,11 @@ def local_pearsonr(y_true, y_pred):
                        greater_is_better=True)
 
 
-def rmse_func(y_true, y_pred):
-    return np.sqrt(((y_true - y_pred) ** 2).mean())
+def rmse_func(y_true, y_pred, **kwargs):
+    if kwargs:
+        return sklearn.metrics.mean_squared_error(y_true, y_pred, squared=False, **kwargs)
```
liangfu, line 459: why squared is False ?

liangfu, line 459: Why not use this directly and drop numpy version ?

liangfu (commented):
Thanks for the contribution. Would be better if you could add a test case for this as well.

core/src/autogluon/core/metrics/__init__.py
```diff
@@ -454,8 +454,11 @@ def local_pearsonr(y_true, y_pred):
                        greater_is_better=True)
 
 
-def rmse_func(y_true, y_pred):
-    return np.sqrt(((y_true - y_pred) ** 2).mean())
+def rmse_func(y_true, y_pred, **kwargs):
```
Innixma, line 457: Please add a simple unit test that calls this function with and without sample weights, verifying that it returns the correct score. You can refer to the existing metric unit tests for log_loss and roc_auc: https://github.com/autogluon/autogluon/blob/master/core/tests/unittests/metrics/test_classification_metrics.py#L130-L222

core/src/autogluon/core/metrics/__init__.py
jjaeyeon: Discussed with Nick, following the implementation of roc_auc, we keep this to use accelerated (skipping validation checks) implementation if no additional argument is given

liangfu:
> I see the test failure from timeseries due to tensors in different device - is this a known issue?

https://github.com/autogluon/autogluon/pull/3058 this has merged, and should fix the issue. Please rebase latest master branch.

Innixma (approved):

PR #3054, opened 2023-03-16
RayResourceManager and ParallelLocalFoldFittingStrategy

yinweisu:
*Issue #, if available:*

*Description of changes:*
* This PR enables distributed training in AG tabular by implementing `RayResourceManager` and `ParallelLocalFoldFittingStrategy`
* `RayResourceManager` will report total resources of the cluster instead of a single hardware. 
  * The functionality requires import of ray. The `try_import` function cannot be used here because it's being defined within `autogluon.core`, which has a dependency on `autogluon.common`. Using it would result in a circular dependency. Here, we only instantiate `RayResourceManager` when a env var `AG_DISTRIBUTED_MODE` is set. This variable is supposed to be set by the `cloud` module
* `ParallelLocalFoldFittingStrategy` will have a different ray init args and requires a s3 bucket for model synchronization between head and worker nodes. Currently the s3 bucket is being defined in a env var `AG_MODEL_SYNC_PATH`. This interface is subject to change.
* Adding unit test is hard as it requires a ray cluster to work properly. The cloud module can test the functionality on its side.

Attach a screenshot of AG being run on a cluster of 2 m5.2xlarge machine.
Take notice on how 2 node_id are present for tasks and each task is utilizing 2 cpus while a single m5.2xlarge only has 8 cpus.
image

`predictor.predict` also works after the `fit` meaning model artifacts have been saved correctly
```bash
Fitting 1 L1 models ...
Fitting model: LightGBM_BAG_L1 ...
parallel_distributed
        Fitting 8 child models (S1F1 - S1F8) | Fitting with ParallelDistributedFoldFittingStrategy
        0.85     = Validation score   (accuracy)
        0.45s    = Training   runtime
        0.04s    = Validation runtime
        0.85     = Validation score   (accuracy)
        0.0s     = Training   runtime
        0.0s     = Validation runtime
AutoGluon training complete, total runtime = 3.73s ... Best model: "WeightedEnsemble_L2"
TabularPredictor saved. To load, use: predictor = TabularPredictor.load("AutogluonModels/ag-20230316_203258/")
6118      <=50K
23204     <=50K
29590     <=50K
18116     <=50K
33964      >50K
          ...  
29128     <=50K
23950     <=50K
13700      >50K
35248     <=50K
24772     <=50K
```

### DIFF
--- a/common/src/autogluon/common/utils/resource_utils.py
+++ b/common/src/autogluon/common/utils/resource_utils.py
@@ -127,7 +127,7 @@ class ResourceManager:
 
     
 class RayResourceManager:
-    """Manager that fetches ray cluster resources info."""
+    """Manager that fetches ray cluster resources info. This class should only be used within a ray cluster."""
 
     @staticmethod
     def _init_ray():
--- a/common/src/autogluon/common/utils/resource_utils.py
+++ b/common/src/autogluon/common/utils/resource_utils.py
@@ -4,6 +4,8 @@ import os
 import shutil
 import subprocess
 
+from typing import Union
+
 from .utils import bytes_to_mega_bytes
 from .lite import disable_if_lite_mode
 
@@ -131,6 +133,7 @@ class RayResourceManager:
 
     @staticmethod
     def _init_ray():
+        """Initialize ray runtime if not already initialized. Will force the existance of a cluster already being spinned up"""
         import ray
         if not ray.is_initialized():
             ray.init(
@@ -140,15 +143,28 @@ class RayResourceManager:
             )
     
     @staticmethod
-    def _get_cluster_sources(key, default_val=0):
+    def _get_cluster_resources(key: str, default_val: Union[int, float]=0):
+        """
+        Get value of resources available in the cluster.
+        
+        Parameter
+        ---------
+        key: str
+            The key of the value you want to get, i.e. CPU
+        default_val: Union[int, float]
+            Default value to get if key not available in the cluster
+        """
         import ray
         RayResourceManager._init_ray()
         return ray.cluster_resources().get(key, default_val)
 
     @staticmethod
     def get_cpu_count():
-        return RayResourceManager._get_cluster_sources("CPU")
+        """Get number of cpu cores (virtual) available in the cluster"""
+        return RayResourceManager._get_cluster_resources("CPU")
 
     @staticmethod
+    # TODO: find a better naming, "all" sounds unnecessary
     def get_gpu_count_all():
-        return RayResourceManager._get_cluster_sources("GPU")
+        """Get number of gpus available in the cluster"""
+        return RayResourceManager._get_cluster_resources("GPU")
--- a/common/src/autogluon/common/utils/resource_utils.py
+++ b/common/src/autogluon/common/utils/resource_utils.py
@@ -6,6 +6,7 @@ import subprocess
 
 from typing import Union
 
+from .distribute_utils import DistributedContext
 from .utils import bytes_to_mega_bytes
 from .lite import disable_if_lite_mode
 
@@ -168,3 +169,8 @@ class RayResourceManager:
     def get_gpu_count_all():
         """Get number of gpus available in the cluster"""
         return RayResourceManager._get_cluster_resources("GPU")
+    
+    
+def get_resource_manager():
+    """Get resource manager class based on the training context"""
+    resource_manager = RayResourceManager if DistributedContext.is_distributed_mode else ResourceManager
--- a/common/src/autogluon/common/utils/resource_utils.py
+++ b/common/src/autogluon/common/utils/resource_utils.py
@@ -1,4 +1,5 @@
 import multiprocessing
+import logging
 import os
 import shutil
 import subprocess
@@ -123,3 +124,31 @@ class ResourceManager:
         gpu_count = cudaDeviceGetCount()
         cudaShutdown()
         return gpu_count
+
+    
+class RayResourceManager:
+    """Manager that fetches ray cluster resources info."""
+
+    @staticmethod
+    def _init_ray():
+        import ray
+        if not ray.is_initialized():
+            ray.init(
+                address="auto",  # Force ray to connect to an existing cluster. There should be one. Otherwise, something went wrong
+                log_to_driver=False,
+                logging_level=logging.ERROR
+            )
+    
+    @staticmethod
+    def _get_cluster_sources(key, default_val=0):
+        import ray
+        RayResourceManager._init_ray()
+        return ray.cluster_resources().get(key, default_val)
+
+    @staticmethod
+    def get_cpu_count():
+        return RayResourceManager._get_cluster_sources("CPU")
+
+    @staticmethod
+    def get_gpu_count_all():
+        return RayResourceManager._get_cluster_sources("GPU")
--- a/common/src/autogluon/common/utils/resource_utils.py
+++ b/common/src/autogluon/common/utils/resource_utils.py
@@ -6,6 +6,8 @@ import subprocess
 
 from typing import Union
 
+from autogluon.common.utils.try_import import try_import_ray
+
 from .distribute_utils import DistributedContext
 from .utils import bytes_to_mega_bytes
 from .lite import disable_if_lite_mode
@@ -135,6 +137,7 @@ class RayResourceManager:
     @staticmethod
     def _init_ray():
         """Initialize ray runtime if not already initialized. Will force the existance of a cluster already being spinned up"""
+        try_import_ray()
         import ray
         if not ray.is_initialized():
             ray.init(
@@ -155,6 +158,7 @@ class RayResourceManager:
         default_val: Union[int, float]
             Default value to get if key not available in the cluster
         """
+        try_import_ray()
         import ray
         RayResourceManager._init_ray()
         return ray.cluster_resources().get(key, default_val)
--- a/common/src/autogluon/common/utils/resource_utils.py
+++ b/common/src/autogluon/common/utils/resource_utils.py
@@ -173,4 +173,4 @@ class RayResourceManager:
     
 def get_resource_manager():
     """Get resource manager class based on the training context"""
-    resource_manager = RayResourceManager if DistributedContext.is_distributed_mode else ResourceManager
+    return RayResourceManager if DistributedContext.is_distributed_mode() else ResourceManager
--- a/common/src/autogluon/common/utils/s3_utils.py
+++ b/common/src/autogluon/common/utils/s3_utils.py
@@ -1,4 +1,9 @@
+import boto3
 import logging
+import os
+import shutil
+
+from typing import List
 
 
 logger = logging.getLogger(__name__)
@@ -23,7 +28,6 @@ def s3_bucket_prefix_to_path(bucket, prefix, version='s3'):
 
 
 def delete_s3_prefix(bucket, prefix):
-    import boto3
     s3 = boto3.resource('s3')
     objects_to_delete = s3.meta.client.list_objects(Bucket=bucket, Prefix=prefix)
 
@@ -41,6 +45,8 @@ def download_s3_folder(
     prefix: str,
     local_path: str,
     error_if_exists: bool = True,
+    delete_if_exists: bool = False,
+    dry_run: bool = False,
     verbose: bool = True
 ):
     """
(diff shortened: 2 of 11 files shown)


core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -564,8 +564,11 @@ def _preprocess_fit_resources(self, silent=False, total_resources=None, parallel
             enforced_num_gpus = kwargs.get('num_gpus', None)
             assert enforced_num_cpus is not None and enforced_num_cpus != 'auto' and enforced_num_gpus is not None and enforced_num_gpus != 'auto'
             return kwargs
-        system_num_cpus = ResourceManager.get_cpu_count()
-        system_num_gpus = ResourceManager.get_gpu_count_all()
+        resource_manager = ResourceManager
+        if os.environ.get("AG_DISTRIBUTED_MODE", False):
+            resource_manager = RayResourceManager
```
Innixma, line 569: move to dedicated method to enable easy overriding.

core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
```diff
@@ -411,28 +411,17 @@ def _set_n_repeat_single(self):
...
-        )
-        current_os = platform.system()
-        fold_fitting_strategy = os_fitting_strategy_map.get(current_os, 'sequential_local')
-        if fold_fitting_strategy == 'sequential_local':
-            warning_msg = f'\tWill use sequential fold fitting strategy because {current_os} OS does not yet support parallel folding.'
+        try:
+            try_import_ray()
+            fold_fitting_strategy = "parallel_local"
+            if os.environ.get("AG_DISTRIBUTED_MODE", False):
+                fold_fitting_strategy = "parallel_distributed"
```
Innixma, line 418: Can be a 1-liner

```
fold_fitting_strategy = "parallel_distributed" if is_distributed_mode() else "parallel_local"
```

core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
```diff
@@ -411,28 +411,17 @@ def _set_n_repeat_single(self):
...
-            Linux='parallel_local',
-        )
-        current_os = platform.system()
-        fold_fitting_strategy = os_fitting_strategy_map.get(current_os, 'sequential_local')
-        if fold_fitting_strategy == 'sequential_local':
-            warning_msg = f'\tWill use sequential fold fitting strategy because {current_os} OS does not yet support parallel folding.'
+        try:
+            try_import_ray()
+            fold_fitting_strategy = "parallel_local"
+            if os.environ.get("AG_DISTRIBUTED_MODE", False):
```
Innixma, line 417: Don't hard-code this in multiple places. Make a function somewhere like `is_distributed_mode()`

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -564,8 +564,11 @@ def _preprocess_fit_resources(self, silent=False, total_resources=None, parallel
             enforced_num_gpus = kwargs.get('num_gpus', None)
             assert enforced_num_cpus is not None and enforced_num_cpus != 'auto' and enforced_num_gpus is not None and enforced_num_gpus != 'auto'
             return kwargs
-        system_num_cpus = ResourceManager.get_cpu_count()
-        system_num_gpus = ResourceManager.get_gpu_count_all()
+        resource_manager = ResourceManager
+        if os.environ.get("AG_DISTRIBUTED_MODE", False):
```
Innixma, line 568: if is_distributed_mode():

core/src/autogluon/core/models/abstract/abstract_model.py
Innixma: can also be 1 liner

core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
```diff
@@ -464,12 +453,14 @@ def _fit_folds(self,
             fold_fitting_strategy = self._get_default_fold_fitting_strategy()
         num_folds_parallel = self.params.get('num_folds_parallel', 'auto')
         disable_parallel_fitting = self.params.get('_disable_parallel_fitting', False)
-        if fold_fitting_strategy == 'parallel_local':
+        if fold_fitting_strategy in ['parallel_local', "parallel_distributed"]:
+            if fold_fitting_strategy == "parallel_local":
+                fold_fitting_strategy = ParallelLocalFoldFittingStrategy
+            else:
+                fold_fitting_strategy = ParallelDistributedFoldFittingStrategy
```
Innixma, line 460: This logic is getting sufficiently complicated that it deserves its own method. All the logic prior to L472 should be moved to its own method.

core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
```diff
@@ -519,9 +510,11 @@ def _fit_folds(self,
             save_folds=save_folds, num_cpus=num_cpus, num_gpus=num_gpus
         )
         # noinspection PyCallingNonCallable
-        if fold_fitting_strategy == ParallelLocalFoldFittingStrategy:
+        if issubclass(fold_fitting_strategy, ParallelFoldFittingStrategy):
             fold_fitting_strategy_args['num_jobs'] = num_folds
             fold_fitting_strategy_args['num_folds_parallel'] = num_folds_parallel
+        if fold_fitting_strategy == ParallelDistributedFoldFittingStrategy:
+            fold_fitting_strategy_args['model_sync_path'] = os.environ.get("AG_MODEL_SYNC_PATH")
```
Innixma, line 517: Don't sneak in os environment variables in random places. All distributed env variable fetching should be in a singular place.

For example (please consider thinking of a better name, and I didn't actually test if the code is correct):
```

class DistributedVariables:
     @staticmethod
     @property
     def model_sync_path() -> str:
           return os.environ.get("AG_MODEL_SYNC_PATH")

     @staticmethod
     @property
     def is_distributed_mode() -> bool:
           return os.environ.get("AG_DISTRIBUTED_MODE", False)

```


```

fold_fitting_strategy_args['model_sync_path'] = DistributedVariables.model_sync_path
is_distributed_mode = DistributedVariables.is_distributed_mode

```

core/src/autogluon/core/models/ensemble/fold_fitting_strategy.py
```diff
@@ -340,19 +341,30 @@ def _fit(self, model_base, time_start_fold, time_limit_fold, fold_ctx, kwargs):
...
+    y,
+    X_pseudo,
+    y_pseudo,
+    fold_ctx,
+    time_limit_fold,
+    save_bag_folds, 
+    resources,
+    kwargs_fold,
+    model_sync_path=None
+):
```
Innixma, line 357: lets add type hints

core/src/autogluon/core/models/ensemble/fold_fitting_strategy.py
```diff
@@ -377,7 +389,10 @@ def _ray_fit(model_base, bagged_ensemble_model_path,
     fold_model.fit_time = time_train_end_fold - time_start_fold
     fold_model, pred_proba = _ray_predict_oof(fold_model, X_val_fold, y_val_fold,
                                               time_train_end_fold, resources['num_cpus'], save_bag_folds)
-    fold_model.save()
+    model_save_path = None
+    if model_sync_path is not None:
+        model_save_path = model_sync_path + f"{fold_model.name}/"
+    fold_model.save(model_save_path)
```
Innixma, line 395: nit: use key=value

core/src/autogluon/core/models/ensemble/fold_fitting_strategy.py
```diff
@@ -470,17 +491,17 @@ def _estimate_data_memory_usage(self):
 
     def schedule_fold_model_fit(self, fold_ctx):
         self.jobs.append(fold_ctx)
+        
+    def _get_ray_init_args(self):
+        return dict(
+            address="auto",
+            logging_level=logging.ERROR,
+            log_to_driver=False
+        )
```
Innixma, line 500: add docstring

core/src/autogluon/core/models/ensemble/fold_fitting_strategy.py
```diff
@@ -470,17 +491,17 @@ def _estimate_data_memory_usage(self):
 
     def schedule_fold_model_fit(self, fold_ctx):
         self.jobs.append(fold_ctx)
+        
+    def _get_ray_init_args(self):
```
Innixma, line 495: nit: add type hint

core/src/autogluon/core/models/ensemble/fold_fitting_strategy.py
```diff
@@ -708,3 +733,31 @@ def _parse_ray_error(self, e):
             logger.warning(default_error_msg)
             e = NotEnoughCudaMemoryError
         return e
+    
+    def _sync_model_artifact(self, local_path, model_sync_path):
+        pass
```
Innixma, line 738: docstring + type hint. Do we even want to allow calling this in the scenarios it doesn't do anything?

core/src/autogluon/core/models/ensemble/fold_fitting_strategy.py
```diff
@@ -708,3 +733,31 @@ def _parse_ray_error(self, e):
...
+class ParallelDistributedFoldFittingStrategy(ParallelFoldFittingStrategy):
+    
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        # Append bag model name in the path
+        self.model_sync_path = self.model_sync_path + os.path.basename(os.path.normpath(self.bagged_ensemble_model.path)) + "/"
+
+    def _sync_model_artifact(self, local_path, model_sync_path):
+        bucket, path = s3_path_to_bucket_prefix(model_sync_path)
+        download_s3_folder(bucket, path, local_path)
```
Innixma, line 763: docstring + type hint

Innixma (changes requested):
Added comments

core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
Innixma: refer to later comment for a more refined suggestion

core/src/autogluon/core/models/abstract/abstract_model.py
Innixma: refer to later comment for a more refined suggestion

common/src/autogluon/common/utils/resource_utils.py
```diff
@@ -1,4 +1,5 @@
 import multiprocessing
+import logging
```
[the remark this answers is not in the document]
yinweisu: Yea, i can do this in a follow-up PR

common/src/autogluon/common/utils/resource_utils.py
```diff
@@ -113,3 +114,31 @@ def _get_gpu_count_cuda():
...
+        RayResourceManager._init_ray()
+        return ray.cluster_resources().get(key, default_val)
+
+    @staticmethod
+    def get_cpu_count():
+        return RayResourceManager._get_cluster_sources("CPU")
+
+    @staticmethod
+    def get_gpu_count_all():
+        return RayResourceManager._get_cluster_sources("GPU")
```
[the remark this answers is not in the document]
yinweisu: This is already done by the ray.init(address="auto"). This line will fail if there's no existing cluster

common/src/autogluon/common/utils/resource_utils.py
```diff
@@ -113,3 +114,31 @@ def _get_gpu_count_cuda():
...
+                logging_level=logging.ERROR
+            )
+    
+    @staticmethod
+    def _get_cluster_sources(key, default_val=0):
+        import ray
+        RayResourceManager._init_ray()
+        return ray.cluster_resources().get(key, default_val)
+
+    @staticmethod
```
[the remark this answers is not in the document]
yinweisu: Added type hints. Ray cluster only work with virtual cores

common/src/autogluon/common/utils/resource_utils.py
```diff
@@ -113,3 +114,31 @@ def _get_gpu_count_cuda():
...
+    def get_cpu_count():
+        return RayResourceManager._get_cluster_sources("CPU")
+
+    @staticmethod
+    def get_gpu_count_all():
```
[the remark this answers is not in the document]
yinweisu: Yea, was trying to keep the naming convention.

common/src/autogluon/common/utils/s3_utils.py
```diff
@@ -29,3 +29,47 @@ def delete_s3_prefix(bucket, prefix):
 
     if len(delete_keys['Objects']) != 0:
         s3.meta.client.delete_objects(Bucket=bucket, Delete=delete_keys)
+        
+
+def download_s3_folder(bucket, prefix, local_path="."):
+    """
+    This util function downloads a s3 folder and maintain its structure.
+    For example, assuming bucket = bar and prefix = foo, and the bucket structure looks like this
```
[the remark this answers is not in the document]
yinweisu: I think I didn't get the question here.
So if you have `s3://bar/foo/a.txt`, you provide `bucket=bar` and `prefix=foo/a.txt`, then it will download `a.txt`
if you have `s3://bar/fooa/foo/a.txt`, you provide `bucket=bar` and `prefix=fooa/foo/a.txt`, then it will download the `a.txt` located in this subdir

common/src/autogluon/common/utils/s3_utils.py
```diff
@@ -29,3 +29,47 @@ def delete_s3_prefix(bucket, prefix):
...
+        └── bar/
+            ├── test.txt
+            └── foo/
+                ├── test2.txt
+                └── temp/
+                    └── test3.txt
+    This util will download foo to `local_path` and maintain the structure:
+        .
+        └── local_path/
+            └── foo/
```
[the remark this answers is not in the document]
yinweisu: Added `keep_root_dir` that default to False. If set to False, only contents of `foo` will be downloaded. If set to True, `foo` will be downloaded along with all its contents

core/src/autogluon/core/models/ensemble/fold_fitting_strategy.py
yinweisu: This is mainly serving as a general interface so that subclass can provide its own implementation. In the case of no syncing is needed, I think it makes sense to just leave an empty implementation there.

common/src/autogluon/common/utils/s3_utils.py
```diff
@@ -45,31 +59,87 @@ def download_s3_folder(bucket, prefix, local_path="."):
...
+        The local path to download the object/folder into
+    error_if_exists: bool
+        Whether to raise an error if the root folder exists already
+    delete_if_exists: bool
+        Whether to delete the root folder and all contents within if the root folder exists already
+    dry_run: bool
+        Whether to perform the directory creation and file downloading.
+        If True, will isntead log every file that will be downloaded and every directory that will be created
+    verbose: bool
+        Whether to log detailed loggings
```
Innixma, line 82: Specify defaults in docstrings when they exist via `error_if_exists: bool, default = True`, ditto for all other instances.

common/src/autogluon/common/utils/s3_utils.py
```diff
@@ -45,31 +59,87 @@ def download_s3_folder(bucket, prefix, local_path="."):
...
+    bucket: str
+        The name of the bucket
+    prefix: str
+        The prefix of the object/folder to be downloaded
+    local_path: str
+        The local path to download the object/folder into
+    error_if_exists: bool
+        Whether to raise an error if the root folder exists already
+    delete_if_exists: bool
+        Whether to delete the root folder and all contents within if the root folder exists already
```
Innixma, line 77: Clarify:

root folder -> local root folder

Add: 

if `error_if_exists=True`, deletion will not occur.

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -20,6 +20,7 @@
 from autogluon.common.utils.lite import disable_if_lite_mode
 from autogluon.common.utils.log_utils import DuplicateFilter
 from autogluon.common.utils.resource_utils import ResourceManager, RayResourceManager
+from autogluon.common.utils.resource_utils import get_resource_manager
```
Innixma, line 23: nit: can be on same line as L22

core/src/autogluon/core/models/ensemble/fold_fitting_strategy.py
```diff
@@ -389,10 +390,8 @@ def _ray_fit(
     fold_model.fit_time = time_train_end_fold - time_start_fold
     fold_model, pred_proba = _ray_predict_oof(fold_model, X_val_fold, y_val_fold,
                                               time_train_end_fold, resources['num_cpus'], save_bag_folds)
-    model_save_path = None
-    if model_sync_path is not None:
-        model_save_path = model_sync_path + f"{fold_model.name}/"
-    fold_model.save(model_save_path)
+    model_save_path = model_sync_path + f"{fold_model.name}/" if model_sync_path is not None else None
```
Innixma, line 393: `f"{fold_model.name}/"` , is there concern with cross-OS compatibility by the use of `/`?

I'd recommend at least adding a docstring for `model_sync_path` as it isn't immediately clear what this is doing.

core/src/autogluon/core/models/ensemble/fold_fitting_strategy.py
```diff
@@ -760,4 +777,4 @@ def __init__(self, **kwargs):
 
     def _sync_model_artifact(self, local_path, model_sync_path):
         bucket, path = s3_path_to_bucket_prefix(model_sync_path)
-        download_s3_folder(bucket, path, local_path)
+        download_s3_folder(bucket, path, local_path, keep_root_dir=True, error_if_exists=False)
```
Innixma, line 780: 1. `keep_root_dir` doesn't exist.
2. Use keyword args

core/src/autogluon/core/utils/try_import.py
```diff
@@ -53,7 +53,7 @@ def try_import_mxnet():
 
 
 def try_import_ray() -> ModuleType:
-    RAY_MAX_VERSION = '2.3.0'
+    RAY_MAX_VERSION = '3.0.0'
```
Innixma, line 56: add inline comment that this is a non-inclusive max version

yinweisu: This is not meant to be here.. It's needed to enable me easily test the distributed functionality. Will revert

core/src/autogluon/core/models/ensemble/fold_fitting_strategy.py
yinweisu: Something probably went wrong when I was building the wheel and testing this functionality on ray cluster. This should be caught. I'll fix it and re-test it.

core/src/autogluon/core/models/abstract/abstract_model.py
yinweisu: I remember when using `black` it will reformat the import to saperate Class imported and function imported. That's why I did it here. I tried to search for the PEP rule but not sure what's the correct way to find it

core/src/autogluon/core/models/ensemble/fold_fitting_strategy.py
yinweisu: This path is a s3 path so we cannot use utils like `os.path.join`. I'll add a comment

core/src/autogluon/core/models/abstract/abstract_model.py
Innixma: ok, not a big deal, I'm not familiar with the official guidelines here

Innixma (approved):
LGTM, great work!

PR #3055, opened 2023-03-16
[AutoMM] Fix NER HPO

zhiqiangdon:
*Issue #, if available:*

*Description of changes:*
`filter_hyperparameters ` requires column type `text_ner` to keep the `ner_text` model. Post-process the `column_types` from `infer_column_types` to replace column type `text` with `text_ner`.

### DIFF
--- a/multimodal/src/autogluon/multimodal/predictor.py
+++ b/multimodal/src/autogluon/multimodal/predictor.py
@@ -86,6 +86,7 @@ from .data.infer_types import (
     infer_problem_type_output_shape,
     infer_rois_column_type,
     is_image_column,
+    infer_ner_column_type,
 )
 from .data.preprocess_dataframe import MultiModalFeaturePreprocessor
 from .matcher import MultiModalMatcher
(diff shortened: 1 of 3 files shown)


multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -710,6 +711,9 @@ def fit(
             label_columns=self._label_column,
             provided_column_types=column_types,
         )
+        if self._problem_type == NER:
+            column_types = infer_ner_column_type(column_types=column_types)
```
liangfu, line 715: Can we make this problem_type agnostic ?

liangfu (approved):
LGTM, thanks for the contribution. This unifies the HPO support for all problem types.

multimodal/src/autogluon/multimodal/predictor.py
zhiqiangdon: Good suggestion! Moved `infer_ner_column_type ` into `infer_column_types`.

cheungdaven (approved):
lgtm

PR #3056, opened 2023-03-16
[eda][api][docs] Introduced dot notation for properties to simplify components customization UX

gradientsky:
*Description of changes:*
- added support for dot-notation config to pass `fig_args` and `chart_args`; updated docs to document supported properties
- updated `quick_fit` to save model to state by default
- `covariate_shift` doesn't render categorical variables to highlight shift since they are not helpful
- updated docs to include `explain` modules
- [bugfix] `explain`: fixed columns handling when `rows` have different column order than the originally trained model so that `shap` doesn't complain

### Examples

Dot notation for composite components now is more compact:
```python
auto.target_analysis(
    train_data=df_train, 
    label=target_col,
    fig_args={
        f'chart.{target_col}.figsize': (4,4),
        'chart.Sex.figsize': (4,4),
        'correlation.figsize': (1,1),
    },
    chart_args={
        f'chart.{target_col}.color': 'orange',
        'chart.Sex.color': 'green',
        'correlation.cmap': 'coolwarm',
    },
)
```

image

### DIFF
--- a/docs/tutorials/eda/components/autogluon.eda.explain.md
+++ b/docs/tutorials/eda/components/autogluon.eda.explain.md
@@ -1,0 +1,62 @@
+```{eval-rst}
+.. role:: hidden
+    :class: hidden-section
+```
+
+# Components: explain
+
+## autogluon.eda.visualization.explain
+
+```{eval-rst}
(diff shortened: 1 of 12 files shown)


gidler (approved):
Looks great to me!

PR #3062, opened 2023-03-17
[timeseries] Implement OOF prediction caching

shchur:
This is part 1 of the improved multi-window backtesting for time series (part 2: #3051).

*Description of changes:*
- Cache the predictions on the validation set for all time series models (as a model attribute + saved to disk). This ensures that we only call `predict` on the `val_data` once for all models. This saves training time and deals with the fact that predictions are non-deterministic for some models.
- Ensure that the reported prediction time for the test/val sets corresponds to the actual **prediction** time (before it included the time spent slicing the data / computing the metric).
- Simplify scoring logic in `AbstractTimeSeriesModel` and `AbstractTimeSeriesTrainer`
- Fix the bug where `AutoGluonTabularModel` would create an empty directory at initialization.

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
+++ b/timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
@@ -2,10 +2,11 @@ import copy
 import logging
 import os
 import time
-from typing import Any, Dict, List, Optional, Union
+from typing import Any, Dict, List, Optional, Tuple, Union
 
 import autogluon.core as ag
 from autogluon.common.savers import save_pkl
+from autogluon.common.loaders import load_pkl
 from autogluon.core.hpo.exceptions import EmptySearchSpace
 from autogluon.core.hpo.executors import HpoExecutor
 from autogluon.core.models import AbstractModel
@@ -53,6 +54,8 @@ class AbstractTimeSeriesModel(AbstractModel):
         Hyperparameters that will be used by the model (can be search spaces instead of fixed values).
         If None, model defaults are used. This is identical to passing an empty dictionary.
     """
+
+    _oof_filename = "oof.pkl"
 
     # TODO: refactor "pruned" methods after AbstractModel is refactored
     predict_proba = None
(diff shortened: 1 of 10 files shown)


timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
```diff
@@ -167,28 +167,31 @@ def __init__(
...
         self.feat_static_cat_cardinality: List[int] = []
 
-    def save(self, path: str = None, **kwargs) -> str:
-        if path is None:
-            path = self.path
-        path = Path(path)
-        path.mkdir(exist_ok=True)
-
+    def save(self, path: str = None, verbose: bool = True) -> str:
+        # The GluonTS predictor will be serialized using custom logic
```
canerturkmen, line 171: is this stating just what's happening here or is it a TODO?

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -336,6 +335,11 @@ def save(self) -> None:
 
         self.models = models
 
+    def get_model_oof_predictions(self, model_name: str) -> TimeSeriesDataFrame:
```
canerturkmen, line 338: should any of this, `slice_for_scoring` and `score_with_predictions` be marked private (single underscore)?

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
```diff
@@ -273,39 +311,53 @@ def predict(
         """
         raise NotImplementedError
 
-    def predict_for_scoring(self, data: TimeSeriesDataFrame, **kwargs):
-        """Given a dataset, truncate the last `self.prediction_length` time steps and forecast these
-        steps with previous history. This method produces predictions for the *last* `self.prediction_length`
-        steps of the *given* time series, in order to be used for validation or scoring.
+    def slice_data_for_scoring(
```
canerturkmen, line 314: these two methods reside both here and in the trainer, and replicate the exact same logic. should they maybe move somewhere else and have only one implementation?

canerturkmen (commented):
looks great! added some comments.

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
shchur: 1. Moved `slice_data_for_scoring` to `TimeSeriesDataFrame.get_model_inputs_for_scoring`
2. Moving `score_with_predictions` seems problematic:
    - We could only keep it as a method of `AbstractTimeSeriesModel`, but then inside the trainer we will need to load each model twice - once to predict & once to compute the score. This can be slow for local models that cache all predictions.
    - Could be moved to a static method in `utils`, but then will require additionally passing `target`, `prediction_length`, `metric` every time the method is invoked. 

     So I think it would be easier to keep `score_with_predictions` duplicated.

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -495,17 +498,12 @@ def _train_and_save(
 
             model = self._train_single(train_data, model, val_data=val_data, time_limit=time_limit)
             fit_end_time = time.time()
-
-            val_score = model.score(val_data) if val_data is not None else None
-            model.val_score = val_score
-
-            pred_end_time = time.time()
-
             model.fit_time = model.fit_time or (fit_end_time - fit_start_time)
```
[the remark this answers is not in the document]
shchur: Addressed in another comment.

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
```diff
@@ -331,17 +383,20 @@ def score(self, data: TimeSeriesDataFrame, metric: str = None, **kwargs) -> floa
...
+
+    def score_and_cache_oof(self, val_data: TimeSeriesDataFrame) -> None:
+        """Compute val_score, predict_time and cache out-of-fold (OOF) predictions."""
+        if self.val_score is not None or self.predict_time is not None or self._oof_predictions is not None:
+            raise ValueError(f"Model {self.name} has already been scored on OOF data!")
+
+        past_data, known_covariates = self.slice_data_for_scoring(val_data)
+        predict_start_time = time.time()
+        self._oof_predictions = self.predict(past_data, known_covariates=known_covariates)
+        self.predict_time = time.time() - predict_start_time
```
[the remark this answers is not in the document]
canerturkmen: OK, let's go with that.

canerturkmen (approved):
LGTM. Thanks!

PR #3064, opened 2023-03-21
[multimodal] compile without saving to disk

liangfu:
*Issue #, if available:*

the /opt/ml/model directory on sagemaker endpoint is on a read-only file system

*Description of changes:*

1. write onnx model file and tensorrt cache on /tmp directory

### DIFF
--- a/multimodal/src/autogluon/multimodal/utils/export.py
+++ b/multimodal/src/autogluon/multimodal/utils/export.py
@@ -1,5 +1,6 @@
 import logging
 import os
+import tempfile
 import warnings
 from collections import defaultdict, namedtuple
 from typing import Dict, List, Optional, Union
@@ -193,16 +194,17 @@ class ExportMixin:
                 raise ValueError(f"unsupported column type: {col_type}")
         data = pd.DataFrame.from_dict(data_dict)
 
-        onnx_path = self.export_onnx(
-            data=data, path=os.path.join("/tmp", os.path.basename(self.path)), truncate_long_and_double=True
-        )
-        onnx_module = OnnxModule(onnx_path, providers)
-        onnx_module.input_keys = self._model.input_keys
-        onnx_module.prefix = self._model.prefix
-        onnx_module.get_output_dict = self._model.get_output_dict
+        onnx_module = None
+        with tempfile.TemporaryDirectory() as tmpdirname:
+            onnx_path = self.export_onnx(data=data, path=tmpdirname, truncate_long_and_double=True)
 
-        # To use the TensorRT module for prediction, simply replace the _model in the predictor
-        self._model = onnx_module
+            onnx_module = OnnxModule(onnx_path, providers)
+            onnx_module.input_keys = self._model.input_keys
+            onnx_module.prefix = self._model.prefix
+            onnx_module.get_output_dict = self._model.get_output_dict
+
+            # To use the TensorRT module for prediction, simply replace the _model in the predictor
+            self._model = onnx_module
 
         # Evaluate and cache TensorRT engine files
         logger.info("Compiling ... (this may take a few minutes)")
--- a/multimodal/src/autogluon/multimodal/utils/export.py
+++ b/multimodal/src/autogluon/multimodal/utils/export.py
@@ -193,7 +193,9 @@ class ExportMixin:
                 raise ValueError(f"unsupported column type: {col_type}")
         data = pd.DataFrame.from_dict(data_dict)
 
-        onnx_path = self.export_onnx(data=data, path=self.path, truncate_long_and_double=True)
+        onnx_path = self.export_onnx(
+            data=data, path=os.path.join("/tmp", os.path.basename(self.path)), truncate_long_and_double=True
+        )
         onnx_module = OnnxModule(onnx_path, providers)
         onnx_module.input_keys = self._model.input_keys
         onnx_module.prefix = self._model.prefix
--- a/multimodal/src/autogluon/multimodal/utils/export.py
+++ b/multimodal/src/autogluon/multimodal/utils/export.py
@@ -1,6 +1,6 @@
+import io
 import logging
 import os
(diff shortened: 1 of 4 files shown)


multimodal/src/autogluon/multimodal/utils/export.py
```diff
@@ -193,14 +194,17 @@ def optimize_for_inference(
                 raise ValueError(f"unsupported column type: {col_type}")
         data = pd.DataFrame.from_dict(data_dict)
 
-        onnx_path = self.export_onnx(data=data, path=self.path, truncate_long_and_double=True)
-        onnx_module = OnnxModule(onnx_path, providers)
-        onnx_module.input_keys = self._model.input_keys
-        onnx_module.prefix = self._model.prefix
-        onnx_module.get_output_dict = self._model.get_output_dict
+        onnx_module = None
+        with tempfile.TemporaryDirectory() as tmpdirname:
```
yinweisu, line 198: I think this method creates a temporary folder under the current directory, i.e. ./temp_abcd.
If the intention of this PR is to write artifacts to /tmp folder, I don't think this does the job.

On the other hand, the cloud module encountered such problem for timeseries predictor as well because it's writing artifacts to the endpoint while inferencing. I think a better way to deal with such an issue is to expose a parameter that allows the user to customize where to save the artifact instead of manually write it to /tmp folder

liangfu: good point. i've moved to utilize BytesIO instead.

multimodal/src/autogluon/multimodal/utils/export.py
```diff
@@ -130,13 +131,14 @@ def export_onnx(
         input_keys = self._model.input_keys
         input_vec = [batch[k] for k in input_keys]
 
-        # Infer default onnx path, and create parent directory if needed
+        # Write to BytesIO if path argument is not provided
         if not path:
```
yinweisu, line 135: I would recommend changing this to `if path is None` explicitly to avoid bugs as python is not a strong typed language and many things could trigger the `if not path` clause.

multimodal/src/autogluon/multimodal/utils/export.py
```diff
@@ -130,13 +131,14 @@ def export_onnx(
         input_keys = self._model.input_keys
         input_vec = [batch[k] for k in input_keys]
 
-        # Infer default onnx path, and create parent directory if needed
+        # Write to BytesIO if path argument is not provided
```
yinweisu, line 134: Let's add some docstring on why we decided to do so as it is non-typical to write to BytesIO if no path is provided. Generally this would mean write to the current folder.

multimodal/tests/unittests/others/test_deployment_onnx.py
```diff
@@ -185,12 +185,7 @@ def test_onnx_optimize_for_inference(dataset_name, model_names, text_backbone, i
         predictor_opt = MultiModalPredictor.load(path=model_path)
         predictor_opt.optimize_for_inference(providers=providers)
 
-        # Check existence of the exported onnx model file and tensorrt cache files
-        onnx_path = os.path.join(predictor_opt.path, "model.onnx")
-        assert os.path.exists(onnx_path), f"onnx model file not found at {onnx_path}"
-        if providers == None or providers == ["TensorrtExecutionProvider"]:
-            trt_cache_dir = os.path.join(predictor_opt.path, "model_trt")
-            assert len(os.listdir(trt_cache_dir)) >= 2, f"tensorrt cache model files are not found in {trt_cache_dir}"
```
yinweisu, line 193: We probably want to test the logic for `export_onnx` when path is provided as well.

multimodal/src/autogluon/multimodal/utils/onnx.py
```diff
@@ -48,7 +48,7 @@ class OnnxModule(object):
     so that we can predict with TensorRT by simply replacing predictor._model with OnnxModule.
     """
 
-    def __init__(self, onnx_path: str, providers: Optional[Union[dict, List[str]]] = None):
+    def __init__(self, onnx_path: Union[str, bytes], providers: Optional[Union[dict, List[str]]] = None):
```
yinweisu, line 51: Update the docstring, and describe what bytes it is expecting

multimodal/src/autogluon/multimodal/utils/onnx.py
```diff
@@ -60,15 +60,21 @@ def __init__(self, onnx_path: str, providers: Optional[Union[dict, List[str]]] =
...
+        else:
+            if not os.path.exists(onnx_path):
+                raise FileNotFoundError(f"failed to located onnx file at {onnx_path}")
 
-        logger.info("Loading ONNX file from path {}...".format(onnx_path))
-        onnx_model = onnx.load(onnx_path)
+            logger.info("Loading ONNX file from path {}...".format(onnx_path))
+            onnx_model = onnx.load(onnx_path)
 
         if providers == None:
```
yinweisu, line 72: use `if providers is None` as a better practice: https://stackoverflow.com/questions/3257919/what-is-the-difference-between-is-none-and-none

liangfu: updated accordingly

multimodal/tests/unittests/others/test_deployment_onnx.py
liangfu: test logic added.

multimodal/src/autogluon/multimodal/utils/export.py
liangfu: changed as suggested.

multimodal/src/autogluon/multimodal/utils/onnx.py
liangfu: docstring updated.

yinweisu (approved):
LGTM!

PR #3070, opened 2023-03-21
[multimodal] fix num_gpus

liangfu:
*Issue #, if available:*

num_gpus always become 1 when calling predictor.evaluate after fit, therefore prevent us from supporting multi-gpu.

*Description of changes:*

### DIFF
--- a/multimodal/src/autogluon/multimodal/utils/inference.py
+++ b/multimodal/src/autogluon/multimodal/utils/inference.py
@@ -482,7 +482,7 @@ def predict(
     if predictor._problem_type == OBJECT_DETECTION:
         strategy = "ddp"
 
-    if strategy == "ddp" and predictor._fit_called:
+    if strategy == "ddp" and predictor._fit_called and predictor._problem_type == OBJECT_DETECTION:
         num_gpus = 1  # While using DDP, we can only use single gpu after fit is called
 
     if num_gpus <= 1:


multimodal/src/autogluon/multimodal/utils/inference.py
```diff
@@ -482,7 +482,7 @@ def predict(
     if predictor._problem_type == OBJECT_DETECTION:
         strategy = "ddp"
 
-    if strategy == "ddp" and predictor._fit_called:
+    if strategy == "ddp" and predictor._fit_called and predictor._problem_type == OBJECT_DETECTION:
```
FANGAreNotGnu, line 485: Have you tested if it works for other problem type? While working on object detection task, the blocker is a cuda barrier which I think may not only affect the detection problem type

liangfu: multi-gpu works for other problem types, since we are using DP. I think @zhiqiangdon would have more detailed.

zhiqiangdon: > multi-gpu works for other problem types, since we are using DP. I think @zhiqiangdon would have more detail.

Here we try to allow using DDP for other problem types. Right?

liangfu: this single-line change isn't quite related to DDP yet, actually it would be DP for other problem types.

zhiqiangdon: This if condition is about `ddp`. Wondering how it affects `dp`.

liangfu: Yes, you're right, this `if` condition is about ddp, but it should only affect OBJECT_DETECTION problem type, not other problem types.

it shouldn't affect dp for other problem types.

PR #3077, opened 2023-03-23
[multimodal] Update layer id for yolox

FANGAreNotGnu:
fix bug in "two stage" setting and enable layerwise decay for yolox, tested on pothole dataset for correctness.


cheungdaven (commented):
do we need a test case for it?

FANGAreNotGnu:
> do we need a test case for it?

Sure, let me combine this PR with next one on data augmentation for yolox and share a test case.

PR #3090, opened 2023-03-29
[AutoMM] Fix tests

zhiqiangdon:
*Issue #, if available:*
https://github.com/autogluon/autogluon/actions/runs/4545834564/jobs/8013765308
Some dataloader workers are killed accidentally. I have run the test locally and didn't see any issue. I suspect some recent updates in our CI environment may cause the issue.

*Description of changes:*
Set `num_workers` and `num_workers_evaluation` to 0 in tests, which follows the practices of unit tests in AutoMM.

### DIFF
--- a/tabular/tests/unittests/models/test_automm.py
+++ b/tabular/tests/unittests/models/test_automm.py
@@ -4,7 +4,7 @@ import pytest
 @pytest.mark.gpu
 def test_automm_sts(fit_helper):
     fit_args = dict(
-        hyperparameters={'AG_AUTOMM': {}},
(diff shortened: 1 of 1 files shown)


liangfu:
Thanks for the fix. I don't think changing test cases would prevent users from being trapped into similar issues.

tonyhoo (approved):

gidler (approved):

zhiqiangdon:
> Thanks for the fix. I don't think changing test cases would prevent users from being trapped into similar issues.

Agree. This is just a quick fix. We can explore the issue further. Feel free to dive into it and give a better solution.

PR #3094, opened 2023-03-29
[multimodal]Add MultiImageMix Dataset for Object Detection

FANGAreNotGnu:
- Refactor Processors
    - Able to only load data in processor, then perform some transforms that requires access to other rows of data in Dataset class
- Add a MultiImageMixDataset for YOLOX
    - A bit hardcoding, since currently the input format of object detection is fixed.
    - Enables Mosaic and Mixup transforms for detection problems, about 2% performance increase.
- Refine layer_id for yolox (enables layerwise_decay and fixes two_stage_lr)
    - There is performance drop if using finer layer id in Pothole dataset, pending larger benchmarking

**Previous tests have covered changes in this PR.**


multimodal/src/autogluon/multimodal/data/process_categorical.py
```diff
@@ -89,6 +89,7 @@ def __call__(
         categorical_features: Dict[str, int],
         feature_modalities: Dict[str, Union[int, float, list]],
         is_training: bool,
+        load_only: bool = False,
```
zhiqiangdon, line 92: Do we need to add this flag in each processor class? How about restricting it only to the related processors and using an if condition when calling processors in `apply_data_processor`?

multimodal/src/autogluon/multimodal/data/datamodule.py
```diff
@@ -31,6 +32,7 @@ def __init__(
         predict_data: Optional[pd.DataFrame] = None,
         id_mappings: Optional[Union[Dict[str, Dict], Dict[str, pd.Series]]] = None,
         val_use_training_mode: bool = False,
+        model_config: Optional[str] = None,
```
zhiqiangdon, line 35: `model_config` doesn't fit the context of datamodule. A better way may be trying to build the special dataset outside and pass it to argument `dataset`. We can detect whether `dataset` is None. Calling it in `set_dataset ` would be the same as calling `BaseDataset`.

multimodal/src/autogluon/multimodal/data/process_categorical.py
FANGAreNotGnu: Done. I was thinking that later we may need to use it in other processors as well. But we can add when necessary.

multimodal/src/autogluon/multimodal/data/datamodule.py
FANGAreNotGnu: This new dataset is only used in training stage, and also this new dataset need model_config as input. An alternative way is to build a new data module.

zhiqiangdon: How about creating the special dataset in `_fit()` of predictor.py and passing it to datamodule initialization?

FANGAreNotGnu: Sure, it works for the current case, but it may not work if we have similar issues in the future.

multimodal/src/autogluon/multimodal/data/process_label.py
```diff
@@ -70,6 +70,7 @@ def __call__(
         labels: Dict[str, Union[int, float]],
         feature_modalities: Dict[str, Union[int, float, list]],
         is_training: bool,
+        load_only: bool = False,
```
zhiqiangdon, line 73: Does `mmdet_image` use label processor? I thought it uses only the mmdet processor.

FANGAreNotGnu: done 2 fixes we discussed offline and added a TODO here

zhiqiangdon (approved):
LGTM!

PR #3100, opened 2023-03-31
[eda] set default sample to 10k for eda.auto functions

gradientsky:
*Description of changes:*
Set default sample to 10k for `eda.auto` functions. This is to cap time needed to run the analysis and encourage fast iterations. The sample can be set to a higher value via `sample` parameter or removed via setting to `None`.


suzhoum (approved):
LGTM!

PR #3124, opened 2023-04-11
[eda] Anomaly detection analysis

gradientsky:
[Open In Colab](https://colab.research.google.com/drive/1E6ZVMYfGJrJT5YDVcs6PovtEYHLA5WTd?usp=sharing)

*Description of changes:*
- Added anomaly detection analysis and visualization components: `AnomalyDetector`, `AnomalyDetectorAnalysis`, `AnomalyScoresVisualization`
- Added `auto.detect_anomalies()` shortcut

### Anomaly Detection

This method is used to identify unusual patterns or behaviors in data that deviate significantly from the norm.
It's best used when finding outliers, rare events, or suspicious activities that could indicate fraud, defects, or system failures.

When interpreting anomaly scores, consider:

- `Threshold`: Determine a suitable threshold to separate normal from anomalous data points, based on domain knowledge or statistical methods.
- `Context`: Examine the context of anomalies, including time, location, and surrounding data points, to identify possible causes.
- `False positives/negatives`: Be aware of the trade-offs between false positives (normal points classified as anomalies) and false negatives (anomalies missed).
- `Feature relevance`: Ensure the features used for anomaly detection are relevant and contribute to the model's performance.
- `Model performance`: Regularly evaluate and update the model to maintain its accuracy and effectiveness.

It's important to understand the context and domain knowledge before deciding on an appropriate approach to deal with anomalies.
The choice of method depends on the data's nature, the cause of anomalies, and the problem being addressed.

The common ways to deal with anomalies:
- `Removal`: If an anomaly is a result of an error, noise, or irrelevance to the analysis, it can be removed from the dataset
    to prevent it from affecting the model's performance.
- `Imputation`: Replace anomalous values with appropriate substitutes, such as the mean, median, or mode of the feature,
    or by using more advanced techniques like regression or k-nearest neighbors.
- `Transformation`: Apply transformations like log, square root, or z-score to normalize the data and reduce the impact of extreme values.
    Absolute dates might be transformed into relative features like age of the item.
- `Capping`: Set upper and lower bounds for a feature, and replace values outside these limits with the bounds themselves.
    This method is also known as winsorizing.
- `Separate modeling`: Treat anomalies as a distinct group and build a separate model for them, or use specialized algorithms designed
    for handling outliers, such as robust regression or one-class SVM.
- `Incorporate as a feature`: Create a new binary feature indicating the presence of an anomaly, which can be useful if anomalies have predictive value.

#### Parameters
```
train_data: DataFrame
    training dataset
label: str
    target variable
test_data: Optional[pd.DataFrame], default = None
    test dataset
val_data: Optional[pd.DataFrame], default = None
    validation dataset
explain_top_n_anomalies: Optional[int], default = None
    explain the anomaly scores for n rows with the highest scores; don't perform analysis if value is `None` or `0`
show_top_n_anomalies: Optional[int], default = 10
    display n rows with highest anomaly scores
threshold_stds: float, default = 3
    specifies how many standard deviations above mean anomaly score considered as anomalies
    (only needed for visualization, does not affect scores calculation)
show_help_text:bool, default = True
    if `True` shows additional information how to interpret the data
state: Union[None, dict, AnalysisState], default = None
    pass prior state if necessary; the object will be updated during `anlz_facets` `fit` call.
sample: Union[None, int, float], default = 10000
    sample size; if `int`, then row number is used;
    `float` must be between 0.0 and 1.0 and represents fraction of dataset to sample;
    `None` means no sampling
    See also :func:`autogluon.eda.analysis.dataset.Sampler`
return_state: bool, default = False
    return state if `True`
fig_args: Optional[Dict[str, Any]], default = None,
    kwargs to pass into visualization component
chart_args: Optional[dict], default = None
    kwargs to pass into visualization component
anomaly_detector_kwargs
    kwargs to pass into :py:class:`~autogluon.eda.analysis.anomaly.AnomalyDetectorAnalysis`
```

#### State attributes
```
anomaly_detection.scores.<dataset>
    scores for each of the datasets passed into analysis (i.e. `train_data`, `test_data`)
state.anomaly_detection.anomalies.<dataset>
    data points considered as anomalies - original rows with added `score` column sorted in descending score order.
    defined by `threshold_stds` parameter
anomaly_detection.anomaly_score_threshold
    anomaly score threshold above which data points are considered as anomalies;
    defined by `threshold_stds` parameter
```

### Examples - `auto.detect_anomalies()`
```python

state = auto.detect_anomalies(
    train_data=x,
    test_data=x_test,
    label=target_col,
    threshold_stds=3,  # specifies how many standard deviations above mean anomaly score considered as anomalies
    show_top_n_anomalies=5,  # show a table with top-5 anomalies
    explain_top_n_anomalies=1,  # visualized top-1 anomaly as a waterfall plot
    return_state=True,
    # show_help_text=False,  # suppress help text if needed
    fig_args={
        'figsize': (6, 4)
    },
    chart_args={
        'normal.color': 'lightgrey',
        'anomaly.color': 'orange',
    }    
)

```
image
image
image

### Examples using raw primitives
```python
import autogluon.eda.analysis as eda
import autogluon.eda.visualization as viz

auto.analyze(
    train_data=train_data,
    test_data=test_data,
    label=label,
    state=state,
    sample=sample,
    anlz_facets=[
        eda.dataset.ProblemTypeControl(),
        eda.transform.ApplyFeatureGenerator(
            category_to_numbers=True,
            children=[
                eda.anomaly.AnomalyDetectorAnalysis(
                    store_explainability_data=store_explainability_data,
                    **anomaly_detector_kwargs
                ),
            ],
        ),
    ],
    viz_facets=[
        viz.anomaly.AnomalyScoresVisualization(
            threshold_stds=threshold_stds,
            headers=True, 
            fig_args=fig_args, 
            **chart_args
        ),
    ]
)
```


eda/src/autogluon/eda/utils/common.py
```diff
@@ -0,0 +1,38 @@
+from typing import Any, Dict
+
+__all__ = ["get_empty_dict_if_none", "expand_nested_args_into_nested_maps"]
+
+
+def get_empty_dict_if_none(value) -> dict:
```
gradientsky, line 6: This was just moved to simplify reuse

eda/src/autogluon/eda/utils/common.py
```diff
@@ -0,0 +1,38 @@
...
+    if value is None:
+        value = {}
+    return value
+
+
+def expand_nested_args_into_nested_maps(args: Dict[str, Any]) -> Dict[str, Any]:
```
gradientsky, line 12: This was just moved to simplify reuse

eda/src/autogluon/eda/analysis/anomaly.py
```diff
@@ -0,0 +1,342 @@
...
+    - `anomaly_detection.explain_rows_fns.<dataset>`
+        if `store_explainability_data=True`, then analysis will store helper functions into this
+        variable. The function can be used later via :py:meth:`~autogluon.eda.auto.simple.explain_rows`
+        and automatically pre-populates `train_data`, `model` and `rows` parameters when called
+        (see example for more details)
+
+
+    Parameters
+    ----------
+    n_folds: int, default = 5
```
yinweisu, line 228: Just curious, why 5 folds

eda/src/autogluon/eda/analysis/anomaly.py
```diff
@@ -0,0 +1,342 @@
...
+        number of folds to use when training detectors; default is 5 folds.
+    store_explainability_data: bool, default = False
+        if true
```
yinweisu, line 231: missing something?

eda/src/autogluon/eda/analysis/anomaly.py
```diff
@@ -0,0 +1,342 @@
...
+        self.store_explainability_data = store_explainability_data
+
+    def can_handle(self, state: AnalysisState, args: AnalysisState) -> bool:
+        args_present = self.all_keys_must_be_present(args, "train_data", "label")
+        no_nans = True
+        if args_present:
+            for ds, df in self.available_datasets(args):
+                cols_with_nas = [c for c in df.columns if df[c].dtype != "object" and df[c].hasnans]
+                if len(cols_with_nas) > 0:
+                    self.logger.warning(
```
yinweisu, line 271: Do we want to raise instead if they should do something before proceed

eda/src/autogluon/eda/state.py
```diff
@@ -57,7 +57,7 @@ def at_least_one_key_must_be_present(self, state: AnalysisState, *keys) -> bool:
             True if at least one key from the `keys` list is present in the state
         """
         for k in keys:
-            if k in state:
+            if k in state and state[k] is not None:
```
yinweisu, line 60: Doesn't matter but maybe this is simpler:
`if state.get(k, None) is not None`

yinweisu (approved):
LGTM! Had some minor comments

eda/setup.cfg
```diff
@@ -58,7 +58,6 @@ commands =
     pytest {posargs}
 
 [testenv:typecheck]
-basepython = python3.8
```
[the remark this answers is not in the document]
gradientsky: It is doing: it's restricting this task only to run on python 3.8. It was not supposed to be there. I removed it because I wanted to test on py3.8/3.9/3.10.

eda/src/autogluon/eda/analysis/anomaly.py
```diff
@@ -0,0 +1,342 @@
...
+
+    def fit_transform(self, train_data: pd.DataFrame) -> pd.Series:
+        """
+        Automatically creates cross-validation splits and fits detectors on each of them.
+        The scores produced for the training data are produced using out-of-folds predictions
+
+        Parameters
+        ----------
+        train_data: pd.DataFrame
+            training data; must be already pre-processed; numeric columns must have NaNs filled
```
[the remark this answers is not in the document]
gradientsky: Yes, the analysis handles it in `can_handle()`.

eda/src/autogluon/eda/analysis/anomaly.py
gradientsky: 80% data for the training. Not too many folds to train.

eda/src/autogluon/eda/analysis/anomaly.py
gradientsky: good catch  - fixed

eda/src/autogluon/eda/analysis/anomaly.py
gradientsky: No, we don't want to block calculation if has no pre-requisites. It just don't do analysis if `can_handle` returns `False` and logging warning with information what's missing.

eda/src/autogluon/eda/state.py
gradientsky: good suggestion; added

PR #3135, opened 2023-04-12
mm update

yinweisu:
*Issue #, if available:*
mm_det and mmocr version not updated in nightly image causing cloud tests to fail

*Description of changes:*
* Update version range of mm_det and mmocr
* Update image to use 0.7 DLC image as the base image


FANGAreNotGnu (approved):
LGTM!

PR #3142, opened 2023-04-14
Maximum resources

yinweisu:
*Issue #, if available:*
torch models training will be slowed by the usage of virtual cores.

*Description of changes:*
* Add maximum resources check when in distributed mode

Example run output with a newly launched cluster of 8 m5.24xlarge machine:
The training time matches a local run and appear to be normal now
```bash
Fitting 1 L1 models ...
Fitting model: NeuralNetFastAI_BAG_L1 ...
        Fitting 8 child models (S1F1 - S1F8) | Fitting with ParallelDistributedFoldFittingStrategy
        0.9224   = Validation score   (accuracy)
        286.06s  = Training   runtime
        3.44s    = Validation runtime
        0.9224   = Validation score   (accuracy)
        0.08s    = Training   runtime
        0.04s    = Validation runtime
AutoGluon training complete, total runtime = 298.45s ... Best model: "WeightedEnsemble_L2"
```

### DIFF
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -550,39 +550,14 @@ class AbstractModel:
             if user_specified_model_level_resource is not None:
                 user_specified_lower_level_resource = min(user_specified_model_level_resource * k_fold, system_resource, user_specified_total_resource)
         return user_specified_lower_level_resource
-
-    def _preprocess_fit_resources(self, silent=False, total_resources=None, parallel_hpo=False, **kwargs):
-        """
-        This function should be called to process user-specified total resources.
-        Sanity checks will be done to user-specified total resources to make sure it's legit.
-        When user-specified resources are not defined, will instead look at model's default resource requirements.
-        """
-        if 'num_cpus' in kwargs and 'num_gpus' in kwargs:
-            # This value will only be passed by autogluon through previous layers(i.e. bagged model to model base).
-            # We respect this value with highest priority
-            # They should always be set to valid values
-            enforced_num_cpus = kwargs.get('num_cpus', None)
-            enforced_num_gpus = kwargs.get('num_gpus', None)
-            assert enforced_num_cpus is not None and enforced_num_cpus != 'auto' and enforced_num_gpus is not None and enforced_num_gpus != 'auto'
-            # The logic below is needed because ray cluster is running some process in the backend even when it's ready to be used
-            # Trying to use all cores on the machine could lead to resource contention situation
-            # TODO: remove this logic if ray team can identify what's going on underneath and how to workaround
-            if DistributedContext.is_distributed_mode():
-                minimum_model_resources = self.get_minimum_resources(
-                    is_gpu_available=(enforced_num_gpus > 0)
-                )
-                minimum_model_num_cpus = minimum_model_resources.get('num_cpus', 1)
-                enforced_num_cpus = max(minimum_model_num_cpus, enforced_num_cpus - 2)  # leave some cpu resources for process running by cluster nodes
-            max_resources = self._get_maximum_resources()
-            max_num_cpus = max_resources.get("num_cpus", None)
-            max_num_gpus = max_resources.get("num_gpus", None)
-            if max_num_cpus is not None:
-                enforced_num_cpus = min(max_num_cpus, enforced_num_cpus)
-            if max_num_gpus is not None:
-                enforced_num_gpus = min(max_num_gpus, enforced_num_gpus)
-            kwargs["num_cpus"] = enforced_num_cpus
-            kwargs["num_gpus"] = enforced_num_gpus
-            return kwargs
+    
+    def _calculate_total_resources(
+        self,
+        silent: bool = False,
+        total_resources: Optional[Dict[str, Union[int, float]]] = None,
+        parallel_hpo: bool = False,
+        **kwargs
+        ):
         resource_manager = get_resource_manager()
         system_num_cpus = resource_manager.get_cpu_count()
         system_num_gpus = resource_manager.get_gpu_count_all()
@@ -676,7 +651,43 @@ class AbstractModel:
         kwargs['num_gpus'] = num_gpus
         if not silent:
             logger.log(15, f"\tFitting {self.name} with 'num_gpus': {kwargs['num_gpus']}, 'num_cpus': {kwargs['num_cpus']}")
+        
         return kwargs
+
+    def _preprocess_fit_resources(self, silent=False, total_resources=None, parallel_hpo=False, **kwargs):
+        """
+        This function should be called to process user-specified total resources.
+        Sanity checks will be done to user-specified total resources to make sure it's legit.
+        When user-specified resources are not defined, will instead look at model's default resource requirements.
+        """
+        if 'num_cpus' in kwargs and 'num_gpus' in kwargs:
+            # This value will only be passed by autogluon through previous layers(i.e. bagged model to model base).
+            # We respect this value with highest priority
+            # They should always be set to valid values
+            enforced_num_cpus = kwargs.get('num_cpus', None)
+            enforced_num_gpus = kwargs.get('num_gpus', None)
+            assert enforced_num_cpus is not None and enforced_num_cpus != 'auto' and enforced_num_gpus is not None and enforced_num_gpus != 'auto'
+            # The logic below is needed because ray cluster is running some process in the backend even when it's ready to be used
+            # Trying to use all cores on the machine could lead to resource contention situation
+            # TODO: remove this logic if ray team can identify what's going on underneath and how to workaround
+            max_resources = self._get_maximum_resources()
+            max_num_cpus = max_resources.get("num_cpus", None)
+            max_num_gpus = max_resources.get("num_gpus", None)
+            if max_num_gpus is not None:
+                enforced_num_gpus = min(max_num_gpus, enforced_num_gpus)
+            if DistributedContext.is_distributed_mode():
+                minimum_model_resources = self.get_minimum_resources(
+                    is_gpu_available=(enforced_num_gpus > 0)
+                )
+                minimum_model_num_cpus = minimum_model_resources.get('num_cpus', 1)
+                enforced_num_cpus = max(minimum_model_num_cpus, enforced_num_cpus - 2)  # leave some cpu resources for process running by cluster nodes
+            if max_num_cpus is not None:
+                enforced_num_cpus = min(max_num_cpus, enforced_num_cpus)
+            kwargs["num_cpus"] = enforced_num_cpus
+            kwargs["num_gpus"] = enforced_num_gpus
+            return kwargs
+        
+        return self._calculate_total_resources(silent=silent, total_resources=total_resources, parallel_hpo=parallel_hpo, **kwargs)
 
     def _register_fit_metadata(self, **kwargs):
         """
@@ -1665,7 +1676,7 @@ class AbstractModel:
         save_json.save(path=json_path, obj=info)
         return info
     
-    def _get_maximum_resources(self) -> Dict[str, Union[int,float]]:
+    def _get_maximum_resources(self) -> Dict[str, Union[int, float]]:
         """
         Get the maximum resources allowed to use for this model.
         This can be useful when model not scale well with resources, i.e. cpu cores.
@@ -1673,7 +1684,9 @@ class AbstractModel:
         
         Return
         ------
-        Dict[str, Union[int,float]]
+        Dict[str, Union[int, float]]
+            key, name of the resource, i.e. `num_cpus`, `num_gpus`
+            value, maximum amount of resources
         """
         return {}
         
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -568,15 +568,15 @@ class AbstractModel:
             # Trying to use all cores on the machine could lead to resource contention situation
             # TODO: remove this logic if ray team can identify what's going on underneath and how to workaround
             if DistributedContext.is_distributed_mode():
-                max_num_cpus = self._get_maximum_resources().get("num_cpus", None)
-                if max_num_cpus is not None:
-                    enforced_num_cpus = min(max_num_cpus, enforced_num_cpus)
                 minimum_model_resources = self.get_minimum_resources(
                     is_gpu_available=(enforced_num_gpus > 0)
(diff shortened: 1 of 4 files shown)


tabular/src/autogluon/tabular/models/knn/knn_model.py
```diff
@@ -242,6 +244,11 @@ def sample_func(chunk, frac):
             idx = set(idx)
             self._X_unused_index = [i for i in range(num_rows_max) if i not in idx]
         return self.model
+    
+    def _get_maximum_resources(self) -> Dict[str, float]:
+        return {
+            "num_cpus": min(32, ResourceManager.get_cpu_count())
+        }
```
Innixma, line 251: Shouldn't need to use `min` here, can just be `32` right?

Also, change default_resources:

```python3
    def _get_default_resources(self):
        # use at most 32 cpus to avoid OpenBLAS error: https://github.com/autogluon/autogluon/issues/1020
        num_cpus = min(32, ResourceManager.get_cpu_count())
        num_gpus = 0
        return num_cpus, num_gpus
```

to

```python3
    def _get_default_resources(self):
        num_cpus = ResourceManager.get_cpu_count()
        num_gpus = 0
        return num_cpus, num_gpus
```

Then, move the ` # use at most 32 cpus to avoid OpenBLAS error: https://github.com/autogluon/autogluon/issues/1020` to the maximum resources method

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -573,7 +573,10 @@ def _preprocess_fit_resources(self, silent=False, total_resources=None, parallel
                 )
                 minimum_model_num_cpus = minimum_model_resources.get('num_cpus', 1)
                 enforced_num_cpus = max(minimum_model_num_cpus, enforced_num_cpus - 2)  # leave some cpu resources for process running by cluster nodes
-                kwargs["num_cpus"] = enforced_num_cpus
+            max_num_cpus = self._get_maximum_resources().get("num_cpus", None)
+            if max_num_cpus is not None:
+                enforced_num_cpus = min(max_num_cpus, enforced_num_cpus)
+            kwargs["num_cpus"] = enforced_num_cpus
```
Innixma, line 579: do this for gpus too (make sure only 1 call to the method though)

Innixma: This actually gives a very elegant way of disabling GPUs for models, we can simply set `ag.max_gpus=0` for models that don't have GPU support.

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -1656,6 +1659,15 @@ def save_info(self) -> dict:
...
         save_json.save(path=json_path, obj=info)
         return info
+    
+    def _get_maximum_resources(self) -> Dict[str, float]:
+        """
+        Get the maximum resources allowed to use for this model.
+        This can be useful when model not scale well with resources, i.e. cpu cores.
+        Return empty dict if no maximum resources needed
+        """
+        return {}
```
Innixma, line 1669: This should check for `ag.max_gpus`, `ag.max_cpus`, however we can add this in a follow-up PR.

Ideally user should be able to do:

```
 # Force each fold to use at most 8 CPUs and never use GPU
hyperparameters = {
     'GBM': {'ag.max_cpus': 8, 'ag.max_gpus': 0}
}
```

```
# Force each fold to use at minimum 8 CPUs (for example, if bagging 8 folds with 32 cores, this forces to only do 4 folds in parallel, since each fold needs 8 CPUs)
hyperparameters = {
     'GBM': {'ag.min_cpus': 8}  
}
```

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -1656,6 +1659,15 @@ def save_info(self) -> dict:
         json_path = self.path + self.model_info_json_name
         save_json.save(path=json_path, obj=info)
         return info
+    
+    def _get_maximum_resources(self) -> Dict[str, float]:
+        """
+        Get the maximum resources allowed to use for this model.
+        This can be useful when model not scale well with resources, i.e. cpu cores.
+        Return empty dict if no maximum resources needed
+        """
```
Innixma, line 1668: Add expected output keys and value types.

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -1656,6 +1659,15 @@ def save_info(self) -> dict:
         json_path = self.path + self.model_info_json_name
         save_json.save(path=json_path, obj=info)
         return info
+    
+    def _get_maximum_resources(self) -> Dict[str, float]:
```
Innixma, line 1663: Union[int, float]

tabular/src/autogluon/tabular/models/tabular_nn/torch/tabular_nn_torch.py
```diff
@@ -547,6 +549,11 @@ def reduce_memory_size(self, remove_fit=True, requires_save=True, **kwargs):
 
     def _get_default_stopping_metric(self):
         return self.eval_metric
+    
+    def _get_maximum_resources(self) -> Dict[str, float]:
+        return {
+            "num_cpus": ResourceManager.get_cpu_count_psutil(logical=False)
+        }
```
Innixma, line 556: add reasoning

tabular/src/autogluon/tabular/models/fastainn/tabular_nn_fastai.py
```diff
@@ -529,6 +530,11 @@ def _estimate_memory_usage(self, X, **kwargs):
     def _get_hpo_backend(self):
         """Choose which backend(Ray or Custom) to use for hpo"""
         return RAY_BACKEND
+    
+    def _get_maximum_resources(self) -> Dict[str, float]:
+        return {
+            "num_cpus": ResourceManager.get_cpu_count_psutil(logical=False)
+        }
```
Innixma, line 537: add reasoning

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -573,7 +573,15 @@ def _preprocess_fit_resources(self, silent=False, total_resources=None, parallel
...
-                kwargs["num_cpus"] = enforced_num_cpus
+            max_resources = self._get_maximum_resources()
+            max_num_cpus = max_resources.get("num_cpus", None)
+            max_num_gpus = max_resources.get("num_gpus", None)
+            if max_num_cpus is not None:
+                enforced_num_cpus = min(max_num_cpus, enforced_num_cpus)
+            if max_num_gpus is not None:
+                enforced_num_gpus = min(max_num_gpus, enforced_num_gpus)
+            kwargs["num_cpus"] = enforced_num_cpus
+            kwargs["num_gpus"] = enforced_num_gpus
```
Innixma, line 584: What if max_num_gpus=0, but enforced_num_gpus=1? Currently it appears like we have `minimum_model_resources` that is passed `enforced_num_gpus > 0`. Should this instead be passed that value after it is updated by `max_num_gpus`?

Innixma, line 584: the logic starting at line 578 should be moved to its own method, currently I'm pretty confused trying to understand in what situations line 578+ is called and in which situations the earlier code is called.

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -1656,6 +1664,19 @@ def save_info(self) -> dict:
         json_path = self.path + self.model_info_json_name
         save_json.save(path=json_path, obj=info)
         return info
+    
+    def _get_maximum_resources(self) -> Dict[str, Union[int,float]]:
```
Innixma, line 1668: nit: missing space between `,` and `float`

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -1656,6 +1664,19 @@ def save_info(self) -> dict:
...
+        """
+        Get the maximum resources allowed to use for this model.
+        This can be useful when model not scale well with resources, i.e. cpu cores.
+        Return empty dict if no maximum resources needed
+        
+        Return
+        ------
+        Dict[str, Union[int,float]]
```
Innixma, line 1676: ditto

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -1656,6 +1664,19 @@ def save_info(self) -> dict:
         json_path = self.path + self.model_info_json_name
         save_json.save(path=json_path, obj=info)
         return info
+    
+    def _get_maximum_resources(self) -> Dict[str, Union[int,float]]:
+        """
+        Get the maximum resources allowed to use for this model.
+        This can be useful when model not scale well with resources, i.e. cpu cores.
+        Return empty dict if no maximum resources needed
```
Innixma, line 1672: This still doesn't tell me the typical keys that are important in the output (aka `num_cpus`, `num_gpus`)

tabular/src/autogluon/tabular/models/fastainn/tabular_nn_fastai.py
```diff
@@ -529,6 +530,12 @@ def _estimate_memory_usage(self, X, **kwargs):
     def _get_hpo_backend(self):
         """Choose which backend(Ray or Custom) to use for hpo"""
         return RAY_BACKEND
+    
+    def _get_maximum_resources(self) -> Dict[str, float]:
+        # fastai model trains slower when utilizing virtual cores and this issue scale up when the number of cpu cores increases
```
Innixma, line 535: Can we tie this to a GitHub issue?

Innixma: ditto for torch

core/src/autogluon/core/models/abstract/abstract_model.py
yinweisu: You right. This does become tricky though cause we are trying to reserve some cpus in the distributed scenario. We do so first and check if that's still above the max. If we reverse it, we'll be taking 2 more from the max. I guess this still sort of makes sense? But to avoid it, we need to check the total cpu available for the system and check if we are already using 2 less than it

tabular/src/autogluon/tabular/models/knn/knn_model.py
```diff
@@ -242,10 +244,16 @@ def sample_func(chunk, frac):
             idx = set(idx)
             self._X_unused_index = [i for i in range(num_rows_max) if i not in idx]
         return self.model
+    
+    def _get_maximum_resources(self) -> Dict[str, float]:
```
Innixma, line 248: typing consistency

tabular/src/autogluon/tabular/models/tabular_nn/torch/tabular_nn_torch.py
```diff
@@ -547,6 +549,12 @@ def reduce_memory_size(self, remove_fit=True, requires_save=True, **kwargs):
 
     def _get_default_stopping_metric(self):
         return self.eval_metric
+    
+    def _get_maximum_resources(self) -> Dict[str, float]:
```
Innixma, line 553: typing consistency

core/src/autogluon/core/models/abstract/abstract_model.py
yinweisu: Logic starting 578+ always called no matter distributed or not

Innixma: We can't take 2 more from the max, imagine if `max_cpus=1`, this would crash.

Innixma: I'm saying to have 

```
if max_num_gpus is not None:
                enforced_num_gpus = min(max_num_gpus, enforced_num_gpus)
```
Prior to the `if DistributedContext` logic.

The `max_num_cpus` logic can be after the `if DistributedContext` logic.

Innixma: Maybe line numbers are not best way of saying it, I mean the code after `return kwargs`

Starting with

```python3
resource_manager = get_resource_manager()
system_num_cpus = resource_manager.get_cpu_count()
...
```

tabular/src/autogluon/tabular/models/fastainn/tabular_nn_fastai.py
yinweisu: https://github.com/autogluon/autogluon/issues/3146

core/src/autogluon/core/models/abstract/abstract_model.py
yinweisu: sure

tabular/src/autogluon/tabular/models/fastainn/tabular_nn_fastai.py
Innixma: Please include github issue link in-line

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -550,39 +550,14 @@ def _process_user_provided_resource_requirement_to_calculate_total_resource_when
...
-            kwargs["num_gpus"] = enforced_num_gpus
-            return kwargs
+    
+    def _calculate_total_resources(
+        self,
+        silent: bool = False,
+        total_resources: Optional[Dict[str, Union[int, float]]] = None,
+        parallel_hpo: bool = False,
+        **kwargs
+        ):
```
Innixma, line 560: add docstring explaining this, add return type

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -676,8 +651,44 @@ def _preprocess_fit_resources(self, silent=False, total_resources=None, parallel
         kwargs['num_gpus'] = num_gpus
         if not silent:
             logger.log(15, f"\tFitting {self.name} with 'num_gpus': {kwargs['num_gpus']}, 'num_cpus': {kwargs['num_cpus']}")
+        
         return kwargs
 
+    def _preprocess_fit_resources(self, silent=False, total_resources=None, parallel_hpo=False, **kwargs):
```
Innixma, line 657: add return type

Innixma (approved):
LGTM! Added a few minor comments

core/src/autogluon/core/models/abstract/abstract_model.py
Innixma, line 657: add type hints

yinweisu:
Merging as unit tests for previous commits have passed, the most recent commit only added comments

PR #1351, opened 2021-10-22
Tabular: Switched KNN and RF to use refit_folds instead of use_child_oof

Innixma:
*Issue #, if available:*

*Description of changes:*

- Switched KNN and RF to use refit_folds instead of use_child_oof during bagging
- This results in the same model at test time for RF and KNN, but different out of fold predictions.
- While being 10x slower to train, this will reduce stack information leakage significantly while keeping inference speed identical. I would guess that assuming enough time is given, this will improve stack quality.

TODO:

- [ ] Benchmark

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #3148, opened 2023-04-15
Optimize refit_folds logic

Innixma:
*Issue #, if available:*

*Description of changes:*
- Optimize refit_folds logic
- This is an updated version of this PR: #1351
- Currently does not enable `refit_folds=True` for any model, but a follow-up PR will test enabling for KNN.
- `refit_folds=True` will fit a bagged ensemble of a given model and then immediately call refit_full, replacing the bag with a single model in a similar manner to `use_child_oof=True`, but supported for any model rather than only models that have `use_child_oof=True` supported.
- Fixed a bug where ParallelBagging ignored `save_folds` (SequentialBagging and ParallelBagging now produce the same result instead of different results when `save_folds=False`, which is relevant to reduce disk usage).
- Removed a redundant predict call on all of the data during refit_full via the new `_skip_oof` argument. This should drastically speed up refit_full for models that are slow in inference, and avoids needlessly storing unused OOF predictions in the model artifact, saving disk space.


yinweisu (approved):
LGTM!

PR #3155, opened 2023-04-17
fix that detection configs are missing in pip install

FANGAreNotGnu:
Fix for #3152 and #3082 

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #3158, opened 2023-04-19
Packages fix

gradientsky:
*Issue #, if available:*
#3155

*Description of changes:*
- Added `__init__.py` to python packages.

### DIFF
--- a/multimodal/setup.py
+++ b/multimodal/setup.py
@@ -83,11 +83,6 @@ if __name__ == "__main__":
         "configs/environment/*.yaml",
         "configs/distiller/*.yaml",
         "configs/matcher/*.yaml",
-        "configs/pretrain/detection/*.py",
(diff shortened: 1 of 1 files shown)


FANGAreNotGnu:
Cli and configs are not directly imported in other files. I'm not sure if we shall treat them as packages.

yinweisu (approved):

multimodal/setup.py
```diff
@@ -83,11 +83,6 @@
         "configs/environment/*.yaml",
         "configs/distiller/*.yaml",
         "configs/matcher/*.yaml",
```
zhiqiangdon, line 85: If we put `__init__.py` in `configs`, can we remove the yaml files here?

gradientsky: yes

PR #3160, opened 2023-04-19
[timeseries] Fix incorrect path used by MultiWindowModel

shchur:
*Description of changes:*
- Fix a bug, where `MultiWindowModel` tries to load the base model from the original path, where the model was saved initially, even if the model got moved.
- Make sure that `TimeSeriesPredictor.path` is correctly updated when loading from disk.

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/multi_window/multi_window_model.py
+++ b/timeseries/src/autogluon/timeseries/models/multi_window/multi_window_model.py
@@ -55,7 +55,6 @@ class MultiWindowBacktestingModel(AbstractTimeSeriesModel):
         self.info_per_val_window = []
 
         self.most_recent_model: AbstractTimeSeriesModel = None
-        self.most_recent_model_path: str = None
         super().__init__(**kwargs)
 
     def _fit(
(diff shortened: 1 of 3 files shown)


timeseries/src/autogluon/timeseries/models/multi_window/multi_window_model.py
```diff
@@ -117,7 +116,6 @@ def _fit(
 
         # Only the model trained on most recent data is saved & used for prediction
         self.most_recent_model = trained_models[0]
-        self.most_recent_model_path = trained_models[0].path
```
shchur, line 120: `self.most_recent_model_path` wasn't updated even if the model was moved to a new location on disk, leading to problems

timeseries/tests/unittests/models/test_multi_window_model.py
```diff
@@ -43,3 +47,18 @@ def test_when_val_data_passed_to_mw_model_fit_then_exception_is_raised(temp_mode
...
+    model = get_multi_window_deepar(path=original_path)
+    model.fit(train_data=DUMMY_TS_DATAFRAME)
+    model.save()
+    new_path = tempfile.mkdtemp() + os.sep
+    shutil.move(os.path.join(original_path, model.name), new_path)
+    loaded_model = model.load(os.path.join(new_path, model.name) + os.sep)
+    assert loaded_model.path.startswith(new_path)
+    assert loaded_model.most_recent_model.path.startswith(new_path)
+
+    shutil.rmtree(original_path)
```
yinweisu, line 63: Not an issue, but same thing could be done with tempfile as a context manager in an easier way:
```python
with tempfile.TemporaryDirectory() as temp_dir:
    ...
    # no need to remove this created dir. Will be cleaned up when getting out of the context
```

yinweisu (approved):
LGTM! Thanks for the fix!

timeseries/tests/unittests/models/test_multi_window_model.py
shchur: Thanks, will keep this in mind in the future.

PR #2865, opened 2023-02-08
Tabular: Fix error when loading with a different OS

Innixma:
*Issue #, if available:*
#2208

*Description of changes:*
Fixes exception during load when TabularPredictor was trained on Windows, and then loaded on MacOS/Linux.
Should also fix exception when trained on MacOS/Linux and loaded on Windows.

Because this PR introduces a new check on load based on a new variable, prior trained predictors on earlier versions will not be able to load using this PR.

**Note**: I have not formally tested this yet as I don't have a Windows machine on hand. Would appreciate if those affected can comment on if this PR fixes their problem.

Follow-up: Add unit test, tracked in #2863

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Issue #2208, opened 2022-10-13
Autogluon Tabular Predictor: IndexError: list index out of range


hkaurut:
Hi , I am using Autogluon Tabular predictor. The predictor is as follows:
predictor= TabularPredictor(
    label="ops",
    verbosity=4,
    problem_type="regression",
    path="agModels-predicttest1”,
    eval_metric='mean_absolute_error'
  ).fit(
    train_data=train_data_transformed.drop(
      ["spotcp", "ad_revenue", "pb_ops", "pb_spotcp"], axis=1),
    feature_generator=IdentityFeatureGenerator(),
    time_limit=3200,
    # visualizer='tensorboard',
    presets='medium_quality',   #'optimize_for_deployment', #'best_quality',
    hyperparameters={"GBM":{'num_boost_round': [ag.space.Int](http://ag.space.int/)(lower=50, upper=2000, default=100),
    'num_leaves': [ag.space.Int](http://ag.space.int/)(lower=64, upper=1024, default=256),
    'learning_rate': ag.space.Real(3e-2, 2e-1, default=1e-1,log=True),
    'max_depth':[ag.space.Int](http://ag.space.int/)(lower=6, upper=10, default=8),
    'early_stopping_round':10,'min_data_in_leaf': [ag.space.Int](http://ag.space.int/)(lower=10, upper=100, default=20)}},
    hyperparameter_tune_kwargs={'num_trials': 10,'searcher': 'auto','scheduler':'local'}
  )
With medium_quality presets and hyper parameter tuning, if I try to load the model after training as follows:
predictor_test1=TabularPredictor.load("/home/hadoop/agModels-predicttest1/“)
I get the following error:
Traceback (most recent call last):
 File "/tmp/1665687139040-0/zeppelin_python.py", line 158, in <module>
  exec(code, _zcUserQueryNameSpace)
 File "<stdin>", line 7, in <module>
 File "/usr/local/lib/python3.7/site-packages/autogluon/tabular/predictor/predictor.py", line 2901, in load
  predictor = cls._load(path=path)
 File "/usr/local/lib/python3.7/site-packages/autogluon/tabular/predictor/predictor.py", line 2833, in _load
  predictor._set_post_fit_vars(learner=learner)
 File "/usr/local/lib/python3.7/site-packages/autogluon/tabular/predictor/predictor.py", line 2787, in _set_post_fit_vars
  self._learner.persist_trainer(low_memory=True)
[...]

PR #3161, opened 2023-04-19
[Tabular] Added compatibility to load older models in different OS

eses-wk:
*Issue #, if available:*
#2208 

*Description of changes:*
Based on #2865, as the previous PR cannot load predictors from earlier versions,
> prior trained predictors on earlier versions will not be able to load using this PR.

Note: predictors trained in different OS can be loaded, yet `predict()` and `predict_proba()` for Ensemble Models still do not work due to `pickle.load` issue


Innixma:
Thanks for the PR @eses-wk! We have updated #2865 with an alternative solution after testing that should work in all cases (including the ensemble). If you get a chance to try it out, please let us know if it resolves your issue.

PR #3177, opened 2023-04-25
[timeseries] Implement recurrent forecasting model based on TabularPredictor and MLForecast

shchur:
*Description of changes:*
- Add a new model `RecursiveTabularModel` that uses the `TabularPredictor` to forecast time series 1 step into the future. The overall forecast is obtained by unrolling the model predictions over time.


timeseries/src/autogluon/timeseries/models/presets.py
```diff
@@ -38,6 +39,7 @@
     AutoETS=AutoETSModel,
     AutoARIMA=AutoARIMAModel,
     DynamicOptimizedTheta=DynamicOptimizedThetaModel,
+    RecursiveTabular=RecursiveTabularModel,
```
tonyhoo, line 42: which presets are we intended to add for this model?

timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
```diff
@@ -0,0 +1,301 @@
...
+
+    def predict(
+        self,
+        data: TimeSeriesDataFrame,
+        known_covariates: TimeSeriesDataFrame = None,
+        **kwargs,
+    ) -> TimeSeriesDataFrame:
+        predictions = self._predict_without_quantiles(data, known_covariates)
+        for q in self.quantile_levels:
+            predictions[str(q)] = predictions["mean"] + self.quantile_adjustments[q]
```
tonyhoo, line 297: Wondering how the quantile adjustment is done

timeseries/src/autogluon/timeseries/models/presets.py
shchur: I plan to update these in a follow-up PR.

timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
shchur: Here we just use the empirical distribution of the quantiles to generate dummy quantiles. I will add another tree-based model that is trained to predict the quantiles using the wQL loss in a future PR.

tonyhoo (approved):

PR #3180, opened 2023-04-26
[multimodal] Fix mmcv reference if import failed

FANGAreNotGnu:
Error message that guide user to install mmcv won't print before this fix.


zhiqiangdon (commented):
Before this fix, what would happen if mmcv is not installed?

FANGAreNotGnu:
> Before this fix, what would happen if mmcv is not installed?

The mmcv will not be imported when we init the dataset and the error will raise for the line `model_config: mmcv.utils.config.Config,`.

zhiqiangdon (commented):
CI is not passed.

FANGAreNotGnu:
> CI is not passed.

Fixed. It was because the import code was wrong.

zhiqiangdon (approved):
LGTM

PR #3183, opened 2023-04-28
[timeseries] Allow creating multiple configurations of the same model using `hyperparameters`

shchur:
*Description of changes:*
- Allow training multiple configurations of the same model using the `hyperparameters` kwarg, similar to how this is done in the `TabularPredictor`. For example, the following code will train 3 models `[ETS, ETS_2, ETS_seasonal]`:
    ```python
    TimeSeriesPredictor().fit(
        data,
        hyperparameters={
            "ETS": [
                {"seasonal": None},  # ETS
                {},  # ETS_2 -> suffix added to avoid name collision
                {"seasonal": "mul", "ag_args": {"name_suffix": "_mul"}},  # ETS_mul -> custom suffix
            ]
        }
    )
    ```
    By default, a suffix `_2`, `_3`, .. is added to avoid name collisions, but custom names/suffixes/prefixes can also be specified using the `ag_args` hyperparameter (just like in Tabular). Currently, only following `ag_args` are supported: `name`, `name_suffix`, `name_prefix`.

   This change will be helpful in the future (e.g., for zero-shot HPO), and brings closer the APIs of TimeSeries and Tabular.
- Move the logic for ensuring that `context_length >= 10` for all GluonTS models from `presets.py` to the `AbstractGluonTSModel` class, where it belongs.
- Set the default model `name` inside `AbstractTimeSeriesModel`, instead of repeating this code in each individual model class.

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/presets.py
+++ b/timeseries/src/autogluon/timeseries/models/presets.py
@@ -89,7 +89,6 @@ DEFAULT_CUSTOM_MODEL_PRIORITY = 0
 
 VALID_AG_ARGS_KEYS = {
     "name",
-    "name_main",
     "name_prefix",
     "name_suffix",
 }
--- a/timeseries/src/autogluon/timeseries/models/presets.py
+++ b/timeseries/src/autogluon/timeseries/models/presets.py
@@ -226,7 +226,9 @@ def get_preset_models(
             ag_args = model_hps.pop(ag.constants.AG_ARGS, {})
             for key in ag_args:
                 if key not in VALID_AG_ARGS_KEYS:
-                    logger.warning(f"WARNING: Unknown ag_args key: {key}")
+                    raise ValueError(
+                        f"Model {model_type} received unknown ag_args key: {key} (valid keys {VALID_AG_ARGS_KEYS})"
+                    )
             model_name_base = get_model_name(ag_args, model_type)
 
             model_type_kwargs = dict(
--- a/timeseries/src/autogluon/timeseries/models/presets.py
+++ b/timeseries/src/autogluon/timeseries/models/presets.py
@@ -1,6 +1,7 @@
 import copy
 import logging
-from typing import Any, Dict, List, Optional, Union
+import re
+from typing import Any, Dict, List, Optional, Type, Union
 
 import autogluon.core as ag
 import autogluon.timeseries as agts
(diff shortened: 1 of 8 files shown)


timeseries/src/autogluon/timeseries/models/presets.py
```diff
@@ -226,35 +216,56 @@ def get_preset_models(
...
-        )
+        model_hps_list = hyperparameters[model]
+        if not isinstance(model_hps_list, list):
+            model_hps_list = [model_hps_list]
+
+        for model_hps in model_hps_list:
+            ag_args = model_hps.pop(ag.constants.AG_ARGS, {})
+            for key in ag_args:
+                if key not in VALID_AG_ARGS_KEYS:
+                    logger.warning(f"WARNING: Unknown ag_args key: {key}")
```
tonyhoo, line 227: Shall we raise error instead of warning to avoid unintended behavior from original expectation?

timeseries/src/autogluon/timeseries/models/presets.py
```diff
@@ -226,35 +216,56 @@ def get_preset_models(
...
-        model_type_kwargs.pop("name", None)
-        increment = 1
-        while model.name in all_assigned_names:
-            increment += 1
-            model = model_type(name=f"{name_stem}_{increment}", **model_type_kwargs)
+            model_type_kwargs.pop("name", None)
+            increment = 1
+            while model.name in all_assigned_names:
+                increment += 1
+                model = model_type(name=f"{model_name_base}_{increment}", **model_type_kwargs)
```
tonyhoo, line 248: See tabs from the review, can you apply black formatter?

timeseries/src/autogluon/timeseries/models/presets.py
```diff
@@ -226,35 +216,56 @@ def get_preset_models(
...
 
-        if multi_window:
-            model = MultiWindowBacktestingModel(model_base=model, name=model.name, **model_type_kwargs)
+            if multi_window:
+                model = MultiWindowBacktestingModel(model_base=model, name=model.name, **model_type_kwargs)
 
-        all_assigned_names.add(model.name)
-        models.append(model)
+            all_assigned_names.add(model.name)
+            models.append(model)
```
tonyhoo, line 254: Same here

timeseries/src/autogluon/timeseries/models/presets.py
shchur: Agreed, changed this to `ValueError`.

timeseries/src/autogluon/timeseries/models/presets.py
shchur: I think this is an intentional change - there a new for loop that moved the indentation one level deeper. 

`black` is already a part of the `lint_check` CI step, so this code should be formatted correctly.

PR #3190, opened 2023-05-02
Add support for python 3.11

### LINKED ISSUE #2687: Support Python 3.11
Add support for Python 3.11 in AutoGluon.

Tentatively planned for AutoGluon v0.8 release in ~June 2023. Unknown if there will be blockers at this time.

- [x] Pre-req: Add Python 3.10 support (https://github.com/autogluon/autogluon/issues/2686)
- [x] https://github.com/autogluon/autogluon/pull/2739
- [x] https://github.com/catboost/catboost/issues/2213
  - [x] Note: Latest CatBoost seems to break on MacOS when Python != 3.11, need to investigate. 
- [x] ray 3.11 support
- [x] torchtext 3.11 support
- [x] torchvision 3.11 support

ddelange:
*Issue #, if available:* closes #2687

*Description of changes:*
- bump dependency versions for which cp311 wheels/support is available
- consequently add support for torch 2.0, torchmetrics 1.0, pytorch-lightning 2.0
- run CI on 3.11
- switch from cuda/cpu plus-notation to using pip's `--force-reinstall` flag


ddelange:
last commit due to `ValueError: You selected an invalid strategy name: `strategy=None`. It must be either a string or an instance of `pytorch_lightning.strategies.Strategy`. Example choices: auto, ddp, ddp_spawn, deepspeed, ... Find a complete list of options in our documentation at https://lightning.ai/`

Innixma:
@ddelange Thanks for giving this a try, this is a great start!

I suspect there might be a non-trivial amount of changes needed since we need to upgrade to torch 2.0 and pytorch-lightning 2.0, but at least we should be able to start testing things now that CatBoost supports Python 3.11.

gradientsky:
This is still broken for Clip models:
```python
import os
import numpy as np
import warnings
warnings.filterwarnings('ignore')
np.random.seed(123)

download_dir = './ag_automm_tutorial'
zip_file = 'https://automl-mm-bench.s3.amazonaws.com/petfinder_kaggle.zip'
from autogluon.core.utils.loaders import load_zip

import pandas as pd
dataset_path = download_dir + '/petfinder_processed'
train_data = pd.read_csv(f'{dataset_path}/train.csv', index_col=0)
test_data = pd.read_csv(f'{dataset_path}/dev.csv', index_col=0)
label_col = 'AdoptionSpeed'

image_col = 'Images'
train_data[image_col] = train_data[image_col].apply(lambda ele: ele.split(';')[0]) # Use the first image for a quick tutorial
test_data[image_col] = test_data[image_col].apply(lambda ele: ele.split(';')[0])


def path_expander(path, base_folder):
    path_l = path.split(';')
    return ';'.join([os.path.abspath(os.path.join(base_folder, path)) for path in path_l])

train_data[image_col] = train_data[image_col].apply(lambda ele: path_expander(ele, base_folder=dataset_path))
test_data[image_col] = test_data[image_col].apply(lambda ele: path_expander(ele, base_folder=dataset_path))

train_data = train_data.sample(500, random_state=0)
test_data = test_data.sample(100, random_state=0)

from autogluon.multimodal import AutoMMPredictor
predictor = AutoMMPredictor(label=label_col)
predictor.fit(
    train_data=train_data,
    hyperparameters={
        "model.names": ["clip", "timm_image", "hf_text", "numerical_mlp", "fusion_mlp"],
        "model.timm_image.checkpoint_name": "swin_small_patch4_window7_224",
        "model.hf_text.checkpoint_name": "google/electra-small-discriminator",
        "env.num_gpus": 1,
    },
    time_limit=20, # seconds
)

scores = predictor.evaluate(test_data, metrics=["accuracy"])
print(scores)
```

```
│ /home/ubuntu/Projects/autogluon/multimodal/src/autogluon/multimodal/models/fusion/fusion_mlp.py: │
│ 126 in input_keys                                                                                │
│                                                                                                  │
│   123 │   def input_keys(self):                                                                  │
│   124 │   │   input_keys = []                                                                    │
│   125 │   │   for m in self.model:                                                               │
│ ❱ 126 │   │   │   assert hasattr(m, "input_keys"), f"invalid model {type(m)}, which doesn't ha   │
│   127 │   │   │   input_keys += m.input_keys                                                     │
│   128 │   │   return input_keys                                                                  │
│   129                                                                                            │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
AssertionError: invalid model <class 'autogluon.multimodal.models.clip.CLIPForImageText'>, which doesn't have a 'input_keys' attribute
```

sxjscience:
Some failed tests point to `utils/map.py`. The version detection may not have taken effect: https://github.com/autogluon/autogluon/blob/7dfefbd5d9733f8dfbd6f77abd0ab6192ad1b932/multimodal/src/autogluon/multimodal/utils/map.py#L6-L7

I think it's safe to always use torchmetrics>0.12.0. @FANGAreNotGnu correct me if I'm wrong.

timeseries/setup.py
```diff
@@ -29,8 +29,8 @@
     "pandas",  # version range defined in `core/_setup_utils.py`
     "statsmodels>=0.13.0,<0.14",
     "gluonts>=0.12.4,<0.13",
-    "torch>=1.9,<1.14",
-    "pytorch-lightning>=1.7.4,<1.10.0",
+    "torch>=1.9,<2.1",
```
sxjscience, line 32: Due to security vulnerabilities of torch, we may need to force "torch>=1.13.1"

sxjscience:
The lightning upgrade will require some verification.

timeseries/setup.py
ddelange: in that case, the minimum torchvision version would go to 0.14.0 and your 0.12 comment becomes obsolete 👍

gradientsky:
> The lightning upgrade will require some verification.

I did smoke tests we usually ran for the containers: CLIP is broken (see the code above), but other scenarios looks ok.

timeseries/setup.py
Innixma: Do we want to be supporting multiple major versions of torch? That sounds intimidating to maintain. Is it possible to drop support of torch 1.x and only support 2.x?

.github/workflows/platform_tests-command.yml
```diff
@@ -45,7 +45,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [macos-latest, windows-latest, ubuntu-latest]
-        python: ["3.8", "3.9", "3.10"]
+        python: ["3.8", "3.9", "3.10", "3.11"]
```
yinweisu, line 48: To trigger a platform test, you should be able to do it with comment on this PR by
`/platform_tests ref=ddelange:cp311`

However, our platform tests are currently failing because of other issues. @gradientsky @Innixma We'll want to fix the platform tests to unblock this PR

.github/workflow_scripts/env_setup.sh
```diff
@@ -16,21 +16,23 @@ function setup_build_contrib_env {
 }
 
 function setup_torch_gpu {
-    # Security-patched torch.
-    python3 -m pip install torch==1.13.1+cu116 torchvision==0.14.1+cu116 torchaudio==0.13.1+cu116 --extra-index-url https://download.pytorch.org/whl/cu116
+    PIP_EXTRA_INDEX_URL=https://download.pytorch.org/whl/cu118 reinstall_torch
```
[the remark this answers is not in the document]
ddelange: sounds like good practice to keep that closed off 💪

fwiw, I don't mind if a maintainer opens a new PR in favour of this one

timeseries/setup.py
ddelange: e.g. torch-lightning supports 1.11-2.0, I think it should be ok to support v1 a while longer:

> torch.compile is the main API for PyTorch 2.0, which wraps your model and returns a compiled model. It is a fully additive (and optional) feature and hence 2.0 is 100% backward compatible by definition.

ref https://pytorch.org/blog/pytorch-2.0-release/

personally I as user would be fine with >=2 as well

FANGAreNotGnu:
> Some failed tests point to `utils/map.py`. The version detection may not have taken effect:
> 
> https://github.com/autogluon/autogluon/blob/7dfefbd5d9733f8dfbd6f77abd0ab6192ad1b932/multimodal/src/autogluon/multimodal/utils/map.py#L6-L7
> 
> I think it's safe to always use torchmetrics>0.12.0. @FANGAreNotGnu correct me if I'm wrong.

Just confirmed that torchmetrics>=0.11.3 works well. I can update the version requirement (also removing the backup codes) if it also works well with other problem type. @zhiqiangdon

zhiqiangdon:
> 

@FANGAreNotGnu Please go ahead cleaning up the old logic for object detection. Other problem types already use the newest torchmetrics.

FANGAreNotGnu:
> Some failed tests point to `utils/map.py`. The version detection may not have taken effect:
> 
> https://github.com/autogluon/autogluon/blob/7dfefbd5d9733f8dfbd6f77abd0ab6192ad1b932/multimodal/src/autogluon/multimodal/utils/map.py#L6-L7
> 
> I think it's safe to always use torchmetrics>0.12.0. @FANGAreNotGnu correct me if I'm wrong.

For torchmetrics >= 0.11.3 the map implementation is bug free. But there's still an issue that it's much slower than 0.6.0 (the implementation copied here). https://github.com/Lightning-AI/torchmetrics/issues/1024 There might be a problem that the old implementation is no more compatible with latest pytorch/torchmetrics. We can check the pytorch version and only use the new but slow implementation if we have to. @zhiqiangdon @sxjscience

multimodal/setup.py
```diff
@@ -33,13 +33,15 @@
...
     "fairscale>=0.4.5,<0.4.14",
-    "scikit-image>=0.19.1,<0.20.0",
-    "pytorch-lightning>=1.9.0,<1.10.0",
+    "scikit-image>=0.19.1,<0.21.0",
+    "pytorch-lightning>=1.9.0,<2.1",
     "text-unidecode>=1.3,<1.4",
-    "torchmetrics>=0.11.0,<0.12.0",
+    # temporary for testing ref https://github.com/autogluon/autogluon/issues/2687#issuecomment-1531122312
+    # "torchmetrics>=0.11.0,<0.12.0",
+    "torchmetrics @ https://github.com/Lightning-AI/torchmetrics/archive/95599c9a7bd5e6c8f0a8039562dbbc832b434031.zip",
```
ddelange, line 44: @FANGAreNotGnu fyi I've [asked](https://github.com/Lightning-AI/torchmetrics/pull/1625#issuecomment-1533615952) when this commit with pt 2.0 support will be released

ddelange:
> [2023-05-03T20:18:48.878Z] E           NotImplementedError: Support for `validation_epoch_end` has been removed in v2.0.0. `MMDetLitModule` implements this method. You can use the `on_validation_epoch_end` hook instead. To access outputs, save them in-memory as instance attributes. You can find migration examples in https://github.com/Lightning-AI/lightning/pull/16520.

ddelange:
> [2023-05-03T20:06:35.747Z] >           self.trainer.val_dataloaders = []  # skip the final validation by setting val_dataloaders empty
[2023-05-03T20:06:35.747Z] E           AttributeError: can't set attribute
[2023-05-03T20:06:35.747Z] src/autogluon/multimodal/optimization/lit_module.py:242: AttributeError

ddelange:
some questions regarding [`test_install.sh`](https://github.com/ddelange/autogluon/blob/6a0387e6ef8cfe838e27093831061909dd7b3985/.github/workflow_scripts/test_install.sh):

1. is `multimodal` [missing](https://github.com/ddelange/autogluon/blob/6a0387e6ef8cfe838e27093831061909dd7b3985/.github/workflow_scripts/env_setup.sh#L60-L70) on purpose?
2. why [does](https://github.com/ddelange/autogluon/blob/6a0387e6ef8cfe838e27093831061909dd7b3985/.github/workflow_scripts/test_install.sh#L9) it install mxnet still?
3. is it possible to test install on cp311 instead of cp38? e.g. [using](https://github.com/ddelange/pycuda/blob/654f713acc8c6dada689f257ca33f5b6895f1b15/buildtime/Dockerfile#L13) apt and the [deadsnakes repo](https://github.com/ddelange/pycuda/blob/654f713acc8c6dada689f257ca33f5b6895f1b15/base/Dockerfile#L36)

multimodal/setup.py
ddelange: `1.0.0rc0` is out

ddelange:
> > Some failed tests point to `utils/map.py`. The version detection may not have taken effect:
> > https://github.com/autogluon/autogluon/blob/7dfefbd5d9733f8dfbd6f77abd0ab6192ad1b932/multimodal/src/autogluon/multimodal/utils/map.py#L6-L7
> > 
> > I think it's safe to always use torchmetrics>0.12.0. @FANGAreNotGnu correct me if I'm wrong.
> 
> Just confirmed that torchmetrics>=0.11.3 works well. I can update the version requirement (also removing the backup codes) if it also works well with other problem type. @zhiqiangdon



> > 
> 
> @FANGAreNotGnu Please go ahead cleaning up the old logic for object detection. Other problem types already use the newest torchmetrics.

fyi I've bumped minimum torchmetrics to `1.0.0rc0` due to https://github.com/autogluon/autogluon/pull/3190/commits/6e626af6a2ce70b3255ea495b07ae12325a7e92d (it's also the first release that supports torch 2.0).

I think with that, the whole `map.py` can be deleted. Is there other cleanup necessary? You mentioned object detection?

ddelange:
```
src/autogluon/multimodal/optimization/lit_mmdet.py:241: in on_validation_epoch_end
    self.log_dict(mAPs, sync_dist=True)
../../opt/venv/lib/python3.8/site-packages/pytorch_lightning/core/module.py:570: in log_dict
    self.log(
../../opt/venv/lib/python3.8/site-packages/pytorch_lightning/core/module.py:444: in log
    value = apply_to_collection(value, (Tensor, numbers.Number), self.__to_tensor, name)
../../opt/venv/lib/python3.8/site-packages/lightning_utilities/core/apply_func.py:51: in apply_to_collection
    return function(data, *args, **kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = MMDetLitModule(
  (model): MMDetAutoModelForObjectDetection(
    (model): YOLOV3(
      (backbone): MobileNetV2(
     ...': 'Normal', 'std': 0.01, 'override': {'name': 'convs_pred'}}
    )
  )
  (validation_metric): MeanAveragePrecision()
)
value = tensor([0, 1, 2, 3, 4, 5, 6, 7, 8, 9], device='cuda:0', dtype=torch.int32)
name = 'val_classes'
    def __to_tensor(self, value: Union[Tensor, numbers.Number], name: str) -> Tensor:
        value = (
            value.clone().detach().to(self.device)
            if isinstance(value, Tensor)
            else torch.tensor(value, device=self.device)
        )
        if not torch.numel(value) == 1:
>           raise ValueError(
                f"`self.log({name}, {value})` was called, but the tensor must have a single element."
                f" You can try doing `self.log({name}, {value}.mean())`"
            )
E           ValueError: `self.log(val_classes, tensor([0, 1, 2, 3, 4, 5, 6, 7, 8, 9], device='cuda:0', dtype=torch.int32))` was called, but the tensor must have a single element. You can try doing `self.log(val_classes, tensor([0, 1, 2, 3, 4, 5, 6, 7, 8, 9], device='cuda:0', dtype=torch.int32).mean())`
../../opt/venv/lib/python3.8/site-packages/pytorch_lightning/core/module.py:617: ValueError
```

anyone ideas here? the [lightning function in question](https://github.com/Lightning-AI/lightning/blob/2.0.2/src/lightning/pytorch/core/module.py#L610-L622) looks unchanged since 1.9.0. can another dependency like torchmetrics cause this error?

timeseries/setup.py
ddelange: because of `pytorch-lightning>=2.0.0,<2.1` (https://github.com/autogluon/autogluon/pull/3190/commits/6a0387e6ef8cfe838e27093831061909dd7b3985), shall I consequently bump

```diff
- torch>=1.9,<2.1
+ torch>=1.11,<2.1
- torchvision>=0.10.0,<0.16
+ torchvision>=0.12.0,<0.16 
```
?

ddelange:
@gradientsky regarding the CLIP error you posted:

- in autogluon [`class CLIPForImageText(nn.Module)`](https://github.com/autogluon/autogluon/blob/v0.7.0/multimodal/src/autogluon/multimodal/models/clip.py#L26) has no `input_keys` attr
- in torch [`class Module`](https://github.com/pytorch/pytorch/blob/v1.13.1/torch/nn/modules/module.py#L249) has no `input_keys` attr
- as a matter of fact, there's only one file in torch 1.13.1 source containing `input_keys` (which seems irrelevant):
  ```console
  $ rg input_keys
  torch/csrc/jit/codegen/cuda/root_domain_map.cpp
  225:    input_keys_.insert(DomainKey(tv_->domain(), id));
  230:    return input_keys_;
  252:      if (input_keys_.find(out_key) == input_keys_.end()) {
  260:      input_keys_.insert(input_key);
  266:  DomainKeySet input_keys_;
  343:    const auto& input_keys = reduction_domain_inputs_.at(reduction_domain);
  351:                     input_keys.begin(), input_keys.end(), consumer_domain) !=
  352:              input_keys.end();
  ```

if this was missed by CI, maybe this error/breaking change was introduced a while ago already in an older torch version?

ddelange:
`test_*` CI jobs are all green 🎉

anything else from my side?

gradientsky:
> * in autogluon [`class CLIPForImageText(nn.Module)`](https://github.com/autogluon/autogluon/blob/v0.7.0/multimodal/src/autogluon/multimodal/models/clip.py#L26) has no `input_keys` attr
> * in torch [`class Module`](https://github.com/pytorch/pytorch/blob/v1.13.1/torch/nn/modules/module.py#L249) has no `input_keys` attr
> * as a matter of fact, ripgrep shows only one file in torch 1.13.1 source containing `input_keys` (which seems irrelevant):

I did a run w/ our 0.7 release which is bundled w/ torch 1.13.1 in our [containers](https://github.com/aws/deep-learning-containers/blob/master/available_images.md#autogluon-training-containers):
```
pytorch-lightning==1.9.4
pytorch-metric-learning==1.7.3
torch @ https://aws-pytorch-unified-cicd-binaries.s3.us-west-2.amazonaws.com/r1.13.1_sm/20230104-203046/4ad54e91be2ccb431cf26e7d7d1a2de10e8d82ac/torch-1.13.1%2Bcu117-cp39-cp39-linux_x86_64.whl
torchaudio @ https://download.pytorch.org/whl/cu117/torchaudio-0.13.1%2Bcu117-cp39-cp39-linux_x86_64.whl
torchdata @ https://download.pytorch.org/whl/test/torchdata-0.5.1-cp39-cp39-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
torchmetrics==0.8.2
torchvision @ https://download.pytorch.org/whl/cu117/torchvision-0.14.1%2Bcu117-cp39-cp39-linux_x86_64.whl
```

<details>
  <summary>Run Output</summary>

```
Downloading ./ag_automm_tutorial/file.zip from https://automl-mm-bench.s3.amazonaws.com/petfinder_kaggle.zip...
100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 2.00G/2.00G [00:51<00:00, 39.1MiB/s]
No path specified. Models will be saved in: "AutogluonModels/ag-20230505_164712/"
AutoMM starts to create your model. ✨

- Model will be saved to "/AutogluonModels/ag-20230505_164712".

- Validation metric is "accuracy".

- To track the learning progress, you can open a terminal and launch Tensorboard:
    # Assume you have installed tensorboard
    tensorboard --logdir /AutogluonModels/ag-20230505_164712

Enjoy your coffee, and let AutoMM do the job ☕☕☕ Learn more at https://auto.gluon.ai

Downloading (…)lve/main/config.json: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████| 4.19k/4.19k [00:00<00:00, 565kB/s]
Downloading pytorch_model.bin: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 605M/605M [00:01<00:00, 427MB/s]
Downloading (…)lve/main/config.json: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████| 665/665 [00:00<00:00, 284kB/s]
Downloading pytorch_model.bin: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 54.2M/54.2M [00:00<00:00, 404MB/s]
Downloading (…)okenizer_config.json: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████| 29.0/29.0 [00:00<00:00, 13.1kB/s]
Downloading (…)solve/main/vocab.txt: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████| 232k/232k [00:00<00:00, 68.1MB/s]
Downloading (…)/main/tokenizer.json: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████| 466k/466k [00:00<00:00, 82.2MB/s]
Downloading: "https://github.com/SwinTransformer/storage/releases/download/v1.0.0/swin_small_patch4_window7_224.pth" to /root/.cache/torch/hub/checkpoints/swin_small_patch4_window7_224.pth
Downloading (…)olve/main/vocab.json: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████| 862k/862k [00:00<00:00, 56.9MB/s]
Downloading (…)olve/main/merges.txt: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████| 525k/525k [00:00<00:00, 119MB/s]
Downloading (…)cial_tokens_map.json: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████| 389/389 [00:00<00:00, 206kB/s]
Downloading (…)okenizer_config.json: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████| 568/568 [00:00<00:00, 296kB/s]
Sanity Checking: 0it [00:00, ?it/s][2023-05-05 16:47:23.594: W smdistributed/modelparallel/torch/nn/predefined_hooks.py:78] Found unsupported HuggingFace version 4.26.1 for automated tensor parallelism. HuggingFace modules will not be automatically distributed. You can use smp.tp_register_with_module API to register desired modules for tensor parallelism, or directly instantiate an smp.nn.DistributedModule. Supported HuggingFace transformers versions for automated tensor parallelism: ['4.17.0', '4.20.1', '4.21.0']
[2023-05-05 16:47:23.620 58e53240b787:16 INFO utils.py:28] RULE_JOB_STOP_SIGNAL_FILENAME: None
[2023-05-05 16:47:23.640 58e53240b787:16 INFO profiler_config_parser.py:111] Unable to find config at /opt/ml/input/config/profilerconfig.json. Profiler is disabled.
Epoch 0:  68%|██████████████████████████████████████████████████████████████████████████████████                                      | 52/76 [01:23<00:38,  1.60s/it, loss=2.96, v_num=]
Start to fuse 2 checkpoints via the greedy soup algorithm.
Predicting DataLoader 0: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 4/4 [00:00<00:00,  4.12it/s]
Predicting DataLoader 0: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 4/4 [00:00<00:00,  4.04it/s]
AutoMM has created your model 🎉🎉🎉

- To load the model, use the code below:
    predictor = MultiModalPredictor.load("/AutogluonModels/ag-20230505_164712")

- You can open a terminal and launch Tensorboard to visualize the training log:
    # Assume you have installed tensorboard
    tensorboard --logdir /AutogluonModels/ag-20230505_164712

- If you are not satisfied with the model, try to increase the training time,
adjust the hyperparameters (https://auto.gluon.ai/stable/tutorials/multimodal/advanced_topics/customization.html),
or post issues on GitHub: https://github.com/autogluon/autogluon


Predicting DataLoader 0: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 4/4 [00:00<00:00,  4.23it/s]
{'accuracy': 0.15}
```
</details>

ddelange:
@gradientsky interesting. could you also run with current autogluon master? so we can rule out https://github.com/autogluon/autogluon/pull/2836 as the culprit?

Innixma:
@gradientsky @zhiqiangdon Can we confirm the "Check HF Model List" test and add the tag if it is satisfied? This is the last failing CI check

gradientsky:
@ddelange I can confirm the CLIP issue is not related to this PR: I created a separate issue https://github.com/autogluon/autogluon/issues/3203

multimodal/setup.py
```diff
@@ -33,13 +33,13 @@
     "evaluate>=0.2.2,<0.4.0",
     "accelerate>=0.9,<0.17",
     "timm>=0.6.12,<0.7.0",
-    "torch>=1.9,<1.14",
-    "torchvision<0.15.0",
+    "torch>=1.11,<2.1",
```
zhiqiangdon, line 36: This change makes user install torch 2.0 by default. Is the CI still testing torch 1.13.1? If so, there is some inconsistency between what we test and what users use.

ddelange: does this thread answer your question? https://github.com/autogluon/autogluon/pull/3190#discussion_r1183987923

after this merges I think there are no explicit tests that run with pytorch 1.x anymore

ddelange:
hmm [looks](https://github.com/autogluon/autogluon/actions/runs/4896586916/jobs/8743608803?pr=3190#step:4:1271) like an `assert_allclose` failed. could someone re-run?

multimodal/src/autogluon/multimodal/utils/object_detection.py
```diff
@@ -807,8 +807,7 @@ def cocoeval_torchmetrics(outputs: List):
     """
 
     import torch
-
-    from . import MeanAveragePrecision
+    from torchmetrics.detection.mean_ap import MeanAveragePrecision
```
FANGAreNotGnu, line 810: We can do this for now for quickly moving to torch 2.0. And later figure out how to speed up the map in torch 2.0, as current torchmetrics version is very slow in map calculation.

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -1475,7 +1475,6 @@ def _fit(
                 log_every_n_steps=OmegaConf.select(config, "optimization.log_every_n_steps", default=10),
                 enable_progress_bar=enable_progress_bar,
                 fast_dev_run=config.env.fast_dev_run,
-                track_grad_norm=OmegaConf.select(config, "optimization.track_grad_norm", default=-1),
```
zhiqiangdon, line 1478: Please update the doc correspondingly: https://github.com/autogluon/autogluon/blob/master/docs/tutorials/multimodal/advanced_topics/customization.ipynb?short_path=c0bd10d#L368

zhiqiangdon: BTW, as lightning has removed `track_grad_norm ` from trainer, do you know if lightning has some equivalent design for it?

multimodal/src/autogluon/multimodal/optimization/lit_module.py
```diff
@@ -239,13 +239,13 @@ def training_step(self, batch, batch_idx):
...
     def on_validation_start(self) -> None:
         if self.skip_final_val and self.trainer.should_stop:
-            self.trainer.val_dataloaders = []  # skip the final validation by setting val_dataloaders empty
             self.log(
                 self.validation_metric_name,
                 self.validation_metric,
                 on_step=False,
                 on_epoch=True,
             )
+            return None
```
zhiqiangdon, line 248: @suzhoum Can you verify the fast build mode still behaves well after this change?

zhiqiangdon (commented):
@ddelange Many thanks for the amazing work in bumping up the dependency versions. As this upgrade involves several key packages, we are trying to do some internal benchmarking on this PR to make sure everything is ok after this upgrade. We will keep you updated.

multimodal/src/autogluon/multimodal/predictor.py
ddelange: ref https://github.com/Lightning-AI/lightning/pull/16745 they suggest an explicit `log_dict`. any thoughts on this?

ddelange: > Please update the doc correspondingly: https://github.com/autogluon/autogluon/blob/master/docs/tutorials/multimodal/advanced_topics/customization.ipynb?short_path=c0bd10d#L368 We can remove `track_grad_norm ` from the doc.

https://github.com/autogluon/autogluon/pull/3190/commits/88d06643d520f7daec300b002429cad4485a860b

multimodal/src/autogluon/multimodal/optimization/lit_module.py
ddelange: more context in the commit message: https://github.com/autogluon/autogluon/pull/3190/commits/18fbb5f095fb39df1d1cdae07d16dccf4615cca8

multimodal/src/autogluon/multimodal/utils/object_detection.py
ddelange: fyi there's an open performance PR worth watching: https://github.com/Lightning-AI/torchmetrics/pull/1389

ddelange:
> there's quite a chance that pt 2.0 will 'just work'. there's now `zero_grad(set_to_none=True)` (default used to be False), for the rest the changes [look](https://github.com/pytorch/pytorch/releases/tag/v2.0.0) all good to me

there are a couple of `zero_grad` occurrences in autogluon. despite green CI, shall I add `set_to_none=False` everywhere just in case?

yinweisu:
Cloned this PR and discovered couple issues:

1. couple multimodal tests failing
2. ray and onnxruntime-gpu doesn't support 3.11 yet.

Ray support would be the major blocker. Will give it an evaluation today

ddelange:
> Cloned this PR and discovered couple issues:
> 
> 1. couple multimodal tests failing
> 2. ray and onnxruntime-gpu doesn't support 3.11 yet.
> 
> Ray support would be the major blocker. Will give it an evaluation today

hmm, interesting. ray [ships](https://pypi.org/project/ray/2.3.0/#files) cp311 wheels as of 2.3.0. is it due to a downstream dependency?

yinweisu:
Yea. I just created a virv on py311 and I was able to install ray on linux machine. Though it seems doesn't support macos and windows: https://docs.ray.io/en/latest/ray-overview/installation.html. According to our unit tests, it seems available on windows but not on macos

Having a deeper look, these are the more detailed issue in the platform tests:
1. core installation on macos 3.11 failed with ray installation cause it's not available: https://github.com/autogluon/autogluon/actions/runs/4918057906/jobs/8783917156. This happens to other module that depends on core too, i.e. eda, tabular
2. eda and tabular tests fail on some weird catboost installation error for all version of python: https://github.com/autogluon/autogluon/actions/runs/4918057906/jobs/8783922246. The most recent nightly platform tests don't have this problem
3. py311 support is marked as experimental. I saw a failure happening at the import step of ray: https://github.com/autogluon/autogluon/actions/runs/4918057906/jobs/8783917961

Innixma:
`onnxruntime-gpu` is probably not a big deal, I think it is an optional dependency to my understanding, but `ray` is a major dependency that is very important.

yinweisu:
Updates: Created a py311 env with latest ray(2.4.0). The `from ray import tune` fails

yinweisu:
Hi @ddelange, after some discussion with the team, we decide to not support python 3.11 as the benefits don't outweigh the required effort / risk. Will keep this PR open and revisit later. Thanks a lot for your work!

ddelange:
Makes sense 👍 I'll be around!

ddelange:
I think now the only blocker left is https://github.com/ray-project/ray/issues/31362

Innixma:
@ddelange I saw you responded to some of the CatBoost GitHub issues regarding MacOS support, have the MacOS CatBoost issues been resolved?

ddelange:
update: it [looks](https://github.com/ray-project/ray/issues/31362#issuecomment-1583812031) like ray will start shipping all 3.11 wheels as of the upcoming 2.6.0 release

multimodal/src/autogluon/multimodal/utils/object_detection.py
ddelange: ^ will probably be superceded, see also the discussion in the linked issue

https://github.com/Lightning-AI/torchmetrics/pull/1832

Innixma:
@ddelange Great to hear! This will definitely be on the top of our priorities for the v0.9 release, and seems like it could be feasible within that timeline.

ddelange:
another 3.11 blocker that'll hopefully be fixed in ray 2.6: https://github.com/ray-project/ray/issues/36319

ddelange:
interesting:

- tensorrt install [errors](https://github.com/autogluon/autogluon/actions/runs/5309652350/jobs/9610616186?pr=3190#step:4:1006) with `/home/ci/opt/venv/bin/python3: No module named pip`
  - command `['/home/ci/opt/venv/bin/python3', '-m', 'pip', 'install', 'tensorrt_libs==8.6.1', '--index-url', 'https://pypi.nvidia.com/']`
- yet autogluon ci [runs](https://github.com/autogluon/autogluon/actions/runs/5309652350/jobs/9610616186?pr=3190#step:4:154) `python3 -m pip install --upgrade pip`
  - which upgrades `/home/ci/opt/venv/lib/python3.8/site-packages`

ddelange:
> another 3.11 blocker that'll hopefully be fixed in ray 2.6: [ray-project/ray#36319](https://github.com/ray-project/ray/issues/36319)

I'm now successfully running training on GPU using this branch with ray 2.4 and cp311 🎉

Innixma:
@ddelange thats awesome! Are there any blockers remaining or should we try running the CI with 3.11?

ddelange:
To run CI on linux cp311, the only blocker I still see is the `/home/ci/opt/venv/bin/python3: No module named pip` error mentioned above.

- tensorrt sdist install [runs `pip install` inside `pip install`](https://github.com/NVIDIA/TensorRT/blob/v8.6.1/python/packaging/frontend_sdist/setup.py#L32), which is undefined/unsupported behaviour from the PyPA side (just like running multiple concurrent pip commands).
- The weird thing is that I can't reproduce the installation issue on my local machine, where I set up my venv basically the same way autogluon CI does (`VIRTUAL_ENV` env var is set and `VIRTUAL_ENV/bin` is prepended to `PATH` env var).
- The way they invoke pip (linked above) is correct: based on `sys.executable`. Very weird to see that one fail with `No module named pip` because `pip` is clearly correctly installed in the venv.

Innixma:
Great! I'll wait and see if TensorRT will respond to your PR/Issue.

ddelange:
So that PR hit a dead end pretty quickly... I guess it's back to inspecting autogluon's virtual environment setup, I really have no clue why pip can't be found by the subprocess that's clearly running from the right venv.

I've tested a bare `pip install tensorrt` in docker [using](https://github.com/ddelange/pycuda/blob/0e534f60f96ea8a02c2ad7a9c5b674377e92d1f2/base/Dockerfile#L34-L38) deadsnakes, a simple [ENV](https://github.com/ddelange/pycuda/blob/0e534f60f96ea8a02c2ad7a9c5b674377e92d1f2/base/Dockerfile#L48-L50), and then the usual venv [creation](https://github.com/ddelange/pycuda/blob/0e534f60f96ea8a02c2ad7a9c5b674377e92d1f2/vanilla/Dockerfile#L8-L9).

Can you maybe spot a meaningful difference from how your AWS Batch is preparing the venv?

Innixma:
@gradientsky Could you maybe take a look at the above comment from @ddelange?

ddelange:
hi all :wave:

another commit went in, the ray team deprecated `ray_lightning` (https://github.com/ray-project/ray/issues/35893) and added support for `pytorch_lightning==2.*` in ray 2.5.

could someone have a look at the `ray_lightning` leftovers in the codebase? feel free to commit to this branch!

ddelange:
On the side:

Running this branch on cp311 with 1 GPU (`g4dn.2xlarge`), I made a quick [performance profile](https://github.com/autogluon/autogluon/assets/14880945/2b811434-e8e1-409f-83a6-4993f5503105) using py-spy. It's an interactive svg, the dev is a wizard :)

The vast majority of time is spent inside [this line](https://github.com/autogluon/autogluon/blob/973289cd14b572dbf5175d52133ddd27c0808836/multimodal/src/autogluon/multimodal/models/huggingface_text.py#L173) (48%) and [this line](https://github.com/autogluon/autogluon/blob/973289cd14b572dbf5175d52133ddd27c0808836/multimodal/src/autogluon/multimodal/models/utils.py#L383) (37%). That's with python (single-core) at 100%, and the GPU running between 95% and 100%. So there's still a slight bottleneck on the CPU side, but in comparison to cp38 (GPU at ~90%) it's already almost gone! 🎉

Innixma:
> On the side:
> 
> Running this branch on cp311 with 1 GPU (`g4dn.2xlarge`), I made a quick [performance profile](https://github.com/autogluon/autogluon/assets/14880945/2b811434-e8e1-409f-83a6-4993f5503105) using py-spy. It's an interactive svg, the dev is a wizard :)
> 
> The vast majority of time is spent inside [this line](https://github.com/autogluon/autogluon/blob/973289cd14b572dbf5175d52133ddd27c0808836/multimodal/src/autogluon/multimodal/models/huggingface_text.py#L173) (48%) and [this line](https://github.com/autogluon/autogluon/blob/973289cd14b572dbf5175d52133ddd27c0808836/multimodal/src/autogluon/multimodal/models/utils.py#L383) (37%). That's with python (single-core) at 100%, and the GPU running between 95% and 100%. So there's still a slight bottleneck on the CPU side, but in comparison to cp38 (GPU at ~90%) it's already almost gone! 🎉

That is great to hear! 95-100% utilization is promising, and identifying the bottlenecks is good for us to keep in mind.
Regarding follow-ups to this PR and your question on pytorch_lightning, our team is just returning from holidays and will be diving into this in more detail either later this week or early next week.

yinweisu:
@ddelange Thanks again for the contribution!

Our aws batch is only testing python 3.9, but unit tests should pass regardless. I can try pull the image used in our CI and directly install tensorrt.

Regarding ray-lightning, yes we should clean it up. I'll take a look soon.

ddelange:
super! this PR is probably the best base for @zhiqiangdon - upgrading had a couple of cascading effects on the other dependencies. feel free to take it over!

yinweisu:
I've created this [PR](https://github.com/autogluon/autogluon/pull/3398) to remove ray lightning.

As of the tensorrt issue, I found it will be fixed if we don't upgrade pip for some reason... Will need to discuss with the team on what we should do

yinweisu:
@ddelange is the tensorrt upgrade required to support 3.11? Want to get some context here

ddelange:
> @ddelange is the tensorrt upgrade required to support 3.11? Want to get some context here

yes, 8.6.0 [has](https://pypi.org/project/tensorrt/8.6.0/#files) only wheels up to cp310

yinweisu:
OK. We could probably just not updating the pip package for now. We are in the process of releasing 0.8 container, which would be the new base image for our CI once published. I did a test on tensorrt installation with the new container, and it worked there. I honestly don't really know why our current container doesn't work.

So i'll try resolving the conflicts you have in this PR and clone it to the other testing PR and retrigger tests.

ddelange:
ray 2.6 was just released with all the cp311 wheels available 🎉

.github/workflow_scripts/env_setup.sh
```diff
@@ -17,21 +17,23 @@ function setup_build_contrib_env {
...
+    setup_torch_gpu
 }
 
 function setup_torch_cpu_non_linux {
-    pip3 install torch==1.13.1 torchvision==0.14.1
+    setup_torch_cpu
+}
+
+function reinstall_torch {
+    pip3 install --force-reinstall torchvision~=0.15.1
```
ddelange, line 36: why `pip3` btw? might it be that the autogluon python setup is not running a virtual environment, in which symlinked `pip` will be available? see also my Dockerfile snippet from the earlier comment

ddelange:
[test_eda](https://github.com/autogluon/autogluon/actions/runs/5637372148/job/15270451763?pr=3190#step:4:1341)
```
[2023-07-23T16:15:05.418Z] E   ValueError: node array from the pickle has an incompatible dtype:
[2023-07-23T16:15:05.418Z] E   - expected: [('left_child', '<i8'), ('right_child', '<i8'), ('feature', '<i8'), ('threshold', '<f8'), ('impurity', '<f8'), ('n_node_samples', '<i8'), ('weighted_n_node_samples', '<f8')]
[2023-07-23T16:15:05.418Z] E   - got     : {'names': ['left_child', 'right_child', 'feature', 'threshold', 'impurity', 'n_node_samples', 'weighted_n_node_samples', 'missing_go_to_left'], 'formats': ['<i8', '<i8', '<i8', '<f8', '<f8', '<i8', '<f8', 'u1'], 'offsets': [0, 8, 16, 24, 32, 40, 48, 56], 'itemsize': 64}
```

[test_tabular](https://github.com/autogluon/autogluon/actions/runs/5637372148/job/15270451833?pr=3190#step:4:1597)
```
[2023-07-23T16:49:54.059Z]         predictor = TabularPredictor(label="class", log_to_file=True).fit(train_data=train_data, hyperparameters={"DUMMY": {}})
[2023-07-23T16:49:54.059Z]         log = TabularPredictor.load_log(predictor_path=predictor.path)
[2023-07-23T16:49:54.059Z] >       assert "TabularPredictor saved." in log[-1]
[2023-07-23T16:49:54.059Z] E       IndexError: list index out of range
```

[test_multimodal](https://github.com/autogluon/autogluon/actions/runs/5637372148/job/15270452009?pr=3190#step:4:1154) still has the tensorrt error. making `pip` binary available in autogluon venv should fix that

.github/workflow_scripts/env_setup.sh
Innixma: I think `pip3` is just because of legacy where we thought `pip3` would be correct more often than `pip` (and in most cases it is identical). However, now it is probably the case that `pip3` and `pip` are almost always the same, so it shouldn't matter if it is changed to `pip`

yinweisu:
Hmm, weird that it fails the tabular log test. Will try to reproduce it locally. The EDA failures should be unrelated. There are also some onnx export tests failures, which I might need to sync with other members.

Just FYI, we are working on a PR to upgrade torch to 2.0 right now

ddelange:
pulled master 👍

ddelange:
@yinweisu fyi this PR bumps torchmetrics to 1.0.0, in order to support (allow) torch 2.0 ref https://github.com/autogluon/autogluon/issues/2687#issuecomment-1531122312

yinweisu:
@ddelange I'm going to work on upgrading our CI base image now. Hope it fixes this issue. Regarding torchmetrics, I believe @zhiqiangdon's PR regarding torch and pytorch lightning upgrades already bumps that. I saw tochmetrics 1.0.2 being installed in our CI env

ddelange:
you're right, the torchmetrics bump is gone from this PR with a recent pull from master :)

yinweisu:
@ddelange I've upgraded our image and triggered a rebuild of the CI. Let's see what does the results look like. Regarding the lower bound, it should be fine. If the user is in python 311 environment, and pypi should auto resolve the dependency to install compatible versions

common/src/autogluon/common/utils/try_import.py
```diff
@@ -124,7 +124,7 @@ def try_import_torch():
         import torch
     except ImportError as e:
         raise ImportError(
-            "Unable to import dependency torch\n" "A quick tip is to install via `pip install torch`.\n" "The minimum torch version is currently 1.6."
+            "Unable to import dependency torch\n" "A quick tip is to install via `pip install torch`.\n" "The minimum torch version is currently 1.11."
```
ddelange, line 127:
```suggestion
            "Unable to import dependency torch\n" "A quick tip is to install via `pip install torch`.\n" "The minimum torch version is currently 1.13."  # sync with core/_setup_utils.py
```

yinweisu:
@ddelange Thanks! Just retriggered

ddelange:
in [test_tabular](https://github.com/autogluon/autogluon/actions/runs/5968741992/job/16198100898?pr=3190#step:4:2551):
```
assert isinstance(predictor._learner.trainer.models["NeuralNetTorch"].processor, TabularNeuralNetTorchOnnxTransformer)
where False = isinstance(ColumnTransformer(...
```

in [test_multimodal](https://github.com/autogluon/autogluon/actions/runs/5968741992/job/16198100397?pr=3190#step:4:1492):
```
./.github/workflow_scripts/test_multimodal.sh: line 1:   910 Segmentation fault      (core dumped) python3 -m pytest --junitxml=results.xml --runslow tests/unittests/"$SUB_FOLDER"/
tests/unittests/others/test_deployment_onnx.py ....
```

ddelange:
Hi all 👋

I'm not sure what these last two failing tests could be related to. Some changes in new versions upstream?

Who can help out here?

Innixma:
An update on this: We are heading towards v1.0 release in November. Currently we are trying to finalize a variety of science innovation work that will be added in v1.0, and in early October we will start focusing on various technical clean up and preparation, including the merging of this PR. Lots to do, sorry if we haven't responded promptly as of late 😅. @ddelange greatly appreciate your support, it means a lot to us!

ddelange:
^ ray 2.7 contains some critical fixes for ray serve specific to python 3.11 ref https://github.com/kserve/kserve/pull/3075#issuecomment-1677630840

Innixma:
@ddelange and @yinweisu, awesome work! Everything appears to be passing now on the clone PR (#3204). We will move the changes over to this PR and merge soon. This is a huge feature, thanks so much!

Innixma (changes requested):
Awaiting changes from #3204 to be merged into this PR by @yinweisu.

yinweisu:
@Innixma Changes merged. I think the code is good to merge functionality wise; though there are some installation instruction changes. Currently, we are prompting for pytorch 1.13 installation, I wonder shall we increase it to pytorch 2.0 and its corresponding torchvision?

We can probably address this in a follow-up PR too given this PR has been around for a long time

docs/install-cpu-pip.md
```diff
@@ -4,7 +4,7 @@ pip install -U setuptools wheel
 
 # CPU version of pytorch has smaller footprint - see installation instructions in
 # pytorch documentation - https://pytorch.org/get-started/locally/
-pip install torch==1.13.1+cpu torchvision==0.14.1+cpu -f https://download.pytorch.org/whl/cpu/torch_stable.html
+pip install torchvision~=0.15.1 --force-reinstall --extra-index-url https://download.pytorch.org/whl/cpu
```
ddelange, line 7: I switched the docs to torch's new `--extra-index-url` notation.

torchvision etc have an exact pin on a torch version, so this command is sufficient (and will reinstall cuda version if a user previously had a cpu version installed amd vice versa)

multimodal/src/autogluon/multimodal/configs/optimization/adamw.yaml
```diff
@@ -16,7 +16,6 @@ optimization:
   skip_final_val: False # Flag to skip the last validation
   gradient_clip_val: 1
   gradient_clip_algorithm: "norm"
-  track_grad_norm: -1 # Whether to check gradient norm. We can set it to 2 to check for gradient norm.
```
zhiqiangdon, line 19: We still support using `track_grad_norm `, so no need to remove it. See https://github.com/autogluon/autogluon/blob/master/multimodal/src/autogluon/multimodal/optimization/lit_module.py#L401-L404

docs/tutorials/multimodal/advanced_topics/customization.ipynb
```diff
@@ -360,29 +360,6 @@
...
-   "metadata": {},
-   "source": [
-    "```\n",
-    "# default used by AutoMM (no tracking)\n",
-    "predictor.fit(hyperparameters={\"optimization.track_grad_norm\": -1})\n",
-    "# track the 2-norm\n",
-    "predictor.fit(hyperparameters={\"optimization.track_grad_norm\": 2})\n",
-    "```\n"
-   ]
-  },
```
zhiqiangdon, line 385: We still support using track_grad_norm , so no need to remove it. See https://github.com/autogluon/autogluon/blob/master/multimodal/src/autogluon/multimodal/optimization/lit_module.py#L401-L404

ddelange: I remember we had some conversation on it in https://github.com/autogluon/autogluon/pull/3190/commits/14e6fee69956d833ad41675ce7cf44bb1b198939 - I'm on holiday, maybe @yinweisu you could take over?

zhiqiangdon: Happy holiday! Lightning's trainer has removed the `track_grad_norm` argument. We are using a suggested workaround of doing that.

yinweisu:
@zhiqiangdon Addressed your comments

docs/tutorials/multimodal/advanced_topics/customization.ipynb
yinweisu: yap. @ddelange Happy holiday and thanks for your contribution!

Innixma (approved):
LGTM!!

zhiqiangdon (approved):
LGTM! Thanks for the contribution!

Innixma:
Merged! Thanks everyone for working to push this through!

multimodal/setup.py
```diff
@@ -64,7 +64,7 @@
     "onnx>=1.13.0,<1.14.0",
     "onnxruntime>=1.15.0,<1.16.0;platform_system=='Darwin'",
     "onnxruntime-gpu>=1.15.0,<1.16.0;platform_system!='Darwin'",
-    "tensorrt>=8.5.3.1,<8.5.4;platform_system=='Linux'",
+    "tensorrt>=8.5.3.1,<8.5.4;platform_system=='Linux' and python_version<'3.11'",  # tensorrt > 8.5.4 cause segfault
```
ddelange, line 67: @yinweisu could you give tensorrt 8.6.1 another shot? to get some more info about the segfault?

if i find some time, I'll try to install it in our py 3.11 staging cluster and try out `optimize_for_inference`

AnirudhDagar: Hi @ddelange, I've been trying to work out that problem with tensorrt 8.6.1 running the tests in `test_deployment_onnx.py`. It still gives me a segfault, I'll update with more info.

AnirudhDagar: Update: I ran some more tests in `python3.11`, and `tensorrt==8.6.1` worked with `cuda12.1` with `autogluon` built from source (current master) and `torch==2.1.1`. I was able to successfully run tests in `test_deployment_onnx.py` with the above versions.

All in all, we can wait for the cuda version to be upgraded inside autogluon requirements to cuda12.1 (https://github.com/autogluon/autogluon/pull/3663), which will then solve the segfault issue on the latest version of `tensorrt` and `cuda11.8`. Only then we can go ahead and update `tensorrt` version in setup.py

ddelange: Hi @AnirudhDagar :wave:

#3982 was merged. could you finalize support for `optimize_for_inference` on python 3.11 to finally close #2687?

AnirudhDagar: Yes, I'll update the issues as soon as we get rid of the py3.11 conditional which is still present there due to a few failing HPO tests ([see comment](https://github.com/autogluon/autogluon/pull/3982/files#r1526574710)). I need to run some tests locally, specifically for `optimize_for_inference`.

ddelange: @AnirudhDagar I saw you added "close https://github.com/autogluon/autogluon/issues/2687" to the release tracker https://github.com/autogluon/autogluon/issues/4027

but that issue tracks `optimize_for_inference` support on 3.11. is your plan to squeeze it in still?

AnirudhDagar: Thanks for the reminder @ddelange, yes I can give the `optimize_for_inference ` testing a shot after merging your ray PR.

ddelange: fwiw in a future version, the `tensorrt` python package seems deprecated (from inspecting the latest prerelease it's now an empty meta package forwarding to https://pypi.org/project/tensorrt-cu12/10.0.0b6/).

So if you hit a snatch, it might be worth to give `tensorrt-cu12` a try. See also the install instructions on the pypi page (to avoid an ugly pip-in-pip pattern ref https://github.com/NVIDIA/TensorRT/pull/3080)

PR #3192, opened 2023-05-03
Move ag space to common

yinweisu:
*Issue #, if available:*

*Description of changes:*
* Move ag space to common so it could be consumed by cloud module
* Also making the way to use search space in our source code more consistent now. Now all usage would be `from autogluon.common.space import XXX`


common/src/autogluon/common/__init__.py
```diff
@@ -1,6 +1,7 @@
 from .version import __version__
 
 from .utils.log_utils import _add_stream_handler, fix_logging_if_kaggle as __fix_logging_if_kaggle
+from .space import Space, Categorical, Real, Int, Bool
```
shchur, line 4: Would it be possible to avoid a breaking change by keeping the spaces available in `autogluon.core` for 1 more release with a deprecation warning?

For example, using something like the following code in `autogluon/core/__init__.py`
```python
import warnings

class DeprecatedSpacesWrapper:
    def __getattr__(self, attr: str) -> Any:
        import autogluon.common as ag

        if attr in ["Categorical", "Real", "Int", "Bool"]:
            warnings.warn(
                "Accessing search spaces as `autogluon.core.space` is deprecated as of v0.8 and won't be supported "
                "in the next release. Please use `autogluon.common.space` instead."
            )
            return getattr(ag.space, attr)
        else:
            raise AttributeError

space = DeprecatedSpacesWrapper()
```

timeseries/src/autogluon/timeseries/models/presets.py
```diff
@@ -3,7 +3,7 @@
 import re
 from typing import Any, Dict, List, Optional, Type, Union
 
-import autogluon.core as ag
+import autogluon.common as ag
 import autogluon.timeseries as agts
```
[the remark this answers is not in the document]
yinweisu: Sure. I think both would work, but I'll update it to keep it in sync with other usages

common/src/autogluon/common/__init__.py
yinweisu: Thanks for bringing it up. I think this is a good idea

Innixma: +1, this is nice

yinweisu: I moved the warning to the init of any `Space` object and imported `Space` related object from common in core instead.

The method @shchur provided wouldn't work if user try to do things like `from autogluon.core import Real`, which is possible currently. After the change, the above usage would also give proper warning message

shchur: Would this result in a warning always being produced when a `Space` is used (e.g., as a part of the presets), even if the user didn't use the deprecated API?

yinweisu: Oh yea... This is tricky. Need to think about how to cover all cases

yinweisu: OK. Added this inside `core.space.py`, and this plus the wrapper @shchur provided should cover all cases and not affect import from common

```python
# TODO: Remove this file for v1.0 as space is moved to common
import warnings
import sys

from typing import Any


this_module = sys.modules[__name__]
spaces = ["Space", "Categorical", "Real", "Int", "Bool"]


class DeprecatedSpaceWrapper:
    def __init__(self, common_space) -> None:
        self.comomn_space = common_space
    
    def __call__(self, *args, **kwargs) -> Any:
        import autogluon.common as ag

        warnings.warn(
            "Accessing search spaces through `autogluon.core` is deprecated as of v0.8 and won't be supported "
            f"in the next release. Please use `from autogluon.common import {self.comomn_space}` instead."
        )
        return getattr(ag.space, self.comomn_space)(*args, **kwargs)


for s in spaces:
    setattr(this_module, s, DeprecatedSpaceWrapper(s))

```

core/src/autogluon/core/__init__.py
```diff
@@ -1,11 +1,29 @@
...
+# TODO: Remove deprecation warning for v1.0
+class DeprecatedSpacesWrapper:
+    def __getattr__(self, attr: str) -> Any:
+        import autogluon.common as ag
+
+        if attr in spaces:
+            warnings.warn(
+                "Accessing search spaces as `autogluon.core.space` is deprecated as of v0.8 and won't be supported "
+                "in the next release. Please use `autogluon.common.space` instead."
+            )
```
Innixma, line 22: Unsure best practice, but I wonder if we should have a class/function specifically that checks `__version__` and errors if `__version__ >= 0.9`, else warn, as some kind of standardized logic across the repo. Currently we track the deprecations in a bunch of different ways, and can miss removing code / logic that was previously deprecated.

Innixma: This wouldn't be something added in this PR, but something we may want to think about (@yinweisu @tonyhoo @gradientsky )

core/src/autogluon/core/hpo/executors.py
```diff
@@ -7,14 +7,15 @@
 import pandas as pd
 import time
 
+import autogluon.common as ag
```
Innixma, line 10: I wonder if this is primarily a bad practice we are continuing from early days of repo. This `import ... as ag` logic was back when we would do `import autogluon as ag` before we were a namespace package.

For v1.0, the names and structure of `space.py` could maybe use a refactor.

Innixma: also, why aren't we doing `from autogluon.common import Space`?

core/src/autogluon/core/hpo/ray_hpo.py
```diff
@@ -2,6 +2,8 @@
 import os
 import shutil
 
+import autogluon.common as ag
```
Innixma, line 5: why not `from autogluon.common import Space`?

core/src/autogluon/core/hpo/space_converter.py
```diff
@@ -1,8 +1,8 @@
+import autogluon.common as ag
```
Innixma, line 1: why not `from autogluon.common import ...`?

core/src/autogluon/core/searcher/local_grid_searcher.py
```diff
@@ -4,8 +4,10 @@
 import numpy as np
 from sklearn.model_selection import ParameterGrid
 
+import autogluon.common as ag
```
Innixma, line 7: why not `from autogluon.common import Space`? (Ditto for all other cases, there are many.

core/src/autogluon/core/__init__.py
yinweisu: +1. I agree we should think about a general way to handle deprecation

shchur: We introduced the [`deprecated` decorator](https://github.com/autogluon/autogluon/blob/master/common/src/autogluon/common/utils/deprecated.py) for methods, but I came across a few cases where it is not applicable (e.g., when an argument to a method is renamed).

A simple approach that would cover these edge cases is to always check all the occurrences of the string `deprecate` within the codebase as a part of the release checklist. Though, this is definitely not elegant or scalable.

timeseries/tests/unittests/test_predictor.py
```diff
@@ -6,7 +6,7 @@
 import pandas as pd
 import pytest
 
-import autogluon.core as ag
+from autogluon.common.space import Bool, Categorical, Int, Real, Space
```
shchur, line 9: I think the following code would still work and require fewer changes.
```python
import autogluon.common as ag

...
ag.space.Int(1, 3)
....
```

I personally also like `ag.space.Categorical` more than `Categorical` since this way it's always clear that this is a search space. This also would help to avoid potential clashes, e.g., with `torch.distributions.Categorical`.

core/src/autogluon/core/__init__.py
```diff
@@ -1,11 +1,29 @@
...
+# TODO: Remove deprecation warning for v1.0
+class DeprecatedSpacesWrapper:
+    def __getattr__(self, attr: str) -> Any:
+        from autogluon.common import space
+
+        if attr in spaces:
+            warnings.warn(
+                "Accessing search spaces as `autogluon.core.space` is deprecated as of v0.8 and won't be supported "
+                "in the next release. Please use `autogluon.common.space` instead."
+            )
```
shchur, line 22:
```suggestion
                "in the next release. Please use `autogluon.common.space` instead.",
                category=DeprecationWarning,
            )
```

shchur (approved):
LGTM on the `timeseries` side, a few minor suggestions only.

timeseries/tests/unittests/test_predictor.py
yinweisu: What you describe makes sense. We are trying to unify the usage of the search space in our code base and we think it's not a good practice to import the whole common module while all we are using are search spaces. Maybe a better alternative would be `from autogluon.common import space` then `space.XXX`. And when there are potential other search namespace, we can do `from autogluon.common import space as ag_space`

shchur: I agree, this totally makes sense

common/src/autogluon/common/__init__.py
Innixma, line 4: Shall we remove these imports here since we want users to do `from autogluon.common.space import ...`?

Innixma (approved):
LGTM! Had 1 comment

PR #3204, opened 2023-05-05
Clone of PR 3190 for testing

yinweisu:
*Issue #, if available:*

*Description of changes:*


yinweisu:
Couple issues discovered so far:
1. couple multimodal tests failing
2. ray and onnxruntime-gpu doesn't support 3.11 yet.

yinweisu:
@ddelange Made some edits based on your PR and all tests are green now. We should be good to go @Innixma

multimodal/setup.py
```diff
@@ -64,7 +64,7 @@
     "onnx>=1.13.0,<1.14.0",
     "onnxruntime>=1.15.0,<1.16.0;platform_system=='Darwin'",
     "onnxruntime-gpu>=1.15.0,<1.16.0;platform_system!='Darwin'",
-    "tensorrt>=8.5.3.1,<8.5.4;platform_system=='Linux'",
+    "tensorrt>=8.5.3.1,<8.5.4;platform_system=='Linux' and python_version<'3.11'",  # tensorrt > 8.5.4 cause segfault
```
ddelange, line 67: is there anything we cab report upstream?

tabular/setup.py
```diff
@@ -38,8 +38,9 @@
     "catboost": [
         # CatBoost wheel build is not working correctly on darwin for CatBoost 1.2, so use old version in this case.
         # https://github.com/autogluon/autogluon/pull/3190#issuecomment-1540599280
-        'catboost>=1.1,<1.2 ; sys_platform == "darwin"',
-        "catboost>=1.1,<1.3",
+        # Catboost 1.2 doesn't have wheel for python 3.11
+        "catboost>=1.1,<1.2 ; sys_platform == 'darwin' and python_version < '3.11'",
```
ddelange, line 42: they have it by now for mac os 11+, i updated my branch

common/src/autogluon/common/utils/try_import.py
```diff
@@ -30,7 +31,7 @@ def try_import_mxboard():
 
 
 def try_import_ray() -> ModuleType:
-    RAY_MAX_VERSION = "2.7.0"
+    RAY_MAX_VERSION = "2.7.0"  # sync with core/setup.py
```
ddelange, line 34: have some updates here as well on my branch, some critocal fixes for ray serve specific to py 3.11 were shipped with 2.7.0

tabular/setup.py
ddelange: I requested they revert their mac pipeline, no response yet though ref https://github.com/catboost/catboost/issues/2481

multimodal/setup.py
yinweisu: Didn't dive deep into why it cause segfault yet.

tabular/setup.py
yinweisu: So we did notice the wheel being available, but for some reason our platform test still tries to build it, which failed

common/src/autogluon/common/utils/try_import.py
yinweisu: ray 2.7 has breaking changes that will break our existing code. I think it's better to have a separate PR for the ray upgrade. I feel it would be a bit messy to have all changes being squashed into this PR

PR #3205, opened 2023-05-05
Deprecation utils

yinweisu:
*Issue #, if available:*

*Description of changes:*
* Reworked deprecation utils and added unit tests


common/src/autogluon/common/utils/deprecated.py
```diff
@@ -1,26 +1,230 @@
...
+            stacklevel=2,
         )
 
-        @functools.wraps(func)
-        def wrapped_func(*args, **kwargs):
-            warnings.warn(message=warning_message, category=DeprecationWarning, stacklevel=2)
-            return func(*args, **kwargs)
 
-        return wrapped_func
+def Deprecated(
```
Innixma, line 34: Is there a convention you are following for why it is capitalized?

common/tests/unittests/test_deprecated.py
```diff
@@ -0,0 +1,93 @@
+import pytest
+
+from autogluon.common.utils.deprecated import Deprecated, Deprecated_args
+
+
+@Deprecated(min_version_to_warn="0.0", min_version_to_error="9999.0")
+class ShouldDeprecateButNoErrorClass:
+    pass
```
Innixma, line 8: Missing test on micro-version changes (even though we probably won't use micro versions for deprecations in general, good to test that it does work correctly)

Innixma: Also: Missing test on pre-release version (Ensure that the logic works correctly in our pre-release version scheme)

common/src/autogluon/common/utils/deprecated.py
```diff
@@ -1,26 +1,230 @@
...
+                version_to_remove=version_to_remove,
+            )
+            return obj(*args, **kwargs)
+
+        return patched_func_with_warning_msg
+
+    return _decorator
+
+
+def rename_kwargs(
```
Innixma, line 133: If only used here, then make it private `_rename_kwargs`

common/tests/unittests/test_deprecated.py
```diff
@@ -0,0 +1,93 @@
...
+        NotDeprecatedClass().should_deprecate_but_no_error_func()
+        should_deprecate_but_no_error_func()
+        new_arg = func_with_deprecated_args_no_error(deprecated_arg=1)
+        assert new_arg == 1
+        deprecated_arg = func_with_deprecated_args_no_replacement(deprecated_arg=1)
+        assert deprecated_arg == 1
+
+
+def test_should_raise_deprecate_error():
+    with pytest.raises(Exception):
```
Innixma, line 78: Why `Exception`? Why not the specific exception we expect?

common/tests/unittests/test_deprecated.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+def test_should_not_deprecate():
+    NotDeprecatedClass().not_deprecated_func()
+    not_deprecated_func()
+    func_with_no_deprecated_args(new_arg=1)
+
+
+def test_should_raise_if_both_new_and_deprecated_args_passed():
+    with pytest.raises(Exception):
```
Innixma, line 92: ditto

common/src/autogluon/common/utils/deprecated.py
```diff
@@ -1,26 +1,230 @@
...
 
-def deprecated(message: str = "", version_removed: Optional[str] = None, *args, **kwargs):
-    def deprecated_decorator(func):
-        warning_message = (
-            f"Calling deprecated function {func.__name__}. This function is deprecated "
-            + "and will be removed in "
-            + (f"v{version_removed}. " if version_removed else "a future release. ")
-            + message
+    if error:
+        raise ValueError(msg)
```
Innixma, line 25: why ValueError? Is this the commonly used error for deprecation errors?

common/src/autogluon/common/utils/deprecated.py
```diff
@@ -1,26 +1,230 @@
...
-import warnings
 import functools
-from typing import Optional
+import inspect
+import warnings
+from typing import Any, Dict, Optional
+
+from packaging import version
 
+from ..version import __version__
```
Innixma, line 8: Should we add logic to be able to alter this value for testing purposes? AKA to pretend we are version `1.0.3`, etc?

common/src/autogluon/common/utils/deprecated.py
yinweisu: I've seen ray warpping it's api call with Capitalized decorator, so trying to mimic here. I feel that makes it pretty noticeable

common/tests/unittests/test_deprecated.py
yinweisu: Sure. Though the version comparison just uses `packaging.version.parse`. So if we assume this package is doing the right thing, there's no need to test that. I can still add a test as a safe guard

common/src/autogluon/common/utils/deprecated.py
yinweisu: Yea. This could come in handy

common/src/autogluon/common/utils/deprecated.py
yinweisu: There's no built-in deprecation error/exception. So just used the general ValueError. For the warning, we are throwing deprecation warning

common/tests/unittests/test_deprecated.py
yinweisu: It'll be ValueError then if you think that's ok. Otherwise, we'll need to create a custom exception

common/tests/unittests/test_deprecated.py
Innixma: "So if we assume this package is doing the right thing" : We should not make such assumptions, otherwise why test any of it at all if we assume everything is doing the right thing?

common/tests/unittests/test_deprecated.py
Innixma: ValueError is probably ok for now, unless someone has a better idea

common/src/autogluon/common/utils/deprecated.py
Innixma: ok sounds good

Innixma (approved):
LGTM once comments are addressed

common/src/autogluon/common/utils/deprecated.py
Innixma: sounds good

common/src/autogluon/common/utils/deprecated.py
```diff
@@ -1,26 +1,238 @@
...
+            stacklevel=2,
         )
 
-        @functools.wraps(func)
-        def wrapped_func(*args, **kwargs):
-            warnings.warn(message=warning_message, category=DeprecationWarning, stacklevel=2)
-            return func(*args, **kwargs)
 
-        return wrapped_func
+def Deprecated(
```
Innixma, line 34: One way to do this per version is to not use `__version__` at all here, and then to wrap the usage of `__version__` for each module.

for example, something like:

```
from autogluon.common.utils.deprecated import AbstractDeprecated

from .. import __version__

Deprecated = AbstractDeprecated(version=__version__)
```

Innixma:
```
from autogluon.tabular... import Deprecated  # uses autogluon.tabular.__version__
from autogluon.common... import Deprecated  # uses autogluon.common.__version__
```

yinweisu: Good idea! Will add in follow-up PR

PR #3209, opened 2023-05-09
Add wrapper to generate deprecated decorator

yinweisu:
*Issue #, if available:*

*Description of changes:*
* Add wrapper to generate deprecated decorator so the util can be used by modules like cloud


Innixma (approved):
LGTM!

Issue #3136, opened 2023-04-13
Add option to disable caching of local models


shchur:
Currently, all predictions made by local models like ETS and Naive are saved to disk. While this saves time when predicting many times for the same item, it introduces an overhead if we only predict once; it also may cause the CloudPredictor to crash when it's invoked in a read-only directory.

Adding a flag for disabling caching would be helpful to avoid these problems.

shchur:
#3214 disabled caching for local models, so this is not relevant anymore

Alex-Wenner-FHR:
I am experiencing this issue when deploying a TS AG model to sagemaker endpoints, with a custom inference script. It seems to fail because of a read only file system.

`OSError: [Errno 30] Read-only file system: 'AutogluonModels'`

and the code trying to load this model artifact directory:
```python
TimeSeriesPredictor(cache_predictions=False, log_to_file=False).load(path="autogluon-model")
```

I tried to just use `TimeSeriesPredictor.load(path="...")` but this failed trying to invoke certain models, with the same error. The two arguments passed into the `TimeSeriesPredictor` class was an attempt to force it to readonly.

- I am using AG latest, so 1.1.1. 

- It was trained on python 3.11.10 and inferenced on 3.11.9 (patch mismatch)

However, I am usure as to how to make AG only read and not write. Let me know if more information is needed! Thank you!
@shchur

Alex-Wenner-FHR:
@shchur apologies, I have fixed this by passing `model.predict(data, use_cache=False)` instead of passing the arguments into the `TimeSeriesPredictor` class itself! This can be closed, this resolved this issue!

Issue #3098, opened 2023-03-30
[BUG] RuntimeError when training AutoETS/AutoARIMA models outside of the name-main idiom


shchur:
- [ ] I have checked that this bug exists on the latest stable version of AutoGluon
- [x] and/or I have checked that this bug exists on the latest mainline of AutoGluon via source installation

**Describe the bug**
A `RuntimeError` is thrown when training StatsForecast models outside the `if __name__ == "__main__":` block in a Python script.
```
RuntimeError:
        An attempt has been made to start a new process before the
        current process has finished its bootstrapping phase.

        This probably means that you are not using fork to start your
        child processes and you have forgotten to use the proper idiom
        in the main module:

            if __name__ == '__main__':
                freeze_support()
```

**Expected behavior**
No exception is raised, all models train normally.

**To Reproduce**
Calling the following line of code in a `.py` script outside the `if __name__ == "__main__":` block raises an exception
```python
predictor = TimeSeriesPredictor().fit(train_data, hyperparameters={"AutoETS": {}, "AutoARIMA": {}})
```
or, equivalently, using presets that include AutoETS/AutoARIMA
```python
predictor = TimeSeriesPredictor().fit(train_data, presets="high_quality")
```

**Installed Versions**
Latest 

**Additional context**

shchur:
A solution to this problem is to move the code inside the name-main idiom, for example:
```python
if __name__ == "__main__":
     train_data = TimeSeriesDataFrame("data.csv")
     predictor = TimeSeriesPredictor().fit(train_data, hyperparameters={"AutoETS": {}, "AutoARIMA": {}})
```
We need to add this information to FAQ / tutorials.

shchur:
Fixed by #3214

Issue #2600, opened 2022-12-23
[BUG] TimeSeriesPredictor accepts invalid fit and __init__ kwargs arguments.


rjattrill:
EDITED by @Innixma:

> Note: The real issue here is that TimeSeriesPredictor is not raising an exception upon being passed an invalid kwarg. TimeSeriesPredictor should adopt TabularPredictor's invalid kwarg handling logic to properly error in this scenario. Code in TabularPredictor for this logic should be moved to a generic location and re-used.

========= Original Message =========

When using `excluded_model_types` argument in the `fit` method of `TimeSeriesPredictor` the excluded model is still used.   Note that `presets` argument has also been used.

Using the sample code from the tutorial and adding 'excluded_model_types':

```
predictor = TimeSeriesPredictor(
    path="autogluon-m4-hourly",  
    target="Value",  
    prediction_length=48,  
    eval_metric="MAPE", 
).fit(
    train_data=ts_dataframe,
    presets="fast_training",  
    time_limit=30,  
    excluded_model_types=["ETS"]
)

```
When executing we see the following:
`Models that will be trained: ['Naive', 'SeasonalNaive', 'ETS', 'Theta', 'ARIMA', 'AutoGluonTabular', 'DeepAR']`

Full code: https://gist.github.com/rjattrill/74d2aee51f02ab98dc7a8aa461699f32
Input file: https://gist.github.com/rjattrill/ff6fb7f06c771919759c33dcb095214e

Note: The reason for trying to exclude this is that the ETS model doesn't seem to observe the time_limit for training - which might be a separate issue.

**Installed Versions**
>>> from autogluon.core.utils import show_versions
>>> show_versions()
NOTE: Redirects are currently not supported in Windows or MacOs.

```
INSTALLED VERSIONS
------------------
date                   : 2022-12-23
time                   : 15:25:59.818857
python                 : 3.9.15.final.0
OS                     : Windows
OS-release             : 10
Version                : 10.0.19044
machine                : AMD64
processor              : Intel64 Family 6 Model 140 Stepping 1, GenuineIntel
num_cores              : 8
cpu_ram_mb             : 16103
cuda version           : None
num_gpus               : 0
gpu_ram_mb             : []
avail_disk_size_mb     : None

accelerate             : 0.13.2
albumentations         : 1.1.0
autogluon.common       : 0.6.1
autogluon.core         : 0.6.1
autogluon.features     : 0.6.1
autogluon.multimodal   : 0.6.1
[...]

PR #3214, opened 2023-05-16
[timeseries] Refactor local forecasting models

shchur:
*Issue #, if available:* #3098, #3136, #2600

*Description of changes:*

This PR unifies the code used by all local models (StatsForecast, statsmodels, naive). Now all models use Joblib for parallelization and no caching of the predictions is done by the models (but predictions are cached by the AbstractTimeSeriesTrainer). 

This leads to several improvements:
- All local models now support the `time_limit` (before this parameter was ignored).
- All models can be safely trained even outside the `if __name__ == "__main__"` idiom (#3098)
- Models from `statsmodels` / `naive` now train faster on large datasets thanks to replacing `loc[item_id]` with the faster `groupby` operation.
- Models from `StatsForecast` now train much faster on smaller datasets thanks to switching to the `loky` backed used by `joblib` (instead of the previously used `multiprocessing` backed used by `StatsForecast`). The main changes can be summarized as:
  - `multiprocessing` (old code) kills the processes after each individual model is trained. Therefore, each time we want to predict with a new model, `numba` needs to recompile the model code, which takes ~30-60 seconds for each model for each call to `predict`.
  - `loky` (this PR) reuses the worker processes when training multiple models (AutoETS, AutoARIMA, Theta). `numba` compiled code is preserved by these processes, which saves time.

In theory, it would be great to overcome this problem by simply caching the compiled `numba` code using `@njit(cache=True)` but, unfortunately, this randomly fails for some runs. I suspect that this is caused by https://github.com/numba/numba/issues/6972, as the crashes occur for models that rely on the [optimization method](https://github.com/Nixtla/statsforecast/blob/444f01e5014c22ced63e18863314bc75e98f68c5/statsforecast/ets.py#L558) that takes other functions as arguments.


timeseries/src/autogluon/timeseries/models/local/abstract_local_model.py
```diff
@@ -88,69 +100,97 @@ def _fit(self, train_data: TimeSeriesDataFrame, time_limit: int = None, **kwargs
...
-
-        self._fit_and_cache_predictions(data, quantile_levels=quantile_levels)
-        item_id_to_prediction = {}
-        for item_id, ts_hash in hash_ts_dataframe_items(data).items():
-            item_id_to_prediction[item_id] = self._cached_predictions[ts_hash]
-        predictions_df = pd.concat(item_id_to_prediction)
-        predictions_df.index.rename([ITEMID, TIMESTAMP], inplace=True)
-        return TimeSeriesDataFrame(predictions_df)
+        except Exception as e:
+            logger.debug(f"{self.name} failed with exception {e}, falling back to SeasonalNaive")
```
tonyhoo, line 155: Let's log `error` instead of `debug`, as this suggests potential issues in the stats model implementation that should not go unnoticed.

timeseries/src/autogluon/timeseries/models/local/statsforecast.py
```diff
@@ -1,112 +1,48 @@
...
+        time_series: pd.Series,
+        local_model_args: dict,
+    ) -> pd.DataFrame:
+        model_type = self._get_model_type()
+        model = model_type(**local_model_args)
+
+        levels = []
+        quantile_to_key = {}
+        for q in self.quantile_levels:
+            level = round(abs(q - 0.5) * 200, 1)
```
tonyhoo, line 35: Can you share some context on this calculation logics?

tonyhoo (approved):

PR #3215, opened 2023-05-16
[Tabular] Better mem approx

yinweisu:
*Issue #, if available:*

*Description of changes:*
* Updated our equation to estimate memory consumption. Now the equation is a linear regression from data points we collected on 15 datasets that have >100k rows. (We don't worry too much about small dataset memory consumption)
  * r2 score for lgb compared to our old heuristic:  0.86 vs -90
  * r2 score for xgb compared to our old heuristic:  0.89 vs -13
  * r2 score for cat compared to our old heuristic:  0.81 vs 0.5

### DIFF
--- a/tabular/src/autogluon/tabular/models/catboost/catboost_model.py
+++ b/tabular/src/autogluon/tabular/models/catboost/catboost_model.py
@@ -14,7 +14,7 @@ from autogluon.common.utils.try_import import try_import_catboost
 
 from .callbacks import EarlyStoppingCallback, MemoryCheckCallback, TimeCheckCallback
 from .catboost_utils import get_catboost_metric_from_ag_metric
-from .hyperparameters.parameters import get_param_baseline
+from .hyperparameters.parameters import get_param_baseline, DEFAULT_ITERATIONS
 from .hyperparameters.searchspaces import get_default_searchspace
 
 logger = logging.getLogger(__name__)
@@ -62,7 +62,8 @@ class CatBoostModel(AbstractModel):
     def _estimate_memory_usage(self, X, **kwargs):
         num_classes = self.num_classes if self.num_classes else 1  # self.num_classes could be None after initialization if it's a regression problem
         data_mem_usage = get_approximate_df_mem_usage(X).sum()
-        approx_mem_size_req = data_mem_usage * 7 + data_mem_usage / 4 * num_classes  # TODO: Extremely crude approximation, can be vastly improved
+        iterations = self.params.get('iterations', DEFAULT_ITERATIONS)
+        approx_mem_size_req = 76.55 * data_mem_usage + 25.24 * num_classes - 0.16 * iterations - 9.38 * len(self.features)  # Linear regression from collected memory profiling data points
         return approx_mem_size_req
 
     # TODO: Use Pool in preprocess, optimize bagging to do Pool.split() to avoid re-computing pool for each fold! Requires stateful + y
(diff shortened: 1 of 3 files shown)


gradientsky (approved):

tabular/src/autogluon/tabular/models/catboost/catboost_model.py
```diff
@@ -62,7 +62,8 @@ def _preprocess_nonadaptive(self, X, **kwargs):
     def _estimate_memory_usage(self, X, **kwargs):
         num_classes = self.num_classes if self.num_classes else 1  # self.num_classes could be None after initialization if it's a regression problem
         data_mem_usage = get_approximate_df_mem_usage(X).sum()
-        approx_mem_size_req = data_mem_usage * 7 + data_mem_usage / 4 * num_classes  # TODO: Extremely crude approximation, can be vastly improved
+        iterations = self.params.get('iterations', DEFAULT_ITERATIONS)
+        approx_mem_size_req = 76.55 * data_mem_usage + 25.24 * num_classes - 0.16 * iterations - 9.38 * len(self.features)  # Linear regression from collected memory profiling data points
```
Innixma, line 66: Because `data_mem_usage` is astronomically larger than all the other variables, it is the only thing that matters in this equation. `76.55` is a very large modifier. I'm not confident that this is going to be correct.

tabular/src/autogluon/tabular/models/lgb/lgb_model.py
```diff
@@ -64,7 +64,8 @@ def _get_stopping_metric_internal(self):
     def _estimate_memory_usage(self, X, **kwargs):
         num_classes = self.num_classes if self.num_classes else 1  # self.num_classes could be None after initialization if it's a regression problem
         data_mem_usage = get_approximate_df_mem_usage(X).sum()
-        approx_mem_size_req = data_mem_usage * 7 + data_mem_usage / 4 * num_classes  # TODO: Extremely crude approximation, can be vastly improved
+        num_boost_round = self.params.get('num_boost_round', DEFAULT_NUM_BOOST_ROUND)
+        approx_mem_size_req = 0.11 * data_mem_usage -0.17 * num_classes + 0.014 * num_boost_round + 1.28 * len(self.features)  # Linear regression from collected memory profiling data points
```
Innixma, line 68: So we are saying LightGBM uses 76.55/0.11 = 800x less memory than CatBoost? I don't think that is true.

tabular/src/autogluon/tabular/models/xgboost/xgboost_model.py
```diff
@@ -180,7 +180,8 @@ def _ag_params(self) -> set:
     def _estimate_memory_usage(self, X, **kwargs):
         num_classes = self.num_classes if self.num_classes else 1  # self.num_classes could be None after initialization if it's a regression problem
         data_mem_usage = get_approximate_df_mem_usage(X).sum()
-        approx_mem_size_req = data_mem_usage * 7 + data_mem_usage / 4 * num_classes  # TODO: Extremely crude approximation, can be vastly improved
+        n_estimators = self.params.get('n_estimators', DEFAULT_NUM_BOOST_ROUND)
+        approx_mem_size_req = -0.19 * data_mem_usage + 6.2 * num_classes + 0.015 * n_estimators + 1.34 * len(self.features)  # Linear regression from collected memory profiling data points
```
Innixma, line 184: so the bigger the data the less memory XGBoost uses???? No.

Innixma (changes requested):

Innixma:
Closing as we found that the memory approximations were not as accurate as other alternatives. (replaced by #3421)

PR #3216, opened 2023-05-16
[Tabular] Update hpo preset mapping

yinweisu:
*Issue #, if available:*

*Description of changes:*
* Add preset mapping so user don't specify `local`, which is weird, when using distributed hpo in cloud


Innixma (approved):
LGTM, but consider adding docstring about this. Currently I think it is hidden.

yinweisu:
> LGTM, but consider adding docstring about this. Currently I think it is hidden.

Yes it is hidden kind of intentionally because user cannot directly use these preset but need to use it with cloud instead.

PR #3217, opened 2023-05-16
Fix multimodal hpo cpu usage

yinweisu:
*Issue #, if available:*
https://github.com/autogluon/autogluon/issues/2366

*Description of changes:*
* Correctly set torch thread now with cpu training. Torch thread doesn't affect gpu training as cpu only used to load data and is controlled by `num_workers`

### DIFF
--- a/multimodal/src/autogluon/multimodal/predictor.py
+++ b/multimodal/src/autogluon/multimodal/predictor.py
@@ -121,7 +121,6 @@ from .utils import (
     assign_feature_column_names,
     average_checkpoints,
     check_if_packages_installed,
-    compute_num_cpus,
     compute_num_gpus,
     compute_score,
     convert_pred_to_xywh,
@@ -825,7 +824,7 @@ class MultiModalPredictor(ExportMixin):
         if hpo_mode:
             # TODO: allow custom gpu
             assert self._resume is False, "You can not resume training with HPO"
-            resources = dict(num_cpus=ResourceManager.get_cpu_count(), num_gpus=ResourceManager.get_gpu_count_torch())
+            resources = dict(num_gpus=ResourceManager.get_gpu_count_torch())
             if _fit_args["max_time"] is not None:
                 _fit_args["max_time"] *= 0.95  # give some buffer time to ray lightning trainer
             _fit_args["predictor"] = self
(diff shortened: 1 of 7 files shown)


multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -1496,11 +1501,12 @@ def _fit(
                 ".* in the `DataLoader` init to improve performance.*",
             )
             warnings.filterwarnings("ignore", "Checkpoint directory .* exists and is not empty.")
-            trainer.fit(
-                task,
-                datamodule=train_dm,
-                ckpt_path=ckpt_path if resume else None,  # this is to resume training that was broken accidentally
-            )
+            with set_torch_num_threads(num_cpus=num_cpus):
```
zhiqiangdon, line 1504: Can we use lightning's trainer's `accelerator` and `devices` to setup cpu training?
https://lightning.ai/docs/pytorch/stable/common/trainer.html#accelerator
https://lightning.ai/docs/pytorch/stable/common/trainer.html#devices

multimodal/src/autogluon/multimodal/utils/hpo.py
```diff
@@ -36,7 +41,8 @@ def hpo_trial(sampled_hyperparameters, predictor, checkpoint_dir=None, **_fit_ar
     if checkpoint_dir is not None:
         _fit_args["resume"] = True
         _fit_args["ckpt_path"] = os.path.join(checkpoint_dir, RAY_TUNE_CHECKPOINT)
-    predictor._fit(**_fit_args)
+    with set_torch_num_threads(num_cpus=num_cpus):
```
zhiqiangdon, line 44: What will happen if multiple trials are running in parallel? Will they have conflicts in using `num_cpus`? Is `num_cpus ` is the number of cpus of this machine?

yinweisu: num_cpus is being calculated already. If 1 machine has 32 cpus and 2 trials running in parallel. Within each trial, the `num_cpus` will be 16

zhiqiangdon: Since `num_cpus` is already calculated before calling this function `hpo_trial()`, why does ray still set `torch.set_num_threads(1)`?

yinweisu: I don't know why ray set that. We tell ray how much cpu to use for each tasks and it's only being used by ray for scheduling purpose. We need to control the amount of compute the task actually used. The same is being done inside tabular too.

We don't want to do a `torch.set_num_threads()` in core as resource control are specific to different models

zhiqiangdon (approved):
LGTM

PR #3220, opened 2023-05-17
[multimodal]Enable backbone freezing to boost finetuning speed and save GPU usage

FANGAreNotGnu:
The current two-stage/layerwise-decay learning rate settings supports use different lr in head and non-head layers. It also allows user to set lr in non-head layers to 0. However,
- Most models are more complicated than backbone + head. And we may want to use two stage on head v.s. non-head, and layer freeze on backbone v.s. non-backbone. (e.g. `backbone_lr = 1e-5, neck_lr=1e-5, head_lr=1e-3` and `backbone_lr = 0, neck_lr=1e-4, head_lr=1e-4` may both converge fast and nice but `backbone_lr = 0, neck_lr=0, head_lr=1e-4` may not be a good one)
- Only setting lr = 0 in two-stage does not save GPU memory. Setting `require_grad=False` for backbone parameters **can save lots of GPU memory and enable us to run a larger batch_size or a larger model.**

So here we introduced a new hyperparameter `model.mmdet.frozen_layers` that disables gradient update for backbone. It can be used together with any lr_choice, i.e. "single_lr", "two_stage", or "layerwise_decay". 

Future work:
Due to bandwidth limit, it's only added to lit_mmdet. Will need benchmarking on other problem types and add to corresponding lit modules. Also the lit modules may need a refactor (add an base module for a better OOP design).


tonyhoo:
Thank @FANGAreNotGnu, can you run some quick tests to make sure the memory footprint has been reduced post this change?

FANGAreNotGnu:
> Thank @FANGAreNotGnu, can you run some quick tests to make sure the memory footprint has been reduced post this change?

Yes I did, the YOLOX-L GPU usage with batch_size=8 is reduced from >16GB to <10GB. (trainable parameters from 54M to 27M)

multimodal/src/autogluon/multimodal/models/mmdet_image.py
```diff
@@ -390,6 +391,18 @@ def get_layer_ids(
 
         return name_to_id
 
+    def get_backbone_layer_names(self):
+        backbone_layer_names = []
+        backbone_layers_patterns = [
```
tonyhoo, line 396: does this cover all backbone layer names used by mmdet?

multimodal/src/autogluon/multimodal/models/mmdet_image.py
```diff
@@ -390,6 +391,18 @@ def get_layer_ids(
...
+            "backbone",
+            "encoder",
+        ]
+        for n, _ in self.named_parameters():
+            for pattern in backbone_layers_patterns:
```
tonyhoo, line 401: What if the layer name has both "backbone" and "encoder" in it? The layer name will be added twice in current impl? Is that intended?

multimodal/src/autogluon/multimodal/models/mmdet_image.py
FANGAreNotGnu: It covers all the common models, i.e. non transformer models use "backbone", transformer based models use "encoder".

multimodal/src/autogluon/multimodal/models/mmdet_image.py
FANGAreNotGnu: Thanks for pointing this out. Made the edits to use `any()`.

Ideally "model.backbone.xxx" parameter name is used by non transformer models and "model.encoder.xxx" is used by transformer based models. So there should be no conflicts. Also it won't cause a problem even if there is.

multimodal/src/autogluon/multimodal/optimization/utils.py
```diff
@@ -628,6 +628,73 @@ def get_trainable_params_efficient_finetune(
     return trainable_param_names
 
 
+def apply_freeze_backbone_lr(
```
zhiqiangdon, line 631: Is freezing backbone is one case of two stages lr? The difference is whether setting the requires_grad as False.

zhiqiangdon:
> The current two-stage/layerwise-decay learning rate settings supports use different lr in head and non-head layers. It also allows user to set lr in non-head layers to 0. However,
> 
> * Most models are more complicated than backbone + head. And we may want to use two stage on head v.s. non-head, and layer freeze on backbone v.s. non-backbone. (e.g. `backbone_lr = 1e-5, neck_lr=1e-5, head_lr=1e-3` and `backbone_lr = 0, neck_lr=1e-4, head_lr=1e-4` may both converge fast and nice but `backbone_lr = 0, neck_lr=0, head_lr=1e-4` may not be a good one)
> * Only setting lr = 0 in two-stage does not save GPU memory. Setting `require_grad=False` for backbone parameters **can save lots of GPU memory and enable us to run a larger batch_size or a larger model.**
> 
> By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Looks like this PR only addresses the second point, i.e., setting the requires_grad=True for backbone parameters.

multimodal/src/autogluon/multimodal/optimization/utils.py
zhiqiangdon: I see. But in your example, the neck_lr and head_lr are still the same, so the corresponding layers including neck can be treated as "head" layers.

zhiqiangdon: I guess most users don't need to set different learning rates for neck and head. So, from the view of learning rate, it's still two-stage.

zhiqiangdon: I think we may need to improve `apply_two_stages_lr` to support requires_grad=False for the parameters whose learning rate is 0. For object detection models, the `head_layer_names` attribute can have both neck and head layers.

FANGAreNotGnu: Our current two-stage lr implementation have two issues:
- Only support using head layer v.s. other layers
- Use base `lr` on other layers, and `lr * lr_mult` on head layers
To support backbone frozen in current two stage implementation, we need to:
- Add hyperparameter providing option to freeze layers with `lr=0.`
- Add hyperparameter to support using head layer v.s. other layers or backbone layer v.s. other layers.
- Use base `lr` on head/non-backbone layers, and `lr / lr_mult` on non-head/backbone layers
The third change involves an implicit change in API and may cause users' confusion.

FANGAreNotGnu: Two stage lr on head layer v.s. other layers is still useful since it converges faster.

zhiqiangdon: Right. `apply_two_stages_lr` can't directly support lr=0 for backbone parameters due to `lr * lr_mult`. I think there are two choices:

- Changing the design `lr_mult` in two-stage lr. Let `lr` be the head lr and the backbone lr is still `lr * lr_mult` but with 0<=`lr_mult`<1.
- Using  `apply_layerwise_lr_decay ` instead of `apply_two_stages_lr`. `apply_layerwise_lr_decay ` can simulate the case where head uses `lr` and backbone use 0 learning rate. In general, layerwise learning rate decay should not perform better than two-stage if we can have a good `lr_decay`.

zhiqiangdon: I think the definition of head layers can be flexible. For object detection models, we can count the neck layers as head since we want to finetune them together the real head layers.

multimodal/src/autogluon/multimodal/optimization/utils.py
```diff
@@ -629,6 +629,58 @@ def get_trainable_params_efficient_finetune(
...
+    for n, p in model.named_parameters():
+        if not is_nonfrozen_layer(n):
+            p.requires_grad = False
+
+    for group_idx, group_param in enumerate(optimizer_grouped_parameters):
+        updated_params = []
+        for p in group_param["params"]:
+            if p.requires_grad:
+                updated_params.append(p)
+        optimizer_grouped_parameters[group_idx]["params"] = updated_params
```
zhiqiangdon, line 679: Filtering out optimizer parameters whose requires_grad is False is general functionality. It doesn't have to be related to freezing the whole backbone. How about making this filtering an independent function?

multimodal/src/autogluon/multimodal/optimization/utils.py
```diff
@@ -629,6 +629,58 @@ def get_trainable_params_efficient_finetune(
...
+    The grouped parameters or their names.
+    """
+    if hasattr(model, "backbone_layer_names"):
+        is_nonfrozen_layer = lambda n: not any(bb in n for bb in model.backbone_layer_names)
+    else:  # freeze all non-head layers if backbone layers are not specified
+        is_nonfrozen_layer = lambda n: any(bb in n for bb in model.head_layer_names)
+
+    for n, p in model.named_parameters():
+        if not is_nonfrozen_layer(n):
+            p.requires_grad = False
```
zhiqiangdon, line 672: A better design may be setting requires_grad to False when initializing the model in `mmdet_image.py`. The model doesn't need the `backbone_layer_names ` attribute.

multimodal/src/autogluon/multimodal/models/mmdet_image.py
```diff
@@ -142,6 +142,7 @@ def _load_checkpoint(self, checkpoint_file):
 
         self.name_to_id = self.get_layer_ids()
         self.head_layer_names = [n for n, layer_id in self.name_to_id.items() if layer_id <= 0]
+        self.backbone_layer_names = self.get_backbone_layer_names()
```
zhiqiangdon, line 145: Instead of adding `backbone_layer_names ` and `get_backbone_layer_names `, we can have one class function to set `requires_grad` to False according to an additional hyperparameter `frozen_parts`.

multimodal/src/autogluon/multimodal/configs/optimization/adamw.yaml
```diff
@@ -3,6 +3,7 @@ optimization:
   learning_rate: 1.0e-4
   weight_decay: 0.001
   lr_choice: "layerwise_decay"
+  freeze_backbone: False
```
zhiqiangdon, line 6: Consider removing this hyperparameter and add `frozen_parts` under the `mmdet_image` model config.

FANGAreNotGnu:
Changed the design based on some offline discussions.

multimodal/src/autogluon/multimodal/optimization/utils.py
```diff
@@ -629,6 +629,31 @@ def get_trainable_params_efficient_finetune(
     return trainable_param_names
 
 
+def remove_parameters_without_grad(
+    grouped_parameters: List,
```
zhiqiangdon, line 633: `List` -> `List[Dict]` to be more concrete.

multimodal/src/autogluon/multimodal/models/utils.py
```diff
@@ -733,3 +733,25 @@ def run_model(model: nn.Module, batch: dict, trt_model: Optional[nn.Module] = No
     else:
         output = model(batch)
     return output
+
+
+def freeze_model_layers(model, frozen_layers):
+    """
+    Get an MMOCR config and model based on a checkpoint name.
```
zhiqiangdon, line 740: Docstrings don't match function.

multimodal/src/autogluon/multimodal/configs/model/fusion_mlp_image_text_tabular.yaml
```diff
@@ -149,6 +149,7 @@ model:
       - "image"
     max_img_num_per_col: 1
     output_bbox_format: "xyxy"  # now support xyxy or xywh, for bbox format details see https://keras.io/api/keras_cv/bounding_box/formats/
+    frozen_layers: null
```
zhiqiangdon, line 152: What if removing `null`? What's the difference between nothing and `null`?

FANGAreNotGnu: I think there's no difference. We are using both in the configs.

zhiqiangdon (approved):
LGTM.

PR #3230, opened 2023-05-22
[timeseries] Update to GluonTS v0.13.0 + add DLinear and PatchTST models

shchur:
*Description of changes:*
- Bump GluonTS dependency to v0.13.0.
- Add two new models: `DLinear` and `PatchTST` (not yet added to the presets).
- Clean up the logic for setting the default `context_length` for GluonTS models.
- Speed up the methods for converting `DistributionForecast` to `QuantileForecast` by vectorizing the code over all `quantile_levels`. Comparison of prediction time in seconds on M4 Quarterly (24000 time series) for SimpleFeedForward model that produces a `DistributionForecast`.
 
    | Version | Prediction time (s) |
    |--------|--------|
    | Old | 241 |
    | New (this PR) | 64 |

    The overall prediction time is still rather slow and probably could be further improved, but that would require digging into the GluonTS internals.


timeseries/src/autogluon/timeseries/models/__init__.py
```diff
@@ -13,6 +13,8 @@
 
 __all__ = [
     "DeepARModel",
+    "DLinearModel",
```
tonyhoo, line 16: I am curious if there is a particular reason or quantitative analysis supporting the inclusion of these two models

timeseries/src/autogluon/timeseries/models/gluonts/torch/models.py
```diff
@@ -124,21 +125,18 @@ def load(cls, path: str, reset_paths: bool = True, verbose: bool = True) -> "Abs
...
         forecast: DistributionForecast, quantile_levels: List[float]
     ) -> QuantileForecast:
-        forecast_arrays = [forecast.mean]
-        forecast_keys = ["mean"]
-
-        for q in quantile_levels:
-            forecast_arrays.append(forecast.quantile(q))
-            forecast_keys.append(str(q))
+        # Compute all quantiles in parallel instead of a for-loop
+        quantiles = torch.tensor(quantile_levels, device=forecast.distribution.mean.device).reshape(-1, 1)
```
tonyhoo, line 129: Any reason why we need to get the device from forecast distribution?

tonyhoo (approved):

PR #3231, opened 2023-05-22
[timeseries] Add excluded_model_types and simplify preset generation logic

shchur:
*Issue number*: #2600

*Description of changes:*
- Gracefully handle the case when users incorrectly add the `Model` suffix to the model name (e.g., specify `DeepARModel` instead of `DeepAR`). This is a common pain point for users since models are listed with the `Model` suffix in the [documentation](https://auto.gluon.ai/stable/tutorials/timeseries/forecasting-model-zoo.html) because of Sphinx.
- New argument `excluded_model_types` for `TimeSeriesPredictor.fit` that allows users to exclude certain models from the presets, replicating the API of `TabularPredictor`.
- Remove some untested/undocumented/probably unsupported functionality for `AbstractTimeSeriesModelFactory` & `GenericGluonTSMXNetModelFactory`.

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/presets.py
+++ b/timeseries/src/autogluon/timeseries/models/presets.py
@@ -168,6 +168,7 @@ def get_default_hps(key):
 
 
 def normalize_model_type_name(model_name: str) -> str:
+    """Remove 'Model' suffix from the end of the string, if it's present."""
     if model_name.endswith("Model"):
         model_name = model_name[: -len("Model")]
     return model_name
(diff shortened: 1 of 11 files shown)


timeseries/src/autogluon/timeseries/models/presets.py
```diff
@@ -193,30 +201,52 @@ def get_preset_models(
             f"hyperparameters must be a dict, a string or None (received {type(hyperparameters)}). "
             f"Please see the documentation for TimeSeriesPredictor.fit"
         )
+    # Handle model names ending with "Model", e.g., "DeepARModel" is mapped to "DeepAR"
+    hyperparameters_clean = {}
+    for key, value in hyperparameters.items():
+        if isinstance(key, str):
+            key = normalize_model_type_name(key)
+        hyperparameters_clean[key] = value
```
Innixma, line 209: What if the user provides the following:

```python3
hyperparameters = {
'DeepAR': {'epochs': 5},
'DeepARModel': {'epochs': 10}
}
```

I believe this edge-case is not accounted for in an expected manner atm.

Valid options:
1. Error
2. hyperparameters_clean = {'DeepAR': [{'epochs': 5}, {'epochs': 10}]}
3. Warn and tell user which one takes priority.

Innixma, line 209: Move to a method/function, can then add a unit test

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -448,6 +449,8 @@ def fit(
                         "num_trials": 5,
                     }
                 )
+        excluded_model_types: List[str], optional
+            Banned subset of model types to avoid training during ``fit()``, even if present in ``hyperparameters``.
```
Innixma, line 453: Add 1-liner example in the docstring to help illustrate to users

Innixma (approved):
LGTM, added a couple comments

timeseries/src/autogluon/timeseries/models/presets.py
shchur: I'm testing that this function works correctly at the level of the trainer (`test_when_some_models_have_incorrect_suffix_then_correct_model_are_trained`). Are there some other test cases you think we should cover?

Innixma: thats fine

Issue #2808, opened 2023-02-01
[TimeSeries] Add AutoGluon version check during load


Innixma:
Identical to #2807, but for `autogluon.timeseries`.

PR #3233, opened 2023-05-23
[timeseries] Check AG-TS version when loading TimeSeriesPredictor

shchur:
*Issue #, if available:* #2808

*Description of changes:*
- Move logic for checking if the current AG version matches the saved version of the predictor to `autogluon.common.utils.utils`
- Use this logic inside `TimeSeriesPredictor` to keep track of the version used to train the model.

### DIFF
--- a/tabular/src/autogluon/tabular/predictor/predictor.py
+++ b/tabular/src/autogluon/tabular/predictor/predictor.py
@@ -18,7 +18,7 @@ from autogluon.common.savers import save_json
 from autogluon.common.utils.file_utils import get_directory_size, get_directory_size_per_file
 from autogluon.common.utils.log_utils import set_logger_verbosity
 from autogluon.common.utils.pandas_utils import get_approximate_df_mem_usage
-from autogluon.common.utils.utils import setup_outputdir, get_autogluon_metadata, compare_autogluon_metadata
+from autogluon.common.utils.utils import setup_outputdir, get_autogluon_metadata, compare_autogluon_metadata, check_saved_predictor_version
 from autogluon.core.constants import BINARY, MULTICLASS, REGRESSION, QUANTILE, AUTO_WEIGHT, BALANCE_WEIGHT, PSEUDO_MODEL_SUFFIX, PROBLEM_TYPES_CLASSIFICATION
 from autogluon.core.data.label_cleaner import LabelCleanerMulticlassToBinary
 from autogluon.core.dataset import TabularDataset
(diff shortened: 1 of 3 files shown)


tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -3157,43 +3157,35 @@ def load(cls, path: str, verbosity: int = None, require_version_match: bool = Tr
 
         try:
             from ..version import __version__
-            version_load = __version__
+            version_current = __version__
```
shchur, line 3160: The existing naming was a bit confusing to me: I would expect `version_load` to correspond to the version of the loaded predictor, but in the old code it referred to the version of AG installed in the system. I changed the names `version_load` -> `version_current` and `version_init` -> `version_saved` to address this, but if you prefer the old naming, I will undo the change.

Innixma: LGTM

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -671,13 +684,34 @@ def load(cls, path: str) -> "TimeSeriesPredictor":
...
+
         logger.info(f"Loading predictor from path {path}")
         learner = AbstractLearner.load(path)
         predictor = load_pkl.load(path=learner.path + cls.predictor_file_name)
         predictor._learner = learner
         predictor.path = learner.path
         return predictor
 
+    def _save_version_file(self):
+        version_file_contents = str(current_ag_version)
```
Innixma, line 711: isn't it already a string?

Innixma (approved):
LGTM!

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -9,15 +9,16 @@
...
-from autogluon.core.utils.loaders import load_pkl
-from autogluon.core.utils.savers import save_pkl
+from autogluon.core.utils.loaders import load_pkl, load_str
+from autogluon.core.utils.savers import save_pkl, save_str
 from autogluon.timeseries.configs import TIMESERIES_PRESETS_CONFIGS
 from autogluon.timeseries.dataset.ts_dataframe import ITEMID, TIMESTAMP, TimeSeriesDataFrame
 from autogluon.timeseries.learner import AbstractLearner, TimeSeriesLearner
 from autogluon.timeseries.trainer import AbstractTimeSeriesTrainer
 from autogluon.timeseries.utils.random import set_random_seed
+from autogluon.timeseries.version import __version__ as current_ag_version
```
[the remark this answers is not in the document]
shchur: Fixed the import to `from autogluon.timeseries import __version__`.

As for absolute imports: we decided to stick to absolute imports in AG-TS wherever possible, as recommended by [PEP8](https://peps.python.org/pep-0008/#imports).

PR #3237, opened 2023-05-25
[WIP] [timeseries] Implement prediction caching for all models

shchur:
*Description of changes:*
- WIP


timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -92,13 +92,12 @@ class TimeSeriesPredictor:
         If True, the predictor will ignore the datetime indexes during both training and testing, and will replace
         the data indexes with dummy timestamps in second frequency. In this case, the forecast output time indexes will
         be arbitrary values, and seasonality will be turned off for local models.
-    learner_type : AbstractLearner, default = TimeSeriesLearner
-        A class which inherits from ``AbstractLearner``. The learner specifies the inner logic of the
-        ``TimeSeriesPredictor``.
-    learner_kwargs : dict, optional
-        Keyword arguments to send to the learner (for advanced users only). Options include ``trainer_type``, a
-        class inheriting from ``AbstractTrainer`` which controls training of multiple models.
-        If ``path`` and ``eval_metric`` are re-specified within ``learner_kwargs``, these are ignored.
```
shchur, line 101: We provide no documentation explaining how these kwargs could be used, so I removed them from the API to reduce clutter.

timeseries/src/autogluon/timeseries/models/ensemble/greedy_ensemble.py
```diff
@@ -110,27 +110,14 @@ def predict(self, data: Dict[str, TimeSeriesDataFrame], **kwargs) -> TimeSeriesD
...
-        if len(failed_models) == len(data):
-            raise RuntimeError(f"All input models failed during prediction, {self.name} cannot predict.")
-        if len(failed_models) > 0:
-            logger.warning(
-                f"Following models failed during prediction: {failed_models}. "
-                f"{self.name} will set the weight of these models to zero and re-normalize the weights when predicting."
-            )
+        for model_name, model_pred in data.items():
+            if model_pred is None:
+                raise RuntimeError(f"{self.name} cannot predict because base model {model_name} failed.")
```
Innixma, line 115: nit: Consider listing all failed model names instead of only the first failed model. This will help to identify bugs during development easier.

timeseries/src/autogluon/timeseries/models/local/statsmodels.py
```diff
@@ -143,7 +143,7 @@ def _predict_with_local_model(
 
         results = [predictions.predicted_mean.rename("mean")]
         coverage_fn = lambda alpha: predictions.pred_int(alpha=alpha)
-        results += get_quantiles_from_statsmodels(coverage_fn=coverage_fn, quantile_levels=quantile_levels)
```
Innixma, line 146: How did this work before? `quantile_levels` appears undefined

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -878,15 +862,174 @@ def score(
...
+            return data
+
+    def get_model_pred_dict(
+        self,
+        model_names: List[str],
+        data: TimeSeriesDataFrame,
+        known_covariates: Optional[TimeSeriesDataFrame] = None,
+        record_pred_time: bool = False,
+        raise_exception_if_failed: bool = True,
+    ) -> Union[Dict[str, TimeSeriesDataFrame], Tuple[Dict[str, TimeSeriesDataFrame], Dict[str, float]]]:
```
Innixma, line 905: I have a very similar method in Tabular. I wonder if we can generalize this and share the implementation. If so, that would make life a lot easier in future. That would be its own PR, and probably require some careful thought.

Innixma: Perhaps add a TODO to unify with Tabular at some point

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -878,15 +862,174 @@ def score(
...
+        known_covariates
+            Future values of the known covariates.
+        record_pred_time
+            If True, will additionally return the total prediction times for all models (including the prediction time
+            for base models). If False, will only return the model predictions.
+        raise_exception_if_failed
+            If True, the method will raise an exception if any model crashes during prediction.
+            If False, error will be logged and predictions for failed models will contain None.
+        """
+        if self.cache_predictions:
```
Innixma, line 923: Consider adding `use_cache` as an argument with default True, so that we can recompute even if we have the cache for the purposes of measuring predict times or as a sanity check.

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -878,15 +862,174 @@ def score(
...
+        pred_time_dict_total = defaultdict(float)
+        for model_name in pred_time_dict_marginal.keys():
+            for base_model in self.get_minimum_model_set(model_name):
+                if pred_time_dict_marginal[base_model] is not None:
+                    pred_time_dict_total[model_name] += pred_time_dict_marginal[base_model]
+        return dict(pred_time_dict_total)
+
+    @property
+    def _cached_predictions_filename(self) -> Path:
+        return Path(self.path) / "cached_predictions.pkl"
```
Innixma, line 980: "cached_predictions.pkl" -> Make a class variable (example: https://github.com/autogluon/autogluon/blob/master/core/src/autogluon/core/trainer/abstract_trainer.py#L45-L49)

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -878,15 +862,174 @@ def score(
...
+        data: TimeSeriesDataFrame, known_covariates: Optional[TimeSeriesDataFrame] = None
+    ) -> str:
+        """Compute a unique string that identifies the time series dataset."""
+        from hashlib import md5
+
+        def hash_df(df: Optional[pd.DataFrame]) -> str:
+            if df is not None:
+                df = pd.DataFrame(data)
+                df.sort_index(inplace=True, axis=0)
+                df.reset_index(inplace=True)
```
Innixma, line 993: Why reset index?

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -878,15 +862,174 @@ def score(
...
+        return Path(self.path) / "cached_predictions.pkl"
+
+    @staticmethod
+    def _compute_dataset_hash(
+        data: TimeSeriesDataFrame, known_covariates: Optional[TimeSeriesDataFrame] = None
+    ) -> str:
+        """Compute a unique string that identifies the time series dataset."""
+        from hashlib import md5
+
+        def hash_df(df: Optional[pd.DataFrame]) -> str:
```
Innixma, line 989: Should this live in autogluon.common? Seems generic. Probably something that Tabular could use to enhance some sanity checks.

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -878,15 +862,174 @@ def score(
...
+    def _save_cached_pred_dicts(
+        self, dataset_hash: str, model_pred_dict: Dict[str, TimeSeriesDataFrame], pred_time_dict: Dict[str, float]
+    ) -> None:
+        if self._cached_predictions_filename.exists():
+            logger.debug(f"Extending existing cached predictions")
+            cached_predictions = load_pkl.load(str(self._cached_predictions_filename))
+        else:
+            cached_predictions = {}
+        # Do not save results for models that failed
+        cached_predictions[dataset_hash] = {
```
Innixma, line 1027: I like this key, that is a good idea

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -878,15 +862,174 @@ def score(
...
+        return {}, {}
+
+    def _save_cached_pred_dicts(
+        self, dataset_hash: str, model_pred_dict: Dict[str, TimeSeriesDataFrame], pred_time_dict: Dict[str, float]
+    ) -> None:
+        if self._cached_predictions_filename.exists():
+            logger.debug(f"Extending existing cached predictions")
+            cached_predictions = load_pkl.load(str(self._cached_predictions_filename))
+        else:
+            cached_predictions = {}
```
Innixma, line 1025: Theoretically this file could get cumbersome if we were to save many caches. Another option would be to save files `cache_N.pkl`, with a dictionary of `dataset_hash -> filename`. Don't worry about actually implementing this though, it is only relevant if such a scenario is realistic.

Innixma (approved):
LGTM!

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
Innixma: If standalone, would be useful to add a unit test (for example, hash a df, save it, load it, hash the loaded, check for equivalence).

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
shchur: Nice idea, thanks!

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
shchur: Moved to `autogluon.common.utils.uitls` & added tests to `autogluon/common/tests/unittests/test_hash_pandas_df.py`.

timeseries/src/autogluon/timeseries/models/local/statsmodels.py
shchur: Fixed in #3252

PR #3239, opened 2023-05-25
Add support for included_model_types

yinweisu:
*Issue #, if available:*
https://github.com/autogluon/autogluon/issues/1458

*Description of changes:*
* Added ModelFilter as a util class to exclude and include models
* Added `included_model_types` support


common/src/autogluon/common/model_filter/model_filter.py
```diff
@@ -0,0 +1,106 @@
...
+        """
+        if isinstance(models, dict):
+            included_models = {model: val for model, val in models.items() if model in included_model_types}
+            missing_models = set(included_model_types) - set(included_models.keys())
+        elif isinstance(models, list):
+            included_models = [model for model in models if model in included_model_types]
+            missing_models = set(included_model_types) - set(included_models)
+        if len(missing_models) > 0:
+            logger.warning(
+                f"The following models: {missing_models} are not present in model list specified by the user: {models.keys() if isinstance(models, dict) else models}. Will ignore."
```
Innixma, line 38: "The models types {missing_models} are not present in the model list specified by the user and will be ignored:"

common/tests/unittests/test_model_filter.py
```diff
@@ -0,0 +1,40 @@
...
+        (["dummy"], ["dummy"], None, ["dummy"]),
+        (["dummy"], ["foo"], None, []),
+        (["dummy", "foo"], ["foo"], None, ["foo"]),
+        ([], ["foo"], None, []),
+        (["dummy"], None, ["foo"], ["dummy"]),
+        ([], None, ["foo"], []),
+        (["dummy"], None, ["dummy"], []),
+        (["dummy", "foo"], None, ["dummy"], ["foo"]),
+    ],
+)
```
Innixma, line 33: It may be easier to have 1 line per argument for each test with proper use of whitespace, otherwise it is very hard to verify correctness by simply looking at this block.

common/src/autogluon/common/model_filter/model_filter.py
```diff
@@ -0,0 +1,106 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class ModelFilter:
+    """Class to filter models given user requirements"""
+
+    @staticmethod
+    def include_models(
+        models: Union[Dict[str, Any], List[str]], included_model_types: List[str]
```
Innixma, line 12: Why support both list and dict input for `models`? When is list ever used? Ditto for `exclude_models`

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -626,6 +626,12 @@ def fit(self,
                 See the `ag_args_ensemble` argument from "Advanced functionality: Custom AutoGluon model arguments" in the `hyperparameters` argument documentation for valid values.
                 Identical to specifying `ag_args_ensemble` parameter for all models in `hyperparameters`.
                 If a key in `ag_args_ensemble` is already specified for a model in `hyperparameters`, it will not be altered through this argument.
+            included_model_types : list, default = None
+                To only include listed model for training during `fit()`.
```
Innixma, line 630: listed model types

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -626,6 +626,12 @@ def fit(self,
...
+                Models that are listed in `included_model_types` but not in `hyperparameters` will be ignored.
+                Reference `hyperparameters` documentation for what models correspond to each value.
+                Useful when a only a subsample of model needs to be trained and the `hyperparameters` dictionary is difficult or time-consuming.
+                    Example: To include both 'KNN' and 'custom' models, specify `included_model_types=['KNN', 'custom']`.
```
Innixma, line 634: nit: Switch example to `['GBM', 'FASTAI']`, which is a more realistic setting a user would pick

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -626,6 +626,12 @@ def fit(self,
                 See the `ag_args_ensemble` argument from "Advanced functionality: Custom AutoGluon model arguments" in the `hyperparameters` argument documentation for valid values.
                 Identical to specifying `ag_args_ensemble` parameter for all models in `hyperparameters`.
                 If a key in `ag_args_ensemble` is already specified for a model in `hyperparameters`, it will not be altered through this argument.
+            included_model_types : list, default = None
+                To only include listed model for training during `fit()`.
+                Models that are listed in `included_model_types` but not in `hyperparameters` will be ignored.
+                Reference `hyperparameters` documentation for what models correspond to each value.
+                Useful when a only a subsample of model needs to be trained and the `hyperparameters` dictionary is difficult or time-consuming.
```
Innixma, line 633: subset of models

common/src/autogluon/common/model_filter/__init__.py
```diff
@@ -0,0 +1 @@
+from .model_filter import ModelFilter
```
Innixma, line 1: rename `model_filter.py` to `_model_filter.py` so that it cleans up import IDE hints if we only care about `ModelFilter`. and can get it here.

Innixma (approved):
LGTM, had some minor comments

common/src/autogluon/common/model_filter/model_filter.py
yinweisu: I had a look at TimeSeries implementation of `excluded_model_types` and it was using list. We could of course change how TimeSeries is implementing the logic but I feel it could be a valid use case while not adding too much difficulty in maintaining

gradientsky: +1 on list support: `include=['RF', 'NN']` is a legit use case.

gradientsky (approved):

PR #3240, opened 2023-05-25
Tabular: Raise regression pred_proba

### LINKED ISSUE #2685: [v0.8] Tabular: Throw Exception when calling predict_proba in regression
We should throw an exception when predict_proba is called in regression instead of returning the same output as predict.

PR to add FutureWarning: #2684

PR to raise exception (v0.8): #3240

Innixma:
*Issue #, if available:*

Resolves #2685 

*Description of changes:*

Update `TabularPredictor` to raise an AssertionError when `predict_proba` and `predict_proba_multi` are called when the problem type is `regression` or `quantile`. This aligns the logic with `MultimodalPredictor`.

### DIFF
--- a/core/src/autogluon/core/problem_type.py
+++ b/core/src/autogluon/core/problem_type.py
@@ -1,0 +1,71 @@
+from typing import Dict, List
+
+from .constants import BINARY, MULTICLASS, REGRESSION, SOFTCLASS, QUANTILE
+
+__all__ = ['problem_type_info']
+
+
+class ProblemType:
+    """Simple class that holds information on what a problem type is capable of doing"""
+    def __init__(self,
+                 can_predict: bool,
+                 can_predict_proba: bool,
+                 is_classification: bool):
+        self.can_predict = can_predict
+        self.can_predict_proba = can_predict_proba
+        self.is_classification = is_classification
+
+
+class ProblemTypeInfo:
+    """Class that stores all problem_type information, and can vend this information via the provided methods."""
+    def __init__(self, problem_type_dict: Dict[str, ProblemType]):
+        self.problem_type_dict = problem_type_dict
+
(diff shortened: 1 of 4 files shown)


core/src/autogluon/core/problem_type.py
```diff
@@ -0,0 +1,71 @@
...
+
+__all__ = ['problem_type_info']
+
+
+class ProblemType:
+    """Simple class that holds information on what a problem type is capable of doing"""
+    def __init__(self,
+                 can_predict: bool,
+                 can_predict_proba: bool,
+                 is_classification: bool):
```
gradientsky, line 13: nit: params docstring

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -1338,7 +1346,7 @@ def fit_pseudolabel(self, pseudo_data: pd.DataFrame, max_iter: int = 5, return_p
                 self.fit_weighted_ensemble()
 
             if return_pred_prob:
-                y_pred_proba = self.predict_proba(pseudo_data)
+                y_pred_proba = self.predict_proba(pseudo_data) if self.can_predict_proba else self.predict(pseudo_data)
```
gradientsky, line 1349: nit: split into if-else to have consistency vs other places?

gradientsky (approved):
LGTM, minor nit comments

tabular/src/autogluon/tabular/predictor/predictor.py
Innixma: Will keep for now, it will be unified once we apply black in future

core/src/autogluon/core/problem_type.py
Innixma: added

PR #3252, opened 2023-05-31
[timeseries] Minor bugfixes & improvements for local forecasting models

shchur:
*Description of changes:*
- Expose `use_fallback_model` as an optional hyperparameter for all local models (default `True`). When set to `False`, fallback model will be disabled, and any exception in the underlying model will propagate. This is important for testing - currently, we had one model that always failed because of a bug, but this wasn't caught by the CI because of the fallback model. 
- Fix typos in docstrings
- All local models are now trained using at most the last 2500 entries of each time series. This allows to significantly reduce the training time without degrading the accuracy:
image
image

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/local/abstract_local_model.py
+++ b/timeseries/src/autogluon/timeseries/models/local/abstract_local_model.py
@@ -28,8 +28,6 @@ class AbstractLocalModel(AbstractTimeSeriesModel):
     ----------
     allowed_local_model_args : List[str]
         Argument that can be passed to the underlying local model.
-    max_ts_length : int, optional
-        If not None, only the last ``max_ts_length`` time steps of each time series will be used to train the model.
     default_n_jobs : Union[int, float]
         Default number of CPU cores used to train models. If float, this fraction of CPU cores will be used.
     init_time_in_seconds : int
@@ -38,7 +36,6 @@ class AbstractLocalModel(AbstractTimeSeriesModel):
     """
 
     allowed_local_model_args: List[str] = []
-    max_ts_length: Optional[int] = None
     default_n_jobs: Union[int, float] = 0.5
     init_time_in_seconds: int = 0
 
@@ -71,6 +68,10 @@ class AbstractLocalModel(AbstractTimeSeriesModel):
             self.n_jobs = n_jobs
         else:
             raise ValueError(f"n_jobs must be a float between 0 and 1 or an integer (received n_jobs = {n_jobs})")
+        # Default values, potentially overridden inside _fit()
+        self.use_fallback_model = True
+        self.max_ts_length = 2500
+
         self._local_model_args: Dict[str, Any] = None
         self._seasonal_period: Optional[int] = None
         self.time_limit: Optional[float] = None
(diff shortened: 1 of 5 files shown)


tonyhoo:
Thank you for addressing the issue. I have a question: why doesn't our CI detect the local model's failure and subsequent use of the naive model, given that this occurs consistently? Does this suggest that the naive model's performance is on par with the local model, or is the issue specific to the dataset being used?

timeseries/src/autogluon/timeseries/models/local/abstract_local_model.py
```diff
@@ -71,6 +68,10 @@ def __init__(
             self.n_jobs = n_jobs
         else:
             raise ValueError(f"n_jobs must be a float between 0 and 1 or an integer (received n_jobs = {n_jobs})")
+        # Default values, potentially overridden inside _fit()
+        self.use_fallback_model = True
+        self.max_ts_length = 2500
```
tonyhoo, line 73: Shall we adjust it based on the freq? For example, for min data, 2500 is less than 2 days which will not be able to capture weekly trends/seasonality

timeseries/src/autogluon/timeseries/models/local/statsforecast.py
```diff
@@ -95,11 +95,14 @@ class AutoARIMAModel(AbstractStatsForecastModel):
         When set to None, seasonal_period will be inferred from the frequency of the training data. Can also be
         specified manually by providing an integer > 1.
         If seasonal_period (inferred or provided) is equal to 1, seasonality will be disabled.
-    n_jobs : int or float, default = -1
+    n_jobs : int or float, default = 0.5
         Number of CPU cores used to fit the models in parallel.
         When set to a float between 0.0 and 1.0, that fraction of available CPU cores is used.
         When set to a positive integer, that many cores are used.
         When set to -1, all CPU cores are used.
+    max_ts_length : int, default = 2500
```
tonyhoo, line 103: same here

timeseries/src/autogluon/timeseries/models/local/abstract_local_model.py
shchur: @tonyhoo Currently, the models that we use (`AutoETS`, `AutoARIMA`, `Theta`, `ETS`, `ARIMA`, `Naive`, `SeasonalNaive`) are anyway unable to capture multiple seasonalities - they only capture the seasonality at the `seasonal_period` that we provide. This is at most 24 * 60 = 1440 for minutely data, but usually [much smaller](https://github.com/autogluon/autogluon/blob/master/timeseries/src/autogluon/timeseries/utils/seasonality.py#L5) (<= 24). 

I think that the example you described would apply to models like [MSTL](https://nixtla.github.io/statsforecast/models.html#mstl-1) that consider multiple seasonalities, and I agree that we would need to increase the `max_ts_length` for such models if we add them.

@gradientsky I've moved the parameter override code from the [`_fit` method](https://github.com/autogluon/autogluon/pull/3252/files#diff-4a906aede4e4fa2d1d793970ec678d975a8522832fa38a54ab6ecd726a5a709dR88-R90) to `__init__`.

timeseries/tests/unittests/models/test_local.py
```diff
@@ -3,6 +3,7 @@
 import pandas as pd
 import pytest
```
shchur, line 4: @tonyhoo Regarding the tests for model performance: Currently, CI for time series does not include any regression tests. The reasoning was the models could be changing quite frequently, and keeping track of the performance ranges for individual models would be tedious. However, I agree that now we should probably be looking to add these tests, as the model set is becoming more stable. Do you think this it's fine if we add these tests after v0.8, or do you think it has higher priority and we should do this asap?

tonyhoo: We can add these post 0.8 and make sure it is incorporated into the benchmark and dashboard project

tonyhoo (approved):

abhishek-iitmadras:
Hi @shchur,
Where to find reliable sources for this Australian electricity demand dataset ?

shchur:
This one is taken from the Monash Forecasting Repository https://arxiv.org/abs/2105.06643.

PR #3259, opened 2023-06-01
Fix plot ensemble model on Windows

yinweisu:
*Issue #, if available:*
https://github.com/autogluon/autogluon/issues/1065

*Description of changes:*
* path attributes already being stored in the node is causing some weird issue to pygraphviz on Windows likely because it somehow treat `\\` separator on Windows as escape character.
* This PR removes all unneeded attributes from the nodes

### DIFF
--- a/tabular/src/autogluon/tabular/predictor/predictor.py
+++ b/tabular/src/autogluon/tabular/predictor/predictor.py
@@ -2994,7 +2994,6 @@ class TabularPredictor:
         A.node_attr.update(shape='rectangle')
 
         for node in A.iternodes():
-            node.attr['label'] = f"{node.name}\nVal score: {float(node.attr['val_score']):.4f}"
 
             if node.name == best_model_node:
                 node.attr['style'] = 'filled'
--- a/tabular/src/autogluon/tabular/predictor/predictor.py
+++ b/tabular/src/autogluon/tabular/predictor/predictor.py
@@ -2982,6 +2982,12 @@ class TabularPredictor:
         best_model_node = self.get_model_best()
 
         A = nx.nx_agraph.to_agraph(G)
+        
+        for node in A.iternodes():
+            # Remove unnecessary attributes
+            label = f"{node.name}\nVal score: {float(node.attr['val_score']):.4f}"
+            node.attr.clear()
+            node.attr['label'] = label
 
         A.graph_attr.update(rankdir='BT')
         A.node_attr.update(fontsize=10)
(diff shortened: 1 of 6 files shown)


tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -2982,6 +2982,12 @@ def plot_ensemble_model(self, prune_unused_nodes=True) -> str:
         best_model_node = self.get_model_best()
 
         A = nx.nx_agraph.to_agraph(G)
+        
+        for node in A.iternodes():
+            # Remove unnecessary attributes
+            label = f"{node.name}\nVal score: {float(node.attr['val_score']):.4f}"
+            node.attr.clear()
+            node.attr['label'] = label
```
Innixma, line 2990: Wouldn't this crash? Don't you need to keep `node.name`? Also, don't you need to remove L2997 which also sets `label` attribute using the currently deleted attributes?

tabular/tests/unittests/test_tabular.py
```diff
@@ -111,6 +111,7 @@ def test_advanced_functionality():
     savedir_predictor_original = savedir + 'predictor/'
     predictor: TabularPredictor = TabularPredictor(label=label, path=savedir_predictor_original).fit(train_data)
     leaderboard = predictor.leaderboard(data=test_data)
+    predictor.plot_ensemble_model()
```
Innixma, line 114: we need to install pygraphviz for unit tests to work, it isn't installed by default.

tabular/src/autogluon/tabular/predictor/predictor.py
yinweisu: `node.name` is not part of the attr I believe. And L2997 should be removed, that was a mistake.

tabular/tests/unittests/test_tabular.py
yinweisu: I see. it might be tricky to install this in our CI env as it requires some C++ library. I'll take a look

tabular/tests/unittests/test_tabular.py
```diff
@@ -219,6 +220,7 @@ def test_advanced_functionality():
 
     assert(predictor.get_model_full_dict() == dict())
     predictor.refit_full()
+    predictor.plot_ensemble_model()
```
Innixma, line 223: can check that the output file exists on disk, probably that is good enough for testing purposes

Innixma (approved):
LGTM assuming the tests pass on Windows

yinweisu:
Have tested on Windows. Will merge now

PR #3262, opened 2023-06-02
[multimodal] Upgrade preset and update the finetuning example for object detection

FANGAreNotGnu:
For benchmarking result for three new presets, check internal quip document "Autogluon Object Detection Benchmarking in 0.8.0".


multimodal/src/autogluon/multimodal/models/mmdet_image.py
zhiqiangdon, line 1: Why to change errors to warnings? Are there cases where class number or classes are missing but still work?

FANGAreNotGnu: Because in `test_preset.py` we did not provide `sample_data_path` and new DINO model does not have a default `CLASSES` value and the test will fail.

zhiqiangdon (approved):
LGTM.

PR #3266, opened 2023-06-02
[bug][multimodal] fix import error in object detection

FANGAreNotGnu:

### DIFF
--- a/multimodal/src/autogluon/multimodal/data/dataset_mmlab/multi_image_mix_dataset.py
+++ b/multimodal/src/autogluon/multimodal/data/dataset_mmlab/multi_image_mix_dataset.py
@@ -839,6 +839,4 @@ try:
             return translation_matrix
 
 except Exception as e:
-    print(e)
-    exit()
     MultiImageMixDataset = None
--- a/multimodal/src/autogluon/multimodal/data/dataset_mmlab/multi_image_mix_dataset.py
+++ b/multimodal/src/autogluon/multimodal/data/dataset_mmlab/multi_image_mix_dataset.py
@@ -17,15 +17,12 @@ logger = logging.getLogger(__name__)
 
 try:
     import mmcv
-
     from mmcv.transforms import BaseTransform
     from mmcv.transforms.utils import cache_randomness
-
+    from mmdet.structures.bbox import autocast_box_type
+    from mmdet.utils import log_img_scale
     from mmengine.dataset import BaseDataset
     from mmengine.config import Config as MMConfig
-
-    from mmdet.structures.bbox import autocast_box_type
-    from mmdet.utils import log_img_scale  # inline import to avoid mmdet uninstall error for other tasks
 
     class MultiImageMixDataset(torch.utils.data.Dataset):
         """
--- a/multimodal/src/autogluon/multimodal/data/dataset_mmlab/multi_image_mix_dataset.py
+++ b/multimodal/src/autogluon/multimodal/data/dataset_mmlab/multi_image_mix_dataset.py
@@ -17,835 +17,828 @@ logger = logging.getLogger(__name__)
 
 try:
     import mmcv
-    from mmcv.utils import Config as MMCVConfig
-except:
-    MMCVConfig = None
 
-try:
     from mmcv.transforms import BaseTransform
     from mmcv.transforms.utils import cache_randomness
-except:
-    BaseTransform = None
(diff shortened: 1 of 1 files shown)


Innixma:
@FANGAreNotGnu Why does an import error fix PR have over 700 lines of code change?

FANGAreNotGnu:
> @FANGAreNotGnu Why does an import error fix PR have over 700 lines of code change?

I wrapped this class in a try except block.

Some mmlab packages (mmcv, mmdet, and mmengine) are not used in most problem types and thus not installed by default. But some mmlab files are used as inherited class or decoration functions here, and have to be wrapped otherwise it will cause error during import in any automm tasks.

multimodal/src/autogluon/multimodal/data/dataset_mmlab/multi_image_mix_dataset.py
```diff
@@ -17,834 +17,823 @@
...
-    def __init__(
-        self,
-        data: pd.DataFrame,
-        preprocessor: List[MultiModalFeaturePreprocessor],
-        processors: List[dict],
-        model_config: MMCVConfig,
-        id_mappings: Optional[Union[Dict[str, Dict], Dict[str, pd.Series]]] = None,
-        is_training: bool = False,
-    ):
+    class MultiImageMixDataset(torch.utils.data.Dataset):
```
Innixma, line 27: Instead of putting the entire code of the file in a try/except block, create a new file `_multi_image_mix_dataset.py` with the original code and no try/except. Then in this file, simply import the classes from that file in a try/except.

```python3
try:
    from ._multi_image_mix_dataset import MultiImageMixDataset
except ImportError as e:
    MultiImageMixDataset = None
```

This will be much cleaner.

multimodal/src/autogluon/multimodal/data/dataset_mmlab/multi_image_mix_dataset.py
```diff
@@ -17,834 +17,823 @@
...
+                [[1, np.tan(x_radian), 0.0], [np.tan(y_radian), 1, 0.0], [0.0, 0.0, 1.0]], dtype=np.float32
+            )
+            return shear_matrix
+
+        @staticmethod
+        def _get_translation_matrix(x: float, y: float) -> np.ndarray:
+            translation_matrix = np.array([[1, 0.0, x], [0.0, 1, y], [0.0, 0.0, 1.0]], dtype=np.float32)
+            return translation_matrix
+
+except Exception as e:
```
Innixma, line 838: Why use `Exception` when we can be more specific such as `ImportError`?

multimodal/src/autogluon/multimodal/data/dataset_mmlab/multi_image_mix_dataset.py
FANGAreNotGnu: Oh sure, or I can also add the try block in __init__?

multimodal/src/autogluon/multimodal/data/dataset_mmlab/multi_image_mix_dataset.py
```diff
@@ -17,834 +17,823 @@
...
+    MultiImageMixDataset = None
```
Innixma, line 839: Why only `MultiImageMixDataset = None`?

What about `Mosaic`, `MixUp`, `RandomAffine`?

multimodal/src/autogluon/multimodal/data/dataset_mmlab/multi_image_mix_dataset.py
Innixma: If you want to add to init, you can do that, but the file with the code source should not have the try/except block for the class definitions

multimodal/src/autogluon/multimodal/data/dataset_mmlab/multi_image_mix_dataset.py
FANGAreNotGnu: Addressed in #3277

multimodal/src/autogluon/multimodal/data/dataset_mmlab/multi_image_mix_dataset.py
FANGAreNotGnu: Addressed in #3277

PR #3268, opened 2023-06-03
Update CatBoost to 1.2

Innixma:
*Issue #, if available:*

*Description of changes:*
Update CatBoost to 1.2

- CatBoost 1.2 has 8% faster training speed in multiclass classification with the log loss metric.
- Will plan to benchmark after merge on mainline during code freeze instead of benchmarking this PR individually, as it will be easy to isolate the root cause of regressions in case this PR is the cause.

TODO:

- [x] Run the CI on Windows/MacOS

### DIFF
--- a/tabular/setup.py
+++ b/tabular/setup.py
@@ -34,7 +34,10 @@ extras_require = {
         'lightgbm>=3.3,<3.4',
     ],
     'catboost': [
-        'catboost>=1.1,<1.2',
+        # CatBoost wheel build is not working correctly on darwin for CatBoost 1.2 on Python 3.8, so use old version in this case.
+        # https://github.com/autogluon/autogluon/pull/3190#issuecomment-1540599280
+        'catboost>=1.1,<1.2 ; python_version <= "3.8" and sys_platform == "darwin"',
+        'catboost>=1.1,<1.3',
     ],
     # FIXME: Debug why xgboost 1.6 has 4x+ slower inference on multiclass datasets compared to 1.4
     #  It is possibly only present on MacOS, haven't tested linux.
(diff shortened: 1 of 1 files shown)


Innixma:
@shchur Unit test failed in TimeSeries. Looks to be potentially unrelated to this PR?

shchur:
@Innixma rebasing should fix this, the bugfix was merged in #3255

Innixma:
@shchur My branch already has that PR merged in, so the CI issue happened post-fix.

shchur:
Should be fixed by #3275

tabular/setup.py
```diff
@@ -34,7 +34,10 @@
         'lightgbm>=3.3,<3.4',
     ],
     'catboost': [
-        'catboost>=1.1,<1.2',
+        # CatBoost wheel build is not working correctly on darwin for CatBoost 1.2 on Python 3.8, so use old version in this case.
+        # https://github.com/autogluon/autogluon/pull/3190#issuecomment-1540599280
+        'catboost>=1.1,<1.2 ; python_version <= "3.8" and sys_platform == "darwin"',
```
yinweisu, line 39: I just did some testing on all python 3.8 - 3.10, and all of them fail to install catboost 1.2.
Let's just not do it for macos no matter which python version it is

yinweisu (changes requested):

tabular/setup.py
Innixma: Thanks for checking! I've updated the PR so that it should avoid installing 1.2 on MacOS, if you can give another try or we can run platform tests

yinweisu (approved):
LGTM

Issue #2806, opened 2023-02-01
[Tabular] Add TabPFN model


Innixma:
Add TabPFN model to Tabular.

KirderfEman:
Interesting, and it uses some FE as well in the model.

> Do not preprocess inputs to TabPFN. TabPFN pre-processes inputs internally. It applies a z-score normalization (x-train_x.mean()/train_x.std()) per feature (fitted on the training set) and log-scales outliers [heuristically](https://github.com/automl/TabPFN/blob/f7402ec1916aa78d953574daf95508045af5953e/tabpfn/utils.py#L201). Finally, TabPFN applies a PowerTransform to all features for every second ensemble member. Pre-processing is important for the TabPFN to make sure that the real-world dataset lies in the distribution of the synthetic datasets seen during training. So to get the best results, do not apply a PowerTransformation to the inputs.

> TabPFN expects scalar values only (you need to encode categoricals as integers e.g. with [OrdinalEncoder](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.OrdinalEncoder.html#sklearn.preprocessing.OrdinalEncoder)). It works best on data that does not contain any categorical or NaN data (see [Appendix B.1](https://arxiv.org/abs/2207.01848)).

> TabPFN ensembles multiple input encodings per default. It feeds different index rotations of the features and labels to the model per ensemble member. You can control the ensembling with TabPFNClassifier(...,N_ensemble_configurations=?)

> TabPFN does not use any statistics from the test set. That means predicting each test example one-by-one will yield the same result as feeding the whole test set together.

> TabPFN is differentiable in principle, only the pre-processing is not and relies on numpy.

Innixma:
Resolved by #3270

PR #3270, opened 2023-06-04
Add TabPFN

Innixma:
*Issue #, if available:*
#2806 

*Description of changes:*
- Adds the [TabPFN model](https://github.com/automl/TabPFN) to AutoGluon.

TODO:
- [x] Run AutoMLBenchmark to verify correctness & performance
- [x] Verify implementation with TabPFN authors


Innixma:
Hello @noahho and @SamuelGabriel, I am looking to add TabPFN to AutoGluon for the upcoming v0.8 release on June 13th!

If you had time, it would be great if you could give a quick look at the PR and mention any suggestions/changes you'd recommend.

In particular, I'd be curious your thoughts on the `sample_rows` value and how to best handle missing values.

tabular/src/autogluon/tabular/models/tabpfn/tabpfn_model.py
```diff
@@ -0,0 +1,142 @@
...
+            X=X,
+            y=y,
+            problem_type=self.problem_type,
+            test_size=num_rows_to_drop,
+            random_state=random_state,
+            min_cls_count_train=1,
+        )
+        return X, y
+
+    # TODO: Should we fillna 0? what about -1 or mean? Does TabPFN automatically fill missing values?
```
SamuelGabriel, line 71: TabPFNClassifier can handle nans and fills them more or less with the mean of the feature. So, I think it is best to just hand the nans over.

tabular/src/autogluon/tabular/models/tabpfn/tabpfn_model.py
```diff
@@ -0,0 +1,142 @@
...
+        max_features = ag_params.get('max_features')
+        max_classes = ag_params.get('max_classes')
+        if max_classes is not None and self.num_classes > max_classes:
+            # TODO: Move to earlier stage when problem_type is checked
+            raise AssertionError(f'Max allowed classes for the model is {max_classes}, '
+                                 f'but found {self.num_classes} classes.')
+
+        # TODO: Make sample_rows generic
+        if sample_rows is not None and len(X) > sample_rows:
+            X, y = self._subsample_train(X=X, y=y, num_rows=sample_rows)
```
SamuelGabriel, line 44: We have a bit of data on this. The best strategy is to something like have multiple trees with a TabPFN in each leaf, but this is not too much worse in our experience.

SamuelGabriel:
Hey :) I have left two comments on your points. It looks good in general, I think, besides the 0 imputation strategy.

Generally I think it is important to know that the main costs of TabPFN are incurred with loading the model to (GPU) memory at the first construction of the TabPFNClassifer and in the fit function, where the full forward pass happens.

Best,
Sam

tabular/src/autogluon/tabular/models/tabpfn/tabpfn_model.py
Innixma: Great! I did a quick test and it appears to work without doing the `fillna`, with comparable performance, so I've updated the PR to not fillna and instead pass the NaNs to TabPFN directly.

tabular/src/autogluon/tabular/models/tabpfn/tabpfn_model.py
Innixma: That makes sense, I had a similar assumption as to what would probably work best.

For now, I've kept as is since doing the `TabPFN Forest` approach is very similar to using bagging in AutoGluon on `TabPFN`. I'd expect `TabPFN_Forest` to be slightly better, but will keep things simple for now. If you do have data that shows a compelling performance improvement I'd definitely take another look.

If `TabPFN` can do this logic internally, that could be a very nice addition. For example, when `N_ensemble_configurations=10`, each TabPFN in N could use a different sample from the train data input, which should maximize diversity.

PR #3272, opened 2023-06-05
[timeseries] Fix high memory usage of the RecursiveTabular model

shchur:
*Description of changes:*
- Currently, the `RecursiveTabular` model based on the `mlforecast` library can result in extremely high memory usage for large datasets (e.g., model crashes with OOM for datasets with >50M rows like `M5`, `Corporacion Favorita`, `London Smart Meters` on `m5.4xlarge` instances with 64 GB of RAM). This PR fixes this problem by selecting a subset of the time series to ensure that the peak memory usage does not exceed 80% of the available memory.
- The approach is based on fitting a linear model of the form `a * num_rows * num_features + b * num_rows + c` to predict the peak memory usage. I fit the model using memory usage statistics for 6 datasets with 10K-10M rows & different choices for `num_features`. The model was validated on on the M5 dataset with >50M rows:
  - Predicted peak memory utilization: `81354.0 MB`
  - Actual peak memory utilization: `79258.9 MB`
  - When setting the max allowed memory utilization to `48029.7 MB` & selecting items based on the heuristic, the resulting peak memory usage was `48122.7 MB`.

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
+++ b/timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
@@ -135,7 +135,7 @@ class RecursiveTabularModel(AbstractTimeSeriesModel):
 
         lags = model_params.get("lags")
         if lags is None:
-            lags = get_lags_for_frequency(self.freq, lag_ub=200)
+            lags = get_lags_for_frequency(self.freq)
 
         date_features = model_params.get("date_features")
         if date_features is None:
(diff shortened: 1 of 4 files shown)


timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
```diff
@@ -333,3 +333,38 @@ def predict(
...
+@dataclass(frozen=True)
+class MLFMemoryUsage:
+    """Estimated peak memory usage of MLForecast.preprocess (in MB) using the following linear model:
+
+    peak_mem_usage = a * num_rows * num_features + b * num_rows * c
+    """
+
+    a = 3.16e-5
+    b = 1.4e-4
+    c = 661
```
tonyhoo, line 347: Can we add test cases to ensure the longevity of this linear model despite updates to our model over time?

PR #3281, opened 2023-06-06
[timeseries] Update presets & bump dependencies for v0.8.0

shchur:
*Description of changes:*
- Update presets to include the new models
- Update description of the presets in `TimeSeriesPredictor` API documentation
- Bump dependency versions, add stricter version caps to protect us from potential breaking changes introduced by dependencies
- Rename `local_only` hyperparameters setting to `fast_training` for consistency


timeseries/setup.py
```diff
@@ -27,13 +27,13 @@
     "numpy",  # version range defined in `core/_setup_utils.py`
     "scipy",  # version range defined in `core/_setup_utils.py`
     "pandas",  # version range defined in `core/_setup_utils.py`
-    "statsmodels>=0.13.0,<0.14",
+    "statsmodels>=0.13.0,<0.15",
```
tonyhoo, line 30: any reason why we lower the upper bound? Any issue with 0.14?

shchur: The upper bound is increased here - I did some testing and `statsmodels==0.14.0` does not affect any of the functionality that we depend on (compared to `v0.13.*`).

I could not increase the upper bound for `statsforecast` to `v1.5.0` though because of a dependency clash with multimodal:
```
[2023-06-07T15:41:18.580Z] ERROR: Cannot install autogluon.multimodal, fugue and fugue-sql-antlr==0.1.6 because these package versions have conflicting dependencies.
[2023-06-07T15:41:18.581Z] The conflict is caused by:
[2023-06-07T15:41:18.581Z]     omegaconf 2.1.1 depends on antlr4-python3-runtime==4.8
[2023-06-07T15:41:18.581Z]     fugue-sql-antlr 0.1.6 depends on antlr4-python3-runtime<4.12 and >=4.11.1
```
This is not critical now but should be investigated in the future. I added a comment for that.

tonyhoo (approved):

PR #3283, opened 2023-06-06
Cleanup diskspace for docker image build

yinweisu:
*Issue #, if available:*

*Description of changes:*
* Nightly image building is failing recently because of the increase of the image size leading to out of disk space for github runners. Adding a step to cleanup the machine to squeeze more space.
* Has been tested [here](https://github.com/yinweisu/autogluon/actions/runs/5191861673)
* To be noticed, if our image size continue to grow, it's likely that we'll need to move away from github hosted runners...

### DIFF
--- a/.github/workflows/build_latest_image.yml
+++ b/.github/workflows/build_latest_image.yml
@@ -25,7 +25,7 @@ jobs:
         sudo swapoff -a
         sudo rm -f /swapfile
         sudo apt clean
-        docker rmi $(docker image ls -aq)
+        docker system prune -af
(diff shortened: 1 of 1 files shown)


tonyhoo (commented):
Looks good to me, a more concerning issue is the root cause of ever growing image size

yinweisu:
> Looks good to me, a more concerning issue is the root cause of ever growing image size

I think we are just keep adding more dependencies, mostly in multimodal

gradientsky (approved):

PR #3285, opened 2023-06-06
Fix platform tests

yinweisu:
*Issue #, if available:*

*Description of changes:*
* install fail because of setup_mmcv.sh not being found. Change to use install_all.sh now as we don't need mxnet anymore
* fix tempfile path. `str(tempfile)` doesn't give a string but still the object itself for some reason causing issue on Windows. Tried to use `tempfile.name` instead but the underlying issue with `TemporaryFile` is that it opens the file automatically, while our saving logic tries to open it again. This will fail on Windows. Change to create a tempdir instead

### DIFF
--- a/common/tests/unittests/test_hash_pandas_df.py
+++ b/common/tests/unittests/test_hash_pandas_df.py
@@ -1,6 +1,6 @@
 import numpy as np
 import pandas as pd
-from tempfile import TemporaryFile
+import tempfile
 
 from autogluon.common.loaders import load_pkl
 from autogluon.common.savers import save_pkl
@@ -19,9 +19,10 @@ def _get_pandas_df(num_rows=20):
 
 def test_when_df_saved_and_loaded_from_disk_then_hash_is_unchanged():
     df = _get_pandas_df()
-    with TemporaryFile() as temp_file:
-        save_pkl.save(temp_file.name, df)
-        loaded_df = load_pkl.load(str(temp_file))
+    with tempfile.TemporaryDirectory() as _:
+        tem_pfile = "temp_file.pkl"
+        save_pkl.save(tem_pfile, df)
+        loaded_df = load_pkl.load(tem_pfile)
     assert hash_pandas_df(df) == hash_pandas_df(loaded_df)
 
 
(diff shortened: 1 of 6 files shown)


common/tests/unittests/test_hash_pandas_df.py
```diff
@@ -19,9 +19,10 @@ def _get_pandas_df(num_rows=20):
 
 def test_when_df_saved_and_loaded_from_disk_then_hash_is_unchanged():
     df = _get_pandas_df()
-    with TemporaryFile() as temp_file:
-        save_pkl.save(str(temp_file), df)
-        loaded_df = load_pkl.load(str(temp_file))
+    with tempfile.TemporaryDirectory() as _:
+        tem_pfile = "temp_file.pkl"
+        save_pkl.save(tem_pfile, df)
+        loaded_df = load_pkl.load(tem_pfile)
```
shchur, line 25: It seems that the current code saves `temp_file.pkl` outside the `TemporaryDirectory`, is this the case?

Also, probably typo in the file name `tem_pfile` -> `temp_file`?

Would the following code work? (also requires adding `from pathlib import Path` in the header)
```suggestion
    with tempfile.TemporaryDirectory() as temp_dir:
        temp_file = str(Path(temp_dir.name) / "temp_file.pkl")
        save_pkl.save(temp_file, df)
        loaded_df = load_pkl.load(temp_file)
```

common/tests/unittests/test_hash_pandas_df.py
```diff
@@ -1,6 +1,6 @@
 import numpy as np
 import pandas as pd
-from tempfile import TemporaryFile
+import tempfile
```
shchur, line 3:
```suggestion
import tempfile
from pathlib import Path
```

shchur (approved):
Thank you for looking into this! 

Added some comments/suggestions, but feel free to merge if my suggestion doesn't work.

common/tests/unittests/test_hash_pandas_df.py
yinweisu: Nice suggestion. Your code doesn't work out of the box. Made some changes

PR #3286, opened 2023-06-06
Add enable_categorical=True support to XGBoost

### LINKED ISSUE #2429: Test XGBoost Categorical Support
XGBoost recently added native categorical feature support.

> XGBoost now [supports the "optimal partitioning" algorithm used by LightGBM](https://xgboost.readthedocs.io/en/stable/tutorials/categorical.html#optimal-partitioning) as opposed to requiring one-hot-encoding.

We should test how it performs compared to the one-hot-encoding we currently use.

Innixma:
*Issue #, if available:*
Resolves #2429 

*Description of changes:*
- Added support for XGBoost's new (as of XGBoost 1.6) support for categorical features via `enable_categorical=True`.
- Currently not enabled by default, will benchmark to consider enabling by default.

### DIFF
--- a/tabular/src/autogluon/tabular/models/xgboost/xgboost_model.py
+++ b/tabular/src/autogluon/tabular/models/xgboost/xgboost_model.py
@@ -57,11 +57,10 @@ class XGBoostModel(AbstractModel):
 
         if is_train:
             if self._ohe:
-                if self._ohe_generator is None:
-                    self._ohe_generator = xgboost_utils.OheFeatureGenerator(max_levels=max_category_levels)
+                self._ohe_generator = xgboost_utils.OheFeatureGenerator(max_levels=max_category_levels)
                 self._ohe_generator.fit(X)
 
-        if self._ohe_generator is not None:
+        if self._ohe:
             X = self._ohe_generator.transform(X)
 
         return X
--- a/tabular/src/autogluon/tabular/models/xgboost/xgboost_model.py
+++ b/tabular/src/autogluon/tabular/models/xgboost/xgboost_model.py
@@ -25,6 +25,7 @@ class XGBoostModel(AbstractModel):
     """
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
+        self._ohe: bool = True
         self._ohe_generator = None
         self._xgb_model_type = None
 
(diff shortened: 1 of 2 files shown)


tabular/src/autogluon/tabular/models/xgboost/xgboost_model.py
```diff
@@ -54,13 +55,14 @@ def get_eval_metric(self):
     def _preprocess(self, X, is_train=False, max_category_levels=None, **kwargs):
         X = super()._preprocess(X=X, **kwargs)
 
-        if self._ohe_generator is None:
-            self._ohe_generator = xgboost_utils.OheFeatureGenerator(max_levels=max_category_levels)
-
         if is_train:
-            self._ohe_generator.fit(X)
+            if self._ohe:
+                if self._ohe_generator is None:
```
gradientsky, line 60: If the model is re-trained (i.e. on the full data), wouldn't this reuse the generator. I think it's better to always create a new one if `is_train and self._ohe`.

tabular/src/autogluon/tabular/models/xgboost/xgboost_model.py
```diff
@@ -83,6 +85,13 @@ def _fit(self,
         if num_cpus:
             params['n_jobs'] = num_cpus
         max_category_levels = params.pop('proc.max_category_levels', 100)
+        enable_categorical = params.get('enable_categorical', False)
+        if enable_categorical:
+            """Skip one-hot-encoding and pass categoricals directly to XGBoost"""
+            self._ohe = False
+        else:
+            """One-hot-encode categorical features"""
+            self._ohe = True
```
gradientsky, line 95: nit:
```python
self._ohe = not enable_categorical
```

tabular/src/autogluon/tabular/models/xgboost/xgboost_model.py
Innixma: No, re-training would be a new class instance, it does not cause issues

tabular/src/autogluon/tabular/models/xgboost/xgboost_model.py
Innixma: Good point, but will keep as is since I have docstrings to describe what True and False mean inline.

Innixma:
Note: CI failures are from AutoMM, unrelated to this PR. All tabular tests and CI builds succeeded.

gradientsky (approved):

PR #3288, opened 2023-06-07
Tabular: Fix crash when save path is absolute

Innixma:
*Issue #, if available:*

*Description of changes:*
- If using an absolute path in predictor init, will crash upon load.

This PR fixes the obvious error so that the below script works, however the current fix is not a good idea long-term, as I have no idea if there will be edge-cases where it fails. I also don't know if it will work when using multiple disk drives in a single process.

The root of the issue is that `StackerEnsembleModel` does something it shouldn't: It has the absolute paths of both itself and its base models stored as variables. Therefore, in order to figure out where the base models are upon loading in a new location, it isn't sufficient to look at the old absolute paths, instead they need to be regenerated based on 1. the old absolute path, 2. the new path.

For example:

```
Model 1: `/abc/def/`
Model 2: `./foo/bar/`


# Model 1 where model 2 is a base model:
self.path_root = `/abc/def/`
self.base_model_paths_dict = {
    'model_2': './foo/bar/'
}
```

When loaded from a new location, this can cause issues. It worked in the past except for cross-os loading.

```
new_path = './new/location/'
new_model_2_path = ?????  # <------- This is the problem, hard to figure out the correct path
```

The solution is to either

1. Delete StackerEnsembleModel and have `trainer` deal with this (Ideal, but significant work)
2. Convert absolute paths to relative paths, or otherwise hack things to avoid the breaking situation (What I've done in this PR, but I don't like it. It **probably** fixes things, but hard to know for sure in all cases).
3. Bandaid solution by try/except on the path conversion: If can't convert because absolute, assume same machine and continue (will crash if cross-OS and absolute path was used during fit).

Reproducible Example:
```python3
from autogluon.tabular import TabularPredictor, TabularDataset


if __name__ == '__main__':
    data_root = 'https://autogluon.s3.amazonaws.com/datasets/Inc/'
    train_data = TabularDataset(data_root + 'train.csv')
    train_data = train_data.sample(500)
    test_data = TabularDataset(data_root + 'test.csv')

    predictor = TabularPredictor(
        label='class',
        path='/tmp/tmpfe_o0p7g/',
    ).fit(
        train_data=train_data,
        hyperparameters={'DUMMY': {}},
        fit_weighted_ensemble=False,
        num_bag_folds=2,
        num_bag_sets=1,
    )
    predictor.persist_models('best')
    predictor.leaderboard(test_data)
```

Exception:
```
    assert not PathConverter._is_absolute(path), "It is ambiguous on how to convert an absolute path. Please provide a relative path instead"
AssertionError: It is ambiguous on how to convert an absolute path. Please provide a relative path instead
```

### DIFF
--- a/common/src/autogluon/common/utils/path_converter.py
+++ b/common/src/autogluon/common/utils/path_converter.py
@@ -1,5 +1,6 @@
+import os
 import platform
-from pathlib import PurePosixPath, PureWindowsPath
+from pathlib import Path, PurePosixPath, PureWindowsPath
 
 
 class PathConverter:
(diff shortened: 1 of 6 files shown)


Creative-Ataraxia:
this bug is still happening as of now, if you run this code from a `.ipynb` on a different drive `d:`
```
predictor = TabularPredictor(
    label=label,
    path='I:/Checkpoints/AutogluonModels/',
).fit(
    train_data,
    presets="high_quality"
)
```

you will get:
`autogluon\core\models\ensemble\stacker_ensemble_model.py:86`
`self.base_model_paths_dict = {key: os.path.relpath(val, self.path) for key, val in base_model_paths_dict.items()}`
`File <frozen ntpath>:791, in relpath(path, start)`
`ValueError: path is on mount 'd:', start on mount 'I:'`

Creative-Ataraxia:
> this bug is still happening as of now, if you run this code from a `.ipynb` on a different drive `d:`
> 
> ```
> predictor = TabularPredictor(
>     label=label,
>     path='I:/Checkpoints/AutogluonModels/',
> ).fit(
>     train_data,
>     presets="high_quality"
> )
> ```
> 
> you will get: `autogluon\core\models\ensemble\stacker_ensemble_model.py:86` `self.base_model_paths_dict = {key: os.path.relpath(val, self.path) for key, val in base_model_paths_dict.items()}` `File <frozen ntpath>:791, in relpath(path, start)` `ValueError: path is on mount 'd:', start on mount 'I:'`


can we maybe not use `relpath()`? and simply replace that line with `self.base_model_paths_dict = {key: val for key, val in base_model_paths_dict.items()}`?

PR #3289, opened 2023-06-07
[AutoMM] Support model parameter number statistics

zhiqiangdon:
*Issue #, if available:*

*Description of changes:*
Support calling `predictor.total_parameters `, `predictor.trainable_parameters `, and `predictor.model_size `.

### DIFF
--- a/multimodal/src/autogluon/multimodal/predictor.py
+++ b/multimodal/src/autogluon/multimodal/predictor.py
@@ -147,6 +147,7 @@ from .utils import (
     infer_scarcity_mode_by_data_size,
     init_df_preprocessor,
     init_pretrained,
+    is_lazy_weight_tensor,
     list_timm_models,
     load_text_tokenizers,
     logits_to_prob,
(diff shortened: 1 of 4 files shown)


multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -508,6 +509,26 @@ def column_types(self):
...
+    def trainable_parameters(self) -> int:
+        return sum(
+            p.numel() if not is_lazy_weight_tensor(p) else 0 for p in self._model.parameters() if p.requires_grad
+        )
+
+    @property
+    def model_size(self) -> float:
+        precision_to_bits = {64: 64, 32: 32, 16: 16, "bf16": 16}
+        if self._config is not None:
+            precision = precision_to_bits.get(self._config.env.precision, 16)
```
tonyhoo, line 526: wondering how mixed precision can be handled here? Shall we use `total_size_in_bytes = sum(p.numel() * p.element_size() for p in  self._model.parameters())` to address it?

multimodal/src/autogluon/multimodal/utils/model.py
```diff
@@ -556,3 +557,15 @@ def modify_duplicate_model_names(
 
 def list_timm_models(pretrained=True):
     return timm.list_models(pretrained=pretrained)
+
+
+def is_lazy_weight_tensor(p: Tensor) -> bool:
+    from torch.nn.parameter import UninitializedParameter
+
+    if isinstance(p, UninitializedParameter):
+        warnings.warn(
```
tonyhoo, line 566: Wondering how frequent this gonna happen? If it is the case, shall we just run one dummy forward pass to make these layers initialized and then calculate?

zhiqiangdon: This refers to pytorch lightning: https://github.com/Lightning-AI/lightning/blob/984f49f7195ddc67e961c7c498ee6e19fc0cecb5/src/lightning/pytorch/utilities/model_summary/model_summary.py#L236
In some cases, parameters are uninitialized.

tonyhoo: ok, as long as we are consistent with lightning, shall be good

multimodal/src/autogluon/multimodal/predictor.py
zhiqiangdon: Good suggestion. Updated.

FANGAreNotGnu (approved):
LGTM

suzhoum (approved):
LGTM!

PR #3293, opened 2023-06-08
Fix path logic

yinweisu:
*Issue #, if available:*

*Description of changes:*
* https://github.com/autogluon/autogluon/pull/3288 introduce logic for saving relative paths for models. This logic affects `timeseries` too and `Trainer` in `timeseries` is not being updated with the relative path logic. This could be an issue in some edge cases, for example:
```
`current_working_dir` is `/home/ubuntu/user/test`
`path` specified to predictor is `/home/ubuntu/timeseries`

`model_paths` after training will contain `../timeseries/model/XXX`
while the `path` being saved in trainer is `/home/ubuntu/timeseries`
Then when we try to find common prefix to reset the context, it fails as there's no common prefix
```


Innixma (approved):
LGTM!

PR #3302, opened 2023-06-12
Fix log to file Windows tests

yinweisu:
*Issue #, if available:*

*Description of changes:*
* tempfile behaves weird on Windows: if you create a file within the TemporaryDirectory, the directory will fail to clean up itself. Minimal reproducible example
```python
def test_foo():
    with tempfile.TemporaryDirectory() as tempdir:
        temp = os.path.join(tempdir, "abc")
        with open(temp, "r") as f:
            pass
```
pytest running test above will fail^^^
* Removed usage of tempfile

### DIFF
--- a/tabular/src/autogluon/tabular/predictor/predictor.py
+++ b/tabular/src/autogluon/tabular/predictor/predictor.py
@@ -3315,6 +3315,7 @@ class TabularPredictor:
             assert predictor_path is not None, "Please either provide `predictor_path` or `log_file_path` to load the log file"
             file_path = os.path.join(predictor_path, "logs", cls._predictor_log_file_name)
         assert os.path.isfile(file_path), f"Log file does not exist at {file_path}"
+        lines = []
         with open(file_path, "r") as f:
             lines = f.readlines()
         return lines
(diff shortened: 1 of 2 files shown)


tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -3315,6 +3315,7 @@ def load_log(cls, predictor_path: str = None, log_file_path: Optional[str] = Non
             assert predictor_path is not None, "Please either provide `predictor_path` or `log_file_path` to load the log file"
             file_path = os.path.join(predictor_path, "logs", cls._predictor_log_file_name)
         assert os.path.isfile(file_path), f"Log file does not exist at {file_path}"
+        lines = []
```
Innixma, line 3318: why is this necessary?

Innixma (approved):
LGTM assuming tests pass, had 1 comment above

tabular/src/autogluon/tabular/predictor/predictor.py
yinweisu: Just found that variable is unbounded, so added a default value. Though it probably doesn’t matter cause it will not be assigned a value only when we failed to read the file, and that would raise an error already

PR #3315, opened 2023-06-14
Fix NaN warning in np.array(X)

Innixma:
*Issue #, if available:*

*Description of changes:*
- Fix NaN warning in np.array(X)


yinweisu (approved):
Approving assuming CI and benchmark passes

PR #3331, opened 2023-06-20
Move PyMuPDF to optional

Innixma:
*Issue #, if available:*

*Description of changes:*
- Move PyMuPDF to optional
Since PyMuPDF isn't available in Conda, better to move to an optional dependency as it is only used in a specific scenario for PDF file reading.

TODO:

- [ ] Merge into master
- [ ] Merge diff into branch 0.8.0 -> use this for conda release, don't use master

### DIFF
--- a/multimodal/setup.py
+++ b/multimodal/setup.py
@@ -75,10 +75,10 @@ extras_require = {
     ]
 }
 
-for test_package in ['PyMuPDF']:
+for test_package in ["PyMuPDF"]:
     tests_require += extras_require[test_package]
 
-extras_require['tests'] = tests_require
+extras_require["tests"] = tests_require
 
 
 if __name__ == "__main__":
--- a/multimodal/setup.py
+++ b/multimodal/setup.py
@@ -55,22 +55,30 @@ install_requires = [
     "jinja2>=3.0.3,<3.2",
     "tensorboard>=2.9,<3",
     "pytesseract>=0.3.9,<0.3.11",
-    "PyMuPDF<=1.21.1",
 ]
 
 install_requires = ag.get_dependency_version_ranges(install_requires)
 
+tests_require = [
+    "black>=22.3,<23.0",
+    "isort>=5.10",
+    "datasets>=2.3.2,<=2.3.2",
+    "onnx>=1.13.0,<1.14.0",
+    "onnxruntime>=1.13.0,<1.14.0;platform_system=='Darwin'",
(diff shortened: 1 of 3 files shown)


multimodal/setup.py
```diff
@@ -55,23 +55,31 @@
...
-        "tensorrt>=8.5.3.1,<8.5.4;platform_system=='Linux'",
+    "PyMuPDF": [
+        "PyMuPDF<=1.21.1",
     ]
 }
 
+for test_package in ["PyMuPDF"]:
+    tests_require += extras_require[test_package]
+
+extras_require["tests"] = tests_require
```
zhiqiangdon, line 81: What's the benefits of adding `tests_require `? It looks more complicated than using only `extras_require `.

zhiqiangdon: Adding `PyMuPDF ` into `tests_require` is kind of confusing since it's not only for test.

multimodal/setup.py
```diff
@@ -55,23 +55,31 @@
...
 install_requires = ag.get_dependency_version_ranges(install_requires)
 
+tests_require = [
+    "black>=22.3,<23.0",
+    "isort>=5.10",
+    "datasets>=2.3.2,<=2.3.2",
+    "onnx>=1.13.0,<1.14.0",
+    "onnxruntime>=1.13.0,<1.14.0;platform_system=='Darwin'",
+    "onnxruntime-gpu>=1.13.0,<1.14.0;platform_system!='Darwin'",
+    "tensorrt>=8.5.3.1,<8.5.4;platform_system=='Linux'",
```
zhiqiangdon, line 69: `onnx` and `tensorrt` are not only for testing.

multimodal/setup.py
Innixma: it is required for tests to pass, therefore I've put it in `tests_require`. Do you have another suggestion?

multimodal/setup.py
Innixma: They were originally in `tests`, I have not modified the logic from the original. You may consider editing this logic in a follow-up PR, as the purpose of this PR is specifically to move PyMuPDF to optional.

multimodal/setup.py
zhiqiangdon: Got it. Maybe we can put `onnx` and `tensorrt` into another category in future.

zhiqiangdon (approved):
LGTM!

Issue #3341, opened 2023-06-21
[BUG] [timeseries] DirectTabularModel fails 


noamanemobidata:
**Bug Report Checklist**

- [ ] I provided code that demonstrates a minimal reproducible example.  
- [ ] I confirmed bug exists on the latest mainline of AutoGluon via source install.  
- [ ] I confirmed bug exists on the latest stable version of AutoGluon.  

**Describe the bug**
Training DirectTabularModel fails : 

Warning: Exception caused DirectTabular to fail during training... Skipping this model.
	'Series' object has no attribute 'set_index'
Traceback (most recent call last):
  File "/local_disk0/.ephemeral_nfs/cluster_libraries/python/lib/python3.10/site-packages/autogluon/timeseries/trainer/abstract_trainer.py", line 495, in _train_and_save
    model = self._train_single(train_data, model, val_data=val_data, time_limit=time_limit)
  File "/local_disk0/.ephemeral_nfs/cluster_libraries/python/lib/python3.10/site-packages/autogluon/timeseries/trainer/abstract_trainer.py", line 408, in _train_single
    model.fit(
  File "/local_disk0/.ephemeral_nfs/cluster_libraries/python/lib/python3.10/site-packages/autogluon/timeseries/models/abstract/abstract_timeseries_model.py", line 235, in fit
    return super().fit(**kwargs)
  File "/local_disk0/.ephemeral_nfs/cluster_libraries/python/lib/python3.10/site-packages/autogluon/core/models/abstract/abstract_model.py", line 811, in fit
    out = self._fit(**kwargs)
  File "/local_disk0/.ephemeral_nfs/cluster_libraries/python/lib/python3.10/site-packages/autogluon/timeseries/models/multi_window/multi_window_model.py", line 105, in _fit
    model.score_and_cache_oof(val_fold, store_val_score=True, store_predict_time=True)
  File "/local_disk0/.ephemeral_nfs/cluster_libraries/python/lib/python3.10/site-packages/autogluon/timeseries/models/abstract/abstract_timeseries_model.py", line 374, in score_and_cache_oof
    self._oof_predictions = self.predict(past_data, known_covariates=known_covariates)
  File "/local_disk0/.ephemeral_nfs/cluster_libraries/python/lib/python3.10/site-packages/autogluon/timeseries/models/autogluon_tabular/direct_tabular.py", line 389, in predict
    predictions.set_index(data_future.index, inplace=True)
[...]

PR #3350, opened 2023-06-26
[timeseries] Fix bugs in DirectTabularModel

shchur:
*Issue #, if available:* #3341

*Description of changes:*
This PR addresses two bugs related to the `DirectTabularModel` introduced in v0.8.
1. `DirectTabular` model fails for non-default values of `eval_metric`.
    - **Reason** 
    Under the hood, `TabularPredictor` is used differently depending on the `eval_metric`. 
        - If `eval_metric="mean_wQuantileLoss"`: we use `problem_type="quantile"`, predictor returns a `pd.DataFrame`.
        - If `eval_metric!="mean_wQuantileLoss"`: we use `problem_type="regression"`, predictor returns a `pd.Series`. Unlike DataFrame, Series has no method `set_index`, which leads to an exception. 
    - **Why didn't we catch this bug before?**
    Our smoke tests did not consider the non-default values for `eval_metric`. When doing a sanity check of the leaderboards produced during benchmark runs (to make sure that no model fails on any dataset), I picked the first folder for each dataset, which happened to correspond to the quantile loss eval metric, where the model didn't fail.
    - **How do we avoid such bugs in the future?**
    I modified the smoke test in `tests/smoketests/test_features_and_covariates.py` to run with both quantile & point forecast metrics. Another good idea is to use a static analysis tool like https://github.com/astral-sh/ruff to catch such bugs in the future.
2. Warnings produced by `AutoARIMA` were not silenced in some cases and flooded the logs (see example below).
    - **Reason** 
    An interaction between Joblib, Python warning management, and the changed model order in the presets. The timeline looks as follows:
        1. We train most local models that all rely on Joblib (`Naive`, `SeasonalNaive`, `AutoETS`, `Theta`). Joblib creates a process pool, where the warnings are silenced using `PYTHONWARNINGS=ignore` environmental variable. This pool is reused by all models in this group, warnings are silenced.
        2. Train other models (e.g., `DeepAR`, `RecursiveTabular`) that don't use Joblib. 
        3. Train `DirectTabular`. This model uses Joblib but never produces any warnings itself.
            - If step (b) took < 300 seconds, this model re-uses the existing process pool. 
            - If step (b) took > 300 seconds, a new process pool is created WITHOUT the `PYTHONWARNINGS=ignore` flag.
        5. Train `AutoARIMA` -> we re-use the process pool used by `DirectTabular`. We always set the flag `PYTHONWARNINGS=ignore` here, but this may have no effect if the pool was created by `DirectTabular` without the flag. In this case, warnings leak into the log.
    - **Why didn't we catch this bug before?**
    I ran the sanity checks on small datasets, where step (b) took < 300 seconds, so the bug did not occur. Out CI does not check if warnings are raised.
    - **How do we avoid such bugs in the future?**
    I made sure that the flag `PYTHONWARNINGS=ignore` is set in the `DirectTabular` model. Ideally, a larger refactor to standardize the usage of warning filters in the code would be helpful, but I would postpone these changes to v0.9. I don't think that there is a waterproof way to check this automatically in the CI, but we should investigate the logs for large benchmark runs more thoroughly.
 
Example warnings produced by AutoARIMA: 
```
/home/shchuro/miniconda3/envs/ag/lib/python3.9/site-packages/statsforecast/arima.py:884: UserWarning: possible convergence problem: minimize gave code 2]
  warnings.warn(
/home/shchuro/miniconda3/envs/ag/lib/python3.9/site-packages/statsforecast/arima.py:884: UserWarning: possible convergence problem: minimize gave code 2]
  warnings.warn(
/home/shchuro/miniconda3/envs/ag/lib/python3.9/site-packages/statsforecast/arima.py:884: UserWarning: possible convergence problem: minimize gave code 2]
  warnings.warn(
```

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/autogluon_tabular/direct_tabular.py
+++ b/timeseries/src/autogluon/timeseries/models/autogluon_tabular/direct_tabular.py
@@ -16,6 +16,7 @@ from autogluon.timeseries.dataset.ts_dataframe import ITEMID, TIMESTAMP, TimeSer
 from autogluon.timeseries.models.abstract import AbstractTimeSeriesModel
 from autogluon.timeseries.models.local.abstract_local_model import AG_DEFAULT_N_JOBS
 from autogluon.timeseries.utils.forecast import get_forecast_horizon_index_ts_dataframe
+from autogluon.timeseries.utils.warning_filters import statsmodels_joblib_warning_filter, statsmodels_warning_filter
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


timeseries/src/autogluon/timeseries/models/autogluon_tabular/direct_tabular.py
```diff
@@ -386,7 +388,7 @@ def predict(
         # Predict for batches (instead of using full dataset) to avoid high memory usage
         batches = features.groupby(np.arange(len(features)) // self.PREDICTION_BATCH_SIZE, sort=False)
         predictions = pd.concat([self.tabular_predictor.predict(batch) for _, batch in batches])
-        predictions.set_index(data_future.index, inplace=True)
+        predictions.index = data_future.index
```
Innixma, line 391: Wow, that is very subtle. I'm surprised Series doesn't have `.set_index`, seems like a strange difference in API. Good catch!

Innixma (approved):
LGTM!

Issue #3260, opened 2023-06-01
[Tabular] Improve path logic


yinweisu:
Currently path are stored as pure string, and `+` operations are used. This can easily lead to issue when dealing cross-os usage.

We would want to
1. Refactor to use `Pathlib.Path` in most of the places (at least in the core area)
2. Consider switching to a logic to store paths in a cascade manner, where a parent store both its path and its children's paths relative to its own path. And whenever a model creates / load its child, it passes the correct path to the child.

Innixma:
Additionally:

- [ ] Delete StackerEnsembleModel, move logic into `AbstractTrainer`: Trainer should be the class to prepare the input and know the connections of the stacker models, because it is the object that stores the DAG of model connections. This should resolve the lions share of the path bugs.

Innixma:
Related: #3325

Innixma:
Related: #3318

PR #3355, opened 2023-06-26
Refactor path

yinweisu:
Issue #, if available:
* https://github.com/autogluon/autogluon/issues/3260

Description of changes:
* Refactor path logic in Tabular/TimeSeries, so that parent only will track relative paths to its children and passing correct path to its children when init/load
* Trainer now stores model's relative path as a list, and dynamically construct them to a path.
* Replaced usage os `os.path.sep` + str concatenation in path with `os.path.join` to avoid bugs

- [x] Cross OS check for tabular
- [x] Cross OS check for timeseries

### DIFF
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -120,7 +120,7 @@ class AbstractModel:
         # v0.9 FIXME: This is a hack, change so we aren't vulnerable to self.path_root breaking things
         self.path_root = PathConverter.to_relative(self.path_root)
 
-        self.path = self.create_contexts(self.path_root + self.path_suffix)  # TODO: Make this path a function for consistency.
+        self.path = self.create_contexts(os.path.join(self.path_root, self.path_suffix))  # TODO: Make this path a function for consistency.
 
         self.num_classes = None
         self.model = None
@@ -1002,7 +1002,7 @@ class AbstractModel:
             Path to the saved model, minus the file name.
             This should generally be a directory path ending with a '/' character (or appropriate path separator value depending on OS).
             If None, self.path is used.
-            The final model file is typically saved to path + self.model_file_name.
+            The final model file is typically saved to os.path.join(path, self.model_file_name).
         verbose : bool, default True
             Whether to log the location of the saved file.
 
@@ -1037,7 +1037,7 @@ class AbstractModel:
         path : str
             Path to the saved model, minus the file name.
             This should generally be a directory path ending with a '/' character (or appropriate path separator value depending on OS).
-            The model file is typically located in path + cls.model_file_name.
+            The model file is typically located in os.path.join(path, cls.model_file_name).
         reset_paths : bool, default True
             Whether to reset the self.path value of the loaded model to be equal to path.
             It is highly recommended to keep this value as True unless accessing the original self.path value is important.
@@ -1708,7 +1708,7 @@ class AbstractModel:
 
     @classmethod
     def load_info(cls, path, load_model_if_required=True) -> dict:
-        load_path = path + cls.model_info_name
+        load_path = os.path.join(path, cls.model_info_name)
         try:
             return load_pkl.load(path=load_path)
         except:
@@ -1721,8 +1721,8 @@ class AbstractModel:
     def save_info(self) -> dict:
         info = self.get_info()
 
-        save_pkl.save(path=self.path + self.model_info_name, object=info)
-        json_path = self.path + self.model_info_json_name
+        save_pkl.save(path=os.path.join(self.path, self.model_info_name), object=info)
+        json_path = os.path.join(self.path, self.model_info_json_name)
         save_json.save(path=json_path, obj=info)
         return info
 
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -1051,6 +1051,9 @@ class AbstractModel:
             Loaded model object.
         """
         file_path = os.path.join(path, cls.model_file_name)
+        print("model load")
+        print(path)
+        print(file_path)
         model = load_pkl.load(path=file_path, verbose=verbose)
         if reset_paths:
             model.set_contexts(path)
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -70,7 +70,7 @@ class AbstractModel:
         If None, a new unique time-stamped directory is chosen.
     name : str, default = None
         Name of the subdirectory inside path where model will be saved.
-        The final model directory will be path+name+os.path.sep()
+        The final model directory will be os.path.join(path, name)
         If None, defaults to the model's class name: self.__class__.__name__
     problem_type : str, default = None
         Type of prediction problem, i.e. is this a binary/multiclass classification or regression problem (options: 'binary', 'multiclass', 'regression').
@@ -112,9 +112,6 @@ class AbstractModel:
         self.path_root = path
         if self.path_root is None:
             path_suffix = self.name
-            if len(self.name) > 0:
-                if self.name[0] != os.path.sep:
-                    path_suffix = os.path.sep + path_suffix
             # TODO: Would be ideal to not create dir, but still track that it is unique. However, this isn't possible to do without a global list of used dirs or using UUID.
             path_cur = setup_outputdir(path=None, create_dir=True, path_suffix=path_suffix)
             self.path_root = path_cur.rsplit(self.path_suffix, 1)[0]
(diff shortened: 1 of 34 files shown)


core/src/autogluon/core/models/ensemble/stacker_ensemble_model.py
```diff
@@ -54,11 +54,7 @@ def __init__(
             base_model_types_dict = {}
         self.base_model_names = base_model_names
         self.base_models_dict: Dict[str, AbstractModel] = base_models_dict  # String name -> Model objects
-        self.base_model_paths_dict = base_model_paths_dict
-
-        # FIXME: DO NOT DO THIS, FIX ASAP
-        self.base_model_paths_dict = {k: PathConverter.to_relative(v) for k, v in self.base_model_paths_dict.items()}
-
+        self.base_model_paths_dict = {key: os.path.relpath(val, self.path) for key, val in base_model_paths_dict.items()}
```
Innixma, line 57: Very nice. I think this may even fix all of the issues in stacker ensemble model.

Innixma (approved):
LGTM, this is an excellent and much needed refactor!

core/src/autogluon/core/trainer/abstract_trainer.py
```diff
@@ -292,23 +291,12 @@ def get_model_level(self, model_name: str) -> int:
...
 
     def set_contexts(self, path_context):
-        self.path, model_paths = self.create_contexts(path_context)
-        for model, path in model_paths.items():
-            self.set_model_attribute(model=model, attribute="path", val=path)
+        self.path = self.create_contexts(path_context)
 
-    def create_contexts(self, path_context: str) -> (str, dict):
-        self.path = PathConverter.to_current(self.path)
+    def create_contexts(self, path_context: str) -> str:
```
shchur, line 296: It seems that the `create_contexts` method always just returns the argument passed to it. Can this method be removed?

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -39,7 +39,6 @@ class SimpleAbstractTrainer:
```
shchur, line 39: @Innixma Around the v0.5.0 release, we planned to move the `SimpleAbstractTrainer` class to `autogluon.core`, which would prevent code duplication between timeseries & other modules (see the comments in lines 31-34 above). Do you think that's something that we should still do?

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -112,18 +112,12 @@ def __init__(self, path: str = None, name: str = None, problem_type: str = None,
...
             # TODO: Would be ideal to not create dir, but still track that it is unique. However, this isn't possible to do without a global list of used dirs or using UUID.
             path_cur = setup_outputdir(path=None, create_dir=True, path_suffix=path_suffix)
             self.path_root = path_cur.rsplit(self.path_suffix, 1)[0]
             logger.log(20, f"Warning: No path was specified for model, defaulting to: {self.path_root}")
 
-        # v0.9 FIXME: This is a hack, change so we aren't vulnerable to self.path_root breaking things
-        self.path_root = PathConverter.to_relative(self.path_root)
-
-        self.path = self.create_contexts(self.path_root + self.path_suffix)  # TODO: Make this path a function for consistency.
+        self.path = self.create_contexts(os.path.join(self.path_root, self.path_suffix))  # TODO: Make this path a function for consistency.
```
shchur, line 120: Should we consider storing `path` as a `pathlib.Path` object (as mentioned in #3260)? Might not be worth it if this is a lot work + the current PR already addresses all the path problems, but just wanted to mention this option.

shchur (approved):
Thank you for the PR, this looks great!

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
Innixma: Yes, lets sync on this later in the week during our 1:1 to see what else needs to be done. Ideally the bulk of trainer and learner logic should be moved to core. I'd probably focus on this in mid-late August / September.

core/src/autogluon/core/models/abstract/abstract_model.py
Innixma: I believe @yinweisu tried to do this before, but for some reason we decided to keep it as string. @yinweisu do you recall the exact reason?

yinweisu: Yea, the reasoning behind is that there are certain downstream calls that expect the path to be a string instead of a `Path` object. Therefore, in order for it to be transparent, what we need to do would be having a path something like this: `str(Path(path) / "path_suffix")`. This looks verbose and unnecessary to me when `os.path.join` can already accomplish the job.

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
yinweisu: +1

core/src/autogluon/core/trainer/abstract_trainer.py
yinweisu: I guess it doesn't matter that much as having this function gives the reader a sense of what it's doing, especially when other abstract classes have similar methods. I'll say let's keep it for now

eromoe:
Similar error:
ValueError: path is on mount 'C:', start on mount 'E:'

I run code in `E` , and set path to `C`
```
TabularPredictor(label='y', eval_metric='precision', path=autgoluon_model_path)
```

```
ValueError                                Traceback (most recent call last)
~\AppData\Local\Temp/ipykernel_17032/2841266324.py in <module>
----> 1 m1 = train(X1_train, y1_train, X1_test, y1_test)

~\AppData\Local\Temp/ipykernel_17032/2534130983.py in train(X_train, y_train, X_test, y_test)
     42         m = model_cls(eval_metric=eval_metric, path=new_model_path)
     43
---> 44         m.fit(X_train, y_train)
     45
     46         if X_test.shape[0]:

e:\workspace\github_me\mlquant\mlquant\my_auto.py in fit(self, X, y, **kwargs)
     30
     31     def fit(self, X, y, **kwargs):
---> 32         return super().fit(X.assign(y=y), hyperparameters=self.hyperparameters, verbosity=0, **kwargs)
     33
     34     def predict_proba(self, data, model=None,):

~\anaconda3\lib\site-packages\autogluon\core\utils\decorators.py in _call(*args, **kwargs)
     29         def _call(*args, **kwargs):
     30             gargs, gkwargs = g(*other_args, *args, **kwargs)
---> 31             return f(*gargs, **gkwargs)
     32
     33         return _call

~\anaconda3\lib\site-packages\autogluon\tabular\predictor\predictor.py in fit(self, train_data, tuning_data, time_limit, presets, hyperparameters, feature_metadata, infer_limit, infer_limit_batch_size, fit_weighted_ensemble, calibrate_decision_threshold, num_cpus, num_gpus, **kwargs)
    984             aux_kwargs["fit_weighted_ensemble"] = False
    985         self.save(silent=True)  # Save predictor to disk to enable prediction and training after interrupt
--> 986         self._learner.fit(
    987             X=train_data,
    988             X_val=tuning_data,

~\anaconda3\lib\site-packages\autogluon\tabular\learner\abstract_learner.py in fit(self, X, X_val, **kwargs)
    157             raise AssertionError("Learner is already fit.")
    158         self._validate_fit_input(X=X, X_val=X_val, **kwargs)
--> 159         return self._fit(X=X, X_val=X_val, **kwargs)
    160
    161     def _fit(

~\anaconda3\lib\site-packages\autogluon\tabular\learner\default_learner.py in _fit(self, X, X_val, X_unlabeled, holdout_frac, num_bag_folds, num_bag_sets, time_limit, infer_limit, infer_limit_batch_size, verbosity, **trainer_fit_kwargs)
    133             time_limit_trainer = None
    134
--> 135         trainer = self.trainer_type(
    136             path=self.model_context,
    137             problem_type=self.label_cleaner.problem_type_transform,

~\anaconda3\lib\site-packages\autogluon\core\trainer\abstract_trainer.py in __init__(self, path, problem_type, eval_metric, num_classes, quantile_levels, low_memory, feature_metadata, k_fold, n_repeats, sample_weight, weight_evaluation, save_data, random_state, verbosity)
    102     ):
    103         self.path = path
--> 104         self.path = PathConverter.to_relative(self.path)
    105         self.problem_type = problem_type
    106         self.feature_metadata = feature_metadata

~\anaconda3\lib\site-packages\autogluon\common\utils\path_converter.py in to_relative(path)
     54             return path
     55         os_path_sep = PathConverter.os_path_sep()
---> 56         path_relative = os.path.relpath(path)
     57         if path.endswith(os_path_sep):
     58             if not path_relative.endswith(os_path_sep):

~\anaconda3\lib\ntpath.py in relpath(path, start)
    701         path_drive, path_rest = splitdrive(path_abs)
    702         if normcase(start_drive) != normcase(path_drive):
--> 703             raise ValueError("path is on mount %r, start on mount %r" % (
    704                 path_drive, start_drive))
    705

ValueError: path is on mount 'C:', start on mount 'E:'
```

PR #3373, opened 2023-07-01
vw version bump

mglowacki100:
bump vw version to 9.8

*Description of changes:*
I've just increased allowed version of wovpal wabbit to 9.8 in file:
https://github.com/autogluon/autogluon/blob/master/common/src/autogluon/common/utils/try_import.py
I've tested it locally.

### DIFF
--- a/common/src/autogluon/common/utils/try_import.py
+++ b/common/src/autogluon/common/utils/try_import.py
@@ -220,10 +220,10 @@ def try_import_vowpalwabbit():
 
         vowpalwabbit_version = parse_version(vowpalwabbit.__version__)
         assert vowpalwabbit_version >= parse_version("9.0.0") and vowpalwabbit_version < parse_version(
-            "9.5.0"
-        ), f"Currently, we only support vowpalwabbit version >=9.0 and <9.5. Found vowpalwabbit version: {vowpalwabbit_version}"
+            "9.9.0"
+        ), f"Currently, we only support vowpalwabbit version >=9.0 and <9.9. Found vowpalwabbit version: {vowpalwabbit_version}"
     except ImportError:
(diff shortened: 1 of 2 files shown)


Innixma (approved):
Thanks for the contribution! I've updated the PR to also bump the setup install version of vowpalwabbit to 9.8 to match the version range update in the import code. Will wait for CI to pass and then merge.

mglowacki100:
Thanks :) 
Btw. vowpal wabbit is not enabled by default, have you tested and it doesn't bring much value to ensemble? 
As far I remember it was sometimes used in old kaggle competitions, as it scaled well with interaction features and could be configured as factorization machine.

Innixma:
@mglowacki100 I believe VowpalWabbit was contributed awhile ago but it probably needs some default hyperparameter tweaks to ensure better results. At present the defaults don't seem to do that well. If you have suggestions on better hyperparameter settings, happy to adjust it and it is easy for me to test.

Merging as CI passed, thanks again!

mglowacki100:
Hi @Innixma 
I've checked a couple of parameters locally and think the biggest impact and quite generic could be adding interactions between all features:
`hyperparameters = {'VW': [{}, {'interactions': '::'}]}`
also you could modify https://github.com/autogluon/autogluon/blob/master/tabular/src/autogluon/tabular/models/vowpalwabbit/vowpalwabbit_model.py line 232:
```
    def _set_default_params(self):
        default_params = {
            "passes": 10,  # TODO: Much better if 500+, revisit this if wanting to use VW to get strong results #
            "bit_precision": 32,
            "ngram": 2,
            "skips": 1,
            "learning_rate": 1,
            "sparse_weights": True,
            "interactions": "::", #added by MG
        }

```

PR #3375, opened 2023-07-02
fallback load model on CPU in TabularNeuralNetTorchModel

s41m0n:
This PR introduces support for loading PyTorch models previously trained on a device (e.g., cuda GPU) not currently available, falling back to CPU mapping of the model state.

Despite being a PyTorch-related issue, we could introduce this temporary fix until the PyTorch community approves the changes, so that it is possible to operate (e.g., visualize, plot, consult architecture, etc.) on GPU-trained models from a device with only CPU (e.g., perform model training/HPO/NAS on GPU-enabled devices/cluster for performance, and then do some post-ops and analysis on the local machine).

For the original PyTorch issue (still open) please refer to https://github.com/pytorch/pytorch/issues/16797

## Previous error message

Attempting to load GPU-trained (Cuda) models on a non-GPU-enabled device:
 
```bash
Loading: withgpu/predictor.pkl
Loading: withgpu/learner.pkl
Loading: withgpu/models/trainer.pkl
Loading: withgpu/models/NeuralNetTorch/cd903054/model.pkl
  File "~/scripts/printArchParams.py", line 41, in <module>
    main()
  File "~/scripts/printArchParams.py", line 23, in main
    model = predictor._trainer.load_model(predictor.get_model_best())
  File "~/.venv/lib/python3.9/site-packages/autogluon/core/trainer/abstract_trainer.py", line 1390, in load_model
    return model_type.load(path=path, reset_paths=self.reset_paths)
  File "~/.venv/lib/python3.9/site-packages/autogluon/tabular/models/tabular_nn/torch/tabular_nn_torch.py", line 604, in load
    model: TabularNeuralNetTorchModel = super().load(path=path, reset_paths=reset_paths, verbose=verbose)
  File "~/.venv/lib/python3.9/site-packages/autogluon/core/models/abstract/abstract_model.py", line 1034, in load
    model = load_pkl.load(path=file_path, verbose=verbose)
  File "~/.venv/lib/python3.9/site-packages/autogluon/common/loaders/load_pkl.py", line 43, in load
    object = pickle.load(fin)
  File "~/.venv/lib/python3.9/site-packages/torch/storage.py", line 240, in _load_from_bytes
    return torch.load(io.BytesIO(b))
  File "~/.venv/lib/python3.9/site-packages/torch/serialization.py", line 795, in load
    return _legacy_load(opened_file, map_location, pickle_module, **pickle_load_args)
  File "~/.venv/lib/python3.9/site-packages/torch/serialization.py", line 1012, in _legacy_load
    result = unpickler.load()
  File "~/.venv/lib/python3.9/site-packages/torch/serialization.py", line 958, in persistent_load
    wrap_storage=restore_location(obj, location),
  File "~/.venv/lib/python3.9/site-packages/torch/serialization.py", line 215, in default_restore_location
    result = fn(storage, location)
  File "~/.venv/lib/python3.9/site-packages/torch/serialization.py", line 182, in _cuda_deserialize
    device = validate_cuda_device(location)
  File "~/.venv/lib/python3.9/site-packages/torch/serialization.py", line 166, in validate_cuda_device
    raise RuntimeError('Attempting to deserialize object on a CUDA '
RuntimeError: Attempting to deserialize object on a CUDA device but torch.cuda.is_available() is False. If you are running on a CPU-only machine, please use torch.load with map_location=torch.device('cpu') to map your storages to the CPU.
```

## Description of Changes

Credits to the original PyTorch issue discussion. The change consists of introducing a custom *Unpickler* class with the *find_class* method that automatically accounts for introducing the *map_location='cpu'* to the *torch.load* function. At this point, the model is loaded in the CPU, despite its original device used.

### DIFF
--- a/common/src/autogluon/common/loaders/load_pkl.py
+++ b/common/src/autogluon/common/loaders/load_pkl.py
@@ -77,3 +77,27 @@ def load_with_fn(path, pickle_fn, format=None, verbose=True):
     with open(path, "rb") as fin:
         object = pickle_fn(fin)
     return object
+
+
+class SafeTorchUnpicklerCPU(pickle.Unpickler):
+    """
+    SafeTorchUnpicklerCPU class for safely unpickling objects with underlying
+    Torch model previously trained on a device not currently available.
+    (e.g., 'cuda' or 'mps')
+
+    PyTorch issue not yet fixed at https://github.com/pytorch/pytorch/issues/16797
+    """
+
+    def find_class(self, module, name):
+        """
+        Method to return the correct method for loading the object.
+        If attempting to load the actual Torch model, return the safe torch.load method using cpu.
+        For all the other objects (e.g., the TabularNeuralNetTorchModel wrapper) use the default method.
+        """
+
+        import torch
+
+        if module == "torch.storage" and name == "_load_from_bytes":
+            return lambda b: torch.load(io.BytesIO(b), map_location=torch.device("cpu"))
+        else:
+            return super().find_class(module, name)
--- a/tabular/src/autogluon/tabular/models/tabular_nn/torch/tabular_nn_torch.py
+++ b/tabular/src/autogluon/tabular/models/tabular_nn/torch/tabular_nn_torch.py
@@ -10,6 +10,7 @@ import numpy as np
 import pandas as pd
 
 from autogluon.common.features.types import R_BOOL, R_CATEGORY, R_FLOAT, R_INT, S_TEXT_AS_CATEGORY, S_TEXT_NGRAM
+from autogluon.common.loaders.load_pkl import SafeTorchUnpicklerCPU
 from autogluon.common.utils.resource_utils import ResourceManager
 from autogluon.common.utils.try_import import try_import_torch
 from autogluon.core.constants import BINARY, MULTICLASS, QUANTILE, REGRESSION, SOFTCLASS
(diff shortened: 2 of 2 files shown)


Innixma:
Thanks for the contribution @s41m0n!! We will target merging this for the next release once we test it out. This feature is much appreciated and will go a long way to making GPU better supported in TabularPredictor!

Innixma:
@yinweisu good catch!

@s41m0n can you test v0.8.2 and see if it fixes the issue you are facing?

@yinweisu The implementation of the fix is pretty different between the two options. We may want to think about which one is more elegant.

s41m0n:
> @yinweisu good catch!
> 
> @s41m0n can you test v0.8.2 and see if it fixes the issue you are facing?
> 
> @yinweisu The implementation of the fix is pretty different between the two options. We may want to think about which one is more elegant.

I have tested this behaviour in the 0.8.2 release today and I am still facing problems.
Despite being an elegant solution, the fix introduced in the previous PR does not let me load a GPU-trained model on a CPU-only device. I also tried to patch the following function with no luck :https://github.com/autogluon/autogluon/blob/526bc1754e67d2dbe7b974fe23e813f1422db897/tabular/src/autogluon/tabular/models/tabular_nn/torch/tabular_nn_torch.py#L575

```python
        import torch

        # Save on CPU to ensure the model can be loaded on a box without GPU
        if self.model is not None:
            device = torch.device('cpu')     # NEWLY INTRODUCED
            self.device = device             # NEWLY INTRODUCED
            self.model.device = device       # NEWLY INTRODUCED
            self.model = self.model.to(device)
        path = super().save(path, verbose)
        # Put the model back to the device after the save
        if self.model is not None:
            self.model.to(self.device)
        return path
```

Still, when loading back the model the same CUDA error appears. Moreover, when attempting to re-load the model, the knowledge of the previous device used for the training would've been lost with these lines.

Please let me know if I am wrong. It seems that `self.model = self.model.to(torch.device('cpu'))` is not enough for ensuring safety when loading back the models.

Otherwise, I would be more than glad to proceed with the PR :)

yinweisu:
This is interesting if the previous commit didn't fix this issue. It's hard to have unit test for cross system saving and loading, I'll do some manual testing to verify this.

Besides, regarding the verbosity of hpo. Currently, we are not exposing that to the users, I'll create an issue to track this. Thanks for reporting!

yinweisu:
@s41m0n I just tried our 0.8.2 release and was able to load a gpu trained model on cpu.

To be more specific, I trained NN_TORCH model with gpu, verified gpu usage through nvidia-smi. Then disable gpu through `export CUDA_VISIBLE_DEVICES=""`. The tried to load the previous trained model after verifying `assert torch.cuda.is_available() == False`. I was able to load the model correctly.

If you look at the commit that fixed this issue, we always save the model to cpu. It's only during loading that we move the model to the correct device. I guess you were trying to load a model trained in older AG version? Since the saving logic was modified to fix this issue, models trained in previous AG version wouldn't work

Innixma:
Closing as we believe the issue has been fixed in v0.8.2 based on @yinweisu's investigation above. Thanks for the contribution @s41m0n, please let us know if the issue is still present in v0.8.2 by opening a new GitHub issue.

PR #3379, opened 2023-07-06
[AutoMM] Support customizing use_fast for AutoTokenizer

zhiqiangdon:
*Issue #, if available:*

*Description of changes:*
Through customizing `hf_text.use_fast`, users can choose whether to use fast tokenizer for hf_text models. If not set, we use the same default as the AutoTokenizer.

### DIFF
--- a/multimodal/src/autogluon/multimodal/data/process_text.py
+++ b/multimodal/src/autogluon/multimodal/data/process_text.py
@@ -95,6 +95,7 @@ class TextProcessor:
         train_augment_types: Optional[List[str]] = None,
         template_config: Optional[DictConfig] = None,
         normalize_text: Optional[bool] = False,
+        use_fast: Optional[bool] = None,
     ):
         """
         Parameters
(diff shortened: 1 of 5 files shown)


multimodal/src/autogluon/multimodal/data/process_text.py
```diff
@@ -420,14 +428,21 @@ def get_pretrained_tokenizer(
...
+            See: https://huggingface.co/docs/transformers/model_doc/auto#transformers.AutoTokenizer.from_pretrained.use_fast
 
         Returns
         -------
         A tokenizer instance.
         """
         try:
             tokenizer_class = ALL_TOKENIZERS[tokenizer_name]
-            return tokenizer_class.from_pretrained(checkpoint_name)
+            if use_fast is None:  # to be backward compatible
```
tonyhoo, line 442: Can we default it to True for backward compatibility?

zhiqiangdon: Done

tonyhoo (approved):

PR #3404, opened 2023-07-17
[AutoMM] Upgrade torch

zhiqiangdon:
*Issue #, if available:*

*Description of changes:*

- Upgrade torch to 2.0 for AutoMM.
- Clean up CI env setup to use AutoMM's setup to pin torch version. This is to ensure the torch version we test is the same as the version users install.


shchur (approved):
Thanks for the PR! Changes to `timeseries` look good to me, but I cannot comment on the workflow scripts.

multimodal/setup.py
```diff
@@ -33,8 +33,8 @@
     "evaluate>=0.2.2,<0.4.0",
     "accelerate>=0.9,<0.17",
     "timm>=0.9.2,<0.10.0",
-    "torch>=1.9,<1.14",
-    "torchvision<0.15.0",
+    "torch>=1.13,<2.1",
```
yinweisu, line 36: Noticed torch requirements appear in multiple modules. Not sure if it's doable to synchronize them into once place especially when torch for tabular is in extra dependency. @Innixma Any idea?

.github/workflow_scripts/env_setup.sh
```diff
@@ -16,24 +16,6 @@ function setup_build_contrib_env {
...
-}
-
-function setup_torch_gpu_non_linux {
-    pip3 install torch==1.13.1+cu117 torchvision==0.14.1+cu117 --extra-index-url https://download.pytorch.org/whl/cu117
-}
-
-function setup_torch_cpu_non_linux {
-    pip3 install torch==1.13.1 torchvision==0.14.1
-}
-
```
Innixma, line 36: Is there a reason we are able to remove all of these without issue?

Innixma (approved):
LGTM!

.github/workflow_scripts/env_setup.sh
yinweisu: Previously these was added because we needed a specific security patched version of torch. Now we don't need them, so safe to remove

Innixma: Sounds good

zhiqiangdon (commented):
/platform_tests

multimodal/setup.py
```diff
@@ -27,14 +27,14 @@
...
+    "torch",  # version range defined in `core/_setup_utils.py`
     "requests>=2.21,<3",
     "jsonschema>=4.14,<4.18",
     "seqeval>=1.2.2,<1.3.0",
     "evaluate>=0.2.2,<0.4.0",
     "accelerate>=0.9,<0.17",
     "timm>=0.9.2,<0.10.0",
-    "torch>=1.9,<1.14",
-    "torchvision<0.15.0",
+    "torchvision>=0.15.0,<0.16.0",
```
ddelange, line 37: fyi `torchvision==0.15.0` pins `torch==2.0.0` so technically this drops support for torch v1

zhiqiangdon: Thanks for pointing it out. I will loose the lower bound to 0.14 in future PRs.

PR #3406, opened 2023-07-18
[fix][eda] Anomaly Detection - disable bps_flag for SUOD

gradientsky:
*Description of changes:*

Don't use `bps_flag=True` - it's using pre-trained models which aren't loading in newer versions of sklearn


tonyhoo (approved):

zhiqiangdon (approved):
Thanks for the quick fix!

PR #3419, opened 2023-07-25
Upgrade lightning to 2.0

zhiqiangdon:
*Issue #, if available:*

*Description of changes:*
- Upgrade lightning to 2.0 for both Multimodal and Timeseries and move the lightning caps to core.
- Re-implement the support of `track_grad_norm`. Refer to https://github.com/Lightning-AI/lightning/pull/16745
- Fix the skip_final_val. Can't set self.trainer.val_dataloaders = [] because val_dataloaders is trainer's property. Refer to https://github.com/Lightning-AI/lightning/blob/6552d29a121b3ec7739436fe12a9b0c3bbd14818/src/lightning/pytorch/callbacks/callback.py#L122
- Default accelerator, devices, strategy to `auto` instead of None, which is required by lightning 2.0.
- Avoid using `strategy_name` attribute since some strategies like DDPStrategy no longer has it.

TODOs:

- replace dp strategy in inference with ddp or ddp_spawn since dp is not supported in lightning 2.0.

### DIFF
--- a/core/src/autogluon/core/_setup_utils.py
+++ b/core/src/autogluon/core/_setup_utils.py
@@ -27,6 +27,7 @@ DEPENDENT_PACKAGES = {
     "tqdm": ">=4.38,<5",  # Major version cap
     "Pillow": ">=9.3,<9.6",  # "<{N+2}" upper cap
     "torch": ">=1.13,<2.1",  # "<{N+1}" upper cap
+    "pytorch-lightning": ">=2.0.0,<2.1",  # "<{N+1}" upper cap
 }
 if LITE_MODE:
     DEPENDENT_PACKAGES = {package: version for package, version in DEPENDENT_PACKAGES.items() if package not in ["psutil", "Pillow", "timm"]}
--- a/core/src/autogluon/core/_setup_utils.py
+++ b/core/src/autogluon/core/_setup_utils.py
@@ -26,6 +26,7 @@ DEPENDENT_PACKAGES = {
     "networkx": ">=3.0,<4",  # Major version cap
     "tqdm": ">=4.38,<5",  # Major version cap
     "Pillow": ">=9.3,<9.6",  # "<{N+2}" upper cap
+    "torch": ">=1.13,<2.1",  # "<{N+1}" upper cap
 }
 if LITE_MODE:
     DEPENDENT_PACKAGES = {package: version for package, version in DEPENDENT_PACKAGES.items() if package not in ["psutil", "Pillow", "timm"]}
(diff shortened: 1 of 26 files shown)


timeseries/setup.py
```diff
@@ -30,7 +30,7 @@
     "torch",  # version range defined in `core/_setup_utils.py`
     "statsmodels>=0.13.0,<0.15",
     "gluonts>=0.13.1,<0.14",
-    "pytorch-lightning>=1.7.4,<1.10.0",
+    "pytorch-lightning>=2.0.0,<2.1",
```
shchur, line 33: Changes to `timeseries` look good to me, I validated that everything works as expected locally with `pytorch-lightning` v2.0.

Feel free to merge once the changes to `multimodal` are approved.

multimodal/src/autogluon/multimodal/optimization/lit_module.py
```diff
@@ -5,6 +5,7 @@
 import torch
 import torch.nn.functional as F
 import torchmetrics
+from pytorch_lightning.utilities import grad_norm
```
shchur, line 8: Should we at some point consider switching to the new API (as described [here](https://github.com/Lightning-AI/lightning/discussions/16688#discussioncomment-4927515)) or should we wait until the deprecation notice is released for the old API?
```python
import lightning as L
```

zhiqiangdon: Good call. We can switch to the new namespace when it becomes more stable.

yinweisu (commented):
Changes to `_setup_utils` look good to me

core/src/autogluon/core/_setup_utils.py
```diff
@@ -27,6 +27,7 @@
     "tqdm": ">=4.38,<5",  # Major version cap
     "Pillow": ">=9.3,<9.6",  # "<{N+2}" upper cap
     "torch": ">=1.13,<2.1",  # "<{N+1}" upper cap
+    "pytorch-lightning": ">=2.0.0,<2.1",  # "<{N+1}" upper cap
```
tonyhoo, line 30: Is it compatible with torch 1.13, which is the minimum requirement we have above

zhiqiangdon: Yes, lightning 2.0 also supports torch 1.13. See the doc: https://lightning.ai/docs/pytorch/latest/versioning.html#compatibility-matrix

tonyhoo (approved):
LGTM

PR #3421, opened 2023-07-25
Tabular: Add model failures tracking support

Innixma:
*Issue #, if available:*

*Description of changes:*
- Add model failures tracking support via `predictor.get_model_failures()`
- Improve LightGBM, CatBoost, and XGBoost memory estimate by estimating the histogram memory usage.
- Improve stability of refit full when memory usage is high by increasing the memory usage cap by 25% for refit models to avoid fitting the original model successfully and then having the refit model cause a NotEnoughMemoryError exception due to requiring slightly more memory.
- PR has been tested extensively across 244 datasets with 1416 model configs. It works well, eliminating all cases of OOM due to incorrect memory estimates for GBM models, and the model failures logic works in all cases.

TODO:
- [x] Add unit tests
- [x] Add ability to not raise error when no models were trained
- [x] Polish docstrings

### DIFF
--- a/core/src/autogluon/core/trainer/abstract_trainer.py
+++ b/core/src/autogluon/core/trainer/abstract_trainer.py
@@ -4,10 +4,12 @@ import copy
 import logging
 import os
 import shutil
+import sys
 import time
+import traceback
 from collections import defaultdict
 from pathlib import Path
-from typing import Dict, List, Optional, Tuple, Union
+from typing import Any, Dict, List, Optional, Tuple, Union
 
 import networkx as nx
 import numpy as np
@@ -172,6 +174,7 @@ class AbstractTrainer:
         self._extra_banned_names = set()  # Names which are banned but are not used by a trained model.
 
         self._models_failed_to_train = []  # List of models which failed to train
+        self._models_failed_to_train_errors = dict()  # Dict of model name -> model failure metadata
 
         # self._exceptions_list = []  # TODO: Keep exceptions list for debugging during benchmarking.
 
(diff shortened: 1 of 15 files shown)


core/src/autogluon/core/trainer/abstract_trainer.py
```diff
@@ -1795,47 +1798,86 @@ def _train_and_save(
...
+                    logger.exception("Detailed Traceback:")
+            crash_time = time.time()
+            total_time = crash_time - fit_start_time
+            tb = traceback.format_exc()
+            model_info = self.get_model_info(model=model)
             self._models_failed_to_train.append(model.name)
-            if self.verbosity > 0:
-                logger.exception("Detailed Traceback:")
-            del model
+            self._models_failed_to_train_errors[model.name] = dict(
```
yinweisu, line 1829: With the addition of `_models_failed_to_train_errors`, we prolly can get rid of `_models_failed_to_train `

core/src/autogluon/core/trainer/abstract_trainer.py
```diff
@@ -2897,7 +2967,93 @@ def leaderboard(self, extra_info=False):
...
+            model_info_inner = self._flatten_model_info(model_info_inner)
+
+            valid_keys = [
+                "exc_type",
+                "exc_str",
+                "exc_traceback",
+                "total_time",
+            ]
+            valid_keys_inner = [
+                # "name",
```
yinweisu, line 2995: Do we want to keep this?

core/src/autogluon/core/trainer/abstract_trainer.py
```diff
@@ -2897,7 +2967,93 @@ def leaderboard(self, extra_info=False):
...
+            data=model_infos,
+        ).T
+
+        explicit_order = [
+            "model",
+            "exc_type",
+            "total_time",
+            "model_type",
+            "child_model_type",
+            # "stack_level",
```
yinweisu, line 3029: ditto, will skip the rest

yinweisu (approved):
LGTM! Added some minor comments

core/src/autogluon/core/trainer/abstract_trainer.py
Innixma: good point, removed

core/src/autogluon/core/trainer/abstract_trainer.py
Innixma: good point! removed

core/src/autogluon/core/trainer/abstract_trainer.py
Innixma: good point! removed

PR #3427, opened 2023-07-27
Update setup.py - LightGBM version bump

mglowacki100:
*Description of changes:*
There is a new version of LightGBM: https://github.com/microsoft/LightGBM/releases/tag/v4.0.0
I think the most important part is quantization that promises 2x speed-up in training (https://arxiv.org/pdf/2207.09682.pdf)
It could be enabled by parameter `use_quantized_grad`  (https://lightgbm.readthedocs.io/en/latest/Parameters.html)

### DIFF
--- a/tabular/setup.py
+++ b/tabular/setup.py
@@ -33,7 +33,7 @@ install_requires = [
 
 extras_require = {
     "lightgbm": [
-        "lightgbm>=3.3,<4.2",  # <{N+1} upper cap, where N is the latest released minor version
+        "lightgbm>=3.3,<4.4",  # <{N+1} upper cap, where N is the latest released minor version
     ],
     "catboost": [
         # CatBoost wheel build is not working correctly on darwin for CatBoost 1.2, so use old version in this case.


mglowacki100:
I've no idea why it fails for `TimeSeries predictor` on AWS. I've tried to reproduce those tests locally on colab:
https://colab.research.google.com/drive/1qG70VA7KChjjagJsq_1VArjrV20inUjW?usp=sharing but it seems to work there correctly.

Innixma:
Wow!! v4.0 is awesome to see for LightGBM. Will definitely be taking a look at this when I find time. I can run a benchmark on AutoMLBenchmark to check performance differences. Thanks @mglowacki100 for creating the PR!

Innixma:
@shchur When you find time could you take a look at the TimeSeries test failures in this PR?

mglowacki100:
@Innixma Thanks! I've checked `use_quantized_grad` set to True with https://auto.gluon.ai/stable/tutorials/tabular/tabular-quick-start.html, but unfortunately I've got error from lightgbm internals: https://github.com/microsoft/LightGBM/issues/4946 
Nevertheless, it'd be interesting to check how often this error occurs and performance differences.
Note, that it doesn't explain problem with time series as `use_quantized_grad` default is False.

shchur:
It seems that the problem is unrelated to `timeseries` but rather related to the change that occurred in LightGBM v4.0.0 and our implementation of the custom early stopping callback.

- In v3.3.5, the second argument to `_format_eval_result` had a default value set to `True` ([source](https://github.com/microsoft/LightGBM/blob/v3.3.5/python-package/lightgbm/callback.py#L39))
- In v4.0.0, the default value got removed ([source](https://github.com/microsoft/LightGBM/blob/v4.0.0/python-package/lightgbm/callback.py#L56))

To fix this problem, we need to replace all occurrences of `_format_eval_result(x)` in https://github.com/autogluon/autogluon/blob/master/tabular/src/autogluon/tabular/models/lgb/callbacks.py with `_format_eval_result(x, show_stdv=True)`. This fixes the test for `timeseries`.

@mglowacki100 can you please update your PR with change?

@Innixma the reason why this problem wasn't caught in `tabular/tests/unittests/models/test_lightgbm.py` ([logs](https://github.com/autogluon/autogluon/actions/runs/5677860158/job/15387017052?pr=3427#step:4:1615)) is that the tests in `tabular` don't set the `time_limit` but the tests in `timeseries` do. In the latter case, the exception was encountered in [L240](https://github.com/autogluon/autogluon/blob/master/tabular/src/autogluon/tabular/models/lgb/callbacks.py#L240). Do you think we need to modify the tabular tests to address this?

mglowacki100:
@shchur Thanks for detailed info how to fix it!

Innixma:
Thanks @mglowacki100! Also, regarding the quantized error, the LightGBM developers are aware of it and have confirmed the issue: https://github.com/microsoft/LightGBM/issues/5982

I think we will need to wait for their next release where it is hopefully fixed to enable `use_quantized_grad`.

Regarding merging this PR, I will first want to run a benchmark to see if there are any performance regressions. I'm going to be on vacation until August 17th, so I will pick this up in the second half of August. I'll also plan to add unit test coverage for time_limit != None scenarios in Tabular as pointed out by @shchur. Thanks!

mglowacki100:
Hi @Innixma,
There is new version of LightGBM and I've re-checked bug https://github.com/microsoft/LightGBM/issues/5982, it seems to be solved at least for this example: https://colab.research.google.com/drive/1IYI9evSJa3OHTh0aZegN8Fq8aGIAYOmE?usp=sharing
So maybe it is a good moment to run test across more datasets with `use_quantized_grad`.

Btw. in the meanwhile there was:
- release of catboost 1.2.1 - it should be picked up automatically except darwin
"catboost": [
        # CatBoost wheel build is not working correctly on darwin for CatBoost 1.2, so use old version in this case.
        # https://github.com/autogluon/autogluon/pull/3190#issuecomment-1540599280
        'catboost>=1.1,<1.2 ; sys_platform == "darwin"',
        "catboost>=1.1,<1.3",
    ],
- release xgboost 2.0.0 - it will be not picked up as in setup.py:
"xgboost": [
        "xgboost>=1.6,<1.8",
    ],
I could make a PR regarding xgboost upgrade.

ddelange:
fwiw lightgbm v4 has a breaking change concerning early stopping rounds: https://github.com/microsoft/LightGBM/issues/5196#issuecomment-1702252267

and the darwin catboost line can be safely removed since they fixed darwin wheels publishing: https://github.com/autogluon/autogluon/pull/3190/files#diff-3caf337ab14835b6293e2c00467eeb4d72ff3e746f934d627bb0ba2314a30883

Innixma:
Re this PR, will plan to wait and see if LightGBM 4.2 releases by end of Oct with the `use_quantized_grad` [fix](https://github.com/microsoft/LightGBM/pull/6092) that is merged into their mainline.

If they don't release in time, we will plan to use LightGBM 4.1 for AutoGluon 1.0 and not use the quantization logic.

prateekdesai04:
Note: Just a heads up LightGBM version bump PR is already merged in main - https://github.com/autogluon/autogluon/pull/3709

Innixma:
Keeping this open in case LightGBM 4.2 releases

mglowacki100:
LightGBM 4.2.0 has been released: https://github.com/microsoft/LightGBM/releases/tag/v4.2.0

Innixma:
@mglowacki100 There is a bug in `lightgbm==4.2.0` causing segfaults when categoricals are present and qunatization is enabled. Will wait for response from devs: https://github.com/microsoft/LightGBM/issues/6257

Borda (approved):
Would be great to have this bump

mglowacki100:
@Borda There is 4.3 release (https://github.com/microsoft/LightGBM/releases/tag/v4.3.0) but I'm not sure if this was fixed: https://github.com/microsoft/LightGBM/issues/6257

Borda:
I would say unless there are some reasons why you need to keep `LightGBM` frozen, please update it, ideally with update and all passing CI it shall be fine to land

Innixma:
> I would say unless there are some reasons why you need to keep `LightGBM` frozen, please update it, ideally with update and all passing CI it shall be fine to land

Fair point! I think we can treat this PR as a verison bump. I've rebased it and updated the version bump to the latest LightGBM release. I've created #3917 to track `use_quantized_grad` support. Will merge once CI passes.

Innixma (approved):
LGTM! Thanks for the contribution @mglowacki100!

PR #3434, opened 2023-08-02
[AutoMM] Unify creating tokenizers for all models

zhiqiangdon:
*Issue #, if available:*

*Description of changes:*

- Previously, tokenizers were created in both model and data processor, which is difficult to maintain. This PR is to unify creating tokenizers for all models. Now tokenizers are created in the model initializations and then passed to data processors.
- Fixed the slow loading issue of M5 tokenizer. `use_fast` is only needed in one place.


tonyhoo (approved):
LGTM

PR #3444, opened 2023-08-07
PR to fix core platform tests

prateekdesai04:
*Description of changes:*

This PR is to fix the floating point calculations which causes the core platform tests to fail on some operating systems due to the way floating point numbers are handled.


yinweisu (approved):
LGTM! Thanks for the fix!

yinweisu (changes requested):
Mark as request changes just to make sure CI passed

PR #3445, opened 2023-08-08
[AutoMM] Use ddp_spawn for muti-GPU inference

zhiqiangdon:
*Issue #, if available:*

*Description of changes:*

- Deprecate `dp` from the default inference since lightning no longer supports it (https://github.com/Lightning-AI/lightning/pull/16748). Now we use the same strategy for both training and inference.
- Change the default strategy from `ddp_spawn` to `ddp_spawn_find_unused_parameters_true`. According to https://github.com/Lightning-AI/lightning/pull/16611, lightning set `find_unused_parameters=False` by default. 

TODOs:
- Evaluate the benefit of greedy soup. Two options: 1) turn it off; 2) keep it and use single gpu for evaluation.

### DIFF
--- a/multimodal/src/autogluon/multimodal/configs/environment/default.yaml
+++ b/multimodal/src/autogluon/multimodal/configs/environment/default.yaml
@@ -11,6 +11,6 @@ env:
   fast_dev_run: False
   deterministic: False
   auto_select_gpus: True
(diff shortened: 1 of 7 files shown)


tonyhoo (approved):
LGTM. Please make sure all tests passing

PR #3454, opened 2023-08-14
[timeseries] Automatically handle irregular timestamps and missing data in TimeSeriesPredictor

### LINKED ISSUE #3439: AttributeError when calling TimeSeriesPredictor.predict with known_covariates
**Description**:

I am encountering an issue when using the `TimeSeriesPredictor` in AutoGluon. The problem arises when I call the `predict` method with the `known_covariates` argument.

**Code**:

Here is a simplified version of my code:

```python
predictor = TimeSeriesPredictor.load("AutogluonModels/ag-20230804_203406/")
train_data = pd.read_csv('train_data.csv')  # train_data has a MultiIndex ('item_id', 'timestamp')
zukunft = pd.read_csv('zukunft.csv')  # zukunft has columns 'item_id', 'timestamp', and 'trading_active'
zukunft.set_index(['item_id', 'timestamp'], inplace=True)

predictions = predictor.predict(train_data, known_covariates=zukunft)
```

**Error**:

When running this, I encounter the following error:

```
AttributeError: 'DataFrame' object has no attribute 'item_ids'
```

The traceback points towards `_align_covariates_with_forecast_index` method in `learner.py`, which appears to try accessing `item_ids` as an attribute of the DataFrame, rather than as a part of the DataFrame's index.

```python
missing_item_ids = data.item_ids.difference(known_covariates.item_ids)
```

My `train_data` and `zukunft` DataFrames both use a MultiIndex with `item_id` and `timestamp`. It seems like the `predict` method is expecting `item_ids` to be a column in the DataFrame, not a part of the index.

**Attempts to resolve**:

I've tried adjusting the structure of my DataFrames and renaming columns, but I am still encountering this error. This makes me believe that the issue may be internal to the `predict` method or the `_align_covariates_with_forecast_index` method.

shchur:
*Description of changes:*

1. Refactor the logic for handling irregular timestamps in `TimeSeriesPredictor`.
    - Old logic (now deprecated):
        - If data contains irregular timestamps, user can set `TimeSeriesPredictor(..., ignore_time_index=True)`. In this case, timestamps for all data passed to the predictor will be replaced with dummy timestamps.
    - New logic (this PR):
        - If data contains irregular timestamps, user can set the desired frequency when creating the predictor `TimeSeriesPredictor(..., freq=freq)`. In this case, all data passed to the predictor will automatically be resampled to the desired frequency.
        - This can also be done outside the predictor with method `TimeSeriesDataFrame.convert_frequency`.
2. If data contains missing values represented by NaN, they are automatically filled by the predictor. 
4. Automatically convert `known_covariates` to a `TimeSeriesDataFrame`, if possible (fixes #3439)
5. Warn users if test data only includes future values and does not include historic values (fixes #3235)
6. Fixed typos in documentation


timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -187,57 +186,109 @@ def __init__(
...
+                    f"frequency when creating the predictor with `TimeSeriesPredictor(freq=...)` or ensure that "
+                    f"the data has a regular time index with `{name}.to_regular_index(freq=...)`"
+                )
+            else:
+                self.freq = df.freq
+                logger.info(f"Inferred data frequency: {df.freq}")
+        else:
+            if df.freq != self.freq:
+                logger.warning(f"{name} with frequency '{df.freq}' has been resampled to frequency '{self.freq}'.")
+                df = df.convert_frequency(freq=self.freq)
```
shchur, line 251: @tonyhoo Suppose user creates `TimeSeriesPredictor(freq="D")` and passes hourly data to it. Which of the two alternatives is more reasonable?
1. Automatically resample hourly data to the daily frequency + notify the user
2. Raise an exception saying that provided data is incompatible with daily frequency

tonyhoo (approved):

PR #3469, opened 2023-08-21
[timeseries] Automatically remove short time series from training data

shchur:
For the TimeSeriesPredictor to function normally, the time series in `train_data` must have length `>= (num_val_windows + 1) * prediction_length + 1`. This ensures that the "actual training set" for each validation fold has time series of length `>= prediction_length + 1`.

**Current solution (master branch)**: 
We warn the user if some time series are too short but then proceed with training. This has two disadvantages:
- User may experience model failures or see cryptic exceptions during the training process.
- Presence of very short time series in the training set can make validation scores extremely noisy because very little training data is available (e.g., this happens on the `cif_2016` dataset)

**Proposed solution (this PR)**: 
We automatically remove time series that are too short for the chosen `prediction_length` and `num_val_windows` from the training set. This ensures that all models can be trained normally and allows us to simplify the logic under the hood. We also take the burden of data preprocessing off the user.

*Description of changes:*
- If `train_data` contains time series that are too short for the `prediction_length` and `num_val_windows` provided by the user, they are automatically removed from `train_data`. 
- If user sets `predictor.fit(..., num_val_windows=None)`, `num_val_windows` is determined automatically based on the length of time series in `train_data`.
  - We need do benchmarking to determine if this should be the default
- Minor changes 
  - Improved log messages & docstrings
  - Fix small bugs related to edge cases in the preprocessing pipeline


timeseries/src/autogluon/timeseries/learner.py
```diff
@@ -149,7 +150,7 @@ def _align_covariates_with_forecast_index(
         missing_item_ids = data.item_ids.difference(known_covariates.item_ids)
         if len(missing_item_ids) > 0:
             raise ValueError(
-                f"known_covariates are missing information for the following item_ids: {missing_item_ids.to_list()}."
+                f"known_covariates are missing information for the following item_ids: {reprlib.repr(missing_item_ids.to_list())}."
```
tonyhoo, line 153: Thank you for utilizing `reprlib` to truncate lengthy inputs in this context.

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -294,47 +289,59 @@ def _check_data_for_evaluation(self, data: TimeSeriesDataFrame, name: str = "dat
...
+        ts_lengths = data.num_timesteps_per_item()
+        median_length = int(ts_lengths.median())
+        min_length = ts_lengths.min()
+        max_length = ts_lengths.max()
+        return (
+            f"{len(data)} rows, {data.num_items} time series. "
+            f"Median time series length is {median_length} (min={min_length}, max={max_length}). "
+        )
+
+    def _recommend_num_val_windows(self, train_data: TimeSeriesDataFrame, max_num_val_windows: int = 3) -> int:
```
tonyhoo, line 303: my personal opinion which is not backed by any data. Shall we set the `max_num_val_windows` default to 5 as the max allowed in Amazon Forecast is the 5 per the documentation: https://docs.aws.amazon.com/forecast/latest/dg/API_EvaluationParameters.html#forecast-Type-EvaluationParameters-NumberOfBacktestWindows

shchur: I also don't have any data and I agree that 5 is a reasonable default, updated this parameter.

tonyhoo (approved):

PR #3472, opened 2023-08-22
Fix for Tabular test timeout on Windows

prateekdesai04:
*Description of changes:*
Skips two tests which cause the timeout issue on Windows platforms.
Added a custom pytest marker.


tabular/tests/unittests/test_tabular.py
```diff
@@ -137,7 +137,6 @@ def test_advanced_functionality():
     savedir_predictor_original = savedir + "predictor/"
     predictor: TabularPredictor = TabularPredictor(label=label, path=savedir_predictor_original).fit(train_data)
     leaderboard = predictor.leaderboard(data=test_data)
-    predictor.plot_ensemble_model()
```
yinweisu, line 140: What's the failure here. We need to fix it instead of removing it

tabular/tests/unittests/test_tabular.py
```diff
@@ -1143,7 +1141,6 @@ def test_tabular_log_to_file():
     predictor = TabularPredictor(label="class", log_to_file=True, log_file_path=log_file).fit(train_data=train_data, hyperparameters={"DUMMY": {}})
     log = TabularPredictor.load_log(log_file_path=log_file)
     assert "TabularPredictor saved." in log[-1]
-    os.remove(log_file)
```
yinweisu, line 1146: ditto

tabular/tests/unittests/test_tabular.py
```diff
@@ -694,7 +692,7 @@ def test_pseudolabeling():
                     assert False, error_msg + "unlabeled test data, best quality" + error_suffix
 
 
-@pytest.mark.slow
+@pytest.mark.platform
```
yinweisu, line 695: We probably want to mark this test both slow and platform. Otherwise, this would change behavior in our regular CI

Issue #3353, opened 2023-06-26
Apply ruff linter


shchur:
This tool could be helpful for improving code quality and catching potential bugs 

https://github.com/astral-sh/ruff

Borda:
I would suggest to replace `flake8` with Ruff as Ruff is much faster and you can with the same tool covered isort linting or bandit issues... Happy to take it if you agree 🐰

PR #3476, opened 2023-08-23
[timeseries] Apply ruff linter to timeseries

shchur:
*Issue #, if available:* #3353

*Description of changes:*
- Apply ruff linter to code in `timeseries/`
- Add ruff lint check to GitHub action


yinweisu (approved):
LGTM! A more general question: what's the pros of switching to ruff linter? Had a quick look, and it seems just a faster lint? Is there any other benefits? Trying to understand this to see if we should apply it to other modules too

shchur:
So far we use `isort` that checks the order of imports and `black` that checks the style (indents, newlines and whitespaces). `ruff` is essentially a faster version of `flake8` and it additionally checks things like unused imports, unused variables, or undefined variables present in the code. This may catch some bugs (unlike `black` and `isort`).

PR #3480, opened 2023-08-24
Tabular: Fix train_test split edge-case

Innixma:
*Issue #, if available:*

*Description of changes:*

In cases where there are an extreme number of classes compared to data samples and the data is large, AutoGluon can crash during train test split due to a bug in scikit-learn's `train_test_split` function.

Example:

```
Train Data Rows:    1528329
Train Data Class Count: 27000
```

Because the test split would be 1% of the total data based on AutoGluon's split logic, it results in the test data having fewer rows than the number of classes, triggering the error:

```
ValueError: The test_size = 14915 should be greater or equal to the number of classes = 27000
```

While scikit-learn claims that test_size must be greater than the number of classes, this isn't technically required fundamentally. This PR adds a work-around for this limitation by disabling stratification during train test split if stratification would lead to the above error.

### DIFF
--- a/core/src/autogluon/core/utils/utils.py
+++ b/core/src/autogluon/core/utils/utils.py
@@ -472,9 +472,24 @@ def generate_train_test_split(
                 y_split = y.drop(index=rare_indices)
                 stratify = y_split
 
-    X_train, X_test, y_train, y_test = train_test_split(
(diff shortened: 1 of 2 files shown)


prateekdesai04 (approved):
LGTM!

PR #3484, opened 2023-08-25
[AutoMM] Support multi-gpu training in notebook

zhiqiangdon:
*Issue #, if available:*

*Description of changes:*

- Supported multi-gpu training in notebook. It can automatically detect interactive env and switch to use ddp_fork.
- Removed the AutoMMPrediction API as warned before.
- Did local tests on notebook with 4 gpus.

### DIFF
--- a/multimodal/src/autogluon/multimodal/predictor.py
+++ b/multimodal/src/autogluon/multimodal/predictor.py
@@ -129,7 +129,6 @@ from .utils import (
     evaluate_coco,
     extract_from_output,
     filter_hyperparameters,
-    get_available_devices,
     get_config,
     get_detection_classes,
     get_dir_ckpt_paths,
@@ -1289,7 +1288,7 @@ class MultiModalPredictor(ExportMixin):
                 validate_data=val_df,
                 val_use_training_mode=val_use_training_mode,
             )
-
+        print(f"cuda initialized  at line 1292: {torch.cuda.is_initialized()}")
         optimization_kwargs = dict(
             optim_type=config.optimization.optim_type,
             lr_choice=config.optimization.lr_choice,
@@ -1370,6 +1369,7 @@ class MultiModalPredictor(ExportMixin):
 
         logger.debug(f"validation_metric_name: {task.validation_metric_name}")
         logger.debug(f"minmax_mode: {minmax_mode}")
+        print(f"cuda initialized  at line 1373: {torch.cuda.is_initialized()}")
 
         checkpoint_callback = AutoMMModelCheckpoint(
             dirpath=save_path,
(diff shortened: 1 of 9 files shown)


multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -1772,21 +1778,23 @@ def _default_predict(
...
             evaluator = pl.Trainer(
                 accelerator="gpu" if num_gpus > 0 else "auto",
-                devices=get_available_devices(num_gpus=num_gpus, auto_select_gpus=self._config.env.auto_select_gpus),
+                devices=num_gpus if num_gpus > 0 else "auto",
                 num_nodes=self._config.env.num_nodes,
                 precision=precision,
                 strategy=strategy,
                 benchmark=False,
-                enable_progress_bar=self._enable_progress_bar,
+                enable_progress_bar=False if barebones else self._enable_progress_bar,
```
suzhoum, line 1792: I think this is disabled on lightning side so we don't have to apply the logic explicitly. https://lightning.ai/docs/pytorch/stable/common/trainer.html

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -2854,12 +2863,13 @@ def list_supported_models(self, pretrained=True):
...
-    def __init__(self, **kwargs):
-        warnings.warn(
-            "AutoMMPredictor has been renamed as 'MultiModalPredictor'. "
-            "Consider to use MultiModalPredictor instead. Using AutoMMPredictor will "
-            "raise an exception starting in v0.7."
-        )
-        super(AutoMMPredictor, self).__init__(**kwargs)
+    def update_strategy_by_env(self):
+        """
+        Set strategy to ddp_for or ddp_notebook if an iterative env is detected.
```
suzhoum, line 2868: "ddp_fork"

multimodal/src/autogluon/multimodal/utils/log.py
```diff
@@ -200,10 +200,13 @@ def get_gpu_message(detected_num_gpus: int, used_num_gpus: int):
     A string with the GPU info.
     """
     gpu_message = f"{detected_num_gpus} GPUs are detected, and {used_num_gpus} GPUs will be used.\n"
-    for i in range(detected_num_gpus):
-        free_memory, total_memory = torch.cuda.mem_get_info(i)
-        gpu_message += f"   - GPU {i} name: {torch.cuda.get_device_name(i)}\n"
-        gpu_message += f"   - GPU {i} memory: {free_memory * 1e-9:.2f}GB/{total_memory * 1e-9:.2f}GB (Free/Total)\n"
+    if not any(s in strategy for s in ["fork", "notebook"]):
```
suzhoum, line 203: Do we not want any logs in notebook?

zhiqiangdon: Calling `torch.cuda.mem_get_info()` initializes CUDA, which would trigger one pytorch error "Cannot re-initialize CUDA in forked subprocess".

multimodal/src/autogluon/multimodal/predictor.py
zhiqiangdon: Thanks. Corrected.

multimodal/src/autogluon/multimodal/predictor.py
zhiqiangdon: I found that It would throw an error if both enable_progress_bar and barebones are set True.

multimodal/src/autogluon/multimodal/utils/log.py
zhiqiangdon: See https://github.com/Lightning-AI/lightning/blob/6b7abda5a3be5ee3204c8e1ca5a8cd644e768ba4/src/lightning/fabric/strategies/launchers/multiprocessing.py#L186-L203

suzhoum: I see.

suzhoum (approved):
LGTM!

PR #3491, opened 2023-08-28
More meaningful HPO error msg without ray

yinweisu:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/common/src/autogluon/common/utils/try_import.py
+++ b/common/src/autogluon/common/utils/try_import.py
@@ -53,7 +53,7 @@ def try_import_ray() -> ModuleType:
         return ray
     except ImportError:
         raise ImportError(
-            "ray is required to train folds in parallel / HPO. "
+            "ray is required to train folds in parallel for TabularPredictor or HPO for MultiModalPredictor. "
             f"A quick tip is to install via `pip install ray=={ray_min_version}`"
         )
 


common/src/autogluon/common/utils/try_import.py
```diff
@@ -53,9 +53,9 @@ def try_import_ray() -> ModuleType:
         return ray
     except ImportError:
         raise ImportError(
-            "ray is required to train folds in parallel. "
+            "ray is required to train folds in parallel / HPO. "
             f"A quick tip is to install via `pip install ray=={ray_min_version}`, "
-            "or use sequential fold fitting by passing `sequential_local` to `ag_args_ensemble` when calling tabular.fit"
+            "or use sequential fold fitting by passing `sequential_local` to `ag_args_ensemble` when calling TabularPredictor fit()"
```
zhiqiangdon, line 58: Can we make this message more general? Now it mainly mentions tabular predictor and ensemble. It can confuse the multimodal predictor users.

multimodal/src/autogluon/multimodal/presets.py
```diff
@@ -1,5 +1,7 @@
 from typing import List, Optional
 
+from autogluon.common.utils.try_import import try_import_ray
```
zhiqiangdon, line 3: I think no need to add `try_import_ray` in presets.py. The one in predictor.py can always catch the error for HPO.

zhiqiangdon (approved):
LGTM!

PR #3498, opened 2023-08-29
Upgrade pandas and scikit-learn

yinweisu:
*Issue #, if available:*

*Description of changes:*
* Upgrade pandas and scikit learn
* Also upgraded skl2onnx and onnxruntime as older version not compatible with latest scikit learn
* Changed `pd.to_datetime` to use `mixed` format, which is no longer the default: https://pandas.pydata.org/docs/dev/whatsnew/v2.0.0.html#datetimes-are-now-parsed-with-a-consistent-format
* Updated onnx operators to allign with new api


shchur:
I've checked locally that the tests pass for both pandas v2.0.x and v2.1.x (since 2.1 also introduced some breaking changes), so the PR is good to go from the time series side.

tonyhoo (approved):
LGTM! Please merge after all tests passed

PR #3510, opened 2023-09-07
[timeseries] Add RMSSE

melopeo:
*Description of changes:*

Add RMSSE metric as defined in https://mofc.unic.ac.cy/m5-competition/

### DIFF
--- a/timeseries/src/autogluon/timeseries/evaluator.py
+++ b/timeseries/src/autogluon/timeseries/evaluator.py
@@ -46,6 +46,11 @@ def mape_per_item(*, y_true: pd.Series, y_pred: pd.Series) -> pd.Series:
     return ((y_true - y_pred) / y_true).abs().groupby(level=ITEMID, sort=False).mean()
 
 
+def rmsse_per_item(*, y_true: pd.Series, y_pred: pd.Series, random_walk_error: pd.Series) -> pd.Series:
+    mse = mse_per_item(y_true=y_true, y_pred=y_pred)
+    return np.sqrt(mse / random_walk_error)
+
+
 def symmetric_mape_per_item(*, y_true: pd.Series, y_pred: pd.Series) -> pd.Series:
     """Compute symmetric Mean Absolute Percentage Error for each item (time series)."""
     return (2 * (y_true - y_pred).abs() / (y_true.abs() + y_pred.abs())).groupby(level=ITEMID, sort=False).mean()
@@ -106,7 +111,7 @@ class TimeSeriesEvaluator:
         :meth:``~autogluon.timeseries.TimeSeriesEvaluator.check_get_evaluation_metric``.
     """
 
-    AVAILABLE_METRICS = ["MASE", "MAPE", "sMAPE", "mean_wQuantileLoss", "MSE", "RMSE", "WAPE"]
+    AVAILABLE_METRICS = ["MASE", "MAPE", "sMAPE", "mean_wQuantileLoss", "MSE", "RMSE", "WAPE", "RMSSE"]
     METRIC_COEFFICIENTS = {
         "MASE": -1,
         "MAPE": -1,
@@ -115,6 +120,7 @@ class TimeSeriesEvaluator:
         "MSE": -1,
         "RMSE": -1,
         "WAPE": -1,
+        "RMSSE": -1,
     }
     DEFAULT_METRIC = "mean_wQuantileLoss"
 
@@ -182,6 +188,11 @@ class TimeSeriesEvaluator:
         abs_error_sum = (mae_per_item(y_true=y_true, y_pred=y_pred) * self.prediction_length).sum()
         abs_target_sum = y_true.abs().sum()
         return abs_error_sum / abs_target_sum
+
+    def _rmsse(self, y_true: pd.Series, predictions: TimeSeriesDataFrame) -> float:
+        y_pred = self._get_median_forecast(predictions)
+        rmsse = rmsse_per_item(y_true, y_pred, self._random_walk_error)
+        return rmsse.mean()
 
     def _get_median_forecast(self, predictions: TimeSeriesDataFrame) -> pd.Series:
         # TODO: Median forecast doesn't actually minimize the MAPE / sMAPE losses
--- a/timeseries/src/autogluon/timeseries/evaluator.py
+++ b/timeseries/src/autogluon/timeseries/evaluator.py
@@ -22,12 +22,12 @@ def get_seasonal_diffs(*, y_past: pd.Series, seasonal_period: int = 1) -> pd.Ser
 
 def in_sample_seasonal_naive_error(*, y_past: pd.Series, seasonal_period: int = 1) -> pd.Series:
     """Compute seasonal naive forecast error (predict value from seasonal_period steps ago) for each time series."""
-    seasonal_diffs = get_seasonal_diffs(y_past, seasonal_period=seasonal_period)
+    seasonal_diffs = get_seasonal_diffs(y_past=y_past, seasonal_period=seasonal_period)
     return seasonal_diffs.groupby(level=ITEMID, sort=False).mean().fillna(1.0)
 
 
 def in_sample_random_walk_error(*, y_past: pd.Series) -> pd.Series:
-    onestep_diffs = get_seasonal_diffs(y_past, seasonal_period=1)
+    onestep_diffs = get_seasonal_diffs(y_past=y_past, seasonal_period=1)
     return onestep_diffs.dropna().pow(2.0).groupby(level=ITEMID, sort=False).mean()
 
 
@@ -140,6 +140,7 @@ class TimeSeriesEvaluator:
 
         self.metric_method = self.__getattribute__("_" + self.eval_metric.lower())
         self._past_naive_error: Optional[pd.Series] = None
+        self._random_walk_error: Optional[pd.Series] = None
 
     @property
     def coefficient(self) -> int:
@@ -191,8 +192,7 @@ class TimeSeriesEvaluator:
 
     def _rmsse(self, y_true: pd.Series, predictions: TimeSeriesDataFrame) -> float:
         y_pred = self._get_median_forecast(predictions)
-        rmsse = rmsse_per_item(y_true, y_pred, self._random_walk_error)
-        return rmsse.mean()
+        return self._safemean(rmsse_per_item(y_true=y_true, y_pred=y_pred, random_walk_error=self._random_walk_error))
 
     def _get_median_forecast(self, predictions: TimeSeriesDataFrame) -> pd.Series:
         # TODO: Median forecast doesn't actually minimize the MAPE / sMAPE losses
--- a/timeseries/src/autogluon/timeseries/evaluator.py
+++ b/timeseries/src/autogluon/timeseries/evaluator.py
@@ -89,6 +89,7 @@ class TimeSeriesEvaluator:
         * ``MSE``: mean squared error
         * ``RMSE``: root mean squared error
         * ``WAPE``: weighted absolute percentage error. See https://docs.aws.amazon.com/forecast/latest/dg/metrics.html#metrics-WAPE
+        * ``RMSSE``: Root Mean Squared Scaled Error . See https://mofc.unic.ac.cy/m5-competition/
 
     prediction_length : int
         Length of the forecast horizon
--- a/timeseries/src/autogluon/timeseries/evaluator.py
+++ b/timeseries/src/autogluon/timeseries/evaluator.py
@@ -16,9 +16,13 @@ from autogluon.timeseries.utils.warning_filters import evaluator_warning_filter
 logger = logging.getLogger(__name__)
 
 
+def get_seasonal_diffs(*, y_past: pd.Series, seasonal_period: int = 1) -> pd.Series:
+    return y_past.groupby(level=ITEMID, sort=False).diff(seasonal_period).abs()
+
+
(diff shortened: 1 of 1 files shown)


timeseries/src/autogluon/timeseries/evaluator.py
```diff
@@ -16,12 +16,21 @@
...
 def in_sample_seasonal_naive_error(*, y_past: pd.Series, seasonal_period: int = 1) -> pd.Series:
     """Compute seasonal naive forecast error (predict value from seasonal_period steps ago) for each time series."""
-    seasonal_diffs = y_past.groupby(level=ITEMID, sort=False).diff(seasonal_period).abs()
+    seasonal_diffs = get_seasonal_diffs(y_past=y_past, seasonal_period=seasonal_period)
     return seasonal_diffs.groupby(level=ITEMID, sort=False).mean().fillna(1.0)
 
 
+def in_sample_random_walk_error(*, y_past: pd.Series) -> pd.Series:
+    onestep_diffs = get_seasonal_diffs(y_past=y_past, seasonal_period=1)
+    return onestep_diffs.dropna().pow(2.0).groupby(level=ITEMID, sort=False).mean()
```
shchur, line 31: 1. I think we should use the seasonal error for both MASE and RMSSE (https://otexts.com/fpp3/accuracy.html#scaled-errors), i.e., not only `seasonal_period=1` for RMSSE.
2. Should we call the methods something like `in_sample_abs_seasonal_error` and `in_sample_squared_seasonal_error` to differentiate between the two cases?

timeseries/src/autogluon/timeseries/evaluator.py
```diff
@@ -219,6 +240,9 @@ def save_past_metrics(self, data_past: TimeSeriesDataFrame):
             y_past=data_past[self.target_column], seasonal_period=seasonal_period
         )
 
+        if self.eval_metric == "RMSSE":
```
shchur, line 243: Should we add a similar check above and only compute the `_past_naive_error` in case `eval_metric` is MASE?

timeseries/src/autogluon/timeseries/evaluator.py
```diff
@@ -174,6 +191,10 @@ def _wape(self, y_true: pd.Series, predictions: TimeSeriesDataFrame) -> float:
         abs_target_sum = y_true.abs().sum()
         return abs_error_sum / abs_target_sum
 
+    def _rmsse(self, y_true: pd.Series, predictions: TimeSeriesDataFrame) -> float:
+        y_pred = self._get_median_forecast(predictions)
```
shchur, line 195: I think we should use the "mean" forecast here since RMSSE is minimized by the conditional mean, not median (similar to RMSE, as described [here](https://otexts.com/fpp3/accuracy.html#scale-dependent-errors))

timeseries/src/autogluon/timeseries/evaluator.py
```diff
@@ -219,6 +240,9 @@ def save_past_metrics(self, data_past: TimeSeriesDataFrame):
             y_past=data_past[self.target_column], seasonal_period=seasonal_period
         )
 
+        if self.eval_metric == "RMSSE":
+            self._random_walk_error = in_sample_random_walk_error(y_past=data_past[self.target_column])
```
shchur, line 244: Similar to the above discussion, something like `_past_abs_seasonal_error` and `_past_squared_seasonal_error`?

timeseries/src/autogluon/timeseries/evaluator.py
shchur: This would also require changing the test https://github.com/autogluon/autogluon/blob/0a00e0652321e67e8b4f9e26eac8bdef21cd4628/timeseries/tests/unittests/test_evaluator.py#L196

Maybe add a similar test that ensures that squared error is computed for RMSSE?

timeseries/src/autogluon/timeseries/evaluator.py
```diff
@@ -75,6 +89,7 @@ class TimeSeriesEvaluator:
         * ``MSE``: mean squared error
         * ``RMSE``: root mean squared error
         * ``WAPE``: weighted absolute percentage error. See https://docs.aws.amazon.com/forecast/latest/dg/metrics.html#metrics-WAPE
+        * ``RMSSE``: Root Mean Squared Scaled Error . See https://mofc.unic.ac.cy/m5-competition/
```
shchur, line 92: - Maybe we can point to https://otexts.com/fpp3/accuracy.html#scaled-errors for both RMSSE and MASE?
- Can you please add the documentation also to `TimeSeriesPredictor`?

timeseries/src/autogluon/timeseries/evaluator.py
```diff
@@ -37,6 +46,11 @@ def mape_per_item(*, y_true: pd.Series, y_pred: pd.Series) -> pd.Series:
     return ((y_true - y_pred) / y_true).abs().groupby(level=ITEMID, sort=False).mean()
 
 
+def rmsse_per_item(*, y_true: pd.Series, y_pred: pd.Series, random_walk_error: pd.Series) -> pd.Series:
+    mse = mse_per_item(y_true=y_true, y_pred=y_pred)
+    return np.sqrt(mse / random_walk_error)
```
shchur, line 51: For `RMSE` we apply `sqrt` after averaging over items to be compatible with `GluonTS`. I'm not sure if this is the best option, but probably we would want to be consistent between RMSE and RMSSE, so we either need to modify our RMSE definition (+ respective test that checks parity with GluonTS), or perform sqrt after averaging in RMSSE.

timeseries/src/autogluon/timeseries/evaluator.py
```diff
@@ -37,6 +46,11 @@ def mape_per_item(*, y_true: pd.Series, y_pred: pd.Series) -> pd.Series:
     return ((y_true - y_pred) / y_true).abs().groupby(level=ITEMID, sort=False).mean()
 
 
+def rmsse_per_item(*, y_true: pd.Series, y_pred: pd.Series, random_walk_error: pd.Series) -> pd.Series:
```
shchur, line 49: Is there any way that we can add a test that ensures the correctness of our RMSSE implementation over time?

timeseries/src/autogluon/timeseries/evaluator.py
melopeo: Agreed and updated!

timeseries/src/autogluon/timeseries/evaluator.py
melopeo: This is an interesting point. According to M5 competition, the aggregated MRSSE is a weighted average of the item ones:

> After estimating the RMSSE for all the 42,840 time series of the competition, the participating methods will be ranked using the Weighted RMSSE (WRMSSE), as described latter in this Guide, using the following formula:
> WRMSSE=∑_(i=1)^42,840▒〖w_i*RMSSE〗,
> where w_i is the weight of the i_th series of the competition. A lower WRMSSE score is better.

timeseries/src/autogluon/timeseries/evaluator.py
melopeo: Agreed and updated.

timeseries/src/autogluon/timeseries/evaluator.py
melopeo: Agreed and updated.

timeseries/src/autogluon/timeseries/evaluator.py
melopeo: I did a very small updated on the test, but I am not sure if I understand enough what needs to be updated here. 
What I did update is `score_with_saved_past_metrics`, given that certain attributed are stored only if certain `eval_metrics` are used.

timeseries/src/autogluon/timeseries/evaluator.py
melopeo: Agreed and updated.

timeseries/src/autogluon/timeseries/evaluator.py
shchur: Nit: Should we call the variable `seasonal_diffs` now?

timeseries/src/autogluon/timeseries/evaluator.py
```diff
@@ -215,10 +238,15 @@ def check_get_evaluation_metric(
 
     def save_past_metrics(self, data_past: TimeSeriesDataFrame):
         seasonal_period = get_seasonality(data_past.freq) if self.seasonal_period is None else self.seasonal_period
-        self._past_naive_error = in_sample_seasonal_naive_error(
+        self._past_abs_seasonal_error = in_sample_abs_seasonal_error(
             y_past=data_past[self.target_column], seasonal_period=seasonal_period
         )
```
shchur, line 243: For the first point, I was suggesting the following change
```suggestion
        if self.eval_metric == "MASE":
            self._past_abs_seasonal_error = in_sample_abs_seasonal_error(
                y_past=data_past[self.target_column], seasonal_period=seasonal_period
            )
```
This would ensure that we only compute the `_past_abs_seasonal_error` for the metrics that need it (i.e. MASE), and don't spend time on this computation for other metrics.

timeseries/src/autogluon/timeseries/evaluator.py
```diff
@@ -228,7 +256,15 @@ def score_with_saved_past_metrics(
         it doesn't require splitting the test data into past/future portions each time (e.g., when fitting ensembles).
         """
         assert (predictions.num_timesteps_per_item() == self.prediction_length).all()
-        assert self._past_naive_error is not None, "Call save_past_metrics before score_with_saved_past_metrics"
+        if self.eval_metric == "MASE":
```
shchur, line 259: Nit: a shorter way to check this would be
```python
if self.eval_metric == "MASE" and self._past_abs_seasonal_error is None:
    raise AssertionError(...)
```

timeseries/tests/unittests/test_evaluator.py
```diff
@@ -192,7 +192,7 @@ def test_given_unavailable_input_and_no_raise_check_get_eval_metric_output_defau
     )
 
 
-@pytest.mark.parametrize("eval_metric", ["MASE", "sMAPE"])
+@pytest.mark.parametrize("eval_metric", ["MASE", "RMSSE"])
```
shchur, line 195: Nice!

timeseries/src/autogluon/timeseries/evaluator.py
shchur: I mean we could pre-compute the RMSSE values for 1 or 2 simple examples (e.g., using a simple for-loop in the test) and add a test that ensures that the output of the TimeSeriesEvaluator matches this naive implementation.

This would ensure that in the future we don't accidentally break our RMSSE calculation in the evaluator because of some minor change.

timeseries/src/autogluon/timeseries/evaluator.py
shchur: I also feel like sqrt before averaging is the more reasonable thing to do, but I would then update the RMSE to do the same.

timeseries/src/autogluon/timeseries/evaluator.py
melopeo: Done!

timeseries/src/autogluon/timeseries/evaluator.py
melopeo: Done!

timeseries/src/autogluon/timeseries/evaluator.py
melopeo: Done!

timeseries/src/autogluon/timeseries/evaluator.py
melopeo: I updated RMSSE to follow the current paradigm.

timeseries/src/autogluon/timeseries/evaluator.py
melopeo: I added tests for this.

timeseries/tests/unittests/common.py
```diff
@@ -32,22 +32,31 @@ def get_data_frame_with_item_index(
...
+    start_date: str = "2022-01-01",
+    columns: List[str] = ["target"],
+    data_generation: str = "random",
 ):
+    assert data_generation in ["random", "sequential"]
+    if data_generation == "random":
+        data = [random.random() for _ in range(len(item_list) * data_length)]
+    elif data_generation == "sequential":
+        data = [e for e in range(len(item_list) * data_length)]
+
```
shchur, line 44: This is very nice!

timeseries/tests/unittests/test_evaluator.py
```diff
@@ -204,7 +204,33 @@ def test_given_historic_data_not_cached_when_scoring_then_exception_is_raised(ev
...
+        columns=["mean"],
+        data_generation="sequential",
+    )
+    ag_evaluator = TimeSeriesEvaluator(
+        eval_metric="RMSSE",
+        prediction_length=prediction_length,
+        eval_metric_seasonal_period=eval_metric_seasonal_period,
+    )
+    ag_value = ag_evaluator(data, predictions)
+    print(f"ag_value:{ag_value}")
```
shchur, line 235: Nit: I guess this can be removed?

shchur (approved):
Looks great, thank you!

timeseries/tests/unittests/test_evaluator.py
melopeo: Yes, just removed it.

PR #3511, opened 2023-09-07
Dynamic folds

yinweisu:
*Issue #, if available:*
https://github.com/autogluon/autogluon/issues/3492

*Description of changes:*
* Dynamically adjust number of folds to be trained in parallel based on memory available

### DIFF
--- a/core/tests/unittests/test_parallel_local_folding.py
+++ b/core/tests/unittests/test_parallel_local_folding.py
@@ -15,7 +15,7 @@ from autogluon.core.searcher import LocalRandomSearcher
 
 class DummyBigModel(AbstractModel):
     def _estimate_memory_usage(self, **kwargs):
-        return ResourceManager.get_available_virtual_mem() / 2.5
+        return 1e+9
 
 
 def _prepare_data():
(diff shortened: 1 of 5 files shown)


core/tests/unittests/test_parallel_local_folding.py
```diff
@@ -88,3 +95,10 @@ def test_resource_allocation_and_time_limit():
     for i in range(num_iterations):
         config = searcher.get_config()
         _test_resource_allocation_and_time_limit(**config)
+
+
+@patch("autogluon.common.utils.resource_utils.ResourceManager.get_available_virtual_mem")
+def test_dynamic_resource_allocation(mock_get_mem):
+    mock_get_mem.return_value = 1073741824
+    fold_fitting_strategy = _construct_dummy_fold_strategy(model_base_cls=DummyBigModel, num_jobs=100, num_folds_parallel=8)
+    assert fold_fitting_strategy.num_parallel_jobs == 2
```
Innixma, line 104: Can you add a unit test where we are training 8 folds, have memory for 7.5, we train two sets of 4

Add a unit test where we are training 10 folds, we have memory for 6, we train 4, 4, 2. (Also add comment as to why it is doing this)

yinweisu: Good point! Added

Innixma (approved):
LGTM after I added updates, benchmark shows that results are stable and similar to mainline, thanks @yinweisu for the contribution!

PR #3513, opened 2023-09-08
[timeseries] Remove statsmodels models

shchur:
*Description of changes:*
- Remove all models from `statsmodels` since these are (1) slower than their counterparts from StatsForecast and (2) not used in any presets.


melopeo:
Quick question: this PR is mainly addressing to move models `ARIMAModel, ETSModel, Theta` from `statsmodels` to `statsforecast`. This seems to be suggested by [this](https://github.com/autogluon/autogluon/pull/3513/files#diff-c4e23c6054af5bb74200daf97e42d6b78cee2e4b3efce23c32b1f9d9e1667ef0L5) line. Is this correct?

timeseries/src/autogluon/timeseries/models/local/statsforecast.py
```diff
@@ -178,6 +246,41 @@ def _get_model_type(self):
         return AutoETS
 
 
+class ETSModel(AutoETSModel):
```
melopeo, line 249: Why is `ETSModel` designed with parent class `AutoETSModel`, whereas `ARIMAModel` has as parent class `AbstractStatsForecastModel`?

docs/tutorials/timeseries/index.md
```diff
@@ -6,7 +6,7 @@ and does not require you to manually deal with cumbersome issues like model sele
 
 Under the hood, AutoGluon combines various state of the art forecasting algorithms.
 These include established statical methods like ETS and ARIMA from
-[`StatsForecast`](https://github.com/Nixtla/statsforecast) and [`statsmodels`](https://statsmodels.org/) libraries,
+[`StatsForecast`](https://github.com/Nixtla/statsforecast),
```
melopeo, line 9: My apologies for this question. Is there anywhere in AGTS documented what are the models available to the user?

melopeo: Ah, it is here!
http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-3513/426f107/tutorials/timeseries/forecasting-model-zoo.html

timeseries/src/autogluon/timeseries/models/local/statsforecast.py
shchur: StatsForecast doesn't have a dedicated class for the ETS model but has one for ARIMA. The `ETSModel` class is [deprecated](https://github.com/Nixtla/statsforecast/blob/main/statsforecast/models.py#L828) and it's just a wrapper for AutoETS (similar to our code).

My idea was to provide `ETSModel` as a convenience for users who used the `ETSModel` from statsmodels before.

docs/tutorials/timeseries/index.md
shchur: Thanks, that's a really good point! I will remove the references from Model Zoo as well.

shchur: Actually, it seems that the references to the models got updated automatically and there is nothing to change there 😄

melopeo: Yes, it is interesting :)

PR #3514, opened 2023-09-08
[timeseries] Avoid importing pytorch_lightning at top level

shchur:
A recent update to `pytorch_lightning` has potential to introduce a massive performance regression to forecasting models in AutoGluon that rely on multiprocessing / joblib. As of `v2.0.8` of `pytorch_lightning`, a multiprocessing lock is created whenever the module is imported. This increased the process initialization time significantly.

For example, total runtime of the above code snippet
```python
import pytorch_lightning as pl

data = TimeSeriesDataFrame('https://autogluon.s3.amazonaws.com/datasets/timeseries/m4_hourly/train.csv')
predictor = TimeSeriesPredictor()
predictor.fit(data, hyperparameters={"Naive": {}})
```

- current `master` branch: 1m 42s
- with this PR: 36s.

*Description of changes:*
- Hide all imports of `pytorch_lightning` and `torch` inside class methods. This ensures that they are not imported by the processes spawned by joblib.
- Create a separate method for setting the random seed that does not rely on `pytorch_lightning`.

### DIFF
--- a/common/src/autogluon/common/utils/utils.py
+++ b/common/src/autogluon/common/utils/utils.py
@@ -184,6 +184,7 @@ def seed_everything(seed: int) -> None:
     np.random.seed(seed)
     try:
         import torch
+
         torch.manual_seed(seed)
     except ImportError:
         pass
--- a/common/src/autogluon/common/utils/utils.py
+++ b/common/src/autogluon/common/utils/utils.py
@@ -7,6 +7,7 @@ from hashlib import md5
 from pathlib import Path
 from typing import Any, Dict, Optional
 
+import numpy as np
 import pandas as pd
 
 from ..version import __version__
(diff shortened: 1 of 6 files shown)


common/src/autogluon/common/utils/utils.py
```diff
@@ -176,3 +177,14 @@ def hash_pandas_df(df: Optional[pd.DataFrame]) -> str:
...
+
+def seed_everything(seed: int) -> None:
+    """Set random seeds for numpy and PyTorch."""
+    np.random.seed(seed)
+    try:
+        import torch
+
+        torch.manual_seed(seed)
+    except ImportError:
+        pass
```
shchur, line 190: @yinweisu do we have this functionality anywhere else in `core` or `common`, or is it fine to add it here?

yinweisu: I think it's fine to add it here

melopeo (approved):

PR #3520, opened 2023-09-11
[AutoMM] Support torch.compile()

zhiqiangdon:
*Issue #, if available:*

*Description of changes:*
Users can turn on compiling their models through `predictor.fit(hyperparameters={"env.compile.turn_on", True})`.

Limitations
- Doesn't work for multi-gpu training.
- Needs pytorch version >= `2.2.0.dev20230908`

### DIFF
--- a/multimodal/src/autogluon/multimodal/predictor.py
+++ b/multimodal/src/autogluon/multimodal/predictor.py
@@ -79,6 +79,7 @@ from .constants import (
     SCORE,
     TEXT,
     TEXT_NER,
+    TORCH_COMPILE_MIN_VERSION,
     UNIFORM_SOUP,
     XYWH,
     Y_PRED,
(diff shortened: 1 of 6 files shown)


multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -1151,6 +1152,19 @@ def _fit(
...
+            assert version.parse(torch.__version__) >= version.parse(TORCH_COMPILE_MIN_VERSION), (
+                f"torch.compile requires torch version >= {TORCH_COMPILE_MIN_VERSION}, "
+                f"but torch version {torch.__version__} is detected."
+            )
+            logger.debug("Using torch.compile() in compiling the model.")
+            model = torch.compile(
+                model,
+                mode=OmegaConf.select(config, "env.compile.mode", default="default"),
+                dynamic=OmegaConf.select(config, "env.compile.dynamic", default=True),
+                backend=OmegaConf.select(config, "env.compile.backend", default="inductor"),
```
suzhoum, line 1165: Do we want to use the default values in `default.yaml` instead? It can help in case we want to update the default values in future.

suzhoum (approved):
LGTM with a minor comment.

multimodal/src/autogluon/multimodal/predictor.py
zhiqiangdon: Yes, `OmegaConf.select()` prefers using the values in default.yaml. Adding the default values in the function call is for backward compatibility, i.e., users load previously saved predictor which doesn't have these keys. Without these default values, it would throw an exception. See the doc: https://omegaconf.readthedocs.io/en/2.3_branch/usage.html#omegaconf-select

suzhoum: I see. That makes sense.

PR #3521, opened 2023-09-13
Set bf16 AMP fallback to fp16 AMP

ddelange:
*Issue #, if available:*

*Description of changes:*

Hi 👋 We noticed a fallback to 32 bit floats on older T4 GPUs (g4dn family) for `presets='multilingual'`. This PR sets that fallback to fp16 AMP instead of falling back to 32.

This way, we still have fast training on the older machines. Or was there some reason to fall back to 32?

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice. ✅

### DIFF
--- a/multimodal/src/autogluon/multimodal/utils/environment.py
+++ b/multimodal/src/autogluon/multimodal/utils/environment.py
@@ -100,10 +100,10 @@ def infer_precision(
             warnings.warn(
                 "bf16 is not supported by the GPU device / cuda version. "
                 "Consider using GPU devices with versions after Amphere or upgrading cuda to be >=11.0. "
-                "MultiModalPredictor is switching precision from bf16 to 32.",
+                "MultiModalPredictor is switching precision from bf16 AMP to fp16 AMP.",
                 UserWarning,
             )
-            precision = 32
+            precision = 16
 
     if as_torch:
         precision_mapping = {


ddelange:
@zhiqiangdon taking the liberty to ping you from the blame :)

multimodal/src/autogluon/multimodal/utils/environment.py
```diff
@@ -100,10 +100,10 @@ def infer_precision(
             warnings.warn(
                 "bf16 is not supported by the GPU device / cuda version. "
                 "Consider using GPU devices with versions after Amphere or upgrading cuda to be >=11.0. "
-                "MultiModalPredictor is switching precision from bf16 to 32.",
+                "MultiModalPredictor is switching precision from bf16 AMP to fp16 AMP.",
                 UserWarning,
             )
-            precision = 32
```
zhiqiangdon, line 106: I remember we set it to 32 on purpose because a model pretrained with bf16 may be difficult to converge if finetuned on fp16. bf16 can represent larger numbers than fp16, which can cause fp16 to overflow.

ddelange: thanks for the quick reply and explanation, will close with that 👍

PR #3536, opened 2023-09-19
[timeseries] Fix smoketests for time series

shchur:
*Description of changes:*
- Gracefully handle the case when `ETSModel` and `AutoETSModel` fail if time series length < 2 * `season_length`.
- Disable damped trend for `ETSModel` (for compatibility with the `ETSModel` from statsmodels that was replaced in #3513).

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/local/statsforecast.py
+++ b/timeseries/src/autogluon/timeseries/models/local/statsforecast.py
@@ -245,6 +245,17 @@ class AutoETSModel(AbstractStatsForecastModel):
 
         return AutoETS
 
+    def _predict_with_local_model(
+        self,
+        time_series: pd.Series,
+        local_model_args: dict,
+    ) -> pd.DataFrame:
+        # Disable seasonality if time series too short for chosen season_length, otherwise model will crash
+        if len(time_series) < 2 * local_model_args["season_length"]:
+            # changing last character to "N" disables seasonality, e.g., model="AAA" -> model="AAN"
+            local_model_args["model"] = local_model_args["model"][:-1] + "N"
+        return super()._predict_with_local_model(time_series=time_series, local_model_args=local_model_args)
+
 
 class ETSModel(AutoETSModel):
     """Exponential smoothing with trend and seasonality.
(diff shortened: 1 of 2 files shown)


timeseries/src/autogluon/timeseries/models/local/statsforecast.py
```diff
@@ -245,6 +245,17 @@ def _get_model_type(self):
 
         return AutoETS
 
+    def _predict_with_local_model(
```
melopeo, line 248: where is this used? If I understand correctly this is a new functionality, but it is unclear where this is used.

timeseries/tests/smoketests/test_features_and_covariates.py
```diff
@@ -79,8 +79,8 @@ def test_predictor_smoke_test(
         "TemporalFusionTransformer": DUMMY_MODEL_HPARAMS,
         "Theta": {},
         # Override default hyperparameters for faster training
-        "AutoARIMA": {"max_p": 2},
-        "AutoETS": {"model": "AAA"},
+        "AutoARIMA": {"max_p": 2, "use_fallback_model": False},
```
melopeo, line 82: Why do we set `use_fallback_model` to `False`?

timeseries/src/autogluon/timeseries/models/local/statsforecast.py
shchur: This method is used [here](https://github.com/autogluon/autogluon/blob/master/timeseries/src/autogluon/timeseries/models/local/abstract_local_model.py#L165) - the method generates forecast for a single time series using the local model.

With this change, we first check if time series is long enough for the provided `season_length`. Then we call the [method](https://github.com/autogluon/autogluon/blob/master/timeseries/src/autogluon/timeseries/models/local/statsforecast.py#L24) that is shared by all models from StatsForecast.

timeseries/tests/smoketests/test_features_and_covariates.py
shchur: We want to make sure that the tests catch the case when some models fail.

Before we had `use_fallback_model=True`, so even if AutoETS failed, we used the fallback model and the test passed. This is the reason why we didn't notice that the model fails if `len(time_series) < 2 * season_length`.

melopeo: Nice! Thanks!

timeseries/src/autogluon/timeseries/models/local/statsforecast.py
melopeo: Thank you!

melopeo (approved):

PR #3552, opened 2023-09-29
0.8.2 post release

tonyhoo:
*Issue #, if available:*

*Description of changes:*

Update the stable website for 0.8.2 release


Innixma (changes requested):
We actually shouldn't make these changes. The problem is not the version in stable, it is the fact that it is pointing to the wrong link.

Innixma:
Opened a PR which should fix the problem: https://github.com/autogluon/autogluon/pull/3553

tonyhoo:
@Innixma We need to update the version for stable anyway as it is still reflecting the 0.8.1 release which is outdated unless we have a reason to do so

gidler:
These look like the correct changes to me. I'm wondering if we missed [these steps](https://github.com/autogluon/autogluon/blob/master/release_instructions/ReleaseInstructions.md#post-release) during the post-release of version `0.8.2`. Maybe that means we need to make them more explicit?

VERSION
```diff
@@ -1 +1 @@
-0.8.2
\ No newline at end of file
+0.8.3
```
Innixma, line 1: This should be 0.8.2 for the stable release

docs/conf.py
```diff
@@ -4,7 +4,7 @@
 sys.path = ['.', '..'] + sys.path
 
 project = 'AutoGluon'
-release = '0.8.2'
+release = '0.8.3'
```
Innixma, line 7: This should be 0.8.2 for the stable release

Innixma (approved):
Updated to address requested changes

PR #3553, opened 2023-09-29
Force versions link to always go to stable versions file

Innixma:
*Issue #, if available:*

*Description of changes:*

- Ensure versions link in website always links to dev documentation. This way it doesn't become outdated over time.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -252,7 +252,7 @@ hidden:
 ---
 
 Cheat Sheets <cheatsheet.rst>
-Versions <versions.rst>
+Versions <https://auto.gluon.ai/dev/versions.html>
 Tabular FAQ <tutorials/tabular/tabular-faq.md>
 Multimodal FAQ <tutorials/multimodal/multimodal-faq.md>
 Time Series FAQ <tutorials/timeseries/forecasting-faq.md>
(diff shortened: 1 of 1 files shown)


tonyhoo:
The updated version page will redirected user to 0.8.3 dev website, is it intended behavior to route users cross versions?

Innixma:
> The updated version page will redirect user to 0.8.3 dev website, is it intended behavior to route users cross versions?

Yes, this is intended. Otherwise how would you go from an old doc page to seeing all the new versions? The alternative is removing versions from the page for dev and stable, so they don't reference incorrect versions on old pages.

gidler:
I agree with Nick's observation that the existing versions page won't provide a link for users to navigate from an old version back to a new version because it only has version links for version that exist when the version was created. Permanently changing this link is an expedient way to fix this.

@Innixma - Is there a reason why you prefer changing the `versions` link to https://auto.gluon.ai/dev/versions.html instead of https://auto.gluon.ai/stable/versions.html? If the stable versions were updated correctly, which @tonyhoo addressed in https://github.com/autogluon/autogluon/pull/3552, I think pointing to stable would be a better long term solution.

Innixma:
@gidler Good point, yes if it is fixed as in #3552, then we can have it point to stable, which would be better.

Innixma:
Updated to stable

gidler (approved):
Approved. Wait to merge for https://github.com/autogluon/autogluon/pull/3552 to have comments addressed and merged.

PR #3559, opened 2023-10-03
[timeseries] Refactor DirectTabularModel and RecursiveTabularModel

shchur:
*Description of changes:*
- Unify the API for `DirectTabularModel` and `RecursiveTabularModel`. Now both models are based on the `mlforecast` library, inherit from the same base class and accept same hyperparameters.
- Reduce preprocessing time by only calling `mlf.preprocess()` once during `fit()`
- Fix OOM errors previously encountered by `RecursiveTabularModel` on large datasets by using `float32` and replacing `df.dropna()` operation with `df.query("y.notnull()")`. Example memory usage for M5 dataset:
  - using float64 + df.dropna (master branch): 63 GB
  - Using float64 + df.query: 35 GB
  - Using float32 + df.query (this PR): 20 GB
- Update `mlforecast` to `v0.9.3` and remove the safeguards around bugs that have been fixed since `v0.7.3`
- Unify the usage of `warning_filter` across the `timeseries` module


tonyhoo (approved):
LGTM. Thanks for the change

Issue #3458, opened 2023-08-16
[timeseries] Update StatsForecast to v1.6.0


shchur:
The new version of StatsForecast contains new models and improvements to existing models, so it would be desirable to upgrade.

Currently the upgrade is blocked by a dependency conflict: 
- `omegaconf` pins the dependency `antlr4-python3-runtime==4.9.*`
- `statsforecast==1.5.0` depends on `fugue`, which in turn depends on `antlr4-python3-runtime>=4.11`

`omegaconf` is required for `pytorch_lightnng`, so it's a hard requirement for AutoGluon. We need to wait until `omegaconf` updates its dependencies.

Relevant PR: https://github.com/omry/omegaconf/pull/1114

shchur:
Completed in #4194

Issue #3518, opened 2023-09-11
 Integration of Advanced Time Series Frequencies and NaN Handling in AutoGluon TimeSeries Models


obwohl:
To the AutoGluon Development Team,

I am writing to propose specific enhancements to the time series functionalities within AutoGluon:

### Custom Frequencies Support:

Integrate the ability to recognize and utilize custom frequencies as provided by Pandas, especially those aligned with specific operational hours.

### Handling of NaN Values:

Enhance models to natively address NaN values, contingent on the inherent capabilities of the underlying model.
Model Documentation:

### Provide concise documentation detailing the capabilities of each model, specifically their support for custom frequencies and NaN values.
The implementation of these features would significantly augment the versatility of AutoGluon in handling diverse time series datasets.

Thank you for your consideration.

Best regards

shchur:
Thank you for the suggestion! These items are on our radar and we are working on incorporating some of them into the next release (e.g., https://github.com/autogluon/autogluon/pull/3454).

obwohl:
That is great! Could you clarify how one could define "CBH", i.e. custom business hours and let AutoGluon use these as freq?

shchur:
In #3563 we've added support for all the default pandas frequencies that can be represented by a `freq` string (not an offset object). To support, e.g., custom business hours with custom opening hours would we'll need to ensure that all models used by AutoGluon support them, which will require more work.

I don't think we can accomplish this by the next release, but we can try to keep in mind for future versions. It would be helpful if you can provide some examples of other libraries that support this feature, so that we can understand how to structure the API / what use cases we need to cover.

shchur:
I would recommend two things:
- converting data to the `B` frequency instead (this will exclude the weekends)
- fitting the predictor with a subset of the data that fits into memory & making predictions for batches of data

eromoe:
@shchur For `making predictions for batches of data` , I don't see `TimeSeriesPredictor` support increamental trainning. So, with small batchs, how to catch long term trend/ seasonality ?

shchur:
Assuming that your dataset contains a large number of time series, you can define batches as groups of time series (e.g., 1 batch = first 1000 time series from the dataset). I would just decrease the # of time series until everything fits nicely into memory at fit time (without splitting individual time series along the time axis).

If you have a single (or few) extremely long time series, I would recommend to just take the most recent observations, though I have to warn you that we have not extensively tested this setting - we mostly focused on datasets with many time series in our benchmarking.

[...]

PR #3563, opened 2023-10-06
[timeseries] Add support for all pandas frequencies

shchur:
*Issue #, if available:* #3518

*Description of changes:*
- `TimeSeriesPredictor` and all underlying models now support all [pandas frequencies](https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#offset-aliases).
- Move logic for generating time features & lag indices from frequency strings from GluonTS to AutoGluon (see `autogluon.timeseries.utils.datetime`).
  - Logic for all frequencies supported by GluonTS is unchanged, but now we also support "non-standard" frequencies like "BH" (business hour) and "SM" (semi-month) that would produce exceptions in GluonTS.
- Add tests for ensuring that all models support all frequencies.
- Fix a bug in `timeseries/tests/smoketests/test_all_models.py` that led to us missing some model failures.


timeseries/src/autogluon/timeseries/models/local/abstract_local_model.py
```diff
@@ -58,6 +58,8 @@ def __init__(
     ):
         if hyperparameters is None:
             hyperparameters = {}
+        else:
+            hyperparameters = hyperparameters.copy()
```
shchur, line 62: In-place modification of `hyperparameters` here led to a bug in `timeseries/tests/smoketests/test_all_models.py`: we popped the `"use_fallback_model"` key from `hyperparameters` (modifying the `DUMMY_MODEL_HPARAMS` dictionary in tests), so all models except the first one actually used the fallback model. 

Because of this, model failures for some models would go unnoticed by the tests.

timeseries/src/autogluon/timeseries/models/presets.py
```diff
@@ -30,6 +29,9 @@
 from .abstract import AbstractTimeSeriesModel
 from .multi_window.multi_window_model import MultiWindowBacktestingModel
 
+# TODO: Enable ARIMAModel after upgrading to StatsForecast >=1.5.0 - currently ARIMA model is broken
```
shchur, line 32: Because of the bug described above (in-place modification of `DUMMY_MODEL_HPARAMS`), we missed the fact that `ARIMA` model doesn't work in StatsForecast v1.4.0 — the model failure was caught by the fallback model. We will enable the model after we resolve #3458 and upgrade StatsForecast to >= v1.5.0.

docs/tutorials/timeseries/forecasting-model-zoo.md
```diff
@@ -48,7 +48,6 @@ Note that some of the models' hyperparameters have names and default values that
    SeasonalNaiveModel
    AverageModel
    SeasonalAverageModel
-   ARIMAModel
```
tonyhoo, line 51: Do we propose to use AutoARIMA as a replacement for the ARIMA model?

timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
```diff
@@ -12,8 +12,12 @@
 from autogluon.tabular import TabularPredictor
 from autogluon.timeseries.dataset.ts_dataframe import ITEMID, TIMESTAMP, TimeSeriesDataFrame
 from autogluon.timeseries.models.abstract import AbstractTimeSeriesModel
+from autogluon.timeseries.utils.datetime import (
+    get_lags_for_frequency,
```
tonyhoo, line 16: Do we need to define additional functions at AG, considering that gluonts already has similar ones (e.g., this [one](https://github.com/awslabs/gluonts/blob/0bc0a459263d1599f5f8eaa6d5e94647917fcef4/src/gluonts/time_feature/lag.py#L29))?

shchur: We used to rely on these methods from GluonTS, but they do not support many of the pandas frequencies (e.g., C, BM, CBM, CBMS, SMS, SM, BQ, BQS, BH, L, U, N). Our implementation of these methods is equivalent to the GluonTS implementations for the frequencies that GluonTS supports, but we also add support for the other pandas frequencies where GluonTS fails.

docs/tutorials/timeseries/forecasting-model-zoo.md
shchur: As of v0.8.2, ARIMA is not used in any of the presets, but I agree that it would be nice to have the plain ARIMA model at some point. The plan would be to return ARIMA after #3458 is resolved and we can upgrade to StatsForecast >=1.5.0 where the plain ARIMA model is introduced. Currently this upgrade is blocked by https://github.com/omry/omegaconf/pull/1114.

tonyhoo (approved):
Please ensure that all tests are passing. Thank you.

timeseries/src/autogluon/timeseries/models/local/abstract_local_model.py
Innixma: nice catch

Innixma (approved):
LGTM!

PR #3569, opened 2023-10-09
[core] Add symmetric_mean_absolute_percentage error

shchur:
*Description of changes:*
- Add `symmetric_mean_absolute_percentage_error` regression metric. This metric corresponds to one of the metrics used in TimeSeries module, and it would be helpful to use this metric inside `DirectTabular` and `RecursiveTabular` forecasting models.

### DIFF
--- a/core/src/autogluon/core/metrics/__init__.py
+++ b/core/src/autogluon/core/metrics/__init__.py
@@ -412,13 +412,13 @@ mean_absolute_percentage_error = make_scorer(
 )
 mean_absolute_percentage_error.add_alias("mape")
 
+
 def smape_func(y_true, y_pred) -> float:
     epsilon = np.finfo(np.float64).eps
     return np.average(np.abs(y_pred - y_true) / np.maximum(np.abs(y_true) + np.abs(y_pred), epsilon))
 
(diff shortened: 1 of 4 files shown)


core/src/autogluon/core/metrics/__init__.py
```diff
@@ -413,6 +413,15 @@ def make_scorer(
 mean_absolute_percentage_error.add_alias("mape")
 
 
+def smape_func(y_true, y_pred) -> float:
+    epsilon = np.finfo(np.float64).eps
+    return np.average(np.abs(y_pred - y_true) / np.maximum(np.abs(y_true) + np.abs(y_pred), epsilon))
+
+
+symmetric_mean_absolute_percentage_error = make_scorer("symmetric_mean_absolute_percentage_error", smape_func, optimum=0.0, greater_is_better=False)
+symmetric_mean_absolute_percentage_error.add_alias("smape")
```
shchur, line 422: If we don't want to add SMAPE as a default metric for `problem_type=REGRESSION`, I will move the metric definition to `autogluon.timeseries`.

Innixma (approved):
LGTM!

PR #3585, opened 2023-10-12
WIP: Generate simulation artifact for TimeSeriesPredictor

shchur:
*Description of changes:*
- Similar to #3555, but for the `TimeSeriesPredictor`

### DIFF
--- a/timeseries/src/autogluon/timeseries/predictor.py
+++ b/timeseries/src/autogluon/timeseries/predictor.py
@@ -290,7 +290,7 @@ class TimeSeriesPredictor:
                 raise ValueError(f"Some time series in {name} consist completely of NaN values. Please remove them.")
         return df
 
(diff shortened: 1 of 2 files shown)


timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -1046,3 +1046,43 @@ def refit_full(self, model: str = "all", set_best_to_refit_full: bool = True) ->
                     f"Training may have failed on the refit model. AutoGluon will default to using '{model_best}' for predict()."
                 )
         return refit_full_dict
+
+    def get_simulation_artifact(self, test_data: TimeSeriesDataFrame) -> dict:
```
Innixma, line 1050: I renamed to `simulation_artifact` for the Tabular code, may want to do the same here.

PR #3591, opened 2023-10-13
Fix model memory usage calculation

Innixma:
*Issue #, if available:*

*Description of changes:*

Fix model memory usage calculation. The bug was introduced in #3033.

The bug makes it so that models are treated as requiring `1/mem_error_threshold` the memory as was intended.

For XGBoost and CatBoost whose `mem_error_threshold` was 1, this had no impact.

For RandomForest/ExtraTrees, whose `mem_error_threshold=0.5`, this caused the models to be treated as requiring 2x the memory as was intended.

For KNN, whose `mem_error_threshold=0.2`, this caused KNN to be treated as requiring 5x the memory.

This PR fixes this bug.

This PR (correctly trains):

```
Fitting model: RandomForest ...
	Warning: Potentially not enough memory to safely train model. Estimated to require 250.000 GB out of 516.989 GB available memory (48.357%)... (50.000% of avail memory is the max safe size)
	To avoid this warning, specify the model hyperparameter "ag.max_memory_usage_ratio" to a larger value (currently 1.0, set to >=1.26 to avoid the warning)
		To set the same value for all models, do the following when calling predictor.fit: `predictor.fit(..., ag_args_fit={"ag.max_memory_usage_ratio": VALUE})`
		Setting "ag.max_memory_usage_ratio" to values above 1 may result in out-of-memory errors. You may consider using a machine with more memory as a safer alternative.
	-0.2944	 = Validation score   (-log_loss)
	1.33s	 = Training   runtime
	0.04s	 = Validation runtime
```

Mainline (incorrectly skips and incorrectly calculates memory percentage):

```
Fitting model: RandomForest ...
	Warning: Not enough memory to safely train model. Estimated to require 250.000 GB out of 517.378 GB available memory (96.641%)... (50.000% of avail memory is the max safe size)
	To force training the model, specify the model hyperparameter "ag.max_memory_usage_ratio" to a larger value (currently 1.0, set to >=1.02 to avoid the error)
		To set the same value for all models, do the following when calling predictor.fit: `predictor.fit(..., ag_args_fit={"ag.max_memory_usage_ratio": VALUE})`
	Not enough memory to train RandomForest... Skipping this model.

```

### DIFF
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -1659,9 +1659,15 @@ class AbstractModel:
             return  # Model is smaller than the min threshold to check available mem
 
         available_mem = ResourceManager.get_available_virtual_mem()
-        ratio = approx_mem_size_req / available_mem
-        min_error_memory_ratio = ratio / mem_error_threshold
-        min_warning_memory_ratio = ratio / mem_warning_threshold
+
+        # The expected memory usage percentage of the model during fit
+        expected_memory_usage_ratio = approx_mem_size_req / available_mem
+
+        # The minimum `max_memory_usage_ratio` values required to avoid an error/warning
+        min_error_memory_ratio = expected_memory_usage_ratio / mem_error_threshold
+        min_warning_memory_ratio = expected_memory_usage_ratio / mem_warning_threshold
+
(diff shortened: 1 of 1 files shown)


prateekdesai04:
LGTM!

prateekdesai04 (approved):

PR #3606, opened 2023-10-19
[timeseries] Update GluonTS to v0.14.0

shchur:
*Description of changes:*
- Bump GluonTS to v0.14.0.
- Change all `pytorch_lightning` references to `lightning.pytorch`.
- Add wrapper for the `WaveNet` model.
- Rename old MXNet argument names `learning_rate` / `epochs` to `lr` / `max_epochs` for compatibility with PyTorch-Lightning naming (both versions are still supported, only docstrings are updated).

------
**To Do:**
- [x] Change dependency in `setup.py`  to `gluonts~=0.14.0`

### DIFF
--- a/timeseries/setup.py
+++ b/timeseries/setup.py
@@ -29,7 +29,7 @@ install_requires = [
     "lightning",  # version range defined in `core/_setup_utils.py`
     "pytorch_lightning",  # version range defined in `core/_setup_utils.py`
     "statsmodels>=0.13.0,<0.15",
-    "gluonts==0.14.0rc2",
+    "gluonts>=0.14.0,<0.15",
     "networkx",  # version range defined in `core/_setup_utils.py`
     # TODO: update statsforecast to v1.5.0 - resolve antlr4-python3-runtime dependency clash with multimodal
     "statsforecast>=1.4.0,<1.5",
--- a/timeseries/setup.py
+++ b/timeseries/setup.py
@@ -29,7 +29,7 @@ install_requires = [
     "lightning",  # version range defined in `core/_setup_utils.py`
     "pytorch_lightning",  # version range defined in `core/_setup_utils.py`
     "statsmodels>=0.13.0,<0.15",
-    "gluonts==0.14.0rc1",
+    "gluonts==0.14.0rc2",
     "networkx",  # version range defined in `core/_setup_utils.py`
     # TODO: update statsforecast to v1.5.0 - resolve antlr4-python3-runtime dependency clash with multimodal
     "statsforecast>=1.4.0,<1.5",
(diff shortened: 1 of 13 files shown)


timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
```diff
@@ -146,7 +146,7 @@ class AbstractGluonTSModel(AbstractTimeSeriesModel):
 
     gluonts_model_path = "gluon_ts"
     # default number of samples for prediction
-    default_num_samples: int = 1000
+    default_num_samples: int = 250
```
shchur, line 149: This parameter is only used by a single model (DeepAR), where it is overridden to 250. We might as well keep the overridden value here, as no other model generates SampleForecast output.

timeseries/setup.py
```diff
@@ -29,7 +29,7 @@
     "lightning",  # version range defined in `core/_setup_utils.py`
     "pytorch_lightning",  # version range defined in `core/_setup_utils.py`
     "statsmodels>=0.13.0,<0.15",
-    "gluonts>=0.13.1,<0.14",
+    "gluonts==0.14.0rc2",
```
canerturkmen, line 32: Did we do a quick benchmark to ensure no performance regressions?

timeseries/src/autogluon/timeseries/models/presets.py
```diff
@@ -44,6 +45,7 @@
     DLinear=DLinearModel,
     PatchTST=PatchTSTModel,
     TemporalFusionTransformer=TemporalFusionTransformerModel,
+    WaveNet=WaveNetModel,
```
canerturkmen, line 48: Shouldn't the model also get a default priority and a place in `best` or does that come after testing?

timeseries/src/autogluon/timeseries/models/gluonts/torch/__init__.py
```diff
@@ -1,9 +0,0 @@
-from .models import DeepARModel, DLinearModel, PatchTSTModel, SimpleFeedForwardModel, TemporalFusionTransformerModel
```
canerturkmen, line 1: nit: why?

shchur: We have the exact same imports listed in `src/autogluon/timeseries/models/gluonts/__init__.py`, this is just duplication of code (now that the MXNet models are gone).

timeseries/src/autogluon/timeseries/models/presets.py
shchur: I plan to add priority + refactor the presets after running zero-shot HPO for portfolio selection.

timeseries/setup.py
shchur: Haven't gotten to running the benchmark yet. If we notice any regressions, we can try to address them in v0.14.1.

PR #3608, opened 2023-10-19
[Tabular][Enhancement] Incorporating folds in benchmarking

prateekdesai04:
*Description of changes:*
Adding folds as a parameter for benchmarking.
Also fixing benchmarking issues, initially AG Bench was picking up wrong benchmark to run.
Adding a new test benchmark (small).

Test Slash Command Link: https://github.com/prateekdesai04/autogluon/pull/26#issuecomment-1769579799


.github/workflows/benchmark-command.yml
```diff
@@ -40,14 +40,14 @@ on:
           - 8h
           - 16h
           - 24h
+      folds:
+        description: 'Number of folds to run'
+        required: true
```
tonyhoo, line 45: My understanding is that the num of folds are not required and we can run benchmark fine without it

.github/workflows/benchmark-command.yml
```diff
@@ -40,14 +40,14 @@ on:
...
       branch_or_pr_number:
         description: 'Branch or PR number to run the benchmark on'
         required: true
-      fork_name:
-        description: 'Get the forked PR repository name'
-        required: true
-      fork_branch:
-        description: 'Get the forked PR repository branch name'
+      fork_info:
+        description: 'Get the forked PR repository name and branch'
```
tonyhoo, line 50: We shall add an example in the description

.github/workflows/benchmark-command.yml
```diff
@@ -89,9 +89,15 @@ jobs:
...
+      - name: Extract Fork Info
+        id: parse_fork_info
+        run: |
+          IFS="|" read -r fork_name fork_branch <<< "${{ github.event.inputs.fork_info }}"
+          echo "FORK_NAME=$fork_name" >> $GITHUB_OUTPUT
+          echo "FORK_BRANCH=$fork_branch" >> $GITHUB_OUTPUT
       - name: Generate AMLB User Dir
         run: |
-          /bin/bash CI/bench/generate_amlb_user_dir.sh ${{ github.event.inputs.fork_name }} ${{ github.event.inputs.fork_branch }} ${{ github.sha }}
+          /bin/bash CI/bench/generate_amlb_user_dir.sh ${{ steps.parse_fork_info.outputs.FORK_NAME }} ${{ steps.parse_fork_info.outputs.FORK_BRANCH }} ${{ github.sha }} "" ${{ github.event.inputs.folds }}
```
tonyhoo, line 100: Thanks for the fix

CI/bench/tabular/generate_framework.py
```diff
@@ -8,16 +8,29 @@
 
 parser.add_argument("--repository", help="git repository to run autogluon on", type=str, required=True)
 parser.add_argument("--branch", help="git branch to run autogluon on", type=str, required=True)
+parser.add_argument("--folds_to_run", help="number of folds to run", type=int, required=True)
```
tonyhoo, line 11: should not be required per my understanding

.github/workflows/benchmark-command.yml
prateekdesai04: Number of folds are pre-defined in the constraints.yaml - https://github.com/autogluon/autogluon/pull/3608/files#diff-b973100f40432a8431c41887e66d9a98912e3864c3e22a7cfdab54fa69259847.
This functionality is added so that we could control it if we wanted, currently I have set it as required. I can change it to False and default values will be used. 
However, the number of folds run in the PR must be equal to number of folds run on master

.github/workflows/benchmark-command.yml
prateekdesai04: Sure! 👍 
This was the benchmarking fix which picked wrong datasets: https://github.com/autogluon/autogluon/pull/3608/files#diff-d7a950e181b92ab6db738323d4bf38363908a944efb0d226ef1aa097d7caff17

.github/workflows/benchmark-command.yml
prateekdesai04: Yes will do!

.github/workflows/benchmark-command.yml
Innixma: required should be False imo.

prateekdesai04: Yup will change this!

.github/workflows/benchmark-command.yml
prateekdesai04: Default fold value?

CI/bench/tabular/amlb_user_dir/benchmarks/small.yaml
```diff
@@ -0,0 +1,12 @@
...
+# Small datasets to test benchmarking - extracted from autogluon-bench/data/metadata/task_metadata.csv
+
+- name: arcene
+  openml_task_id: 190412
+
+- name: tecator
+  openml_task_id: 359934
+
+- name: boston
+  openml_task_id: 359950
```
Innixma, line 12: datasets should be changed to the following as they are the fastest to train on:

```
# binary
- name: Australian
  openml_task_id: 146818

# multiclass
- name: vehicle
  openml_task_id: 190146

# regression
- name: sensory
  openml_task_id: 359931
```

.github/workflows/benchmark-command.yml
Innixma: default `fork_info`

CI/bench/generate_amlb_user_dir.sh
```diff
@@ -4,9 +4,10 @@ REPOSITORY=$1
 BRANCH=$(basename $2)
 SHORT_SHA=$3
 PR_NUMBER=$4
+FOLDS=$5
 
 # generate tabular configs
-python $(dirname "$0")/tabular/generate_framework.py --repository https://github.com/$REPOSITORY.git --branch $BRANCH
+python $(dirname "$0")/tabular/generate_framework.py --repository https://github.com/$REPOSITORY.git --branch $BRANCH --folds_to_run $FOLDS
```
Innixma, line 10: Won't this crash if folds isn't specified?

Innixma: `--folds_to_run $FOLDS` should only exist in the command if `$FOLDS` is specified

.github/workflows/benchmark-command.yml
Innixma: but I see now that `fork_info` is required, so no need I suppose.

Innixma (approved):
LGTM, consider cleaning up the script that takes in the arguments to avoid passing argument key values when they are None.

.github/workflows/benchmark-command.yml
```diff
@@ -89,9 +89,15 @@ jobs:
           role-to-assume: arn:aws:iam::369469875935:role/AutoGluonCIBenchmarkConfig
           role-duration-seconds: 3600
           aws-region: us-east-1
+      - name: Extract Fork Info
+        id: parse_fork_info
+        run: |
+          IFS="|" read -r fork_name fork_branch <<< "${{ github.event.inputs.fork_info }}"
```
tonyhoo, line 95: Though `fork_name` and `fork_branch` are not directly run as a script in this step, please make sure they are not being used for script execution in subsequent steps to precent the risk of [script injection](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#understanding-the-risk-of-script-injections)

CI/bench/generate_amlb_user_dir.sh
prateekdesai04: added more robust check

.github/workflows/benchmark-command.yml
prateekdesai04: The benchmark workflow from master is run and that too only by maintainers thus preventing this. 
But still made changes and used the env method to mitigate this.

tonyhoo (approved):

Issue #2779, opened 2023-01-29
Fix Stack Information Leakage


Innixma:
Stack Information Leakage is a complex problem unique to multi-layer stack ensembling.

I first discovered this problem in 2021 during benchmark results analysis where I found that in certain datasets AutoGluon would produce very poor results due to heavy validation score overfitting without an obvious reason. I have since identified the cause and named it "Stack Information Leakage".

[Here is a Colab Notebook](https://colab.research.google.com/drive/126XwY0G_qJdPKVC004mDNKzSXVYsP6z1?usp=sharing) that describes and demonstrates the issue.

AutoGluon-Tabular currently suffers from stack information leakage, reducing the predictive quality of binary and regression problems, and introducing extreme worst-case scenarios of intensely overfit models.

2-5:
Is this issue present if `tuning_data` is passed to `fit` together with `use_bag_holdout=True`?

Innixma:
> Is this issue present if `tuning_data` is passed to `fit` together with `use_bag_holdout=True`?

When `use_bag_holdout=True`, the validation scores of the models, their ensembling, and the selection of the best model will use the holdout data and thus not be overfit. However, the multi-layer stack ensemble will still suffer reduced test (and therefore holdout) performance if the leak occurs, since they are still fed the out-of-fold predictions during training that is unrelated to the holdout data.

`use_bag_holdout=True` is a method to _detect_ the leak, but it does not _fix_ the leak. Detection is still very useful however, as it avoids the worst-case scenario of picking the model that suffered the most from the leak (and thus has very poor test score)

hanxuh-hub:
Sounds promising! Is there a timeline for solving this issue?

mglowacki100:
Hi @Innixma ,
Interesting problem, I think I have a potential solution for binary classification case that prevents overfitting, it is easy to implement and independent from model type. Basically, we need to round OOF probabilities.  Here is an example, based on your notebook (section 'potential fix'):  https://colab.research.google.com/gist/mglowacki100/fd8e28cb4e19ff8a1b21703de7866472/autogluoncomplexdataleak_v07_leak_fix.ipynb ):
```
ROUND = 0.1
l1_oof2 = l1_oof.copy()
l1_test2 = l1_test.copy()
l1_oof2[' >50K'] = l1_oof2[' >50K'].apply(lambda x: round(x / ROUND) * ROUND)
[...]

PR #3616, opened 2023-10-21
Support Dynamic Stacking to Avoid Stacked Overfitting

LennartPurucker:
Rel.: #2779 

### Description of changes:
This PR adds support for dynamic stacking. The idea of dynamic stacking is to avoid stacked overfitting (a.k.a. stacked information leakage) by fitting AutoGluon at least twice. All but the last fits are done on a subset of the training data whereby a holdout set is used to determine if stacked overfitting occurs for the provided data. Afterwards, we run the last fit with or without  multi-layer stacking depending on whether stacked overfitting has not occurred in a previous fit.

We understand dynamic stacking as a baseline solution to avoid stacked overfitting. Nevertheless, so far, it is the best solution we have benchmarked due to it being most consistent (which is very important for AutoML systems). Moreover, due to having a holdout set as a result of this approach, we have an empirical source of truth to determine if stacked overfitting occurred. While this source of truth may not necessarily generalize to the test data, it has so far still been better than any heuristic alternative or adjustment to the out-of-fold predictions. 

The default version of the code in this PR has an accuracy (_!preliminary numbers!_) of correctly determining whether stacked overfitting occurs for each fold of the AutoML benchmark of ~74% (balanced accuracy: ~69%). This translates into better predictive performance on average.

### Code Example 

<details>
<summary>
Script (outdated, see tests for newest examples)
</summary>

```python
import numpy as np
import openml
import pandas as pd


def get_data(tid: int, fold: int):
    # Get Task and dataset from OpenML and return split data
    oml_task = openml.tasks.get_task(tid, download_splits=True, download_data=True, download_qualities=False, download_features_meta_data=False)

    train_ind, test_ind = oml_task.get_train_test_split_indices(fold)
    X, *_ = oml_task.get_dataset().get_data(dataset_format="dataframe")

    return (
        X.iloc[train_ind, :].reset_index(drop=True),
        X.iloc[test_ind, :].reset_index(drop=True),
        oml_task.target_name,
        oml_task.task_type != "Supervised Classification",
    )


def _print_lb(leaderboard, task_id):
    with pd.option_context("display.max_rows", None, "display.max_columns", None, "display.width", 1000):
        print(leaderboard[leaderboard["model"].str.endswith("L1")].sort_values(by="score_val", ascending=True))
        print(leaderboard[~leaderboard["model"].str.endswith("L1")].sort_values(by="score_val", ascending=False))


def _run(task_id, metric, fold=0, print_res=True, enable_test=False):
    train_data, test_data, label, regression = get_data(task_id, fold)
    n_max_cols = 100
    n_max_train_instances = 500
    n_max_test_instances = 200

    # Sub sample instances
    train_data = train_data.sample(n=min(len(train_data), n_max_train_instances), random_state=0).reset_index(drop=True)
    test_data = test_data.sample(n=min(len(test_data), n_max_test_instances), random_state=0).reset_index(drop=True)

    # Sub sample columns
    cols = list(train_data.columns)
    cols.remove(label)
    if len(cols) > n_max_cols:
        cols = list(np.random.RandomState(42).choice(cols, replace=False, size=n_max_cols))
    train_data = train_data[cols + [label]]
    test_data = test_data[cols + [label]]

    # Run AutoGluon
    print(f"### task {task_id} and fold {fold} and shape {(train_data.shape, test_data.shape)}.")
    predictor = TabularPredictor(eval_metric=metric, label=label, verbosity=2)
    predictor.fit(
        train_data=train_data,
        hyperparameters={
            "FASTAI": [{}],
            "NN_TORCH": [{}],
            "RF": [{}],
            "XT": [{}],
            "GBM": [{}],
        },
        num_bag_sets=2,
        num_bag_folds=2,
        num_gpus=0,
        num_stack_levels=1,
        fit_weighted_ensemble=True,
        dynamic_stacking=True,
        time_limit=240,
        ds_args=dict(use_holdout=True, detection_time_frac=1 / 4, holdout_frac=1 / 9),
    )
    leaderboard = predictor.leaderboard(train_data, silent=True)[["model", "score_test", "score_val"]].sort_values(by="model").reset_index(drop=True)

    if print_res:
        _print_lb(leaderboard, task_id)

    return leaderboard


if __name__ == "__main__":
    _run(359955, "roc_auc")
    _run(146217, "log_loss")
    _run(359938, "mse")
```

</details>

### TODOs and Open Questions for Merge

- Determine how to include this in presets and decide on default split ratio

Write tests:

- No time limit and time limit 
- Extra fit works with dynamic stacking at first fit 
- holdout, cv, repeated cv, custom val data. 

---

### DIFF
--- a/tabular/src/autogluon/tabular/predictor/predictor.py
+++ b/tabular/src/autogluon/tabular/predictor/predictor.py
@@ -1234,8 +1234,8 @@ class TabularPredictor:
         if holdout_data is None:
             ag_fit_kwargs["X"] = X
         else:
-            logger.log(20, "Concatenating holdout data from dynamic stacking to the training for the full fit.")
-            ag_fit_kwargs["X"] = pd.concat([X, holdout_data])
+            logger.log(20, "Concatenating holdout data from dynamic stacking to the training data for the full fit (and reset the index).")
+            ag_fit_kwargs["X"] = pd.concat([X, holdout_data], ignore_index=True)
 
         ag_fit_kwargs["X_val"] = X_val
         ag_fit_kwargs["X_unlabeled"] = X_unlabeled
--- a/tabular/src/autogluon/tabular/predictor/predictor.py
+++ b/tabular/src/autogluon/tabular/predictor/predictor.py
@@ -14,7 +14,6 @@ from typing import List, Optional, Tuple, Union
 import networkx as nx
 import numpy as np
 import pandas as pd
-import ray
 
 from autogluon.common.loaders import load_json
 from autogluon.common.savers import save_json
--- a/tabular/src/autogluon/tabular/predictor/predictor.py
+++ b/tabular/src/autogluon/tabular/predictor/predictor.py
@@ -9,8 +9,8 @@ import pprint
 import shutil
 import time
 import warnings
-from typing import List, Optional, Tuple, Union
 from concurrent.futures import ProcessPoolExecutor
+from typing import List, Optional, Tuple, Union
 
 import networkx as nx
 import numpy as np
@@ -43,6 +43,7 @@ from autogluon.core.metrics import Scorer
 from autogluon.core.problem_type import problem_type_info
 from autogluon.core.pseudolabeling.pseudolabeling import filter_ensemble_pseudo, filter_pseudo
 from autogluon.core.scheduler.scheduler_factory import scheduler_factory
+from autogluon.core.stacked_overfitting.utils import check_stacked_overfitting_from_leaderboard
 from autogluon.core.trainer import AbstractTrainer
 from autogluon.core.utils import (
     get_pred_from_proba_df,
@@ -54,7 +55,6 @@ from autogluon.core.utils.decorators import apply_presets
 from autogluon.core.utils.loaders import load_pkl, load_str
 from autogluon.core.utils.savers import save_pkl, save_str
 from autogluon.core.utils.utils import default_holdout_frac
-from autogluon.core.stacked_overfitting.utils import check_stacked_overfitting_from_leaderboard
 
 from ..configs.feature_generator_presets import get_default_feature_generator
 from ..configs.hyperparameter_configs import get_hyperparameter_config
--- a/tabular/src/autogluon/tabular/predictor/predictor.py
+++ b/tabular/src/autogluon/tabular/predictor/predictor.py
@@ -1113,7 +1113,8 @@ class TabularPredictor:
         clean_up_fits: bool,
         holdout_data: Optional[Union[str, pd.DataFrame, None]] = None,
     ):
-        """Dynamically determines if stacking is used or not by validating the behavior of a sub-fit of AutoGluon that uses stacking on held out data."""
+        """Dynamically determines if stacking is used or not by validating the behavior of a sub-fit of AutoGluon that uses stacking on held out data.
+        See `ds_args` in the docstring of `fit()` for details on the parameters."""
         time_start = time.time()
         time_limit_og = ag_fit_kwargs["time_limit"]
         org_num_stack_levels = ag_fit_kwargs["num_stack_levels"]
--- a/tabular/src/autogluon/tabular/predictor/predictor.py
+++ b/tabular/src/autogluon/tabular/predictor/predictor.py
@@ -755,31 +755,35 @@ class TabularPredictor:
                 See the `ag_args_ensemble` argument from "Advanced functionality: Custom AutoGluon model arguments" in the `hyperparameters` argument documentation for valid values.
                 Identical to specifying `ag_args_ensemble` parameter for all models in `hyperparameters`.
                 If a key in `ag_args_ensemble` is already specified for a model in `hyperparameters`, it will not be altered through this argument.
-            ds_args : dict, default = {'detection_time_frac': 1/4, 'use_holdout': True, 'holdout_frac': 1/9, 'memory_safe_fits': True, 'clean_up_fits': True},
+            ds_args : dict, see bleow for default,
                 Keyword arguments for dynamic stacking, only used if `dynamic_stacking=True`. These keyword arguments control the behavior of dynamic stacking
                 and determine how AutoGluon tries to detect stacked overfitting. To detect stacked overfitting, AutoGluon will fit itself (so called sub-fits)
                 on a subset (for holdout) or multiple subsets (for repeated cross-validation) and use the predictions of AutoGluon on the validation data to
                 detect stacked overfitting. The sub-fits stop and stacking will be disabled if any sub-fit shows stacked overfitting.
                 Allowed keys and values are:
-                    `detection_time_frac` : float in (0,1)
+                    `detection_time_frac` : float in (0,1), default = 1/4
                         Determines how much of the original training time is used for detecting stacked overfitting.
                         If no time limit is given to AutoGluon, this parameter is ignored and AutoGluon is fit without a time limit in the sub-fit.
                         When using (repeatd) cross-validation, each sub-fit will be fit for `1/n_splits * detection_time_frac * time_limit`.
-                    `use_holdout` : bool
+                    `use_holdout` : bool, default = True
                         Whether to use holdout validation or not.
-                    `holdout_frac` : float in (0,1)
+                    `holdout_frac` : float in (0,1), default = 1/9
                         Determines how much of the original training data is used for the holdout data.
-                    `n_folds` : int in [2, +inf)
+                    `n_folds` : int in [2, +inf), default = 2
                         Number of folds to use for cross-validation.
-                    `n_repeats` : int [1, +inf)
+                    `n_repeats` : int [1, +inf), default = 1
                         Number of repeats to use for repeated cross-validation.
-                    `memory_safe_fits` : bool
+                    `memory_safe_fits` : bool, default = True
                         If True, AutoGluon runs each sub-fit in a ray-based subprocess to avoid memory leakage that exist due to Python's lackluster
                         garbage collector.
-                    `clean_up_fits` : bool
+                    `clean_up_fits` : bool, default = True
                         If True, AutoGluon will remove all saved information from sub-fits from disc.
                         If False, the sub-fits are kept on disc and `self._sub_fits` will store paths to the sub-fits, which can be loaded just like any other
                         predictor from disc using `TabularPredictor.load()`.
+                    `custom_val_data`: str or :class:`TabularDataset` or :class:`pd.DataFrame`, default = None
+                        Another dataset containing validation data reserved for detecting stacked overfitting. This dataset should be in the same format as
+                        `train_data`. If str is passed, `custom_val_data` will be loaded using the str value as the file path.
+                        If `custom_val_data` is not None, the sub-fit fit on all of `train_data`.
             included_model_types : list, default = None
                 To only include listed model types for training during `fit()`.
                 Models that are listed in `included_model_types` but not in `hyperparameters` will be ignored.
@@ -1109,6 +1113,7 @@ class TabularPredictor:
         n_repeats=None,
         memory_safe_fits=None,
         clean_up_fits=None,
+        custom_val_data=None,
     ):
         time_start = time.time()
         org_time_limit = ag_fit_kwargs["time_limit"]
@@ -1154,9 +1159,15 @@ class TabularPredictor:
             )
             logger.info(f"Starting holdout-based sub-fit for dynamic stacking. Context path is: {ds_fit_kwargs['ds_fit_context']}.")
             stacked_overfitting = self._sub_fit_memory_save_wrapper(X, time_limit, ds_fit_kwargs, inner_ag_fit_kwargs, ag_post_fix_kwargs)
+        elif custom_val_data is not None:
+            # Custom val data support
+            _, custom_val_data, _ = self._validate_fit_data(X, tuning_data=custom_val_data)
+            ds_fit_kwargs["ds_fit_context"] = ds_fit_context + "/sub_fit_custom_ho"
+            logger.info(f"Starting holdout-based sub-fit for dynamic stacking with custom validation data. Context path is: {ds_fit_kwargs['ds_fit_context']}.")
+            stacked_overfitting = self._sub_fit_memory_save_wrapper(
+                X, time_limit, ds_fit_kwargs, inner_ag_fit_kwargs, ag_post_fix_kwargs, custom_val_data=custom_val_data
+            )
         else:
-            # TODO: support custom val data
-
             # Holdout is false, use (repeated) cross-validation
             is_stratified = self.problem_type in [REGRESSION, QUANTILE, SOFTCLASS]
             self._learner._validate_groups(X, X_val)  # Validate splits before splitting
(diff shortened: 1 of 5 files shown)


tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -646,6 +652,15 @@ def fit(
...
             WeightedEnsembleModel if stacking is disabled.
+        dynamic_stacking: bool, default = False
+            If True and `num_stack_levels` > 0, AutoGluon will dynamically determine whether to use stacking or not by first validating AutoGluon's stacking
+            behavior. This is done to avoid so-called stacked overfitting that can make traditional multi-layer stacking, as used in AutoGluon, fail drastically
+            and produce unreliable validation scores.
+            It is recommended to keep this value set to True when using stacking, as long as it is unknown whether the data is affected by stacked overfitting.
+            If it is known that the data is unaffected by stacked overfitting, then setting this value to False is expected to maximize predictive quality.
+            If enabled, by default, AutoGluon performs dynamic stacking by spending 25% of the provided time limit for detection and all remaining
+            time for fitting AutoGluon. This can be adjusted by setting by passing `ds_args` with different parameters to `fit()`.
```
Innixma, line 662: `by setting by passing` -> `by specifying`

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -740,11 +755,40 @@ def fit(
                 See the `ag_args_ensemble` argument from "Advanced functionality: Custom AutoGluon model arguments" in the `hyperparameters` argument documentation for valid values.
                 Identical to specifying `ag_args_ensemble` parameter for all models in `hyperparameters`.
                 If a key in `ag_args_ensemble` is already specified for a model in `hyperparameters`, it will not be altered through this argument.
+            ds_args : dict, see bleow for default
```
Innixma, line 758: `bleow` -> `below`

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -740,11 +755,40 @@ def fit(
...
+                Keyword arguments for dynamic stacking, only used if `dynamic_stacking=True`. These keyword arguments control the behavior of dynamic stacking
+                and determine how AutoGluon tries to detect stacked overfitting. To detect stacked overfitting, AutoGluon will fit itself (so called sub-fits)
+                on a subset (for holdout) or multiple subsets (for repeated cross-validation) and use the predictions of AutoGluon on the validation data to
+                detect stacked overfitting. The sub-fits stop and stacking will be disabled if any sub-fit shows stacked overfitting.
+                Allowed keys and values are:
+                    `detection_time_frac` : float in (0,1), default = 1/4
+                        Determines how much of the original training time is used for detecting stacked overfitting.
+                        If no time limit is given to AutoGluon, this parameter is ignored and AutoGluon is fit without a time limit in the sub-fit.
+                        When using (repeatd) cross-validation, each sub-fit will be fit for `1/n_splits * detection_time_frac * time_limit`.
```
Innixma, line 767: repeatd -> repeated

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -740,11 +755,40 @@ def fit(
...
+                    `use_holdout` : bool, default = True
+                        Whether to use holdout validation or not.
```
Innixma, line 769: Clarify what happens if False

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -740,11 +755,40 @@ def fit(
...
+                        Determines how much of the original training data is used for the holdout data.
+                    `n_folds` : int in [2, +inf), default = 2
+                        Number of folds to use for cross-validation.
+                    `n_repeats` : int [1, +inf), default = 1
+                        Number of repeats to use for repeated cross-validation.
+                    `memory_safe_fits` : bool, default = True
+                        If True, AutoGluon runs each sub-fit in a ray-based subprocess to avoid memory leakage that exist due to Python's lackluster
+                        garbage collector.
+                    `clean_up_fits` : bool, default = True
+                        If True, AutoGluon will remove all saved information from sub-fits from disc.
```
Innixma, line 780: `disc` -> `disk`

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -740,11 +755,40 @@ def fit(
...
+                        If False, the sub-fits are kept on disc and `self._sub_fits` will store paths to the sub-fits, which can be loaded just like any other
```
Innixma, line 781: `disc` -> `disk`

Innixma: I see other occurrences, I'd. recommend find-and-replace for all of them.

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -740,11 +755,40 @@ def fit(
...
+                        predictor from disc using `TabularPredictor.load()`.
+                    `custom_val_data`: str or :class:`TabularDataset` or :class:`pd.DataFrame`, default = None
+                        Another dataset containing validation data reserved for detecting stacked overfitting. This dataset should be in the same format as
+                        `train_data`. If str is passed, `custom_val_data` will be loaded using the str value as the file path.
+                        If `custom_val_data` is not None, the sub-fit fit on all of `train_data`.
```
Innixma, line 786: Clarify that `custom_val_data` is incorporated into the final fit.
Rename to `tuning_data` to be consistent with the fit() API.

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -1023,9 +1067,217 @@ def fit(
             calibrate_decision_threshold=calibrate_decision_threshold,
             infer_limit=infer_limit,
         )
-        self.save()
+        if dynamic_stacking:
+            logger.info(
```
Innixma, line 1071: Update to `logger.log(20, "Dynamic ...`

Puts it higher in logging level so it shows up by default.

Innixma: ```f"Dynamic stacking is enabled (dynamic_stacking={dynamic_stacking}). AutoGluon will ...```

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -1023,9 +1067,217 @@ def fit(
...
+        ag_post_fix_kwargs,
+        use_holdout=None,
+        detection_time_frac=None,
+        holdout_frac=None,
+        n_folds=None,
+        n_repeats=None,
+        memory_safe_fits=None,
+        clean_up_fits=None,
+        custom_val_data=None,
+    ):
```
Innixma, line 1108: add a full docstring with type hints

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -1023,9 +1067,217 @@ def fit(
...
+        self._fit(ag_fit_kwargs, ag_post_fix_kwargs)
+
         return self
 
+    def _fit(self, ag_fit_kwargs, ag_post_fix_kwargs):
+        self.save(silent=True)  # Save predictor to disk to enable prediction and training after interrupt
+        self._learner.fit(**ag_fit_kwargs)
+        self._set_post_fit_vars()
+        self._post_fit(**ag_post_fix_kwargs)
+        self.save()
```
Innixma, line 1094: type hints

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -3871,6 +4124,16 @@ def _fit_extra_kwargs_dict(self):
...
+            ds_args=dict(
+                use_holdout=True,
+                detection_time_frac=1 / 4,
+                holdout_frac=1 / 9,
+                n_folds=2,
+                n_repeats=1,
+                memory_safe_fits=True,
+                clean_up_fits=True,
+                custom_val_data=None,
+            ),
```
Innixma, line 4136: If it isn't supported at fit_extra time, why is part of the fit_extra_kwargs_dict?

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -3927,6 +4190,24 @@ def _validate_fit_extra_kwargs(self, kwargs, extra_valid_keys=None):
...
+        for arg_name in ["use_holdout", "memory_safe_fits", "clean_up_fits"]:
+            if (arg_name in ds_args) and (not isinstance(ds_args[arg_name], bool)):
+                raise ValueError(f"`{arg_name}` in `ds_args` must be bool.  Got: {type(ds_args[arg_name])}")
+        for arg_name in ["detection_time_frac", "holdout_frac"]:
+            if (arg_name in ds_args) and ((not isinstance(ds_args[arg_name], float)) or (ds_args[arg_name] >= 1) or (ds_args[arg_name] <= 0)):
+                raise ValueError(f"`{arg_name}` in `ds_args` must be float in (0,1).  Got: {type(ds_args[arg_name])}, {ds_args[arg_name]}")
+        if ("n_folds" in ds_args) and ((not isinstance(ds_args["n_folds"], int)) or (ds_args["n_folds"] < 2)):
+            raise ValueError(f"`n_folds` in `ds_args` must be int in [2, +inf).  Got: {type(ds_args['n_folds'])}, {ds_args['n_folds']}")
+        if ("n_repeats" in ds_args) and ((not isinstance(ds_args["n_repeats"], int)) or (ds_args["n_repeats"] < 1)):
+            raise ValueError(f"`n_repeats` in `ds_args` must be int in [1, +inf).  Got: {type(ds_args['n_repeats'])}, {ds_args['n_repeats']}")
```
Innixma, line 4209: move to its own method that is called here.

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -4085,7 +4366,11 @@ def _sanitize_stack_args(self, num_bag_folds, num_bag_sets, num_stack_levels, ti
                 num_bag_sets = 1
         if not isinstance(num_bag_sets, int):
             raise ValueError(f"num_bag_sets must be an integer. (num_bag_sets={num_bag_sets})")
-        return num_bag_folds, num_bag_sets, num_stack_levels
+
+        if dynamic_stacking and num_stack_levels < 1:
+            logger.info("Dynamic stacking was enabled but stacking is disabled. Setting dynamic stacking to False.")
```
Innixma, line 4371: `logger.log(20,`

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -4274,3 +4559,59 @@ def from_learner(cls, learner: AbstractTabularLearner):
         predictor = cls(label=learner.label, path=learner.path)
         predictor._set_post_fit_vars(learner=learner)
         return predictor
+
+
+def _sub_fit(self, train_data, time_limit, ds_fit_kwargs, ag_fit_kwargs, ag_post_fix_kwargs, custom_val_data=None):
```
Innixma, line 4564: docstring + type hints

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -4274,3 +4559,59 @@ def from_learner(cls, learner: AbstractTabularLearner):
...
+        val_data = train_data.iloc[val_indices]
+        train_data = train_data.iloc[train_indices]
+    elif custom_val_data is not None:
+        train_data = train_data
+        val_data = custom_val_data
+    else:
+        raise ValueError("Unsupported validation procedure during dynamic stacking!")
+
+    set_logger_verbosity(ag_fit_kwargs["verbosity"])
+    org_learner = copy.deepcopy(self._learner)
```
Innixma, line 4586: nit: `learner_og`

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -4274,3 +4559,59 @@ def from_learner(cls, learner: AbstractTabularLearner):
...
+
+    ag_fit_kwargs["X"] = train_data
+    ag_fit_kwargs["time_limit"] = time_limit
+    ds_fit_context = ds_fit_kwargs.get("ds_fit_context")
+    clean_up_fits = ds_fit_kwargs.get("clean_up_fits")
+
+    self._learner.set_contexts(ds_fit_context)
+    self._fit(ag_fit_kwargs, ag_post_fix_kwargs)
```
Innixma, line 4594: `ag_post_fix` is this a typo for `ag_post_fit`?

Innixma: always call by keyword arguments. Improves readability and reduces risk of bugs.

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -1023,9 +1067,217 @@ def fit(
...
+        ag_fit_kwargs["X"] = X
+        ag_fit_kwargs["X_val"] = X_val
+        ag_fit_kwargs["X_unlabeled"] = X_unlabeled
+
+        logger.info(f"Starting full fit now with num_stack_levels {num_stack_levels}.")
+        return num_stack_levels, time_limit_fit_full
+
+    def _sub_fit_memory_save_wrapper(self, train_data, time_limit, ds_fit_kwargs, ag_fit_kwargs, ag_post_fix_kwargs, custom_val_data=None):
+        """Tries to run the sub-fit in a subprocess (managed by ray). Similar to AutoGluon's parallel fold fitting strategies,
+        this code does not shut down ray after usage. Otherwise, we would also kill outer-scope ray usage."""
```
Innixma, line 1226: type hints

tabular/src/autogluon/tabular/predictor/predictor.py
Innixma, line 4586: Why are we creating a copy of the learner instead of a copy of the predictor? (both are probably fine, but wanted to know the reasoning)

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -1023,9 +1067,217 @@ def fit(
...
+
+            del splits
+
+        if clean_up_fits:
+            shutil.rmtree(ds_fit_context)
+
+        # -- Determine rest time and new num_stack_levels
+        time_spend_sub_fits = int(time.time() - time_start)
+        logger.info(f"Spend {time_spend_sub_fits} seconds for the sub-fit(s) during dynamic stacking.")
+        if org_time_limit is None:
```
Innixma, line 1206: time_limit_og

Innixma (commented):
Added initial review

tabular/src/autogluon/tabular/predictor/predictor.py
LennartPurucker: `custom_val_data` is so far not incorporated into the final fit. Will add that. But I am not sure about naming `tuning_data`. Tuning data sounds very misleading here.

tabular/src/autogluon/tabular/predictor/predictor.py
LennartPurucker: the argument for this function are validated in `_validate_fit_extra_kwargs` and the docstring/documentation is in the docstring/documentation of `fit()`. 

Do we want to have this in multiple places?

tabular/src/autogluon/tabular/predictor/predictor.py
LennartPurucker: Because that is called/part of `_validate_fit_extra_kwargs`. I thought this dict refers to the extra kwargs of fit and not the kwargs of fit_extra. There is no function like `_validate_fit_kwargs` AFAIK.

LennartPurucker: Moved it to `_validate_fit_kwargs`

tabular/src/autogluon/tabular/predictor/predictor.py
LennartPurucker: To my understanding the only part of the predictor that is affected / needed for a reset to the original state before a fit is the learner. I think copying the predictor would be more expensive (copying useless stuff) than copying the learner. But for safety, we could also copy the predictor and reset the object this way.

tabular/src/autogluon/tabular/predictor/predictor.py
LennartPurucker: As the docstrting for the parameters is in the docstring of`fit()`, I only added a "smaller" docstring here to explain the method but not the full docstring. 

Should I add a docstrting for the parameters here too, i.e., at two places?

Innixma: As long as it explains where one can find the definitions, it is fine.

Innixma:
Benchmark results look good! Thanks for the amazing contribution @LennartPurucker!

Innixma (approved):
LGTM!

PR #3626, opened 2023-10-27
[Feature][Timeseries] Benchmark Dashboard Integration

prateekdesai04:
*Description of changes:*
This PR incorporates benchmarking and visualization of AG Timeseries.
Benchmark on the Master Branch are ran sequentially as updating the same CloudFormation bench stack simultaneously can lead to errors and making two stacks will be overkill of resources.

These changes use the latest AG-Bench and not the PyPi release.
This PR works with the PR in AG Bench related to timeseries: https://github.com/autogluon/autogluon-bench/pull/64

### DIFF
--- a/common/src/autogluon/common/utils/pandas_utils.py
+++ b/common/src/autogluon/common/utils/pandas_utils.py
@@ -47,7 +47,7 @@ def get_approximate_df_mem_usage(df: DataFrame, sample_ratio=0.2):
                 num_categories = len(df[column].cat.categories)
                 num_categories_sample = math.ceil(sample_ratio * num_categories)
                 sample_ratio_cat = num_categories_sample / num_categories
-                memory_usage[column] = (
+                memory_usage[column] = int(
                     df[column].cat.codes.dtype.itemsize * num_rows
                     + df[column].cat.categories[:num_categories_sample].memory_usage(deep=True) / sample_ratio_cat
                 )


.github/workflow_scripts/run_benchmark.sh
```diff
@@ -14,31 +14,31 @@ source $(dirname "$0")/env_setup.sh
 setup_benchmark_env
 
 /bin/bash CI/bench/generate_bench_config.sh $MODULE $PRESET $BENCHMARK $TIME_LIMIT $BRANCH_OR_PR_NUMBER
-agbench run $MODULE"_cloud_configs.yaml" --wait
+agbench run $MODULE"_cloud_configs.yaml" --dev-branch https://github.com/autogluon/autogluon-bench.git#master --wait
```
tonyhoo, line 17: What is the intention of passing the -dev-branch explicitly? What is the default

.github/workflow_scripts/run_benchmark.sh
```diff
@@ -14,31 +14,31 @@ source $(dirname "$0")/env_setup.sh
...
 
 # If it is a PR, fetch the cleaned file of master-evaluation
 if [ $BRANCH_OR_PR_NUMBER != "master" ]
 then
     # Capture the name of the file, rename it and store it in ./results
-    master_cleaned_file=$(aws s3 ls s3://autogluon-ci-benchmark/cleaned/master/latest/ | awk '{print $NF}')
+    master_cleaned_file=$(aws s3 ls s3://autogluon-ci-benchmark/cleaned/$MODULE/master/latest/ | awk '{print $NF}')
     new_master_cleaned_file="master_${master_cleaned_file}"
-    aws s3 cp --recursive s3://autogluon-ci-benchmark/cleaned/master/latest/ ./results
+    aws s3 cp --recursive s3://autogluon-ci-benchmark/cleaned/$MODULE/master/latest/ ./results
```
tonyhoo, line 25: Is it correct to assume that the `MODULE` here pertains to the categories of `tabular`, `multimodal`, and `timeseries`?

.github/workflows/benchmark-command.yml
```diff
@@ -25,12 +26,15 @@ on:
           - high
           - good
           - medium
+          - dev
```
tonyhoo, line 29: What does the dev means here in terms of preset? Do we have such preset supported in AG and bench?

.github/workflows/benchmark-command.yml
```diff
@@ -25,12 +26,15 @@ on:
...
       benchmark:
         description: 'Benchmark to run'
         required: true
         options:
           - full
           - test
+          - small
+          - timeseries_test
```
tonyhoo, line 37: same here, these value options are a bit confusing

tonyhoo:
Thank you for your effort. Could you include a PR link demonstrating a successful timeseries run with dashboard link?

.github/workflows/benchmark-command.yml
prateekdesai04: I think here it is the versions of AG.
Yes, we have such preset supported for timeseries.

.github/workflow_scripts/run_benchmark.sh
prateekdesai04: Yes, the assumption is correct.

.github/workflows/benchmark-command.yml
```diff
@@ -40,6 +44,7 @@ on:
           - 8h
           - 16h
           - 24h
+          - test101
```
shchur, line 47: This might apply more to [`autogluon.bench`](https://github.com/autogluon/autogluon-bench/blob/master/sample_configs/amlb_configs/constraints.yaml), but should we make the constraint names more consistent with other AMLB constraints (e.g., `10m4c` for 10 minutes, 4 cores rather than `test101`)?

.github/workflows/benchmark-command.yml
prateekdesai04: Yes

CI/bench/timeseries/amlb_user_dir/frameworks_template.yaml
```diff
@@ -0,0 +1,18 @@
...
+### AutoML frameworks ###
+#########################
+
+######### Do Not Remove #########
+AutoGluon:
+  version: "latest"
+######### Do Not Remove #########
+
+
+AutoGluon_dev:
```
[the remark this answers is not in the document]
prateekdesai04: Yes we need it, using this template file, the repo and branch is overwritten with that of the PR repo and branch.

prateekdesai04:
> Thank you for your effort. Could you include a PR link demonstrating a successful timeseries run with dashboard link?

https://github.com/prateekdesai04/autogluon/pull/30#issuecomment-1786683574
(this was put locally in order to test, sometimes the dashboard may change if a newer benchmark is run)

CI/bench/timeseries/amlb_user_dir/constraints.yaml
```diff
@@ -0,0 +1,7 @@
+---
+
+test101:
```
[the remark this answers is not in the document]
prateekdesai04: Yes, we need the constraint as this will be picked up by Bench running in our CI.
Basically the flow is as follows: we use our own custom configs to specify constraints, presets, frameworks and benchmark datasets, and for bench to pick this up we need to store all this in something called _amlb-user-dir_, then a cloud config file is generated which ensures the custom configs are picked up rather than the default AMLB configs.
If nothing is provided, AMLB Configs are picked up by default.

.github/workflows/benchmark-command.yml
prateekdesai04: Yes, sure I can make this change!

.github/workflows/benchmark-command.yml
prateekdesai04: I can make changes to keep a better naming convention

.github/workflows/benchmark-command.yml
tonyhoo: I'd appreciate some clarification on the need for this new dev preset. Could you provide further information on why it is being introduced?

.github/workflow_scripts/run_benchmark.sh
tonyhoo: Thanks for clarification

.github/workflows/benchmark-command.yml
prateekdesai04: It's because AG-Bench uses this preset for timeseries - https://github.com/autogluon/autogluon-bench/blob/master/sample_configs/amlb_configs/frameworks_example.yaml
And hence adding dev will make it pick up the version specified in the frameworks.yaml file

CI/bench/timeseries/amlb_user_dir/constraints.yaml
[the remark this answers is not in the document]
shchur: I see, so [these configs](https://github.com/autogluon/autogluon-bench/blob/master/sample_configs/amlb_configs/constraints.yaml#L3) are available in the `autogluon.bench` repo, but they aren't automatically downloaded when we install bench with `pip install autogluon.bench`. So we need to store these files somewhere to use them, right?

shchur (approved):
Thanks for addressing the comments, @prateekdesai04!

You might also want to wait for @tonyhoo's approval since I don't have much context about the workings of this part of the codebase.

PR #3632, opened 2023-10-30
[timeseries] Add scaled quantile loss metric

shchur:
*Description of changes:*
- Add scaled quantile loss metric
- Use new GluonTS metrics from `gluonts.ev` instead of the old Evaluator from `gluonts.evaluation` in the parity checks.


canerturkmen (approved):
LGTM!

PR #3635, opened 2023-10-30
[timeseries] Improve handling of static_features by TimeSeriesDataFrame

shchur:
## Problem

Currently, there are only two ways for the user to create a `TimeSeriesDataFrame` with `static_features`.

1. After creating a TSDF in any way (e.g., using `from_data_frame`, `from_path`), we attach the static features as
    ```python
    tsdf = TimeSeriesDataFrame.from_data_frame(...)
    tsdf.static_features = static_df
    ```
2. Set at creation time (this approach is not documented)
    ```python
    tsdf = TimeSeriesDataFrame(..., static_features=static_df)
   ```

In both of these cases, `static_df` must be a `pd.DataFrame` with `item_id` **set as index**. This is rather restrictive and is not compatible with other methods for TSDF creation such as `TimeSeriesDataFrame.from_data_frame`:
- What if `item_id` is a column and not the index?
- What if the item ID column is not called `item_id`?

## Proposed solution

This PR adds support for new ways to attach `static_features` to a `TimeSeriesDataFrame`.

3. Using [`from_data_frame`](https://auto.gluon.ai/stable/api/autogluon.timeseries.TimeSeriesDataFrame.from_data_frame.html) constructor
    ```python
    tsdf = TimeSeriesDataFrame.from_data_frame(data, id_column="custom_id_column", static_features_df=static_df)
    ```
    This method assumes that `static_df` has a column with name `"custom_id_column"` that contains the item ids.
4. Using [`from_path`](https://auto.gluon.ai/stable/api/autogluon.timeseries.TimeSeriesDataFrame.from_path.html) constructor
    ```python
    tsdf = TimeSeriesDataFrame.from_path(data_path, id_column="custom_id_column", static_features_path="static.csv")
    ```
    This method assumes that `static.csv` contains a column with name `"custom_id_column"` that contains the item ids.
5. Methods (1) and (2) described above now also work if `item_id` is a column of `static_df` and not necessarily the index.
--------
## This PR

*Description of changes*:
- Enable passing `static_features` in `TimeSeriesDataFrame.from_data_frame` and `TimeSeriesDataFrame.from_path` methods.
- Automatically use the `item_id` column as index, if it's present in `static_features`.
- Support passing paths as `pathlib.Path` type (before only `str` was supported).
- Update `forecasting-indepth` tutorial to reflect new changes to the API.
- Simplify the internal logic of `TimeSeriesDataFrame` creation.

### DIFF
--- a/timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
+++ b/timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
@@ -32,7 +32,7 @@ class TimeSeriesDataFrame(pd.DataFrame):
 
     Parameters
     ----------
-    data : Any
+    data : pd.DataFrame, str, pathlib.Path or Iterable
         Time series data to construct a ``TimeSeriesDataFrame``. The class currently supports four input formats.
 
         1. Time series data in a pandas DataFrame format without multi-index. For example::
@@ -48,9 +48,13 @@ class TimeSeriesDataFrame(pd.DataFrame):
                 7        2 2019-01-02       7
                 8        2 2019-01-03       8
 
-        2. Time series data in pandas DataFrame format with multi-index on ``item_id`` and ``timestamp``. For example::
+            You can also use :meth:`~autogluon.timeseries.TimeSeriesDataFrame.from_data_frame` for loading data in such format.
 
-                                        target
+        2. Path to a data file in CSV or Parquet format. The file must contain columns ``item_id`` and ``timestamp``, as well as columns with time series values. This is similar to Option 1 above (pandas DataFrame format without multi-index). Both remote (e.g., S3) and local paths are accepted. You can also use :meth:`~autogluon.timeseries.TimeSeriesDataFrame.from_path` for loading data in such format.
+
+        3. Time series data in pandas DataFrame format with multi-index on ``item_id`` and ``timestamp``. For example::
+
+                                    target
                 item_id timestamp
                 0       2019-01-01       0
                         2019-01-02       1
@@ -62,8 +66,6 @@ class TimeSeriesDataFrame(pd.DataFrame):
                         2019-01-02       7
                         2019-01-03       8
 
-        3. Path to a data file in CSV or Parquet format. The file must contain columns ``item_id`` and ``timestamp``, as well as columns with time series values. This is similar to Option 1 above (pandas DataFrame format without multi-index). Both remote (e.g., S3) and local paths are accepted.
-
         4. Time series data in Iterable format. For example::
 
                 iterable_dataset = [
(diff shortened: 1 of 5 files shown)


timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
```diff
@@ -107,7 +129,7 @@ def __init__(self, data: Any, static_features: Optional[pd.DataFrame] = None, *a
                 self._validate_multi_index_data_frame(data)
             else:
                 data = self._construct_pandas_frame_from_data_frame(data)
-        elif isinstance(data, str):
+        elif isinstance(data, (str, Path)):
```
canerturkmen, line 132: why not do `static_features: Any` and also accept a path there while we're at it?

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
```diff
@@ -3,7 +3,9 @@
 import copy
```
canerturkmen, line 3: This class seems to be getting a little entangled with dependencies on how it constructs things. Now looks like a good time to overhaul this. Specifically, 
- `from_` constructors seem to be able to call `__init__`. 
- Everyone seems to be able to call internal `_construct_*` methods, and `load_pd`
- `from_` constructors seem to be able to call each other.

This has also resulted in some redundancies such as 
- `_load_data_frame_from_file`.. which seems to have only one caller and its use appears redundant.
- `from_iterable` looks like it may not need a call to `_construct_.._from_iterable` as the call to `cls(...)` would anyway handle that. 

Normally `from_` constructors are there to explicitly prevent us from overloading the constructor. If we're allowing the constructor to take `Any`, then `from_` is not necessary unless for reasons for backward compatibility. This is why maybe we could redesign to ensure `from_` constructors just wrap `__init__` for now, and don't call private methods.

canerturkmen (commented):
Looks good overall and I agree with the intent. Left some comments...

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
shchur: Good idea, I've implemented this change.

Now `TimeSeriesDataFrame` accepts `data` in `DataFrame` / `str` / `Path` / `Iterable` formats. `from_*` methods are just thin wrappers that call `TimeSeriesDataFrame.__init__` and don't use internal methods.

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -127,7 +128,7 @@ def __init__(
         freq: str = None,
         eval_metric: Union[str, TimeSeriesScorer, None] = None,
         eval_metric_seasonal_period: Optional[int] = None,
-        path: Optional[str] = None,
+        path: Optional[Union[str, Path]] = None,
```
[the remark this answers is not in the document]
shchur: For TS, the only user-exposed methods that deal with paths are the following. 
- `TimeSeriesPredictor`:
  - `__init__`
  - `load`
- `TimeSeriesDataFrame`
  - `__init__`
  - `from_path`
  - `from_pickle`
All of them support `pathlib.Path` objects now

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
```diff
@@ -136,92 +181,81 @@ def _constructor_from_mgr(self, mgr, axes):
...
-            self._cached_freq = IRREGULAR_TIME_INDEX_FREQSTR
-            return None
+        cls._validate_data_frame(df)
+        return df.set_index([ITEMID, TIMESTAMP])
 
-        freq = freq.freqstr if isinstance(freq, pd._libs.tslibs.BaseOffset) else freq
-        self._cached_freq = freq
-        return freq
+    @classmethod
+    def _load_tsdf_from_path(
```
canerturkmen, line 213: looks like this method is only used in `__init__` ? do we need it?

canerturkmen (approved):
LGTM! Thanks for the updates!

PR #3639, opened 2023-10-30
[AutoMM] Avoid running DDP multiple times in one process

FANGAreNotGnu:

### DIFF
--- a/multimodal/src/autogluon/multimodal/matcher.py
+++ b/multimodal/src/autogluon/multimodal/matcher.py
@@ -92,6 +92,7 @@ from .utils import (
     init_df_preprocessor,
     load_text_tokenizers,
     predict,
+    run_ddp_only_once,
     save_pretrained_model_configs,
     save_text_tokenizers,
     select_model,
@@ -944,6 +945,9 @@ class MultiModalMatcher:
         # save artifacts for the current running, except for model checkpoint, which will be saved in trainer
         self.save(save_path)
 
+        # After first DDP run, change following DDP runs to single GPU
+        num_gpus, strategy = run_ddp_only_once(num_gpus, strategy)
+
         blacklist_msgs = ["already configured with model summary"]
         log_filter = LogFilter(blacklist_msgs)
         with apply_log_filter(log_filter):
@@ -1277,6 +1281,9 @@ class MultiModalMatcher:
             signature=signature,
             match_label=match_label,
         )
+
+        # After first DDP run, change following DDP runs to single GPU
+        num_gpus, strategy = run_ddp_only_once(num_gpus, strategy)
 
         blacklist_msgs = []
         if self._verbosity <= 3:  # turn off logging in prediction
--- a/multimodal/src/autogluon/multimodal/predictor.py
+++ b/multimodal/src/autogluon/multimodal/predictor.py
@@ -160,6 +160,7 @@ from .utils import (
     modify_duplicate_model_names,
     object_detection_data_to_df,
     predict,
+    run_ddp_only_once,
     save_ovd_result_df,
     save_pretrained_model_configs,
     save_result_df,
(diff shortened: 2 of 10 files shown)


multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -1816,7 +1823,7 @@ def _default_predict(
         with apply_log_filter(log_filter):
             evaluator = pl.Trainer(
                 accelerator="gpu" if num_gpus > 0 else "auto",
-                devices=num_gpus if num_gpus > 0 else "auto",
+                devices=get_available_devices(num_gpus, self._config.env.auto_select_gpus),
```
zhiqiangdon, line 1826: `self._config` - > `config`

multimodal/src/autogluon/multimodal/predictor.py
```diff
@@ -1505,6 +1506,9 @@ def _fit(
         # save artifacts for the current running, except for model checkpoint, which will be saved in trainer
         self.save(save_path, standalone=standalone)
 
+        # After first DDP run, change following DDP runs to single GPU
+        num_gpus, strategy = run_ddp_only_once(num_gpus, strategy)
```
zhiqiangdon, line 1510: Need to move this above line 1502 because `num_gpus` needs update in `config`.

multimodal/src/autogluon/multimodal/matcher.py
```diff
@@ -944,6 +945,9 @@ def _fit(
         # save artifacts for the current running, except for model checkpoint, which will be saved in trainer
         self.save(save_path)
 
+        # After first DDP run, change following DDP runs to single GPU
+        num_gpus, strategy = run_ddp_only_once(num_gpus, strategy)
```
zhiqiangdon, line 949: Need to move this above line 941 because num_gpus needs update in config.

multimodal/src/autogluon/multimodal/utils/environment.py
```diff
@@ -322,7 +322,11 @@ def get_available_devices(num_gpus: int, auto_select_gpus: bool):
     -------
     The available devices.
     """
-    if num_gpus > 0:
+    if num_gpus == 1:
+        # for single GPU just return GPU 0,
+        # find_usable_cuda_devices can cause error if there is previous DDP run
```
zhiqiangdon, line 327: What's the reason of error?

multimodal/src/autogluon/multimodal/utils/environment.py
```diff
@@ -347,3 +351,15 @@ def _get_mmlab_installation_guide(package_name):
         raise ValueError("Available package_name are: mmdet, mmcv, mmengine.")
 
     return err_msg
+
+
+def run_ddp_only_once(num_gpus, strategy):
+    if strategy in DDP_STRATEGIES:
+        global FIRST_DDP_RUN  # Use the global variable to make sure it is tracked per process
```
zhiqiangdon, line 358: Is this global variable shared across processes? If ddp creates 1 additional process running from the beginning, the new process would be set `num_gpus` 1 and strategy `auto` when running this function?

multimodal/src/autogluon/multimodal/predictor.py
FANGAreNotGnu: But at next time it's loaded it will run in single GPU mode?

multimodal/src/autogluon/multimodal/utils/environment.py
FANGAreNotGnu: I think it's still because there's extra GPU usage in unused GPUs in previous DDP run. And future access to other GPUs after the DDP run will be blocked.

multimodal/src/autogluon/multimodal/utils/environment.py
FANGAreNotGnu: The new DDP process can't see the previous global variable.

FANGAreNotGnu:
Fixed in #3665

PR #3646, opened 2023-11-01
[timeseries] Support passing multiple evaluation metrics to TimeSeriesPredictor.evaluate

shchur:
*Description of changes:*
-  The behavior of `TimeSeriesPredictor.evaluate` now aligns with Tabular/Multimodal predictors:
   - it now returns a dictionary where keys = metric names, values = scores along each metric
   - user can now pass multiple metrics to `evaluate`
- Deprecate the method  `TimeSeriesPredictor.score` as it's just an alias for `TimeSeriesPredictor.evaluate`.
- Do not log a warning if user provides the old name for the `mean_wQuantileLoss` metric. 


Innixma (approved):
LGTM!

PR #3653, opened 2023-11-01
[timeseries] Fix bug caused by `known_covariates_names` provided as a pd.Index instead of list

shchur:
*Description of changes:*
- If user provides `known_covariates_names` as a `np.ndarray` or `pd.Index` instead of `List`, this causes a bug for [MLForecsat models](https://github.com/autogluon/autogluon/blob/master/timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py#L212C7-L212C7). Instead of concatenating two lists like `["1", "2"] + ["B"] = ["1", "2", "B"]`, we accidentally append to each entry in the first list as `np.array(["1", "2"]) + ["B"] = ["1B", "2B"]`. Then, the model fails because it cannot find these columns if the dataframe.
- Fix the test that is supposed to catch this bug.

### DIFF
--- a/timeseries/src/autogluon/timeseries/predictor.py
+++ b/timeseries/src/autogluon/timeseries/predictor.py
@@ -156,7 +156,7 @@ class TimeSeriesPredictor:
             )
         if self.target in known_covariates_names:
             raise ValueError(f"Target column {self.target} cannot be one of the known covariates.")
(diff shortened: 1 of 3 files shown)


canerturkmen (approved):
LGTM!

PR #3655, opened 2023-11-02
[timeseries] Implement timeseries deprecations scheduled for v1.0

shchur:
*Description of changes:*
- Deprecate `TimeSeriesDataFrame.get_reindexed_view` and `TimeSeriesDataFrame.to_regular_index` methods.
- Deprecate all old splitter classes located in `autogluon.timeseries.splitter`
  - Move new validation splitter to `autogluon.timeseries.val_splitter` to avoid clashes with the deprecated classes.
  - Update the in-depth tutorial to use the new splitter class.
- Deprecate `ignore_time_index` argument to `TimeSeriesPredictor.__init__()`
- Add error message for deprecated method `TimeSeriesPredictor.score()`
- Remove `TimeSeriesEvaluator` from `autogluon.timeseries` namespace 

Deprecations that I decided not to implement:
- I decided to keep the old name `TimeSeriesPredictor.fit(enable_ensemble: bool = True)` instead of renaming the argument to `TimeSeriesPredictor.fit(fit_weighted_ensemble: bool = True)`, as done in Tabular. We might implement ensembling strategies that are more general than weighted ensemble in time series, so I'd like to keep the more general name.

*To do in a future PR:*
- Make sure that deprecated method don't show up in API documentation on the website.


timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
```diff
@@ -952,3 +877,23 @@ def convert_frequency(
         )
         resampled_df.static_features = self.static_features
         return resampled_df
+
+    def __dir__(self) -> List[str]:
+        # This hides method from IPython autocomplete, but not VSCode autocomplete
+        deprecated = ["get_reindexed_view", "to_regular_index"]
+        return [d for d in super().__dir__() if d not in deprecated]
```
shchur, line 884: @Innixma Here is a partial solution that I found for hiding some methods from autocomplete (inspired by [this SO answer](https://stackoverflow.com/questions/2531479/hide-deprecated-methods-from-tab-completion)). I verified that this works for IPython, but it doesn't work for VSCode/Intellisense, unfortunately.

Please let me know if you find anything more general that deals with all autocomplete engines.

canerturkmen: nice! but why is it `super().__dir__` and not `self.__dir__` ? also, is there a solution to hardcoding the method names?

timeseries/src/autogluon/timeseries/evaluator.py
```diff
@@ -8,7 +8,10 @@
 @Deprecated(
     min_version_to_warn="1.0",
     min_version_to_error="1.1",
```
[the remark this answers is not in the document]
canerturkmen: Now is the time to gut it I think. Especially as we have the eval stuff in docs anyway and it shouldn't be very hard to migrate even if someone was using it?

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -1048,3 +1058,12 @@ def refit_full(self, model: str = "all", set_best_to_refit_full: bool = True) ->
                     f"Training may have failed on the refit model. AutoGluon will default to using '{model_best}' for predict()."
                 )
         return refit_full_dict
+
+    def __dir__(self) -> List[str]:
+        # This hides method from IPython autocomplete, but not VSCode autocomplete
+        deprecated = ["score"]
+        return [d for d in super().__dir__() if d not in deprecated]
+
+    @Deprecated(min_version_to_warn="1.0", min_version_to_error="1.0", new="evaluate")
```
canerturkmen, line 1067: so it both warns and errors?

timeseries/src/autogluon/timeseries/val_splitter.py
```diff
@@ -0,0 +1,51 @@
+from typing import Iterator, Optional, Tuple
```
canerturkmen, line 1: Why the change in namespace? Aren't these objects just generic splitters? Can they only be used during validation? Are there different splitters for use in training?

canerturkmen (approved):
LGTM! I'm just a little confused about the namespace renaming of `splitter`, but not a blocker. Thank you!

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
shchur: I think `self.__dir__` would just go into infinite recursion, did you mean `self.__dict__.keys()` or something like that? `super().__dir__` should take the default implementation of the dir method by `object`, as it's the only parent class that implements it.

Could you elaborate on hardcoding the method names?

timeseries/src/autogluon/timeseries/predictor.py
shchur: It just errors in this case. This decorator comes with functionality for both warning & raising errors. 

Actually, I think it would be cleaner if we just raise the ValueErrors ourselves, since there is no warning necessary for the breaking changes. This way we'll be certain that the exceptions are raised correctly.

timeseries/src/autogluon/timeseries/val_splitter.py
shchur: The only reason for this change is to avoid confusion among new users. 

If `ExpandingWindowSplitter` stays in the `autogluon.timeseries.splitter` namespace and user types `from autogluon.timeseries.splitter import ...` + clicks autocomplete, they will receive suggestions that include deprecated classes (`MultiWindowSplitter`, `LastWindowSplitter`). This seems confusing to me, so I thought about moving all non-deprecated classes to `autogluon.timeseries.val_splitter` and never telling the users about the `autogluon.timeseries.splitter` namespace at all.

Do you think this is a reasonable enough justification?

timeseries/src/autogluon/timeseries/predictor.py
canerturkmen: I rather meant, if we're just going to error why do we still have warn as a kwarg?

timeseries/src/autogluon/timeseries/val_splitter.py
canerturkmen: mm, I don't think so. Even if so, I'd put the deprecated stuff to another namespace and remove it later. But I'm OK with just keeping everything in `splitter` too.

timeseries/src/autogluon/timeseries/predictor.py
shchur: I've updated the PR to not use this decorator, so this shouldn't be a problem anymore.

I think the motivation for the decorator was to introduce DeprecationWarnings that are later replaced by breaking changes. This isn't our use case - we just want to break everything immediately.

timeseries/src/autogluon/timeseries/val_splitter.py
shchur: Fair enough, I agree that having intuitive names for namespaces is more important than autocomplete. Moved all splitter classes back to `autogluon.timeseries.splitter`.

PR #3657, opened 2023-11-03
[timeseries] Split time limit evenly among forecasting models

shchur:
Some time series models have a hard time following the time limit:
1. Local models like `AutoETS` may take a very long time to train and completely consume the allocated time limit.
2. GluonTS models like DeepAR and TFT always train for max # of epochs until the time limit is reached. 

Both of these facts can cause problems: If each model is given 100% of the remaining time limit, it may happen that
1. `AutoETS` consumes all of the time and still fails when the time limit is exceeded because the model has no early stopping.
2. `DeepAR` takes all of training time. Under a strict time limit constraint, this time would be better spent training other models. For example, it's better to train DeepAR for 20 epochs + TFT for 20 epochs, rather than only train DeepAR for 40 epochs.

Currently, as a solution to this problem, we use a hack to evenly distribute the training time across all models in the `best_quality` presets: we define a dummy search space for DeepAR, which forces the trainer to evenly split the time.

*Description of changes:*

With this PR, we remove the need for this hack. Now the time limit is always split evenly among models.

- Simplify logic of the for-loop across individual models in `AbstractTimeSeriesTrainer._train_multi`.
- Fix the bug, where incorrect `time_limit` was displayed during HPO.

### DIFF
--- a/timeseries/src/autogluon/timeseries/predictor.py
+++ b/timeseries/src/autogluon/timeseries/predictor.py
@@ -377,6 +377,7 @@ class TimeSeriesPredictor:
         refit_every_n_windows: int = 1,
         refit_full: bool = False,
         enable_ensemble: bool = True,
+        split_time_limit: Optional[bool] = None,
         random_seed: Optional[int] = None,
         verbosity: Optional[int] = None,
     ) -> "TimeSeriesPredictor":
@@ -572,6 +573,10 @@ class TimeSeriesPredictor:
         enable_ensemble : bool, default = True
             If True, the ``TimeSeriesPredictor`` will fit a simple weighted ensemble on top of the models specified via
             ``hyperparameters``.
+        split_time_limit : bool, optional
+            If True, ``time_limit`` will be distributed evenly among all models. If False, each model can be trained up
+            to the full remaining total time limit. If None (default), will be automatically set to True if
+            ``hyperparameter_tune_kwargs`` are provided, or False otherwise.
         random_seed : int, optional
             If provided, fixes the seed of the random number generator for all models. This guarantees reproducible
             results for most models (except those trained on GPU because of the non-determinism of GPU operations).
@@ -604,6 +609,7 @@ class TimeSeriesPredictor:
             num_val_windows=num_val_windows,
             refit_full=refit_full,
             enable_ensemble=enable_ensemble,
+            split_time_limit=split_time_limit,
             random_seed=random_seed,
             verbosity=verbosity,
         )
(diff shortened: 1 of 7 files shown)


timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -367,6 +367,7 @@ def fit(
         refit_every_n_windows: int = 1,
         refit_full: bool = False,
         enable_ensemble: bool = True,
+        split_time_limit: Optional[bool] = None,
```
canerturkmen, line 370: I don't follow why this is needed. This is a complication Tabular did not need to have. What's unique to time series that requires this? Can't we default to what Tabular is doing? This feels like a choice to be made in favor of cross-AG consistency at least as much as some of the others we've looked at.

Moreover, I want to force us to think well about the API here. Is this the only way in which users can wish to play with the time limit? Or is this a `time_limit_policy="split"` that can take different values in the future?

Also, in the comments we point out that the best quality search space is a dummy. Is it? I think several models are still trained for DeepAR so hyperparameter search is performed.

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -558,54 +559,51 @@ def _train_multi(
...
+            if time_limit is None:
+                time_left = None
+            else:
+                time_left = time_limit - (time.time() - time_start)
+                if time_left <= 0:
+                    logger.info(f"Stopping training due to lack of time remaining. Time left: {time_left:.2f} seconds")
+                    break
+
+            if time_left is not None and split_time_limit:
+                time_left_for_model = time_left / (num_models - i)
```
canerturkmen, line 579: so this is a different behavior still than what I would expect from the docstring. you're not giving each model equal shots from the beginning, but deciding on the fly which model gets what. However in principle, if models obey time limits strictly, then there should be no regressions I guess.

canerturkmen (commented):
Thanks! Added some comments.

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
shchur: Some models (e.g., Naive or SeasonalNaive) may complete training before the time limit is exceeded. With this logic we can redistribute the extra time among the remaining models.

timeseries/src/autogluon/timeseries/predictor.py
shchur: I agree with you, the API becomes really confusing with this change. I have updated the PR to always split the time limit and remove the option to control it. I will do benchmarking before merging this PR to see whether this change is beneficial.

canerturkmen (approved):
LGTM! Thanks!!

shchur:
This PR achieves 57% win rate over the master branch (58 datasets, 1h limit), so I'm merging this.

PR #3660, opened 2023-11-04
[AutoMM] Fix a bug in FusionMLP when getting output from single models

taoyang1122:
*Issue #, if available:*

*Description of changes:*
- Fix a bug in FusionMLP when getting output from single models.
- Previously, only the output of the last single model will be returned.


multimodal/src/autogluon/multimodal/models/fusion/fusion_mlp.py
```diff
@@ -181,7 +181,7 @@ def get_output_dict(self, features: torch.Tensor, logits: torch.Tensor, multimod
                 per_output = {per_model.prefix: {}}
                 per_output[per_model.prefix][WEIGHT] = torch.tensor(self.loss_weight).to(per_logits.dtype)
                 per_output[per_model.prefix][LOGITS] = per_logits
-            output.update(per_output)
+                output.update(per_output)
```
zhiqiangdon, line 184: Can we add a test to avoid future similar issue? Consider `loss_weight` is None or not in the test.

zhiqiangdon:
Already done in https://github.com/autogluon/autogluon/pull/3690

PR #3662, opened 2023-11-06
[AutoMM] Add the few shot learner

zhiqiangdon:
*Issue #, if available:*

*Description of changes:*

- Refactor previous `FewShotSVMPredictor` into `FewShotSVMLearner`. Now users can import `MultiModalPredictor` to do few shot learning. The API is unified and easier to use.
- Add new problem type `few_shot_classification` and the corresponding presets.
- Refactor the few shot tutorial accordingly, which is more clean and easier to read.
- Move some object detection specific attributes into the object detection learner.
- Remove the `few_shot_text_classification` problem type in favor of `few_shot_classification`.
- Fix multi-gpu training and prediction for the CLIP model.


multimodal/src/autogluon/multimodal/learners/base.py
```diff
@@ -161,13 +161,10 @@ def __init__(
         hyperparameters: Optional[dict] = None,
```
suzhoum, line 161: Unrelated to this line, but do we want to add a default `eval_metric` for this problem type?

zhiqiangdon: Good call! We can add the accuracy as the default.

zhiqiangdon: Added accuracy as the default and tested in unit tests.

multimodal/src/autogluon/multimodal/learners/base.py
```diff
@@ -193,7 +190,7 @@ def __init__(
             - 'image_text_similarity': Text-image similarity problem
             - 'feature_extraction': Extracting feature (only support inference)
             - 'zero_shot_image_classification': Zero-shot image classification (only support inference)
-            - 'few_shot_text_classification': (experimental) Few-shot text classification
+            - 'few_shot_classification': Few-shot classification for image or text data.
```
tonyhoo, line 193: [minor] does this work for matching and document classification?

tonyhoo (approved):

suzhoum (approved):
LGTM! Thanks for the change!

multimodal/src/autogluon/multimodal/learners/base.py
zhiqiangdon: Not tested for document data yet.

zhiqiangdon: Doesn't support matching.

PR #3663, opened 2023-11-06
[DRAFT] Increase the upper bound of torch and lightning to accept 2.1 version

tonyhoo:
*Issue #, if available:*

*Description of changes:*
Increase the upper bound of torch and lightning to accept 2.1 version


core/src/autogluon/core/_setup_utils.py
```diff
@@ -26,9 +26,9 @@
     "networkx": ">=3.0,<4",  # Major version cap
     "tqdm": ">=4.38,<5",  # Major version cap
     "Pillow": ">=9.3,<9.6",  # "<{N+2}" upper cap
-    "torch": ">=2.0,<2.1",  # "<{N+1}" upper cap, sync with common/src/autogluon/common/utils/try_import.py
-    "lightning": ">=2.0.0,<2.1",  # "<{N+1}" upper cap
-    "pytorch_lightning": ">=2.0.0,<2.1",  # "<{N+1}" upper cap, capping `lightning` does not cap `pytorch_lightning`!
+    "torch": ">=2.0,<2.2",  # "<{N+1}" upper cap, sync with common/src/autogluon/common/utils/try_import.py
```
shchur, line 29: As far as I understand, the problem boils down to the following: Running `pip install torch==2.0.0` installs PyTorch compiled with CUDA 11.7 (see [here](https://pytorch.org/get-started/previous-versions/#linux-and-windows-3)). In contrast, running `pip install torch==2.1.0` or just `pip install torch` (as of Nov 7, 2023) installs PyTorch compiled with CUDA 12. Since the environment that we use to run tests comes with CUDA 11.8 installed, it cannot be used to run PyTorch that was compiled with CUDA12.

I don't it's related to any specific code that we have in AutoGluon. 

To fix this problem, we would need to ensure that during tests we install the correct PyTorch version with something like
```
pip install torch~=2.1.0 --index-url https://download.pytorch.org/whl/cu118
```
rather than
```
pip install torch~=2.1.0
```
that we currently use.

Some other things that I noticed:
1. During installation of `multimodal`, actually `torch-2.0.1` [is installed](https://github.com/autogluon/autogluon/actions/runs/6775228028/job/18421678754?pr=3663#step:4:1000), even though we increase the cap in this PR. Potentially, this happens because one of the `multimodal` dependencies caps torch <2.1. Therefore, the multimodal tests pass.
2. I tried creating a fresh Python 3.10 environment on a `p3` instance (with V100 GPU) and installing PyTorch 2.1 in it. Trying to use CUDA results in the same error as shown in [timeseries logs](https://github.com/autogluon/autogluon/actions/runs/6775228028/job/18421679644?pr=3663#step:4:75724). To reproduce:
    ```
    conda env create -n cuda12 python=3.10
    conda activate cuda12
    pip install torch~=2.1.0
    python -c "import torch; torch.zeros(1).cuda()"
   ```
    Output:
    ```
    RuntimeError: The NVIDIA driver on your system is too old (found version 11080). Please update your GPU driver by downloading and installing a new version from the URL: http://www.nvidia.com/Download/index.aspx Alternatively, go to: https://pytorch.org/ to install a PyTorch version that has been compiled with your version of the CUDA driver
    ```

ddelange:
from a quick look at the [available wheels](https://download.pytorch.org/whl/torch/), all should be fine as long as the end user requests the correct cuda version in the (extra) index url.

This will work fine:

```sh
# cu118 has wheels for 2.0.0 through 2.1.2
pip install autogluon.multimodal --force-reinstall --extra-index-url https://download.pytorch.org/whl/cu118
# cu121 has wheels for 2.1.0 through 2.1.2
pip install autogluon.multimodal --force-reinstall --extra-index-url https://download.pytorch.org/whl/cu121
```

Apart from maybe a docs update, are there any other blockers here?

Asking because this PR [blocks](https://github.com/autogluon/autogluon/pull/3190#discussion_r1407576081) support for latest tensorrt (which completes autogluon py3.11 support).

ddelange:
cc @Innixma @AnirudhDagar

Innixma:
@tonyhoo Any updates on the status of this PR?

AnirudhDagar:
thanks for the notif @ddelange

PR #3671, opened 2023-11-08
[Feature][AutoMM] Benchmark integration for multimodal text and text-tabular

prateekdesai04:
*Description of changes:*
This PR incorporates benchmarking and visualization of AG Multimodal Text and Text-Tabular.
These changes use the latest AG-Bench and not the PyPi release.

Test Master Branch Reference: https://github.com/prateekdesai04/autogluon/actions/runs/6788655846
Test PR Reference: https://github.com/prateekdesai04/autogluon/pull/32#issuecomment-1802917450 (text), https://github.com/prateekdesai04/autogluon/pull/32#issuecomment-1803201706 (text-tabular)


.github/workflow_scripts/run_benchmark.sh
```diff
@@ -14,31 +14,44 @@ source $(dirname "$0")/env_setup.sh
...
         aws s3 rm --recursive s3://autogluon-ci-benchmark/cleaned/$MODULE/master/latest/
         aws s3 cp "$file" s3://autogluon-ci-benchmark/cleaned/$MODULE/master/latest/$(basename "$file")
+    elif [[ "$(basename "$file")" != "master"* ]] && [ $MODULE == "multimodal" ]; then
+        aws s3 cp "$file" "s3://autogluon-ci-benchmark/cleaned/$MODULE/$BENCHMARK/$BRANCH_OR_PR_NUMBER/$SHA/$(basename "$file")"
+        aws s3 rm --recursive s3://autogluon-ci-benchmark/cleaned/$MODULE/$BENCHMARK/$BRANCH_OR_PR_NUMBER/latest/
+        aws s3 cp "$file" s3://autogluon-ci-benchmark/cleaned/$MODULE/$BENCHMARK/$BRANCH_OR_PR_NUMBER/latest/$(basename "$file")
+    elif [[ "$(basename "$file")" == "master"* ]] && [ $MODULE == "multimodal" ]; then
+        aws s3 cp "$file" "s3://autogluon-ci-benchmark/cleaned/$MODULE/$BENCHMARK/master/$SHA/$(basename "$file")"
+        aws s3 rm --recursive s3://autogluon-ci-benchmark/cleaned/$MODULE/$BENCHMARK/master/latest/
+        aws s3 cp "$file" s3://autogluon-ci-benchmark/cleaned/$MODULE/$BENCHMARK/master/latest/$(basename "$file")
```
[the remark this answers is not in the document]
prateekdesai04: Because for tabular and timeseries we store it in bucket-name->cleaned->{tabular/timeseries}->master.
We don't have different types of data like text, text-tabular, image as we do for multimodal.
Hence, for better readability and segregation I have it as the following for multimodal:
bucket-name->cleaned->multimodal->{text/text-tabular/image}->master, so we can know which benchmark was run.

suzhoum:
A general comment on good practice of making git commits, we should try to make the commit message more informative, and squash / fixup commits when necessary, so that anyone who refers to the PR can quickly refer to the relevant code.

.github/workflow_scripts/run_benchmark.sh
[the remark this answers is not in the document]
suzhoum: That makes sense. Also I think you can simplify the logic by decoupling the two conditions as below:
```
if [[ "$(basename "$file")" == "master"* ]]; then
          BRANCH_NAME="master"
else
          BRANCH_NAME=$BRANCH_OR_PR_NUMBER
fi
if [[ $MODULE == "multimodal" ]]; then
          SUB_FOLDER="$BENCHMARK/"
else
          SUB_FOLDER=""
fi
```
Then you don't have to repeat similar commands for 4 times.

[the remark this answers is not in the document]
prateekdesai04: We also have to add the check to see if the file does not start with master everytime because we do not want to copy that result as it is the master branch's cleaned result and not the PR branch and with the above code it will not do that, hence we have these many checks

.github/workflow_scripts/env_setup.sh
```diff
@@ -19,7 +19,8 @@ function setup_build_contrib_env {
 
 function setup_benchmark_env {
     # pip install -U autogluon.bench - revert back when latest update in PyPi
-    git clone https://github.com/autogluon/autogluon-bench.git
+    # git clone https://github.com/autogluon/autogluon-bench.git
+    git clone -b automm https://github.com/prateekdesai04/autogluon-bench.git
```
suzhoum, line 23: It's not safe to clone from a master branch. It would be better to pin to a release tag.

.github/workflow_scripts/run_benchmark.sh
```diff
@@ -14,31 +14,50 @@ source $(dirname "$0")/env_setup.sh
 setup_benchmark_env
 
 /bin/bash CI/bench/generate_bench_config.sh $MODULE $PRESET $BENCHMARK $TIME_LIMIT $BRANCH_OR_PR_NUMBER
-agbench run $MODULE"_cloud_configs.yaml" --dev-branch https://github.com/autogluon/autogluon-bench.git#master --wait
+agbench run $MODULE"_cloud_configs.yaml" --dev-branch https://github.com/prateekdesai04/autogluon-bench.git#automm --wait
```
suzhoum, line 17: We should get rid of `--dev-branch`

.github/workflow_scripts/run_benchmark.sh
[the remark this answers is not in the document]
suzhoum: Try something like below? The code should be simplified to reduce redundancy and improve readability:

```
CLEANED_PATH="s3://autogluon-ci-benchmark/cleaned/$MODULE"
EVALUATION_PATH="s3://autogluon-ci-benchmark/evaluation/$MODULE"
BRANCH_NAME="master"
if [[ "$(basename "$file")" != "master"* ]]; then
    BRANCH_NAME="$BRANCH_OR_PR_NUMBER"
fi

if [ $MODULE == "multimodal" ]; then
    CLEANED_PATH="$CLEANED_PATH/$BENCHMARK"
    EVALUATION_PATH="$EVALUATION_PATH/$BENCHMARK"
fi

aws s3 cp "$file" "$CLEANED_PATH/$BRANCH_NAME/$SHA/$(basename "$file")"
aws s3 rm --recursive "$CLEANED_PATH/$BRANCH_NAME/latest/"
aws s3 cp "$file" "$CLEANED_PATH/$BRANCH_NAME/latest/$(basename "$file")"

if [[ "$(basename "$file")" != "master"* ]]; then
    aws s3 cp --recursive ./evaluate "$EVALUATION_PATH/$BRANCH_NAME/$SHA/"
    aws s3 rm --recursive "$EVALUATION_PATH/$BRANCH_NAME/latest/"
    aws s3 cp --recursive ./evaluate "$EVALUATION_PATH/$BRANCH_NAME/latest/"
fi
```

.github/workflows/benchmark_master.yml
```diff
@@ -155,4 +155,86 @@ jobs:
         run: |
           nvm install 20
           npm install -g aws-cdk
-          /bin/bash ./.github/workflow_scripts/run_benchmark.sh multimodal master_medium none g4_12x ${{ env.AG_BRANCH_NAME }} ${{ github.sha }}
+          /bin/bash ./.github/workflow_scripts/run_benchmark.sh multimodal master_medium automm-text g4_12x ${{ env.AG_BRANCH_NAME }} ${{ github.sha }}
```
suzhoum, line 158: Do we only run on multi-gpu?

prateekdesai04: Yes as of now, we can change the constraint, and running both would take a lot of time

suzhoum: Cost wise 12x is 3 times more expensive than 4x. I'm wondering if we want to default to 4x unless the benchmarking time is significantly faster considering the overheads for instance startups.

prateekdesai04: We run this once a week, so it should not be too much.
If we end up using this too much I can always change the constraint

suzhoum: Another aspect to consider is that what our customers care most? Do they mostly run on single GPU or multi-GPUs? Do we even want to run on both on weekly basis?@tonyhoo @zhiqiangdon might chime in if you have a chance.

prateekdesai04: For now only the maintainers can run this to prevent misuse of resources.

suzhoum (approved):
Approving the PR but we might need to adjust the constraint post release. Thanks for the work!

.github/workflows/benchmark_master.yml
suzhoum: Are tabular and time series using the default instance type (g4dn.2x) for benchmark with your current setup?

prateekdesai04: no they're using m5.2x

suzhoum: For tabular, I believe we want to use `m6i.2xlarge`, but I'm not sure about time series.

PR #3672, opened 2023-11-08
[AutoMM] Add a flag to disable realtime for pytest matcher with multi GPU

FANGAreNotGnu:


multimodal/src/autogluon/multimodal/learners/base.py
```diff
@@ -166,6 +166,7 @@ def __init__(
         enable_progress_bar: Optional[bool] = None,
         pretrained: Optional[bool] = True,
         validation_metric: Optional[str] = None,
+        realtime_enabled: Optional[bool] = True,
```
zhiqiangdon, line 169: Instead of adding the new flag, how about setting `realtime=False` for the `predict`, `predict_proba`, `extract_embedding`, and `evaluate` APIs?

zhiqiangdon: BTW, test_predictor should have the same issue for multi-gpu testing. Consider fixing it also in this PR. I think `realtime=False` also works.

FANGAreNotGnu:
Move to #3684

PR #3679, opened 2023-11-09
[AutoMM] Remove the config arg from fit API

zhiqiangdon:
*Issue #, if available:*

*Description of changes:*
Remove `config` from `fit` API to align with tabular and timeseries predictors. Users need to use `hyperparamters` instead.


suzhoum (approved):
LGTM!

multimodal/tests/unittests/predictor/test_predictor.py
```diff
@@ -440,149 +440,6 @@ def test_customizing_model_names(
             assert hasattr(predictor._learner._config.model, per_name)
 
 
-def test_model_configs():
```
taoyang1122, line 443: Do we need to add these to something like test_hyperparameters() ?

zhiqiangdon: This is to test passing a model config through `config`. The hyperparamters are already tested in other tests.

tonyhoo (approved):
LGTM, please make sure all tests passing before merge

PR #3681, opened 2023-11-09
[AutoMM] Support multi-column input for few shot learning

zhiqiangdon:
*Issue #, if available:*

*Description of changes:*

- Support a dataframe has multiple feature columns in few shot classification.
- Add unit tests.


taoyang1122 (approved):
LGTM!

multimodal/src/autogluon/multimodal/learners/few_shot_svm.py
```diff
@@ -379,21 +381,27 @@ def fit_per_run(
...
             if len(features) == 1:
                 return next(iter(features.values()))
-            if self._config.data.aggregate_column_features == "concat":
-                return torch.cat(features.values(), dim=1)
-            elif self._config.data.aggregate_column_features == "mean":
-                return torch.mean(torch.stack(features.values()), dim=0)
+            if column_features_pooling_mode == "concat":
+                return np.concatenate(list(features.values()), axis=1)
+            elif column_features_pooling_mode == "mean":
+                return np.mean(list(features.values()), axis=0)
```
tonyhoo, line 402: Wondering how features with different dimensions can be supported here (e.g. image feature versus text feature from different backbones)

zhiqiangdon: Currently, we only support single modality in few shot learning. The feature dimension should be consistent for different columns. The default is `concat` which is more flexible.

tonyhoo (approved):

PR #3685, opened 2023-11-10
[AutoMM] Improve learner modularization

zhiqiangdon:
*Issue #, if available:*

*Description of changes:*

- Removed the util function calls, including `compute_num_gpus`, `infer_precision`, and `compute_inference_batch_size`, from `fit_per_run` and `predict_per_run`. Now all function calls use the class methods, which makes it easier for other learners to inherit from base learner.
- Avoided setting the `_output_shape` attribute in ner learner's `fit_per_run` to respect the design that learner attributes should not be updated inside `fit_per_run`.
- Organized the logic for getting `num_gpus` and `strategy` into one class method, which can be use in both fit and predict, as well as different learners.


FANGAreNotGnu (approved):
LGTM

PR #3689, opened 2023-11-10
Update black to v23.

jaheba:
Black updates its formatting each year. The version for 2023 is `v23`.

In addition, this PR adds `pyproject.toml` files for subfolders, which use other line length settings than the root folder. This ensure we can just run `black <folder>` on each folder with the desired line length setting.

### DIFF
--- a/timeseries/setup.py
+++ b/timeseries/setup.py
@@ -49,7 +49,7 @@ extras_require = {
         "flaky>=3.7,<4",
         "pytest-timeout>=2.1,<3",
         "isort>=5.10",
-        "black>=22.3,<23.0",
+        "black~=23.0",
     ],
 }
 
--- a/.github/workflow_scripts/env_setup.sh
+++ b/.github/workflow_scripts/env_setup.sh
@@ -2,7 +2,7 @@ function setup_build_env {
     python3 -m pip install --upgrade pip
     python3 -m pip install tox
     python3 -m pip install flake8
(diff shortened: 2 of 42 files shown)


timeseries/setup.py
```diff
@@ -49,7 +49,7 @@
         "flaky>=3.7,<4",
         "pytest-timeout>=2.1,<3",
         "isort>=5.10",
-        "black>=22.3,<23.0",
+        "black~=23.0",
```
shchur, line 52: I think that `setup.py` for multimodal should be updated as well https://github.com/autogluon/autogluon/blob/master/multimodal/setup.py#L61 .

Maybe there is a way to keep all test dependency versions in one place, similar to https://github.com/autogluon/autogluon/blob/master/core/src/autogluon/core/_setup_utils.py#L18C1-L18C19? Alternatively, maybe we could just put the union of all dependencies necessary for tests in one place and then import/reuse them in each module's `setup.py`?

shchur: There seems to already be a TODO for that [here](https://github.com/autogluon/autogluon/blob/master/core/src/autogluon/core/_setup_utils.py#L39C1-L39C14).

jaheba: I've updated the missing requirements, I think unifying is a good idea, but should we do it as part of a different PR?

shchur (approved):
Thanks for the PR! LGTM

PR #3692, opened 2023-11-10
Fix memory size estimate after AutoMM refactor

Innixma:
*Issue #, if available:*

*Description of changes:*

- Fix memory size estimate after AutoMM refactor
- Mainline crashes if training an AutoMM model via tabular and then calling `predictor.persist_models()` because `self.model._model` does not exist anymore.


taoyang1122 (approved):
LGTM.

PR #3701, opened 2023-11-13
Tabular: v1.0 API Deprecations

Innixma:
*Issue #, if available:*

*Description of changes:*

Implements planned API deprecations and changes.

Breaking Changes:
- Remove `autogluon.core.spaces`, which was previously deprecated and replaced by `autogluon.common.spaces`.

Backwards Compatible Changes to TabularPredictor (Keeps an alias method with the old name and include a deprecation warning):
- Switch `leaderboard(silent=False)` to `leaderboard(display=False)` -> Now leaderboard is silent by default. `silent` will still work if passed, so this is not a breaking change.
- `get_model_names_persisted()` -> Move logic to `model_names(persisted=True)`
- `compile_models` -> `compile`
- `get_model_failures` -> `model_failures`
- `persist_models` -> `persist`
- `unpersist_models` -> `unpersist`
- `get_model_best` -> `model_best`
- `get_pred_from_proba` -> `predict_from_proba`
- `get_oof_pred_proba` -> `predict_proba_oof`
- `get_oof_pred` -> `predict_oof`
- `get_model_full_dict` -> `model_refit_map`
- `get_size_disk` -> `size`
- `get_size_disk_per_file` -> `size_per_file`

Internal Changes:
- Removed `autogluon.core.task.base.base_task.py`, relocated relevant logic to appropriate locations.
- Removed `autogluon.core.utils.miscs.in_ipynb`, due to no usage in project.

TODO:
- [x] `evaluate` -> add `display` argument
- [x] `evaluate_predictions` -> add `display` argument


tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -4657,6 +4690,61 @@ def _assert_is_fit(self, message_suffix: str = None):
                 error_message = f"{error_message} `.{message_suffix}`."
             raise AssertionError(error_message)
 
+    @Deprecated(min_version_to_warn="0.8", min_version_to_error="1.1", version_to_remove="1.1", new="persist")
```
shchur, line 4693: What about the risk that erroring in v1.1 will break semantic versioning for AutoGluon?

For example, if a user upgrades to v1.0 and for some reason misses the DeprecationWarning (e.g., code is invoked outside of `if __name__ == "__main__":`, so deprecations are silenced [by default](https://peps.python.org/pep-0565/)), they could get upset when the code breaks after the v1.1 upgrade.

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -4657,6 +4690,61 @@ def _assert_is_fit(self, message_suffix: str = None):
...
+    def persist_models(self, **kwargs) -> List[str]:
+        """Deprecated method. Use `persist` instead."""
+        return self.persist(**kwargs)
+
+    @Deprecated(min_version_to_warn="0.8", min_version_to_error="1.1", version_to_remove="1.1", new="unpersist")
+    def unpersist_models(self, **kwargs) -> List[str]:
```
shchur, line 4699: Maybe change all deprecated method signatures to `def method(self, *args, **kwargs):` to avoid breaking calls that use positional arguments?

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -4657,6 +4690,61 @@ def _assert_is_fit(self, message_suffix: str = None):
...
+        """Deprecated method. Use `unpersist` instead."""
+        return self.unpersist(**kwargs)
+
+    @Deprecated(min_version_to_warn="0.8", min_version_to_error="1.1", version_to_remove="1.1", new="model_names")
+    def get_model_names(self, **kwargs) -> List[str]:
```
shchur, line 4704: We will also need to update https://github.com/autogluon/autogluon/blob/master/docs/api.rst to ensure that all deprecated methods don't show up in the side bar on https://auto.gluon.ai/stable/api/autogluon.tabular.TabularPredictor.html

shchur (commented):
Looks good overall, a few minor comments / questions.

core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
```diff
@@ -1186,7 +1186,7 @@ def reduce_memory_size(self, remove_fit_stack=False, remove_fit=True, remove_inf
                 if requires_save and self.low_memory:
                     self.save_child(model=model)
 
-    def _get_model_names(self):
+    def _model_names(self):
```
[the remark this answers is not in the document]
Innixma: Yes, since normal models don't have `_get_model_names`

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -3596,13 +3599,13 @@ def delete_models(self, models_to_keep=None, models_to_delete=None, allow_delete
...
 
-    def get_size_disk(self) -> int:
+    def size(self) -> int:
         """
         Returns the combined size of all files under the `predictor.path` directory in bytes.
         """
         return get_directory_size(self.path)
 
-    def get_size_disk_per_file(self, *, sort_by: str = "size", include_path_in_name: bool = False) -> pd.Series:
+    def size_per_file(self, *, sort_by: str = "size", include_path_in_name: bool = False) -> pd.Series:
```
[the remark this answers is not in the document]
Innixma: Updated to `disk_size_per_file`, good suggestion

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -3596,13 +3599,13 @@ def delete_models(self, models_to_keep=None, models_to_delete=None, allow_delete
             dry_run=dry_run,
         )
 
-    def get_size_disk(self) -> int:
+    def size(self) -> int:
```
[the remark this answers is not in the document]
Innixma: Changed to `disk_size`, good suggestion!

tabular/src/autogluon/tabular/predictor/predictor.py
Innixma: Added `*args`, good catch!

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -3019,9 +3022,9 @@ def refit_full(self, model="all", set_best_to_refit_full=True):
                 model_to_set_best = set_best_to_refit_full
             else:
                 model_to_set_best = model_best
-            model_full_dict = self._trainer.get_model_full_dict()
-            if model_to_set_best in model_full_dict:
-                self._trainer.model_best = model_full_dict[model_to_set_best]
+            model_refit_map = self._trainer.model_full_dict()
```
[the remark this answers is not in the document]
Innixma: I've now switched to using the consistent `model_refit_map` name everywhere. I originally didn't do this because TimeSeries also used `model_full_dict` and I was avoiding changing TimeSeries, but I have made the refactor for TimeSeries as well now.

tabular/src/autogluon/tabular/predictor/interpretable_predictor.py
```diff
@@ -55,15 +55,17 @@ def _validate_fit_extra_kwargs(self, kwargs, extra_valid_keys=None) -> dict:
             raise ValueError(f"{self.__class__.__name__} does not support `auto_stack`.")
         return kwargs
 
-    def leaderboard_interpretable(self, silent=False, **kwargs) -> pd.DataFrame:
+    def leaderboard_interpretable(self, verbose: bool = False, silent: bool = None, **kwargs) -> pd.DataFrame:
```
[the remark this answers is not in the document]
Innixma: Good point! Added.

core/src/autogluon/core/trainer/abstract_trainer.py
```diff
@@ -1358,7 +1358,7 @@ def refit_ensemble_full(self, model="all") -> dict:
             ensemble_set = self.get_minimum_model_set(model)
         existing_models = self.get_model_names()
         ensemble_set_valid = []
-        model_full_dict = self.get_model_full_dict()
+        model_full_dict = self.model_full_dict()
```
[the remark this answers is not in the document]
Innixma: Updated to be consistent

tabular/src/autogluon/tabular/predictor/predictor.py
Innixma: Do you have an alternative proposal? Does Pandas/NumPy do it differently?

tabular/src/autogluon/tabular/predictor/predictor.py
Innixma: @gidler Do you know how we can hide certain methods from appearing in the side bar?

gidler: Do you want the deprecated methods to not be documented in the methods table as well as the side bar, or is it just the side bar that we are concerned about?

I don't think this is as easy as one would wish it was. We are avoiding including `__init__` method documentation [here](https://github.com/autogluon/autogluon/blob/master/docs/_templates/autosummary/class.rst?plain=1#L16-L17) by filtering on the name. We could add the deprecated function names manually but that doesn't seem very maintainable and any function with that name across our whole project would get dropped.

Innixma: Ok, we can keep it for now, and if we can find a way to elegantly filter them, we can add it

tabular/src/autogluon/tabular/predictor/predictor.py
shchur: Pandas seems to always introduce deprecations in minor releases, and use major releases for breaking changes (and no new deprecations) https://pandas.pydata.org/docs/development/policies.html#version-policy. It seems that this wouldn't apply to us since our next release is a major version.

The way I see it, we can either 1/ break the API now or 2/ keep the old method names around until v2.0 to follow semantic versioning.

shchur (approved):
LGTM! Feel free to merge after the `verbose` -> `display` change & closing the remaining TODOs.

tabular/src/autogluon/tabular/predictor/predictor.py
Innixma: Discussed offline and decided on removing in v1.2

PR #3705, opened 2023-11-14
[timeseries] Enable Ray Tune for timeseries

canerturkmen:
*Issue #, if available:*

*Description of changes:*

We enable Ray Tune as the default HPO backend for GluonTS models except TFT. 

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
+++ b/timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
@@ -413,6 +413,8 @@ class AbstractTimeSeriesModel(AbstractModel):
         # logic here, and set to minimum num_gpus to 1 if it is set to 0 when GPUs are available
         kwargs["num_gpus"] = 0 if not self._is_gpu_available() else max(kwargs.get("num_gpus", 1), 1)
 
+        # we use k_fold=1 to circumvent autogluon.core logic to manage resources during parallelization
+        # of different folds
         hpo_executor.register_resources(self, k_fold=1, **kwargs)
         return self._hyperparameter_tune(hpo_executor=hpo_executor, **kwargs)
 
--- a/timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
+++ b/timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
@@ -386,17 +386,35 @@ class AbstractTimeSeriesModel(AbstractModel):
         """
         return train_fn_kwargs
 
+    def _is_gpu_available(self) -> bool:
+        return False
+
     def hyperparameter_tune(
         self, hyperparameter_tune_kwargs="auto", hpo_executor: HpoExecutor = None, time_limit: float = None, **kwargs
     ):
-        k_fold = kwargs.pop("k_fold", 1)
-        return super().hyperparameter_tune(
-            hyperparameter_tune_kwargs=hyperparameter_tune_kwargs,
-            hpo_executor=hpo_executor,
-            time_limit=time_limit,
-            k_fold=k_fold,
-            **kwargs,
+        if hpo_executor is None:
+            hpo_executor = self._get_default_hpo_executor()
+            default_num_trials = kwargs.pop("default_num_trials", None)
+        hpo_executor.initialize(
+            hyperparameter_tune_kwargs, default_num_trials=default_num_trials, time_limit=time_limit
         )
+        kwargs = self.initialize(time_limit=time_limit, **kwargs)
+
+        self._register_fit_metadata(**kwargs)
+        self._validate_fit_memory_usage(**kwargs)
+
+        kwargs = self._preprocess_fit_resources(
+            parallel_hpo=hpo_executor.executor_type == "ray", silent=True, **kwargs
+        )
+        self.validate_fit_resources(**kwargs)
+
+        # autogluon.core runs a complicated logic to determine the final number of gpus
+        # used in trials, which results in unintended setting of num_gpus=0. We override this
+        # logic here, and set to minimum num_gpus to 1 if it is set to 0 when GPUs are available
+        kwargs["num_gpus"] = 0 if not self._is_gpu_available() else max(kwargs.get("num_gpus", 1), 1)
+
+        hpo_executor.register_resources(self, k_fold=1, **kwargs)
+        return self._hyperparameter_tune(hpo_executor=hpo_executor, **kwargs)
 
     def _hyperparameter_tune(
         self,
@@ -444,14 +462,15 @@ class AbstractTimeSeriesModel(AbstractModel):
         if self.estimate_memory_usage is not None:
             model_estimate_memory_usage = self.estimate_memory_usage(**kwargs)
 
+        minimum_resources = self.get_minimum_resources(is_gpu_available=self._is_gpu_available())
         hpo_context = disable_stdout if isinstance(hpo_executor, RayHpoExecutor) else nullcontext
         with hpo_context(), warning_filter():  # prevent Ray from outputting its results to stdout with print
             hpo_executor.execute(
                 model_trial=model_trial,
                 train_fn_kwargs=train_fn_kwargs,
                 directory=directory,
-                minimum_cpu_per_trial=self.get_minimum_resources().get("num_cpus", 1),
-                minimum_gpu_per_trial=self.get_minimum_resources().get("num_gpus", 0),
+                minimum_cpu_per_trial=minimum_resources.get("num_cpus", 1),
+                minimum_gpu_per_trial=minimum_resources.get("num_gpus", 0),
                 model_estimate_memory_usage=model_estimate_memory_usage,
                 adapter_type="timeseries",
             )
--- a/timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
+++ b/timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
@@ -395,9 +395,10 @@ class AbstractTimeSeriesModel(AbstractModel):
         if hpo_executor is None:
             hpo_executor = self._get_default_hpo_executor()
             default_num_trials = kwargs.pop("default_num_trials", None)
-        hpo_executor.initialize(
-            hyperparameter_tune_kwargs, default_num_trials=default_num_trials, time_limit=time_limit
-        )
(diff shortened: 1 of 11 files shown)


timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
```diff
@@ -384,6 +386,38 @@ def _get_hpo_train_fn_kwargs(self, **train_fn_kwargs) -> dict:
...
+    def _is_gpu_available(self) -> bool:
+        return False
+
+    def hyperparameter_tune(
+        self, hyperparameter_tune_kwargs="auto", hpo_executor: HpoExecutor = None, time_limit: float = None, **kwargs
+    ):
+        if hpo_executor is None:
+            hpo_executor = self._get_default_hpo_executor()
+            default_num_trials = kwargs.pop("default_num_trials", None)
+        hpo_executor.initialize(
```
shchur, line 398: In `core.models.abstract.AbstractModel.hyperparameter_tune` this line is inside the `if` condition, why do we need to change it here?

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
```diff
@@ -429,21 +463,27 @@ def _hyperparameter_tune(
...
-            adapter_type="timeseries",
-        )
 
-        return hpo_executor.get_hpo_results(
-            model_name=self.name,
-            model_path_root=self.path_root,
-            time_start=time_start,
-        )
+        minimum_resources = self.get_minimum_resources(is_gpu_available=self._is_gpu_available())
+        hpo_context = disable_stdout if isinstance(hpo_executor, RayHpoExecutor) else nullcontext
```
shchur, line 468: Nice!

timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
```diff
@@ -302,6 +304,18 @@ def _get_estimator(self) -> GluonTSEstimator:
             **init_args,
         )
 
+    def _is_gpu_available(self) -> bool:
+        import torch.cuda
+
+        return torch.cuda.is_available()
+
+    def get_minimum_resources(self, is_gpu_available: bool = False) -> Dict[str, Union[int, float]]:
```
shchur, line 312: Is my understanding correct?
- `get_minimum_resources` method is necessary to tell Ray "please always assign 1 GPU to this job if there is GPU available"
- all GluonTS models will always try to use the GPU if GPU is available, regardless of `num_gpus` passed to `model.fit` or anything else

timeseries/src/autogluon/timeseries/models/multi_window/multi_window_model.py
```diff
@@ -58,6 +58,23 @@ def __init__(
...
+    def _get_model_base(self):
+        if self.most_recent_model is not None:
+            return self.most_recent_model
+        elif self.model_base is not None:
+            return self.model_base
+        else:
+            raise ValueError("Base model cannot be retrieved from MultiWindowModel")
+
+    def _get_hpo_backend(self) -> str:
+        return self._get_model_base()._get_hpo_backend()
```
shchur, line 70: Why not just use `self.model_base._get_hpo_backend()`? The constructor of `MultiWindowBacktestingModel` should ensure that `self.model_base` is an object of type `AbstractTimeSeriesModel`. If  `self.model_base is None`, we will have even bigger problems (`get_child_model()` below will fail, so fit won't work).

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -512,28 +512,36 @@ def fit(
...
-
-            Setting this parameter to string ``"random"`` performs 10 trials of random search per model.
-
-            We can change the number of random search trials per model by passing a dictionary as `hyperparameter_tune_kwargs`.
-            The dict must include the following keys
-
-            - ``"num_trials"``: int, number of configurations to train for each tuned model
-            - ``"searcher"``: currently, the only supported option is ``"random"`` (random search)
-            - ``"scheduler"``: currently, the only supported option is ``"local"`` (all models trained on the same machine)
+        hyperparameter_tune_kwargs : str or dict, default = None
```
shchur, line 515: Nit: We usually use `str or dict, optional` to annotate the type `Optional[Union[str, dict]]`

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -512,28 +512,36 @@ def fit(
...
+            Hyperparameter tuning strategy and kwargs (for example, how many HPO trials to run).
+            If None, then hyperparameter tuning will not be performed.
+
+            If type is ``str``, then this argument specifies a preset.
+                Valid preset values:
+                    'auto': Performs HPO via bayesian optimization search on GluonTS-backed neural forecasting models and
```
shchur, line 521: Nit: The majority of the docstrings in this file user `"` instead of `'` to denote strings

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -512,28 +512,36 @@ def fit(
...
-            The dict must include the following keys
-
-            - ``"num_trials"``: int, number of configurations to train for each tuned model
-            - ``"searcher"``: currently, the only supported option is ``"random"`` (random search)
-            - ``"scheduler"``: currently, the only supported option is ``"local"`` (all models trained on the same machine)
+        hyperparameter_tune_kwargs : str or dict, default = None
+            Hyperparameter tuning strategy and kwargs (for example, how many HPO trials to run).
+            If None, then hyperparameter tuning will not be performed.
+
+            If type is ``str``, then this argument specifies a preset.
```
shchur, line 519: Will the nested paragraph structure be preserved once the docs are rendered? We might need to use bulleted lists to ensure that the text is displayed in a readable manner in the docs.

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -512,28 +512,36 @@ def fit(
...
+                Valid preset values:
+                    'auto': Performs HPO via bayesian optimization search on GluonTS-backed neural forecasting models and
+                        random search on other models using local scheduler.
+                    'random': Performs HPO via random search.
+            You can also provide a dict to specify searchers and schedulers
+                Valid keys:
+                    'num_trials': How many HPO trials to run
```
shchur, line 526: Same as above, maybe use bulleted list for nicer rendering?

timeseries/tests/unittests/models/test_models.py
```diff
@@ -412,6 +413,12 @@ def test_when_custom_metric_passed_to_model_then_model_can_hyperparameter_tune(m
         },
         eval_metric=CustomMetric(),
     )
+    backend = model._get_hpo_backend()
+    if backend is RAY_BACKEND:
+        # Ray has trouble keeping references to the custom metric in the test namespace. We therefore
+        # skip this test.
+        pytest.skip()
```
shchur, line 420: Can we do something like [this](https://auto.gluon.ai/stable/tutorials/tabular/advanced/tabular-custom-metric.html#ensuring-metric-is-serializable) to ensure that Ray works with our implementation of custom metrics?

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
canerturkmen: Thanks for catching this!

timeseries/src/autogluon/timeseries/models/multi_window/multi_window_model.py
canerturkmen: `._get_model_base()` is an `AbstractModel` method that ray also relies on (assuming that when the base model is not equal to the main model, then you're using bagging) and `AbstractModel._get_default_hpo_executor()` expects to be working properly. I can have `_get_model_base()` just return `self.model_base` but I would rather keep the idiom of `_get_model_base` for consistency. wdyt?

timeseries/src/autogluon/timeseries/predictor.py
canerturkmen: sure. but, how do we imply it's None by default?

timeseries/src/autogluon/timeseries/predictor.py
canerturkmen: This mirrored Tabular. I updated this and will check the rendering.

timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
canerturkmen: That matches my understanding, yes.

timeseries/src/autogluon/timeseries/predictor.py
shchur: If the default value was not None, we would write 
```
hyperparameter_tune_kwargs : str or dict, default = "my_default_value"
```
while 
```
hyperparameter_tune_kwargs : str or dict, optional
```
implies that the default value is None. At least that's how it's used in the rest of TimeSeriesPredictor / TimeSeriesDataFrame documentation.

This is also a convention that Numpy uses - see, e.g., docs for `dtype` and `like` arguments [here](https://numpy.org/doc/stable/reference/generated/numpy.ones.html)

timeseries/src/autogluon/timeseries/models/multi_window/multi_window_model.py
shchur: Always returning `self.model_base` from `_get_model_base()` sounds good to me.

timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
shchur: Thanks!

timeseries/tests/unittests/models/test_models.py
canerturkmen: Hmm :/ I don't think so. I don't think the problem is serializability, but rather the fact that during tests Ray doesn't include timeseries.tests.unittests in the pythonpath and find the `CustomMetric` object to unpickle with. If I'm thinking about it, I don't think there's anything more serializable / unserializable about `CustomMetric` than our own metric implementations. I haven't had any trouble so far on this end.

shchur (approved):
LGTM, thank you! Feel free to merge after the platform tests pass

PR #3706, opened 2023-11-14
[timeseries] add intermittent demand models from Statsforecast

canerturkmen:
*Issue #, if available:*

*Description of changes:*

- Add intermittent demand forecasting models from Statsforecast. 
- Add the "Zero forecaster"
- Add AutoCES
- Add "naive pooled" conformalization for models which Statsforecast does not provide probabilistic forecasts. We prefer naive pooled over unpooled since in practice calibration sets are small. i.e., strictly speaking, with 5 calibration samples per time step in the horizon, conformal quantiles for many alpha values would simply default to (-inf, +inf). Moreover, since we apply conformalization often to models targeting sparse datasets, the sample quantiles of nonconformity scores (absolute residuals) have high variance and can result in collapsing critical nonconformity scores (hence, the intervals) to zero.


timeseries/src/autogluon/timeseries/models/__init__.py
```diff
@@ -8,12 +8,19 @@
     WaveNetModel,
 )
 from .local import (
+    ADIDAModel,
     ARIMAModel,
     AutoARIMAModel,
+    AutoCESModel,
     AutoETSModel,
     AverageModel,
+    ConformalizedZeroModel,
```
shchur, line 17: Should it also be listed under `__all__` below?

timeseries/src/autogluon/timeseries/models/local/statsforecast.py
```diff
@@ -21,14 +22,35 @@ def _update_local_model_args(self, local_model_args: Dict[str, Any]) -> Dict[str
...
+    def _get_point_forecast(
+        self,
+        time_series: pd.Series,
+        local_model_args: Dict,
+    ):
+        return self._get_local_model(local_model_args).forecast(
+            h=self.prediction_length, y=time_series.values.ravel()
+        )["mean"]
+
     def _predict_with_local_model(
```
shchur, line 38: Why not make this method abstract (raise `NotImplementedError`) for the abstract class, and then implement it for each of `AbstractProbabilisticStatsForecastModel` and `AbstractConformalizedStatsForecastModel`?

timeseries/src/autogluon/timeseries/models/local/statsforecast.py
```diff
@@ -138,10 +162,10 @@ def _get_model_type(self):
         return AutoARIMA
 
 
-class ARIMAModel(AbstractStatsForecastModel):
+class ARIMAModel(AbstractProbabilisticStatsForecastModel):
     """Autoregressive Integrated Moving Average (ARIMA) model with fixed parameters.
 
-    Based on `statsforecast.models.ARIMA <https://nixtla.github.io/statsforecast/models.html#arima>`_.
+    Based on `statsforecast.models.ARIMA <https://nixtla.mintlify.app/statsforecast/src/core/models.html#arima>`_.
```
shchur, line 168: Thanks for the fix!

timeseries/src/autogluon/timeseries/models/local/statsforecast.py
```diff
@@ -391,3 +415,343 @@ def _get_model_type(self):
...
+    ----------
+    .. [Svetunkov2022] Svetunkov, Ivan, Nikolaos Kourentzes, and John Keith Ord. "Complex exponential
+        smoothing." Naval Research Logistics (NRL) 69.8 (2022): 1108-1123.
+
+
+    Other Parameters
+    ----------------
+    model : str, default = "Z"
+        Defines type of CES model, "N" for simple CES, "S" for simple seasonality, "P" for partial seasonality
+        (without complex part), $F$ for full seasonality. When "Z" is selected, the best model is selected using
```
shchur, line 511:
```suggestion
        (without complex part), "F" for full seasonality. When "Z" is selected, the best model is selected using
```

timeseries/src/autogluon/timeseries/models/local/statsforecast.py
```diff
@@ -391,3 +415,343 @@ def _get_model_type(self):
...
+
+        point_forecast = self._get_point_forecast(time_series, local_model_args)
+        predictions = {
+            "mean": point_forecast,
+            **{str(q): point_forecast + q_sign[i] * ehat[i] for i, q in enumerate(levels)},
+        }
+        return pd.DataFrame(predictions)
+
+
+class AutoCESModel(AbstractConformalizedStatsForecastModel):
```
shchur, line 494: According to https://nixtla.mintlify.app/statsforecast/src/core/models_intro#automatic-forecasting, `AutoCESModel` is able to produce probabilistic forecasts. Should this model inherit from `AbstractProbabilisticStatsForecastModel` instead?

timeseries/src/autogluon/timeseries/models/local/statsforecast.py
```diff
@@ -391,3 +415,343 @@ def _get_model_type(self):
...
+            # AutoCES does not handle "tiny" datasets, fall back to naive
+            return np.full(self.prediction_length, time_series.values[-1])
+        if len(time_series) < 2 * local_model_args["season_length"] + 1 or local_model_args["season_length"] == 1:
+            local_model_args["model"] = "N"
+        return super()._get_point_forecast(time_series, local_model_args)
+
+
+class AbstractStatsForecastIntermittentDemandModel(AbstractConformalizedStatsForecastModel):
+    allowed_local_model_args = []
+    max_num_conformalization_windows = 5
```
shchur, line 557: Nit: Both attributes are already defined in the superclass?

timeseries/src/autogluon/timeseries/models/local/statsforecast.py
shchur: This change would also allow us to remove the `AbstractStatsForecastIntermittentDemandModel` class and make all intermittent models directly inherit from `AbstractConformalizedStatsForecastModel`.

timeseries/src/autogluon/timeseries/models/local/statsforecast.py
```diff
@@ -391,3 +415,343 @@ def _get_model_type(self):
...
+    max_ts_length : int, default = 2500
+        If not None, only the last ``max_ts_length`` time steps of each time series will be used to train the model.
+        This significantly speeds up fitting and usually leads to no change in accuracy.
+    """
+
+    def _get_model_type(self):
+        raise NotImplementedError
+
+    def _get_local_model(self, local_model_args: Dict):
+        raise NotImplementedError
```
shchur, line 750: Why not just remove these methods?

timeseries/src/autogluon/timeseries/models/local/statsforecast.py
```diff
@@ -391,3 +415,343 @@ def _get_model_type(self):
...
+        This significantly speeds up fitting and usually leads to no change in accuracy.
+    """
+
+    def _get_model_type(self):
+        from statsforecast.models import IMAPA
+
+        return IMAPA
+
+
+class ConformalizedZeroModel(AbstractStatsForecastIntermittentDemandModel):
```
shchur, line 730: Maybe change the name to `ZeroModel`? We don't use the `Conformalized*` prefix for other models that are also based on conformalization, so maybe drop it here as well?

timeseries/tests/unittests/models/test_local.py
```diff
@@ -266,3 +287,136 @@ def test_when_npts_fit_with_default_seasonal_features_then_predictions_match_glu
     assert (pred_gts.mean == pred_ag["mean"]).all()
     for q in npts_ag.quantile_levels:
         assert (pred_gts.quantile(str(q)) == pred_ag[str(q)]).all()
+
+
+class MockConformalModel(AbstractConformalizedStatsForecastModel):
```
shchur, line 292: Nice!

timeseries/src/autogluon/timeseries/models/__init__.py
```diff
@@ -22,15 +29,22 @@
 )
 
 __all__ = [
```
shchur, line 31: Can you please also add the new models to the following files?
- `docs/tutorials/timeseries/forecasting-model-zoo.md` - probably ZeroModel under "Baselines" & add a new section for sparse models
- add at least some models to `tests/smoketests/test_all_models.py`
- assign aliases & priorities in `src/autogluon/timeseries/models/presets.py`

timeseries/src/autogluon/timeseries/models/local/statsforecast.py
```diff
@@ -21,14 +22,35 @@ def _update_local_model_args(self, local_model_args: Dict[str, Any]) -> Dict[str
...
 
+    def _get_local_model(self, local_model_args: Dict):
+        model_type = self._get_model_type()
+        return model_type(**local_model_args)
+
+    def _get_point_forecast(
+        self,
+        time_series: pd.Series,
+        local_model_args: Dict,
+    ):
```
shchur, line 33: Nit:
```suggestion
    ) -> np.ndarray:
```

timeseries/src/autogluon/timeseries/models/local/statsforecast.py
```diff
@@ -391,3 +415,343 @@ def _get_model_type(self):
...
+
+    References
+    ----------
+    .. [Svetunkov2022] Svetunkov, Ivan, Nikolaos Kourentzes, and John Keith Ord. "Complex exponential
+        smoothing." Naval Research Logistics (NRL) 69.8 (2022): 1108-1123.
+
+
+    Other Parameters
+    ----------------
+    model : str, default = "Z"
```
shchur, line 509: Nit:
```suggestion
    model : {"Z", "N", "S", "P", "F"}, default = "Z"
```

timeseries/src/autogluon/timeseries/models/local/statsforecast.py
```diff
@@ -391,3 +415,343 @@ def _get_model_type(self):
...
+        local_model_args = super()._update_local_model_args(local_model_args)
+        local_model_args.setdefault("model", "Z")
+        return local_model_args
+
+    def _get_point_forecast(self, time_series: pd.Series, local_model_args: Dict):
+        # Disable seasonality if time series too short for chosen season_length or season_length == 1,
+        # otherwise model will crash
+        if len(time_series) < 5:
+            # AutoCES does not handle "tiny" datasets, fall back to naive
+            return np.full(self.prediction_length, time_series.values[-1])
```
shchur, line 549: Related to above comments: If we change `AutoCES` to inherit from `AbstractProbabilisticStatsForecastModel`, we can use the `seasonal_naive_forecast` method from `autogluon.timeseries.models.local.abstract_local_model` here.

timeseries/src/autogluon/timeseries/models/local/statsforecast.py
canerturkmen: I believe this was only added with v1.5.0 to CES.

timeseries/src/autogluon/timeseries/models/local/statsforecast.py
canerturkmen: hmm :/ explicit is better than implicit? I'll remove `_get_local_model` and also comment what I'm trying to say in `_get_model_type`

timeseries/src/autogluon/timeseries/models/local/statsforecast.py
canerturkmen: I wanted to highlight that it's got a little more intellectual depth than just Zero, but fair enough :)

timeseries/src/autogluon/timeseries/models/local/statsforecast.py
shchur: Okay, then we can keep the structure as is.

Can you please add a TODO for this change for when we can finally upgrade to StatsForecast v1.5.0?

timeseries/src/autogluon/timeseries/models/__init__.py
canerturkmen: Which model are you referring to? Is this skipping one?

timeseries/src/autogluon/timeseries/models/local/statsforecast.py
canerturkmen: I agree.

timeseries/src/autogluon/timeseries/models/__init__.py
shchur: My bad, I didn't see the `ConformalizedZeroModel` in the list below, but it's actually there. Please disregard the comment.

shchur (approved):
LGTM!

PR #3714, opened 2023-11-15
[timeseries] API Deprecations for v1.0

shchur:
*Description of changes:*
- Mirror the deprecations in #3701. For `get_model_names` and `model_best` we only perform the deprecations at the level of Predictor and not Trainer since Trainer object already has an attribute `model_best`.
  - in `leaderboard`: `silent` -> `display`
  - `get_model_names` -> `model_names`
  - `get_model_best` -> `model_best`
- Fix warning in GluonTS models
- Set `random_seed=123` in `TimeSeriesPredictor.fit` to produce deterministic results, similar to `TabularPredictor`.


Innixma (approved):
LGTM!

PR #3725, opened 2023-11-16
[AutoMM] Fix onnx export for un-initialized predictor & example onnx_text.py

AnirudhDagar:
*Issue*
https://github.com/autogluon/autogluon/blob/master/examples/automm/production/onnx_text.py is currently broken and has not been updated to handle some api changes.

1. `pipeline` arg is no longer used.
2. `predictor.export_onnx` always returns the onnx path now.
3. `get_processed_batch_for_deployment` has been refactored

Besides a even bigger problem was that if the predictor is not initialized, _model is set to `None` [here](https://github.com/autogluon/autogluon/blob/53a653f6c1f6d4fc8c026d3c60224cab22bda30a/multimodal/src/autogluon/multimodal/learners/base.py#L278) and then when `export_onnx` is called directly, it throws the following error due to [this](https://github.com/autogluon/autogluon/blob/53a653f6c1f6d4fc8c026d3c60224cab22bda30a/multimodal/src/autogluon/multimodal/utils/export.py#L117):
```
NotImplementedError: export_onnx doesn't support model type <class 'NoneType'>
```

*Description of changes:*

1. Fix func `export_onnx` in `predictor.py` by making sure predictor is prediction ready.
2. Fix old api usage in `onnx_text.py` example

### DIFF
--- a/examples/automm/production/onnx_text.py
+++ b/examples/automm/production/onnx_text.py
@@ -32,11 +32,11 @@ def eval_cosine(predictor, df, onnx_session):
 if __name__ == "__main__":
     parser = argparse.ArgumentParser()
     parser.add_argument("--checkpoint_name", default="sentence-transformers/msmarco-MiniLM-L-12-v3", type=str)
-    parser.add_argument("--onnx_path", default=None, type=str)
+    parser.add_argument("--model_path", default=None, type=str)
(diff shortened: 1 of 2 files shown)


tonyhoo (approved):
Thanks for the fix

PR #3727, opened 2023-11-16
[timeseries] Bugfixes for MLForecast models

shchur:
*Description of changes:*
- Avoid high memory usage by shortening time series before calling `MLForecast.preprocess`
   - Increase default `num_items` to 20K because we don't suffer from OOM problems anymore with the above change
- Generate dummy forecast for time series that are shorter than `sum(differences)` (before we used to get NaN predictions for these series)
- Add dummy quantiles to predictions of `DirectTabular` model if point forecast metric is provided (before there was a bug, where mean forecast was just repeated for all quantiles)
- Expose `damped` parameter of `AutoETS` model from StatsForecast

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
+++ b/timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
@@ -76,7 +76,8 @@ class AbstractMLForecastModel(AbstractTimeSeriesModel):
         from mlforecast import MLForecast
         from mlforecast.target_transforms import BaseTargetTransform
 
-        self._required_ts_length: Optional[int] = None
+        self._sum_of_differences: int = 0  # number of time steps removed from each series by differencing
+        self._max_ts_length: Optional[int] = None
         self._target_lags: Optional[List[int]] = None
         self._date_features: Optional[List[str]] = None
         self._mlf: Optional[MLForecast] = None
@@ -88,7 +89,7 @@ class AbstractMLForecastModel(AbstractTimeSeriesModel):
 
     def _get_model_params(self) -> dict:
         model_params = super()._get_model_params().copy()
-        model_params.setdefault("max_num_items", 10_000)
+        model_params.setdefault("max_num_items", 20_000)
         model_params.setdefault("max_num_samples", 1_000_000)
         model_params.setdefault("tabular_hyperparameters", {"GBM": {}})
         model_params.setdefault("tabular_fit_kwargs", {})
@@ -127,7 +128,7 @@ class AbstractMLForecastModel(AbstractTimeSeriesModel):
 
         if len(differences) > 0:
             target_transforms.append(Differences(differences))
-            self._required_ts_length = sum(differences)
+            self._sum_of_differences = sum(differences)
 
         scaler_name = model_params.get("scaler")
         if scaler_name is None:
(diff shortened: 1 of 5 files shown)


timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
```diff
@@ -279,6 +295,52 @@ def _get_scale_per_item(self, item_ids: pd.Index) -> pd.Series:
...
+        Add quantile levels assuming that residuals follow normal distribution
+        """
+        from scipy.stats import norm
+
+        scale_per_item = self._get_scale_per_item(repeated_item_ids.unique())
+        num_items = int(len(predictions) / self.prediction_length)
+        sqrt_h = np.sqrt(np.arange(1, self.prediction_length + 1))
+        # Series where normal_scale_per_timestep.loc[item_id].loc[N] = sqrt(1 + N) for N in range(prediction_length)
+        normal_scale_per_timestep = pd.Series(np.tile(sqrt_h, num_items), index=repeated_item_ids)
+        std_per_timestep = self._avg_residuals_std * scale_per_item * normal_scale_per_timestep
```
canerturkmen, line 339: Just saw that the default value for `self._avg_residuals_std` is 1.0. Why? Looks to me like this variable is better off without a default?

canerturkmen: also, just to cross check, this value is computed on one-step residuals right?

canerturkmen (approved):
Left some comments, LGTM otherwise. Thanks!

timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
```diff
@@ -279,6 +295,52 @@ def _get_scale_per_item(self, item_ids: pd.Index) -> pd.Series:
...
+            data_long = data.query("item_id not in @short_series")
+            if known_covariates is not None:
+                known_covariates_long = known_covariates.query("item_id not in @short_series")
+            else:
+                known_covariates_long = None
+        else:
+            data_long = data
+            known_covariates_long = known_covariates
+            forecast_for_short_series = None
+        return data_long, known_covariates_long, forecast_for_short_series
```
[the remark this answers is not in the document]
shchur: I have renamed the method and added the following docstring:
```python
    def _remove_short_ts_and_generate_fallback_forecast(
        self,
        data: TimeSeriesDataFrame,
        known_covariates: Optional[TimeSeriesDataFrame] = None,
    ) -> Tuple[TimeSeriesDataFrame, Optional[TimeSeriesDataFrame], Optional[TimeSeriesDataFrame]]:
        """Remove series that are too short for chosen differencing from data and generate naive forecast for them.

        Returns
        -------
        data_long : TimeSeriesDataFrame
            Data containing only time series that are long enough for the model to predict.
        known_covariates_long : TimeSeriesDataFrame or None
            Future known covariates containing only time series that are long enough for the model to predict.
        forecast_for_short_series : TimeSeriesDataFrame or None
            Seasonal naive forecast for short series, if there are any in the dataset.
        """
```

timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
shchur: Good point, set it to None (we should receive an error if model is not fit).

Indeed, it's computed on one-step-ahead forecast residuals.

PR #3732, opened 2023-11-17
[AutoMM] Update model presets

taoyang1122:
*Issue #, if available:*

*Description of changes:*
- Updated tabular data model presets from numerical/categorical MLP to ft-transformer.

### DIFF
--- a/multimodal/src/autogluon/multimodal/presets.py
+++ b/multimodal/src/autogluon/multimodal/presets.py
@@ -69,8 +69,7 @@ def default(presets: str = DEFAULT):
     """
     hyperparameters = {
         "model.names": [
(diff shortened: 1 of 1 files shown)


multimodal/src/autogluon/multimodal/presets.py
```diff
@@ -69,8 +69,7 @@ def default(presets: str = DEFAULT):
     """
     hyperparameters = {
         "model.names": [
-            "categorical_mlp",
-            "numerical_mlp",
+            "ft_transformer",
```
zhiqiangdon, line 72: Can we do the same change for ner presets?

zhiqiangdon (approved):
LGTM!

PR #3733, opened 2023-11-17
Fix LightGBM callback crash with v4.0+

Innixma:
*Issue #, if available:*

*Description of changes:*
- Fix LightGBM callback crash with v4.0+
- In v4.0+, `show_stdv` was changed from having a default value to being required to specify by the user in `_format_eval_results`. Because we weren't specifying it before, it started causing crashes when we upgraded to LightGBM v4.0+.


prateekdesai04 (approved):
LGTM!

PR #3740, opened 2023-11-19
[timeseries] Use per-item residuals in MLForecast models

shchur:
*Description of changes:*
- Estimate the standard deviation of the residuals separately for each item in `DirectTabular` model (with point forecast metric) and `RecursiveTabular` model. Before, we used to estimate a single standard deviation across all items.
- Fix typo in docstring.

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
+++ b/timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
@@ -291,7 +291,6 @@ class AbstractMLForecastModel(AbstractTimeSeriesModel):
     def _compute_residuals_std(self, val_df: pd.DataFrame) -> float:
         residuals = val_df[MLF_TARGET] - self._mlf.models_["mean"].predict(val_df)
         return residuals.pow(2).groupby(val_df[MLF_ITEMID]).mean().pow(0.5)
-        # return np.sqrt(residuals.pow(2.0).mean())
 
     def _get_scale_per_item(self, item_ids: pd.Index) -> pd.Series:
         """Extract the '_scale' values from the scaler object, if available."""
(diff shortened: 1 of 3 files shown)


timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
```diff
@@ -280,13 +281,15 @@ def _fit(
         self._mlf.models = {"mean": estimator}
 
         with warning_filter():
-            self._mlf.fit_models(X=train_df.drop(MLF_TARGET, axis=1), y=train_df[MLF_TARGET])
+            self._mlf.fit_models(X=train_df.drop(columns=[MLF_TARGET, MLF_ITEMID]), y=train_df[MLF_TARGET])
 
-        self._avg_residuals_std = self._compute_residuals_std(val_df)
+        self._residuals_std_per_item = self._compute_residuals_std(val_df)
+        self._avg_residuals_std = np.mean(self._residuals_std_per_item)
```
canerturkmen, line 287: The previous version seems to me the correct one. Averaging $\sigma := \frac{1}{N} \sum \sigma_i$ is _not_ the standard deviation of a random (equivalently, mixture of) time series drawn from the sample. I believe we're looking for $\sigma := \sqrt{\frac{1}{N} \sum \sigma_i^2}$ instead? 

Sanity check:
```python
import numpy as np
import scipy.stats as sst

n, m = 10000, 100
samples = np.full((n, m), fill_value=0)
for i in range(n):
    coin = np.random.choice(2)
    samples[i] = np.random.randn(m) * (44 if coin else 3)
    
print(np.std(samples))  # 30.7
print(np.mean([44, 3]))  # 23.5
print(np.sqrt(np.square([44, 3]).mean()))  # 31.2

# check coverage
lb, ub = sst.norm(0, 31.2).ppf([0.05, 0.95])
print(np.mean(np.logical_and(lb < samples, samples < ub)))  # 0.88

lb, ub = sst.norm(0, 23.5).ppf([0.05, 0.95])
print(np.mean(np.logical_and(lb < samples, samples < ub)))  # 0.81
```

canerturkmen (approved):
LGTM. Feel free to merge after we give the comment a bit of thought.

timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
shchur: That's a good point (and also I didn't know that GitHub now renders math)!. I have updated the code to correctly compute the avg residuals std.

canerturkmen: yeah TIL :)

canerturkmen (approved):
LGTM. Feel free to merge after addressing unit tests.

PR #3741, opened 2023-11-19
[timeseries] Adjust how time limit is split across models

shchur:
*Description of changes:*
- Do not allocate more than 600 seconds for fitting ensemble
- Evenly split the time limit across all validation windows in MultiWindowBacktestingModel
- Fix a bug where incorrect model suffix was saved if no refit happened in the most recent window


canerturkmen (approved):
LGTM. Thanks!

PR #3746, opened 2023-11-20
[timeseries] Fix GluonTS models using high context_length

shchur:
*Description of changes:*
- Currently, if user provides a very large `prediction_length` (more than 1000), it will result in extremely high `context_length` used by GluonTS models. This may lead to OOM errors (e.g., for TemporalFusionTransformer model that scales quadratically with sequence length).
- We clip the `context_length` to be at most 512 to avoid these issues.


canerturkmen (approved):
LGTM!

PR #3748, opened 2023-11-20
Update setup.py to remove s3fs

Innixma:
*Issue #, if available:*

*Description of changes:*
- Update setup.py to remove s3fs
- Can add back later after we figure out how to avoid slowing down install


shchur (approved):
LGTM!

PR #3750, opened 2023-11-20
Add ensemble weight logging

Innixma:
*Issue #, if available:*

*Description of changes:*
- Add ensemble weight logging

### DIFF
--- a/core/src/autogluon/core/trainer/abstract_trainer.py
+++ b/core/src/autogluon/core/trainer/abstract_trainer.py
@@ -1779,6 +1779,11 @@ class AbstractTrainer:
                     return model_names_trained
                 if self._time_limit is not None and self._time_train_start is not None:
                     time_left_total = self._time_limit - (fit_start_time - self._time_train_start)
+                    # If only a very small amount of time remains, skip training
+                    min_time_required = min(self._time_limit * 0.01, 10)
+                    if (time_left_total < min_time_required) and (time_limit < min_time_required):
+                        logger.log(15, f"Skipping {model.name} due to lack of time remaining.")
+                        return model_names_trained
                 else:
                     time_left_total = time_limit
                 fit_log_message += f" Training model for up to {round(time_limit, 2)}s of the {round(time_left_total, 2)}s of remaining time."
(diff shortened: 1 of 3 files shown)


core/src/autogluon/core/trainer/abstract_trainer.py
```diff
@@ -1779,6 +1779,11 @@ def _train_and_save(
                     return model_names_trained
                 if self._time_limit is not None and self._time_train_start is not None:
                     time_left_total = self._time_limit - (fit_start_time - self._time_train_start)
+                    # If only a very small amount of time remains, skip training
+                    min_time_required = min(self._time_limit * 0.01, 10)
```
shchur, line 1783: Not sure about the intended logic, but should this be `max` instead?

core/src/autogluon/core/trainer/abstract_trainer.py
```diff
@@ -2002,6 +2007,19 @@ def _update_model_attr(self, model: str, **attributes):
...
+            model_weights = model._get_model_weights()
+            model_weights = {k: round(v, 3) for k, v in model_weights.items()}
+            msg_weights = ""
+            is_first = True
+            for key, value in sorted(model_weights.items(), key=lambda x: x[1], reverse=True):
+                if not is_first:
+                    msg_weights += ", "
+                msg_weights += f"'{key}': {value}"
+                is_first = False
+            logger.log(20, f"\tEnsemble Weights: {{{msg_weights}}}")
```
shchur, line 2022: Nit: Why do we need this logic for printing, would something like the following be enough?
```python
model_weights = model._get_model_weights()
model_weights = {k: round(v, 3) for k, v in sorted(model_weights.items(), key=lambda item: item[1], reverse=True)}
logger.log(20, f:\tEnsemble Weights: {str(model_weights)}")
```

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -221,16 +221,16 @@ class TabularPredictor:
...
-        verbosity=2,
-        log_to_file=False,
-        log_file_path="auto",
-        sample_weight=None,
-        weight_evaluation=False,
-        groups=None,
+        label: str,
+        problem_type: str = None,
+        eval_metric: str | Scorer = None,
+        path: str = None,
```
shchur, line 227: Should optional arguments be of type `str | None = None` instead?

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -3068,6 +3085,7 @@ def refit_full(self, model: str | List[str] = "all", set_best_to_refit_full: boo
         logger.log(20, f'Refit complete, total runtime = {round(te - ts, 2)}s ... Best model: "{self._trainer.model_best}"')
         return refit_full_dict
 
+    @property
```
shchur, line 3088: Maybe we can keep `model_best` as a method as well? Changing it to a property could be potentially confusing since `predictor.model_names` and `predictor.features` are methods, but `predictor.model_best` is now a property.

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -913,6 +913,7 @@ def info(self) -> Dict[str, Any]:
         """Returns a dictionary of objects each describing an attribute of the training process and trained models."""
         return self._learner.get_info(include_model_info=True)
 
+    @property
     def model_best(self) -> str:
```
shchur, line 917: Same as above comment in `tabular.predictor`

core/src/autogluon/core/trainer/abstract_trainer.py
Innixma: `min` on purpose, as if a user gives 10 seconds for runtime, we should try to train the models with 3 seconds left, but if given 10000 seconds, probably skip the model if 3 seconds are left.

shchur (approved):
LGTM!

PR #3754, opened 2023-11-22
[timeseries] Fix incorrect scale used by DirectTabular model

shchur:
*Description of changes:*
- Currently, `DirectTabular` model trained with point forecast metric applies scaling twice when computing quantile forecasts. This results in extremely wide prediction intervals.
image

- This PR fixes this bug and ensures that scaling is applied once for all models.


shchur:
Same image as above, but after this PR:
image

canerturkmen (approved):
Nice! thank you!

PR #3756, opened 2023-11-23
[timeseries] Speed up TimeSeriesDataFrame.convert_frequency using joblib

shchur:
*Description of changes:*
- `tsdf.convert_frequency` can be extremely slow for large datasets. In this PR, we parallelize resampling using joblib. 
- This results in a noticeable speedup. For example, for M4 Monthly dataset (48000 items):
  - Current sequential implementation based on `df.groupby(...).resample(...)` - `4m11s`.
  - This PR (32 cores, chunk_size=100) -`31s`.

Note that chunking critical for the speedup. If we don't use chunking, runtime for M4 Monthly goes up to `4m30s`, which is comparable to the sequential version.

I have considered [other methods](https://stackoverflow.com/questions/15799162/resampling-within-a-pandas-multiindex) for speeding up resampling, however:
 - `pd.Grouper` isn't applicable in case original data has irregular timestamps that are a subset of `date_range` (it's only appropriate for subsampling).
 - unstack/stack won't work in case items have different start-end dates.

The `convert_frequency` method is [extensively tested](https://github.com/autogluon/autogluon/blob/master/timeseries/tests/unittests/test_ts_dataset.py#L893-L981), so this change is low-risk addition for the v1.0 release.

### DIFF
--- a/timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
+++ b/timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
@@ -962,6 +962,8 @@ class TimeSeriesDataFrame(pd.DataFrame):
                 resampled_dfs.append(pd.concat({item_id: resampled_df}, names=[ITEMID]))
             return pd.concat(resampled_dfs)
 
+        # Resampling time for 1 item < overhead time for a single parallel job. Therefore, we group items into chunks
+        # so that the speedup from parallelization isn't dominated by the communication costs.
         chunks = split_into_chunks(pd.DataFrame(self).groupby(level=ITEMID, sort=False), chunk_size)
         resampled_chunks = Parallel(n_jobs=num_cpus)(delayed(resample_chunk)(chunk) for chunk in chunks)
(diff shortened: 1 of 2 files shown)


timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
```diff
@@ -939,9 +951,22 @@ def convert_frequency(
             else:
                 aggregation[col] = agg_categorical
 
-        resampled_df = TimeSeriesDataFrame(
-            self.groupby(level=ITEMID, sort=False).resample(freq, level=TIMESTAMP, **kwargs).agg(aggregation)
-        )
+        def split_into_chunks(iterable: Iterable, size: int) -> Iterable[Iterable]:
+            iterable = iter(iterable)
+            return iter(lambda: tuple(islice(iterable, size)), ())
```
canerturkmen, line 956: nice alien sorcery :)

canerturkmen (approved):
LGTM!

Issue #3545, opened 2023-09-25
[BUG] Quantile Mode seems to behave like classification in some bad ways


SamuelGabriel:
Hi There,

Nice to have seen some of you at Automl-conf and thanks for this package.

I run into the following errors when using quantile (@jwmueller). I give a little script you can copy into ipython to see it for yourself below.
It happens on any kind of dataset for me.

```
Fitting model: RandomForestMSE_BAG_L1 ... Training model for up to 21.86s of the 21.86s of remaining time.
	Fitting 5 child models (S1F1 - S1F5) | Fitting with ParallelLocalFoldFittingStrategy
	Warning: Exception caused RandomForestMSE_BAG_L1 to fail during training... Skipping this model.
		ray::_ray_fit() (pid=2904225, ip=10.5.166.216)
  File "/home/muellesa/miniconda3/envs/prior-fitting-24-autogluon/lib/python3.10/site-packages/autogluon/core/models/ensemble/fold_fitting_strategy.py", line 402, in _ray_fit
    fold_model.fit(X=X_fold, y=y_fold, X_val=X_val_fold, y_val=y_val_fold, time_limit=time_limit_fold, **resources, **kwargs_fold)
  File "/home/muellesa/miniconda3/envs/prior-fitting-24-autogluon/lib/python3.10/site-packages/autogluon/core/models/abstract/abstract_model.py", line 829, in fit
    out = self._fit(**kwargs)
  File "/home/muellesa/miniconda3/envs/prior-fitting-24-autogluon/lib/python3.10/site-packages/autogluon/tabular/models/rf/rf_model.py", line 195, in _fit
    model = model.fit(X, y, sample_weight=sample_weight)
  File "/home/muellesa/miniconda3/envs/prior-fitting-24-autogluon/lib/python3.10/site-packages/autogluon/tabular/models/rf/rf_quantile.py", line 513, in fit
    y_train_leaves = est.y_train_leaves_
AttributeError: 'DecisionTreeQuantileRegressor' object has no attribute 'y_train_leaves_'
Detailed Traceback:
Traceback (most recent call last):
  File "/home/muellesa/miniconda3/envs/prior-fitting-24-autogluon/lib/python3.10/site-packages/autogluon/core/trainer/abstract_trainer.py", line 1733, in _train_and_save
    model = self._train_single(X, y, model, X_val, y_val, total_resources=total_resources, **model_fit_kwargs)
[...]

PR #3761, opened 2023-11-26
WIP: [tabular] Fix quantile regression models

shchur:
*Issue #, if available:* #3545

*Description of changes:*
- Ensure that only models that support quantile regression are trained for `problem_type=QUANTILE`
- Fix `FutureWarning` for `max_features="auto"`

*To Do:*
- [ ] Fix quantile regression RF models failing with the following exception 
```
    y_train_leaves = est.y_train_leaves_
AttributeError: 'ExtraTreeQuantileRegressor' object has no attribute 'y_train_leaves_'
```


tabular/src/autogluon/tabular/models/rf/rf_quantile.py
```diff
@@ -334,15 +334,15 @@ def __init__(
...
+class ExtraTreeQuantileRegressor(BaseTreeQuantileRegressor, ExtraTreeRegressor):
     def __init__(
         self,
         criterion="squared_error",
         splitter="random",
         max_depth=None,
         min_samples_split=2,
         min_samples_leaf=1,
-        max_features="auto",
+        max_features=1.0,
```
Innixma, line 345: What was the future warning? Isn't this a change in the model that would change results?

shchur: The "auto" value has been deprecated and 1.0 is the new default, here is the relevant announcement
https://github.com/scikit-learn/scikit-learn/blob/f44291b775cb67998c0965a72a8eeebc0beed37b/doc/whats_new/v1.1.rst?plain=1#L1274

Innixma: Gotcha, makes sense!

Innixma (approved):
LGTM!

PR #3767, opened 2023-11-27
[AutoMM] Fix HPO tests

zhiqiangdon:
*Issue #, if available:*

*Description of changes:*
Removed the time_limit as the training is capped by max_epochs=1.

### DIFF
--- a/multimodal/tests/unittests/others_2/test_hpo.py
+++ b/multimodal/tests/unittests/others_2/test_hpo.py
@@ -47,7 +47,6 @@ def predictor_hpo(searcher, scheduler, presets=None):
     predictor_hpo = predictor.fit(
         train_data=dataset.train_df,
         hyperparameters=hyperparameters,
-        time_limit=30,
         save_path=save_path,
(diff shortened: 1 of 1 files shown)


multimodal/tests/unittests/others_2/test_hpo.py
```diff
@@ -47,7 +47,6 @@ def predictor_hpo(searcher, scheduler, presets=None):
     predictor_hpo = predictor.fit(
         train_data=dataset.train_df,
         hyperparameters=hyperparameters,
-        time_limit=30,
```
tonyhoo, line 50: Did the `time_limit` prevent us from persisting the `config.yaml` file? If that's the case, should we consider increasing it instead of removing it, as it may extend the duration of test runs?

zhiqiangdon: The actual time may depend on the machine status. Increasing the time may still have the issue. Since the training is capped by max_epochs=1, it shouldn't take too much more time.

tonyhoo (approved):
LGTM! Thanks for the fix

PR #3770, opened 2023-11-28
[1.0?] Fix AutoMM warnings

FANGAreNotGnu:
- Refactor warning filters and add one more filter for object detection
- Remove messaging in DDP subprocesses

### DIFF
--- a/multimodal/src/autogluon/multimodal/learners/matching.py
+++ b/multimodal/src/autogluon/multimodal/learners/matching.py
@@ -221,6 +221,8 @@ class MultiModalMatcher(BaseLearner):
         self._total_train_time = None
         self._best_score = None
 
+        self._log_filters = []
+
     @property
     def query(self):
         return self._query
--- a/multimodal/src/autogluon/multimodal/learners/matching.py
+++ b/multimodal/src/autogluon/multimodal/learners/matching.py
@@ -221,7 +221,10 @@ class MultiModalMatcher(BaseLearner):
         self._total_train_time = None
         self._best_score = None
 
-        self._log_filters = []
+        self._log_filters = [
+            ".*does not have many workers.* in the `DataLoader` init to improve performance.*",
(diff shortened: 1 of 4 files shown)


multimodal/src/autogluon/multimodal/learners/matching.py
```diff
@@ -221,6 +221,8 @@ def __init__(
         self._total_train_time = None
         self._best_score = None
 
+        self._log_filters = []
```
zhiqiangdon, line 224: Matcher shares the same `run_trainer` API with base learner. Using an empty log filter is overwriting the log filters in base learner.

FANGAreNotGnu: Fixed with correct filters. Matcher did not call `super.__init__`, so we'll need to initialize `self._log_filters`.

zhiqiangdon (approved):
LGTM!

zhiqiangdon (commented):
The logs in the built docs have issues.

zhiqiangdon (commented):
Looks like `from pytorch_lightning.utilities import rank_zero_only` makes our `apply_log_filter` can't work properly. See http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-3770/ef6637e/tutorials/multimodal/text_prediction/beginner_text.html#training 
So, I removed this import.

FANGAreNotGnu:
> Looks like `from pytorch_lightning.utilities import rank_zero_only` makes our `apply_log_filter` can't work properly. See http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-3770/ef6637e/tutorials/multimodal/text_prediction/beginner_text.html#training So, I removed this import.

I have no idea why `rank_zero_only` breaks our `apply_log_filter`. Let me check their source code. And we might need to remove ddp subprocess warning after 1.0 release.

PR #3775, opened 2023-11-28
Fix get_precision_context

ddelange:
*Issue #, if available:* N/A

*Description of changes:* Avoid crash when model `config.yaml` contains integer `precision: 16`

```pytb
File "/usr/share/python3/app/lib/python3.11/site-packages/autogluon/multimodal/predictor.py", line 644, in predict_proba
    return self._learner.predict_proba(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/share/python3/app/lib/python3.11/site-packages/autogluon/multimodal/learners/base.py", line 2000, in predict_proba
    outputs = self.predict_per_run(
              ^^^^^^^^^^^^^^^^^^^^^
  File "/usr/share/python3/app/lib/python3.11/site-packages/autogluon/multimodal/learners/base.py", line 1724, in predict_per_run
    outputs = self.realtime_predict(
              ^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/share/python3/app/lib/python3.11/site-packages/autogluon/multimodal/learners/base.py", line 1631, in realtime_predict
    output = self.predict_batch(
             ^^^^^^^^^^^^^^^^^^^
  File "/usr/share/python3/app/lib/python3.11/site-packages/autogluon/multimodal/utils/inference.py", line 180, in predict_batch
    precision_context = get_precision_context(precision=precision, device_type=device_type)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/share/python3/app/lib/python3.11/site-packages/autogluon/multimodal/utils/environment.py", line 245, in get_precision_context
    return torch.autocast(device_type=device_type, dtype=torch.bfloat16 if "bf16" in precision else torch.half)
                                                                           ^^^^^^^^^^^^^^^^^^^
TypeError: argument of type 'int' is not iterable

```

### DIFF
--- a/multimodal/src/autogluon/multimodal/utils/environment.py
+++ b/multimodal/src/autogluon/multimodal/utils/environment.py
@@ -68,6 +68,43 @@ def compute_num_gpus(config_num_gpus: Union[int, float, List], accelerator: str)
     return num_gpus
 
 
+def convert_to_torch_precision(precision: Union[int, str]):
+    """
+    Convert a precision integer or string to the corresponding torch precision.
+
+    Parameters
+    ----------
+    precision
+    a precision integer or string from the config.
+
+    Returns
+    -------
+    A torch precision object.
+    """
+    precision_mapping = {
+        16: torch.half,
+        "16": torch.half,
+        "16-mixed": torch.half,
+        "16-true": torch.half,
+        "bf16": torch.bfloat16,
+        "bf16-mixed": torch.bfloat16,
+        "bf16-true": torch.bfloat16,
+        32: torch.float32,
+        "32": torch.float32,
+        "32-true": torch.float32,
+        64: torch.float64,
+        "64": torch.float64,
+        "64-true": torch.float64,
+    }
+
+    if precision in precision_mapping:
+        precision = precision_mapping[precision]
+    else:
+        raise ValueError(f"Unknown precision: {precision}")
+
+    return precision
+
+
 def infer_precision(
     num_gpus: int, precision: Union[int, str], as_torch: Optional[bool] = False, cpu_only_warning: bool = True
 ):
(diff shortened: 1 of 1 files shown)


multimodal/src/autogluon/multimodal/utils/environment.py
```diff
@@ -238,15 +238,14 @@ def get_precision_context(precision: Union[int, str], device_type: Optional[str]
     -------
     A precision context manager.
     """
-    if precision == 32:
+    precision = infer_precision(num_gpus=1, precision=precision, as_torch=True)
```
ddelange, line 241: hard-coded num_gpus here instead of `num_gpus = int(device_type == 'gpu')`

because I'm not sure if `get_precision_context` is supposed to support 64bit floats on cpu, and `infer_precision` disallows it.

multimodal/src/autogluon/multimodal/utils/environment.py
```diff
@@ -238,12 +257,14 @@ def get_precision_context(precision: Union[int, str], device_type: Optional[str]
...
+    if precision == torch.float32:
         assert torch.get_default_dtype() == torch.float32
         return contextlib.nullcontext()
-    elif precision in [16, "16", "16-mixed", "bf16", "bf16-mixed"]:
-        return torch.autocast(device_type=device_type, dtype=torch.bfloat16 if "bf16" in precision else torch.half)
-    elif precision == 64:
+    elif precision == torch.float64:
         return double_precision_context()
     else:
         raise ValueError(f"Unknown precision: {precision}")
```
ddelange, line 270: this line will never be reached (will now be raised in `convert_to_torch_precision`) so probably no more need for elif/else

thanks for the quick action, LGTM!

zhiqiangdon: Thanks. Maybe we can use that line as a safe guard just in case that we add new precisions in precision_map, but torch.autocast can't handle it.

zhiqiangdon (approved):
LGTM! Thanks @ddelange for the fix!

PR #3778, opened 2023-11-28
Fix Platform tests

prateekdesai04:
*Description of changes:*
PR to fix platform tests
Link to test run: https://github.com/prateekdesai04/autogluon/actions/runs/7023137621

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #3782, opened 2023-11-29
[core] Add async_timeout to requirements

shchur:
*Description of changes:*
- Ray fails on Python 3.11 if GPU is present. #3778 fixes this problem in our test environment, but the error seems to still be present outside of our test environment (e.g., if we install AG with `pip install autogluon` in a fresh Python 3.11 env). This PR fixes the problem by adding `async-timeout` to the `core[ray]` requirements.


prateekdesai04 (approved):
LGTM!

ddelange:
ref https://github.com/ray-project/ray/issues/41267#issuecomment-1819554001

will be fixed upstream in ray 2.9

Issue #3814, opened 2023-12-14
[BUG] TimeSeriesPredictor's refit_fill method can't get parameter (model="all")


hykhhijk:
**Bug Report Checklist**
- [ ] I provided code that demonstrates a minimal reproducible example.  
- [ ] I confirmed bug exists on the latest mainline of AutoGluon via source install.  
- [x] I confirmed bug exists on the latest stable version of AutoGluon.  

**Describe the bug**
![image](https://github.com/autogluon/autogluon/assets/58303938/5d0b9859-fe7a-4ccf-9ee9-74f350885617)
![image](https://github.com/autogluon/autogluon/assets/58303938/0779c504-1167-496f-b69b-d4b6bac64d4e)
I have used previous version of refit_full method TimeSeriesPredictor but when I upgrade to ver1.0 TimeSeriesPredictor.refit_full() can't get (model="all") parameter and output error like this

**Expected behavior**
Want to use model="all" parameter again?

**Screenshots / Logs**
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Cell In[24], [line 1](vscode-notebook-cell:?execution_count=24&line=1)
----> [1](vscode-notebook-cell:?execution_count=24&line=1) predictor.refit_full()

File [~/anaconda3/envs/ag/lib/python3.10/site-packages/autogluon/timeseries/predictor.py:1080](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Dacon/Jeju/~/anaconda3/envs/ag/lib/python3.10/site-packages/autogluon/timeseries/predictor.py:1080), in TimeSeriesPredictor.refit_full(self, model, set_best_to_refit_full)
   [1074](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Dacon/Jeju/~/anaconda3/envs/ag/lib/python3.10/site-packages/autogluon/timeseries/predictor.py:1074) logger.info(
   [1075](https://vscode-remote+wsl-002bubuntu.vscode-resource.vscode-cdn.net/mnt/d/Dacon/Jeju/~/anaconda3/envs/ag/lib/python3.10/site-packages/autogluon/timeseries/predictor.py:1075)     "Refitting models via `refit_full` using all of the data (combined train and validation)...\n"
[...]

PR #3820, opened 2023-12-15
[timeseries] Fix refit_full bug

shchur:
*Issue #, if available:* #3814

*Description of changes:*
- Fix bugs when calling `predictor.refit_full()` multiple times
- Add test that verifies correctness of repeated calls to `refit_full()`
- Add `can_refit_full` tag to `DirectTabular` and `RecursiveTabular` models that was incorrectly removed in v1.0


canerturkmen (approved):
LGTM!

Issue #3497, opened 2023-08-29
[BUG] SHAP waterfall plot does not consistently show positive class


callinSwitzer:
**Bug Report Checklist**

- [ ] I provided code that demonstrates a minimal reproducible example.  
- [ ] I confirmed bug exists on the latest mainline of AutoGluon via source install.  
- [X] I confirmed bug exists on the latest stable version of AutoGluon.  

**Describe the bug**
When using ShapAnalysis (https://github.com/autogluon/autogluon/blob/master/eda/src/autogluon/eda/analysis/explain.py#L120), to create a waterfall plot, I see that sometimes the positive class is represented as increasing along the horizontal axis, and other times the negative class is increasing along the horizontal axis. Furthermore, I would expect features that increase the probability of being in the positive class to be red, and features that increase the probability of being in the negative class to be blue. 

**Expected behavior**
I would expect that moving right along the horizontal axis of the waterfall plot would mean increasing probability of being in the positive class. The [SHAP documentation](https://shap.readthedocs.io/en/latest/example_notebooks/api_examples/plots/waterfall.html#:~:text=The%20bottom%20of%20a%20waterfall%20plot%20starts%20as%20the%20expected%20value%20of%20the%20model%20output%2C%20and%20then%20each%20row%20shows%20how%20the%20positive%20(red)%20or%20negative%20(blue)%20contribution%20of%20each%20feature%20moves%20the%20value%20from%20the%20expected%20model%20output%20over%20the%20background%20dataset%20to%20the%20model%20output%20for%20this%20prediction.) shows red as a positive contribution.

**To Reproduce**
```
import autogluon.eda.auto as auto

# get predictor, training data, and select a row to explain
predictor = ...
train_data = ...
explain_row = ...

auto.explain_rows(
    train_data=train_data,
    model=predictor,
    display_rows=True,
    rows=explain_row,
    plot="waterfall",
)
```

**Screenshots / Logs**

![image (14)](https://github.com/autogluon/autogluon/assets/7850472/956c9a28-ca4c-4072-99a3-177e6844862b)

**Installed Versions**

```python
INSTALLED VERSIONS
------------------
date                   : 2023-08-29
time                   : 16:24:25.698539
python                 : 3.10.12.final.0
OS                     : Linux
OS-release             : 5.4.250-173.369.amzn2int.x86_64
Version                : #1 SMP Tue Aug 1 20:57:12 UTC 2023
machine                : x86_64
processor              : x86_64
[...]

PR #3839, opened 2023-12-24
[eda] added positive class parameters and autodetection

gradientsky:
*Issue #, if available:* #3497

*Description of changes:*
- Added positive class selection - see #3497
- binary classification will use `model.positive_class`
- multiclass uses the last column in prediction probabilities
```python
auto.explain_rows(
    train_data=df_train,
    model=state.model,
    display_rows=True,
    return_state=True,
    positive_class='banana', # specify positive class
    rows=state.model_evaluation.highest_error[:1]
)
```

- Libraries updates: bumped versions for shap and pyod; updated changed shap API calls.


tonyhoo (approved):
LGTM! Thanks for the fix

PR #3849, opened 2024-01-05
[timeseries] Return empty leaderboard if no models were trained during fit

shchur:
*Description of changes:*
- Currently, if all models failed during `fit()`, calling `predictor.leaderobard()` crashes with a `KeyError`. With this PR, we will return an empty leaderboard instead.


canerturkmen (approved):
LGTM!

PR #3866, opened 2024-01-19
[FIX] Avoid Ray-Deadlock due to Dynamic Stacking's Memory Save Sub-fit 

LennartPurucker:
This short PR is to fix a bug related to dynamic stacking's memory save subfit.

If one starts a ray cluster outside of AutoGluon, passing the resources to the remote function of the memory save fit can result in incorrect ray scheduling.  If one uses a GPU, this can make Ray think that fewer resources are available than there should be; resulting in a ray-related deadlock. 

As a temporary workaround, one can alternatively disable memory save fitting of dynamic stacking. 

---


LennartPurucker:
I am not too sure that this does not create other problems. 

While this fixes my current problem, I suspect it creates the same kind of deadlock if the ray has not been initialized in the outer scope. Or this might be just my setup. It's really hard to tell right now.

LennartPurucker:
Closing this for now as this seems to have created too many other problems in my own tests.

PR #3867, opened 2024-01-20
[AutoMM] add a new column type 'image_base64_str'

Harry-zzh:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/multimodal/src/autogluon/multimodal/utils/misc.py
+++ b/multimodal/src/autogluon/multimodal/utils/misc.py
@@ -68,6 +68,7 @@ def path_to_bytearray_expander(path, base_folder):
 def shopee_dataset(
     download_dir: str,
     is_bytearray=False,
+    is_base64str=False,
 ):
     """
     Download Shopee dataset for demo.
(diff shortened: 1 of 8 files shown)


multimodal/src/autogluon/multimodal/utils/misc.py
```diff
@@ -87,12 +88,16 @@ def shopee_dataset(
     load_zip.unzip(zip_file, unzip_dir=download_dir)
 
     dataset_path = os.path.join(download_dir, "shopee")
-    train_data = pd.read_csv(f"{dataset_path}/train.csv")
-    test_data = pd.read_csv(f"{dataset_path}/test.csv")
+    if is_base64str:
+        train_data = pd.read_csv(f"{dataset_path}/train_base64_str.csv")
+        test_data = pd.read_csv(f"{dataset_path}/test_base64_str.csv")
```
zhiqiangdon, line 93: If the shopee dataset was downloaded in a machine, `load_zip` function would not download it again, resulting in the error `FileNotFoundError: [Errno 2] No such file or directory: './shopee/train_base64_str.csv' `. Is it possible that we generate the base64 str format online like the way we build bytearray images?

zhiqiangdon (approved):
LGTM

PR #3868, opened 2024-01-20
[FIX] Imporve Error Handling and Init for Ray for Memory Save Sub-fits in Dynamic Stacking 

LennartPurucker:
While searching for another issue, I realized that the current implementation of memory-safe fits will not do memory-safe fits in the edge case where Ray was never initialized on the current system (e.g., in a fresh container).

Due to using `address="auto"`, this would fail and then be silently handled by except without informing the user. 
This PR is a short, straightforward fix for this and provides a warning now if save memory fits should have been used but were not. 

---

### DIFF
--- a/tabular/src/autogluon/tabular/predictor/predictor.py
+++ b/tabular/src/autogluon/tabular/predictor/predictor.py
@@ -1283,8 +1283,9 @@ class TabularPredictor(TabularPredictorDeprecatedMixin):
             try:
                 _ds_ray = try_import_ray()
                 if not _ds_ray.is_initialized():
-                    _ds_ray.init(address="auto", logging_level=logging.ERROR, log_to_driver=False)
-            except:
+                    _ds_ray.init(logging_level=logging.ERROR, log_to_driver=False)
+            except Exception as e:
+                warnings.warn(f"Failed to use ray for memory safe fits. Falling back to normal fit. Error: {repr(e)}", stacklevel=2)
                 _ds_ray = None
 
             if _ds_ray is not None:


LennartPurucker:
We should also add a check to disable memory save sub-fits if we use GPUs. 

Following a rather long debugging session on this, my current understanding is that Ray does not free GPU resources when using nested remote functions ([but only CPU resources](https://docs.ray.io/en/latest/ray-core/tasks/nested-tasks.html#yielding-resources-while-blocked)). 

### Details 

Assume you have 1 GPU. In the case of memory save sub-fits, we have first a remote function (A) that is the sub-fit and then a remote function (B) for fitting models in the case of a ParallelLocalFoldFittingStrategy. 

Now, (A) will allocate the GPU by design. Then, within (A), the code tries to call (B), and if the model requires the GPU, this will result in a deadlock as (A) does not yield its allocated GPU resources. 

Now, one could not pass GPU resources to (A), but then the sub-fit would crash if a model uses `SequentialLocalFoldFittingStrategy` and requires a GPU (as the sub-fit does not know a GPU exists). 


----

#### Somewhat Related Sidenote

While digging through the code to debug my Ray issues, I found that refitting has severe resource management issues. I have not yet fully understood the problem, but when you, for example, tell AutoGluon to use 4 CPUs and 4 GPUs, then AutoGluon will only use these resources with Ray during a regular call to fit.  However, during refitting (after fit or after cross-validation via `refit_folds`), Ray no longer respects these constraints and defaults to use all available CPUs (and no GPUs).

I hypothesize that Ray no longer respects these resources, as they are never passed to Ray during calls when refitting. This could be fixed by setting them in the `core_kwargs` in calls to fit_single methods during the refit. 

Sadly, I do not have a reproducible example for this because it heavily depends on my SLURM cluster setup.

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -1283,8 +1283,9 @@ def _sub_fit_memory_save_wrapper(
             try:
                 _ds_ray = try_import_ray()
                 if not _ds_ray.is_initialized():
-                    _ds_ray.init(address="auto", logging_level=logging.ERROR, log_to_driver=False)
-            except:
+                    _ds_ray.init(logging_level=logging.ERROR, log_to_driver=False)
```
Innixma, line 1286: Is there any downside to not specifying an `address` value vs setting to `"auto"`?

Innixma:
Thanks for the PR! If you can create separate issues for the 2 additional bugs you mention, that would help to be able to track them.

tabular/src/autogluon/tabular/predictor/predictor.py
LennartPurucker: In this setting, there is only an upside or, rather, avoid a failure case.

If `auto` is set, Ray assumes that there is an existing cluster and will throw an Error otherwise. While not providing the `address` results in creating a local Ray instance instead of throwing an error in that case. 

Generally, one should only use `auto` if we know that a cluster should exist or has existed before. In the edge cases described above, neither is the case.

Innixma (approved):

PR #3872, opened 2024-01-22
[Fix] Fix tests on scikit-learn upgrade

prateekdesai04:
*Description of changes:*
Upgrading to scikit-learn 1.4.0 has brought on some changes in the API. PR to address the same.

### DIFF
--- a/tabular/src/autogluon/tabular/models/rf/rf_quantile.py
+++ b/tabular/src/autogluon/tabular/models/rf/rf_quantile.py
@@ -510,7 +510,7 @@ class BaseForestQuantileRegressor(ForestRegressor):
                 bootstrap_indices = np.arange(len(y))
 
             est_weights = np.bincount(bootstrap_indices, minlength=len(y))
-            y_train_leaves = est.tree_.apply(X)
+            y_train_leaves = est.y_train_leaves_
             # Normalize the bootstrap weights such that the total weight of each leaf sums up to 1
             # Relabel leaves starting from zero in order to efficiently count the total sum per leaf with bincount
             leaves_starting_from_zero = np.unique(y_train_leaves, return_inverse=True)[1]
--- a/tabular/src/autogluon/tabular/models/rf/rf_quantile.py
+++ b/tabular/src/autogluon/tabular/models/rf/rf_quantile.py
@@ -510,7 +510,7 @@ class BaseForestQuantileRegressor(ForestRegressor):
                 bootstrap_indices = np.arange(len(y))
 
             est_weights = np.bincount(bootstrap_indices, minlength=len(y))
-            y_train_leaves = est.y_train_leaves_
+            y_train_leaves = est.tree_.apply(X)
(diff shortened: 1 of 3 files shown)


tabular/src/autogluon/tabular/models/rf/rf_quantile.py
```diff
@@ -510,7 +510,7 @@ def fit(self, X, y, sample_weight=None):
                 bootstrap_indices = np.arange(len(y))
 
             est_weights = np.bincount(bootstrap_indices, minlength=len(y))
-            y_train_leaves = est.y_train_leaves_
+            y_train_leaves = est.tree_.apply(X)
```
Innixma, line 513: This situation is not so simple. While this does indeed appear to fix the problem, the issue goes deeper and should be investigated.

Specifically, [according to multi-inheritance rules in Python](https://www.programiz.com/python-programming/multiple-inheritance), `BaseTreeQuantileRegressor.fit` should be called when `DecisionTreeQuantileRegressor(BaseTreeQuantileRegressor, DecisionTreeRegressor)` calls `super().fit`.

This is the case in scikit-learn 1.3.2, which is why `est.y_train_leaves` exists, because the final line in the `.fit` call sets it: `self.y_train_leaves_ = self.tree_.apply(X)`

However, in scikit-learn 1.4, this mysteriously changes, and `BaseTreeQuantileRegressor.fit` is never called. You can verify this by adding `raise AssertionError` to the start of `BaseTreeQuantileRegressor.fit`. With scikit-learn 1.3.2 it raises the AssertionError, in scikit-learn 1.4 it never errors, meaning the code is never entered, despite the intention of the code to be entered.

We can confirm this by removing entirely the class from the code:

```
class DecisionTreeQuantileRegressor(BaseTreeQuantileRegressor, DecisionTreeRegressor):

->

class DecisionTreeQuantileRegressor(DecisionTreeRegressor)
```

This edit works without error. (Note: It also works for scikit-learn 1.3.2)

We can go even further:

```
        super(RandomForestQuantileRegressor, self).__init__(
            DecisionTreeQuantileRegressor(),

->

        super(RandomForestQuantileRegressor, self).__init__(
            DecisionTreeRegressor(),
```

This also works with identical results, removing the entire Quantile regressor tree logic, although we may want to keep it in case we wanted to fit a tree rather than a forest.

I have been unable to determine why the `fit` is not called in `scikit-learn==1.4`. My understanding of Python leads me to believe the code should be entered, and yet it isn't. I also haven't figured out what fundamentally changed in `scikit-learn==1.4` to warrant this.

The current solution in this PR is merely a bandaid and likely isn't a proper solution, as it replicates what should have been set during `est.fit(...)`, but doing so later on.

prateekdesai04: Yes, I observed the same things while working on the fix.
So for now should we just cap the `scikit-learn` version to `1.3.2` to get the CI running, until we figure the cause for this?

Innixma (approved):
LGTM!

Issue #3766, opened 2023-11-27
TimeSeriesPredictor redirect verbose output to file


lucical:
### Discussed in https://github.com/autogluon/autogluon/discussions/3765

<div type='discussions-op-text'>

<sup>Originally posted by **lucical** November 27, 2023</sup>
Hi,

How can I save the verbose output from the terminal when I train the `TimeSeriesPredictor` to a file? I found [these answers](https://github.com/autogluon/autogluon/issues/2325) for `TabularPredictor` but none of it works for Time Series prediction and there is no support for logging implemented like there `predictor = TabularPredictor(..., log_to_file=True)`.

Thanks!

</div>

canerturkmen:
Hi @lucical, the issue should now be fixed and will be released with the next AG release. Thank you again for bringing it to our attention.

PR #3877, opened 2024-01-23
[timeseries] enable logging to file, clean up set_logger_verbosity calls

canerturkmen:
*Issue #, if available:*

#3766 

*Description of changes:*

- Enable logging to file, including a custom log file name following the API of `TabularPredictor`.
- Clean up unused verbosity arguments and set_logger_verbosity calls throughout the library. 

Note: Previously, timeseries model objects (`AbstractTimeSeriesModel`) had their own verbosity levels provided in the call to `fit`, which was propagated from the Trainer `fit` call to individual models. While this created clutter, it also afforded the flexibility of models potentially using the verbosity level to set dependency log output levels (e.g., GluonTS). We keep this as is: GluonTS models still receive `verbosity` as a fit argument.

However, logger verbosity is now only controlled by the Predictor, which sets the verbosity (either during init or fit) to the `autogluon.timeseries` parent logger, which is inherited by all `logging.getLogger(__name__)` calls in the timeseries module. 

### DIFF
--- a/timeseries/src/autogluon/timeseries/predictor.py
+++ b/timeseries/src/autogluon/timeseries/predictor.py
@@ -9,7 +9,7 @@ from typing import Any, Dict, List, Optional, Tuple, Type, Union
 import pandas as pd
 
 from autogluon.common.utils.deprecated_utils import Deprecated
-from autogluon.common.utils.log_utils import set_logger_verbosity
+from autogluon.common.utils.log_utils import add_log_to_file, set_logger_verbosity
 from autogluon.common.utils.system_info import get_ag_system_info
 from autogluon.common.utils.utils import check_saved_predictor_version, seed_everything, setup_outputdir
 from autogluon.core.utils.decorators import apply_presets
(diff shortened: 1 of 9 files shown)


timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -229,6 +239,14 @@ def __init__(
     def _trainer(self) -> AbstractTimeSeriesTrainer:
         return self._learner.load_trainer()  # noqa
 
+    def _setup_log_to_file(self, log_to_file, log_file_path):
```
shchur, line 242: Maybe add type hints?

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -149,6 +156,8 @@ def __init__(
         eval_metric_seasonal_period: Optional[int] = None,
         path: Optional[Union[str, Path]] = None,
         verbosity: int = 2,
+        log_to_file: bool = True,
+        log_file_path: str = "auto",
```
shchur, line 160: Nit: `Union[str, Path]` type?

shchur (approved):
Thank you, this looks great! Feel free to merge if the tests are passing.

PR #3889, opened 2024-01-26
[timeseries] Add method for plotting forecasts

shchur:
*Description of changes:*
- Add method `TimeSeriesPredictor.plot()` that can plot time series (potentially, with forecasts) using `matplotlib`.
- Update forecasting quickstart tutorial to use the `plot` method for visualizing forecasts.


timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -1187,3 +1189,112 @@ def select_target(ts_df: TimeSeriesDataFrame) -> TimeSeriesDataFrame:
...
+            Dictionary describing the plot style that will be passed to [`matplotlib.pyplot.rc_context`](https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.rc_context.html).
+            See [matplotlib documentation](https://matplotlib.org/stable/users/explain/customizing.html#the-default-matplotlibrc-file) for the list of available options.
+        """
+        import matplotlib.pyplot as plt
+
+        data = self._check_and_prepare_data_frame(data)
+        if item_ids is None:
+            item_ids = list(np.random.choice(data.item_ids, size=min(max_num_item_ids, data.num_items), replace=False))
+        else:
+            item_ids = list(item_ids)[:max_num_item_ids]
```
canerturkmen, line 1235: So If I give you 9 item_ids you will not plot the 9th... The `item_ids` docstring does not spell that out very well, and I would expect this not to be the case.

shchur: I have updated the docstrings to communicate this more clearly.

I have two arguments in favor of keeping the current argument name & effect:

1. I think that limiting the max number of plots to `max_num_item_ids=8` by default is a reasonable safeguard in case the user calls
```python
predictor.plot(data, item_ids=data.item_ids)
```
which may generate thousands of plots and kill the jupyter kernel.
2. If `max_num_item_ids` only applies to the case when `item_ids=None`, we'll need to rename to it something like `max_num_random_item_ids` or `max_num_default_item_ids`, which doesn't sound as intuitive as the current name.

PR #3911, opened 2024-02-09
[Feature][CI] Add Concurrency control to increase efficiency, save resources and bring down cost

prateekdesai04:
*Description of changes:*
This change adds concurrency control to our CI.
If a same PR has multiple commits within a short span of time then the older CI runs are cancelled and the latest one is only run. The same thing would happen for synchronization as well.
This would massively bring down the cost incurred by the AG Team for using AWS Batch for the CI.
It would also free up the vCPUs that are used unnecessarily for old runs, which we can then dedicate to new runs, thus increasing the capacity of our CI.

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -4,6 +4,10 @@ on:
   push:
   pull_request_target:
 
+concurrency:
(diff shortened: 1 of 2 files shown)


Innixma (commented):
LGTM, awesome find!

Innixma (approved):

PR #3933, opened 2024-02-17
[WIP] Code implementation of Conv-LoRA

Harry-zzh:
*Issue #, if available:*

*Description of changes:*


Harry-zzh:
> Great work! One real quick question, have you tried conv-lora on standard text tasks instead of image/SAM tasks? If so, how was it? If havnt tried, do you think it is a more general-purpose PEFT method, or it is more of a SAM/CV-specific approach? Thanks a lot!

Thank you for your question. While I haven't tried text tasks yet, my understanding is that Conv-LoRA is primarily designed for image tasks.

Conv-LoRA incorporates local priors into image features at appropriate scales, considering potential variations in object scale. This involves interpolating image features to larger scales than default and subsequently employing convolution operations for injecting local priors. In our paper, we find that interpolating features to larger scales for local prior injection is more beneficial, given that features in the Vision Transformer (ViT) are downscaled by a factor (e.g., 16) from the original continuous image.

However, texts are 1-D discrete sequences and lack a concept akin to "object scale". Consequently, considering the feature processing in Conv-LoRA and its motivation, it is unsuitable for text tasks.

multimodal/src/autogluon/multimodal/models/conv_lora/moe_conv.py
```diff
@@ -0,0 +1,238 @@
+from typing import Any, Callable, Dict, Iterator, List, Mapping, Optional, Set, Tuple, TypeVar, Union, overload
+
+import numpy as np
+import torch
+from torch import nn
+from torch.distributions.normal import Normal
+
+
+class MoEConv(nn.Module):
```
zhiqiangdon, line 9: Need a more accurate name? `MoEGate`? This class doesn't contain convolutions and is to determine gates.

multimodal/src/autogluon/multimodal/models/conv_lora/modeling_sam.py
zhiqiangdon, line 1: How about creating a folder `custom_hf_models` and put this file there with a new name `modeling_sam_for_conv_lora`? The consideration is that we may encounter similar situations of hosting customized hf model definitions in the future, so all these kinds of models can go here.

multimodal/src/autogluon/multimodal/models/conv_lora/adaptation_layers.py
zhiqiangdon, line 1: Since we already have a `multimodal/src/autogluon/multimodal/models/adaptation_layers.py`, how about we put the definitions of `ConvLoRALinear`, `MoEConv`, and `SparseDispatcher` in that file?

multimodal/src/autogluon/multimodal/configs/optimization/default.yaml
```diff
@@ -34,6 +34,7 @@ optimization:
       - "^o$"
     r: 8
     alpha: 8
+    conv_lora_expert_num: 8 # default setting for Conv-LoRA
```
zhiqiangdon, line 37: Any other important hyperparameters for conv-lora?

multimodal/src/autogluon/multimodal/constants.py
```diff
@@ -139,6 +139,7 @@
 MAE = "mae"
 BER = "ber"
 IOU = "iou"
+MOE_LOSS = "moe_loss"
```
zhiqiangdon, line 142: We can add a new section `# loss` and put `MOE_LOSS = "moe_loss"` there.

multimodal/src/autogluon/multimodal/models/conv_lora/adaptation_layers.py
```diff
@@ -0,0 +1,133 @@
...
+
+    def __init__(
+        self,
+        in_features: int,
+        out_features: int,
+        r: int = 0,
+        lora_alpha: int = 1,
+        lora_dropout: float = 0.0,
+        fan_in_fan_out: bool = False,  # Set this to True if the layer to replace stores weight like (fan_in, fan_out)
+        merge_weights: bool = False,
```
2catycm, line 50: Is Conv-Lora reparameterizable? 
It is more complicated than lora, lora just merge the weights by multiplying the matrices.
but here we have convolutions.

Harry-zzh: Conv-LoRA is not reparameterizable mainly due to its interpolation operation. 
Actually convolutions is not the main reason because convolution layer could be re-parameterized into FC layer in some cases. You could refer to papers about structural re-parameterization for more details.

multimodal/src/autogluon/multimodal/models/sam.py
```diff
@@ -359,6 +369,15 @@ def mask_label_key(self):
...
 
+    def train(self, mode: bool = True):
+        super().train(mode)
+        for module in self.modules():
+            if isinstance(module, ConvLoRALinear):
+                self.output_moe_loss = True
+                return self
+
+        return self
+
```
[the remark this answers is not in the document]
Harry-zzh: We need the MoE loss when calculating the validation loss. During the validation process, the module mode is set to "eval". So we cannot distinguish the validation and inference process here.

zhiqiangdon (approved):
LGTM!

Issue #3923, opened 2024-02-15
 Unable to Affect Model Training with Different random_seed Values in Time-series Module


kristinakupf:
## Description
I'm working with the Time-series module in AutoGluon v0.8.2 and v1.0.0, aiming to assess performance sensitivity to random initialization by feeding alternative seeds during model training. Despite attempts, changing the random_seed or model-specific seed parameter doesn't affect the outcomes (identical results) for models like DeepAR, TemporalSpatialTransformer, etc.

The [documentation]((https://auto.gluon.ai/0.7.0/_modules/autogluon/timeseries/predictor.html)) suggests that random_seed should ensure reproducible results across most models, except for non-deterministic GPU operations. However, my attempts to modify the seed as per the documentation have not yielded different results.

> random_seed : int, optional
>             If provided, fixes the seed of the random number generator for all models. This guarantees reproducible
>             results for most models (except those trained on GPU because of the non-determinism of GPU operations).
> 

Thanks for your help! 

### Attempts to resolve
I've tried setting the seed both directly in the fit method and within hyperparameters, as shown below, without any change in the outcomes:
```
      predictor.fit(
          train_data,
          random_seed=random_seed,
          hyperparameters={
              model_type: model_params,
          },
      )
```

```
model_params = {"seed":random_seed)
      predictor.fit(
          train_data,
          hyperparameters={
              model_type: model_params,
          },
      )
```
Note: model_type and model_params are placeholders for the specific models and parameters used.

### Expected vs. Actual Behavior
I expected that changing the random_seed or seed would result in variations in model training outcomes, reflecting sensitivity to initialization. However, the results remain unchanged regardless of the seed value. I did however observe differences in forecasts from v0.8.2 to v1.0.0.

### Environment
AutoGluon version: v0.8.2, v1.0.0
Python version: 3.7
Operating System: Linux

### References
I've reviewed similar issues (e.g., #425, #1454) without finding a solution that addresses my problem.

### References
I've reviewed similar issues (e.g., #425, #1454) without finding a solution that addresses my problem.

canerturkmen:
Hi @kristinakupf, thanks for raising this issue! Yes, it looks like the random seed lost its effect when the high_quality preset or `DirectTabular` model was used with `NN_TORCH`. I am not sure if this applied to your case as well, but I suspect it might have been the root cause. We now merged a fix for this, and I will close this issue. Please feel free to reopen it if you find it persists. Thanks!!

Issue #3896, opened 2024-01-30
Timeseries: Random Seed Variability for "high_quality" vs "medium_quality" Presets


samuel-woerz:
Hi,
when experimenting with Autogluon, I noticed that when using the "medium_quality" preset and different random seeds, my scores fluctuate somewhat. But when using the "high_quality" preset and different random seeds, the scores are exactly the same (most of the time). The same holds for the ensemble weights: they fluctuate for "medium_quality", but are the same for "high_quality".

I reproduced this behavior using the [In-Depth Tutorial Notebook](https://auto.gluon.ai/stable/tutorials/timeseries/forecasting-indepth.html) as a starting point: [link to colab notebook with results](https://colab.research.google.com/drive/1-ngYlUZu4m78FXPwIfKWJGt1v8jfQr19#scrollTo=r1-aQzKsLhx3) (runtime ist ~40 minutes on T4) 

I also noticed that this behavior seems to correlate with the inclusion of the TemporalFusionTransformer in the final ensemble.

As I understand it, the only difference between the presets are the models considered for the ensemble. 
Therefore, I have no good explanation for this difference in variability. 
Any chance you could share some insights?
Thanks for your time.

canerturkmen:
Hi @samuel-woerz, thank you for raising this issue! We now merged a fix which should take care of it. I will close this issue for now, feel free to reopen if you find it persists. Thanks!!

PR #3934, opened 2024-02-19
[timeseries] fix random seed, reset seed for every model

canerturkmen:
*Issue #, if available:*

#3896 
#3923

*Description of changes:*

Previously, `seed_everything` was called once at the beginning of `TimeSeriesPredictor` `fit` and `predict` methods which led to two undesirable side effects:
- if any of the models reset the seeds in their internal logic, they could effectively cancel the `random_seed` provided by the user. This was done, e.g., in `RecursiveTabular`'s `NN_TORCH` affecting `high_quality` presets.
- The fitting order of models could affect the outcomes of individual models. i.e., the random state at the beginning of each model depended on previous models.

With this change, we propagate the random seed to the learner and trainer both at training and prediction time. We set the seed again, per model, when fitting single models or predicting with single models.

Also fixes a minor issue in predictor unit tests.

### DIFF
--- a/timeseries/tests/unittests/test_predictor.py
+++ b/timeseries/tests/unittests/test_predictor.py
@@ -14,6 +14,7 @@ import pytest
 
 from autogluon.common import space
 from autogluon.common.utils.log_utils import verbosity2loglevel
+from autogluon.common.utils.utils import seed_everything
 from autogluon.timeseries.dataset import TimeSeriesDataFrame
 from autogluon.timeseries.dataset.ts_dataframe import ITEMID, TIMESTAMP
 from autogluon.timeseries.metrics import DEFAULT_METRIC_NAME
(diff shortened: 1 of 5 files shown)


timeseries/tests/unittests/test_predictor.py
```diff
@@ -1087,3 +1095,66 @@ def test_when_predictor_fit_with_verbosity_then_verbosity_overridden_and_propaga
...
+
+    def train_save_side_effect(self, *args, **kwargs):
+        assert torch.get_rng_state().numpy()[0] == random_seed
+
+        # mess with the seed
+        seed_everything(66)
+
+        return "mock_model"
+
+    with mock.patch("autogluon.timeseries.trainer.AbstractTimeSeriesTrainer._train_and_save") as mock_train_save:
```
shchur, line 1114: Nice!

timeseries/tests/unittests/test_predictor.py
```diff
@@ -1034,26 +1036,32 @@ def test_when_log_to_file_set_then_predictor_logs_to_file(temp_model_path):
...
-    # check if the log contains text
-    with open(log_path, "r") as f:
-        log_text = f.read()
-    assert "Naive" in log_text
+        # check if the log contains text
+        with open(log_path, "r") as f:
+            log_text = f.read()
+        assert "Naive" in log_text
+    finally:
+        log_path.unlink(missing_ok=True)
```
shchur, line 1064: Why is this change necessary?

canerturkmen: The test left the `custom_log.txt` file in the local folder. This makes the test cleans up after itself.

shchur (approved):
LGTM! Feel free to merge once the tests pass.

PR #3937, opened 2024-02-20
[timeseries] Add option to keep lightning_logs after training

### LINKED ISSUE #3853: Feature Request: Make Deletion of lightning_logs Directory Optional in TimeSeries Models
## Description
The current implementation of AutoGluon's time series models, specifically those using the GluonTS torch backend, automatically deletes the `lightning_logs` directory after each training run. This directory contains logs that are essential for users who utilize TensorBoard to monitor and compare different training sessions. The automatic deletion of these logs makes it difficult to use TensorBoard effectively, as it relies on historical log data for comparison.

## Proposal
I propose a feature for the `timeseries` module where the deletion of the `lightning_logs` directory is made optional. This could be implemented by adding a parameter to the model training functions and the `fit()` method, allowing users to choose whether to preserve the logs. By default, this parameter could be `True` to maintain the current behavior, but when set to `False`, it would keep the logs intact for further analysis with TensorBoard.

## Code
this code is responsible for deletion of the logs:
 if lightning_logs_dir.exists() and lightning_logs_dir.is_dir(): logger.debug(f"Removing lightning_logs directory {lightning_logs_dir}") shutil.rmtree(lightning_logs_dir)

shchur:
*Issue #, if available:* Fixes #3853

*Description of changes:*
- Add hyperparameter `keep_lightning_logs` to GluonTS models. If set to `True`, `lightning_logs` directory won't be removed after training.

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
+++ b/timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
@@ -393,6 +393,7 @@ class AbstractGluonTSModel(AbstractTimeSeriesModel):
         self._check_fit_params()
         # update auxiliary parameters
         init_args = self._get_estimator_init_args()
+        keep_lightning_logs = init_args.pop("keep_lightning_logs", False)
         callbacks = self._get_callbacks(
             time_limit=time_limit,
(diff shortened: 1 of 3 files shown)


canerturkmen (approved):
LGTM!

Issue #3878, opened 2024-01-23
Add RMSLE-metric


obwohl:
## Please add RMSLE-metric

Here is a proposal (this is my my_metrics.py file I am using right now)

import numpy as np
from autogluon.core.metrics import make_scorer

def rmsle_func(y_true: np.ndarray, y_pred: np.ndarray) -> float:
    # Check for non-positive values to avoid log domain error
    if np.any(y_true < 0) or np.any(y_pred < 0):
        raise ValueError("Root Mean Squared Logarithmic Error cannot be computed for negative values.")
    # Compute the Mean Squared Logarithmic Error
    msle = ((np.log(1 + y_true) - np.log(1 + y_pred)) ** 2).mean()
    # Take the square root to get the RMSLE
    return np.sqrt(msle)

RMSLE_scorer = make_scorer(name='root_mean_squared_log_error',
                           score_func=rmsle_func,
                           optimum=0,
                           greater_is_better=False)

canerturkmen:
Assuming the original request referred to the timeseries module, where RMSLE is most used, I believe this issue has now been addressed. Feel free to reopen it otherwise. Thanks for raising this @obwohl !

PR #3938, opened 2024-02-20
[timeseries] add RMSLE to metrics

canerturkmen:
*Issue #, if available:*

#3878

*Description of changes:*

RMSLE is a forecast error metric popular with some Kaggle competitions, that is specifically insensitive to effects of scale and trends and has an asymmetric penalty for overprediction. For example, the "getting started" Sales Prediction competition uses RMSLE. This PR includes RMSLE to AG-TS.

Discussion points: 
- [x] the current implementation "clips" predictions to 0. This may not be an assumption we would like to make. 

### DIFF
--- a/timeseries/src/autogluon/timeseries/metrics/__init__.py
+++ b/timeseries/src/autogluon/timeseries/metrics/__init__.py
@@ -2,7 +2,7 @@ from pprint import pformat
 from typing import Type, Union
 
 from .abstract import TimeSeriesScorer
-from .point import MAE, MAPE, MASE, MSE, RMSE, RMSSE, SMAPE, WAPE
+from .point import MAE, MAPE, MASE, MSE, RMSE, RMSLE, RMSSE, SMAPE, WAPE
 from .quantile import SQL, WQL
 
 __all__ = [
@@ -12,6 +12,7 @@ __all__ = [
     "SMAPE",
     "MSE",
     "RMSE",
+    "RMSLE",
     "RMSSE",
     "SQL",
     "WAPE",
@@ -25,6 +26,7 @@ AVAILABLE_METRICS = {
     "MAPE": MAPE,
     "SMAPE": SMAPE,
     "RMSE": RMSE,
+    "RMSLE": RMSLE,
     "RMSSE": RMSSE,
     "WAPE": WAPE,
     "SQL": SQL,
--- a/timeseries/src/autogluon/timeseries/metrics/point.py
+++ b/timeseries/src/autogluon/timeseries/metrics/point.py
@@ -306,3 +306,33 @@ class RMSSE(TimeSeriesScorer):
 
         mse_per_item = (y_true - y_pred).pow(2.0).groupby(level=ITEMID, sort=False).mean()
         return np.sqrt(self._safemean(mse_per_item / self._past_squared_seasonal_error))
+
+
+class RMSLE(TimeSeriesScorer):
+    r"""Root mean squared logarithmic error.
+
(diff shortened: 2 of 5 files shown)


timeseries/src/autogluon/timeseries/metrics/point.py
```diff
@@ -306,3 +306,33 @@ def compute_metric(
...
+    References
+    ----------
+    - `Scikit-learn: <https://scikit-learn.org/stable/modules/model_evaluation.html#mean-squared-log-error>`_
+    """
+
+    def compute_metric(self, data_future, predictions, target, **kwargs):
+        y_true, y_pred = self._get_point_forecast_score_inputs(data_future, predictions, target=target)
+        y_pred = np.clip(y_pred, a_min=0.0, a_max=None)
+
+        return np.sqrt(np.power(np.log1p(y_pred) - np.log1p(y_true), 2).mean())
```
shchur, line 338: Should we add a sanity check inside `compute_metric` or `__call__` (the latter won't slow down ensemble, but requires more code)?
```python
    def __call__(
        self,
        data: TimeSeriesDataFrame,
        predictions: TimeSeriesDataFrame,
        prediction_length: int = 1,
        target: str = "target",
        seasonal_period: Optional[int] = None,
        **kwargs,
    ) -> float:
        if (data[target] < 0).any():
            raise ValueError(f"{self.name} cannot be used if target time series contains negative values!")
        return super().__call__(data=data, predictions=predictions, prediction_length=prediction_length, target=target, seasonal_period=seasonal_period, **kwargs)
```

timeseries/src/autogluon/timeseries/metrics/__init__.py
```diff
@@ -25,6 +26,7 @@
     "MAPE": MAPE,
     "SMAPE": SMAPE,
     "RMSE": RMSE,
+    "RMSLE": RMSLE,
     "RMSSE": RMSSE,
     "WAPE": WAPE,
     "SQL": SQL,
```
shchur, line 32: We also need to add the new metric to the following files
- [ ] `timeseries/src/autogluon/timeseries/predictor.py` (line 90)
- [ ] `docs/tutorials/timeseries/forecasting-metrics.md` (lines 46, 141, 187)

timeseries/src/autogluon/timeseries/metrics/__init__.py
```diff
@@ -2,7 +2,7 @@
 from typing import Type, Union
```
shchur, line 2: Should we add a unittest to ensure that our implementation doesn't break over time (similar to https://github.com/autogluon/autogluon/blob/master/timeseries/tests/unittests/test_metrics.py#L219)?

timeseries/src/autogluon/timeseries/metrics/point.py
canerturkmen: cool. wdyt about clipping `y_pred` ?

shchur: I think that's reasonable. Another alternative is to take the absolute value, but clipping to zero feels more natural (if model predicts negative demand, the user will likely clip it to 0 rather than take the absolute value).

timeseries/src/autogluon/timeseries/metrics/__init__.py
shchur: Maybe we should have some checklists in documentation for adding new metrics & new models

shchur (approved):
LGTM! Thanks!

PR #3956, opened 2024-02-29
Fix windows python 3.11 issue by removing ray

tonyhoo:
*Issue #, if available:*

To fix AutoGluon installation [issue](https://github.com/autogluon/autogluon/issues/3807) on windows with Python 3.11

*Description of changes:*
Currently the workaround is not to install ray till ray release a version which support windows installation

### DIFF
--- a/core/setup.py
+++ b/core/setup.py
@@ -5,6 +5,8 @@ import importlib.util
 import os
 
 from setuptools import setup
+import sys
+import platform
 
 filepath = os.path.abspath(os.path.dirname(__file__))
 filepath_import = os.path.join(filepath, "..", "core", "src", "autogluon", "core", "_setup_utils.py")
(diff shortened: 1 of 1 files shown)


core/setup.py
```diff
@@ -45,18 +47,23 @@
...
-    ],
-    "raytune": [
-        "ray[default,tune]>=2.6.3,<2.7",
-        # TODO: consider alternatives as hyperopt is not actively maintained.
-        "hyperopt>=0.2.7,<0.2.8",  # This is needed for the bayes search to work.
-        # 'GPy>=1.10.0,<1.11.0'  # TODO: Enable this once PBT/PB2 are supported by ray lightning
-    ],
-}
+# Check for Windows and Python 3.11
+is_windows_and_py311 = platform.system().lower() == "windows" and sys.version_info.major == 3 and sys.version_info.minor == 11
```
Innixma, line 51: I don't think this will solve the problem when doing a pip install. The `setup.py` to my understanding is not executed in pip install.

tonyhoo: changed to environment markers and tested with `pip install autogluon-tonyhu-test==1.0.5b20240302`

Innixma (approved):
LGTM!

core/setup.py
```diff
@@ -3,6 +3,8 @@
 # This code block is a HACK (!), but is necessary to avoid code duplication. Do NOT alter these lines.
 import importlib.util
 import os
+import platform
+import sys
```
Innixma, line 7: unnecessary imports

core/setup.py
Innixma: Is “win32” the correct identifier? What are the other “winXYZ” markers? Is there a list somewhere?

PR #3962, opened 2024-03-04
[CI][Feature] Package Version Comparator

prateekdesai04:
*Issue #, if available:* https://github.com/autogluon/autogluon/issues/3952

*Description of changes:*
This feature compares the versions of all the dependencies installed while installing AutoGluon.
1. A file listing all the dependencies is prepared for every CI run
2. A file from the previous latest CI run is fetched and compared with the current CI run, following which differences are listed

Currently the difference can be viewed in the CI run (**_package_diff_**) task
I will also be putting a follow-up PR which will enhance the user experience, by displaying the diff in github comments once a PR is raised.


Innixma (approved):
LGTM, excited to see this one in action! Thanks for the contribution

zhiqiangdon (commented):
LGTM. Thanks for enabling this automatic detection. Looking forward to follow-up PRs to make this easier to use.

.github/workflows/continuous_integration.yml
```diff
@@ -46,6 +46,44 @@ jobs:
...
+      - name: Store package versions
+        run: |
+          cur_date=$(date +%Y-%m-%d-%H:%M:%S)
+          echo "Timestamp: $cur_date" > ./package_versions_${cur_date}.txt
+          echo "Package Versions:" >> ./package_versions_${cur_date}.txt
+          pip freeze >> ./package_versions_${cur_date}.txt
+      - name: Fetch previous version file and compare
+        run: |
+          /bin/bash ./.github/workflow_scripts/version_diff.sh
+        continue-on-error: true
```
tonyhoo, line 83: Why should we proceed with the workflow in spite of a step failing? Is it acceptable for issues to remain unnoticed for a period of time?

.github/workflow_scripts/version_diff.sh
```diff
@@ -0,0 +1,41 @@
...
+        latest_timestamp=$timestamp_seconds
+        latest_file=$filename
+    fi
+done <<< "$files"
+
+aws s3 cp s3://autogluon-ci/package_versions/$latest_file ./
+old_latest_file="old_${latest_file}"
+mv "./$latest_file" "./$old_latest_file"
+
+diff ./package_versions_* ./old_package_versions_* > ./diff_output.txt
```
tonyhoo, line 30: The diff command compares files line by line in a sequential manner. If the packages are listed in a different order between two files, diff will report these as differences, even if the same packages and versions are present. This can make the output noisy and harder to understand, especially in case of package addition or deletion

.github/workflows/continuous_integration.yml
prateekdesai04: Because if the folder in the S3 bucket is empty and there is no initial file to compare with then we should still proceed to the next step which is to upload the current pip freeze file to S3 and this can be the first file.
Package diff failing is acceptable as we only use this to diff and not to pass or fail the entire workflow.

.github/workflow_scripts/version_diff.sh
prateekdesai04: `pip freeze` will always list the packages in a sorted order so the `diff` command will compare two sorted files.
If a new package is added or deleted then yes the output may be noisy, that's where we would download the current and previous file, remove the line containing the latest added or deleted package and then `diff`, this effort may have to be manual. It seems to be okay as we do not add or delete packages often

.github/workflows/continuous_integration.yml
tonyhoo: My primary interest lies in learning how to distinguish between expected failures, such as package discrepancies or non-existent s3 files, and unexpected ones, like tool malfunctions or missing credentials. Without this differentiation, and without the necessary manual upkeep, the tool is likely to cease functioning over time, primarily due to these unforeseen issues.

.github/workflow_scripts/version_diff.sh
tonyhoo: The current implementation is acceptable as it primarily functions as a support tool to help authors determine if functional failures are associated with package version changes.

.github/workflows/continuous_integration.yml
prateekdesai04: `(continue-on-error: true)` is only valid for `[Fetch previous version file and compare]` step it's not applicable to any steps before or after it. If steps before or after fail, then the workflow will fail and will give out the specific error which will help us understand unexpected failures (like missing credentials or error while configuring credentials).

Issue #3886, opened 2024-01-26
[timeseries] Add first-class support for missing values


shchur:
Currently, `TimeSeriesPredictor` deals with missing values in the data by first imputing them via forward-/backward-filling, and then training all models as if there are no missing values. This strategy may lead to poor accuracy on datasets with a large portion of missing values since training data will include regions with constant values arising from ffilling. 

A better alternative is to keep the missing values represented by `nan` in the data, and let models handle it. This requires following modifications to the code:
- [x] Update metric implementations to handle missing values in target (#3966)
- [ ] Update preprocessing logic in `TimeSeriesPredictor._check_and_prepare_data_frame`
- [ ] Make sure that all models can handle missing values. This means, all models can train normally and produce forecasts with no NaN values, even if training data contains NaNs.
  - [ ] GluonTS models (DeepAR, TFT, PatchTST, DLinear)
  - [ ] StatsForecast models (AutoETS, AutoARIMA, Theta, intermittent demand models)
  - [ ] Baseline models (Naive, SeasonalNaive, Average, SeasonalAverage, Zero)
  - [ ] MLForecast models (DirectTabular, RecursiveTabular)

PR #3966, opened 2024-03-05
[timeseries] Ensure that all metrics handle missing values in the target

shchur:
*Issue #, if available:* #3886

*Description of changes:*
- Raise exception in `TimeSeriesScorer` if predictions contain NaN values
- Update implementation of WQL metric to handle NaN values in target column
- Update our implementation of MASE and SQL metrics to align with GluonTS in case missing values are present
  - As part of this update, we moved from `pandas + groupby` to `numpy + reshape`, which results in faster metric computations & lower fit time for WeightedEnsemble (for M4 Monthly with 48K time series, ensemble fit time decreases 200s -> 160s)
  - Update the RMSSE implementation to follow the same missing values handling logic
- Add new tests & update some old tests to use dataframes with missing values

### DIFF
--- a/timeseries/src/autogluon/timeseries/metrics/point.py
+++ b/timeseries/src/autogluon/timeseries/metrics/point.py
@@ -247,8 +247,11 @@ class MASE(TimeSeriesScorer):
             raise AssertionError("Call `save_past_metrics` before `compute_metric`")
 
         num_items = len(self._past_abs_seasonal_error)
-        mae_per_item = np.abs(y_true.values - y_pred.values).reshape([num_items, -1])  # [num_items, prediction_length]
-        return np.nanmean(mae_per_item / self._past_abs_seasonal_error.values[:, None])
+        # Reshape MAE values into [num_items, prediction_length] to normalize per item without groupby
+        mae_reshaped = np.abs(y_true.values - y_pred.values).reshape([num_items, -1])
+        # We assume that items are in the same order in both arrays because predictor sorts item_ids
+        mase_per_item = mae_reshaped / self._past_abs_seasonal_error.values[:, None]
+        return np.nanmean(mase_per_item)
 
 
 class RMSSE(TimeSeriesScorer):
--- a/timeseries/src/autogluon/timeseries/metrics/point.py
+++ b/timeseries/src/autogluon/timeseries/metrics/point.py
@@ -249,9 +249,8 @@ class MASE(TimeSeriesScorer):
         num_items = len(self._past_abs_seasonal_error)
         # Reshape MAE values into [num_items, prediction_length] to normalize per item without groupby
         mae_reshaped = np.abs(y_true.values - y_pred.values).reshape([num_items, -1])
-        # We assume that items are in the same order in both arrays because predictor sorts item_ids
-        mase_per_item = mae_reshaped / self._past_abs_seasonal_error.values[:, None]
-        return np.nanmean(mase_per_item)
+        # We assume that items are in the same order in both arrays because predictor sorts by item_id
+        return self._safemean(mae_reshaped / self._past_abs_seasonal_error.values[:, None])
 
 
 class RMSSE(TimeSeriesScorer):
--- a/timeseries/src/autogluon/timeseries/metrics/point.py
+++ b/timeseries/src/autogluon/timeseries/metrics/point.py
@@ -306,7 +306,7 @@ class RMSSE(TimeSeriesScorer):
         if self._past_squared_seasonal_error is None:
             raise AssertionError("Call `save_past_metrics` before `compute_metric`")
 
-        num_items = len(self._past_abs_seasonal_error)
+        num_items = len(self._past_squared_seasonal_error)
         # Reshape squared errors into [num_items, prediction_length] to normalize per item without groupby
         squared_errors = ((y_true.values - y_pred.values) ** 2.0).reshape([num_items, -1])
         # We assume that items are in the same order in both arrays because predictor sorts by item_id
--- a/timeseries/src/autogluon/timeseries/metrics/point.py
+++ b/timeseries/src/autogluon/timeseries/metrics/point.py
@@ -246,8 +246,9 @@ class MASE(TimeSeriesScorer):
         if self._past_abs_seasonal_error is None:
             raise AssertionError("Call `save_past_metrics` before `compute_metric`")
 
-        mae_per_item = (y_true - y_pred).abs().groupby(level=ITEMID, sort=False).mean()
-        return self._safemean(mae_per_item / self._past_abs_seasonal_error)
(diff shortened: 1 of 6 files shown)


timeseries/src/autogluon/timeseries/metrics/point.py
```diff
@@ -246,8 +246,11 @@ def compute_metric(
         if self._past_abs_seasonal_error is None:
             raise AssertionError("Call `save_past_metrics` before `compute_metric`")
 
-        mae_per_item = (y_true - y_pred).abs().groupby(level=ITEMID, sort=False).mean()
-        return self._safemean(mae_per_item / self._past_abs_seasonal_error)
+        num_items = len(self._past_abs_seasonal_error)
+        # Reshape MAE values into [num_items, prediction_length] to normalize per item without groupby
+        mae_reshaped = np.abs(y_true.values - y_pred.values).reshape([num_items, -1])
+        # We assume that items are in the same order in both arrays because predictor sorts by item_id
+        return self._safemean(mae_reshaped / self._past_abs_seasonal_error.values[:, None])
```
shchur, line 253: This change is necessary to make sure that our MASE/SQL implementation aligns with GluonTS if missing values are present. For example, suppose that `prediction_length=100` and we have two items: 
- item A with 1 observed value in the forecast horizon
- item B with 100 observed values in the forecast horizon

In our old implementation with groupby, we'd first compute the mean for each item, and then take mean across items. This would assign 100x more weight to the single value for item A compared to the values in item B.

In the new implementation, each observed value in the forecast horizon is weighted equally (same as in GluonTS).

timeseries/tests/unittests/common.py
```diff
@@ -185,9 +197,12 @@ def clear_past_metrics(self) -> None:
...
 
-def get_prediction_for_df(data):
-    predictor = TimeSeriesPredictor(prediction_length=5).fit(data, hyperparameters={"Naive": {}})
-    return predictor.predict(data)
+def get_prediction_for_df(data, prediction_length=5):
+    forecast_index = get_forecast_horizon_index_ts_dataframe(data, prediction_length=prediction_length)
+    columns = ["mean", "0.1", "0.5", "0.9"]
+    return TimeSeriesDataFrame(
+        pd.DataFrame(np.random.normal(size=[len(forecast_index), len(columns)]), index=forecast_index, columns=columns)
+    )
```
shchur, line 205: This change is necessary because `TimeSeriesPredictor.fit()` currently fails if some time series in the training set contain of all NaNs.

timeseries/tests/unittests/test_metrics.py
```diff
@@ -24,24 +24,29 @@
...
-def get_ag_and_gts_metrics() -> List[Tuple[str, str, GluonTSMetric]]:
-    # Each entry is a tuple (ag_metric_name, gts_metric_name, gts_metric_object)
+def get_ag_and_gts_metrics() -> List[Tuple[str, GluonTSMetric]]:
+    # Each entry is a tuple (ag_metric_name, gts_metric_object)
     default_quantile_levels = [0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9]
     # Metric that have different names in AutoGluon and GluonTS
     ag_and_gts_metrics = [
-        ("WQL", "mean_weighted_sum_quantile_loss", MeanWeightedSumQuantileLoss(default_quantile_levels)),
-        ("SQL", "average_mean_scaled_quantile_loss", AverageMeanScaledQuantileLoss(default_quantile_levels)),
-        ("WAPE", "ND[mean]", ND("mean")),
```
shchur, line 39: The second argument in the tuple was unused, so I removed it.

timeseries/src/autogluon/timeseries/metrics/point.py
```diff
@@ -246,8 +245,11 @@ def compute_metric(
         if self._past_abs_seasonal_error is None:
             raise AssertionError("Call `save_past_metrics` before `compute_metric`")
 
-        mae_per_item = (y_true - y_pred).abs().groupby(level=ITEMID, sort=False).mean()
-        return self._safemean(mae_per_item / self._past_abs_seasonal_error)
+        num_items = len(self._past_abs_seasonal_error)
+        # Reshape abs errors into [num_items, prediction_length] to normalize per item without groupby
+        abs_errors = np.abs(y_true.values - y_pred.values).reshape([num_items, -1])
+        # We assume that items are in the same order in both arrays because predictor sorts by item_id
+        return self._safemean(abs_errors / self._past_abs_seasonal_error.values[:, None])
```
shchur, line 252: This change is necessary to make sure that our MASE/SQL/RMSSE implementation aligns with GluonTS if missing values are present. For example, suppose that `prediction_length=100` and we have two items: 
- item A with 1 observed value in the forecast horizon
- item B with 100 observed values in the forecast horizon

In our old implementation with groupby, we'd first compute the mean for each item, and then take mean across items. This would assign 100x more weight to the single value for item A compared to the values in item B.

In the new implementation, each observed value in the forecast horizon is weighted equally (same as in GluonTS).

canerturkmen (commented):
Looks good overall. Just one comment. Thanks for the housekeeping!! 🧹

timeseries/src/autogluon/timeseries/metrics/point.py
shchur: One will need to get really creative to violate this assumption. Even if one uses a custom `TimeSeriesScorer` outside of the predictor, the following code will work because `predictor.predict` preserved the item_id order
```python
data = TimeSeriesDataFrame(...)  # load data with unsorted item_ids
predictions = TimeSeriesPredictor(...).fit(data).predict(data)  # `predictions` have the same item_id order as `data`
mase = MASE()(data, predictions, ...)  # this works
```
One will need to explicitly shuffle `predictions`, or shuffle `data` before calling the scorer to violate this assumption.

canerturkmen: OK so more generally, data and predictions should have the same index order. As long as it's documented clearly I'm OK.

shchur: You are absolutely right, `TimeSeriesScorer` always assumes that predictions & data have the same index, otherwise an AssertionError is raised https://github.com/autogluon/autogluon/blob/master/timeseries/src/autogluon/timeseries/metrics/abstract.py#L80. 

I guess I can completely remove the following comment
```python
# We assume that items are in the same order in both arrays because predictor sorts by item_id
```

canerturkmen: Sure. Also making it clearer in the docstrings would be great.

shchur: Done :+1:

canerturkmen (approved):
Thanks!!

PR #3970, opened 2024-03-09
[AutoMM] Improve DINO performance

FANGAreNotGnu:
Update the settings to which used in automl maper


multimodal/src/autogluon/multimodal/presets.py
```diff
@@ -390,26 +389,40 @@ def object_detection(presets: str = DEFAULT):
...
+                "env.batch_size": 32,
+                "env.per_gpu_batch_size": 1,
+                "env.num_workers": 2,
+                "optimization.learning_rate": 1e-5,
+                "optimization.weight_decay": 1e-4,
+                "optimization.lr_mult": 10,
+                "optimization.lr_choice": "two_stages",  # TODO: changed
+                "optimization.lr_schedule": "multi_step",
+                "optimization.gradient_clip_val": 0.1,
+                "optimization.max_epochs": 60,
```
zhiqiangdon, line 405: Since high and best share these same hyperparameters, how setting them in the shared hyperparameters at the top of this function and moving the previous hyperparameters into the medium quality preset? i.e., reducing duplicate code.

multimodal/src/autogluon/multimodal/models/mmdet_image.py
```diff
@@ -341,8 +344,50 @@ def get_layer_ids(
...
+        Returns
+        -------
+        A dictionary mapping the layer names (keys) to their ids (values).
+        """
+
+        # two stage lr: backbone v.s. else, or head v.s. else
+        if self.backbone_layers:
+            return self.get_layer_ids_by_backbone()
+        else:
+            return self.get_layer_ids_by_head()
```
zhiqiangdon, line 356: Why needs to change the logic of getting layer ids? The `two_stage` lr choice doesn't rely on layer ids.

multimodal/src/autogluon/multimodal/models/mmdet_image.py
```diff
@@ -341,8 +344,50 @@ def get_layer_ids(
...
+        A dictionary mapping the layer names (keys) to their ids (values).
+        """
+        name_to_id = {}
+        backbones = self.backbone_layers
+
+        for n, _ in self.named_parameters():
+            name_to_id[n] = 0
+            for backbone in backbones:
+                if backbone in n:
+                    name_to_id[n] = 1
```
zhiqiangdon, line 379: Wrong logic of using a nested for loop?

multimodal/src/autogluon/multimodal/presets.py
```diff
@@ -349,7 +349,6 @@ def object_detection(presets: str = DEFAULT):
         "model.names": ["mmdet_image"],
         "model.mmdet_image.checkpoint_name": "yolox_s",
         "env.eval_batch_size_ratio": 1,
-        "env.precision": 32,
```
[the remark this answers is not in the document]
FANGAreNotGnu: In our offline benchmarking, using 16 mixed can achieve similar performance with half training time.

multimodal/src/autogluon/multimodal/optimization/lit_mmdet.py
```diff
@@ -239,15 +240,21 @@ def configure_optimizers(self):
...
-            num_warmup_steps=warmup_steps,
-            lr_schedule=self.hparams.lr_schedule,
-            end_lr=self.hparams.end_lr,
-        )
 
-        sched = {"scheduler": scheduler, "interval": "step"}
+        # TODO: Integrate MultiStepLR into get_lr_scheduler and add milestones, gamma, and interval into hyperparameters
+        if self.hparams.lr_schedule == "multi_step":
+            scheduler = torch.optim.lr_scheduler.MultiStepLR(optimizer=optimizer, milestones=[30, 55], gamma=0.1)
+            sched = {"scheduler": scheduler, "interval": "epoch"}
```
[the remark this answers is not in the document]
FANGAreNotGnu: the interval is different

multimodal/src/autogluon/multimodal/optimization/lit_mmdet.py
```diff
@@ -239,15 +240,21 @@ def configure_optimizers(self):
...
-            num_max_steps=max_steps,
-            num_warmup_steps=warmup_steps,
-            lr_schedule=self.hparams.lr_schedule,
-            end_lr=self.hparams.end_lr,
-        )
 
-        sched = {"scheduler": scheduler, "interval": "step"}
+        # TODO: Integrate MultiStepLR into get_lr_scheduler and add milestones, gamma, and interval into hyperparameters
+        if self.hparams.lr_schedule == "multi_step":
+            scheduler = torch.optim.lr_scheduler.MultiStepLR(optimizer=optimizer, milestones=[30, 55], gamma=0.1)
```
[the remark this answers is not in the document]
FANGAreNotGnu: We need to add a few hyperparameters like milestones and gamma value if we make it more general and add to lr_scheduler. Shall we add those hyperparameters into our config given that this multistep lr schedule is only used in DINO?

multimodal/src/autogluon/multimodal/models/mmdet_image.py
FANGAreNotGnu: In mmdet_model, we assign `self.name_to_id = self.get_layer_ids()` and `self.head_layer_names = [n for n, layer_id in self.name_to_id.items() if layer_id <= 0]`

multimodal/src/autogluon/multimodal/models/mmdet_image.py
FANGAreNotGnu: If any backbone pattern is included in parameter names, it's considered as a backbone layer.

multimodal/src/autogluon/multimodal/presets.py
FANGAreNotGnu: Done

examples/automm/object_detection/benchmarking.py
```diff
@@ -0,0 +1,50 @@
+import argparse
+import uuid
+import os
+import time
+
+from autogluon.multimodal import MultiModalPredictor
+
+BENCH_ROOT = "/media/ag/data/AutoMLDetBench"
```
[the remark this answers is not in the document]
FANGAreNotGnu: Moved to arguments

multimodal/src/autogluon/multimodal/optimization/lit_mmdet.py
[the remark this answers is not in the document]
zhiqiangdon: We can add the default arguments `gamma` and `milestones` in the `get_lr_scheduler` function. Later, we can make them customizable through the config.

multimodal/src/autogluon/multimodal/presets.py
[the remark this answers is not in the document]
zhiqiangdon: Curious about what changes make 16-mixed work now but not before? 16-mixed is the default in config, which you should have tried before.

multimodal/src/autogluon/multimodal/models/mmdet_image.py
```diff
@@ -68,7 +68,10 @@ def __init__(
         self.checkpoint_name = checkpoint_name
         self.config_file = config_file
         self.classes = classes
-        self.frozen_layers = frozen_layers
+        # Based on our offline benchmarking results, instead of freezing layers,
+        # Setting backbone to a smaller learning rate achieves better results,
+        self.frozen_layers = []
+        self.backbone_layers = frozen_layers
```
zhiqiangdon, line 74: 1. Assigning `frozen_layers` to `backbone_layers` looks confusing. How about having a new attribute `backbone_layers`?
2. This change also deprecates the previous functionality of supporting freezing layers. Is this deprecation a good choice? Maybe we still support freezing layers as a choice.

examples/automm/object_detection/benchmarking.py
```diff
@@ -0,0 +1,51 @@
...
+    )
+    train_end = time.time()
+    print("The finetuning takes %.2f seconds." % (train_end - start))
+
+    predictor.evaluate(test_path)
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-r", "--benchmark_root", default="/media/ag/data/AutoMLDetBench", type=str)
```
zhiqiangdon, line 40: How about using a relative path by default? I guess some users may not have the permission of using this path.

multimodal/src/autogluon/multimodal/presets.py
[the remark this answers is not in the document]
FANGAreNotGnu: Previously was trying to get the highest performance without considering the cost.

multimodal/src/autogluon/multimodal/models/mmdet_image.py
FANGAreNotGnu: Normally frozen_layers and backbone_layers should refer to the same layers. It's just are we freezing the backbone layers or using a small lr. Adding a new hyperparameter seems redundant.

multimodal/src/autogluon/multimodal/optimization/lit_mmdet.py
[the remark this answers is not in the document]
FANGAreNotGnu: Done.

examples/automm/object_detection/benchmarking.py
FANGAreNotGnu: Done.

multimodal/src/autogluon/multimodal/models/mmdet_image.py
zhiqiangdon: I think we need to make names clear to improve code readability. Otherwise, others probably can't understand them correctly. If we no longer support freezing layers, then we should remove it to reduce confusion.

FANGAreNotGnu: Sent an up date to remain frozen layers, and set up backbone to have low lr for dino

zhiqiangdon (approved):
LGTM. Need to address the TODO items later. The specific config of DINO should be reflected in the preset hyperparameters instead of the hard coding.

PR #3975, opened 2024-03-12
Add note on pd.to_datetime

Innixma:
*Issue #, if available:*

*Description of changes:*
Add note on pd.to_datetime describing when it can be slow.


suzhoum (approved):
Thanks for looking into this!

PR #3977, opened 2024-03-12
Fix refit_full crash when OOM

### LINKED ISSUE #3936: Disable dynamic_stacking when `use_bag_holdout=True`
Disable dynamic_stacking when `use_bag_holdout=True`.

bag holdout enables to pick the correct models and avoids the issue with stack info leakage. `dynamic_stacking` is redundant in this scenario.

Resolved by #3977

### LINKED ISSUE #3946: Tabular: Fallback to bag fold model for refit if out-of-memory / not-enough-memory error occurs when fitting `_FULL` model.
Fallback to bag fold model for refit if out-of-memory / not-enough-memory error occurs when fitting `_FULL` model.

This will safeguard against low-memory scenarios where there isn't sufficient memory to train the `_FULL` model, but refit_full is specified. Currently this will lead to a later exception. Instead, we can fallback to the first fold model trained in the bag to ensure it doesn't crash.

Related: This issue is a band-aid solution to #3935, which is the overarching memory error problem.

Innixma:
*Issue #, if available:*

Resolves #3936, Resolves #3946

*Description of changes:*
- Fix refit_full crash when OOM. This will detect if a model fit failed during refit and fallback to cloning the first fold of the parent model instead, avoiding an exception.
- Note: This increases peak disk usage by ~8x for `good` and `high` quality presets if save_bag_folds=True, which is mostly unavoidable to ensure refit_full works without failure.

Additionally:
- Optimized calibrate_decision_threshold and WeightedEnsemble fit to be 4x faster on binary classification via an optimization to `get_pred_from_proba`.
- Reduced WeightedEnsemble ensemble_size from 100 to 25, speeding up training time by 4x.
- Adding subsample_size arguments to calibrate_decision_threshold and WeightedEnsemble, with default to 1 million. This will provide a linear speedup on calibration and ensemble fitting for large datasets with over 1 million rows. For 2 million rows, it will be a 2x speedup. For 50 million rows, it will be a 50x speedup, etc.

### DIFF
--- a/tabular/src/autogluon/tabular/configs/presets_configs.py
+++ b/tabular/src/autogluon/tabular/configs/presets_configs.py
@@ -5,7 +5,7 @@ tabular_presets_dict = dict(
     # Aliases: best
     best_quality={
         "auto_stack": True,
-        "dynamic_stacking": True,
+        "dynamic_stacking": "auto",
         "num_bag_sets": 1,
         "hyperparameters": "zeroshot",
         "time_limit": 3600,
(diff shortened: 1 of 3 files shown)


tabular/src/autogluon/tabular/configs/presets_configs.py
```diff
@@ -21,10 +21,10 @@
         "time_limit": 3600,
         "refit_full": True,
         "set_best_to_refit_full": True,
-        "_save_bag_folds": False,
+        # "_save_bag_folds": False,
```
suzhoum, line 24: Is this just a temporary change? I see there is a hack related to this option.

Innixma: Temporary change to check performance.

yinweisu:
| :---: | :---: |
| lightning-utilities==0.10.1 | lightning-utilities==0.11.0 |
| lightning-utilities==0.10.1 | lightning-utilities==0.11.0 |

yinweisu:
| :---: | :---: |
| botocore==1.34.68 | botocore==1.34.69 |
| boto3==1.34.68 | boto3==1.34.69 |
| botocore==1.34.68 | botocore==1.34.69 |
| boto3==1.34.68 | boto3==1.34.69 |

yinweisu:
| :---: | :---: |

yinweisu:
| :---: | :---: |
| botocore==1.34.75 | botocore==1.34.76 |
| boto3==1.34.75 | boto3==1.34.76 |
| contourpy==1.2.0 | contourpy==1.2.1 |
| botocore==1.34.75 | botocore==1.34.76 |
| boto3==1.34.75 | boto3==1.34.76 |
| contourpy==1.2.0 | contourpy==1.2.1 |

tabular/src/autogluon/tabular/configs/presets_configs.py
Innixma: Change has been reverted after testing

prateekdesai04 (approved):
LGTM!

PR #3978, opened 2024-03-13
[timeseries] Add Chronos pretrained forecasting model

canerturkmen:
*Issue #, if available:*

*Description of changes:*

Add [Chronos](https://huggingface.co/amazon/chronos-t5-base), a pretrained time series model for forecasting. 

Remaining todos:
- [x] revise docstring
- [x] add model-specific unit tests
- [x] revise presets and document them
- [x] unify setup requirements with other modules (see #3984)
- [x] add openvino support and test cases
- [x] set logger verbosity and silence huggingface / optimum logs 
- [x] test with AG-Cloud 
- [x] benchmark dynamic context lengths

### DIFF
--- a/timeseries/src/autogluon/timeseries/configs/presets_configs.py
+++ b/timeseries/src/autogluon/timeseries/configs/presets_configs.py
@@ -1,4 +1,5 @@
 """Preset configurations for autogluon.timeseries Predictors"""
+from autogluon.timeseries.models.presets import get_default_hps
 
 # TODO: change default HPO settings when other HPO strategies (e.g., Ray tune) are available
 # TODO: add refit_full arguments once refitting is available
@@ -8,4 +9,39 @@ TIMESERIES_PRESETS_CONFIGS = dict(
     high_quality={"hyperparameters": "default"},
     medium_quality={"hyperparameters": "light"},
     fast_training={"hyperparameters": "very_light"},
+    chronos_tiny={
+        "hyperparameters": {"Chronos": {"model_path": "amazon/chronos-t5-tiny", "skip_validation": True}},
+        "disable_length_check": True,
+    },
+    chronos_mini={
+        "hyperparameters": {"Chronos": {"model_path": "amazon/chronos-t5-mini", "skip_validation": True}},
+        "disable_length_check": True,
+    },
+    chronos_small={
+        "hyperparameters": {"Chronos": {"model_path": "amazon/chronos-t5-small", "skip_validation": True}},
+        "disable_length_check": True,
+    },
+    chronos_base={
+        "hyperparameters": {"Chronos": {"model_path": "amazon/chronos-t5-base", "skip_validation": True}},
+        "disable_length_check": True,
+    },
+    chronos_large={
+        "hyperparameters": {
+            "Chronos": {"model_path": "amazon/chronos-t5-large", "skip_validation": True, "batch_size": 8}
+        },
+        "disable_length_check": True,
+    },
+    chronos_mini_ensemble={
+        "hyperparameters": {"Chronos": {"model_path": "amazon/chronos-t5-mini"}} | get_default_hps("default"),
+    },
+    chronos_small_ensemble={
+        "hyperparameters": {"Chronos": {"model_path": "amazon/chronos-t5-small"}} | get_default_hps("default"),
+    },
+    chronos_base_ensemble={
+        "hyperparameters": {"Chronos": {"model_path": "amazon/chronos-t5-base"}} | get_default_hps("default"),
+    },
+    chronos_large_ensemble={
+        "hyperparameters": {"Chronos": {"model_path": "amazon/chronos-t5-large", "batch_size": 8}}
+        | get_default_hps("default"),
+    },
 )
--- a/timeseries/src/autogluon/timeseries/configs/presets_configs.py
+++ b/timeseries/src/autogluon/timeseries/configs/presets_configs.py
@@ -10,29 +10,29 @@ TIMESERIES_PRESETS_CONFIGS = dict(
     medium_quality={"hyperparameters": "light"},
     fast_training={"hyperparameters": "very_light"},
     chronos_tiny={
-        "hyperparameters": {"Chronos": {"model_path": "amazon/chronos-t5-tiny"}},
+        "hyperparameters": {"Chronos": {"model_path": "tiny"}},
     },
     chronos_mini={
-        "hyperparameters": {"Chronos": {"model_path": "amazon/chronos-t5-mini"}},
+        "hyperparameters": {"Chronos": {"model_path": "mini"}},
     },
     chronos_small={
-        "hyperparameters": {"Chronos": {"model_path": "amazon/chronos-t5-small"}},
+        "hyperparameters": {"Chronos": {"model_path": "small"}},
     },
     chronos_base={
-        "hyperparameters": {"Chronos": {"model_path": "amazon/chronos-t5-base"}},
+        "hyperparameters": {"Chronos": {"model_path": "base"}},
     },
     chronos_large={
-        "hyperparameters": {"Chronos": {"model_path": "amazon/chronos-t5-large", "batch_size": 8}},
+        "hyperparameters": {"Chronos": {"model_path": "large", "batch_size": 8}},
     },
     chronos_ensemble={
         "hyperparameters": {
-            **{"Chronos": {"model_path": "amazon/chronos-t5-small"}},
+            **{"Chronos": {"model_path": "small"}},
             **get_default_hps("default"),
         }
     },
     chronos_large_ensemble={
         "hyperparameters": {
-            **{"Chronos": {"model_path": "amazon/chronos-t5-large", "batch_size": 8}},
+            **{"Chronos": {"model_path": "large", "batch_size": 8}},
             **get_default_hps("default"),
         }
     },
--- a/timeseries/src/autogluon/timeseries/configs/presets_configs.py
+++ b/timeseries/src/autogluon/timeseries/configs/presets_configs.py
@@ -26,19 +26,19 @@ TIMESERIES_PRESETS_CONFIGS = dict(
     },
     chronos_ensemble={
         "hyperparameters": {
-            **{"Chronos": {"model_path": "small"}},
+            "Chronos": {"model_path": "small"},
             **get_default_hps("default"),
         }
     },
     chronos_large_ensemble={
         "hyperparameters": {
-            **{"Chronos": {"model_path": "large", "batch_size": 8}},
+            "Chronos": {"model_path": "large", "batch_size": 8},
             **get_default_hps("default"),
         }
     },
 )
 
-timeseries_presets_aliases = dict(
+TIMESERIES_PRESETS_ALIASES = dict(
     chronos="chronos_small",
     best="best_quality",
     high="high_quality",
@@ -51,5 +51,5 @@ timeseries_presets_aliases = dict(
 # update with aliases
 TIMESERIES_PRESETS_CONFIGS = {
     **TIMESERIES_PRESETS_CONFIGS,
-    **{k: TIMESERIES_PRESETS_CONFIGS[v].copy() for k, v in timeseries_presets_aliases.items()},
+    **{k: TIMESERIES_PRESETS_CONFIGS[v].copy() for k, v in TIMESERIES_PRESETS_ALIASES.items()},
 }
--- a/timeseries/src/autogluon/timeseries/configs/presets_configs.py
+++ b/timeseries/src/autogluon/timeseries/configs/presets_configs.py
@@ -10,26 +10,19 @@ TIMESERIES_PRESETS_CONFIGS = dict(
     medium_quality={"hyperparameters": "light"},
     fast_training={"hyperparameters": "very_light"},
     chronos_tiny={
-        "hyperparameters": {"Chronos": {"model_path": "amazon/chronos-t5-tiny", "skip_validation": True}},
-        "disable_length_check": True,
+        "hyperparameters": {"Chronos": {"model_path": "amazon/chronos-t5-tiny"}},
     },
     chronos_mini={
-        "hyperparameters": {"Chronos": {"model_path": "amazon/chronos-t5-mini", "skip_validation": True}},
-        "disable_length_check": True,
+        "hyperparameters": {"Chronos": {"model_path": "amazon/chronos-t5-mini"}},
     },
     chronos_small={
-        "hyperparameters": {"Chronos": {"model_path": "amazon/chronos-t5-small", "skip_validation": True}},
-        "disable_length_check": True,
+        "hyperparameters": {"Chronos": {"model_path": "amazon/chronos-t5-small"}},
     },
     chronos_base={
-        "hyperparameters": {"Chronos": {"model_path": "amazon/chronos-t5-base", "skip_validation": True}},
-        "disable_length_check": True,
+        "hyperparameters": {"Chronos": {"model_path": "amazon/chronos-t5-base"}},
     },
     chronos_large={
-        "hyperparameters": {
-            "Chronos": {"model_path": "amazon/chronos-t5-large", "skip_validation": True, "batch_size": 8}
-        },
-        "disable_length_check": True,
+        "hyperparameters": {"Chronos": {"model_path": "amazon/chronos-t5-large", "batch_size": 8}},
     },
     chronos_ensemble={
         "hyperparameters": {
--- a/timeseries/src/autogluon/timeseries/configs/presets_configs.py
+++ b/timeseries/src/autogluon/timeseries/configs/presets_configs.py
@@ -31,17 +31,32 @@ TIMESERIES_PRESETS_CONFIGS = dict(
         },
         "disable_length_check": True,
     },
-    chronos_mini_ensemble={
-        "hyperparameters": {"Chronos": {"model_path": "amazon/chronos-t5-mini"}} | get_default_hps("default"),
-    },
-    chronos_small_ensemble={
-        "hyperparameters": {"Chronos": {"model_path": "amazon/chronos-t5-small"}} | get_default_hps("default"),
-    },
-    chronos_base_ensemble={
-        "hyperparameters": {"Chronos": {"model_path": "amazon/chronos-t5-base"}} | get_default_hps("default"),
+    chronos_ensemble={
+        "hyperparameters": {
+            **{"Chronos": {"model_path": "amazon/chronos-t5-small"}},
+            **get_default_hps("default"),
+        }
     },
     chronos_large_ensemble={
-        "hyperparameters": {"Chronos": {"model_path": "amazon/chronos-t5-large", "batch_size": 8}}
-        | get_default_hps("default"),
+        "hyperparameters": {
+            **{"Chronos": {"model_path": "amazon/chronos-t5-large", "batch_size": 8}},
+            **get_default_hps("default"),
+        }
     },
 )
+
+timeseries_presets_aliases = dict(
+    chronos="chronos_small",
+    best="best_quality",
+    high="high_quality",
+    medium="medium_quality",
+    bq="best_quality",
+    hq="high_quality",
+    mq="medium_quality",
+)
+
+# update with aliases
+TIMESERIES_PRESETS_CONFIGS = {
+    **TIMESERIES_PRESETS_CONFIGS,
+    **{k: TIMESERIES_PRESETS_CONFIGS[v].copy() for k, v in timeseries_presets_aliases.items()},
+}
--- a/timeseries/src/autogluon/timeseries/models/chronos/chronos.py
+++ b/timeseries/src/autogluon/timeseries/models/chronos/chronos.py
@@ -1,0 +1,402 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# Original Source: https://github.com/amazon-science/chronos-forecasting
+# Author: Lorenzo Stella <stellalo@amazon.com>
+
+import warnings
+from dataclasses import dataclass
+from typing import Any, Dict, List, Literal, Optional, Tuple, Union
+
+import torch
+import torch.nn as nn
+from transformers import AutoConfig, AutoModelForCausalLM, AutoModelForSeq2SeqLM, GenerationConfig, PreTrainedModel
+
+
+@dataclass
+class ChronosConfig:
+    """
+    This class holds all the configuration parameters to be used
+    by ``ChronosTokenizer`` and ``ChronosPretrainedModel``.
+    """
+
+    tokenizer_class: str
+    tokenizer_kwargs: Dict[str, Any]
+    n_tokens: int
+    n_special_tokens: int
+    pad_token_id: int
+    eos_token_id: int
+    use_eos_token: bool
+    model_type: Literal["causal", "seq2seq"]
+    context_length: int
+    prediction_length: int
+    num_samples: int
+    temperature: float
+    top_k: int
+    top_p: float
+
+    def __post_init__(self):
+        assert (
+            self.pad_token_id < self.n_special_tokens and self.eos_token_id < self.n_special_tokens
+        ), f"Special token id's must be smaller than {self.n_special_tokens=}"
+
+    def create_tokenizer(self) -> "ChronosTokenizer":
+        if self.tokenizer_class == "MeanScaleUniformBins":
+            return MeanScaleUniformBins(**self.tokenizer_kwargs, config=self)
+        raise ValueError
+
+
+class ChronosTokenizer:
+    """
+    A ``ChronosTokenizer`` defines how time series are mapped into token IDs
+    and back.
+
+    For details, see the ``input_transform`` and ``output_transform`` methods,
+    which concrete classes must implement.
+    """
+
+    def input_transform(self, context: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor, Any]:
+        """
+        Turn a batch of time series into token IDs, attention map, and scale.
+
+        Parameters
+        ----------
+        context
+            A tensor shaped (batch_size, time_length), containing the
+            timeseries to forecast. Use left-padding with ``torch.nan``
+            to align time series of different lengths.
+
+        Returns
+        -------
+        token_ids
+            A tensor of integers, shaped (batch_size, time_length + 1)
+            if ``config.use_eos_token`` and (batch_size, time_length)
+            otherwise, containing token IDs for the input series.
+        attention_mask
+            A boolean tensor, same shape as ``token_ids``, indicating
+            which input observations are not ``torch.nan`` (i.e. not
+            missing nor padding).
+        decoding_context
+            An object that will be passed to ``output_transform``.
+            Contains the relevant context to decode output samples into
+            real values, such as location and scale parameters.
+        """
+        raise NotImplementedError()
+
+    def output_transform(self, samples: torch.Tensor, decoding_context: Any) -> torch.Tensor:
+        """
+        Turn a batch of sample token IDs into real values.
+
+        Parameters
+        ----------
+        samples
+            A tensor of integers, shaped (batch_size, num_samples, time_length),
+            containing token IDs of sample trajectories.
+        decoding_context
+            An object returned by ``input_transform`` containing
+            relevant context to decode samples, such as location and scale.
+            The nature of this depends on the specific tokenizer.
+
+        Returns
+        -------
+        forecasts
+            A real tensor, shaped (batch_size, num_samples, time_length),
+            containing forecasted sample paths.
+        """
+        raise NotImplementedError()
+
+
+class MeanScaleUniformBins(ChronosTokenizer):
+    def __init__(self, low_limit: float, high_limit: float, config: ChronosConfig) -> None:
+        self.config = config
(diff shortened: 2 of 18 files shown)


timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -0,0 +1,366 @@
...
+    Other Parameters
+    ----------------
+    model_path: str, default = "amazon/chronos-t5-small"
+        Model path used for the model, i.e., a HuggingFace transformers `name_or_path`. Can be a
+        compatible model name on HuggingFace Hub or a local path to a model directory.
+    batch_size : int, default = 16
+        Size of batches used during inference
+    num_samples : int, default = 20
+        Number of samples used during inference
+    skip_validation : bool, default = False
```
shchur, line 179: I think this logic should be absorbed into a single boolean parameter that we pass to `TimeSeriesPredictor.fit` (currently `disable_length_check`). Otherwise, the user could write
```python
predictor.fit(data, hyperparameters={"Chronos": {"skip_validation": True}, "Naive": {}})
```
which would break our model selection logic.

The interpretation of the parameter passed to `TimeSeriesPredictor.fit` (I will call it `skip_validation`, but can be a different name) would be "I only want to train a single model and use it for prediction. I don't want to perform model selection, so I don't need to compute the validation score". If the user sets `predictor.fit(..., skip_validation=True)` but also provides >= 2 models via `hyperparameters` or provides a search space for any model, then an exception will be raised.

We could also split this functionality into a separate PR, and limit this PR to adding the Chronos model (that works like any other forecasting model).

What do you think?

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -0,0 +1,366 @@
...
+        Optimization strategy to use for inference on CPUs. If None, the model will use the default implementation.
+        If `onnx`, the model will be converted to ONNX and the inference will be performed using ONNX. If `ovm`,
+        inference will be performed with the model compiled to OpenVINO.
+    """
+
+    # default number of samples for prediction
+    default_num_samples: int = 20
+    default_batch_size: int = 16
+    default_context_length: int = 512
+    bucket_name = "chronos-release"
```
shchur, line 194: Redundant parameter

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -0,0 +1,366 @@
...
+    ):
+        hyperparameters = hyperparameters if hyperparameters is not None else {}
+        self.batch_size = hyperparameters.get("batch_size", self.default_batch_size)
+        self.num_samples = hyperparameters.get("num_samples", self.default_num_samples)
+        self.model_path = hyperparameters.get("model_path", self.default_model_path)
+        self.skip_validation = hyperparameters.get("skip_validation", False)
+        self.device = hyperparameters.get("device", None)
+        self.optimization_strategy: Optional[Literal["onnx", "ovm"]] = hyperparameters.get(
+            "optimization_strategy", None
+        )
```
shchur, line 215: Nit: Replace `get(x, None)` with just `get(x)`?

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -0,0 +1,366 @@
...
+            hyperparameters=hyperparameters,
+            **kwargs,
+        )
+
+        self.model_pipeline: Optional[OptimizedChronosPipeline] = None
+
+    def save(self, path: str = None, verbose: bool = True) -> str:
+        pipeline = self.model_pipeline
+        self.model_pipeline = None
+        path = Path(super().save(path=path, verbose=verbose))
```
shchur, line 236: Nit: No need to convert to `Path` if we return a `str` anyway

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -0,0 +1,366 @@
...
+            self._oof_predictions = [None]
+            self.val_score = None
+            self.predict_time = None
+        else:
+            super().score_and_cache_oof(val_data, store_val_score, store_predict_time)
+
+    def _is_gpu_available(self) -> bool:
+        import torch.cuda
+
+        return torch.cuda.is_available()
```
shchur, line 266: Let's open an issue to track various Chronos features to integrate into AG? Some ideas:
- Zero-shot inference (this PR)
- Automatic batch size selection
- Multi-GPU inference
- Fine-tuning

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -0,0 +1,366 @@
...
+        minimum_resources = {"num_cpus": 1}
+        # if GPU is available, we train with 1 GPU per trial
+        if is_gpu_available:
+            minimum_resources["num_gpus"] = MODEL_CONFIGS[self.model_path].get("num_gpus", 0)
+        return minimum_resources
+
+    def _get_model_pipeline(self):
+        gpu_available = self._is_gpu_available()
+
+        assert self.model_path in MODEL_CONFIGS or Path.is_dir(
```
shchur, line 278: Do I understand correctly that we only support loading weights from 1/ official Amazon HF repos or 2/ local directory?

What if we allow arbitrary HF paths using logic similar to https://github.com/allenai/cached_path?tab=readme-ov-file#usage? 
- `hf://model/path` - HF repo
- `file://model/path` or `model/path` - local path
- `small` / `tiny` / `mini` / `base` / `large` - automatically resolves to the official Amazon weights on HF?

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -0,0 +1,366 @@
...
+        time_limit: int = None,
+        **kwargs,
+    ) -> None:
+        self._check_fit_params()
+        self.model_pipeline = self._get_model_pipeline()
+
+    def get_inference_data_loader(
+        self,
+        data: TimeSeriesDataFrame,
+        num_workers: int = 1,
```
shchur, line 312: Should this be exposed as a model hyperparameter?

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -0,0 +1,366 @@
...
+                "`import torch; torch.cuda.is_available()` returns `True`."
+            )
+
+        device = self.device or ("cuda" if gpu_available else "auto")
+
+        pipeline = OptimizedChronosPipeline.from_pretrained(
+            self.model_path,
+            device_map=device,
+            optimization_strategy=self.optimization_strategy,
+        )
```
shchur, line 295: Should we also expose the `torch_dtype` argument as a hyperparameter?

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -0,0 +1,366 @@
...
+        self.model_pipeline = None
+        path = Path(super().save(path=path, verbose=verbose))
+        self.model_pipeline = pipeline
+
+        return str(path)
+
+    @classmethod
+    def load(cls, path: str, reset_paths: bool = True, verbose: bool = True) -> "ChronosModel":
+        model = load_pkl.load(path=os.path.join(path, cls.model_file_name), verbose=verbose)
+        model.model_pipeline = model._get_model_pipeline()
```
shchur, line 244: Just to check my understanding, will this load the weights to GPU memory immediately? Or are they only loaded at prediction time?

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -0,0 +1,366 @@
...
+                    num_samples=self.num_samples,
+                    limit_prediction_length=False,
+                )
+                .detach()
+                .cpu()
+                .numpy()
+                for batch in self.get_inference_data_loader(data=data)
+            )
+
+        samples = np.concatenate([c.T for c in chain.from_iterable(prediction_samples)], axis=0)
```
shchur, line 355: If each entry in `prediction_samples` is an array of shape `[batch_size, prediction_length, num_samples]`, could we get the same result with `np.concatenate(prediction_samples, axis=0)`?

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -0,0 +1,366 @@
...
+    """A container for time series datasets that implements the ``torch.utils.data.Dataset`` interface"""
+
+    def __init__(
+        self,
+        target_df: TimeSeriesDataFrame,
+        context_length: int,
+        target_column: str = "target",
+    ):
+        assert target_df is not None
+        assert target_df.freq, "Initializing GluonTS data sets without freq is not allowed"
```
shchur, line 46: Are these checks necessary?

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -0,0 +1,366 @@
...
+        """Replace inefficient groupby ITEMID with indptr that stores start:end of each time series"""
+
+        # TODO: refactor into generic iterator
+
+        item_id_index = target_df.index.get_level_values(ITEMID)
+        indices_sizes = item_id_index.value_counts(sort=False)
+        self.item_ids = indices_sizes.index  # shape [num_items]
+        cum_sizes = indices_sizes.values.cumsum()
+        self.indptr = np.append(0, cum_sizes).astype(np.int32)
+        self.start_timestamps = target_df.reset_index(TIMESTAMP).groupby(level=ITEMID, sort=False).first()[TIMESTAMP]
```
shchur, line 64: Do we need to store/return `item_ids` and `start_timestamps`?

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -0,0 +1,366 @@
...
+        assert len(self.item_ids) == len(self.start_timestamps)
+
+    def __len__(self):
+        return len(self.indptr) - 1  # noqa
+
+    def _get_context(self, a: np.ndarray, pad_value=np.nan):
+        assert self.context_length > 0
```
shchur, line 71: Nit: Should this happen in `__init__`?

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -0,0 +1,366 @@
...
+        a = a[-self.context_length :]
+        pad_size = self.context_length - len(a)
+        if pad_size > 0:
+            pad = np.full(shape=(pad_size,), fill_value=pad_value)
+            a = np.concatenate((pad, a))
+        return a
+
+    def get_full_item(self, idx) -> Dict[str, Any]:
```
shchur, line 79: This method seems to be unused

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -0,0 +1,366 @@
...
+        self,
+        target_df: TimeSeriesDataFrame,
+        context_length: int,
+        target_column: str = "target",
+    ):
+        assert target_df is not None
+        assert target_df.freq, "Initializing GluonTS data sets without freq is not allowed"
+
+        self.target_column = target_column
+        self.context_length = context_length
```
shchur, line 49: Nit: A potential optimization to save GPU memory is to set `context_length = min(context_length, min(indices_sizes))` (i.e., no longer than the shortest TS in the dataset)

shchur (commented):
Looks great, and thanks for the amazing CPU inference speed optimizations! 

My main question is related to merging `skip_validation` and `disable_length_check` kwargs, the rest are minor comments.

timeseries/src/autogluon/timeseries/models/chronos/model.py
canerturkmen: Here for completeness. We will likely have to extend this when covariates come into play.

timeseries/src/autogluon/timeseries/models/chronos/model.py
canerturkmen: will do. thanks!!

timeseries/src/autogluon/timeseries/models/chronos/model.py
canerturkmen: It loads them immediately.

timeseries/src/autogluon/timeseries/models/chronos/model.py
canerturkmen: Do you mean longest TS in the dataset? 

So in the current construction, by time the data is provided the model will already be initialized. i.e., as above, we're not doing deferred initialization for the model.

timeseries/src/autogluon/timeseries/models/chronos/model.py
canerturkmen: `chain.from_iterable` processes item by item in each batch, thus reshaping as well. The alternative is something along the lines of 
```
X2 = np.concatenate(list(pred_samples), axis=0)
res2 = X2.swapaxes(0, 1).reshape(X2.shape[1], -1)
```

for which I did not see a consistent speedup.

timeseries/src/autogluon/timeseries/models/chronos/model.py
canerturkmen: why not? (i.e., I would like to keep `get_full_item` method as well.

shchur: My reasoning was that 1/ Chronos won't use timestamp/item_id information until at least AG v1.2 and 2/ even when it does, it won't necessarily expect the GluonTS dataset format with start+freq fields (e.g., passing the timestamps as `pd.DatetimeIndex` might be more efficient). Therefore, these variables as well as `get_full_item` method would be dead code until v1.2. If we need this logic in the future, we can copy it from `SimpleGluonTSDataset`.

timeseries/src/autogluon/timeseries/models/chronos/model.py
canerturkmen: I agree to separating to a different PR.

timeseries/src/autogluon/timeseries/models/chronos/model.py
shchur: You are right, sorry, I meant the longest TS in the dataset.

As far as I understood, we are instantiating a new `ChronosInferenceDataset` every time `model.predict(data)` is called. I meant `context_length` as a property of this `ChronosInferenceDataset` object that affects how much padding we add (or how short we cut all the time series). If user calls `predict` with `data` where all TS have length 100, the current code will pad them all to 512, which would increase the memory consumption. If we set `context_length = min(context_length, max(indices_sizes))`, then `context_length` will be automatically shortened to 100.

This might be a premature optimization that is not needed, so feel free to ignore my comment.

timeseries/src/autogluon/timeseries/models/chronos/model.py
canerturkmen: We can definitely implement the shorthands. I'm not sure about the value the other formats bring. Are you also suggesting that we use the `cached_path` library? (i.e., HF already provides filesystem caching out of the box)

shchur: I didn't mean using the `cached_path` library, rather using the `hf://` and `file://` prefixes to differentiate between huggingface & local paths

timeseries/src/autogluon/timeseries/models/chronos/model.py
shchur: Okay, so just to make sure that we're on the same page, you will remove the `skip_validation` and `disable_length_check` arguments in this PR?

yinweisu:
| :---: | :---: |

timeseries/src/autogluon/timeseries/models/chronos/model.py
canerturkmen: Right now HuggingFace logic seems to be to look for a local path first. If there's a clash between hf_hub and local path, then local path is preferred. I think this is highly unlikely, and would like to keep the current implementation.

timeseries/src/autogluon/timeseries/models/chronos/model.py
canerturkmen: Included this logic. Looks a bit shaky to me in hindsight, it might use some @shchur reviewing 👀 .

yinweisu:
| :---: | :---: |
| botocore==1.34.65 | botocore==1.34.66 |
| boto3==1.34.65 | boto3==1.34.66 |
| botocore==1.34.65 | botocore==1.34.66 |
| boto3==1.34.65 | boto3==1.34.66 |

timeseries/src/autogluon/timeseries/configs/presets_configs.py
```diff
@@ -8,4 +9,47 @@
...
+        "hyperparameters": {"Chronos": {"model_path": "base"}},
+    },
+    chronos_large={
+        "hyperparameters": {"Chronos": {"model_path": "large", "batch_size": 8}},
+    },
+    chronos_ensemble={
+        "hyperparameters": {
+            **{"Chronos": {"model_path": "small"}},
+            **get_default_hps("default"),
+        }
```
shchur, line 31: Nit: the following would also work
```python
        "hyperparameters": {
            "Chronos": {"model_path": "small"}, **get_default_hps("default")
        }

```

timeseries/src/autogluon/timeseries/configs/presets_configs.py
```diff
@@ -8,4 +9,47 @@
...
+    bq="best_quality",
+    hq="high_quality",
+    mq="medium_quality",
+)
+
+# update with aliases
+TIMESERIES_PRESETS_CONFIGS = {
+    **TIMESERIES_PRESETS_CONFIGS,
+    **{k: TIMESERIES_PRESETS_CONFIGS[v].copy() for k, v in timeseries_presets_aliases.items()},
+}
```
shchur, line 55: Nice!

timeseries/src/autogluon/timeseries/models/chronos/chronos.py
```diff
@@ -0,0 +1,470 @@
...
+
+        if prediction_length > self.model.config.prediction_length:
+            msg = (
+                f"We recommend keeping prediction length <= {self.model.config.prediction_length}. "
+                f"The quality of longer predictions may degrade since the model is not optimized for it. "
+            )
+            if limit_prediction_length:
+                msg += "You can turn off this check by setting `limit_prediction_length=False`."
+                raise ValueError(msg)
+            warnings.warn(msg)
```
shchur, line 354: Nit: I think we should set `warnings.warn(msg, stacklevel=2)` for an informative warning?

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -0,0 +1,319 @@
...
+    ):
+        assert context_length > 0
+        self.target_column = target_column
+        self.context_length = context_length
+        self.target_array = target_df[target_column].to_numpy(dtype=np.float32)
+        self.freq = target_df.freq
+
+        # store pointer to start:end of each time series
+        indices_sizes = target_df.index.get_level_values(ITEMID).value_counts(sort=False)
+        cum_sizes = indices_sizes.values.cumsum()
```
shchur, line 55: A slightly less verbose alternative is 
```python
cum_sizes = target_df.num_timesteps_per_item().values.cumsum()
```

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -0,0 +1,319 @@
...
+    """A container for time series datasets that implements the ``torch.utils.data.Dataset`` interface"""
+
+    def __init__(
+        self,
+        target_df: TimeSeriesDataFrame,
+        context_length: int,
+        target_column: str = "target",
+    ):
+        assert context_length > 0
+        self.target_column = target_column
```
shchur, line 48: Unused variable?

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -0,0 +1,319 @@
...
+            model.set_contexts(path)
+        return model
+
+    def score_and_cache_oof(
+        self,
+        val_data: TimeSeriesDataFrame,
+        store_val_score: bool = False,
+        store_predict_time: bool = False,
+    ) -> None:
+        super().score_and_cache_oof(val_data, store_val_score, store_predict_time)
```
shchur, line 200: Method can be removed since we anyway use the parent implementation

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -0,0 +1,319 @@
...
+
+    def _is_gpu_available(self) -> bool:
```
shchur, line 202: Nit: Should we move this method to `AbstractTimeSeriesModel` since multiple models rely on this method (also `AbstractGluonTSModel`)?

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -0,0 +1,319 @@
...
+
+    def load_model_pipeline(self, context_length: Optional[int] = None):
+        from .chronos import OptimizedChronosPipeline
+
+        gpu_available = self._is_gpu_available()
+
+        if not gpu_available and MODEL_CONFIGS.get(self.model_path, {}).get("num_gpus", 0) > 0:
+            raise RuntimeError(
+                f"{self.name} requires a GPU to run, but no GPU was detected. "
+                "Please make sure that you are using a GPU instance and "
```
shchur, line 222: Nit: `GPU machine`?

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -0,0 +1,319 @@
...
+    ) -> TimeSeriesDataFrame:
+        import torch
+
+        # We defer initialization of the model pipeline. i.e., the model is only loaded to device memory
+        # during inference. We also infer the maximum length of the time series in the inference data set
+        # and use that to determine the context length of the model. If the context length is specified
+        # during initialization, this is always used. If not, the context length is set to the longest
+        # item length. The context length is always capped by self.maximum_context_length.
+        context_length = self.context_length or min(
+            data.index.get_level_values(ITEMID).value_counts(sort=False).max(),
```
shchur, line 282: Similar to before, `data.num_timesteps_per_items().max()`

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -496,10 +496,19 @@ def fit(
...
+              and ``DirectTabular``. These models are fast to train but may not be very accurate.
+            - ``"medium_quality"``: all models mentioned above + deep learning model ``TemporalFusionTransformer``. Default setting that produces good forecasts
+              with reasonable training time.
+            - ``"high_quality"``: All ML models available in AutoGluon + additional statistical models (``NPTS``, ``AutoETS``, ``AutoARIMA``, ``CrostonSBA``,
+              ``DynamicOptimizedTheta``). Much more accurate than ``medium_quality``, but takes longer to train.
             - ``"best_quality"``: Same models as in ``"high_quality"`, but performs validation with multiple backtests. Usually better than ``high_quality``, but takes even longer to train.
+            - ``"chronos_{model_size}"``: where model size is one of ``tiny,mini,small,base,large``. Uses the Chronos pretrained model for zero-shot forecasting.
+              See the documentation for ``ChronosModel`` or see `Hugging Face <https://huggingface.co/collections/amazon/chronos-models-65f1791d630a8d57cb718444>`_ for more information.
```
shchur, line 507: TIL that Hugging Face is officially spelled with a space 😮

timeseries/tests/unittests/models/test_chronos.py
```diff
@@ -0,0 +1,284 @@
...
+    ],
+)
+def default_chronos_tiny_model(request) -> ChronosModel:
+    model = ChronosModel(
+        hyperparameters={
+            "model_path": "amazon/chronos-t5-tiny",
+            "num_samples": 3,
+            "context_length": 16,
+            "device": "cpu",
+            **request.param,
```
shchur, line 59: Nice!

timeseries/tests/unittests/models/test_chronos.py
```diff
@@ -0,0 +1,284 @@
...
+def test_when_on_cpu_then_chronos_model_can_score_and_cache_oof(data, default_chronos_tiny_model):
+    default_chronos_tiny_model.score_and_cache_oof(data)
+    assert default_chronos_tiny_model._oof_predictions is not None
+
+
+@pytest.mark.parametrize("data", DATASETS)
+def test_when_on_cpu_then_chronos_model_can_infer(data, default_chronos_tiny_model):
+    predictions = default_chronos_tiny_model.predict(data)
+    assert all(
+        predictions.index.get_level_values("item_id").unique() == data.index.get_level_values("item_id").unique()
```
shchur, line 92: Nit: `predictions.item_ids` and `data.item_ids` accomplishes the same

timeseries/tests/unittests/models/test_models.py
```diff
@@ -147,11 +153,7 @@ def test_given_hyperparameter_spaces_when_tune_called_then_tuning_output_correct
         path=temp_model_path,
         freq="H",
         quantile_levels=[0.1, 0.9],
-        hyperparameters={
-            "epochs": space.Int(1, 3),
-            "num_batches_per_epoch": 1,
-            "use_fallback_model": False,
-        },
+        hyperparameters=DUMMY_HYPERPARAMETERS | {"epochs": space.Int(1, 3)},
```
shchur, line 156: This won't work on Python 3.8 during platform tests

shchur (commented):
Mostly nitpicks from my side, I think we are almost there. Also the tests are a work of art 💪

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -496,10 +496,19 @@ def fit(
...
+            - ``"fast_training"``: fit simple statistical models (``ETS``, ``Theta``, ``Naive``, ``SeasonalNaive``) + fast tree-based models ``RecursiveTabular``
+              and ``DirectTabular``. These models are fast to train but may not be very accurate.
+            - ``"medium_quality"``: all models mentioned above + deep learning model ``TemporalFusionTransformer``. Default setting that produces good forecasts
+              with reasonable training time.
+            - ``"high_quality"``: All ML models available in AutoGluon + additional statistical models (``NPTS``, ``AutoETS``, ``AutoARIMA``, ``CrostonSBA``,
+              ``DynamicOptimizedTheta``). Much more accurate than ``medium_quality``, but takes longer to train.
             - ``"best_quality"``: Same models as in ``"high_quality"`, but performs validation with multiple backtests. Usually better than ``high_quality``, but takes even longer to train.
+            - ``"chronos_{model_size}"``: where model size is one of ``tiny,mini,small,base,large``. Uses the Chronos pretrained model for zero-shot forecasting.
```
shchur, line 506: Add a "section header" for Chronos presets? Something like
```
...
- ``"best_quality"``: Same models as in ``"high_quality"`, but performs validation with multiple backtests. Usually better than ``high_quality``, but takes even longer to train.

Presets involving Chronos model
- ``"chronos_{model_size}"``: ...
```

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -496,10 +496,19 @@ def fit(
...
+              See the documentation for ``ChronosModel`` or see `Hugging Face <https://huggingface.co/collections/amazon/chronos-models-65f1791d630a8d57cb718444>`_ for more information.
+              Note that a GPU is required for model sizes ``small``, ``base`` and ``large``.
+            - ``"chronos"``: alias for ``"chronos_small"``.
+            - ``"chronos_ensemble"``: builds an ensemble of the models specified in ``"high_quality"`` and ``"chronos_small"``.
+            - ``"chronos_large_ensemble"``: builds an ensemble of the models specified in ``"high_quality"`` and ``"chronos_large"``.
```
shchur, line 511: Merge these two into a single bullet point?

timeseries/src/autogluon/timeseries/models/chronos/model.py
canerturkmen: It's already there, returning `False` by default. I guess the idea was to not make mention of torch if it's not needed.

yinweisu:
| :---: | :---: |

yinweisu:
| :---: | :---: |

timeseries/src/autogluon/timeseries/models/chronos/model.py
shchur: I see. I guess that in the ideal world it would be fine to use the `torch` import inside this method since only models that support GPU will ever call`_is_gpu_available()`. However, because of the idiosyncrasies of the HPO logic with ray this method is actually called by all models. I'm fine keeping it as is, or maybe adding a TODO comment if you feel like it (to remove `_is_gpu_available` from HPO logic and use a tag indicating if model supports GPU instead).

yinweisu:
| :---: | :---: |
| graphviz==0.20.2 | graphviz==0.20.3 |
| graphviz==0.20.2 | graphviz==0.20.3 |

yinweisu:
| :---: | :---: |

yinweisu:
| :---: | :---: |

yinweisu:
| :---: | :---: |

canerturkmen:
@shchur Ready for your final review! 🚀

yinweisu:
| :---: | :---: |
| typer==0.9.0 | typer==0.9.4 |
| typer==0.9.0 | typer==0.9.4 |

shchur (approved):
LGTM! 🚀

PR #3979, opened 2024-03-13
[timeseries] Fix path and device bugs

shchur:
*Description of changes:*
- Fix a bug, where GluonTS models trained on GPU could not be loaded on a CPU-only machine
- Fix a bug, where passing path to data as a `str` raised an exception
  - Fix the respective test that was supposed to catch this bug
- Allow passing the path to data as a `pathlib.Path` when using `TimeSeriesPredictor`

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
+++ b/timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
@@ -204,7 +204,7 @@ class AbstractGluonTSModel(AbstractTimeSeriesModel):
             model = load_pkl.load(path=os.path.join(path, cls.model_file_name), verbose=verbose)
             if reset_paths:
                 model.set_contexts(path)
-            model.gts_predictor = PyTorchPredictor.deserialize(Path(path) / cls.gluonts_model_path)
+            model.gts_predictor = PyTorchPredictor.deserialize(Path(path) / cls.gluonts_model_path, device="auto")
         return model
 
     def _get_hpo_backend(self):
--- a/timeseries/src/autogluon/timeseries/predictor.py
+++ b/timeseries/src/autogluon/timeseries/predictor.py
@@ -252,12 +252,12 @@ class TimeSeriesPredictor(TimeSeriesPredictorDeprecatedMixin):
 
     def _to_data_frame(
         self,
-        data: Union[TimeSeriesDataFrame, pd.DataFrame, str],
+        data: Union[TimeSeriesDataFrame, pd.DataFrame, Path, str],
         name: str = "data",
     ) -> "TimeSeriesDataFrame":
         if isinstance(data, TimeSeriesDataFrame):
             return data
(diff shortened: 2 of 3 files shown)


timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
```diff
@@ -204,7 +204,7 @@ def load(cls, path: str, reset_paths: bool = True, verbose: bool = True) -> "Abs
             model = load_pkl.load(path=os.path.join(path, cls.model_file_name), verbose=verbose)
             if reset_paths:
                 model.set_contexts(path)
-            model.gts_predictor = PyTorchPredictor.deserialize(Path(path) / cls.gluonts_model_path)
+            model.gts_predictor = PyTorchPredictor.deserialize(Path(path) / cls.gluonts_model_path, device="auto")
```
shchur, line 207: I couldn't figure out a way to test that this works. Modifying `CUDA_VISIBLE_DEVICES` after torch has been imported has no effect, so I couldn't disable CUDA inside the tests. Do you think it's fine if we leave this untested?

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -1195,8 +1196,8 @@ def select_target(ts_df: TimeSeriesDataFrame) -> TimeSeriesDataFrame:
 
     def plot(
         self,
-        data: Union[TimeSeriesDataFrame, pd.DataFrame, str],
-        predictions: Optional[Union[TimeSeriesDataFrame, pd.DataFrame, str]] = None,
+        data: Union[TimeSeriesDataFrame, pd.DataFrame, Path, str],
+        predictions: Optional[Union[TimeSeriesDataFrame, pd.DataFrame, Path, str]] = None,
```
canerturkmen, line 1200: Is this being handled?

timeseries/tests/unittests/test_predictor.py
```diff
@@ -522,15 +522,17 @@ def test_given_data_is_in_dataframe_format_then_predictor_works(temp_model_path)
     assert isinstance(predictions, TimeSeriesDataFrame)
 
 
-def test_given_data_is_in_str_format_then_predictor_works(temp_model_path):
+@pytest.mark.parametrize("path_format", [str, Path])
```
canerturkmen, line 525: nice

canerturkmen (commented):
Thanks! One comment.

timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
canerturkmen: Assuming you've tested offline, sure.

timeseries/src/autogluon/timeseries/predictor.py
shchur: Good catch, thanks! Fixing this now

canerturkmen (approved):
LGTM. Thanks!

PR #3984, opened 2024-03-15
[core] move transformers to setup_utils, bump dependency version

canerturkmen:
*Issue #, if available:*

*Description of changes:*

Chronos implementation (cf #3978) requires `transformers>=4.36`. This PR moves transformers and accelerate version ranges to `_setup_utils.py` and bumps the version. 

### DIFF
--- a/core/src/autogluon/core/_setup_utils.py
+++ b/core/src/autogluon/core/_setup_utils.py
@@ -30,6 +30,8 @@ DEPENDENT_PACKAGES = {
     "lightning": ">=2.1,<2.2",  # Major version cap
     "pytorch_lightning": ">=2.1,<2.2",  # Major version cap, capping `lightning` does not cap `pytorch_lightning`!
     "async_timeout": ">=4.0,<5",  # Major version cap
+    "transformers[sentencepiece]": ">=4.36.0,<4.39.0",
+    "accelerate": ">=0.21.0,<0.22.0",
 }
 if LITE_MODE:
     DEPENDENT_PACKAGES = {package: version for package, version in DEPENDENT_PACKAGES.items() if package not in ["psutil", "Pillow", "timm"]}
--- a/core/src/autogluon/core/_setup_utils.py
+++ b/core/src/autogluon/core/_setup_utils.py
@@ -30,7 +30,7 @@ DEPENDENT_PACKAGES = {
     "lightning": ">=2.1,<2.2",  # Major version cap
     "pytorch_lightning": ">=2.1,<2.2",  # Major version cap, capping `lightning` does not cap `pytorch_lightning`!
     "async_timeout": ">=4.0,<5",  # Major version cap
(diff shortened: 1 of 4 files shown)


canerturkmen:
In Chronos, we would like to depend on `optimum-intel` for fast CPU inference, and this requires `transformers>=4.36`. We can ship without `optimum-intel` support too, but this results in 2-3x slower inference on some models. 

Currently, bumping the `transformers` version appears to be breaking a specific unit test. If this is easy to mitigate for MM, then we can include that fix here or in another PR to enable Chronos.

Looking forward to your comments.

tonyhoo:
@prateekdesai04  we should test this change with torch update

timeseries/setup.py
```diff
@@ -55,7 +57,9 @@
     ],
 }
 
-extras_require["all"] = []
+extras_require["all"] = [
+    "optimum[openvino,onnxruntime,nncf]>=1.17,<1.18",  # for faster CPU inference in pretrained models
```
tonyhoo, line 61: Do we need [neuron suppport](https://huggingface.co/docs/optimum-neuron/index) to speed up inference on Inf2 architecture?

prateekdesai04:
> @prateekdesai04 we should test this change with torch update

@tonyhoo Just pushed the latest Torch + CUDA image, running this PR https://github.com/autogluon/autogluon/pull/3982
Once successful we can merge it and then run Caner's PR with the latest changes

timeseries/setup.py
canerturkmen: Not at the moment, though this will come in v1.2.

canerturkmen:
Converted to draft, pending the PyTorch upgrade.

yinweisu:
| :---: | :---: |

yinweisu:
| :---: | :---: |

canerturkmen:
@zhiqiangdon for your review of test fails

core/src/autogluon/core/_setup_utils.py
```diff
@@ -30,6 +30,8 @@
     "lightning": ">=2.1,<2.2",  # Major version cap
     "pytorch_lightning": ">=2.1,<2.2",  # Major version cap, capping `lightning` does not cap `pytorch_lightning`!
     "async_timeout": ">=4.0,<5",  # Major version cap
+    "transformers[sentencepiece]": ">=4.36.0,<4.39.0",
```
zhiqiangdon, line 33: Why do we use 4.36 instead of 4.38 as the lower cap?

canerturkmen: This is the lowest required by optimum for now. Wanted to leave it wide for more flexibility.

yinweisu:
| :---: | :---: |

tonyhoo (approved):

zhiqiangdon (approved):
LGTM

PR #4011, opened 2024-03-27
[DRAFT][CI][Fix] tenssort installation fix, capping pip

prateekdesai04:
*Description of changes:*
Fix for sudden failure in installing tensorrt - `RuntimeError: Didn't find wheel for nvidia-cudnn-cu12 9.0.0.312`
If there happens to be a better fix in the future then there will be a follow-up PR.
Below is the detailed log: https://github.com/autogluon/autogluon/actions/runs/8443664550/job/23127826077?pr=4007

Links to similar issues:
https://github.com/NVIDIA/TensorRT/issues/2933
https://github.com/NVIDIA/TensorRT/issues/3078
https://github.com/NVIDIA/TensorRT/pull/3080


yinweisu:
| :---: | :---: |

.github/workflow_scripts/env_setup.sh
```diff
@@ -1,5 +1,5 @@
 function setup_build_env {
-    python -m pip install --upgrade pip
+    python -m pip install pip==21.3.1 # Capping pip to prevent tensorrt install failure
```
tonyhoo, line 2: Could you please provide some context regarding the need for version pinning here? An 'TODO' comment with additional details would be helpful.

zhiqiangdon, line 2: Looks like the tesorrt installation error is gone. Do we still need this change? Would it affect other packages?

prateekdesai04: We do not need it at the moment.
Ideally it should not affect other packages because it's just pip.
We will treat this PR as a draft PR in case we see this failure in the future.
(As this failure was experienced by a lot of open-source teams outside AutoGluon at some point - in the shared links)

ddelange: the proper fix would be to add a `PIP_EXTRA_INDEX_URL='https://pypi.nvidia.com'` environment variable to your CI environment (see the description [here](https://pypi.org/project/tensorrt/8.6.1.post1/))

for more details, see https://github.com/NVIDIA/TensorRT/pull/3080

ddelange: ah, at closer inspection of the failure, it does not seem to be an issue with tensorrt installation (the pip-inside-pip pattern of the tensorrt setup.py seems to be doing its job).

The flaky error was rather in some other custom wheel download code, namely from `nvidia_stub`:
```pytb
  File "/tmp/pip-build-env-yu1rd47j/overlay/lib/python3.10/site-packages/nvidia_stub/wheel.py", line 146, in download_wheel
    return download_manual(wheel_directory, distribution, version)
  File "/tmp/pip-build-env-yu1rd47j/overlay/lib/python3.10/site-packages/nvidia_stub/wheel.py", line 113, in download_manual
    raise RuntimeError(f"Didn't find wheel for {distribution} {version}")
RuntimeError: Didn't find wheel for nvidia-cudnn-cu12 9.0.0.312
```

That may have been a temporary host resolution error or so. In any case it looks like they're attempting to bypass the standardized install mechanisms provided by PyPA, probably for similar reasons as described in the PR discussion I linked above.

prateekdesai04: Yup, it seemed to be temporary host resolution error.
Will look to add the proper fix if we see this failure again!
Thanks for the solution!

Issue #3924, opened 2024-02-15
Request: Implement Feature Importance Explainability for Time-Series Module


kristinakupf:
### Summary:
The AutoGluon time-series module has proven to be a powerful tool for forecasting tasks. However, one area that could significantly enhance its utility is the inclusion of feature importance explainability in terms of both global training as well as inclusion as covariates, akin to what is currently available in the AutoGluon tabular module. This feature would greatly aid in understanding model decisions, facilitating a more intuitive analysis and improvement of models by highlighting which features contribute most to predictions.

### Detail:
The tabular module in AutoGluon offers an insightful feature importance mechanism that helps users understand the impact of each feature on the model's predictions. This is not only crucial for model interpretation but also for improving model performance by focusing on the most influential features. Implementing a similar feature for the time-series module would provide users with a comprehensive tool for time-series forecasting that is not only powerful but also interpretable.

- Model Transparency: Provides clear insights into how and why predictions are made, increasing trust in the model.

- Feature Engineering: Identifies which features are most valuable, guiding users on where to focus their feature engineering efforts.

- Model Improvement: Helps in diagnosing model performance issues by highlighting features that are less important or potentially noisy.

## Suggested Implementation:
It would be extremely helpful for the time-series module to incorporate a feature importance mechanism. This could potentially leverage some modified version of existing frameworks like SHAP (SHapley Additive exPlanations) or permutation importance, similar to the approach used in the tabular module. 

The addition of feature importance explainability to the AutoGluon time-series module would be a valuable enhancement, making the module not only a powerful forecasting tool but also an interpretable and transparent one. It would align with the growing need for explainable AI in critical applications and facilitate a deeper understanding and trust in AI-driven forecasting models.

Thank you for considering this feature request. I believe it would make a significant contribution to the AutoGluon toolkit and its user community.

731315163:
very good

dshetyo:
Following

PR #4033, opened 2024-04-03
[timeseries] add feature importance to TimeSeriesPredictor

canerturkmen:
*Issue #, if available:*

#3924 

*Description of changes:*

This draft PR introduces the high level design for `TimeSeriesPredictor.feature_importance`. 

Additional to-dos:
- [x] Fix feature transformation logic to prevent shuffling the held-out future
- [x] Add override logic to make sure overall unused features are not computed for and 0 importance is assigned
- [x] Add unit tests
- ~(stretch) Revisit prediction caching logic to prevent models that don't use features from re-inferring~

### DIFF
--- a/timeseries/src/autogluon/timeseries/learner.py
+++ b/timeseries/src/autogluon/timeseries/learner.py
@@ -198,13 +198,82 @@ class TimeSeriesLearner(AbstractLearner):
 
     def evaluate(
         self,
-        data: Union[TimeSeriesDataFrame, pd.DataFrame, str],
+        data: TimeSeriesDataFrame,
         model: Optional[str] = None,
         metrics: Optional[Union[str, TimeSeriesScorer, List[Union[str, TimeSeriesScorer]]]] = None,
         use_cache: bool = True,
     ) -> Dict[str, float]:
         data = self.feature_generator.transform(data)
         return self.load_trainer().evaluate(data=data, model=model, metrics=metrics, use_cache=use_cache)
+
+    def get_feature_importance(
+        self,
+        data: Optional[TimeSeriesDataFrame] = None,
+        model: Optional[str] = None,
+        metric: Optional[Union[str, TimeSeriesScorer]] = None,
+        features: Optional[List[str]] = None,
+        time_limit: Optional[float] = None,
+        method: Literal["naive", "permutation"] = "permutation",
+        subsample_size: int = 50,
+        num_iterations: int = 1,
+        random_seed: Optional[int] = None,
+        relative_scores: bool = False,
+        include_confidence_band: bool = True,
+        confidence_level: float = 0.99,
+    ) -> pd.DataFrame:
+        trainer = self.load_trainer()
+        if data is None:
+            data = trainer.load_val_data() or trainer.load_train_data()
+
+        # if features are provided in the dataframe, check that they are valid features in the covariate metadata
+        provided_static_columns = [] if data.static_features is None else data.static_features.columns
+        unused_features = [
+            f
+            for f in set(provided_static_columns).union(set(data.columns) - {self.target})
+            if f not in self.feature_generator.covariate_metadata.all_features
+        ]
+
+        if features is None:
+            features = self.feature_generator.covariate_metadata.all_features
+        else:
+            if len(features) == 0:
+                raise ValueError(
+                    "No features provided to compute feature importance. At least some valid features should be provided."
+                )
+            for fn in features:
+                if fn not in self.feature_generator.covariate_metadata.all_features and fn not in unused_features:
+                    raise ValueError(f"Feature {fn} not found in covariate metadata or the dataset.")
+
+        if len(set(features)) < len(features):
+            logger.warning(
+                "Duplicate feature names provided to compute feature importance. This will lead to unexpected behavior. "
+                "Please provide unique feature names across both static features and covariates."
+            )
+
+        data = self.feature_generator.transform(data)
+
+        importance_df = trainer.get_feature_importance(
+            data=data,
+            features=features,
+            model=model,
+            metric=metric,
+            time_limit=time_limit,
+            method=method,
+            subsample_size=subsample_size,
+            num_iterations=num_iterations,
+            random_seed=random_seed,
+            relative_scores=relative_scores,
+            include_confidence_band=include_confidence_band,
+            confidence_level=confidence_level,
+        )
+
+        for feature in set(features).union(unused_features):
+            if feature not in importance_df.index:
+                importance_df.loc[feature] = (
+                    [0, 0, 0] if not include_confidence_band else [0, 0, 0, float("nan"), float("nan")]
+                )
+
+        return importance_df
 
     def leaderboard(self, data: Optional[TimeSeriesDataFrame] = None, use_cache: bool = True) -> pd.DataFrame:
         if data is not None:
--- a/timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
+++ b/timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
@@ -76,7 +76,6 @@ class AbstractTimeSeriesModel(AbstractModel):
 
     supports_known_covariates: bool = False
     supports_past_covariates: bool = False
-    supports_cat_covariates: bool = False
     supports_static_features: bool = False
 
     def __init__(
--- a/timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
+++ b/timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
@@ -74,6 +74,11 @@ class AbstractTimeSeriesModel(AbstractModel):
     _preprocess_nonadaptive = None
     _preprocess_set_features = None
 
+    supports_known_covariates: bool = False
+    supports_past_covariates: bool = False
+    supports_cat_covariates: bool = False
+    supports_static_features: bool = False
+
     def __init__(
         self,
         freq: Optional[str] = None,
--- a/timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
+++ b/timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
@@ -425,6 +425,9 @@ class DirectTabularModel(AbstractMLForecastModel):
         end of each time series).
     """
 
+    supports_known_covariates = True
+    supports_static_features = True
+
     @property
     def is_quantile_model(self) -> bool:
         return self.eval_metric.needs_quantile
@@ -575,6 +578,9 @@ class RecursiveTabularModel(AbstractMLForecastModel):
         If not None, training dataset passed to TabularPredictor will contain at most this many rows (starting from the
         end of each time series).
     """
+
+    supports_known_covariates = True
+    supports_static_features = True
 
     def _get_model_params(self) -> dict:
         model_params = super()._get_model_params()
--- a/timeseries/src/autogluon/timeseries/predictor.py
+++ b/timeseries/src/autogluon/timeseries/predictor.py
@@ -570,7 +570,7 @@ class TimeSeriesPredictor(TimeSeriesPredictorDeprecatedMixin):
             Valid preset values:
 
             * "auto": Performs HPO via bayesian optimization search on GluonTS-backed neural forecasting models and
-                random search on other models using local scheduler.
+              random search on other models using local scheduler.
             * "random": Performs HPO via random search.
 
             You can also provide a dict to specify searchers and schedulers
@@ -892,6 +892,137 @@ class TimeSeriesPredictor(TimeSeriesPredictorDeprecatedMixin):
             logger.info("Evaluations on test data:")
             logger.info(json.dumps(scores_dict, indent=4))
         return scores_dict
+
+    def feature_importance(
+        self,
+        data: Optional[Union[TimeSeriesDataFrame, pd.DataFrame, Path, str]] = None,
+        model: Optional[str] = None,
+        metric: Optional[Union[str, TimeSeriesScorer]] = None,
+        features: Optional[List[str]] = None,
+        time_limit: Optional[float] = None,
+        method: Literal["naive", "permutation"] = "permutation",
+        subsample_size: int = 50,
+        num_iterations: Optional[int] = None,
+        random_seed: Optional[int] = 123,
+        relative_scores: bool = False,
+        include_confidence_band: bool = True,
+        confidence_level: float = 0.99,
+    ):
+        """
+        Calculates feature importance scores for the given model via replacing each feature by a shuffled version of the same feature
+        (also known as permutation feature importance) or by assigning a constant value representing the median or mode of the feature,
+        and computing the relative decrease in the model's predictive performance.
+
+        A feature's importance score represents the performance drop that results when the model makes predictions on a perturbed copy
+        of the data where this feature's values have been randomly shuffled across rows. A feature score of 0.01 would indicate that the
+        predictive performance dropped by 0.01 when the feature was randomly shuffled or replaced. The higher the score a feature has,
+        the more important it is to the model's performance.
+
+        If a feature has a negative score, this means that the feature is likely harmful to the final model, and a model trained with
+        the feature removed would be expected to achieve a better predictive performance. Note that calculating feature importance can
+        be a computationally expensive process, particularly if the model uses many features. In many cases, this can take longer than
+        the original model training. Roughly, this will equal to the number of features in the data multiplied by ``num_iterations``
+        (or, 1 when ``method="naive"``) and time taken when ``evaluate()`` is called on a dataset with ``subsample_size``.
+
+        Parameters
+        ----------
+        data : TimeSeriesDataFrame, pd.DataFrame, Path or str, optional
+            The data to evaluate feature importances on. The last ``prediction_length`` time steps of the data set, for each
+            item, will be held out for prediction and forecast accuracy will be calculated on these time steps.
+            More accurate feature importances will be obtained from new data that was held-out during ``fit()``.
+
+            If ``known_covariates_names`` were specified when creating the predictor, ``data`` must include the columns
+            listed in ``known_covariates_names`` with the covariates values aligned with the target time series.
+            This data must contain the label column with the same column name as specified during ``fit()``.
+
+            If ``train_data`` used to train the predictor contained past covariates or static features, then ``data``
+            must also include them (with same column names and dtypes).
+
+            If provided data is an instance of pandas DataFrame, AutoGluon will attempt to automatically convert it
+            to a ``TimeSeriesDataFrame``. If str or Path is passed, ``data`` will be loaded using the str value as the file path.
+
+            If ``data`` is not provided, then validation (tuning) data provided during training (or the held out data used for
+            validation if ``tuning_data`` was not explicitly provided ``fit()``) will be used.
+        model : str, optional
+            Name of the model that you would like to evaluate. By default, the best model during training
+            (with highest validation score) will be used.
+        metric : str or TimeSeriesScorer, optional
+            Metric to be used for computing feature importance. If None, the ``eval_metric`` specified during initialization of
+            the ``TimeSeriesPredictor`` will be used.
+        features : List[str], optional
+            List of feature names that feature importances are calculated for and returned. By default, all feature importances
+            will be returned.
+        method : {"permutation", "naive"}, default = "permutation"
+            Method to be used for computing feature importance.
+
+            * ``naive``: computes feature importance by replacing the values of each feature by a constant value and computing
+              feature importances as the relative improvement in the evaluation metric. The constant value is the median for
+              real-valued features and the mode for categorical features, for both covariates and static features, obtained from the
+              feature values in ``data`` provided.
+            * ``permutation``: computes feature importance by naively shuffling the values of the feature across different items
+              and time steps. Each feature is shuffled for ``num_iterations`` times and feature importances are computed as the
+              relative improvement in the evaluation metric. Refer to https://explained.ai/rf-importance/ for an explanation of
+              permutation importance.
+
+        subsample_size : int, default = 50
+            The number of items to sample from `data` when computing feature importance. Larger values increase the accuracy of
+            the feature importance scores. Runtime linearly scales with `subsample_size`.
+        time_limit : float, optional
+            Time in seconds to limit the calculation of feature importance. If None, feature importance will calculate without early stopping.
+            If ``method="permutation"``, a minimum of 1 full shuffle set will always be evaluated. If a shuffle set evaluation takes longer than
+            ``time_limit``, the method will take the length of a shuffle set evaluation to return regardless of the `time_limit`.
+        num_iterations : int, optional
+            The number of different iterations of the data that are evaluated. If ``method="permutation"``, this will be interpreted
+            as the number of shuffle sets (equivalent to ``num_shuffle_sets`` in :meth:`TabularPredictor.feature_importance`). If ``method="naive"``, the
+            constant replacement approach is repeated for ``num_iterations`` times, and a different subsample of data (of size ``subsample_size``) will
+            be taken in each iteration.
+            Default is 1 for ``method="naive"`` and 5 for ``method="permutation"``. The value will be ignored if ``method="naive"`` and the subsample
+            size is greater than the number of items in ``data`` as additional iterations will be redundant.
+            Larger values will increase the quality of the importance evaluation.
+            It is generally recommended to increase ``subsample_size`` before increasing ``num_iterations``.
+            Runtime scales linearly with ``num_iterations``.
+        random_seed : int or None, default = 123
+            If provided, fixes the seed of the random number generator for all models. This guarantees reproducible
+            results for feature importance.
+        relative_scores : bool, default = False
+            By default, this method will return expected average *absolute* improvement in the eval metric due to the feature. If True, then
+            the statistics will be computed over the *relative* (percentage) improvements.
+        include_confidence_band: bool, default = True
+            If True, returned DataFrame will include two additional columns specifying confidence interval for the true underlying importance value of
+            each feature. Increasing ``subsample_size`` and ``num_iterations`` will tighten the confidence interval.
+        confidence_level: float, default = 0.99
+            This argument is only considered when ``include_confidence_band=True``, and can be used to specify the confidence level used
+            for constructing confidence intervals. For example, if ``confidence_level`` is set to 0.99, then the returned DataFrame will include
+            columns ``p99_high`` and ``p99_low`` which indicates that the true feature importance will be between ``p99_high`` and ``p99_low`` 99% of
+            the time (99% confidence interval). More generally, if ``confidence_level`` = 0.XX, then the columns containing the XX% confidence interval
+            will be named ``pXX_high`` and ``pXX_low``.
+
+        Returns
+        -------
+        :class:`pd.DataFrame` of feature importance scores with 2 columns:
+            index: The feature name.
+            'importance': The estimated feature importance score.
+            'stddev': The standard deviation of the feature importance score. If NaN, then not enough ``num_iterations`` were used.
+        """
+        if data is not None:
+            data = self._check_and_prepare_data_frame(data)
+            self._check_data_for_evaluation(data)
+
+        fi_df = self._learner.get_feature_importance(
+            data=data,
+            model=model,
+            metric=metric,
+            features=features,
+            time_limit=time_limit,
+            method=method,
+            subsample_size=subsample_size,
+            num_iterations=num_iterations,
+            random_seed=random_seed,
+            relative_scores=relative_scores,
+            include_confidence_band=include_confidence_band,
+            confidence_level=confidence_level,
+        )
+        return fi_df
 
     @classmethod
     def _load_version_file(cls, path: str) -> str:
--- a/timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
+++ b/timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
@@ -23,7 +23,11 @@ from autogluon.timeseries.models.abstract import AbstractTimeSeriesModel
 from autogluon.timeseries.models.ensemble import AbstractTimeSeriesEnsembleModel, TimeSeriesGreedyEnsemble
 from autogluon.timeseries.models.presets import contains_searchspace
 from autogluon.timeseries.splitter import AbstractWindowSplitter, ExpandingWindowSplitter
-from autogluon.timeseries.utils.features import CovariateMetadata
+from autogluon.timeseries.utils.features import (
+    ConstantReplacementFeatureImportanceTransform,
+    CovariateMetadata,
+    PermutationFeatureImportanceTransform,
+)
 from autogluon.timeseries.utils.warning_filters import disable_tqdm
 
 logger = logging.getLogger("autogluon.timeseries.trainer")
@@ -241,6 +245,9 @@ class SimpleAbstractTrainer:
 
 class AbstractTimeSeriesTrainer(SimpleAbstractTrainer):
     _cached_predictions_filename = "cached_predictions.pkl"
+
+    max_rel_importance_score: float = 1e5
+    eps_abs_importance_score: float = 1e-5
 
     def __init__(
         self,
@@ -849,7 +856,9 @@ class AbstractTimeSeriesTrainer(SimpleAbstractTrainer):
                 unpersisted_models.append(model)
         return unpersisted_models
 
-    def _get_model_for_prediction(self, model: Optional[Union[str, AbstractTimeSeriesModel]] = None) -> str:
+    def _get_model_for_prediction(
+        self, model: Optional[Union[str, AbstractTimeSeriesModel]] = None, verbose: bool = True
+    ) -> str:
         """Given an optional identifier or model object, return the name of the model with which to predict.
 
         If the model is not provided, this method will default to the best model according to the validation score.
@@ -858,10 +867,11 @@ class AbstractTimeSeriesTrainer(SimpleAbstractTrainer):
             if self.model_best is None:
                 best_model_name: str = self.get_model_best()
                 self.model_best = best_model_name
-            logger.info(
-                f"Model not specified in predict, will default to the model with the "
-                f"best validation score: {self.model_best}",
-            )
+            if verbose:
+                logger.info(
+                    f"Model not specified in predict, will default to the model with the "
+                    f"best validation score: {self.model_best}",
+                )
             return self.model_best
         else:
             if isinstance(model, AbstractTimeSeriesModel):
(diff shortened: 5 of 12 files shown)


timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -893,6 +893,113 @@ def evaluate(
...
+        self,
+        data: Union[TimeSeriesDataFrame, pd.DataFrame, Path, str],
+        model: Optional[str] = None,
+        metric: Optional[Union[str, TimeSeriesScorer]] = None,
+        features: Optional[List[str]] = None,
+        time_limit: Optional[float] = None,
+        method: Literal["naive", "permutation"] = "permutation",
+        subsample_size: int = 50,
+        num_iterations: Optional[int] = None,
+    ):
```
shchur, line 906: Should we fix the seed here as well for reproducibility (both when selecting `subsample_size` items & permutations)?

Maybe set seed to `123` by default, and then for each of `num_iterations` increment the seed by 1 to ensure reproducible results.

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -893,6 +893,113 @@ def evaluate(
...
+            If ``train_data`` used to train the predictor contained past covariates or static features, then ``data``
+            must also include them (with same column names and dtypes).
+
+            If provided data is an instance of pandas DataFrame, AutoGluon will attempt to automatically convert it
+            to a ``TimeSeriesDataFrame``. If str or Path is passed, ``data`` will be loaded using the str value as the file path.
+        model : str, optional
+            Name of the model that you would like to evaluate. By default, the best model during training
+            (with highest validation score) will be used.
+        metric : str or TimeSeriesScorer, optional
+            Metric to be used for computing feature importance. If None, the ``eval_metric`` specified during ``fit()`` will be used.
```
shchur, line 943: Specified inside `__init__()` / when creating the predictor

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -893,6 +893,113 @@ def evaluate(
...
+        features : List[str], optional
+            List of feature names that feature importances are calculated for and returned. By default, all feature importances
+            will be returned.
+        method : {"naive", "permutation"}, default = "permutation"
```
shchur, line 947: Nitissimo: default argument goes first in numpy docstring style

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -893,6 +893,113 @@ def evaluate(
...
+            Method to be used for computing feature importance.
+
+            * ``naive``: computes feature importance by replacing the values of each feature by a constant value and computing
+              feature importances as the relative improvement in the evaluation metric. The constant value is the median for
+              real-valued features and the mode for categorical features, for both covariates and static features.
```
shchur, line 952: Mention that median/mode are obtained from the training set?

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -893,6 +893,113 @@ def evaluate(
             logger.info(json.dumps(scores_dict, indent=4))
         return scores_dict
 
+    def feature_importance(
+        self,
+        data: Union[TimeSeriesDataFrame, pd.DataFrame, Path, str],
```
shchur, line 898: Make this an optional argument and use validation data (i.e., full `train_data`) saved inside the trainer by default?

yinweisu:
| :---: | :---: |

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -893,6 +893,113 @@ def evaluate(
...
+            List of feature names that feature importances are calculated for and returned. By default, all feature importances
+            will be returned.
+        method : {"naive", "permutation"}, default = "permutation"
+            Method to be used for computing feature importance.
+
+            * ``naive``: computes feature importance by replacing the values of each feature by a constant value and computing
+              feature importances as the relative improvement in the evaluation metric. The constant value is the median for
+              real-valued features and the mode for categorical features, for both covariates and static features.
+            * ``permutation``: computes feature importance by naively shuffling the values of the feature across different items
+              and time steps. Each feature is shuffled for `num_shuffle_steps` times and feature importances are computed as the
```
shchur, line 954: `num_iterations`?

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -893,6 +893,113 @@ def evaluate(
...
+        subsample_size : int, default = 50
+            The number of items to sample from `data` when computing feature importance. Larger values increase the accuracy of
+            the feature importance scores. Runtime linearly scales with `subsample_size`.
+        time_limit : float, optional
+            Time in seconds to limit the calculation of feature importance. If None, feature importance will calculate without early stopping.
+            If ``method="permutation"``, a minimum of 1 full shuffle set will always be evaluated. If a shuffle set evaluation takes longer than
+            ``time_limit``, the method will take the length of a shuffle set evaluation to return regardless of the `time_limit`.
+        num_iterations : int, optional
+            The number of different iterations of the data that are evaluated. If ``method="permutation"``, this will be interpreted
+            as the number of shuffle sets (equivalent to ``num_shuffle_sets`` in :class:`TabularPredictor`). If ``method="naive"``, the
```
shchur, line 967: `in :meth:TabularPredictor.feature_importance`?

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -893,6 +893,113 @@ def evaluate(
...
+            constant replacement approach is repeated for ``num_iterations`` times.
```
shchur, line 968: Maybe clarify that non-determinism comes from subsampling different items when `method="naive"`? (this wasn't obvious to me on the first read)

timeseries/src/autogluon/timeseries/utils/features.py
```diff
@@ -48,6 +52,14 @@ def covariates_real(self) -> List[str]:
     def covariates_cat(self) -> List[str]:
         return self.known_covariates_cat + self.past_covariates_cat
 
+    @property
+    def real_features(self) -> List[str]:
+        return self.static_features_real + self.covariates_real
```
shchur, line 57: Should we have the same method for `cat_features` for completeness?

timeseries/src/autogluon/timeseries/learner.py
```diff
@@ -198,14 +202,58 @@ def score(
...
+                if fn not in self.feature_generator.covariate_metadata.all_features:
+                    raise ValueError(f"Feature {fn} not found in covariate metadata.")
+
+        feature_importance_transform = (
+            {
+                "permutation": PermutationFeatureImportanceTransform,
+                "naive": ConstantReplacementFeatureImportanceTransform,
+            }
+            .get(method)
+            .from_feature_generator(self.feature_generator)
```
shchur, line 240: Maybe we can keep all logic related to feature importance inside `trainer.get_feature_importance`? We should have all the necessary information available there (e.g., `covariate_metadata` is available inside the trainer as `self.covariate_metadata`). This would reduce the mental load on developers if we decide to extend this functionality in the future.

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -936,6 +936,56 @@ def evaluate(
...
+            logger.info(
+                f"Subsample_size {subsample_size} is larger than the number of items in the data and will be ignored"
+            )
+            subsample_size = data.num_items
+
+        relative_drops = defaultdict(list)
+        for n in range(num_iterations):
+            if subsample_size < data.num_items:
+                item_ids_sampled = data.item_ids.to_series().sample(subsample_size)
+                data_sample = data.loc[item_ids_sampled]
```
shchur, line 970: `data.query("item_id in @item_ids_sampled")` is for some reason much faster than `loc` for large dataframes

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -936,6 +936,56 @@ def evaluate(
...
+            else:
+                data_sample = data
+
+            base_score = self.evaluate(data=data_sample, model=model, metrics=metric).get(metric.name)
+
+            for feature in features:
+                data_sample_replaced = feature_importance_transform.transform(data_sample, target_feature=feature)
+                score = self.evaluate(data=data_sample_replaced, model=model, metrics=metric).get(metric.name)
+                relative_drops[feature].append((score - base_score) / -base_score)
```
shchur, line 979: What happens if the model achieves 0 error for data_sample?

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -936,6 +936,56 @@ def evaluate(
...
+                item_ids_sampled = data.item_ids.to_series().sample(subsample_size)
+                data_sample = data.loc[item_ids_sampled]
+            else:
+                data_sample = data
+
+            base_score = self.evaluate(data=data_sample, model=model, metrics=metric).get(metric.name)
+
+            for feature in features:
+                data_sample_replaced = feature_importance_transform.transform(data_sample, target_feature=feature)
+                score = self.evaluate(data=data_sample_replaced, model=model, metrics=metric).get(metric.name)
```
shchur, line 978: Should we set `use_cache=False` here to be safe?

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -936,6 +936,56 @@ def evaluate(
...
+
+        relative_drops = defaultdict(list)
+        for n in range(num_iterations):
+            if subsample_size < data.num_items:
+                item_ids_sampled = data.item_ids.to_series().sample(subsample_size)
+                data_sample = data.loc[item_ids_sampled]
+            else:
+                data_sample = data
+
+            base_score = self.evaluate(data=data_sample, model=model, metrics=metric).get(metric.name)
```
shchur, line 974: Nit: Does it make sense to persist the models here to speed up inference (and unpersist later)?

timeseries/src/autogluon/timeseries/utils/features.py
```diff
@@ -284,3 +296,63 @@ def _check_required_columns_are_present(
...
+    ) -> "AbstractFeatureImportanceTransform":
+        return cls(
+            forecast_target=feature_generator.target,
+            covariate_metadata=feature_generator.covariate_metadata,
+        )
+
+    def _apply_transform(self, data: TimeSeriesDataFrame, target_feature: str) -> TimeSeriesDataFrame:
+        raise NotImplementedError
+
+    def transform(self, data: TimeSeriesDataFrame, target_feature: str, **kwargs) -> TimeSeriesDataFrame:
```
shchur, line 326: Maybe rename `target_feature` -> `feature_name` to avoid confusion with the target time series?

timeseries/src/autogluon/timeseries/utils/features.py
```diff
@@ -284,3 +296,63 @@ def _check_required_columns_are_present(
...
+            data[target_feature] = data[target_feature].sample(frac=1)
+        return data
+
+
+class ConstantReplacementFeatureImportanceTransform(AbstractFeatureImportanceTransform):
+    """Replaces a target feature with the median if it's a real-valued feature, and the mode if it's a
+    categorical feature."""
+
+    def _apply_transform(self, data: TimeSeriesDataFrame, target_feature: str) -> TimeSeriesDataFrame:
+        aggregation = "median" if target_feature in self.covariate_metadata.real_features else "mode"
```
shchur, line 353: I think it would be more prudent to use the mode/median of training data for masking here rather than mode/median of `data`. `data` will usually contain a subsample of 50 items from the full dataset, so the values used for imputation may change significantly based on the subset. 

Using statistics of the training data also has a minor advantage of being similar to our logic for imputing all-missing values.

shchur (commented):
This looks very elegant!

The main points that I want to discuss:
1. Moving all feature importance logic to Trainer
2. Using statistics of train data for masking
3. Use full train_data/tuning_data by default
4. Fixing seed for determinism

timeseries/src/autogluon/timeseries/predictor.py
canerturkmen: Why does the seed need to be incremented (and reset) for each iteration?

shchur: Never mind, just fixing the seed should be fine.

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
canerturkmen: Nice!

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
canerturkmen: Why? I would like to make the cache aware of feature use, but what could go wrong now?

timeseries/src/autogluon/timeseries/utils/features.py
canerturkmen: I'm not sure about 'training' data. I understand if it's not a subsample but the entire data provided during `feature_importance`. We do the imputations before shuffling, and I don't think there is an easy way of escaping

> so the values used for imputation may change significantly based on the subset

yinweisu:
| :---: | :---: |
| pydantic_core==2.16.3 | - |
| ray==2.10.0 | ray==2.6.3 |
| pydantic==2.6.4 | pydantic==1.10.15 |
| annotated-types==0.6.0 | - |
| platformdirs==4.2.0 | platformdirs==3.11.0 |
| virtualenv==20.25.1 | virtualenv==20.21.0 |
| ray==2.10.0 | ray==2.6.3 |
| - | gpustat==1.1.1 |
| pydantic==2.6.4 | pydantic==1.10.15 |
| - | blessed==1.20.0 |
| - | nvidia-ml-py==12.535.133 |
| platformdirs==4.2.0 | platformdirs==3.11.0 |
| virtualenv==20.25.1 | virtualenv==20.21.0 |

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
shchur: Right, if we update the caching logic, it should be fine.

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
shchur: Should we report absolute numbers + log a warning in that case?

Also, should we in general report unnormalized feature importance values (or at least add a flag `normalize_score=True`)?

timeseries/src/autogluon/timeseries/utils/features.py
```diff
@@ -284,3 +300,53 @@ def _check_required_columns_are_present(
...
+
+class PermutationFeatureImportanceTransform(AbstractFeatureImportanceTransform):
+    """Naively shuffles a given feature."""
+
+    def _apply_transform(self, data: TimeSeriesDataFrame, feature_name: str) -> TimeSeriesDataFrame:
+        if feature_name in self.covariate_metadata.static_features:
+            data.static_features[feature_name] = data.static_features[feature_name].sample(frac=1)
+        else:
+            data[feature_name] = data[feature_name].sample(frac=1)
+        return data
```
Innixma, line 339: Add random seed. Importance should be deterministic given a seed

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -936,6 +940,88 @@ def evaluate(
...
+        data: TimeSeriesDataFrame,
+        features: Optional[List[str]] = None,
+        model: Optional[Union[str, AbstractTimeSeriesModel]] = None,
+        metric: Optional[Union[str, TimeSeriesScorer]] = None,
+        time_limit: Optional[float] = None,
+        method: Literal["naive", "permutation"] = "permutation",
+        subsample_size: int = 50,
+        num_iterations: int = 1,
+        random_seed: Optional[int] = None,
+    ) -> pd.DataFrame:
```
Innixma, line 954: Is there a reason that we aren't re-using Tabular's feature importance logic? https://github.com/autogluon/autogluon/blob/master/core/src/autogluon/core/utils/utils.py#L688-L929

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -893,6 +893,120 @@ def evaluate(
             logger.info(json.dumps(scores_dict, indent=4))
         return scores_dict
 
+    def feature_importance(
+        self,
+        data: Optional[Union[TimeSeriesDataFrame, pd.DataFrame, Path, str]],
+        model: Optional[str] = None,
+        metric: Optional[Union[str, TimeSeriesScorer]] = None,
+        features: Optional[List[str]] = None,
```
Innixma, line 901: Note: Can support feature-group-importance if using Tabular's logic: https://github.com/autogluon/autogluon/blob/master/tabular/src/autogluon/tabular/predictor/predictor.py#L2851-L2857

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -893,6 +893,120 @@ def evaluate(
...
+            results for feature importance.
+
+        Returns
+        -------
+        :class:`pd.DataFrame` of feature importance scores with 2 columns:
+            index: The feature name.
+            'importance': The estimated feature importance score.
+            'stddev': The standard deviation of the feature importance score. If NaN, then not enough ``num_iterations`` were used.
+        """
+        # TODO: include confidence interval and t-test computation
```
Innixma, line 992: No confidence band arguments?
https://github.com/autogluon/autogluon/blob/master/tabular/src/autogluon/tabular/predictor/predictor.py#L2817-L2818

One option is to move the logic in Tabular to core and re-use: 
https://github.com/autogluon/autogluon/blob/master/tabular/src/autogluon/tabular/predictor/predictor.py#L2933-L2959

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -893,6 +893,120 @@ def evaluate(
             logger.info(json.dumps(scores_dict, indent=4))
         return scores_dict
 
+    def feature_importance(
```
Innixma, line 896: No unit test / No usage in tutorial?

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -936,6 +940,88 @@ def evaluate(
...
+            for feature in features:
+                data_sample_replaced = importance_transform.transform(data_sample, feature_name=feature)
+                score = self.evaluate(data=data_sample_replaced, model=model, metrics=metric).get(metric.name)
+                relative_drops[feature].append((score - base_score) / -base_score)
+
+            if time_limit is not None and time.time() - time_start > time_limit:
+                logger.info(f"Time limit reached, stopping feature importance computation after {n} iterations")
+                break
+
+        self.unpersist(model_names="all")
```
Innixma, line 1019: What if the models were already persisted by the user? They probably wouldn't expect their models to be unpersisted by calling `feature_importance`.

Innixma: Better to track which models were persisted at the start of the logic so we know what to unpersist to ensure the same state at the end in terms of models that are persisted.

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
Innixma: Absolute numbers should be reported in all cases IMO. relative numbers can be reported as an additional column.

Innixma: Tabular reports absolute numbers

timeseries/src/autogluon/timeseries/utils/features.py
```diff
@@ -284,3 +300,53 @@ def _check_required_columns_are_present(
...
+
+    def transform(self, data: TimeSeriesDataFrame, feature_name: str, **kwargs) -> TimeSeriesDataFrame:
+        if feature_name not in self.covariate_metadata.all_features:
+            raise ValueError(f"Target feature {feature_name} not found in covariate metadata")
+
+        # feature transform works on a shallow copy of the main time series data frame
+        # but a deep copy of the static features.
+        data = data.copy(deep=False)
+
+        return self._apply_transform(data, feature_name, **kwargs)
```
Innixma, line 328: This is similar to my initial implementation of feature importance in Tabular, but it turns out to be compute inefficient as it does excessive data mutations (I think). If you run into performance issues, consider checking the Tabular code implementation I linked elsewhere in my review.

timeseries/src/autogluon/timeseries/utils/features.py
Innixma, line 339: BUG: To properly calculate importance that is comparable across features, the shuffling needs to be done with the same indices for each feature. This requires using a seed.

Otherwise, some features might just get lucky and get a favorable `.sample` order, which pollutes the correctness of the feature importance score.

Innixma (commented):
Added initial review

timeseries/src/autogluon/timeseries/learner.py
```diff
@@ -198,14 +202,58 @@ def score(
...
+        )
+
+        if features is None:
+            features = feature_importance_transform.covariate_metadata.all_features
```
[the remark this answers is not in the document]
Innixma: This is what is done in Tabular I believe

canerturkmen:
> Added initial review

Thanks for the review @Innixma ! This was the first cut trying to pin down the design. I'll base my replies on that.

timeseries/src/autogluon/timeseries/utils/features.py
canerturkmen: I'm surprised. In expectation over shuffles and considering that features are independent by assumption I would assume this effect to be negligible if any. This is also reflected in the pseudocode in our main [reference](https://explained.ai/rf-importance/#4). In the [eli5](https://github.com/eli5-org/eli5/blob/master/eli5/permutation_importance.py) implementation, it's only mentioned as a computational trick. It's also not in the Breiman paper.

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
canerturkmen: The data format... We have both "static" and "dynamic" features that need to be treated separately as well as those dynamic features that should only be shuffled in the past or those only in the future. As usual, the `X, y` API does not carry over in this case.

timeseries/src/autogluon/timeseries/predictor.py
canerturkmen: I'll punt if you don't think this is critical.

timeseries/src/autogluon/timeseries/predictor.py
canerturkmen: will do reimplement here. thanks for pointing it out!

timeseries/src/autogluon/timeseries/predictor.py
canerturkmen: TK :)

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
canerturkmen: 👍 thanks for catching this!!

timeseries/src/autogluon/timeseries/utils/features.py
canerturkmen: Thanks! I worked on this a little and so far it looks good because we take care of shallow copies in the custom `TimeSeriesDataFrame`. I'll keep this in mind.

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
canerturkmen: will revise to keep consistent with Tabular, and add option so it can become relative if `normalize_score=True`.

yinweisu:
| :---: | :---: |

timeseries/src/autogluon/timeseries/learner.py
```diff
@@ -198,14 +198,44 @@ def score(
...
+        features: Optional[List[str]] = None,
+        time_limit: Optional[float] = None,
+        method: Literal["naive", "permutation"] = "permutation",
+        subsample_size: int = 50,
+        num_iterations: int = 1,
+        random_seed: Optional[int] = None,
+    ) -> pd.DataFrame:
+        trainer = self.load_trainer()
+        if data is None:
+            data = trainer.load_val_data()
```
shchur, line 224: This will return `None` if `tuning_data` was not provided at fit time, so we need a fallback mechanism that loads `train_data` if `val_data` is None.

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -893,6 +893,120 @@ def evaluate(
...
+        self,
+        data: Optional[Union[TimeSeriesDataFrame, pd.DataFrame, Path, str]],
```
shchur, line 898: Defaults to `None` (same as in `TabularPredictor.feature_importance` and `TimeSeriesPredictor.leaderboard`)?

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -893,6 +893,120 @@ def evaluate(
...
+        time_limit : float, optional
+            Time in seconds to limit the calculation of feature importance. If None, feature importance will calculate without early stopping.
+            If ``method="permutation"``, a minimum of 1 full shuffle set will always be evaluated. If a shuffle set evaluation takes longer than
+            ``time_limit``, the method will take the length of a shuffle set evaluation to return regardless of the `time_limit`.
+        num_iterations : int, optional
+            The number of different iterations of the data that are evaluated. If ``method="permutation"``, this will be interpreted
+            as the number of shuffle sets (equivalent to ``num_shuffle_sets`` in :meth:`TabularPredictor.feature_importance`). If ``method="naive"``, the
+            constant replacement approach is repeated for ``num_iterations`` times, and a different subsample of data (of size ``subsample_size``) will
+            be taken in each iteration.
+            Default is 1 for ``method="naive"`` and 5 for ``method="permutation"``. The value will be ignored f ``method="naive"`` and the subsample
```
shchur, line 977: ignored if

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -893,6 +893,120 @@ def evaluate(
...
+        Returns
+        -------
+        :class:`pd.DataFrame` of feature importance scores with 2 columns:
+            index: The feature name.
+            'importance': The estimated feature importance score.
+            'stddev': The standard deviation of the feature importance score. If NaN, then not enough ``num_iterations`` were used.
+        """
+        # TODO: include confidence interval and t-test computation
+
+        data = self._check_and_prepare_data_frame(data)
```
shchur, line 998: This is going to error out if `data is None`. Replace with
```python
if data is not None:
    data = self._check_and_prepare_data_frame(data)
    slef._check_data_for_evaluation(data)
```

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -936,6 +940,88 @@ def evaluate(
...
+        model: Optional[Union[str, AbstractTimeSeriesModel]] = None,
+        metric: Optional[Union[str, TimeSeriesScorer]] = None,
+        time_limit: Optional[float] = None,
+        method: Literal["naive", "permutation"] = "permutation",
+        subsample_size: int = 50,
+        num_iterations: int = 1,
+        random_seed: Optional[int] = None,
+    ) -> pd.DataFrame:
+        assert method in ["naive", "permutation"], f"Invalid feature importance method {method}."
+        metric = check_get_evaluation_metric(metric) if metric else self.eval_metric
```
shchur, line 960: Are we sure that `TimeSeriesScorer` objects are correctly interpreted as booleans? Otherwise we could be safe and check `if metric is not None`

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -936,6 +940,88 @@ def evaluate(
...
+
+        # start timer and cap subsample size if it's greater than the number of items in the provided data set
+        time_start = time.time()
+        if subsample_size > data.num_items:
+            logger.info(
+                f"Subsample_size {subsample_size} is larger than the number of items in the data and will be ignored"
+            )
+            subsample_size = data.num_items
+
+        # set default number of iterations andcap iterations if the number of items in the data is smaller
```
shchur, line 976: Nit: and cap

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -936,6 +940,88 @@ def evaluate(
...
+
+            if time_limit is not None and time.time() - time_start > time_limit:
+                logger.info(f"Time limit reached, stopping feature importance computation after {n} iterations")
+                break
+
+        self.unpersist(model_names=persisted_models)
+
+        return (
+            pd.DataFrame(relative_drops).agg(["mean", "std"]).T.rename(columns={"mean": "importance", "std": "stdev"})
+        )
```
shchur, line 1038: What happens if a static feature & covariate have the same name? Should we add some suffix/prefix to the static features in that case?

timeseries/src/autogluon/timeseries/utils/features.py
```diff
@@ -284,3 +301,100 @@ def _check_required_columns_are_present(
...
+        # but a deep copy of the static features.
+        data = data.copy(deep=False)
+
+        is_categorical = feature_name in self.covariate_metadata.cat_features
+
+        if feature_name in self.covariate_metadata.past_covariates:
+            # we'll have to work on the history of the data alone
+            data[feature_name] = data[feature_name].copy()
+            feature_data = (
+                data[feature_name].groupby(level=ITEMID, sort=False).nth(slice(None, -self.prediction_length))
```
shchur, line 338: Nit: `.groupy(...).head(-self.prediction_length)` is slightly faster than `nth`

timeseries/src/autogluon/timeseries/utils/features.py
```diff
@@ -284,3 +301,100 @@ def _check_required_columns_are_present(
...
+            )
+            data[feature_name].update(self._transform_series(feature_data, is_categorical=is_categorical))
+        else:
```
shchur, line 341: Nit: Would the following logical flow work as well?
```python
if feature_name in self.covariate_metadata.past_covariates:
    ....
elif feature_name in self.covariate_metadata.static_features:
    ...
else:
   ...
```

timeseries/src/autogluon/timeseries/utils/features.py
```diff
@@ -284,3 +301,100 @@ def _check_required_columns_are_present(
...
+            if feature_name in self.covariate_metadata.static_features:
+                feature_data = data.static_features[feature_name].copy()
+                feature_data.index = np.ones(
```
shchur, line 344: Nit: `feature_data.reset_index(drop=True)` would work as well

timeseries/tests/unittests/test_features.py
```diff
@@ -169,3 +174,252 @@ def test_when_static_features_contain_missing_values_then_they_are_filled_during
 
     data_transformed = feat_generator.fit_transform(data)
     assert not data_transformed.static_features.isna().any(axis=None)
+
+
+@pytest.mark.parametrize(
+    "item_id_to_length",
+    [
+        {"A": 10, "B": 15},
+        {"A": 10, "B": 10, "C": 10},
```
shchur, line 183: To be extra safe, use item_ids that are not sorted alphabetically (e.g. B, C, A) to avoid weird bugs where groupby accidentally sorts the index.

yinweisu:
| :---: | :---: |
| botocore==1.34.77 | botocore==1.34.78 |
| boto3==1.34.77 | boto3==1.34.78 |
| botocore==1.34.77 | botocore==1.34.78 |
| boto3==1.34.77 | boto3==1.34.78 |

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
canerturkmen: Thanks for this callout! Would this clash be handled by the feature generator otherwise?

yinweisu:
| :---: | :---: |

timeseries/src/autogluon/timeseries/utils/features.py
Innixma: I agree that it probably doesn't make a large difference, but since we are scoring each of the features, we should ensure that the way we score them is aligned, so it doesn't hurt to use a fixed sample order. I'd strongly prefer doing a fixed order as it avoids strange scenarios (for example, if the user asks for the importance of features (A, B, D) and in a separate call features (A, B, C, D), they will get a different feature importance for D in the two calls, because it will use a different sample order.

yinweisu:
| :---: | :---: |

yinweisu:
| :---: | :---: |

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
canerturkmen: We currently don't handle this in the inner loop of feature importance either.

yinweisu:
| :---: | :---: |

timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
```diff
@@ -576,6 +579,9 @@ class RecursiveTabularModel(AbstractMLForecastModel):
         end of each time series).
     """
 
+    supports_known_covariates = True
```
shchur, line 582: Should we just add this to the base class `AbstractMLForecastModel`?

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -936,6 +946,172 @@ def evaluate(
...
+        if method == "naive" and data.num_items <= subsample_size:
+            num_iterations = 1
+
+        # if features are provided, check that they are valid features in the covariate metadata
+        provided_static_columns = [] if data.static_features is None else data.static_features.columns
+        unused_features = [
+            f
+            for f in set(provided_static_columns).union(set(data.columns) - {self.target})
+            if f not in self.metadata.all_features
+        ]
```
shchur, line 993: 1. Unused columns have been filtered out when we called `feature_generator.transform(data)` inside Learner, so I think that this set will always be empty. Probably, we need to move the unused features logic as to the learner as well as a postprocessing setup before returning the FI DF.
2. This assumes that there are no duplicate columns in `data` & `static_features`, but I'm not sure if this will matter after we fix (1)

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -936,6 +946,172 @@ def evaluate(
...
+
+        importance_samples = defaultdict(list)
+        for n in range(num_iterations):
+            if subsample_size < data.num_items:
+                item_ids_sampled = data.item_ids.to_series().sample(subsample_size)  # noqa
+                data_sample = data.query("item_id in @item_ids_sampled")
+            else:
+                data_sample = data
+
+            base_score = self.evaluate(data=data_sample, model=model, metrics=metric, use_cache=False).get(metric.name)
```
shchur, line 1031: Probably nit: Should we use `[metric.name]` instead of `get` to rule out some weird (probably impossible) situations where `metric.name` is not in the output?

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -936,6 +946,172 @@ def evaluate(
...
+        models_with_ancestors = set(self.get_minimum_model_set(model))
+
+        if feature in self.metadata.static_features:
+            return any(self.load_model(m).supports_static_features for m in models_with_ancestors)
+        elif feature in self.metadata.known_covariates:
+            return any(self.load_model(m).supports_known_covariates for m in models_with_ancestors)
+        elif feature in self.metadata.past_covariates:
+            return any(self.load_model(m).supports_past_covariates for m in models_with_ancestors)
+
+        return False
```
shchur, line 1086: Very nice!

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -936,6 +946,172 @@ def evaluate(
...
+
+    def _add_ci_to_feature_importance(
+        self, importance_df: pd.DataFrame, confidence_level: float = 0.99
+    ) -> pd.DataFrame:
+        """Add confidence intervals to the feature importance."""
+        import scipy.stats
+
+        if confidence_level <= 0.5 or confidence_level >= 1.0:
+            raise ValueError("confidence_level must lie between 0.5 and 1.0")
+        ci_str = "{:0.0f}".format(confidence_level * 100)
```
shchur, line 1096: Nit: a slightly cleaner option
```python
ci_str = f"{confidence_level:.0%}"
```

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
canerturkmen: wontfix for now, we will be raising a warning

yinweisu:
| :---: | :---: |
| typing_extensions==4.10.0 | typing_extensions==4.11.0 |
| typing_extensions==4.10.0 | typing_extensions==4.11.0 |

yinweisu:
| :---: | :---: |
| lazy_loader==0.3 | lazy_loader==0.4 |
| lazy_loader==0.3 | lazy_loader==0.4 |

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
```diff
@@ -74,6 +74,11 @@ class AbstractTimeSeriesModel(AbstractModel):
     _preprocess_nonadaptive = None
     _preprocess_set_features = None
 
+    supports_known_covariates: bool = False
+    supports_past_covariates: bool = False
+    supports_cat_covariates: bool = False
+    supports_static_features: bool = False
```
shchur, line 80: This is a GluonTS-specific flag that essentially says "please one-hot encode the categorical covariates", so I don't think it belongs here.

Alternatively, we need to set this to True as well for MLForecast models since they also support cat features.

shchur (approved):
LGTM, thank you for taking care of this massive feature! 🚀

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
shchur: I mean the `supports_cat_covariates` flag

yinweisu:
| :---: | :---: |

PR #4038, opened 2024-04-04
[AutoBench] Fix Auto benchmarking and add Report generation feature

prateekdesai04:
*Description of changes:*

This PR addresses 3 things:

1. Fixes auto-benchmarking feature which started failing due to added changes in `AutoGluon Bench 0.4.3`, initially we were using `0.4.2`
2. This PR also adds the report generation feature, after every benchmark run and artifact can be downloaded which has Win rate, train time and inference times of the current master branch which is compared to the `AutoGluon v1.0`. This helps determine which version is better
3. This also adds packages like `pyarrow` which are not present in `0.4.3`, this package is needed to run the latest `autogluon-bench` evaluation scripts

Note: This code has also been refactored, there may be spots where more refactoring may be needed, which may be done after v1.1 release as this is not a direct AG feature.

[Link to successful benchmark run will be added here once it's finished]


yinweisu:
| :---: | :---: |

yinweisu:
| :---: | :---: |

.github/workflow_scripts/env_setup.sh
```diff
@@ -18,6 +18,7 @@ function setup_build_contrib_env {
 
 function setup_benchmark_env {
     pip install -U autogluon.bench
+    pip install pyarrow
```
tonyhoo, line 21: Is this a temporary fix? If yes, please insert a TODO note for its subsequent removal.

.github/workflow_scripts/run_benchmark.sh
```diff
@@ -32,7 +34,10 @@ if [ $BRANCH_OR_PR_NUMBER != "master" ]; then
     fi
 fi
 
-python CI/bench/evaluate.py --config_path ./ag_bench_runs/$MODULE/ --module_name $MODULE --time_limit $TIME_LIMIT --branch_name $BRANCH_OR_PR_NUMBER
+python CI/bench/evaluate.py --config_path ./ag_bench_runs/$MODULE/ --module_name $MODULE --time_limit $TIME_LIMIT --branch_name $BRANCH_OR_PR_NUMBER --benchmark_type $BENCHMARK
+
+echo "Deleting version1.0_file"
+rm -f ./results/version1.0*
```
tonyhoo, line 40: Why these logics are needed?

.github/workflows/benchmark-command.yml
```diff
@@ -28,7 +28,7 @@ on:
           - good
           - medium
           - dev
-          - master_medium
+          - master_best
```
tonyhoo, line 31: Aren't the `best` on line 26 already benchmark against master?

.github/workflows/benchmark-command.yml
```diff
@@ -28,7 +28,7 @@ on:
           - good
           - medium
           - dev
```
tonyhoo, line 30: What does this signify? I'm trying to align others with the appropriate presets while questioning if `dev` aligns with the same concept.

.github/workflow_scripts/run_benchmark.sh
prateekdesai04: This is needed for report generation, we download the version 1.0 cleaned benchmark file and evaluate the current cleaned benchmark run against this file, and then we delete the version1.0 file.
We then copy the current cleaned file to S3, thus we do not need the version1.0 file.

.github/workflows/benchmark-command.yml
prateekdesai04: That `best` preset is for Tabular.
The `master_best` preset is for AutoMM

More on that here: https://github.com/autogluon/autogluon/blob/master/CI/bench/multimodal/custom_user_dir/multimodal_frameworks_template.yaml
https://github.com/autogluon/autogluon-bench/blob/master/sample_configs/resources/multimodal_frameworks.yaml

I changed it to `master_best` because we were benchmarking on `medium_quality` earlier, best will give better results

.github/workflows/benchmark-command.yml
prateekdesai04: Yup `dev` aligns with the same concept as well, these are all presets for different modules.
`dev` in particular is for timeseries, we can change the names but I have tried to keep this similar to the ones in AG-Bench repo to maintain uniformity.

You can find it in Bench here: https://github.com/autogluon/autogluon-bench/blob/master/sample_configs/amlb_configs/frameworks_example.yaml

and in the main repo here: https://github.com/autogluon/autogluon/blob/master/CI/bench/timeseries/amlb_user_dir/frameworks_template.yaml

It's listed as `_dev`

.github/workflow_scripts/env_setup.sh
prateekdesai04: This is a temporary fix, version 0.4.3 of AG bench required this package in order for evaluation scripts to work.
But it was released without this package being added
In order to avoid another release process of AG Bench, I added it here.
Inserted TODO

yinweisu:
| :---: | :---: |

CI/bench/evaluate.py
```diff
@@ -1,142 +1,250 @@
...
-
-    for framework in frameworks:
-        modified_list_frameworks.append('--frameworks-run')
-        modified_list_frameworks.append(framework)
+def process_results(eval_flag: bool):
+    try:
+        paths = []
+        frameworks = []
+        for file in os.listdir("./results"):
+            if file.endswith(".csv") and not file.endswith("_min.csv") and not file.endswith(".parquet"):
```
suzhoum, line 14: I think `not file.endswith(".parquet")` is not necessary here

CI/bench/evaluate.py
```diff
@@ -1,142 +1,250 @@
...
+                "./results",
+            ],
+            check=True,
+        )
+
+        subprocess.run(
+            [
+                "sh",
+                "-c",
+                "rm -f ./results/*_min.csv ./results/*_min.parquet ./results/*.parquet",
```
suzhoum, line 155: If these are removed, is the logic `if file.endswith(".csv") and not file.endswith("_min.csv") and not file.endswith(".parquet"):` still necessary?

.github/workflows/benchmark-command.yml
tonyhoo: The current name is somewhat confusing. It would be beneficial to rename it, ensuring that its purpose is distinctly clear to all team members.

.github/workflows/benchmark-command.yml
tonyhoo: ditto

CI/bench/multimodal/custom_user_dir/multimodal_constraints.yaml
```diff
@@ -11,3 +11,4 @@ g4_12x:
   MAX_MACHINE_NUM: 200  
   BLOCK_DEVICE_VOLUME: 300  
   RESERVED_MEMORY_SIZE: 8000
+  TIME_LIMIT: 3600
```
suzhoum, line 14: Just a note that the `TIME_LIMIT` will be passed to predictor.fit(time_limt=TIME_LIMIT), which terminates the training depending on which comes first, time limit or max_epochs

.github/workflows/benchmark-command.yml
prateekdesai04: Yup makes sense, can I put a follow-up PR because I have already started a benchmark run on this.
Touching framework/preset names means restarting all the benchmarking tasks.

CI/bench/evaluate.py
prateekdesai04: Adding this for robustness in case evaluation scripts are changed and more files are added.

CI/bench/evaluate.py
prateekdesai04: Adding this for robustness in case evaluation scripts are changed and more files are added.

suzhoum: anything `.endswith(".csv")` will not end with `.parquet`, so the last part is redundant.

tonyhoo (approved):
Please ensure proper renaming following this merge

PR #4054, opened 2024-04-06
[timeseries] Clarify documentation related to `test_data`

### LINKED ISSUE #4050: Add an explanation of requirements for data input in TimeSeriesPredictor.leaderboard()
### Describe the issue linked to the documentation

Hello, 

Thanks for the great library that you built! I'm relatively new to AutoGluon Time Series and recently we ran into an issue while running `predictor.leaderboard()` on a `test_data`, by reading [the source code](https://github.com/autogluon/autogluon/blob/master/timeseries/src/autogluon/timeseries/predictor.py#L1216) we noticed that it is required the number of timesteps in the data input to this function must be more than `prediction_length` value, I think this is an important information that needs to be mentioned in the documentation both [here](https://auto.gluon.ai/stable/tutorials/timeseries/forecasting-quick-start.html#evaluating-the-performance-of-different-models) and [here](https://auto.gluon.ai/stable/api/autogluon.timeseries.TimeSeriesPredictor.leaderboard.html#autogluon.timeseries.TimeSeriesPredictor.leaderboard), and additionally I'm also wondering if you could explain the reason for such requirement and why can't `test_data` have exactly `prediction_length` number of timesteps? 

Thanks in advance!

### Suggest a potential alternative/fix

Add a note to `data` variable saying something like below and explain why there is such requirement
> *Please make sure that data includes both historic and future data, and that all time series have length > `prediction_length` (at least `prediction_length `+ 1)*

Thank you!

shchur:
*Issue #, if available:* fixes #4050

*Description of changes:*
- Clearly describe that test data must include both historic & future time series values

### DIFF
--- a/docs/tutorials/timeseries/forecasting-indepth.ipynb
+++ b/docs/tutorials/timeseries/forecasting-indepth.ipynb
@@ -2030,8 +2030,8 @@
    "metadata": {},
    "source": [
     "We obtained two `TimeSeriesDataFrame`s from our original data:\n",
-    "- `test_data` contains exactly the same data as the original `data`\n",
-    "- In `train_data`, the last `prediction_length` time steps are removed from the end of each time series"
+    "- `test_data` contains exactly the same data as the original `data` (i.e., it contains both historic data and the forecast horizon)\n",
+    "- In `train_data`, the last `prediction_length` time steps are removed from the end of each time series (i.e., it contains only historic data)"
    ]
   },
   {
--- a/docs/tutorials/timeseries/forecasting-quick-start.ipynb
+++ b/docs/tutorials/timeseries/forecasting-quick-start.ipynb
@@ -240,6 +240,8 @@
     "We provide the test data set to the leaderboard function to see how well our fitted models are doing on the unseen test data.\n",
     "The leaderboard also includes the validation scores computed on the internal validation dataset.\n",
     "\n",
+    "Note the test data includes both the forecast horizon (last `prediction_length` values of each time series) as well as the historic data (all except the last `prediction_last` values).\n",
+    "\n",
     "In AutoGluon leaderboards, higher scores always correspond to better predictive performance.\n",
     "Therefore our MASE scores are multiplied by `-1`, such that higher \"negative MASE\"s correspond to more accurate forecasts."
    ]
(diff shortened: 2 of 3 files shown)


yinweisu:
| :---: | :---: |

canerturkmen (approved):
LGTM! Thanks!

lleiou:
Thank you so much for moving so quickly to adopt my suggested changes! 

Just want to follow up on the reason for requiring historical period in `test_data`: let's say the `prediction_length` of my `predictor` is 30 days, why can't I pass a `test_data` which also has 30 days of data to `.leaderboard()` since that is the only data I want to calculate the evaluation score on, thanks!

shashankumar2812:
I have the same question: why is it required that ``data`` must be at least `prediction_length + 1`? Why not `prediction_length` instead of `prediction_length+1`?

> Thank you so much for moving so quickly to adopt my suggested changes!
> 
> Just want to follow up on the reason for requiring historical period in `test_data`: let's say the `prediction_length` of my `predictor` is 30 days, why can't I pass a `test_data` which also has 30 days of data to `.leaderboard()` since that is the only data I want to calculate the evaluation score on? Thanks!

shchur:
@lleiou @shashankumar2812 When we call `predictor.evaluate(data)`, essentially, the following happens:
1. We split `data` into two disjoint parts: 
    - `future_data` (last `prediction_length` time steps of each time series)
    - `past_data` (everything before the last `prediction_length` steps of each time series). 
3. We call `forecast = predictor.predict(past_data)`
4. We compare forecast & future data `score = eval_metric(future_data, forecast)`.

If we only provide the last `prediction_length` observations of each time series to `predictor.evaluate`, then after the split our `past_data` will be empty, so the predictor won't be able to generate a forecast for the forecast horizon.

We opted for this more general design (where both past data + future data is required when calling evaluate) since it allows us to evaluate the predictor on different time series / different train-test splits. If we went for the alternative approach, where `predictor.evaluate` accepts just the future data, this would mean that we can only evaluate the predictor on the exact same time series used for training, using only the future values starting from the end of the train data.

shashankumar2812:
Thanks @shchur for your answer

PR #4070, opened 2024-04-09
[timeseries] Fix prediction failing if context contains < 3 observations

shchur:
*Description of changes:*
- Many models currently rely on `data.freq` of the data passed to `model.predict(data)` when generating the forecast. It may happen that some time series in `data` passed to `predict` contain just a single observation. In this case, `freq` inference is impossible and the model fails.
- With this PR, we
  1. Rely on `self.freq` attribute of the model instead of `data.freq` whenever frequency is required
  2. Make sure that `data.fill_missing_values()` works even no `freq` is available (as this is too strong of a requirement for this method)

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
+++ b/timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
@@ -46,6 +46,7 @@ class SimpleGluonTSDataset(GluonTSDataset):
     def __init__(
         self,
         target_df: TimeSeriesDataFrame,
+        freq: str,
         target_column: str = "target",
         feat_static_cat: Optional[np.ndarray] = None,
         feat_static_real: Optional[np.ndarray] = None,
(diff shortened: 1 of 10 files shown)


yinweisu:
| :---: | :---: |

yinweisu:
| :---: | :---: |

yinweisu:
| :---: | :---: |

yinweisu:
| :---: | :---: |

timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
```diff
@@ -234,13 +234,6 @@ def _get_hpo_backend(self):
 
     def _deferred_init_params_aux(self, dataset: TimeSeriesDataFrame) -> None:
         """Update GluonTS specific parameters with information available only at training time."""
-        self.freq = dataset.freq or self.freq
-        if not self.freq:
```
canerturkmen, line 238: 1/ Why remove `self.freq = dataset.freq or self.freq`? We never depend on deferred initialization for freq in any other model? Is freq always set for all models by the trainer?
2/ Why remove this assertion? maybe `assert self.freq`?

shchur: 1. No other model currently depends on deferred initialization of `freq` (both local & MLForecast models assume that `freq` is provided at creation time); trainer always sets `freq` when creating models (https://github.com/autogluon/autogluon/blob/master/timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py#L576)
2. Assuming you are referring to the assertion inside `SimpleGluonTSDataset.__init__`: after pandas 2.2 upgrade, we will anyway need to remove this part completely and always feed a dummy frequency `W` to GluonTS. So the assertion would be removed anyway after #4074 is merged.

PR #4069, opened 2024-04-09
[timeseries] Add scale feature and normalize covariates for MLForecast models

shchur:
*Description of changes:*
- Some lightweight feature engineering for MLForecast-based models:
  - Add two static features, encoding mean & std of the target time series for each item
  - For each continuous real-valued covariate, add a copy of this covariate that is mean-abs-scaled per item

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #4086, opened 2024-04-11
[timeseries] Fix incorrect scaling used by MLForecast models

shchur:
*Description of changes:*
- The covariate scaling logic introduced in #4069 had a mistake - we used a different for past & future data. This could result in a pathological case: when `prediction_length=1`, scaled covariates will just become equal to 1. This PR fixes this problem and uses the same scale for computed on past data when scaling the covariates. To implement the above change, we merge the `preprocess` & `preprocess_known_covariates` methods into one.
- Use `pd.option_context` to silence pandas `SettingWithCopyWarning` instead of unnecessary copies.

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
+++ b/timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
@@ -3,7 +3,7 @@ import os
 import re
 import time
 from contextlib import nullcontext
-from typing import Dict, List, Optional, Union
+from typing import Dict, List, Optional, Tuple, Union
 
 from autogluon.common import space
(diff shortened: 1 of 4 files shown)


yinweisu:
| :---: | :---: |
| ruff==0.3.5 | ruff==0.3.6 |
| ruff==0.3.5 | ruff==0.3.6 |

yinweisu:
| :---: | :---: |
| pydantic_core==2.16.3 | pydantic_core==2.18.1 |
| pydantic==2.6.4 | pydantic==2.7.0 |
| pydantic_core==2.16.3 | pydantic_core==2.18.1 |
| pydantic==2.6.4 | pydantic==2.7.0 |

yinweisu:
| :---: | :---: |

PR #4088, opened 2024-04-11
[Enhancement][AutoBench] Adding more columns for detailed report 

prateekdesai04:
*Description of changes:*
This is an enhancement PR which gives out a more detailed report of all the benchmarks of Master vs v1.0.
Initially, the report had `Winrate`, `training time`, `inference time` and `rank`.
This may not provide sufficient information to determine if the framework is necessarily better hence this PR is to give out a more detailed report (upload the entire evaluation as a zipped artifact for every module).

P.S. We can add it as a 1.1 milestone item if required, may not be necessary.

### DIFF
--- a/CI/bench/evaluate.py
+++ b/CI/bench/evaluate.py
@@ -106,7 +106,6 @@ def main():
     branch_name = args.branch_name
     benchmark_type = args.benchmark_type
     df1 = pd.DataFrame()
-    df2 = pd.DataFrame()
 
     try:
         for root, dirs, files in os.walk(config_path):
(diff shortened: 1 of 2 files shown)


CI/bench/evaluate.py
```diff
@@ -188,17 +187,9 @@ def main():
...
-                        df1 = pd.read_csv(file_path, usecols=["time_train_s", "time_infer_s","rank"])
-
                     if file.startswith("AutoGluon") and file.endswith(".csv") and "pairwise" in root:
                         file_path = os.path.join(root, file)
-                        df2 = pd.read_csv(file_path, usecols=["framework", "Winrate"])
-                        df1 = df1.assign(Winrate=None, framework=None)
-                        df1["Winrate"] = df2["Winrate"]
-                        df1["framework"] = df2["framework"]
-                        df1 = df1.reindex(columns=["framework", "Winrate", "rank", "time_train_s", "time_infer_s"])
+                        df1 = pd.read_csv(file_path, usecols=["framework", "Winrate", "time_train_s", "time_infer_s","rank"])
```
prateekdesai04, line 192: @Innixma I still keep this in case we have a request to remove some columns/files in the future.

yinweisu:
| :---: | :---: |

CI/bench/evaluate.py
Innixma: Sounds good

Innixma (approved):
LGTM!

PR #4096, opened 2024-04-14
Optionally use Poppler/Pdf2Image for PDF Image Conversion

zkalson:
*Issue #, if available:*
[3800](https://github.com/autogluon/autogluon/issues/3800)

*Description of changes:*
Check if poppler is installed and in PATH. If so, get the DPI and convert the PDF with pdf2image.
Adds an optional dependency pdf2image, which can be installed in place of PyMuPDF.

### DIFF
--- a/multimodal/setup.py
+++ b/multimodal/setup.py
@@ -54,6 +54,7 @@ install_requires = [
     "tensorboard>=2.9,<3",
     "pytesseract>=0.3.9,<0.3.11",
     "nvidia-ml-py3==7.352.0",
+    "pdf2image>=1.17.0,<1.19",
 ]
 
 install_requires = ag.get_dependency_version_ranges(install_requires)
@@ -68,19 +69,7 @@ tests_require = [
     "tensorrt>=8.6.0,<8.6.2;platform_system=='Linux' and python_version<'3.11'",
 ]
 
-extras_require = {
-    "PyMuPDF": [
-        "PyMuPDF<=1.21.1",
-    ],
-    "pdf2image": [
-        "pdf2image<=1.17.0",
-    ],
-}
-
-for test_package in ["PyMuPDF", "pdf2image"]:
-    tests_require += extras_require[test_package]
-
-extras_require["tests"] = tests_require
+extras_require = {"tests": tests_require}
 
 
 if __name__ == "__main__":
--- a/multimodal/setup.py
+++ b/multimodal/setup.py
@@ -71,10 +71,13 @@ tests_require = [
 extras_require = {
     "PyMuPDF": [
         "PyMuPDF<=1.21.1",
-    ]
+    ],
+    "pdf2image": [
+        "pdf2image<=1.17.0",
+    ],
 }
 
-for test_package in ["PyMuPDF"]:
+for test_package in ["PyMuPDF", "pdf2image"]:
     tests_require += extras_require[test_package]
 
 extras_require["tests"] = tests_require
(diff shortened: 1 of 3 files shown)


multimodal/src/autogluon/multimodal/data/process_document.py
```diff
@@ -275,16 +279,56 @@ def process_one_sample(
             try:
```
tonyhoo, line 279: I think we can replace the fitz logics here totally with pdf2image as following
```
for per_col_name, per_col_image_features in document_features.items():
            try:
                # Process PDF documents.
                if feature_modalities[per_col_name] == DOCUMENT_PDF:
                    from pdf2image import convert_from_path

                    # Convert PDF to PIL images.
                    doc_images = convert_from_path(per_col_image_features[0])
                    if doc_images:
                        doc_image = doc_images[0].convert(image_mode)
                        words, normalized_word_boxes = self.get_ocr_features(per_col_image_features[0], doc_image)
                    else:
                        raise ValueError("Failed to convert PDF to images.")
                else:
                    # Process document image.
                    with PIL.Image.open(per_col_image_features[0]) as doc_image:
                        doc_image = doc_image.convert(image_mode)
                        words, normalized_word_boxes = self.get_ocr_features(per_col_image_features[0], doc_image)
            except ImportError as e:
                raise e
            except Exception as e:
                if self.missing_value_strategy.lower() == "zero":
                    logger.debug(f"Using a zero image due to '{e}'")
                    doc_image = PIL.Image.new(image_mode, (self.size, self.size), color=0)
                    doc_image = doc_image.convert(image_mode)
                    words = ""  # empty words
                    normalized_word_boxes = [self.pad_token_box]
                else:
                    raise e
 ```

File: multimodal/setup.py [hunk 2]
```diff
@@ -71,10 +71,13 @@
 extras_require = {
     "PyMuPDF": [
```
tonyhoo, line 72: We shall remove PyMuPDF from here

multimodal/setup.py
```diff
@@ -71,10 +71,13 @@
...
         "PyMuPDF<=1.21.1",
-    ]
+    ],
+    "pdf2image": [
```
tonyhoo, line 75: Considering the small size of the pdf2image package, is it possible to include it in the `install_requires` without causing significant churns?

multimodal/src/autogluon/multimodal/data/process_document.py
tonyhoo: Thanks Zain for your contribution. We highly appreciate it. However due to license incompatibility, removing the usage of fitz will be more appropriate instead of creating direct/in-direct linkage to it

yinweisu:
| :---: | :---: |
| safetensors==0.4.2 | safetensors==0.4.3 |
| safetensors==0.4.2 | safetensors==0.4.3 |

multimodal/src/autogluon/multimodal/data/process_document.py
tonyhoo: Thank you for your contributions and comments; they are greatly appreciated. Due to licensing inconsistencies, I removed `PyMuPDF` and transitioned to `pdf2image` implementation. Detailed instructions on how to install Poppler on various platforms have been provided for your convenience. I hope this helps.

yinweisu:
| :---: | :---: |

tonyhoo (approved):

suzhoum (approved):
Thanks for the fix! I'm curious if the performance will be impacted for document tasks.

PR #4099, opened 2024-04-15
[timeseries] Use different random seed for each model

shchur:
*Description of changes:*
- Increment the random seed used to train each consecutive model by 1. This fixes a weird regression for GluonTS models: when we use the same seed for all GluonTS models, their performance becomes correlated and sometimes leads to an overall performance regression. This likely happens because, when the seed is fixed, the GluonTS dataloaders returns batches in the same order.


yinweisu:
| :---: | :---: |

canerturkmen (approved):
LGTM! Thanks 🦸

PR #4137, opened 2024-04-24
[tabular] Add "auto" option for calibrate_decision_threshold

Innixma:
*Issue #, if available:*

*Description of changes:*

- Add "auto" option for calibrate_decision_threshold that sets to True for valid metrics in Binary Classification, otherwise False.
- Improve Scorer information on knowing if metrics are for regression or classification tasks (via the new `needs_class` property). This is important because while we technically can run metrics like root_mean_squared_error through decision_threshold_calibration and not crash, it would not lead to the intended functionality. With `needs_class`, we know that RMSE is not valid for threshold calibration and avoid it when `calibrate_decision_threshold="auto"`.
- In a future PR I intend to switch `calibrate_decision_threshold` to `"auto"` as the new default in TabularPredictor.

### DIFF
--- a/core/src/autogluon/core/metrics/__init__.py
+++ b/core/src/autogluon/core/metrics/__init__.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import json
 from abc import ABCMeta, abstractmethod
 from functools import partial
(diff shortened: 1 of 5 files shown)


rey-allan (approved):
Ship it!

yinweisu:
| :---: | :---: |
| botocore==1.34.91 | botocore==1.34.92 |
| boto3==1.34.91 | boto3==1.34.92 |
| ruff==0.4.1 | ruff==0.4.2 |
| botocore==1.34.91 | botocore==1.34.92 |
| boto3==1.34.91 | boto3==1.34.92 |
| ruff==0.4.1 | ruff==0.4.2 |

PR #4145, opened 2024-04-26
[tabular] Remove unnecessary duplicate call to `_validate_calibrate_decision_threshold`

rey-allan:
*Issue #, if available:*

*Description of changes:*
- We are validating the `calibrate_decision_threshold` inside `fit` (L974). And then calling the validation function again inside `_post_fit`.
- Since `fit` calls `_post_fit`, there's no need to call the validation function again.
- This won't improve much, but the duplicated call is unnecessary.


yinweisu:
| :---: | :---: |
| fastcore==1.5.29 | fastcore==1.5.31 |
| fastcore==1.5.29 | fastcore==1.5.31 |

Innixma (approved):
LGTM, good catch!

PR #4160, opened 2024-05-02
[timeseries] Ensure that prediction_length is stored as an integer

shchur:
*Issue #, if available:* Fixes #4012

*Description of changes:*
- Currently, predictor.fit() fails is `prediction_length` is provided as a `float` instead of `int`. This PR ensures that `prediction_length` is always stored as an `int`.


yinweisu:
| :---: | :---: |

AnirudhDagar (approved):
Thanks for the fix @shchur!

PR #4175, opened 2024-05-07
[timeseries] Fix tabular models sometimes failing because of a bug in preprocessing logic

shchur:
Tabular forecasting models would occasionally fail when a non-boolean known real covariate was mistakenly interpreted as a boolean covariate.

MWE:
```python
import pandas as pd
from autogluon.timeseries import TimeSeriesPredictor
N = 30
df = pd.DataFrame(
    {
        "item_id": ["A"] * N,
        "timestamp": pd.date_range("2020-01-01", freq="D", periods=N),
        "target": np.random.normal(size=N),
        "feat": np.tile([5, 0, 0], int(N/3)),
    }
)
predictor = TimeSeriesPredictor(known_covariates_names=["feat"]).fit(df, hyperparameters={"RecursiveTabular": {}})
```
This code will fail with the following exception
```
	Warning: Exception caused RecursiveTabular to fail during training... Skipping this model.
	"['__scaled_feat'] not in index"
```

This happens because during `fit()`, the feature `feat` is interpreted as non-boolean, so a scaled copy of the feature is added. At predict time, when transforming `known_covariates` (containing one row with only the `0` value), the feature was interpreted as boolean, so no scaled version of the feature was added. This results in the model failing at prediction time.

*Description of changes:*
- This PR fixes the above bug.

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
+++ b/timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
@@ -86,6 +86,7 @@ class AbstractMLForecastModel(AbstractTimeSeriesModel):
         self._residuals_std_per_item: Optional[pd.Series] = None
         self._avg_residuals_std: Optional[float] = None
         self._train_target_median: Optional[float] = None
+        self._non_boolean_real_covariates: List[str] = []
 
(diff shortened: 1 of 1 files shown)


yinweisu:
| :---: | :---: |
| tenacity==8.2.3 | tenacity==8.3.0 |
| tenacity==8.2.3 | tenacity==8.3.0 |

PR #4198, opened 2024-05-14
[timeseries] Use AutoGluon forks of Chronos models

shchur:
*Description of changes:*
- Use `autogluon/chronos-t5-*` path to load Chronos models

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/chronos/model.py
+++ b/timeseries/src/autogluon/timeseries/models/chronos/model.py
@@ -16,27 +16,27 @@ logger = logging.getLogger(__name__)
 
 # allowed HuggingFace model paths with custom parameter definitions
 MODEL_CONFIGS = {
-    "amazon/chronos-t5-tiny": {
+    "chronos-t5-tiny": {
         "num_gpus": 0,  # minimum number of required GPUs
         "default_torch_dtype": "auto",
         "default_batch_size": 16,
     },
-    "amazon/chronos-t5-mini": {
+    "chronos-t5-mini": {
         "num_gpus": 0,
         "default_torch_dtype": "auto",
         "default_batch_size": 16,
     },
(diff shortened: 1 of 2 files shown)


yinweisu:
| :---: | :---: |
| pyarrow==16.0.0 | pyarrow==16.1.0 |
| pyarrow==16.0.0 | pyarrow==16.1.0 |

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -185,9 +185,10 @@ def _is_gpu_available(self) -> bool:
     @property
     def ag_default_config(self) -> Dict[str, Any]:
         """The default configuration of the model used by AutoGluon if the model is one of those
-        defined in MODEL_CONFIGS. For now, these are ``amazon/chronos-t5-*`` family of models.
+        defined in MODEL_CONFIGS. For now, these are ``autogluon/chronos-t5-*`` family of models.
         """
-        return MODEL_CONFIGS.get(self.model_path, {})
+        model_name = str(self.model_path).split("/")[-1]
```
canerturkmen, line 190: if the intention is to make it work also on local paths, this will not work for Windows.

canerturkmen (approved):
One minor comment, looks good otherwise. Thanks!!

timeseries/src/autogluon/timeseries/models/chronos/model.py
shchur: The intention is to make this work with both `amazon/chronos-t5-*` and `autogluon/chronos-t5-*` paths, so should be fine.

canerturkmen: Then note that this will now include local paths on *nix that are named according to the amazon-t5-* convention, but not on windows.

PR #4203, opened 2024-05-15
[tabular, timeseries] Rename cached version file to `version.txt` from `__version__`

### LINKED ISSUE #4161: __version__ file not uploadable to Kaggle
[User Report](https://www.kaggle.com/competitions/playground-series-s4e5/discussion/499495#2789917)

When uploading a trained AutoGluon TabularPredictor artifact to Kaggle, the `__version__` file in the metadata is not able to be uploaded, because Kaggle for some unknown reason does not allow this specific file name to be uploaded. This results in a confusing warning message when a user tries to then load the predictor via `TabularPredictor.load`:

```
WARNING: Could not find version file at "/kaggle/input/test-ag/ag-20230301_033436/__version__".
This means that the predictor was fit in a version `<=0.3.1`.
```

To resolve, we can rename the file from `__version__` to `version.txt`, or some equivalent.

- [x] Ensure backwards compatibility, check for `__version__` if `version.txt` doesn't exist.
- [x] Check if the same problem exists for TimeSeries and Mulitmodal: Yes, fixed in #4203 for TimeSeries and Tabular. Multimodal does not check version on load.

Innixma:
*Issue #, if available:*
Resolves #4161 

*Description of changes:*

- Previously AutoGluon's TabularPredictor and TimeSeriesPredictor would save a file named `__version__` that contained the AutoGluon version as a string at the time of saving. However, Kaggle doesn't allow files named `__version__` to be uploaded for an unknown reason.
- This PR renames the file to `version.txt`, while keeping backwards compatibility.

### DIFF
--- a/tabular/src/autogluon/tabular/predictor/predictor.py
+++ b/tabular/src/autogluon/tabular/predictor/predictor.py
@@ -214,7 +214,7 @@ class TabularPredictor(TabularPredictorDeprecatedMixin):
 
     Dataset = TabularDataset
     predictor_file_name = "predictor.pkl"
-    _predictor_version_file_name = "__version__"
+    _predictor_version_file_name = "version.txt"
     _predictor_metadata_file_name = "metadata.json"
     _predictor_log_file_name = "predictor_log.txt"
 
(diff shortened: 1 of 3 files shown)


yinweisu:
| :---: | :---: |

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -4152,6 +4152,16 @@ def _load_version_file(cls, path) -> str:
         version = load_str.load(path=version_file_path)
         return version
 
+    @classmethod
+    def _load_version_file_old(cls, path) -> str:
```
shchur, line 4156: Maybe we could add the logic of trying both old & new filenames to `_load_version_file` instead of introducing a new method `_load_version_file_old`? I feel like the namespace of `Predictor` already contains quite a few methods.

Innixma: fair point, updated

yinweisu:
| :---: | :---: |
| botocore==1.34.106 | botocore==1.34.107 |
| boto3==1.34.106 | boto3==1.34.107 |
| botocore==1.34.106 | botocore==1.34.107 |
| boto3==1.34.106 | boto3==1.34.107 |

tonyhoo (approved):
LGTM!

shchur (approved):
LGTM, thanks!

PR #4210, opened 2024-05-21
Fix dynamic_stacking hanging with num_gpus>0

### LINKED ISSUE #3870: [BUG] Dynamic Stacking's Memory Save Sub-fit Runs Into a Deadlock for GPU resources (AutoGluon-Tabular).
**Bug Report Checklist**

- [ ] I provided code that demonstrates a minimal reproducible example.  **(unable to due to being very dependent on a local cluster setup with shared resources)**.  
- [X] I confirmed bug exists on the latest stable version of AutoGluon.  

**Describe the bug**
We should add a check to disable memory save sub-fits if we use GPUs.

Following a rather long debugging session on this, my current understanding is that Ray does not free GPU resources when using nested remote functions ([but only CPU resources](https://docs.ray.io/en/latest/ray-core/tasks/nested-tasks.html#yielding-resources-while-blocked)).

Assume you have 1 GPU. In the case of memory save sub-fits, we have first a remote function (A) that is the sub-fit and then a remote function (B) for fitting models in the case of a ParallelLocalFoldFittingStrategy.

Now, (A) will allocate the GPU by design. Then, within (A), the code tries to call (B), and if the model requires the GPU, this will result in a deadlock as (A) does not yield its allocated GPU resources.

Now, one could not pass GPU resources to (A), but then the sub-fit would crash if a model uses SequentialLocalFoldFittingStrategy and requires a GPU (as the sub-fit does not know a GPU exists).

**Expected behavior**
The GPU should be usable for a fitting method using `ParallelLocalFoldFittingStrategy`, and AutoGluon's Ray should not run into a deadlock.

### LINKED ISSUE #4196: [BUG] GPU training not working with non-default tabular presets
**Bug Report Checklist**

**Describe the bug**
Fitting the dataset using `num_gpus=1` only works when `presets='medium_quality'`. 
This might have been introduced by the recent `ray` version bump #3774 

In the ray dashboard I see these events, and the tasks are stuck: 
`Warning: The following resource request cannot be scheduled right now: {'CPU': 12.0, 'GPU': 0.5}. This is likely due to all cluster resources being claimed by actors. Consider creating fewer actors or adding more nodes to this Ray cluster.`

I only have 1 gpu, setting `num_gpus` to some value <1 seems to at least let the ray tasks run, albeit this doesn’t really work since `num_gpus>=1` is expected downstream

From what I can tell this issue is present on google colab too, so it should be easily reproducible.

**Expected behavior**
GPU training should work even for more complex tabular model presets

**To Reproduce**
Run [the tabular quickstart example notebook](https://github.com/autogluon/autogluon/blob/master/docs/tutorials/tabular/tabular-quick-start.ipynb) with `predictor = TabularPredictor(label=label).fit(train_data, num_gpus=1, presets='good_quality')` or better presets

**Installed Versions**
`autogluon==1.1.0`

Innixma:
*Issue #, if available:*

Resolves #4196 
Resolves #3870

*Description of changes:*

- Fix dynamic_stacking hanging with num_gpus>0
- Refer to the root cause here: https://docs.ray.io/en/latest/ray-core/tasks/nested-tasks.html#yielding-resources-while-blocked

TODO in follow-up PR: #4214 

Details:

Seems like nested ray jobs don't properly handle `num_gpus`.

For example:

```
sub_fit_caller = _ds_ray.remote(max_calls=1)(_sub_fit)
sub_fit_caller.options(num_cpus=num_cpus, num_gpus=1).remote(...)
```

Inside this `_sub_fit` call in ray, if we try to launch nested ray jobs, such as using `num_gpus=1` to fit a fold model, it will deadlock.

This is because the parent call already reserved the `num_gpus=1`, and thus there are no more GPU resources available to launch the nested job that requires GPU resources, so it will wait for the parent call to finish and release the GPU resources before the child call starts, which never happens.

However, `num_cpus` doesn't work like this, and it appears as if the `num_cpus` value of the parent for some reason is fully ignored by child calls. The parent can have `num_cpus=1` and the child can still run with `num_cpus=16` without issue.

Output of `ray status` during dynamic stacking subfit deadlock (parent num_gpus=1, num_cpus=32):

```
Resources
---------------------------------------------------------------
Usage:
 0.0/32.0 CPU
 1.0/1.0 GPU
 0B/74.92GiB memory
 482.93KiB/36.10GiB object_store_memory

Demands:
 {'CPU': 16.0, 'GPU': 0.5}: 2+ pending tasks/actors
```

### DIFF
--- a/tabular/src/autogluon/tabular/predictor/predictor.py
+++ b/tabular/src/autogluon/tabular/predictor/predictor.py
@@ -1351,7 +1351,11 @@ class TabularPredictor(TabularPredictorDeprecatedMixin):
 
                 # Call sub fit in its own subprocess via ray
                 sub_fit_caller = _ds_ray.remote(max_calls=1)(_sub_fit)
-                ref = sub_fit_caller.options(num_cpus=num_cpus, num_gpus=num_gpus).remote(
+                # FIXME: For some reason ray does not treat `num_cpus` and `num_gpus` the same.
+                #  For `num_gpus`, the process will reserve the capacity and is unable to share it to child ray processes, causing a deadlock.
+                #  For `num_cpus`, the value is completely ignored by children, and they can even use more num_cpus than the parent.
+                #  Because of this, num_gpus is set to 0 here to avoid a deadlock, but num_cpus does not need to be changed.
+                ref = sub_fit_caller.options(num_cpus=num_cpus, num_gpus=0).remote(
                     predictor=predictor_ref,
                     train_data=train_data_ref,
                     time_limit=time_limit,
--- a/tabular/src/autogluon/tabular/predictor/predictor.py
+++ b/tabular/src/autogluon/tabular/predictor/predictor.py
@@ -1355,6 +1355,7 @@ class TabularPredictor(TabularPredictorDeprecatedMixin):
                 #  For `num_gpus`, the process will reserve the capacity and is unable to share it to child ray processes, causing a deadlock.
                 #  For `num_cpus`, the value is completely ignored by children, and they can even use more num_cpus than the parent.
                 #  Because of this, num_gpus is set to 0 here to avoid a deadlock, but num_cpus does not need to be changed.
+                #  For more info, refer to Ray documentation: https://docs.ray.io/en/latest/ray-core/tasks/nested-tasks.html#yielding-resources-while-blocked
                 ref = sub_fit_caller.options(num_cpus=num_cpus, num_gpus=0).remote(
                     predictor=predictor_ref,
                     train_data=train_data_ref,


yinweisu:
| :---: | :---: |

LennartPurucker:
This is likely related: #3870

yinweisu:
| :---: | :---: |
| nvidia-nvjitlink-cu12==12.4.127 | nvidia-nvjitlink-cu12==12.5.40 |
| nvidia-nvjitlink-cu12==12.4.127 | nvidia-nvjitlink-cu12==12.5.40 |

rey-allan (approved):
LGTM. Such a weird behavior by Ray!

PR #4224, opened 2024-05-23
[tabular] Fix RuntimeWarning in LinearModel

Innixma:
*Issue #, if available:*

*Description of changes:*

- Fix RuntimeWarning in LinearModel
- From what I can tell the RuntimeWarning is harmless, I can't find a situation where it is doing something incorrectly.
- @LennartPurucker's explanation below provides a good hypothesis for what is going on. Because of this, I will opt to silence the warning instead of reimplementing the logic.

RuntimeWarning:

```
/opt/conda/envs/ag-111-310/lib/python3.10/site-packages/pandas/core/arrays/categorical.py:1664: RuntimeWarning: invalid value encountered in cast
  return np.asarray(ret, dtype)
```

### DIFF
--- a/features/src/autogluon/features/generators/one_hot_encoder.py
+++ b/features/src/autogluon/features/generators/one_hot_encoder.py
@@ -9,6 +9,7 @@ from sklearn.preprocessing import OneHotEncoder
 from autogluon.common.features.types import R_CATEGORY, R_INT, S_BOOL, S_SPARSE
 
 from .abstract import AbstractFeatureGenerator
+from .label_encoder import LabelEncoderFeatureGenerator
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


yinweisu:
| :---: | :---: |

rey-allan (approved):
LGTM!

LennartPurucker:
For documentation purposes: this seems to happen because  `.to_numpy` first transforms the data to the desired type and then fills nan values. 

The RuntimeWarning is triggered when we transform data with `nan` values to `unit8` as `nan` values are represented as float values that do not exist in `uint8`.

This is not a problem because, after the faulty conversion, all nan value indices of the original array (not the transformed one) are set to the specified `na_value`.

LennartPurucker:
My workaround for now is this:


```python
def pd_to_np(self, X: DataFrame) -> np.ndarray:
    with warnings.catch_warnings():
        if np.issubdtype(self._dtype, np.integer):
            warnings.filterwarnings("ignore", category=RuntimeWarning)
        return X.to_numpy(dtype=self._dtype, na_value=self.fillna_val, copy=True)
```

Innixma:
@LennartPurucker Great explanation! I will move forward with your code snippet instead of my prior reimplementation for simplicity.

yinweisu:
| :---: | :---: |

PR #4256, opened 2024-06-11
update sagemaker base images to use pytorch 2.2

suzhoum:
*Issue #, if available:*

*Description of changes:*
The original pytorch2.1 sagemaker base images installs `torchaudio` 2.1.x, however,  AutoGluon master installs `torch` 2.2.x will introduce incompatibility issues. Updating the sagemaker base image version would resolve this issue.

### DIFF
--- a/CI/docker/Dockerfile.cpu-inference
+++ b/CI/docker/Dockerfile.cpu-inference
@@ -1,4 +1,4 @@
-FROM 763104351884.dkr.ecr.us-east-1.amazonaws.com/pytorch-inference:2.1.0-cpu-py310-ubuntu20.04-sagemaker
+FROM 763104351884.dkr.ecr.us-east-1.amazonaws.com/pytorch-inference:2.2.0-cpu-py310-ubuntu20.04-sagemaker
 
 RUN apt-get update \
  && apt-get -y upgrade \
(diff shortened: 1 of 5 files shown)


yinweisu:
| :---: | :---: |
| pydantic==2.7.3 | pydantic==2.7.4 |
| pydantic==2.7.3 | pydantic==2.7.4 |

CI/docker/Dockerfile.cpu-inference
prateekdesai04, line 1: does torchaudio get installed along with torch?
or is it separate?

suzhoum: `torchaudio` is included in the sagemaker base image, previously we had the torch 2.1 base image, so everything related to torch installs 2.1, but since our master is already installing torch 2.2 by default, and we don't use torchaudio, torchaudio stays at 2.1.x. Somehow it's causing issues when I ran the tests in AutoGluon Cloud https://github.com/autogluon/autogluon-cloud/actions/runs/9456879067/job/26091603325

prateekdesai04: got it, so how is it resolved just by updating to torch 2.2 ?

suzhoum: The sagemaker torch 2.2 base image would align with AG's torch version 2.2 and installs torchaudio 2.2.x, so that torchaudio would also be compatible with other torch modules.

prateekdesai04 (approved):

PR #4277, opened 2024-06-17
[tabular] Validate CPU/GPU settings

Innixma:
*Issue #, if available:*

*Description of changes:*

Adds additional checks to ensure CPU and GPU settings are valid.

- raise TypeError if num_cpus and num_gpus are invalid types.
- raise ValueError if num_cpus<1
- raise ValueError if num_gpus<0
- Fixes a cryptic failure caused when passing `num_cpus` as a float that could be converted to an int such as `12.0`, which will fail the process but not provide a useful error message to understand the root cause. Now it will fail immediately stating that `num_cpus` must be an int.


yinweisu:
| :---: | :---: |

rey-allan (approved):
LGTM!

PR #4278, opened 2024-06-18
ChronosAugmentedGeneration: Two steps chronos with covariates

ikvision:
It is a first attempt to extend the foundation model chronos from univariate to univariate with past_covariates.

*Issue #, if available:*

*Description of changes:*
Extending chronos to use covariates input time series: ChronosAugmentedGeneration where the we augment the raw features with the learn representation of chornos.
It is zero-shot embedding the each time series and train an LGBM regressor on top of those embeddings

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/chronos_augmented_generation/model.py
+++ b/timeseries/src/autogluon/timeseries/models/chronos_augmented_generation/model.py
@@ -1,12 +1,12 @@
+import logging
 import os
 from typing import Any, Dict, Literal, Optional, Union
 
-import logging
 import numpy as np
 import pandas as pd
-from sklearn.multioutput import MultiOutputRegressor
-from lightgbm import LGBMRegressor
 import torch
+from lightgbm import LGBMRegressor
+from sklearn.multioutput import MultiOutputRegressor
 
 from autogluon.tabular.models.lgb.lgb_model import LGBModel
 from autogluon.timeseries import TimeSeriesDataFrame
@@ -27,18 +27,17 @@ class ChronosAugmentedGeneration(AbstractTimeSeriesModel):
     default_num_samples: int = 20
     default_model_path = "autogluon/chronos-t5-mini"
     default_torch_dtype = "bfloat16"
-    default_device ='cpu'
-
+    default_device = "cpu"
 
     maximum_context_length = 512
     default_batch_size = 8
     default_use_raw_features = True
 
     lgbm_default_params = {
-            "n_estimators": 300,
-            "n_jobs": -1,
-            "random_state": 0,
-        }
+        "n_estimators": 300,
+        "n_jobs": -1,
+        "random_state": 0,
+    }
 
     def __init__(
         self,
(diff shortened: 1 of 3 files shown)


timeseries/src/autogluon/timeseries/models/chronos_augmented_generation/model.py
```diff
@@ -0,0 +1,228 @@
+import logging
```
shchur, line 1: @ikvision I've linted the code and added the model to unittests in `test_models.py` to make sure that the tests are running

ikvision (commented):
Thank you so. much for adding it to the test_models. It will help me test is in a better wya

timeseries/src/autogluon/timeseries/models/chronos_augmented_generation/model.py
```diff
@@ -0,0 +1,228 @@
...
+        )
+        self.time_limit: Optional[float] = None
+
+        self.embedding_model = ChronosPipeline.from_pretrained(
+            self.model_path, device_map=hyperparameters["device_map"]
+        )
+
+        lgbm_params = kwargs.get("lgbm_params", self.lgbm_default_params)
+
+        lgbm = LGBModel()  # TODO: this is the correct lgbm to use in autogluon, get train_data as input
```
shchur, line 98: I think the `LGBModel` from AutoGluon expects `X`, `y` as input ([code](https://github.com/ikvision/autogluon/blob/620d49d2f36a0c61453790eedad9460331643eab/tabular/src/autogluon/tabular/models/lgb/lgb_model.py#L97)), so the API should be very similar to `LGBMRegressor`.

timeseries/src/autogluon/timeseries/models/chronos_augmented_generation/model.py
```diff
@@ -0,0 +1,228 @@
...
+        num_gpus: Optional[int] = None,
+        verbosity: int = 2,
+        **kwargs,
+    ) -> None:
+        print("Entering the `_fit` method")
+        self.time_limit = time_limit
+
+        # train_data - long dataframe
+        # X_train = train_data.slice_by_timestep(None, -self.prediction_length)
+        X = self._preprocess(train_data, is_train=True)
```
shchur, line 174: For me to give advice on implementation, I would need to understand what the expected shape of `X` and `y` should be for the model. For example, assuming `train_data` contains `N` time series, each with length `T` and the forecast horizon is `H`, what will be the shape of `X` and `y` as a function of `N`, `T`, `H`?

If I understand the current code correctly, `X` has shape `[N, D]` (where `D = chronos_embedding_dim + ...`) and `y` has shape `[N, H]`. In other words, there is only a single training example for each time series. Is this the correct intention?

jobs-git:
This would be great! Any progress @ikvision @shchur? Will this automatically cover `chronos-bolt` models?

shchur:
Hi @jobs-git, I'm not sure if currently planning to proceed with this PR. We might add an option to extract embeddings from Chronos models and, potentially, use them as features for some other models. However, this likely won't happen before 1.4.

jobs-git:
> Hi @jobs-git, I'm not sure if currently planning to proceed with this PR. We might add an option to extract embeddings from Chronos models and, potentially, use them as features for some other models. However, this likely won't happen before 1.4.

I see, how about past-co-variate feature similar to:

[TemporalFusionTransformerModel](https://auto.gluon.ai/dev/tutorials/timeseries/forecasting-model-zoo.html#autogluon.timeseries.models.TemporalFusionTransformerModel)

I saw somewhere that there are efforts to add these types of features (i.e. ) to other models too

shchur:
It seems that we won't be proceeding with this PR, but we might revisit the idea of using Chronos embeddings as inputs to a forecasting model at some point. Closing for now.

PR #4290, opened 2024-06-24
[tabular] Fix Stacker max_models logic

Innixma:
*Issue #, if available:*

*Description of changes:*

- Fix a long standing major bug where L2+ stacker models that are not weighted ensembles will not respect the `max_models` and `max_models_per_type` parameters, and will instead use all base models during fit.
- This bug has been present forever, since the original code was written, but only started majorly impacting AutoGluon in v1.0+ with the introduction of Zeroshot-HPO.
- Fixing this should lead to faster training and inference times for stack layers, and a potential quality improvement.

Results: 3x inference speedup with no noticeable model quality drop:

pr4290_boxplot

TODO:

- [x] Benchmark `best` and `high` quality presets.
- [x] Consider changing defaults for `max_models` and `max_models_per_type`. Currently they are 25 and 5 respectively. Likely `max_models` should be increased to ~50.
- [x] Add additional documentation for `get_feature_metadata` in Trainer.
- [x] Add advanced unit tests.

Follow-up PRs:

- [ ] Potentially delete StackerEnsembleModel, move logic to Trainer.
- [ ] Add option for user to specify WeightedEnsemble max_models and max_models_per_type
- [ ] Remove `base_models_dict` parameter.
- [x] Bug: `sample_weight` column is present in `X` in the `model.fit` call, it should not be present. (Created issue: https://github.com/autogluon/autogluon/issues/4304)
- [x] Bug: `sample_weight` column is passed to `feature_prune` if both sample_weight and feature_prune are enabled. This shouldn't happen. (Created issue: https://github.com/autogluon/autogluon/issues/4304)

### DIFF
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -457,7 +457,7 @@ class AbstractModel:
             feature_metadata = self._infer_feature_metadata(X=X)
         else:
             feature_metadata = copy.deepcopy(feature_metadata)
-        feature_metadata = self._update_feature_metadata(feature_metadata=feature_metadata)
+        feature_metadata = self._update_feature_metadata(X=X, feature_metadata=feature_metadata)
         get_features_kwargs = self.params_aux.get("get_features_kwargs", None)
         if get_features_kwargs is not None:
             valid_features = feature_metadata.get_features(**get_features_kwargs)
@@ -505,7 +505,7 @@ class AbstractModel:
         if error_if_no_features and not self._features_internal:
             raise NoValidFeatures
 
-    def _update_feature_metadata(self, feature_metadata: FeatureMetadata) -> FeatureMetadata:
+    def _update_feature_metadata(self, X: pd.DataFrame, feature_metadata: FeatureMetadata) -> FeatureMetadata:
         """
         [Advanced] Method that performs updates to feature_metadata during initialization.
         Primarily present for use in stacker models.
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -454,9 +454,10 @@ class AbstractModel:
             self.features = list(X.columns)
         # TODO: Consider changing how this works or where it is done
         if feature_metadata is None:
-            feature_metadata = FeatureMetadata.from_df(X)
+            feature_metadata = self._infer_feature_metadata(X=X)
         else:
             feature_metadata = copy.deepcopy(feature_metadata)
+        feature_metadata = self._update_feature_metadata(feature_metadata=feature_metadata)
         get_features_kwargs = self.params_aux.get("get_features_kwargs", None)
         if get_features_kwargs is not None:
             valid_features = feature_metadata.get_features(**get_features_kwargs)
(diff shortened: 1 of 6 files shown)


prateekdesai04:
/benchmark module=tabular preset=tabular_best benchmark=tabular_full time_limit=1h

prateekdesai04:
/benchmark module=tabular preset=tabular_best benchmark=tabular_full time_limit=1h

prateekdesai04:
/benchmark module=tabular preset=tabular_best benchmark=tabular_full time_limit=1h

prateekdesai04:
/benchmark module=tabular preset=tabular_best benchmark=tabular_full time_limit=4h

prateekdesai04:
/benchmark module=tabular preset=tabular_best benchmark=tabular_full time_limit=4h

Innixma:
Benchmarks look very good, results in around a 3x mean inference speedup with no drop in model quality. 
pr4290_boxplot

core/src/autogluon/core/models/ensemble/stacker_ensemble_model.py
```diff
@@ -88,7 +94,49 @@ def _initialize(self, **kwargs):
...
+        elif num_rows < 35000:
+            max_models_per_type = 8
+        elif num_rows < 40000:
+            max_models_per_type = 9
+        elif num_rows < 45000:
+            max_models_per_type = 10
+        elif num_rows < 50000:
+            max_models_per_type = 11
+        else:
+            max_models_per_type = 12
```
prateekdesai04, line 125: How did we arrive at 12?
Also, what is the reason behind changing `max_models_per_type` based on the no. of rows in the data?

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -562,11 +573,11 @@ def _initialize(self, X=None, y=None, feature_metadata=None, num_classes=None, *
 
         self._init_misc(X=X, y=y, feature_metadata=feature_metadata, num_classes=num_classes, **kwargs)
 
+        self._init_params()
```
prateekdesai04, line 576: qq: why did we move it here?

prateekdesai04 (approved):
Left a few comments, overall LGTM!

core/src/autogluon/core/models/ensemble/stacker_ensemble_model.py
Innixma: 12 is an arbitrary stopping point, but the intuition is that 12 should be close to approximating uncapped, and will behave much much better than if it is left uncapped if a user uses HPO and fits 1000s of models as an example.

This logic is not optimal, but it is better than a fixed value. We can iterate on the exact thresholds and values in future PRs.

The reason for changing the value depending on num_rows is that doing this shows improvement over a fixed value on TabRepo simulations. The intuition is that larger `max_models_per_type` leads to more overfitting, and you can counter overfitting by having more data.

core/src/autogluon/core/models/abstract/abstract_model.py
Innixma: Good question!

- In `self._preprocess_set_features` I added logic in this PR which updates `self.features to reflect which base_models are being used. Previously `self.features` wasn't correct and contained unused base models.
- In order to know which `base_models` are being used, I need to run the logic which filters base_models based on `max_models` and `max_models_per_type`.
- In order to run the filtering logic, I first need to know the hyperparameter values of `max_models` and `max_models_per_type`.
- In order to know those hyperparameter values, I need the hyperparameters initialized.
- `self._init_params()` initializes the parameters so I can use them in `self._preprocess_set_features`.

PR #4294, opened 2024-06-25
Replace ray-default with necessary submodules

suzhoum:
*Issue #, if available:*

*Description of changes:*

`ray-default` installs `NodeJS` as a dependency while we are not really making use of the UI component of Ray in AutoGluon. By removing `NodeJS`, we could help reduce AutoGluon's surface area in services which has AG as a dependency. 

### DIFF
--- a/core/setup.py
+++ b/core/setup.py
@@ -48,10 +48,10 @@ install_requires = (
 
 extras_require = {
     "ray": [
-        "ray[default]>=2.10.0,<2.11",  # sync with common/src/autogluon/common/utils/try_import.py
+        "ray[air,core,train]>=2.10.0,<2.11",  # sync with common/src/autogluon/common/utils/try_import.py
     ],
     "raytune": [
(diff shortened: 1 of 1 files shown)


Innixma:
Can you provide a description for why this change is being made?

suzhoum:
> Can you provide a description for why this change is being made?

I have updated the description for the PR.

suzhoum:
It looks like there is no way to get around `ray-default` in conda build since one of our required submodules ray-tune ( https://github.com/autogluon/autogluon/blob/ea7b4fba7695a5a09deafb10fd8d034fb2265916/core/src/autogluon/core/hpo/ray_hpo.py#L15would) requires ray-default (https://github.com/conda-forge/ray-packages-feedstock/blob/94f217b781ea4c1c606c169f9e6eb47cdeb0ff27/recipe/meta.yaml#L256)

PR #4302, opened 2024-06-27
[TEST] Ray upgrade

suzhoum:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/core/setup.py
+++ b/core/setup.py
@@ -48,10 +48,10 @@ install_requires = (
 
 extras_require = {
     "ray": [
-        "ray[default]>=2.10.0,<2.11",  # sync with common/src/autogluon/common/utils/try_import.py
+        "ray[default]>=2.10.0,<2.40",  # sync with common/src/autogluon/common/utils/try_import.py
     ],
     "raytune": [
-        "ray[default,tune]>=2.10.0,<2.11",  # sync with common/src/autogluon/common/utils/try_import.py
+        "ray[default,tune]>=2.10.0,<2.40",  # sync with common/src/autogluon/common/utils/try_import.py
         # TODO: consider alternatives as hyperopt is not actively maintained.
         "hyperopt>=0.2.7,<0.2.8",  # This is needed for the bayes search to work.
         # 'GPy>=1.10.0,<1.11.0'  # TODO: Enable this once PBT/PB2 are supported by ray lightning


suzhoum:
/benchmark module=tabular preset=tabular_medium benchmark=tabular_full time_limit=1h

suzhoum:
/benchmark module=tabular preset=tabular_medium benchmark=tabular_full time_limit=1h

common/src/autogluon/common/utils/try_import.py
```diff
@@ -30,7 +30,7 @@ def try_import_mxboard():
 
 
 def try_import_ray() -> ModuleType:
-    RAY_MAX_VERSION = "2.11.0"  # sync with core/setup.py
+    RAY_MAX_VERSION = "2.40.0"  # sync with core/setup.py
```
Innixma, line 33: 2.31 is the latest ray version, so this should be changed to "2.32"

core/setup.py
```diff
@@ -48,10 +48,10 @@
 
 extras_require = {
     "ray": [
-        "ray[default]>=2.10.0,<2.11",  # sync with common/src/autogluon/common/utils/try_import.py
+        "ray[default]>=2.10.0,<2.40",  # sync with common/src/autogluon/common/utils/try_import.py
```
Innixma, line 51: 2.31 is the latest ray version, so this should be changed to "2.32"

core/setup.py
```diff
@@ -48,10 +48,10 @@
...
     ],
     "raytune": [
-        "ray[default,tune]>=2.10.0,<2.11",  # sync with common/src/autogluon/common/utils/try_import.py
+        "ray[default,tune]>=2.10.0,<2.40",  # sync with common/src/autogluon/common/utils/try_import.py
```
Innixma, line 54: 2.31 is the latest ray version, so this should be changed to "2.32"

common/src/autogluon/common/utils/try_import.py
Innixma, line 33: `tabular_medium` in AG-Bench I assume means `medium_quality`. `medium_quality` does not use Ray. We need to test `best_quality` which uses Ray.

suzhoum: That's good to know. Let me run a benchmark with `best_quality`

Innixma (approved):
LGTM, thanks for running the benchmark!

PR #4311, opened 2024-07-03
Update infer_types.py to say the default handling is zero

zkalson:
*Description of changes:*

According to the documentation (https://auto.gluon.ai/stable/tutorials/multimodal/advanced_topics/customization.html#data-image-missing-value-strategy) and process_image.py (autogluon/multimodal/src/autogluon/multimodal/data
/process_image.py), Autogluon uses an image with zero pixels if an image is missing. There is a warning in infer_types.py (multimodal/src/autogluon/multimodal/data/infer_types.py) that says the default behavior is to skip samples, which I updated.


Innixma (approved):

PR #4327, opened 2024-07-18
[tabular] Add initial callbacks support

Innixma:
*Issue #, if available:*

*Description of changes:*

- Add initial callbacks support to TabularPredictor
- Callbacks allow users to inject custom logic into the training process, and theoretically allow the user to completely override the training logic with their own custom logic.

Example Code:

```
import pandas as pd

from autogluon.core.callbacks import EarlyStoppingCallback


if __name__ == '__main__':
    label = 'class'
    train_data = pd.read_csv('https://autogluon.s3.amazonaws.com/datasets/Inc/train.csv')

    callbacks = [EarlyStoppingCallback(patience=3)]
    predictor = TabularPredictor(label=label).fit(train_data=train_data, callbacks=callbacks)
```

Example Output:

```
...
User-specified callbacks (1): ['EarlyStoppingCallback']
Fitting 13 L1 models ...
Fitting model: KNeighborsUnif ...
	0.7736	 = Validation score   (accuracy)
	1.63s	 = Training   runtime
	0.02s	 = Validation runtime
EarlyStoppingCallback: Best Score: 0.7736 | Patience: 0/3 | Best Model: KNeighborsUnif (New Best)
Fitting model: KNeighborsDist ...
	0.7652	 = Validation score   (accuracy)
	0.2s	 = Training   runtime
	0.01s	 = Validation runtime
EarlyStoppingCallback: Best Score: 0.7736 | Patience: 1/3 | Best Model: KNeighborsUnif
Fitting model: LightGBMXT ...
	0.8792	 = Validation score   (accuracy)
	1.74s	 = Training   runtime
	0.0s	 = Validation runtime
EarlyStoppingCallback: Best Score: 0.8792 | Patience: 0/3 | Best Model: LightGBMXT (New Best)
Fitting model: LightGBM ...
	0.8824	 = Validation score   (accuracy)
	1.37s	 = Training   runtime
	0.0s	 = Validation runtime
EarlyStoppingCallback: Best Score: 0.8824 | Patience: 0/3 | Best Model: LightGBM (New Best)
Fitting model: RandomForestGini ...
	0.8612	 = Validation score   (accuracy)
	0.91s	 = Training   runtime
	0.08s	 = Validation runtime
EarlyStoppingCallback: Best Score: 0.8824 | Patience: 1/3 | Best Model: LightGBM
Fitting model: RandomForestEntr ...
	0.8584	 = Validation score   (accuracy)
	1.0s	 = Training   runtime
	0.09s	 = Validation runtime
EarlyStoppingCallback: Best Score: 0.8824 | Patience: 2/3 | Best Model: LightGBM
Fitting model: CatBoost ...
	0.8824	 = Validation score   (accuracy)
	6.89s	 = Training   runtime
	0.01s	 = Validation runtime
EarlyStoppingCallback: Best Score: 0.8824 | Patience: 3/3 | Best Model: LightGBM
EarlyStoppingCallback: Early stopping trainer fit for level=1. Reason: No score_val improvement in the past 3 models.
	Ensemble Weights: {'CatBoost': 0.5, 'LightGBM': 0.45, 'KNeighborsUnif': 0.05}
	0.886	 = Validation score   (accuracy)
	0.06s	 = Training   runtime
	0.0s	 = Validation runtime
EarlyStoppingCallback: Best Score: 0.8860 | Patience: 0/3 | Best Model: WeightedEnsemble_L2 (New Best)
AutoGluon training complete, total runtime = 15.16s ... Best model: WeightedEnsemble_L2 | Estimated inference throughput: 69200.6 rows/s (2500 batch size)
TabularPredictor saved. To load, use: predictor = TabularPredictor.load("AutogluonModels/ag-20240730_021147")
```

TODO:
- [x] Add PR description
- [x] Add example callback
- [x] Add unit tests
- [x] Add docstrings
- [x] Add tutorial: Moved to a future PR, tracking issue: #4351

### DIFF
--- a/core/src/autogluon/core/callbacks/_abstract_callback.py
+++ b/core/src/autogluon/core/callbacks/_abstract_callback.py
@@ -11,7 +11,19 @@ class AbstractCallback(object, metaclass=ABCMeta):
     Abstract callback class for AutoGluon's TabularPredictor.
     The inner API and logic within `trainer` is considered private API. It may change without warning between releases.
 
-    # TODO: Docstring for init args
+    Parameters
+    ----------
+    allow_recursive_calls : bool, default = False
+        If True, will allow recursive calls to this callback.
+        For example, a recursive call can happen if inside `self.before_fit` or `self.after_fit`, the callback initiates a model fit in trainer.
+        This model fit will then trigger `self.before_fit` again, which could lead to an infinite loop if the callback is not implemented carefully.
+        If False, guarantees that the callback logic will be skipped if it is part of a recursive call.
+    skip_if_trainer_stopped : bool, default = False
+        If True, will skip self.before_fit and self.after_fit logic if `early_stop=True` was returned from any callback,
+        indicating that the trainer is stopping training.
+        This matters if you have 2 or more callbacks, and the first callback returns `early_stop=True`.
+        If the second callback has `skip_if_trainer_stopped=True`, it will skip its callback logic.
+        Otherwise, its callback logic will still trigger.
 
     Examples
     --------
--- a/core/src/autogluon/core/callbacks/_abstract_callback.py
+++ b/core/src/autogluon/core/callbacks/_abstract_callback.py
@@ -17,11 +17,11 @@ class AbstractCallback(object, metaclass=ABCMeta):
     ----------
     allow_recursive_calls : bool, default = False
         If True, will allow recursive calls to this callback.
-        For example, a recursive call can happen if inside `self.before_fit` or `self.after_fit`, the callback initiates a model fit in trainer.
-        This model fit will then trigger `self.before_fit` again, which could lead to an infinite loop if the callback is not implemented carefully.
+        For example, a recursive call can happen if inside `self.before_model_fit` or `self.after_model_fit`, the callback initiates a model fit in trainer.
+        This model fit will then trigger `self.before_model_fit` again, which could lead to an infinite loop if the callback is not implemented carefully.
         If False, guarantees that the callback logic will be skipped if it is part of a recursive call.
     skip_if_trainer_stopped : bool, default = False
-        If True, will skip self.before_fit and self.after_fit logic if `early_stop=True` was returned from any callback,
+        If True, will skip self.before_model_fit and self.after_model_fit logic if `early_stop=True` was returned from any callback,
         indicating that the trainer is stopping training.
         This matters if you have 2 or more callbacks, and the first callback returns `early_stop=True`.
         If the second callback has `skip_if_trainer_stopped=True`, it will skip its callback logic.
--- a/core/src/autogluon/core/callbacks/_abstract_callback.py
+++ b/core/src/autogluon/core/callbacks/_abstract_callback.py
@@ -11,6 +11,8 @@ class AbstractCallback(object, metaclass=ABCMeta):
     Abstract callback class for AutoGluon's TabularPredictor.
     The inner API and logic within `trainer` is considered private API. It may change without warning between releases.
 
+    # TODO: Docstring for init args
+
     Examples
     --------
     >>> from autogluon.core.callbacks import ExampleCallback
@@ -20,8 +22,17 @@ class AbstractCallback(object, metaclass=ABCMeta):
     >>> label = 'class'
     >>> predictor = TabularPredictor(label=label).fit(train_data, callbacks=callbacks)
     """
+    def __init__(
+        self,
+        allow_recursive_calls: bool = False,
+        skip_if_trainer_stopped: bool = False,
+    ):
+        assert isinstance(allow_recursive_calls, bool)
+        assert isinstance(skip_if_trainer_stopped, bool)
+        self.allow_recursive_calls = allow_recursive_calls
+        self.skip_if_trainer_stopped = skip_if_trainer_stopped
+        self._skip = False
 
-    @abstractmethod
     def before_fit(
         self,
         trainer: AbstractTrainer,
@@ -65,9 +76,34 @@ class AbstractCallback(object, metaclass=ABCMeta):
             if False, the trainer continues with its normal logic.
             Ignored if `early_stop=True`.
         """
-        raise NotImplementedError
+        if self._skip or (self.skip_if_trainer_stopped and trainer._callback_early_stop):
+            return False, False
+        if not self.allow_recursive_calls:
+            self._skip = True
+        early_stop, skip_model = self._before_fit(
+            trainer=trainer,
+            model=model,
+            logger=logger,
+            time_limit=time_limit,
+            stack_name=stack_name,
+            level=level,
+        )
+        if not self.allow_recursive_calls:
+            self._skip = False
+        return early_stop, skip_model
 
     @abstractmethod
+    def _before_fit(
+        self,
+        trainer: AbstractTrainer,
+        model: AbstractModel,
+        logger: Logger,
+        time_limit: float | None = None,
+        stack_name: str = "core",
+        level: int = 1,
+    ) -> Tuple[bool, bool]:
+        raise NotImplementedError
+
     def after_fit(
         self,
         trainer: AbstractTrainer,
@@ -108,4 +144,28 @@ class AbstractCallback(object, metaclass=ABCMeta):
             If True, the trainer stops training additional models and ends the fit process immediately.
             If False, the trainer continues with its normal logic.
         """
+        if self._skip or (self.skip_if_trainer_stopped and trainer._callback_early_stop):
+            return False
+        if not self.allow_recursive_calls:
+            self._skip = True
+        early_stop = self._after_fit(
+            trainer=trainer,
+            model_names=model_names,
+            logger=logger,
+            stack_name=stack_name,
+            level=level,
+        )
+        if not self.allow_recursive_calls:
+            self._skip = False
+        return early_stop
+
+    @abstractmethod
+    def _after_fit(
+        self,
+        trainer: AbstractTrainer,
+        model_names: List[str],
+        logger: Logger,
+        stack_name: str = "core",
+        level: int = 1,
+    ) -> bool:
         raise NotImplementedError
--- a/core/src/autogluon/core/callbacks/_abstract_callback.py
+++ b/core/src/autogluon/core/callbacks/_abstract_callback.py
@@ -1,11 +1,13 @@
-from abc import ABCMeta, abstractmethod
-from logging import Logger
-from typing import List, Tuple
+from __future__ import annotations
+
+from abc import ABCMeta
 
 from ..models import AbstractModel
 from ..trainer import AbstractTrainer
 
 
+# TODO: Open design questions:
+#  1. Should trainer be a class variable for ease of access?
 class AbstractCallback(object, metaclass=ABCMeta):
     """
     Abstract callback class for AutoGluon's TabularPredictor.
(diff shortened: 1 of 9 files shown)


core/src/autogluon/core/trainer/abstract_trainer.py
```diff
@@ -2335,9 +2390,58 @@ def _train_single_full(
...
+        for callback in self.callbacks:
+            callback_early_stop = callback.after_fit(
+                self,
+                model_names=model_names,
+                logger=logger,
+                stack_name=stack_name,
+                level=level,
+            )
+            if callback_early_stop:
+                self._callback_early_stop = True
```
eddiebergman, line 2443: If there's no side effect to care about with the `.after_fit()`, i.e. you don't care about callbacks state after this call, then you could break early.

```python
    def _callbacks_after_fit(
        self,
        *,
        model_names: List[str],
        stack_name: str,
        level: int,
    ):
        for callback in self.callbacks:
            callback_early_stop = callback.after_fit(
                self,
                model_names=model_names,
                logger=logger,
                stack_name=stack_name,
                level=level,
            )
            if callback_early_stop:
                self._callback_early_stop = True
                break
```

Or if you're feeling functional ;)

```python
    def _callbacks_after_fit(
        self,
        *,
        model_names: List[str],
        stack_name: str,
        level: int,
    ):
        should_stop_itr = (
            callback.after_fit(
                self,
                model_names=model_names,
                logger=logger,
                stack_name=stack_name,
                level=level,
            )
            for callback in self.callbacks
        )
        self._callback_early_stop = any(should_stop_itr)
```

eddiebergman (commented):
Sorry for the unsolicited PR review, was just curious to see how callbacks get implemented!

core/src/autogluon/core/callbacks/_example_callback.py
```diff
@@ -0,0 +1,60 @@
...
+        model: AbstractModel,
+        logger: Logger,
+        time_limit: float | None = None,
+        stack_name: str = "core",
+        level: int = 1,
+        **kwargs,
+    ) -> Tuple[bool, bool]:
+        time_limit_trainer = trainer._time_limit
+        if time_limit_trainer is not None and trainer._time_train_start is not None:
+            time_left_total = time_limit_trainer - (time.time() - trainer._time_train_start)
```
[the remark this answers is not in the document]
Innixma: Nice find! #4342

Innixma:
@eddiebergman unsolicited PR reviews are among my favorite kinds of code reviews :)

core/src/autogluon/core/trainer/abstract_trainer.py
Innixma: My thought process is that some callbacks could simply be logging related, and I wouldn't want a callback that early stops to prevent a logging callback from doing its logging.

The other thing is technically the later callbacks can check if `trainer._callback_early_stop == True` and skip their logic.

This is a good idea though, and I thought about it too while implementing. I've added the following `self.skip_if_trainer_stopped` parameter to callbacks. When true, the callback logic will be skipped when `trainer._callback_early_stop == True`. This should give the best of both worlds.

eddiebergman: Sounds good! If you want to make it clearer to users, you could make a shallow class `LoggingCallback`, or conversely, `EarlyStoppingCallback` which just internally sets this flag, i.e. user never sets it. I use the pattern a good bit where the flag is actually set as a class variables, e.g. `LoggingCallback` should not be skipped and `EarlyStoppingCallback` can be stopped.

Gives you the freedom to later change this to a dict or things, or whatever, as long as `LoggingCallback` is never skipped and `EarlyStoppimhCallback` is. Also one less parameter for someone to think about

core/src/autogluon/core/trainer/abstract_trainer.py
```diff
@@ -243,6 +244,9 @@ def __init__(
 
         # self._exceptions_list = []  # TODO: Keep exceptions list for debugging during benchmarking.
 
+        self.callbacks: List[callable] = []
```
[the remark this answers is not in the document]
Innixma: Good catch! I updated the PR and I created a GitHub issue to track replacement throughout the project: #4349

core/src/autogluon/core/trainer/abstract_trainer.py
```diff
@@ -389,6 +393,7 @@ def train_multi_levels(
         level_time_modifier=0.333,
         infer_limit=None,
         infer_limit_batch_size=None,
+        callbacks: List[callable] = None,
```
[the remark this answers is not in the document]
Innixma: ditto

core/src/autogluon/core/callbacks/_early_stopping_callback.py
```diff
@@ -0,0 +1,77 @@
...
+            if self.last_improvement == 0:
+                msg += " (New Best)"
+            self._log(logger, 20, msg=msg)
+        if self.verbose and early_stop:
+            msg = f"Stopping trainer fit due to callback early stopping. Reason: No score_val improvement in the past {self.last_improvement} models."
+            self._log(logger, 20, msg=msg)
+        return early_stop
+
+    def _calc_new_best(self, trainer: AbstractTrainer):
+        leaderboard = trainer.leaderboard()
```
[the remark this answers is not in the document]
Innixma: Good call! I actually needed to revamp this logic to make it work in cases where the user specified `infer_limit` to ensure that the model returned satisfies the `infer_limit`.

core/src/autogluon/core/callbacks/_abstract_callback.py
```diff
@@ -0,0 +1,111 @@
...
+    @abstractmethod
+    def before_fit(
+        self,
+        trainer: AbstractTrainer,
+        model: AbstractModel,
+        logger: Logger,
+        time_limit: float | None = None,
+        stack_name: str = "core",
+        level: int = 1,
+    ) -> Tuple[bool, bool]:
```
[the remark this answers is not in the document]
Innixma: Nice! Updated

core/src/autogluon/core/trainer/abstract_trainer.py
```diff
@@ -656,6 +693,8 @@ def stack_new_level_core(
         If self.bagged_mode, then models will be trained as StackerEnsembleModels.
         The data provided in this method should not contain stack features, as they will be automatically generated if necessary.
         """
+        if self._callback_early_stop:
+            return []
```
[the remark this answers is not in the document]
Innixma: Interesting idea, will probably keep as is for now and add overwrite logic if it becomes useful.

core/src/autogluon/core/trainer/abstract_trainer.py
```diff
@@ -470,10 +477,40 @@ def train_multi_levels(
             model_names_fit += base_model_names + aux_models
         if self.model_best is None and len(model_names_fit) != 0:
             self.model_best = self.get_model_best(can_infer=True, infer_limit=infer_limit, infer_limit_as_child=True)
-        self._time_limit = None
+        self._fit_cleanup()
```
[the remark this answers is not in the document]
Innixma: This is an interesting idea. I think for now I will keep it as is, but this could be something worth adopting in future. I've made essentially zero use of `yield` in AutoGluon so far, mostly due to my unfamiliarity.

core/src/autogluon/core/trainer/abstract_trainer.py
Innixma: That is a great idea, updated the code, I think this is a good practice I should adopt more often to separate user-level args with developer level args (aka those that are only relevant to specify when subclassing)

core/src/autogluon/core/callbacks/_abstract_callback.py
```diff
@@ -0,0 +1,205 @@
...
+        Example usages of this method include logging a final summary of the training, saving information to disk, or executing additional post-fit logic.
+        """
+        pass
+
+    def before_model_fit(
+        self,
+        trainer: AbstractTrainer,
+        model: AbstractModel,
+        time_limit: float | None = None,
+        stack_name: str = "core",
```
shchur, line 77: Should this be a `typing.Literal`?

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -709,6 +711,21 @@ def fit(
             setting the memory limit (and any other resources) on systems with shared resources that are controlled by the operating system (e.g., SLURM and
             cgroups). Otherwise, AutoGluon might wrongly assume more resources are available for fitting a model than the operating system allows,
             which can result in model training failing or being very inefficient.
+        callbacks : List[AbstractCallback], default = None
+            [Experimental] Callback support is preliminary, targeted towards developers, and is subject to change.
```
shchur, line 715: Nit: Should we use [admonitions](https://myst-parser.readthedocs.io/en/latest/syntax/admonitions.html) to highlight experimental features?

```
:::{warning}
This is an experimental feature and may change in the future releases without warning.
:::
```

core/src/autogluon/core/trainer/abstract_trainer.py
```diff
@@ -468,12 +521,51 @@ def train_multi_levels(
...
+        """
+        Prepare the trainer state at the start of / prior to a fit call.
+        Should be paired with a `self._fit_cleanup()` at the conclusion of the fit call.
+        """
+        self._time_train_start = time.time()
+        self._time_train_start_last = self._time_train_start
+        self._time_limit = time_limit
+        self.reset_callbacks()
+        if callbacks is not None:
+            assert isinstance(callbacks, list), f"`callbacks` must be a list. Found invalid type: `{type(callbacks)}`."
```
shchur, line 541: Does it make sense to verify that callbacks are of type `AbstractCallback`?

core/src/autogluon/core/trainer/abstract_trainer.py
```diff
@@ -243,6 +249,9 @@ def __init__(
 
         # self._exceptions_list = []  # TODO: Keep exceptions list for debugging during benchmarking.
 
+        self.callbacks: List[AbstractCallback] = []
+        self._callback_early_stop = False
```
shchur, line 253: I'm not very familiar with the rest of the `AbstractTrainer` code, so this might be a bad suggestion, but would it be feasible to communicate the early stopping / interruption without altering the state of the trainer? I can imagine that in some scenarios such as distributed training it will be really tricky to reason about the state of this variable.

core/src/autogluon/core/callbacks/_abstract_callback.py
Innixma: in this case no because stack_name can be any valid string.

core/src/autogluon/core/trainer/abstract_trainer.py
Innixma: Tough to say, but I think I would prefer to edit the trainer state, as the intention is to not pass the trainer object to worker threads, and therefore it remains a singular source of truth. We can change it later if we find this has limitations.

core/src/autogluon/core/trainer/abstract_trainer.py
Innixma: Good idea! Added

tabular/src/autogluon/tabular/predictor/predictor.py
Innixma: Nice! Added

core/src/autogluon/core/trainer/abstract_trainer.py
```diff
@@ -457,12 +507,55 @@ def train_multi_levels(
                 additional_full_weighted_ensemble=additional_full_weighted_ensemble,
             )
             model_names_fit += base_model_names + aux_models
-        if self.model_best is None and len(model_names_fit) != 0:
+        if (self.model_best is None or infer_limit is not None) and len(model_names_fit) != 0:
```
prateekdesai04, line 510: qq: could you explain this check ?

prateekdesai04 (approved):

core/src/autogluon/core/trainer/abstract_trainer.py
Innixma: Pseudocode breakdown of the check:

```
if (trainer has not specified which model is best) and (any model exists):
    select the best valid model based on validation score
elif (user specified infer limit) and (any model exists):
    select the best valid model based on validation score that satisfies the infer limit
```

The added infer_limit check ensures that we always return a model that satisfies the infer_limit constraints. Sometimes, a non-ensemble model can get a better score while satisfying infer limit than an ensemble model, and by default only ensemble models are set to `self.model_best` prior to this call. This call ensures that the best model will be picked, even if it isn't an ensemble.

PR #4333, opened 2024-07-22
[Proposal] Add Support for Custom Memory (Soft) Limit

LennartPurucker:
Heyho,

This proposal PR allows users to set a custom memory limit (RAM, not VRAM). 

I have not found a prior issue or PR related to this, but I might have missed it. 
Feel free to link any previous discussions. 

### Context

So far, AutoGluon assumes that the memory limit equals all available memory on the system. AutoGluon uses this memory limit at several points in the program flow to control the fitting behavior at runtime. 

This includes (but is likely not limited to): whether to fit certain models at all (based on estimating their required memory) or early stopping models that would go over the memory limit when fitting them further. 

### Problem & Impact

If AutoGluon is used in an environment with shared resources (e.g., SLURM), or if one would like to avoid fitting very memory-expensive models, then the current behavior is "bugged" or suboptimal, respectively.

From studying some papers that have used AutoGluon, I also suspect several researchers fell into this trap and used AutoGluon incorrectly. To illustrate, in the case of SLURM, this leads to AutoGluon trying to fit models that require more memory than allocated for the SLURM job, resulting in the fit process being killed by SLURM. AutoGluon does not expect this process to be killed, so it cannot just stop fitting early or skip the model in the first place. 

Thus, this is quite an important feature we should add and communicate to the user (just like the num_cpus or num_gpus parameters). Moreover, I would love to use this as an official feature instead of my monkey patch workaround, which I have been using for a while now (see below).

### Proposed Solution

I opted to use an environment variable in the current format of the resource manager (which is all static).
If we want to make this a parameter like num_cpus or num_gpus, we need to make the resource manager dynamic, requiring many changes. Or add the custom memory limit to all places that require it. 

### Mokey Ptach Workaround for Custom Memory Limit 

To monkey patch the memory limit (e.g., when running on SLURM), call the following function before calling any AutoGluon code or importing AutoGluon. Essentially, call this function before doing anything else in your script. Check the output of AutoGluon to see if the patch worked.

Kudos to @eddiebergman for cleaning up my monkey patch code and turning it into a usable function! 

```python
"""Due to AutoGluon's use of psutil to infer memory limit,
we need to patch psutil to return the required memory limit.

This method is easier than patching AutoGluon's internals
and psutil is more likely to be stable.

The use of this code will enforce AutoGluon's memory usage
but may impact any other tools relying on psutil's memory
information.

The use of this function should not be used in production and
is mostly required for systems in which there is a process
enforced memory limit that is not directly accessible to
psutil, i.e. SLURM.
"""

import warnings

import psutil


def patch_memory_limit(gb: float) -> None:
    """Patch psutil to enforce a maximum memory limit when AutoGluon reads memory
    available.

    Args:
        gb (float): The maximum memory limit in GB.
    """
    psutil._virtual_memory = psutil.virtual_memory  # type: ignore
    MAX_MEMORY_BYTES = int(gb * (1024.0**3))

    def patch_virtual_memory():
        r = psutil._virtual_memory()  # type: ignore
        p = psutil.Process()
        usable_memory = MAX_MEMORY_BYTES - p.memory_info().rss
        return r._replace(
            available=usable_memory,
            free=usable_memory,
            total=MAX_MEMORY_BYTES,
            percent=round(
                (MAX_MEMORY_BYTES - usable_memory) / MAX_MEMORY_BYTES * 100,
                1,
            ),
        )

    psutil.virtual_memory = patch_virtual_memory
    warnings.warn(
        f"Psutil's Max Memory is enforced to be {gb} GB. "
        "This might be higher than the actual available memory, depending on the"
        " machine/SLURM setup.",
        stacklevel=2,
    )
```
--- 


Innixma:
Very nice! Is there any way that SLURM clusters traditionally deal with this for other packages that have memory checks? Is there any way AutoGluon could automatically pick up that there are limited resources? If not then I would guess this is the best solution.

eddiebergman:
Re AutoSklearn:

There method there was just to have it be user specified and then use [`pynisher`](https://github.com/automl/pynisher) to set process memory limits. No support for this method [on Mac](https://github.com/automl/pynisher?tab=readme-ov-file#features) though :(

Innixma:
@eddiebergman 

The concept of a wrapper for memory handling is very interesting:

```
model_trainer = limit(
    train_memory_hungry_model,
    memory=(500, "MB"),
    wall_time=(1.5, "h")  # 1h30m
)
```
Couple questions:

1. Does this correctly capture memory usage for models such as LightGBM / CatBoost that are written in C with their own multiprocessing logic?
2. Does it work for ray worker threads?
3. Can I use a variant of this logic to store the peak memory usage of a function call?

common/src/autogluon/common/utils/resource_utils.py
```diff
@@ -159,9 +159,22 @@ def _get_gpu_count_cuda():
         cudaShutdown()
         return gpu_count
 
+    @staticmethod
+    def _get_custom_memory_size():
+        memory_limit = float(os.environ.get("AG_MEMORY_LIMIT_IN_GB"))
```
Innixma, line 164: Would it make sense for `TabularPredictor.fit` or `init` to have a `max_memory` argument that sets `os.environ["AG_MEMORY_LIMIT_IN_GB"] = str(max_memory)` internally?

common/src/autogluon/common/utils/resource_utils.py
```diff
@@ -159,9 +159,22 @@ def _get_gpu_count_cuda():
...
+
+        if memory_limit <= 0:
+            raise ValueError("Memory set via `AG_MEMORY_LIMIT_IN_GB` must be greater than 0!")
+
+        # Transform to bytes and return
+        return max(int(memory_limit * (1024.0**3)), 1)
```
Innixma, line 170: Note: If AutoGluon's memory estimates are off for a model, and the model uses more memory than it is supposed to, this could still lead to memory over-allocation in the SLURM cluster. Unsure if there is a good work-around for this besides "don't incorrectly estimate memory usage", maybe the wrapper logic that errors a function if using too much memory via `pynisher`, but I suspect that has a performance tradeoff.

Innixma: I wonder if Ray has any logic when creating a worker to limit memory usage of the worker? Couldn't find anything from a quick google search.

For example, Ray kills workers if >95% memory usage, but probably would ignore the memory specified in this PR's patch: https://docs.ray.io/en/latest/ray-core/scheduling/ray-oom-prevention.html

common/tests/unittests/test_memory_limit.py
```diff
@@ -0,0 +1,21 @@
+def test_memory_mocking():
+    import psutil
+    from autogluon.common.utils.resource_utils import ResourceManager
+
+    assert psutil.virtual_memory().total == ResourceManager._get_memory_size()
+    assert psutil.virtual_memory().available == ResourceManager._get_available_virtual_mem()
```
Innixma, line 6: I imagine this unit test could randomly fail if total/available memory changes between psutil calls.

common/tests/unittests/test_memory_limit.py
```diff
@@ -0,0 +1,21 @@
+def test_memory_mocking():
```
Innixma, line 1: linting is failing. Can you run black/isort?

common/tests/unittests/test_memory_limit.py
```diff
@@ -0,0 +1,21 @@
...
+    assert psutil.virtual_memory().available == ResourceManager._get_available_virtual_mem()
+
+
+def test_custom_memory_soft_limit():
+    import os
+
+    import psutil
+
+    p = psutil.Process()
+    os.environ["AG_MEMORY_LIMIT_IN_GB"] = "48"
```
Innixma, line 15: Does this only work if done prior to AutoGluon import?

common/tests/unittests/test_memory_limit.py
```diff
@@ -0,0 +1,21 @@
...
+    from autogluon.common.utils.resource_utils import ResourceManager
+
+    assert ResourceManager.get_memory_size(format="GB") == 48
+    assert (48 * (1024.0**3)) - p.memory_info().rss == ResourceManager.get_available_virtual_mem(format="B")
+
+    del os.environ["AG_MEMORY_LIMIT_IN_GB"]
```
Innixma, line 21: Maybe best to ensure this is called even if assert fails?

common/tests/unittests/test_memory_limit.py
```diff
@@ -0,0 +1,21 @@
...
+    import os
+
+    import psutil
+
+    p = psutil.Process()
+    os.environ["AG_MEMORY_LIMIT_IN_GB"] = "48"
+    from autogluon.common.utils.resource_utils import ResourceManager
+
+    assert ResourceManager.get_memory_size(format="GB") == 48
+    assert (48 * (1024.0**3)) - p.memory_info().rss == ResourceManager.get_available_virtual_mem(format="B")
```
Innixma, line 19: Ditto that theoretically value returned from `p.memory_info().rss` could change between the two calls, leading to a failed assert.

Innixma (commented):
Generally looks very nice! Added some comments for further improvement / test cleanup / linting

common/src/autogluon/common/utils/resource_utils.py
LennartPurucker: Jup, that would be more clean. I will look into it.

common/tests/unittests/test_memory_limit.py
LennartPurucker: No, this should always work because. Fixed the order.

LennartPurucker:
> Very nice! Is there any way that SLURM clusters traditionally deal with this for other packages that have memory checks (Ex: Auto-sklearn)? Is there any way AutoGluon could automatically pick up that there are limited resources? If not then I would guess this is the best solution.

Typically, SLURM workloads are not machine learning tasks but instead other high-performance computing applications. If they are (GPU) machine learning tasks, they tend to use an entire node (i.e., making it non-shared). In these settings, memory checks, tracking, and allocations are usually baked into the system (due to being high-performance jobs).

I am unaware of a general framework that supports this nicely regarding memory. Most people ignore these kinds of problems. It should be possible in Linux, but there is nothing available for Python AFAIK. 

If you can use a virtualized run on SLURM (via a resources-limited container), you could fake it so that psutil and similar checks only see as many resources as you would give the container. Unfortunately, resources-limited containers are often not supported by SLURM deployments (due to red tape, not technically limitations).

common/src/autogluon/common/utils/resource_utils.py
LennartPurucker: That case would also error/kill a normal run without SLURM. Which is then caught, and the model is simply disregarded as failed (as only the fitting process is killed if the ray is used). 

One can adjust Ray's soft and hard limits. However, we would likely need to monkey patch its memory estimations via psutil again. I would be interested to see how far one could allocate memory with Ray jobs, too. In theory, it could have the same API as CPUs/GPUs, but I am not aware that it is supported so far.

common/tests/unittests/test_memory_limit.py
LennartPurucker: Proposing a fix based on a moneky patch in updated code.

common/tests/unittests/test_memory_limit.py
LennartPurucker: Fixed.

common/tests/unittests/test_memory_limit.py
LennartPurucker: Added a potential workaround.

eddiebergman:
> @eddiebergman
> 
> The concept of a wrapper for memory handling is very interesting:
> 
> ```
> model_trainer = limit(
>     train_memory_hungry_model,
>     memory=(500, "MB"),
>     wall_time=(1.5, "h")  # 1h30m
> )
> ```
> 
> Couple questions:
> 
>     1. Does this correctly capture memory usage for models such as LightGBM / CatBoost that are written in C with their own multiprocessing logic?
> 
>     2. Does it work for ray worker threads?
> 
>     3. Can I use a variant of this logic to store the peak memory usage of a function call? This would be awesome for unit tests / benchmarking / logging (Although I'd guess it comes with a performance hit due to subprocess usage)

Just to give some more details on how it works:

Wrap everything into a subprocess, and the first thing the wrapper will do once launched inside the subprocess is set RLIMIT's (Linux) or communicate to `pywin32` to set actual memory limits on the process. Note this isn't a polling system, it's a syscall that literally prevents more allocation than a certain size. For example, `np.ones(BIG_BIG_NUMBER)` will just not be allowed allocate if it's over the limit. Once this occurs, Python throws an error and the wrapper uses pipes between the main subprocesses and worker subprocess to communicate what happened (could also be a timeout or a more general user/code error).

There's some issues:
* If an import causes it to go over the memory limit, you get an `ImportError` and not a `Memory` based error, which is hard to track that something when wrong.
* The memory limiting on Linux is weird, from what I gathered, the only memory you could limit for a child process is one that includes both its resident memory (that which only the worker has access to) **+** the memory it shares with the main process from `fork` or `fork-server` (the memory not yet copied over to the worker but will be copied as soon as the worker accesses it).
* Linux has one way of doing things, Windows has another, and Mac memory-limiting and signals is a dumpster-fire which is better left burning.

Now to the main points:
1. It should, it uses system's memory for that process, nothing Python specific other than how the syscall is made. (I think it keeps track of children memory too, would need to double check)
2. The way to approach this would be to send work to the ray worker/thread and have it launch the wrapped function. You could also send over the wrapped function but I'm not sure how pickling works there. In general though, since `pynisher` will essentially launch a child process just for that function call, it should be fairly okay to just send the wrapped function to the worker and it will handle it there. I'm not a user of Ray but if the ray worker can not spawn subprocesses, you may have issues.
3. There might be a way where you can set a limit and a signal handler to catch the over memory use, ignore the signal so the process can continue and track the memory somewhere. However I was just using this with polling: https://github.com/automl/pynisher/blob/608ae2f7a102b399d28d7643a7263c6cdff7af10/pynisher/util.py#L218-L240

eddiebergman:
You might also want to check out this thread: https://github.com/giampaolo/psutil/issues/1096

common/src/autogluon/common/utils/resource_utils.py
LennartPurucker: I just proposed a solution that still fits the PR

Innixma (approved):
LGTM. This is an awesome feature add!

Innixma:
@eddiebergman Thanks for the detailed reply! I think for now the solution in the PR works for solving the overall problem, although doing direct limiting of process memory would be very interesting as a potential future improvement to AutoGluon's overall resilience to crashes.

Notably, the ability to track overall memory usage (over time and peak) including all child processes will be very useful for stress testing the system, finding memory leaks, and seeing the impact of optimizations. Will hopefully circle back on this later when we find bandwidth.

eddiebergman:
Happy to chime in where I can :)

I would imagine Ray has some better functionality for this built-in, given it has some C/C++ in there (I think).

If not, all of this information is actually easier to track with conventional tools on Linux for process monitoring given Python is just a C program. Note, **tracking**, trying to do memory limiting in Python is usually a patchwork of syscalls on the interpreters process then a thing Python can manage itself in Python land. 

If I come across a tool for general process monitoring, I'll let you know!

PR #4352, opened 2024-07-30
[tabular] Fix file path for RF ONNX

Innixma:
*Issue #, if available:*

*Description of changes:*

Previously, the ONNX compilation code would crash if RandomForest's model directory was not yet created. This didn't trigger in normal usage, but it does if we want to compile prior to saving the original model. This PR fixes this by ensuring the parent directory is created.


prateekdesai04 (approved):
LGTM!

PR #4353, opened 2024-07-30
[tabular] Remove sklearnex RF logic that raises exception

Innixma:
*Issue #, if available:*

*Description of changes:*

- Remove sklearnex RF logic that raises exception
- Previously this code worked and halved the disk usage / memory usage of the RandomForest model.
- Now the code crashes, due to a change at some point in sklearnex. Therefore, we have to remove the code.


prateekdesai04 (approved):
LGTM!

PR #4354, opened 2024-07-30
[tabular] Add `ag.compile` parameter to models

Innixma:
*Issue #, if available:*

*Description of changes:*

- Add `ag.compile` parameter to models
- This allows the user to specify model compilation to occur directly after the model is fit, rather than in a separate `predictor.compile` call.

### DIFF
--- a/tabular/tests/unittests/models/test_tabular_nn.py
+++ b/tabular/tests/unittests/models/test_tabular_nn.py
@@ -1,4 +1,4 @@
-import pytest
+import shutil
 
 from autogluon.tabular.models.tabular_nn.torch.tabular_nn_torch import TabularNeuralNetTorchModel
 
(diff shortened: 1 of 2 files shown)


tabular/tests/unittests/models/test_tabular_nn.py
```diff
@@ -51,6 +51,19 @@ def test_tabular_nn_binary_compile_onnx(fit_helper):
...
 
+def test_tabular_nn_binary_compile_onnx_as_ag_arg(fit_helper):
+    fit_args = dict(
+        hyperparameters={TabularNeuralNetTorchModel: {"ag.compile": {"compiler": "onnx"}}},
+    )
+    dataset_name = "adult"
+    predictor = fit_helper.fit_and_validate_dataset(dataset_name=dataset_name, fit_args=fit_args, refit_full=True, delete_directory=False)
+    from autogluon.tabular.models.tabular_nn.compilers.onnx import TabularNeuralNetTorchOnnxTransformer
+
+    assert isinstance(predictor._learner.trainer.load_model("NeuralNetTorch").processor, TabularNeuralNetTorchOnnxTransformer)
```
prateekdesai04, line 62: QQ: Just wanted to understand the difference here between `NeuralNetTorch` and `NeuralNetTorch_FULL`

prateekdesai04 (approved):
LGTM

tabular/tests/unittests/models/test_tabular_nn.py
Innixma: NeuralNetTorch is the original model fit with train and validation data.

NeuralNetTorch_FULL is the refit model (via `refit_full=True`) which combines the train and validation data together into a combined train data and fits the model again. The unit test checks that compilation works correctly for both the original and _FULL model.

PR #4366, opened 2024-08-06
[core] Add `flake8` to the `[test]` requirements in `core/setup.py`

nathanaelbosch:
*Issue #, if available:*
Tests in `core` failed due to a missing `flake8` dependency.

*Description of changes:*
I added `flake8` to the test requirements in `core/setup.py`.

### DIFF
--- a/core/setup.py
+++ b/core/setup.py
@@ -63,6 +63,7 @@ tests_require = [
     "types-requests",
     "types-setuptools",
     "pytest-mypy",
+    "flake8",
 ]
 
 all_requires = []


core/setup.py
```diff
@@ -63,6 +63,7 @@
     "types-requests",
     "types-setuptools",
     "pytest-mypy",
+    "flake8",
```
shchur, line 66: It seems strange that the tests from `core/tests/test_check_style.py` that require `flake8` weren't triggered / didn't fail in our CI. I think that we need to update our CI config to make sure that the tests are triggered and, maybe, switch from flake8 to ruff (#3353) while we are at it.

shchur (approved):
LGTM!

PR #4367, opened 2024-08-06
[core] Speed up `EnsembleSelection._fit`

nathanaelbosch:
*Issue #, if available:*
The current implementation of the greedy ensembling is unnecessarily slowed down, especially for large ensemble_sizes, as the `weighted_ensemble_prediction` computation here
https://github.com/autogluon/autogluon/blob/f5c7b7400596e1449fad2b4107fde55809a09bac/core/src/autogluon/core/models/greedy_ensemble/ensemble_selection.py#L113-L125
is done in every iteration of the loop from scratch. This could be done on a rolling basis instead, which would reduce the complexity of this function from quadratic in the ensemble size to linear.

*Description of changes:*
This PR changes the current implementation to instead accumulate into the `ensemble_prediction` at each iteration. On a small local example I see a decrease in the ensembling time from 1.4s to 0.9s for the standard `ensemble_size=100`, and for `ensemble_size=1000` I see a decrease from 66s to 8.4s.

### DIFF
--- a/core/src/autogluon/core/models/greedy_ensemble/ensemble_selection.py
+++ b/core/src/autogluon/core/models/greedy_ensemble/ensemble_selection.py
@@ -110,19 +110,12 @@ class EnsembleSelection(AbstractWeightedEnsemble):
         round_scores = False
         epsilon = 1e-4
         round_decimals = 6
+        ensemble_prediction = np.zeros(predictions[0].shape)
         for i in range(ensemble_size):
             scores = np.zeros((len(predictions)))
             s = len(ensemble)
-            if s == 0:
-                weighted_ensemble_prediction = np.zeros(predictions[0].shape)
-            else:
-                # Memory-efficient averaging!
-                ensemble_prediction = np.zeros(ensemble[0].shape)
-                for pred in ensemble:
-                    ensemble_prediction += pred
-                ensemble_prediction /= s
 
-                weighted_ensemble_prediction = (s / float(s + 1)) * ensemble_prediction
+            weighted_ensemble_prediction = (s / float(s + 1)) * ensemble_prediction
             fant_ensemble_prediction = np.zeros(weighted_ensemble_prediction.shape)
             for j, pred in enumerate(predictions):
                 fant_ensemble_prediction[:] = weighted_ensemble_prediction + (1.0 / float(s + 1)) * pred
@@ -175,6 +168,10 @@ class EnsembleSelection(AbstractWeightedEnsemble):
             trajectory.append(best_score)
             order.append(best)
             used_models.add(best)
+
+            ensemble_prediction *= s
+            ensemble_prediction += ensemble[-1]
+            ensemble_prediction /= s+1
 
             # Handle special case
             if len(predictions) == 1:
--- a/core/src/autogluon/core/models/greedy_ensemble/ensemble_selection.py
+++ b/core/src/autogluon/core/models/greedy_ensemble/ensemble_selection.py
@@ -111,17 +111,21 @@ class EnsembleSelection(AbstractWeightedEnsemble):
         epsilon = 1e-4
         round_decimals = 6
         ensemble_prediction = np.zeros(predictions[0].shape)
+        weighted_ensemble_prediction = np.zeros(predictions[0].shape)
+        fant_ensemble_prediction = np.zeros(predictions[0].shape)
         for i in range(ensemble_size):
             scores = np.zeros((len(predictions)))
             s = len(ensemble)
 
-            weighted_ensemble_prediction = (s / float(s + 1)) * ensemble_prediction
-            fant_ensemble_prediction = np.zeros(weighted_ensemble_prediction.shape)
+            if s != 0:
+                ensemble_prediction *= (s - 1) / s
+                ensemble_prediction += ensemble[-1] / s
+                weighted_ensemble_prediction[:] = (s / float(s + 1)) * ensemble_prediction
             for j, pred in enumerate(predictions):
                 fant_ensemble_prediction[:] = weighted_ensemble_prediction + (1.0 / float(s + 1)) * pred
                 if self.problem_type in ["multiclass", "softclass"]:
                     # Renormalize
-                    fant_ensemble_prediction[:] = fant_ensemble_prediction / fant_ensemble_prediction.sum(axis=1)[:, np.newaxis]
+                    fant_ensemble_prediction /= fant_ensemble_prediction.sum(axis=1)[:, np.newaxis]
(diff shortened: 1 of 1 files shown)


core/src/autogluon/core/models/greedy_ensemble/ensemble_selection.py
```diff
@@ -176,6 +169,10 @@ def _fit(self, predictions: List[np.ndarray], labels: np.ndarray, time_limit=Non
             order.append(best)
             used_models.add(best)
 
+            ensemble_prediction *= s
+            ensemble_prediction += ensemble[-1]
+            ensemble_prediction /= s+1
```
shchur, line 174: I wonder if this may lead to loss of numerical precision when we multiple by a large number `s` and then divide it back by `s+1`. Is this code equivalent to the following?
```python
ensemble_prediction =  ensemble_prediction * s / (s + 1) + ensemble[-1] / (s + 1)
```

core/src/autogluon/core/models/greedy_ensemble/ensemble_selection.py
```diff
@@ -110,19 +110,12 @@ def _fit(self, predictions: List[np.ndarray], labels: np.ndarray, time_limit=Non
         round_scores = False
```
shchur, line 110: The lint check is currently failing, can you please run `black` on this PR? https://github.com/autogluon/autogluon/actions/runs/10262339335/job/28391811858?pr=4367

core/src/autogluon/core/models/greedy_ensemble/ensemble_selection.py
nathanaelbosch: Good point, yes this should be equivalent. I will replace this part with your code.

core/src/autogluon/core/models/greedy_ensemble/ensemble_selection.py
nathanaelbosch: Should also be fixed now.

core/src/autogluon/core/models/greedy_ensemble/ensemble_selection.py
```diff
@@ -176,6 +169,8 @@ def _fit(self, predictions: List[np.ndarray], labels: np.ndarray, time_limit=Non
             order.append(best)
             used_models.add(best)
 
+            ensemble_prediction = ensemble_prediction * (s / (s + 1)) + ensemble[-1] / (s + 1)
```
nathanaelbosch, line 172: Just because it was present in the code from before, maybe explicit casting as `float` is needed here too?
```suggestion
            ensemble_prediction = ensemble_prediction * (s / float(s + 1)) + ensemble[-1] / float(s + 1)
```

shchur: This should be fine, in Python >=3 integer division always results in a float.

core/src/autogluon/core/models/greedy_ensemble/ensemble_selection.py
```diff
@@ -110,25 +110,22 @@ def _fit(self, predictions: List[np.ndarray], labels: np.ndarray, time_limit=Non
         round_scores = False
         epsilon = 1e-4
         round_decimals = 6
+        ensemble_prediction = np.zeros(predictions[0].shape)
+        weighted_ensemble_prediction = np.zeros(predictions[0].shape)
+        fant_ensemble_prediction = np.zeros(predictions[0].shape)
```
Innixma, line 115: I updated the code a bit to minimize the redundant np.ndarray creations and use in-place operations whenever possible.

core/src/autogluon/core/models/greedy_ensemble/ensemble_selection.py
```diff
@@ -110,25 +110,22 @@ def _fit(self, predictions: List[np.ndarray], labels: np.ndarray, time_limit=Non
...
-                    ensemble_prediction += pred
-                ensemble_prediction /= s
-
-                weighted_ensemble_prediction = (s / float(s + 1)) * ensemble_prediction
-            fant_ensemble_prediction = np.zeros(weighted_ensemble_prediction.shape)
+
+            if s != 0:
+                ensemble_prediction *= (s - 1) / s
+                ensemble_prediction += ensemble[-1] / s
+                weighted_ensemble_prediction[:] = (s / float(s + 1)) * ensemble_prediction
```
Innixma, line 123: I moved the `ensemble_prediction` logic to before `weighted_ensemble_prediction` is updated to avoid one unused `ensemble_prediction` operation on the final iteration of the for loop.

core/src/autogluon/core/models/greedy_ensemble/ensemble_selection.py
```diff
@@ -110,25 +110,22 @@ def _fit(self, predictions: List[np.ndarray], labels: np.ndarray, time_limit=Non
...
-                for pred in ensemble:
-                    ensemble_prediction += pred
-                ensemble_prediction /= s
-
-                weighted_ensemble_prediction = (s / float(s + 1)) * ensemble_prediction
-            fant_ensemble_prediction = np.zeros(weighted_ensemble_prediction.shape)
+
+            if s != 0:
+                ensemble_prediction *= (s - 1) / s
+                ensemble_prediction += ensemble[-1] / s
```
Innixma, line 122: Using in-place operations since it is slightly faster and less memory usage

Innixma (approved):
Excellent work and good eye to catch this!

Indeed this is a major speedup for large ensemble_size values, and probably is a big deal for large multiclass datasets with few model trainings and a fast metric (such as log_loss), where I expect the bulk of runtime was in the inefficient for loop that has been removed in this PR.

I confirmed equivalence to mainline by running TabRepo ensemble simulations and achieving identical results with this PR.

LennartPurucker:
Great work! 

Two minor comments:

- I wonder how far the new code will be less numerically stable than the old code due to `ensemble_prediction *= (s - 1) / s`, any thoughts? 
- Also, if we need further speedups for ensemble selection, we can always consider parallelizing the internal for loop for roughly a `len(predictions) // n_jobs` speedup at the cost of much higher memory usage.

Innixma:
@LennartPurucker I tested the PR on TabRepo, no noticeable numerical precision loss:

```
method & normalized-error & rank & time fit (s) & time infer (s) \\
PortfolioPR4367--N200 (ensemble) (4h) & 0.355 & 195.5 & 6335.7 & 0.036 \\
PortfolioMainline--N200 (ensemble) (4h) & 0.355 & 195.5 & 6335.7 & 0.036 \\
```

PR #4379, opened 2024-08-12
[tabular] Fix exception when >20 configs present in medium quality preset

Innixma:
*Issue #, if available:*

*Description of changes:*

-  Fix exception when >20 configs present in medium quality preset
- The old hyperparameters configs had single models be in a special format where they aren't in a list, such as:

```
{
    'XGB': {},
}
```

This will cause an exception if logged in an edge-case where there are more than 20 configs in the hyperparameters dict.
This impacts medium and good quality presets.

This PR fixes this issue by converting to list before logging.


prateekdesai04 (approved):
LGTM!

PR #4389, opened 2024-08-14
[Tabular] Support for Raising an Error When No Models Were Trained Successfully During fit()

LennartPurucker:
This PR enables users to make AutoGluon raise an error if no models are trained successfully during `fit`. 

Without this, people sometimes write their own code to catch this or ignore it and then believe AutoGluon is broken, although it is most likely a result of a broken environment or their memory being insufficient. 

So far, this defaults to False, but I would argue this should default to True in the future. 


LennartPurucker:
@Innixma, I think you need to re-trigger the CI with your authentication. Otherwise, I am lost in what the error message could be.

Innixma:
> @Innixma, I think you need to re-trigger the CI with your authentication. Otherwise, I am lost in what the error message could be.

Fun fact, the CI failed because all of GitHub was down yesterday for over 30 minutes: https://www.githubstatus.com/incidents/kz4khcgdsfdv

PR #4391, opened 2024-08-15
Addressed Unexpected Torch Seed Behavior with Dataloader

adibiasio:
#### Summary of Changes:
This pull request addresses unexpected torch seed behavior observed when using the autogluon `TabularTorchDataset` object, which leverages Torch Dataloaders. Using different seeds for the data loading process can lead to unexpected behavior when various functions rely on data loader behavior. For instance, in the `TabularNeuralNetTorchModel` training code, calling `self.score()` multiple times led to different results, despite calculating the same metric on the same datasets. This issue was caused because of differences in the batch loading process utilized within `self.score()`.

To address these issues, the following PyTorch article on ensuring reproducibility was implemented: https://pytorch.org/docs/stable/notes/randomness.html

### DIFF
--- a/tabular/src/autogluon/tabular/models/tabular_nn/torch/tabular_torch_dataset.py
+++ b/tabular/src/autogluon/tabular/models/tabular_nn/torch/tabular_torch_dataset.py
@@ -1,5 +1,6 @@
 import logging
 import os
+import random
 
 import numpy as np
 import torch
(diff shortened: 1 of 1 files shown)


Innixma (approved):
LGTM! Awesome catch!

Innixma:
For clarification, the issue is that `self.score` would change the torch.rand value, thus changing the training process in future epochs due to`torch.rand` impacting how logic such as batch order, dropout, etc. work.

PR #4396, opened 2024-08-15
[Tabular,Core] Fix Temperature Scaling

LennartPurucker:
Heyho,

I have noticed that the predictions of my best model were often quite suboptimal compared to other models for metrics that were affected by temperature scaling. 

After looking more into it, I believe I found a bug and two crucial improvements to the temperature scaling logic. 

The proposed PR now does the following:
1. We take the best-found temperature instead of the last temperature when tuning temperature. 
2. We double-check that temperature improves performance on the target metric.

I am certain that 1. is very important, although it might overfit more.

For 2. it is a bit more complicated, as it is sometimes beneficial to even accept better calibration if it does not improve the target metric on validation data. Moreover, I am not too sure that my code works in all cases of all eval metrics (I also did not find an easy way to add sample weights into the metric call here). 

# Failure Case Examplee

I recently ran into this unhappy result when using default AutoGluon:

Default
```bash
Scores Before Tuning Temp: {'roc_auc_ovo': 0.9153388674076931, 'accuracy': 0.7211895910780669, 'balanced_accuracy': 0.7160878147673921, 'log_loss': -0.7753828493979228, 'mcc': 0.6275299996789895}
Temp After Tuning: -1242.176513671875
Scores After Tuning Temp: {'roc_auc_ovo': 0.09863829315371282, 'accuracy': 0.01486988847583643, 'balanced_accuracy': 0.014253449992886616, 'log_loss': -1.3871372218846423, 'mcc': -0.3297512402994374}
```

Here, the temperature is so low because, after 10 out of 100 iterations, the gradient-based optimizer started to diverge, and the exponential learning rate made it irrecoverable. Then, the old code picked the last iteration's temperature and returned it.
Btw. this is also likely the reason why this function sometimes returns None. 

My Code
```bash
Scores Before Tuning Temp: {'roc_auc_ovo': 0.9153388674076931, 'accuracy': 0.7211895910780669, 'balanced_accuracy': 0.7160878147673921, 'log_loss': -0.7753828493979228, 'mcc': 0.6275299996789895}
Temp After Tuning: 0.9741883873939514
Scores After Tuning Temp: {'roc_auc_ovo': 0.9154014511133995, 'accuracy': 0.7211895910780669, 'balanced_accuracy': 0.7160878147673921, 'log_loss': -0.7687201466930794, 'mcc': 0.6275299996789895}
```

---

### DIFF
--- a/core/src/autogluon/core/calibrate/temperature_scaling.py
+++ b/core/src/autogluon/core/calibrate/temperature_scaling.py
@@ -61,7 +61,10 @@ def tune_temperature_scaling(y_val_probs: np.ndarray, y_val: np.ndarray, init_va
 
     optimizer.step(temperature_scale_step)
 
-    best_loss_index = np.array(optimizer_trajectory)[:, 0].argmin()
+    try:
+        best_loss_index = np.nanargmin(np.array(optimizer_trajectory)[:, 0])
+    except ValueError:
+        return None
     temperature_scale = float(np.array(optimizer_trajectory)[best_loss_index, 1])
 
     if np.isnan(temperature_scale):
--- a/core/src/autogluon/core/calibrate/temperature_scaling.py
+++ b/core/src/autogluon/core/calibrate/temperature_scaling.py
@@ -47,6 +47,8 @@ def tune_temperature_scaling(y_val_probs: np.ndarray, y_val: np.ndarray, init_va
     optimizer = torch.optim.LBFGS([temperature_param], lr=lr, max_iter=max_iter)
     scheduler = torch.optim.lr_scheduler.ExponentialLR(optimizer, gamma=0.99)
 
+    optimizer_trajectory = []
+
     def temperature_scale_step():
         optimizer.zero_grad()
         temp = temperature_param.unsqueeze(1).expand(logits.size(0), logits.size(1))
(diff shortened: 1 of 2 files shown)


core/src/autogluon/core/calibrate/temperature_scaling.py
```diff
@@ -47,18 +47,26 @@ def tune_temperature_scaling(y_val_probs: np.ndarray, y_val: np.ndarray, init_va
...
+    optimizer_trajectory = []
+
     def temperature_scale_step():
         optimizer.zero_grad()
         temp = temperature_param.unsqueeze(1).expand(logits.size(0), logits.size(1))
         new_logits = logits / temp
         loss = nll_criterion(new_logits, y_val_tensor)
         loss.backward()
         scheduler.step()
+        optimizer_trajectory.append((loss.item(), temperature_param.item()))
```
Innixma, line 59: Is this the correct location for this line? Or should it be before the `loss.backward()` call?

LennartPurucker: I think it does not matter, as `backward` should only affect the connected computation graph, not the loss value stored in the tensor. In [this ](https://pytorch.org/tutorials/beginner/introyt/trainingyt.html#the-training-loop) example code, they also call `loss.item()` after `loss.backward()`.

Innixma (approved):
LGTM! Excellent find!

PR #4397, opened 2024-08-15
Skip temperature scaling if negative temperature, increase learning rate

Innixma:
*Issue #, if available:*

*Description of changes:*

- Skip temperature scaling if zero or negative temperature. A negative temperature could be selected if the model has a worse validation score than a model that predicts a constant `0.5` value. This can happen when there are very few rows of validation data and is an artifact of validation overfitting.
- increase learning rate by 10x to make convergence faster, reduce max iters by 5x to compensate. We can afford a more aggressive learning rate now that we pick the best iteration instead of the final iteration, so divergence is less problematic.


prateekdesai04 (approved):
LGTM

PR #4400, opened 2024-08-15
[tabular] raise_on_no_models_fitted=True

Innixma:
*Issue #, if available:*

*Description of changes:*

- Switch the default behavior of TabularPredictor to raise a RuntimeError if no models were successfully fit in a fit call. Most users would prefer an error to be raised in this scenario, as not having an error raise may confuse them into thinking a model was fit successfully.


prateekdesai04 (approved):
LGTM

PR #4407, opened 2024-08-18
[Tabular] Fix ROC AUC Average for Macro is Macro

LennartPurucker:
Unfortunately, so far in AutoGluon, when using the multiclass ROC AUC metric via the string key, the call to sklearn has been misconfigured. 

Instead of using `macro`, `weighted` has been passed to sklearn's `average` argument of ROC AUC.
This is unexpected by default and also occurs when explicitly using the `macro` string key for the metric.

*Description of changes:*
This bug resulted from re-using a previously initialized loop variable, which was set to `weighted` after a prior loop had finished.
I replaced the variable with an explicit statement, as there is no need for a variable. 


Innixma (approved):
LGTM! Very good find, this is a major bug.

PR #4409, opened 2024-08-19
[timeseries] Fix SeasonalNaive producing NaN forecasts if leading NaNs are present

shchur:
*Description of changes:*
- Currently, `SeasonalNaive` fails to produce forecasts if the first `y[:-(seasonal_period+2)]` observations contain NaN values, which leads to the predictor failing since the fallback model does not work. This PR fixes this issue by falling back to avg fill in this case.


canerturkmen (approved):
LGTM!

PR #4412, opened 2024-08-20
Replace black and isort with ruff

shchur:
*Issue #, if available:* #3353

*Description of changes:*
- This PR replaces all references to `black` and `isort` with the `ruff` linter.

*Left for a future PR:*
- Replacing flake8 with ruff


core/pyproject.toml
```diff
@@ -1,2 +1,34 @@
...
+    "smart_open",
+    "lightning",
+    "torchmetrics",
+    "transformers",
+    "nptyping",
+    "omegaconf",
+    "pytorch-metric-learning",
+    "nlpaug",
+    "nltk",
+]
```
shchur, line 34: We need to duplicate the `tool.ruff.lint.isort` sections in all `pyproject.toml` files for the ruff import sorter to work correctly.

core/pyproject.toml
```diff
@@ -1,2 +1,34 @@
-[tool.black]
+[tool.ruff]
```
prateekdesai04, line 1: @Innixma @shchur would this change detect all the src paths for all modules after pip install ?
as that was the problem earlier

prateekdesai04 (approved):
LGTM

core/pyproject.toml
Innixma: Probably not. But we can check after. My current plan is to remove the `pyproject.toml` files except for the root one after we converge on 119 line-length.

Innixma (approved):
LGTM! Thanks for sending this one out!

Innixma:
@shchur I didn't see any edits to the `check_style` tests, example: https://github.com/autogluon/autogluon/blob/master/tabular/tests/test_check_style.py

These tests run `flake8`. Maybe they should run `ruff` instead? Will they crash given that we are trying to remove `flake8` as a test time dependency?

pyproject.toml
```diff
@@ -1,10 +1,14 @@
-[tool.black]
+[tool.ruff]
 line-length = 119
-target-version = ['py38', 'py39', 'py310']
-
-[tool.isort]
-known_first_party = "autogluon"
-known_third_party = [
+target-version = "py38"
```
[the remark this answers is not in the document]
shchur: Ruff requires specifying the lower bound ([docs](https://docs.astral.sh/ruff/settings/#target-version)), so `target-version = "py38"` means that all versions `>= 3.8` are supported.

shchur:
@Innixma my plan is to deal with `flake8` in a separate PR since the current one already introduces a few changes (mostly related to formatting).

I agree that we can completely get rid of the `flake8` dependency, but that would require changing the logic in tests, which I would rather do in a separate PR to ensure that we don't miss anything.

PR #4415, opened 2024-08-21
[timeseries] Add method to convert a TimeSeriesDataFrame to a regular pd.DataFrame

shchur:
*Issue #, if available:*

*Description of changes:*
- Add method to convert a TimeSeriesDataFrame to a regular pd.DataFrame

### DIFF
--- a/timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
+++ b/timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
@@ -1008,3 +1008,7 @@ class TimeSeriesDataFrame(pd.DataFrame, TimeSeriesDataFrameDeprecatedMixin):
         # This hides method from IPython autocomplete, but not VSCode autocomplete
         deprecated = ["get_reindexed_view", "to_regular_index"]
         return [d for d in super().__dir__() if d not in deprecated]
+
+    def to_data_frame(self) -> pd.DataFrame:
+        """Convert `TimeSeriesDataFrame` to a `pandas.DataFrame`"""
+        return pd.DataFrame(self)
--- a/timeseries/tests/unittests/test_ts_dataset.py
+++ b/timeseries/tests/unittests/test_ts_dataset.py
@@ -1011,3 +1011,10 @@ def test_when_timestamps_have_datetime64_type_then_tsdf_can_be_constructed(dtype
     df[TIMESTAMP] = df[TIMESTAMP].astype(dtype)
     assert df[TIMESTAMP].dtype == dtype
     TimeSeriesDataFrame.from_data_frame(df)
+
+
+def test_when_to_data_frame_called_then_return_values_is_a_pandas_df():
+    tsdf = SAMPLE_TS_DATAFRAME.copy()
+    df = tsdf.to_data_frame()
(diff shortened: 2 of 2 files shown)


timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
```diff
@@ -1008,3 +1008,7 @@ def __dir__(self) -> List[str]:
         # This hides method from IPython autocomplete, but not VSCode autocomplete
         deprecated = ["get_reindexed_view", "to_regular_index"]
         return [d for d in super().__dir__() if d not in deprecated]
+
+    def to_data_frame(self) -> pd.DataFrame:
```
hsbadr, line 1012:
```suggestion
    def to_pandas(self) -> pd.DataFrame:
```

I'd rename this method since `TimeSeriesDataFrame` is a `DataFrame`.

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
```diff
@@ -1008,3 +1008,7 @@ def __dir__(self) -> List[str]:
...
+        """Convert `TimeSeriesDataFrame` to a `pandas.DataFrame`"""
+        return pd.DataFrame(self)
```
hsbadr, line 1014: Do we need to reset the index of the DataFrame?

timeseries/tests/unittests/test_ts_dataset.py
```diff
@@ -1011,3 +1011,10 @@ def test_when_timestamps_have_datetime64_type_then_tsdf_can_be_constructed(dtype
     df[TIMESTAMP] = df[TIMESTAMP].astype(dtype)
     assert df[TIMESTAMP].dtype == dtype
     TimeSeriesDataFrame.from_data_frame(df)
+
+
+def test_when_to_data_frame_called_then_return_values_is_a_pandas_df():
```
hsbadr, line 1016:
```suggestion
def test_when_to_pandas_called_then_return_values_is_a_pandas_df():
```

timeseries/tests/unittests/test_ts_dataset.py
```diff
@@ -1011,3 +1011,10 @@ def test_when_timestamps_have_datetime64_type_then_tsdf_can_be_constructed(dtype
...
+    tsdf = SAMPLE_TS_DATAFRAME.copy()
+    df = tsdf.to_data_frame()
```
hsbadr, line 1018: De we need to test that `df` can be directly converted to `TimeSeriesDataFrame`?
With or without resetting the index?

canerturkmen (approved):
LGTM!

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
shchur: I'd prefer to keep the current name to be consistent with `from_data_frame()`

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
shchur: The users should be able to do that themselves afterwards

PR #4436, opened 2024-08-27
[tabular] Add AdaptiveES support to NN_TORCH

Innixma:
*Issue #, if available:*

*Description of changes:*

- Add AdaptiveES support to NN_TORCH
- Switch default ES to use adaptive early stopping
- Update default num_epochs from 500 to 1000
- Code cleanup & documentation


tabular/src/autogluon/tabular/models/tabular_nn/torch/tabular_nn_torch.py
```diff
@@ -37,6 +39,10 @@
 class TabularNeuralNetTorchModel(AbstractNeuralNetworkModel):
     """
     PyTorch neural network models for classification/regression with tabular data.
+
```
prateekdesai04, line 42: qq: in reference to -> Switch default ES to use adaptive early stopping
this switch is just being made to NN_Torch correct ?

prateekdesai04 (approved):
LGTM!

tabular/src/autogluon/tabular/models/tabular_nn/torch/tabular_nn_torch.py
Innixma: Yes only NN_TORCH

PR #4444, opened 2024-08-29
[tabular] Add `valid_stacker` and `use_orig_features` options

Innixma:
*Issue #, if available:*

*Description of changes:*

- Add `valid_stacker` and `use_orig_features` options
- `valid_stacker` allows to disable model in L2+ layers
- `use_orig_features="never"` allows to disable model in L1 layer.
- Improved exception logging in HPO
- Added docstrings


tonyhoo:
The logic seems sound. We should incorporate tests to encompass these newly implemented parameters, if we haven't already done so.

Innixma:
> The logic seems sound. We should incorporate tests to encompass these newly implemented parameters, if we haven't already done so.

Thanks! Created a GitHub issue to add tests: https://github.com/autogluon/autogluon/issues/4448

Since this logic is advanced and not enabled by default, I plan to add unit tests as a follow-up.

FANGAreNotGnu (approved):
LGTM

PR #4446, opened 2024-08-29
[tabular] Set num_bag_sets to 1 by default

Innixma:
*Issue #, if available:*

*Description of changes:*

- Make num_bag_sets == 1 by default.
- Ever since v1.0, we did not use num_bag_sets > 1 in any of the presets, however if the user specified bagging manually and set a time limit, it would actually be set to 5 which most users would not want since it would 5x the runtime for minimal gain.
- This change has no impact when presets are used.


prateekdesai04 (approved):
LGTM

PR #4450, opened 2024-09-01
[tabular] Fix incorrect return type in `predict_multi` for regression

Innixma:
*Issue #, if available:*

*Description of changes:*

- Fix incorrect return type in `predict_multi` for regression
- Previously would return np.ndarray when `as_pandas=True` when it should return `pd.Series`.


prateekdesai04 (approved):
LGTM

Issue #2969, opened 2023-02-23
AutoGluonTabular model ignores the hyperparameters


shchur:
### Discussed in https://github.com/autogluon/autogluon/discussions/2968

<div type='discussions-op-text'>

<sup>Originally posted by **fsty** February 23, 2023</sup>
tabular_hyperparmeters hase no effect, all default models will be trained  
```
predictor = TimeSeriesPredictor(
    path="ts/", 
    prediction_length=12,
    target="QTE",
    eval_metric="RMSE", 
    known_covariates_names=["act"],
    quantile_levels=[0.1, 0.5, 0.9], 
    validation_splitter=splitter, 
    verbosity=4
).fit(
    train_data, 
     hyperparameters={
      "AutoGluonTabular": {"tabular_hyperparmeters": {"XGB":{}}  },
      }, 
      time_limit=1200
)
```</div>

**Comment by shchur** [comment]
No bug, just a typo.

**Comment by shchur** [comment]
Fix the docstring in https://auto.gluon.ai/stable/tutorials/timeseries/forecasting-model-zoo.html#autogluon.timeseries.models.AutoGluonTabularModel

**Comment by shchur** [comment]
Currently `timeseries` models do not raise an exception if an unknown HP is passed to them via `hyperparameters`. I agree that this should be fixed

**Comment by bryanwhiting** [comment]
This is a substantive issue that bad inputs don't yield any errors.

```
predictor = TimeSeriesPredictor(
    prediction_length=21,
    freq='h',
    target = 'x',
    eval_metric="MAE",
    quantile_levels = {},
)

predictor.fit(
    train_data=train_data, 
    hyperparameters={
        "DeepAR": {},
        "RecursiveTabular": {
            "model_type": space.Categorical(""),
            "n_estimators1": space.Int(50, 500),
            "max_depth2": space.Int(3, 10),
            "learning_rate3": space.Real(1e-3, 1e-1, log=True),
            "subsampl4e": space.Real(0.5, 1.0),
            "colsample_bytr1ee": space.Real(0.5, 1.0)
        }
    },
    hyperparameter_tune_kwargs={
        "scheduler": "local",
        "searcher": "auto",
        "num_trials": 5,
    },
    enable_ensemble = False,
    time_limit = 600 
)

def get_hyperparameters(model_name: str) -> dict:
    # https://github.com/autogluon/autogluon/issues/4152#issuecomment-2085256152
    model = predictor._trainer.load_model(model_name)
[...]

PR #4460, opened 2024-09-08
[timeseries] Refactor scaler implementation

shchur:
*Issue #, if available:*

*Description of changes:*
- Add a new parameter `target_scaler` supported by all models. If provided, a target scaler is applied to the target time series before any model-specific preprocessing.
- Upgrade to `mlforecast==0.13`
  - This required some modifications to the preprocessing logic because of breaking changes introduced in `mlforecast-0.11`
  - Instead of hacking the scaler private API with `_get_scale_per_item`, we now rely only on the public API of the scaler
- Add a property `allowed_hyperparameters` to `AbstractTimeSeriesModel` that lists the supported hyperparameters. This is necessary to silence warnings when local models receive the `target_scaler` argument. 
  - In the future, we can extend it to all models to deal with #2969


timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
```diff
@@ -167,19 +165,10 @@ def _get_mlforecast_init_args(self, train_data: TimeSeriesDataFrame, model_param
...
-        else:
-            logger.warning(
-                f"Unrecognized `scaler` {scaler_name} (supported options: ['standard', 'mean_abs', None]). Scaling disabled."
-            )
-
-        if self._scaler is not None:
+        # Support "scaler" for backward compatibility
+        scaler_type = model_params.get("target_scaler", model_params.get("scaler"))
+        if scaler_type is not None:
+            self._scaler = MLForecastScaler(scaler_type=scaler_type)
```
shchur, line 171: To discuss: Previously, `DirectTabular` and `RecursiveTabular` had a `"scaler"` hyperparameter that behaved very similarly to `target_scaler`. With the current change, the models "hijack" the `target_scaler` argument and implement custom scaling option (different from what's done in `AbstractTimeSeriesModel`). This is necessary to ensure that scaling is performed after differencing (inside the MLForecast object), which usually results in better accuracy.

Does this make sense?

timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
```diff
@@ -472,8 +468,9 @@ def is_quantile_model(self) -> bool:
 
     def _get_model_params(self) -> dict:
         model_params = super()._get_model_params()
-        model_params.setdefault("scaler", "mean_abs")
-        model_params.setdefault("differences", [])
+        model_params.setdefault("target_scaler", "mean_abs")
+        if "differences" not in model_params or model_params["differences"] is None:
+            model_params["differences"] = []
```
shchur, line 473: This change is needed so that `{"differences": None}` is interpreted as "auto differences". Currently (without the fix), `{"differences": None}` raises an error.

timeseries/src/autogluon/timeseries/transforms/scaler.py
```diff
@@ -0,0 +1,121 @@
...
+    """Applies min/max scaling to each time series in the dataset."""
+
+    def _compute_loc_scale(self, target_series: pd.Series) -> Tuple[pd.Series, pd.Series]:
+        stats = target_series.abs().groupby(level=ITEMID, sort=False).agg(["min", "max"])
+        scale = (stats["max"] - stats["min"]).clip(lower=self.min_scale)
+        loc = stats["min"] / scale
+        return loc, scale
+
+
+class LocalRobustScaler(LocalTargetScaler):
```
nathanaelbosch, line 86: All scalers above have a docstring. Is there a short one-line description that could add some information here?

timeseries/src/autogluon/timeseries/models/gluonts/torch/models.py
```diff
@@ -85,6 +85,8 @@ class DeepARModel(AbstractGluonTSModel):
         If True, ``lightning_logs`` directory will NOT be removed after the model finished training.
     """
 
+    # TODO: Should we rename "scaling: bool" to, e.g., "window_scaler": {"mean_abs", None} and have a hyperparameter names for all models?
+
```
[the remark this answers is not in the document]
nathanaelbosch: I am not very familiar with the details, but the categorization into local scaling and window scaling seems very plausible to me.

timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
nathanaelbosch: I am not quite sure what "This is necessary to ensure that scaling is performed after differencing" means, but overall it seems plausible to me that we transform the previously-used scaler into an `MLForecastScaler` now.

timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
```diff
@@ -553,15 +555,10 @@ def _postprocess_predictions(self, predictions: np.ndarray, repeated_item_ids: p
             predictions["mean"] = predictions["0.5"]
         else:
             predictions = pd.DataFrame(predictions, columns=["mean"])
-            predictions = self._add_gaussian_quantiles(predictions, repeated_item_ids=repeated_item_ids)
```
nathanaelbosch, line 556: Why did this part change with the scaling?

timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
nathanaelbosch: What does "auto differences" mean?

timeseries/src/autogluon/timeseries/models/autogluon_tabular/transforms.py
```diff
@@ -0,0 +1,60 @@
...
+        self.ag_scaler = get_target_scaler(name=self.scaler_type, target=self.target_col)
+        transformed = self.ag_scaler.fit_transform(self._df_to_tsdf(df)).reset_index()
+        return self._tsdf_to_df(transformed)
+
+    def inverse_transform(self, df: pd.DataFrame) -> pd.DataFrame:
+        transformed = self.ag_scaler.inverse_transform(self._df_to_tsdf(df))
+        return self._tsdf_to_df(transformed)
+
+
+def apply_inverse_transform(
```
nathanaelbosch, line 43: I was a bit surprised at first about this function not being inside of the class above, but I suppose that this comes from the fact that we need to handle both `BaseTargetTransform` and `_BaseGroupedArrayTargetTransform`?

nathanaelbosch (commented):
Thanks for the PR, and in particular also for your comments! As you know I am not too familiar with the codebase, but I left some comments and questions and hopefully resolving them will help make sure that we caught everything.

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
```diff
@@ -241,11 +243,31 @@ def fit(
         model: AbstractTimeSeriesModel
             The fitted model object
         """
+        self.initialize(**kwargs)
+        self.target_scaler = self._get_target_scaler()
```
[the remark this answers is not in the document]
shchur: I see two reasons for this choice:
1. `target_scaler` is a special kind of preprocessing, since we also need to apply the `inverse_transform` on the predictions. In contrast, `preprocess` only modifies the data before it's consumed by the model.
2. `self.preprocess` contains model-specific preprocessing logic that can be modified by the subclasses. By keeping the `target_scaler` out of it, we don't need to rely on the person implementing the model to remember using the `target_scaler` appropriately.

timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
shchur: "auto differences" = we set differences to `[get_seasonality(model.freq)]` by default.

timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
shchur: Previously, we had a rather convoluted logic that can be summarized as

- in fit(): original data -> scaled data -> compute residuals -> save "scaled" residuals
- in predict(): original data -> scaled data -> scaled predictions -> add scaled residuals -> undo scaling

We now simplified this to
- in fit(): original data -> scaled data -> compute residuals -> undo scaling on residuals -> save residuals that are on the same scaled as original data
- in predict(): original data -> scaled data -> scaled predictions -> undo scaling -> add residuals

The latter approach is slightly cleaner and does not requires us to mess with the internals of the scaler.

timeseries/src/autogluon/timeseries/models/autogluon_tabular/transforms.py
shchur: Exactly, we need a function that works both with `BaseTargetTransform` and `_BaseGroupedArrayTargetTransform` classes that have different expectations for the input format: the former works on `pd.DataFrame` while the latter only works on `GroupedArray`.

timeseries/src/autogluon/timeseries/transforms/scaler.py
shchur: Good idea, added some docstrings.

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
```diff
@@ -241,11 +243,31 @@ def fit(
...
+                val_data = self.target_scaler.transform(val_data)
             val_data = self.preprocess(val_data, is_train=False)
         return super().fit(train_data=train_data, val_data=val_data, **kwargs)
 
+    @property
+    def allowed_hyperparameters(self) -> List[str]:
+        """List of hyperparameters allowed by the model."""
+        return ["target_scaler"]
+
+    def _get_target_scaler(self) -> Optional[LocalTargetScaler]:
```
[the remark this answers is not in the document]
shchur: The purpose of this method is to create a `LocalTargetScaler` object based on the value of the `self.params["target_scaler"]` hyperparameter. I guess that a better name would be `_initialize_target_scaler` or `_create_target_scaler`, changing this.

There is a slightly complicated logic of how the hyperparameters are processed by the model. For example, we cannot just check the value of `self.params["target_scaler"]` during `__init__` and create the scaler there. The reason is that `self.params` might contain search spaces, and get the "concrete" hyperparameter value only after we call `self.initialize()` inside `fit()`. For this reason, we can only assign the `target_scaler` attribute during `fit()`.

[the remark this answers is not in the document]
shchur: I've renamed the method to `_create_target_scaler` and added a docstring explaining the purpose.

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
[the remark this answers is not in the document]
nathanaelbosch: This seems plausible to me, thanks a lot for clarifying!

timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
nathanaelbosch: Thank you for clarifying!

Innixma (approved):
LGTM!

Note that I'm not the most familiar with these changes, as it is somewhat hard to fully understand given the amount of code edits, but assuming benchmark results look fine and tests pass, should be ok.

PR #4463, opened 2024-09-09
fix platform tests

suzhoum:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -98,7 +98,7 @@ celerybeat-schedule
 .env
 
 # virtualenv
-.venv
+.venv*
 venv/
 ENV/
 
--- a/core/tests/unittests/test_parallel_local_folding.py
+++ b/core/tests/unittests/test_parallel_local_folding.py
@@ -8,9 +8,15 @@ import pandas as pd
 from autogluon.common import space
 from autogluon.common.utils.resource_utils import ResourceManager
 from autogluon.core.models import AbstractModel
-from autogluon.core.models.ensemble.bagged_ensemble_model import BaggedEnsembleModel
-from autogluon.core.models.ensemble.fold_fitting_strategy import ParallelLocalFoldFittingStrategy
+from autogluon.core.models.ensemble.bagged_ensemble_model import \
+    BaggedEnsembleModel
+from autogluon.core.models.ensemble.fold_fitting_strategy import (
+    FoldFittingStrategy, ParallelLocalFoldFittingStrategy)
+from autogluon.core.ray.resources_calculator import CpuResourceCalculator
 from autogluon.core.searcher import LocalRandomSearcher
+
+NUM_CPU = 8
+NUM_GPU = 1
 
 
 class DummyBigModel(AbstractModel):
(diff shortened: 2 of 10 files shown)


.gitignore
```diff
@@ -98,7 +98,7 @@ celerybeat-schedule
 .env
 
 # virtualenv
```
prateekdesai04, line 100: Why do we need * here ?

core/tests/unittests/test_parallel_local_folding.py
```diff
@@ -47,8 +55,8 @@ def _construct_dummy_fold_strategy(num_jobs, model_base_cls=AbstractModel, time_
         oof_pred_proba=np.array([]),
         oof_pred_model_repeats=np.array([]),
         save_folds=True,
-        num_cpus=ResourceManager.get_cpu_count(),
-        num_gpus=ResourceManager.get_gpu_count(),
+        num_cpus=NUM_CPU,
```
prateekdesai04, line 58: Why did we move away from calling the method ?

multimodal/setup.py
```diff
@@ -63,8 +63,9 @@
     "ruff",
     "datasets>=2.10.0,<2.15.0",
     "onnx>=1.13.0,<1.14.0",
-    "onnxruntime>=1.15.0,<1.16.0;platform_system=='Darwin'",
+    "onnxruntime>=1.15.0,<1.16.0",
     "onnxruntime-gpu>=1.15.0,<1.16.0;platform_system!='Darwin'",
+    "onnx>=1.13.0,<1.14.0",
```
prateekdesai04, line 68: Why do we add onnx twice?

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -230,7 +230,7 @@ def load_model_pipeline(self, context_length: Optional[int] = None):
                 "`import torch; torch.cuda.is_available()` returns `True`."
             )
 
-        device = self.device or ("cuda" if gpu_available else "auto")
+        device = self.device or ("cuda" if gpu_available else "cpu")
```
prateekdesai04, line 233: Why do we remove auto ?

core/tests/unittests/test_parallel_local_folding.py
suzhoum: The cpu number needs to be mocked, otherwise the results can vary depending on the Github runner machine type.

timeseries/src/autogluon/timeseries/models/chronos/model.py
suzhoum: This is being cherry-picked from https://github.com/autogluon/autogluon/pull/4565. Timeseries tests were failing on MacOS due to MPS memory overflow.

.gitignore
suzhoum: This is just to ignore virtual environment folders created such as `.venv3.11` or `.venv3.10` instead of just `.venv`.

multimodal/setup.py
suzhoum: Will remove it

timeseries/src/autogluon/timeseries/models/chronos/model.py
prateekdesai04: The timeout may mean that the tests may be getting stuck

suzhoum: MutiGPU tests passed when switching back from `uv pip install` to `pip install`.

.gitignore
suzhoum: This is just for testing purpose. We can name the virtualenv to anything we'd like.

prateekdesai04 (approved):

Issue #4453, opened 2024-09-03
[tabular] Add stratification for Regression tasks


Innixma:
For regression tasks, we currently do random splits without considering the label column.

Instead, we could do stratified splits by binning the label column and then doing stratified=True in the CV splitter.

This might lead to better results, especially when outliers exist.

Open question: How many bins?

LennartPurucker:
Random thought: One could use a clustering method that automatically finds the number of clusters to determine this.

Also, regarding the implementation, maybe treating this as a group is cleaner in the code.

LennartPurucker:
Rel.: https://github.com/scikit-learn/scikit-learn/issues/4757

Innixma:
Implemented here: #4586

PR #4478, opened 2024-09-17
[core] Enhance generate_train_test_split

Innixma:
*Issue #, if available:*

*Description of changes:*

- Add `train_size` argument to `generate_train_test_split`. This allows the user to specify the train size directly, akin to the argument `train_size` in sklearn's `train_test_split`.
- Add `stratify` argument to `generate_train_test_split`. This allows the user to specify the stratification logic, akin to the argument `stratify` in sklearn's `train_test_split`.
- Made `problem_type` argument optional. Defaults to the equivalent of `regression`.
- Added extensive unit tests
- This logic is partly preparation for resolving #4453 in a future PR, as we need the `stratify` argument.

### DIFF
--- a/core/src/autogluon/core/utils/utils.py
+++ b/core/src/autogluon/core/utils/utils.py
@@ -359,8 +359,15 @@ def extract_label(data: DataFrame, label: str) -> (DataFrame, Series):
 
 
 def generate_train_test_split_combined(
-    data: DataFrame, label: str, problem_type: str, test_size: float = 0.1, random_state: int = 0, min_cls_count_train: int = 1
-) -> (DataFrame, DataFrame):
+    data: DataFrame,
+    label: str,
+    problem_type: str = None,
+    test_size: int | float = None,
+    train_size: int | float = None,
+    random_state: int = 0,
+    stratify: bool | Series = None,
+    min_cls_count_train: int = 1,
+) -> Tuple[DataFrame, DataFrame]:
     """
     Generate a train test split from a DataFrame that contains the label column.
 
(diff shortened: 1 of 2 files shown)


core/src/autogluon/core/utils/utils.py
```diff
@@ -435,20 +486,42 @@ def generate_train_test_split(
...
         raise ValueError(f"Cannot split data into train/val as it contains only one sample.")
-    if isinstance(test_size, float):
-        if (test_size <= 0.0) or (test_size >= 1.0):
-            raise ValueError("fraction of data to hold-out must be specified between 0 and 1")
-    elif isinstance(test_size, int):
-        assert test_size > 0
-    valid_problem_types = [BINARY, MULTICLASS, REGRESSION, SOFTCLASS, QUANTILE]
-    assert problem_type in valid_problem_types, f'Unknown problem type "{problem_type}" | Valid problem types: {valid_problem_types}'
+    if test_size is None and train_size is None:
+        test_size = 0.1
```
prateekdesai04, line 490: qq: why do we go with a 90/10 and not 80/20

core/src/autogluon/core/utils/utils.py
```diff
@@ -484,23 +557,20 @@ def generate_train_test_split(
...
         if len(y_test) >= len(y_split.unique()):
             # This should never occur, otherwise the original exception is not an expected one
             raise
-    if problem_type != SOFTCLASS:
-        y_train = pd.Series(y_train, index=X_train.index)
-        y_test = pd.Series(y_test, index=X_test.index)
-    else:
-        y_train = pd.DataFrame(y_train, index=X_train.index)
-        y_test = pd.DataFrame(y_test, index=X_test.index)
+    cls_y = type(y)
```
prateekdesai04, line 565: qq: why do we remove the check here?

prateekdesai04 (approved):
LGTM!

core/src/autogluon/core/utils/utils.py
Innixma: Because I made that decision 5 or 6 years ago and now it is legacy. I felt that 90/10 is better than 80/20. No quantitative reason.

core/src/autogluon/core/utils/utils.py
Innixma: Because this check is really about the type of y_train/y_test. SOFTCLASS is the only problem type that is in the DataFrame format. The new logic generalizes and removes unnecessary hardcoding.

PR #4480, opened 2024-09-19
Add new model: Explainable Boosting Machine

paulbkoch:
*Description of changes:*
Add new model: Explainable Boosting Machine


paulbkoch:
The lint check seems to be failing because the dependency interpret-core was not installed.  I added interpret-core as an optional flagged dependency to the package and also to the optional tests flag, but at the time that linting occurs it doesn't look like the autogluon package has been built yet, and thus the new test requirement is not installed.  On my local machine these requirements install correctly using the optional installation flags.

I guess I could add it explicitly for now to the env_setup.sh file.  Please advise on your preferred solution.

paulbkoch:
I tried adding "python -m pip install interpret-core" to setup_build_env, but from the logs it doesn't appear to have been run.  I presume this is running in a container and I don't have permission to update the script being run in the action.

prateekdesai04:
Hi @paulbkoch you would first have to pass the lint_check in order to run the further remaining tests.
The lint_check seems to fail due to some formatting issue, you can try updating your PR after which we can run the CI.

prateekdesai04:
Hi @paulbkoch , 
Is this the run where you first observed the lint failure: https://github.com/autogluon/autogluon/actions/runs/10946844689/job/30394271765 ?

If yes, I believe this is because the message seems to be incorrectly formatted: https://github.com/autogluon/autogluon/pull/4480/files#diff-9d8ccb4278e0f2df479992439def2c704b877c2a12244f9175b4a17c3de21a01

It should work once you format it correctly, you can use `black` for formatting.

You can add interpret-core as an optional dependency in `setup.py` rather than using `env_setup.sh` once you make the above changes you can try running the CI again. Let us know if that helps!

prateekdesai04:
Apologies - I believe we have moved on from black to ruff - https://github.com/autogluon/autogluon/pull/4412
initially it was mixed

paulbkoch:
Yep, figured and updated.

paulbkoch:
Thanks @prateekdesai04.  Seems to have passed the lint so you were correct about the formatting being the issue.

Innixma:
Hi @paulbkoch, many thanks for sending the PR!! We have been looking to add the EBM model and this is a fantastic start to this addition. We will plan to merge this for the next release once we do additional testing and benchmarking to verify its stability and impact on the ensemble result.

@LennartPurucker You may be interested in this given you implemented EBM in AutoGluon as a prototype earlier

Innixma:
We will plan to do extensive testing via [TabRepo](https://github.com/autogluon/tabrepo) once we merge our improved model testing code.

Some high priority additions to the current baseline (recommended to work on after we get initial results from TabRepo to track progress):

- [ ] Respect time_limit
- [ ] Leverage validation data for early stopping?
- [ ] Preprocessing to handle edge-cases / NaNs / improve performance?
- [ ] Memory usage estimation prior to fit
- [ ] Strong Hyperparameter Defaults
- [ ] Hyperparameter Search Space
- [ ] Feature Type Filters (Drop Images / Text / etc.)

LennartPurucker:
Heyho everyone, @paulbkoch thank you very much for contributing and starting this push for adding EBMs! 

Here are some immediate pointers from my side regarding the prototype.


## Requirements 

I advise against depending on the full `interpret` package as it has a lot of dependencies and functionalities we actually do not need for AutoGluon or as part of the code for running EBMs.

To solve this, we need to depend only on the following:
```
interpret-core[required,ebm]
``` 

We could even install with `--no-dependencies`, but that might be too much hassle for pip. 

## AutoGluon Model EBM Code 

I created a running prototype of EBMs for AutoGluon a while back. Your PR motivated me to clean it up a bit and add it to a branch of AutoGluon; see here: https://github.com/LennartPurucker/autogluon/blob/ebm_model/tabular/src/autogluon/tabular/models/ebm/ebm_model.py

Feel free to merge/pull my model in your code. 

It is a bit longer than your code because it supports the following:
* preprocessing,
* validation data,
* early stopping support (+metric),
* passing column types to EBMs,
* some changes to the default behavior of EBMs to make them effective for AutoGluon,
* some other AutoGluon-related housekeeping. 

## Open TODOs

I added some open todos to the code, roughly matching Nick's requirements.

paulbkoch:
> * [ ]  Respect time_limit
> * [ ]  Leverage validation data for early stopping?
> * [ ]  Preprocessing to handle edge-cases / NaNs / improve performance?
> * [ ]  Memory usage estimation prior to fit
> * [ ]  Strong Hyperparameter Defaults
> * [ ]  Hyperparameter Search Space
> * [ ]  Feature Type Filters (Drop Images / Text / etc.)

Hi @Innixma -- Thanks for the feedback.  These look like great changes.  For background, I should mention that I'm one of the InterpretML developers.  I had planned on adding hyperparameters after the baseline was merged.  The default EBM parameters are generally pretty good since we have our own benchmark system, and the defaults were chosen to optimize the global performance on a variety of datasets.  We haven't had a lot of experience optimizing the parameters for individual datasets though.  I've speculatively listed some recommendations here based on the results from the global optimization: https://interpret.ml/docs/hyperparameters.html.

We might be able to strike out most preprocessing as already completed.  Categoricals are detected automatically if the pandas dataframe uses CategoricalDType, or non-numeric strings.  InterpretML includes handling for missing values, and EBMs do not require imputation, although it might be interesting to benchmark it.

I'll take a look at @LennartPurucker's PR for the rest.

After you have a chance to benchmark it with TabRepo, I'd be curious to hear your findings.  I think you will be surprised at how well EBMs perform.  In our own benchmarks, which have only use default parameters so far, EBMs outperform XGB, and on some metrics LightGBM too.

paulbkoch:
> Heyho everyone, @paulbkoch thank you very much for contributing and starting this push for adding EBMs!
> 
> Here are some immediate pointers from my side regarding the prototype.
> 
> ## Requirements
> I advise against depending on the full `interpret` package as it has a lot of dependencies and functionalities we actually do not need for AutoGluon or as part of the code for running EBMs.
> 
> To solve this, we need to depend only on the following:
> 
> ```
> interpret-core[required,ebm]
> ```
> 
> We could even install with `--no-dependencies`, but that might be too much hassle for pip.
> 
> ## AutoGluon Model EBM Code
> I created a running prototype of EBMs for AutoGluon a while back. Your PR motivated me to clean it up a bit and add it to a branch of AutoGluon; see here: https://github.com/LennartPurucker/autogluon/blob/ebm_model/tabular/src/autogluon/tabular/models/ebm/ebm_model.py
> 
> Feel free to merge/pull my model in your code.
> 
> It is a bit longer than your code because it supports the following:
> 
> * preprocessing,
> * validation data,
> * early stopping support (+metric),
> * passing column types to EBMs,
> * some changes to the default behavior of EBMs to make them effective for AutoGluon,
> * some other AutoGluon-related housekeeping.
> 
> ## Open TODOs
> I added some open todos to the code, roughly matching Nick's requirements.

Hi @LennartPurucker -- Thanks for working on this earlier EBM implementation!  I had a look and most of it looks good.  While reading it over I saw the following that I would modify:

1) No need to impute.  EBMs handle missing values.
2) I'm not 100% sure, but I think the LabelEncoderFeatureGenerator isn't required in this case since it looks like autogluon is passing in categoricals as CategoricalDType, which InterpretML already understands.  A simple concatenation of X and X_val and leaving it as pandas should work, I think.
3) f_type = "None" should be f_type = None, or you can use "auto" if you want a string.  I would think this raises an exception?
4) The line: validation_size=len(X)    should be    validation_size=len(X_val)    It doesn't matter though since bags is also specified so validation_size is ignored.  I mention it in case you have other code that parallels this that doesn't specify bags.


I agree the interpret package is unnecessary for autogluon.  I called the optional tag "interpret" in autogluon for clarity, but if you look at the optional dependencies in setup.py, you'll see that I used interpret-core.  The interpret tags [required,ebm] are no longer required in the latest interpret-core releases.

LennartPurucker:
> No need to impute. EBMs handle missing values.

Very nice :) 

>  LabelEncoderFeatureGenerator

Yes, this stems from a default preprocessing for models with an sklearn API, which is likely suboptimal. 
Fixing it. 

> f_type = "None" should be f_type = None, or you can use "auto" if you want a string. I would think this raises an exception?

This worked fine for the last months. Maybe because I worked on numpy arrays? I changed it to "auto".

> validation_size=len(X) 

Well spotted; this was a copy-paste error. But then I will remove `validation_size` if it is unnecessary in this case.

>I used interpret-core.

Ah very cool; I did not see that. Great! 


I pushed the changes to the branch. 
---

One comment regarding your previous reply: I played quite a bit with dataset-specific hyperparameters and have some pointers, but I think we should learn them via TabRepo. However, I noticed the performance to be very task-specific (i.e., regression vs. binary vs. multiclass) due to how interactions work, as far as I understand. 

And one question since you are an InterpretML developer: First of all, great work! 
Could you support a timeout callback for EBMs? I.e., stopping in the rounds/iterations early based on time.

paulbkoch:
Thanks @LennartPurucker.  I'll have a look again.  Regarding hyperparameters, what version of InterpretML were you experimenting on previously?  There were some big changes in v0.5.1 that targeted improved performance, and then another change in v0.6.1 that I think would impact multiclass significantly.  In general EBMs are strongest on binary classification.  For multiclass the default implementation doesn't include pairs.  It's possible to enable multiclass pairs, so for Autogluon where the interpretability isn't important, we might want to try that.  You're probably right that for regression, pairs and possibly higher order interaction become more important.  We can certainly try scaling those up to see what impact it has.  We haven't experimented too much with that ourselves since having excessive pairs breaks interpretability, as do 3-way and higher interactions, but that's not a concern here.

LennartPurucker:
I mainly worked with v0.6.1+. 

>  strongest on binary classification

This matches my findings. I found binary > regression > multiclass.


>  It's possible to enable multiclass pairs

That sounds good. We should add that to the search space for TabRepo / as an alternative default similar to the multiple defaults for LightGBM. 

> Autogluon where the interpretability isn't important, 

I would not rule out interpretability entirely. AutoGluon has a (somewhat unsupported) interpretability preset, which EBMs could significantly enhance. Ideally, the users can trade off interpretability with performance if needed.

paulbkoch:
Hi @LennartPurucker -- I looked at the updated code and I have two further recommendations:

If the type hints to the _fit function are correct, then X and y are already pd.DataFrame and pd.Series, so using pd.DataFrame(X) and pd.Series(y) here is less memory efficient as they get duplicated:
https://github.com/LennartPurucker/autogluon/blob/ebdc463c645f8032ae410af339fc403bbc5c56cd/tabular/src/autogluon/tabular/models/ebm/ebm_model.py#L64-L65

Similar memory concerns here.  Given X and y are a pandas objects, vstack/hstack will convert the dataframe/series to an object array, which will be slower to process and consume a lot more memory if there are categorical types:
https://github.com/LennartPurucker/autogluon/blob/ebdc463c645f8032ae410af339fc403bbc5c56cd/tabular/src/autogluon/tabular/models/ebm/ebm_model.py#L110-L111

I would recommend changing to:
fit_X = pd.concat([X, X_val], ignore_index=True)
fit_y = pd.concat([y, y_val], ignore_index=True)

Other than these two things, your implementation is clearly more complete, so we can probably just replace mine.  My preference would be to do this after benchmarking the simple implementation as @Innixma suggested to get an initial baseline.  I think you and @Innixma are correct that setting the outer bags to 1 and using the validation set will be beneficial, but these kinds of changes make me worried without a good benchmark.

paulbkoch:
I had a further question on the n_jobs parameter.  The current EBM implementation parallelizes the creation of the outer bags.  Our default is to create 14 outer bags, so on typical hardware we get fairly good utilization since most computers have less than 14 cores.  If we set outer bags to 1, then the current EBM implementation will only be able to use 1 core.  That might not be bad if AutoGluon can efficiently use the other cores to build other models in parallel.

paulbkoch:
To answer some of the questions above that I haven't answered yet:

I agree adding a timeout parameter would be useful, and especially in the context of AutoGluon.  I'll have to give it some thought on the best way to do it for EBMs.  Since we parallelize on the outer bags, we might have a scenario where some of the outer bags are completed, some are in process, and some have not even begun.  In some cases it might be better to throw away the incomplete bags and rely only on the completed bags, and in other cases it might make more sense to incorporate them.  These are only really issues for non-AutoGluon workloads with multiple outer bags.  Eventually we want to move to a finer grained parallelism which would allow us to more efficiently use all the cores across multiple outer bags.  This would also make timed stopping much cleaner.

In terms of memory estimation: I'd verify this, but with the default outer_bags=14, my rule of thumb would be to take the size of the dataframe and multiply by 3-5x.  When using 1 outer bag, I'd change that to multiplying by 2x.  The EBM implementation bit packs data internally, so it shouldn't use much more than the original dataframe, and our internal data structures aren't very big either.

In terms of time estimation, EBMs take about 20x longer than XGBoost, 50x longer than LightGBM, and maybe 2x longer than catboost (from memory), so perhaps we can leverage their time estimation initially.  There are algorithmic differences though that could make the numbers fairly different on specific datasets, and I'm unsure how the outer_bags=1 change will affect things.

In addition to adding multiclass pairs, we might consider bumping up the default number of pairs substantially.  For interpretability reasons we limit the number of pairs, but I've observed increasing performance as the number of pairs go up.  The upper pair limit is probably more dependent on how much time we want to spend on them.

I'm curious about the interpretability aspect of AutoGluon.  What's the value-add?  Is it mostly for hyperparameter tuning?

LennartPurucker:
> using pd.DataFrame(X) and pd.Series(y) here is less memory efficient /  vstack/hstack 

Yes, these were both still remnants from the using numpy arrays. Fixed. 



I would recommend changing to:
fit_X = pd.concat([X, X_val], ignore_index=True)
fit_y = pd.concat([y, y_val], ignore_index=True)


> benchmarking the simple implementation

Nick knows best how to approach this, but my two cents are that I am less in favor. For knowing how good a model is, we require models to abide by time limits for proper benchmarking in (Auto)ML.  Moreover, ignoring the validation data passed to the model is also not optional, given that the model would otherwise ignore 1/8% of the data per fit. Lastly, to make a fair comparison in that case, we would also need to implement internal cross-validation and bagging for all other methods. 

--- 

> If we set outer bags to 1, then the current EBM implementation will only be able to use 1 core. That might not be bad if AutoGluon can efficiently use the other cores to build other models in parallel.

AutoGluon would parallelize the outer bags using Ray. For larger datasets, however, AutoGluon might give the model multiple cores; ideally, the model should still be able to use multiple cores. 

This is likely a methodological question, but could EBMs parallelize something other than the outer bags? I guess the number of rounds has to be sequential. 

> In terms of memory estimation

We can also verify this empirical, i.e., fixing it once it crashes somewhere. 

> EBMs take about 20x longer than XGBoost, 50x longer than LightGBM, and maybe 2x longer than catboost (from memory),

In my limited experience, this is very much due to outer_bags. Most of these are similarly fast, especially if we use early stopping and callbacks for time, as all of these do in AutoGluon. 

---

> I'm curious about the interpretability aspect of AutoGluon. What's the value-add? Is it mostly for hyperparameter tuning?

I am not sure I got your questions.  I think the original interpretability preset mainly offered a version that supports only interpretable models that are then correctly ensembled and validated. But Nick would know more as I only found it in the code 😅

paulbkoch:
> AutoGluon would parallelize the outer bags using Ray. For larger datasets, however, AutoGluon might give the model multiple cores; ideally, the model should still be able to use multiple cores.
> 
> This is likely a methodological question, but could EBMs parallelize something other than the outer bags? I guess the number of rounds has to be sequential.

There is a way to parallelize work within a single bag, but it's non-trivial to implement.  It's on our roadmap, but since things mostly work today, it has been a lower priority than other items.  It sounds like it won't be an issue in AutoGluon either, so that's nice.  We might consider changing n_jobs to 1 in this code since joblib will avoid spawning a new process then, and perhaps joblib might also use that information somehow when scheduling other work.

> In my limited experience, this is very much due to outer_bags. Most of these are similarly fast, especially if we use early stopping and callbacks for time, as all of these do in AutoGluon.

True.  Those numbers are when constructing 14 outer bags, so EBMs should be in the same ballpark for a single bag if the CPUs can be used efficiently elsewhere.

Innixma:
@paulbkoch 

> I'm curious about the interpretability aspect of AutoGluon. What's the value-add? Is it mostly for hyperparameter tuning?

For now that remains an open question. We support feature importance and SHAP as black-box methods, but we don't have a clear vision for what interpretability in an AutoML system would look like beyond those techniques. We have added various things in the past, such as an entire `autogluon.eda` package, which you can see here: https://www.kaggle.com/code/samvelkoch/s4e9-eda-by-autogluon-v-1

The tricky part is how to prioritize which pieces of information to show the user, otherwise without smart prioritization, it just becomes an avalanche of information that for 99% of users won't be actionable.

I think the first priority for EBM's in AG is performance improvement if it adds diversity to the ensemble. The second priority would be seeing if we can leverage any of EBM's nice properties for improved visualizations to the user if they are interested in them.

paulbkoch:
> For now that remains an open question. We support feature importance and SHAP as black-box methods, but we don't have a clear vision for what interpretability in an AutoML system would look like beyond those techniques. We have added various things in the past, such as an entire `autogluon.eda` package, which you can see here: https://www.kaggle.com/code/samvelkoch/s4e9-eda-by-autogluon-v-1

Having blackbox methods like SHAP in AG makes a lot of sense.  For glassbox models, you pretty much can't combine them, so that certainly takes away a lot of the utility within the AG context.  Maybe there's some utility in being able to compare across various interpretable models quickly and simultaneously look for hyperparameters.  FIGS in imodels is often a high performer, as is GAMI-net, so people might want to compare them, and of course linear models, etc.  These are things that people could do of course using only the separate packages, but often there is a benefit to having a unified API and being able to switch pipelines from blackbox to glassbox using a few parameters.

> I think the first priority for EBM's in AG is performance improvement if it adds diversity to the ensemble. The second priority would be seeing if we can leverage any of EBM's nice properties for improved visualizations to the user if they are interested in them.

Yes, completely agree.  In terms of AG, my primary motivation is to push the performance aspect of EBMs without regard to interpretability.  I'm happy to consider the interpretability angle later though if you think there would be a benefit.

Innixma:
Update: I have high hopes for this model, as some preliminary tests show that it does improve AutoGluon's ensemble when added! I wasn't able to find bandwidth to get it in for the v1.2 release, so I'll be targeting adding in the v1.3 release. Happy holidays!

paulbkoch:
Very nice.  Out of curiosity, if it's easy to retrieve, what was the improvement observed on the preliminary tests, and what version of interpret was tested?

After a basic version has been merged, it would be interesting to benchmark a few variations if there's time before the v1.3 release.  We've already touched on a few above, but since we don't care about interpretability in this application, we can also add things like 3-way and higher interactions, extend the number of pairs by large amounts, etc.  It would be interesting to see how each change improves the ensemble.

Innixma:
@paulbkoch I observed a 54% win-rate across 110 classification tasks by adding a single default hyperparameter EBM model to AutoGluon's portfolio, but I only tested on datasets with <=10000 rows as this was part of my experiments for our now released [TabPFNMix](https://huggingface.co/autogluon/tabpfn-mix-1.0-classifier) model.

As I work on TabRepo 2.0 in Q1 2025, I'll be incorporating EBM as one of the first additional models.
Here is the very early and very hacky WIP TabRepo 2.0 code for EBMs: https://github.com/autogluon/tabrepo/pull/77/files#diff-41e8b957082dd6383a87921817cccf865cb3db8a0d441971930355960b294441

Innixma:
@paulbkoch Thanks for your patience and also for updating the PR to add EBM to the model register!

I've been focused on TabRepo 2.0, and am at the state where I can start accepting search space recommendations from model authors for me to test.

If you'd like to sent a PR to TabRepo with your recommended EBM search space, here are a few example PRs from others for different models:

TabICL: https://github.com/autogluon/tabrepo/pull/113/files
RealMLP: https://github.com/autogluon/tabrepo/pull/105/files
XGBoost: https://github.com/autogluon/tabrepo/pull/106/files

Innixma:
You can also send a PR if you have improvements to my current reference EBM implementation in TabRepo: https://github.com/autogluon/tabrepo/blob/main/tabrepo/benchmark/models/ag/ebm/ebm_model.py

Innixma:
@paulbkoch yes the existing implementation in TabRepo does work. Thanks for the search space PR!

paulbkoch:
@Innixma -- How are you finding the current fitting time?  Most of the changes I would want to make would increase the fitting time.

Innixma:
@paulbkoch current fit time is acceptable. We plan to benchmark with a 1 hour time limit in TabRepo, so if the model requires >1 hour to train on a dataset, then it will need to be early stopped before being fully fit.

tabular/setup.py
```diff
@@ -139,6 +142,7 @@
 
 test_requires = []
 for test_package in [
+    "interpret",
```
LennartPurucker, line 146: We could also add it to the tabarena requirements above, I think.

paulbkoch: it seems to be called "ebm" in tabrepo:

https://github.com/paulbkoch/tabrepo/blob/2155d4f1c7f778abd08b43f28e4113f5ed91d008/setup.py#L50

paulbkoch: ah, gotcha, you mean here:

https://github.com/paulbkoch/autogluon/blob/master/tabular/setup.py#L134-L137

LennartPurucker:
FYI: i introduced a bug in my refactor. I fixed it here https://github.com/autogluon/tabrepo/commit/34374d73999e4180fbfeefc6b4d7e7ed6d9c44c2

paulbkoch:
> FYI: i introduced a bug in my refactor. I fixed it here [autogluon/tabrepo@34374d7](https://github.com/autogluon/tabrepo/commit/34374d73999e4180fbfeefc6b4d7e7ed6d9c44c2)

Thanks.  Fixed.

paulbkoch:
@LennartPurucker & @Innixma -- I'm done tinkering with this code.  Ready for review.

paulbkoch:
> LGTM, one thought: should we call the folder of the model code `interpret` or `ebm`?
> 
> If there is another model from `interpret` at some point, then we would still need another folder for this model. Also so far I think most model folders have the name of the model itself.

That makes a lot of sense.  I changed the directory to ebm

tabular/tests/unittests/registry/test_model_registry.py
```diff
@@ -134,6 +137,7 @@
     HSTreeModel: 0,
     BoostedRulesModel: 0,
     DummyModel: 0,
+    EBMModel: 0,
```
paulbkoch: Should I set the priority to something other than zero?

paulbkoch: I changed the priority to 35 since it seemed like EBMs were in between linear models and XGBoost.  Let me know if you wanted something different.

LennartPurucker: sounds good to me!

tabular/tests/unittests/models/test_ebm.py
```diff
@@ -0,0 +1,25 @@
+from autogluon.tabular.models import EBMModel
+from autogluon.tabular.testing import FitHelper
+
+
```
Innixma, line 13:
```suggestion
toy_model_params = {}


def test_ebm():
    model_cls = EBMModel
    model_hyperparameters = toy_model_params

    FitHelper.verify_model(model_cls=model_cls, model_hyperparameters=model_hyperparameters)


```

tabular/src/autogluon/tabular/models/ebm/ebm_model.py
```diff
@@ -0,0 +1,241 @@
...
+    def __call__(self, *args, **kwargs):
+        if self.end_time is None:
+            self.end_time = time.monotonic() + self.seconds
+            return False
+        return time.monotonic() > self.end_time
+
+
+class EBMModel(AbstractModel):
+    ag_key = "EBM"
+    ag_name = "EBMModel"
```
Innixma: Some options:

1. "EBMModel"
2. "EBM"
3. "ExplainableBM"
4. "ExplainBM"

Feel free to choose the one you prefer (or suggest an alternative). This will be the name users will see in things such as `predictor.leaderboard()` and will be the name of the saved model artifact directory.

tabular/src/autogluon/tabular/models/ebm/ebm_model.py
```diff
@@ -0,0 +1,241 @@
...
+            self.stopping_metric,
+            num_cpus,
+            time_limit,
+        )
+
+        # Init Class
+        model_cls = get_class_from_problem_type(self.problem_type)
+        self.model = model_cls(random_state=self.random_seed, **params)
+
+        # Handle validation data format for EBM
```
Innixma, line 94: nit: More a comment regarding the EBM package, it might be useful to have a simpler interface for users to pass validation data. No need to address for this PR, just a thought I had.

tabular/src/autogluon/tabular/models/ebm/ebm_model.py
```diff
@@ -0,0 +1,241 @@
...
+    def supported_problem_types(cls) -> list[str] | None:
+        return ["binary", "multiclass", "regression"]
+
+    @classmethod
+    def _class_tags(cls) -> dict:
+        return {"can_estimate_memory_usage_static": True}
+
+    def _more_tags(self) -> dict:
+        """EBMs do not yet support refit full."""
+        return {"can_refit_full": False}
```
Innixma: This might be very easy to add? I think you can switch this to True and it would work immediately, unless validation data is being used for something very important during fit such that you expect performance to be worse without the validation data.

I'd expect the refit_full logic to be quite desirable for this model due to the fast inference speed.

tabular/setup.py
```diff
@@ -49,6 +49,9 @@
     "realmlp": [
         "pytabkit>=1.6,<1.7",
     ],
+    "interpret": [
+        "interpret-core>=0.7.2",
```
Innixma: Lets add a upper bound too. Perhaps `<0.8`?

This mostly depends on how strict you intend to adhere to semantic versioning and backwards compatibility. We try to ensure as best as possible that releases of AutoGluon continue to work long after their original release, and to do so we are often rather strict on upper bounds for packages.

tabular/src/autogluon/tabular/models/ebm/ebm_model.py
```diff
@@ -0,0 +1,241 @@
...
+        self.end_time: float | None = None
+
+    def __call__(self, *args, **kwargs):
+        if self.end_time is None:
+            self.end_time = time.monotonic() + self.seconds
+            return False
+        return time.monotonic() > self.end_time
+
+
+class EBMModel(AbstractModel):
```
Innixma, line 33: Might be good to include a docstring describing the model (refer to the docstrings I added for the other recent models such as TabPFNv2, TabM, RealMLP, etc.)

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -562,6 +563,7 @@ def fit(
                     ],
                     'CAT': {},
                     'XGB': {},
+                    'EBM': {},
```
Innixma: Not (yet) part of the default hyperparameters

```suggestion
```

Innixma (commented):
This looks great @paulbkoch! I added a few comments. Should be good to merge soon after addressing.

tabular/src/autogluon/tabular/models/ebm/ebm_model.py
paulbkoch: I like 'EBM'.  I'm assuming this is for the ag_name attribute?

Updated in the code.

tabular/src/autogluon/tabular/models/ebm/ebm_model.py
paulbkoch: We use bagging in order to generate the error bounds on our graphs (see: https://raw.githubusercontent.com/interpretml/interpret/refs/heads/main/docs/readme/ebm-global.png).  If we were passing in a single validation set then I would use the same interface as other models, but the current method of passing in an array of integers indicating how many times to include each sample in each bag is the best interface I've come up with so far.  It's also very memory efficient, which is a strong consideration.

tabular/src/autogluon/tabular/models/ebm/ebm_model.py
paulbkoch: done

tabular/setup.py
paulbkoch: done

tabular/src/autogluon/tabular/models/ebm/ebm_model.py
paulbkoch: done

paulbkoch:
The build failure in test_common looks like a permissions issue outside of my control.  The failure in test_tabular looks like mine though.  I pushed a fix for the test_tabular failure.

paulbkoch:
@Innixma, @LennartPurucker -- I've completed all the items from @Innixma's review, and cleaned up a few additional things I noticed while looking at the EBM docs.  Unless there's anything else on your end, I believe I'm done with changes to this PR.

tabular/src/autogluon/tabular/models/ebm/ebm_model.py
Innixma: Sounds good!

tabular/src/autogluon/tabular/models/ebm/ebm_model.py
Innixma: Yes, sounds good!

tabular/src/autogluon/tabular/models/ebm/ebm_model.py
```diff
@@ -0,0 +1,263 @@
...
+    @classmethod
+    def supported_problem_types(cls) -> list[str] | None:
+        return ["binary", "multiclass", "regression"]
+
+    @classmethod
+    def _class_tags(cls) -> dict:
+        return {"can_estimate_memory_usage_static": True}
+
+    def _more_tags(self) -> dict:
+        """EBMs do not yet support refit full."""
```
Innixma:
```suggestion
        """EBMs support refit full."""
```

paulbkoch: :)  Thanks!

Issue #4152, opened 2024-04-30
[timeseries] Show model hyperparmeters in the leaderboard


shchur:
## Description
It should be possible to view the model hyperparameters after HPO is performed. For example, this can be done via the `TimeSeriesPredictor.leaderboard` method, similar to how `TabularPredictor.leaderboard` returns such information when `extra_info=True`.

shchur:
Here is a workaround that can be used in v1.1 to get this information 
```python
def get_hyperparameters(model_name: str) -> dict:
    model = predictor._trainer.load_model(model_name)
    if hasattr(model, "most_recent_model"):
        model = model.most_recent_model
    return model._get_params()

leaderboard = predictor.leaderboard()
leaderboard["hyperparameters"] = [get_hyperparameters(m) for m in leaderboard["model"]]
```

dany4142:
Can we please add it as a feature?

shchur:
@dany4142 we will try to include it in the v1.2 release.

shchur:
Completed in #4481

PR #4481, opened 2024-09-20
[timeseries] Align TimeSeriesPredictor.leaderboard API with TabularPredictor

shchur:
*Issue #, if available:* #4152

*Description of changes:*
- Two new arguments to `TimeSeriesPredictor.leaderboard()`:
  - `extra_metrics` allows computing additional metrics on the test set
  - `extra_info` returns hyperparameters selected by hyperparameter tuning for each model

### DIFF
--- a/.github/workflow_scripts/lint_check.sh
+++ b/.github/workflow_scripts/lint_check.sh
@@ -22,6 +22,6 @@ function lint_check_all {
     # lint_check tabular
 }
 
-bandit -r multimodal/src -ll
+bandit -r multimodal/src -ll --skip B614
 lint_check_all
 ruff check timeseries/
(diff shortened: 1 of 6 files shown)


timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -1236,6 +1238,17 @@ def leaderboard(
...
+            If True, the leaderboard will contain an additional column `hyperparameters` with the hyperparameters used
+            by each model during training. An empty dictionary `{}` means that the model was trained with default
+            hyperparameters.
+        extra_metrics : List[Union[str, TimeSeriesScorer]], optional
+            A list of metrics to calculate scores for and include in the output DataFrame.
+
+            Only valid when `data` is specified. The scores refer to the scores on `data` (same data as used to
+            calculate the `score_test` column).
+
+            This list can contain any values which would also be valid for `eval_metric` when creating a `TimeSeriesPredictor`.
```
Innixma, line 1251: nit: Mention the column name they will have in the output. I think this is also in the Tabular docstring

.github/workflow_scripts/lint_check.sh
```diff
@@ -22,6 +22,6 @@ function lint_check_all {
     # lint_check tabular
 }
 
-bandit -r multimodal/src -ll
+bandit -r multimodal/src -ll --skip B614
```
shchur, line 25: @Innixma skipping the new rule that marks `torch.save` / `torch.load` as unsafe to unblock the CI

PR #4483, opened 2024-09-20
[WIP] [tabular] Add class call tracker

Innixma:
*Issue #, if available:*

*Description of changes:*

WIP Draft PR, not ready for review.

Adds logic which can monitor the call trace of a class. This can be extremely powerful for writing unit tests / compatibility / correctness tests, similar to [sklearn.check_estimator](https://scikit-learn.org/stable/modules/generated/sklearn.utils.estimator_checks.check_estimator.html), but far more extensive.

Example call trace output after fitting a LightGBM model, calling predict, and fetching model info:

## Code

```python
def test_lightgbm_binary_model_with_tracer(model_fit_helper):
    fit_args = dict()
    dataset_name = "adult"
    from autogluon.common.testing.track_calls import track_all_calls
    model_cls = track_all_calls(LGBModel)
    model = model_cls()
    model = model_fit_helper.fit_and_validate_dataset(dataset_name=dataset_name, model=model, fit_args=fit_args)
    call_order = model.get_call_order()
    print(call_order)
```

### call_order output

ag_tracer

ag_tracer2

### DIFF
--- a/common/src/autogluon/common/testing/track_calls.py
+++ b/common/src/autogluon/common/testing/track_calls.py
@@ -1,0 +1,96 @@
+import functools
+
+import pandas as pd
+
+
+# TODO: Adjust code doc
+# TODO: Use this for AbstractModel to simplify integration of new models so users can check correctness!
+# TODO: Ex with qualname: "Missing call to AbstractModel.fit, maybe it is overwritten?"
+def track_all_calls(cls):
+    """ Recursively applies call tracking to a class and its parent classes """
+
+    original_methods = {}
+    # Store original methods to avoid double wrapping if parent classes are also processed separately
+    for key, value in vars(cls).items():
+        if callable(value):
+            original_methods[key] = value
+
+    # Function to wrap methods
+    def count_calls(method):
+        @functools.wraps(method)
+        def wrapper(self, *args, **kwargs):
+            if not hasattr(self, "_method_calls"):
+                self._method_calls = {}
+                self._method_call_order = []
+                self._total_calls = 0
+                self._method_depth = 0
+            # name = method.__name__
+            name = method.__qualname__
+            if name not in self._method_calls:
+                self._method_calls[name] = 0
+            self._total_calls += 1
+            cur_depth = self._method_depth
+            self._method_depth += 1
+            cur_calls = self._total_calls
+            self._method_calls[name] += 1
+            pos = len(self._method_call_order)
+            parent_name = None
+            if cur_depth != 0:
+                for i in range(1, pos+1):
+                    i = -i
+                    if self._method_call_order[i][3] == (cur_depth - 1):
+                        parent_name = self._method_call_order[i][0]
+                        # parent_cls = parent_name.split(".", 1)[0]
+                        break
+            else:
+                parent_name = "root"
+            if parent_name is None:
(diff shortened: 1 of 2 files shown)


Innixma:
@eddiebergman, you might find this interesting

eddiebergman (commented):
Very cool! Happy to see some developer tools or AutoGluon! I experimented with something similar a long while back for timing + logging call traces as a `@decorator`. Fun fact incase you weren't aware, the function you have would also work as a decorator.

Some other pointers you might want to know about:
* I never considered it for testing and it was more of a development tool. I since rely on something like [`py-spy`](https://github.com/benfred/py-spy) since it tracked relative time and call order without being invasive. Again, only for personal development and profiling, `py-spy` would not work for tests.
* This might not work in a multi-processing scenario (not certain). Main issue is the inner defined functions which don't have a global namespace which can be pickled. Best case, it would crash and you're aware, worse case, you are silently unaware of the actual behaviour. I believe the `@functool.wrap` resets the namespace to the underlying wrapped definition, (e.g. "module_foo.bar.GoobClass"), meaning that pickle will pick up and export the underlying function/class and omit the wrapped instance of that definition. None of this is an issue if you are aware, it's just that if trying to track calls that will happen in subprocesses, you might get unexpected results.
* I don't know any details but [`sentry`](https://docs.sentry.io/platforms/python/) ([sdk](https://github.com/getsentry/sentry-python)) is a service for web which essentially does a lot of dynamic wrapping for profiling. If there are any edge-cases, they've probably encountered them and could be a good reference to consider.
* Considering recursive behaviour is tricky, namely due to the `callable()` condition. For example, consider you wrap the base class and a subclass, which wrapped definition holds the counts? Might be fine and I have not looked deeply enough to come up with edge-cases. There are more explicit ways to check for methods of a class ([`inspect.ismethod()`](https://docs.python.org/3/library/inspect.html#inspect.ismethod)) that might help prevent this. Not an actual problem is used correctly but worth documenting if you plan to expose this to external users.

Also one other thing you might want to track is the caller through [`inspect.stack()`](https://stackoverflow.com/a/8663885/5332072). Pretty sure it's the same data Python uses for generating tracebacks and call sites for `warnings.warn(..., stacklevel=<num>)`

eddiebergman:
Just some additional commentary on `py-spy`. If your goal is for profiling to find runtime optimizations, I highly recommend it. It's non-invasive and will actually read the memory of the python interpreter and log it in a seperate process (Rust btw). It's very specific to the use case and less flexible than what you've written, but as a benefit it can go and also track internal C calls (numpy, torch, etc...).
     * Bonus: As a result of `py-spy`, I found out if caching indices and it matters time wise, you should use `np.uintp` or `np.intp` (p for pointer) instead of `np.int64` :) [Saved some whole milli-seconds lol](https://github.com/automl/ConfigSpace/blob/d5b82ae3767f389d340e7a1b2d24a04ea3669f0f/src/ConfigSpace/_condition_tree.py#L97-L101)
     * Combing your method to identify high-traffic hot-paths with some further optimizations with `py-py` could be beneficial for ensembling optimization, if that's something worth shaving time off

Innixma:
Closing to clean up the open PR queue. I might revisit this in future, but currently lack bandwidth to make it all come together nicely.

PR #4485, opened 2024-09-20
Fix py38 import

Innixma:
*Issue #, if available:*

*Description of changes:*

A recent merged commit leads to a import error with Python 3.8 due to the usage of `|` type hints, which were introduced in Python 3.9. This PR fixes this by adding `from __future__ import annotations` for backwards compatibility.

### DIFF
--- a/core/src/autogluon/core/utils/utils.py
+++ b/core/src/autogluon/core/utils/utils.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import logging
 import math
 import pickle


suzhoum (approved):
Thanks for the fix! In future, do we still need to add this on top of every file that we used the type hints? Or is it pre-imported on package level?

Innixma:
@suzhoum I believe it needs to be at the top of every file that uses the type hints. Once we stop supporting Python 3.8, the frequency of needing to do this will lessen

PR #4495, opened 2024-09-27
[tabular] Make refit_full respect user `num_cpus` and `num_gpus`

### LINKED ISSUE #3871: [BUG] Refit Does Not Respect Resource Constraints (AutoGluon-Tabular).
**Bug Report Checklist**

- [X] I confirmed bug exists on the latest stable version of AutoGluon.  

**Describe the bug**
While digging through the code to debug my Ray issues, I found that refitting has severe resource management issues. I have not yet fully understood the problem, but when you, for example, tell AutoGluon to use 4 CPUs and 4 GPUs, then AutoGluon will only use these resources with Ray during a regular call to fit. However, during refitting (after fit or after cross-validation via `refit_folds`), Ray no longer respects these constraints and defaults to use all available CPUs (and no GPUs).

I hypothesize that Ray no longer respects these resources, as they are never passed to Ray during refitting calls. This could be fixed by setting them in the `core_kwargs` in calls to fit_single methods during the refit.

Sadly, I do not have a reproducible example for this because it heavily depends on my SLURM cluster setup.

**Expected behavior**
Refit only uses the resource constraints specified at fit (or refit) time.

Innixma:
*Issue #, if available:*

Resolves #3871

Follow-up #4496 

*Description of changes:*

- Make refit_full respect user `num_cpus` and `num_gpus`.
- Make WeightedEnsemble respect user `num_cpus` and `num_gpus`.
- User can now check the resources used to fit each model via `predictor.info()`

Now when the user specifies refit_full along with resource constraints, those resource constraints are respected during refit_full:

```python
predictor = TabularPredictor(label='class').fit(
    train_data=train_data, num_cpus=17, num_gpus=0, refit_full=True
)
```

The above code would previously fit the original models with the constraints, and then during refit_full it would ignore the constraints. For example, if you had a GPU, it would forcibly use the GPU during refit.

With this PR, this is no longer the case, and the refit models respect the constraints.

## Mainline

```
Fitting model: NeuralNetTorch_BAG_L1 ...
	Fitting NeuralNetTorch_BAG_L1 with 'num_gpus': 0, 'num_cpus': 17
	Fitting WeightedEnsemble_L2 with 'num_gpus': 0, 'num_cpus': 96
Fitting model: NeuralNetTorch_BAG_L1_FULL ...
	Fitting NeuralNetTorch_BAG_L1_FULL with 'num_gpus': 4, 'num_cpus': 96
Fitting model: WeightedEnsemble_L2_FULL | Skipping fit via cloning parent ...
```

## This PR

```
Fitting model: NeuralNetTorch_BAG_L1 ...
	Fitting NeuralNetTorch_BAG_L1 with 'num_gpus': 0, 'num_cpus': 17
	Fitting WeightedEnsemble_L2 with 'num_gpus': 0, 'num_cpus': 17
Fitting model: NeuralNetTorch_BAG_L1_FULL ...
	Fitting NeuralNetTorch_BAG_L1_FULL with 'num_gpus': 0, 'num_cpus': 17
Fitting model: WeightedEnsemble_L2_FULL | Skipping fit via cloning parent ...
```

### DIFF
--- a/tabular/src/autogluon/tabular/predictor/predictor.py
+++ b/tabular/src/autogluon/tabular/predictor/predictor.py
@@ -1192,7 +1192,12 @@ class TabularPredictor(TabularPredictorDeprecatedMixin):
             "excluded_model_types": excluded_model_types,
             "feature_prune_kwargs": kwargs.get("feature_prune_kwargs", None),
         }
-        aux_kwargs = {}
+        aux_kwargs = {
+            "total_resources": {
+                "num_cpus": num_cpus,
+                "num_gpus": num_gpus,
+            },
+        }
         if fit_weighted_ensemble is False:
             aux_kwargs["fit_weighted_ensemble"] = False
         aux_kwargs["fit_full_last_level_weighted_ensemble"] = fit_full_last_level_weighted_ensemble
@@ -1791,7 +1796,12 @@ class TabularPredictor(TabularPredictorDeprecatedMixin):
         ag_args = self._set_hyperparameter_tune_kwargs_in_ag_args(kwargs["hyperparameter_tune_kwargs"], ag_args, time_limit=time_limit)
 
         fit_new_weighted_ensemble = False  # TODO: Add as option
-        aux_kwargs = {}
+        aux_kwargs = {
+            "total_resources": {
+                "num_cpus": num_cpus,
+                "num_gpus": num_gpus,
+            },
+        }
         if fit_weighted_ensemble is False:
             aux_kwargs = {"fit_weighted_ensemble": False}
         aux_kwargs["fit_full_last_level_weighted_ensemble"] = fit_full_last_level_weighted_ensemble
@@ -3663,6 +3673,11 @@ class TabularPredictor(TabularPredictorDeprecatedMixin):
             y = trainer.load_y_val()
             fit = False
 
+        total_resources = {
+            "num_cpus": num_cpus,
+            "num_gpus": num_gpus,
+        }
+
         stack_name = "aux1"
         if base_models is None:
             base_models = trainer.get_model_names(stack_name="core")
@@ -3688,6 +3703,7 @@ class TabularPredictor(TabularPredictorDeprecatedMixin):
                     base_model_names=models_to_check_now,
                     name_suffix=name_suffix + "_Pareto" + str(i),
                     time_limit=time_limit,
+                    total_resources=total_resources,
                 )
 
         max_base_model_level = max([trainer.get_model_level(base_model) for base_model in base_models])
@@ -3700,6 +3716,7 @@ class TabularPredictor(TabularPredictorDeprecatedMixin):
             base_model_names=base_models,
             name_suffix=name_suffix,
             time_limit=time_limit,
+            total_resources=total_resources,
         )
 
         if refit_full:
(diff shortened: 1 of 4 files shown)


tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -3569,7 +3603,16 @@ def info(self):
     # TODO: Move code logic to learner/trainer
     # TODO: Add fit() arg to perform this automatically at end of training
     # TODO: Consider adding cutoff arguments such as top-k models
-    def fit_weighted_ensemble(self, base_models: list = None, name_suffix="Best", expand_pareto_frontier=False, time_limit=None, refit_full=False):
+    def fit_weighted_ensemble(
+        self,
+        base_models: list = None,
+        name_suffix="Best",
```
lovvge, line 3619: nitpick on consistency: there are arguments with / without type, and with / without space around =

tabular/tests/unittests/models/test_advanced.py
```diff
@@ -17,3 +18,44 @@ def test_bagged_predict_children(model_fit_helper):
         check_predict_children=True,
         sample_size=100,
     )
+
+
+def test_resource_constraints(fit_helper, dataset_loader_helper):
+    """
+    Verify that num_cpus and num_gpus are respected when specified in the fit call.
+    Also verifies that constraints are respected for weighted ensemble models and during refit_full.
+    Also verifies that specifying num_cpus > max_cpus will use max_cpus.
```
lovvge, line 27: just curious, why only testing this for CPU but not for GPU?

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -3441,13 +3477,21 @@ def refit_full(
...
+        }
+
         if train_data_extra is not None:
             assert kwargs.get("X_pseudo", None) is None, f"Cannot pass both train_data_extra and X_pseudo arguments"
             assert kwargs.get("y_pseudo", None) is None, f"Cannot pass both train_data_extra and y_pseudo arguments"
             X_pseudo, y_pseudo, _ = self._sanitize_pseudo_data(pseudo_data=train_data_extra, name="train_data_extra")
             kwargs["X_pseudo"] = X_pseudo
             kwargs["y_pseudo"] = y_pseudo
-        refit_full_dict = self._learner.refit_ensemble_full(model=model, **kwargs)
+        refit_full_dict = self._learner.refit_ensemble_full(model=model, total_resources=total_resources, **kwargs)
```
suzhoum, line 3494: Is `total_resources` getting used inside the `AbstractTabularLearner.refit_ensemble_full`?

tabular/src/autogluon/tabular/predictor/predictor.py
Innixma: Good point. I was initially not planning to touch the original code when adding the additional arguments beyond satisfying the linter, but I have went and updated to add type hints.

Regarding the spaces, this is because when you do type hints, the correct linting is to add spaces, but without type hint, you should not have spaces.

tabular/src/autogluon/tabular/predictor/predictor.py
Innixma: Yes, it is passed as a kwarg down into the trainer's fit logic

tabular/tests/unittests/models/test_advanced.py
Innixma: We don't have GPU tests for tabular currently. This is something I hope to add in 1.2 release if we find time to dive into better GPU support. I have created #4500 to track as well as added a TODO in the unit test.

prateekdesai04 (approved):
LGTM!

PR #4516, opened 2024-10-03
[Master] Update installation instructions to use uv temporarily

tonyhoo:
This change is a temporary measure to mitigate pip installation issues.

*Issue #, if available:* #4515

*Description of changes:*

- Updated installation instructions to use uv instead of pip
- Added uv installation step in all installation documentation
- Replaced pip install commands with uv pip install
- Updated full_install.sh to check for uv and install it if not present
- Used uv for installing AutoGluon and its dependencies

This change addresses the recent installation failures reported in issue #4515, where users are encountering problems with onnx and optimum dependencies. By using uv, we aim to provide a more stable installation process until a permanent solution can be implemented.

### DIFF
--- a/full_install.sh
+++ b/full_install.sh
@@ -1,10 +1,19 @@
 #!/usr/bin/env bash
 set -euo pipefail
-python3 -m pip install -e common/[tests]
-python3 -m pip install -e core/[all,tests]
-python3 -m pip install -e features/
-python3 -m pip install -e tabular/[all,tests]
-python3 -m pip install -e multimodal/[tests]
-python3 -m pip install -e timeseries/[all,tests]
-python3 -m pip install -e eda/
-python3 -m pip install -e autogluon/
+
+# Check if uv is installed
+if ! command -v uv &> /dev/null
+then
+    echo "uv could not be found. Installing uv..."
+    python3 -m pip install uv
+fi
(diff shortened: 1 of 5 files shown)


full_install.sh
```diff
@@ -1,10 +1,19 @@
...
-python3 -m pip install -e core/[all,tests]
-python3 -m pip install -e features/
-python3 -m pip install -e tabular/[all,tests]
-python3 -m pip install -e multimodal/[tests]
-python3 -m pip install -e timeseries/[all,tests]
-python3 -m pip install -e eda/
-python3 -m pip install -e autogluon/
+
+# Check if uv is installed
+if ! command -v uv &> /dev/null
```
suzhoum, line 5: Would there be an issue if we pre-install `uv` in this script?

tonyhoo: We have to pre-install `uv` if it is not available in the system, otherwise pip install won't work

suzhoum (approved):

Innixma (approved):
LGTM! I especially like using uv in the `full_install.sh`, should make it a lot faster

Innixma:
Note: I wonder if we can speed up `full_install.sh` by doing all the installs in a single line with uv.

PR #4517, opened 2024-10-03
[Stable] Update installation instructions to use uv temporarily

tonyhoo:
This change is a temporary measure to mitigate pip installation issues.

*Issue #, if available:* #4515

*Description of changes:*

- Updated installation instructions to use uv instead of pip
- Added uv installation step in all installation documentation
- Replaced pip install commands with uv pip install
- Updated full_install.sh to check for uv and install it if not present
- Used uv for installing AutoGluon and its dependencies

This change addresses the recent installation failures reported in issue #4515, where users are encountering problems with onnx and optimum dependencies. By using uv, we aim to provide a more stable installation process until a permanent solution can be implemented.

### DIFF
--- a/.github/workflow_scripts/lint_check.sh
+++ b/.github/workflow_scripts/lint_check.sh
@@ -20,6 +20,6 @@ function lint_check_all {
     lint_check tabular
 }
 
-bandit -r multimodal/src -ll
+bandit -r multimodal/src -ll --skip B614
 lint_check_all
 ruff check timeseries/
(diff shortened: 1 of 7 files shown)


prateekdesai04 (approved):

Innixma (commented):
I think you will need to remove `ray` and only keep `raytune` in the optional dependenceis used in `core[all]` to avoid the installation error seen in the CI

PR #4518, opened 2024-10-03
[Bug] Bump optimum version for AG-Timeseries

prateekdesai04:
*Issue #, if available:* https://github.com/autogluon/autogluon/issues/4515

*Description of changes:*


timeseries/setup.py
```diff
@@ -58,7 +58,7 @@
         "optimum[openvino,nncf]>=1.17,<1.19",
     ],
     "chronos-onnx": [  # for faster CPU inference in pretrained models with ONNX
-        "optimum[onnxruntime]>=1.17,<1.19",
+        "optimum[onnxruntime]>=1.20,<1.22.1",
```
Innixma, line 61: Maybe as an alternative or additionally: we should add a version range to `onnx` here as well, since `optimum` doesn't provide an onnx version range: https://github.com/huggingface/optimum/blob/main/setup.py#L51-L58

PR #4520, opened 2024-10-03
[Bug] Cap pyarrow to fix source installation issue

prateekdesai04:
*Issue #, if available:*
https://github.com/autogluon/autogluon/issues/4519

*Description of changes:*
Source installation now works on every supported version of Python - 3.9, 3.10, 3.11.

### DIFF
--- a/core/setup.py
+++ b/core/setup.py
@@ -51,6 +51,7 @@ extras_require = {
         "ray[default]>=2.10.0,<2.32",  # sync with common/src/autogluon/common/utils/try_import.py
     ],
     "raytune": [
+        "pyarrow>=15.0.0,<17.0.1",  # cap Pyarrow to fix source installation
         "ray[default,tune]>=2.10.0,<2.32",  # sync with common/src/autogluon/common/utils/try_import.py
         # TODO: consider alternatives as hyperopt is not actively maintained.
         "hyperopt>=0.2.7,<0.2.8",  # This is needed for the bayes search to work.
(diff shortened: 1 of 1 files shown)


tonyhoo (approved):
LGTM and please push once all tests passed

core/setup.py
```diff
@@ -51,6 +51,7 @@
         "ray[default]>=2.10.0,<2.32",  # sync with common/src/autogluon/common/utils/try_import.py
     ],
     "raytune": [
+        "pyarrow>=17.0.0",  # cap Pyarrow to fix source installation - https://github.com/autogluon/autogluon/issues/4519
```
suzhoum, line 54: I think 17.0.0 is the latest version. Is it safer to cap it to a lower version? I tested that capping to `12.0.0` also works.

prateekdesai04: I tested all python versions on 17.0 and they passed.
@Innixma suggested that we keep this cap

suzhoum (approved):

core/setup.py
tonyhoo: Also cap it with upperbound as well in case future release broke autogluon

Innixma: @tonyhoo since raytune depends on pyarrow, this is an implicit dependency, and we rely on ray to set a proper upperbound. This is only installed when paired with ray to limit the version ranges pip will search through. Therefore, I don't think we need an upperbound.

Innixma:
@prateekdesai04 Maybe we can put lower bound to 15.0.0, to be a bit more flexible

prateekdesai04:
@Innixma lowered the bound

Innixma (approved):
LGTM!

PR #4527, opened 2024-10-07
[tabular] Enhance predict_proba_multi

Innixma:
*Issue #, if available:*

*Description of changes:*

- Update predict_proba_multi docstring
- Update predict_multi docstring, add overrides to improve type hinting based on the user's value of `as_pandas`
- Added helper arg to trainer.get_model_oof to use parent model if refit for OOF loading
- Added logic to ensure predict_proba_multi will not crash if getting OOF for refit models. Previously it would crash.
- Added unit test to verify that the logic works correctly.

### DIFF
--- a/tabular/src/autogluon/tabular/learner/abstract_learner.py
+++ b/tabular/src/autogluon/tabular/learner/abstract_learner.py
@@ -267,6 +267,7 @@ class AbstractTabularLearner(AbstractLearner):
         as_multiclass: bool = True,
         transform_features: bool = True,
         inverse_transform: bool = True,
+        use_refit_parent_oof: bool = True,
     ) -> dict:
         """
         Returns a dictionary of prediction probabilities where the key is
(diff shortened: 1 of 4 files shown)


tabular/src/autogluon/tabular/learner/abstract_learner.py
```diff
@@ -302,6 +303,8 @@ def predict_proba_multi(
         inverse_transform : bool, default = True
             If True, will return prediction probabilities in the original format.
             If False (advanced), will return prediction probabilities in AutoGluon's internal format.
+        use_refit_parent_oof: bool = True
```
prateekdesai04, line 306: unrelated but `_post_process_predict()` and `_post_process_predict_proba()` is what we do in TabRepo's abstract class as well if my understanding is correct?

prateekdesai04 (approved):

tabular/src/autogluon/tabular/learner/abstract_learner.py
Innixma: yes, pretty much

PR #4533, opened 2024-10-09
[timeseries] Fix refit_every_n_windows=None raising an exception if num_val_windows=1

shchur:
*Issue #, if available:*

*Description of changes:*
- Currently, calling `predictor.fit(..., num_val_windows=1, refit_every_n_windows=None)` raises an exception. This PR fixes the bug.

### DIFF
--- a/timeseries/src/autogluon/timeseries/predictor.py
+++ b/timeseries/src/autogluon/timeseries/predictor.py
@@ -416,7 +416,7 @@ class TimeSeriesPredictor(TimeSeriesPredictorDeprecatedMixin):
         excluded_model_types: Optional[List[str]] = None,
         num_val_windows: int = 1,
         val_step_size: Optional[int] = None,
-        refit_every_n_windows: int = 1,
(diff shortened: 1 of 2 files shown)


canerturkmen (approved):
LGTM! Thanks!

trenchantai:
Hey Andrew,
I've got a quick question related to this.  I've read through all of the documentation from Autogluon and it still isn't clear to me.

When I set the prediction_length in TimeSeriesPredictor, is it evaluating each timestep independently and using all available information up to that timestep?  So, a simple use case, currently I'm forecasting hourly data and I want to compare the actual to the predicted values for 24 hours into the "future".  Currently I set prediction_length = 1 and loop through the train data one timestep at a time to be sure it has all of the historical information up to each time step.  This feels cludgy and I haven't seen anyone else do it but the forecast seems  acceptable.  If I try setting the prediction_length = 24. the prediction is completely wrong.
Please let me know what I'm missing and thanks in advance for any guidance you can offer.

Issue #4522, opened 2024-10-04
[BUG] [timeseries] Deep learning models fail if `item_id` is of type `string[python]`


shchur:
**Bug Report Checklist**

- [x] I provided code that demonstrates a minimal reproducible example.  
- [x] I confirmed bug exists on the latest mainline of AutoGluon via source install.  

**Describe the bug**
If the `item_id` column is of type `string[python]` (i.e., `pd.StringDtype()`), all deep learning models fail during training

**Expected behavior**
Models train normally.

**To Reproduce**
```python
N = 100
df = pd.DataFrame(
    {
        "item_id": ["A"] * N,
        "timestamp": pd.date_range("2020-01-01", freq="D", periods=N),
        "target": range(N),
    },
).astype({"item_id": pd.StringDtype()})

tsdf = TimeSeriesDataFrame(df)
TimeSeriesPredictor(prediction_length=5).fit(tsdf, hyperparameters={"DeepAR": {"epochs": 2}})
```

**Screenshots / Logs**
```
...
Models that will be trained: ['DeepAR']
Training timeseries model DeepAR. 

        Warning: Exception caused DeepAR to fail during training... Skipping this model.
        'IntegerArray' object has no attribute 'cumsum'
```

**Installed Versions**
Current master branch.

shchur:
Current workaround for this problem: Convert `item_id` to `str` dtype before creating the `TimeSeriesDataFrame`:
```python
df = pd.DataFrame(...)
tsdf = TimeSeriesDataFrame.from_data_frame(df.astype({"item_id": "str"}))
```

rsj123:
The problem is caused in 

https://github.com/autogluon/autogluon/blob/65d8d6e6017d7e175459ea8ac0eae43375ecb489/timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py#L72C3-L78C7
There are the min code to Reproduce
``` python
import pandas as pd
from autogluon.timeseries import TimeSeriesDataFrame

N=100
df = pd.DataFrame(
    {
        "item_id": ["A"] * N,
        "timestamp": pd.date_range("2020-01-01", freq="D", periods=N),
        "target": range(N),
    },
).astype({"item_id": pd.StringDtype()})
tsdf = TimeSeriesDataFrame(df)
ITEMID = "item_id"
item_id_index = tsdf.index.get_level_values(ITEMID)
indices_sizes = item_id_index.value_counts(sort=False)
[...]

PR #4539, opened 2024-10-13
[timeseries] Fix IntegerArray to ndarray

rsj123:
Fix IntegerArray to ndarray Conversion in Value Counts Cumsum Calculation

*Issue #, if available:*
#4522 

*Description of changes:*
Fix IntegerArray to ndarray Conversion in Value Counts Cumsum Calculation


shchur (approved):
Hi @rsj123, thanks a lot for identifying the source of the problem and opening the PR! I've added a unittest to make sure that we don't encounter similar issues in the future. LGTM!

PR #4540, opened 2024-10-15
[timeseries] Update logic for inferring the time series frequency

### LINKED ISSUE #4372: [BUG] TimeSeriesPredictor fails to train any models if timestamps are irregular/duplicated for item_ids above 100
**Bug Report Checklist**

- [x] I confirmed bug exists on the latest mainline of AutoGluon via source install.  

**Describe the bug**
Current frequency inference mechanism in AutoGluon only checks the frequency of the first 100 items ([code](https://github.com/autogluon/autogluon/blob/master/timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py#L491)). Therefore, if items beyond the first 100 have irregular frequency or duplicated timestamps, AutoGluon will not catch this problem. This, however, will result in all models failing when computing the scores.

**Expected behavior**
An informative error indicating the data problem is raised.

**To Reproduce**
```python
import pandas as pd

def get_df(i, N=100, freq="D"):
    return pd.DataFrame({"item_id": i, "timestamp": pd.date_range("2020-01-01", freq=freq, periods=N), "target": range(N)})

# Entries for item_ids 100-150 are duplicated
df = TimeSeriesDataFrame(pd.concat([get_df(i) for i in range(150)] + [get_df(i) for i in range(100, 150)]))

pred = TimeSeriesPredictor().fit(df, hyperparameters={"Naive": {}})
```

**Screenshots / Logs**
```
Beginning AutoGluon training...
AutoGluon will save models to 'AutogluonModels/ag-20240809_111936'
=================== System Info ===================
AutoGluon Version:  1.1.0
Python Version:     3.10.14
Operating System:   Linux
Platform Machine:   x86_64
Platform Version:   #1 SMP Tue Jul 30 16:40:37 UTC 2024
CPU Count:          32
GPU Count:          0
Memory Avail:       230.51 GB [/](https://vscode-remote+ssh-002dremote-002bdesk.vscode-resource.vscode-cdn.net/) 239.85 GB (96.1%)
Disk Space Avail:   988.23 GB [/](https://vscode-remote+ssh-002dremote-002bdesk.vscode-resource.vscode-cdn.net/) 1968.52 GB (50.2%)
===================================================

Fitting with arguments:
{'enable_ensemble': True,
 'eval_metric': WQL,
 'hyperparameters': {'Naive': {}},
 'known_covariates_names': [],
 'num_val_windows': 1,
 'prediction_length': 1,
 'quantile_levels': [0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9],
 'random_seed': 123,
 'refit_every_n_windows': 1,
 'refit_full': False,
 'skip_model_selection': False,
 'target': 'target',
 'verbosity': 2}

Inferred time series frequency: 'D'
Provided train_data has 20000 rows, 150 time series. Median time series length is 100 (min=100, max=200). 

Provided data contains following columns:
	target: 'target'

AutoGluon will gauge predictive performance using evaluation metric: 'WQL'
	This metric's sign has been flipped to adhere to being higher_is_better. The metric score can be multiplied by -1 to get the metric value.
===================================================

Starting training. Start time is 2024-08-09 11:19:36
Models that will be trained: ['Naive']
Training timeseries model Naive. 
	Warning: Exception caused Naive to fail during training... Skipping this model.
	Prediction and data indices do not match.
Not fitting ensemble as no models were successfully trained.
Training complete. Models trained: []
Total runtime: 0.31 s
Trainer has no fit models that can predict.
```
**Installed Versions**
AG v1.1.1 / master branch

</details>

shchur:
*Issue #, if available:*  Fixes #4372

*Description of changes:*
- Move the logic for frequency inference from the attribute `TimeSeriesDataFrame.freq` to a new method `TimeSeriesDataFrame.infer_frequency()`. The new method allows us to control how many items are used for frequency inference, and whether an informative error should be raised if frequency inference fails.
  - The new method now also randomly selects the items for checking the frequency instead of always using the first 100 items.
- Ensure that the frequency is checked for all items in `train_data` during `predictor.fit()` to avoid failures described in #4372.
- Extend test coverage for irregular frequencies to include cases when 1/ some time series have less than 3 observations, 2/ different items have different frequencies.


timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
```diff
@@ -476,28 +477,82 @@ def static_features(self, value: Optional[pd.DataFrame]):
...
+
+            If some items have an irregular frequency or if different items have different frequencies, returns string
+            `IRREG`.
+        """
+
+        df = pd.DataFrame(self)
+        if num_items is not None:
+            all_item_ids = self.item_ids
+            if len(all_item_ids) > num_items:
+                items_subset = all_item_ids.to_series().sample(n=num_items, random_state=123)
```
abdulfatir, line 507: Probably overkill but should random_state be also customizable?

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -299,7 +299,10 @@ def _check_and_prepare_data_frame(
 
         # Ensure that data has a regular frequency that matches the predictor frequency
         if self.freq is None:
-            if df.freq is None:
+            try:
+                # Use all items for inferring the frequency
+                self.freq = df.infer_frequency(num_items=None, raise_if_irregular=True)
+            except ValueError:
```
abdulfatir, line 305: Should we pass on the raised error forward for details? This will help debugging.

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
shchur: I don't think a user will ever need to access it. If submsampling with our fixed seed failed to detect problems, the user can run `df.infer_frequency(num_items=None)` to compute the frequency on all items in a deterministic and 100% reliable way. 

Do you think that's reasonable?

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
```diff
@@ -476,28 +477,82 @@ def static_features(self, value: Optional[pd.DataFrame]):
...
+                        f"{pformat(items_with_irregular_freq.index.tolist())}"
+                    )
+                else:
+                    raise ValueError(
+                        "Cannot infer frequency. Multiple frequencies detected in the dataset: "
+                        f"{freq_for_each_item.unique().tolist()}"
+                    )
+            return IRREGULAR_TIME_INDEX_FREQSTR
+        else:
+            return freq.freqstr if isinstance(freq, pd._libs.tslibs.BaseOffset) else freq
```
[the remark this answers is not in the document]
shchur: Good point, replaced with public API only.

timeseries/src/autogluon/timeseries/predictor.py
shchur: This should already happen by default. 

When we raise an exception inside the `except` block, the previous exception is also forwarded (at least in Python 3.11):
```python
try:
    raise ValueError("Hello")
except:
    raise RuntimeError("World")
```
Output:
image

The first exception would only be silenced if we changed the code to
```python
except:
    raise RuntimeError("World") from None
```

abdulfatir: Right, thanks

abdulfatir (approved):

PR #4552, opened 2024-10-17
[tabular] Add support for controlling AutoGluon's repeated cross-validation behavior

LennartPurucker:
> [!CAUTION]
> This PR changes the default behavior **if** a user manually has set `num_bag_sets != 1`

### Description of changes

This PR enables support for controlling when bagging is repeated in cross-validation.

### Argument For Change In Default Behavior

I also suggest changing the default behavior for two reasons:

- it is very unintuitive that if a user increases the number of bag sets (i.e., repeats) under a time limit, they might not even see a change in the results due to the entire portfolio being fit first;
- if we have enabled repeats, it is likely more important to do more repeats than evaluate more models to avoid overfitting, especially for small data;
- the current default AutoGluon behavior does not change, as we have  `num_bag_sets == 1` since AutoGluon 1.0.

Happy to discuss this in more detail. 

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution under the terms of your choice.

### DIFF
--- a/core/src/autogluon/core/trainer/abstract_trainer.py
+++ b/core/src/autogluon/core/trainer/abstract_trainer.py
@@ -2675,6 +2675,7 @@ class AbstractTrainer:
         n_repeats=None,
         n_repeat_start=0,
         time_limit=None,
+        delay_bag_sets: bool = False,
         **kwargs,
     ) -> List[str]:
         """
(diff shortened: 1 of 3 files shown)


Innixma (approved):

PR #4557, opened 2024-10-18
[timeseries] Optimize TimeSeriesFeatureGenerator preprocessing logic

shchur:
*Issue #, if available:*

*Description of changes:*

This PR optimizes the runtime and memory consumption of the `fit` and `transform` methods of the `TimeSeriesFeatureGenerator`. 

For example, comparison on the M5 dataset (30K items, 47M rows, 7 known real covariates, 1 past real covariate, 5 static categorical features)

  | Before this PR |   | This PR |  
-- | -- | -- | -- | --
method name | Peak mem usage (GB) | Runtime (s) | Peak mem usage (GB) | Runtime (s)
fit() | 21.6 | 26.29 | 11.8 | 15.18
fit_transform() | 21.6 | 49.09 | 11.8 | 15.18
transform() | 16.6 | 23.13 | 10.4 | 7.41

Previously, the fit method resulted in OOM errors on machines with <16 GB RAM, which is now fixed.

The specific changes primarily involve:
- Avoiding duplicate operations during `fit_predict`
- Using `float32` dtype for the covariates (since anyway all models that consume covariates convert them to float32 during preprocessing)
- Work with `pd.DataFrame`s without a `MultiIndex` during intermediate steps
- Using inplace operations, where possible
- Converting `bool` attributes to `float` instead of categoricals


timeseries/src/autogluon/timeseries/utils/features.py
```diff
@@ -73,7 +74,7 @@ class ContinuousAndCategoricalFeatureGenerator(PipelineFeatureGenerator):
     Imputes missing categorical features with the most frequent value in the training set.
     """
 
-    def __init__(self, verbosity: int = 0, minimum_cat_count=2, float_dtype: str = "float64", **kwargs):
```
abdulfatir, line 76: Why not keep the `float_dtype` param and default it ti float32 instead?

timeseries/src/autogluon/timeseries/utils/features.py
```diff
@@ -209,9 +227,47 @@ def fit(self, data: TimeSeriesDataFrame) -> None:
...
+        # Median of real-valued covariates will be used for missing value imputation
+        if self.num_samples is not None and len(df) > self.num_samples:
+            df = df.sample(n=self.num_samples, replace=True)
+        self._train_covariates_real_median = df[self.covariate_metadata.covariates_real].median()
+
+        self.fit_time = time.monotonic() - start_time
         self._is_fit = True
 
+        df_out = self._concat_dfs(dfs_to_concat)
+        df_out.index = index
```
abdulfatir, line 240: Are we sure that no prior ops change the index? Doesn't look like it.

timeseries/src/autogluon/timeseries/utils/features.py
shchur: I've moved the logic for converting numeric columns to `float_dtype` from `ContinuousAndCategoricalFeatureGenerator` (a separate instance is created for past, known, and static covariates) to `TimeSeriesFeatureGenerator` (a single object processes all columns). This way we reduced the number of `df.astype()` calls from 2 to 1, decreasing the runtime and reducing the memory usage.

timeseries/src/autogluon/timeseries/utils/features.py
shchur: Do you mean that `df_out` has the same rows / in the same order as the original `df`? I am almost certain that this is the case since none of the steps in `ContinuousAndCategoricalFeatureGenerator` change the order of the rows (if they did, this would also break AutoGluon-Tabular). 

If you rather mean "are we sure that the `index` object (defined on [L163](https://github.com/autogluon/autogluon/pull/4557/files#diff-ac94905e2a5e94bcd2c6b86f47354f212aa012314125dbad34c9d196fa432b12R163)) is not modified by any transformations" - I've just added an extra assertion to the [test](https://github.com/autogluon/autogluon/pull/4557/files#diff-a634dd5750d533af15c67f8e7cb99046d5156bf43c70de32216a8ff4136dbb5cR123) to ensure that the index object is not modified.

abdulfatir: I meant the former. Thanks for adding the test.

PR #4559, opened 2024-10-18
[tabular] Pipeline presets with automatic cross-validation preset

LennartPurucker:
## Description of changes:

A refactor included in this PR introduces the first version of a "control center" in the form of `pipeline_presets` for all ML pipeline-related variables we might want to control before running AutoGluon. 

The pipeline presets can allow us to control, hopefully, all aspects of how we run the AutoML system in the future, with options such as the validation method, when to use (dynamic) stacking, and whether to refit. 

In addition, the PR includes a first proposal for adding a heuristic selection of validation methods to this control center. I believe this heuristic would already be very impactful for users, as this can avoid overfitting or speeding up training.

Obviously, we need to benchmark this and, ideally, in the future, learn which AutoML configuration to use and when. 

The PR only changes the default behavior if `auto_stack is True` (e.g., for `best_quality`) 

---

### DIFF
--- a/core/src/autogluon/core/utils/utils.py
+++ b/core/src/autogluon/core/utils/utils.py
@@ -184,7 +184,7 @@ def normalize_multi_probas(y_predprob, eps):
     return y_predprob
 
 
-def default_holdout_frac(num_train_rows, hyperparameter_tune: bool =False):
+def default_holdout_frac(num_train_rows, hyperparameter_tune: bool = False):
     """Returns default holdout_frac used in fit().
     Between row count 5,000 and 25,000 keep 0.1 holdout_frac, as we want to grow validation set to a stable 2500 examples.
     """
--- a/core/src/autogluon/core/utils/utils.py
+++ b/core/src/autogluon/core/utils/utils.py
@@ -184,7 +184,7 @@ def normalize_multi_probas(y_predprob, eps):
     return y_predprob
 
 
-def default_holdout_frac(num_train_rows, hyperparameter_tune=False):
+def default_holdout_frac(num_train_rows, hyperparameter_tune: bool =False):
     """Returns default holdout_frac used in fit().
     Between row count 5,000 and 25,000 keep 0.1 holdout_frac, as we want to grow validation set to a stable 2500 examples.
     """
--- a/tabular/src/autogluon/tabular/configs/pipeline_presets.py
+++ b/tabular/src/autogluon/tabular/configs/pipeline_presets.py
@@ -74,7 +74,7 @@ def get_validation_and_stacking_method(
 ) -> tuple[int, int, int, bool, bool, float, bool]:
     """Get the validation method for AutoGluon via a heuristic.
 
-    Input variables are `None` if they were not specified by the user or have an explict default.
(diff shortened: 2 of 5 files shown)


tabular/src/autogluon/tabular/configs/pipeline_presets.py
```diff
@@ -0,0 +1,149 @@
...
+        num_bag_sets = 4
+    elif num_train_rows < 10_000:
+        num_bag_sets = 2
+    elif num_train_rows < USE_BAG_HOLDOUT_AUTO_THRESHOLD:
+        # TODO(improvement): go towards 4-fold cv for more than 100k rows?
+        pass
+    else:
+        # Only use holdout validation afterward
+        num_bag_folds = 0
+        num_bag_sets = 0
```
Innixma, line 35: interesting idea, however this would presumably disable stacking then. Maybe that is ok?

Innixma: `USE_BAG_HOLDOUT_AUTO_THRESHOLD` still uses bagging, it just enables `use_bag_holdout`, so this is a change in logic. Unsure if intended

core/src/autogluon/core/utils/utils.py
```diff
@@ -184,7 +184,7 @@ def normalize_multi_probas(y_predprob, eps):
     return y_predprob
 
 
-def default_holdout_frac(num_train_rows, hyperparameter_tune=False):
+def default_holdout_frac(num_train_rows, hyperparameter_tune: bool =False):
```
Innixma, line 187:
```suggestion
def default_holdout_frac(num_train_rows, hyperparameter_tune: bool = False):
```

tabular/src/autogluon/tabular/configs/pipeline_presets.py
LennartPurucker: I intended to disable cross-validation, but I somehow wrongly assumed that `USE_BAG_HOLDOUT_AUTO_THRESHOLD` also disabled it. I did not consider that this also disables stacking.

My thought process was that the larger the number of rows, the fewer folds/bags we would need to avoid overfitting. Thus, no bagging would be needed once we hit a critical point. But currently, we do not support stacking without bagging. 

What was the original motivation for `USE_BAG_HOLDOUT_AUTO_THRESHOLD`? Avoid stacked overfitting in these cases or use less expensive validation.

Innixma: The original motivation was to avoid stack overfitting without needing dynamic stacking when the number of rows were large. The only motivation was that the bagged model strength would be better understood by the weighted ensemble when taking the bagged predictions as input, potentially leading to better weights.

LennartPurucker: I see, I changed it such that we keep the old logic.

LennartPurucker:
The behavior change did not result in better generalization, likely due to overfitting on val data being still too dominant.

We will do another PR without a behavior change to use this as a general refactor/code clean-up.

PR #4566, opened 2024-10-22
[timeseries] Add CovariatesRegressor for forecasting models

### LINKED ISSUE #3508: [timeseries] Add support for known covariates in StatsForecast models
It would be helpful to add support for known covariates to the models from StatsForecast.

We could add a flag that allows users to turn the covariates on/off, similar to how we do it for GluonTS models  in https://github.com/autogluon/autogluon/blob/master/timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py#L220C13-L220C37.

### Discussed in https://github.com/autogluon/autogluon/discussions/3501

<div type='discussions-op-text'>

<sup>Originally posted by **darhat** September  1, 2023</sup>
Since known covariates are passed through TimeSeriesPredictor, why are they also not passed into AutoARIMA when forecasting? https://github.com/autogluon/autogluon/blob/master/timeseries/src/autogluon/timeseries/models/local/statsforecast.py#L42. According to statsforecast documentation, this is supported https://nixtla.github.io/statsforecast/src/core/models.html#autoarima.forecast</div>

shchur:
*Issue #, if available:*
Fixes #3508

*Description of changes:*
- Add an option to fit a tabular regression model on the covariates & forecasting on the residuals of the tabular models. The relevant logic is incapsulated into the `CovariateRegressor` object.

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
+++ b/timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
@@ -168,6 +168,8 @@ class AbstractTimeSeriesModel(AbstractModel):
     def _initialize(self, **kwargs) -> None:
         self._init_params_aux()
         self._init_params()
+        self.target_scaler = self._create_target_scaler()
+        self.covariate_regressor = self._create_covariate_regressor()
 
     def _compute_fit_metadata(self, val_data: TimeSeriesDataFrame = None, **kwargs):
         fit_metadata = dict(
@@ -253,11 +255,9 @@ class AbstractTimeSeriesModel(AbstractModel):
         """
         start_time = time.monotonic()
         self.initialize(**kwargs)
-        self.target_scaler = self._create_target_scaler()
         if self.target_scaler is not None:
             train_data = self.target_scaler.fit_transform(train_data)
 
-        self.covariate_regressor = self._create_covariate_regressor()
         if self.covariate_regressor is not None:
             train_data = self.covariate_regressor.fit_transform(
                 train_data,
--- a/timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
+++ b/timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
@@ -15,7 +15,7 @@ from autogluon.core.hpo.executors import HpoExecutor, RayHpoExecutor
 from autogluon.core.models import AbstractModel
 from autogluon.timeseries.dataset import TimeSeriesDataFrame
 from autogluon.timeseries.metrics import TimeSeriesScorer, check_get_evaluation_metric
-from autogluon.timeseries.regressor import CovariatesRegressor
+from autogluon.timeseries.regressor import CovariateRegressor
 from autogluon.timeseries.transforms import LocalTargetScaler, get_target_scaler_from_name
 from autogluon.timeseries.utils.features import CovariateMetadata
 from autogluon.timeseries.utils.forecast import get_forecast_horizon_index_ts_dataframe
@@ -261,7 +261,7 @@ class AbstractTimeSeriesModel(AbstractModel):
         if self.covariates_regressor is not None:
             train_data = self.covariates_regressor.fit_transform(
                 train_data,
-                time_limit=time_limit * 0.5 if time_limit is not None else time_limit,
+                time_limit=0.5 * time_limit if time_limit is not None else None,
             )
 
         train_data = self.preprocess(train_data, is_train=True)
@@ -290,7 +290,7 @@ class AbstractTimeSeriesModel(AbstractModel):
         else:
             return None
 
-    def _create_covariates_regressor(self) -> Optional[CovariatesRegressor]:
+    def _create_covariates_regressor(self) -> Optional[CovariateRegressor]:
         """Create a CovariatesRegressor object based on the value of the `covariates_regressor` hyperparameter."""
         covariates_regressor = self._get_model_params().get("covariates_regressor")
         if covariates_regressor is not None:
@@ -301,9 +301,13 @@ class AbstractTimeSeriesModel(AbstractModel):
                 return None
             else:
                 if isinstance(covariates_regressor, str):
-                    return CovariatesRegressor(covariates_regressor, target=self.target, metadata=self.metadata)
-                elif isinstance(covariates_regressor, CovariatesRegressor):
-                    logger.warning("Using a custom CovariatesRegressor object is experimental functionality!")
+                    return CovariateRegressor(covariates_regressor, target=self.target, metadata=self.metadata)
+                elif isinstance(covariates_regressor, CovariateRegressor):
+                    logger.warning(
+                        "Using a custom CovariatesRegressor object is experimental functionality that may break in the future!"
+                    )
+                    covariates_regressor.target = self.target
+                    covariates_regressor.metadata = self.metadata
                     return covariates_regressor
                 else:
                     raise ValueError(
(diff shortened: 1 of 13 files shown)


timeseries/src/autogluon/timeseries/regressor.py
```diff
@@ -0,0 +1,135 @@
...
+        self.model: Optional[AbstractModel] = None
+        self.metadata = metadata or CovariateMetadata()
+
+    def fit(self, data: TimeSeriesDataFrame, time_limit: Optional[float] = None, **kwargs) -> "CovariatesRegressor":
+        """Fit the tabular regressor on the target column using covariates as features."""
+        tabular_df = self._get_tabular_df(data, static_features=data.static_features)
+        tabular_df = tabular_df.query(f"{self.target}.notnull()")
+
+        median_ts_length = data.num_timesteps_per_item().median()
+        if self.validation_fraction is not None:
```
canerturkmen, line 73: according to the above, self.validation_fraction will never be None.

timeseries/src/autogluon/timeseries/regressor.py
```diff
@@ -0,0 +1,135 @@
...
+
+    def inverse_transform(
+        self,
+        predictions: TimeSeriesDataFrame,
+        known_covariates: TimeSeriesDataFrame,
+        static_features: Optional[pd.DataFrame],
+    ) -> TimeSeriesDataFrame:
+        """Add the tabular regressor predictions to the target column."""
+        y_pred = self._predict(known_covariates, static_features=static_features)
+        return predictions.assign(**{col: predictions[col] + y_pred for col in predictions.columns})
```
canerturkmen, line 115: cool!

timeseries/src/autogluon/timeseries/models/local/abstract_local_model.py
```diff
@@ -113,8 +113,11 @@ def _fit(self, train_data: TimeSeriesDataFrame, time_limit: Optional[int] = None
         local_model_args = {}
         # TODO: Move filtering logic to AbstractTimeSeriesModel
         for key, value in raw_local_model_args.items():
-            if key in self.allowed_hyperparameters:
+            if key in self.allowed_local_model_args:
```
canerturkmen, line 116: was this a bug?

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
```diff
@@ -243,22 +251,35 @@ def fit(
         model: AbstractTimeSeriesModel
             The fitted model object
         """
+        start_time = time.monotonic()
         self.initialize(**kwargs)
         self.target_scaler = self._create_target_scaler()
         if self.target_scaler is not None:
             train_data = self.target_scaler.fit_transform(train_data)
 
+        self.covariates_regressor = self._create_covariates_regressor()
```
canerturkmen, line 260: Let's cover in model tests with each model?

canerturkmen (commented):
Looks great! Made a bunch of comments, but I think we can move to adding tests. Thanks!

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
```diff
@@ -243,22 +251,35 @@ def fit(
...
+                train_data,
+                time_limit=time_limit * 0.5 if time_limit is not None else time_limit,
+            )
+
         train_data = self.preprocess(train_data, is_train=True)
         if self._get_tags()["can_use_val_data"] and val_data is not None:
             if self.target_scaler is not None:
                 val_data = self.target_scaler.transform(val_data)
+            if self.covariates_regressor is not None:
+                val_data = self.covariates_regressor.transform(val_data)
```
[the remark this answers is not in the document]
shchur: You are right, but I'm not quite sure how to set the time_limit for `fit_transform` in this case. In practice, I don't think it will ever be relevant since only GluonTS models even use `val_data` during `fit()`, and for those model we probably don't want to use the "regressor" anyway.

The real "refit on val_data" will happen during `predict` for all models when they call `score_and_cache_oof`.

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
```diff
@@ -324,18 +367,42 @@ def predict(
...
+                data = self.covariates_regressor.transform(data)
 
         data = self.preprocess(data, is_train=False)
         known_covariates = self.preprocess_known_covariates(known_covariates)
+
+        # FIXME: Avoid copying covariates_regressor across processes if self._predict uses joblib.
+        # FIXME: The clean solution is to convert all methods executed in parallel to @classmethod
+        covariates_regressor = self.covariates_regressor
+        self.covariates_regressor = None
         predictions = self._predict(data=data, known_covariates=known_covariates, **kwargs)
```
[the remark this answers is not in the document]
shchur: Yes, the `_predict` method will expect the covariates to be present in accordance with the `CovariateMetadata` (and the `_predict` method ideally shouldn't be aware that a "regressor" object even exists). It will mostly be user's and our responsibility not to do silly things such as using both the "regressor" in combination with a model that natively handles the covariates.

timeseries/src/autogluon/timeseries/models/local/abstract_local_model.py
shchur: Yes, but this wasn't caught because we didn't have tests that added `target_scaler` to all models. Will fix that in this PR.

timeseries/src/autogluon/timeseries/regressor.py
```diff
@@ -0,0 +1,135 @@
...
+import numpy as np
+import pandas as pd
+
+from autogluon.core.models import AbstractModel
+from autogluon.tabular.trainer.model_presets.presets import MODEL_TYPES as TABULAR_MODEL_TYPES
+from autogluon.timeseries.dataset.ts_dataframe import ITEMID, TIMESTAMP, TimeSeriesDataFrame
+from autogluon.timeseries.utils.features import CovariateMetadata
+
+
+class CovariatesRegressor:
```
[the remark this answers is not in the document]
canerturkmen: got it `CovariateRegressor` sounds good to me.

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -293,7 +293,7 @@ def _check_and_prepare_data_frame(
         df = self._to_data_frame(data, name=name)
         if not pd.api.types.is_numeric_dtype(df[self.target]):
             raise ValueError(f"Target column {name}['{self.target}'] has a non-numeric dtype {df[self.target].dtype}")
-        df[self.target] = df[self.target].astype("float64")
+        df = df.assign(**{self.target: df[self.target].astype("float64")})
```
shchur, line 296: This is necessary to silence a pandas CopyOnWrite warning

timeseries/src/autogluon/timeseries/transforms/scaler.py
```diff
@@ -85,7 +85,7 @@ class LocalMinMaxScaler(LocalTargetScaler):
     def _compute_loc_scale(self, target_series: pd.Series) -> Tuple[pd.Series, pd.Series]:
         stats = target_series.abs().groupby(level=ITEMID, sort=False).agg(["min", "max"])
         scale = (stats["max"] - stats["min"]).clip(lower=self.min_scale)
-        loc = stats["min"] / scale
+        loc = stats["min"]
```
shchur, line 88: There was a bug in the min-max scaler, fixed now.

timeseries/src/autogluon/timeseries/regressor.py
```diff
@@ -0,0 +1,135 @@
+from typing import Any, Dict, Optional
```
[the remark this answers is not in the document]
shchur: I tried my best, let me know if you have ideas for other tests

canerturkmen (approved):
LGTM! Thanks!

PR #4569, opened 2024-10-22
[WIP] Silent NLTK Download

FANGAreNotGnu:
Fix https://github.com/autogluon/autogluon/issues/4556

### DIFF
--- a/multimodal/src/autogluon/multimodal/data/utils.py
+++ b/multimodal/src/autogluon/multimodal/data/utils.py
@@ -616,7 +616,7 @@ def image_mean_std(norm_type: str):
 def silent_nltk_download(package):
     import sys
 
-    is_jupyter = 'ipykernel' in sys.modules
+    is_jupyter = "ipykernel" in sys.modules
     if is_jupyter:
         from IPython.utils.io import capture_output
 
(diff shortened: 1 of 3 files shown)


multimodal/src/autogluon/multimodal/data/utils.py
```diff
@@ -613,3 +613,9 @@ def image_mean_std(norm_type: str):
         return CLIP_IMAGE_MEAN, CLIP_IMAGE_STD
     else:
         raise ValueError(f"unknown image normalization: {norm_type}")
+
+
+def silent_nltk_download(package):
+    with open(os.devnull, "w") as devnull:
+        with redirect_stdout(devnull):
+            return nltk.download(package, quiet=True)
```
tonyhoo, line 621: Curious why the `quiet=True` flag alone is not sufficient here anymore?

FANGAreNotGnu: Didn't find the root cause. Source code shows that it should be skipped if quiet is True: https://github.com/nltk/nltk/blob/9a5622f8a5b228df9499cd03181d9f8491e39f17/nltk/downloader.py#L818-L825

tonyhoo (approved):

multimodal/src/autogluon/multimodal/data/utils.py
tonyhoo: Looks like the logs are still there post this change http://autogluon-staging.s3-website-us-west-2.amazonaws.com/PR-4569/0c9c435/tutorials/multimodal/text_prediction/beginner_text.html

Issue #4567, opened 2024-10-22
[BUG] Incorrect number of CPUs


rsj123:
**Bug Report Checklist**

- [x] I provided code that demonstrates a minimal reproducible example.  
- [ ] I confirmed bug exists on the latest mainline of AutoGluon via source install.  
- [x] I confirmed bug exists on the latest stable version of AutoGluon.  

**Describe the bug**
In my execution, AutoGluon incorrectly detected the actual number of available CPUs (possibly related to Docker). My container is limited to 32 vCPUs, but AutoGluon reads the number of CPUs from the host machine, resulting in an incorrect CPU allocation.
Here are some of the cpython community discussions: https://github.com/python/cpython/issues/80235#issuecomment-1093815926
**Expected behavior**
AutoGluon detected actual number of available CPUs correctly
**To Reproduce**

**Screenshots / Logs**

(autoscaler +14m43s) Error: No available node types can fulfill resource request {'CPU': 384.0}. Add suitable node types to this cluster to resolve this issue.

**Installed Versions**

<details>

INSTALLED VERSIONS
------------------
date                   : 2024-10-23
time                   : 00:41:41.426556
python                 : 3.10.8.final.0
OS                     : Linux
OS-release             : 5.4.0-153-generic
Version                : #170-Ubuntu SMP Fri Jun 16 13:43:31 UTC 2023
machine                : x86_64
processor              : x86_64
num_cores              : 384
cpu_ram_mb             : 1547877.46875
cuda version           : None
num_gpus               : 0
gpu_ram_mb             : []
avail_disk_size_mb     : 39242

accelerate             : 0.21.0
autogluon              : 1.1.1
autogluon.common       : 1.1.1
autogluon.core         : 1.1.1
autogluon.features     : 1.1.1
autogluon.multimodal   : 1.1.1
autogluon.tabular      : 1.1.1
autogluon.timeseries   : 1.1.1
boto3                  : 1.35.45
catboost               : 1.2.7
defusedxml             : 0.7.1
evaluate               : 0.4.3
fastai                 : 2.7.18
gluonts                : 0.15.1
hyperopt               : 0.2.7
imodels                : None
jinja2                 : 3.1.2
joblib                 : 1.4.2
jsonschema             : 4.21.1
lightgbm               : 4.3.0
lightning              : 2.3.3
matplotlib             : 3.6.3
mlforecast             : 0.10.0
networkx               : 3.4.2
nlpaug                 : 1.1.11
[...]

PR #4571, opened 2024-10-23
Update cpu count to current process can use

rsj123:
*Issue #, if available:*
#4567 
*Description of changes:*
Update cpu count to current process can use

### DIFF
--- a/common/src/autogluon/common/utils/resource_utils.py
+++ b/common/src/autogluon/common/utils/resource_utils.py
@@ -17,7 +17,7 @@ class ResourceManager:
 
     @staticmethod
     def get_cpu_count():
-        return multiprocessing.cpu_count()
+        return len(os.sched_getaffinity(0))
 
     @staticmethod
     @disable_if_lite_mode(ret=1)


tonyhoo:
Thank you for reporting the issue and your contribution! The proposed change will work on Linux-like systems but may not perform as expected on Mac and Windows. Could you take a look?

rsj123:
psutil is a good method, but it will not be imported in lite_mode

tonyhoo:
It seems the solution is still not accurately retrieving the provisioned CPU count from the Docker environment on both MacOS and Ubuntu. 

For clarity, the cgroup info indicates that 2 CPUs are allocated, as shown with the command `docker run --cpus=2 ...`:

```
root@2decc059608a:/# cat /sys/fs/cgroup/cpu/cpu.cfs_quota_us
200000
```

However, the code yields:

```
In [1]: import os

In [2]: len(os.sched_getaffinity(0))
Out[2]: 96
```

This suggests that the intended functionality is not working as expected within the Docker container. Try with this [image](https://hub.docker.com/r/autogluon/autogluon)

Let’s investigate further!

rsj123:
> It seems the solution is still not accurately retrieving the provisioned CPU count from the Docker environment on both MacOS and Ubuntu.
> 
> For clarity, the cgroup info indicates that 2 CPUs are allocated, as shown with the command `docker run --cpus=2 ...`:
> 
> ```
> root@2decc059608a:/# cat /sys/fs/cgroup/cpu/cpu.cfs_quota_us
> 200000
> ```
> 
> However, the code yields:
> 
> ```
> In [1]: import os
> 
> In [2]: len(os.sched_getaffinity(0))
> Out[2]: 96
> ```
> 
> This suggests that the intended functionality is not working as expected within the Docker container. Try with this [image](https://hub.docker.com/r/autogluon/autogluon)
> 
> Let’s investigate further!

I think --cpus won't affect AutoGluon's normal operation because it's based on time slice restrictions, allowing Ray to allocate normally. 
However, if --cpuset-cpus is used, which binds to specific cores, Ray cannot allocate CPUs correctly in that case. We might only need to consider this latter situation.

Innixma:
Moving this to 1.3 release timeline as it has a potentially wide-ranging impact that will need testing

tonyhoo:
Moving to 1.4 release as 1.3 is in progress

Innixma:
@tonyhoo Has this been resolved already in a separate PR?

Innixma:
Thanks for the contribution! We implemented a solution using joblib for AutoGluon 1.4 release that should address this issue. Please open a new issue if the problem persists in AG 1.4. Cheers!

PR #4576, opened 2024-10-24
Fix log loss calculation for `MultiModalPredictor.evaluate()`.

mtdiedrich:
*Description of changes:*

Log loss is calculated using probabilities. The existing implementation instead uses labels. This change corrects the existing implementation to use probabilities.

### DIFF
--- a/multimodal/src/autogluon/multimodal/utils/metric.py
+++ b/multimodal/src/autogluon/multimodal/utils/metric.py
@@ -18,6 +18,7 @@ from ..constants import (
     F1,
     FEW_SHOT_CLASSIFICATION,
     IOU,
+    LOG_LOSS,
     MAP,
     MATCHING_METRICS,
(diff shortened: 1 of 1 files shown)


FANGAreNotGnu:
Seems that log_loss is defined as needs_proba in https://github.com/mtdiedrich/autogluon/blob/7a60f1d05be94b57b434f879a06bd46f1eb071e6/core/src/autogluon/core/metrics/__init__.py#L644. Shall we do 
```
if metric.needs_proba:
        return metric._sign * metric(metric_data[Y_TRUE], metric_data[Y_PRED_PROB])
```
 instead? @Innixma

FANGAreNotGnu:
Thank you for identifying this bug! I've submitted a fix that includes a more general refactor of the `compute_score` function (https://github.com/autogluon/autogluon/pull/4629). Please feel free to review and share any questions or suggestions for further improvements.

PR #4583, opened 2024-10-25
Fix edgecase label crash

Innixma:
*Issue #, if available:*

*Description of changes:*

- Fix rare crash that occurs when the label column is a categorical with categories defined that have 0 rows with that value. AutoGluon treats it as a missing label and crashes. This PR fixes this to ignore such categories.
- Added type hints


shchur (approved):
I'm not very familiar with the label processing logic so I might miss some edge case, but this looks reasonable.

suzhoum (approved):
LGTM!

PR #4594, opened 2024-10-29
[timeseries] Add weighted cumulative error forecasting metric

shchur:
*Issue #, if available:*

*Description of changes:*
- Add experimental `WCD` (Weighted Cumulative Discrepancy) forecasting metric

### DIFF
--- a/timeseries/src/autogluon/timeseries/metrics/point.py
+++ b/timeseries/src/autogluon/timeseries/metrics/point.py
@@ -362,33 +362,31 @@ class RMSLE(TimeSeriesScorer):
         )
 
 
-class CE(TimeSeriesScorer):
-    r"""Cumulative error.
+class WCD(TimeSeriesScorer):
+    r"""Weighted cumulative discrepancy.
 
-    This error measures the discrepancy between the cumulative sum of the forecast and the cumulative sum of the actual
-    values.
+    Measures the discrepancy between the cumulative sum of the forecast and the cumulative sum of the actual values.
 
     .. math::
 
-        \operatorname{CE} = \frac{1}{N} \frac{1}{H} \sum_{i=1}^{N} \sum_{t=T+1}^{T+H} \alpha1 \cdot \max(0, -d_{i, t}) + \alpha2 \cdot \max(0, d_{i, t})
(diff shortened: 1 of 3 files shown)


timeseries/src/autogluon/timeseries/metrics/point.py
```diff
@@ -359,3 +360,54 @@ def __call__(
...
+
+    def compute_metric(
+        self, data_future: TimeSeriesDataFrame, predictions: TimeSeriesDataFrame, target: str = "target", **kwargs
+    ) -> float:
+        y_true, y_pred = self._get_point_forecast_score_inputs(data_future, predictions, target=target)
+        cumsum_true = self._fast_cumsum(y_true.to_numpy())
+        cumsum_pred = self._fast_cumsum(y_pred.to_numpy())
+        diffs = cumsum_pred - cumsum_true
+        error = diffs * np.where(diffs < 0, -self.alpha1, self.alpha2)
+        return self._safemean(error)
```
shchur, line 413: Previously, we computed the metric as `self._safemean(error) * self.num_items` (equivalent to removing the 1/H scaling factor from the formula above). Which of the two versions should we keep?

timeseries/src/autogluon/timeseries/metrics/point.py
```diff
@@ -359,3 +360,54 @@ def __call__(
...
+
+    Parameters
+    ----------
+    alpha1 : float
+        Penalty assigned to underpredictions (when cumulative forecast is below the cumulative actual value).
+    alpha2 : float
+        Penalty assigned to overpredictions (when cumulative forecast is above the cumulative actual value).
+    """
+
+    def __init__(self, alpha1: float = 0.75, alpha2: float = 0.25):
```
andborth, line 389: Let's have a single alpha which is equivalent to alpha1.  alpha2 would be replaced with 1 - alpha.

shchur: Should we then maybe multiply the overall score by 2.0 so that in case alpha=0.5 we recover `MAE(np.cumsum(y_true, np.cumsum(y_pred))`?

andborth: Yes, I like the idea.  That would make the score more intuitive.

canerturkmen (approved):
LGTM!

PR #4565, opened 2024-10-21
[timeseries] disable non-CUDA backends for chronos

canerturkmen:
*Issue #, if available:*

#4544 

*Description of changes:*

The MPS backend leads to bugs for Chronos in PyTorch 2.2 and 2.3. This PR disables all non-CUDA backends for Chronos.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #4608, opened 2024-11-04
[timeseries] Add fine-tuning support for Chronos models 

abdulfatir:
*Issue #, if available:* N/A

*Description of changes:* This PR adds support for fine-tuning Chronos models. 

- [x] Unify fine-tuning support for Chronos and Bolt models.
- [x] Fix tests to run with fine-tuned models
- [x] Add new tests
- [x] Add Smoke tests
- [x] Update docstrings and model zoo docs
- [x]  #4625 has been merged

Should be merged after #4625.


timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -254,7 +286,93 @@ def _fit(
         **kwargs,
     ) -> None:
         self._check_fit_params()
-        self.time_limit = time_limit
+
+        fine_tune_time_limit = time_limit // 2 if self.fine_tune else 0
```
canerturkmen, line 290: We don't need to do this by dividing in half explicitly. As long as the entire workflow respects the time limit, we can allow fine tuning to take as much time as it needs.

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -254,7 +286,93 @@ def _fit(
...
+                target_df=train_data,
+                freq=self.freq,
+                target_column=self.target,
+                context_length=context_length,
+                prediction_length=self.prediction_length,
+                mode="training",
+            )
+            callbacks = [TimeLimitCallback(time_limit=fine_tune_time_limit), EvaluateAndSaveFinalStepCallback()]
+
+            tokenizer_val_dataset = (
```
canerturkmen, line 335: why are we using the validation data? in order to do early stopping or only for selecting checkpoints? If we're not doing early-stopping I think we can skip this entirely to prevent val data from leaking into the training here (the same data will be used to fit the ensemble).

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -290,15 +415,12 @@ def _predict(
...
-            data.num_timesteps_per_item().max(),
-            self.maximum_context_length,
-        )
+        context_length = self._get_context_length(data)
 
         with warning_filter(all_warnings=True):
             import torch
 
-            if self.model_pipeline is None:
+            if self.model_pipeline is None or self.optimization_strategy is not None:
```
canerturkmen, line 423: Why this? Can we make the reason more explicit, also in code? Btw if optimization strategy stuff is too much overhead, we can remove it. We won't be needing CPU acceleration as much anymore.

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -254,7 +286,93 @@ def _fit(
...
+            best_train_eval_loss = log_history_df["eval_loss"].min()
+
+            if best_train_eval_loss <= zero_shot_eval_loss:
+                fine_tuned_ckpt_path = Path(self.path) / self.fine_tuned_ckpt_name
+                logger.debug(
+                    f"Validation loss improved after fine-tuning. Saving fine-tuned model to {fine_tuned_ckpt_path}"
+                )
+                self.model_pipeline.model.model.save_pretrained(Path(self.path) / self.fine_tuned_ckpt_name)
+
+            if not self.keep_transformers_logs:
```
canerturkmen, line 373: I might have missed it, but I think the `model_path` needs to be overwritten somewhere here so that the Predictor will pick up the fine tuned checkpoint during inference?

canerturkmen (commented):
Thanks!!! A couple of more pointers to add this to, just to keep track.

- smoke tests: https://github.com/autogluon/autogluon/blob/master/timeseries/tests/smoketests/test_all_models.py
- model tests https://github.com/autogluon/autogluon/blob/master/timeseries/tests/unittests/models/test_chronos.py#L27
- model zoo docs and docstrings

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -1,13 +1,21 @@
 import logging
 import os
+import shutil
+from pathlib import Path
 from typing import Any, Dict, Literal, Optional, Union
 
 import numpy as np
 import pandas as pd
+from transformers import Trainer, TrainingArguments
```
shchur, line 9: Ideally, we should hide the imports for all "large" libraries (like torch, transformers) inside class methods to avoid slow import times for `from autogluon.timeseries import ...`. We have to pay this overhead each time multiprocessing is used, so this adds up.

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -1,13 +1,21 @@
...
 
 from autogluon.common.loaders import load_pkl
 from autogluon.timeseries.dataset.ts_dataframe import TimeSeriesDataFrame
 from autogluon.timeseries.models.abstract import AbstractTimeSeriesModel
+from autogluon.timeseries.models.chronos.utils import (
+    ChronosFineTuningDataset,
+    EvaluateAndSaveFinalStepCallback,
+    TimeLimitCallback,
+)
```
shchur, line 18: Same comment regarding the imports: `.utils` imports `torch` so we should hide it inside methods

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -124,6 +137,14 @@ def __init__(
         model_path_input = hyperparameters.get("model_path", self.default_model_path)
         self.model_path = MODEL_ALIASES.get(model_path_input, model_path_input)
 
+        self.fine_tune = hyperparameters.get("fine_tune", False)
+        self.fine_tune_batch_size = hyperparameters.get("fine_tune_batch_size", self.default_batch_size)
+        self.fine_tune_lr = hyperparameters.get("fine_tune_lr", self.default_fine_tune_lr)
+        self.fine_tune_steps = hyperparameters.get("fine_tune_steps", self.default_fine_tune_steps)
+        self.logging_steps = hyperparameters.get("logging_steps", self.default_logging_steps)
+        self.eval_and_save_steps = hyperparameters.get("eval_and_save_steps", self.default_eval_and_save_steps)
+        self.keep_transformers_logs = hyperparameters.get("keep_transformers_logs", False)
```
shchur, line 146: Nit: These hyperparameters are not needed during `__init__`, so I would recommend just accessing them inside `_fit` using `self._get_model_params()`. The default values for the hyperparameters are also only used once, so we probably don't need to store them as class attributes.

Also see my comment below regarding the `TrainerArguments`.

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -218,12 +245,12 @@ def get_minimum_resources(self, is_gpu_available: bool = False) -> Dict[str, Uni
...
         return minimum_resources
 
-    def load_model_pipeline(self, context_length: Optional[int] = None):
+    def load_model_pipeline(self, context_length: Optional[int] = None, is_training: bool = False):
         from .pipeline import OptimizedChronosPipeline
 
         gpu_available = self._is_gpu_available()
 
-        if not gpu_available and self.min_num_gpus > 0:
+        if not gpu_available and (self.min_num_gpus > 0 or is_training):
```
shchur, line 253: Do we require GPU for fine-tuning because HF will raise an error if no GPU provided, or will the fine-tuning just be extremely slow?

If it's the latter, is fine-tuning on CPU also slow for the patch-based model?

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -254,7 +286,93 @@ def _fit(
...
+                max_steps=self.fine_tune_steps,
+                gradient_accumulation_steps=1,
+                dataloader_num_workers=self.data_loader_num_workers,
+                tf32=self._has_tf32(),
+                save_only_model=True,
+                load_best_model_at_end=True,
+                metric_for_best_model="eval_loss",
+                prediction_loss_only=True,
+                save_total_limit=1,
+            )
```
shchur, line 323: Maybe it makes sense to group all the fine-tuning related parameters into a single dictionary with default values? 

Something like
```python
trainer_kwargs = dict(
    output_dir=str(output_dir),
    per_device_train_batch_size=self.default_batch_size,
    learning_rate=1e-3,
    lr_scheduler_type="linear",
    warmup_ratio=0.0,
    optim="adamw_torch_fused",
    logging_dir=str(output_dir),
    logging_strategy="steps",
    logging_steps=self.logging_steps,
    save_strategy="steps",
    save_steps=50,
    evaluation_strategy="steps",
    eval_steps=50,
    report_to="none",
    max_steps=1000,
    gradient_accumulation_steps=1,
    dataloader_num_workers=self.data_loader_num_workers,
    tf32=self._has_tf32(),
    save_only_model=True,
    load_best_model_at_end=True,
    metric_for_best_model="eval_loss",
    prediction_loss_only=True,
    save_total_limit=1,
)

user_trainer_kwargs = self._get_model_params().get("trainer_kwargs", {})
trainer_kwargs.update(user_trainer_kwargs)

training_args = TrainingArguments(**trainer_kwargs)
```
This will be similar to how we expose the `trainer_kwargs` for [GluonTS models](https://auto.gluon.ai/dev/tutorials/timeseries/forecasting-model-zoo.html#autogluon.timeseries.models.DeepARModel).

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -254,7 +286,93 @@ def _fit(
         **kwargs,
     ) -> None:
         self._check_fit_params()
-        self.time_limit = time_limit
+
+        fine_tune_time_limit = time_limit // 2 if self.fine_tune else 0
+        self.time_limit = time_limit - fine_tune_time_limit  # inference time budget
```
shchur, line 291: Note that `time_limit` can also be `None`, which will raise an exception here.

timeseries/src/autogluon/timeseries/models/chronos/model.py
shchur: We need some guardrails here since otherwise finetuning may consume 100% of the time limit, no time will be left for inference, and the Chronos model will just raise an exception when predicting/computing the val score.

I guess that allocating 50% of the time limit for fine-tuning is a reasonable value, but maybe there exist better ways to handle the overall time limit.

timeseries/src/autogluon/timeseries/models/chronos/utils.py
```diff
@@ -1,11 +1,177 @@
...
+        target_column: str = "target",
+        context_length: int = 512,
+        prediction_length: int = 64,
+        mode: Literal["training", "validation"] = "training",
+    ) -> None:
+        super().__init__()
+
+        assert mode in ("training", "validation")
+
+        self.gluonts_dataset = SimpleGluonTSDataset(target_df=target_df, freq=freq, target_column=target_column)
```
shchur, line 113: Let's just put a dummy `freq` here since it's anyway not used by the model, otherwise GluonTS might freak out about some non-standard frequencies.

timeseries/src/autogluon/timeseries/models/chronos/utils.py
```diff
@@ -1,11 +1,177 @@
...
+        self.tokenizer = tokenizer
+        self.context_length = context_length
+        self.prediction_length = prediction_length
+        self.mode = mode
+
+    def preprocess_entry(self, entry: dict) -> dict:
+        entry = {f: entry[f] for f in [FieldName.START, FieldName.TARGET]}
+        entry[FieldName.TARGET] = np.asarray(entry[FieldName.TARGET], dtype=np.float32)
+
+        return entry
```
shchur, line 123: Nit: Is this necessary? I believe that `SimpleGluonTSDataset` already returns entries in this format (except that `item_id` is also included)

timeseries/src/autogluon/timeseries/models/chronos/utils.py
```diff
@@ -1,11 +1,177 @@
...
+            "labels": labels.squeeze(0),
+        }
+
+    def __iter__(self) -> Iterator:
+        preprocessed_dataset = Map(self.preprocess_entry, self.gluonts_dataset)
+
+        if self.mode == "training":
+            iterable = self._create_training_data(preprocessed_dataset)
+        elif self.mode == "validation":
+            iterable = self._create_validation_data(preprocessed_dataset)
```
shchur, line 171: Is there a way to limit the # of validation batches (similar to https://lightning.ai/docs/pytorch/stable/common/trainer.html#limit-val-batches)? Otherwise we will just spend most of the time computing the validation loss for datasets with a large number of items.

timeseries/src/autogluon/timeseries/models/chronos/utils.py
```diff
@@ -55,6 +221,38 @@ def __iter__(self):
...
+        Parameters
+        ----------
+        time_limit: int
+            maximum time allowed for training in seconds.
+        """
+        self.time_limit = time_limit
+        self.start_time = None
+
+    def on_train_begin(self, args, state, control, **kwargs):
+        self.start_time = time.time()
```
shchur, line 248: Nit: `time.monotonic()` is a better alternative that avoids some edge cases for `time.time()` (e.g., if system clock changes during training)

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -254,7 +286,93 @@ def _fit(
...
+                else None
+            )
+
+            trainer = Trainer(
+                model=self.model_pipeline.model.model,
+                args=training_args,
+                train_dataset=tokenizer_train_dataset,
+                eval_dataset=tokenizer_val_dataset,
+                callbacks=callbacks,
+            )
```
shchur, line 355: Can we control the verbosity of the HF trainer based on the `verbosity` level provided to `_fit`?

Here is what the output currently looks like in a jupyter notebook:

image

We should try to completely suppress the HF logs + warnings for `verbosity<=2`.

If possible, it would be nice to show the progress bar + DF for `verbosity>=3` (but not critical).

shchur (commented):
Just tried this and it works like a charm ✨, thanks a lot for the PR!

timeseries/src/autogluon/timeseries/models/chronos/utils.py
shchur: Alternatively, we can just randomly subsample the items when creating the `tokenizer_val_dataset` above.

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -254,7 +286,93 @@ def _fit(
...
+                context_length=context_length,
+                prediction_length=self.prediction_length,
+                mode="training",
+            )
+            callbacks = [TimeLimitCallback(time_limit=fine_tune_time_limit), EvaluateAndSaveFinalStepCallback()]
+
+            tokenizer_val_dataset = (
+                ChronosFineTuningDataset(
+                    tokenizer=self.model_pipeline.tokenizer,
+                    target_df=val_data,
```
shchur, line 338: If we use the `val_data` for validation, we should also add the respective tag [here](https://github.com/autogluon/autogluon/blob/master/timeseries/src/autogluon/timeseries/models/chronos/model.py#L339-L340) to ensure that preprocessing is applied to validation data (same as in https://github.com/autogluon/autogluon/blob/master/timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py#L695-L696)

core/src/autogluon/core/_setup_utils.py
```diff
@@ -32,7 +32,7 @@
     "pytorch_lightning": ">=2.2,<2.4",  # Major version cap, capping `lightning` does not cap `pytorch_lightning`!
     "async_timeout": ">=4.0,<5",  # Major version cap
     "transformers[sentencepiece]": ">=4.38.0,<4.41.0",
-    "accelerate": ">=0.21.0,<0.22.0",
+    "accelerate": ">=0.32.0,<1",
```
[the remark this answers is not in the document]
abdulfatir: This is temporary, I will remove this change.

timeseries/src/autogluon/timeseries/models/chronos/utils.py
```diff
@@ -1,11 +1,177 @@
 import time
-from typing import Callable, Optional
+from typing import Callable, Iterator, Literal, Optional
 
 import numpy as np
 import torch
+from gluonts.dataset.field_names import FieldName
+from gluonts.itertools import Cyclic, Map
```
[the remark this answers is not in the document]
abdulfatir: Removed `Cycle` and `Map`. I left `FieldName` because that's what's also done for GluonTS models. I don't think this should be an overhead.

timeseries/src/autogluon/timeseries/models/chronos/utils.py
```diff
@@ -55,6 +221,38 @@ def __iter__(self):
...
+        self.time_limit = time_limit
+        self.start_time = None
+
+    def on_train_begin(self, args, state, control, **kwargs):
+        self.start_time = time.time()
+
+    def on_step_end(self, args, state, control, **kwargs):
+        elapsed_time = time.time() - self.start_time
+        if elapsed_time >= self.time_limit:
+            control.should_training_stop = True
```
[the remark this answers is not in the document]
abdulfatir: Logged as a warning.

timeseries/src/autogluon/timeseries/models/chronos/utils.py
```diff
@@ -1,11 +1,177 @@
...
+
+    def _create_validation_data(self, data):
+        data = self._create_instance_splitter("validation").apply(data, is_train=False)
+        return data
+
+    def to_hf_format(self, entry: dict) -> dict:
+        past_target = torch.tensor(entry[f"past_{FieldName.TARGET}"]).unsqueeze(0)
+        input_ids, attention_mask, scale = self.tokenizer.context_input_transform(past_target)
+        future_target = torch.tensor(entry[f"future_{FieldName.TARGET}"]).unsqueeze(0)
+        labels, labels_mask = self.tokenizer.label_input_transform(future_target, scale)
```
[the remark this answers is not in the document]
abdulfatir: Yes, will look into this.

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -108,6 +116,11 @@ class ChronosModel(AbstractTimeSeriesModel):
     default_num_samples: int = 20
     default_model_path = "autogluon/chronos-t5-small"
     maximum_context_length = 512
+    fine_tuned_ckpt_name: str = "fine-tuned-ckpt"
+    default_fine_tune_lr: float = 0.001
```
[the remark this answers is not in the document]
abdulfatir: This will be changed later based on experiments. For now I have found that 5K steps with 1e-4 initial rate is good default value. This may also change though.

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -108,6 +116,11 @@ class ChronosModel(AbstractTimeSeriesModel):
...
+    default_fine_tune_steps: int = 1000
+    default_logging_steps: int = 50
```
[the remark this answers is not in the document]
abdulfatir: Filtered logs from transformers to respect the verbosity level.

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -175,10 +196,16 @@ def load(cls, path: str, reset_paths: bool = True, verbose: bool = True) -> "Chr
...
+        fine_tune_ckpt_path = Path(model.path) / cls.fine_tuned_ckpt_name
+        if fine_tune_ckpt_path.exists():
+            logger.debug(f"Fine-tuned checkpoint exists, setting model_path to {fine_tune_ckpt_path}")
+            model.model_path = fine_tune_ckpt_path
+
         return model
 
     def _is_gpu_available(self) -> bool:
-        import torch.cuda
+        import torch
```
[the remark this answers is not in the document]
abdulfatir: Reverted.

timeseries/src/autogluon/timeseries/models/chronos/model.py
abdulfatir: I will keep it like this for now and we can discuss again.

timeseries/src/autogluon/timeseries/models/chronos/model.py
abdulfatir: We are selecting the best checkpoint based on validation. I discussed this with @shchur and we decided that we will have an option to perform such validation but will turn in off by default.

timeseries/src/autogluon/timeseries/models/chronos/model.py
abdulfatir: `load` checks if there's a fine-tuned checkpoint and loads that one. I just realized that reversion to pretrained ckpt will not work if the object is in memory. Fixed that.

timeseries/src/autogluon/timeseries/models/chronos/model.py
abdulfatir: Done.

timeseries/src/autogluon/timeseries/models/chronos/model.py
abdulfatir: It will be slow. Should we allow CPU fine-tuning? (I think for Bolt CPU fine-tuning should be possible at a reasonable speed)

timeseries/src/autogluon/timeseries/models/chronos/model.py
abdulfatir: Good idea. Doing that.

timeseries/src/autogluon/timeseries/models/chronos/model.py
abdulfatir: Add a check.

timeseries/src/autogluon/timeseries/models/chronos/utils.py
abdulfatir: Done.

timeseries/src/autogluon/timeseries/models/chronos/utils.py
abdulfatir: Removed this function as it was not needed.

timeseries/src/autogluon/timeseries/models/chronos/model.py
abdulfatir: This is what I went with in the end:

```
# verbosity < 3: all logs and warnings from transformers will be supressed
# verbosity >= 3: progress bar and loss logs will be logged
# verbosity 4: everything will be logged
```

timeseries/src/autogluon/timeseries/models/chronos/utils.py
abdulfatir: Fixed.

timeseries/src/autogluon/timeseries/models/chronos/model.py
shchur: I think CPU fine-tuning is reasonable for Bolt, so I would allow this.

timeseries/src/autogluon/timeseries/models/chronos/model.py
abdulfatir: I took a stab at this. Lmk what you think.

timeseries/src/autogluon/timeseries/models/chronos/model.py
abdulfatir: Addressed.

timeseries/src/autogluon/timeseries/models/chronos/model.py
abdulfatir: I kept the option to perform validation but turned it off by default.

timeseries/src/autogluon/timeseries/models/chronos/model.py
abdulfatir: Fixed.

timeseries/src/autogluon/timeseries/models/chronos/utils.py
abdulfatir: There's no such option in `transformers` it seems, so I did it manually.

timeseries/src/autogluon/timeseries/models/chronos/pipeline/chronos_bolt.py
```diff
@@ -375,6 +381,9 @@ class ChronosBoltPipeline(BaseChronosPipeline):
     def __init__(self, model: ChronosBoltModelForForecasting):
         self.model = model
 
+        # for easy access to the inner HF-style model
+        self.inner_model = model
```
canerturkmen, line 385: I like this, however in this case we should maybe get rid of `self.model` to avoid confusion? Also, I would optionally add somehow add this, maybe as an instance variable to `BaseChronosPipeline.__init__` to hint subclasses will implement this.

timeseries/tests/unittests/models/chronos/test_model.py
```diff
@@ -52,7 +61,14 @@ def chronos_bolt_model(*args, hyperparameters=None, **kwargs):
...
 
-TESTABLE_MODELS = [ChronosModel, chronos_bolt_model]
+def chronos_bolt_model_with_finetuning(*args, hyperparameters=None, **kwargs):
+    hyperparameters = copy.deepcopy(hyperparameters or {})
+    hyperparameters |= {"model_path": CHRONOS_BOLT_TEST_MODEL_PATH, "fine_tune": True, "fine_tune_steps": 10}
+    kwargs["hyperparameters"] = hyperparameters
+    return ChronosModel(*args, **kwargs)
+
+
+TESTABLE_MODELS = [ChronosModel, chronos_bolt_model, chronos_bolt_model_with_finetuning]
```
canerturkmen, line 71: cool. note also that this doesn't cover chronos classic fine tuning (I think we can skip for now) and also we should still add to smoke tests.

timeseries/src/autogluon/timeseries/models/chronos/pipeline/utils.py
```diff
@@ -1,14 +1,210 @@
...
 import numpy as np
 import torch
+from gluonts.dataset.field_names import FieldName
+from gluonts.transform import ExpectedNumInstanceSampler, InstanceSplitter, ValidationSplitSampler
+from torch.utils.data import IterableDataset
+from transformers import TrainerCallback
 
 from autogluon.core.utils.exceptions import TimeLimitExceeded
 from autogluon.timeseries.dataset.ts_dataframe import TimeSeriesDataFrame
+from autogluon.timeseries.models.gluonts.abstract_gluonts import SimpleGluonTSDataset
```
canerturkmen, line 17: a todo to drop in `SimpleGluonTSDataset`, maybe it makes sense to move this to the `dataset` namespace now that multiple model classes are depending on it. cc: @shchur

timeseries/src/autogluon/timeseries/models/chronos/pipeline/utils.py
```diff
@@ -1,14 +1,210 @@
...
+
+if TYPE_CHECKING:
```
canerturkmen, line 19: very cool. although if I was to nit, the fact that this circular dependency is here hints that ChronosFineTuningDataset should not depend on the Tokenizer, but rather the pipeline object should know about how it will do tokenization. we can drop a TODO for later.

timeseries/src/autogluon/timeseries/models/chronos/pipeline/utils.py
```diff
@@ -1,14 +1,210 @@
...
+            is_pad_field=FieldName.IS_PAD,
+            start_field=FieldName.START,
+            forecast_start_field=FieldName.FORECAST_START,
+            instance_sampler=instance_sampler,
+            past_length=self.context_length,
+            future_length=self.prediction_length,
+            dummy_value=np.nan,
+        )
+
+    def _infinite_iterable(self, data: Iterable[dict]):
```
canerturkmen, line 126: nit: using `itertools.cycle` saves a bit of code and makes the intent clearer.

timeseries/src/autogluon/timeseries/models/chronos/pipeline/utils.py
```diff
@@ -1,14 +1,210 @@
...
+
+        return {"context": past_target, "target": future_target}
+
+    def __iter__(self) -> Iterator:
+        if self.mode == "training":
+            iterable = self._create_training_data(self.gluonts_dataset)
+        elif self.mode == "validation":
+            iterable = self._create_validation_data(self.gluonts_dataset)
+
+        format_transform_fn = self.to_chronos_format if self.tokenizer is not None else self.to_chronos_bolt_format
```
canerturkmen, line 192: along similar lines to the tokenizer comment above, this line also suggests that it should rather be the pipelines who know how to handle their data formats...

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -273,15 +304,198 @@ def persist(self) -> "ChronosModel":
...
+                args=training_args,
+                train_dataset=tokenizer_train_dataset,
+                eval_dataset=tokenizer_val_dataset,
+                callbacks=callbacks,
+            )
+
+            if verbosity < 3:
+                # remove PrinterCallback from callbacks which logs to the console via a print() call,
+                # so cannot be handled by setting the log level
+                trainer.pop_callback(PrinterCallback)
```
canerturkmen, line 464: does this mean that if verbosity > 3 we will print TF stuff to stdout? if you feel like it, it might be worth overriding the `PrinterCallback` with `ChronosLogCallback(PrinterCallback)` that logs to our logger defined above.

canerturkmen (commented):
Thanks!! Looks good to me overall just some nits here and there. Yours to merge after fixing the tests.

timeseries/src/autogluon/timeseries/models/chronos/model.py
canerturkmen: sounds good.

timeseries/src/autogluon/timeseries/models/chronos/model.py
abdulfatir: How about this: if fine-tuning is used, we ignore the optimization_strategy? This way the model won't be reloaded every time. Someone who can afford to fine-tune on CPU can probably afford to wait for the slower inference.

abdulfatir: I am going with ignoring `optimization_strategy` when fine-tuning. Added a FIXME, just in case.

timeseries/src/autogluon/timeseries/models/chronos/pipeline/chronos_bolt.py
abdulfatir: Responded above.

abdulfatir: Done.

timeseries/src/autogluon/timeseries/models/chronos/pipeline/utils.py
abdulfatir: Added.

timeseries/src/autogluon/timeseries/models/chronos/pipeline/utils.py
abdulfatir: `itertools.cycle` caches the data which we don't want here.

timeseries/src/autogluon/timeseries/models/chronos/pipeline/utils.py
abdulfatir: Left a TODO.

timeseries/src/autogluon/timeseries/models/chronos/model.py
abdulfatir: Added `LoggerCallback`.

timeseries/src/autogluon/timeseries/models/chronos/pipeline/utils.py
canerturkmen: then something like `chain.from_iterable(cycle([data]))` is still cleaner. this defines a protected instance method for this simple functionality. there is nothing in this definition that is associated with the class or an instance, or anything to suggest this can be the case in the future. it's not shared across methods, or doesn't serve to break up a method...

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -107,9 +110,28 @@ class ChronosModel(AbstractTimeSeriesModel):
...
+        The number of gradient update steps to fine-tune for
+    fine_tune_batch_size : int, default = 16
+        The batch size to use for fine-tuning
+    fine_tune_shuffle_buffer_size : int, default = 10000
+        The size of the shuffle buffer to shuffle the data during fine-tuning. If None, shuffling will
+        be turned off.
+    eval_during_fine_tune : bool, default = False
+        If True, validation will be performed during fine-tuning to select the best checkpoint.
+        Setting this argument to True may result in slower fine-tuning.
+    fine_tune_eval_max_items : int, default = 100
```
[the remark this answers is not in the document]
shchur: 256 looks reasonable (=1 batch for bolt)

canerturkmen:
@shchur should we run platform tests?

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -107,9 +114,28 @@ class ChronosModel(AbstractTimeSeriesModel):
...
+    fine_tune_eval_max_items : int, default = 256
+        The maximum number of randomly-sampled time series to use from the validation set for evaluation
+        during fine-tuning. If None, the entire validation dataset will be used.
+    fine_tune_trainer_kwargs : dict, optional
+        Extra keyword arguments passed to ``transformers.TrainingArguments``
     device : str, default = None
-        Device to use for inference. If None, model will use the GPU if available. For larger model sizes
-        `small`, `base`, and `large`; inference will fail if no GPU is available.
+        Device to use for inference (and fine-tuning, if enabled). If None, model will use the GPU if available.
+        For larger model sizes `small`, `base`, and `large`; inference will fail if no GPU is available.
```
shchur, line 138: Nit: Let's put all the fine-tuning related arguments at the end (right before `keep_transformers_logs`)?

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -272,15 +308,229 @@ def persist(self) -> "ChronosModel":
...
+            if isinstance(self.model_pipeline, ChronosPipeline):
+                # Update prediction_length of the model
+                # NOTE: We only do this for ChronosPipeline because the prediction length of ChronosBolt models
+                # is fixed due to direct multistep forecasting setup
+                self.model_pipeline.model.config.prediction_length = fine_tune_prediction_length
+                self.model_pipeline.inner_model.config.chronos_config["prediction_length"] = (
+                    fine_tune_prediction_length
+                )
+
+            if isinstance(self.model_pipeline, ChronosBoltPipeline):
```
shchur, line 421: Should this be an if-else?

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -272,15 +308,229 @@ def persist(self) -> "ChronosModel":
...
+
+        eval_during_fine_tune = val_data is not None and fine_tune_args["eval_during_fine_tune"]
+
+        start_time = time.monotonic()
+        if do_fine_tune:
+            context_length = self._get_context_length(train_data)
+            # load model pipeline to device memory
+            self.load_model_pipeline(is_training=True)
+
+            extra_fine_tune_trainer_kwargs = {}
```
shchur, line 408: Nit: It seems that this can go into the `if isinstance(self.model_pipeline, ChronosPipeline):` condition

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -366,7 +621,7 @@ def _predict(
         return TimeSeriesDataFrame(df)
 
     def _more_tags(self) -> Dict:
-        return {"allow_nan": True}
+        return {"allow_nan": True, "can_use_val_data": True}
```
shchur, line 624:
```suggestion
        return {"allow_nan": True, "can_use_val_data": self._get_model_params()["fine_tune"]}
```
This will allow us to save some time on preprocessing val_data if fine-tuning is disabled

shchur (approved):
Thank you! Just a couple of nitpicks, otherwise LGTM! 🚀

canerturkmen:
Looks like the models are landing in MPS on mac again... [logs](https://productionresultssa16.blob.core.windows.net/actions-results/3a25b203-ded6-4d40-9a95-7e04c25ec1e4/workflow-job-run-f5b814b4-b405-5f2e-e8fb-2758af10ab75/logs/job/job-logs.txt?rsct=text%2Fplain&se=2024-11-12T15%3A27%3A10Z&sig=y%2FzhdZo5djrEKoi%2F2ZN%2BCN7oGcTju5pEorStiYx4QG0%3D&ske=2024-11-13T00%3A47%3A23Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2024-11-12T12%3A47%3A23Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2024-08-04&sp=r&spr=https&sr=b&st=2024-11-12T15%3A17%3A05Z&sv=2024-08-04) this will keep breaking platform tests. 

See also #4565

abdulfatir:
@canerturkmen it looks like this is happening for other models as well or is this only related to Chronos. 

```
2024-11-12T15:13:28.9118750Z ERROR tests/unittests/models/test_models.py::test_given_context_has_1_observation_when_model_predicts_then_model_can_predict[5-SimpleFeedForwardModel] - RuntimeError: MPS backend out of memory (MPS allocated: 712.00 MB, other allocations: 369.14 MB, max allowed: 7.93 GB). Tried to allocate 256 bytes on shared pool. Use PYTORCH_MPS_HIGH_WATERMARK_RATIO=0.0 to disable upper limit for memory allocations (may cause system failure).
2024-11-12T15:13:28.9120400Z ERROR tests/unittests/models/test_models.py::test_given_context_has_1_observation_when_model_predicts_then_model_can_predict[5-TemporalFusionTransformerModel] - RuntimeError: MPS backend out of memory (MPS allocated: 712.00 MB, other allocations: 369.14 MB, max allowed: 7.93 GB). Tried to allocate 256 bytes on shared pool. Use PYTORCH_MPS_HIGH_WATERMARK_RATIO=0.0 to disable upper limit for memory allocations (may cause system failure).
```

canerturkmen:
> @canerturkmen it looks like this is happening for other models as well or is this only related to Chronos.
> 
> ```
> 2024-11-12T15:13:28.9118750Z ERROR tests/unittests/models/test_models.py::test_given_context_has_1_observation_when_model_predicts_then_model_can_predict[5-SimpleFeedForwardModel] - RuntimeError: MPS backend out of memory (MPS allocated: 712.00 MB, other allocations: 369.14 MB, max allowed: 7.93 GB). Tried to allocate 256 bytes on shared pool. Use PYTORCH_MPS_HIGH_WATERMARK_RATIO=0.0 to disable upper limit for memory allocations (may cause system failure).
> 2024-11-12T15:13:28.9120400Z ERROR tests/unittests/models/test_models.py::test_given_context_has_1_observation_when_model_predicts_then_model_can_predict[5-TemporalFusionTransformerModel] - RuntimeError: MPS backend out of memory (MPS allocated: 712.00 MB, other allocations: 369.14 MB, max allowed: 7.93 GB). Tried to allocate 256 bytes on shared pool. Use PYTORCH_MPS_HIGH_WATERMARK_RATIO=0.0 to disable upper limit for memory allocations (may cause system failure).
> ```

At least for now let's work with the assumption that it's due to this PR, and we can't garbage collect between different tests.

abdulfatir:
@canerturkmen platform tests pass. The two windows failures are unrelated to this PR.


```
FAILED tests\unittests\models\chronos\test_model.py::test_when_chronos_scores_oof_and_time_limit_is_exceeded_then_exception_is_raised[default-0] - autogluon.core.utils.exceptions.TimeLimitExceeded
```
This fails because `time_limit=0.001` is too short and TLE is raised in the call to `fit()` even before calling `score_and_cache_oof()`. 

```
FAILED tests\unittests\test_predictor.py::test_when_plot_called_then_figure_contains_correct_number_of_subplots[1-None-1-mean-10-quantile_levels0-None] - _tkinter.TclError: Can't find a usable init.tcl in the following directories: 
```
This is some Tkinter error but given that this test passed in the previous platform test, I don't think this is related to the current PR.

canerturkmen (approved):
Thank you!

PR #4610, opened 2024-11-04
Add AdamW support to NN_TORCH

Innixma:
*Issue #, if available:*

*Description of changes:*

- Add AdamW support to NN_TORCH


tabular/src/autogluon/tabular/models/tabular_nn/torch/tabular_nn_torch.py
```diff
@@ -769,6 +769,8 @@ def _init_optimizer(self, optimizer, learning_rate, weight_decay):
             optimizer = torch.optim.SGD(params=self.model.parameters(), lr=learning_rate, weight_decay=weight_decay)
         elif optimizer == "adam":
             optimizer = torch.optim.Adam(params=self.model.parameters(), lr=learning_rate, weight_decay=weight_decay)
+        elif optimizer == "adamw":
```
tonyhoo, line 772: Can we add a test to cover this use case if it’s not already in place?

Innixma: While technically this would make sense, I'd prefer not to test every possible hyperparameter option, as it will be very time consuming for not much benefit. Rather, we can identify during large-scale hyperparameter sweeps as done in TabRepo whether a given option is performing well or has issues/bugs.

tonyhoo (approved):

PR #4613, opened 2024-11-04
Refactor TabularDataset

Innixma:
*Issue #, if available:*

*Description of changes:*

- refactor TabularDataset so that it no longer is a subclass of a pandas DataFrame, but rather *is* a pandas DataFrame.
- This simplifies a lot of logic, such that we only need to consider input types for pandas DataFrame rather than for TabularDataset.
- This also fixes IDE visualizers such as PyCharm which wouldn't properly display TabularDataset because it didn't recognize it as a pandas DataFrame.
- The extra attribute that TabularDataset used to have was never used, so this change is backwards compatible.
- Updated docstrings and type hints in TabularPredictor to remove mention of TabularDataset and cleanup / improve existing type hints.
- Moved TabularDataset from `core` to `common`, since it only depends on pandas and `autogluon.common`.

### DIFF
--- a/common/src/autogluon/common/__init__.py
+++ b/common/src/autogluon/common/__init__.py
@@ -1,3 +1,4 @@
+from .dataset import TabularDataset
 from .features.feature_metadata import FeatureMetadata
 from .utils.log_utils import _add_stream_handler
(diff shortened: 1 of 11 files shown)


prateekdesai04 (approved):
LGTM!

PR #4625, opened 2024-11-07
[timeseries] add Chronos bolt

canerturkmen:
*Issue #, if available:*

*Description of changes:*
- update directory structure 
- remove OptimizedChronosPipeline (merge into main `ChronosPipeline` class)
- remove `causal` models in ChronosPipeline
- add abstract `ForecastPipeline` (refactored with new `PipelineRegistry`)
- add Chronos Bolt source code ⚡ 
- temporarily add a "toy" Chronos-Bolt 350K model as a test fixture

TODOs
- [ ] remove test fixture once models are released
- [ ] add Bolt to model tests
- [ ] add Bolt to smoke tests
- [ ] add Bolt to `ChronosModel` defaults


timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -310,23 +312,38 @@ def _predict(
...
-                    .cpu()
-                    .numpy()
-                    for batch in inference_data_loader
-                ]
-
-        samples = np.concatenate(prediction_samples, axis=0).swapaxes(1, 2).reshape(-1, self.num_samples)
-
-        mean = samples.mean(axis=-1, keepdims=True)
-        quantiles = np.quantile(samples, self.quantile_levels, axis=-1).T
+                if self.model_pipeline.forecast_type == ForecastType.SAMPLES:
```
shchur, line 315: With this design we are implicitly making a decision that I'd like to discuss to make sure it's the right one. `ChronosModel` will be a generic wrapper for HF models that are loaded with `from_pretrained` and implement the `predict(context: torch.Tensor, prediction_length: int) -> torch.Tensor` API. To control which model is used, the users need to specify the `model_path` hyperparameter.
```python
hyperparameters={"Chronos": {"model_path": "autogluon/chronos-bolt-small"}}
```

The alternative option would involve creating a separate class (e.g., `ChronosBoltModel`) for the new set of Chronos models that directly return the quantile forecasts. This could be slightly more intuitive for the users since one hyperparameter completely changes the model architecture
```python
hyperparameters={"ChronosBolt": {"model_path": "small"}}
```

The first option seems cleaner and more future proof to me, but I anyway wanted to bring this up to make sure that we don't miss any important limitations of the chosen design.

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -310,23 +312,38 @@ def _predict(
...
+                            prediction_length=self.prediction_length,
+                            limit_prediction_length=False,
+                        )
+                        .detach()
+                        .cpu()
+                        .numpy()
+                        for batch in inference_data_loader
+                    ]
+                    model_quantiles = self.model_pipeline.quantiles
+                    quantiles = np.concatenate(predictions, axis=0).swapaxes(1, 2).reshape(-1, len(model_quantiles))
```
shchur, line 345: We will need to implement the quantile interpolation from `model_pipeline.quantiles ` to `self.quantile_levels` for the unit tests to pass

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -310,23 +312,38 @@ def _predict(
...
-                    .numpy()
-                    for batch in inference_data_loader
-                ]
-
-        samples = np.concatenate(prediction_samples, axis=0).swapaxes(1, 2).reshape(-1, self.num_samples)
-
-        mean = samples.mean(axis=-1, keepdims=True)
-        quantiles = np.quantile(samples, self.quantile_levels, axis=-1).T
+                if self.model_pipeline.forecast_type == ForecastType.SAMPLES:
+                    prediction_samples = [
```
abdulfatir, line 316: Maybe we can put the inference call/quantile computation in a separate function for clarity?

timeseries/src/autogluon/timeseries/models/chronos/pipeline/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .forecast_pipeline import ForecastPipeline, ForecastType
+from .chronos import ChronosPipeline
+from .chronos_bolt import PatchedT5Pipeline
```
abdulfatir, line 3: Naming? Let's go with `ChronosBoltPipeline`.

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -310,23 +312,38 @@ def _predict(
...
+                        self.model_pipeline.predict(
+                            batch,
+                            prediction_length=self.prediction_length,
+                            num_samples=self.num_samples,
+                            limit_prediction_length=False,
```
shchur, line 321: Nit: We can remove this line since the default value is changed to `False`

timeseries/src/autogluon/timeseries/models/chronos/pipeline/chronos_bolt.py
```diff
@@ -0,0 +1,486 @@
...
+    T5Stack,
+)
+from transformers.utils import ModelOutput
+
+from . import ForecastPipeline, ForecastType
+from .utils import left_pad_and_stack_1D
+
+
+@dataclass
+class PatchedT5Config:
```
abdulfatir, line 26:
```suggestion
class ChronosBoltConfig:
```

timeseries/src/autogluon/timeseries/models/chronos/pipeline/chronos_bolt.py
```diff
@@ -0,0 +1,486 @@
...
+
+        return (x - loc) / scale, (loc, scale)
+
+    def inverse(self, x: torch.Tensor, loc_scale: Tuple[torch.Tensor, torch.Tensor]) -> torch.Tensor:
+        loc, scale = loc_scale
+        return x * scale + loc
+
+
+@dataclass
+class PatchedT5Output(ModelOutput):
```
abdulfatir, line 85:
```suggestion
class ChronosBoltOutput(ModelOutput):
```

timeseries/src/autogluon/timeseries/models/chronos/pipeline/chronos.py
```diff
@@ -11,14 +11,17 @@
 
 import torch
 import torch.nn as nn
-from transformers import AutoConfig, AutoModelForCausalLM, AutoModelForSeq2SeqLM, GenerationConfig, PreTrainedModel
+from transformers import AutoConfig, AutoModelForSeq2SeqLM, GenerationConfig, PreTrainedModel
 
 from autogluon.timeseries.utils.warning_filters import set_loggers_level
 
+from . import ForecastPipeline, ForecastType
```
shchur, line 18:
```suggestion
from .forecast_pipeline import ForecastPipeline, ForecastType
```
Does the current code result in circular imports error?

timeseries/src/autogluon/timeseries/models/chronos/pipeline/chronos_bolt.py
```diff
@@ -0,0 +1,486 @@
...
+        res = self.residual_layer(x)
+
+        out = out + res
+
+        if self.use_layer_norm:
+            return self.layer_norm(out)
+        return out
+
+
+class BasePatchedModelForForecasting(T5PreTrainedModel):
```
abdulfatir, line 126:
```suggestion
class BaseChronosBoltModelForForecasting(T5PreTrainedModel):
```

timeseries/src/autogluon/timeseries/models/chronos/pipeline/chronos_bolt.py
```diff
@@ -0,0 +1,486 @@
...
+            encoder_hidden_states=hidden_states,
+            encoder_attention_mask=attention_mask,
+            output_attentions=output_attentions,
+            return_dict=True,
+        )
+
+        return decoder_outputs.last_hidden_state  # sequence_outputs, b x 1 x d_model
+
+
+class PatchedT5EncoderForForecasting(BasePatchedModelForForecasting):
```
abdulfatir, line 377: Are we keeping this? Let's remove this for clarify since we're not using it.

timeseries/src/autogluon/timeseries/models/chronos/pipeline/chronos_bolt.py
```diff
@@ -0,0 +1,486 @@
...
+            predictions.append(prediction)
+            remaining -= prediction.shape[-1]
+
+            if remaining <= 0:
+                break
+
+            central_idx = torch.abs(torch.tensor(self.model.config.chronos_config["quantiles"]) - 0.5).argmin()
+            central_prediction = prediction[:, central_idx]
+
+            context_tensor = torch.cat([context_tensor, central_prediction], dim=-1)
```
abdulfatir, line 458: Are we going with the 🎄 effect for now? Or should we use a different style of unrolling?

timeseries/src/autogluon/timeseries/models/chronos/pipeline/forecast_pipeline.py
```diff
@@ -0,0 +1,87 @@
...
+
+import torch
+
+
+class ForecastType(Enum):
+    SAMPLES = "samples"
+    QUANTILES = "quantiles"
+
+
+class PipelineRegistry(type):
```
abdulfatir, line 17: Nice. You really like registries.

timeseries/src/autogluon/timeseries/models/chronos/pipeline/chronos_bolt.py
canerturkmen: I think yes as we haven't evaluated with anything else. I can drop a todo here.

timeseries/src/autogluon/timeseries/models/chronos/model.py
canerturkmen: I am also comfortable with (1) for now. A "ChronosModel" in this sense is anything that can be loaded by a `ForecastPipeline` (which I think we should also rename to something like `BaseChronosPipeline`). That means it is compatible with our Pipeline API, as well as persisted with a Chronos config or `chronos_pipeline_class`. We could alias to get the effect of (2), though I'm a bit afraid that could complicate things.

timeseries/src/autogluon/timeseries/models/chronos/model.py
shchur: I'm fine with just clipping to the max quantile available for the model for extrapolation. Have we tried training the base models with the [0.05, 0.10, ..., 0.95] quantile levels?

canerturkmen: No we didn't. I think clipping is quite a silent side effect, should we warn/err out instead?

shchur: Warning in the logs is fine, I don't think that we should error since this will completely remove the model in case the user cares about high/low quantiles.

timeseries/src/autogluon/timeseries/models/chronos/pipeline/chronos_bolt.py
```diff
@@ -449,16 +449,23 @@ def predict_quantiles(
...
-            mean = predictions[:, :, self.quantiles.index("0.5")]
-            return quantiles, mean
+            quantiles = predictions[..., [self.quantiles.index(q) for q in quantile_levels_labels]]
         else:
-            # TODO: Handle interpolation for arbitrary quantiles
-            raise NotImplementedError()
+            # we rely on torch for interpolating quantiles if quantiles that
+            # Chronos Bolt was trained on were not provided
+            augmented_predictions = torch.cat(
+                [predictions[..., [0]], predictions, predictions[..., [-1]]],
```
shchur, line 464: This piece of code seems to linearly interpolate between the P10 and P90 predictions, so the interpolated P51 might be far from the predicted P50. Maybe we can use pandas for interpolation instead?
```python
import pandas as pd

data = {
    0.1: [1, 2, 3],
    0.2: [1.5, 2.5, 3.5],
    0.3: [2, 3, 4],
    0.4: [2.5, 3.5, 4.5],
    0.5: [3, 4, 5],
    0.6: [3.5, 4.5, 5.5],
    0.7: [4, 5, 6],
    0.8: [4.5, 5.5, 6.5],
    0.9: [5, 6, 7]
}
df = pd.DataFrame(data)
query_quantiles = [0.05, 0.15, 0.25, 0.5, 0.75]

df = df.reindex(columns=sorted(set(df.columns).union(query_quantiles))).interpolate(axis=1)
df = df.bfill(axis=1).ffill(axis=1)  # Extrapolate quantiles outside of the [0.1, 0.9] range by "clipping"
result_df = df[query_quantiles]
```

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -50,22 +65,30 @@
...
+    base, and large variants a GPU is required to perform inference efficiently. Chronos takes a minimalistic approach to
+    pretraining time series models, by discretizing time series data directly into bins which are treated as tokens,
+    effectively performing regression by classification. This results in a simple and flexible framework
     for using any language model in the context of time series forecasting. See [Ansari2024]_ for more information.
 
+    The newer Chronos-Bolt variants enable much faster inference by first "patching" the time series. The resulting
+    time series is then fed into a T5 model for forecasting. The Chronos-Bolt variants are capable of performing inference
+    on CPUs, and are available on Hugging Face <https://huggingface.co/collections/>`_.
+    Chronos-Bolt variants are capable of much faster inference, especially on CPUs. Chronos-Bolt variants can be used with
+
```
shchur, line 91: The last sentence seems broken?

timeseries/src/autogluon/timeseries/models/chronos/pipeline/utils.py
```diff
@@ -8,6 +11,44 @@
...
+        if not force and dest.exists():
+            continue
+        if not dest.parent.exists():
+            dest.parent.mkdir(exist_ok=True, parents=True)
+        bucket.download_file(obj.key, str(dest))
+
+
+def cache_model_from_s3(s3_uri: str, force=False):
+    assert re.match("^s3://([^/]+)/(.*?([^/]+)/?)$", s3_uri) is not None, f"Not a valid S3 URI: {s3_uri}"
+    cache_home = Path(os.environ.get("XGD_CACHE_HOME", os.path.expanduser("~/.cache")))
```
shchur, line 45: 1. Should this be `XDG_CACHE_HOME` instead of `XGD_CACHE_HOME`?
2. Does this work on Windows (especially the second part with both `~`, `/` and `.`)?

timeseries/src/autogluon/timeseries/models/chronos/pipeline/utils.py
```diff
@@ -8,6 +11,44 @@
...
+    padded = []
+    for c in tensors:
+        assert isinstance(c, torch.Tensor)
+        assert c.ndim == 1
+        padding = torch.full(size=(max_len - len(c),), fill_value=torch.nan, device=c.device)
+        padded.append(torch.concat((padding, c), dim=-1))
+    return torch.stack(padded)
+
+
+def download_prefix(bucket, prefix, local_path, force: bool = False, boto3_session=None) -> None:
```
shchur, line 25: Maybe we can use some of the existing functionality in `autogluon.common.loaders.load_s3` (or at least move this method to `common`)?

timeseries/src/autogluon/timeseries/models/chronos/pipeline/base.py
```diff
@@ -0,0 +1,146 @@
...
+        """
+        Load the model, either from a local path or from the HuggingFace Hub.
+        Supports the same arguments as ``AutoConfig`` and ``AutoModel``
+        from ``transformers``.
+
+        When a local path is provided, supports both a folder or a .tar.gz archive.
+        """
+        from transformers import AutoConfig
+
+        if str(pretrained_model_name_or_path).startswith("s3://"):
```
shchur, line 125: Is this just for internal usage, or do we plan to expose this in the user-facing documentation?

timeseries/src/autogluon/timeseries/utils/warning_filters.py
```diff
@@ -70,3 +71,22 @@ def disable_stdout():
     sys.stdout = io.StringIO()
     yield
     sys.stdout = save_stdout
+
+
+class DuplicateLogFilter:
```
shchur, line 76: Nice!

timeseries/src/autogluon/timeseries/utils/warning_filters.py
```diff
@@ -70,3 +71,22 @@ def disable_stdout():
...
+        self.max_count = max_count
+
+    def filter(self, record):
+        count = self.messages[record.msg]
+        self.messages[record.msg] += 1
+        return count < self.max_count
+
+
+@contextlib.contextmanager
+def disable_duplicate_logs(logger, max_count: int = 1):
```
shchur, line 88: Where do the repeated log messages originate from?

timeseries/src/autogluon/timeseries/models/chronos/pipeline/chronos_bolt.py
```diff
@@ -0,0 +1,507 @@
...
+        if set(quantile_levels).issubset(set(training_quantile_levels)):
+            # no need to perform intra/extrapolation
+            quantiles = predictions[..., [training_quantile_levels.index(q) for q in quantile_levels]]
+        else:
+            # we rely on torch for interpolating quantiles if quantiles that
+            # Chronos Bolt was trained on were not provided
+            if min(quantile_levels) < min(training_quantile_levels) or max(quantile_levels) > max(
+                training_quantile_levels
+            ):
+                logger.warn(
```
shchur, line 463: Nit: `logger.warn` is deprecated, `logger.warning` should be used instead

timeseries/src/autogluon/timeseries/models/chronos/pipeline/chronos_bolt.py
```diff
@@ -0,0 +1,507 @@
...
+                    f"\tQuantiles to be predicted ({quantile_levels}) are not within the range of "
+                    f"quantiles that Chronos-Bolt was trained on ({training_quantile_levels}). "
+                    "Quantile predictions will be set to the minimum/maximum levels at which Chronos-Bolt "
+                    "was trained on. This may significantly affect the quality of the predictions."
+                )
+
+            augmented_predictions = torch.cat(
+                [predictions[..., [0]], predictions, predictions[..., [-1]]],
+                dim=-1,
+            )
```
shchur, line 473: Do I understand correctly that this logic only works if `training_quantile_levels == [0.1, 0.2, ..., 0.9]`? 

If that is the case, should we add a TODO here indicating that this part needs to be modified if we train models with other quantile levels? (since other places in the code, e.g., `central_idx` above seem to be generic)

timeseries/src/autogluon/timeseries/models/chronos/pipeline/chronos_bolt.py
```diff
@@ -0,0 +1,507 @@
...
+            quantiles = torch.quantile(
+                augmented_predictions, q=torch.tensor(quantile_levels, dtype=augmented_predictions.dtype), dim=-1
+            ).permute(1, 2, 0)
+        mean = predictions[:, :, training_quantile_levels.index(0.5)]
+        return quantiles, mean
```
shchur, line 478: Should we sort the quantiles tensor along the 1st dim to ensure that the quantiles are not crossing?

.gitignore
```diff
@@ -137,3 +137,6 @@ eda/src/autogluon/eda/version.py
 AutogluonModels
 lightning_logs
 VERSION.minor
+
+# timeseries text fixtures
+!timeseries/tests/fixtures/**/config.json
```
shchur, line 142: Do we need this?

shchur (commented):
Thanks a lot for the PR! Looks great, just a few minor comments

timeseries/tests/unittests/models/test_models.py
```diff
@@ -28,7 +28,7 @@
     get_data_frame_with_item_index,
     to_supported_pandas_freq,
 )
```
shchur, line 30: Should we run platform tests for a good measure?

timeseries/src/autogluon/timeseries/models/chronos/pipeline/base.py
canerturkmen: Great point. I thought internal for now.

timeseries/src/autogluon/timeseries/models/chronos/pipeline/chronos.py
```diff
@@ -442,42 +424,33 @@ def predict(
...
+                context,
+                prediction_length=prediction_length,
+                num_samples=num_samples,
+            )
+            .detach()
+            .cpu()
+            .swapaxes(1, 2)
+        )
+        mean = prediction_samples.mean(axis=-1, keepdims=True)
+        quantiles = torch.quantile(
```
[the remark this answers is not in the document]
abdulfatir: I think `.permute(1, 2, 0)` would suffice like in `chronos_bolt`.

timeseries/src/autogluon/timeseries/models/chronos/pipeline/chronos_bolt.py
canerturkmen: Right. As long as the predictions are along an equidistant grid along the quantile axis, then I think the logic survives. But you're absolutely right. I'll drop a todo here.

timeseries/src/autogluon/timeseries/utils/warning_filters.py
canerturkmen: They come from the fact that predict_quantiles is called repeatedly by `ChronosModel`. 

Currently the `chronos` package takes care to only minimally change the model implementations from the upstream sources (in the `pipeline` package) and wrap them in `ChronosModel`. i.e., only `ChronosModel` is aware of its data loader which is not passed down to the pipeline implementations. This is an artifact of our choice to wrap pipeline implementations. However, this also results in this quirk: the bolt pipeline implementation will only know that it's being called with the wrong quantiles if it's going to be called multiple times with the same quantiles in the dataloader loop.

timeseries/src/autogluon/timeseries/models/chronos/pipeline/chronos_bolt.py
canerturkmen: addressed for now in code, also benchmarking it.

shchur (approved):
LGTM! ⚡

timeseries/tests/unittests/models/test_models.py
canerturkmen: platform tests look good at https://github.com/autogluon/autogluon/actions/runs/11781956077

timeseries/src/autogluon/timeseries/models/chronos/pipeline/chronos_bolt.py
canerturkmen: benchmark looks good 👍

PR #4629, opened 2024-11-09
[AutoMM] Fix Logloss Bug and Refine Compute Score Logics

FANGAreNotGnu:

### DIFF
--- a/multimodal/src/autogluon/multimodal/utils/metric.py
+++ b/multimodal/src/autogluon/multimodal/utils/metric.py
@@ -234,7 +234,9 @@ def compute_score(
 
     y = metric_data[Y_TRUE]
     if metric.needs_proba or metric.needs_threshold:
-        return metric.convert_score_to_original(compute_metric(y=y, y_pred_proba=metric_data[Y_PRED_PROB][:, pos_label], metric=metric, weights=None))
+        return metric.convert_score_to_original(
+            compute_metric(y=y, y_pred_proba=metric_data[Y_PRED_PROB][:, pos_label], metric=metric, weights=None)
+        )
     else:
         y_pred = metric_data[Y_PRED]
 
--- a/multimodal/src/autogluon/multimodal/utils/metric.py
+++ b/multimodal/src/autogluon/multimodal/utils/metric.py
@@ -17,8 +17,6 @@ from ..constants import (
     BINARY,
     DIRECT_LOSS,
     F1,
-    F1_MACRO,
-    F1_MICRO,
     FEW_SHOT_CLASSIFICATION,
     IOU,
     MAP,
@@ -239,7 +237,7 @@ def compute_score(
     else:
         y_pred = metric_data[Y_PRED]
 
-    if metric.name in [F1, F1_MACRO, F1_MICRO]:  # only for binary classification
+    if metric.name == F1:  # only for binary classification
         y = (metric_data[Y_TRUE] == pos_label).astype(int)
         y_pred = (y_pred == pos_label).astype(int)
     else:
--- a/multimodal/src/autogluon/multimodal/utils/metric.py
+++ b/multimodal/src/autogluon/multimodal/utils/metric.py
@@ -244,12 +244,8 @@ def compute_score(
         y_pred = (y_pred == pos_label).astype(int)
     else:
         y = metric_data[Y_TRUE]
-    
-    return metric.convert_score_to_original(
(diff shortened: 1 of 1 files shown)


multimodal/src/autogluon/multimodal/utils/metric.py
```diff
@@ -231,15 +232,18 @@ def compute_score(
...
+    if metric.needs_proba or metric.needs_threshold:
+        y_pred = metric_data[Y_PRED_PROB][:, pos_label]
     else:
-        try:
-            return metric._sign * metric(metric_data[Y_TRUE], metric_data[Y_PRED], y_prob=metric_data[Y_PRED_PROB])
-        except:
-            return metric._sign * metric(metric_data[Y_TRUE], metric_data[Y_PRED])
+        y_pred = metric_data[Y_PRED]
+
+    if metric.name == F1:  # only for binary classification
```
Innixma, line 240: Add:

```python
# TODO: This is a hack. Doesn't support `f1_macro`, `f1_micro`, `f1_weighted`, or custom `f1` metrics with different names.
# TODO: Longterm the solution should be to have the input data to this function use the internal representation without the original class names. This way `pos_label` would not need to be specified.
```

multimodal/src/autogluon/multimodal/utils/metric.py
```diff
@@ -231,15 +232,18 @@ def compute_score(
...
+        y = (metric_data[Y_TRUE] == pos_label).astype(int)
+        y_pred = (y_pred == pos_label).astype(int)
+    else:
+        y = metric_data[Y_TRUE]
```
Innixma, line 244: Put this under the `else` statement above, since it only applies when not `y_pred_proba`.

Move `y = metric_data[Y_TRUE]` to the top of the function.

Innixma (approved):
LGTM!

multimodal/src/autogluon/multimodal/utils/metric.py
```diff
@@ -231,15 +231,22 @@ def compute_score(
...
-            return metric._sign * metric(metric_data[Y_TRUE], metric_data[Y_PRED], y_prob=metric_data[Y_PRED_PROB])
-        except:
-            return metric._sign * metric(metric_data[Y_TRUE], metric_data[Y_PRED])
+        y_pred = metric_data[Y_PRED]
+
+        # TODO: This is a hack. Doesn't support `f1_macro`, `f1_micro`, `f1_weighted`, or custom `f1` metrics with different names.
+        # TODO: Longterm the solution should be to have the input data to this function use the internal representation without the original class names. This way `pos_label` would not need to be specified.
+        if metric.name == F1:  # only for binary classification
+            y = (metric_data[Y_TRUE] == pos_label).astype(int)
+            y_pred = (metric_data[Y_PRED] == pos_label).astype(int)
```
Innixma, line 247:
```suggestion
            y = (y == pos_label).astype(int)
            y_pred = (y_pred == pos_label).astype(int)
```

multimodal/src/autogluon/multimodal/utils/metric.py
```diff
@@ -231,15 +231,22 @@ def compute_score(
...
     metric = get_metric(metric)
-    if metric.name in [ROC_AUC, AVERAGE_PRECISION]:
-        return metric._sign * metric(metric_data[Y_TRUE], metric_data[Y_PRED_PROB][:, pos_label])
-    elif metric.name in [F1]:  # only for binary classification
-        return f1_score(metric_data[Y_TRUE], metric_data[Y_PRED], pos_label=pos_label)
+
+    y = metric_data[Y_TRUE]
+    if metric.needs_proba or metric.needs_threshold:
+        return metric.convert_score_to_original(
+            compute_metric(y=y, y_pred_proba=metric_data[Y_PRED_PROB][:, pos_label], metric=metric, weights=None)
```
Innixma, line 238: [Major] Won't this crash for `multiclass` problem_type?

For example, `log_loss` with multiclass, due to `[:, pos_label]` , as there is no such thing as `pos_label`.

If yes, please add unit test to catch this.

multimodal/src/autogluon/multimodal/utils/metric.py
Innixma, line 247: F1 can also be used for multiclass classification. Specifically `f1_macro`, `f1_micro`, `f1_weighted`.

Ensure you also check `problem_type == BINARY`

PR #4634, opened 2024-11-09
[timeseries] Move covariate scaling logic into a separate class

shchur:
*Issue #, if available:*

*Description of changes:*
- Combine methods `preprocess` and `preprocess_known_covariates` into a single method. This simplifies certain types of preprocessing such as converting past_covariates into known_covariates using lags.
- Move GluonTS preprocessing logic into a standalone class `CovariateScaler` that can be used by any `AbstractTimeSeriesModel`
- Extend test coverage for the covariate scaling logic

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
+++ b/timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
@@ -16,7 +16,12 @@ from autogluon.core.models import AbstractModel
 from autogluon.timeseries.dataset import TimeSeriesDataFrame
 from autogluon.timeseries.metrics import TimeSeriesScorer, check_get_evaluation_metric
 from autogluon.timeseries.regressor import CovariateRegressor
-from autogluon.timeseries.transforms import LocalTargetScaler, get_target_scaler_from_name
+from autogluon.timeseries.transforms import (
+    CovariateScaler,
+    LocalTargetScaler,
+    get_covariate_scaler_from_name,
+    get_target_scaler_from_name,
+)
 from autogluon.timeseries.utils.features import CovariateMetadata
 from autogluon.timeseries.utils.forecast import get_forecast_horizon_index_ts_dataframe
 from autogluon.timeseries.utils.warning_filters import disable_stdout, warning_filter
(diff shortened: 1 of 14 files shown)


timeseries/src/autogluon/timeseries/transforms/covariate_scaler.py
```diff
@@ -0,0 +1,158 @@
...
+        self.metadata = metadata
+        self.use_known_covariates = use_known_covariates
+        self.use_past_covariates = use_past_covariates
+        self.use_static_features = use_static_features
+
+    def fit_transform(
+        self,
+        data: TimeSeriesDataFrame,
+        known_covariates: Optional[TimeSeriesDataFrame] = None,
+    ) -> Tuple[TimeSeriesDataFrame, Optional[TimeSeriesDataFrame]]:
```
canerturkmen, line 36: The fact that `LocalTargetScaler.transform() -> TimeSeriesDataFrame` and `CovariateScaler.transform() -> Tuple` is not trivial. It might be worth documenting this in the transform methods of both classes.

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
```diff
@@ -290,6 +303,20 @@ def _create_target_scaler(self) -> Optional[LocalTargetScaler]:
         else:
             return None
 
+    def _create_covariate_scaler(self) -> Optional[CovariateScaler]:
+        """Create a CovariateScaler object based on the value of the `covariate_scaler` hyperparameter."""
+        covariate_scaler_type = self._get_model_params().get("covariate_scaler")
```
canerturkmen, line 308: just to check: is `"global"` set as the default somewhere, i.e., are we setting it as the default for all `AbstractTimeSeriesModel`s or not? (I assume not)

canerturkmen (approved):
LGTM overall. I assumed no logic changed during the refactor of covariate scaling. Thanks a lot for raising this!

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
shchur: We only set it as default for GluonTS models (https://github.com/autogluon/autogluon/pull/4634/files#diff-28a0cc813c0c8a1cd39a4e8c358ee689718114207a2655af00cb4ae7675d7665R293) so the behavior is equivalent to before the refactor

timeseries/src/autogluon/timeseries/transforms/covariate_scaler.py
shchur: Good point, will add a docstring. The alternative to returning a tuple is saving an internal state and applying to `known_covariates` - which is a bit more tedious

shchur: You are right, we can actually split the preprocessing for data & known_covariates into separate methods, making the API cleaner.

PR #4635, opened 2024-11-09
Fix Torch accidentally being imported immediately

Innixma:
*Issue #, if available:*

*Description of changes:*

- Fix Torch accidentally being imported immediately on `import autogluon.tabular`


cheungdaven (approved):
LGTM

PR #4640, opened 2024-11-12
[timeseries] refactor GluonTS default parameter handling, update TiDE parameters

canerturkmen:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
+++ b/timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
@@ -166,6 +166,7 @@ class AbstractGluonTSModel(AbstractTimeSeriesModel):
     # default number of samples for prediction
     default_num_samples: int = 250
     supports_cat_covariates: bool = False
+    _default_hyperparameters = {}
 
     def __init__(
         self,
@@ -287,10 +288,6 @@ class AbstractGluonTSModel(AbstractTimeSeriesModel):
 
         self.negative_data = (dataset[self.target] < 0).any()
 
-    @property
-    def default_context_length(self) -> int:
-        return min(512, max(10, 2 * self.prediction_length))
-
     def preprocess(self, data: TimeSeriesDataFrame, is_train: bool = False, **kwargs) -> TimeSeriesDataFrame:
         # Copy data to avoid SettingWithCopyWarning from pandas
         data = data.copy()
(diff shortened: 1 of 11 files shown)


timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
```diff
@@ -372,9 +378,6 @@ def _get_model_params(self) -> dict:
                 callbacks=self.callbacks,
             )
         )
-        # Support MXNet kwarg names for backwards compatibility
-        init_args.setdefault("lr", init_args.get("learning_rate", 1e-3))
-        init_args.setdefault("max_epochs", init_args.get("epochs", 100))
```
shchur, line 377: Are we intentionally deprecating the backward-compatible argument names `learning_rate` and `epochs`?

If that's the case, maybe we should log a warning when unsupported arguments are present (so that we don't just quietly change the behavior for users who use the old args)?

timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
```diff
@@ -372,9 +378,6 @@ def _get_model_params(self) -> dict:
                 callbacks=self.callbacks,
             )
         )
```
shchur, line 380: I think we can fold `freq`, `prediction_length` and `quantiles` into `_get_default_params` as well.

As for callbacks, we can probably just directly set them in `_get_estimator` as
```python
        default_trainer_kwargs = {
           ...
            "callbacks": self.callbacks,
           ...
        }
```

timeseries/src/autogluon/timeseries/models/gluonts/torch/models.py
```diff
@@ -496,13 +496,13 @@ class TiDEModel(AbstractGluonTSModel):
...
         Number of epochs the model will be trained for
-    batch_size : int, default = 64
+    batch_size : int, default = 256
         Size of batches used during training
     predict_batch_size : int, default = 500
         Size of batches used during prediction.
     num_batches_per_epoch : int, default = 50
         Number of batches processed every epoch
-    lr : float, default = 1e-3,
+    lr : float, default = 1e-4,
```
shchur, line 505: Interesting!

timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
```diff
@@ -357,13 +353,23 @@ def preprocess_known_covariates(
             known_covariates[columns] = self._real_column_transformers["known"].transform(known_covariates[columns])
         return known_covariates
 
+    def _get_default_params(self):
+        """Gets default parameters for GluonTS estimator initialization that are available after
+        AbstractTimeSeriesModel initialization (i.e., before deferred initialization)
+        """
+        # Override library-wide defaults with specific hyperparameter defaults provided by the model
```
shchur, line 360: Nit: Should this go into the docstring?

shchur (commented):
Thanks, I really like this refactor! Just a few minor comments

timeseries/src/autogluon/timeseries/models/gluonts/torch/models.py
shchur: Did you also try values between 1e-3 and 1e-4?

canerturkmen: not really. but the difference between 1e-3 and 1e-4 was significant.

PR #4641, opened 2024-11-12
[timeseries] Estimate the regressor prediction time

shchur:
*Issue #, if available:*

*Description of changes:*
- Disable the CovariateRegressor if predicting with the regressor is expected to exceed the time_limit
- Skip transforming the `train_data` with the covariate regressor if the model does not use the `train_data` (e.g., Chronos without fine-tuning)
- Make it possible to pass a `dict` to the `covariate_regressor` argument to simplify preset definition

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
+++ b/timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
@@ -303,19 +303,19 @@ class AbstractTimeSeriesModel(AbstractModel):
         covariate_regressor = self._get_model_params().get("covariate_regressor")
         if covariate_regressor is not None:
             if len(self.metadata.known_covariates + self.metadata.static_features) == 0:
-                logger.debug(
-                    "Skipping CovariateRegressor since the dataset contains no covariates or static features."
+                logger.info(
+                    "\tSkipping covariate_regressor since the dataset contains no covariates or static features."
                 )
                 return None
             else:
                 if isinstance(covariate_regressor, str):
                     return CovariateRegressor(covariate_regressor, target=self.target, metadata=self.metadata)
-                elif isinstance(covariate_regressor, (dict, CovariateRegressor)):
+                elif isinstance(covariate_regressor, dict):
+                    return CovariateRegressor(**covariate_regressor, target=self.target, metadata=self.metadata)
+                elif isinstance(covariate_regressor, CovariateRegressor):
                     logger.warning(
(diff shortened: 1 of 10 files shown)


timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
```diff
@@ -60,6 +60,7 @@ class AbstractTimeSeriesModel(AbstractModel):
     """
 
     _oof_filename = "oof.pkl"
+    _regressor_fit_time_fraction: float = 0.5
```
canerturkmen, line 63: `_covariate_regressor_fit_time_fraction`? Also, if we have this as a class variable, shouldn't we already yhink about some models where it can be higher and override?

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
```diff
@@ -255,16 +257,22 @@ def fit(
...
-                time_limit=0.5 * time_limit if time_limit is not None else None,
+                time_limit=regressor_time_limit,
+                verbosity=kwargs.get("verbosity", 2) - 1,
             )
 
-        train_data = self.preprocess(train_data, is_train=True)
+        if self._get_tags()["can_use_train_data"]:
+            if self.covariate_regressor is not None:
+                train_data = self.covariate_regressor.transform(train_data)
+            train_data = self.preprocess(train_data, is_train=True)
```
canerturkmen, line 274: is it now the case that `train_data` can be undefined during the `super().fit()` call? i.e., should we default it to `train_data=None` above?

canerturkmen (approved):
Very nice. Added two comments. Thanks!

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
shchur: The `train_data` will always be a `TimeSeriesDataFrame`. However, in case `{"can_use_train_data": False}`, we will pass an un-preprocessed dataframe to `super()._fit()` since the dataframe anyway won't be used inside (to save time on preprocessing).

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
shchur: I don't have an idea for what a reasonable override will be since this significantly depends on the hardware, model type, dataset size, absolute time_limit. 

I will add a TODO so that we think about it at some point in the future.

PR #4643, opened 2024-11-13
[AutoMM] Fix Proba Metrics for Multiclass

FANGAreNotGnu:
https://github.com/autogluon/autogluon/pull/4629 follow up

### DIFF
--- a/multimodal/src/autogluon/multimodal/utils/metric.py
+++ b/multimodal/src/autogluon/multimodal/utils/metric.py
@@ -234,9 +234,13 @@ def compute_score(
 
     y = metric_data[Y_TRUE]
     if metric.needs_proba or metric.needs_threshold:
-        y_pred_proba = metric_data[Y_PRED_PROB] 
-        y_pred_proba = y_pred_proba if y_pred_proba.shape[1] > 2 else y_pred_proba[:, pos_label]  # only use pos_label for binary classification
-        return metric.convert_score_to_original(compute_metric(y=y, y_pred_proba=y_pred_proba, metric=metric, weights=None))
+        y_pred_proba = metric_data[Y_PRED_PROB]
+        y_pred_proba = (
+            y_pred_proba if y_pred_proba.shape[1] > 2 else y_pred_proba[:, pos_label]
+        )  # only use pos_label for binary classification
+        return metric.convert_score_to_original(
+            compute_metric(y=y, y_pred_proba=y_pred_proba, metric=metric, weights=None)
+        )
     else:
         y_pred = metric_data[Y_PRED]
 
--- a/multimodal/src/autogluon/multimodal/utils/metric.py
+++ b/multimodal/src/autogluon/multimodal/utils/metric.py
@@ -248,7 +248,7 @@ def compute_score(
         # TODO: Longterm the solution should be to have the input data to this function use the internal representation without the original class names. This way `pos_label` would not need to be specified.
         if metric.name == F1:  # only for binary classification
             y = (y == pos_label).astype(int)
-            y_pred = (y == pos_label).astype(int)
+            y_pred = (y_pred == pos_label).astype(int)
 
         return metric.convert_score_to_original(compute_metric(y=y, y_pred=y_pred, metric=metric, weights=None))
 
--- a/multimodal/src/autogluon/multimodal/utils/metric.py
+++ b/multimodal/src/autogluon/multimodal/utils/metric.py
@@ -234,17 +234,17 @@ def compute_score(
 
     y = metric_data[Y_TRUE]
     if metric.needs_proba or metric.needs_threshold:
-        return metric.convert_score_to_original(
-            compute_metric(y=y, y_pred_proba=metric_data[Y_PRED_PROB][:, pos_label], metric=metric, weights=None)
-        )
+        y_pred_proba = metric_data[Y_PRED_PROB] 
+        y_pred_proba = y_pred_proba if y_pred_proba.shape[1] > 2 else y_pred_proba[:, pos_label]  # only use pos_label for binary classification
+        return metric.convert_score_to_original(compute_metric(y=y, y_pred_proba=y_pred_proba, metric=metric, weights=None))
     else:
         y_pred = metric_data[Y_PRED]
 
         # TODO: This is a hack. Doesn't support `f1_macro`, `f1_micro`, `f1_weighted`, or custom `f1` metrics with different names.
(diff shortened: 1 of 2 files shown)


multimodal/src/autogluon/multimodal/utils/metric.py
```diff
@@ -234,17 +234,21 @@ def compute_score(
...
+            compute_metric(y=y, y_pred_proba=y_pred_proba, metric=metric, weights=None)
         )
     else:
         y_pred = metric_data[Y_PRED]
 
         # TODO: This is a hack. Doesn't support `f1_macro`, `f1_micro`, `f1_weighted`, or custom `f1` metrics with different names.
         # TODO: Longterm the solution should be to have the input data to this function use the internal representation without the original class names. This way `pos_label` would not need to be specified.
         if metric.name == F1:  # only for binary classification
-            y = (metric_data[Y_TRUE] == pos_label).astype(int)
-            y_pred = (metric_data[Y_PRED] == pos_label).astype(int)
```
cheungdaven, line 247: why replace Y_TRUE with Y_PRED?

multimodal/tests/unittests/others/test_classification.py
```diff
@@ -57,3 +57,33 @@ def test_focal_loss_multiclass(checkpoint_name):
...
+        hyperparameters={
+            "model.mmdet_image.checkpoint_name": checkpoint_name,
+            "env.num_gpus": -1,
+            "optimization.max_epochs": 1,
+        },
+        train_data=train_data,
+        time_limit=30,  # seconds
+    )  # you can trust the default config, e.g., we use a `swin_base_patch4_window7_224` model
+
+    score = predictor.evaluate(train_data)
```
cheungdaven, line 87: should you assert-check the score?

FANGAreNotGnu: It's not deterministic, especially when we have a time_limit of 30 seconds.

cheungdaven: Since this issue is relevant to the correctness of metric calculation, it would be better if there is some sort of guardrail here.

Innixma: Perhaps get the train predictions via `predictor.predict / predict_proba`, then ensure manual metric calculation matches `evaluate` output?

Innixma: The above is what I do in tabular's tests

multimodal/src/autogluon/multimodal/utils/metric.py
```diff
@@ -234,17 +234,21 @@ def compute_score(
 
     y = metric_data[Y_TRUE]
     if metric.needs_proba or metric.needs_threshold:
+        y_pred_proba = metric_data[Y_PRED_PROB]
+        y_pred_proba = (
+            y_pred_proba if y_pred_proba.shape[1] > 2 else y_pred_proba[:, pos_label]
+        )  # only use pos_label for binary classification
```
Innixma, line 240: What if it is multiclass classification with 2 classes? Is that a valid scenario?

multimodal/src/autogluon/multimodal/utils/metric.py
```diff
@@ -234,17 +234,21 @@ def compute_score(
...
     else:
         y_pred = metric_data[Y_PRED]
 
         # TODO: This is a hack. Doesn't support `f1_macro`, `f1_micro`, `f1_weighted`, or custom `f1` metrics with different names.
         # TODO: Longterm the solution should be to have the input data to this function use the internal representation without the original class names. This way `pos_label` would not need to be specified.
         if metric.name == F1:  # only for binary classification
-            y = (metric_data[Y_TRUE] == pos_label).astype(int)
-            y_pred = (metric_data[Y_PRED] == pos_label).astype(int)
+            y = (y == pos_label).astype(int)
+            y_pred = (y == pos_label).astype(int)
```
Innixma, line 251: Bug? Why `y_pred = (y == pos_label).astype(int)` and not `y_pred = (y_pred == pos_label).astype(int)

multimodal/src/autogluon/multimodal/utils/metric.py
Innixma: Why was this marked as resolved? Looks like a bug that still exists to me

multimodal/tests/unittests/others/test_classification.py
```diff
@@ -57,3 +57,33 @@ def test_focal_loss_multiclass(checkpoint_name):
...
+
+
+@pytest.mark.parametrize(
+    "checkpoint_name,eval_metric",
+    [
+        ("swin_tiny_patch4_window7_224", "log_loss"),
+        ("swin_tiny_patch4_window7_224", "f1_micro"),
+    ],
+)
+def test_metrics_multiclass(checkpoint_name, eval_metric):
```
Innixma, line 69: This doesn't appear to be a particularly useful test at the moment, as it doesn't check anything beyond the logic not crashing. We need to test for correctness.

Innixma, line 69: Add docstring explaining the purpose of the unit test. It is currently unclear why we choose the specific metrics and model, and what exactly we are trying to verify in the test

multimodal/src/autogluon/multimodal/utils/metric.py
FANGAreNotGnu: That should be binary classification. @cheungdaven correct me if it's wrong: For NER, it's rare but possble to have a multiclass setting of 2 classes if customer is using NER_TOKEN_F1, which is a multiclass F1 score, which does not need proba

FANGAreNotGnu: Instead we can change problem type to binary if there's only 2 classes. Let me check if it's already implemented.

multimodal/src/autogluon/multimodal/utils/metric.py
FANGAreNotGnu: fixed the typo

multimodal/tests/unittests/others/test_classification.py
FANGAreNotGnu: Added the docstring

multimodal/tests/unittests/others/test_classification.py
FANGAreNotGnu: Added an assert-check to the scores

Innixma (approved):
LGTM!

PR #4650, opened 2024-11-17
[timeseries] prune timeseries unit and smoke tests

canerturkmen:
*Issue #, if available:*

*Description of changes:*

This PR reduces overall test time by ~30%, primarily by 
- reducing the test case grid in smoke tests
- removing expensive but unused local models from smoke and model tests
- removing some Chronos hyperparameter configs

### DIFF
--- a/timeseries/tests/smoketests/test_all_models.py
+++ b/timeseries/tests/smoketests/test_all_models.py
@@ -55,12 +55,16 @@ def generate_train_and_test_data(
     return train_data, test_data
 
 
-DUMMY_MODEL_HPARAMS = {"max_epochs": 1, "num_batches_per_epoch": 1, "use_fallback_model": False}
+DUMMY_MODEL_HPARAMS = {
+    "max_epochs": 1,
+    "num_batches_per_epoch": 1,
+    "use_fallback_model": False,
+}
 
 ALL_MODELS = {
     "ADIDA": DUMMY_MODEL_HPARAMS,
     "Average": DUMMY_MODEL_HPARAMS,
-    "AutoCES": DUMMY_MODEL_HPARAMS,
+    # "AutoCES": {**DUMMY_MODEL_HPARAMS, "model": "S"},
     "Chronos": {},
     "Croston": DUMMY_MODEL_HPARAMS,
     "DLinear": DUMMY_MODEL_HPARAMS,
@@ -68,7 +72,7 @@ ALL_MODELS = {
     "DirectTabular": DUMMY_MODEL_HPARAMS,
     "DynamicOptimizedTheta": DUMMY_MODEL_HPARAMS,
     "IMAPA": DUMMY_MODEL_HPARAMS,
-    "ETS": DUMMY_MODEL_HPARAMS,
+    "AutoETS": {**DUMMY_MODEL_HPARAMS, "model": "ANN"},
     "NPTS": DUMMY_MODEL_HPARAMS,
     "Naive": DUMMY_MODEL_HPARAMS,
     "PatchTST": DUMMY_MODEL_HPARAMS,
@@ -77,7 +81,7 @@ ALL_MODELS = {
     "SeasonalNaive": DUMMY_MODEL_HPARAMS,
     "SimpleFeedForward": DUMMY_MODEL_HPARAMS,
     "TemporalFusionTransformer": DUMMY_MODEL_HPARAMS,
-    "Theta": DUMMY_MODEL_HPARAMS,
+    # "Theta": DUMMY_MODEL_HPARAMS,  # covered by DynamicOptimizedTheta
     "TiDE": DUMMY_MODEL_HPARAMS,
     "WaveNet": DUMMY_MODEL_HPARAMS,
     "Zero": DUMMY_MODEL_HPARAMS,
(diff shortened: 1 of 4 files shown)


timeseries/tests/smoketests/test_all_models.py
```diff
@@ -55,20 +54,24 @@ def generate_train_and_test_data(
...
+    "max_epochs": 1,
+    "num_batches_per_epoch": 1,
+    "use_fallback_model": False,
+}
 
 ALL_MODELS = {
     "ADIDA": DUMMY_MODEL_HPARAMS,
     "Average": DUMMY_MODEL_HPARAMS,
-    "AutoCES": DUMMY_MODEL_HPARAMS,
+    # "AutoCES": {**DUMMY_MODEL_HPARAMS, "model": "S"},
```
shchur, line 66: Should we remove this line completely instead of commenting out?

timeseries/tests/smoketests/test_all_models.py
```diff
@@ -182,8 +181,43 @@ def test_all_models_can_handle_all_covariates(
...
+        {"DeepAR": DUMMY_MODEL_HPARAMS},
+        {"DirectTabular": DUMMY_MODEL_HPARAMS},
+        {"AutoETS": {**DUMMY_MODEL_HPARAMS, "model": "AAA"}},
+        {"NPTS": DUMMY_MODEL_HPARAMS},
+        {"Naive": DUMMY_MODEL_HPARAMS},
+        {"PatchTST": DUMMY_MODEL_HPARAMS},
+        {"RecursiveTabular": DUMMY_MODEL_HPARAMS},
+        {"SeasonalAverage": DUMMY_MODEL_HPARAMS},
+        {"SeasonalNaive": DUMMY_MODEL_HPARAMS},
+        {"SimpleFeedForward": DUMMY_MODEL_HPARAMS},
```
shchur, line 213: I think we can drop SimpleFeedForward and WaveNet here as well since these model are not used in any presets.

timeseries/tests/smoketests/test_all_models.py
```diff
@@ -182,8 +181,43 @@ def test_all_models_can_handle_all_covariates(
...
+        "QE",
+        "SME",
+        "W",
+        "D",
+        "B",
+        "bh",
+        "h",
+        "min",
+        "s",
+    ],
```
shchur, line 197: Nit: If we are specifying these manually, I would suggest trying some nonstandard combinations like `2W`, `3D` `4h`, `100s`

timeseries/tests/smoketests/test_all_models.py
```diff
@@ -166,7 +165,7 @@ def test_all_models_can_handle_all_covariates(
         target=TARGET_COLUMN,
         prediction_length=prediction_length,
         known_covariates_names=known_covariates_names if len(known_covariates_names) > 0 else None,
-        eval_metric=eval_metric,
+        eval_metric="WQL",
```
shchur, line 168: Can we add a test to `models/test_mlforecast.py` that ensures that `RecursiveTabular` and `DirectTabular` work with all covariate combinations for both point & quantile eval_metrics? These models change their behavior depending on whether point or quantile metric is used, so I think we should have a test that validates their behavior for all covariate + metric combinations.

timeseries/tests/smoketests/test_all_models.py
```diff
@@ -55,20 +54,24 @@ def generate_train_and_test_data(
...
 
 
-DUMMY_MODEL_HPARAMS = {"max_epochs": 1, "num_batches_per_epoch": 1, "use_fallback_model": False}
+DUMMY_MODEL_HPARAMS = {
+    "max_epochs": 1,
+    "num_batches_per_epoch": 1,
+    "use_fallback_model": False,
+}
 
 ALL_MODELS = {
```
shchur, line 63: I wonder whether we can reduce the space even further here. There are many models (e.g., all local models) that don't change their behavior when the covariates are present. For them, it's probably fine to just test "works with no covariates" and "works with all covariates". We can reserve the more expensive tests with all covariate combinations to other models.

No need to address this now, but we can add a TODO in case we want to tackle this later.

timeseries/tests/smoketests/test_all_models.py
shchur: If we are going for the approach "some models are basically exchangeable when it comes to how they handle covariates", we can only keep one representative from each group in these tests. For example, ADIDA, IMAPA, Croston -> only keep Croston.

timeseries/tests/smoketests/test_all_models.py
```diff
@@ -157,7 +156,7 @@ def test_all_models_can_handle_all_covariates(
         use_known_covariates=use_known_covariates,
         use_past_covariates=use_past_covariates,
         use_static_features_continuous=use_static_features_continuous,
-        use_static_features_categorical=use_static_features_categorical,
+        use_static_features_categorical=False,
```
shchur, line 159: I would prefer to set `use_static_features_continuous=False` and keep `use_static_features_categorical` as a test case so that we test at least some categorical features.

timeseries/tests/smoketests/test_all_models.py
canerturkmen: ADIDA, IMAPA, etc. are very fast to run, so will leave them in for now.

timeseries/tests/smoketests/test_all_models.py
canerturkmen: again, will leave them in as they do not hurt greatly in runtime.

timeseries/tests/smoketests/test_all_models.py
canerturkmen: replacing some with these (not adding new) as this leads to a proportionate increase in runtime

shchur (approved):
Nice, thank you for the fix!

PR #4667, opened 2024-11-20
add val operation when skip hpo

echowve:
When skipping hpo, validation data was not used in the call to the model's fit method, which affected the printing during the training process and the evaluation of the trained model. I have corrected this issue.


Innixma:
Thanks for the fix! This was indeed a bug. I've updated the PR to reflect a solution that fixes the problem without causing a bug present in the initial PR which assigns X_val twice as an arg when HPO is performed. Will merge after the CI passes.

Innixma (approved):
LGTM, thanks for the contribution and spotting the bug!

PR #4672, opened 2024-11-21
[timeseries] Change Chronos-Bolt model aliases

shchur:
*Issue #, if available:*

*Description of changes:*
- Currently, there is an inconsistency between the preset names (e.g., `bolt_small`) and model aliases (e.g., `"Chronos": {"model_path": "bolt-small"}`. This PR changes the aliases for bolt models to `bolt_size` instead of `bolt-size` to be consistent with the preset names. 


canerturkmen (approved):
Thanks!

PR #4676, opened 2024-11-22
[core] Bump accelerate version to >=0.34,<1.0

shchur:
*Issue #, if available:*

*Description of changes:*
- Currently, Chronos fine-tuning fails for `accelerate<0.34` with error `'AdamW' has no attribute 'train'` (see https://github.com/huggingface/transformers/issues/33620#issuecomment-2375857846). This error was fixed in `accelerate-0.34.0`, so we need to raise the lower bound.


canerturkmen:
interesting. over to @tonyhoo . another (lesser) solution is to cap `transformers`.

tonyhoo (approved):

PR #4677, opened 2024-11-22
[timeseries] Change verbosity of warnings for Chronos models

shchur:
*Issue #, if available:*

*Description of changes:*
- Clean up the logging & warning outputs for Chronos models at different verbosity levels

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/chronos/model.py
+++ b/timeseries/src/autogluon/timeseries/models/chronos/model.py
@@ -18,7 +18,7 @@ logger = logging.getLogger("autogluon.timeseries.models.chronos")
 
 # TODO: Replace `evaluation_strategy` with `eval_strategy` when upgrading to `transformers>=4.41` + remove warning filter
 warnings.filterwarnings("ignore", category=FutureWarning, message="`evaluation_strategy` is deprecated")
-# # TODO: Remove warning filter when upgrading to `transformers>=4.40`
+# TODO: Remove warning filter when upgrading to `transformers>=4.40`
 warnings.filterwarnings("ignore", category=FutureWarning, message="Passing the following arguments to ")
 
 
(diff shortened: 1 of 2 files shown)


timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -342,6 +348,7 @@ def _get_model_params(self) -> dict:
             logging_dir=str(output_dir),
             logging_strategy="steps",
             logging_steps=100,
+            disable_tqdm=True,
```
abdulfatir, line 351: Why is tqdm always disabled?

timeseries/src/autogluon/timeseries/models/chronos/pipeline/utils.py
```diff
@@ -317,7 +317,7 @@ def on_train_begin(self, args, state, control, **kwargs):
     def on_step_end(self, args, state, control, **kwargs):
         elapsed_time = time.monotonic() - self.start_time
         if elapsed_time > self.time_limit:
-            logger.info("\tStopping fine-tuning since time_limit is reached")
+            logger.log(15, "Stopping fine-tuning since time_limit is reached")
```
abdulfatir, line 320: Why not `info`?

shchur: Info (default) usually should only report the val_loss, fit time and predict time. Information about early stopping seems much more detailed, so we reserve it for verbosity=3

shchur: Also, stopping training when the time_limit is reached is the default behavior for most models (e.g., all GluonTS models), and we don't usually log that.

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -15,6 +16,11 @@
 
 logger = logging.getLogger("autogluon.timeseries.models.chronos")
 
+# TODO: Replace `evaluation_strategy` with `eval_strategy` when upgrading to `transformers>=4.41` + remove warning filter
+warnings.filterwarnings("ignore", category=FutureWarning, message="`evaluation_strategy` is deprecated")
+# # TODO: Remove warning filter when upgrading to `transformers>=4.40`
```
abdulfatir, line 21:
```suggestion
# TODO: Remove warning filter when upgrading to `transformers>=4.40`
```

abdulfatir (approved):
Left a small suggestion.

PR #4680, opened 2024-11-23
[WIP] [Tabular] Support Refit for fit_strategy="parallel"

LennartPurucker:
We require the memory estimate of the child models to run refit in parallel with the current logic. To enable this, this PR caches the memory requirement of the base models and uses it for refit. Then refit works again.

This feels a bit hacky but could be a good solution if we adjust the mem estimate for the training data set size change from fit to refit. 

(Also, there are some formatting changes that ruff did by default, so there are more changes than actually content) 

---

### DIFF
--- a/core/src/autogluon/core/trainer/abstract_trainer.py
+++ b/core/src/autogluon/core/trainer/abstract_trainer.py
@@ -1495,7 +1495,7 @@ class AbstractTrainer:
             # FIXME: Need a common utility class for initializing ray so we don't duplicate code
             if not ray.is_initialized():
                 ray.init(log_to_driver=False, logging_level=logging.ERROR)
-
+            assert self._refit_mem_cache is not None, "Memory cache must be available for parallel refit. This should not happen, please open an issue!"
             distributed_manager = ParallelFitManager(
                 mode="refit",
                 func=_remote_refit_single_full,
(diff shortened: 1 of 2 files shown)


core/src/autogluon/core/trainer/abstract_trainer.py
```diff
@@ -3102,6 +3107,9 @@ def _train_multi_fold_parallel(
                 # Re-schedule jobs
                 unfinished_job_refs += distributed_manager.schedule_jobs()
 
+        # FIXME: this might be slightly off due to less training data used for the bagging fit (depending on the validation method).
+        self._refit_mem_cache = distributed_manager.refit_model_child_mem_estimate_cache.copy()
```
Innixma, line 3111: I don't think this will work properly for multi-layer stacking, because you will overwrite the previous layer's mem cache.

Innixma: Should also be probably using the DAG to store this info

LennartPurucker: Ture, will need to fix this (by extending the dict). 

Yes, we can get it as a return value as well. Looking into it.

Innixma:
Closing in favor of #5850

PR #4730, opened 2024-12-11
[timeseries] Do not try to compute OOF evaluations for "_FULL" models

nathanaelbosch:
*Description of changes:*
When setting `refit_full=True`, compatible models are re-trained on the full dataset, and the resulting models are saved with a `"_FULL"` suffix in the list of base models. This creates an issue in `predictor._simulation_artifact`, where we simply retrieve the already-computed OOF evaluations of each available model from disk, to be then stored in the `simulation_artifact` dict. These are not available for the `"_FULL"` models, and thus currently the code fails with an error of the form
```
FileNotFoundError: [Errno 2] No such file or directory: '/local/home/user/autogluon/AutogluonModels/ag-20241211_105202/models/DeepAR/DeepAR_FULL/utils/oof.pkl'
```
This PR fixes this issue by skipping any `"_FULL"` model in this step.

### DIFF
--- a/timeseries/src/autogluon/timeseries/predictor.py
+++ b/timeseries/src/autogluon/timeseries/predictor.py
@@ -1429,17 +1429,17 @@ class TimeSeriesPredictor(TimeSeriesPredictorDeprecatedMixin):
         trainer = self._trainer
         train_data = trainer.load_train_data()
         val_data = trainer.load_val_data()
-        base_models = trainer.get_model_names(level=0)
+        base_model_names = trainer.get_model_names(level=0)
         pred_proba_dict_val: Dict[str, List[TimeSeriesDataFrame]] = {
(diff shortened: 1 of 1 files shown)


timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -1432,6 +1432,7 @@ def select_target(ts_df: TimeSeriesDataFrame) -> TimeSeriesDataFrame:
         base_models = trainer.get_model_names(level=0)
         pred_proba_dict_val: Dict[str, List[TimeSeriesDataFrame]] = {
             model: trainer._get_model_oof_predictions(model) for model in base_models
+            if "_FULL" not in model
         }
```
shchur, line 1436: Maybe let's also rename the variables `base_models` -> `base_model_names` and `model` -> `model_name` to make it clearer what the variables mean?

PR #4732, opened 2024-12-11
Fix DeBERTa FP16 overflow in gradient checkpointing test

abhishek-iitmadras:
*Description of changes:*
Change DeBERTa precision to 32-true in tests

**Fixed RuntimeError**: "_value cannot be converted to type at::Half without overflow_" in test_predictor_gradient_checkpointing when using DeBERTa with FP16 mixed precision.
Changed precision from "16-mixed" to "32-true" for DeBERTa tests to avoid attention mask value overflow.

**Note :** The error occurred when setting masked attention scores to minimum FP16 value, which is too small to be represented in half precision. Using 32-true provides the necessary numeric range for attention masking operations.

cc @Innixma @shchur


abhishek-iitmadras:
Need to add label:"module: multimodal" and label:run-multi-gpu

cc @Innixma @shchur

abhishek-iitmadras:
The failing test on test_multimodal_others_2 handled well in https://github.com/autogluon/autogluon/pull/4731 i.e
_`FAILED tests/unittests/others_2/test_ner_chinese.py::test_ner_chinese - Unico...`_

cc @Innixma @shchur

abhishek-iitmadras:
Closed this PR as added this commit to https://github.com/autogluon/autogluon/pull/4731 for easily review and CI will passed in one place

PR #4734, opened 2024-12-11
Fix setup_outputdir

suzhoum:
*Issue #, if available:*
Fixes https://github.com/autogluon/autogluon/issues/4733

*Description of changes:*

### DIFF
--- a/common/src/autogluon/common/utils/utils.py
+++ b/common/src/autogluon/common/utils/utils.py
@@ -62,7 +62,6 @@ def setup_outputdir(path, warn_if_exist=True, create_dir=True, path_suffix=None)
                 f'Warning: path already exists! This predictor may overwrite an existing predictor! path="{path}"'
             )
     path = os.path.expanduser(path)  # replace ~ with absolute path if it exists
-    path = os.path.abspath(path)
     return path
 
 
--- a/common/src/autogluon/common/utils/utils.py
+++ b/common/src/autogluon/common/utils/utils.py
@@ -2,7 +2,7 @@ import logging
 import os
 import platform
 import sys
-from datetime import datetime
+from datetime import datetime, timezone
 from hashlib import md5
 from pathlib import Path
 from typing import Any, Dict, Optional
@@ -23,23 +23,28 @@ LITE_MODE: bool = __lite__ is not None and __lite__
 
 
 def setup_outputdir(path, warn_if_exist=True, create_dir=True, path_suffix=None):
+    is_s3_path = False
     if path:
-        assert isinstance(path, (str, Path)), (
-            f"Only str and pathlib.Path types are supported for path, got {path} of type {type(path)}."
-        )
+        assert isinstance(
+            path, (str, Path)
+        ), f"Only str and pathlib.Path types are supported for path, got {path} of type {type(path)}."
+        
+        is_s3_path = str(path).lower().startswith("s3://")        
+    
     if path_suffix is None:
         path_suffix = ""
-    if path_suffix and path_suffix[-1] == os.path.sep:
+    if path_suffix and path_suffix[-1] == os.path.sep if not is_s3_path else "/":
         path_suffix = path_suffix[:-1]
+    
     if path is not None:
         path = f"{path}{path_suffix}"
-    if path is None:
-        utcnow = datetime.utcnow()
+    else:
+        utcnow = datetime.now(timezone.utc)
         timestamp = utcnow.strftime("%Y%m%d_%H%M%S")
         path = os.path.join("AutogluonModels", f"ag-{timestamp}{path_suffix}")
         for i in range(1, 1000):
             try:
-                if create_dir:
+                if create_dir and not is_s3_path:
                     os.makedirs(path, exist_ok=False)
                     break
                 else:
@@ -51,7 +56,8 @@ def setup_outputdir(path, warn_if_exist=True, create_dir=True, path_suffix=None)
         else:
             raise RuntimeError("more than 1000 jobs launched in the same second")
         logger.log(25, f'No path specified. Models will be saved in: "{path}"')
-    elif warn_if_exist:
+    
+    if warn_if_exist and not is_s3_path:
         try:
             if create_dir:
                 os.makedirs(path, exist_ok=False)
@@ -61,7 +67,7 @@ def setup_outputdir(path, warn_if_exist=True, create_dir=True, path_suffix=None)
             logger.warning(
                 f'Warning: path already exists! This predictor may overwrite an existing predictor! path="{path}"'
             )
-    if not path.lower().startswith("s3://"):
+    if not is_s3_path:
         path = os.path.expanduser(path)  # replace ~ with absolute path if it exists
         path = os.path.abspath(path)
     return path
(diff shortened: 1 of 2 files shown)


common/src/autogluon/common/utils/utils.py
```diff
@@ -62,7 +62,6 @@ def setup_outputdir(path, warn_if_exist=True, create_dir=True, path_suffix=None)
                 f'Warning: path already exists! This predictor may overwrite an existing predictor! path="{path}"'
             )
     path = os.path.expanduser(path)  # replace ~ with absolute path if it exists
-    path = os.path.abspath(path)
```
shchur, line 65: Maybe we can wrap this block into
```python
if not path.startswith("s3://"):
    path = os.path.expanduser(path)  # replace ~ with absolute path if it exists
    path = os.path.abspath(path)
```
I believe there are some edge cases against which the current `path = os.path.abspath(path)` command protects (though it does introduce another bug with loading from S3).

LennartPurucker: Yes, this looks like the best solution.

LennartPurucker:
Ah, my bad! I did not think of s3 when adding this line. 

It fixes a bug in parallel and distributed fitting where the workers/processes use a relative path but should use the absolute path in the (shared) file system.

common/src/autogluon/common/utils/utils.py
LennartPurucker: What happens if one calls `.expanduser()` on an S3 path? I think likely nothing, as S3 paths do not contain/start with `~`. 

Otherwise, the current suggested if-case would change this behavior, too.

suzhoum: In this case I think what @shchur proposed should accommodate S3 paths. I will update the PR.

shchur (approved):

LennartPurucker (approved):
LGTM

common/src/autogluon/common/utils/utils.py
```diff
@@ -61,8 +61,9 @@ def setup_outputdir(path, warn_if_exist=True, create_dir=True, path_suffix=None)
             logger.warning(
                 f'Warning: path already exists! This predictor may overwrite an existing predictor! path="{path}"'
             )
-    path = os.path.expanduser(path)  # replace ~ with absolute path if it exists
-    path = os.path.abspath(path)
+    if not path.lower().startswith("s3://"):
```
Innixma, line 64: If we plan to support s3, then we should also edit the code above this when `warn_if_exist` to not call `os.makedirs` when the path is an s3 path.

Innixma: Also, if we plan to add support for s3, lets add a short docstring that mentions this

Innixma: Several things need to be done:

1. If s3, then `os.path.sep` needs to instead be `/` when updating path_suffix.
2. bool flag near start of function to indicate if s3 path. If flag set, skip `elif warn_if_exist`.
3. Why `path.lower()`? This doesn't seem valid.

common/src/autogluon/common/utils/utils.py
```diff
@@ -61,8 +61,9 @@ def setup_outputdir(path, warn_if_exist=True, create_dir=True, path_suffix=None)
...
+        path = os.path.expanduser(path)  # replace ~ with absolute path if it exists
```
Innixma, line 65: Add unit test for s3 path input, since it shouldn't make any folder.

Innixma (changes requested):

common/src/autogluon/common/utils/utils.py
suzhoum: Added unit test.

common/src/autogluon/common/utils/utils.py
suzhoum: Refactored the function to address comments.

common/src/autogluon/common/utils/utils.py
```diff
@@ -23,23 +23,28 @@
...
+    is_s3_path = False
     if path:
-        assert isinstance(
-            path, (str, Path)
-        ), f"Only str and pathlib.Path types are supported for path, got {path} of type {type(path)}."
+        assert isinstance(path, (str, Path)), (
+            f"Only str and pathlib.Path types are supported for path, got {path} of type {type(path)}."
+        )
+
+        is_s3_path = str(path).lower().startswith("s3://")
```
Innixma, line 32: why `.lower()`?

common/src/autogluon/common/utils/utils.py
```diff
@@ -23,23 +23,28 @@
...
-    if path is None:
-        utcnow = datetime.utcnow()
+    else:
+        utcnow = datetime.now(timezone.utc)
         timestamp = utcnow.strftime("%Y%m%d_%H%M%S")
         path = os.path.join("AutogluonModels", f"ag-{timestamp}{path_suffix}")
         for i in range(1, 1000):
             try:
-                if create_dir:
+                if create_dir and not is_s3_path:
```
Innixma, line 47: No need to check `is_s3_path` when path is None.

common/tests/unittests/test_setup_outputdir.py
```diff
@@ -28,6 +29,13 @@ def test(self):
         returned_path = setup_outputdir(path, warn_if_exist=True, create_dir=False, path_suffix=None)
         assert str(Path(returned_path)) == str(path)
 
+        # checks that setup_outputdir handles S3 paths correctly
+        with patch("os.makedirs") as mock_makedirs:  # Mock os.makedirs to ensure no local directory is created
+            path = "s3://test-bucket/test-folder"
+            returned_path = setup_outputdir(path, warn_if_exist=True, create_dir=False, path_suffix=None)
```
Innixma, line 35: `create_dir` must be `True` to properly test this.

common/src/autogluon/common/utils/utils.py
```diff
@@ -23,23 +23,28 @@
...
+        assert isinstance(path, (str, Path)), (
+            f"Only str and pathlib.Path types are supported for path, got {path} of type {type(path)}."
+        )
+
+        is_s3_path = str(path).lower().startswith("s3://")
+
     if path_suffix is None:
         path_suffix = ""
-    if path_suffix and path_suffix[-1] == os.path.sep:
+    if path_suffix and path_suffix[-1] == os.path.sep if not is_s3_path else "/":
```
Innixma, line 36: would be good to add a unit test for this, it is relevant for Windows OS

common/src/autogluon/common/utils/utils.py
suzhoum: This is just to incorporate cases when people use capitalized `S3` in the path.

common/src/autogluon/common/utils/utils.py
suzhoum: Added more unit tests for this.

common/tests/unittests/test_setup_outputdir.py
suzhoum: Fixed.

Innixma (approved):
LGTM, thanks for the fix!

PR #4168, opened 2024-05-03
[WIP][AutoMM] CLIP Model for late fusion on multimodal tasks.

Harry-zzh:
*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #4737, opened 2024-12-13
[AutoMM] Bag of Tricks

zhiqiangdon:
*Issue #, if available:*

*Description of changes:*

This PR implemented a bag of tricks for multimodal AutoML involving multimodal model fusion strategies, multimodal data augmentation, cross-modal alignment, converting tabular data into text, handling missing modalities with modality dropout and learnable embeddings, and an ensemble learner to integrate the bag of tricks for optimal performance. To better implement these tricks, we also refactored AutoMM's codebase accordingly. For technical details, see [this doc](https://quip-amazon.com/LPtpAL5nbUWG/Bag-of-Tricks-for-Multimodal-AutoML-with-Image-Text-and-Tabular-Data)

Here is one example of using bag of tricks:
```
predictor = MultiModalPredictor(label="label", use_ensemble=True)
predictor.fit(train_data=train_data)
```

For users who used AutoMM previously, please change your configurations as bellow.  No action is needed if you didn’t customize these hyperparameters.

```
optimization -> optim
optimization.learning_rate -> optim.lr
optimization.efficient_finetune -> optim.peft
optimization.loss_function -> optim.loss_func
env.num_workers_evaluation -> env.num_workers_inference
env.eval_batch_size_ratio -> env.inference_batch_size_ratio
data.label.numerical_label_preprocessing -> data.label.numerical_preprocessing
model.categorical_mlp.drop_rate -> model.categorical_mlp.dropout
model.numerical_mlp.drop_rate -> model.numerical_mlp.dropout
model.numerical_mlp.d_token -> model.numerical_mlp.token_dim
model.timm_image.max_img_num_per_col -> model.timm_image.max_image_num_per_column
model.clip.max_img_num_per_col -> model.clip.max_image_num_per_column
model.fusion_mlp.weight -> model.fusion_mlp.aux_loss_weight
model.fusion_mlp.drop_rate -> model.fusion_mlp.dropout
model.fusion_transformer.n_blocks -> model.fusion_transformer.num_blocks
model.fusion_transformer.attention_n_heads -> model.fusion_transformer.attention_num_heads
model.fusion_transformer.ffn_d_hidden -> model.fusion_transformer.ffn_hidden_size
model.ft_transformer.attention_n_heads -> model.ft_transformer.attention_num_heads

```


zhiqiangdon:
@Innixma @tonyhoo @FANGAreNotGnu @suzhoum Please help review and approve. Please also help add the necessary labels to run multi-gpu jobs since my developer permission was revoked by someone.

George suggested adding the bag of tricks into AutoMM. The corresponding paper describing the details will be published soon. This PR is large due to the backlog.

zhiqiangdon:
> Great work! Just really a curious question (not as a blocker in anyway but just a question): is it really needed to change the parameter names ("optimization -> optim", "learning rate -> lr" etc)? As this will make the updated AutoMM library not backward compatible.

Great observation and question! To better implement the bag of tricks, we have refactored several places in the data preprocessor and processors, which are not backward compatible. Therefore, we also took this opportunity to pay tech debts and optimize some designs including the config name changes. As the optimization configs have multi-level hyperparameters such as `optim.lemda.consist_threshold`, shortening some key word lengths would make users' life easier. Note that in AutoGluon's [previous release notes](https://github.com/autogluon/autogluon/releases) (< 1.2), it usually mentioned "Loading models trained on older versions of AutoGluon is not supported." Moreover, the `optim` acronym was already adopted by Pytorch (https://pytorch.org/docs/stable/optim.html).

Innixma:
Thanks @zhiqiangdon! Taking a look at this now as I was at NeurIPS last week.

Do you have a recommendation for how we should benchmark this PR? Will the existing `best_quality` preset incorporate the changes automatically, or do you have a recommended configuration we should try?

Innixma:
Given the changes / additions to the hyperparameters / params that a user may specify, it would be good if we can write a migration guide so users can know how to modify existing scripts to be compatible with this version of the code. For example, mentioning to change `optimization -> optim`, etc.

multimodal/src/autogluon/multimodal/learners/ensemble.py
```diff
@@ -0,0 +1,719 @@
...
+            predictions
+        ), f"len(learners) {len(learners)} doesn't match len(predictions) {len(predictions)}"
+
+    def fit_per_ensemble(
+        self,
+        predictions: List[np.ndarray],
+        labels: np.ndarray,
+    ):
+        weighted_ensemble = EnsembleSelection(
+            ensemble_size=self._ensemble_size * len(predictions),
```
Innixma, line 230: Why `* len(predictions)`? This probably isn't necessary and would slow down the training a lot. I'd recommend just always using a smallish value such as `ensemble_size=40`.

multimodal/src/autogluon/multimodal/learners/ensemble.py
```diff
@@ -0,0 +1,719 @@
...
+            Whether to raise warning if the specified path already exists.
+        enable_progress_bar
+            Whether to show progress bar. It will be True by default and will also be
+            disabled if the environment variable os.environ["AUTOMM_DISABLE_PROGRESS_BAR"] is set.
+        ensemble_size
+            A multiple of number of models in the ensembling pool (Default 2). The actual ensemble size = ensemble_size * the model number
+        ensemble_mode
+            The mode of conducting ensembling:
+            - `one_shot`: the classic ensemble selection
+            - `sequential`: iteratively calling the classic ensemble selection with each time growing the model zoo by the best next model.
```
Innixma, line 102: `sequential` is a cool idea! It might make the most sense for this to be added in `core` so it isn't exclusive to AutoMM? That way we could simply call `EnsembleSelection(mode="sequential")` to do it in Tabular and TimeSeries as well.

Innixma: Note: can be a follow-up PR

zhiqiangdon:
> Given the changes / additions to the hyperparameters / params that a user may specify, it would be good if we can write a migration guide so users can know how to modify existing scripts to be compatible with this version of the code. For example, mentioning to change `optimization -> optim`, etc.

Good point. I have updated the PR description by adding the migration guideline.

multimodal/src/autogluon/multimodal/learners/ensemble.py
zhiqiangdon: len(predictions) is the number of models, not the number of samples. For example, if we train 10 models, then it would be 10. `ensemble_size` is a multiplier on this number, making the actual ensemble size adaptable to the number of models.

multimodal/src/autogluon/multimodal/learners/ensemble.py
zhiqiangdon: It's not exclusive to AutoMM. We can integrate it into EnsembleSelection later if needed.

multimodal/src/autogluon/multimodal/learners/ensemble.py
Innixma: Yeah I was aware of len(predictions) being the # of models, but I'll mention that in my testing with TabRepo, there is virtually zero benefit going beyond `ensemble_size=40`. In fact, going beyond this value is harmful due to validation overfitting on small datasets, and you are more likely to lose the sparseness property as you have very high ensemble_size, which will slow down your inference speed.

Refer to fig 4 here: https://github.com/autogluon/tabrepo/blob/main/data/sensitivity.png

Innixma: Similarly, it isn't necessary to use very small values. If I'm not mistaken, with 2 models the code currently only uses ensemble_size=4. This will be suboptimal, as 4 isn't enough to get good weights between the 2 models. I'd recommend always using between 25 and 40, but selecting a single value for simplicity. Tabular uses 40.

CI/bench/multimodal/custom_user_dir/multimodal_frameworks_template.yaml
```diff
@@ -4,4 +4,4 @@ AutoGluon_multimodal_best:
   params:  # MultimodalPredictor.fit(params) # can add the actual job time limit here
     presets: best_quality
     hyperparameters:
-      optimization.max_epochs: 10 # 10 is default
+      optim.max_epochs: 10 # 10 is default
```
Innixma, line 7: General question: minor bug fixes / updates have been made over the last several months in AutoMM mainline. Does this PR incorporate those changes as well (via rebasing during the period of development), or does the PR overwrite them?

If it rebased on them, were there any major merge conflicts resolved that are part of this PR worth mentioning?

zhiqiangdon:
> Thanks @zhiqiangdon! Taking a look at this now as I was at NeurIPS last week.
> 
> Do you have a recommendation for how we should benchmark this PR? Will the existing `best_quality` preset incorporate the changes automatically, or do you have a recommended configuration we should try?
> 
> This PR also adds quite a few changes and thus would be challenging to review (>10k LoC changed across 188 files). Do you have a recommendation for how we should approach this PR for review?

Good questions. Here's some context and recommendations:

Regarding benchmarking, the best_quality preset currently trains only a single model. To incorporate the bag of tricks, you’ll need to set use_ensemble=True when initializing the predictor. At this point, autogluon-bench does not support benchmarking the bag of tricks due to the additional setup required for datasets and certain model checkpoints (e.g., meta-transformer). For now, I suggest referring to the results documented[ here](https://quip-amazon.com/LPtpAL5nbUWG/Bag-of-Tricks-for-Multimodal-AutoML-with-Image-Text-and-Tabular-Data).

This PR is large because of management issues earlier this year that led to an unexpected backlog. All unit tests and tutorial builds have passed successfully. To proceed, you can benchmark the three existing presets (single model without bag of tricks) and compare the results with previous releases. Many parts of this PR involving configuration changes should be straightforward to skim through.

CI/bench/multimodal/custom_user_dir/multimodal_frameworks_template.yaml
zhiqiangdon: This PR is based on the most recent AutoMM mainline. I didn't notice other major conflicts except for the configuration changes.

multimodal/src/autogluon/multimodal/learners/ensemble.py
zhiqiangdon: Great suggestion! We used `2* model_num` in previous experiments. We can try 40 later to see whether it can improve performance.

Innixma:
While referring to the paper results is good, it is still results from a snapshot in time that doesn't necessarily represent the state of the system in practice today. The current idea I have is that once a few more people are back from vacation in mid Jan, we can run a benchmark to verify stability and performance with the presets.

If it looks good and there aren't any other major concerns that come up, we can look to merge. This could take some time though, as the changes are significant.

Innixma:
@zhiqiangdon How do the following two PRs relate to this PR?

https://github.com/autogluon/autogluon/pull/4167
https://github.com/autogluon/autogluon/pull/4168

They are currently still open.
Does this PR include those changes?
What is your recommended resolution?

zhiqiangdon:
> @zhiqiangdon How do the following two PRs relate to this PR?
> 
> #4167 #4168
> 
> They are currently still open. Does this PR include those changes? What is your recommended resolution?

@Innixma The functionalities of the 2 PRs are included in this PR. We can close them.

Innixma:
@zhiqiangdon Sounds good, I closed the other 2 PRs.

zhiqiangdon:
> @zhiqiangdon To move forward, we will need to run a reasonable benchmark to compare this with current mainline. I would prefer to avoid the alternative, which is breaking the PR down into components, as I believe this would be too much effort and we would in the end need to benchmark the code anyways.
> 
> You mentioned that this PR is not compatible with our prior benchmarking tooling. In order for us to move forward, we will need to make it work so we can test it. I don't know the exact benchmarks we used for evaluating AutoMM in past releases. Was it AG-Bench? If so, please work with @suzhoum on how to configure the changes in this PR to run the tests.

@Innixma Thank you for your engagement in this PR. Feel free to benchmark this PR on existing datasets and compare results with past releases. I don't know how AutoMM was tested in the recent releases (>1.1). Maybe @suzhoum @FANGAreNotGnu @tonyhoo have more context and can help.

suzhoum:
> > @zhiqiangdon To move forward, we will need to run a reasonable benchmark to compare this with current mainline. I would prefer to avoid the alternative, which is breaking the PR down into components, as I believe this would be too much effort and we would in the end need to benchmark the code anyways.
> > You mentioned that this PR is not compatible with our prior benchmarking tooling. In order for us to move forward, we will need to make it work so we can test it. I don't know the exact benchmarks we used for evaluating AutoMM in past releases. Was it AG-Bench? If so, please work with @suzhoum on how to configure the changes in this PR to run the tests.
> 
> @Innixma Thank you for your engagement in this PR. Feel free to benchmark this PR on existing datasets and compare results with past releases. I don't know how AutoMM was tested in the recent releases (>1.1). Maybe @suzhoum @FANGAreNotGnu @tonyhoo have more context and can help.

I have synced with @zhiqiangdon and will be helping to run the tests on AG-Bench.

Innixma:
Sounds reasonable.

@zhiqiangdon, there is a minor merge conflict as we fixed some bugs in the CI related to two unit tests. You can see them in the conflicting files for this PR. If you can rebase to address these merge conflicts, this will allow the CI to run so that the tests pass. That way we can benchmark on a version of the code that is directly merge-able without further edits.

Innixma:
Also, perhaps as a follow-up (but necessary) PR: It may be a good idea to ensure backwards compatibility by supporting the old hyperparameter names, internally converting them to the new hyperparameter names. You can have a deprecation warning for users who use the old names in the 1.3 release, and we can turn it into an exception in the 1.4 release.

One way to do this is to scan through user-provided hyperparameters for the changed keys at the very start of the predictor's fit call, and log a warning for each matching key, along with converting it in-place to the new key.

Since the number of changed values is around 20, it should be sufficient to have a mapping dict manually created for this purpose.

suzhoum:
We have benchmarked both the current master and this PR with 3 presets (`best_quality`, `high_quality`, `medium_quality`) on 5 image-tabular, image-text, image-text-tabular datasets, 5 text-tabular datasets, and 6 image only datasets. 

These are the raw benchmark results:
[results_automlbenchmark_None_ag_bench_bot.csv](https://github.com/user-attachments/files/18681334/results_automlbenchmark_None_ag_bench_bot.csv)
where `famework` with prefix `AutoGluon_bot_` denotes the code branch from this PR, and prefix `AutoGluon_` denotes the code branch from master.

These are the evaluation results:
[evaluation_results.zip](https://github.com/user-attachments/files/18681375/evaluation_results.zip)
Bag-of-tricks branch performance looks to be comparable with master with single model (`use_ensemble=False`)

zhiqiangdon:
> We have benchmarked both the current master and this PR with 3 presets (`best_quality`, `high_quality`, `medium_quality`) on 5 image-tabular, image-text, image-text-tabular datasets, 5 text-tabular datasets, and 6 image only datasets.
> 
> These are the raw benchmark results: [results_automlbenchmark_None_ag_bench_bot.csv](https://github.com/user-attachments/files/18681334/results_automlbenchmark_None_ag_bench_bot.csv) where `famework` with prefix `AutoGluon_bot_` denotes the code branch from this PR, and prefix `AutoGluon_` denotes the code branch from master.
> 
> These are the evaluation results: [evaluation_results.zip](https://github.com/user-attachments/files/18681375/evaluation_results.zip) Bag-of-tricks branch performance looks to be comparable with master with single model (`use_ensemble=False`)

Thank @suzhoum for benchmarking this PR.

zhiqiangdon:
> Sounds reasonable.
> 
> @zhiqiangdon, there is a minor merge conflict as we fixed some bugs in the CI related to two unit tests. You can see them in the conflicting files for this PR. If you can rebase to address these merge conflicts, this will allow the CI to run so that the tests pass. That way we can benchmark on a version of the code that is directly merge-able without further edits.

@Innixma I have resolved the conflicts with the master branch. Also, Su finished the benchmarking showing similar results when using the same configurations as before.

Innixma:
@suzhoum @zhiqiangdon Looking at the eval results, the medium quality comparison is incorrect.

It is comparing mainline medium vs PR high. It should be mainline medium vs PR medium.

Innixma:
@zhiqiangdon Can we run a configuration you would expect outperforms the current mainline AutoMM based on the findings in the paper?

Currently all of the results point to similar performance, but I don't see a performance improvement. Was the goal of these runs to show matching performance by performing the same logic, or did the runs for this PR include the bag-of-tricks enhancements in the paper?

suzhoum:
> @suzhoum @zhiqiangdon Looking at the eval results, the medium quality comparison is incorrect.
> 
> It is comparing mainline medium vs PR high. It should be mainline medium vs PR medium.
> 
> ```
> framework	                                Winrate	>	<	=	% Loss Reduction	% Loss Reduction (median)	Avg Fit Speed Diff	Avg Inf Speed Diff	time_train_s	metric_error	time_infer_s	bestdiff
> AutoGluon_bot_high_master_g4_12x_ag_bench_bot	0.1875	3	13	0	-26.57424979	-17.07415627	-1.650422655	-0.168812526	1926.6	35.82963563	54.7875	0.003491426
> AutoGluon_medium_master_g4_12x_ag_bench_bot	0.5	0	0	16	0	0	0	0	713.3	36.08120188	46.06875	0.269233924
> ```
> 
> Can we get the medium vs medium results?

Sorry for the confusion. Here is the updated comparison:

framework,Winrate,>,<,=,% Loss Reduction,% Loss Reduction (median),Avg Fit Speed Diff,Avg Inf Speed Diff,time_train_s,metric_error,time_infer_s,bestdiff,loss_rescaled,time_train_s_rescaled,time_infer_s_rescaled,rank,rank=1_count,rank=2_count,rank=3_count,rank>3_count,error_count
AutoGluon_medium_master_g4_12x_ag_bench_bot,0.5,0,0,16,0.0,0.0,0.0,0.0,713.3,36.081201875000005,46.06875,0.00927391575355948,0.0625,1.0470561660139157,1.003246794748479,1.25,9,7,0,0,0
AutoGluon_bot_medium_master_g4_12x_ag_bench_bot,0.75,9,1,6,1.2011959568014077,0.5747288361275027,0.014269784414941894,-0.0003807718518358677,710.325,36.458584375,46.06875,0.021285875321573555,0.5625,1.0327863815989737,1.003627566600315,1.75,1,15,0,0,0

Innixma:
@suzhoum this is hard for me to parse, can you upload a file instead?

suzhoum:
> @suzhoum this is hard for me to parse, can you upload a file instead?

Updated above.

zhiqiangdon:
> @zhiqiangdon Can we run a configuration you would expect outperforms the current mainline AutoMM based on the findings in the paper?
> 
> Currently all of the results point to similar performance, but I don't see a performance improvement. Was the goal of these runs to show matching performance by performing the same logic, or did the runs for this PR include the bag-of-tricks enhancements in the paper?

@Innixma Su's benchmarking is to validate the changes in this PR do not affect previous performance, addressing concerns of this big PR. Regarding the performance improvement, we can refer to the doc results for now. Those benchmark datasets haven't been integrated into ag-bench yet. 

Currently, `use_ensemble=True` only works for multimodal datasets. The logic of doing ensembling for unimodal tasks would be different and needs more research. Of course, we can benchmark `use_ensemble=True` on the current supported multimodal datasets in ag-bench, but we need to download the meta-transformer to local and specify the local checkpoint path as follows:

```
hyperparameters = {
        "early_fusion": {
            "model.meta_transformer.checkpoint_path": args.meta_transformer_ckpt_path,
        }
    }
```

suzhoum:
> > @zhiqiangdon Can we run a configuration you would expect outperforms the current mainline AutoMM based on the findings in the paper?
> > Currently all of the results point to similar performance, but I don't see a performance improvement. Was the goal of these runs to show matching performance by performing the same logic, or did the runs for this PR include the bag-of-tricks enhancements in the paper?
> 
> @Innixma Su's benchmarking is to validate the changes in this PR do not affect previous performance, addressing concerns of this big PR. Regarding the performance improvement, we can refer to the doc results for now. Those benchmark datasets haven't been integrated into ag-bench yet.
> 
> Currently, `use_ensemble=True` only works for multimodal datasets. The logic of doing ensembling for unimodal tasks would be different and need more research. Of course, we can benchmark `use_ensemble=True` on the current supported multimodal datasets in ag-bench, but we need to download the [meta-transformer](https://github.com/invictus717/MetaTransformer) to local and specify the local checkpoint path as follows:
> 
> ```
> hyperparameters = {
>         "early_fusion": {
>             "model.meta_transformer.checkpoint_path": args.meta_transformer_ckpt_path,
>         }
>     }
> ```

I have ran benchmark with `use_ensemble=True` (it errored out due to the missing checkpoint), but upon checking with @zhiqiangdon , he stated that only single model mode is required for benchmark. 

Quoted from an earlier reply:
"This PR is large because of management issues earlier this year that led to an unexpected backlog. All unit tests and tutorial builds have passed successfully. To proceed, you can benchmark the three existing presets (single model without bag of tricks) and compare the results with previous releases. Many parts of this PR involving configuration changes should be straightforward to skim through."

I can try getting the checkpoint saved in the benchmarking instances and have another round of benchmark with `use_ensemble=True` if that's necessary.

multimodal/src/autogluon/multimodal/learners/matching.py
```diff
@@ -2013,52 +2011,37 @@ def _load_metadata(
         with open(os.path.join(path, "assets.json"), "r") as fp:
             assets = json.load(fp)
 
-        query_config = upgrade_config(query_config, assets["version"])
-        response_config = upgrade_config(response_config, assets["version"])
-
         with open(os.path.join(path, "df_preprocessor.pkl"), "rb") as fp:
-            df_preprocessor = CustomUnpickler(fp).load()
+            df_preprocessor = pickle.load(fp)  # nosec B301
```
Innixma, line 2015: What is CustomUnpickler and why did we stop using it in many areas of the code?

Innixma:
@zhiqiangdon I am aligned that the results show that performance is largely the same between mainline and this PR. If the PR was purely code cleanup / renaming then at this point I would be inclined to merge it.

> This PR implemented a bag of tricks for multimodal AutoML involving multimodal model fusion strategies, multimodal data augmentation, cross-modal alignment, converting tabular data into text, handling missing modalities with modality dropout and learnable embeddings, and an ensemble learner to integrate the bag of tricks for optimal performance

If we merged the PR as-is, then we would be merging it without testing any of these added features. I understand that the paper exists, but we need a way to replicate the results so that in future when we make changes we can ensure we are going in the right direction and aren't making performance regressions. At the very least one single configuration that shows benefit.

@zhiqiangdon Regarding meta-transformer, is there a reason `use_ensemble` doesn't work out of the box? Why don't we just download from HuggingFace when the user specifies `use_ensemble`?

multimodal/src/autogluon/multimodal/learners/matching.py
zhiqiangdon: CustomUnpickler is a wrapper on top of pickle. It was a legacy used to load models saved when automm was still within autogluon.text. Since we autogluon.text was deprecated a long time ago, we no longer need to support it.

zhiqiangdon:
> @zhiqiangdon I am aligned that the results show that performance is largely the same between mainline and this PR. If the PR was purely code cleanup / renaming then at this point I would be inclined to merge it.
> 
> > This PR implemented a bag of tricks for multimodal AutoML involving multimodal model fusion strategies, multimodal data augmentation, cross-modal alignment, converting tabular data into text, handling missing modalities with modality dropout and learnable embeddings, and an ensemble learner to integrate the bag of tricks for optimal performance
> 
> If we merged the PR as-is, then we would be merging it without testing any of these added features. I understand that the paper exists, but we need a way to replicate the results so that in future when we make changes we can ensure we are going in the right direction and aren't making performance regressions. At the very least one single configuration that shows benefit.
> 
> @zhiqiangdon Regarding meta-transformer, is there a reason `use_ensemble` doesn't work out of the box? Why don't we just download from HuggingFace when the user specifies `use_ensemble`?

@Innixma I agree that we'd better benchmark use_ensemble=True. HuggingFace doesn't host me-transformer checkpoints, so we can't download from it like other commonly used models. To make use_ensemble=True work out of box, we have two options in follow-up PRs:

1. Removing it from the presets to avoid errors. If users want to use it, they need to manually config it.
2. Hosting the checkpoints in our S3 bucket, and adding logic to auto-download them when users use me-transformer.

 @suzhoum Could you download the meta-transformer checkpoints and run this benchmark on existing multimodal (>1 modalities) datasets supported by ag-bench? Thanks.

Innixma:
@zhiqiangdon re meta-transformer. [2] is unlike what we have done in the past, correct? I'm a bit hesitant to do that. Re removing it, can we test the performance impact of removing it?

zhiqiangdon:
> @zhiqiangdon re meta-transformer. [2] is unlike what we have done in the past, correct? I'm a bit hesitant to do that. Re removing it, can we test the performance impact of removing it?

@Innixma I remember we hosted the pretrained checkpoints of ft-transformer from [XTab](https://arxiv.org/abs/2305.06090). I proposed this solution because it should be more stable than auto-downloading from the meta-transformer's website.

Innixma:
@zhiqiangdon maybe we can instead ask the authors to add the weights to HuggingFace?

suzhoum:
@zhiqiangdon There seems to have a bug when using `hyperparameters` during `.fit()`:

```
import os
import numpy as np
import warnings
warnings.filterwarnings('ignore')
np.random.seed(123)

download_dir = './ag_automm_tutorial'
zip_file = 'https://automl-mm-bench.s3.amazonaws.com/petfinder_for_tutorial.zip'
from autogluon.core.utils.loaders import load_zip

import pandas as pd
dataset_path = download_dir + '/petfinder_for_tutorial'
train_data = pd.read_csv(f'{dataset_path}/train.csv', index_col=0)
label_col = 'AdoptionSpeed'

image_col = 'Images'
train_data[image_col] = train_data[image_col].apply(lambda ele: ele.split(';')[0]) # Use the first image for a quick tutorial

def path_expander(path, base_folder):
    path_l = path.split(';')
    return ';'.join([os.path.abspath(os.path.join(base_folder, path)) for path in path_l])

train_data[image_col] = train_data[image_col].apply(lambda ele: path_expander(ele, base_folder=dataset_path))

predictor = MultiModalPredictor(label=label_col)
hyperparameters = {
    "optimization.max_epochs": 5,
    "optimization.learning_rate": 0.005
}
predictor.fit(
    train_data=train_data,
    time_limit=120, # seconds
    hyperparameters=hyperparameters
)
```

Traceback
```
  File "/home/ubuntu/src/autogluon-bench_zs/autogluon-bench/test.py", line 49, in <module>
    predictor.fit(
  File "/home/ubuntu/autogluon-bench-zq-pr/autogluon/multimodal/src/autogluon/multimodal/predictor.py", line 529, in fit
    self._learner.fit(
  File "/home/ubuntu/autogluon-bench-zq-pr/autogluon/multimodal/src/autogluon/multimodal/learners/base.py", line 665, in fit
    fit_returns = self.execute_fit()
                  ^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/autogluon-bench-zq-pr/autogluon/multimodal/src/autogluon/multimodal/learners/base.py", line 577, in execute_fit
    attributes = self.fit_per_run(**self._fit_args)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/autogluon-bench-zq-pr/autogluon/multimodal/src/autogluon/multimodal/learners/base.py", line 1275, in fit_per_run
    config = self.get_config_per_run(config=config, hyperparameters=hyperparameters)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/autogluon-bench-zq-pr/autogluon/multimodal/src/autogluon/multimodal/learners/base.py", line 703, in get_config_per_run
    config = get_config(
             ^^^^^^^^^^^
  File "/home/ubuntu/autogluon-bench-zq-pr/autogluon/multimodal/src/autogluon/multimodal/utils/config.py", line 198, in get_config
    config = apply_omegaconf_overrides(config, overrides=overrides, check_key_exist=True)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/autogluon-bench-zq-pr/autogluon/multimodal/src/autogluon/multimodal/utils/config.py", line 479, in apply_omegaconf_overrides
    raise KeyError(
KeyError: '"optimization.max_epochs" is not found in the config. You may need to check the overrides. overrides={\'optimization.max_epochs\': 5, \'optimization.learning_rate\': 0.005}'
```

zhiqiangdon:
> @zhiqiangdon There seems to have a bug when using `hyperparameters` during `.fit()`:
> 
> ```
> import os
> import numpy as np
> 
> download_dir = './ag_automm_tutorial'
> zip_file = 'https://automl-mm-bench.s3.amazonaws.com/petfinder_for_tutorial.zip'
> from autogluon.core.utils.loaders import load_zip
> load_zip.unzip(zip_file, unzip_dir=download_dir)
> 
> import pandas as pd
> dataset_path = download_dir + '/petfinder_for_tutorial'
> train_data = pd.read_csv(f'{dataset_path}/train.csv', index_col=0)
> label_col = 'AdoptionSpeed'
> 
> image_col = 'Images'
> train_data[image_col] = train_data[image_col].apply(lambda ele: ele.split(';')[0]) # Use the first image for a quick tutorial
> 
> def path_expander(path, base_folder):
>     path_l = path.split(';')
>     return ';'.join([os.path.abspath(os.path.join(base_folder, path)) for path in path_l])
> 
> train_data[image_col] = train_data[image_col].apply(lambda ele: path_expander(ele, base_folder=dataset_path))
> 
> from autogluon.multimodal import MultiModalPredictor
> 
> predictor = MultiModalPredictor(label=label_col)
> hyperparameters = {
>     "optimization.max_epochs": 5,
>     "optimization.learning_rate": 0.005
> }
> predictor.fit(
>     train_data=train_data,
>     time_limit=120, # seconds
>     hyperparameters=hyperparameters
> )
> ```
> 
> Traceback
> 
> ```
> Traceback (most recent call last):
>   File "/home/ubuntu/src/autogluon-bench_zs/autogluon-bench/test.py", line 49, in <module>
>     predictor.fit(
>   File "/home/ubuntu/autogluon-bench-zq-pr/autogluon/multimodal/src/autogluon/multimodal/predictor.py", line 529, in fit
>     self._learner.fit(
>   File "/home/ubuntu/autogluon-bench-zq-pr/autogluon/multimodal/src/autogluon/multimodal/learners/base.py", line 665, in fit
>     fit_returns = self.execute_fit()
>                   ^^^^^^^^^^^^^^^^^^
>   File "/home/ubuntu/autogluon-bench-zq-pr/autogluon/multimodal/src/autogluon/multimodal/learners/base.py", line 577, in execute_fit
>     attributes = self.fit_per_run(**self._fit_args)
>                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/home/ubuntu/autogluon-bench-zq-pr/autogluon/multimodal/src/autogluon/multimodal/learners/base.py", line 1275, in fit_per_run
>     config = self.get_config_per_run(config=config, hyperparameters=hyperparameters)
>              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/home/ubuntu/autogluon-bench-zq-pr/autogluon/multimodal/src/autogluon/multimodal/learners/base.py", line 703, in get_config_per_run
>     config = get_config(
>              ^^^^^^^^^^^
>   File "/home/ubuntu/autogluon-bench-zq-pr/autogluon/multimodal/src/autogluon/multimodal/utils/config.py", line 198, in get_config
>     config = apply_omegaconf_overrides(config, overrides=overrides, check_key_exist=True)
>              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/home/ubuntu/autogluon-bench-zq-pr/autogluon/multimodal/src/autogluon/multimodal/utils/config.py", line 479, in apply_omegaconf_overrides
>     raise KeyError(
> KeyError: '"optimization.max_epochs" is not found in the config. You may need to check the overrides. overrides={\'optimization.max_epochs\': 5, \'optimization.learning_rate\': 0.005}'
> ```

@suzhoum How about trying this change: `optimization` -> `optim`? This PR requires using the simplified the key.

zhiqiangdon:
> @zhiqiangdon maybe we can instead ask the authors to add the weights to HuggingFace?

@Innixma There were discussions regarding this. They haven't added the model into huggingface/transformers because some head weights are not published. See https://github.com/huggingface/transformers/issues/25725

Innixma:
@zhiqiangdon lets figure this out after testing the performance. In the meantime, if you can add an informative error message telling the user how to get the weights if they don't have them locally, that would be a good intermediate solution where the feature remains experimental until we find an easy way to get it working out of the box.

Innixma:
I'd also like to understand why we need the MetaTransformer model in the first place.

zhiqiangdon:
> I'd also like to understand why we need the MetaTransformer model in the first place.

@Innixma MetaTransformer serves as the early fusion model in our ensemble. It was pretrained on several modalities including image, text, and tabular data. There are many pretrained early fusion models for image and text data, but usually not including tabular data. The bag of tricks we proposed are for any combinations of the three modalities.

zhiqiangdon:
> @zhiqiangdon lets figure this out after testing the performance. In the meantime, if you can add an informative error message telling the user how to get the weights if they don't have them locally, that would be a good intermediate solution where the feature remains experimental until we find an easy way to get it working out of the box.

@Innixma A warning message is added. If no local MetaTransformer checkpoint is provided, we remove it from ensembling to avoid the error. The message also provides the guidance to download the checkpoints and configure using it.

Innixma:
@zhiqiangdon Thanks, warning msg looks good to me, and usage of MetaTransformer for early fusion makes sense.

Now will wait for benchmark results for `use_ensemble=True`.

suzhoum:
@zhiqiangdon @Innixma The benchmark is done for `MultiModalPredictor(use_ensemble=True)`. I'm attaching both raw results file and the pairwise evaluation results for reference. 

One thing to note is that `rmse` we got from `predictor.evaluate()["ensemble"]` are all negative. Maybe @zhiqiangdon can take a look on the logic again.

After removing regression datasets from the evaluation, the performances seem to be comparable with master.

[results_automlbenchmark_g4_12x_ag_bench_bot_2025_02_19 (1).csv](https://github.com/user-attachments/files/18874899/results_automlbenchmark_g4_12x_ag_bench_bot_2025_02_19.1.csv)
[AutoGluon_medium_master_g4_12x_ag_bench_bot_2025_02_19.csv](https://github.com/user-attachments/files/18874894/AutoGluon_medium_master_g4_12x_ag_bench_bot_2025_02_19.csv)
[AutoGluon_high_master_g4_12x_ag_bench_bot_2025_02_19.csv](https://github.com/user-attachments/files/18874896/AutoGluon_high_master_g4_12x_ag_bench_bot_2025_02_19.csv)
[AutoGluon_best_master_g4_12x_ag_bench_bot_2025_02_19.csv](https://github.com/user-attachments/files/18874897/AutoGluon_best_master_g4_12x_ag_bench_bot_2025_02_19.csv)

Innixma:
@zhiqiangdon @suzhoum Thanks for the result files! Comments:

1. What are we reporting in the `result` column? Is it score or error?
2. Why does RMSE flip sign in the PR? r2 doesn't flip sign, only RMSE. Need to identify cause.
3. Why are there 10 benchmark results for PR_best and 16 benchmark results for mainline_best? What are the failures?
4. Why does the comparison on best only show 13 datasets, and only 7 datasets with success by PR_best? Were there failures on both sides? How many datasets are there total in the benchmark?
5. PR_best is running for >10x longer than mainline_best. @zhiqiangdon does this PR's ensemble mode respect time limit?
6. Can we give equal time to each method? For example, 4 hours?

zhiqiangdon:
> @zhiqiangdon @Innixma The benchmark is done for `MultiModalPredictor(use_ensemble=True)`. I'm attaching both raw results file and the pairwise evaluation results for reference.
> 
> One thing to note is that `rmse` we got from `predictor.evaluate()["ensemble"]` are all negative. Maybe @zhiqiangdon can take a look on the logic again.
> 
> After removing regression datasets from the evaluation, the performances seem to be comparable with master.
> 
> [results_automlbenchmark_g4_12x_ag_bench_bot_2025_02_19 (1).csv](https://github.com/user-attachments/files/18874899/results_automlbenchmark_g4_12x_ag_bench_bot_2025_02_19.1.csv) [AutoGluon_medium_master_g4_12x_ag_bench_bot_2025_02_19.csv](https://github.com/user-attachments/files/18874894/AutoGluon_medium_master_g4_12x_ag_bench_bot_2025_02_19.csv) [AutoGluon_high_master_g4_12x_ag_bench_bot_2025_02_19.csv](https://github.com/user-attachments/files/18874896/AutoGluon_high_master_g4_12x_ag_bench_bot_2025_02_19.csv) [AutoGluon_best_master_g4_12x_ag_bench_bot_2025_02_19.csv](https://github.com/user-attachments/files/18874897/AutoGluon_best_master_g4_12x_ag_bench_bot_2025_02_19.csv)

@suzhoum Thanks for benchmarking and identifying the metric sign issue. I have fixed it in the new commit.

zhiqiangdon:
> @zhiqiangdon @suzhoum Thanks for the result files! Comments:
> 
> 1. What are we reporting in the `result` column? Is it score or error?
> 2. Why does RMSE flip sign in the PR? r2 doesn't flip sign, only RMSE. Need to identify cause.
> 3. Why are there 10 benchmark results for PR_best and 16 benchmark results for mainline_best? What are the failures?
> 4. Why does the comparison on best only show 13 datasets, and only 7 datasets with success by PR_best? Were there failures on both sides? How many datasets are there total in the benchmark?
> 5. PR_best is running for >10x longer than mainline_best. @zhiqiangdon does this PR's ensemble mode respect time limit?
> 6. Can we give equal time to each method? For example, 4 hours?

@Innixma I have fixed the metric sign issue in the new commit. It is due to directly calling the score function from autogluon.core to compute rmse, which gives negative values. Now ensemble learner uses the same way as base learner in computing this, which should resolve the issue. 

We do support time limit, but it is still for each individual model rather than the whole ensemble. If time_limit=1 hour, it would train each model for one hour.  So far, I don't think ensembling can improve baseline if given equal time.  In future, we can optimize this logic to respect the time of ensembling as a whole and maximize the performance given an overall time budget.

Innixma:
@zhiqiangdon Thanks for addressing some of the comments. If it isn't already mentioned in the doc, it would be good to add documentation that use_ensemble will run for up to time_limit * N seconds. Otherwise people will probably not be happy about the behavior. Maybe you can consider running until the time_limit is over, and simply ensembling whichever models happened to have finished, and skipping the remainder? This is what we do in AutoGluon Tabular and AutoGluon TimeSeries.

The other question remaining is why there were task failures. It seems only 7 shared datasets contain results?
Additionally, either we should re-run the regression tasks or flip the sign on RMSE in the result files so it is incorporated into the evaluation logic for checking performance with mainline.

zhiqiangdon:
> @zhiqiangdon Thanks for addressing some of the comments. If it isn't already mentioned in the doc, it would be good to add documentation that use_ensemble will run for up to time_limit * N seconds. Otherwise people will probably not be happy about the behavior. Maybe you can consider running until the time_limit is over, and simply ensembling whichever models happened to have finished, and skipping the remainder? This is what we do in AutoGluon Tabular and AutoGluon TimeSeries.
> 
> The other question remaining is why there were task failures. It seems only 7 shared datasets contain results? Additionally, either we should re-run the regression tasks or flip the sign on RMSE in the result files so it is incorporated into the evaluation logic for checking performance with mainline.

@Innixma Following your suggestion, I have added the documentation for time_limit used in ensembling. Aligning the logic with Taublar makes sense. We may leave it for future PRs considering the complexity of this PR. 

We may not need to re-run the benchmark on regression tasks since the negative sign happens only during evaluation, not affecting training and ensembling. Regarding other aspects of benchmarking, @suzhoum may have more context.

suzhoum:
Hi @zhiqiangdon @Innixma, to address your comments, I have reprocessed the results and attaching here. I have removed single modality datasets from the mainline benchmark, so there are in total 10 multi-modality datasets (2 image-tabular, 1 image-text, 2 image-text-tabular, 5 text-tabular) for evaluation. There was however one benchmark missing from the mainline_high_quality, but it should not affect the overall performance evaluation much. 
[AutoGluon_medium_master_g4_12x_ag_bench_bot_2025_02_19.csv](https://github.com/user-attachments/files/18896190/AutoGluon_medium_master_g4_12x_ag_bench_bot_2025_02_19.csv)
[AutoGluon_high_master_g4_12x_ag_bench_bot_2025_02_19.csv](https://github.com/user-attachments/files/18896191/AutoGluon_high_master_g4_12x_ag_bench_bot_2025_02_19.csv)
[AutoGluon_best_master_g4_12x_ag_bench_bot_2025_02_19.csv](https://github.com/user-attachments/files/18896192/AutoGluon_best_master_g4_12x_ag_bench_bot_2025_02_19.csv)

Innixma:
@suzhoum can you compare all 6 in the same table? I'm curious if ex: `en_medium > mainline_high`

Innixma:
In general results look very good, nice work @zhiqiangdon! And thanks for running the benchmarks @suzhoum!

I think we are nearly there for merging. The next steps are to minimize friction for existing users.

@zhiqiangdon The renaming of hyperparameters should include backwards compatibility, at least until v1.4 release. Can you create a separate PR that points to this PR (so it is a new branch building on-top of this PR's branch), which adds the backwards compatibility for hyperparameters? I mention a separate PR so it is easy to review isolated from the other changes in this PR.

This way we won't have a gap period between merging this PR and adding the backwards compatibility where existing users who are running AutoGluon via source install will have their code broken due to the breaking change.

You can consider this PR to be mostly fixed, I don't expect to request much changes to it before merge.

Innixma:
re backwards compatibility, one idea is to check the dictionary specified by the user at the start of the `predictor.fit` call and warn the user if using a old hyperparameter name, and then replacing it with the new one.

suzhoum:
> @suzhoum can you compare all 6 in the same table? I'm curious if ex: `en_medium > mainline_high`

Please see attached.
[AutoGluon_bot_vs_master_g4_12x_ag_bench_bot_2025_02_19.csv](https://github.com/user-attachments/files/18896900/AutoGluon_bot_vs_master_g4_12x_ag_bench_bot_2025_02_19.csv)

Innixma:
Thanks @suzhoum! This seems to indicate that en_medium > main_best by a notable margin. If we can resolve respecting time limit while maintaining most of the benefit, this could be a compelling option for users.

zhiqiangdon:
> In general results look very good, nice work @zhiqiangdon! And thanks for running the benchmarks @suzhoum!
> 
> I think we are nearly there for merging. The next steps are to minimize friction for existing users.
> 
> @zhiqiangdon The renaming of hyperparameters should include backwards compatibility, at least until v1.4 release. Can you create a separate PR that points to this PR (so it is a new branch building on-top of this PR's branch), which adds the backwards compatibility for hyperparameters? I mention a separate PR so it is easy to review isolated from the other changes in this PR.
> 
> This way we won't have a gap period between merging this PR and adding the backwards compatibility where existing users who are running AutoGluon via source install will have their code broken due to the breaking change.
> 
> You can consider this PR to be mostly fixed, I don't expect to request much changes to it before merge.

@Innixma I have added the backward compatibility in this PR. Opening another PR doesn't seem to save time. You can check the logic through the new commit.

multimodal/src/autogluon/multimodal/utils/config.py
```diff
@@ -812,3 +813,59 @@ def update_ensemble_hyperparameters(
...
+            )
+
+        for old_k, new_k in key_pairs.items():
+            if k == old_k:
+                overrides[new_k] = overrides.pop(provided_k)
+                logger.warning(
+                    f"The hyperparameter name {provided_k} is depreciated. "
+                    f"We recommend using the new name {new_k} instead."
+                )
+                break
```
Innixma, line 869: nit: `if k in key_pairs:` is simpler and avoids the for loop.

multimodal/src/autogluon/multimodal/utils/config.py
```diff
@@ -812,3 +813,59 @@ def update_ensemble_hyperparameters(
         hyperparameters = presets_hyperparameters
 
     return hyperparameters
+
+
+def make_overrides_backward_compatible(overrides: Dict):
```
Innixma, line 818: should we do a `overrides = copy.deepcopy(overrides)` at the start of this function to avoid in-place mutation of the user's dictionary?

multimodal/src/autogluon/multimodal/utils/config.py
```diff
@@ -796,3 +785,87 @@ def split_hyperparameters(hyperparameters: Dict):
...
+                f"Please replace `optimization.` with `optim.` when customizing the optimization hyperparameters."
+            )
+
+        for old_k, new_k in key_pairs.items():
+            if k == old_k:
+                overrides[new_k] = overrides.pop(provided_k)
+                logger.warning(
+                    f"The hyperparameter name {provided_k} is depreciated. "
+                    f"We recommend using the new name {new_k} instead."
+                )
```
Innixma, line 868:
```suggestion
                logger.warning(
                    f"The hyperparameter name {provided_k} is depreciated. "
                    f"We recommend using the new name {new_k} instead. "
                    f"The deprecated hyperparameter will raise an exception starting in AutoGluon 1.4.0"
                )
```

Innixma:
@zhiqiangdon thanks! I added comments to the backward compatibility code

multimodal/src/autogluon/multimodal/utils/config.py
zhiqiangdon: Thanks! Adopted the suggestion.

multimodal/src/autogluon/multimodal/utils/config.py
zhiqiangdon: A `copy.deepcopy` is applied before calling function `apply_omegaconf_overrides`, so no need to copy them again. See line 188 here: https://github.com/autogluon/autogluon/pull/4737/files#diff-a10c8783a894991a02709f19d3b1a2653731c04d798ec8b5234b0c9dc2d55bcdR188

multimodal/src/autogluon/multimodal/utils/config.py
zhiqiangdon: Updated.

Innixma:
@zhiqiangdon can you resolve the minor merge conflict? Once resolved, I think we are good to merge.

tabular/tests/unittests/models/test_automm.py
```diff
@@ -9,8 +9,7 @@ def test_automm_sts(fit_helper):
         # Skip test if no GPU available
         pytest.skip("Skip, no GPU available.")
     fit_args = dict(
-        hyperparameters={"AG_AUTOMM": {"env.num_workers": 0, "env.num_workers_evaluation": 0}},
-        time_limit=60,
+        hyperparameters={"AG_AUTOMM": {"env.num_workers": 0, "env.num_workers_inference": 0}},
```
Innixma, line 12: Please keep the `time_limit=60`, otherwise this takes arounds 10 minutes which slows down the CI

zhiqiangdon: Updated

zhiqiangdon:
> @zhiqiangdon can you resolve the minor merge conflict? Once resolved, I think we are good to merge.

@Innixma The conflict was resolved, and tests have passed.

Innixma (approved):
Looks good to me! Thanks @zhiqiangdon for the major improvements and cleanup, as well as the quick follow-ups!

Issue #4590, opened 2024-10-28
[BUG] TabularPredictor: Validation raises an exception for a 2-class string target column


th-omas-i:
**Bug Report Checklist**

- [x] I provided code that demonstrates a minimal reproducible example.  
- [ ] I confirmed bug exists on the latest mainline of AutoGluon via source install.  
- [x] I confirmed bug exists on the latest stable version of AutoGluon.  

**Describe the bug**
When a TabularPredictor is trained on a 2-class problem in "multiclass" mode and then evaluated, the evaluation fails with an autogluon-internal exception during balanced accuracy calculation (see below). This especially happens for a target column which is of type string.

**Expected behavior**
A 2-class TabularPredictor in "multiclass" mode can be trained and evaluated without exceptions, just as a 3-class or 4-class prediction.

**To Reproduce**
```
from autogluon.tabular import TabularPredictor
from autogluon.core.metrics.classification_metrics import balanced_accuracy
import numpy as np
import pandas as pd

x1 = [1,2,1,2,3,4,3,2,1,2,3,4,3,2,3,4,5,3,2,1]
x2 = [4,3,2,3,2,1,4,5,6,3,7,3,2,8,4,2,3,5,4,1]
solution = np.array(["ABC","DEF","ABC","ABC","DEF","ABC","DEF","ABC","ABC","DEF","ABC","DEF","ABC","ABC","DEF","ABC","DEF","ABC","ABC","DEF"])
full_data = pd.DataFrame({"x1":x1, "x2":x2, "label":solution})
model = TabularPredictor("label", problem_type="multiclass")
model.fit(full_data, time_limit=60)

probas = model.predict_proba(full_data)
eval_res = model.evaluate_predictions(full_data["label"], probas, detailed_report=True)  # THIS CRASHES
```

Reproduction of the actual library function call:

```
from autogluon.tabular import TabularPredictor
from autogluon.core.metrics.classification_metrics import balanced_accuracy
import numpy as np
import pandas as pd

# 3 classes: works
solution = np.array(["ABC","DEF","ABC","ABC","DEF","ABC","DEF","ABC","ABC","DEF","ABC","DEF","ABC","ABC","DEF","ABC","DEF","ABC","ABC","DEF","GHI"])
[...]

PR #4750, opened 2024-12-23
[tabular] Fix possible type error in balanced_accuracy

rsj123:
*Issue #, if available:*
#4590 4590
*Description of changes:*

In a multicategorization problem, predictions will not be converted to onehot to be passed to `balanced_accuracy`.
This causes an error when a test dataset containing only two categories is used in a multiclass task:
1. since it is a multiclass problem, the data will not be transformed before being passed to `balanced_accuracy`
2. since it contains only two categories, `balanced_accuracy` will guess that it is a binary problem and the data will not be transformed
So the data without onehot conversion is passed to the evaluation section

I think this modification may not be perfect, the best way would be to pass the problem type before passing it to the evaluation function, instead of relying on the data to guess.

### DIFF
--- a/core/src/autogluon/core/metrics/classification_metrics.py
+++ b/core/src/autogluon/core/metrics/classification_metrics.py
@@ -23,11 +23,7 @@ def balanced_accuracy(solution, prediction):
     if y_type not in ["binary", "multiclass", "multilabel-indicator"]:
         raise ValueError(f"{y_type} is not supported")
 
-    if y_type == "binary":
-        # Do not transform into any multiclass representation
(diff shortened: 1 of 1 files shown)


Innixma:
Thanks for the PR! This was helpful in informing me of the issue.

The issue is a bit more complicated it turns out, and checking for string values isn't quite enough. For example, the values [3, 7] would also raise an exception (or if not raise an exception, they would calculate the score incorrectly).

I've created an alternative solution in #4775 to address all edge-cases and add unit tests. I'll close this PR for now.

PR #4775, opened 2025-01-08
[tabular] Fix balanced_accuracy edge-case

### LINKED ISSUE #4590: [BUG] TabularPredictor: Validation raises an exception for a 2-class string target column

Innixma:
*Issue #, if available:*

Resolves #4590 
Closes #4750

*Description of changes:*

Fix balanced_accuracy edge-case exception detailed in #4590.

Additionally adds many metric unit tests to better detect bugs like this in future. The new unit tests catch the previous bug.

### DIFF
--- a/core/src/autogluon/core/metrics/classification_metrics.py
+++ b/core/src/autogluon/core/metrics/classification_metrics.py
@@ -25,7 +25,13 @@ def balanced_accuracy(solution, prediction):
 
     if y_type == "binary":
         # Do not transform into any multiclass representation
-        pass
+        unique_sol = np.unique(solution)
+        unique_pred = np.unique(prediction)
+        classes = np.unique(np.concatenate((unique_sol, unique_pred)))
+        if set(classes) != {0, 1}:
+            pos_class = classes[-1]
+            solution = np.array([1 if i == pos_class else 0 for i in solution])
+            prediction = np.array([1 if i == pos_class else 0 for i in prediction])
 
     elif y_type == "multiclass":
         n = len(solution)
--- a/core/tests/unittests/metrics/test_metrics.py
+++ b/core/tests/unittests/metrics/test_metrics.py
@@ -7,9 +7,21 @@ import sklearn
 from autogluon.core.constants import BINARY, MULTICLASS, QUANTILE, REGRESSION
 from autogluon.core.metrics import METRICS, Scorer, make_scorer, rmse_func
 
+METRICS_NEEDS_CLASS = {}
+METRICS_NEEDS_PROBA = {}
+METRICS_NEEDS_THRESHOLD = {}
+NOT_METRICS_NEEDS_THRESHOLD = {}
+for problem_type in METRICS:
+    METRICS_NEEDS_CLASS[problem_type] = [k for k, v in METRICS[problem_type].items() if v.needs_class]
+    METRICS_NEEDS_PROBA[problem_type] = [k for k, v in METRICS[problem_type].items() if v.needs_proba or v.needs_threshold]
+    METRICS_NEEDS_THRESHOLD[problem_type] = [k for k, v in METRICS[problem_type].items() if v.needs_threshold]
+    NOT_METRICS_NEEDS_THRESHOLD[problem_type] = [k for k, v in METRICS[problem_type].items() if not v.needs_threshold]
+
 BINARY_METRICS = list(METRICS[BINARY].keys())
 MULTICLASS_METRICS = list(METRICS[MULTICLASS].keys())
 REGRESSION_METRICS = list(METRICS[REGRESSION].keys())
+
+BINARY_METRICS_NEEDS_POS_LABEL = [k for k, v in METRICS[BINARY].items() if v.needs_pos_label]
 
 EXPECTED_BINARY_METRICS = {
     "acc",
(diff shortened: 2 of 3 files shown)


core/tests/unittests/metrics/test_metrics.py
```diff
@@ -140,6 +152,60 @@ def test_metrics_perfect_multiclass(metric: str):
     _assert_perfect_score(scorer=METRICS[MULTICLASS][metric])
 
 
+@pytest.mark.skip(reason="average_precision doesn't raise an exception here when it should, so this test currently fails")
```
suzhoum, line 155: Do we know why it fails?

core/tests/unittests/metrics/test_metrics.py
```diff
@@ -140,6 +152,60 @@ def test_metrics_perfect_multiclass(metric: str):
...
+@pytest.mark.parametrize("metric", METRICS_NEEDS_THRESHOLD[BINARY], ids=METRICS_NEEDS_THRESHOLD[BINARY])  # noqa
+def test_metrics_perfect_raises_binary_single_sample(metric: str):
+    with pytest.raises(Exception):
+        # threshold metrics should not be able to predict on a single sample
+        _assert_perfect_score_single_sample(scorer=METRICS[BINARY][metric])
+
+
+@pytest.mark.skip(reason="mcc, quadradic_kappa, and pac produce unexpected values here, so this test currently fails")
```
suzhoum, line 163: Do we plan to fix these tests in a follow up PR?

core/src/autogluon/core/metrics/classification_metrics.py
```diff
@@ -25,7 +25,13 @@ def balanced_accuracy(solution, prediction):
 
     if y_type == "binary":
         # Do not transform into any multiclass representation
-        pass
+        unique_sol = np.unique(solution)
+        unique_pred = np.unique(prediction)
+        classes = np.unique(np.concatenate((unique_sol, unique_pred)))
+        if set(classes) != {0, 1}:
+            pos_class = classes[-1]
```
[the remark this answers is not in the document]
Innixma: `max(classes)` will raise an exception if the values are strings, which they can be. In the balanced_accuracy metric which class is the positive class doesn't matter, since they have equal weight, so the selection mechanism for `pos_class` is fine as long as it is consistent.

core/tests/unittests/metrics/test_metrics.py
Innixma: This is an edge-case where the user tries to score a single sample's prediction.

It is the output of the scikit-learn metric, so it isn't a bug caused by AutoGluon's wrapper. It isn't a new issue, the new tests simply showed that this failure case exists.

It is likely because average precision (probably?) should be undefined for a single sample. https://scikit-learn.org/1.5/modules/model_evaluation.html#precision-recall-f-measure-metrics

However, sklearn decided instead of raising an exception to return a score of 1.0 if y_true = [1], ignoring the prediction (y_score), and similarly return a score of 0.0 if y_true != [1], ignoring the prediction.

```python
from sklearn.metrics import average_precision_score

y_score = [0]
y_true = [1]
score = average_precision_score(y_true=y_true, y_score=y_score)
print(score)  # 1.0

y_score = [1]
y_true = [1]
score = average_precision_score(y_true=y_true, y_score=y_score)
print(score)  # 1.0

y_score = [0.5]
y_true = [1]
score = average_precision_score(y_true=y_true, y_score=y_score)
print(score)  # 1.0

y_score = [0]
y_true = [0]
score = average_precision_score(y_true=y_true, y_score=y_score)
print(score)  # 0.0

y_score = [1]
y_true = [0]
score = average_precision_score(y_true=y_true, y_score=y_score)
print(score)  # 0.0

y_score = [0.5]
y_true = [0]
score = average_precision_score(y_true=y_true, y_score=y_score)
print(score)  # 0.0
```

Innixma: Because it is a threshold metric, and my understanding is that threshold metrics should fail if used to predict only a single sample, it behaves differently than the other threshold metrics which raise exceptions as expected, like roc_auc:

```python
from sklearn.metrics import roc_auc_score

y_score = [1]
y_true = [1]

score = roc_auc_score(y_true=y_true, y_score=y_score)
```

Output:

```
ValueError: Only one class present in y_true. ROC AUC score is not defined in that case.
```

Innixma: If I were to fix this in AutoGluon, users might complain that AutoGluon's metric raises an exception when scikit-learn's does not.

core/tests/unittests/metrics/test_metrics.py
Innixma: Possibly, however as in the prior comment on average_precision the details are quite nuanced for each, while the scenarios this comes up are very niche.

core/tests/unittests/metrics/test_metrics.py
Innixma: This is an example of the issue in sklearn: https://github.com/scikit-learn/scikit-learn/issues/30147#issuecomment-2438417120

suzhoum: Thanks for the context!

suzhoum (commented):
LGTM!

PR #4787, opened 2025-01-10
Increase tabpfn version min

Innixma:
*Issue #, if available:*

*Description of changes:*

Increase tabpfn version min as lower versions are broken after the TabPFN v2 release.


suzhoum (approved):
LGTM!

PR #4788, opened 2025-01-10
[timeseries] address type checker issues in metrics and trainer

canerturkmen:
*Issue #, if available:*

*Description of changes:*

Start addressing static type checker (pyright) issues in AbstractTrainer and the metrics module.

### DIFF
--- a/timeseries/src/autogluon/timeseries/metrics/abstract.py
+++ b/timeseries/src/autogluon/timeseries/metrics/abstract.py
@@ -1,4 +1,4 @@
-from typing import Optional, Tuple, Union, cast
+from typing import Optional, Tuple, Union
 
 import numpy as np
 import pandas as pd
@@ -161,7 +161,7 @@ class TimeSeriesScorer:
     @staticmethod
     def _safemean(array: Union[np.ndarray, pd.Series]) -> float:
         """Compute mean of a numpy array-like object, ignoring inf, -inf and nan values."""
-        return cast(float, np.mean(array[np.isfinite(array)]))
+        return float(np.mean(array[np.isfinite(array)]))
 
     @staticmethod
     def _get_point_forecast_score_inputs(
--- a/timeseries/src/autogluon/timeseries/metrics/abstract.py
+++ b/timeseries/src/autogluon/timeseries/metrics/abstract.py
@@ -1,4 +1,4 @@
-from typing import Optional, Tuple, Union
+from typing import Optional, Tuple, Union, cast
 
 import numpy as np
 import pandas as pd
@@ -161,7 +161,7 @@ class TimeSeriesScorer:
     @staticmethod
     def _safemean(array: Union[np.ndarray, pd.Series]) -> float:
         """Compute mean of a numpy array-like object, ignoring inf, -inf and nan values."""
-        return np.mean(array[np.isfinite(array)])
+        return cast(float, np.mean(array[np.isfinite(array)]))
 
     @staticmethod
     def _get_point_forecast_score_inputs(
--- a/timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
+++ b/timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
@@ -138,7 +138,6 @@ class AbstractTimeSeriesModel(AbstractModel):
         self.target_scaler: Optional[LocalTargetScaler] = None
         self.covariate_scaler: Optional[CovariateScaler] = None
         self.covariate_regressor: Optional[CovariateRegressor] = None
-        assert self.fit_time is None
         self.fit_time: Optional[float]
 
     def __repr__(self) -> str:
--- a/timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
+++ b/timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
@@ -3,7 +3,7 @@ import os
 import re
 import time
 from contextlib import nullcontext
-from typing import Dict, List, Optional, Tuple, Union
+from typing import Any, Dict, List, Optional, Tuple, Union
 
 import pandas as pd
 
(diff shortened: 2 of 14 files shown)


timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -212,7 +212,7 @@ def save_info(self, include_model_info: bool = False):
         save_json.save(path=os.path.join(self.path, self.trainer_info_json_name), obj=info)
         return info
 
-    def get_model_best(self, *args, **kwargs) -> AbstractModel:
+    def get_model_best(self, *args, **kwargs) -> Union[str, AbstractModel]:
```
shchur, line 215: Should this always return `str`?

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -760,7 +763,7 @@ def fit_ensemble(
         for window_idx, data in enumerate(data_per_window):
             predictions = ensemble.predict({n: model_preds[n][window_idx] for n in ensemble.model_names})
             score_per_fold.append(self._score_with_predictions(data, predictions))
-        ensemble.val_score = np.mean(score_per_fold)
+        ensemble.val_score = cast(float, np.mean(score_per_fold))
```
shchur, line 766: Same as before, maybe `float(np.mean(score_per_fold))`?

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -818,6 +821,8 @@ def leaderboard(
                 raise_exception_if_failed=False,
                 use_cache=use_cache,
             )
+            assert isinstance(model_predictions, dict)
+            assert isinstance(pred_time_dict, dict)
```
shchur, line 825: Maybe a slightly cleaner option would be to always return `Tuple[Dict[...], Dict[...]]` from `get_model_pred_dict` (equivalent to setting `record_pred_time=True`)? This would, however, require some modifications to the tests and all usages of `get_model_pred_dict` throughout the codebase.

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -941,7 +946,11 @@ def predict(
             use_cache=use_cache,
             random_seed=random_seed,
         )
-        return model_pred_dict[model_name]
+        assert isinstance(model_pred_dict, dict)
+        predictions = model_pred_dict[model_name]
+        if predictions is None:
+            raise ValueError(f"Model {model_name} failed to predict. Please check the model's logs.")
```
shchur, line 952: Interesting, this block can actually never be reached since `raise_exception_if_failed=True` in `get_model_pred_dict`. Maybe just adding a type annotation or type: ignore would do the trick?
```python
predictions: TimeSeriesDataFrame = model_pred_dict[model_name] 
```

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -1178,7 +1191,9 @@ def get_model_pred_dict(
         raise_exception_if_failed: bool = True,
         use_cache: bool = True,
         random_seed: Optional[int] = None,
-    ) -> Union[Dict[str, TimeSeriesDataFrame], Tuple[Dict[str, TimeSeriesDataFrame], Dict[str, float]]]:
+    ) -> Union[
+        Dict[str, Optional[TimeSeriesDataFrame]], Tuple[Dict[str, Optional[TimeSeriesDataFrame]], Dict[str, float]]
+    ]:
```
shchur, line 1196: Related to the above comment: Should this always be `Tuple[Dict[str, Optional[TimeSeriesDataFrame]], Dict[str, float]]`?

shchur (commented):
Thank you for the cleanup! Just a bunch of minor comments

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
```diff
@@ -138,6 +138,7 @@ def __init__(
         self.target_scaler: Optional[LocalTargetScaler] = None
         self.covariate_scaler: Optional[CovariateScaler] = None
         self.covariate_regressor: Optional[CovariateRegressor] = None
+        self.fit_time: Optional[float]
```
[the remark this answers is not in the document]
canerturkmen: The superclass already sets this. This just adds a type hint.

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
```diff
@@ -402,7 +403,7 @@ def predict(
 
         Parameters
         ----------
-        data: Union[TimeSeriesDataFrame, Dict[str, TimeSeriesDataFrame]]
+        data: Union[TimeSeriesDataFrame, Dict[str, Optional[TimeSeriesDataFrame]]]
```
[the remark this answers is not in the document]
canerturkmen: I think that's fair. Let's add it to the backlog.

timeseries/src/autogluon/timeseries/metrics/abstract.py
```diff
@@ -161,7 +161,7 @@ def error(self, *args, **kwargs):
     @staticmethod
     def _safemean(array: Union[np.ndarray, pd.Series]) -> float:
         """Compute mean of a numpy array-like object, ignoring inf, -inf and nan values."""
-        return np.mean(array[np.isfinite(array)])
+        return cast(float, np.mean(array[np.isfinite(array)]))
```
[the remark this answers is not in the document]
canerturkmen: cast doesn't have any effects at runtime, whereas float results in a copy operation. however I agree that this is cleaner and will change it.

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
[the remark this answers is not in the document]
shchur: The `assert` statement looks a bit odd. Maybe we can overwrite with the type hint `self.fit_time: Optional[float] = None` or add a type hint in the superclass?

This is really a nit, feel free to ignore.

shchur (approved):
One tiny nit, otherwise LGTM!

PR #4799, opened 2025-01-15
Add `test_version.py`

Innixma:
*Issue #, if available:*

*Description of changes:*

- Add `test_version.py` to ensure that we don't accidentally do a release with a version that doesn't include an explicit micro / minor version. For example, we released `1.2` instead of `1.2.0`. While `1.2` is technically valid, it causes issues with some AWS logic such as docker builds that expect an explicit micro version.
- Update the way we create our `version.py` file so the content aligns with the ruff formater. Previously it complained that the version file lacked a newline and should use `"` instead of `'` for string values. This would lead to unnecessary clutter and confusing when users would perform automated linting prior to submitting a PR.

Example of the previous confusing lint message:

```
ruff format --diff "timeseries/"
--- timeseries/src/autogluon/timeseries/version.py
+++ timeseries/src/autogluon/timeseries/version.py
@@ -1,2 +1,3 @@
 """This is the autogluon version file."""
-__version__ = '1.2'
+
+__version__ = "1.2"

1 file would be reformatted, 65 files already formatted

```

### DIFF
--- a/common/tests/unittests/test_version.py
+++ b/common/tests/unittests/test_version.py
@@ -1,0 +1,18 @@
+from packaging import version
+
+from autogluon.common import __version__
+
+
+def test_version_has_major_minor_micro():
+    """
+    Verifies that the version contains a major minor and micro version explicitly.
+
+    For example, `__version__ = "1.2"` will fail this test because it does not contain an explicit micro version.
+    """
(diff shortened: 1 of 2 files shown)


shchur (approved):
Thanks for the fix!

PR #4801, opened 2025-01-15
[timeseries] address type checker issues in transforms, learner and others

canerturkmen:
*Issue #, if available:*

*Description of changes:*

```
pyright timeseries/src/autogluon/timeseries/{configs,trainer,metrics,transforms,regressor.py,splitter.py,learner.py,evaluator.py}
```

### DIFF
--- a/timeseries/src/autogluon/timeseries/learner.py
+++ b/timeseries/src/autogluon/timeseries/learner.py
@@ -51,31 +51,15 @@ class TimeSeriesLearner(AbstractLearner):
             target=self.target, known_covariates_names=self.known_covariates_names
         )
 
-    def load_trainer(self) -> AbstractTimeSeriesTrainer:
+    def load_trainer(self) -> AbstractTimeSeriesTrainer:  # type: ignore
         """Return the trainer object corresponding to the learner."""
-        return super().load_trainer()  # noqa
+        return super().load_trainer()  # type: ignore
 
     def fit(
         self,
         train_data: TimeSeriesDataFrame,
-        val_data: TimeSeriesDataFrame = None,
-        hyperparameters: Union[str, Dict] = None,
-        hyperparameter_tune_kwargs: Optional[Union[str, dict]] = None,
-        **kwargs,
-    ) -> None:
-        return self._fit(
-            train_data=train_data,
-            val_data=val_data,
-            hyperparameters=hyperparameters,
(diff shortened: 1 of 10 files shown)


timeseries/src/autogluon/timeseries/learner.py
```diff
@@ -145,7 +131,7 @@ def _align_covariates_with_forecast_index(
 
         If some of the item_ids or timestamps are missing, an exception is raised.
         """
-        if len(self.known_covariates_names) == 0:
+        if len(self.known_covariates_names) == 0 or known_covariates is None:
```
shchur, line 134: This looks potentially buggy to me: If `len(self.known_covariates_names) != 0` but `known_covariates is None` we want to raise an exception (see the next `if` block), but this block will just return `None`.

timeseries/src/autogluon/timeseries/learner.py
```diff
@@ -51,31 +51,15 @@ def __init__(
...
-            train_data=train_data,
-            val_data=val_data,
-            hyperparameters=hyperparameters,
-            hyperparameter_tune_kwargs=hyperparameter_tune_kwargs,
-            **kwargs,
-        )
-
-    def _fit(
-        self,
-        train_data: TimeSeriesDataFrame,
```
shchur, line 76: Nice!

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -577,26 +576,21 @@ def _train_multi(
...
             self.is_data_saved = True
 
-        if models is None:
-            models = self.construct_model_templates(
-                hyperparameters=hyperparameters,
-                hyperparameter_tune=hyperparameter_tune_kwargs is not None,  # TODO: remove hyperparameter_tune
-                freq=train_data.freq,
-                multi_window=self.val_splitter.num_val_windows > 0,
-                excluded_model_types=excluded_model_types,
-            )
```
shchur, line 599: Nice, as we get rid of dead branches like these, I feel that we are getting value out of the type checker

timeseries/src/autogluon/timeseries/learner.py
canerturkmen: thanks for catching! put up a new version...

shchur (approved):
LGTM!

PR #4804, opened 2025-01-16
[core] refactor AbstractTrainer

canerturkmen:
*Issue #, if available:*

*Description of changes:*

This is the first PR of a few for a major refactor of `AbstractTrainer` unifying the interfaces of `AbstractTimeSeriesTrainer` and tabular's Trainer class into one abstract class in core.

This PR factors out `AbstractTrainer` in `core`, including model management behavior that is common in time series and tabular, as well as common interfaces. 

Subsequent PRs will
- Move `AbstractTrainer` in core to tabular, since it is now almost exclusively tabular specific concretizations.
- Remove `SimpleAbstractTrainer` from timeseries, and move it to core.


core/src/autogluon/core/trainer/abstract_trainer.py
```diff
@@ -21,22 +19,27 @@
...
+    AbstractModel,
+    BaggedEnsembleModel,
+    GreedyWeightedEnsembleModel,
+    SimpleWeightedEnsembleModel,
+    StackerEnsembleModel,
+    WeightedEnsembleModel,
+)
+from autogluon.core.pseudolabeling.pseudolabeling import assert_pseudo_column_match
+from autogluon.core.ray.distributed_jobs_managers import ParallelFitManager
+from autogluon.core.utils import (
```
Innixma, line 42: Why switch from relative to absolute imports? Tabular, core, common, and features all use relative imports.

Is there a general guideline you are following for favoring absolute over relative?

core/src/autogluon/core/trainer/abstract_trainer.py
```diff
@@ -4092,7 +4200,7 @@ def _delete_model_from_graph(self, model: str):
         shutil.rmtree(path=path_attr_model, ignore_errors=True)
 
     @classmethod
-    def load(cls, path, reset_paths=False):
+    def load(cls, path: str, reset_paths: bool = False) -> AbstractTabularTrainer:
```
Innixma, line 4203: ditto on `Self` return type

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -34,16 +35,15 @@
...
-
+# TODO: This class will be removed after the behavior is reconciled with core
+class SimpleAbstractTrainer(AbstractTrainer):
     def __init__(self, path: str, low_memory: bool, save_data: bool, *args, **kwargs):
+        super().__init__(
+            path=path,
+            low_memory=low_memory,
+            save_data=save_data,
+        )
+
```
Innixma, line 46: For the init code below this, you should be able to remove pretty much all of it right? Looks like code-dupe atm.

Innixma (approved):
Very nice! Added some minor comments but overall this is a great step in the right direction for unifying the API of trainer.

core/src/autogluon/core/trainer/abstract_trainer.py
canerturkmen: Nothing particular in this case, it's just that these will move into tabular anyway. I can switch to relative imports for the ones that remain after that.

core/src/autogluon/core/trainer/abstract_trainer.py
```diff
@@ -45,18 +48,213 @@
...
+
+    def get_model_names(self, **kwargs) -> list[str]:
+        """Get all model names that are registered in the model graph"""
+        raise NotImplementedError
+
+    def save(self) -> None:
+        raise NotImplementedError
+
+    @classmethod
+    def load(cls, path: str, reset_paths: bool = False) -> AbstractTrainer:
```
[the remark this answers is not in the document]
canerturkmen: 👍 👍 Thanks!

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
canerturkmen: Yes. I will also get rid of this class altogether shortly.

core/src/autogluon/core/trainer/abstract_trainer.py
```diff
@@ -45,18 +48,213 @@
...
+    def save_info(self, include_model_info=False) -> dict[str, Any]:
+        info = self.get_info(include_model_info=include_model_info)
+
+        save_pkl.save(path=os.path.join(self.path, self.trainer_info_name), object=info)
+        save_json.save(path=os.path.join(self.path, self.trainer_info_json_name), obj=info)
+        return info
+
+    def construct_model_templates(
+        self, hyperparameters: str | dict[str, Any], **kwargs
+    ) -> tuple[Sequence[AbstractModel], dict] | Sequence[AbstractModel]:
```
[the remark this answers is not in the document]
canerturkmen: Yeah TIL. because with list typing complains 

```
  /local/home/atturkm/workplace/autogluon/timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py:1315:9 - error: Method "construct_model_templates" overrides class "AbstractTrainer" in an incompatible manner
    Return type mismatch: base method returns type "tuple[list[AbstractModel], dict[Unknown, Unknown]] | list[AbstractModel]", override returns type "List[AbstractTimeSeriesModel]"
      Type "List[AbstractTimeSeriesModel]" is not assignable to type "tuple[list[AbstractModel], dict[Unknown, Unknown]] | list[AbstractModel]"
        "List[AbstractTimeSeriesModel]" is not assignable to "tuple[list[AbstractModel], dict[Unknown, Unknown]]"
        "List[AbstractTimeSeriesModel]" is not assignable to "list[AbstractModel]"
          Type parameter "_T@list" is invariant, but "AbstractTimeSeriesModel" is not the same as "AbstractModel"
          Consider switching from "list" to "Sequence" which is covariant (reportIncompatibleMethodOverride)
```

i.e., Sequence is covariant and will be OK with a more specific type (`AbstractTimeSeriesModel < AbstractModel`) but list is invariant.

core/src/autogluon/core/trainer/abstract_trainer.py
```diff
@@ -45,18 +48,213 @@
...
+        self.path = self.create_contexts(path_context)
+
+    def create_contexts(self, path_context: str) -> str:
+        path = path_context
+        return path
+
+    def save_model(self, model: AbstractModel, **kwargs) -> None:
+        model.save()
+        if not self.low_memory:
+            self.models[model.name] = model
```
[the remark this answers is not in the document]
canerturkmen: agreed. 👍

core/src/autogluon/core/trainer/abstract_trainer.py
shchur: [PEP8](https://peps.python.org/pep-0008/#imports) recommends absolute imports but I would vote in favor of consistency with other modules here.

core/src/autogluon/core/trainer/abstract_trainer.py
```diff
@@ -45,18 +49,213 @@
...
+        if attribute == "path":
+            return os.path.join(*self.model_graph.nodes[model][attribute])
+        return self.model_graph.nodes[model][attribute]
+
+    def set_model_attribute(self, model, attribute: str, val):
+        if not isinstance(model, str):
+            model = model.name
+        self.model_graph.nodes[model][attribute] = val
+
+    def get_minimum_model_set(self, model, include_self=True) -> list:
```
shchur, line 159: Nit: Some methods are missing type hints.

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -160,28 +117,10 @@ def load_model(
...
-            model = model.name
-        minimum_model_set = list(nx.bfs_tree(self.model_graph, model, reverse=True))
-        if not include_self:
-            minimum_model_set = [m for m in minimum_model_set if m != model]
-        return minimum_model_set
-
-    def get_models_info(self, models: Optional[List[str]] = None) -> Dict[str, Any]:
-        if models is None:
-            models = self.get_model_names()
+    def get_models_info(self, models: Optional[List[str | AbstractModel]] = None) -> Dict[str, Any]:
```
shchur, line 120: What is the reason for the difference between this method and `AbstractTrainer.get_models_info`?

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -34,38 +35,19 @@
...
-# TODO: This class is meant to be moved to `core`, where it will likely
-# TODO: be renamed `AbstractTrainer` and the current `AbstractTrainer`
-# TODO: will inherit from this class.
-# TODO: add documentation for abstract methods
-class SimpleAbstractTrainer:
-    trainer_file_name = "trainer.pkl"
-    trainer_info_name = "info.pkl"
-    trainer_info_json_name = "info.json"
-
+# TODO: This class will be removed after the behavior is reconciled with core
```
shchur, line 38: So this class currently only contains methods that will be completely removed in a follow-up PR?

core/src/autogluon/core/trainer/abstract_trainer.py
```diff
@@ -3573,7 +3710,7 @@ def get_base_model_names(self, model) -> list:
         base_model_set = list(self.model_graph.predecessors(model))
         return base_model_set
 
-    def model_refit_map(self, inverse=False) -> Dict[str, str]:
+    def model_refit_map(self, inverse=False) -> dict[str, str]:
```
shchur, line 3713: Could you add a TODO next to `AbstractTimeSeriesTrainer.model_refit_map` in the `__init__` method that we should reconcile the method / object?

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -240,9 +157,6 @@ def get_info(self, include_model_info: bool = False) -> Dict[str, Any]:
 
         return info
 
-    def predict(self, *args, **kwargs):
-        raise NotImplementedError
-
 
 class AbstractTimeSeriesTrainer(SimpleAbstractTrainer):
     _cached_predictions_filename = "cached_predictions.pkl"
```
shchur, line 162: Nit: 

This class also contains some methods that could, in theory, be moved to `AbstractTrainer` and unified with Tabular like`persist` and `unpersist`.

Also `load_data` and `construct_model_templates` could be abstract methods in the `AbstractTrainer` with `NotImplementedError`.

shchur (approved):
Thank you for this herculean effort! Only a several questions and minor comments

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
canerturkmen: will do in a subsequent PR

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
canerturkmen: will be reconciled in the follow-up PR

core/src/autogluon/core/trainer/abstract_trainer.py
shchur: I was really hoping that by specializing in Python instead of C++ I will be able to avoid learning what a generic is... welp

PR #4820, opened 2025-01-22
[core] make AbstractTrainer generic in model type, remove SimpleAbstractTrainer

canerturkmen:
… SimpleAbstractTrainer

*Issue #, if available:*

*Description of changes:*

- Makes `AbstractTrainer` generic in the model type, allowing tighter type checking for timeseries. This signals to users that `AbstractTrainer`s will always be coupled to an `AbstractModel` object that they work with. This is also a good first step for moving shared stuff out of `AbstractModel` and refactoring `AbstractTimeSeriesModel` which still improperly inherits from `AbstractModel`.
- Removes the redundant `SimpleAbstractTrainer` from timeseries. 


timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -201,9 +82,8 @@ def __init__(
 
         self.verbosity = verbosity
 
-        # Dict of normal model -> FULL model. FULL models are produced by
-        # self.refit_single_full() and self.refit_full().
-        # TODO: reconcile this with abstract classes
+        #: Dict of normal model -> FULL model. FULL models are produced by
+        #: self.refit_single_full() and self.refit_full().
```
shchur, line 86: `#:` is a typo?

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -35,130 +35,7 @@
...
-        }
-
-        if include_model_info:
-            info["model_info"] = self.get_models_info()
-
-        return info
-
-
-class AbstractTimeSeriesTrainer(SimpleAbstractTrainer):
+class AbstractTimeSeriesTrainer(AbstractTrainer[AbstractTimeSeriesModel]):
```
shchur, line 38: Why do we have a separate class for `AutoTrainer(AbstractTabularTrainer)` and `AutoTimeSeriesTrainer(AbstractTimeSeriesTrainer)`? Should these be merged into respective single classes as well?

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -254,6 +138,17 @@ def save(self) -> None:
...
+    @classmethod
+    def load(cls, path: str, reset_paths: bool = False) -> Self:
+        load_path = os.path.join(path, cls.trainer_file_name)
+        if not reset_paths:
+            return load_pkl.load(path=load_path)
+        else:
+            obj = load_pkl.load(path=load_path)
+            obj.set_contexts(path)
+            obj.reset_paths = reset_paths
+            return obj
```
shchur, line 150: This looks equivalent to `AbstractTabularTrainer.load()`. Should we move the method to `AbstractTrainer`?

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
```diff
@@ -344,6 +239,55 @@ def get_model_names(self, level: Optional[int] = None, **kwargs) -> List[str]:
...
+        for model in models_:
+            if isinstance(model, str):
+                if model in self.models.keys():
+                    model = self.models[model]
+                model_type = self.get_model_attribute(model=model, attribute="type")
+                model_path = os.path.join(self.path, self.get_model_attribute(model=model, attribute="path"))
+                model_info_dict[model] = model_type.load_info(path=model_path)
+            else:
+                model_info_dict[model.name] = model.get_info()
+        return model_info_dict
```
shchur, line 264: These two methods could, in theory, be reconciled between `AbstractTabularTrainer` and `AbstractTimeSeriesTrainer`. Do you think that's possible in this PR? Or should we at least add a TODO?

core/src/autogluon/core/trainer/abstract_trainer.py
```diff
@@ -68,7 +68,10 @@
 logger = logging.getLogger(__name__)
 
 
-class AbstractTrainer:
+ModelTypeT = TypeVar("ModelTypeT", bound=AbstractModel)
```
shchur, line 71: Amazing!

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
canerturkmen: apparently this is a [doc comment](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#doc-comments-and-docstrings) that I thought we might start using for the non-public stuff (will remove it from the other PR)

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
canerturkmen: Right now this class hierarchy for timeseries is there for consistency with tabular. However I agree we can merge into a single `TimeSeriesTrainer` for timeseries. Wdyt?

canerturkmen: i.e., now I think we have everything in place to do that.

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
canerturkmen: I don't think we should try extra hard. In principle the two implementations serve (i.e., are responsible to) different users. The contract is there to 1/ make it easier for users to switch from one module to another and 2/ force us to not drift apart too much in behavior and not do very obvious code duplication.

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
canerturkmen: thanks!

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
shchur: These two methods looked pretty close, but I'm also fine with keeping them as-is.

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
shchur: Let's merge them, it's something that bothered me for a while.

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
shchur: TIL! Are there some cases in which Sphinx will build the documentation for the Trainer class?

shchur (dismissed):
LGTM!

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
canerturkmen: Not really, but we still follow sphinx conventions.

timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py
canerturkmen: Worked more on it. `get_models_attribute_dict` changes behavior for our `fit_summary()` API method ever so slightly, and quite likely doesn't affect tabular's behavior. Note that we can see this as a fix. Tabular worked inconsistently when `get_models_attribute_dict("path", models=None)` and `get_models_attribute_dict("path", models=models)` was called.

timeseries/src/autogluon/timeseries/trainer.py
```diff
@@ -1313,21 +1221,69 @@ def refit_full(self, model: str = "all") -> Dict[str, str]:
...
+        random_seed: Optional[int] = None,
+    ):
+        """
+        Fit a set of timeseries models specified by the `hyperparameters`
+        dictionary that maps model names to their specified hyperparameters.
 
-    # TODO: persist and unpersist models
+        Parameters
+        ----------
+        train_data: TimeSeriesDataFrame
```
shchur, line 1264: Nit: Let's remove the type hints from the docstrings, they are already available in the method signature?

core/src/autogluon/core/trainer/abstract_trainer.py
```diff
@@ -167,31 +172,31 @@ def get_minimum_model_set(self, model: str | AbstractModel, include_self: bool =
             minimum_model_set = [m for m in minimum_model_set if m != model]
         return minimum_model_set
 
-    def get_model_info(self, model: str | AbstractModel) -> dict[str, Any]:
+    def get_model_info(self, model: str | ModelTypeT) -> dict[str, Any]:
         if isinstance(model, str):
             if model in self.models.keys():
                 model = self.models[model]
-        if isinstance(model, str):
```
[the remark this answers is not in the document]
canerturkmen: aargh. yes. thanks!

timeseries/src/autogluon/timeseries/trainer.py
```diff
@@ -1313,21 +1221,69 @@ def refit_full(self, model: str = "all") -> Dict[str, str]:
...
-        self, hyperparameters: Union[str, Dict[str, Any]], multi_window: bool = False, **kwargs
-    ) -> List[AbstractTimeSeriesModel]:
-        """Constructs a list of unfit models based on the hyperparameters dict."""
-        raise NotImplementedError
+    def construct_model_templates(self, hyperparameters: Union[str, Dict], multi_window: bool = False, **kwargs):
+        path = kwargs.pop("path", self.path)
+        eval_metric = kwargs.pop("eval_metric", self.eval_metric)
+        eval_metric_seasonal_period = kwargs.pop("eval_metric", self.eval_metric_seasonal_period)
+        quantile_levels = kwargs.pop("quantile_levels", self.quantile_levels)
+        hyperparameter_tune = kwargs.get("hyperparameter_tune", False)
```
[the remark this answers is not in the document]
canerturkmen: the default values are instance variables, though I'll do it anyway for type safety.

timeseries/src/autogluon/timeseries/trainer.py
```diff
@@ -1313,21 +1221,69 @@ def refit_full(self, model: str = "all") -> Dict[str, str]:
         logger.info(f"Total runtime: {time.time() - time_start:.2f} s")
         return copy.deepcopy(self.model_refit_map)
 
-    def construct_model_templates(
-        self, hyperparameters: Union[str, Dict[str, Any]], multi_window: bool = False, **kwargs
-    ) -> List[AbstractTimeSeriesModel]:
-        """Constructs a list of unfit models based on the hyperparameters dict."""
-        raise NotImplementedError
+    def construct_model_templates(self, hyperparameters: Union[str, Dict], multi_window: bool = False, **kwargs):
```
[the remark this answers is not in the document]
canerturkmen: How do you mean? isn't the return type of `AbstractTrainer` `tuple[list[ModelTypeT], dict] | list[ModelTypeT]` ?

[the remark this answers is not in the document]
shchur: Sorry, my bad. I missed `|` and only noticed the `tuple`

core/src/autogluon/core/trainer/abstract_trainer.py
```diff
@@ -167,31 +172,31 @@ def get_minimum_model_set(self, model: str | AbstractModel, include_self: bool =
...
             model_type = self.get_model_attribute(model=model, attribute="type")
             model_path = self.get_model_attribute(model=model, attribute="path")
             model_info = model_type.load_info(path=os.path.join(self.path, model_path))
         else:
             model_info = model.get_info()
         return model_info
+    
+    def get_model_names(self, **kwargs) -> list[str]:
+        """Get all model names that are registered in the model graph"""
+        return list(self.model_graph.nodes)
```
Innixma, line 188: Would recommend adding a docstring detail on if there are any guarantees on the ordering of the names in the output. For example, is it the order it was added to the graph, or is it alphanumeric?

core/src/autogluon/core/trainer/abstract_trainer.py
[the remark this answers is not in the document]
Innixma: Unresolving this as the current PR still needs to address this

core/src/autogluon/core/trainer/abstract_trainer.py
canerturkmen: I added "in no particular order." Underlying implementation appears to be a dict, for model_graph which doesn't guarantee ordering. Moreover, even if it were, I don't think the interface should constrain this.

timeseries/src/autogluon/timeseries/trainer.py
```diff
@@ -1314,20 +1232,79 @@ def refit_full(self, model: str = "all") -> Dict[str, str]:
...
+        *,
+        multi_window: bool = False,
+        path: Optional[str] = None,
+        freq: Optional[str] = None,
+        eval_metric: Optional[str] = None,
+        excluded_model_types: Optional[List[str]] = None,
+        eval_metric_seasonal_period: Optional[int] = None,
+        quantile_levels: Optional[List[float]] = None,
+        hyperparameter_tune: bool = False,
+        **kwargs,
```
shchur, line 1246: Looking at how this method is used, should we even expose `eval_metric`, `path`, `eval_metric_seasonal_period`  and `quantile_levels` as kwargs here? These attributes are set during the `__init__` of the trainer. Maybe we can just always set them to `self.*` when calling `get_preset_models`?

https://github.com/autogluon/autogluon/blob/f1bd5f42b2da0099c8d7319f38f811127446d9af/timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py#L502-L508

shchur (approved):
One small comment, otherwise LGTM!

Innixma (approved):
Looks great, awesome work!

timeseries/src/autogluon/timeseries/trainer.py
canerturkmen: agreed!

PR #4821, opened 2025-01-22
[timeseries] address static type checker issues in TimeSeriesPredictor

canerturkmen:
*Issue #, if available:*

*Description of changes:*

- Also, remove deprecations and "DeprecationMixins"


timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -194,7 +177,9 @@ def __init__(
         self.freq = freq
         if self.freq is not None:
             # Standardize frequency string (e.g., "T" -> "min", "Y" -> "YE")
-            std_freq = pd.tseries.frequencies.to_offset(self.freq).freqstr
+            offset = pd.tseries.frequencies.to_offset(self.freq)
+            assert offset is not None
+            std_freq = offset.freqstr
```
shchur, line 182: Disappointing that `pyright` cannot infer that `offset is not None` from the method signature :(

image

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -161,7 +145,6 @@ def __init__(
         log_file_path: Union[str, Path] = "auto",
         quantile_levels: Optional[List[float]] = None,
         cache_predictions: bool = True,
-        learner_type: Optional[Type[AbstractLearner]] = None,
         learner_kwargs: Optional[dict] = None,
```
shchur, line 148: Should we also deprecate `learner_kwargs` while we are at it? It seems that pretty much all the arguments are simply forwarded from the predictor to the learner, I don't see why anyone would want to override them.

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
```diff
@@ -870,6 +854,9 @@ def assign(self, **kwargs) -> TimeSeriesDataFrame:
         """Assign new columns to the time series dataframe. See :meth:`pandas.DataFrame.assign` for details."""
         return super().assign(**kwargs)  # type: ignore
 
+    def sort_index(self, *args, **kwargs) -> TimeSeriesDataFrame:
```
shchur, line 857: Maybe add a comment explaining that this method was only added for pyright? Similarly for other methods following this pattern.

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -375,7 +362,7 @@ def _filter_useless_train_data(
         train_data: TimeSeriesDataFrame,
         num_val_windows: int,
         val_step_size: int,
-    ) -> Tuple[TimeSeriesDataFrame, Optional[TimeSeriesDataFrame]]:
+    ) -> TimeSeriesDataFrame:
```
shchur, line 365: Nice catch

timeseries/src/autogluon/timeseries/learner.py
```diff
@@ -61,7 +61,7 @@ def fit(
         hyperparameters: Union[str, Dict],
         val_data: Optional[TimeSeriesDataFrame] = None,
         hyperparameter_tune_kwargs: Optional[Union[str, dict]] = None,
-        time_limit: Optional[int] = None,
+        time_limit: Optional[float] = None,
```
shchur, line 64: After the initial review of all "breaking" changes, let's also change the type hints to `dict`, `list`, etc. This will make the docs look better https://auto.gluon.ai/stable/api/autogluon.timeseries.TimeSeriesPredictor.html

canerturkmen: I'm planning to do that library-wide in a future PR if that's OK

Issue #4792, opened 2025-01-14
[timeseries] Clarify the documentation for `known_covariates` during `predict()`


shchur:
## Description
We should clarify which values should be provided as `known_covariates` during prediction time.  

The [current documentation](https://auto.gluon.ai/stable/tutorials/timeseries/forecasting-indepth.html) says:
"The timestamp index must include the values for prediction_length many time steps into the future from the end of each time series in train_data".
This formulation is ambiguous.

- [ ] impove wording in the documentation
- [ ] add a method `TimeSeriesPredictor.get_forecast_index(data) -> pd.MultiIndex` that returns the `item_id, timestamp` index that should be covered by the `known_covariates` during `predict()`.

PR #4834, opened 2025-01-23
[timeseries] Remove _cached_freq attribute for TimeSeriesDataFrame

### LINKED ISSUE #4843: [timeseries] reference to python 3.8 should be removed/updated
Autogluon no longer supports Python 3.8. The minimum version is now 3.9, so the reference to Python 3.8 below should be removed or updated:

https://github.com/autogluon/autogluon/blob/f1bd5f42b2da0099c8d7319f38f811127446d9af/timeseries/tests/unittests/utils/test_features.py#L50

cc: @shchur @canerturkmen

### LINKED ISSUE #4844: [Question] Clarification on TiDE model's support for past_known_cov
The documentation states that both TFT and TiDE have the parameter disable_past_covariates, which implies support for past_known_cov. However, the model zoo table does not show a checkmark ```(✅)``` under the "past covariates" column for TiDE.

Could you clarify if this is an oversight in the model zoo table or if TiDE indeed does not support past_known_cov despite the presence of disable_past_covariates in the parameters?

shchur:
*Issue #, if available:*
- The frequency caching mechanism sometimes results in very weird and hard-to-debug situations when an incorrect data frequency is cached even when resampling is applied.

*Description of changes:*
- Originally, we used `data.freq` to infer the data frequency throughout the codebase, so caching this computation made sense. We can safely remove the caching mechanism now since
  1. Currently, we only infer the frequency once when data is passed to the predictor with `data.infer_frequency()`. In the rest of the code, we pass around the `freq` as a string (e.g., setting it as a model attribute).
  2. The new implementation of `data.freq` is much faster than before, so there is minimal overhead.
- Always use all items when inferring the data frequency inside the `TimeSeriesPredictor`.
- Move the method for creating the forecast horizon index into `AbstractTimeSeriesModel.get_forecast_horizon_index` to avoid code duplication across different models.
- Minor documentation fixes:
  - Fixes #4843
  - Fixes #4844

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
+++ b/timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
@@ -442,10 +442,7 @@ class AbstractTimeSeriesModel(AbstractModel):
 
         if self.covariate_regressor is not None:
             if known_covariates is None:
-                forecast_index = get_forecast_horizon_index_ts_dataframe(
-                    data, prediction_length=self.prediction_length, freq=self.freq
-                )
-                known_covariates = pd.DataFrame(index=forecast_index, dtype="float32")
+                known_covariates = pd.DataFrame(index=self.get_forecast_horizon_index(data), dtype="float32")
 
             predictions = self.covariate_regressor.inverse_transform(
                 predictions,
@@ -456,6 +453,10 @@ class AbstractTimeSeriesModel(AbstractModel):
         if self.target_scaler is not None:
             predictions = self.target_scaler.inverse_transform(predictions)
         return predictions
+
+    def get_forecast_horizon_index(self, data: TimeSeriesDataFrame) -> pd.MultiIndex:
+        """For each item in the dataframe, get timestamps for the next `prediction_length` time steps into the future."""
+        return get_forecast_horizon_index_ts_dataframe(data, prediction_length=self.prediction_length, freq=self.freq)
 
     def _predict(
         self,
--- a/timeseries/src/autogluon/timeseries/models/chronos/pipeline/utils.py
+++ b/timeseries/src/autogluon/timeseries/models/chronos/pipeline/utils.py
@@ -253,7 +253,6 @@ class ChronosInferenceDataset:
         assert context_length > 0
         self.context_length = context_length
         self.target_array = target_df[target_column].to_numpy(dtype=np.float32)
-        self.freq = target_df.freq
 
         # store pointer to start:end of each time series
         cum_sizes = target_df.num_timesteps_per_item().values.cumsum()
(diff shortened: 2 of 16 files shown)


timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -278,25 +278,25 @@ def _check_and_prepare_data_frame(
         # MultiIndex.is_monotonic_increasing checks if index is sorted by ["item_id", "timestamp"]
         if not df.index.is_monotonic_increasing:
             df = df.sort_index()
-            df._cached_freq = None  # in case frequency was incorrectly cached as IRREGULAR_TIME_INDEX_FREQSTR
```
canerturkmen, line 281: Great!

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
```diff
@@ -457,6 +454,10 @@ def predict(
             predictions = self.target_scaler.inverse_transform(predictions)
         return predictions
 
+    def get_forecast_horizon_index(self, data: TimeSeriesDataFrame) -> pd.MultiIndex:
```
canerturkmen, line 457: design nit: Is there any reason left for `get_forecast_horizon_index_ts_dataframe` to exist in `utils`? If so, why are we having the models depend on it through their superclass?

timeseries/src/autogluon/timeseries/models/chronos/pipeline/utils.py
```diff
@@ -253,7 +253,6 @@ def __init__(
         assert context_length > 0
         self.context_length = context_length
         self.target_array = target_df[target_column].to_numpy(dtype=np.float32)
-        self.freq = target_df.freq
```
canerturkmen, line 256: Nice!

canerturkmen (approved):
LGTM! Only one question.

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
shchur: We have showed this method to the users in the [in-depth tutorial](https://auto.gluon.ai/stable/tutorials/timeseries/forecasting-indepth.html#time-varying-covariates) and I have seen a few examples of customers using it in their code. The functionality is non-trivial to implement on your own and I don't want to break this without warning. How about the following plan?
1. We add the deprecation decorator to the method in `utils.forecast` and hard-remove it in v2.0
2. We add a method `get_forecast_horizon_index` to `TimeSeriesPredictor` so that users still have this functionality (see #4792)

canerturkmen: Sounds good to me.

shchur: Will do the deprecation in a separate PR. For now just added a TODO

Issue #4776, opened 2025-01-08
[BUG] Chronos & Chronos-Bolt unable to hyperparameter tune, and fails to output best found parameters when it succeeds.


GhostDog98:
**Bug Report Checklist**

- [ ] I provided code that demonstrates a minimal reproducible example.  
- [ ] I confirmed bug exists on the latest mainline of AutoGluon via source install.  
- [ ] I confirmed bug exists on the latest stable version of AutoGluon.  

**Describe the bug**

When attempting to tune hyperparameters, none of the hypertuning options work.

**To Reproduce**

When running:
```
predictor.fit(
    train_data,
    hyperparameters={
        "ChronosModel": {
            "model_path": "bolt_base",
            "context_length": 2048,
        },
    },
    enable_ensemble=False,
)
```
Training occurs correctly. However, when running:
```
predictor.fit(
    train_data,
    hyperparameters={
        "ChronosModel": {
            "model_path": "bolt_base",
            "context_length": space.Int(64, 4096),
        },
    },
    hyperparameter_tune_kwargs="auto",
    enable_ensemble=False,
)
```
It fails due with:
```
Traceback (most recent call last):
  File "/mnt/bc457ffc-58c4-4dfe-a922-2b44ae3fa37e/gluon/train.py", line 212, in <module>
    predictor.fit(
  File "/home/ghostdog/miniconda3/envs/autogluon/lib/python3.11/site-packages/autogluon/core/utils/decorators.py", line 31, in _call
    return f(*gargs, **gkwargs)
           ^^^^^^^^^^^^^^^^^^^^
  File "/home/ghostdog/miniconda3/envs/autogluon/lib/python3.11/site-packages/autogluon/timeseries/predictor.py", line 753, in fit
    self._learner.fit(
  File "/home/ghostdog/miniconda3/envs/autogluon/lib/python3.11/site-packages/autogluon/timeseries/learner.py", line 66, in fit
    return self._fit(
           ^^^^^^^^^^
  File "/home/ghostdog/miniconda3/envs/autogluon/lib/python3.11/site-packages/autogluon/timeseries/learner.py", line 126, in _fit
    self.trainer.fit(
  File "/home/ghostdog/miniconda3/envs/autogluon/lib/python3.11/site-packages/autogluon/timeseries/trainer/auto_trainer.py", line 67, in fit
    self._train_multi(
[...]

PR #4838, opened 2025-01-24
[timeseries] Bugfix to allow tuning of `context_length` in Chronos

abdulfatir:
*Issue #, if available:* #4776

*Description of changes:* This PR fixes a bug which prevented tuning of the `context_length` hyperparameter in Chronos models. Attributes are moved from `__init__` to another helper function which is called inside `_fit`. 
 

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/chronos/model.py
+++ b/timeseries/src/autogluon/timeseries/models/chronos/model.py
@@ -9,6 +9,7 @@ import numpy as np
 import pandas as pd
 
 from autogluon.common.loaders import load_pkl
+from autogluon.common.space import Space
 from autogluon.timeseries.dataset.ts_dataframe import TimeSeriesDataFrame
 from autogluon.timeseries.models.abstract import AbstractTimeSeriesModel
 from autogluon.timeseries.utils.forecast import get_forecast_horizon_index_ts_dataframe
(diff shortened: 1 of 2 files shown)


abdulfatir:
@shchur I initially planned to defer all initialization logic from `__init__()`. However, initialization is actually not the problem. The parameters initialized in `__init__()` are also later concretized (a bit magical, not sure if this is intended). Only validation was problematic, so I moved it to another place.

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -577,6 +570,15 @@ def _get_inference_data_loader(
...
     def _get_context_length(self, data: TimeSeriesDataFrame) -> int:
+        # we validate the context_length here because the value is concrete,
+        # unlike in the constructor where it may be a search space
+        if self.context_length is not None and self.context_length > self.maximum_context_length:
+            logger.info(
+                f"\tContext length {self.context_length} exceeds maximum context length {self.maximum_context_length}."
+                f"Context length will be set to {self.maximum_context_length}."
+            )
+            self.context_length = self.maximum_context_length
+
```
shchur, line 581: Should we maybe move all the `self.` attributes set during `__init__` (`batch_size`, `num_samples`, `device`, `torch_dtype`, `optimization_strategy`, `data_loader_num_workers` and `context_length`) to the `_fit` method? They all can potentially run into issues with HPO. 

We should try to use the access pattern with `model_params = self._get_model_params()` called only inside `_fit` (as in https://github.com/autogluon/autogluon/blob/master/timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py#L235). These are guaranteed to contain "concrete" values, unlike the `hyperparameters` dict in `__init__` that may contain search spaces.

cc: @canerturkmen

shchur: @abdulfatir, sorry, I only now noticed your comment above. Please ignore my previous message.

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -216,13 +216,6 @@ def __init__(
...
         self.context_length = hyperparameters.get("context_length")
 
-        if self.context_length is not None and self.context_length > self.maximum_context_length:
-            logger.info(
-                f"\tContext length {self.context_length} exceeds maximum context length {self.maximum_context_length}."
-                f"Context length will be set to {self.maximum_context_length}."
-            )
-            self.context_length = self.maximum_context_length
-
         # we truncate the name to avoid long path errors on Windows
```
shchur, line 219: Should we add some unit tests to make sure that non of the other hyperparameters (excluding `model_path`) break HPO?

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -193,39 +194,11 @@ def __init__(
...
-                f"\tContext length {self.context_length} exceeds maximum context length {self.maximum_context_length}."
-                f"Context length will be set to {self.maximum_context_length}."
-            )
-            self.context_length = self.maximum_context_length
-
-        # we truncate the name to avoid long path errors on Windows
-        model_path_safe = str(model_path_input).replace("/", "__").replace(os.path.sep, "__")[-50:]
-        name = (name if name is not None else "Chronos") + f"[{model_path_safe}]"
+        name = name if name is not None else "Chronos"
+        if not isinstance(model_path_input, Space):
```
abdulfatir, line 198: This is needed because if the user decides to "tune" model size, the model name looks quite ugly like `Chronos[Categorical[.., ...]`.

shchur (dismissed):
LGTM, thanks for the fix!

abdulfatir:
@shchur The changes to `__init__` broke persist which was leading to test failures. I have now modified `persist()` to first `initialize()` and then assign model attributes. What do you think?

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -333,6 +306,9 @@ def load_model_pipeline(self, is_training: bool = False):
         self._model_pipeline = pipeline
 
     def persist(self) -> "ChronosModel":
+        self.initialize()
+        model_params = self._get_model_params()
+        self._validate_and_assign_attributes(model_params)
```
shchur, line 311: I think the actual problem is that `persist` should only be called after `fit()` is complete (for any model, not just Chronos).

A short-term solution would be to just add the fit() call to the unit tests that use persist. Another option would be to assert that the model has been fit before persist is called (e.g. by storing an attribute that is set to True during fit).

Option 1 + adding a TODO for later would be fine with me.

abdulfatir: @shchur Thanks! I went with option 1 for now and opened an issue to keep track of this. #4861

shchur (approved):
LGTM!

timeseries/src/autogluon/timeseries/models/chronos/model.py
shchur: Thanks!

Issue #4832, opened 2025-01-23
[BUG] [timeseries] Covariate regressor creates an empty folder under `AutogluonModels/ag-{TIMESTAMP}CatBoostModel``


shchur:
**Bug Report Checklist**

- [x] I provided code that demonstrates a minimal reproducible example.  
- [x] I confirmed bug exists on the latest mainline of AutoGluon via source install.  
- [ ] I confirmed bug exists on the latest stable version of AutoGluon.  

**Describe the bug**
When fitting a forecasting model with a `covariate_regressor` inside a `TimeSeriesPredictor`, AutoGluon creates an empty directory under `AutogluonModels/ag-**********ModelName`. 

**Expected behavior**
No directory should be created. Even if the regressor must be saved to disk, it should be done in the folder corresponding to the predictor / model.

**To Reproduce**
```python
from autogluon.timeseries import TimeSeriesDataFrame, TimeSeriesPredictor

data = TimeSeriesDataFrame.from_path(
    "https://autogluon.s3.amazonaws.com/datasets/timeseries/grocery_sales/test.csv",
)
predictor = TimeSeriesPredictor(
    prediction_length=8,
    target="unit_sales",
    known_covariates_names=["scaled_price", "promotion_email", "promotion_homepage"],
    path="test_predictor",
    verbosity=0,
)
predictor.fit(data, hyperparameters={"Chronos": {"covariate_regressor": "CAT"}}, time_limit=10)
```
**Screenshots / Logs**
Directory structure before fit:
```
└── reproduce.py
```
Directory structure after fit:
```
├── AutogluonModels
│   └── ag-20250123_124514CatBoostModel
├── reproduce.py
└── test_predictor
    ├── learner.pkl
    ├── logs
    │   └── predictor_log.txt
    ├── models
    │   ├── Chronos[autogluon__chronos-bolt-small]
    │   │   ├── W0
    │   │   │   └── model.pkl
    │   │   ├── model.pkl
    │   │   └── utils
    │   │       └── oof.pkl
    │   └── trainer.pkl
    ├── predictor.pkl
    ├── utils
    │   └── data
    │       └── train.pkl
    └── version.txt
```
Note the folder `AutogluonModels/ag-20250123_124514CatBoostModel` that was created. This happened even though the predictor was saved to `test_predictor/`.

**Root cause**
When a tabular regression model is created in 
[...]

PR #4845, opened 2025-01-27
[timeseries] Refactor covariate support properties

### LINKED ISSUE #4808: [BUG] [timeseries] TimeSeriesPredictor.feature_importance outputting 0 when covariate is used by regressor
**Bug Report Checklist**

- [x] I confirmed bug exists on the latest mainline of AutoGluon via source install.  

**Describe the bug**
When `TimeSeriesPredictor.feature_importance` is called given only Chronos with CatBoost covariate regressor, feature importance is computed as 0 even though CatBoost assigns 100% feature importance to one of the covariates. 

set up:
```
import numpy as np
import pandas as pd

df = pd.read_csv("https://autogluon.s3.amazonaws.com/datasets/timeseries/m4_daily_subset/train.csv")
df.head()

static_features_df = pd.read_csv("https://autogluon.s3.amazonaws.com/datasets/timeseries/m4_daily_subset/metadata.csv")
static_features_df.head()

train_data = TimeSeriesDataFrame.from_data_frame(
    df,
    id_column="item_id",
    timestamp_column="timestamp",
    static_features_df=static_features_df,
)
train_data.head()

train_data["log_target"] = np.log(train_data["target"])

WEEKEND_INDICES = [5, 6]
timestamps = train_data.index.get_level_values("timestamp")
train_data["weekend"] = timestamps.weekday.isin(WEEKEND_INDICES).astype(float)

train_data.head()

predictor = TimeSeriesPredictor(
    prediction_length=14,
    target="target",
    known_covariates_names=["weekend"],
).fit(
    train_data,
    hyperparameters={
        "Chronos": {
            "model_path": "bolt_small",
            "covariate_regressor": "CAT",
            "target_scaler": "standard",
            "ag_args": {"name_suffix": "WithRegressor"},    
        },
    }
)

predictor.feature_importance(train_data)
```

```python
trainer = predictor._learner.load_trainer()
model = trainer.load_model(
    trainer.get_model_best()
)
reg = model._get_model_base().covariate_regressor.fit(train_data).model.model

dict(zip(reg.feature_names_, reg.feature_importances_))  # {'weekend': 100.0}
```

Similar issue: https://github.com/autogluon/autogluon/issues/4322

```python
show_versions()
```

</details>

canerturkmen:
*Issue #, if available:*

Fix #4808

*Description of changes:*

Now `supports_x_covariates` are properties computed at runtime and are not settable. If a covariate regressor is made available by the `_create_covariate_regressor` method, then the model objects report they support known and static features.


timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
```diff
@@ -171,6 +172,22 @@ def load_oof_predictions(cls, path: str, verbose: bool = True) -> List[TimeSerie
         """Load the cached OOF predictions from disk."""
         return load_pkl.load(path=os.path.join(path, "utils", cls._oof_filename), verbose=verbose)
 
+    @property
+    def supports_known_covariates(self) -> bool:
+        return self._create_covariate_regressor() is not None or self.__class__._supports_known_covariates
```
shchur, line 177: I would recommend checking for something like `self._get_model_params().get("covariate_regressor") is not None` to avoid any potential side effects of  `_create_covariate_regressor()` (like #4832).

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
```diff
@@ -87,9 +87,10 @@ class AbstractTimeSeriesModel(AbstractModel):
     _preprocess_nonadaptive = None
     _preprocess_set_features = None
 
-    supports_known_covariates: bool = False
-    supports_past_covariates: bool = False
-    supports_static_features: bool = False
+    _supports_known_covariates: bool = False
+    _supports_past_covariates: bool = False
+    _supports_static_features: bool = False
+    _supports_cat_covariates: bool = False
```
shchur, line 93: This attribute is only relevant for GluonTS models. Should we keep it there without the underscore (and maybe add a comment)?

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
canerturkmen: I thought about this a bit. User-defined parameters are not a great "source of truth," as we have a habit of overriding them. It's not unrealistic that in the near future some models override user-provided "CAT" to None, or None to "CAT" or even override during deferred init. I agree regarding the side effect but I think this deserves some thought.

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
canerturkmen: If we can imagine it being relevant for other models I would rather keep it at the top level?

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
```diff
@@ -87,9 +87,10 @@ class AbstractTimeSeriesModel(AbstractModel):
     _preprocess_nonadaptive = None
     _preprocess_set_features = None
 
-    supports_known_covariates: bool = False
```
shchur, line 90: We should also update the custom model tutorial https://auto.gluon.ai/dev/tutorials/timeseries/advanced/forecasting-custom-model.html#implement-the-custom-model

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
shchur: I feel that even the `supports_cat_covariates` flag for GluonTS models is kind of a hack because it's unclear if this applies to known, past, or both types of covariates. It currently happens that the only model where this is set to `True` (TFT) is also the only GluonTS model that supports past covariates, but this doesn't have to be true in general. 

I would prefer to keep it limited to GluonTS until we at least explore other backends for DL models.

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
shchur: I don't currently see such scenario (overriding user params for `covariate_regressor`) as very likely and have a slight preference for the hyperparameter check, but I'm also okay with the current solution - up to you.

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
canerturkmen: Got it. back it goes.

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
canerturkmen: OK, will switch to hyperparameters for now.

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
canerturkmen: I sent it back to gluonts, but kept it as a property to keep it private and consistent.

shchur (approved):
LGTM!

PR #4851, opened 2025-01-30
[core] Move AbstractTabularTrainer to tabular

canerturkmen:
*Issue #, if available:*

*Description of changes:*

- Minor typing tweaks in `AbstractTrainer`, `TimeSeriesTrainer`
- Minor edits in `AbstractTabularTrainer`: ~formatting~, typing additions to method signatures or removing dead code.

*Note*: After your approval I will move the `AbstractTabularTrainer` out of core and into tabular. I am opening the PR in this state to make code review easier. @shchur @Innixma 


core/src/autogluon/core/trainer/abstract_trainer.py
```diff
@@ -189,13 +194,15 @@ def load_model(self, model_name: str | ModelTypeT, path: str | None = None, mode
...
-                path = self.get_model_attribute(model=model_name, attribute="path")  # get relative location of the model to the trainer
+                path = self.get_model_attribute(
+                    model=model_name, attribute="path"
+                )  # get relative location of the model to the trainer
                 assert path is not None
             if model_type is None:
                 model_type = self.get_model_attribute(model=model_name, attribute="type")
                 assert model_type is not None
             return model_type.load(path=os.path.join(self.path, path), reset_paths=self.reset_paths)
-
```
shchur, line 198: Maybe open a separate PR first to enable formatting for core/tabular with `ruff`? This would simplify the reviewing imo.

timeseries/src/autogluon/timeseries/trainer.py
```diff
@@ -761,7 +761,6 @@ def predict(
         model: Optional[Union[str, AbstractTimeSeriesModel]] = None,
         use_cache: bool = True,
         random_seed: Optional[int] = None,
-        **kwargs,
```
shchur, line 764: I think we can keep this door open in case we allow passing extra `kwargs` to models during predict. We can also add the kwargs back in the future if a need arises, so this is just a nit.

core/src/autogluon/core/trainer/abstract_trainer.py
canerturkmen: agreed and done 👍

canerturkmen: i.e., reverted the formatting-only changes for now.

timeseries/src/autogluon/timeseries/trainer.py
canerturkmen: What could those kwargs be? I am now realizing we use it as a placeholder for `subclasses or we might use this in the future`, but we actually want the extra `unexpected kwargs` check. i.e., `kwargs` doesn't really signal "the user can pass some other parameters" but rather "the user can pass whatever they want."

core/src/autogluon/core/trainer/abstract_trainer.py
```diff
@@ -1152,7 +1161,14 @@ def _get_best(self) -> str:
...
     # Note: model_pred_proba_dict is mutated in this function to minimize memory usage
-    def get_inputs_to_model(self, model: str | AbstractModel, X: pd.DataFrame, model_pred_proba_dict: dict[str, np.ndarray] = None, fit=False, preprocess_nonadaptive=False):
+    def get_inputs_to_model(
+        self,
+        model: str | AbstractModel,
+        X: pd.DataFrame,
+        model_pred_proba_dict: dict[str, np.ndarray] | None = None,
+        fit: bool = False,
+        preprocess_nonadaptive: bool = False,
+    ):
```
Innixma, line 1171: For Tabular the output of this method is a pd.DataFrame. Unsure what it is for TS, but should we have a type hint for the return type?

core/src/autogluon/core/trainer/abstract_trainer.py
```diff
@@ -1689,70 +1729,6 @@ def refit_single_full(
...
-                    models_trained_full_level.append(model_trained)
-
-                    logger.log(20,f"Finished refit model for {refit_full_parent}")
-                    unfinished_job_refs += distributed_manager.schedule_jobs()
-
-                logger.log(20, f"Finished distributed refitting for {len(models_trained_full_level)} L{level} models.")
-                models_trained_full += models_trained_full_level
-                distributed_manager.clean_job_state(unfinished_job_refs=unfinished_job_refs)
-
-            distributed_manager.clean_up_ray()
```
Innixma, line 1755: Why are we deleting this code?

timeseries/src/autogluon/timeseries/trainer.py
shchur: Currently two models support `time_limit` as a kwarg to `predict`:
https://github.com/autogluon/autogluon/blob/ee40fa4a87bffbdb1ba60bf354ba0106e5d90cce/timeseries/src/autogluon/timeseries/models/chronos/model.py#L608

https://github.com/autogluon/autogluon/blob/ee40fa4a87bffbdb1ba60bf354ba0106e5d90cce/timeseries/src/autogluon/timeseries/models/local/abstract_local_model.py#L163
Let's make this explicit?

There is a minor concern that keeping `kwargs` in the template for custom model implementations makes them more future-proof (i.e., if we start passing a new kwarg to `model.predict`, this won't break custom implementations). But this is a really minor concern - I'm fine removing all kwargs for now (but keeping the `time_limit`).

core/src/autogluon/core/trainer/abstract_trainer.py
canerturkmen: It's dead code (i.e., is never used)

canerturkmen: cf. below. type checker warned code is unreachable, up to you if you'd like to keep it.

https://github.com/autogluon/autogluon/blob/bd24491f6c7e0ea4468644d6781573c578a73d4a/core/src/autogluon/core/trainer/abstract_trainer.py#L1675-L1677

Innixma: I agree it is dead code, but I would like to keep it since this was a stretch goal for 1.2 release that didn't quite make it since I didn't have time to comprehensively test it. The plan is it will be used in 1.3 release, since it is a major speedup, but it needs to be made memory safe (current implementation isn't necessarily memory safe).

Maybe instead of removing you could add a FIXME / TODO at the top of the code mentioning that this is currently dead code.

timeseries/src/autogluon/timeseries/trainer.py
Innixma: Another thing that probably should be supported eventually for predict calls would be `num_cpus` and `num_gpus`.

Also some models, such as TabPFN, can actually be ensembled during inference by randomly shuffling the data, so theoretically they could want to be passed this as a kwarg in the predict call (although this one is starting to get pretty ambitious)

core/src/autogluon/core/trainer/abstract_trainer.py
Innixma: Basically, I don't want to forget that a lot of work towards this feature has already been done, which could very well happen if we delete the code.

Innixma: the WIP follow-up PR that uses this code: https://github.com/autogluon/autogluon/pull/4680

core/src/autogluon/core/trainer/abstract_trainer.py
canerturkmen: added type hint

timeseries/src/autogluon/timeseries/trainer.py
canerturkmen: We haven't really unified the model and trainer interfaces yet, so I'm not sure if the comparisons to models apply.

> There is a minor concern that keeping kwargs in the [template for custom model implementations](https://auto.gluon.ai/dev/tutorials/timeseries/advanced/forecasting-custom-model.html#implement-the-custom-model) makes them more future-proof (i.e., if we start passing a new kwarg to model.predict, this won't break custom implementations). But this is a really minor concern - I'm fine removing all kwargs for now (but keeping the time_limit).

I think this is what we _shouldn't_ do. We are using kwargs in function arguments as a catch-all for "futureproofing." We can add the keyword arguments in as we introduce them, as long as those keyword arguments have defaults, it's backwards compatible. If we find this is getting out of hand, this should in fact signal to us that our objects are exposing their innards too much and we should find another solution. So the interfaces stay tight and if something ends up breaking we fix it.

Independently of the futureproofing argument, custom model tutorial can keep the `**kwargs` there, there's nothing stopping that. Adding `**kwargs` in a concretion is OK. Adding it to abstractions, however, signals that anyone implementing the abstraction will need to deal with arbitrary parameters. That is,

```python

# not ok! Child also needs to expose `**kwargs`

class Parent:    
    def foo(self, a, **kwargs):
        pass
   
 
class Child(Parent):
    def foo(self, a, b, c = 0):
        pass

# ok. this is like the child implements two methods foo(a) and foo(a, b) thereby satisfying substitutability

class Parent:    
    def foo(self, a):
        pass
   
class Child(Parent):
    def foo(self, a, b = 1):
        pass

# also ok.

class Parent:    
    def foo(self, a):
        pass
   
class Child(Parent):
    def foo(self, a, **kwargs):
        pass
        
```

About adding `time_limit`, I think adding it now and not implementing it is worse than adding it when we actually make it work.

canerturkmen: One reason for confusion, TIL, is how we are used to think of `*args, **kwargs` in Python (instead of just `**kwargs`).

There is a legacy habit of 
```
class Parent:    
    def foo(self, *args, **kwargs):
        pass
```

to mean "I absolutely don't care what the subclass does" instead of "all subclasses must accept arbitrarily many positional and keyword arguments." The first is convention, the second is what Liskov substitution principle implies. However it looks like the first wrong interpretation is so common static type checker implementers just decided to [go with it](https://github.com/python/mypy/issues/5876).

shchur: Thanks for the clarification, I somehow completely missed the point of this discussion but now I think I'm on the same page. Here is my takeaway:
- We remove `**kwargs` from `TimeSeriesTrainer.predict`
- We keep `**kwargs` for the custom model in the custom model tutorial
- We may add `time_limit`, `num_cpus` and `num_gpus` as optional arguments to `AbstractTimeSeriesModel` in the future (does not concern this PR)

shchur (approved):
LGTM!

Innixma (approved):
LGTM, awesome work!

Issue #4859, opened 2025-02-02
[BUG]


alexkovaceski:
**Bug Report Checklist**
Seeing multiple errors appear when running Tabular preset= experimental_quality

**Describe the bug**
Getting the following for information  errors when running tabular with experimental quality:

Train Epoch: [0/200] Train Loss: 3.7328 Validation r2: -0.00:  49%|████▉     | 155/314 [01:21<01:23,  1.90it/s]/home/alex/PycharmProjects/algolotl-py/venv/lib/python3.10/site-packages/torch/nn/modules/loss.py:535: UserWarning:
(_ray_fit pid=133433) 
(_ray_fit pid=133433) Using a target size (torch.Size([512])) that is different to the input size (torch.Size([512, 1])). This will likely lead to incorrect results due to broadcasting. Please ensure they have the same size.
(_ray_fit pid=133433) 

(_ray_fit pid=131398) /home/alex/PycharmProjects/algolotl-py/venv/lib/python3.10/site-packages/xgboost/core.py:160: UserWarning: [repeated 6x across cluster]
(_ray_fit pid=131398) [18:22:00] WARNING: /workspace/src/common/error_msg.cc:45: `gpu_id` is deprecated since2.0.0, use `device` instead. E.g. device=cpu/cuda/cuda:0 [repeated 2x across cluster]
(_ray_fit pid=131398) [18:22:00] WARNING: /workspace/src/common/error_msg.cc:27: The tree method `gpu_hist` is deprecated since 2.0.0. To use GPU training, set the `device` parameter to CUDA instead. [repeated 3x across cluster]
(_ray_fit pid=131398)     E.g. tree_method = "hist", device = "cuda" [repeated 3x across cluster]

and 

itting model: Transformer_BAG_L5 ... Training model for up to 10495.69s of the 10495.68s of remaining time.
	Fitting 5 child models (S1F1 - S1F5) | Fitting with ParallelLocalFoldFittingStrategy (1.0 workers, per: cpus=1, gpus=1, memory=0.00%)
(_ray_fit pid=133433) /home/alex/PycharmProjects/algolotl-py/venv/lib/python3.10/site-packages/autogluon/tabular/models/tab_transformer/tab_transformer_encoder.py:85: FutureWarning:
(_ray_fit pid=133433) 
(_ray_fit pid=133433) unique with argument that is not not a Series, Index, ExtensionArray, or np.ndarray is deprecated and will raise in a future version.
(_ray_fit pid=133433) 
[...]

PR #4866, opened 2025-02-05
Fix XGBoost GPU warnings

Innixma:
*Issue #, if available:*

#4859

*Description of changes:*

Gets rid of deprecation warnings when using GPU with XGBoost, as the arguments changed in the XGBoost 2.0 release.

This will also remove support for XGBoost 1.x. We could consider keeping it by having an `if version<2:` logic, but unless XGB 1.x is important for some reason, I'll plan to just drop it for simplicity.

Note that I haven't tested this PR yet, so I don't know if it works properly. Would be good for someone with a GPU machine to sanity check it.


prateekdesai04:
@Innixma I'll try this out on a GPU instance and let you know!

prateekdesai04 (approved):

PR #4872, opened 2025-02-07
[timeseries] refactor model tests

canerturkmen:
*Issue #, if available:*

*Description of changes:*

Our model tests have lately gotten quite tangled. In this PR, we refactor the model tests package to work primarily with pytest.fixtures for dependency injection, and having all tests depend on a high level `common` module inverting their dependency. This gets rid of a lot of clutter with repeating dummy hyperparameters and model parametrizations. Moreover, it allows customizing dummy hyperparameters per model, giving tighter control on tests--without this we also can't refactor AbstractModel to tighten its interfaces.

I will follow this PR up with a new one, further organizing the tests and adding new ones for pinning down the `AbstractTimeseriesModel` functionality. 

The PR also includes some tweaks here and there to shave off some test run time, the minor adjustment in `src/` should not affect functionality.


pyproject.toml
```diff
@@ -55,4 +55,5 @@ extraPaths = [
     "common/src",
     "core/src",
     "tabular/src",
+    "timeseries/src",
```
canerturkmen, line 58: For pyright to properly work with timeseries tests.

timeseries/tests/unittests/models/test_models.py
```diff
@@ -26,649 +23,479 @@
...
-    score = model.score(DUMMY_TS_DATAFRAME, metric)
+        yield model
 
-    assert isinstance(score, float)
+    @pytest.mark.parametrize("metric", AVAILABLE_METRICS)
+    def test_when_score_called_then_scores_can_be_computed(self, trained_model, metric):
+        score = trained_model.score(DUMMY_TS_DATAFRAME, metric)
+        assert isinstance(score, float)
 
+    def test_when_val_score_accessed_then_value_is_set(self, trained_model):
```
shchur, line 76: Nit: This reads like "value is set because we access the val_score". What about something like `test_when_model_is_fit_then_val_score_is_set`?

Similarly for `test_when_predict_time_accessed_then_value_is_set`.

timeseries/tests/unittests/models/test_models.py
```diff
@@ -26,649 +23,479 @@
...
-@pytest.mark.parametrize("model_class", TESTABLE_MODELS)
-@pytest.mark.parametrize("prediction_length", [1, 5])
-def test_when_score_and_cache_oof_called_then_attributes_are_saved(model_class, prediction_length, trained_models):
-    model = trained_models[(prediction_length, repr(model_class))]
-    assert isinstance(model.val_score, float)
-    assert isinstance(model.predict_time, float)
+    def test_when_predict_time_accessed_then_value_is_set(self, trained_model):
+        assert isinstance(trained_model.predict_time, float)
 
+    def test_given_score_and_cache_oof_called_when_get_oof_predictions_called_then_oof_predictions_are_saved(
```
shchur, line 82:
```suggestion
    def test_given_score_and_cache_oof_called_when_get_oof_predictions_called_then_oof_predictions_are_available(
```

timeseries/tests/unittests/models/test_models.py
```diff
@@ -26,649 +23,479 @@
...
+    @pytest.mark.parametrize(
+        "test_data",
+        [
+            get_data_frame_with_item_index(["A", "B"], data_length=15),
+            mask_entries(get_data_frame_with_item_index(["C", "D"], data_length=60)),
+            get_data_frame_with_item_index(["A"], data_length=10),
+            get_data_frame_with_item_index([0, 1, 2, 3], data_length=15),
+        ],
+    )
+    def test_when_predict_called_then_predictor_inference_aligns_index_aligns_with_expected_index(
```
shchur, line 130: Not sure if I understand the meaning of `predictor_inference` here. Maybe `test_when_predict_called_then_forecast_index_aligns_with_expected_index`?

timeseries/tests/unittests/models/test_models.py
```diff
@@ -26,649 +23,479 @@
...
 
-        for j in DUMMY_TS_DATAFRAME.item_ids:
-            assert np.allclose(call_df.loc[j], DUMMY_TS_DATAFRAME.loc[j][:-prediction_length], equal_nan=True)
+        data = TimeSeriesDataFrame.from_iterable_dataset(
+            [{"target": [1], "start": pd.Period("2020-01-01", freq="D")} for _ in range(5)]
+        )
+        predictions = trained_model.predict(data)
+        assert len(predictions) == data.num_items * trained_model.prediction_length
+
+    def test_when_itemid_has_string_dtype_then_model_can_predict(self, trained_model):
```
shchur, line 157:
```suggestion
    def test_when_itemid_has_arrow_string_dtype_then_model_can_predict(self, trained_model):
```

timeseries/tests/unittests/models/test_models.py
```diff
@@ -26,649 +23,479 @@
...
+    """
 
+    @pytest.mark.parametrize(
+        "quantile_levels",
+        [
+            [0.1, 0.44, 0.72],
+            [0.1, 0.5, 0.9],
+        ],
+    )
+    def test_when_fit_called_then_models_train_and_returned_predictor_inference_has_mean_and_correct_quantiles(
```
shchur, line 311: Again not sure about `predictor_inference` here. Maybe `predictions` or `forecast`?

timeseries/tests/unittests/models/test_models.py
canerturkmen: changed to `value_is_returned`. 

This is a confusion I see we have in the tests often, of which I'm clearly guilty as well. Given-when-then is the build-operate-check steps of the test.

- Given refers to the dependency of the test, often the injected dependency. So it's "given a model that is fit,"
- When refers to what you jimmy in the test
- Then is the asserted criterion.

Here given is ommitted because it is rather clear from the test class now 👍

timeseries/tests/unittests/models/test_local.py
```diff
@@ -34,56 +23,24 @@
...
-    ZeroModel,
-    CrostonModel,
-    IMAPAModel,
-]
-TESTABLE_MODELS_LOCAL = SEASONAL_TESTABLE_MODELS + SEASONAL_TESTABLE_MODELS_LOCAL_ONLY + NONSEASONAL_TESTABLE_MODELS
-SEASONAL_TESTABLE_MODELS_LOCAL = SEASONAL_TESTABLE_MODELS + SEASONAL_TESTABLE_MODELS_LOCAL_ONLY
-TESTABLE_MODELS = SEASONAL_TESTABLE_MODELS + NONSEASONAL_TESTABLE_MODELS
-
-# Restrict to single core for faster training on small datasets
 DEFAULT_HYPERPARAMETERS = {"n_jobs": 1, "use_fallback_model": False}
```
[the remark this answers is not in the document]
canerturkmen: Leaving it in until we do a reorg of the test cases similar to models..

PR #4828, opened 2025-01-23
Add xdist to timeseries

canerturkmen:
*Issue #, if available:*

*Description of changes:*

Add pytest-xdist for accelerating time series tests.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #4877, opened 2025-02-10
[CI] fix timeseries platform tests

canerturkmen:
*Issue #, if available:*

*Description of changes:*

#4828 started causing time series platform tests to fail. This disables xdist in platform tests.

### DIFF
--- a/.github/workflow_scripts/test_timeseries.sh
+++ b/.github/workflow_scripts/test_timeseries.sh
@@ -15,6 +15,7 @@ python -m pip install --upgrade pytest-xdist
 
 export PYTHONHASHSEED=0  # for consistency in xdist tests
 
+cd timeseries/
 if [ "$IS_PLATFORM_TEST" = 1 ]; then
     python -m pytest --junitxml=results.xml --runslow tests  # run platform tests without multiprocessing
 else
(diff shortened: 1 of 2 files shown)


.github/workflow_scripts/test_timeseries.sh
```diff
@@ -11,12 +13,10 @@ export CUDA_VISIBLE_DEVICES=0
 install_local_packages "common/[tests]" "core/[all,tests]" "features/" "tabular/[all,tests]" "timeseries/[all,tests]"
 python -m pip install --upgrade pytest-xdist
 
-export PYTHONHASHSEED=0
+export PYTHONHASHSEED=0  # for consistency in xdist tests
 
-cd timeseries/
```
abdulfatir, line 16: `cd timeseries` is not needed now?

.github/workflows/platform_tests-command.yml
```diff
@@ -260,7 +260,7 @@ jobs:
         shell: bash -l {0}
         run: |
           conda install --channel conda-forge pygraphviz
-          chmod +x ./.github/workflow_scripts/test_timeseries.sh && ./.github/workflow_scripts/test_timeseries.sh "-m not gpu" "true"
```
abdulfatir, line 263: Previously only tests not marked as `gpu` tests were running, now we only have a flag for platform tests. Is this intended? Also, was the `"true"` supposed to be the flag for platform tests previously?

PR #4887, opened 2025-02-13
bump ray to latest version 2.42

celestinoxp:
*Issue #, if available:*

Closes: https://github.com/autogluon/autogluon/issues/4883

*Description of changes:*

### DIFF
--- a/common/src/autogluon/common/utils/try_import.py
+++ b/common/src/autogluon/common/utils/try_import.py
@@ -31,7 +31,7 @@ def try_import_mxboard():
 
 
 def try_import_ray() -> ModuleType:
-    RAY_MAX_VERSION = "2.41.0"  # sync with core/setup.py
+    RAY_MAX_VERSION = "2.43.0"  # sync with core/setup.py
     ray_max_version_os_map = dict(
         Darwin=RAY_MAX_VERSION,
         Windows=RAY_MAX_VERSION,
--- a/common/src/autogluon/common/utils/try_import.py
+++ b/common/src/autogluon/common/utils/try_import.py
@@ -37,7 +37,7 @@ def try_import_ray() -> ModuleType:
         Windows=RAY_MAX_VERSION,
         Linux=RAY_MAX_VERSION,
     )
-    ray_min_version = "2.10.0"
+    ray_min_version = "2.41.0"
     current_os = platform.system()
     ray_max_version = ray_max_version_os_map.get(current_os, RAY_MAX_VERSION)
     strict_ray_version = os.environ.get("AG_LOOSE_RAY_VERSION", "False") != "True"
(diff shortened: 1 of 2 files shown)


core/setup.py
```diff
@@ -48,11 +48,11 @@
 
 extras_require = {
     "ray": [
-        "ray[default]>=2.10.0,<2.41",  # sync with common/src/autogluon/common/utils/try_import.py
+        "ray[default]>=2.41.0,<2.43",  # sync with common/src/autogluon/common/utils/try_import.py
     ],
     "raytune": [
         "pyarrow>=15.0.0",  # cap Pyarrow to fix source installation - https://github.com/autogluon/autogluon/issues/4519
-        "ray[default,tune]>=2.10.0,<2.41",  # sync with common/src/autogluon/common/utils/try_import.py
+        "ray[default,tune]>=2.41.0,<2.43",  # sync with common/src/autogluon/common/utils/try_import.py
```
Innixma, line 55: Seems the CVE is withdrawn, unsure if that means the lower bound bump isn't necessary anymore? https://github.com/advisories/GHSA-xg2h-7cxj-3gvh

celestinoxp: I'm not sure if the cve was removed because it was a duplicate of the older cve. However, I don't know if versions up to 2.40 are still affected.

common/src/autogluon/common/utils/try_import.py
```diff
@@ -31,13 +31,13 @@ def try_import_mxboard():
...
 def try_import_ray() -> ModuleType:
-    RAY_MAX_VERSION = "2.41.0"  # sync with core/setup.py
+    RAY_MAX_VERSION = "2.43.0"  # sync with core/setup.py
     ray_max_version_os_map = dict(
         Darwin=RAY_MAX_VERSION,
         Windows=RAY_MAX_VERSION,
         Linux=RAY_MAX_VERSION,
     )
-    ray_min_version = "2.10.0"
+    ray_min_version = "2.41.0"
```
Innixma, line 40:
```suggestion
    ray_min_version = "2.10.0"
```

core/setup.py
```diff
@@ -48,11 +48,11 @@
 
 extras_require = {
     "ray": [
-        "ray[default]>=2.10.0,<2.41",  # sync with common/src/autogluon/common/utils/try_import.py
+        "ray[default]>=2.41.0,<2.43",  # sync with common/src/autogluon/common/utils/try_import.py
```
Innixma, line 51:
```suggestion
        "ray[default]>=2.10.0,<2.43",  # sync with common/src/autogluon/common/utils/try_import.py
```

core/setup.py
Innixma, line 55:
```suggestion
        "ray[default,tune]>=2.10.0,<2.43",  # sync with common/src/autogluon/common/utils/try_import.py
```

Innixma:
I took a look at the CVE, I think it is a disputed CVE that might impact all versions of ray but is considered outside the scope of ray by the developers. For now, I don't think any CVE points to the current versions we use, so I've recommended keeping the `2.10` minimum bound.

Innixma (approved):
LGTM, thanks for the contrib!

PR #4890, opened 2025-02-14
Repin accelerate to <2.0 after 1.2.0 release

timostrunk:
*Description of changes:*
Afaik the accelerate pin is the only dependency keeping numpy below 2.0, when installing all autogluon packages.

I did not run any tests yet and am lifting this pin to see what happens.


Innixma:
@tonyhoo @suzhoum can we check how this impacts us? Unsure if the version bump will be reflected in the PR CI run since it is coming from a different branch. If this is a blocker for numpy 2.0, we should prioritize seeing if this PR allows us to use numpy 2.0

FireballDWF:
As I had merged this PR into #5000, which was merged to master earlier today, this PR is already included

PR #4901, opened 2025-02-18
[tabular] Add predictor.model_hyperparameters()

Innixma:
*Issue #, if available:*

*Description of changes:*

Adding extra features to better support TabRepo 2.0:

- Add `predictor.model_hyperparameters(model)`
- Add `predictor.model_info(model)`
- Add `include_feature_metadata` boolean to `model.get_info()`, as `feature_metadata` is dependent on AutoGluon unlike all other outputs of `get_info` which are python native types.

TODO:
- [x] Add unit test for holdout
- [x] Add unit test for bagging
- [ ] (Stretch) Add `predictor.export_hyperparameters(models)` to return a usable `hyperparameters` dict that can be passed into a new `predictor.fit(... hyperparameters=hyperparameters)` call to reproduce a given AutoGluon stack architecture on new tasks.

### DIFF
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -1632,9 +1632,7 @@ class AbstractModel:
         path = self.path_root
         problem_type = self.problem_type
         eval_metric = self.eval_metric
-        hyperparameters = self._user_params.copy()
-        if self._user_params_aux:
-            hyperparameters[AG_ARGS_FIT] = self._user_params_aux.copy()
+        hyperparameters = self.get_hyperparameters_init()
 
         args = dict(
             path=path,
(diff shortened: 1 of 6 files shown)


core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -2243,6 +2255,7 @@ def get_info(self) -> dict:
             "predict_time": self.predict_time,
             "val_score": self.val_score,
             "hyperparameters": self.params,
+            "hyperparameters_user": self.get_hyperparameters_init(),
```
prateekdesai04, line 2258: Why is this needed ?

prateekdesai04 (approved):

core/src/autogluon/core/models/abstract/abstract_model.py
Innixma: This contains the exact hyperparameters specified by the user for the model. This way we can retrieve what the user specified the hyperparameters to be, which is different from what the hyperparameters actually are (since the hyperparameters of the model include the default hyperparameters for those unspecified by the user)

PR #4905, opened 2025-02-18
[tabular] Add FutureWarning for delete_models dry_run

Innixma:
*Issue #, if available:*

*Description of changes:*

Add FutureWarning for TabularPredictor.delete_models when dry_run is not specified.

The idea is that when users call `delete_models`, it is likely they would prefer for the deletion to occur rather than it being a dry run. It may be hard for users to consistently remember to set `dry_run=False`, so this PR makes it that starting in v1.4.0, dry_run will default to False instead of True.

@shchur I notice that there is no `delete_models` equivalent in TimeSeriesPredictor. Might be something interesting to have in future.


prateekdesai04 (approved):
LGTM!

PR #4909, opened 2025-02-19
[timeseries] decouple AbstractTimeSeriesModel from core

canerturkmen:
*Issue #, if available:*

*Description of changes:*

This PR contains all the key design choices for removing `AbstractTimeSeriesModel`'s tight coupling to core / AbstractModel.

Initially, `AbstractTimeSeriesModel` was tightly coupled to `AbstractModel` as both a quick solution to ensure the rest of core development does not break timeseries + to force us to replicate most of AG's core patterns in time series. It was also evident at the time that inheritance was not the correct path since the subtyping relation `AbstractTimeSeriesModel <: AbstractModel` was broken in most cases--starting from the type of data they operated on. In time, this coupling continued to cause problems and has recently made timeseries increasingly hard to develop (cf. the Chronos and Chronos-Bolt experiences, new ensembling, etc.). Our recent adventure to enable type checking on `timeseries` also yielded that this wasn't possible with the current design.

This PR proposes to change tack entirely, and only share very lightweight "interfaces" between timeseries and the rest of AG. This both gives us back the much needed control of model objects, but also relieves some other painful issues such as a circular dependency between HPO and model implementations. It segragates the high level contracts and concerns other core modules expect (see, `Tunable`) from the model behavior. 

### Description of Changes

Specifically the PR,

- Introduces a new `ModelBase` (to be renamed) abstract class that is a lightweight interface for abstract model objects in `core`. 
- Segregates some of the other concern out into other abstract mixins: `Tunable`, `Taggable`.
- Decouples `AbstractTimeSeriesModel` entirely from `AbstractModel`, moves some of the shared logic down into timeseries as these should anyway not be "de-duplicated" across modules.
- minor fix in `setup_outputdir` typing @Innixma 

The most important side effect is that the resource usage calculation logic in `AbstractModel` is dropped from time series. Indeed, our experience has been that we tripped over it more than we used it. However this point is up for discussion. I believe this is a pragmatic fix.

### Future PRs

Future PRs will continue this refactoring in multiple avenues. We will,

- Refactor `AbstractTimeSeriesModel`. It is now only decoupled, and we haven't taken care of the simplifications this decoupling would imply.
- Revisit all time series models for typing violations.
- Refactor `core.HPO` to invert its dependency on model implementations, instead depending only on an abstract `Tunable` interface (currently stubbed). Refactor the `model_trial` logic, which hides a lot of these dependencies.
- Move model objects not needed by TS into Tabular: AbstractModel, and ensemble model objects (unless these are needed by MultiModal) This is optional, as the transitive dependents of AbstractModel anyway make up most of `core`.

### Todos in this PR
- [ ] discussion
- [ ] ~format `core.hpo` after review is complete~

### DIFF
--- a/common/src/autogluon/common/utils/utils.py
+++ b/common/src/autogluon/common/utils/utils.py
@@ -27,6 +27,9 @@ LITE_MODE: bool = __lite__ is not None and __lite__
 def setup_outputdir(
     path: str | Path | None, warn_if_exist: bool = True, create_dir: bool = True, path_suffix: str | None = None
 ) -> str:
+    if isinstance(path, Path):
+        path = str(path)
+    
     is_s3_path = False
     if path:
         assert isinstance(path, (str, Path)), (
--- a/common/src/autogluon/common/utils/utils.py
+++ b/common/src/autogluon/common/utils/utils.py
@@ -25,7 +25,7 @@ LITE_MODE: bool = __lite__ is not None and __lite__
 
 
 def setup_outputdir(
-    path: str | None, warn_if_exist: bool = True, create_dir: bool = True, path_suffix: str | None = None
+    path: str | Path | None, warn_if_exist: bool = True, create_dir: bool = True, path_suffix: str | None = None
 ) -> str:
     is_s3_path = False
     if path:
--- a/common/src/autogluon/common/utils/utils.py
+++ b/common/src/autogluon/common/utils/utils.py
@@ -29,7 +29,7 @@ def setup_outputdir(
 ) -> str:
     if isinstance(path, Path):
         path = str(path)
-    
+
     is_s3_path = False
     if path:
         assert isinstance(path, (str, Path)), (
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -99,11 +99,11 @@ class Taggable(ABC):
 # TODO: refactor this class as a clean interface HPO works with. The methods below are not
 # an exhaustive set of all methods the HPO module needs!
 class Tunable(ABC):
-    def estimate_memory_usage(self) -> float | None:
+    def estimate_memory_usage(self, *args, **kwargs) -> float | None:
         """Return the estimated memory usage of the model. None if memory usage cannot be
         estimated.
         """
-        return None    
+        return None
 
     def get_minimum_resources(self, is_gpu_available: bool = False) -> Dict[str, Union[int, float]]:
         return {
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -99,21 +99,21 @@ class Taggable(ABC):
 # TODO: refactor this class as a clean interface HPO works with. The methods below are not
 # an exhaustive set of all methods the HPO module needs!
 class Tunable(ABC):
-    @property
-    def is_ensemble(self) -> bool:
-        """Return True if the model is an ensemble model or a container of multiple models."""
-        return False
-
     def estimate_memory_usage(self) -> float | None:
         """Return the estimated memory usage of the model. None if memory usage cannot be
         estimated.
         """
-        return None
+        return None    
 
     def get_minimum_resources(self, is_gpu_available: bool = False) -> Dict[str, Union[int, float]]:
         return {
             "num_cpus": 1,
         }
+
+    # TODO: remove. this is needed by hpo to determine if the model is an ensemble.
+    @abstractmethod
+    def _get_model_base(self) -> "Tunable":
+        pass
 
     @abstractmethod
     def get_params(self) -> dict:
@@ -2601,11 +2601,6 @@ class AbstractModel(ModelBase):
     @property
     def _features(self) -> List[str]:
         return self._features_internal
-
-    @property
-    def is_ensemble(self) -> bool:
-        """Return True if the model is an ensemble model or a container of multiple models."""
-        return (self._get_model_base() is self)
 
     def _get_model_base(self):
         return self
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -9,10 +9,12 @@ import os
 import pickle
 import sys
 import time
+from abc import ABC, abstractmethod
 from typing import Any, Dict, List, Optional, Tuple, Union
 
 import numpy as np
 import pandas as pd
+from typing_extensions import Self
 
 from autogluon.common.features.feature_metadata import FeatureMetadata
 from autogluon.common.space import Space
@@ -51,7 +53,119 @@ dup_filter = DuplicateFilter()
 logger.addFilter(dup_filter)
 
 
-class AbstractModel:
+class Taggable(ABC):
+    @classmethod
+    def _class_tags(cls) -> dict:
+        return _DEFAULT_CLASS_TAGS
+
+    def _more_tags(self) -> dict:
+        return _DEFAULT_TAGS
+
+    def _get_tags(self) -> dict:
+        """
+        Tags are key-value pairs assigned to an object.
+        These can be accessed after initializing an object.
+        Tags are used for identifying if an object supports certain functionality.
+        """
+        # first get class tags, which are overwritten by any object tags
+        collected_tags = self._get_class_tags()
+        for base_class in reversed(inspect.getmro(self.__class__)):
+            if hasattr(base_class, "_more_tags"):
+                # need the if because mixins might not have _more_tags
+                # but might do redundant work in estimators
+                # (i.e. calling more tags on BaseEstimator multiple times)
+                more_tags = base_class._more_tags(self)
+                collected_tags.update(more_tags)
+        return collected_tags
+
+    @classmethod
+    def _get_class_tags(cls) -> dict:
+        """
+        Class tags are tags assigned to a class that are fixed.
+        These can be accessed prior to initializing an object.
+        Tags are used for identifying if an object supports certain functionality.
+        """
+        collected_tags = {}
+        for base_class in reversed(inspect.getmro(cls)):
+            if hasattr(base_class, "_class_tags"):
+                # need the if because mixins might not have _class_tags
+                # but might do redundant work in estimators
+                # (i.e. calling more tags on BaseEstimator multiple times)
+                more_tags = base_class._class_tags()
+                collected_tags.update(more_tags)
+        return collected_tags
+
+
+# TODO: refactor this class as a clean interface HPO works with. The methods below are not
+# an exhaustive set of all methods the HPO module needs!
+class Tunable(ABC):
+    @property
+    def is_ensemble(self) -> bool:
+        """Return True if the model is an ensemble model or a container of multiple models."""
+        return False
+
+    def estimate_memory_usage(self) -> float | None:
+        """Return the estimated memory usage of the model. None if memory usage cannot be
+        estimated.
+        """
+        return None
+
+    def get_minimum_resources(self, is_gpu_available: bool = False) -> Dict[str, Union[int, float]]:
+        return {
+            "num_cpus": 1,
+        }
+
+    @abstractmethod
+    def get_params(self) -> dict:
+        """Return a clean copy of constructor parameters that can be used to
+        clone the current model.
+        """
+        pass
+        
+    @abstractmethod
+    def hyperparameter_tune(self, *args, **kwargs) -> tuple:
+        pass
+
+
+# TODO: will be renamed to AbstractModel
+class ModelBase(Taggable, Tunable, ABC):
+    def __init__(
+        self,
+        path: str | None = None,
+        name: str | None = None,
+        hyperparameters: dict[str, Any] | None = None,
+    ):
+        self.name: str
+        self.path: str
+
+    @abstractmethod
+    def rename(self, name: str) -> None:
+        pass
+
+    @abstractmethod
+    def get_info(self, *args, **kwargs) -> dict[str, Any]:
+        pass 
+
+    @abstractmethod
+    def fit(self, *args, **kwargs) -> Self:
+        pass
+
+    @abstractmethod
+    def predict(self, *args, **kwargs) -> Any:
+        pass
+
+    @abstractmethod
+    def save(self, path: str | None = None, verbose: bool = True) -> str:
+        pass
+
+    @classmethod
+    @abstractmethod
+    def load(cls, path: str, reset_paths: bool = True) -> Self:
+        pass
+
+
+# TODO: move to tabular, rename AbstractTabularModel
+class AbstractModel(ModelBase):
     """
     Abstract model implementation from which all AutoGluon models inherit.
 
@@ -2488,49 +2602,10 @@ class AbstractModel:
     def _features(self) -> List[str]:
         return self._features_internal
 
-    def _get_tags(self) -> dict:
-        """
-        Tags are key-value pairs assigned to an object.
-        These can be accessed after initializing an object.
-        Tags are used for identifying if an object supports certain functionality.
-        """
-        # first get class tags, which are overwritten by any object tags
-        collected_tags = self._get_class_tags()
-        for base_class in reversed(inspect.getmro(self.__class__)):
-            if hasattr(base_class, "_more_tags"):
-                # need the if because mixins might not have _more_tags
-                # but might do redundant work in estimators
-                # (i.e. calling more tags on BaseEstimator multiple times)
-                more_tags = base_class._more_tags(self)
-                collected_tags.update(more_tags)
-        return collected_tags
-
-    @classmethod
-    def _get_class_tags(cls) -> dict:
-        """
-        Class tags are tags assigned to a class that are fixed.
-        These can be accessed prior to initializing an object.
-        Tags are used for identifying if an object supports certain functionality.
-        """
-        collected_tags = {}
-        for base_class in reversed(inspect.getmro(cls)):
-            if hasattr(base_class, "_class_tags"):
-                # need the if because mixins might not have _class_tags
-                # but might do redundant work in estimators
-                # (i.e. calling more tags on BaseEstimator multiple times)
-                more_tags = base_class._class_tags()
-                collected_tags.update(more_tags)
-        return collected_tags
-
-    @classmethod
-    def _class_tags(cls) -> dict:
-        """
-        [Advanced] Optional tags used to communicate model capabilities to AutoML systems, such as if the model supports text features.
-        """
-        return _DEFAULT_CLASS_TAGS
-
-    def _more_tags(self) -> dict:
-        return _DEFAULT_TAGS
+    @property
+    def is_ensemble(self) -> bool:
+        """Return True if the model is an ensemble model or a container of multiple models."""
+        return (self._get_model_base() is self)
 
     def _get_model_base(self):
         return self
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -127,8 +127,8 @@ class Tunable(ABC):
         pass
 
 
-# TODO: will be renamed to AbstractModel
 class ModelBase(Taggable, Tunable, ABC):
+    @abstractmethod
     def __init__(
         self,
         path: str | None = None,
--- a/timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
+++ b/timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
@@ -219,7 +219,6 @@ class AbstractTimeSeriesModel(ModelBase):
 
         # TODO: remove the variables below
         self.model = None
-        self.problem_type = "timeseries"
 
         self._is_initialized = False
         self._user_params, self._user_params_aux = check_and_split_hyperparameters(hyperparameters)
@@ -371,7 +370,6 @@ class AbstractTimeSeriesModel(ModelBase):
         return dict(
             path=self.path_root,
             name=self.name,
-            problem_type=self.problem_type,
             eval_metric=self.eval_metric,
             hyperparameters=hyperparameters,
             freq=self.freq,
@@ -848,7 +846,7 @@ class AbstractTimeSeriesModel(ModelBase):
                 directory=self.path,
                 minimum_cpu_per_trial=minimum_cpu_per_trial,
                 minimum_gpu_per_trial=minimum_resources.get("num_gpus", 0),
-                model_estimate_memory_usage=None,  # type: ignore
+                model_estimate_memory_usage=None,
                 adapter_type="timeseries",
             )
 
--- a/timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
+++ b/timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
@@ -1,9 +1,12 @@
+from __future__ import annotations
+
+import copy
 import logging
 import os
 import re
 import time
 from contextlib import nullcontext
-from typing import Any, Callable, Dict, List, Optional, Tuple, Union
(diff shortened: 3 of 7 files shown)


common/src/autogluon/common/utils/utils.py
```diff
@@ -25,7 +25,7 @@
 
 
 def setup_outputdir(
-    path: str | None, warn_if_exist: bool = True, create_dir: bool = True, path_suffix: str | None = None
+    path: str | Path | None, warn_if_exist: bool = True, create_dir: bool = True, path_suffix: str | None = None
```
canerturkmen, line 28: @Innixma

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -51,7 +53,119 @@
 logger.addFilter(dup_filter)
 
 
-class AbstractModel:
+class Taggable(ABC):
```
shchur, line 56: I understand the reasoning behind `Tunable`, but why do we need `Taggable` to be a separate interface instead of defining `_class_tags`, `_more_tags`, etc as method of `ModelBase`?

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -51,7 +53,119 @@
...
+# TODO: will be renamed to AbstractModel
+class ModelBase(Taggable, Tunable, ABC):
+    def __init__(
+        self,
+        path: str | None = None,
+        name: str | None = None,
+        hyperparameters: dict[str, Any] | None = None,
+    ):
+        self.name: str
+        self.path: str
```
shchur, line 139: Why are we declaring these here?

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -51,7 +53,119 @@
...
+        pass
+
+    @abstractmethod
+    def save(self, path: str | None = None, verbose: bool = True) -> str:
+        pass
+
+    @classmethod
+    @abstractmethod
+    def load(cls, path: str, reset_paths: bool = True) -> Self:
+        pass
```
shchur, line 164: To better understand the intent here, what methods does `ModelBase` need to define? Are these limited to methods that are touched by `AbstractTrainer`?

Also, do I understand correctly that we intentionally do not put any implementations here, even if they are shared between Tabular and TS models? For example, creating an empty folder during `__init__()` is shared, same as the class attributes like `model_file_name ` and `model_info_name`.

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
```diff
@@ -32,9 +41,88 @@
 from .model_trial import model_trial, skip_hpo
 
 logger = logging.getLogger(__name__)
+dup_filter = DuplicateFilter()
+logger.addFilter(dup_filter)
```
shchur, line 45: Why do we need this?

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
```diff
@@ -32,9 +41,88 @@
...
+
 
+# TODO: refactor and move to util. We do not need to use "params_aux" in time series
```
shchur, line 48: Not sure if `params_aux` would be the best tool for that, but it would be nice to keep the mechanism for setting `max_time_limit` for individual models

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
```diff
@@ -67,49 +155,52 @@ class AbstractTimeSeriesModel(AbstractModel):
...
-        freq: Optional[str] = None,
-        prediction_length: int = 1,
         path: Optional[str] = None,
         name: Optional[str] = None,
+        hyperparameters: Optional[Dict[str, Any]] = None,
+        freq: Optional[str] = None,
+        prediction_length: int = 1,
         metadata: Optional[CovariateMetadata] = None,
+        target: str = "target",
+        quantile_levels: List[float] = [0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9],
```
shchur, line 179: Mutable default values can be dangerous?

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
```diff
@@ -67,49 +155,52 @@ class AbstractTimeSeriesModel(AbstractModel):
...
-            path=path,
-            name=name,
-            problem_type=None,
-            eval_metric=None,
-            hyperparameters=hyperparameters,
-        )
+        self.name = name or re.sub(r"Model$", "", self.__class__.__name__)
+
+        self.path_root = path
+        if self.path_root is None:
```
shchur, line 186: Would it make sense to keep this logic in `ModelBase` since it's shared between Tabular and TimeSeries?

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
```diff
@@ -67,49 +155,52 @@ class AbstractTimeSeriesModel(AbstractModel):
         If None, model defaults are used. This is identical to passing an empty dictionary.
     """
 
+    model_file_name = "model.pkl"
+    model_info_name = "info.pkl"
```
shchur, line 159: Should these be attributes of `ModelBase` since they are shared?

common/src/autogluon/common/utils/utils.py
Innixma: Looks good to me, but maybe we also do `path = str(path)`  inside `if path:` just so we ensure we are only working with a single type (reduces chance of bugs)

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -51,7 +53,119 @@
...
+    @abstractmethod
+    def get_params(self) -> dict:
+        """Return a clean copy of constructor parameters that can be used to
+        clone the current model.
+        """
+        pass
+        
+    @abstractmethod
+    def hyperparameter_tune(self, *args, **kwargs) -> tuple:
+        pass
```
Innixma, line 127: Curious if you think hyperparameter_tune belongs in the model class or if it should be removed and handled by something different. I've neglected HPO for a long time in AG, and whatever is implemented inside the AbstractModel class is probably not the most seamless logic.

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -2488,49 +2602,10 @@ def _ag_params(self) -> set:
...
-        [Advanced] Optional tags used to communicate model capabilities to AutoML systems, such as if the model supports text features.
-        """
-        return _DEFAULT_CLASS_TAGS
-
-    def _more_tags(self) -> dict:
-        return _DEFAULT_TAGS
+    @property
+    def is_ensemble(self) -> bool:
+        """Return True if the model is an ensemble model or a container of multiple models."""
+        return (self._get_model_base() is self)
```
Innixma, line 2608: `self._get_model_base` can take time, such as in bagged ensemble model where it could be loading from disk:

```
    def _get_model_base(self):
        if self.model_base is None:
            return self.load_model_base()
        else:
            return self.model_base
```

Better to identify `is_ensemble` via a tag. The only ones that are ensembles are weighted ensembles, bagged ensembles, and stack ensembles. All of them inherit from bagged ensemble, so we just add a tag in bagged ensemble and it should work.

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
```diff
@@ -67,49 +155,52 @@ class AbstractTimeSeriesModel(AbstractModel):
...
+            # TODO: Would be ideal to not create dir, but still track that it is unique. However, this isn't possible to do without a global list of used dirs or using UUID.
+            path_cur = setup_outputdir(path=None, create_dir=True, path_suffix=path_suffix)
+            self.path_root = path_cur.rsplit(self.name, 1)[0]
+            logger.log(20, f"Warning: No path was specified for model, defaulting to: {self.path_root}")
+
+        self.path = os.path.join(self.path_root, self.name)
+
         self.eval_metric: TimeSeriesScorer = check_get_evaluation_metric(eval_metric)
         self.eval_metric_seasonal_period = eval_metric_seasonal_period
         self.problem_type = "timeseries"
```
shchur, line 197: Do we need this attribute?

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
```diff
@@ -232,6 +381,18 @@ def get_params(self) -> dict:
             target=self.target,
         )
 
+    @classmethod
+    def load_info(cls, path: str, load_model_if_required: bool = True) -> dict:
```
shchur, line 385: Do we actually save or load info in TS models? Looking at existing trained predictors, we never save `info.pkl`.

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
```diff
@@ -334,15 +495,12 @@ def fit(  # type: ignore
...
-        kwargs = self._preprocess_fit_resources(**kwargs)
-        self.validate_fit_resources(**kwargs)
-
         self._fit(
             train_data=train_data,
             val_data=val_data,
             time_limit=time_limit,
-            **kwargs,
+            verbosity=verbosity,
+            **(self._get_system_resources() | kwargs),
```
shchur, line 503: Can we make this more explicit by replacing `kwargs` in the `fit()` signature with `num_cpus` and `num_gpus`? Currently, the meaning of this line is not very clear and requires checking a few places in the code to understand what's going on.

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
```diff
@@ -369,6 +527,31 @@ def _preprocess_time_limit(self, time_limit: float) -> float:
...
+        **kwargs,
+    ) -> None:
+        """Private method for `fit`. See `fit` for documentation of arguments. Apart from
+        the model training logic, `fit` additionally implements other logic such as keeping
+        track of the time limit, etc.
+        """
+        # TODO: Make the models respect `num_cpus` and `num_gpus` parameters
+        raise NotImplementedError
+
+    def _check_fit_params(self):
```
shchur, line 547: Let's add a TODO to perform this check inside `AbstractTimeSeriesModel.fit()` instead of calling it explicitly in every subclass (like [here](https://github.com/autogluon/autogluon/blob/master/timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py#L484))

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
```diff
@@ -764,6 +939,32 @@ def convert_to_refit_full_via_copy(self) -> Self:
...
+        if "hyperparameters" not in params:
+            params["hyperparameters"] = dict()
+
+        if AG_ARGS_FIT not in params["hyperparameters"]:
+            params["hyperparameters"][AG_ARGS_FIT] = dict()
+
+        # Increase memory limit by 25% to avoid memory restrictions during fit
+        params["hyperparameters"][AG_ARGS_FIT]["max_memory_usage_ratio"] = (
+            params["hyperparameters"][AG_ARGS_FIT].get("max_memory_usage_ratio", 1.0) * 1.25
+        )
```
shchur, line 960: TS models do not use `max_memory_usage_ratio` [here](https://github.com/autogluon/autogluon/pull/4909/files#diff-72dcfbecb24aa875b49f6e776b10cbe58ce394cb5f13f884d7473caa89e6e168L183). Do we still need to keep that here?

core/src/autogluon/core/models/abstract/abstract_model.py
canerturkmen: In this case, for reuse. If we can reasonably imagine non-`ModelBase` objects as being taggable one day, it makes sense to me to separate them now.

core/src/autogluon/core/models/abstract/abstract_model.py
canerturkmen: Not really the interface of our models trainers depend on (although that too would be a great justification, I agree) but to fix the high level policy on what we call a model. It's Taggable, Tunable, Serializable (which we could make another abstract class on the same grounds as Taggable) implements _some_ methods called `fit` and `predict`  with certain return types. They have some other convenience methods which the AbstractTrainer depends on. Also they have a path and a name. 

Using ABC instead of the `NotImplementedError` idiom we now strictly enforce this high-level policy at runtime. 

One could imagine also separating out `Trainable`, although our `AbstractTrainer` isn't really opinionated about models, and anyway it makes more sense to wire things together in the timeseries module itself.

And yes, we very purposely do not put behavior here. De-duping code in this instance is not an absolute good: it doesn't stand for "I'm keeping DRY and doing some reuse" but "I want all modules in AG to play within these constraints." Unless these are prescribed by equally high-level policies (other modules in core everyone depends on) I don't see a reason to do this coupling. What duping we have is likely [accidental](https://stackoverflow.com/questions/54542274/single-responsibility-principle-without-code-duplication-how-to-do) and we'll anyway refactor them and they will be gone in a few weeks 🙏

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
canerturkmen: carried over from the original `AbstractModel`. Will probably be gone when I touch the model.

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
canerturkmen: Noted. Now that `params_aux` is not really needed, we can also think of other things lurking as hyperparameters and constructor arguments and pull them into an immutable dataclass..

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
canerturkmen: Yes, thanks for flagging. Changing to 
```
quantile_levels: Sequence[float] = (0.1, ..., 0.9)
```

and coercing to list inside. the constructor.

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
canerturkmen: For reasons flagged above, I would keep them here. Again, they are not shared in the sense that "we definitely want to do this like this and when tabular makes a change we want to inherit it"

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
canerturkmen: cf. above

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
canerturkmen: haven't really started touching the model yet, just separated out. But I moved this one below to `TODO: remove` 👍

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
canerturkmen: will refactor.

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
canerturkmen: I agree. In a future PR, I'll chop away through all the `**kwargs` in timeseries and see what happens :)

core/src/autogluon/core/models/abstract/abstract_model.py
canerturkmen: So that if any subclass of `ModelBase` decides to change their mind about the type of `path` or `name`, pyright will get angry.

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -51,7 +53,119 @@
...
+
+
+# TODO: will be renamed to AbstractModel
+class ModelBase(Taggable, Tunable, ABC):
+    def __init__(
+        self,
+        path: str | None = None,
+        name: str | None = None,
+        hyperparameters: dict[str, Any] | None = None,
+    ):
```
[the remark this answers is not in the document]
canerturkmen: In this case to establish the contract: models will always have their path, name, and hyperparameters injected with these types. Any violation will be caught by the type checker. If your question is more general, as in why do we even need to do enforce this, we can discuss.

core/src/autogluon/core/models/abstract/abstract_model.py
canerturkmen: It should indeed be handled by something different imho. This is also evident in the way we call the `HpoExecutor`, where we pass in the `model_trial` callables. In other words, the models have to know too much: they don't only know how to fit an ML model, but how to cook up a callable that does a full training run of "itself" and also knows how to pass this into the HpoExecutor. I'll try to take things in a more flexible direction in timeseries and we can see where that leads us 👍 .

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
canerturkmen: added a todo 👍

core/src/autogluon/core/models/abstract/abstract_model.py
canerturkmen: Thanks for flagging this, I reverted the changes to the hpo module and left a todo in the `Tunable`

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
canerturkmen: removed entirely.

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
```diff
@@ -67,49 +155,51 @@ class AbstractTimeSeriesModel(AbstractModel):
...
-        )
+        self.name = name or re.sub(r"Model$", "", self.__class__.__name__)
+
+        self.path_root = path
+        if self.path_root is None:
+            path_suffix = self.name
+            # TODO: Would be ideal to not create dir, but still track that it is unique. However, this isn't possible to do without a global list of used dirs or using UUID.
+            path_cur = setup_outputdir(path=None, create_dir=True, path_suffix=path_suffix)
+            self.path_root = path_cur.rsplit(self.name, 1)[0]
+            logger.log(20, f"Warning: No path was specified for model, defaulting to: {self.path_root}")
```
shchur, line 191: I know that this code is copy-pasted from `AbstractModel`, but do we actually need to create the folder at model initialization? That seems like a rather strange side-effect of model creation, imo it would make a bit more sense to do only create a folder during `save` (if no folder exists).

As a side note, I also don't think that we ever create a TS model with `path=None` during normal usage of AG-TS (this only happens during unit tests).

Not sure what the best resolution would be, just adding a TODO here would be fine.

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
```diff
@@ -191,27 +320,46 @@ def _get_default_auxiliary_params(self) -> dict:
...
+
+    def _init_params(self):
+        """Initializes model hyperparameters"""
+        hyperparameters = self._user_params
+        self.nondefault_params = []
+        if hyperparameters is not None:
+            self.params.update(hyperparameters)
+            self.nondefault_params = list(hyperparameters.keys())[
+                :
+            ]  # These are hyperparameters that user has specified.
```
shchur, line 336: Nanonit: Can we move the comment above? The formatting hurts my eyes a bit.
```suggestion
            # These are hyperparameters that user has specified.
            self.nondefault_params = list(hyperparameters.keys())[:]  
```

shchur (approved):
LGTM!

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
canerturkmen: noted

core/src/autogluon/core/models/abstract/abstract_model.py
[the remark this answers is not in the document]
Innixma: Interesting. Is there a PEP / best practice that I can refer to for this? Normally I assume that doing this will make the IDE complain that we aren't calling `super().__init__()` in the subclass, even though we shouldn't be calling it since this does nothing.

I agree with the intent making sense, my question is more about the nuances of Python.

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
Innixma: @shchur @canerturkmen I believe the reason I create it during init is because of a few legacy reasons:

1. A model might create intermediate artifacts, and I wanted the artifacts to save without issue (such as if the artifacts would crash during save if no folder exists) -> This is probably not necessary, as we can simply ensure that if artifacts are created during fit, we ensure that it works properly, lazily creating the folder only when it is needed.
2. To ensure a warning if the folder already exists is logged at the appropriate time. -> This can be resolved without creating the folder by having a flag in the model at init that checks if the folder already exists and using that for the logging (but doesn't create it if it doesn't).

There might have been other reasons, but this setup_outputdir logic goes all the way back to 2018.

If we can avoid the folder being created at init, that would be nice.

Innixma (approved):
Looks great, ship it!

+ Added a comment regarding the creation of directory on model init.

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
Innixma: Also, the bulk of intermediate artifacts used to be the checkpoints of neural network models. However I've since optimized this and the checkpoints are stored in-memory, so I don't think it is very common for models to have disk artifacts prior to saving.

This would actually be a really good unit test to check if a model saves anything before calling `model.save()` (by checking if the directory exists)

Innixma: Created a GitHub issue to track: https://github.com/autogluon/autogluon/issues/4915

core/src/autogluon/core/models/abstract/abstract_model.py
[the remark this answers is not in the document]
canerturkmen: Actually, you are right. Going headfirst into Python nuances:
- Looks like `pylint` won't complain `super-init-not-called` if the `ModelBase.__init__` is also marked an abstractmethod 🤷 
- And turns out -> type checkers do not enforce substitutability for constructor arguments -- since constructors are not part of the object's runtime behavior. But I also don't know a better way to signal this. :shrug:

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
canerturkmen: This goes into the "what is the lifecycle of a model object?" discussion. In timeseries now that's going to be super simplified: no lazy init, not much state, etc. So we can actually get away with not saving until saving. I wouldn't be quite so sure about tabular though.

Innixma: "no lazy init" is an interesting departure from tabular. I'd be curious to see how it works out. I forget all the reasons why I do it in tabular, but I do believe there are some good reasons, at least with the existing design. Lazy init definitely makes then code more complex though, no doubt.

PR #4913, opened 2025-02-19
[tabular] Add ag_model_register

Innixma:
*Issue #, if available:*

*Description of changes:*

- Add `ag_key`, `ag_name` and `ag_priority` class variables to all AbstractModel subclasses.
- Add ag_model_register and ModelRegister to simplify model book keeping and enable registering custom models easily.
- Add 134 unit tests for ModelRegister to ensure matching functionality with mainline.
- This logic will be extremely helpful with making TabRepo 2.0 adoption lightweight.
- Add name for DummyModel -> "Dummy", made relevant unit test changes. Previously it was the only unnamed model.

WIP:
- [ ] Replace logic in `autogluon.tabular.trainer.model_presets.presets` to use `ag_model_register`. I decided not to do that in this PR as it would expand the scope to the point of being hard to review. So I will do this as a follow-up PR.

## Example:

```python
import pandas as pd

from autogluon.core.models import AbstractModel
from autogluon.tabular.register import ag_model_register


class MyCustomModel(AbstractModel):
    ag_key = "MYMODEL"
    ag_name = "MyModel"
    ag_priority = 70


ag_model_register.add(MyCustomModel)
print(f"Class mapped to 'MYMODEL' key: {ag_model_register.key_to_cls(key='MYMODEL')}")

print("Contents of ag_model_register:")
with pd.option_context("display.max_rows", None, "display.max_columns", None, "display.width", 1000):
    print(ag_model_register.to_frame())

```

### Output:

```
Class mapped to 'MYMODEL' key: <class '__main__.MyCustomModel'>
Contents of ag_model_register:
                                       model_cls                    ag_name ag_priority
ag_key                                                                                 
RF                                       RFModel               RandomForest          80
XT                                       XTModel                 ExtraTrees          60
KNN                                     KNNModel                 KNeighbors         100
GBM                                     LGBModel                   LightGBM          90
CAT                                CatBoostModel                   CatBoost          70
XGB                                 XGBoostModel                    XGBoost          40
NN_TORCH              TabularNeuralNetTorchModel             NeuralNetTorch          25
LR                                   LinearModel                LinearModel          30
FASTAI                      NNFastAiTabularModel            NeuralNetFastAI          50
TRANSF                       TabTransformerModel                Transformer           0
AG_TEXT_NN                    TextPredictorModel              TextPredictor           0
AG_IMAGE_NN                  ImagePredictorModel             ImagePredictor           0
AG_AUTOMM               MultiModalPredictorModel        MultiModalPredictor           0
FT_TRANSFORMER                FTTransformerModel              FTTransformer           0
TABPFN                               TabPFNModel                     TabPFN         110
TABPFNMIX                         TabPFNMixModel                  TabPFNMix          45
FASTTEXT                           FastTextModel                   FastText           0
VW                             VowpalWabbitModel               VowpalWabbit          10
ENS_WEIGHTED         GreedyWeightedEnsembleModel           WeightedEnsemble           0
SIMPLE_ENS_WEIGHTED  SimpleWeightedEnsembleModel           WeightedEnsemble           0
IM_RULEFIT                          RuleFitModel                    RuleFit           0
IM_GREEDYTREE                    GreedyTreeModel                 GreedyTree           0
IM_FIGS                                FigsModel                       Figs           0
IM_HSTREE                            HSTreeModel  HierarchicalShrinkageTree           0
IM_BOOSTEDRULES                BoostedRulesModel               BoostedRules           0
DUMMY                                 DummyModel                      Dummy           0
MYMODEL                            MyCustomModel                    MyModel          70
```

### DIFF
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -1022,10 +1022,7 @@ class AbstractModel(ModelBase):
         **kwargs :
             Any additional fit arguments a model supports.
         """
-        if "time_limit" in kwargs and kwargs["time_limit"] is not None:
-            time_start = time.time()
-        else:
-            time_start = None
+        time_start = time.time()
         kwargs = self.initialize(
             **kwargs
         )  # FIXME: This might have to go before self._preprocess_fit_args, but then time_limit might be incorrect in **kwargs init to initialize
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -205,7 +205,7 @@ class AbstractModel(ModelBase):
     ag_key: str | None = None  # set to string value for subclasses for use in AutoGluon
     ag_name: str | None = None  # set to string value for subclasses for use in AutoGluon
     ag_priority: int = 0  # set to int value for subclasses for use in AutoGluon
-    ag_priority_by_problem_type: dict[str, int] = {}
+    ag_priority_by_problem_type: dict[str, int] = {}  # if not set, we fall back to ag_priority
 
     model_file_name = "model.pkl"
     model_info_name = "info.pkl"
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -10,6 +10,7 @@ import pickle
 import sys
 import time
 from abc import ABC, abstractmethod
+from types import MappingProxyType
 from typing import Any, Dict, List, Optional, Tuple, Union
 
 import numpy as np
@@ -121,7 +122,7 @@ class Tunable(ABC):
         clone the current model.
         """
         pass
-        
+
     @abstractmethod
     def hyperparameter_tune(self, *args, **kwargs) -> tuple:
         pass
@@ -144,7 +145,7 @@ class ModelBase(Taggable, Tunable, ABC):
 
     @abstractmethod
     def get_info(self, *args, **kwargs) -> dict[str, Any]:
-        pass 
+        pass
 
     @abstractmethod
     def fit(self, *args, **kwargs) -> Self:
@@ -205,7 +206,7 @@ class AbstractModel(ModelBase):
     ag_key: str | None = None  # set to string value for subclasses for use in AutoGluon
     ag_name: str | None = None  # set to string value for subclasses for use in AutoGluon
     ag_priority: int = 0  # set to int value for subclasses for use in AutoGluon
-    ag_priority_by_problem_type: dict[str, int] = {}  # if not set, we fall back to ag_priority
+    ag_priority_by_problem_type: dict[str, int] = MappingProxyType({})  # if not set, we fall back to ag_priority. Use MappingProxyType to avoid mutation.
 
     model_file_name = "model.pkl"
     model_info_name = "info.pkl"
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -202,6 +202,10 @@ class AbstractModel(ModelBase):
         Hyperparameters that will be used by the model (can be search spaces instead of fixed values).
         If None, model defaults are used. This is identical to passing an empty dictionary.
     """
+    ag_key: str | None = None  # set to string value for subclasses for use in AutoGluon
+    ag_name: str | None = None  # set to string value for subclasses for use in AutoGluon
+    ag_priority: int = 0  # set to int value for subclasses for use in AutoGluon
+    ag_priority_by_problem_type: dict[str, int] = {}
 
     model_file_name = "model.pkl"
     model_info_name = "info.pkl"
@@ -2624,3 +2628,14 @@ class AbstractModel(ModelBase):
     def fit_num_gpus_child(self) -> float:
         """Number of GPUs used for fitting one model (i.e. a child model)"""
         return self.fit_num_gpus
+
+    @classmethod
+    def get_ag_priority(cls, problem_type: str | None = None) -> int:
+        """
+        Returns the AutoGluon fit priority,
+        defined by `cls.ag_priority` and `cls.ag_priority_by_problem_type`.
+        """
+        if problem_type is None:
+            return cls.ag_priority
+        else:
+            return cls.ag_priority_by_problem_type.get(problem_type, cls.ag_priority)
--- a/tabular/src/autogluon/tabular/models/rf/rf_model.py
+++ b/tabular/src/autogluon/tabular/models/rf/rf_model.py
@@ -30,9 +30,6 @@ class RFModel(AbstractModel):
     ag_key = "RF"
     ag_name = "RandomForest"
     ag_priority = 80
-    ag_priority_by_problem_type = {
-        SOFTCLASS: 80
-    }
 
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
--- a/tabular/src/autogluon/tabular/models/rf/rf_model.py
+++ b/tabular/src/autogluon/tabular/models/rf/rf_model.py
@@ -27,6 +27,12 @@ class RFModel(AbstractModel):
     """
     Random Forest model (scikit-learn): https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.RandomForestClassifier.html
     """
+    ag_key = "RF"
+    ag_name = "RandomForest"
+    ag_priority = 80
+    ag_priority_by_problem_type = {
+        SOFTCLASS: 80
+    }
 
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
--- a/tabular/src/autogluon/tabular/models/xt/xt_model.py
+++ b/tabular/src/autogluon/tabular/models/xt/xt_model.py
@@ -10,7 +10,6 @@ class XTModel(RFModel):
     ag_key = "XT"
     ag_name = "ExtraTrees"
     ag_priority = 60
-    ag_priority_by_problem_type = {}
 
     def _get_model_type(self):
         if self.problem_type == REGRESSION:
--- a/tabular/src/autogluon/tabular/models/xt/xt_model.py
+++ b/tabular/src/autogluon/tabular/models/xt/xt_model.py
@@ -7,6 +7,10 @@ class XTModel(RFModel):
     """
     Extra Trees model (scikit-learn): https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.ExtraTreesClassifier.html#sklearn.ensemble.ExtraTreesClassifier
     """
+    ag_key = "XT"
+    ag_name = "ExtraTrees"
+    ag_priority = 60
+    ag_priority_by_problem_type = {}
 
     def _get_model_type(self):
         if self.problem_type == REGRESSION:
--- a/tabular/src/autogluon/tabular/register/_ag_model_register.py
+++ b/tabular/src/autogluon/tabular/register/_ag_model_register.py
@@ -1,0 +1,66 @@
+from autogluon.core.models import (
+    DummyModel,
+    GreedyWeightedEnsembleModel,
+    SimpleWeightedEnsembleModel,
+)
+
+from . import ModelRegister
+from ..models import (
+    BoostedRulesModel,
+    CatBoostModel,
+    FastTextModel,
+    FigsModel,
+    FTTransformerModel,
+    GreedyTreeModel,
+    HSTreeModel,
+    ImagePredictorModel,
+    KNNModel,
+    LGBModel,
+    LinearModel,
+    MultiModalPredictorModel,
+    NNFastAiTabularModel,
+    RFModel,
+    RuleFitModel,
+    TabPFNMixModel,
+    TabPFNModel,
+    TabularNeuralNetTorchModel,
+    TextPredictorModel,
+    VowpalWabbitModel,
+    XGBoostModel,
+    XTModel,
+)
+from ..models.tab_transformer.tab_transformer_model import TabTransformerModel
+
+
+# When adding a new model officially to AutoGluon, the model class should be added to the bottom of this list.
+REGISTERED_MODEL_CLS_LST = [
+    RFModel,
+    XTModel,
+    KNNModel,
+    LGBModel,
+    CatBoostModel,
+    XGBoostModel,
+    TabularNeuralNetTorchModel,
+    LinearModel,
+    NNFastAiTabularModel,
+    TabTransformerModel,
+    TextPredictorModel,
+    ImagePredictorModel,
+    MultiModalPredictorModel,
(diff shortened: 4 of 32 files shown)


Innixma:
@shchur This logic might be applicable to TimeSeries as well. Curious on your thoughts.

LennartPurucker:
Nit: How far is the priority of the model class connected or disconnected from the priority of a model configuration? 
Would it be clearer to name this the "default" priority?

LennartPurucker (approved):
LGTM!

Innixma:
> Nit: How far is the `ag_priority` of the model class connected or disconnected from the priority of a model configuration? Would it be clearer to name this the `ag_default_priority`?

Model configs [can have specified priorities](https://github.com/autogluon/autogluon/blob/master/tabular/src/autogluon/tabular/configs/zeroshot/zeroshot_portfolio_2023.py) via `"ag_args": {"priority": N}`, as can the model name.

While it is true that this is only the default, it would be overly verbose to put "default" in the name IMO, since the same argument could be said for `ag_name` -> `ag_default_name` just feels a bit clunky.

LennartPurucker:
Agree. Maybe then a comment somewhere clarifying this to users that have to set this value

tabular/src/autogluon/tabular/models/rf/rf_model.py
```diff
@@ -27,6 +27,12 @@ class RFModel(AbstractModel):
     """
     Random Forest model (scikit-learn): https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.RandomForestClassifier.html
     """
+    ag_key = "RF"
+    ag_name = "RandomForest"
+    ag_priority = 80
+    ag_priority_by_problem_type = {
+        SOFTCLASS: 80
+    }
```
shchur, line 35: Is this a typo?

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -88,6 +88,10 @@ class AbstractModel:
         Hyperparameters that will be used by the model (can be search spaces instead of fixed values).
         If None, model defaults are used. This is identical to passing an empty dictionary.
     """
+    ag_key: str | None = None  # set to string value for subclasses for use in AutoGluon
+    ag_name: str | None = None  # set to string value for subclasses for use in AutoGluon
+    ag_priority: int = 0  # set to int value for subclasses for use in AutoGluon
+    ag_priority_by_problem_type: dict[str, int] = {}  # if not set, we fall back to ag_priority
```
shchur, line 94: Nit: Setting a mutable object as the default value might lead to some problems in the future.

tabular/src/autogluon/tabular/models/xt/xt_model.py
```diff
@@ -7,6 +7,10 @@ class XTModel(RFModel):
     """
     Extra Trees model (scikit-learn): https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.ExtraTreesClassifier.html#sklearn.ensemble.ExtraTreesClassifier
     """
+    ag_key = "XT"
+    ag_name = "ExtraTrees"
+    ag_priority = 60
+    ag_priority_by_problem_type = {}
```
shchur, line 14: Is this a typo?

tabular/src/autogluon/tabular/register/_model_register.py
```diff
@@ -0,0 +1,157 @@
+from __future__ import annotations
+
+from typing import Type
+
+import pandas as pd
+
+from autogluon.core.models import AbstractModel
+
+
+# TODO: Move to core? Maybe TimeSeries can reuse?
```
shchur, line 10: In TimeSeries, we don't have model aliases. The name that needs to be specified by the user is the same as the model class (e.g., `hyperparameters={"DeepAR": {}}` to use the model class `DeepARModel`). I guess it makes sense to have a register to keep track of different available models, but we probably don't need to rely on the same implementation given the different requirements.

tabular/src/autogluon/tabular/register/_ag_model_register.py
```diff
@@ -0,0 +1,66 @@
...
+    RuleFitModel,
+    GreedyTreeModel,
+    FigsModel,
+    HSTreeModel,
+    BoostedRulesModel,
+    DummyModel,
+]
+
+# TODO: Replace logic in `autogluon.tabular.trainer.model_presets.presets` with `ag_model_register`
+ag_model_register = ModelRegister(model_cls_list=REGISTERED_MODEL_CLS_LST)
```
shchur, line 66: A potentially less verbose way to implement this pattern is using a decorator.

```python
# in _model_register.py
ALIAS_TO_CLS: dict[str, type] = {}

def register(alias: str):
    def decorator(cls):
        ALIAS_TO_CLS[alias] = cls
        return cls
    return decorator

# in lgb_model.py
@register(alias="LGB")
class LGBModel(AbstractTimeSeriesModel):
    # ...
```
This could also be easier to use by the users that implement custom models.

`ALIAS_TO_CLS` can be replaced with something more advanced like the actual `ModelRegistry` class, or a dict where each alias corresponds to a tuple `[model_cls, priority, priority_dict]`.

tabular/src/autogluon/tabular/register/_model_register.py
```diff
@@ -0,0 +1,157 @@
...
+                f"No registered model exists with provided key: {key}"
+                f"\n\tValid keys: {list(self.key_to_cls_map().keys())}"
+            )
+        return self.key_to_cls_map()[key]
+
+    def priority_map(self, problem_type: str | None = None) -> dict[Type[AbstractModel], int]:
+        return {model_cls: self.priority(model_cls, problem_type=problem_type) for model_cls in self._model_cls_list}
+
+    def key_to_priority_map(self, problem_type: str | None = None) -> dict[str, int]:
+        return {self.key(model_cls): self.priority(model_cls, problem_type=problem_type) for model_cls in self._model_cls_list}
```
shchur, line 112: It is convenient to have both the `alias -> *` and `cls -> *` mappings available here, but I think that for maintainability, it might actually be better to only keep one (either class or alias). This way future users / contributors won't need to guess which of these needs to be used, and there will be less code to be maintained.

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -88,6 +88,10 @@ class AbstractModel:
         Hyperparameters that will be used by the model (can be search spaces instead of fixed values).
         If None, model defaults are used. This is identical to passing an empty dictionary.
     """
+    ag_key: str | None = None  # set to string value for subclasses for use in AutoGluon
+    ag_name: str | None = None  # set to string value for subclasses for use in AutoGluon
```
shchur, line 92: I feel that both `ag_key` and `ag_name` serve a similar purpose, and one of them could potentially be removed. For example, we could generate the `ag_name` automatically from the class name with something like `self.name = re.sub(r"Model$", "", self.__class__.__name__)`.

shchur (commented):
Looks good to me, just several tangential thoughts / ideas and a few nits

canerturkmen:
> @shchur This logic might be applicable to TimeSeries as well. Curious on your thoughts.

I was about to start doing this.

canerturkmen:
Another common [way](https://charlesreid1.github.io/python-patterns-the-registry.html) to do this in Python. cf. [here](https://github.com/amazon-science/chronos-forecasting/blob/94e20ea7e510ac4d665492b8bed8836a5143f16e/src/chronos/base.py#L25).

i.e., you don't even need a decorator. anyone who inherits from your Model class will be registered. You can still have a decorator if this is important though.

canerturkmen:
Another design consideration is if "priority" is an intrinsic property of the model itself, or is it something the Trainer assigns to the model?

core/src/autogluon/core/models/abstract/abstract_model.py
Innixma: Any thoughts on how to avoid this? We could move it to a property instead, but it would look less elegant.

Innixma:
> Another design consideration is if "priority" is an intrinsic property of the model itself, or is it something the Trainer assigns to the model?

@canerturkmen This is a good question. For now I think it can be a value that is overwritten by the trainer if needed, but the reason I have it as part of the model is so that a model contribution would only require the contributor to make edits to the model file and nowhere else. Also, this means someone can bring their custom model to AutoGluon and easily specify the priority relative to other models when fitting.

Luckily this logic is rather easy to change later and deprecate if needed.

Innixma:
> Another common [way](https://charlesreid1.github.io/python-patterns-the-registry.html) to do this in Python. cf. [here](https://github.com/amazon-science/chronos-forecasting/blob/94e20ea7e510ac4d665492b8bed8836a5143f16e/src/chronos/base.py#L25).
> 
> i.e., you don't even need a decorator. anyone who inherits from your Model class will be registered. You can still have a decorator if this is important though.

@canerturkmen Do you have unit tests / example code showing what you can do with this?

One thing I might be wary of with automatic register on import is that I want to ensure no duplicate keys, and also it becomes a bit weird in some ways because whether the user's code works or not will now depend on what has been imported elsewhere, something the IDE wouldn't track and could make script sharing confusing. Just a hypothetical though, I don't know how real this concern is.

core/src/autogluon/core/models/abstract/abstract_model.py
Innixma: Agreed that they serve a similar purpose. For now I will keep both for legacy reasons, but we can consider standardizing. Maybe we can have `self.ag_name` default to your regix logic if it is `None`? Such as `self._ag_name = None` with def ag_name()` as a property?

Innixma: Note that AG Tabular doesn't have the same orderly name structure of the classes as TimeSeries (but I probably should do it soon...):

For example:

RFModel -> RandomForest
XTModel -> ExtraTrees
LGBModel -> LightGBM

tabular/src/autogluon/tabular/models/rf/rf_model.py
Innixma: Not a typo, but it ends up being redundant. This logic was here to perfectly replicate the old code, but since 80 is also the default, I've removed this now.

tabular/src/autogluon/tabular/models/xt/xt_model.py
Innixma: Not a typo, to avoid it inheriting RandomForest's ag_priority_by_problem_type, but since the RF value as redundant and I removed it, I will also remove this.

tabular/src/autogluon/tabular/register/_model_register.py
Innixma: fair, I've removed the key_to_priority and key_to_priority_map functions

tabular/src/autogluon/tabular/register/_ag_model_register.py
Innixma: One possible downside is this links all models to a global object inherently, whereas the ModelRegister is something that isn't inherently global and theoretically a user could pass their own ModelRegister object to the predictor. No idea if that is something that is desirable though.

canerturkmen:
> > Another common [way](https://charlesreid1.github.io/python-patterns-the-registry.html) to do this in Python. cf. [here](https://github.com/amazon-science/chronos-forecasting/blob/94e20ea7e510ac4d665492b8bed8836a5143f16e/src/chronos/base.py#L25).
> > i.e., you don't even need a decorator. anyone who inherits from your Model class will be registered. You can still have a decorator if this is important though.
> 
> @canerturkmen Do you have unit tests / example code showing what you can do with this?
> 
> One thing I might be wary of with automatic register on import is that I want to ensure no duplicate keys, and also it becomes a bit weird in some ways because whether the user's code works or not will now depend on what has been imported elsewhere, something the IDE wouldn't track and could make script sharing confusing. Just a hypothetical though, I don't know how real this concern is.

[This](https://github.com/amazon-science/chronos-forecasting/blob/94e20ea7e510ac4d665492b8bed8836a5143f16e/src/chronos/base.py#L156) is the caller. 

You could de-duplicate keys in the `__new__` method, and raise an informative exception if there is a clash. The IDE will know where to find the exception. Other than that, I think it's a usability issue, I guess an argument could be made it's "too magic" but I would beg to differ. It's just one extra thing to remember and one extra place to touch when adding a new model. 

I'm also pretty surprised this is automerging. I touched abstract_model quite a bit and it might be worth a rebase. 

Other random consideration: I would pull it up in the namespace if you would like it to be used often and it's a part of the public API: `autogluon.tabular.model_register`, etc.

core/src/autogluon/core/models/abstract/abstract_model.py
shchur: They only alternative idea that comes to mind is turning this into an instance attribute and calling `copy` during `__init__`. This also feels less elegant though.

Innixma: MappingProxyType looks reasonable, I've updated the PR to use it

Innixma:
@canerturkmen re namespace: The example I had was outdated, here is the current path:

```
from autogluon.tabular.register import ag_model_register
```

I could maybe move it to
```
from autogluon.tabular import ag_model_register
```
but will wait until I figure out how exactly I want users to interact with the register in the follow-up PR.

Innixma:
> [This](https://github.com/amazon-science/chronos-forecasting/blob/94e20ea7e510ac4d665492b8bed8836a5143f16e/src/chronos/base.py#L156) is the caller.
> 
> You could de-duplicate keys in the __new__ method, and raise an informative exception if there is a clash. The IDE will know where to find the exception. Other than that, I think it's a usability issue, I guess an argument could be made it's "too magic" but I would beg to differ. It's just one extra thing to remember and one extra place to touch when adding a new model.
> 
> I'm also pretty surprised this is automerging. I touched abstract_model quite a bit and it might be worth a rebase.
> 
> Other random consideration: I would pull it up in the namespace if you would like it to be used often and it's a part of the public API: autogluon.tabular.model_register, etc.

This is a solid option, the only concern really is the "magic" aspect and oddities related to seemingly unused imports actually impacting whether a script works or not. I'll keep as is for now but we can always change it to be auto-add in a follow-up PR if we decide it is ultimately more user friendly.

The following situation is what I'm not a huge fan of:

```
# Fails due to not knowing `CUSTOM_MODEL_KEY`

predictor = TabularPredictor(label).fit(..., hyperparameters={"CUSTOM_MODEL_KEY": {...}})
```

```
# Succeeds because of the "unused" AGCustomModel import, which adds AGCustomModel to the registry

from ag_custom_model import AGCustomModel

predictor = TabularPredictor(label).fit(..., hyperparameters={"CUSTOM_MODEL_KEY": {...}})
```

canerturkmen (approved):
LGTM!

tabular/src/autogluon/tabular/register/_ag_model_register.py
Innixma: merging the current logic as-is, we can re-visit this alternative if we think it is cleaner in future.

PR #4925, opened 2025-02-24
Log a wraning if mlflow autologging is enabled

### LINKED ISSUE #4914: [BUG] TimeSeriesPredictor problems training tree-based models RecursiveTabular, DirectTabular and ChronosFineTuned[bolt_small]
**Bug Report Checklist**

**Describe the bug**
I have used the TimeSeriesPredictor with the following configurations. The eval_metric is a build-in one: 

```
predictor = TimeSeriesPredictor(
          prediction_length=predict_params["horizon"],
          target="ist",
          eval_metric="MAE",
          known_covariates_names=predict_params["future_covariates"]
      ).fit(train_data, hyperparameters=hyperparameters, presets="fast_training", enable_ensemble=True)
```
But for the tree-based models RecursiveTabular and DirectTabular as well as for the  ChronosFineTuned[bolt_small] I haved received the error that the function  mean_absolute_error cannot be pickled.

```
Models that will be trained: ['RecursiveTabular', 'DirectTabular']
Training timeseries model RecursiveTabular. 
	Warning: Exception caused RecursiveTabular to fail during training... Skipping this model.
	Can't pickle <function mean_absolute_error at 0x7fe1e8a21da0>: it's not the same object as sklearn.metrics._regression.mean_absolute_error
Training timeseries model DirectTabular. 
INFO:py4j.clientserver:Received command c on object id p0
	Warning: Exception caused DirectTabular to fail during training... Skipping this model.
	Can't pickle <function mean_absolute_error at 0x7fe1e8a21da0>: it's not the same object as sklearn.metrics._regression.mean_absolute_error

Saving fine-tuned model to /Workspace/Users/src/AutogluonModels/ag-20250220_211739/models/ChronosFineTuned[bolt_small]/W0/fine-tuned-ckpt
	-0.0787       = Validation score (-MAE)
	154.17  s     = Training runtime
	0.05    s     = Validation (prediction) runtime
	Warning: Exception caused ChronosFineTuned[bolt_small] to fail during training... Skipping this model.
	Can't pickle <function mean_absolute_error at 0x7f4abb83ede0>: it's not the same object as sklearn.metrics._regression.mean_absolute_erro
```

**Expected behavior**
I would have expected it to run the model training as for the other models. 

**Installed Versions**

<details>

```python
show_versions()
```

</details>

shchur:
*Issue #, if available:* Fixes #4914

*Description of changes:*
- MLflow autologging can break AutoGluon model pickling, completely breaking the training process. This PR logs a warning if MLflow autologging is detected.

### DIFF
--- a/common/src/autogluon/common/utils/log_utils.py
+++ b/common/src/autogluon/common/utils/log_utils.py
@@ -193,3 +193,32 @@ def reset_logger_for_remote_call(verbosity: int):
     # Limiting the verbosity of the BaggedEnsembleModel logger to 10 to avoid
     # duplicated messages about fitting.
     set_logger_verbosity(verbosity=min(verbosity, 1), logger=bem_logger)
+
+
+def warn_if_mlflow_autologging_is_enabled(logger: Optional[logging.Logger] = None):
+    """Log a warning if MLflow autologging is enabled.
+
+    MLflow autologging monkey-patches the sklearn metrics, which leads to a PicklingError when AutoGluon models
+    that have a metric as an instance attribute are saved to disk.
+
+    Related issues:
+        - https://github.com/mlflow/mlflow/issues/6268
+        - https://github.com/autogluon/autogluon/issues/4914
+    """
+    if logger is None:
+        logger = _logger_ag
+    try:
+        import mlflow
+
+        try:
+            if not mlflow.utils.autologging_utils.autologging_is_disabled("sklearn"):
+                logger.warning(
+                    "⚠️ Warning: MLflow autologging is enabled. This will likely break AutoGluon model training. "
+                    "Please disable autologging by executing `import mlflow; mlflow.autolog(disable=True)` before importing AutoGluon."
+                )
+        except Exception:
+            # gracefully handle cases where autologging_is_disabled is not available
+            pass
+    except Exception:
+        # mlflow not installed, all good
+        pass
--- a/tabular/src/autogluon/tabular/predictor/predictor.py
+++ b/tabular/src/autogluon/tabular/predictor/predictor.py
@@ -21,7 +21,7 @@ from autogluon.common.loaders import load_json
 from autogluon.common.savers import save_json
 from autogluon.common.utils.file_utils import get_directory_size, get_directory_size_per_file
 from autogluon.common.utils.hyperparameter_utils import get_hyperparameter_str_deprecation_msg, is_advanced_hyperparameter_format
-from autogluon.common.utils.log_utils import add_log_to_file, set_logger_verbosity
+from autogluon.common.utils.log_utils import add_log_to_file, set_logger_verbosity, warn_if_mlflow_autologging_is_enabled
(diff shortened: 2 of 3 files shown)


common/src/autogluon/common/utils/log_utils.py
```diff
@@ -193,3 +193,32 @@ def reset_logger_for_remote_call(verbosity: int):
...
+        try:
+            if not mlflow.utils.autologging_utils.autologging_is_disabled("sklearn"):
+                logger.warning(
+                    "⚠️ Warning: MLflow autologging is enabled. This will likely break AutoGluon model training. "
+                    "Please disable autologging by executing `import mlflow; mlflow.autolog(disable=True)` before importing AutoGluon."
+                )
+        except Exception:
+            # gracefully handle cases where autologging_is_disabled is not available
+            pass
+    except Exception:
```
shchur, line 222: This could, in theory, be narrowed down to an `ImportWarning`, but I don't want to accidentally break the entire AG training process if `import mlflow` fails with some other spectacular exception.

common/src/autogluon/common/utils/log_utils.py
```diff
@@ -193,3 +193,32 @@ def reset_logger_for_remote_call(verbosity: int):
     # Limiting the verbosity of the BaggedEnsembleModel logger to 10 to avoid
     # duplicated messages about fitting.
     set_logger_verbosity(verbosity=min(verbosity, 1), logger=bem_logger)
+
+
+def warn_if_mlflow_autologging_is_enabled(logger: Optional[logging.Logger] = None):
```
shchur, line 198: Any ideas for how/whether we can test this method in our unittests without installing mlflow?

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -1043,6 +1043,7 @@ def fit(
 
         verbosity = kwargs.get("verbosity", self.verbosity)
         set_logger_verbosity(verbosity)
+        warn_if_mlflow_autologging_is_enabled(logger=logger)
```
canerturkmen, line 1046: is it enough to call this in `fit`? why not somewhere in an __init__ file?

common/src/autogluon/common/utils/log_utils.py
canerturkmen: rather test the good path?

(pseudocode)

```
def test():
    try:
         import mlflow
         pytest.fail("mlflow not expected to be installed in test env")
    except:
         assert warn_if_mlflow_autologging_is_enabled()  does not log
```

canerturkmen (approved):
Looks good to me as is. Couple of comments. Thanks!

tabular/src/autogluon/tabular/predictor/predictor.py
shchur: The patched metrics only cause problems when saving objects with pickle (which is only required during training). Calling methods like `leaderboard` or `evaluate` should work normally.

Innixma (approved):
LGTM!

PR #4937, opened 2025-02-26
[tabular] Add raise_on_model_failure flag

Innixma:
*Issue #, if available:*

*Description of changes:*

-  Add raise_on_model_failure flag (default False)
- When True, will immediately raise the original exception if a model raises an exception during fit.
- Setting to True is very helpful when using a debugger to try to figure out why a model is failing, as otherwise exceptions are handled by AutoGluon which isn't desired while debugging.
- Docstring / type hint improvements


tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -990,6 +990,14 @@ def fit(
                 to any amount of labeled data.
             verbosity : int
                 If specified, overrides the existing `predictor.verbosity` value.
+            raise_on_model_failure: bool, default = False
+                If True, will raise on any exception during model training.
+                    This is useful when using a debugger during development to identify the cause of model failures.
+                    This should only be used for debugging.
+                If False, will try to skip to the next model if an exception occurred during model training.
+                    This is the default logic and is a core principal of AutoGluon's design.
```
canerturkmen, line 998: typo -> principle.

canerturkmen (commented):
LGTM overall. I would only encourage adding tests. If somebody else decides to catch the model errors (i.e., the learner starts catching stuff), the behavior will break.

canerturkmen (approved):

tabular/src/autogluon/tabular/predictor/predictor.py
Innixma: updated

Innixma:
added unit test, good suggestion @canerturkmen !

PR #4942, opened 2025-02-26
[WIP] [tabular] speed up tests

Innixma:
*Issue #, if available:*

*Description of changes:*

WIP: Do not merge.

Massively speed up unit tests for tabular models

- Switch from real datasets to toy synthetic datasets
- Use toy model hyperparameters when faster
- Avoid calling `predictor.info()` for each model fit
- Avoid calling `predictor.leaderboard(..., extra_info=True)` for each model fit
- Only call refit_full for a given model family once, no need to test refit_full on every dataset
- Only test bagging for a given model family once, no need to test bagging on every dataset
- Ensure we test bagging for every model family (previously was inconsistent)
- Ensure we are testing every supported problem_type (previously it was inconsistent)

## Speedup

- LightGBM tests: 130s -> 16s 
- Dummy tests: 23s -> 12s
- NeuralNetTorch tests: 140s -> 8s
- NeuralNetFastAI tests: 15s -> 2s
- RandomForest tests: 87s -> 8s
- ExtraTrees tests: ?s -> 7s
- XGBoost tests: ?s -> 4s
- CatBoost tests: ?s -> 3s
- Linear tests: ?s -> 6s

TODO:

- [ ] Speed up remaining models
- [ ] Speed up `test_tabular.py`
- [ ] Adopt a generic model test function?
- [ ] Verify categorical, missing value, etc. handling for models?

### DIFF
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -2486,6 +2486,9 @@ class AbstractModel(ModelBase):
         """
         Dictionary of customization options related to meta properties of the model such as its name, the order it is trained, and the problem types it is valid for.
         """
+        supported_problem_types = cls.supported_problem_types()
+        if supported_problem_types is not None:
+            return {"problem_types": supported_problem_types}
         return {}
 
     @classmethod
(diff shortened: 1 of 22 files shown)


Innixma:
Moving to #4944 to enable CI to work with new tests

PR #4943, opened 2025-02-26
add notebook env check for colab and fix the source installation issue

### LINKED ISSUE #4934: [BUG] Source install broken on Colab
Source installation is broken on Colab with latest mainline.

[Reproducible Colab Notebook](https://colab.research.google.com/drive/1j4OatXgV1iUvkAU-v_cpkOVFbKst7soM?usp=sharing)

Specifically, imports from the root init file of each submodule fail to work, such as `from autogluon.common import __version__`

Unsure why this is happening.

tonyhoo:
*Issue #, if available:*
Fixes #4934

*Description of changes:*

Adds Jupyter/Colab environment detection to the full_install.sh script. When running in Colab or other Jupyter environments, the script automatically adds the `--system` flag to all `uv pip install` commands.

This addresses the issue where AutoGluon installation with `uv pip` was failing in Colab because packages were being installed in a virtual environment not accessible from the notebook's Python kernel. The `--system` flag ensures packages are installed in the system Python path, making them importable in notebook environments.

Colab testing [link](https://colab.research.google.com/drive/16zmGdMylUbLYYVuEgOcEkip8-ah3GH5b?usp=sharing)

### DIFF
--- a/full_install.sh
+++ b/full_install.sh
@@ -2,6 +2,7 @@
 set -euo pipefail
 
 EDITABLE="true"
+UV_FLAGS=""
 
 while test $# -gt 0
 do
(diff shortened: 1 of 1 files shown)


suzhoum (approved):
Thanks for working on the fix!

PR #4944, opened 2025-02-27
[tabular] speed up tests

Innixma:
*Issue #, if available:*

Same as #4942, moved code to autogluon branch so unit tests work correctly in CI

*Description of changes:*

# Tabular Tests Speedup: 3060s -> 743s (4x speedup)

Massively speed up unit tests for tabular models

- Switch from real datasets to toy synthetic datasets
- Use toy model hyperparameters when faster
- Avoid calling `predictor.info()` for each model fit
- Avoid calling `predictor.leaderboard(..., extra_info=True)` for each model fit
- Only call refit_full for a given model family once, no need to test refit_full on every dataset
- Only test bagging for a given model family once, no need to test bagging on every dataset
- Ensure we test bagging for every model family (previously was inconsistent)
- Ensure we are testing every supported problem_type (previously it was inconsistent)

## Enhancements

- Add classmethod to each model to get the supported problem types

## Speedup

- AutoMM tests: 600s -> 60s
- LightGBM tests: 130s -> 16s 
- Dummy tests: 23s -> 12s
- NeuralNetTorch tests: 140s -> 8s
- NeuralNetFastAI tests: 15s -> 2s
- RandomForest tests: 87s -> 8s
- ExtraTrees tests: 108s -> 7s
- XGBoost tests: 32s -> 4s
- CatBoost tests: 61s -> 3s
- Linear tests: 23s -> 6s
- test_tabular: 233s -> 25s
- test_advanced_functionality: 67s -> 33s
- test_tabularHPO: 199s -> 106s
- test_tabular_parallel_local_bagging: 38s -> 35s
- test_tabular_sequential_local_bagging: 16s -> 8s
- test_calibrate.py: 26s -> 11s
- test_learning_curves.py: 70s -> 40s
- test_pseudolabeling: 68s -> 36s
- test_tabularHPObagstack: 204s -> 51s
- test_tabular_bagstack: 169s -> 6s
- test_tabular_bagstack_use_bag_holdout: 168s -> 6s
- test_sample_weight: 52s -> 15s

## Fixes

- test_tabular_feature_prune wasn't actually testing feature pruning, now it does
- Fixed several bugs with oof predictions on quantile problem type

TODO:

- [ ] Verify categorical, missing value, etc. handling for models (Follow-up, currently still have some tests using real datasets to verify this)

### DIFF
--- a/tabular/src/autogluon/tabular/models/fastainn/tabular_nn_fastai.py
+++ b/tabular/src/autogluon/tabular/models/fastainn/tabular_nn_fastai.py
@@ -650,6 +650,10 @@ class NNFastAiTabularModel(AbstractModel):
         return minimum_resources
 
     @classmethod
+    def supported_problem_types(cls) -> list[str] | None:
+        return ["binary", "multiclass", "regression", "quantile"]
+
+    @classmethod
     def _class_tags(cls):
         return {"can_estimate_memory_usage_static": True}
 
--- a/tabular/tests/unittests/test_tabular.py
+++ b/tabular/tests/unittests/test_tabular.py
@@ -558,17 +558,18 @@ def verify_predictor(predictor, train_data, test_data, crash_in_oof, run_distill
         predictor.distill(time_limit=60, augment_args={"size_factor": 0.5})
 
 
-def run_tabular_benchmarks(fast_benchmark, subsample_size, perf_threshold, seed_val, fit_args, dataset_indices=None, run_distill=False, crash_in_oof=False, fit_helper=None):
+def run_tabular_benchmarks(fast_benchmark, subsample_size, perf_threshold, seed_val, fit_args, datasets=None, dataset_indices=None, run_distill=False, crash_in_oof=False, fit_helper=None):
     print("Running fit with args:")
     print(fit_args)
 
     if fit_helper is not None:
-        datasets = [
-            "toy_binary_10",
-            "toy_multiclass_10",
-            "toy_regression_10",
-            "toy_quantile_10",
-        ]
+        if datasets is None:
+            datasets = [
+                "toy_binary_10",
+                "toy_multiclass_10",
+                "toy_regression_10",
+                "toy_quantile_10",
+            ]
         for dataset_name in datasets:
             predictor = fit_helper.fit_and_validate_dataset(
                 dataset_name=dataset_name,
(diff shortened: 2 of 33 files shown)


Innixma:
Note: I am aware that the code in `test_tabular.py` is still messy (but a lot less messy than it was...).

I plan to do a follow-up PR to clean this up. I avoided doing that in this PR to avoid it being increasingly hard to review.

tabular/src/autogluon/tabular/models/fastainn/tabular_nn_fastai.py
```diff
@@ -649,6 +649,10 @@ def get_minimum_resources(self, is_gpu_available=False):
             minimum_resources["num_gpus"] = 0.5
         return minimum_resources
 
+    @classmethod
+    def supported_problem_types(cls) -> list[str] | None:
+        return ["binary", "multiclass", "regression", "quantile"]
```
shchur, line 654: Should we use constants instead? (also applies to the implementation of `supported_problem_types` for other models)

tabular/tests/unittests/test_tabular.py
```diff
@@ -543,9 +484,107 @@ def get_benchmark_sets():
...
     print("Running fit with args:")
     print(fit_args)
+
+    if fit_helper is not None:
+        if datasets is None:
+            datasets = [
+                "toy_binary_10",
+                "toy_multiclass_10",
+                "toy_regression_10",
+                "toy_quantile_10",
```
shchur, line 571: Should there also be a softclass example included here?

shchur (approved):
I'm not 100% familiar with all of the tabular tests, so I might have missed something, but overall looks good to me.

tabular/src/autogluon/tabular/models/fastainn/tabular_nn_fastai.py
Innixma: I think it is a good idea, and that might be something I add such as `ProblemType.binary` so that it is easier to type check rather than `BINARY`.

tabular/tests/unittests/test_tabular.py
Innixma: Yes, however none of our tests previously tested softclass as a standalone problem. Currently softclass is only used for distillation of multiclass tasks, and it isn't implemented in a way that works out of the box as a standalone problem type (I think, haven't checked). I'll look to add it in future.

Issue #4941, opened 2025-02-26
Refit full with additional train rows throwing an AssertionError


Romchyk-S:
**Bug Report Checklist**

- [ ] I provided code that demonstrates a minimal reproducible example.  
- [ ] I confirmed bug exists on the latest mainline of AutoGluon via source install.  
- [ ] I confirmed bug exists on the latest stable version of AutoGluon.  

**Describe the bug**
I try to run a refit_full of the model while passing some additional training rows to it. The columns are exactly identical which I verified by the same method you use in pseudolabeling.assert_pseudo_column_match(X=X, X_pseudo=X_pseudo). Target column is just named "target". And the issue lies with it, because by the time the dataset reaches the assert_pseudo_column_match() function, the "target" column in the original dataset is changed into None, while for additional training rows it gets dropped completely, which causes an
 AssertionError: X and X_pseudo columns are mismatched!
Unexpected Columns in X_pseudo: []
Missing Columns in X_pseudo: [None]

**Expected behavior**
The columns to be identical after all the processing and assert_pseudo_column_match() not to raise an AssertionError.

**To Reproduce**
Hard to give the code, since it also requires having the specific datasets I am using and preprocessing in a certain way.  Will think of a way if I am not able to provide a clear enough description of an issue.

**Screenshots / Logs**
This screenshot shows None appearing instead of "target".
![Image](https://github.com/user-attachments/assets/4a6aac20-c71d-4368-8456-e17e739b8475)

Ran a check through ChatGPT.
![Image](https://github.com/user-attachments/assets/24c1f0e1-1870-47a1-934d-7e8aae19091c)

This screenshot shows "target" getting removed well, without creating None for the new datarows. Which seems to be the desired effect.
![Image](https://github.com/user-attachments/assets/eb3574f4-f35e-4717-9f95-6e210b93de01)

Ran another check through ChatGPT
![Image](https://github.com/user-attachments/assets/d8ca6a80-17f8-44e7-839b-8011b862e335)

```python
# Replace this code with the output of the following:
INSTALLED VERSIONS
------------------
date                   : 2025-02-26
time                   : 20:46:34.552237
python                 : 3.12.6.final.0
OS                     : Windows
OS-release             : 11
Version                : 10.0.26100
machine                : AMD64
processor              : AMD64 Family 25 Model 117 Stepping 2, AuthenticAMD
num_cores              : 16
cpu_ram_mb             : 15661.2265625
cuda version           : None
num_gpus               : 0
[...]

PR #4948, opened 2025-02-28
[tabular] Fix .refit_full(train_data_extra)

Innixma:
*Issue #, if available:*

Related: #4941

*Description of changes:*

Fix a bug where refit_full(train_data_extra) fails when categorical features are present. This is because when concating two dataframes with identical dtypes, the dtypes for category features are changed to float/int. This causes downstream errors in models. This PR fixes this by doing `.astype(X.dtypes)` after the concat.


prateekdesai04 (approved):
LGTM

PR #4954, opened 2025-03-03
Corrects the ReadMe on line counting

jobs-git:
This is technically four lines of codes:

```python
predictor = TabularPredictor(label="class").fit("train.csv")
predictions = predictor.predict("test.csv")
```

readme just chained the `fit()` method into the second line. However, if autogluon goes this path. We can technically write it in one line! :) 

or probably two by chaining the `predict()` method and leaving the `import` on its own line.

This PR corrects this issue.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 <div align="center">
 <img src="https://user-images.githubusercontent.com/16392542/77208906-224aa500-6aba-11ea-96bd-e81806074030.png" width="350">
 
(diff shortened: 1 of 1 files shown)


Innixma:
Hi @jobs-git, thanks for your PR! I'm closing this PR as we prefer the 3-line version with chained `.fit` call.

PR #4964, opened 2025-03-06
Fix source installation issue on colab

tonyhoo:
This reverts commit 626abf77a5603bf1f33e9363ea13e06d9c35df54.

*Issue #, if available:*
N/A

*Description of changes:*
Reverting changes to `full_install.sh` that added `cd "$script_dir"`. This change caused problems with editable installations in Colab environments, where the src-layout structure combined with changing directories resulted in Python being unable to find the correct modules.

The revert ensures that users installing AutoGluon from source in Colab can properly import modules like `TabularPredictor` without additional workarounds. This issue primarily affects researchers and developers who install from source in notebook environments.

Testing verified at: [Colab](https://colab.research.google.com/drive/16zmGdMylUbLYYVuEgOcEkip8-ah3GH5b?usp=sharing)


full_install.sh
```diff
@@ -1,12 +1,6 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-# Get the directory of the script
-script_dir=$(dirname "$0")
-
-# Change the current directory to the script's directory
-cd "$script_dir"
-
```
Innixma, line 9: Maybe instead we only do this logic if not in a colab environment, instead of removing it?

Alternative, don't remove it, but if colab, force `--non-editable` flag setting. This should also fix the issue, since there is no need to have editable install in colab.

full_install.sh
```diff
@@ -4,12 +4,25 @@ set -euo pipefail
...
 
-EDITABLE="true"
-UV_FLAGS=""
+# Set installation type based on environment
+if [ "$IN_COLAB" == "true" ]; then
+    EDITABLE="false"
+    echo "Colab detected - forcing non-editable install"
+else
+    EDITABLE="true"
+    cd "$script_dir"
```
Innixma, line 22: can we try moving the `cd "$script_dir"` outside of the if/else, so it always happens? This should fix the below:

Ideally this should work in colab:

```
!git clone -b revert_change https://github.com/tonyhoo/autogluon.git
!./autogluon/full_install.sh
```

But it doesn't currently:

```
error: Distribution not found at: file:///content/common
```

Innixma (approved):
Works on colab!

PR #4969, opened 2025-03-09
[timeseries] typing improvements for gluonts models

canerturkmen:
*Issue #, if available:*

*Description of changes:*

Pyright typing improvements for abstract models and gluonts models

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
+++ b/timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
@@ -3,7 +3,7 @@ import os
 import shutil
 from datetime import timedelta
 from pathlib import Path
-from typing import Any, Callable, Dict, Iterator, List, Optional, Type, Union, cast, overload
+from typing import TYPE_CHECKING, Any, Callable, Dict, Iterator, List, Optional, Type, Union, cast, overload
 
 import gluonts
 import gluonts.core.settings
@@ -26,6 +26,9 @@ from autogluon.timeseries.dataset.ts_dataframe import ITEMID, TIMESTAMP, TimeSer
 from autogluon.timeseries.models.abstract import AbstractTimeSeriesModel
 from autogluon.timeseries.utils.datetime import norm_freq_str
 from autogluon.timeseries.utils.warning_filters import disable_root_logger, warning_filter
+
+if TYPE_CHECKING:
+    from gluonts.torch.model.forecast import DistributionForecast
 
 # NOTE: We avoid imports for torch and lightning.pytorch at the top level and hide them inside class methods.
 # This is done to skip these imports during multiprocessing (which may cause bugs)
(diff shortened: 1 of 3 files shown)


timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
```diff
@@ -577,14 +607,23 @@ def _stack_sample_forecasts(self, forecasts: List[SampleForecast], item_ids: pd.
     def _stack_distribution_forecasts(self, forecasts: List[Forecast], item_ids: pd.Index) -> pd.DataFrame:
```
shchur, line 607: Would it be possible to remove the need for the many asserts by:
- changing the type hint here to `forecasts: List["DistributionForecast"]` 
- adding a "masked" import of `DistributionForecast` on top of the file with
```python
if typing.TYPE_CHECKING:
    from gluonts.torch.model.forecast import DistributionForecast
```

Maybe this would also the need to add type hints for `item_id_to_forecast` and `dist_forecasts` below?

timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
```diff
@@ -593,30 +632,30 @@ def stack_distributions(distributions: List[Distribution]) -> Distribution:
...
-            for key in dist.arg_constraints.keys():
+            for key in last_dist.arg_constraints.keys():
                 stacked_params[key] = torch.cat([p[key] for p in params_per_dist])
-            return dist.__class__(**stacked_params)
+            return last_dist.__class__(**stacked_params)
 
-        if not isinstance(forecasts[0].distribution, AffineTransformed):
-            raise AssertionError("Expected forecast.distribution to be an instance of AffineTransformed")
+        if not isinstance(dist_forecasts[0].distribution, AffineTransformed):
+            raise AssertionError()
```
shchur, line 640: Could this be removed given the new assertion above? https://github.com/autogluon/autogluon/pull/4969/files#diff-28a0cc813c0c8a1cd39a4e8c358ee689718114207a2655af00cb4ae7675d7665R619-R621

timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
```diff
@@ -629,9 +668,9 @@ def _gluonts_forecasts_to_data_frame(
 
         item_ids = forecast_index.unique(level=ITEMID)
         if isinstance(forecasts[0], SampleForecast):
-            forecast_df = self._stack_sample_forecasts(forecasts, item_ids)
+            forecast_df = self._stack_sample_forecasts(cast(List[SampleForecast], forecasts), item_ids)
         elif isinstance(forecasts[0], QuantileForecast):
-            forecast_df = self._stack_quantile_forecasts(forecasts, item_ids)
+            forecast_df = self._stack_quantile_forecasts(cast(List[QuantileForecast], forecasts), item_ids)
         elif isinstance(forecasts[0], DistributionForecast):
             forecast_df = self._stack_distribution_forecasts(forecasts, item_ids)
```
shchur, line 675: What if instead of adding assertions for `all(isinstance(...))` in the respective methods, we use these checks as branches for the if conditions?
```python
if all(isinstance(f, SampleForecast) for f in forecasts):
   ....
else:
    raise AssertionError("Expected all forecasts to be of SampleForecast, QuantileForecast or DistributionForecast type")
```
If my intuitive understanding of how type checkers work is correct, this could save us some assertions / `cast` methods down the line.

shchur (approved):
LGTM! Just a few nits

timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
canerturkmen: cool, thanks!

timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py
canerturkmen: I don't think it's that easy to pull off. For starters, the type checker won't be satisfied with the all(isinstance()) check. Hence skipping this for now.

PR #5005, opened 2025-04-02
[timeseries] Gracefully handled corrupted cached predictions

### LINKED ISSUE #4992: [timeseries] Gracefully handle the case when cached predictions are corrupted
## Description
Currently, if `cached_predictions.pkl` stored in the predictor folder is corrupted, model prediction will fail
```
File ~/anaconda3/envs/pytorch_p310/lib/python3.10/site-packages/autogluon/timeseries/trainer/abstract_trainer.py:1276, in AbstractTimeSeriesTrainer._get_cached_pred_dicts(self, dataset_hash)
   1274 """Load cached predictions for given dataset_hash from disk, if possible. Otherwise returns empty dicts."""
   1275 if self._cached_predictions_path.exists():
-> 1276     cached_predictions = load_pkl.load(str(self._cached_predictions_path))
   1277     if dataset_hash in cached_predictions:
   1278         model_pred_dict = cached_predictions[dataset_hash]["model_pred_dict"]

File ~/anaconda3/envs/pytorch_p310/lib/python3.10/site-packages/autogluon/common/loaders/load_pkl.py:44, in load(path, format, verbose, **kwargs)
     42 if compression_fn in compression_fn_map:
     43     with compression_fn_map[compression_fn]["open"](validated_path, "rb", **compression_fn_kwargs) as fin:
---> 44         object = pickle.load(fin)
     45 else:
     46     raise ValueError(
     47         f"compression_fn={compression_fn} or compression_fn_kwargs={compression_fn_kwargs} are not valid."
     48         f" Valid function values: {compression_fn_map.keys()}"
     49     )

EOFError: Ran out of input
```

A better approach would be to log an warning and do inference from scratch without caching in this case.

Additionally, we might consider disabling the caching logic by default, given that we are now moving towards models that all have decent inference speed in our default presets.

shchur:
*Issue #, if available:* Fixes #4992

*Description of changes:*
- Gracefully handle the cases when loading cached predictions fails (e.g., because of EOFError, FileNotFoundError, UnpicklingError, incorrect keys stored in the cached predictions dicts).

### DIFF
--- a/timeseries/src/autogluon/timeseries/trainer.py
+++ b/timeseries/src/autogluon/timeseries/trainer.py
@@ -1,6 +1,7 @@
 import copy
 import logging
 import os
+import pickle
 import time
 import traceback
 from collections import defaultdict
@@ -53,6 +54,7 @@ class TimeSeriesTrainer(AbstractTrainer[AbstractTimeSeriesModel]):
         verbosity: int = 2,
         val_splitter: Optional[AbstractWindowSplitter] = None,
         refit_every_n_windows: Optional[int] = 1,
+        # TODO: Set cache_predictions=False by default once all models in default presets have a reasonable inference speed
         cache_predictions: bool = True,
         ensemble_model_type: Optional[Type] = None,
         **kwargs,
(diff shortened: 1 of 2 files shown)


canerturkmen (approved):
LGTM! Thanks!

PR #5007, opened 2025-04-02
Robust model config save logics for AutoMM

tonyhoo:
*Issue #, if available:*
Fix automm tests [failures](https://github.com/autogluon/autogluon/actions/runs/14049312647/job/39738436798)

*Description of changes:*
Robust model config save logics for AutoMM

### DIFF
--- a/multimodal/src/autogluon/multimodal/utils/config.py
+++ b/multimodal/src/autogluon/multimodal/utils/config.py
@@ -362,9 +362,19 @@ def save_pretrained_model_configs(
         model = model.model
     for per_model in model:
         if per_model.prefix.lower().startswith(HF_MODELS):
-            per_model.config.save_pretrained(os.path.join(path, per_model.prefix))
-            model_config = getattr(config.model, per_model.prefix)
-            model_config.checkpoint_name = os.path.join("local://", per_model.prefix)
+            # Skip if config is None
+            if not hasattr(per_model, 'config') or per_model.config is None:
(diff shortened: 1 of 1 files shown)


multimodal/src/autogluon/multimodal/utils/config.py
```diff
@@ -362,9 +362,19 @@ def save_pretrained_model_configs(
...
+            # Skip if config is None
+            if not hasattr(per_model, 'config') or per_model.config is None:
+                logger.warning(f"Skipping saving config for {per_model.prefix} as its config is None")
+                continue
+
+            try:
+                per_model.config.save_pretrained(os.path.join(path, per_model.prefix))
+                model_config = getattr(config.model, per_model.prefix)
+                model_config.checkpoint_name = os.path.join("local://", per_model.prefix)
+            except Exception as e:
```
cheungdaven, line 374: would it be better to error out instead of capturing it quietly?

suzhoum: It's probably fine skipping the model if it failed to save here.

cheungdaven (approved):
lgtm

suzhoum (approved):
Thanks for working on the fix!

multimodal/src/autogluon/multimodal/utils/config.py
tonyhoo: I have reverted this change after figuring out the root cause is from latest transformers release

PR #5019, opened 2025-04-08
[timeseries] Change `evaluation_strategy` to `eval_strategy`

### LINKED ISSUE #5018: [timeseries][BUG] Chronos fine-tuning fails for transformers>=4.41
## Description
Forwarded issue from a user:

> Heads up team, I had this problem happening on the fit
> ```
> Training timeseries model ChronosFineTuned[bolt_base]. 
> 	Warning: Exception caused ChronosFineTuned[bolt_base] to fail during training... Skipping this model.
> 	TrainingArguments.__init__() got an unexpected keyword argument 'evaluation_strategy'
> ```
> I saw that in the source code there is this TODO
> ```
> # TODO: Replace `evaluation_strategy` with `eval_strategy` when upgrading to `transformers>=4.41` + remove warning filter
> ```
> So I had to downgrade to the earlier version of transformers, moved to 4.40 and it seems to be working 

We need to either update the version range in [_setup_utils](https://github.com/autogluon/autogluon/blob/master/core/src/autogluon/core/_setup_utils.py) or automatically change the kwarg name based on the `transformers` version

abdulfatir:
*Issue #, if available:* fixes #5018

*Description of changes:* This PR changes the `evaluation_strategy` training argument to `eval_strategy`, if the `transformers` version is greater than or equal to `4.46`.

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/chronos/model.py
+++ b/timeseries/src/autogluon/timeseries/models/chronos/model.py
@@ -408,6 +408,8 @@ class ChronosModel(AbstractTimeSeriesModel):
         time_limit: Optional[int] = None,
         **kwargs,
     ) -> None:
+        import transformers
+        from packaging import version
         from transformers.trainer import PrinterCallback, Trainer, TrainingArguments
 
         from .pipeline import ChronosBoltPipeline, ChronosPipeline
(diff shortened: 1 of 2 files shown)


timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -496,6 +498,10 @@ def _fit(
                 fine_tune_trainer_kwargs["load_best_model_at_end"] = False
                 fine_tune_trainer_kwargs["metric_for_best_model"] = None
 
+            if version.parse(transformers.__version__) >= version.parse("4.46"):
+                # transformers changed the argument name from `evaluation_strategy` to `eval_strategy`
+                fine_tune_trainer_kwargs["eval_strategy"] = fine_tune_trainer_kwargs.pop("evaluation_strategy")
```
shchur, line 503: Is it guaranteed that the key `evaluation_strategy` is present in `fine_tune_trainer_kwargs`? Maybe we can add an extra check like the following to be safe?
```
            if version.parse(transformers.__version__) >= version.parse("4.46"):
                # transformers changed the argument name from `evaluation_strategy` to `eval_strategy`
                if "evaluation_strategy" in fine_tune_trainer_kwargs:
                    fine_tune_trainer_kwargs["eval_strategy"] = fine_tune_trainer_kwargs.pop("evaluation_strategy")

```

abdulfatir: We are construction `fine_tune_trainer_kwargs` by ourselves in https://github.com/abdulfatir/autogluon/blob/493c044c634d6a154bf65b9c9f5fa7627d73fe1f/timeseries/src/autogluon/timeseries/models/chronos/model.py#L362, so I would say that the check is not needed.

shchur: All right, should be fine as long as this is pinned by unit tests.

shchur (approved):
Thanks!

PR #5024, opened 2025-04-09
[Tabular] Add mem opt simulation artifact support

Innixma:
*Issue #, if available:*

*Description of changes:*

- Add mem opt simulation artifact support, necessary for upcoming TabArena
- Mem opt simulation artifact is up to 4x smaller than non-optimized artifact.
- Tests pass locally

### DIFF
--- a/common/src/autogluon/common/utils/simulation_utils.py
+++ b/common/src/autogluon/common/utils/simulation_utils.py
@@ -1,6 +1,7 @@
 from collections import defaultdict
 from typing import Any, Dict, Tuple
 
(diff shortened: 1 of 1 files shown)


prateekdesai04 (approved):
Essentially we are converting numpy arrays to Series or DF for memory optimization ?

Innixma:
> Essentially we are converting numpy arrays to Series or DF for memory optimization ?

Yeah, numpy is a lot more memory efficient, so I added support for the input to be an optimized variant

Issue #5034, opened 2025-04-11
[timeseries] Do not require the data to be formatted as `TimeSeriesDataFrame` when using the `TimeSeriesPredictor`


shchur:
## Description
Currently, if the users want to provide static features with their data, the only way to accomplish this is to create a `TimeSeriesDataFrame` with the `static_features` attribute. This doesn't have to be the case. All functionality of the `TimeSeriesPredictor` should be compatible with non-AG types such as `pd.DataFrame` and, potentially, `pl.DataFrame`.

Similarly, we can simplify our tutorials and remove any mentions of `TimeSeriesDataFrame`.

PR #5042, opened 2025-04-14
[timeseries] Expose static_features in the predictor API

shchur:
*Issue #, if available:* Addresses #5034 

*Description of changes:*
- Make it possible to provide `static_features` as an attribute to `TimeSeriesPredictor` methods `fit`, `predict`, `evaluate`, `leaderboard`, `feature_importance`.
- Make the docstrings for different predictor methods more concise and consistent.

### DIFF
--- a/timeseries/src/autogluon/timeseries/predictor.py
+++ b/timeseries/src/autogluon/timeseries/predictor.py
@@ -720,7 +720,9 @@ class TimeSeriesPredictor:
 
         if tuning_data is not None:
             # TODO: Allow passing separate static_features to tuning_data?
-            tuning_data = self._check_and_prepare_data_frame(tuning_data, name="tuning_data")
+            tuning_data = self._check_and_prepare_data_frame(
+                tuning_data, static_features=static_features, name="tuning_data"
+            )
             self._check_data_for_evaluation(tuning_data, name="tuning_data")
             logger.info(f"Provided tuning_data has {self._get_dataset_stats(tuning_data)}")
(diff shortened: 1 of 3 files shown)


timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -706,7 +719,10 @@ def fit(
             )
 
         if tuning_data is not None:
-            tuning_data = self._check_and_prepare_data_frame(tuning_data, name="tuning_data")
+            # TODO: Allow passing separate static_features to tuning_data?
```
canerturkmen, line 722: this is problematic imo. did we have anything constraining tuning data to be the same set of items as the training data before? if not, this implicitly does this and is a breaking change.

canerturkmen (commented):
I'm generally not in favor of this change. I believe it introduces too much clutter to the API. Regarding the argument in the original issue: the need to separate dataset implementation does not mean the interface has to change. We could still have a `TimeSeriesDataFrame` object or even a `TimeSeriesData` object that encapsulates dynamic and static data and ensures consistency, but can be implemented in different backends. In any case, this choice should not leak into the API, we should change the API for API reasons alone, i.e., there is a significant need to provide static features through the API, including the introduction of `tuning_data_static_features`

timeseries/src/autogluon/timeseries/predictor.py
shchur: All right, let's postpone this discussion to a later point. I will move the docstring updates to a separate PR.

PR #5057, opened 2025-04-17
Adopt ag_model_register

Innixma:
*Issue #, if available:*

*Description of changes:*

- Adopt ag_model_register in all places to avoid using multiple derivative global variables to track model class metadata.

@shchur I made one edit in timeseries as one of the global variables was used in timeseries. The change should have no impact.


shchur (approved):
LGTM!

PR #5059, opened 2025-04-17
[timeseries] Fix time_limit when val_data is provided #5046

### LINKED ISSUE #5046: [BUG] Time limit modified in AbstractTimeSeriesTrainer since ver 1.2.0 causing unexpected TimeLimitExceed
**Bug Report Checklist**

**Describe the bug**

Since ver 1.2.0, there is a change that wrongly modifies `time_limit` in `._train_and_save()` method as in [v1.2.0](https://github.com/autogluon/autogluon/blob/v1.2.0/timeseries/src/autogluon/timeseries/trainer/abstract_trainer.py#L529-L530) and [master](https://github.com/autogluon/autogluon/blob/master/timeseries/src/autogluon/timeseries/trainer.py#L363-L364). It incurs unexpected TimeLimitExceed errors when it takes milliseconds to train while much longer to evaluate/cache simple local models. 

**Expected behavior**

`time_limit` should only be override when not provided or not sufficient.

**To Reproduce**

~
**Screenshots / Logs**

Image

**Installed Versions**

<details>

```python
show_versions()
```

</details>
v1.2.0

shchur:
*Issue #, if available:* Fixes #5046

*Description of changes:*
- Fix the bug introduced in [#4653](https://github.com/autogluon/autogluon/pull/4653/files#diff-d2a9d7490bd4601b8f3613cfea6c5e37455fb05cc4a025420240a6d47e45de0fR523) that incorrectly set the `time_limit` if `tuning_data` was provided to `predictor.fit()`.

### DIFF
--- a/timeseries/src/autogluon/timeseries/trainer.py
+++ b/timeseries/src/autogluon/timeseries/trainer.py
@@ -361,7 +361,7 @@ class TimeSeriesTrainer(AbstractTrainer[AbstractTimeSeriesModel]):
             model.fit_time = model.fit_time or (fit_end_time - fit_start_time)
 
             if time_limit is not None:
-                time_limit = fit_end_time - fit_start_time
+                time_limit = time_limit - (fit_end_time - fit_start_time)
             if val_data is not None and not self.skip_model_selection:
                 model.score_and_cache_oof(
                     val_data, store_val_score=True, store_predict_time=True, time_limit=time_limit
--- a/timeseries/tests/unittests/test_trainer.py
+++ b/timeseries/tests/unittests/test_trainer.py
@@ -17,6 +17,7 @@ from autogluon.common.loaders import load_pkl
 from autogluon.timeseries.dataset import TimeSeriesDataFrame
 from autogluon.timeseries.models import DeepARModel, ETSModel
 from autogluon.timeseries.models.ensemble.greedy_ensemble import TimeSeriesGreedyEnsemble
+from autogluon.timeseries.splitter import ExpandingWindowSplitter
 from autogluon.timeseries.trainer import TimeSeriesTrainer
 
 from .common import DATAFRAME_WITH_COVARIATES, DUMMY_TS_DATAFRAME, dict_equal_primitive, get_data_frame_with_item_index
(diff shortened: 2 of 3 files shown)


timeseries/src/autogluon/timeseries/trainer.py
```diff
@@ -361,7 +361,7 @@ def _train_and_save(
             model.fit_time = model.fit_time or (fit_end_time - fit_start_time)
 
             if time_limit is not None:
-                time_limit = fit_end_time - fit_start_time
```
abdulfatir, line 364: 😬

timeseries/tests/unittests/test_trainer.py
```diff
@@ -653,3 +654,25 @@ def test_when_add_ci_to_feature_importance_called_then_confidence_bands_correct(
 
             assert np.isclose(r[lower_ci_name], expected_lower)
             assert np.isclose(r[upper_ci_name], expected_upper)
+
+
+def test_when_val_data_and_time_limit_are_provided_then_models_receive_correct_time_limit(temp_model_path):
+    trainer = TimeSeriesTrainer(
+        path=temp_model_path,
+        prediction_length=3,
+        val_splitter=ExpandingWindowSplitter(prediction_length=3, num_val_windows=0),
```
abdulfatir, line 663: Not sure I understand what this is doing with `num_val_windows=0`.

timeseries/tests/unittests/test_trainer.py
```diff
@@ -653,3 +654,25 @@ def test_when_add_ci_to_feature_importance_called_then_confidence_bands_correct(
...
+        "autogluon.timeseries.models.local.abstract_local_model.AbstractLocalModel", fit=mock_fit, predict=mock_predict
+    ):
+        trainer.fit(
+            train_data=data,
+            val_data=data,
+            hyperparameters={"Naive": {"n_jobs": 1}, "Average": {"n_jobs": 1}},
+            time_limit=10000,
+        )
+        for call_args in mock_fit.call_args_list + mock_predict.call_args_list:
+            assert call_args[1]["time_limit"] > 1000
```
abdulfatir, line 678: minor: Could hard-coding 1000 cause issues? I guess for simple methods like Naive and Average and small enough dataframe it's a reasonable number.

abdulfatir (commented):
Thanks @shchur! Left a couple of comments.

timeseries/tests/unittests/test_trainer.py
shchur: There is currently an [assertion](https://github.com/autogluon/autogluon/blob/master/timeseries/src/autogluon/timeseries/models/multi_window/multi_window_model.py#L103-L104) that raises an error if we both provide `val_data` and use a `val_splitter` with `val_splitter.num_val_windows > 0`. Passing a splitter with `num_val_windows=0` essentially disables the use of `MultiWindowBacktestingModel` and leads to "normal" training with `val_data`. 

I hope to get rid of this weirdness after we refactor away the `MultiWindowBacktestingModel`. Added a TODO in the meantime.

timeseries/tests/unittests/test_trainer.py
shchur: This check is only good for catching the exact failure mode that we had before the fix. In that case, these models would receive a `time_limit<1` during predict.

Now that I think about, probably a better idea would be just to test that the `TimeSeriesPredictor` works end-to-end when `tuning_data` is provided, instead of explicitly checking the time limits for individual models. Let me update the PR.

timeseries/tests/unittests/test_trainer.py
shchur: Never mind, removed this test completely.

abdulfatir (approved):

Issue #5032, opened 2025-04-11
[timeseries] Separate the AbstractTimeSeriesEnsemble class from the AbstractTimeSeriesModel


shchur:
## Descriptuon
[`AbstractTimeSeriesEnsemble`](https://github.com/autogluon/autogluon/blob/master/timeseries/src/autogluon/timeseries/models/ensemble/abstract_timeseries_ensemble.py#L11) and `AbstractTimeSeriesModel` expect different inputs to their `fit` and `predict` methods, and therefore should not be connected via inheritance.

PR #5062, opened 2025-04-17
[timeseries] refactor ensemble base class and add new ensembles

canerturkmen:
*Issue #, if available:*

#5032 

*Description of changes:*
- refactor the AbstractTimeSeriesEnsemble class
- add new SimpleAverage and PerformanceWeighted global ensembles


timeseries/src/autogluon/timeseries/models/ensemble/abstract.py
```diff
@@ -77,10 +83,48 @@ def remap_base_models(self, model_refit_map: Dict[str, str]) -> None:
...
+        # Make sure that all predictions have same shape
+        assert len(set(pred.shape for pred in data.values())) == 1
+
+        weighted_predictions = [data[model_name] * weight for model_name, weight in self.model_to_weight.items()]
+
+        return functools.reduce(lambda x, y: x + y, weighted_predictions)
+
+    def get_info(self) -> dict:
+        info = super().get_info()
+        info["model_weights"] = self.model_to_weight
```
shchur, line 122: Let's return a copy to make sure that the user doesn't accidentally break the ensemble by fiddling with the info?

timeseries/src/autogluon/timeseries/models/ensemble/basic.py
```diff
@@ -0,0 +1,66 @@
...
+    (inverse) validation scores, using the method followed by Pawlikowski and Chorowska [PC2020]_.
+
+    References
+    ----------
+    .. [PC2020] Pawlikowski, Maciej, and Agata Chorowska.
+        "Weighted ensemble of statistical models." International Journal of Forecasting
+        36.1 (2020): 93-97.
+    """
+
+    def __init__(self, name: Optional[str] = None, weight_scheme: Literal["sq", "exp"] = "sq", **kwargs):
```
shchur, line 44: Is there a mechanism for the user to select the `weight_scheme`, or is this more forward looking to when we add an API to configure the ensembles via `TimeSeriesPredictor.fit` kwargs? If it's the latter, why not use the `TimeSeriesModelBase.get_hyperparameters` mechanism to get these ensemble hyperparameters instead?

timeseries/src/autogluon/timeseries/models/ensemble/greedy.py
```diff
@@ -98,22 +130,28 @@ def _calculate_regret(self, y_true, y_pred_proba, metric=None, sample_weight=Non
...
+    References
+    ----------
+    .. [Car2024] Caruana, Rich, et al. "Ensemble selection from libraries of models."
+        Proceedings of the twenty-first international conference on Machine learning. 2004.
+    """
 
     def __init__(self, name: Optional[str] = None, ensemble_size: int = 100, **kwargs):
         if name is None:
-            name = "WeightedEnsemble"
+            name = "GreedyWeightedEnsemble"
```
shchur, line 145: Will this result in the model being called `GreedyWeightedEnsemble` in the `TimeSeriesTrainer` model graph? If yes, this could be a breaking change for some users who expect the `WeightedEnsemble` to be the name of the final ensemble model in AG. 

Let's add a TODO here and revisit this once we expose the API for ensemble configuration in v1.4+?

timeseries/src/autogluon/timeseries/models/ensemble/__init__.py
```diff
@@ -1,2 +1,3 @@
-from .abstract_timeseries_ensemble import AbstractTimeSeriesEnsembleModel
-from .greedy_ensemble import TimeSeriesGreedyEnsemble
+from .abstract import AbstractTimeSeriesEnsembleModel
+from .greedy import TimeSeriesGreedyEnsemble
+from .basic import TimeSeriesSimpleAverageEnsemble, TimeSeriesPerformanceWeightedEnsemble
```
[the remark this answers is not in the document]
canerturkmen: ditto

timeseries/src/autogluon/timeseries/models/ensemble/abstract.py
```diff
@@ -1,27 +1,30 @@
...
 from autogluon.core.utils.exceptions import TimeLimitExceeded
 from autogluon.timeseries.dataset import TimeSeriesDataFrame
-from autogluon.timeseries.models.abstract import AbstractTimeSeriesModel
+from autogluon.timeseries.models.abstract import TimeSeriesModelBase
 
 logger = logging.getLogger(__name__)
 
 
-class AbstractTimeSeriesEnsembleModel(AbstractTimeSeriesModel):
+class AbstractTimeSeriesEnsembleModel(TimeSeriesModelBase):
```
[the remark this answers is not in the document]
canerturkmen: absolutely

timeseries/src/autogluon/timeseries/models/ensemble/abstract.py
```diff
@@ -48,26 +53,27 @@ def fit_ensemble(
...
+    def _fit(
         self,
         predictions_per_window: Dict[str, List[TimeSeriesDataFrame]],
         data_per_window: List[TimeSeriesDataFrame],
-        time_limit: Optional[int] = None,
-        **kwargs,
+        model_scores: Optional[Dict[str, float]] = None,
+        time_limit: Optional[float] = None,
     ):
         """Private method for `fit_ensemble`. See `fit_ensemble` for documentation of arguments. Apart from the model
```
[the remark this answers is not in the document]
canerturkmen: thanks!

timeseries/src/autogluon/timeseries/models/ensemble/basic.py
```diff
@@ -0,0 +1,66 @@
...
+import numpy as np
+
+from autogluon.timeseries.dataset import TimeSeriesDataFrame
+
+from .abstract import AbstractWeightedTimeSeriesEnsembleModel
+
+
+class TimeSeriesSimpleAverageEnsemble(AbstractWeightedTimeSeriesEnsembleModel):
+    """Constructs a weighted ensemble using a simple average of the constituent
+    models' predictions.
```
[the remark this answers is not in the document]
canerturkmen: I generally just go with the flow but you are right. Updated this instance.

timeseries/src/autogluon/timeseries/models/ensemble/greedy.py
canerturkmen: OK reverting this.

timeseries/src/autogluon/timeseries/models/ensemble/greedy.py
```diff
@@ -43,13 +44,33 @@ def __init__(
...
         labels: List[TimeSeriesDataFrame],
-        time_limit: Optional[int] = None,
-        sample_weight=None,
+        time_limit: Optional[float] = None,
+    ):
+        return super().fit(
+            predictions=predictions,  # type: ignore
+            labels=labels,  # type: ignore
+            time_limit=time_limit,
+        )
```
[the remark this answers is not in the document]
canerturkmen: I wanted to make it explicit for the type overrides. another way to do this is to add a typing stub. up to you.

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
```diff
@@ -324,7 +324,7 @@ def _score_with_predictions(
             seasonal_period=self.eval_metric_seasonal_period,
         )
```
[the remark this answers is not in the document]
canerturkmen: will do

timeseries/src/autogluon/timeseries/models/ensemble/basic.py
```diff
@@ -0,0 +1,66 @@
...
+        model_scores: Optional[Dict[str, float]] = None,
+        time_limit: Optional[float] = None,
+    ):
+        assert model_scores is not None
+
+        self.model_to_weight = {}
+        score_transform = np.square if self.weight_scheme == "sq" else np.exp
+
+        for model_name in predictions_per_window.keys():
+            self.model_to_weight[model_name] = 1.0 / (score_transform(model_scores[model_name]) + 1e-5)
```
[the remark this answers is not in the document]
canerturkmen: ha! asleep at the wheel!

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
```diff
@@ -709,3 +692,80 @@ def _get_search_space(self):
...
+            Dataset used for scoring.
+        metric: str
+            String identifier of evaluation metric to use, from one of
+            `autogluon.timeseries.utils.metric_utils.AVAILABLE_METRICS`.
+
+        Other Parameters
+        ----------------
+        num_samples: int
+            Number of samples to use for making evaluation predictions if the probabilistic
+            forecasts are generated by forward sampling from the fitted model.
```
shchur, line 730: The docstring is outdated, this comment can be removed.

timeseries/src/autogluon/timeseries/models/ensemble/abstract.py
```diff
@@ -0,0 +1,136 @@
...
+        predictions_per_window : Dict[str, List[TimeSeriesDataFrame]]
+            Dictionary that maps the names of component models to their respective predictions for each validation
+            window.
+        data_per_window : List[TimeSeriesDataFrame]
+            Observed ground truth data used to train the ensemble for each validation window. Each entry in the list
+            includes both the forecast horizon (for which the predictions are given in ``predictions``), as well as the
+            "history".
+        model_scores : Optional[Dict[str, float]]
+            Scores (higher is better) for the models that will constitute the ensemble.
+        time_limit : Optional[int]
```
shchur, line 45: Nit: `Optional[float]` instead of `Optional[int]`.

In general, I think we can safely skip the type annotations in docstrings for internal classes that are not shown in the API section on https://auto.gluon.ai. The type hints are already available in the method signature, and duplicating them in the docstrings just creates room for typos.

timeseries/src/autogluon/timeseries/models/ensemble/abstract.py
```diff
@@ -0,0 +1,136 @@
...
+        self,
+        predictions_per_window: Dict[str, List[TimeSeriesDataFrame]],
+        data_per_window: List[TimeSeriesDataFrame],
+        model_scores: Optional[Dict[str, float]] = None,
+        time_limit: Optional[float] = None,
+    ):
+        """Private method for `fit`. See `fit` for documentation of arguments. Apart from the model
+        training logic, `fit` additionally implements other logic such as keeping track of the time limit.
+        """
+        raise NotImplementedError
```
shchur, line 77: Nit: Should this be an `@abstractmethod` instead of raising `NotImplementedError`?

timeseries/src/autogluon/timeseries/models/ensemble/abstract.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def _predict(self, data: Dict[str, TimeSeriesDataFrame], **kwargs) -> TimeSeriesDataFrame:
+        raise NotImplementedError
+
+    def remap_base_models(self, model_refit_map: Dict[str, str]) -> None:
+        """Update names of the base models based on the mapping in model_refit_map.
+
+        This method should be called after performing refit_full to point to the refitted base models, if necessary.
+        """
+        raise NotImplementedError
```
shchur, line 102: Similarly, should be `@abstractmethod`?

timeseries/src/autogluon/timeseries/models/ensemble/abstract.py
```diff
@@ -0,0 +1,136 @@
...
+    def model_names(self) -> List[str]:
+        return list(self.model_to_weight.keys())
+
+    @property
+    def model_weights(self) -> np.ndarray:
+        return np.array(list(self.model_to_weight.values()), dtype=np.float64)
+
+    def _predict(self, data: Dict[str, TimeSeriesDataFrame], **kwargs) -> TimeSeriesDataFrame:
+        weighted_predictions = [data[model_name] * weight for model_name, weight in self.model_to_weight.items()]
+        return functools.reduce(lambda x, y: x + y, weighted_predictions)
```
shchur, line 124: Nice!

timeseries/src/autogluon/timeseries/models/ensemble/basic.py
```diff
@@ -0,0 +1,72 @@
...
+        predictions_per_window: Dict[str, List[TimeSeriesDataFrame]],
+        data_per_window: List[TimeSeriesDataFrame],
+        model_scores: Optional[Dict[str, float]] = None,
+        time_limit: Optional[float] = None,
+    ):
+        assert model_scores is not None
+
+        weight_scheme = self.get_hyperparameters()["weight_scheme"]
+        self.model_to_weight = {}
+        score_transform = np.square if weight_scheme == "sq" else (lambda x: np.exp(np.clip(x, a_min=0, a_max=70)))
```
shchur, line 66: In our experiments we found that normalizing the weights (so that they sum up to 1) before applying `exp` can be quite beneficial:
image
See the difference between `exp-lnorm` (errors normalized before applying exp) and `exp` (no normalization, same as the code here).

Normalization only matters in case of the `exp` "activation" function, otherwise the scores will anyway be normalized because of the ratio.

If we are dealing with metrics like MASE / RMSE, the raw score values can have scores >>1000, which will lead to uniform scores assigned to all models (after clipping).

In short, the algorithm is 
1. Normalize model errors to sum up to 1
2. Apply the nonlinear transformation (inverse / square / exp)
3. Normalize the weights to sump up to 1

timeseries/src/autogluon/timeseries/models/ensemble/greedy.py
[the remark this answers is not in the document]
shchur: All right, I guess it's fine as is.

timeseries/tests/unittests/models/test_ensemble.py
```diff
@@ -1,21 +1,49 @@
...
 
 from ..common import DUMMY_TS_DATAFRAME, PREDICTIONS_FOR_DUMMY_TS_DATAFRAME, get_data_frame_with_item_index
 
 
+@pytest.fixture(
+    params=itertools.product(
+        [
+            {"model1": -0.2, "model2": -0.3, "model3": -1500},
+            {"model1": -0.5, "model2": -0.2, "model3": 0},
+            {"model1": -3.0, "model2": -1.0, "model3": -2.0},
```
shchur, line 24: Let's also have a test case when some models have `float("nan")` scores?

shchur (approved):
Thanks, looks great! Just a few more nits

timeseries/src/autogluon/timeseries/models/ensemble/abstract.py
canerturkmen: I know we discussed this before. Can you maybe open an issue for this? I am not addressing this or other typing issues on the fly in order to address them all in one go for consistency.

shchur: Sure, makes sense #5074

timeseries/src/autogluon/timeseries/models/ensemble/basic.py
```diff
@@ -62,11 +65,21 @@ def _fit(
...
+        )
+
+        score_transform = {
+            "sq": lambda x: np.square(np.reciprocal(x)),
+            "inv": lambda x: np.reciprocal(x),
+            "sqrt": lambda x: np.sqrt(np.reciprocal(x)),
+        }[weight_scheme]
+
+        self.model_to_weight = {
+            model_name: score_transform(-model_scores[model_name] + 1e-5) for model_name in model_scores.keys()
```
shchur, line 82: This line assumes that all model scores are negative (otherwise `np.sqrt` will be undefined). This currently happens to be true, but is not guaranteed in general. Maybe add an assertion above?

timeseries/tests/unittests/models/test_ensemble.py
```diff
@@ -46,6 +48,13 @@ def _fit(self, predictions_per_window, data_per_window, model_scores=None, time_
     def _predict(self, data, **kwargs):
         return PREDICTIONS_FOR_DUMMY_TS_DATAFRAME
 
+    def remap_base_models(self, model_refit_map: Dict[str, str]) -> None:
+        pass
+
+    @property
+    def model_names(self) -> List[str]:
+        return list(PREDICTIONS_FOR_DUMMY_TS_DATAFRAME.keys())
```
shchur, line 56: This looks a bit suspicious to me. If I understand correctly, `PREDICTIONS_FOR_DUMMY_TS_DATAFRAME` is a `pd.DataFrame`, so `list(PREDICTIONS_FOR_DUMMY_TS_DATAFRAME.keys())` is equivalent to `list(PREDICTIONS_FOR_DUMMY_TS_DATAFRAME.columns)`, which is the same as `["mean", "0.1", "0.2", ..., "0.9"]`. Is this intended?

canerturkmen: ha. asleep at the wheel again!

PR #5077, opened 2025-04-22
Revert "[timeseries] Add support for `horizon_weight` in time series forecasting metrics"

shchur:
Reverts autogluon/autogluon#5058


canerturkmen (approved):
Thanks!

PR #5078, opened 2025-04-22
[timeseries] Add cutoff support in evaluate and leaderboard

### LINKED ISSUE #4356: [timeseries] Allow choosing custom cutoff in `evaluate` and `leaderboard`
## Description
Currently, methods like `evaluate()` and `leaderboard()` always use the last `prediction_length` time steps of each time series as the test. If users want to perform some more sophisticated evaluation, they need to define complex splitting logic outside of the predictor. We can simplify this process a lot by adding a `cutoff` parameter to these methods that defaults to `-1 * prediction_length` (current behavior).

abdulfatir:
*Issue #, if available:* fixes #4356 

*Description of changes:* This PR adds support for `cutoff` in `TimeSeriesPredictor.evaluate` and `TimeSeriesPredictor.leaderboard`. 

TODOs:

- [x] Add support in `leaderboard`
- [x] Add tests
- [x] Update docstrings

### DIFF
--- a/timeseries/src/autogluon/timeseries/predictor.py
+++ b/timeseries/src/autogluon/timeseries/predictor.py
@@ -861,9 +861,9 @@ class TimeSeriesPredictor:
     def evaluate(
         self,
         data: Union[TimeSeriesDataFrame, pd.DataFrame, Path, str],
-        cutoff: Optional[int] = None,
         model: Optional[str] = None,
         metrics: Optional[Union[str, TimeSeriesScorer, List[Union[str, TimeSeriesScorer]]]] = None,
+        cutoff: Optional[int] = None,
         display: bool = False,
         use_cache: bool = True,
     ) -> Dict[str, float]:
@@ -894,17 +894,17 @@ class TimeSeriesPredictor:
 
             If provided data is a `pandas.DataFrame`, AutoGluon will attempt to convert it to a `TimeSeriesDataFrame`.
             If a `str` or a `Path` is provided, AutoGluon will attempt to load this file.
-        cutoff : int, optional
-            A _negative_ integer less than or equal to ``-1 * prediction_length`` denoting the time step in ``data``
-            where the forecast evaluation starts, i.e., time series are evaluated from the ``-cutoff``-th to the
-            ``-cutoff + prediction_length``-th time step. Defaults to ``-1 * prediction_length`, using the last
-            ``prediction_length`` time steps of each time series for evaluation.
         model : str, optional
             Name of the model that you would like to evaluate. By default, the best model during training
             (with highest validation score) will be used.
         metrics : str, TimeSeriesScorer or List[Union[str, TimeSeriesScorer]], optional
             Metric or a list of metrics to compute scores with. Defaults to ``self.eval_metric``. Supports both
             metric names as strings and custom metrics based on TimeSeriesScorer.
+        cutoff : int, optional
+            A _negative_ integer less than or equal to ``-1 * prediction_length`` denoting the time step in ``data``
+            where the forecast evaluation starts, i.e., time series are evaluated from the ``-cutoff``-th to the
+            ``-cutoff + prediction_length``-th time step. Defaults to ``-1 * prediction_length`, using the last
+            ``prediction_length`` time steps of each time series for evaluation.
         display : bool, default = False
             If True, the scores will be printed.
         use_cache : bool, default = True
--- a/timeseries/src/autogluon/timeseries/predictor.py
+++ b/timeseries/src/autogluon/timeseries/predictor.py
@@ -305,14 +305,31 @@ class TimeSeriesPredictor:
                 df = df.convert_frequency(freq=self.freq)
         return df
 
-    def _check_data_for_evaluation(self, data: TimeSeriesDataFrame, name: str = "data") -> None:
-        """Make sure that provided evaluation data includes both historical and future time series values."""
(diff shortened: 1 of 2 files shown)


timeseries/src/autogluon/timeseries/trainer.py
```diff
@@ -809,12 +811,15 @@ def score(
     def evaluate(
         self,
         data: TimeSeriesDataFrame,
+        cutoff: Optional[int] = None,
         model: Optional[Union[str, AbstractTimeSeriesModel]] = None,
         metrics: Optional[Union[str, TimeSeriesScorer, List[Union[str, TimeSeriesScorer]]]] = None,
         use_cache: bool = True,
     ) -> Dict[str, float]:
```
shchur, line 818: I think we could limit the `cutoff` argument to the `evaluate` method in `predictor` / `learner` / `trainer`, and keep other classes / methods unchanged. For example, we could appropriately slice the `data` at the start of `evaluate` and keep the rest of the code unchanged. Something like
```python
if cutoff is None:
    cutoff = -1 * self.prediction_length
if cutoff < -self.prediction_length:
    data = data.slice_by_timestep(None, cutoff + self.prediction_length)
# the rest of the code in evaluate stays the same
```
Would this work?

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -893,7 +894,17 @@ def evaluate(
...
+            cutoff = -1 * self.prediction_length
+
+        if not (isinstance(cutoff, int) and cutoff <= -self.prediction_length):
+            raise ValueError(f"cutoff should be a negative integer <= -prediction_length, got: {cutoff=}")
+
         data = self._check_and_prepare_data_frame(data)
+
+        if cutoff < -self.prediction_length:
+            data = data.slice_by_timestep(None, cutoff + self.prediction_length)
+
```
shchur, line 907: Now that I think about it, maybe we can even avoid passing the `cutoff` to `learner` / `trainer` and just handle the cutoff inside the `predictor`? Code duplication looks a bit strange, but I also see that we need to have it everywhere.

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -893,7 +894,17 @@ def evaluate(
...
         self._check_data_for_evaluation(data)
```
shchur, line 908: My bad, we probably need to update `_check_data_for_evaluation` as well taking the `cutoff` into account. Otherwise the error message raised when `abs(cutoff) > data.num_timesteps_per_item().max()` will be really cryptic to the user.

abdulfatir: Yep, I realized that and have fixed this.

timeseries/src/autogluon/timeseries/predictor.py
abdulfatir: Yes, that's what I am doing now.

abdulfatir: Re: code duplication. Would you prefer having a utility function attached to `TimeSeriesDataFrame` that handles the cutoff?

shchur: I would rather not introduce this functionality to `TimeSeriesDataFrame` since `TimeSeriesDataFrame.slice_by_timestep` is already very close to the required API. I think we can add a private method to `TimeSeriesPredictor` (something like `_check_and_prepare_data_for_evaluation`) that combines validation from `_check_data_for_evaluation` with slicing based on the cutoff.

abdulfatir: I have already modified `_check_data_for_evaluation` to support cutoff. What do you think?

shchur: I mean that we could, in principle, completely remove any changes to `learner` and `trainer` and 100% localize this PR to the `predictor`. 

We can replace the method `_check_data_for_evaluation` with a new method `_check_and_prepare_data_for_evaluation(self, data: TimeSeriesDataFrame, cutoff: int | None) -> TimeSeriesDataFrame` that checks the `cutoff`, the `data`, and, if necessary, does the slicing on the `data`. We then can call `_check_and_prepare_data_for_evaluation` from inside both `leaderboard` and `evaluate` methods.

abdulfatir: Misunderstood you before. How about now?

shchur (approved):
Looks great, thanks!

timeseries/tests/unittests/test_predictor.py
```diff
@@ -1003,6 +1003,54 @@ def test_when_custom_metric_passed_to_score_then_predictor_can_evaluate(temp_mod
     assert isinstance(scores[eval_metric.name], float)
 
 
+@pytest.mark.parametrize("cutoff, prediction_length", [(-8, 9), (-9.0, 9), (9, 9), ("2020-01-01", 9)])
```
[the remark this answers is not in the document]
abdulfatir: Done.

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -844,6 +861,7 @@ def predict(
     def evaluate(
         self,
         data: Union[TimeSeriesDataFrame, pd.DataFrame, Path, str],
+        cutoff: Optional[int] = None,
```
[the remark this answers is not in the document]
abdulfatir: I moved it after `metrics`.

PR #5079, opened 2025-04-22
[timeseries] Enable HPO over `target_scaler`, `covariate_scaler` and `covariate_regressor` arguments

### LINKED ISSUE #4957: [timeseries] Models currently cannot handle HPO in transform arguments
See [discussion](https://github.com/autogluon/autogluon/pull/4946#discussion_r1978900693) and test case (marked xfail) `test_when_hyperparameter_spaces_of_transforms_provided_to_init_then_model_can_tune`

shchur:
*Issue #, if available:* Fixes #4957

Waiting for #5062 to be complete before this PR can be merged.

*Description of changes:*
- Move the call to `_initialize_transforms_and_regressor` from `__init__` to `fit`.
- Move the hyperparameter manipulations from `__init__` to `_fit` in the `AbstractLocalModel` class
- Replace the test `test_when_hyperparameter_spaces_of_transforms_provided_to_init_then_model_can_tune` performed for each model by a single test for the `NaiveModel` to avoid unnecessary repeated tests.


timeseries/src/autogluon/timeseries/models/local/abstract_local_model.py
```diff
@@ -82,9 +63,12 @@ def __init__(
             **kwargs,
         )
 
-        self._local_model_args: Dict[str, Any] = None
-        self._seasonal_period: Optional[int] = None
-        self._dummy_forecast: Optional[pd.DataFrame] = None
+        self.n_jobs: int
```
canerturkmen, line 66: if these are indeed "hyperparameters" can they safely be removed as instance variables? I think this is a cause for inconsistency that's currently spread across model implementations so we should get rid of them as much as possible. Ideally, hyperparameters should be fixed during the lifetime of a model and should only be accessed through `get_hyperparameters`?

timeseries/src/autogluon/timeseries/models/local/abstract_local_model.py
```diff
@@ -105,19 +89,35 @@ def preprocess(
...
     def _fit(self, train_data: TimeSeriesDataFrame, time_limit: Optional[int] = None, **kwargs):
         self._check_fit_params()
 
         if time_limit is not None and time_limit < self.init_time_in_seconds:
             raise TimeLimitExceeded
 
-        # Initialize parameters passed to each local model
-        raw_local_model_args = self.get_hyperparameters().copy()
+        hyperparameters = self.get_hyperparameters()
+        # TODO: Replace with 'num_cpus' argument passed to fit (after predictor API is changed)
```
canerturkmen, line 106: Similar to the above, can these go inside `self.get_hyperparameters` I wonder

timeseries/src/autogluon/timeseries/models/local/abstract_local_model.py
shchur: I agree with removing the instance attributes `n_jobs`, `max_ts_length` and `use_fallback_model`, but I would prefer to keep `_local_model_args` as-is for now.

timeseries/src/autogluon/timeseries/models/local/abstract_local_model.py
shchur: I would prefer to perform the "preprocessing" of `_local_model_args` inside `_fit()` instead of moving it to `get_hyperparameters()`. If we do move it to `get_hyperparameters()`, the warning 
```
                f"{self.name} ignores following hyperparameters: {unused_local_model_args}. "
                f"See the docstring of {self.name} for the list of supported hyperparameters."
```
will be logged each time the user predicts with the model (e.g., during `leaderboard` or `evaluate`). I think that's unnecessary and the warning should only be logged once during `predictor.fit()`.

canerturkmen (approved):
LGTM! Thanks!

PR #5085, opened 2025-04-23
[timeseries] typing improvements for mlforecast models

canerturkmen:
*Issue #, if available:*

*Description of changes:*

typing improvements for tabular models

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
+++ b/timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
@@ -2,7 +2,7 @@ import logging
 import math
 import os
 import time
-from typing import Any, Callable, Dict, List, Optional, Tuple, Union
+from typing import Any, Callable, Collection, Dict, List, Optional, Tuple, Union
 
 import numpy as np
 import pandas as pd
@@ -12,6 +12,7 @@ from typing_extensions import Self
 import autogluon.core as ag
 from autogluon.tabular import TabularPredictor
 from autogluon.timeseries.dataset.ts_dataframe import ITEMID, TIMESTAMP, TimeSeriesDataFrame
+from autogluon.timeseries.metrics.abstract import TimeSeriesScorer
 from autogluon.timeseries.metrics.utils import in_sample_squared_seasonal_error
 from autogluon.timeseries.models.abstract import AbstractTimeSeriesModel
 from autogluon.timeseries.models.local import SeasonalNaiveModel
@@ -67,9 +68,9 @@ class AbstractMLForecastModel(AbstractTimeSeriesModel):
         prediction_length: int = 1,
         path: Optional[str] = None,
         name: Optional[str] = None,
-        eval_metric: Optional[str] = None,
+        eval_metric: Optional[Union[str, TimeSeriesScorer]] = None,
         hyperparameters: Optional[Dict[str, Any]] = None,
-        **kwargs,  # noqa
+        **kwargs,
     ):
         super().__init__(
             path=path,
@@ -86,7 +87,7 @@ class AbstractMLForecastModel(AbstractTimeSeriesModel):
         self._sum_of_differences: int = 0  # number of time steps removed from each series by differencing
         self._max_ts_length: Optional[int] = None
         self._target_lags: np.ndarray
-        self._date_features: Optional[List[Callable]] = None
+        self._date_features: List[Callable]
         self._mlf: MLForecast
         self._scaler: Optional[BaseTargetTransform] = None
         self._residuals_std_per_item: pd.Series
@@ -170,7 +171,7 @@ class AbstractMLForecastModel(AbstractTimeSeriesModel):
 
         target_transforms = []
         differences = model_params.get("differences")
-        assert isinstance(differences, list)
+        assert isinstance(differences, Collection)
 
         ts_lengths = train_data.num_timesteps_per_item()
         required_ts_length = sum(differences) + 1
--- a/timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
+++ b/timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
@@ -395,6 +395,7 @@ class AbstractTimeSeriesModel(TimeSeriesModelBase, TimeSeriesTunable, ABC):
         eval_metric: Union[str, TimeSeriesScorer, None] = None,
         eval_metric_seasonal_period: Optional[int] = None,
     ):
+        # TODO: make freq a required argument in AbstractTimeSeriesModel
         super().__init__(
             path=path,
             name=name,


timeseries/src/autogluon/timeseries/models/autogluon_tabular/transforms.py
```diff
@@ -13,29 +13,32 @@
...
 from .utils import MLF_ITEMID, MLF_TIMESTAMP
 
 
 class MLForecastScaler(BaseTargetTransform):
-    def __init__(self, scaler_type: Literal["standard", "robust", "min_max", "mean_abs"]):
+    def __init__(self, scaler_type: Literal["standard", "min_max", "mean_abs", "robust"]):
         # For backward compatibility
-        self.scaler_type = scaler_type
-        self.ag_scaler: Optional[LocalTargetScaler] = None
+        self.scaler_type: Literal["standard", "min_max", "mean_abs", "robust"] = scaler_type
```
shchur, line 24: Why is this annotation necessary?

timeseries/src/autogluon/timeseries/models/autogluon_tabular/transforms.py
```diff
@@ -13,29 +13,32 @@
...
+            df.rename(columns={self.id_col: ITEMID, self.time_col: TIMESTAMP}).set_index([ITEMID, TIMESTAMP])
+        )
 
     def _tsdf_to_df(self, ts_df: TimeSeriesDataFrame) -> pd.DataFrame:
         return pd.DataFrame(ts_df).reset_index().rename(columns={ITEMID: self.id_col, TIMESTAMP: self.time_col})
 
-    def fit_transform(self, df: pd.DataFrame) -> pd.DataFrame:
+    def fit_transform(self, df: pd.DataFrame) -> pd.DataFrame:  # type: ignore
         self.ag_scaler = get_target_scaler(name=self.scaler_type, target=self.target_col)
-        transformed = self.ag_scaler.fit_transform(self._df_to_tsdf(df)).reset_index()
```
shchur, line 35: Was this a bug before?

timeseries/src/autogluon/timeseries/models/autogluon_tabular/transforms.py
```diff
@@ -46,7 +49,9 @@ def apply_inverse_transform(
 ) -> pd.DataFrame:
     """Apply inverse transformation to a dataframe, converting to GroupedArray if necessary"""
     if isinstance(transform, BaseTargetTransform):
-        return transform.inverse_transform(df=df)
+        inverse_transformed = transform.inverse_transform(df=df)
+        assert isinstance(inverse_transformed, pd.DataFrame)
+        return inverse_transformed
```
shchur, line 54: Nit: Can we accomplish this with some comment like a `# type: pd.DataFrame` comment instead of an assertion ([PEP 484](https://peps.python.org/pep-0484/#type-comments))? As far as I understand, we are only using the `assert` to tell pyright "this is definitely a pd.DataFrame, otherwise this code will explode". Maybe we can just keep the first part without the risk of raising an exception

timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
```diff
@@ -62,8 +67,8 @@ def __init__(
         prediction_length: int = 1,
         path: Optional[str] = None,
         name: Optional[str] = None,
-        eval_metric: str = None,
-        hyperparameters: Dict[str, Any] = None,
+        eval_metric: Optional[str] = None,
```
shchur, line 70: `eval_metric: Optional[Union[str, TimeSeriesScorer]]` for consistency with `AbstractTimeSeriesModel`?

timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
```diff
@@ -80,11 +85,11 @@ def __init__(
 
         self._sum_of_differences: int = 0  # number of time steps removed from each series by differencing
         self._max_ts_length: Optional[int] = None
-        self._target_lags: Optional[List[int]] = None
-        self._date_features: Optional[List[str]] = None
-        self._mlf: Optional[MLForecast] = None
+        self._target_lags: np.ndarray
+        self._date_features: Optional[List[Callable]] = None
```
shchur, line 89: Similar to above, this could this be `self._date_features: List[Callable]` (without `Optional` and `= None`)?

timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
```diff
@@ -131,24 +139,27 @@ def preprocess(
...
+    def _get_mlforecast_init_args(
+        self, train_data: TimeSeriesDataFrame, model_params: Dict[str, Any]
+    ) -> Dict[str, Any]:
         from mlforecast.target_transforms import Differences
 
         from .transforms import MLForecastScaler
 
         lags = model_params.get("lags")
         if lags is None:
+            assert self.freq is not None
```
shchur, line 162: Side note: This is something that we need to assert in `AbstractTimeSeriesModel` (or make `freq` a require argument to `AbstractTimeSeriesModel.__init__`). Most models implicitly assume that `self.freq` is available, e.g., to generate the forecast horizon index.

timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
```diff
@@ -159,6 +170,7 @@ def _get_mlforecast_init_args(self, train_data: TimeSeriesDataFrame, model_param
 
         target_transforms = []
         differences = model_params.get("differences")
+        assert isinstance(differences, list)
```
shchur, line 173: Can we either cast or add a comment describing this? I can imagine that some users will provide this as a `np.ndarray` - we probably don't want to break the model for them just to appease the type checker.

timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
```diff
@@ -304,6 +318,7 @@ def _fit(
         model_params = self.get_hyperparameters()
 
         mlforecast_init_args = self._get_mlforecast_init_args(train_data, model_params)
+        assert self.freq is not None
```
shchur, line 321: We have this assertion multiple times. Is this intended, or should we maybe move it to `__init__`?

timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
```diff
@@ -542,32 +569,38 @@ def _predict(
...
-        predictions = TimeSeriesDataFrame(predictions.rename(columns={MLF_ITEMID: ITEMID, MLF_TIMESTAMP: TIMESTAMP}))
+        predictions_tsdf: TimeSeriesDataFrame = TimeSeriesDataFrame(
+            predictions.rename(columns={MLF_ITEMID: ITEMID, MLF_TIMESTAMP: TIMESTAMP})
+        )
 
         if forecast_for_short_series is not None:
-            predictions = pd.concat([predictions, forecast_for_short_series])
-            predictions = predictions.reindex(original_item_id_order, level=ITEMID)
-        return predictions
+            predictions_tsdf = pd.concat([predictions_tsdf, forecast_for_short_series])  # type: ignore
```
shchur, line 585: `#type: TimeSeriesDataFrame`?

shchur (commented):
Thanks! Just a few nits.

timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
canerturkmen: added a TODO

timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
canerturkmen: converted to `Collection`

timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
canerturkmen: Yeah pyright doesn't recognize if the assert is in `__init__`.

timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
canerturkmen: the alternative is a `cast`, not a type annotation. `type: TimeSeriesDataFrame` is equivalent, afaiu, to 

```
predictions_tsdf: TimeSeriesDataFrame = ...
```

timeseries/src/autogluon/timeseries/models/autogluon_tabular/transforms.py
canerturkmen: it's not necessary per se but also not bad?

timeseries/src/autogluon/timeseries/models/autogluon_tabular/transforms.py
canerturkmen: see above about type comments. we actually want to raise an exception. the alternative is a polars DataFrame in this case, which we do not want.

shchur: I did some background reading, turns out my understanding of PEP 484 was incorrect. Disregard my comment.

timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
shchur: Never mind, my understanding was incorrect.

shchur (approved):
LGTM!

timeseries/src/autogluon/timeseries/models/autogluon_tabular/transforms.py
canerturkmen: no, but it happens anyway within `_tsdf_to_df`

PR #5087, opened 2025-04-23
Improve Dockerfile for better layer caching and workflow speed

itayporezky:
*Description of changes:* This PR hopefully will improve docker layer caching which has many benefits.

1. Shorten workflow execution time, lowering Github cost
2. Better usage of docker layer caching which will: reduce hosting storage cost, lower client's storage cost, speedup client's pull time.

**Note, the workflow always pushes `:latest` tag, building this PR will probably(?) break your deployments. You can modify my commits to push a different tag for testing.**

I can't really test this myself due to all of your dependencies on AWS and private base docker image(`FROM`), I've tested partially using base image of `python`.

Additional possible future improvement is to change the order of COPY if you have some python modules which are not updated frequently and separate it to a different `RUN pip install xxxxx`, again to improve caching.


prateekdesai04:
Thank you for your contribution @itayporezky!
Curious if you happen to have any speed-up numbers for individual dockerfile builds before and after your changes?

itayporezky:
I didn't test specific size differences, I'm speaking mostly from a theoretical point of view.
Previously, there was 
```Dockerfile
RUN git clone https://github.com/autogluon/autogluon.git
COPY full_install_image.sh autogluon/
RUN cd autogluon && chmod +x full_install_image.sh && ./full_install_image.sh
```
This means that any single byte change in the whole git repo would cause the final layer of the image to be recreated.
After my changes, `docker build` will `COPY` only the relevant folders, so changes in other folders(docs, CI, .git, etc) would not cause the final layer to be different and `docker push` would complete almost instantly and push nothing.

This is significant since the final layer contains cuda and other very heavy packages.

itayporezky:
I'm getting some new ideas to improve this more. I'll change the PR to draft for now

Innixma:
Closing as stale

PR #5089, opened 2025-04-24
increase pytorch cap to 2.8 to enable 2.7

FireballDWF:
*Description of changes:* increase pytorch cap to 2.8 to enable 2.7


FireballDWF:
Encourage anyone to help triage and fix the 3 test failures, as I will not be able to triage for at least two weeks.

FireballDWF:
I am going to wait for pytorch-lightning to officially release Pytorch 2.7 support before attempting triage of the current test failures.

FireballDWF:
Going to wait for the release of [PyTorch 2.7.1](https://dev-discuss.pytorch.org/t/pytorch-release-2-7-1/2963) as[ Commit 7f79222](https://github.com/pytorch/pytorch/commit/7f79222992d72b35d5f51e78db6c7ed5491b4641) should include update to [NCCL 2.26.5](https://docs.nvidia.com/deeplearning/nccl/release-notes/rel_2-26-5.html#rel_2-26-5) which may resolve the core dumps occurring in all 3 test case failures.

FireballDWF:
Error Log https://github.com/autogluon/autogluon/actions/runs/15449884138/job/43489311199?pr=5089 contained 
```
[2025-06-04T18:50:47.608Z] E                       RuntimeError: cuDNN version incompatibility: PyTorch was compiled  against (9, 5, 1) but found runtime version (9, 1, 0). PyTorch already comes bundled with cuDNN. One option to resolving this error is to ensure PyTorch can find the bundled cuDNN. one possibility is that there is a conflicting cuDNN in LD_LIBRARY_PATH.
```
Will see if I can figure out what to change, but encourage anyone more familiar with the test setup to provide suggestions or additional commits

FireballDWF:
@tonyhoo @Innixma Based on preliminary analysis, should either add something like 
```
export LD_LIBRARY_PATH=$(python -c "import torch; print(torch._C._cuda_getLibPath())"):$LD_LIBRARY_PATH:
```
before executing Pytorch tests, or if there are no non-pytorch tests which require CUDA, then start with an image that does not have CUDA pre-installed, so that only the one bundled with PyTorch is used.  Or upgrade to a CUDA 12.6 based image, however I think the better option is to use the one bundled with PyTorch, which would support future tests where the version higher than current default is used (for example to have a new test against the 12.8 CUDA version bundled with the cu128 wheel of PyTorch)

tonyhoo:
@FireballDWF I am taking a look at the test failure issue and trying to reproduce it locally

tonyhoo:
I've updated the Docker image to run natively on Torch 2.7.1. Meanwhile, I've fixed the test cases to account for sensitive numeric differences on short training samples

tonyhoo (approved):
Thanks @FireballDWF for your contributions and for pushing this to completion! Great work! 🚀

PR #5100, opened 2025-04-28
[timeseries] add model registry and fix presets typing

canerturkmen:
*Issue #, if available:*

*Description of changes:*

- [x] Adds `ModelRegistry` to time series. By default, all models which inherit from `AbstractTimeSeriesModel` will be registered. 
- [x] Moves `default_priority` to a field defined by models (WIP).
- [x] Some typing improvements to presets (refactor WIP).


timeseries/src/autogluon/timeseries/models/presets.py
```diff
@@ -244,7 +192,7 @@ def get_preset_models(
                 f"Keys of the `hyperparameters` dictionary must be strings or types, received {type(model)}."
             )
 
-        for model_hps in hyperparameters[model]:
+        for model_hps in hyperparameter_dict[model]:
             ag_args = model_hps.pop(constants.AG_ARGS, {})
```
shchur, line 196: Thanks, the PR looks great! I guess we can also move this logic closer to the `TimeSeriesTrainer` at some point. IMO, there should be just one "preset_configs" file that contains 
- equivalent of the `get_default_hps` method
- preset configurations

Everything else feels like it should be a part of the trainer.

timeseries/src/autogluon/timeseries/models/registry.py
```diff
@@ -0,0 +1,62 @@
...
+from typing import Dict
+
+
+@dataclass
+class ModelRecord:
+    model_class: type
+    default_priority: int
+
+
+class ModelRegistry(type):
```
shchur, line 12: Nit: Should we align this with [Tabular](https://github.com/autogluon/autogluon/blob/master/tabular/src/autogluon/tabular/register/_model_register.py#L14)? I think `Registry` is the more precise term, I wonder if we should also change `ModelRegister` -> `ModelRegistry` on the tabular side.

Innixma: I'm ok with changing to `ModelRegistry` so long as we do it before the 1.3 release. Feel free to open a PR for it

Innixma: Created #5107 to align

timeseries/src/autogluon/timeseries/models/registry.py
```diff
@@ -0,0 +1,62 @@
...
+    @classmethod
+    def get_model_priority(cls, alias: str):
+        return cls.REGISTRY[alias].default_priority
+
+    @classmethod
+    def available_aliases(cls):
+        return list(cls.REGISTRY.keys())
+
+
+class RegisteredABCMeta(ModelRegistry, ABCMeta):
```
shchur: I think we could get away with not introducing `RegisteredABCMeta` and `RegisteredABC` and only having a single `ModelRegistry` class:
```python
# registry.py
class ModelRegistry(ABCMeta):
    REGISTRY: Dict[str, ModelRecord] = {}
    ...

# abstract_timeseries_model.py
class AbstractTimeSeriesModel(TimeSeriesModelBase, TimeSeriesTunable, metaclass=ModelRegistry):
    ...
```
This way all classes inheriting from `AbstractTimeSeriesModel` will automatically be registered.

timeseries/src/autogluon/timeseries/models/registry.py
```diff
@@ -0,0 +1,62 @@
...
+        record = ModelRecord(
+            model_class=new_cls,
+            default_priority=getattr(new_cls, "default_priority", 0),
+        )
+
+        if name is not None:
+            cls._add(name, record)
+
+            # additionally, if the name is like XXModel, alias it as "XX"
+            if name.endswith("Model") and len(name) > 5:
```
shchur: Currently `ModelRegistry.available_aliases()` includes the abstract types, and also two copies of each model (with and without `Model` suffix):
```python
['RegisteredABC',
 'AbstractTimeSeriesModel',
 'AbstractTimeSeries',
 'AbstractLocalModel',
 'AbstractLocal',
 'NaiveModel',
 'Naive',
 'SeasonalNaiveModel',
 'SeasonalNaive',
  ...
```
As I understand, we would want to show the output of `ModelRegistry.available_aliases()` to the user if they provide an invalid model alias in `hyperparameters`, something like
```python
class ModelRegistry:
    @classmethod
    def get_model_class(cls, alias: str):
        if alias not in cls.REGISTRY:
            raise ValueError("Model '{alias}' not found. Available models: {cls.available_aliases()}")
        return cls.REGISTRY[alias].model_class
```
In this case, I would suggest that 
1. We exclude abstract classes from the registry by checking `inspect.isabstract(new_cls)` before registering.
2. We only apply `alias.removesuffix("Model")` inside methods like `get_model_class` (i.e., we only register `Naive` in `REGISTRY`, and do not register the key `NaiveModel`) to keep the list of available models more concise.

timeseries/src/autogluon/timeseries/models/registry.py
```diff
@@ -0,0 +1,62 @@
+from abc import ABCMeta
+from dataclasses import dataclass
+from typing import Dict
+
+
+@dataclass
+class ModelRecord:
+    model_class: type
```
shchur, line 9: Nit: Maybe `Type[AbstractTimeSeriesModel]` instead of `type`?

canerturkmen: introduces a circular dependency...

timeseries/src/autogluon/timeseries/models/registry.py
canerturkmen: It's a little trickier than this. Because `ABCMeta` is also defined by a custom `__new__` method, the new `__new__` method overrides the `ABCMeta` one, removing the "abstractness."

```python
import inspect
from abc import ABCMeta
from typing import Any, Dict

class Registry(ABCMeta):
    REGISTRY: Dict[str, Any] = {}

    def __new__(cls, name, bases, attrs):
        new_cls = type.__new__(cls, name, bases, attrs)

        if name is not None:
            cls._add(name, new_cls)

        return new_cls

    @classmethod
    def _add(cls, alias: str, record: Any):
        if alias in cls.REGISTRY:
            raise ValueError(f"You are trying to define a new model with {alias}, but this model already exists.")
        cls.REGISTRY[alias] = record

class Model(metaclass=Registry):
    pass

assert Registry.REGISTRY
assert not inspect.isabstract(Model)  # !
```

canerturkmen: actually, now realizing that `inspect.isabstract` expects at least one abstract method.

timeseries/src/autogluon/timeseries/models/registry.py
shchur: Oops, indeed. Never mind

timeseries/src/autogluon/timeseries/models/registry.py
shchur: So since `Model` will have `@abstractmethod`s `_fit` and `_predict`, `inspect.isabstract(Model) ` will return `True` and this will work?

canerturkmen: yes. the only other gotcha is that instead of `type.__new__` one needs `super().__new__`. fixed now.

timeseries/src/autogluon/timeseries/models/presets.py
canerturkmen: this has already gotten a little big so I'll defer to a fast-follow PR.

timeseries/src/autogluon/timeseries/models/presets.py
```diff
@@ -9,108 +9,14 @@
...
-    RecursiveTabularModel,
-    SeasonalAverageModel,
-    SeasonalNaiveModel,
-    SimpleFeedForwardModel,
-    TemporalFusionTransformerModel,
-    ThetaModel,
-    TiDEModel,
-    WaveNetModel,
-    ZeroModel,
-)
```
shchur, line 40: Love changes like these!

timeseries/src/autogluon/timeseries/models/presets.py
```diff
@@ -269,22 +153,60 @@ def get_preset_models(
...
+        logger.info(f"Excluded model types: {excluded_model_types}")
+        for model in excluded_model_types:
+            if not isinstance(model, str):
+                raise ValueError(f"Each entry in `excluded_model_types` must be a string, received {type(model)}")
+            excluded_models.add(normalize_model_type_name(model))
+    return excluded_models
+
+
+def get_hyperparameter_dict(
+    hyperparameters: Optional[Union[str, Dict[str, Union[ModelHyperparameters, List[ModelHyperparameters]]]]],
```
shchur: Nitissimo: `Union[str, ..., None]` feels more natural to me than `Optional[str, ...]` if there is no default value set to `None`.

timeseries/src/autogluon/timeseries/models/presets.py
```diff
@@ -269,22 +153,60 @@ def get_preset_models(
...
+                model = MultiWindowBacktestingModel(model_base=model, name=model.name, **model_type_kwargs)  # type: ignore
 
-            all_assigned_names.add(model.name)
+            all_assigned_names.append(model.name)
             models.append(model)
 
     return models
 
 
+def get_excluded_models(excluded_model_types):
```
shchur: Type hints missing here

timeseries/src/autogluon/timeseries/models/registry.py
```diff
@@ -0,0 +1,59 @@
...
+    def get_model_class(cls, alias: Union[str, type]) -> type:
+        return cls._get_model_record(alias).model_class
+
+    @classmethod
+    def get_model_priority(cls, alias: Union[str, type]) -> int:
+        return cls._get_model_record(alias).default_priority
+
+    @classmethod
+    def available_aliases(cls) -> List[str]:
+        return list(cls.REGISTRY.keys())
```
shchur: How about `sorted(cls.REGISTRY.keys())` to make the potential error message more readable?

timeseries/src/autogluon/timeseries/models/registry.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+@dataclass
+class ModelRecord:
+    model_class: type
+    default_priority: int
+
+
+class ModelRegistry(ABCMeta):
+    REGISTRY: Dict[str, ModelRecord] = {}
```
shchur, line 21: Nit: Should we add a short docstring explaining the purpose of the class?

timeseries/src/autogluon/timeseries/models/gluonts/models.py
```diff
@@ -199,6 +203,8 @@ class TemporalFusionTransformerModel(AbstractGluonTSModel):
         If True, ``lightning_logs`` directory will NOT be removed after the model finished training.
     """
 
+    default_priority = 45
```
shchur: Let's add `_aliases = ["TFT"]` for the TFT model while we are at it?

timeseries/src/autogluon/timeseries/models/registry.py
```diff
@@ -0,0 +1,59 @@
...
+from typing import Dict, List, Union
+
+
+@dataclass
+class ModelRecord:
+    model_class: type
+    default_priority: int
+
+
+class ModelRegistry(ABCMeta):
```
shchur, line 13: Should we add some unit tests for the ModelRegistry? E.g. defining a new class with an existing alias fails, etc.

timeseries/tests/unittests/models/test_registry.py
```diff
@@ -0,0 +1,84 @@
...
+
+
+def test_when_multiple_models_with_same_alias_registered_then_value_error_raised():
+    try:
+
+        class BazModel(metaclass=ModelRegistry):
+            _aliases = ["Qux"]
+            pass
+
+        class QuxModel(metaclass=ModelRegistry):
```
shchur, line 67: Should we use `with pytest.raises(...)` here to assert that the exception is raised?

canerturkmen: huh. I had done this, interesting the branch didn't sync...

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
```diff
@@ -376,11 +377,13 @@ def convert_to_refit_full_template(self) -> Self:
         return template
 
 
-class AbstractTimeSeriesModel(TimeSeriesModelBase, TimeSeriesTunable, ABC):
+class AbstractTimeSeriesModel(TimeSeriesModelBase, TimeSeriesTunable, metaclass=ModelRegistry):
     """Abstract base class for all time series models that take historical data as input and
     make predictions for the forecast horizon.
     """
 
+    default_priority: int = 0
```
Innixma: Should we align the name of the priority var across tabular and timeseries?

Similarly, should we have `ag_name` and `ag_key`? This is what I use for the model registry logic in Tabular.

```python
class CatBoostModel(AbstractModel):
    ag_key = "CAT"
    ag_name = "CatBoost"
    ag_priority = 70
```

canerturkmen: My answer again to the "should we align?" question is again "not without a compelling reason." Independently, I like `ag_priority` better so will use that. Analogues for key and name don't exist anyway so won't use those.

PR #5102, opened 2025-04-29
[timeseries] reorganize gluonts namespace, fix minor typing issue

canerturkmen:
*Issue #, if available:*

- Reorganizes `gluonts` namespace, separating out `dataset` and removing the redundant `torch.*` namespace.
- Fixes minor typing issue (use of pipe instead of Union) in `metrics`.

*Description of changes:*

### DIFF
--- a/timeseries/src/autogluon/timeseries/metrics/abstract.py
+++ b/timeseries/src/autogluon/timeseries/metrics/abstract.py
@@ -240,12 +240,12 @@ class TimeSeriesScorer:
     @overload
     @staticmethod
     def check_get_horizon_weight(
-        horizon_weight: Sequence[float] | np.ndarray, prediction_length: int
+        horizon_weight: Union[Sequence[float], np.ndarray], prediction_length: int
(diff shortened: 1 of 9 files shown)


timeseries/src/autogluon/timeseries/metrics/abstract.py
```diff
@@ -240,12 +240,12 @@ def check_get_horizon_weight(horizon_weight: None, prediction_length: int) -> No
     @overload
     @staticmethod
     def check_get_horizon_weight(
-        horizon_weight: Sequence[float] | np.ndarray, prediction_length: int
+        horizon_weight: Union[Sequence[float], np.ndarray], prediction_length: int
```
abdulfatir, line 243: Should we `from __future__ import annotations` instead?

canerturkmen: this is to keep consistent across the module for now. we should make the switch all at once imo.

PR #5116, opened 2025-05-03
Fix formula typo

PangLuo:
*Issue #, if available:*

*Description of changes:*
The original formula was misleading.


PangLuo:
I think the original formula was good. My previous understanding was incorrect.

PR #5131, opened 2025-05-20
[timeseries] Avoid masking the 'scaler' param with the default 'target_scaler' value

shchur:
*Issue #, if available:*

*Description of changes:*
- Prior to v1.3.0, users could set the `scaler` hyperparameter to choose the scaling option in MLForecast models. In v1.3.0, we renamed this hyperparameter to `target_scaler` for consistency with other models. We intended to support the `scaler` hyperparameter as well for backwards compatibility. However, because of a bug in the logic, `scaler` was always shadowed by the default value of the `target_scaler`, so the old parameter name got effectively deprecated. This can result in unexpected performance changes to the users. This PR fixes this problem.

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
+++ b/timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
@@ -2,6 +2,7 @@ import logging
 import math
 import os
 import time
+import warnings
 from typing import Any, Callable, Collection, Dict, List, Optional, Tuple, Union
 
 import numpy as np
@@ -190,8 +191,17 @@ class AbstractMLForecastModel(AbstractTimeSeriesModel):
             target_transforms.append(Differences(differences))
             self._sum_of_differences = sum(differences)
 
+        # TODO: Remove 'scaler' hyperparameter in v2.0.0
         # Support "scaler" for backward compatibility
-        scaler_type = model_params.get("target_scaler", model_params.get("scaler"))
+        if "scaler" in model_params:
+            warnings.warn(
+                f"Hyperparameter 'scaler' for {self.__class__.__name__} has been deprecated "
+                "and will be removed in v2.0. Please use 'target_scaler' instead.",
+                category=FutureWarning,
+            )
+            scaler_type = model_params["scaler"]
+        else:
+            scaler_type = model_params.get("target_scaler")
         if scaler_type is not None:
             self._scaler = MLForecastScaler(scaler_type=scaler_type)
             target_transforms.append(self._scaler)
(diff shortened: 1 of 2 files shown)


timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
```diff
@@ -191,7 +191,10 @@ def _get_mlforecast_init_args(
             self._sum_of_differences = sum(differences)
 
         # Support "scaler" for backward compatibility
-        scaler_type = model_params.get("target_scaler", model_params.get("scaler"))
+        if "scaler" in model_params:
+            scaler_type = model_params["scaler"]
```
shchur, line 195: We could, in theory, add a `FutureWarning` here and deprecate this parameter in v2.0. However, this looks like such a minor overhead to maintain that I'm leaning towards just keeping this around forever.

timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
```diff
@@ -191,7 +191,10 @@ def _get_mlforecast_init_args(
             self._sum_of_differences = sum(differences)
 
         # Support "scaler" for backward compatibility
-        scaler_type = model_params.get("target_scaler", model_params.get("scaler"))
+        if "scaler" in model_params:
```
abdulfatir, line 194: What happens if user provides both `scaler` and `target_scaler`? Should we raise or warn?

timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
abdulfatir: It should be okay to keep this around imo.

timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
abdulfatir: Also, does this imply any documentation changes?

shchur: Before this PR: `scaler` is always ignored, only `target_scaler` matters.
This PR (at the time of review): `scaler` takes precedence over `target_scaler` (probably not ideal)

I have updated the PR so that `target_scaler` takes precedence over `scaler` - that seems more reasonable to me.

shchur: I've also added a warning to avoid introducing a potentially breaking change in a hotfix release.

abdulfatir (approved):
Looks good to me, but at this stage maybe it's better to just make it a deprecation warning for future clarity. I'll let you decide.

timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
shchur: @abdulfatir, regarding the warnings category, `DeprecationWarning`s are often silenced by the default warning filters (see #3463 or [PEP565](https://peps.python.org/pep-0565/)), so I would prefer to keep the `UserWarning` here to ensure that it's displayed.

PR #5132, opened 2025-05-20
[timeseries] Fix ensemble weights format for printing

shchur:
*Issue #, if available:*

*Description of changes:*
- With `numpy>2.0` the ensemble weights are printed as follows:
    ```
    Ensemble weights: {'Chronos[bolt_small]': np.float64(0.74), 'DirectTabular': np.float64(0.04), 'ETS': np.float64(0.02), 'RecursiveTabular': np.float64(0.19)}
    ```
    This PR ensures that the weights are displayed as floats (and not np.float64 dtype).

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/ensemble/greedy.py
+++ b/timeseries/src/autogluon/timeseries/models/ensemble/greedy.py
@@ -182,5 +182,5 @@ class GreedyEnsemble(AbstractWeightedTimeSeriesEnsembleModel):
             if weight != 0:
                 self.model_to_weight[model_name] = weight
 
-        weights_for_printing = {model: round(weight, 2) for model, weight in self.model_to_weight.items()}
+        weights_for_printing = {model: round(float(weight), 2) for model, weight in self.model_to_weight.items()}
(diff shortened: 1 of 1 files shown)


timeseries/src/autogluon/timeseries/models/ensemble/greedy.py
```diff
@@ -182,5 +182,5 @@ def _fit(
             if weight != 0:
                 self.model_to_weight[model_name] = weight
 
-        weights_for_printing = {model: round(weight, 2) for model, weight in self.model_to_weight.items()}
+        weights_for_printing = {model: round(float(weight), 2) for model, weight in self.model_to_weight.items()}
```
abdulfatir, line 185: The nittest of picks: Not sure if `weight.item()` is better than casting.

abdulfatir (approved):

timeseries/src/autogluon/timeseries/models/ensemble/greedy.py
shchur: I think `float(something)` is slightly better here because it will work even if the `weight` values have type `float` (and not `np.float64`).

PR #5150, opened 2025-06-04
Use pyarrow by default, remove fastparquet

Innixma:
*Issue #, if available:*

*Description of changes:*

- Use pyarrow by default, remove fastparquet

This aligns with panda's choice to adopt pyarrow as a required dependency in pandas 3.0: https://pandas.pydata.org/pdeps/0010-required-pyarrow-dependency.html

Additionally, fastparquet has been deprecated by dask, who are the authors of fastparquet and recommend pyarrow: https://stackoverflow.com/a/77844771

Finally, I noticed that loading parquet files from https using `pd.read_parquet` led to different results depending on if fastparquet or pyarrow was used. In the case I observed, pyarrow loaded the parquet file correctly, whereas fastparquet loaded the file incorrectly.

For the pyarrow version ranges, I chose `>=7` as the lower bound to align with pandas 3.0's planned minimum version bound for pyarrow.

### DIFF
--- a/common/src/autogluon/common/loaders/load_pd.py
+++ b/common/src/autogluon/common/loaders/load_pd.py
@@ -132,12 +132,7 @@ def load(
             multiprocessing_method=multiprocessing_method,
         )
     elif format == "parquet":
-        try:
-            # TODO: Deal with extremely strange issue resulting from torch being present in package,
-            #  will cause read_parquet to either freeze or Segmentation Fault when performing multiprocessing
-            df = pd.read_parquet(path, columns=columns_to_keep, engine="fastparquet")
-        except:
-            df = pd.read_parquet(path, columns=columns_to_keep, engine="pyarrow")
+        df = pd.read_parquet(path, columns=columns_to_keep)
         column_count_full = len(df.columns)
     elif format == "csv":
         df = pd.read_csv(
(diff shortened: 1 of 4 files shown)


common/src/autogluon/common/loaders/load_pd.py
```diff
@@ -132,12 +132,7 @@ def load(
             multiprocessing_method=multiprocessing_method,
         )
     elif format == "parquet":
-        try:
-            # TODO: Deal with extremely strange issue resulting from torch being present in package,
-            #  will cause read_parquet to either freeze or Segmentation Fault when performing multiprocessing
-            df = pd.read_parquet(path, columns=columns_to_keep, engine="fastparquet")
-        except:
-            df = pd.read_parquet(path, columns=columns_to_keep, engine="pyarrow")
+        df = pd.read_parquet(path, columns=columns_to_keep)
```
Innixma, line 135: note, if engine is unspecified, pandas uses pyarrow by default

common/src/autogluon/common/savers/save_pd.py
```diff
@@ -104,12 +104,7 @@ def save(
         if verbose:
             logger.log(15, "Saved " + str(path) + " | Columns = " + str(column_count) + " | Rows = " + str(row_count))
     elif type == "parquet":
-        try:
-            df.to_parquet(
-                path, compression=compression, engine="fastparquet"
-            )  # TODO: Might be slower than pyarrow in multiprocessing
-        except:
-            df.to_parquet(path, compression=compression, engine="pyarrow")
+        df.to_parquet(path, compression=compression)
```
Innixma, line 107: note, if engine is unspecified, pandas uses pyarrow by default

prateekdesai04 (approved):

PR #5151, opened 2025-06-04
upgrade CI containers to use PyTorch 2.6.0, with CUDA 12.6 for gpu variants

FireballDWF:
Upgrades the Docker containers to use PyTorch 2.6.0, with CUDA 12.6 for gpu variants.

With the CUDA 12.6 included, expect it to avoid the incompatible cuDNN error seen at https://github.com/autogluon/autogluon/pull/5089#issuecomment-2941166029 when attempting to use PyTorch 2.7 which the gpu wheel is packaged with CUDA 12.6


FireballDWF:
@tonyhoo Requesting your review/approval/merge.  Not only will this bring the current CICD container base up to date with the current release of autogluon's usage of PyTorch 2.6 which used CUDA 12.6, it will also resolve a conflict with the version of CUDA that PyTorch 2.7 depends on CUDA 12.6 which is the source of the errors for the failing CI tests when attempt to test with PyTorch 2.7 where the current CI image's CUDA 12.4 has version incompatibilities due to being loaded before the CUDA built into the PyTorch wheel.

FireballDWF:
Replaced by https://github.com/autogluon/autogluon/pull/5168

PR #5157, opened 2025-06-09
[Fix] Resolve AttributeError in LinearModel when using RAPIDS cuML models

tonyhoo:
*Issue #, if available:*
N/A

*Description of changes:*

### Problem
RAPIDS cuML linear models (LinearRapidsModel) fail during training with an `AttributeError: n_iter_` because they don't expose the `n_iter_` attribute that scikit-learn models provide. This causes the LinearModel base class to crash when trying to access `model.n_iter_` for iteration tracking.

### Root Cause
- RAPIDS cuML models use GPU-optimized algorithms that don't track iterations the same way as scikit-learn
- The `n_iter_` attribute is not available in cuML implementations
- AutoGluon's LinearModel base class assumes all linear models have this attribute

### Solution
Replace direct attribute access with safe `getattr()` calls:
- Use `getattr(model, 'n_iter_', None)` instead of `model.n_iter_`

### Testing
- Verified with RAPIDS cuML lr models
- Confirmed scikit-learn models continue to work as expected
- No regression in existing LinearModel functionality
```from autogluon.tabular.models.knn.knn_rapids_model import KNNRapidsModel
from autogluon.tabular.models.lr.lr_rapids_model import LinearRapidsModel
from autogluon.tabular.models.lr.lr_model import LinearModel
predictor = TabularPredictor(label=label).fit(train_data,
                                              hyperparameters={
        KNNRapidsModel: {},
        LinearRapidsModel: {},
        LinearModel: {},
        
    },
                                              )
                                              
predictor.leaderboard(test_data)
```

| model | score_test | score_val | eval_metric | pred_time_test | pred_time_val | fit_time | pred_time_test_marginal | pred_time_val_marginal | fit_time_marginal | stack_level | can_infer | fit_order |
|-------|------------|-----------|-------------|----------------|---------------|----------|-------------------------|------------------------|-------------------|-------------|-----------|-----------|
| LinearModel | 0.839697 | 0.80 | accuracy | 0.027756 | 0.018365 | 0.644307 | 0.027756 | 0.018365 | 0.644307 | 1 | True | 2 |
| LinearModel_2 | 0.839595 | 0.80 | accuracy | 0.018174 | 0.011734 | 1.089793 | 0.018174 | 0.011734 | 1.089793 | 1 | True | 3 |
| WeightedEnsemble_L2 | 0.829051 | 0.84 | accuracy | 0.034355 | 0.281068 | 8.620535 | 0.001574 | 0.000757 | 0.036132 | 2 | True | 4 |
| KNeighbors | 0.726072 | 0.73 | accuracy | 0.005024 | 0.261946 | 7.940096 | 0.005024 | 0.261946 | 7.940096 | 1 | True | 1 |

                                      

#

### DIFF
--- a/tabular/src/autogluon/tabular/models/lr/lr_rapids_model.py
+++ b/tabular/src/autogluon/tabular/models/lr/lr_rapids_model.py
@@ -1,7 +1,5 @@
 import logging
 
-import numpy as np
-
 from autogluon.common.utils.try_import import try_import_rapids_cuml
 from autogluon.core.constants import REGRESSION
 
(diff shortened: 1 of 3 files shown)


tonyhoo:
> Note: this may introduce its own bug. Need to verify if cuML is actually fitting correctly in the incremental way we do it in AutoGluon.

Noted! Great catch. I just checked the [cuML API](https://docs.rapids.ai/api/cuml/stable/api/#linear-regression), and it seems there's no `warm_start` support. This could lead to issues with our incremental training method in AutoGluon and disrupt our time management. I'll investigate further and look for an alternative solution that won't affect our current code.

tonyhoo:
@Innixma  Implemented the customized `_fit` function based on the Rapids cuML API

tabular/src/autogluon/tabular/models/lr/lr_rapids_model.py
```diff
@@ -51,10 +49,52 @@ def _set_default_params(self):
...
         X = super()._preprocess(X=X, **kwargs)
-        if not isinstance(X, np.ndarray):
+        if hasattr(X, 'toarray'):  # Check if it's a sparse matrix
             X = X.toarray()
         return X
 
     def _fit(self, X, y, **kwargs):
-        kwargs.pop("sample_weight", None)  # sample_weight is not supported
-        super()._fit(X=X, y=y, **kwargs)
+        """
```
Innixma, line 57: Are we not using `num_gpus` argument to the `_fit` call? How do we tell cuML how many GPUs to use?

tabular/src/autogluon/tabular/models/lr/lr_rapids_model.py
```diff
@@ -51,10 +49,52 @@ def _set_default_params(self):
...
+            if 'C' in self.params:
+                filtered_params['alpha'] = 1.0 / self.params['C']
+        else:
+            # For classification, keep C parameter
+            if 'C' in self.params:
+                filtered_params['C'] = self.params['C']
+
+        # Create and fit cuML model - let cuML handle its own error messages
+        self.model = model_cls(**filtered_params)
+        self.model.fit(X, y)
```
Innixma, line 97: We should also change how Rapids interacts with bagging. Probably safest to disable parallel bagging for now until we understand how rapids assigns GPUs to models if >1 GPU exists.


In `autogluon.tabular.models._utils.rapids_utils.RapidsModelMixin`:

```python
    @classmethod
    def _get_default_ag_args_ensemble(cls, **kwargs) -> dict:
        default_ag_args_ensemble = super()._get_default_ag_args_ensemble(**kwargs)
        extra_ag_args_ensemble = {"use_child_oof": False}
        default_ag_args_ensemble.update(extra_ag_args_ensemble)
        return default_ag_args_ensemble
```

change to:

```python
    @classmethod
    def _get_default_ag_args_ensemble(cls, **kwargs) -> dict:
        default_ag_args_ensemble = super()._get_default_ag_args_ensemble(**kwargs)
        extra_ag_args_ensemble = {"use_child_oof": False, "fold_fitting_strategy": "sequential_local"}
        default_ag_args_ensemble.update(extra_ag_args_ensemble)
        return default_ag_args_ensemble

```

tabular/src/autogluon/tabular/models/lr/lr_rapids_model.py
tonyhoo: cuML automatically detects available GPUs using CUDA runtime. The linear model only support single GPU and we currently set it at [here](https://github.com/autogluon/autogluon/blob/master/tabular/src/autogluon/tabular/models/_utils/rapids_utils.py#L19)

Issue #5155, opened 2025-06-08
[BUG] learning_curves.plot_curves incorrect method signature


james-yun:
> Tagging @adibiasio who added the learning_curve feature in https://github.com/autogluon/autogluon/pull/4411, perhaps this is something you could clean up?

**Describe the bug**
Been playing around with the new learning curve (it's a pretty cool feature!)

However, there is a contradiction in the [`plot_curves` function](https://github.com/autogluon/autogluon/blob/b1ab1acc5503320df39e40f33ad6be8d88a4fe5c/core/src/autogluon/core/learning_curves/plot_curves.py#L8-L17) signature, namely the `learning_curves` parameter.

The function signature states `learning_curves` this should be a `dict`.

However, function documentation states `learning_curves` should be a `Tuple[dict, dict]`.

I was able to get it to work by passing in the `Tuple[dict, dict]` returned by `TabularPredictor.learning_curves()`.

In the source code, the first dict in the tuple is dropped. 

I see two options:
1. Function signature needs to be changed to `Tuple[dict, dict]`. Documentation and example likewise need to be updated to reflect how it currently can be used.
2. [BREAKING] update function to documentation and implementation to use the `dict` parameter. This will break existing usages of `learning_curves()`.

Given that 2 is a breaking change, Option 1 is probably the safer move.

Out of scope for this issue, but would be nice to have some sort of "type-checking" unit test to catch incongruities like this one. 

---

(a slightly different issue)

I also had to `pip install seaborn` to get it to work, but not sure if it's worth adding that as a dependency ([it's listed as a dependency in the `autogluon.eda` sub-package](https://github.com/autogluon/autogluon/blob/b1ab1acc5503320df39e40f33ad6be8d88a4fe5c/eda/setup.py#L30C6-L30C13), but that [appears to be deprecated](https://auto.gluon.ai/stable/install.html)).

**Expected behavior**

**To Reproduce**
```python
from autogluon.tabular import TabularPredictor
predictor = TabularPredictor(label="age").fit(
    train_data='https://autogluon.s3.amazonaws.com/datasets/Inc/train.csv',
    learning_curves=True,
)
curves = predictor.learning_curves()

from autogluon.core.learning_curves.plot_curves import plot_curves
plot_curves(curves, "XGBoost", "root_mean_squared_error")
```

**Screenshots / Logs**
![Image](https://github.com/user-attachments/assets/6add5543-af6d-4f3c-8bd0-2b1841a4a326)

[...]

PR #5158, opened 2025-06-10
[Fix] Changed Plot Curves Method Signature Type

adibiasio:
**Problem:** Inconsistent method signature types in plot_curves.py (see issue #5155)

**Solution:** Updated the [plot_curves ](https://github.com/autogluon/autogluon/blob/b1ab1acc5503320df39e40f33ad6be8d88a4fe5c/core/src/autogluon/core/learning_curves/plot_curves.py#L8-L17) method signature type for the learning_curves parameter to align with the current implementation. Switched from `dict` to `tuple[dict, dict]` and updated documentation.

### DIFF
--- a/core/src/autogluon/core/learning_curves/plot_curves.py
+++ b/core/src/autogluon/core/learning_curves/plot_curves.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import matplotlib.pyplot as plt
 import numpy as np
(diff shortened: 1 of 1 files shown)


Innixma:
Thanks for the contribution @adibiasio and the suggestion @james-yun! Incorporated the suggestion and merging in to be included in the v1.4 release. Cheers!

PR #5159, opened 2025-06-11
[timeseries] Optimize groupby operations using indptr

shchur:
*Issue #, if available:*

**Description of changes:**
This PR optimizes the implementation of several methods throughout the TS module that resulted in noticeable overhead when working with large datasets (>10M rows). These optimizations result in up to 7x E2E speedups when using just a single model, and even bigger speedups when using multiple models.

Improved methods:
- `TimeSeriesDataFrame`
  - `_validate_multi_index_data_frame` - avoid copying the index when checking the dtype of item_ids
  - `num_timesteps_per_item` - replace `groupby` with `value_counts`
  - `slice_by_timestep` - replace `groupby` with a clever algorithm + `indptr` tricks
  - `infer_frequency` - replace `groupby` with `indptr` tricks
- `AbstractLocalModel`
  - `_get_dummy_forecast` - subsample the data when computing statistics for the dummy forecast
  - `_predict` - replace `groupby` with `indptr` tricks
- `TimeSeriesPredictor`
  - `make_future_data_frame` - replace `groupby` with `indptr` tricks
- API documentation: hide some internal TSDF methods from the sidebar (`get_indptr`, `sort_index`, `assign`).
- Use the spelling "dataframe" across the docs for consistency (previously we mixed "data frame" and "dataframe").
-----

**Runtime analysis**
- London Smart Meters (LSM) - 166M rows, 5.5K items, no covariates
- M5 - 46M rows, 30K items, 8 real-valued covariates

Total runtime of `TimeSeriesPredictor(...).fit(data, hyperparameters=hyperparameters)` for different values of `hyperparameters`:
Runtime (s) | M5 |   | LSM |  
:-- | -- | -- | -- | --
**hyperparameters** | **before** | **this PR** | **before** | **this PR**
`{"SeasonalNaive": {}}` | 113 | 43 | 248 | 39
`{"Chronos": {}}` | 88 | 39 | 172 | 40
`{"SeasonalNaive": {}, "Naive": {}, "Chronos": {}}` | 312 | 126 | 630 | 94

Runtime of different individual `TimeSeriesDataFrame` methods:

Runtime (s) | M5 |   | LSM |  
:-- | -- | -- | -- | --
**TSDF method** | **before** | **this PR** | **before** | **this PR**
`_validate_multi_index_data_frame(df)` | 0.9 | 0.0 | 3.2 | 0.0
`num_timesteps_per_item()` | 3.2 | 0.1 | 9.6 | 0.5
`slice_by_timestep(None, -10)` | 4.8 | 1.4 | 13.9 | 2.9
`infer_frequency()` | 8.0 | 3.2 | 13.6 | 3.2

-----

**Breaking changes**
This PR does not introduce any breaking changes to the API. Some methods such as `TimeSeriesDataFrame.infer_frequency()` become slower on data with an unsorted index, but that should not affect the vast majority of the users.

----

### DIFF
--- a/timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
+++ b/timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
@@ -726,6 +726,7 @@ class TimeSeriesDataFrame(pd.DataFrame):
             # We put 1 at the slice_start index for each item and -1 at the slice_end index for each item.
             # After we apply cumsum we get the indicator mask selecting values between slice_start and slice_end
             # cumsum([0, 0, 1, 0, 0, -1, 0]) -> [0, 0, 1, 1, 1, 0, 0]
+            # We need array of size len(self) + 1 in case events[starts + slice_end] tries to access position len(self)
             events = np.zeros(len(self) + 1, dtype=np.int8)
             events[starts + slice_start] += 1
             events[starts + slice_end] -= 1


timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
```diff
@@ -469,49 +469,52 @@ def infer_frequency(self, num_items: Optional[int] = None, raise_if_irregular: b
             If some items have an irregular frequency or if different items have different frequencies, returns string
             `IRREG`.
         """
+        self = self._ensure_index_is_sorted(self)
```
canerturkmen: Several issues here. 

- It's generally bad style to assign to self/this/etc. since it's highly misleading. It's really easy to start assuming you've mutated the self object looking at this line but you haven't. this just rebinds the name `self` which is a special name and shouldn't be rebound.  
- "ensure index is sorted" sounds a lot like a mutative method, but it isn't. it's a static method. (Yet you're passing in `self` which is confusing.) I would just call this `sort_index` or something similar.
- Third (nit), if we insist it's a static method, it's better to call it through the class namespace. `self.method_name()` kind of suggests you're calling it on an instance, or the class of an instance, etc.

so this should be something like
```
sorted_tsdf = sort_index(self)
```

or 
```
self.sort_index(inplace=True)
```

etc.

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
```diff
@@ -731,9 +773,23 @@ def from_pickle(cls, filepath_or_buffer: Any) -> TimeSeriesDataFrame:
         except Exception as err:  # noqa
             raise IOError(f"Could not load pickled data set due to error: {str(err)}")
 
+    @staticmethod
+    def _ensure_index_is_sorted(ts_df: TimeSeriesDataFrame, method_name: Optional[str] = None) -> TimeSeriesDataFrame:
```
canerturkmen: I generally disagree with passing in `method_name` just for the purposes of logging. It immediately signals to me "index will be sorted differently based on which method is calling" or "this specifies the sorting algorithm for sorting". I don't think we lose anything if the log doesn't feature the method name.

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
```diff
@@ -780,19 +836,14 @@ def fill_missing_values(self, method: str = "auto", value: float = 0.0) -> TimeS
                 2019-02-07     4.0
 
         """
+        self = self._ensure_index_is_sorted(self, method_name="fill_missing_values")
```
canerturkmen: Same comments for `self`s as above.

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
```diff
@@ -1044,6 +1096,13 @@ def to_data_frame(self) -> pd.DataFrame:
         """Convert `TimeSeriesDataFrame` to a `pandas.DataFrame`"""
         return pd.DataFrame(self)
 
+    def get_indptr(self) -> np.ndarray:
```
canerturkmen, line 1109: Why is this method in the Public API? Should we just make it private? If we decide to go the "make TimeSeriesDataFrame" a facade route, this will not be relevant.

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
```diff
@@ -593,7 +598,8 @@ def slice_by_timestep(
         This operation is equivalent to selecting a slice ``[start_index : end_index]`` from each time series, and then
         combining these slices into a new ``TimeSeriesDataFrame``. See examples below.
 
-        Returns a copy of the original data. This is useful for constructing holdout sets for validation.
+        .. note::
```
canerturkmen: I think this is a potentially big breaking change. Same comments for `fill_missing_values`. 

I understand the predictor behavior is unaffected, but two counterarguments:
- It's quite a side-effect to ask to slice something and get something sorted as a necessary side effect. 
- Pushing ourselves towards a case where "everything is sorted for consistency" may be cornering us. What if we end up with datasets that can only be memmapped? Will we sort on disk first? i.e., adopting this convention is both breaking and a one-way choice.

If there's any plausible way of making this work without auto-sorting, I would urge us towards that. 

Also, do the speedups already incorporate sorting compute and memory overhead?

canerturkmen (commented):
Thanks for this! Impressive speedups. I have a few concerns both in implementation and the breaking changes themselves.

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
shchur: We use this method in Chronos and GluonTS models, so making it public seems like the obvious choice. We can hide it in the docs to make it less likely that users discover and depend on it.

shchur: Regarding the "making TSDF a facade route" - if we do that and adopt composition over inheritance, this will anyway be very significant breaking change. If the new class looks something like this:
```python
class NewTimeSeriesDataFrame:
    data: pd.DataFrame
    static_features: pd.DataFrame | None
    ...
```
then most existing methods like `fillna`, `dropna`, `to_csv`, selecting a column, inserting a column, will break.

The only ways to preserve this functionality that I see are:
1. manually implement all methods that we currently inherit from `pd.DataFrame` - looks tedious and does not cover operations like select / insert column
2. inherit from pd.DataFrame - this is the current design that I thought we should get away from

In summary, I wouldn't worry about the potential future change now since it will likely introduce a new dataset class with a different API  - unless you have another idea for how to implement this change.

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
shchur: Currently both `fill_missing_values` and `slice_by_timestep` require the entire dataset to be loaded into memory, and I'm not sure how we can easily get away from this requirement. There can always be rows at the bottom of the data frame corresponding to the items above, which will affect the result.
```
item_id  timestamp target
A            2020-01-01
A            2020-01-02
A            2020-01-03
B            2020-01-01            
...  # suppose we have 1B rows here
Z            2020-01-01
A            2020-01-04
A            2020-01-05
```
This means `fill_missing_values` and `slice_by_timestep` will always require loading the entire dataset into memory, as they do now.

Regarding the potential future refactor issue - see my comment below. I think that a future refactor will in any case not be backward-compatible and will require introducing a new class, so it should not be a limiting factor for us now.

IMO the main purpose of the `TimeSeriesDataFrame` class is to format the data such that it can be consumed by the `TimeSeriesPredictor`, and the operations should be optimized for their use inside the predictor. Since we anyway sort the data in the predictor, it's fine to optimize the operations for sorted data. Also, many operations already depend on the data being sorted and will break otherwise - e.g., Chronos / GluonTS models, making the forecast index.

As an alternative to automatically sorting, we can warn + try to perform the operation anyway, even if the data is not sorted. This, however, may return nonsense results (also true for the current master branch, even before this PR). Do you that would be a better option?

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
shchur: Thanks, this makes sense.

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
shchur: Re: taking sorting time into account. We always apply sorting (if the index is not already sorted) in the `TimeSeriesPredictor._check_and_prepare_data_frame`, so I don't think that's relevant for the comparison.

- If we decide that `slice_by_timestep` and `fill_missing_values` sort automatically if the index is not sorted, then this PR may increase their runtime when applied on unsorted data outside of the predictor.
- If we decide that `slice_by_timestep` and `fill_missing_values` only warn on unsorted data and attempt the operation (potentially returning nonsense results), this PR will only speed them up.

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
```diff
@@ -680,10 +691,52 @@ def slice_by_timestep(
...
+            # After we apply cumsum we get the indicator mask selecting values between slice_start and slice_end
+            # cumsum([0, 0, 1, 0, 0, -1, 0]) -> [0, 0, 1, 1, 1, 0, 0]
+            events = np.zeros(len(self) + 1, dtype=np.int8)
+            events[starts + slice_start] += 1
+            events[starts + slice_end] -= 1
+            mask = np.cumsum(events)[:-1].astype(bool)
+            # loc[mask] returns a view of the original data - modifying it will produce a SettingWithCopyWarning
+            return self.loc[mask]
+        else:
+            # Fall back to a slow groupby operation
```
shchur, line 737: We have this option available for backward compatibility. For example, the old groupby slicing will work correctly on the following data
```
item_id  timestamp   target
A        2025-01-01  ...
B        2025-01-01  ...
A        2025-01-02  ...
B        2025-01-02  ...
A        2025-01-03  ...
B        2025-01-03  ...
```
but the new indptr slicing will produce an incorrect result.

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
```diff
@@ -529,8 +534,13 @@ def num_items(self):
         return len(self.item_ids)
 
     def num_timesteps_per_item(self) -> pd.Series:
-        """Length of each time series in the dataframe."""
-        return self.groupby(level=ITEMID, sort=False).size()
+        """Number of observations in each time series in the dataframe.
+
+        Returns a `pandas.Series` with item_id as index and number of observations per item as values.
+        """
+        counts = pd.Series(self.index.codes[0]).value_counts(sort=False)
```
abdulfatir, line 541: Nice! Can we also do `codes[1]` and then do `max` + 1?

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
```diff
@@ -680,10 +691,52 @@ def slice_by_timestep(
...
+                # Return empty dataframe with same structure
+                return self.loc[np.zeros(len(self), dtype=bool)]
+
+            starts = starts[valid_slices]
+            slice_start = slice_start[valid_slices]
+            slice_end = slice_end[valid_slices]
+
+            # We put 1 at the slice_start index for each item and -1 at the slice_end index for each item.
+            # After we apply cumsum we get the indicator mask selecting values between slice_start and slice_end
+            # cumsum([0, 0, 1, 0, 0, -1, 0]) -> [0, 0, 1, 1, 1, 0, 0]
```
abdulfatir, line 728: Creative!

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
```diff
@@ -680,10 +691,52 @@ def slice_by_timestep(
...
+            events = np.zeros(len(self) + 1, dtype=np.int8)
```
abdulfatir, line 730: Why +1?

shchur: We need array of size `len(self) + 1` here in case `events[starts + slice_end] -= 1` tries to access the index  `len(self)` to "terminate" the selection indicator for the last time series. This happens whenever `end_index=None`. If we didn't have the +1 here, we would get an IndexError (index `len(self)` is out of bounds for axis with size `len(df)`).

Added a comment to clarify this.

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
shchur: 🤖

timeseries/src/autogluon/timeseries/dataset/ts_dataframe.py
shchur: Could you clarify this part? Not sure about `codes[1]` (that would return the codes for the `timestamp` index level).

abdulfatir: Nevermind. I misunderstood.

PR #5168, opened 2025-06-17
Upgrade image to use torch 2.7.1

tonyhoo:
*Issue #, if available:*

*Description of changes:*
This should resolve CI failure for torch upgrade at [here](https://github.com/autogluon/autogluon/actions/runs/15697314474/job/44296695040) which is due to cuDNN imcompataibilty


prateekdesai04 (approved):
Thanks for the fix!

Issue #4998, opened 2025-03-27
[BUG][timeseries] TFT fails if a new categorical value is encountered in the last prediction_length time steps of the training data


shchur:
**Bug Report Checklist**

- [ ] I provided code that demonstrates a minimal reproducible example.  
- [x] I confirmed bug exists on the latest mainline of AutoGluon via source install.  
- [x] I confirmed bug exists on the latest stable version of AutoGluon.  

**Describe the bug**
During model training, we hold out the last `prediction_length` time steps of each time series as the validation set. If the hold-out contains a categorical known covariate not seen in the training set, the TFT model will fail. Specifically, the `nn.Embedding` layer will raise an `IndexError` because it receives an integer value larger than the expected cardinality.

This happens because TFT infers the feature cardinalities using only the training split (not including the hold-out data!):
https://github.com/autogluon/autogluon/blob/a87a425d90c7acb3858b3742544ca755da370aad/timeseries/src/autogluon/timeseries/models/gluonts/abstract_gluonts.py#L269

One potential fix is to store the categorical feature cardinalities computed on the entire training data inside the `CovariateMetadata` object.

**To Reproduce**
```python
from autogluon.timeseries import TimeSeriesDataFrame, TimeSeriesPredictor
import numpy as np

data = TimeSeriesDataFrame.from_path("https://autogluon.s3.amazonaws.com/datasets/timeseries/m4_hourly/train.csv",)
data = data.loc[data.item_ids[[0]]]
h = 48
values = np.random.choice(["foo", "bar"], size=len(data)).astype("O")
values[-h:] = "hello"
data["feat"] = values
predictor = TimeSeriesPredictor(prediction_length=h, known_covariates_names=["feat"])
predictor.fit(data, hyperparameters={"TemporalFusionTransformer": {"trainer_kwargs": {"accelerator": "cpu"}}}, time_limit=30)
```

**Screenshots / Logs**
```
	Warning: Exception caused TemporalFusionTransformer to fail during training... Skipping this model.
	index out of range in self
Traceback (most recent call last):
  File "/home/shchuro/uv_envs/stable/lib/python3.11/site-packages/autogluon/timeseries/trainer/abstract_trainer.py", line 525, in _train_and_save
    model = self._train_single(train_data, model, val_data=val_data, time_limit=time_limit)
[...]

PR #5176, opened 2025-06-26
[WIP] [timeseries] Add fine-tuning with covariates support for Chronos-Bolt

abdulfatir:
*Issue #, if available:* N/A

*Description of changes:* TBD. 

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/chronos/pipeline/utils.py
+++ b/timeseries/src/autogluon/timeseries/models/chronos/pipeline/utils.py
@@ -134,6 +134,8 @@ class CustomTrainer(Trainer):
             else:
                 optimizer_cls, optimizer_kwargs = self.get_optimizer_cls_and_kwargs(self.args, opt_model)
 
+            print(optimizer_kwargs)
+
             self.optimizer = optimizer_cls(optimizer_grouped_parameters, **optimizer_kwargs)
 
         return self.optimizer


timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -410,6 +438,34 @@ def _validate_and_assign_attributes(self, model_params: dict):
             )
             self.context_length = self.maximum_context_length
 
+    def _deferred_init_covariate_hyperparameters(self, model_params: dict, dataset: TimeSeriesDataFrame) -> None:
```
shchur, line 443: How about we just save the `static_cardinalities`, `past_dynamic_cardinalities` and `dynamic_cardinalities` inside the `CovariateMetadata` when we fit the `TimeSeriesFeatureGenerator`? Then inside the model we can set all of 
```
            dynamic_dims=self.dynamic_dims,
            past_dynamic_dims=self.past_dynamic_dims,
            static_dims=self.static_dims,
            static_cardinalities=self.static_cardinalities,
            dynamic_cardinalities=self.dynamic_cardinalities,
            past_dynamic_cardinalities=self.past_dynamic_cardinalities,
``` 
at model initialization time from the `self.covariate_metadata`.

This logic where the cardinalities are inferred based on the training slice of the data can produce some annoying bugs (#4998), and it feels more elegant if every modes does not need to figure out the "dataset schema" on its own.

shchur: If go down this route, this entire method could be skipped.

timeseries/src/autogluon/timeseries/models/chronos/pipeline/chronos.py
```diff
@@ -529,6 +529,16 @@ def from_pretrained(cls, *args, **kwargs):
         """
         kwargs = kwargs.copy()
 
+        # Remove all covariate-related kwargs because Chronos does not support fine-tuning with covariates
+        dynamic_dims = kwargs.pop("dynamic_dims", 0)
+        past_dynamic_dims = kwargs.pop("past_dynamic_dims", 0)
+        static_dims = kwargs.pop("static_dims", 0)
+        static_cardinalities = kwargs.pop("static_cardinalities", [])
+        dynamic_cardinalities = kwargs.pop("dynamic_cardinalities", [])
+        past_dynamic_cardinalities = kwargs.pop("past_dynamic_cardinalities", [])
```
shchur, line 538: Do we need to set these to variables if they are just ignored afterwards? Maybe some filtering would be simpler:
```python
ignored_kwargs = ["dynamic_dims", "past_dynamic_dims", ...]
kwargs = {k: v for k, v in kwargs.items() if k not in ignored_kwargs}
```

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -212,6 +223,14 @@ def __init__(
...
         self._model_pipeline: Optional[Any] = None  # of type BaseChronosPipeline
 
+        # Following attributes may be overridden during fit() based on train_data & model
+        self.dynamic_dims = 0
+        self.past_dynamic_dims = 0
+        self.static_dims = 0
+        self.static_cardinalities: List[int] = []
+        self.dynamic_cardinalities: List[int] = []
+        self.past_dynamic_cardinalities: List[int] = []
+
```
shchur, line 233: Probably a nit: How about we name these variables more in line with their names in autogluon, something like
```python
self.num_static_real = 0
self.card_static_cat: List[int] = []

self.num_known_real = 0
self.card_known_cat: List[int] = []

self.num_past_real = 0
self.card_past_cat: List[int] = []
```
I think the GluonTS names come from the MXNet implementation of TFT, and I'm not a huge fan: dims sounds like it's a `list[int]` and not an `int`, and `dynamic_...` vs. `past_dynamic_...` sounds like they are two parts of the same object (even though they are unrelated).

timeseries/src/autogluon/timeseries/models/chronos/pipeline/chronos_bolt.py
```diff
@@ -223,20 +325,108 @@ def _init_weights(self, module):
...
+        elif isinstance(module, SelectionBlock):
+            d_model = self.config.d_model
+            key_value_proj_dim = self.config.d_kv
+            n_heads = self.config.num_heads
+            module.q.weight.data.normal_(mean=0.0, std=factor * ((d_model * key_value_proj_dim) ** -0.5))
+            module.k.weight.data.normal_(mean=0.0, std=factor * (d_model**-0.5))
+            module.v.weight.data.normal_(mean=0.0, std=factor * (d_model**-0.5))
+            module.o.weight.data.normal_(mean=0.0, std=factor * ((n_heads * key_value_proj_dim) ** -0.5))
+
+    def _past_feat_patch_and_embed(self, tensor: torch.Tensor) -> torch.Tensor:
```
shchur, line 337: I haven't checked this part of the code since I don't have a strong intuition about it yet. I will take a closer look once we have some reasonable benchmark numbers

timeseries/src/autogluon/timeseries/models/chronos/pipeline/utils.py
```diff
@@ -241,38 +376,99 @@ def cache_model_from_s3(s3_uri: str, force=False):
     return str(bucket_cache_path / prefix)
 
 
-class ChronosInferenceDataset:
+class ChronosInferenceDataset(IterableDataset):
```
shchur, line 457: Do you think it's feasible to have just one `ChronosDataset` class? Currently it feels like there is a lot of duplication between the two classes. We could set `inference=True/False` in the constructor, which will only affect instance splitter - choosing between returning the last `context_length` observations from each series (+ future known covariates) if `inference=True`, and doing random slicing during training when `inference=False`.

shchur:
Thanks a lot! Left some initial thoughts.

PR #5177, opened 2025-06-27
[timeseries] Improve heuristic for long-horizon forecasts with Chronos-Bolt

abdulfatir:
*Issue #, if available:* N/A

*Description of changes:* This PR improves the heuristic for forecasting beyond the model's default prediction length. Previously, we only used the median which led to uncertainty collapse and Christmas-tree-like prediction intervals. This PR improves the heuristic to take the entire distribution into account.

I tested this functionality using the following snippet.

```py
import pandas as pd
import numpy as np

df = pd.read_csv(
    "https://raw.githubusercontent.com/AileenNielsen/TimeSeriesAnalysisWithPython/master/data/AirPassengers.csv"
)
df["id"] = "0"
data = TimeSeriesDataFrame.from_data_frame(df, id_column="id", timestamp_column="Month")

predictor = TimeSeriesPredictor(prediction_length=108, target="#Passengers", verbosity=2).fit(
    data,
    hyperparameters={
        "Chronos": [{"model_path": "bolt_base"}],
    },
    enable_ensemble=False,
    skip_model_selection=True,
    time_limit=60,
)
forecast = predictor.predict(data)
low, median, high = forecast["0.1"].to_numpy(), forecast["0.5"].to_numpy(), forecast["0.9"].to_numpy()

plt.plot(median)
plt.fill_between(np.arange(len(median)), low, high, alpha=0.3)
```

### Before

image

### After

image


timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -628,14 +634,26 @@ def _predict(
         with warning_filter(all_warnings=True):
             import torch
 
+            self.model_pipeline.model.eval()
+            batch_size = self.batch_size
+            if (
+                isinstance(self.model_pipeline, ChronosBoltPipeline)
+                and self.prediction_length > self.model_pipeline.model_prediction_length
+            ):
+                batch_size = max(1, batch_size // 8)
```
abdulfatir: What do you think about the batch_size strategy? Is it too conservative?

shchur: That sounds reasonable to me. If you feel like it, you could check how high the memory consumption ends up being on a g5.xlarge instance (smallest non-deprecated instance that also supports bfloat16) w/ and w/o this fix. If we are always < 50% utilization, we could even increase this.

shchur (approved):
LGTM! Let me run another benchmark before we merge this PR, so that we can keep track of its effect on the performance.

timeseries/src/autogluon/timeseries/models/chronos/model.py
abdulfatir: I checked on a g5.12x (same GPU). `chronos-bolt-base` reaches < 40% (of 24GB) memory usage with a batch size of 64 and long horizon forecast (i.e., effective batch size of 64 * 9). This tells us that maybe we can change the factor to `// 4` at least?

abdulfatir: Maybe even `// 2`? With `128 * 9` effective batch size, the GPU memory usage is at less than 70%.

shchur: What if we add a try-except block that progressively reduces the batch size if the call to `pipeline.predict_quantiles` produced an OutOfMemoryError? Then we can definitely start with 256//2. 

Alternatively, we could again add a try-except block and raise an informative error message like "try reducing batch size by setting `predictor.fit(..., hyperparameters={"Chronos": {"batch_size": 128, ...}})`"

abdulfatir: I am not a fan of the iterative approach because I have seen sometimes the GPU goes into a weird state where the memory is not deallocated and it keeps running OOM. I have set 256//2 and then added a try catch with informative error.

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -628,23 +634,42 @@ def _predict(
...
+                            prediction_length=self.prediction_length,
+                            quantile_levels=self.quantile_levels,
+                            num_samples=self.num_samples,
+                        )
+                    except torch.OutOfMemoryError as ex:
+                        logger.error(
+                            "The call to predict() resulted in an out of memory error. Try reducing the batch_size by setting:"
+                            f" predictor.fit(..., hyperparameters={{'Chronos': {{'batch_size': {batch_size // 2}, ...}}}})"
+                        )
+                        raise ex
```
abdulfatir, line 672: @shchur Is logger.error and then re-raise good or you would do something else?

shchur: Perfect!

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -640,7 +640,7 @@ def _predict(
                 isinstance(self.model_pipeline, ChronosBoltPipeline)
                 and self.prediction_length > self.model_pipeline.model_prediction_length
             ):
-                batch_size = max(1, batch_size // 8)
+                batch_size = max(1, batch_size // 2)
```
shchur: On a second thought, let's maybe reduce this to `// 4`? I feel like not that many people work with super large datasets where the inference speed difference is that noticeable, but randomly getting an OOM error after upgrading to v1.4 feels like a pretty bad user experience (e.g., if someone has a consumer-grade GPU).

abdulfatir: Done.

PR #5182, opened 2025-07-03
[common] Replace `pkg_resources.parse_version` with `packaging.version.parse`

shchur:
*Issue #, if available:*

*Description of changes:*
- Currently, code that uses `pkg_resources.parse_version` produces warnings such as the following:
    ```
    /home/shchuro/workspace/autogluon/common/src/autogluon/common/utils/try_import.py:69: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
    ```
    This PR switches to the non-deprecated `packaging.version.parse` alternative.

### DIFF
--- a/common/src/autogluon/common/utils/try_import.py
+++ b/common/src/autogluon/common/utils/try_import.py
@@ -66,11 +66,11 @@ def try_import_ray() -> ModuleType:
 def try_import_catboost():
     try:
         import catboost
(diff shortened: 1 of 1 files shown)


Innixma (approved):
Looks great, thanks for fixing, I was also seeing this warning!

PR #5184, opened 2025-07-03
[timeseries] Update hyperparameter handling in MLForecast models

shchur:
*Issue #, if available:*

*Description of changes:*
- Update version range for `mlforecast`, `coreforecast`, `utilsforecast`
- Log a warning if unsupported hyperparameters are provided to `DirectTabular`, `RecursiveTabular`, or `Chronos` models.
- Instead of creating a `TabularPredictor` inside `DirectTabular` and `RecursiveTabular` models, create a single tabular model (instance of `autogluon.core.models.AbstractModel`). This simplifies our internal logic and makes the API of the model compatible with the upcoming PerStepTabular model.
- Allow passing `lag_transforms` to `RecursiveTabular` model.

*To do:*
- [ ] Run benchmark to check for regressions

### DIFF
--- a/timeseries/setup.py
+++ b/timeseries/setup.py
@@ -33,7 +33,7 @@ install_requires = [
     "gluonts>=0.15.0,<0.17",
     "networkx",  # version range defined in `core/_setup_utils.py`
     "statsforecast>=1.7.0,<2.0.2",
-    "mlforecast==1.0.2",
+    "mlforecast>=1.0.0,<1.0.3",
     "utilsforecast>=0.2.3,<0.2.12",  # to prevent breaking changes that propagate through mlforecast's dependency
     "coreforecast>=0.0.12,<0.0.17",  # to prevent breaking changes that propagate through mlforecast's dependency
     "fugue>=0.9.0",  # prevent dependency clash with omegaconf
--- a/timeseries/setup.py
+++ b/timeseries/setup.py
@@ -33,7 +33,7 @@ install_requires = [
     "gluonts>=0.15.0,<0.17",
     "networkx",  # version range defined in `core/_setup_utils.py`
     "statsforecast>=1.7.0,<2.0.2",
-    "mlforecast>=1.0.0,<1.0.3",
+    "mlforecast>=0.14.0,<0.15.0",  # cannot upgrade since v0.15.0 introduced a breaking change to DirectTabular
     "utilsforecast>=0.2.3,<0.2.12",  # to prevent breaking changes that propagate through mlforecast's dependency
     "coreforecast>=0.0.12,<0.0.17",  # to prevent breaking changes that propagate through mlforecast's dependency
     "fugue>=0.9.0",  # prevent dependency clash with omegaconf
--- a/timeseries/setup.py
+++ b/timeseries/setup.py
@@ -33,9 +33,9 @@ install_requires = [
     "gluonts>=0.15.0,<0.17",
     "networkx",  # version range defined in `core/_setup_utils.py`
     "statsforecast>=1.7.0,<2.0.2",
-    "mlforecast>0.13,<0.14",
-    "utilsforecast>=0.2.3,<0.2.11",  # to prevent breaking changes that propagate through mlforecast's dependency
-    "coreforecast>=0.0.12,<0.0.16",  # to prevent breaking changes that propagate through mlforecast's dependency
+    "mlforecast==1.0.2",
+    "utilsforecast>=0.2.3,<0.2.12",  # to prevent breaking changes that propagate through mlforecast's dependency
+    "coreforecast>=0.0.12,<0.0.17",  # to prevent breaking changes that propagate through mlforecast's dependency
     "fugue>=0.9.0",  # prevent dependency clash with omegaconf
     "tqdm",  # version range defined in `core/_setup_utils.py`
     "orjson~=3.9",  # use faster JSON implementation in GluonTS
--- a/timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
+++ b/timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
@@ -29,7 +29,7 @@ from .utils import MLF_ITEMID, MLF_TARGET, MLF_TIMESTAMP
 logger = logging.getLogger(__name__)
 
 
-class TabularEstimator(BaseEstimator):
+class TabularModel(BaseEstimator):
     """A scikit-learn compatible wrapper for arbitrary autogluon.tabular models"""
 
     def __init__(self, model_class: Type[AbstractTabularModel], model_kwargs: Optional[dict] = None):
@@ -46,7 +46,6 @@ class TabularEstimator(BaseEstimator):
 
     def get_params(self, deep=True):
         params = {"model_class": self.model_class, "model_kwargs": self.model_kwargs}
-
         if deep:
             return copy.deepcopy(params)
         else:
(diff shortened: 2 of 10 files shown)


timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
```diff
@@ -143,17 +125,42 @@ def preprocess(
                 data[self.target] = data[self.target].fillna(value=self._train_target_median)
         return data, known_covariates
 
-    def _get_extra_tabular_init_kwargs(self) -> Dict[str, Any]:
-        raise NotImplementedError
+    def _process_deprecated_hyperparameters(self, model_params: Dict[str, Any]) -> Dict[str, Any]:
```
shchur, line 128: Previously `DirectTabular` and `RecursiveTabular` created a `TabularPredictor` under the hood. The users could specify `tabular_hyperparameters` and `tabular_fit_kwargs` that were passed to the predictor as
```python
TabularPredictor(...).fit(..., hyperparameters=tabular_hyperparameters, **tabular_fit_kwargs)
```
This potentially allowed the users to fit multiple models inside the predictor, but I doubt that anyone really used this functionality. Using the entire predictor is really clunky (e.g., it saves a lot of files to disk, including training data, fits a separate preprocessing pipeline, 100+ lines are logged to the debug log stream).

With the updated API, the time series models create a single tabular model under the hood using `model_name` and `model_hyperparameters`. The proposed deprecation path is:
- `tabular_fit_kwargs` is always ignored
- If `tabular_hyperparameters` contains a single model, it's automatically converted to `model_name` + `model_hyperparameters`
- If `tabular_hyperparameters` more than one model, an exception is raised

We could, in theory, introduce a 100% backward-compatible option for `tabular_hyperparameters` with >= 2 models, but that will quite complicated and may introduce some bugs. Therefore, I am leaning to this slightly breaking change that, hopefully, won't affect many users.

timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
```diff
@@ -484,15 +493,15 @@ class DirectTabularModel(AbstractMLForecastModel):
...
-        Hyperparameters dictionary passed to ``TabularPredictor.fit``. Contains the names of models that should be fit.
-        Defaults to ``{"GBM": {}}``.
-    tabular_fit_kwargs : Dict[str, Any], optional
-        Additional keyword arguments passed to ``TabularPredictor.fit``. Defaults to an empty dict.
+    model_name : str, default = "GBM"
+        Name of the tabular regression model. See `autogluon.tabular.registry.ag_model_registry` or
+        `the documentation <https://auto.gluon.ai/stable/api/autogluon.tabular.models.html>`_ for the list of available
+        tabular models.
+    model_hyperparameters : Dict[str, Any], optional
+        Hyperparameters passed to the tabular regression model.
```
shchur, line 505: Parameter names `model_name` and `model_hyperparameters` were chosen for consistency with the covariate regressor
https://github.com/autogluon/autogluon/blob/2c1bb33448a180c49247d4a574b0a40e81af1d2e/timeseries/src/autogluon/timeseries/regressor.py#L43-L47

timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
```diff
@@ -29,34 +29,27 @@
...
 
-    def predict(self, X: pd.DataFrame) -> np.ndarray:
-        assert isinstance(X, pd.DataFrame)
-        return self.predictor.predict(X).values  # type: ignore
+    def predict(self, *args, **kwargs):
+        return self.model.predict(*args, **kwargs)
+
+    def get_params(self, deep=True):
+        params = {"model_class": self.model_class, "model_kwargs": self.model_kwargs}
+        if deep:
```
abdulfatir, line 49: Why not always deepcopy?

timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
abdulfatir: Users can still use multiple tabular models through the `TimeSeriesPredictor`, right? Would that approach be somehow different from what was allowed before?

timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
```diff
@@ -337,35 +346,35 @@ def _fit(
...
+            y=train_df[MLF_TARGET],
+            X_val=val_df.drop(columns=[MLF_TARGET, MLF_ITEMID]),
+            y_val=val_df[MLF_TARGET],
+            time_limit=(None if time_limit is None else time_limit - (time.time() - fit_start_time)),
         )
-        self._mlf.models = {"mean": estimator}  # type: ignore
 
-        with warning_filter():
-            self._mlf.fit_models(X=train_df.drop(columns=[MLF_TARGET, MLF_ITEMID]), y=train_df[MLF_TARGET])  # type: ignore
+        self._mlf.models_ = {"mean": tabular_model}
```
abdulfatir, line 361: Is it fine to fit the models first and then assign `models_` attribute over calling the `fit_models` method? Also, `.models` vs `.models_`: is this a recent change?

timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
```diff
@@ -720,8 +740,13 @@ def _predict(
...
-            "problem_type": ag.constants.REGRESSION,
-            "eval_metric": self.eval_metric.equivalent_tabular_regression_metric or "mean_absolute_error",
-        }
+    def _create_tabular_model(self, model_name: str, model_hyperparameters: Dict[str, Any]) -> TabularModel:
+        model_class = ag_model_registry.key_to_cls(model_name)
+        return TabularModel(
+            model_class=model_class,
+            model_kwargs={
+                "hyperparameters": model_hyperparameters,
+                "problem_type": ag.constants.REGRESSION,
```
abdulfatir, line 750: Quantile Regression is not supported here?

timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
shchur: The method and its signature, including the `deep` kwarg, are required by [sklearn.estimator.BaseEstimator](https://scikit-learn.org/stable/modules/generated/sklearn.base.BaseEstimator.html#sklearn.base.BaseEstimator.get_params)

timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
shchur: - `MLForecast.models` - list / dict of model "templates" (e.g., `LGBRegressor()`) that get cloned when actually fitting the model
- `MLForecast.models_` - list / dict of actually fitted model objects (e.g., output of `LGBRegressor().fit(X, y)`)

Calling fit_models is essentially equivalent to assigning models_[model_name] = model.fit(...)
https://github.com/Nixtla/mlforecast/blob/f674b450b19c6cde3e8d3498544d60f8740917aa/mlforecast/forecast.py#L308
The only difference is that we want to call fit(X, y, X_val, y_val) instead of the default fit(X, y) implemented inside fit_models. I will add a comment explaining this aspect.

timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
shchur: Quantile regression does not really work for recursive forecasting because we train a model to make one-step-ahead predictions. If we unroll these predictions, we get weird prediction intervals. 

Our current solution is to fit a zero-mean normal distribution to the val residuals of the model (per item), and use them to create prediction intervals. This still feels really hacky, but I don't yet have a better idea.

timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
shchur: Yes, fitting multiple different tabular models still works, e.g.
```python
TimeSeriesPredictor(...).fit(..., hyperparameters={"RecursiveTabular": [{"model_name": "GBM"}, {"model_name": "CAT"}])
```

timeseries/setup.py
```diff
@@ -33,9 +33,9 @@
     "gluonts>=0.15.0,<0.17",
     "networkx",  # version range defined in `core/_setup_utils.py`
     "statsforecast>=1.7.0,<2.0.2",
-    "mlforecast>0.13,<0.14",
-    "utilsforecast>=0.2.3,<0.2.11",  # to prevent breaking changes that propagate through mlforecast's dependency
-    "coreforecast>=0.0.12,<0.0.16",  # to prevent breaking changes that propagate through mlforecast's dependency
+    "mlforecast>=0.14.0,<0.15.0",  # cannot upgrade since v0.15.0 introduced a breaking change to DirectTabular
```
shchur, line 36: @abdulfatir, FYI, v0.15.0 introduced a breaking change that broke one of our models, so capping the version for now. It seems that there were no important fixes or improvements since 0.14.0 that are relevant for us.

PR #5189, opened 2025-07-07
[timeseries] Add a tabular forecasting model that fits a separate tabular model per horizon step

shchur:
*Issue #, if available:*

*Description of changes:*
- Add `PerStepTabularModel` that fits a separate `autogluon.tabular.AbstractModel` for each time step in the forecast horizon. 
  - This class implements a lot of custom logic, so it does not inherit from `AbstractMLForecastModel` (currently used by `DirectTabular` and `RecursiveTabular` models).
  - We do not use the built-in direct forecasting capabilities of MLForecast because 1/ MLForecast incorrectly handles known_covariates, 2/ it requires saving trained models as instance attributes, which can result in slow sequential inference.
- Minor fixes
  - Fix typos in the docstrings of `mlforecast` models
  - Correctly set the default number of parallel jobs (`AG_DEFAULT_N_JOBS`) to the # of physical CPU cores. Previously we used a heuristic of `multiprocessing.cpu_count() * 0.5` that works for most AWS instances, but sometimes returns an incorrect result.

To do:
- [x] Benchmark
- [x] Add unit tests for the model

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/autogluon_tabular/per_step.py
+++ b/timeseries/src/autogluon/timeseries/models/autogluon_tabular/per_step.py
@@ -50,7 +50,7 @@ class PerStepTabularModel(AbstractTimeSeriesModel):
     seasonal_lags: List[int], default = None
         Seasonal lags of the target used as features. Unlike trailing lags, seasonal lags are not shifted
         but filtered by availability: model for step `h` uses `[lag for lag in seasonal_lags if lag > h]`.
-        If None, determined automatically based on data frequency (e.g., [24, 48, 168] for hourly data).
+        If None, determined automatically based on data frequency.
     date_features : List[Union[str, Callable]], default = None
         Features computed from the dates. Can be pandas date attributes or functions that will take the dates as input.
         If None, will be determined automatically based on the frequency of the data.


timeseries/src/autogluon/timeseries/models/autogluon_tabular/per_step.py
```diff
@@ -0,0 +1,431 @@
...
+        date_features: list[Callable],
+        time_limit: Optional[float],
+        num_cpus: int,
+        verbosity: int,
+    ) -> str:
+        from mlforecast import MLForecast
+
+        start_time = time.monotonic()
+
+        mlf = MLForecast(models=[], freq="D", lags=lags, date_features=date_features)
```
abdulfatir: The freq is dummy here?

timeseries/src/autogluon/timeseries/models/autogluon_tabular/per_step.py
```diff
@@ -0,0 +1,431 @@
...
+
+        features_df = mlf.preprocess(train_df, static_features=[], dropna=False)
+        del train_df
+        del mlf
+        # Sort chronologically for efficient train/test split
+        features_df.sort_values(by=MLF_TIMESTAMP)
```
abdulfatir: `inplace=True`?

timeseries/src/autogluon/timeseries/models/autogluon_tabular/per_step.py
```diff
@@ -0,0 +1,431 @@
...
+        mem_usage_per_column = get_approximate_df_mem_usage(train_df)
+        num_columns = len(train_df.columns)
+        mem_usage_per_job = mem_usage_per_column.sum()
+        try:
+            mem_usage_per_job += model_cls.estimate_memory_usage_static(
+                X=train_df, hyperparameters=model_hyperparameters, problem_type="regression"
+            )
+        except NotImplementedError:
+            mem_usage_per_job *= 2
+        mem_usage_per_job *= overhead_factor + (num_extra_dynamic_features + num_columns) / num_columns
```
abdulfatir, line 207: Why divide by `num_columns` in the end?

timeseries/src/autogluon/timeseries/models/autogluon_tabular/per_step.py
```diff
@@ -0,0 +1,431 @@
...
+            item_ids = data.index.get_level_values(level=ITEMID)
+            scale_per_column: dict[str, pd.Series] = {}
+            columns_grouped = data[self._non_boolean_real_covariates].abs().groupby(item_ids)
+            for col in self._non_boolean_real_covariates:
+                scale_per_column[col] = columns_grouped[col].mean()
+            data = data.assign(**{f"__scaled_{col}": data[col] / scale for col, scale in scale_per_column.items()})
+            if known_covariates is not None:
+                known_covariates = known_covariates.assign(
+                    **{f"__scaled_{col}": known_covariates[col] / scale for col, scale in scale_per_column.items()}
+                )
```
abdulfatir, line 235: Can this thing not be handled by the covariate scaler?

timeseries/src/autogluon/timeseries/models/autogluon_tabular/per_step.py
```diff
@@ -0,0 +1,431 @@
...
+            max_num_items=model_params["max_num_items"],
+            max_num_samples=model_params["max_num_samples"],
+        )
+
+        # Initialize MLForecast arguments
+        assert self.freq is not None
+        trailing_lags = model_params.get("trailing_lags")
+        if trailing_lags is None:
+            trailing_lags = list(range(1, 13))
+        self.trailing_lags = [int(lag) for lag in trailing_lags]
```
abdulfatir: Why do we need the cast?

timeseries/src/autogluon/timeseries/models/local/abstract_local_model.py
```diff
@@ -19,7 +19,7 @@
 
 
 # We use the same default n_jobs across AG-TS to ensure that Joblib reuses the process pool
-AG_DEFAULT_N_JOBS = max(int(cpu_count() * 0.5), 1)
+AG_DEFAULT_N_JOBS = max(cpu_count(only_physical_cores=True), 1)
```
abdulfatir, line 22: Can using all physical cores make the system unresponsive?

abdulfatir (commented):
Thanks @shchur. Left some questions.

timeseries/src/autogluon/timeseries/models/autogluon_tabular/per_step.py
shchur: Yes

timeseries/src/autogluon/timeseries/models/autogluon_tabular/per_step.py
shchur: Thanks!

timeseries/src/autogluon/timeseries/models/autogluon_tabular/per_step.py
shchur: We need to ensure that the lags are of type `list[int]` and not e.g. `np.ndarray`, which would break MLForecast's input validation. Added a comment

timeseries/src/autogluon/timeseries/models/autogluon_tabular/per_step.py
shchur: This one adds new columns and the covariate scaler only modifies the columns inplace

timeseries/src/autogluon/timeseries/models/autogluon_tabular/per_step.py
shchur: `(num_extra_dynamic_features + num_columns) / num_columns` is a scaling factor that tells us how much bigger the DF will be after lag & date features are added

timeseries/src/autogluon/timeseries/models/local/abstract_local_model.py
shchur: The new default is exactly equivalent to the old default on AWS machines where 2 vCPU cores = 1 physical CPU core. So far we have not received any complaints about using too much CPU resources for the TS module. The new method `joblib.cpu_count()` also takes into account the constraints on the # of CPUs (e.g., imposed by SLURM), so it should be even more robust than the old logic.

timeseries/tests/unittests/models/test_per_step_tabular.py
```diff
@@ -0,0 +1,99 @@
...
+from ..common import DUMMY_TS_DATAFRAME
+
+DUMMY_HYPERPARAMETERS = {"model_name": "DUMMY", "n_jobs": 1}
+
+
+@pytest.mark.parametrize(
+    "seasonal_lags, trailing_lags, expected_lags_per_step",
+    [
+        ([], [1, 2, 3, 4], [[1, 2, 3, 4], [2, 3, 4, 5], [3, 4, 5, 6]]),
+        ([1, 2, 3, 4], [], [[1, 2, 3, 4], [2, 3, 4], [3, 4]]),
```
abdulfatir, line 17: Sorry, can you explain again what's happening with the seasonal lags in this case? :D

timeseries/tests/unittests/models/test_per_step_tabular.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+@pytest.mark.parametrize(
+    "n_jobs, time_limit, expected_time_limit_per_model",
+    [
+        (1, 3.0, 1.0),
+        (3, 10.0, 10.0),
+    ],
+)
+def test_when_models_are_fitten_then_time_limit_is_distributed_evenly(
```
abdulfatir:
```suggestion
def test_when_models_are_fitted_then_time_limit_is_distributed_evenly(
```

abdulfatir (approved):
Looks good to me. Thanks @shchur. Are there any more tests that you're planning to add?

timeseries/tests/unittests/models/test_per_step_tabular.py
shchur: Updated the docstring this behavior:

    seasonal_lags: List[int], default = None
        Seasonal lags of the target used as features. Unlike trailing lags, seasonal lags are not shifted
        but filtered by availability: model for step `h` uses `[lag for lag in seasonal_lags if lag > h]`.

PR #5193, opened 2025-07-08
[tabular] Add TabICL Model

Innixma:
*Issue #, if available:*

*Description of changes:*

- Add TabICL Model to AutoGluon

Paper: [TabICL: A Tabular Foundation Model for In-Context Learning on Large Data](https://arxiv.org/abs/2502.05564)
Authors of TabICL: [Jingang Qu](https://arxiv.org/search/cs?searchtype=author&query=Qu,+J), [David Holzmüller](https://arxiv.org/search/cs?searchtype=author&query=Holzm%C3%BCller,+D), [Gaël Varoquaux](https://arxiv.org/search/cs?searchtype=author&query=Varoquaux,+G), [Marine Le Morvan](https://arxiv.org/search/cs?searchtype=author&query=Morvan,+M+L)
Authors of TabICLModel code in TabArena: @LennartPurucker, @Innixma


tabular/src/autogluon/tabular/models/tabicl/tabicl_model.py
```diff
@@ -0,0 +1,174 @@
...
+        return model_cls
+
+    @staticmethod
+    def _get_batch_size(n_cells: int):
+        if n_cells <= 4_000_000:
+            return 8
+        elif n_cells <= 6_000_000:
+            return 4
+        else:
+            return 2
```
Innixma, line 46: Is there a reason this doesn't go above batch size 8 and below batch size 2?

tabular/src/autogluon/tabular/models/tabicl/tabicl_model.py
```diff
@@ -0,0 +1,174 @@
...
+            "random_state": 42,
+        }
+        for param, val in default_params.items():
+            self._set_default_param_value(param, val)
+
+    def _get_default_auxiliary_params(self) -> dict:
+        default_auxiliary_params = super()._get_default_auxiliary_params()
+        default_auxiliary_params.update(
+            {
+                "max_rows": 100000,
```
Innixma, line 101: Do we really want to be fitting TabICL on 100k sample datasets? Is it still cost efficient and performant? What should the cutoff be?

tabular/src/autogluon/tabular/models/tabicl/tabicl_model.py
dholzmueller: We put 8 as default in TabICL (I think to avoid having too many OOMs), and below 2 (or maybe even 4) was not necessary in TabArena. But also, with the default norm_methods=["none", "power"], none and power will be executed sequentially because their parallel execution is not easily compatible with another optimization that is implemented. So batch_size=2 will in this case not use more RAM than batch_size=1.

tabular/src/autogluon/tabular/models/tabicl/tabicl_model.py
dholzmueller: based on the plot in the TabICL repo (which uses worse baselines than the ones in TabArena) performance-wise it is not great after 30k-50k. Maybe it could still give good results on some datasets after that, I don't know. Runtime-wise it will probably not be great at these sizes (might still be acceptable at least in terms of training time, probably in the 100s of seconds on GPU and maybe even a good CPU, but that is also the inference time).

tabular/src/autogluon/tabular/models/tabicl/tabicl_model.py
```diff
@@ -0,0 +1,174 @@
...
+            logger.log(
+                40,
+                f"\tFailed to import tabicl! To use the TabICL model, "
+                f"do: `pip install autogluon.tabular[tabicl]=={__version__}`.",
+            )
+            raise err
+
+        from torch.cuda import is_available
+
+        device = "cuda" if num_gpus != 0 else "cpu"
```
tonyhoo, line 68: Will it be able to support other accelerator such as Apple Silicon given we have its native support in AutoMM now? Is it easy to test?

tabular/src/autogluon/tabular/models/tabicl/tabicl_model.py
```diff
@@ -0,0 +1,174 @@
...
+            }
+        )
+        return default_auxiliary_params
+
+    @classmethod
+    def supported_problem_types(cls) -> list[str] | None:
+        return ["binary", "multiclass"]
+
+    def _get_default_resources(self) -> tuple[int, int]:
+        num_cpus = ResourceManager.get_cpu_count_psutil(logical=False)
```
tonyhoo: Any specific reason we have to use psutil here? With this [PR](https://github.com/autogluon/autogluon/pull/5197/), we shall have more robust impl with `ResourceManager.get_cpu_count()`

Innixma: updated to loky

tabular/src/autogluon/tabular/models/tabicl/tabicl_model.py
Innixma: We can test if we find time after merging the initial PRs. I haven't personally tried.

Issue #5180, opened 2025-07-01
CPU Oversubscription / Ray Issues Megathread


Innixma:
Megathread tracking all issues and PRs related to CPU oversubscription issues, particularly in environments such as SLURM and docker where AutoGluon uses more resources than it should, leading to stalled / extremely slow runtimes or memory errors.

This problem has plagued AutoGluon for years, but is tricky to resolve as it only occurs on non-standard environments, often in clusters or shared development environments. 

## Problem Description

The core of the issue seems to be that `psutil.cpu_count()` returns the overall system resources, regardless of the scope of the process it is called with.

For example, imagine a system has 128 CPUs total and is a shared development environment with 4 different users, each with 32 CPUs allocated. If user A runs AutoGluon without specifying `num_cpus`, AutoGluon will detect available resources by calling `psutil.cpu_count()`, which returns 128. AutoGluon will then proceed to use all 128 CPUs on the machine to fit models. If, at the same time, user B is running a separate process than uses 32 CPUs, this leads to CPU oversubscription on the machine, as AutoGluon is trying to use more CPUs than are freely available. As demonstrated in https://github.com/autogluon/autogluon/issues/5160#issuecomment-2993855451, a CPU oversubscription of as little as 12 CPUs can cause a 400x slowdown of AutoGluon (15 seconds -> 150 minutes).

## Ideal Solution

When AutoGluon identifies available resources, it should return only the available resources the process should have available. In the above example, this should be 32 CPUs, not 128.

## Related Issues / PRs

- [ ] #4571 
- [ ] #5160 
- [ ] #5106 
- [ ] #4952
- [ ] #4895
- [ ] #4567

### Semi-Related

- [ ] #5088
- [ ] #4900
- [ ] #4497
- [ ] #4496

### Get available CPUs with `loky`

From https://github.com/autogluon/autogluon/issues/5160#issuecomment-3022808734

### Get available CPUs with `len(os.sched_getaffinity(0))`

From #4571

YuanfengZhang:
Yeah, I was just about to open an issue because I found that the resource limits do not work well, especially in HPC clusters. Autogluon will always detect and return all the available CPUs and memory of one computing node, ignoring the limits I specified in .fit params like below:
```python
    predictor = (TabularPredictor(label='actual_beta', eval_metric='root_mean_squared_error',
                                  path=output_path / fname, log_to_file=True,
                                  log_file_path=output_path / fname / f'{fname}.log')
[...]

PR #5197, opened 2025-07-09
[WIP] Add CPU utility functions for better CPU detection in restrained env such as docker and slurm cluster

tonyhoo:
*Issue #, if available:*
#5180 

*Description of changes:*
Add CPU utility functions for better CPU detection in restrained env such as docker and slurm cluster

- Introduced a new module  to provide functions for accurate CPU core detection in containerized environments and SLURM clusters.
- Implemented  to retrieve CPU count based on cgroup limits.
- Added  to determine the number of available CPU cores considering various factors like environment variables and CPU affinity.
- Updated  to utilize the new  function for improved resource management.
- Updated dependencies in  to include  for CPU detection support.

### DIFF
--- a/common/src/autogluon/common/utils/cpu_utils.py
+++ b/common/src/autogluon/common/utils/cpu_utils.py
@@ -1,0 +1,148 @@
+"""CPU utilities for accurate resource detection in constrained environments.
+
+This module provides functions for determining the correct number of available CPU cores
+in containerized environments, SLURM clusters, and other resource-constrained systems.
+"""
+
+import os
+import math
+import multiprocessing
+
+import logging
+import loky
+import psutil
+
+logger = logging.getLogger(__name__)
+
+def get_cpu_count_cgroup(os_cpu_count):
+    """Get CPU count from cgroup limits (commonly used in Docker)
+    
+    Parameters
+    ----------
+    os_cpu_count : int
+        Default CPU count to return if no cgroup limits are found
+        
+    Returns
+    -------
+    int
+        CPU count based on cgroup limits or the provided os_cpu_count if no limits found
+    """
+    cpu_max_fname = "/sys/fs/cgroup/cpu.max"
+    cfs_quota_fname = "/sys/fs/cgroup/cpu/cpu.cfs_quota_us"
+    cfs_period_fname = "/sys/fs/cgroup/cpu/cpu.cfs_period_us"
+
+    if os.path.exists(cpu_max_fname):
+        # cgroup v2
+        try:
+            with open(cpu_max_fname) as fh:
+                cpu_quota_us, cpu_period_us = fh.read().strip().split()
+                if cpu_quota_us == "max":
+                    return os_cpu_count
+                else:
+                    cpu_quota_us = int(cpu_quota_us)
+                    cpu_period_us = int(cpu_period_us)
+                    if cpu_quota_us > 0 and cpu_period_us > 0:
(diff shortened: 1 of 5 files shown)


tonyhoo:
/platform_tests c579eac3851d534cbf666212681a01b4a32fc102
[Platform Tests Output][1]

[1]: https://github.com/autogluon/autogluon/actions/runs/16160442860

common/src/autogluon/common/utils/resource_utils.py
```diff
@@ -17,7 +18,7 @@ class ResourceManager:
 
     @staticmethod
     def get_cpu_count():
-        return multiprocessing.cpu_count()
+        return get_available_cpu_count()
```
shchur: Is it possible to modify this method to optionally only detect the # of available physical cores? In TS I observed that setting the # of parallel jobs (e.g., when fitting local models like ETS) to the # of physical cores results in much faster runtime than setting n_jobs to the # of available CPU cores, which defaults to the vCPU count. Currently we are using `joblib.cpu_count(only_physical_cores=True)` for this.

tonyhoo: Good point, added a new flag for `only_physical_cores` in this method and implemented using `psutil.cpu_count(logical=False)`

common/src/autogluon/common/utils/cpu_utils.py
```diff
@@ -0,0 +1,175 @@
...
+            psutil_affinity_count = len(p.cpu_affinity())
+            logger.debug(f"CPU affinity count (psutil): {psutil_affinity_count}")
+            # Affinity works on logical cores, so only use for logical detection
+            if not only_physical_cores:
+                available_counts.append(psutil_affinity_count)
+    except Exception as e:
+        logger.debug(f"Error getting CPU affinity via psutil: {e}")
+
+    # 5. Check cgroup limits (Docker containers with --cpus=N parameter)
+    cgroup_count = get_cpu_count_cgroup(default_cpu_count)
```
shchur: It seems that [`loky.cpu_count()`](https://github.com/joblib/loky/blob/15dfcd77faadfb853841162696b486233e540b0b/loky/backend/context.py#L143) already takes into account the cgrop CPU count. What are the cases when our method `get_cpu_count_cgroup` returns a different result compared to loky?

common/src/autogluon/common/utils/cpu_utils.py
```diff
@@ -0,0 +1,175 @@
...
+                    # Environment variable overrides everything
+                    return count
+            except ValueError:
+                pass
+
+    # 2. Try loky's CPU count which handles various container scenarios
+    loky_count = loky.cpu_count()
+    logger.debug(f"loky.cpu_count(): {loky_count}")
+    # Note: loky doesn't distinguish between physical/logical cores, so we use it as-is
+    # for logical cores, but skip it for physical cores to avoid mixing different types
```
shchur: I'm a bit confused about this comment. Why cannot we call [`loky.cpu_count(only_physical_cores=only_physical_cores)`](https://github.com/joblib/loky/blob/15dfcd77faadfb853841162696b486233e540b0b/loky/backend/context.py#L78) if `only_physical_cores=True` in our method?

tonyhoo: Good call. I am not aware that loky takes physical_core as parameter as well

Innixma (approved):
LGTM, awesome change!

PR #5203, opened 2025-07-09
[tabular] Respect num_cpus/num_gpus in sequential_local fit

Innixma:
*Issue #, if available:*

Related: #5180 

*Description of changes:*

**Major bug fix**
- Previously if the model is fit in bagged mode with `fold_fitting_strategy="sequential_local"`, the system level CPU and GPU arguments from `predictor.fit(num_cpus=..., num_gpus=...)` were ignored.
- This led to major problems, such as GPUs being used when they shouldn't be, or far more CPUs being used than the user expected.
- Now the system level CPU and GPU arguments are respected.
- Additionally fixed an issue in `fold_fitting_strategy="sequential_local"` where if the user specified `ag_args_fit` with only one of `num_cpus` or `num_gpus`, the other unspecified value would be set to an incorrect value, ignoring the model's default resource suggestion.
- Added logging in the bagged model to clearly show the num_cpus and num_gpus being used to fit when in sequential local mode. (previously was hidden, which was why this bug was undetected for so long)

### DIFF
--- a/core/src/autogluon/core/models/ensemble/fold_fitting_strategy.py
+++ b/core/src/autogluon/core/models/ensemble/fold_fitting_strategy.py
@@ -297,17 +297,24 @@ class SequentialLocalFoldFittingStrategy(FoldFittingStrategy):
 
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
-        if self.user_ensemble_resources is None:
-            if self.user_resources_per_job is None:
-                self.num_cpus, self.num_gpus = self.model_base._get_default_resources()
-            else:
-                self.num_cpus = self.user_resources_per_job.get("num_cpus", self.num_cpus)
-                self.num_gpus = self.user_resources_per_job.get("num_gpus", self.num_gpus)
+        total_num_cpus = self.num_cpus
+        total_num_gpus = self.num_gpus
+
+        default_num_cpus, default_num_gpus = self.model_base._get_default_resources()
+        if self.user_resources_per_job is None:
+            fit_num_cpus, fit_num_gpus = default_num_cpus, default_num_gpus
         else:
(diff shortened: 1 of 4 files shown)


core/src/autogluon/core/models/ensemble/fold_fitting_strategy.py
```diff
@@ -297,17 +297,24 @@ class SequentialLocalFoldFittingStrategy(FoldFittingStrategy):
...
-                self.num_gpus = self.user_resources_per_job.get("num_gpus", self.num_gpus)
-        self.resources = {"num_cpus": self.num_cpus, "num_gpus": self.num_gpus}
+            fit_num_cpus = self.user_resources_per_job.get("num_cpus", default_num_cpus)
+            fit_num_gpus = self.user_resources_per_job.get("num_gpus", default_num_gpus)
+
+        # ensure that we never use more resources than the total system resources provided
+        fit_num_cpus = min(fit_num_cpus, total_num_cpus)
+        fit_num_gpus = min(fit_num_gpus, total_num_gpus)
+
+        assert fit_num_cpus >= 1
```
LennartPurucker, line 314: I think, in theory, we could give a job 0.X CPUs (with Ray). But this should be a good check anyhow.

LennartPurucker (approved):
Great fix, ty!

core/src/autogluon/core/models/ensemble/fold_fitting_strategy.py
Innixma: IMO better to raise an exception if this somehow happens so we can decide if it is intended, as I haven't coded the models with non-integer num_cpus in mind.

tonyhoo (approved):

PR #5206, opened 2025-07-10
[timeseries] Minor verbosity fixes for v1.4

shchur:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
+++ b/timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
@@ -482,7 +482,7 @@ class AbstractTimeSeriesModel(TimeSeriesModelBase, TimeSeriesTunable, ABC):
             self.covariate_regressor.fit(
                 train_data,
                 time_limit=covariate_regressor_time_limit,
-                verbosity=verbosity,
+                verbosity=verbosity - 1,
             )
 
         if self._get_tags()["can_use_train_data"]:
--- a/timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
+++ b/timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
@@ -356,6 +356,7 @@ class AbstractMLForecastModel(AbstractTimeSeriesModel):
                 X_val=val_df.drop(columns=[MLF_TARGET, MLF_ITEMID]),
                 y_val=val_df[MLF_TARGET],
                 time_limit=(None if time_limit is None else time_limit - (time.time() - fit_start_time)),
+                verbosity=verbosity - 1,
             )
 
         # We directly insert the trained model into models_ since calling _mlf.fit_models does not support X_val, y_val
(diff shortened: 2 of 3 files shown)


timeseries/src/autogluon/timeseries/models/multi_window/multi_window_model.py
```diff
@@ -129,7 +130,11 @@ def _fit(
                     num_refits_remaining = math.ceil(
                         (val_splitter.num_val_windows - window_index) / refit_every_n_windows
                     )
-                    time_left_for_window = time_left / num_refits_remaining
+                    if num_refits_remaining == 1:
+                        time_left_for_window = time_left
+                    else:
+                        # Allocate less time if it's not the last refit
+                        time_left_for_window = time_left / (num_refits_remaining + (self.last_window_time_factor - 1))
```
shchur: This one is experimental, currently benchmarking this. Other changes are just bugfixes that we definitely want to merge.

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
```diff
@@ -482,7 +482,7 @@ def fit(
             self.covariate_regressor.fit(
                 train_data,
                 time_limit=covariate_regressor_time_limit,
-                verbosity=verbosity,
+                verbosity=verbosity - 1,
```
abdulfatir, line 485: What about when verbosity is 0?

timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
```diff
@@ -531,7 +532,9 @@ def _mask_df(self, df: pd.DataFrame) -> pd.DataFrame:
         """Apply a mask that mimics the situation at prediction time when target/covariates are unknown during the
         forecast horizon.
         """
-        num_hidden = np.random.randint(0, self.prediction_length, size=len(df))
+        # Fix seed to make the model deterministic
+        rng = np.random.default_rng(seed=123)
```
abdulfatir, line 536: Is there a global seed somewhere as well? If that's the case, should we use the global seed instead?

abdulfatir (approved):
No major concerns in general but I left a couple of comments which may need some fixes.

Innixma (approved):

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
shchur: The checks usually look like
https://github.com/autogluon/autogluon/blob/76e60d3a91d5bc18f896595bfecc129f3f8714aa/tabular/src/autogluon/tabular/models/lgb/lgb_model.py#L145
so it's fine

timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
shchur: The underlying tabular models (e.g. LightGBM or CatBoost) are 100% deterministic since in tabular everything is deterministic, and there is no way to modify their random seed. For this reason I think it's fine to just make all tabular forecasting models 100% deterministic and freeze this component as well.

In principle, there is a global seed and `model` receives `seed + i`:
https://github.com/autogluon/autogluon/blob/76e60d3a91d5bc18f896595bfecc129f3f8714aa/timeseries/src/autogluon/timeseries/trainer.py#L461-L462

However, for DirectTabular model this specific random component (which entries get masked during training) may have a noticeable effect on the model performance, which sometimes leads to fluctuations in benchmarks.

Issue #5207, opened 2025-07-10
[BUG] Model distillation failed for RF, CatBoost and NN in tutorial


tonyhoo:
**Bug Report Checklist**

- [x] I provided code that demonstrates a minimal reproducible example.  
- [ ] I confirmed bug exists on the latest mainline of AutoGluon via source install.  
- [ ] I confirmed bug exists on the latest stable version of AutoGluon.  

**Describe the bug**

see [tutorial](https://auto.gluon.ai/dev/tutorials/tabular/tabular-indepth.html#model-distillation)

**Expected behavior**

**To Reproduce**

**Screenshots / Logs**

**Installed Versions**

<details>

```python
# Replace this code with the output of the following:
from autogluon.core.utils import show_versions
show_versions()
```

</details>

Innixma:
Seems to be fixed now.

PR #5208, opened 2025-07-10
[bug fix] Enhance spunge_augment and munge_augment functions for model distillation

tonyhoo:
*Issue #, if available:*
#5207 
## Issue

When using model distillation with the SPUNGE augmentation method (the default), various models would fail with categorical feature errors:

1. **LightGBM Error**: 
   ```
   ValueError: pandas dtypes must be int, float or bool.
   Fields with bad pandas dtypes: workclass: object, education: object, marital-status: object...
   ```

2. **CatBoost Error**:
   ```
   CatBoostError: features data: pandas.DataFrame column 'workclass' has dtype 'category' 
   but is not in cat_features list
   ```

3. **NeuralNetTorch Error**:
   ```
   ValueError: Found array with 0 feature(s) (shape=(4800, 0)) while a minimum of 1 is required.
   ```

## Root Cause

The issue was in the implementation of the `spunge_augment` and `munge_augment` functions in `autogluon/core/augmentation/distill_utils.py`.

During data augmentation:
1. The code added a special `"__NaN__"` category to categorical columns
2. Performed data augmentation operations
3. Removed the `"__NaN__"` category from categorical columns

The problem was in step 3 - when removing the `"__NaN__"` category, the code didn't properly restore the categorical feature properties, causing inconsistent states when used by models.

## Fix Applied

The fix makes two key changes:

1. **Track original categories**:
   ```python
   # Store original categories to restore later
   original_categories = {}
   for feature in category_featnames:
       current_categories = X[feature].cat.categories
       original_categories[feature] = current_categories
   ```

2. **Properly restore categorical features after augmentation**:
   ```python
   for feature in category_featnames:
       # Properly restore categorical features to their original state
       if nan_category not in original_categories[feature]:
           # Need to convert to string temporarily to handle NaN values correctly
           X_aug[feature] = X_aug[feature].astype(str)
           X_aug.loc[X_aug[feature] == nan_category, feature] = np.nan
           X_aug[feature] = pd.Categorical(X_aug[feature], categories=original_categories[feature])
       else:
           # Keep the NaN category but ensure we have the original categorical structure
           X_aug[feature] = pd.Categorical(X_aug[feature], categories=original_categories[feature])
   ```

The same approach was applied to both the `spunge_augment` and `munge_augment` functions to ensure consistent handling of categorical features.

## Testing
Validate the model distillation working fine in local notebook

### DIFF
--- a/core/src/autogluon/core/augmentation/distill_utils.py
+++ b/core/src/autogluon/core/augmentation/distill_utils.py
@@ -80,8 +80,11 @@ def spunge_augment(X, feature_metadata: FeatureMetadata, num_augmented_samples=1
     X = X.copy()
     nan_category = "__NaN__"
     category_featnames = feature_metadata.get_features(valid_raw_types=[R_CATEGORY])
+    # Store original categories to restore later
+    original_categories = {}
     for feature in category_featnames:
         current_categories = X[feature].cat.categories
+        original_categories[feature] = current_categories
         if nan_category in current_categories:
             X[feature] = X[feature].fillna(nan_category)
         else:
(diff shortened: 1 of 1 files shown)


FANGAreNotGnu (approved):
LGTM

Innixma (approved):
LGTM, great fix!

PR #5212, opened 2025-07-11
[timeseries] Add AutoMLPipelineFeatureGenerator to MLForecast-based models

shchur:
*Issue #, if available:*

*Description of changes:*
- This PR adds an `AutoMLPipelineFeatureGenerator` to DirectTabular and RecursiveTabular models. This fixes a regression on some datasets with categorical features that was introduced in #5184. 

This object ensures that the features received by model during fit and predict have the same dtypes and category ordering. For some reason, removing this check led to worse performance of LightGBM on two datasets.

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
+++ b/timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
@@ -345,16 +345,26 @@ class AbstractMLForecastModel(AbstractTimeSeriesModel):
             max_num_items=model_params["max_num_items"],
             max_num_samples=model_params["max_num_samples"],
         )
+        from autogluon.features import AutoMLPipelineFeatureGenerator
+
+        self.pipeline = AutoMLPipelineFeatureGenerator()
+        X = train_df.drop(columns=[MLF_TARGET, MLF_ITEMID])
+        y = train_df[MLF_TARGET]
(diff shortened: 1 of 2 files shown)


timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
```diff
@@ -364,12 +369,12 @@ def _fit(
...
         self._save_residuals_std(val_df)
 
-    def get_tabular_model(self) -> AbstractTabularModel:
+    def get_tabular_model(self) -> TabularModel:
         """Get the unerlyin tabular regression model."""
         assert "mean" in self._mlf.models_, "Call `fit` before calling `get_tabular_model`"
         mean_estimator = self._mlf.models_["mean"]
         assert isinstance(mean_estimator, TabularModel)
-        return mean_estimator.model
+        return mean_estimator
```
abdulfatir, line 377: Does this affect any other callers?

abdulfatir (approved):

timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
shchur: ~~No - there are only 2 calls to this method in `mlforecast.py` and no other calls besides that.~~
That's a good point, this also affects a unit test. Fixed that now.

PR #5214, opened 2025-07-12
[tabular] Add tabular foundational model cache from s3 to benchmark avoid rate limit issue from HF

tonyhoo:
*Issue #, if available:*

*Description of changes:*
[tabular] Add tabular foundational model cache from s3 to benchmark avoid rate limit issue from HF


Innixma (approved):
LGTM!

PR #5216, opened 2025-07-13
[tabular] Fix default loss_function for CatBoostModel with problem_type='regression'

shchur:
*Issue #, if available:*

*Description of changes:*
- Currently, the `CatBoostModel` in AutoGluon always uses the default `loss_function` [(RMSE)](https://catboost.ai/docs/en/concepts/python-reference_catboostregressor) when `problem_type="regression"` is used. This means, the model always tries to estimate the conditional mean, which can lead to poor results for datasets with skewed targets if the user actually cares about the conditional median (e.g., cares about the metric like MAE).

  This PR updates the default value of the `loss_function` used by CatBoostRegressor to match the user-defined `eval_metric`. 

On a small benchmark of 31 time series forecasting tasks, the `PerStepTabular` model (based on CatBoost) achieves 77% win rate and up to 30% error reduction (median: 4% reduction) compared to the default CatBoost version with the RMSE loss.

### DIFF
--- a/tabular/src/autogluon/tabular/models/catboost/catboost_model.py
+++ b/tabular/src/autogluon/tabular/models/catboost/catboost_model.py
@@ -134,7 +134,7 @@ class CatBoostModel(AbstractModel):
             params.setdefault("loss_function",  SoftclassObjective.SoftLogLossObjective())
             params["eval_metric"] = SoftclassCustomMetric.SoftLogLossMetric()
         elif self.problem_type in [REGRESSION, QUANTILE]:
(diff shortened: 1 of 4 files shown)


Innixma (approved):
Great change!

You’ll want to extend this logic to include any other metrics you think are needed where RMSE isn’t a good proxy: https://github.com/autogluon/autogluon/blob/master/tabular/src/autogluon/tabular/models/catboost/catboost_utils.py#L63-L70

also, can you make all the other assignment to loss_function use .setdefault? I think this is better than what I had, as it lets the user override it if necessary

PR #5218, opened 2025-07-13
Mitra

xiyuanzh:
*Issue #, if available:*

*Description of changes:* Add num_cpus, add deterministic for SDPA


xiyuanzh:
/platform_tests https://github.com/autogluon/autogluon/commit/943b886ff03422e06a5c2aee01c6ca86a0ac6656

xiyuanzh:
/platform_tests 34d71e26ec3ba8ae02819619a296c22dcbca4e3d
[Platform Tests Output][1]

[1]: https://github.com/autogluon/autogluon/actions/runs/16313303917

tabular/src/autogluon/tabular/models/mitra/mitra_model.py
```diff
@@ -1,3 +1,6 @@
+import os
+os.environ['CUBLAS_WORKSPACE_CONFIG'] = ':4096:8'
```
tonyhoo, line 6: can we add a TODO comment here to visit this and provide some context why we need to do so now?

tabular/src/autogluon/tabular/models/mitra/mitra_model.py
```diff
@@ -41,6 +44,16 @@ def _fit(
...
+        need_to_reset_torch_threads = False
+        torch_threads_og = None
+        if num_cpus is not None and isinstance(num_cpus, (int, float)):
+            torch_threads_og = torch.get_num_threads()
+            if torch_threads_og != num_cpus:
+                # reset torch threads back to original value after fit
+                torch.set_num_threads(num_cpus)
+                need_to_reset_torch_threads = True
+
         model_cls = self.get_model_cls()
```
tonyhoo, line 62: Any context why we need to reset the threads manually here? Probably worth some comments for context in future

tabular/src/autogluon/tabular/models/mitra/mitra_model.py
xiyuanzh: Thanks a lot! Added the following:

TODO: To ensure deterministic operations we need to set torch.use_deterministic_algorithms(True) and os.environ['CUBLAS_WORKSPACE_CONFIG'] = ':4096:8'. The CUBLAS environment variable configures the workspace size for certain CUBLAS operations to ensure reproducibility when using CUDA >= 10.2. Both settings are required to ensure deterministic behavior in operations such as matrix multiplications.

tabular/src/autogluon/tabular/models/mitra/mitra_model.py
xiyuanzh: Thanks! The previous code does not use num_cpus, and here I follow the scripts in tabpfnmix to reset the threads based on num_cpus: https://github.com/autogluon/autogluon/blob/master/tabular/src/autogluon/tabular/models/tabpfnmix/tabpfnmix_model.py#L193. 

I will add the comment: 
TODO: Reset the number of threads based on the specified num_cpus

tonyhoo (approved):
LGTM! Thanks

tabular/src/autogluon/tabular/models/mitra/mitra_model.py
hohoCode: Just you so know, when forcing such deterministic, it may significant lower the performance/speed. In my other experiments such slowness is quite significant. Hence better provide a setting to tell users the trade-off.

dcmaddix: Yes @xiyuanzh maybe we should leave it as a setting. For Chronos, I checked and we don't have these flags.

PR #5221, opened 2025-07-15
[tabular] Mitra Python 3.9 Support

tonyhoo:
*Issue #, if available:*

*Description of changes:*
* Python 3.9 Support
* Support device detection to work with CPU, MPS and CUDA

### DIFF
--- a/tabular/src/autogluon/tabular/models/mitra/_internal/config/config_run.py
+++ b/tabular/src/autogluon/tabular/models/mitra/_internal/config/config_run.py
@@ -1,12 +1,12 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from typing import Self
 
 import torch
 
 from ..._internal.config.config_pretrain import ConfigSaveLoadMixin
 from ..._internal.config.enums import ModelName
+
 
 @dataclass
 class ConfigRun(ConfigSaveLoadMixin):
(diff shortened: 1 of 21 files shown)


tonyhoo:
/platform_tests 578428db2030a10ea686705e4b43cc5633c47e86
[Platform Tests Output][1]

[1]: https://github.com/autogluon/autogluon/actions/runs/16289805663

tabular/src/autogluon/tabular/models/mitra/_internal/config/config_run.py
```diff
@@ -1,13 +1,13 @@
 from __future__ import annotations
 
 from dataclasses import dataclass
-from typing import Self
```
Innixma, line 4: nit: you can also do `from typing_extensions import Self`

tonyhoo: Thanks for the call! Let's keep it this way for now—there's a good chance we’ll drop Python 3.9 support in the next release since it will be [sunset](https://devguide.python.org/versions/) in Oct 2025. We can reintroduce these syntaxes by then!

tonyhoo:
/platform_tests 66d4b1f0f1f143c525e0f1f54bc8190032015928
[Platform Tests Output][1]

[1]: https://github.com/autogluon/autogluon/actions/runs/16300786842

tonyhoo:
/platform_tests a6b2abd39abe430783af7fdb82f395bae8e80c9b
[Platform Tests Output][1]

[1]: https://github.com/autogluon/autogluon/actions/runs/16301942221

PR #5226, opened 2025-07-16
[tabular] Enhance GPU resource handling across foundational tab models with error logging

tonyhoo:
*Issue #, if available:*

*Description of changes:*
This PR addresses a issue identified in the previous PR (#5221) where `torch` was being imported directly in the `_get_default_resources()` method across several foundational tabular models (RealMLP, TabPFNv2, TabM, TabICL, and Mitra).

The `_get_default_resources()` method is called very early in the AutoGluon workflow - during model initialization and resource calculation phases, before proper error handling mechanisms are established. When torch import fails at this stage, it can:
* Bypass error handling logic - The failure occurs before AutoGluon's try/catch mechanisms are set up
* Crash the entire AutoGluon process - Instead of gracefully falling back to CPU-only mode
* Prevent proper user feedback - Users don't get helpful error messages about missing dependencies

### DIFF
--- a/tabular/src/autogluon/tabular/models/mitra/mitra_model.py
+++ b/tabular/src/autogluon/tabular/models/mitra/mitra_model.py
@@ -1,12 +1,14 @@
+import logging
 import os
 from typing import List, Optional
 
 import pandas as pd
-import torch
 
 from autogluon.common.utils.resource_utils import ResourceManager
 from autogluon.core.models import AbstractModel
+from autogluon.tabular import __version__
 
+logger = logging.getLogger(__name__)
 
 # TODO: Needs memory usage estimate method
 class MitraModel(AbstractModel):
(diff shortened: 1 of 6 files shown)


tabular/src/autogluon/tabular/models/mitra/mitra_model.py
```diff
@@ -142,13 +144,23 @@ def _get_default_ag_args_ensemble(cls, **kwargs) -> dict:
...
         # Only request GPU if CUDA is available
-        if torch.cuda.is_available():
-            num_gpus = 1
-        else:
+        try:
+            import torch
+            if torch.cuda.is_available():
+                num_gpus = 1
+            else:
+                num_gpus = 0
```
Innixma: What if CUDA is available but the user doesn't have any GPUs? Is this possible? We should have a utility function to avoid code dupe.

tabular/src/autogluon/tabular/models/realmlp/realmlp_model.py
```diff
@@ -274,9 +274,20 @@ def _get_default_stopping_metric(self):
     def _get_default_resources(self) -> tuple[int, int]:
         # Use only physical cores for better performance based on benchmarks
         num_cpus = ResourceManager.get_cpu_count(only_physical_cores=True)
+        
         # Only request GPU if CUDA is available (RealMLP doesn't support MPS)
-        import torch
-        num_gpus = 1 if torch.cuda.is_available() else 0
+        try:
+            import torch
+            num_gpus = 1 if torch.cuda.is_available() else 0
```
Innixma: can we instead add a flag to the `ResourceManager.get_gpu_count_torch` to ignore the MPS logic? This seems like it would be cleaner.

tabular/src/autogluon/tabular/models/mitra/mitra_model.py
tonyhoo: Done

tabular/src/autogluon/tabular/models/realmlp/realmlp_model.py
tonyhoo: Done

PR #5232, opened 2025-07-21
Mitra Deterministic

xiyuanzh:
*Issue #, if available:*

*Description of changes:* Move fixing seed to sklearn_interface.py for deterministic operations regardless of the entry file.


tabular/src/autogluon/tabular/models/mitra/sklearn_interface.py
```diff
@@ -114,6 +115,7 @@ def __init__(self,
         self.train_time = 0
         self.seed = seed
 
+        set_seed(self.seed)
```
Innixma, line 118: What if seed is None?

xiyuanzh: Thanks for the comment! By default, we pass seed=0 during model initialization. Will the user be able to explicitly override it with seed=None when calling MitraModel? If so, do we want to interpret that as setting the seed to the default seed=0?

Innixma: Ok, we can merge this as is for now. Essentially None would be non-deterministic mode, which currently crashes if specified.

Innixma (approved):
LGTM!

tabular/src/autogluon/tabular/models/mitra/sklearn_interface.py
xiyuanzh: Thanks! Yeah we might want to add an option to enable or disable deterministic behavior.

PR #5267, opened 2025-07-30
[Tabular] Support different random seeds per fold

LennartPurucker:
This PR ensures that each fold gets a different random seed. I implemented it as part of the kwargs to avoid users setting random seeds as a hyperparameter. 

> By default, we keep a static seed of 0 to ensure reproducibility. When using a bagged model, a static seed of 0 to n_splits-1 is used to ensure that each fold has a different seed that is consistent across different models.

Note, one can disable this behavior and use a static seed across all folds (similar to the prior implementation via):

TODO:

- [x] Add code to support this for all other models. 

--- 
One can control the seed behavior via `ag_args_ensemble`:

```python
# Enable behavior of varying the seeds across folds
predictor = TabularPredictor(label="class", path=str("./ag_path")).fit(
    train_data=train_data,
    hyperparameters={'GBM':{"num_boost_round": 10}},
    ag_args_ensemble={"fold_fitting_strategy": 'sequential_local', "vary_seed_across_folds": True}
)

# Set your own random seed for all models
predictor = TabularPredictor(label="class", path=str("./ag_path")).fit(
    train_data=train_data,
    hyperparameters={'GBM':{"num_boost_round": 10}},
    ag_args_ensemble={"fold_fitting_strategy": 'sequential_local', "model_random_seed": 42}
)

# Set your own random seed per configuration
predictor = TabularPredictor(label="class", path=str("./ag_path")).fit(
    train_data=train_data,
    hyperparameters={'GBM':{"num_boost_round": 10, "ag_args_ensemble": {"model_random_seed": 42}}},
)

```

### DIFF
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -1034,6 +1034,11 @@ class AbstractModel(ModelBase, Tunable):
             verbosity 2: logs only important information.
             verbosity 1: logs only warnings and exceptions.
             verbosity 0: logs only exceptions.
+        random_seed : int, default = 0
+            The random seed used to control the randomness during fitting the model.
+            By default, we keep a static seed of 0 to ensure reproducibility.
+            When using a bagged model, a static seed of 0 to n_splits-1 is used to ensure that each fold has a
+            different seed that is consistent across different models.
         log_resources: bool, default = False
             If True, will log information about the number of CPUs, GPUs, and memory usage during fit.
         **kwargs :
@@ -1179,6 +1184,7 @@ class AbstractModel(ModelBase, Tunable):
         num_cpus: int = None,
         num_gpus: int = None,
         verbosity: int = 2,
+        random_seed: int = 0,
         **kwargs,
     ):
         """
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -214,6 +214,8 @@ class AbstractModel(ModelBase, Tunable):
     model_info_json_name = "info.json"
     learning_curve_file_name = "curves.json"
 
+    default_random_seed: int | None = 0
+
     def __init__(
         self,
         path: str | None = None,
@@ -279,7 +281,8 @@ class AbstractModel(ModelBase, Tunable):
 
         self._compiler = None
 
-        self.random_seed: int | None = None
+        # None is a valid value, "NOTSET" indicates `.init_random_seed` was not called yet.
+        self.random_seed: int | None | str = "NOTSET"
 
     @classmethod
     def _init_user_params(
@@ -699,6 +702,7 @@ class AbstractModel(ModelBase, Tunable):
 
         kwargs.pop("feature_metadata", None)
         kwargs.pop("num_classes", None)
+        kwargs.pop("random_seed", None)
         return kwargs
 
     @classmethod
@@ -728,6 +732,8 @@ class AbstractModel(ModelBase, Tunable):
         self._init_misc(X=X, y=y, feature_metadata=feature_metadata, num_classes=num_classes, **kwargs)
 
         self._init_params()
+
+        self.init_random_seed(random_seed=kwargs.get("random_seed", "auto"), hyperparameters=self.params)
 
         if X is not None:
             self._preprocess_set_features(X=X, feature_metadata=feature_metadata)
@@ -1036,9 +1042,10 @@ class AbstractModel(ModelBase, Tunable):
             verbosity 2: logs only important information.
             verbosity 1: logs only warnings and exceptions.
             verbosity 0: logs only exceptions.
-        random_seed : int, default = 0
+        random_seed : int | None | str, default = "auto"
             The random seed value provided by AutoGluon that can be used to control the randomness of the model (e.g.,
-            init, training, etc.). By default, AutoGluon's random_seed is 0 to ensure reproducibility.
+            init, training, etc.). Note, this parameter is not passed to `._fit` but used in `_initialize`!
+            If "auto", the model will use a default random seed of 0.
             When using a bagged model, this value differs per fold model. The first fold model uses `model_random_seed`,
             the second uses `model_random_seed + 1`, and the last uses `model_random_seed+n_splits-1` where `n_splits`.
             The start value `model_random_seed` can be set via `ag_args_ensemble` in the model's hyperparameters.
@@ -1187,7 +1194,6 @@ class AbstractModel(ModelBase, Tunable):
         num_cpus: int = None,
         num_gpus: int = None,
         verbosity: int = 2,
-        random_seed: int = 0,
         **kwargs,
     ):
         """
@@ -1208,29 +1214,40 @@ class AbstractModel(ModelBase, Tunable):
 
     # TODO: add model-tag to check if the model can work with `None` random seed?
     # TODO: add check that int seed is smaller than `int(np.iinfo(np.int32).max)`?
-    def init_random_seed(self, random_seed: int | None = None, hyperparameters: dict | None = None):
-        """Initialize the random seed used by the model for, e.g., training.
+    def init_random_seed(self, random_seed: int | None | str, hyperparameters: dict | None = None):
+        """Initialize the random seed used by the model by setting `self.random_seed`.
 
-        Following convention, a random seed can be either an integer or None.
+        The random seed can be used to control the randomness of the model (e.g., init, training, etc.).
+        By default, AutoGluon's random_seed is 0 to ensure reproducibility. Following convention,
+        a random seed can be either an integer or None.
+
+        When using a bagged model, this value differs per fold model. The first fold model uses `model_random_seed`,
+        the second uses `model_random_seed + 1`, and the last uses `model_random_seed+n_splits-1` where `n_splits`.
+        The start value `model_random_seed` can be set via `ag_args_ensemble` in the model's hyperparameters.
 
         Parameters
         ----------
-        random_seed
-            The random seed to set. This should be the random_seed provided to the model during `_fit`
-            by AutoGluon.
+        random_seed:
+            The random seed passed to `fit`. If "auto", the model will use a default random seed of 0.
+            Otherwise, it will set the model's random seed to the provided value.
         hyperparameters
             The hyperparameters of the model, which may or may not contain a random_seed.
             If the hyperparameters contain a random_seed, it will be used to set the model's random seed and
             thus override the random_seed provided in `random_seed`.
         """
+        # Set default random seed
+        if random_seed == "auto":
+            random_seed = self.default_random_seed
+
+        # Overwrite random seed based on hyperparameters, if available
         if hyperparameters is not None:
-            has_random_seed, hp_rs = self._get_random_seed_from_hyperparameters(hyperparameters=hyperparameters)
-            if has_random_seed:
+            hp_rs = self._get_random_seed_from_hyperparameters(hyperparameters=hyperparameters)
+            if not isinstance(hp_rs, str):
                 random_seed = hp_rs
 
         self.random_seed = random_seed
 
-    def _get_random_seed_from_hyperparameters(self, hyperparameters: dict | None = None) -> tuple[bool, int | None]:
+    def _get_random_seed_from_hyperparameters(self, hyperparameters: dict | None = None) -> int | None | str:
         """Extract the random seed from the hyperparameters if available.
 
         A model implementation may override this method to extract the random seed from the hyperparameters such that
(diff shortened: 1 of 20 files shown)


core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
```diff
@@ -60,6 +60,8 @@ class BaggedEnsembleModel(AbstractModel):
     """
 
     _oof_filename = "oof.pkl"
+    _vary_seed_across_folds: bool = True
```
Innixma: can be an `ag_args_fit` param to `BaggedEnsembleModel` with default True.

this can then be specified via `ag_args_ensemble` by the user.

tabular/src/autogluon/tabular/models/lgb/lgb_model.py
```diff
@@ -285,8 +284,16 @@ def _fit(self, X, y, X_val=None, y_val=None, time_limit=None, num_gpus=0, num_cp
...
+
+        if "seed_value" in params:
+            logger.log(
+                40,
+                "\t'seed_value' found in hyperparamters is ignored!"
+                "AutoGluon controls the random seed for the model training.",
+            )
+
+        if random_seed is not None:
+            train_params["params"]["seed"] = random_seed
```
Innixma: I wonder if we should have a dedicated method for handling seed so it is easier to check if a model has implemented seed support.

tabular/src/autogluon/tabular/models/lgb/lgb_model.py
```diff
@@ -285,8 +284,16 @@ def _fit(self, X, y, X_val=None, y_val=None, time_limit=None, num_gpus=0, num_cp
             train_params["params"]["num_classes"] = self.num_classes
         elif self.problem_type == QUANTILE:
             train_params["params"]["quantile_levels"] = self.quantile_levels
-        if seed_val is not None:
-            train_params["params"]["seed"] = seed_val
+
+        if "seed_value" in params:
+            logger.log(
+                40,
+                "\t'seed_value' found in hyperparamters is ignored!"
```
Innixma: nit: hyperparameters

tabular/src/autogluon/tabular/models/lgb/lgb_model.py
```diff
@@ -285,8 +284,16 @@ def _fit(self, X, y, X_val=None, y_val=None, time_limit=None, num_gpus=0, num_cp
...
+                "AutoGluon controls the random seed for the model training.",
```
Innixma: Probably not ideal. Users should be able to specify random seeds. If they specify it, then we just use it instead of `random_seed`.

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -1034,6 +1034,11 @@ def fit(self, *, log_resources: bool = False, **kwargs):
             verbosity 2: logs only important information.
             verbosity 1: logs only warnings and exceptions.
             verbosity 0: logs only exceptions.
+        random_seed : int, default = 0
+            The random seed used to control the randomness during fitting the model.
+            By default, we keep a static seed of 0 to ensure reproducibility.
+            When using a bagged model, a static seed of 0 to n_splits-1 is used to ensure that each fold has a
+            different seed that is consistent across different models.
```
Innixma: I'm a bit torn here. random_seed is a hyperparameter, but is not being treated as one by this logic. Instead we could consider using `ag_args_fit` to store the value?

LennartPurucker: Mhm, also to the above, I think we need to answer for ourselves if we want to treat random_seed as an HP.

In all cases where this can be helpful, it indicates that the model is extremely unstable during training. Not sure if we will ever have such a model. At the same time, one could give each HP config a different (start) seed and thus likely get better (val-overfitted) MLPs. 

Adding the random_seed start valuet to ag_args_fit

LennartPurucker:
@Innixma I have reworked the PR based on your feedback. Again, with some larger design choices that need your feedback. 

Moreover, I have added a test case and am now extending it to all other models in AG

core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
```diff
@@ -108,6 +108,8 @@ def _set_default_params(self):
             "stratify": "auto",
             "bin": "auto",
             "n_bins": None,
+            "vary_seed_across_folds": True, # If True, the seed used for each fold will be varied across folds.
```
Innixma: For this PR, lets set it to default False so we enable the ability to use the logic, but make it so the logic doesn't yet change AutoGluon's default behavior until we benchmark things and decide if it is beneficial.

core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
```diff
@@ -938,9 +945,13 @@ def _generate_fold_configs(
                     folds_to_fit=folds_to_fit,
                     folds_finished=fold - fold_start,
                     folds_left=fold_end - fold,
+                    random_seed=random_seed_value
                 )
 
                 fold_fit_args_list.append(fold_ctx)
+                if vary_seed_across_folds:
+                    random_seed_value += 1
```
Innixma: This is technically incorrect in the scenario where `n_repeat_start` and `fold_in_set_start` don't equal 0, such as when fitting an extra repeat of the bag in a future fit call.

The way to do it correctly would be to ensure the Nth child model gets a random_seed_value of `n-1`. Code will be a bit more complicated, but pretty sure there is a smart way to do it, as I've already had to deal with this same problem in the past somewhere in the bagged ensemble code.

tabular/src/autogluon/tabular/models/lgb/lgb_model.py
```diff
@@ -128,7 +128,12 @@ def _estimate_memory_usage_static(
         approx_mem_size_req = data_mem_usage_bytes + histogram_mem_usage_bytes + mem_size_estimators
         return approx_mem_size_req
 
-    def _fit(self, X, y, X_val=None, y_val=None, time_limit=None, num_gpus=0, num_cpus=0, sample_weight=None, sample_weight_val=None, verbosity=2, **kwargs):
+    def _get_random_seed_from_hyperparameters(self, hyperparameters: dict | None = None) -> tuple[bool, int | None]:
+        if "seed_value" in hyperparameters:
+            return True, hyperparameters["seed_value"]
```
Innixma: Is it only `seed_value`? Unsure why, but the AG code also uses `seed` as in:

```python
        if seed_val is not None:
            train_params["params"]["seed"] = seed_val
```

Maybe I wasn't comprehensive when I originally coded it.

tabular/tests/unittests/models/advanced/test_model_random_seed.py
```diff
@@ -0,0 +1,91 @@
...
+    Tests 2 fit types:
+    sequential fit + sequential bag
+    sequential fit + parallel bag
+    parallel fit + parallel bag
+    parallel fit + sequential bag
+    """
+    sample_size = 50
+    dataset_name = "adult"
+    directory_prefix = "./datasets/"
+    train_data, _, dataset_info = FitHelper.load_dataset(name=dataset_name, directory_prefix=directory_prefix)
```
Innixma: nit: we can use the toy datasets I generate for the other unit tests to be faster and avoid an s3 download.

```
dataset_name = "toy_binary"
train_data, test_data, dataset_info = FitHelper.load_dataset(name=dataset_name, directory_prefix=directory_prefix)
```

tabular/src/autogluon/tabular/models/lgb/lgb_model.py
```diff
@@ -285,8 +289,10 @@ def _fit(self, X, y, X_val=None, y_val=None, time_limit=None, num_gpus=0, num_cp
             train_params["params"]["num_classes"] = self.num_classes
         elif self.problem_type == QUANTILE:
             train_params["params"]["quantile_levels"] = self.quantile_levels
-        if seed_val is not None:
-            train_params["params"]["seed"] = seed_val
+
+        self.init_random_seed(random_seed=random_seed, hyperparameters=params)
```
Innixma: I'd like to avoid having the model developer always need to call this. Why not do it in `fit` instead of `_fit`? That way we can do it alongside all the other initialization logic. Maybe in the `self._init_params()` call?

core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
```diff
@@ -800,6 +802,8 @@ def _fit_folds(
             k_fold_end=k_fold_end,
             n_repeat_start=n_repeat_start,
             n_repeat_end=n_repeats,
+            vary_seed_across_folds=self.params["vary_seed_across_folds"],
+            random_seed_start_value=self.params["model_random_seed"],
```
Innixma: So currently BaggedEnsembleModel ignores `random_seed` in its own `_fit` call. This seems strange, as that means that nested `BaggedEnsembleModel` would not handle `random_seed` properly, as whatever is specified by the parent `BaggedEnsembleModel` would be ignored by the children `BaggedEnsembleModel`.

Maybe we don't care enough to address this gap, but good to document it if we decide not to fix this.

tabular/src/autogluon/tabular/models/lgb/lgb_model.py
LennartPurucker: ~~Okay, but do we have the correct hyperparameters at that time? 
I was not sure but thought that some model might only get them in the `_fit` call.~~

Should work

tabular/src/autogluon/tabular/models/lgb/lgb_model.py
LennartPurucker: Fixing both is needed for correct backward compatibility.
Here, seed_val was overwritting seed

tabular/src/autogluon/tabular/models/lgb/lgb_model.py
Innixma: correct hyperparameter values are determined in the `self._init_params()` call, as that is when they are set

core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
```diff
@@ -109,12 +109,15 @@ def _set_default_params(self):
...
             "vary_seed_across_folds": False, # If True, the seed used for each fold will be varied across folds.
-            "model_random_seed": 0,  # Start value for the random seed used for the seeds of the fold models.
+            "model_random_seed": 0,
         }
         for param, val in default_params.items():
             self._set_default_param_value(param, val)
         super()._set_default_params()
 
+    def _get_random_seed_from_hyperparameters(self, hyperparameters: dict | None = None) -> int | None | str:
+        return hyperparameters.get("model_random_seed", "N/A")
```
Innixma: Crashes if `hyperparameters is None`?

LennartPurucker: Ah nvm, should only be called if it is not None, fixing types...

core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
LennartPurucker: This case should be supported now as well

LennartPurucker:
Important to note: when we refit, we reset the seed to the default value, as we are not in a bagged mode anymore. That should be fine however, as we do not know the "correct" seed to set for refitting. Moreover, models with `use_child_oof=True` will also only use the default seed.

core/src/autogluon/core/models/ensemble/bagged_ensemble_model.py
LennartPurucker: Should be fixed with the latest update. If I understand the code correctly, `fold` gives the `n` of all splits.

LennartPurucker:
The current version should now be close to the final version

Innixma:
looks like test is failing

LennartPurucker:
Should be fixed now

LennartPurucker:
The latest version now also supports child_oof and refit_folds to have different seeds if one passed `model_random_seed` to `ag_args_ensemble`.

tabular/src/autogluon/tabular/models/tabular_nn/torch/tabular_nn_torch.py
```diff
@@ -164,6 +164,10 @@ def _prepare_params(params):
 
         return processor_kwargs, optimizer_kwargs, fit_kwargs, loss_kwargs, params
 
+    def _get_random_seed_from_hyperparameters(self, hyperparameters: dict) -> int | None | str:
+        return hyperparameters.get("seed_value", "N/A")
+
+
```
Innixma:
```suggestion
```

PR #5272, opened 2025-08-02
Mitra HF Args

xiyuanzh:
*Issue #, if available:*

*Description of changes:* Pass HuggingFace model checkpoint args to MitraClassifier and MitraRegressor


tabular/src/autogluon/tabular/models/mitra/sklearn_interface.py
```diff
@@ -274,6 +274,8 @@ def __init__(self,
             fine_tune_steps=DEFAULT_FINE_TUNE_STEPS,
             metric=DEFAULT_CLS_METRIC,
             state_dict=None,
+            hf_general_model=DEFAULT_GENERAL_MODEL,
+            hf_cls_model=DEFAULT_CLS_MODEL,
```
Innixma: If we are going to avoid passing `hf_reg_model`, we shouldn't even have `hf_cls_model`, and we would rename `hf_general_model` to `hf_model`? Same idea for Regressor.

Innixma: This code is still going to crash if the user specifies both `hf_cls_model` and `hf_reg_model` through AutoGluon. You'll need to change the code in `MitraModel` as well.

Innixma (changes requested):
Refer to above comments

tabular/src/autogluon/tabular/models/mitra/sklearn_interface.py
xiyuanzh: Yeah we initially discussed about having `hf_general_model`, `hf_cls_model` and `hf_reg_model`, but it is not common practice to pass a classification checkpoint to a regressor model and vice versa. I have removed `hf_cls_model` and `hf_reg_model`, and renamed `hf_general_model` to `hf_model`, thanks a lot!

xiyuanzh: Yes updated as above, thanks!

tabular/src/autogluon/tabular/models/mitra/sklearn_interface.py
```diff
@@ -200,20 +196,14 @@ def _train_ensemble(self, X_train, y_train, X_valid, y_valid, task, dim_output,
...
-                            if self.hf_cls_model is not None:
-                                model = Tab2D.from_pretrained(self.hf_cls_model, device=self.device)
-                            elif self.hf_general_model is not None:
-                                model = Tab2D.from_pretrained(self.hf_general_model, device=self.device)
-                            else:
-                                model = Tab2D.from_pretrained("autogluon/mitra-classifier", device=self.device)
+                        if self.hf_model is not None:
+                            model = Tab2D.from_pretrained(self.hf_model, device=self.device)
+                        elif task == 'classification':
+                            model = Tab2D.from_pretrained("autogluon/mitra-classifier", device=self.device)
```
Innixma: We still need the ability to specify classifier and regressor models as different checkpoints / locations. We should move that logic to the AutoGluon MitraModel class.

PR #5279, opened 2025-08-12
Chronos: allow default downloading of weights

gabrielsinger2:
*Issue #, if available:*
Hello, for time series foundation models, sometimes PyTorch random initialization can be a strong baseline.
In the chronos-forecasting repository you already allowed random initialization, but here in AutoGluon it’s not the case.
*Description of changes:*
  I just updated this feature in the AutoGluon library for Chronos models by adding an init_random parameter.
I modified the load_model_pipeline method to include a condition on the init_random flag:

    1) If init_random is True, the model is loaded directly from its configuration, and it checks whether we are in training or evaluation mode (disabled during evaluation to ensure the correct model is loaded).

    2) If init_random is False, the original loading behavior is used.

I have conducted many tests to ensure I can reproduce results and verify that everything is correctly loaded during both training and inference.

Any feedback is welcome, do you think this addition would be useful for your libraries :) ?

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/chronos/model.py
+++ b/timeseries/src/autogluon/timeseries/models/chronos/model.py
@@ -12,14 +12,21 @@ from autogluon.common.loaders import load_pkl
 from autogluon.common.space import Space
 from autogluon.timeseries.dataset.ts_dataframe import TimeSeriesDataFrame
 from autogluon.timeseries.models.abstract import AbstractTimeSeriesModel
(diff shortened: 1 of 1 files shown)


shchur:
Hi @gabrielsinger2, this looks more like an advanced research use case, so I don't think that we need to add & maintain this code in AutoGluon given a relatively simple workaround for the users that are interested:
1. Randomly initialize the model using the Chronos model from`chronos-forecasting`
2. Save the checkpoint to disk
3. Load the randomly initialized checkpoint using `model_path` in AutoGluon

PR #5285, opened 2025-08-15
[Tabular] Changing the default output dir's base path

LennartPurucker:
*Description of changes:*

By default, AutoGuon saves any artifacts into a folder within the working directory when no path is given. This logic is quite useful, but in some cases, users might want to save the models in a different folder while keeping AutoGluon's logic to determine how to create a new time-stamped subfolder. This PR adds some documentation and support for such functionality.

In particular, the use case I have is that I run a lot of AutoGluon models and want to write the artifacts by default to `/tmp` instead of the working directory while keeping the multiprocessing-safe time-stamp logic. 

### DIFF
--- a/tabular/src/autogluon/tabular/predictor/predictor.py
+++ b/tabular/src/autogluon/tabular/predictor/predictor.py
@@ -165,7 +165,7 @@ class TabularPredictor:
             trainer_type : AbstractTabularTrainer, default = AutoTrainer
                 A class inheriting from `AbstractTabularTrainer` that controls training/ensembling of many models.
                 If you don't know what this is, keep it as the default.
-        default_base_path : str | Path | bool, default = None
+        default_base_path : str | Path | None, default = None
             A default base path to use for the time-stamped folder if `path` is None.
             If None, defaults to `AutogluonModels`. Only used if `path` is None, and thus
             only used for local paths, not s3 paths.


tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -201,7 +205,7 @@ def __init__(
                 f"We do not recommend specifying weight_evaluation when sample_weight='{self.sample_weight}', instead specify appropriate eval_metric."
             )
         self._validate_init_kwargs(kwargs)
-        path = setup_outputdir(path)
+        path = setup_outputdir(path=path, default_base_path=kwargs.get("default_base_path"))
```
Innixma, line 208: Missing default if not specified?

Innixma (commented):
Added a comment, looks good to me otherwise!

tabular/src/autogluon/tabular/predictor/predictor.py
Innixma: Nevermind, this works

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -165,6 +165,10 @@ class TabularPredictor:
             trainer_type : AbstractTabularTrainer, default = AutoTrainer
                 A class inheriting from `AbstractTabularTrainer` that controls training/ensembling of many models.
                 If you don't know what this is, keep it as the default.
+        default_base_path : str | Path | bool, default = None
```
Innixma:
```suggestion
        default_base_path : str | Path | None, default = None
```

PR #5292, opened 2025-08-19
[common] minor typing fixes

canerturkmen:
*Issue #, if available:*

*Description of changes:*

Minor improvements to typing, primarily for time series. 

### DIFF
--- a/common/src/autogluon/common/loaders/load_pkl.py
+++ b/common/src/autogluon/common/loaders/load_pkl.py
@@ -13,7 +13,7 @@ from ..utils import compression_utils, s3_utils
 logger = logging.getLogger(__name__)
 
 
-def load(path: str, format: str | None = None, verbose: bool = True, **kwargs) -> object:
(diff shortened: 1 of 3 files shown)


common/src/autogluon/common/loaders/load_pkl.py
```diff
@@ -13,7 +13,7 @@
 logger = logging.getLogger(__name__)
 
 
-def load(path: str, format: str | None = None, verbose: bool = True, **kwargs) -> object:
+def load(path: str, format: str | None = None, verbose: bool = True, **kwargs):
```
canerturkmen: This line forced the type checker to break in pretty much all invocations of this function, and not for a helpful reason. Having an `object` return type tells the type checker to force all consumers of this function to handle _any_ object. Therefore, in order not to have the type checker break, one has to assert the type at runtime for all invocations which I think unnecessarily pollutes the code... I think this is a good place to leave this return type `Unknown`.

PR #5297, opened 2025-08-20
[timeseries] refactor trainable model set build logic

canerturkmen:
*Issue #, if available:*

*Description of changes:*

This PR intends to refactor the model set building logic (`get_preset_models`) into classes with cleaner encapsulation and move them into the Trainer which depend on them.

- [ ] move classes closer to trainer


timeseries/src/autogluon/timeseries/models/presets.py
```diff
@@ -131,150 +177,137 @@ def get_preset_models(
...
-
-def check_and_clean_hyperparameters(
-    hyperparameters: Dict[str, Union[ModelHyperparameters, List[ModelHyperparameters]]],
-    must_contain_searchspace: bool,
-) -> Dict[str, List[ModelHyperparameters]]:
-    """Convert the hyperparameters dictionary to a unified format:
-    - Remove 'Model' suffix from model names, if present
-    - Make sure that each value in the hyperparameters dict is a list with model configurations
-    - Checks if hyperparameters contain searchspaces
+class HyperparameterBuilder:
```
shchur: I'm a bit on the edge about moving this logic to two new classes. As far as I understand, both classes are essentially stateless (except the `banned_model_names`) and serve only for bundling the main public method with the helper methods. Could we achieve the same by 1/ adding a leading underscore to the methods that shouldn't be used outside of `models/presets.py` and 2/ exposing two public methods `resolve_hyperparameters` and `create_model_templates`?

Currently when I see 
```python
from .presets import HyperparameterBuilder, TrainableModelSetBuilder
```
I expect these classes to keep track of some state or each provide multiple methods, but then it turns out that each class only offers one public method that is called once.

canerturkmen: I understand the urge, but I would push back that class -> mutable state. Here they are meant to bundle methods with some cohesion, and the cohesion is through what lives (and may in the future live) in `self`. The other perhaps slightly more pythonic but slightly less elegant solution I see is to make each class a module. However, here the problem is that functions will pass around `freq`, `excluded_model_names` etc. to each other in a "bucket brigade" and every time we decide to add a new field (e.g., `covariate_metadata`) the impact will be harder to contain, etc.

shchur: Let me try breaking down the motivation here to see if I understand it correctly.

The main purpose of introducing the classes is to bundle the methods together. Bundling could also be accomplished by making `get_hyperparameters` and `get_model_set` classmethods. 
```python
model_set = TrainableModelSetBuilder.get_model_set(hyperparameters, freq=freq, prediction_length=prediction_length, ...)
```
However, that would require passing all "task properties" into each method call, which is tedious. Probably the cleanest option if we were designing from scratch would be to incapsulate all of these properties into a `Task` class, but we don't want to do that now.

As a compromise, we put the task properties into the `__init__` call to highlight that they are different from e.g. the `hyperparameters` that are the "main input" to the method
```python
model_set = TrainableModelSetBuilder(freq=freq, prediction_length=prediction_length, ...).get_model_set(hyperparameters, ...)
```

If that's our motivation, then I'm on board with it. The only suggestion I have would be to merge `HyperparameterBuilder` and `TrainableModelSetBuilder` into a single model builder class that maps raw hyperparameters -> trainable model set. Currently, I feel that introducing two classes whose lifetime is limited to a single method that is called once during the training process is a bit of an overkill.

timeseries/src/autogluon/timeseries/models/presets.py
```diff
@@ -18,109 +19,154 @@
...
+        self,
+        hyperparameters: Dict[str, List[ModelHyperparameters]],
+    ) -> List[TimeSeriesModelBase]:
+        """Create a list of models according to given resolved and canonicalized dictionary of hyperparameters.
+        Hyperparameters can be built using the `HyperparameterBuilder` class.
+        """
+        models = []
+
+        model_priority_list = sorted(
+            hyperparameters.keys(), key=lambda x: ModelRegistry.get_model_priority(x), reverse=True
```
shchur: The following check in [L172](https://github.com/autogluon/autogluon/pull/5297/files#diff-ed1c700fc46c7469e27774fb55d79f04acab317fdfa60f8f817c758ee4248219R172)
```python
if not issubclass(model, AbstractTimeSeriesModel):
```
is unreachable because L124 will already raise an exception if `x` is a custom model type that does not inherit from `AbstractTimeSeriesModel` and therefore is not registered in the ModelRegistry

https://github.com/autogluon/autogluon/blob/e6d7faee18fe957fadabd8390cb3c8e9cde8be01/timeseries/src/autogluon/timeseries/models/registry.py#L51

timeseries/src/autogluon/timeseries/models/presets.py
```diff
@@ -18,109 +19,154 @@
...
+        multi_window: bool,
+    ):
+        self.path = path
+        self.freq = freq
+        self.prediction_length = prediction_length
+        self.eval_metric = eval_metric
+        self.target = target
+        self.quantile_levels = quantile_levels
+        self.covariate_metadata = covariate_metadata
+        self.banned_model_names = banned_model_names.copy()
```
shchur: If the purpose of `__init__` is to separate the task definition from the "actual inputs needed to build the model set", I would recommend making this class stateless by passing `banned_model_names` as an argument to `get_model_set`. This looks like a small price to pay to make the method pure and avoid ever needing to think about the state of the model builder object.

canerturkmen: I agree and will do it. However, I want to point out also the price of using classes to do this, if just to further the discussion. We can do it now, but there's nothing that enforces statelessness in the language. It will always be up to the user to know how to use the class, without assuming it is stateless. I guess this is partly what the "functional programming imposes discipline on state/assignment." quote means.

timeseries/src/autogluon/timeseries/models/presets.py
canerturkmen: middle ground: I didn't want to merge back two relatively well separated concerns, but had the TrainableModelSetBuilder hide HyperparameterBuilder from Trainer.

timeseries/src/autogluon/timeseries/typing.py
```diff
@@ -0,0 +1,3 @@
+from typing import Any
+
+ModelHyperparameters = dict[str, Any]
```
shchur: Congrats on crossing the milestone of having our own `.typing` module 😄

timeseries/src/autogluon/timeseries/configs/hyperparameter_presets.py
```diff
@@ -0,0 +1,62 @@
+from typing import Union
+
+from autogluon.timeseries.typing import ModelHyperparameters
+
+HYPERPARAMETER_PRESETS: dict[str, dict[str, Union[ModelHyperparameters, list[ModelHyperparameters]]]] = {
```
shchur: How about we convert `HYPERPARAMETER_PRESETS` -> `_HYPERPARAMETER_PRESETS` into a private variable and keep accessing it via a method that forces a deepcopy?

Otherwise we need to be very careful to always perform a deepcopy on `HYPERPARAMETER_PRESETS` before usage - I can see how we forget about this at some point in the future and introduce some really nasty bugs that change the behavior of repeated calls to `predictor.fit`.

timeseries/src/autogluon/timeseries/configs/predictor_presets.py
```diff
@@ -1,11 +1,11 @@
 """Preset configurations for autogluon.timeseries Predictors"""
 
-from autogluon.timeseries.models.presets import get_default_hps
+from . import HYPERPARAMETER_PRESETS
 
 # TODO: change default HPO settings when other HPO strategies (e.g., Ray tune) are available
 # TODO: add refit_full arguments once refitting is available
 
-TIMESERIES_PRESETS_CONFIGS = dict(
+TIMESERIES_PREDICTOR_PRESETS = dict(
```
shchur: Same comment as above: should we hide this variable behind a method that enforces a deepcopy?

timeseries/src/autogluon/timeseries/trainer/model_set_builder.py
```diff
@@ -0,0 +1,247 @@
...
+                f"hyperparameters must be a dict, a string or None (received {type(self.hyperparameters)}). "
+                f"Please see the documentation for TimeSeriesPredictor.fit"
+            )
+
+        return self._check_and_clean_hyperparameters(hyperparameter_dict)
+
+    def _check_and_clean_hyperparameters(
+        self,
+        hyperparameters: dict[str, Union[ModelHyperparameters, list[ModelHyperparameters]]],
+    ) -> dict[str, list[ModelHyperparameters]]:
```
shchur: This method (as well as `get_hyperparameters`) are not guaranteed to return `dict[str, list[ModelHyperparameters]]` since the keys in the `hyperparameters` dictionary could be custom model types of type `Type[AbstractTimeSeriesModel]`.

timeseries/src/autogluon/timeseries/trainer/model_set_builder.py
```diff
@@ -0,0 +1,247 @@
...
+
+        # resolve and normalize hyperparameters
+        model_hp_map: dict[str, list[ModelHyperparameters]] = HyperparameterBuilder(
+            hyperparameters=hyperparameters,
+            hyperparameter_tune=hyperparameter_tune,
+            excluded_model_types=excluded_model_types,
+        ).get_hyperparameters()
+
+        model_priority_list = sorted(
+            model_hp_map.keys(), key=lambda x: ModelRegistry.get_model_priority(x), reverse=True
```
shchur, line 77: Sorry, I'm a bit confused, it looks like this can still error out if `model_hp_map` contains custom model types as keys. Is this case handled somewhere now?

shchur: Should we add a unit test that checks that `TrainableModelSetBuilder` errors correctly when a custom model is provided that does not inherit from `AbstractTimeSeriesModel`?

timeseries/src/autogluon/timeseries/trainer/model_set_builder.py
```diff
@@ -0,0 +1,247 @@
...
+        self.covariate_metadata = covariate_metadata
+        self.multi_window = multi_window
+
+    def get_model_set(
+        self,
+        hyperparameters: Union[str, dict, None],
+        hyperparameter_tune: bool,
+        excluded_model_types: Optional[list[str]],
+        banned_model_names: Optional[list[str]] = None,
+    ) -> list[TimeSeriesModelBase]:
```
shchur: It currently looks a bit odd that this method returns `TimeSeriesModelBase`, but all other checks in the class are only concerned with `AbstractTimeSeriesModel`. Let's keep everything as `AbstractTimeSeriesModel` for now and then change to `TimeSeriesModelBase` if we refactor `MultiWindowBacktestingModel` to only inherit from `TimeSeriesModelBase` or decide to include the ensemble classes in the trainable model set as well?

timeseries/src/autogluon/timeseries/models/presets.py
shchur: This looks great, thanks!

timeseries/src/autogluon/timeseries/trainer/model_set_builder.py
canerturkmen: makes sense but the impact is slightly larger than you'd imagine. The `core.AbstractTrainer` interface kind of enforced this. I removed `construct_model_templates` from `AbstractTrainer` and also renamed the method while I'm at it.

timeseries/src/autogluon/timeseries/trainer/model_set_builder.py
canerturkmen: Think it's handled within ModelRegistry?

timeseries/src/autogluon/timeseries/trainer/model_set_builder.py
canerturkmen: removed typing module and fixed these

timeseries/src/autogluon/timeseries/trainer/model_set_builder.py
shchur: The error message here 
https://github.com/autogluon/autogluon/blob/8a275e8f373d7c6c0adfa4f3822290234ac9056c/timeseries/src/autogluon/timeseries/models/registry.py#L46-L53
is imo a lot less clear than the old error message that we had [here](https://github.com/autogluon/autogluon/pull/5297/files#diff-ed1c700fc46c7469e27774fb55d79f04acab317fdfa60f8f817c758ee4248219L126-L127)

```python
            if not issubclass(model, AbstractTimeSeriesModel):
                raise ValueError(f"Custom model type {model} must inherit from `AbstractTimeSeriesModel`.")
```

shchur (approved):
Just one small comment, otherwise LGTM!

Issue #5074, opened 2025-04-22
[timeseries] Clean up type annotations in the `timeseries` module


shchur:
## Description
- [x] Remove type annotations from docstrings of non-public classes and methods (e.g. [`AbstractTimeSeriesModel`](https://github.com/autogluon/autogluon/blob/28a242ebe8d55ba770c991b9db153ab4623c9abd/timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py#L39-L64)). The type annotations are already available in the method/class signatures, and there is no need to duplicate them in the docstrings.
- [x] Now that we dropped support for Python 3.8, we can safely use the newer type aliases (e.g., replace `List` with `list`)

PR #5300, opened 2025-08-21
[timeseries] use builtin generics for typing, remove types in internal docstrings

canerturkmen:
*Issue #, if available:*

#5074 

*Description of changes:*

- Removes types in docstrings in internal APIs (considering `TimeSeriesScorer` and `CovariateRegressor` as public API)
- Switch to builtin generics (`set`, `tuple`, `list`, `dict`) timeseries-wide.


timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py
```diff
@@ -493,26 +493,26 @@ class DirectTabularModel(AbstractMLForecastModel):
 
     Other Parameters
     ----------------
-    lags : List[int], default = None
```
shchur, line 496: Looks like an erroneous change, model hyperparameters should preserve the type hints since they are displayed in https://auto.gluon.ai/stable/tutorials/timeseries/forecasting-model-zoo.html

timeseries/src/autogluon/timeseries/models/ensemble/basic.py
```diff
@@ -34,7 +34,7 @@ class PerformanceWeightedEnsemble(AbstractWeightedTimeSeriesEnsembleModel):
 
     Other Parameters
     ----------------
-    weight_scheme: Literal["sq", "inv", "loginv"], default = "loginv"
```
shchur: I would preserve the type hints for ensemble models params as well in case we expose them similarly to https://auto.gluon.ai/stable/tutorials/timeseries/forecasting-model-zoo.html in a future release

timeseries/src/autogluon/timeseries/models/autogluon_tabular/per_step.py
```diff
@@ -49,34 +49,34 @@ class PerStepTabularModel(AbstractTimeSeriesModel):
 
     Other Parameters
     ----------------
-    trailing_lags : List[int], default = None
```
shchur, line 52: Seems that the docstrings should be updated for a few more models:
- `mlforecast/per_step.py`
- `chronos/model.py`

PR #5302, opened 2025-08-22
Add rare category fillna logic

Innixma:
*Issue #, if available:*

*Description of changes:*

- Add rare category fillna logic
- Add unit test for rare category logic
- Cleanup unit test

### DIFF
--- a/features/src/autogluon/features/generators/category.py
+++ b/features/src/autogluon/features/generators/category.py
@@ -46,7 +46,7 @@ class CategoryFeatureGenerator(AbstractFeatureGenerator):
             'original' : Keep the original order. If the feature was originally an object, this is equivalent to 'alphanumeric'.
             'alphanumeric' : Sort the categories alphanumerically.
             'count' : Sort the categories by frequency (Least frequent in front with code of 0)
-    minimum_cat_count : int, default None
+    minimum_cat_count : int, default 2
         The minimum number of occurrences a category must have in the training data to avoid being considered a rare category.
         Rare categories are removed and treated as missing values.
         If None, no minimum count is required. This includes categories that never occur in the data but are present in the category object
@@ -60,6 +60,7 @@ class CategoryFeatureGenerator(AbstractFeatureGenerator):
         Valid values:
             None : Keep missing values as is. They will appear as NaN and have no category assigned to them.
             'mode' : Set missing values to the most frequent category in their feature.
+            'rare' : Set missing values to the rarest category in their feature that satisfies `minimum_cat_count` and `maximum_num_cat`.
     **kwargs :
         Refer to :class:`AbstractFeatureGenerator` documentation for details on valid key word arguments.
     """
@@ -85,8 +86,8 @@ class CategoryFeatureGenerator(AbstractFeatureGenerator):
         self._maximum_num_cat = maximum_num_cat
         self.category_map = None
         if fillna is not None:
-            if fillna not in ["mode"]:
-                raise ValueError(f"fillna={fillna} is not a valid value. Valid values: {[None, 'mode']}")
+            if fillna not in ["mode", "rare"]:
+                raise ValueError(f"fillna={fillna} is not a valid value. Valid values: {[None, 'mode', 'rare']}")
         self._fillna = fillna
         self._fillna_flag = self._fillna is not None
         self._fillna_map = None
@@ -169,11 +170,27 @@ class CategoryFeatureGenerator(AbstractFeatureGenerator):
                     category_list.sort()
                     X_category[column] = X_category[column].astype(CategoricalDtype(categories=category_list))
                     X_category[column] = X_category[column].cat.reorder_categories(category_list)
-                category_map[column] = copy.deepcopy(X_category[column].cat.categories)
                 if self._fillna_flag:
+                    nan_count = X_category[column].isna().sum()
                     if self._fillna == "mode":
                         if len(rank) > 0:
                             fill_nan_map[column] = list(rank.index)[-1]
+                    elif self._fillna == "rare":  # fill by the rarest category
+                        if len(rank) > 0:
+                            if self._minimum_cat_count is None or (nan_count >= self._minimum_cat_count):
+                                category_list = list(X_category[column].cat.categories)
+                                if all([isinstance(c, int) for c in category_list]):
+                                    fillna_category = max(list(rank.index)) + 1
+                                else:
+                                    fillna_category = "_NaN_"
+                                category_list.append(fillna_category)
+                                X_category[column] = X_category[column].astype(CategoricalDtype(categories=category_list))
+                                X_category[column] = X_category[column].cat.reorder_categories(category_list)
+                                # create a new category for NaNs
+                                fill_nan_map[column] = fillna_category
+                            else:
+                                fill_nan_map[column] = list(rank.index)[0]
+                category_map[column] = copy.deepcopy(X_category[column].cat.categories)
             if not self._fillna_flag:
                 fill_nan_map = None
             return X_category, category_map, fill_nan_map
--- a/features/src/autogluon/features/generators/category.py
+++ b/features/src/autogluon/features/generators/category.py
@@ -184,7 +184,9 @@ class CategoryFeatureGenerator(AbstractFeatureGenerator):
                                 else:
                                     fillna_category = "_NaN_"
                                 category_list.append(fillna_category)
-                                X_category[column] = X_category[column].astype(CategoricalDtype(categories=category_list))
+                                X_category[column] = X_category[column].astype(
+                                    CategoricalDtype(categories=category_list)
+                                )
                                 X_category[column] = X_category[column].cat.reorder_categories(category_list)
                                 # create a new category for NaNs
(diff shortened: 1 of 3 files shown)


features/src/autogluon/features/generators/category.py
```diff
@@ -85,8 +86,8 @@ def __init__(
         self._maximum_num_cat = maximum_num_cat
         self.category_map = None
         if fillna is not None:
-            if fillna not in ["mode"]:
-                raise ValueError(f"fillna={fillna} is not a valid value. Valid values: {[None, 'mode']}")
+            if fillna not in ["mode", "rare"]:
+                raise ValueError(f"fillna={fillna} is not a valid value. Valid values: {[None, 'mode', 'rare']}")
```
eddiebergman: nit:

```python
valid_values = ("mode", "rare", None)
if fillna not in valid_methods:
    raise ValueError(f"{fillna=} is not a valid value. Valid values: {valid_values}")
```

features/src/autogluon/features/generators/category.py
```diff
@@ -169,11 +170,29 @@ def _generate_category_map(self, X: DataFrame) -> (DataFrame, dict):
...
+                                category_list.append(fillna_category)
+                                X_category[column] = X_category[column].astype(
+                                    CategoricalDtype(categories=category_list)
+                                )
+                                X_category[column] = X_category[column].cat.reorder_categories(category_list)
+                                # create a new category for NaNs
+                                fill_nan_map[column] = fillna_category
+                            else:
+                                fill_nan_map[column] = list(rank.index)[0]
+                category_map[column] = copy.deepcopy(X_category[column].cat.categories)
```
eddiebergman: nit:
* `nan_count` only required in the `elif self._fillna == "rare"` block.
* if `all(isinstance(c, int) for c in category_list)`, no need to have an explicit `[]` around it.

features/src/autogluon/features/generators/category.py
```diff
@@ -169,11 +170,29 @@ def _generate_category_map(self, X: DataFrame) -> (DataFrame, dict):
...
                 if self._fillna_flag:
+                    nan_count = X_category[column].isna().sum()
                     if self._fillna == "mode":
                         if len(rank) > 0:
                             fill_nan_map[column] = list(rank.index)[-1]
+                    elif self._fillna == "rare":  # fill by the rarest category
+                        if len(rank) > 0:
+                            if self._minimum_cat_count is None or (nan_count >= self._minimum_cat_count):
+                                category_list = list(X_category[column].cat.categories)
+                                if all([isinstance(c, int) for c in category_list]):
```
eddiebergman: To be annoying, you could end up having `[1, 2.0, 3]` as categories, meaning you hit the `_NaN_` case, where it seems like integer categories might still makes sense, but I feel like handling that might be too much.

However, you could technically have categories `[1.1, 2.2, 3.05]`, and still add `+1` on to them and have valid categories. Not sure of the surrounding context and if that makes sense.

eddiebergman (approved):
... it said review needed, hello :)

Innixma:
Rebased onto master (this branch was ~320 commits behind; the rebase was clean) and pushed one follow-up commit. Summary of what changed and why.

## The `test_features` failure that was blocking this

`test_category_feature_generator` was failing, and it was this PR's own doing. The test used to assert `list(output) in [variant_a, variant_b]`; this PR tightened it to exact equality. The two variants existed for a reason: the test data is `["a","b","a","d","d","d","c",nan,nan]`, so **`b` and `c` are tied at count 1**, and `rank` was built with

```python
X_category[column].value_counts().sort_values(ascending=True)
```

`sort_values` defaults to `kind="quicksort"`, which is **not stable**, so the resolved order of tied categories varies by pandas/numpy build. Locally this produced `[2, 1, 2, 3, 3, 3, 0, nan, nan]` against the expected `[2, 0, 2, 3, 3, 3, 1, nan, nan]`.

Rather than loosen the assertions back, I made the ranking deterministic:

```python
rank = X_category[column].value_counts().sort_index().sort_values(ascending=True, kind="stable")
```

`sort_index()` gives a deterministic starting order regardless of `value_counts()` internals, and `kind="stable"` preserves it through the count sort, so ties now break alphanumerically on every platform. This produces exactly the values the PR already expected, and `features/tests/features` is green (89 passed).

## Does this change default behaviour?

**The original PR content does not** — it is purely additive:

- `fillna` still defaults to `None`; `"rare"` is opt-in.
- The `minimum_cat_count : int, default None` → `default 2` edit is a **docstring correction**; the signature on master already reads `minimum_cat_count: int = 2`.
- Moving `category_map[column] = copy.deepcopy(...)` below the fillna block is a no-op for `fillna=None` (block skipped) and for `fillna="mode"` (the block only writes `fill_nan_map`, never `X_category[column]`). It is only load-bearing for `"rare"`, which needs the newly-appended category to reach transform time.

**The tie-break fix does, but not under the default config.** `rank`'s ordering feeds four things: the `maximum_num_cat` truncation `rank[-max:]` (i.e. which categories *survive*), `cat_order="count"` code assignment, `fillna="mode"`'s `rank.index[-1]`, and the new `"rare"` fallback's `rank.index[0]`. Under `AutoMLPipelineFeatureGenerator`'s `CategoryFeatureGenerator()` (`cat_order="original"`, `maximum_num_cat=None`, `fillna=None`) `rank` is used *only* for the order-independent `>= minimum_cat_count` filter, so nothing changes. It does change behaviour for `AutoMLInterpretablePipelineFeatureGenerator` — the only place `maximum_num_cat=10` is set — and for opt-in `cat_order="count"`.

That change seems clearly worth making: with eleven categories tied at a `maximum_num_cat` boundary, *which one gets dropped* was previously decided by an unstable sort over a `value_counts()` order that is itself a pandas implementation detail, so the same data and config could yield different features on different machines. Ties are the common case, not an exotic one — in a long tail every count-1 category is tied with every other. Happy to split it into its own PR if you would rather keep this one purely additive.

## Integer categories were getting a string category

`all([isinstance(c, int) for c in category_list])` is `False` for `np.int64`, and a category index is normally backed by numpy dtypes. So integer columns fell through to the `"_NaN_"` branch and ended up with a **mixed-type category index** — exactly what the integer branch was written to avoid:

```
CategoryFeatureGenerator(minimum_cat_count=None, fillna="rare", minimize_memory=False)
before: categories [1, 2, 3, '_NaN_']
after:  categories [1, 2, 3, 4]
```

Now `isinstance(c, (int, np.integer))`. Note `bool` is a subclass of `int`, so boolean categories keep their existing behaviour (`[False, True, 2]`) rather than silently changing.

## Docstring

The `'rare'` entry described only the *fallback* path ("set missing values to the rarest category…"). The primary path adds a **new** category; the rarest-existing-category behaviour only applies when fewer than `minimum_cat_count` values are missing. Both are now documented, along with the fact that values dropped by `minimum_cat_count`/`maximum_num_cat` are absorbed into the same bucket (which is the point of the feature), and that nothing happens when no category satisfies the constraints.

## Review nits (thanks @eddiebergman)

All three addressed:

1. Single `valid_fillna_values = (None, "mode", "rare")` tuple as the source of truth in the error message — which also let the `if fillna is not None:` wrapper go, since `None` is now a member.
2. `nan_count` moved into the `elif self._fillna == "rare"` branch; `fillna="mode"` no longer pays for a value it never reads. Also dropped the redundant list-comprehension brackets and the `list()` around `rank.index`.
3. On `[1, 2.0, 3]` / `[1.1, 2.2, 3.05]`: those land as `float64` category indexes, so they take the `"_NaN_"` branch and still get a mixed-type index (`[1.1, 2.2, 3.05, '_NaN_']`). Same for `bool` (`[False, True, 2]`). Both are unchanged from master and only reachable when a pre-existing categorical is handed in — `get_default_infer_features_in_args` is `[R_OBJECT, R_CATEGORY, R_BOOL]`, so no float column reaches this generator directly — and both are harmless under the default `minimize_memory=True`, which remaps categories to ordinals. Extending `max(...) + 1` to cover floats is a one-line change if you think it is worth it; I left it out to keep this a bug fix rather than a behaviour change.

## Tests

Three new focused tests, plus the existing suite:

- `test_category_feature_generator_fillna_rare_integer_categories` — pins the integer fix; **fails without it** (verified).
- `test_category_feature_generator_fillna_rare_object_categories` — the `"_NaN_"` path, including that a category unseen at fit time is grouped into the rare bucket at transform time (which is the desired behaviour here).
- `test_category_feature_generator_invalid_fillna` — the `ValueError`.

`features/tests/features` 89 passed. Also ran `test_lightgbm.py`, `test_catboost.py`, `test_xgboost.py` and `tabular/tests/unittests/experimental` (24 passed) to exercise the tie-break change downstream. `ruff format --check` and `ruff check --select I` clean.

## One open question on behaviour

`fillna="rare"` silently does nothing when every category is filtered out (`len(rank) == 0`, e.g. `minimum_cat_count=4` on the test data): the column comes back all-NaN. That is arguably the case where a single rare bucket is *most* useful. It is currently asserted as a no-op by `generator_9` in the test, so it reads as intentional — flagging it in case it was not.

LennartPurucker (approved):

PR #5308, opened 2025-08-26
[timeseries] typing improvements to multiwindow model

canerturkmen:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/timeseries/src/autogluon/timeseries/metrics/abstract.py
+++ b/timeseries/src/autogluon/timeseries/metrics/abstract.py
@@ -2,7 +2,6 @@ import warnings
 from typing import Optional, Sequence, Union, overload
 
 import numpy as np
-import numpy.typing as npt
 import pandas as pd
 
 from autogluon.timeseries import TimeSeriesDataFrame
@@ -242,12 +241,12 @@ class TimeSeriesScorer:
     @staticmethod
     def check_get_horizon_weight(
         horizon_weight: Union[Sequence[float], np.ndarray], prediction_length: int
-    ) -> npt.NDArray[np.float64]: ...
+    ) -> np.ndarray: ...
 
     @staticmethod
     def check_get_horizon_weight(
         horizon_weight: Union[Sequence[float], np.ndarray, None], prediction_length: int
-    ) -> Optional[npt.NDArray[np.float64]]:
+    ) -> Optional[np.ndarray]:
         """Convert horizon_weight to a non-negative numpy array that sums up to prediction_length.
         Raises an exception if horizon_weight has an invalid shape or contains invalid values.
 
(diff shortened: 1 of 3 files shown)


timeseries/src/autogluon/timeseries/metrics/abstract.py
```diff
@@ -242,12 +241,12 @@ def check_get_horizon_weight(horizon_weight: None, prediction_length: int) -> No
     @staticmethod
     def check_get_horizon_weight(
         horizon_weight: Union[Sequence[float], np.ndarray], prediction_length: int
-    ) -> npt.NDArray[np.float64]: ...
```
canerturkmen, line 245: this caused issues with a newer install, so removing for consistency also with the rest of the library

timeseries/src/autogluon/timeseries/models/multi_window/multi_window_model.py
```diff
@@ -238,32 +244,35 @@ def save(self, path: str = None, verbose=True) -> str:
             most_recent_model.save()
         return save_path
 
-    def persist(self):
+    def persist(self) -> AbstractTimeSeriesModel:
         if self.most_recent_model is None:
             raise ValueError(f"{self.name} must be fit before persisting")
-        self.most_recent_model.persist()
+        return self.most_recent_model.persist()
```
shchur: Should this method return `self` instead of `self.most_recent_model` (for consistency with `AbstractTimeSeriesModel.persist`)?
```suggestion
    def persist(self) -> Self:
        if self.most_recent_model is None:
            raise ValueError(f"{self.name} must be fit before persisting")
        self.most_recent_model.persist()
        return self
```

timeseries/src/autogluon/timeseries/models/multi_window/multi_window_model.py
```diff
@@ -111,11 +110,15 @@ def _fit(
...
         oof_predictions_per_window = []
         global_fit_start_time = time.time()
+        model: Optional[AbstractTimeSeriesModel] = None
 
         for window_index, (train_fold, val_fold) in enumerate(val_splitter.split(train_data)):
             logger.debug(f"\tWindow {window_index}")
+
             # refit_this_window is always True for the 0th window
             refit_this_window = window_index % refit_every_n_windows == 0
+            assert not window_index == 0 or refit_this_window
```
shchur: Maybe 
```python
assert window_index != 0 or refit_this_window
```
for better readability? I keep forgetting the order of `not` with other operations

canerturkmen: changed. fun fact is that `~p or q` is the only way I can remember `p -> q` so in my own way this was more readable to me :)

shchur (approved):
LGTM!

PR #5313, opened 2025-08-28
[timeseries] factor prediction cache out of trainer

canerturkmen:
*Issue #, if available:*

*Description of changes:*

- Moves caching logic out of Trainer and abstracts it behind a Protocol.
- Add unit tests of data hashing.

### DIFF
--- a/timeseries/src/autogluon/timeseries/trainer/prediction_cache.py
+++ b/timeseries/src/autogluon/timeseries/trainer/prediction_cache.py
@@ -1,0 +1,147 @@
+import logging
+from pathlib import Path
+from typing import Any, Dict, Optional, Protocol
+
+from autogluon.common.utils.utils import hash_pandas_df
+from autogluon.core.utils.loaders import load_pkl
+from autogluon.core.utils.savers import save_pkl
+from autogluon.timeseries import TimeSeriesDataFrame
+
+logger = logging.getLogger(__name__)
+
+
+class PredictionCache(Protocol):
+    """A prediction cache is an abstract key-value store for time series predictions. The storage is keyed by
+    (data, known_covariates) pairs and stores (model_pred_dict, pred_time_dict) pair values. In this stored pair,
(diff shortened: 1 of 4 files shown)


timeseries/src/autogluon/timeseries/trainer/prediction_cache.py
```diff
@@ -0,0 +1,147 @@
...
+        return DummyPredictionCache(root_path=root_path)
+
+
+def compute_dataset_hash(data: TimeSeriesDataFrame, known_covariates: Optional[TimeSeriesDataFrame] = None) -> str:
+    """Compute a unique string that identifies the time series dataset."""
+    combined_hash = hash_pandas_df(data) + hash_pandas_df(known_covariates) + hash_pandas_df(data.static_features)
+    return combined_hash
+
+
+class DummyPredictionCache(PredictionCache):
```
shchur: Nit: Maybe something like `NoPredictionCache`, `DisabledPredictionCache` or `NoOpPredictionCache`? To me it wasn't immediately clear that dummy = disabled in this case.

timeseries/src/autogluon/timeseries/trainer/prediction_cache.py
```diff
@@ -0,0 +1,147 @@
...
+    def clear(self) -> None:
+        pass
+
+
+class FileBasedPredictionCache(PredictionCache):
+    """A file-backed cache of model predictions."""
+
+    _cached_predictions_filename = "cached_predictions.pkl"
+
+    def __init__(self, root_path: str):
```
shchur: Does it make sense to move this implementation of `__init__` to the parent class?

shchur (approved):
Nice, LGTM!

timeseries/src/autogluon/timeseries/trainer/prediction_cache.py
canerturkmen: converted to abstract class.

PR #5315, opened 2025-09-02
[timeseries] Fine-tune Chronos-Bolt on user-provided `quantile_levels`

shchur:
*Issue #, if available:*

*Description of changes:*
- Add hyperparameter `fine_tune_expand_quantiles` to Chronos-Bolt model. When set to `True`, the output layer of the model will be expanded to include the `quantile_levels` provided by the user that will be used during fine-tuning. These new quantile levels will be used during training, enabling the user to train models with arbitrary quantiles (below 0.1 and above 0.9).

*To do:*
- [x] Add tests


timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -485,6 +490,8 @@ def _fit(
                         f"\tChronosBolt models can only be fine-tuned with a maximum prediction_length of {model_prediction_length}. "
                         f"Fine-tuning prediction_length has been changed to {fine_tune_prediction_length}."
                     )
+                if model_params["fine_tune_expand_quantiles"]:
```
shchur: To discuss: Should this behavior be hidden behind a flag, or should this be the default behavior?

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -424,7 +429,7 @@ def _fit(
         from packaging import version
         from transformers.trainer import PrinterCallback, Trainer, TrainingArguments
 
-        from .pipeline import ChronosBoltPipeline, ChronosPipeline
+        from .pipeline import ChronosBoltPipeline, ChronosPipeline, patch_chronos_bolt_output_quantiles
```
shchur: Toy example: sine wave + gaussian noise, 3 model variants:
1. zero-shot
2. fine-tuned without expanding quantile levels
3. fine-tuned with expanding quantile levels

The first two have their predictions clipped at P10/P90. The last one clearly can learn the extreme quantiles (and also achieves better validation SQL).

<details> 

<summary> Code </summary>

```python
import pandas as pd
import numpy as np
import numpy as np
from scipy import stats

N = 10_000
h = 45

t = np.arange(N)
noise_scale = 0.6
period = 30
sin_wave = np.sin(2 * np.pi * t / period)
np.random.seed(123)
noise = np.random.normal(0, noise_scale, size=N)
target = sin_wave + noise

df = TimeSeriesDataFrame(
    pd.DataFrame(
        {"item_id": 0, "timestamp": pd.date_range("2020-01-01", freq="D", periods=N), "target": target}
    )
)
past = df.slice_by_timestep(None, -h)
future = df.slice_by_timestep(-h, None)

predictor = TimeSeriesPredictor(prediction_length=h, eval_metric="SQL", quantile_levels=[0.01, 0.05, 0.5, 0.95, 0.99])
predictor.fit(past, hyperparameters={
    "Chronos": [
            {"model_path": "bolt_small", "ag_args": {"name_suffix": "(zero-shot)"}},
            {"model_path": "bolt_small", "fine_tune": True, "ag_args": {"name_suffix": "(fine-tuned)"}},
            {"model_path": "bolt_small", "fine_tune": True, "fine_tune_expand_quantiles": True, "ag_args": {"name_suffix": "(fine-tuned + expanded quantiles)"}},
        ]
    },
    enable_ensemble=False,
    time_limit=90
)

models = predictor.model_names()
predictions = [predictor.predict(past, model=model) for model in models]

# Calculate theoretical quantiles
h = predictor.prediction_length
true_future_sin = np.sin(2 * np.pi * np.arange(len(past), len(past) + h) / period)
theoretical_quantiles = {}
for q in predictor.quantile_levels:
    theoretical_quantiles[str(q)] = true_future_sin + stats.norm.ppf(q, loc=0, scale=noise_scale)

history = past.iloc[-200:].reset_index()
forecast_steps = np.arange(len(history), len(history) + h)

# Create plot
fig, axes = plt.subplots(len(models), 1, figsize=(15, 5*len(models)))
colors = ['#FF0000', '#FF8C00', '#32CD32', '#8B4513', '#4169E1']


for i, (model, preds) in enumerate(zip(models, predictions)):
    ax = axes[i]

    # Plot historical and true future
    ax.plot(range(len(history)), history['target'], 'k-', label='Historical', alpha=0.8, linewidth=1)

    # Plot quantiles
    for j, q in enumerate(predictor.quantile_levels):
        q_str = str(q)
        ax.plot(forecast_steps, preds[q_str], color=colors[j], label=f'Q{q:.2f}', linewidth=1.5)
        ax.plot(forecast_steps, theoretical_quantiles[q_str], color=colors[j],
                linestyle='--', alpha=0.8, linewidth=1,
                label='True quantiles' if j == 0 else "")

    ax.set_title(f'{model} vs True Conditional Quantiles', fontsize=14)
    ax.legend(bbox_to_anchor=(1.05, 1), loc='upper left')
    ax.grid(True, alpha=0.3)

plt.tight_layout()
plt.show()

```

</details>

Image

timeseries/src/autogluon/timeseries/models/chronos/pipeline/chronos_bolt.py
```diff
@@ -528,3 +528,65 @@ def from_pretrained(cls, *args, **kwargs):
 
         model = class_.from_pretrained(*args, **kwargs)
         return cls(model=model)
+
+
+def patch_chronos_bolt_output_quantiles(model: ChronosBoltModelForForecasting, new_quantiles: list[float]) -> None:
```
shchur, line 533: To discuss: Should this move to an instance method of `ChronosBoltModelForForecasting` or `ChronosBoltPipeline`?

timeseries/src/autogluon/timeseries/models/chronos/pipeline/chronos_bolt.py
```diff
@@ -371,6 +371,66 @@ def decode(
...
+
+        self.chronos_config.quantiles = all_quantiles
+        self.num_quantiles = len(all_quantiles)
+        self.register_buffer("quantiles", torch.tensor(all_quantiles, dtype=self.dtype), persistent=False)
+
+        old_output_layer = self.output_patch_embedding
+        new_output_layer = ResidualBlock(
+            in_dim=self.config.d_model,
+            h_dim=self.config.d_ff,
+            out_dim=len(all_quantiles) * self.chronos_config.prediction_length,
```
shchur: To discuss: Should the patched model include 1/ only the user-provided `new_quantiles` or 2/ `set(old_quantiles).union(set(new_quantiles))`?

timeseries/src/autogluon/timeseries/models/chronos/model.py
canerturkmen: if quantiles are provided and don't match defaults I think we should warn and default to this.

timeseries/src/autogluon/timeseries/models/chronos/pipeline/chronos_bolt.py
canerturkmen: I would go for `new_quantiles`. The alternative leaves me with no option if I'm interested in only P50, P90 for example.

timeseries/src/autogluon/timeseries/models/chronos/pipeline/chronos_bolt.py
```diff
@@ -371,6 +371,66 @@ def decode(
 
         return decoder_outputs.last_hidden_state  # sequence_outputs, b x 1 x d_model
 
+    def expand_output_quantiles(self, new_quantiles: list[float]) -> None:
+        """In-place expands ChronosBolt model to support additional quantile levels by copying weights from closest existing quantiles."""
```
canerturkmen: Good idea to initialize it with the old weights!

timeseries/src/autogluon/timeseries/models/chronos/pipeline/chronos_bolt.py
```diff
@@ -371,6 +371,66 @@ def decode(
...
+            # Copy existing quantile weights to their new positions
+            for old_idx, old_q in enumerate(old_quantiles):
+                new_idx = all_quantiles.index(old_q)
+                copy_quantile_weights(old_idx, new_idx)
+
+            # Copy closest quantile weights for new quantiles
+            for new_q in new_quantiles:
+                if new_q not in old_quantiles:
+                    closest_q = min(old_quantiles, key=lambda x: abs(x - new_q))
+                    closest_idx = old_quantiles.index(closest_q)
```
canerturkmen: don't get why we need two loops. isn't the closest old quantile still the closest new quantile?

timeseries/src/autogluon/timeseries/models/chronos/pipeline/chronos_bolt.py
```diff
@@ -371,6 +371,66 @@ def decode(
...
+                    new_idx = all_quantiles.index(new_q)
+                    copy_quantile_weights(closest_idx, new_idx)
+
+        self.output_patch_embedding = new_output_layer
```
canerturkmen, line 420: Would be great to see a test for serde after this.

canerturkmen (approved):
Some comments, but LGTM!

timeseries/src/autogluon/timeseries/models/chronos/model.py
```diff
@@ -162,6 +162,9 @@ class ChronosModel(AbstractTimeSeriesModel):
         during fine-tuning. If None, the entire validation dataset will be used.
     fine_tune_trainer_kwargs : dict, optional
         Extra keyword arguments passed to ``transformers.TrainingArguments``
+    fine_tune_expand_quantiles : bool, default False
```
abdulfatir: Maybe a better name for this option? `new_quantiles` or `extra_quantiles`? Or maybe just `fine_tune_quantiles` and we expand internally?

timeseries/src/autogluon/timeseries/models/chronos/pipeline/chronos_bolt.py
```diff
@@ -371,6 +371,66 @@ def decode(
...
+        new_output_layer = ResidualBlock(
+            in_dim=self.config.d_model,
+            h_dim=self.config.d_ff,
+            out_dim=len(all_quantiles) * self.chronos_config.prediction_length,
+            act_fn_name=self.config.dense_act_fn,
+            dropout_p=self.config.dropout_rate,
+        )
+
+        # hidden_layer is shared across all quantiles
+        new_output_layer.hidden_layer.weight.data.copy_(old_output_layer.hidden_layer.weight.data)
```
abdulfatir, line 396: minor: Can we factor out the weight copying logic to a helper function?

timeseries/src/autogluon/timeseries/models/chronos/pipeline/chronos_bolt.py
shchur: Good point, only kept the `min(..., key=...)` option.

timeseries/src/autogluon/timeseries/models/chronos/model.py
shchur: As discussed above, changed the behavior to always replace the Chronos-Bolt quantiles with the user-specified `quantile_levels` if `fine_tune=True`, so there is no need for the `fine_tune_expand_quantiles` hyperparameter anymore.

timeseries/src/autogluon/timeseries/models/chronos/pipeline/chronos_bolt.py
shchur: It looks like this method is only called twice (for `hidden_layer.weight` and `hidden_layer.bias`), and for other layers we change the values based on indexing/slicing. Do you think we should still create a new method like the following?
```python
@staticmethod
def _copy_weights(old_weight: nn.Parameter, new_weight: nn.Parameter) -> None:
    new_weight.data.copy_(old_weight.data)
```

timeseries/src/autogluon/timeseries/models/chronos/pipeline/chronos_bolt.py
shchur: I added a test that checks that the model can predict after loading and that the quantiles are preserved in [1](test_when_chronos_bolt_fine_tuned_with_custom_quantiles_then_loaded_model_has_custom_quantiles). Is that sufficient, or did you have some other test case in mind?

PR #5321, opened 2025-09-15
[timeseries] add Toto model

canerturkmen:
*Issue #, if available:*

*Description of changes:*

Add Toto to the model zoo. Apart from very minor changes the code in `_internal/` is from the original [repository](https://github.com/DataDog/toto).

TODOS:
- [x] benchmark on fev
- [x] add docstrings
- [x] add documentation to model zoo
- [x] add missing value handling via forward filling
- [x] ~~add to smoketests?~~ did not add to smoketests as timeseries tests in HF_HUB_OFFLINE mode. this means the model has to be downloaded and cached beforehand, which works differently for Toto (it isn't compatible with AutoModel).

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/toto/dataloader.py
+++ b/timeseries/src/autogluon/timeseries/models/toto/dataloader.py
@@ -1,8 +1,8 @@
 import time
-from typing import Any, Callable, Iterator, Optional, Union
+from typing import Any, Callable, Iterator, Literal, Optional, Union
 
 import numpy as np
-import torch.utils.data
+import torch
 
 from autogluon.core.utils.exceptions import TimeLimitExceeded
 from autogluon.timeseries import TimeSeriesDataFrame
@@ -54,6 +54,7 @@ class TotoDataLoader:
         batch_size: int = 1,
         time_limit: Optional[Union[int, float]] = None,
         device: Any = None,
+        missing_value_handling: Literal["ffill", None] = "ffill",
     ):
         self.batch_loader = torch.utils.data.DataLoader(
             dataset=dataset,
@@ -63,6 +64,7 @@ class TotoDataLoader:
         self.device = torch.device(device)
 
         self.freq: str = freq or dataset.freq or "h"
+        self.missing_value_handling: Optional[str] = missing_value_handling
 
     @staticmethod
     def _get_timeout_callback(seconds: Optional[float]) -> Callable:
@@ -74,9 +76,19 @@ class TotoDataLoader:
 
         return callback
 
+    @staticmethod
+    def _ffill(tensor):
+        assert tensor.ndim > 1
+        nan_mask = torch.isnan(tensor)
+        indices = torch.where(nan_mask, 0, torch.arange(tensor.shape[-1], device=tensor.device).expand_as(tensor))
+        last_valid = torch.cummax(indices, dim=-1).values
+        return torch.gather(tensor, dim=-1, index=last_valid)
+
     def __iter__(self) -> Iterator[MaskedTimeseries]:
+        fill_fn = self._ffill if self.missing_value_handling == "ffill" else lambda x: x
+
         for batch in self.batch_loader:
-            time_series = batch.unsqueeze(1).to(self.device)
+            time_series = fill_fn(batch.unsqueeze(1).to(self.device))
             current_batch_size = batch.shape[0]
 
             mts = MaskedTimeseries(
--- a/timeseries/src/autogluon/timeseries/models/toto/dataloader.py
+++ b/timeseries/src/autogluon/timeseries/models/toto/dataloader.py
@@ -1,3 +1,4 @@
+import functools
 import time
 from typing import Any, Callable, Iterator, Optional, Union
 
@@ -14,11 +15,11 @@ class TotoInferenceDataset(torch.utils.data.Dataset):
     def __init__(
         self,
         target_df: TimeSeriesDataFrame,
-        context_length: int,
+        max_context_length: int,
         target_column: str = "target",
     ):
-        assert context_length > 0
-        self.context_length = context_length
+        assert max_context_length > 0
+        self.max_context_length = max_context_length
         self.target_array = target_df[target_column].to_numpy(dtype=np.float32)
 
         # store pointer to start:end of each time series
@@ -29,19 +30,14 @@ class TotoInferenceDataset(torch.utils.data.Dataset):
     def __len__(self):
         return len(self.indptr) - 1  # noqa
 
-    def _get_context(self, a: np.ndarray, pad_value=np.nan):
-        a = a[-self.context_length :]
-        pad_size = self.context_length - len(a)
-        if pad_size > 0:
-            pad = np.full(shape=(pad_size,), fill_value=pad_value)
-            a = np.concatenate((pad, a))
-        return a
-
     def __getitem__(self, idx) -> np.ndarray:
         start_idx = self.indptr[idx]
         end_idx = self.indptr[idx + 1]
 
-        return self._get_context(self.target_array[start_idx:end_idx])
+        if end_idx - start_idx > self.max_context_length:
+            start_idx = end_idx - self.max_context_length
+
+        return self.target_array[start_idx:end_idx]
 
 
 class TotoDataLoader:
@@ -53,12 +49,13 @@ class TotoDataLoader:
         time_limit: Optional[Union[int, float]] = None,
         device: Any = None,
     ):
+        self.device = torch.device(device)
         self.batch_loader = torch.utils.data.DataLoader(
             dataset=dataset,
             batch_size=batch_size,
+            collate_fn=functools.partial(self._collate, device=self.device),
         )
         self.on_batch = self._get_timeout_callback(time_limit) if time_limit is not None else (lambda *a, **k: None)
-        self.device = torch.device(device)
 
         self.freq: str = freq or dataset.freq or "h"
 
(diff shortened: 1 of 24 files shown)


timeseries/src/autogluon/timeseries/models/toto/dataloader.py
```diff
@@ -0,0 +1,116 @@
...
+        self,
+        target_df: TimeSeriesDataFrame,
+        context_length: int,
+        target_column: str = "target",
+    ):
+        assert context_length > 0
+        self.context_length = context_length
+        self.target_array = target_df[target_column].to_numpy(dtype=np.float32)
+
+        target_df = target_df.sort_index()
```
shchur: We assume that the index is always sorted inside the predictor, so this in not necessary (lots of other operations depend on this as well). If you still prefer keeping this, I would at least hide this inside
```python
if not target_df.index.is_monotonic_increasing:
    target_df = target_df.sort_index()
```

timeseries/src/autogluon/timeseries/models/toto/dataloader.py
```diff
@@ -0,0 +1,116 @@
...
+
+class TotoDataLoader:
+    def __init__(
+        self,
+        dataset: TotoInferenceDataset,
+        freq: Optional[str] = None,
+        batch_size: int = 1,
+        time_limit: Optional[Union[int, float]] = None,
+        device: Any = None,
+        missing_value_handling: Literal["ffill", None] = "ffill",
```
shchur: Why do we need the custom ffill logic here? It looks like this is never accessed outside of the dataloader.

If we really need the ffill, we could alternatively use `tsdf.fill_missing_values()` before passing the tsdf to the dataloader, which could be a bit more concise.

timeseries/src/autogluon/timeseries/models/toto/model.py
```diff
@@ -0,0 +1,232 @@
...
+class TotoModel(AbstractTimeSeriesModel):
+    """Toto (Time-Series-Optimized Transformer for Observability) [CohenKhwajaetal2025]_ pretrained time series forecasting model.
+
+    Toto is a 151M parameter model trained on over 1T data points from DataDog's internal observability systems, as well as
+    the GIFT-eval pretrain, Chronos pretraining, and synthetically generated time series corpora. It is a decoder-only
+    architecture that autoregressively outputs parametric distribution forecasts. More details can be found on
+    `Hugging Face <https://huggingface.co/Datadog/Toto-Open-Base-1.0>`_ and `GitHub <https://github.com/DataDog/toto>.
+
+    The AutoGluon implementation of Toto is on a port of the original implementation. It is optimized for easy maintenance
+    with the rest of the AutoGluon model zoo, and does not feature some important optimizations such as xformers and flash-attention
```
shchur, line 29: Did you happen to measure how much this affects the inference speed?

timeseries/src/autogluon/timeseries/models/toto/model.py
```diff
@@ -0,0 +1,232 @@
...
+    ):
+        hyperparameters = hyperparameters if hyperparameters is not None else {}
+
+        self.model_path = hyperparameters.get("model_path", self.default_model_path)
+
+        name = name or "Toto"
+
+        if not isinstance(self.model_path, Space):
+            # we truncate the name to avoid long path errors on Windows
+            model_path_suffix = "[" + str(self.model_path).replace("/", "__").replace(os.path.sep, "__")[-50:] + "]"
```
shchur: What do you think about just keeping just `Toto` as the name and tracking the `model_path` in model hyperparameters? Especially for Toto there is no way to finetune, so I don't think there will be too much ambiguity as to which model weights are loaded.

Looking back, I am not a huge fan of this dynamic name modification that we introduced for Chronos. It has the weird side-effect of breaking the mapping between the keys in the `hyperparameters` dictionary and the resulting names of trained models in `leaderboard`. If we do this for pretrained models, I think it would be even more justified for Recursive/DirectTabular models, where the switch from LightGBM to CatBoost as the underlying model has arguably an even bigger effect.

timeseries/src/autogluon/timeseries/models/toto/dataloader.py
canerturkmen: cool, thanks!

timeseries/src/autogluon/timeseries/models/toto/model.py
canerturkmen: I didn't measure myself, but my intuition is that with an effective sequence length of ~64 (4096 / 64) this should be modest, not enough to justify the dependency.

timeseries/src/autogluon/timeseries/models/toto/model.py
canerturkmen: 👍

timeseries/src/autogluon/timeseries/models/toto/dataloader.py
```diff
@@ -0,0 +1,114 @@
...
+    def _get_timeout_callback(seconds: Optional[float]) -> Callable:
+        start_time = time.monotonic()
+
+        def callback() -> None:
+            if seconds is not None and time.monotonic() - start_time > seconds:
+                raise TimeLimitExceeded
+
+        return callback
+
+    @staticmethod
```
shchur, line 72: I think this method could be replaced with [torch.nn.utils.rnn.pad_sequence](https://docs.pytorch.org/docs/stable/generated/torch.nn.utils.rnn.pad_sequence.html#torch-nn-utils-rnn-pad-sequence)

timeseries/src/autogluon/timeseries/models/toto/model.py
```diff
@@ -0,0 +1,234 @@
...
+
+        hyperparameters = self.get_hyperparameters()
+        pretrained_model = TotoPretrainedModel.from_pretrained(
+            self.model_path,
+            config=TotoConfig.from_pretrained(self.model_path),
+            device_map=hyperparameters["device"],
+        )
+
+        if hyperparameters["compile_model"]:
+            pretrained_model.model.compile()
```
shchur, line 133: I'm a bit confused here, will we compile the model each time we call `self.load_forecaster()`?

timeseries/src/autogluon/timeseries/models/toto/model.py
```diff
@@ -0,0 +1,234 @@
...
+        "This Time is Different: An Observability Perspective on Time Series Foundation Models."
+        https://arxiv.org/abs/2505.14766
+
+
+    Other Parameters
+    ----------------
+    model_path : str, default = "Datadog/Toto-Open-Base-1.0"
+        Model path used for the model, i.e., a HuggingFace transformers ``name_or_path``. Can be a
+        compatible model name on HuggingFace Hub or a local path to a model directory.
+    batch_size : int, default = 16
```
shchur: Doesn't match the default in `_default_hyperparameters`

timeseries/src/autogluon/timeseries/models/toto/model.py
```diff
@@ -0,0 +1,234 @@
...
+        Size of batches used during inference.
+    num_samples : int, default = 256
+        Number of samples used during inference.
+    device : str, default = "cuda"
+        Device to use for inference. Toto requires a CUDA-compatible GPU to run.
+    context_length : int or None, default = 4096
+        The context length to use in the model. Shorter context lengths will decrease model accuracy, but result
+        in faster inference.
```
shchur, line 52: The way I see it, assuming that the dataset has sufficiently many items / historical observations, `context_length * batch_size` is limited by the available GPU memory. We can decrease the default `context_length` to potentially hurt the accuracy at the cost of increasing the inference speed. My intuition tells me that setting the default `context_length` at 1024 or 2048 while increasing the default batch size by 4x or 2x would be a reasonable tradeoff, given that Toto has a pretty bad inference speed by default. We could also set the default `num_samples` to 128 to further boost the inference speed.

What do you think?

timeseries/src/autogluon/timeseries/models/toto/model.py
```diff
@@ -0,0 +1,234 @@
...
+logger = logging.getLogger(__name__)
+
+
+class TotoModel(AbstractTimeSeriesModel):
+    """Toto (Time-Series-Optimized Transformer for Observability) [CohenKhwajaetal2025]_ pretrained time series forecasting model.
+
+    Toto is a 151M parameter model trained on over 1T data points from DataDog's internal observability systems, as well as
+    the GIFT-eval pretrain, Chronos pretraining, and synthetically generated time series corpora. It is a decoder-only
+    architecture that autoregressively outputs parametric distribution forecasts. More details can be found on
+    `Hugging Face <https://huggingface.co/Datadog/Toto-Open-Base-1.0>`_ and `GitHub <https://github.com/DataDog/toto>.
```
shchur: Should be as follows for the hyperlink to render correctly
```
`GitHub <https://github.com/DataDog/toto>`_
```

timeseries/src/autogluon/timeseries/models/toto/model.py
```diff
@@ -0,0 +1,234 @@
...
+    ----------------
+    model_path : str, default = "Datadog/Toto-Open-Base-1.0"
+        Model path used for the model, i.e., a HuggingFace transformers ``name_or_path``. Can be a
+        compatible model name on HuggingFace Hub or a local path to a model directory.
+    batch_size : int, default = 16
+        Size of batches used during inference.
+    num_samples : int, default = 256
+        Number of samples used during inference.
+    device : str, default = "cuda"
+        Device to use for inference. Toto requires a CUDA-compatible GPU to run.
```
shchur, line 49: If `"cuda"` is the only accepted value here, should we even expose it?

timeseries/src/autogluon/timeseries/models/toto/dataloader.py
canerturkmen: cool. could only replace the implementation as the `_collate` function also handled casting to Tensor with device.

timeseries/src/autogluon/timeseries/models/toto/model.py
canerturkmen: now I'm confused too. isn't it by design when one loads `from_pretrained` that they have to compile again?

timeseries/src/autogluon/timeseries/models/toto/model.py
canerturkmen: thanks!!

timeseries/src/autogluon/timeseries/models/toto/model.py
canerturkmen: Let's leave the defaults as in the original and punt this to our exploration for presets?

timeseries/src/autogluon/timeseries/models/toto/model.py
canerturkmen: thanks!

timeseries/src/autogluon/timeseries/models/toto/model.py
canerturkmen: so that you can provide `"cuda:0"` etc. as well maybe?

PR #5325, opened 2025-09-17
[tabular] Add logs for LightGBM CUDA device

Innixma:
*Issue #, if available:*

*Description of changes:*

- Add logs for LightGBM CUDA device
- Keeping "gpu" as the default device for GPUs since "cuda" is not yet tested and is not included in the default lightgbm install on pip. Ref: https://github.com/microsoft/LightGBM/issues/6828

### DIFF
--- a/tabular/src/autogluon/tabular/models/lgb/lgb_model.py
+++ b/tabular/src/autogluon/tabular/models/lgb/lgb_model.py
@@ -305,16 +305,27 @@ class LGBModel(AbstractModel):
             try:
                 self.model = train_lgb_model(early_stopping_callback_kwargs=early_stopping_callback_kwargs, **train_params)
             except LightGBMError:
-                if train_params["params"].get("device", "cpu") != "gpu":
+                if train_params["params"].get("device", "cpu") not in ["gpu", "cuda"]:
                     raise
                 else:
-                    logger.warning(
-                        "Warning: GPU mode might not be installed for LightGBM, GPU training raised an exception. Falling back to CPU training..."
-                        "Refer to LightGBM GPU documentation: https://github.com/Microsoft/LightGBM/tree/master/python-package#build-gpu-version"
-                        "One possible method is:"
-                        "\tpip uninstall lightgbm -y"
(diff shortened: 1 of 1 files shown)


tabular/src/autogluon/tabular/models/lgb/lgb_model.py
```diff
@@ -305,16 +305,27 @@ def _fit(self, X, y, X_val=None, y_val=None, time_limit=None, num_gpus=0, num_cp
...
+                        logger.warning(
+                            "Warning: GPU mode might not be installed for LightGBM, "
+                            "GPU training raised an exception. Falling back to CPU training..."
+                            "Refer to LightGBM GPU documentation: "
+                            "https://github.com/Microsoft/LightGBM/tree/master/python-package#build-gpu-version"
+                            "One possible method is:"
+                            "\tpip uninstall lightgbm -y"
+                            "\tpip install lightgbm --install-option=--gpu"
+                        )
+                    elif train_params["params"]["device"] == "cuda":
```
ncclementi, line 321: Just a comment here, if autogluon, and therefore LightGBM are installed via conda, `device="cuda"` works just fine.

tabular/src/autogluon/tabular/models/lgb/lgb_model.py
```diff
@@ -305,16 +305,27 @@ def _fit(self, X, y, X_val=None, y_val=None, time_limit=None, num_gpus=0, num_cp
...
+                        # Current blocker for using CUDA over GPU: https://github.com/microsoft/LightGBM/issues/6828
```
Innixma, line 322:
```suggestion
                        # Current blocker for using CUDA over GPU: https://github.com/microsoft/LightGBM/issues/6828
                        # Note that device="cuda" works if AutoGluon (and therefore LightGBM) is installed via conda.
```

tabular/src/autogluon/tabular/models/lgb/lgb_model.py
Innixma: Good point, added to inline comment

PR #5328, opened 2025-09-22
[install] Use standalone uv in full_install.sh

Innixma:
*Issue #, if available:*

*Description of changes:*

If the user doesn't have `pip` installed, the current `full_install.sh` source install script will fail even if the user has `uv` installed standalone:

```
$ ./autogluon/full_install.sh
uv could not be found. Installing uv...
/usr/bin/python: No module named pip
```

This PR fixes this by checking if a standalone uv install exists, and if so it will use it without requiring `pip` to be installed.

[Colab example](https://colab.research.google.com/drive/124DXEjBffUCDA9L7L6oq_Tfx31GoiTkp?usp=sharing) of this PR's install script: 


ncclementi:
@Innixma Thanks for putting this up. I tested this install (using the non-pip install of uv), and I run a small example to make sure things run and things run normally.

Innixma:
Thanks @ncclementi! Merging

PR #5330, opened 2025-09-26
Add multilabel support to multimodal package. Initial framework setup…

hubenjm:
… for handling ignore label values with custom BCE loss function, but further updates required.

*Issue #, if available:*

*Description of changes:*
* Overloads `label` input to be either string column name (current behavior) or list of string column names (new multilabel behavior)
* Adds `MULTILABEL` problem type currently with just `accuracy` metrics. Modifies backend lightning module to use torchmetrics `Accuracy(task = 'multilabel', ...)` metric when problem type is multilabel
* Validation check to require only two distinct values in each column
* Add `ignore_label_value` to default config, to be fully supported in future update
* Add custom BCE loss to multimodal/src/autogluon/multimodal/optim/losses/bce_loss.py to handle ignore labels, to be fully supported in future update

### DIFF
--- a/multimodal/src/autogluon/multimodal/configs/data/default.yaml
+++ b/multimodal/src/autogluon/multimodal/configs/data/default.yaml
@@ -17,6 +17,7 @@ data:
   label:
     numerical_preprocessing: "standardscaler"  # The mode of numerical label preprocessing for . Support "standardscaler" or "minmaxscaler" or None (means no transform).
   pos_label:  # The name of binary classification's positive class. It's used in computing some metrics, e.g., roc_auc. If not provided, then use label_encoder.classes_[1],
+  ignore_label:  # The name of multi-label classification ignore class, indicating when label value is unknown. Can be provided as single scalar value for all label columns or as a map from column name to its respective ignore label value
   column_features_pooling_mode: "concat"  # How to pool multi-column features into one feature vector. Currently only support "concat" or "mean" for few shot classification.
   mixup:
     turn_on: False  # The total control of mixup.
--- a/multimodal/src/autogluon/multimodal/constants.py
+++ b/multimodal/src/autogluon/multimodal/constants.py
@@ -25,6 +25,7 @@ ZERO_SHOT = "zero_shot"
 CLASSIFICATION = "classification"
 BINARY = "binary"
 MULTICLASS = "multiclass"
+MULTILABEL = "multilabel"
 REGRESSION = "regression"
 NER = "ner"
 NAMED_ENTITY_RECOGNITION = "named_entity_recognition"
@@ -190,6 +191,7 @@ EVALUATION_METRICS = {
     # Use evaluation metrics from METRICS for these types
     BINARY: list(METRICS[BINARY].keys()) + [COVERAGE],
     MULTICLASS: METRICS[MULTICLASS].keys(),
+    MULTILABEL: ["accuracy", "acc"], # TO DO: Add additional metrics if relevant
     REGRESSION: METRICS[REGRESSION].keys(),
     OBJECT_DETECTION: DETECTION_METRICS,
     SEMANTIC_SEGMENTATION: [IOU, BER, SM],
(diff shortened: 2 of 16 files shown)


Innixma:
Thanks for submitting the PR @hubenjm! We will try to take a look and see if this can be integrated, but note that since it is a rather complex addition to the codebase and we aren't heavily developing the multimodal module at present, this may take some time to process on our end.

@FANGAreNotGnu some things we would need to do to add this:

1. Add unit tests
2. Add tutorial
3. Possibly do a sanity check on a few real multilabel datasets

Innixma:
@zhiqiangdon happy to hear your thoughts on this

Innixma:
Hi @hubenjm, would you be able to provide a simple script that uses a public (or synthetic) dataset to demonstrate the multilabel functionality? This can be used as a point of reference in our review

hubenjm:
Hi guys,

Thank you for your comments so far on this. And sorry it took me a bit of time to get back to you. I think I will work on adding some unit tests and other things to address those concerns. Part of my challenge so far was that when I tried to run existing multimodal tests, the interpreter just hangs and doesn't do anything, so maybe something is wrong with my local environment.

hubenjm:
> Hi @hubenjm, would you be able to provide a simple script that uses a public (or synthetic) dataset to demonstrate the multilabel functionality? This can be used as a point of reference in our review

Yeah I will look into doing this. I see some existing tests use the pets dataset but I am not sure that would be suitable for a multi label example. So I will look around for something simple that could work.

FANGAreNotGnu:
Thanks for submitting this PR @hubenjm ! I'll review it in detail soon. One initial observation: some multilabel datasets are more complex than the multiple binary classification supported here. For instance, the Memotion dataset (https://www.kaggle.com/datasets/williamscott701/memotion-dataset-7k) contains both binary and multiclass label columns simultaneously. It would be great if we could explore a more general design to accommodate these mixed-label-type scenarios.

hubenjm:
> Thanks for submitting this PR @hubenjm ! I'll review it in detail soon. One initial observation: some multilabel datasets are more complex than the multiple binary classification supported here. For instance, the Memotion dataset (https://www.kaggle.com/datasets/williamscott701/memotion-dataset-7k) contains both binary and multiclass label columns simultaneously. It would be great if we could explore a more general design to accommodate these mixed-label-type scenarios.

This is a good observation and I did briefly consider something similar early on. But I kind of abandoned going the full generality route to keep things from getting too complex right off the bat. But now with a working preliminary implementation of the binary multilabel case, it is a bit clearer to me the marginal additional complexity to extend this to the full mixed-label-type case. In particular, this is going to require at least:
1. Having some internal representation for the mixed-label signature, likely as a map from label column name to either 'binary' or 'multiclass' value. The `ignore_label` parameter that is optionally specified by user in config params should be applicable to either of these cases.
2. Custom loss function that handles generic mixed-label type scenario. Also need to update the utils function that applies activation function (softmax or sigmoid) to logits. It becomes more complex than before now since we need to iterate over column_name -> label_type map and apply sigmoid or softmax activation per group independently.
3. The Torchmetrics metric used for training now cannot be handled by built in `Accuracy` class. Have to build some composite class or list of Accuracy metric classes (one per label column_name) and have logic in Lightning Module to compute each. Then the question is, how do we reduce these multiple metrics into a single value that is used for the purposes of determining what model is best? Weighted average of some kind, and if so what would the weights be?
4. Figure out what the various evaluation metrics are going to be. Similar to torchmetrics issue above, might have to reconcile aggregation of multiple separate accuracy metrics into one value? Or maybe just output all separately.

Innixma:
My suggestion is we first add unit tests to verify the already implemented binary scenario, and then once we confirm this is working we can edit the code to add the more complex scenarios while checking that the already implemented binary tests continue to pass.

This might also be something that would be best as 2 separate PRs, but I will defer to @FANGAreNotGnu for his opinion.

FANGAreNotGnu:
Thanks @hubenjm! I agree we can merge this as a beta feature for multilabel support once unit tests are added. I can help with adding the tests.

Given the scope, we might want to target a later release instead of 1.5 to ensure proper validation.

Innixma:
Due to lack of development bandwidth, we won't be able to add / review major feature contributions to multimodal at this time, apologies. Closing the PR.

PR #5332, opened 2025-10-02
[timeseries] reorganize ensembles, add base class for array-based ensemble learning

canerturkmen:
*Issue #, if available:*

*Description of changes:*

- reorganize ensemble classes into weighted (legacy) and array-based.
- add abstract base class for array-based ensembles.


timeseries/src/autogluon/timeseries/models/ensemble/abstract.py
```diff
@@ -1,21 +1,49 @@
...
 from autogluon.timeseries.models.abstract import TimeSeriesModelBase
+from autogluon.timeseries.utils.features import CovariateMetadata
 
 logger = logging.getLogger(__name__)
 
 
 class AbstractTimeSeriesEnsembleModel(TimeSeriesModelBase, ABC):
     """Abstract class for time series ensemble models."""
 
+    _default_model_name = None
```
shchur: It looks like at least for the currently listed models we can stick to the default `name` based on the class name. In what cases do we need the `_default_model_name` attribute?

timeseries/src/autogluon/timeseries/models/ensemble/array_based/abstract.py
```diff
@@ -0,0 +1,209 @@
...
+        df
+            TimeSeriesDataFrame to convert to an array.
+
+        Returns
+        -------
+        array
+            of shape (items, nr_timesteps, quantiles).
+        """
+        df = df.sort_index()
+        array = df.values
```
shchur: Nit: `df.to_numpy()` is future-proof if we ever switch to pyarrow dtypes

timeseries/src/autogluon/timeseries/models/ensemble/array_based/abstract.py
```diff
@@ -0,0 +1,209 @@
...
+        nr_items = len(df.index.get_level_values(0).unique())
```
shchur: How about this?
```python
num_items = df.num_items
```
More efficient than `get_level_values` (creates new index object, can be slow) and better aligned with naming

timeseries/src/autogluon/timeseries/models/ensemble/array_based/abstract.py
```diff
@@ -0,0 +1,209 @@
...
+            nr_items,  # nr_items
+            df.shape[0] // nr_items,  # timesteps per item
+            df.shape[1],  # number of quantiles
+        )
+        return array.reshape(shape)
+
+    def _split_data_per_window(
+        self,
+        data_per_window: list[TimeSeriesDataFrame],
+    ):
```
shchur:
```suggestion
    ) -> tuple[list[TimeSeriesDataFrame], list[TimeSeriesDataFrame]]:
```

timeseries/src/autogluon/timeseries/models/ensemble/array_based/abstract.py
```diff
@@ -0,0 +1,209 @@
...
+        """Split the given `data_per_window` into ground truth for that window (fold) and the past data."""
+        past_data_per_window = [y.slice_by_timestep(None, -self.prediction_length) for y in data_per_window]
+        ground_truth_per_window = [y.slice_by_timestep(-self.prediction_length, None) for y in data_per_window]
```
shchur: Should we have some check that ensures that enough observations are available for each item? This logic could also live in the trainer when we get the data for fitting the ensemble.

timeseries/src/autogluon/timeseries/models/ensemble/array_based/abstract.py
```diff
@@ -0,0 +1,209 @@
...
+        model_scores: Optional[dict[str, float]],
+    ) -> dict[str, list[TimeSeriesDataFrame]]:
+        """Filter out failed models based on detect_and_ignore_failures setting."""
+        if not self.get_hyperparameters()["detect_and_ignore_failures"]:
+            return predictions_per_window
+
+        if model_scores is None or len(model_scores) == 0:
+            return predictions_per_window
+
+        valid_scores = {k: v for k, v in model_scores.items() if np.isfinite(v)}
```
shchur, line 221: Should we filter out models with NaN loss beforehand inside the trainer? Maybe even marking the model as failed if loss equals NaN/Inf?

Currently we check that none of the predictions are equal to NaN inside the `TimeSeriesScorer`. This means the loss can only be NaN / Inf in degenerate cases like WAPE or MAPE on all-zero targets, where all models will get a bad score. I think predictor should fail earlier if that happens.

timeseries/src/autogluon/timeseries/models/ensemble/array_based/abstract.py
```diff
@@ -0,0 +1,209 @@
...
+        self,
+        predictions_per_window: dict[str, list[TimeSeriesDataFrame]],
+        data_per_window: list[TimeSeriesDataFrame],
+        model_scores: Optional[dict[str, float]] = None,
+        time_limit: Optional[float] = None,
+    ) -> None:
+        ground_truth_per_window, _ = self._split_data_per_window(data_per_window=data_per_window)
+        labels = np.stack(
+            [self.to_array(gt) for gt in ground_truth_per_window], axis=0
+        )  # (nr_windows, items, prediction_length, quantiles)
```
shchur: Nit: Shape of labels is likely just `[num_windows, num_items, prediction_length]` (without quantiles)

timeseries/src/autogluon/timeseries/models/ensemble/array_based/abstract.py
```diff
@@ -0,0 +1,209 @@
...
+
+        filtered_predictions = self._filter_failed_models(predictions_per_window, model_scores)
+
+        self._model_names = list(filtered_predictions.keys())
+
+        self.ensemble_regressor = self._regressor_type(**self.get_hyperparameters())
+        self.ensemble_regressor.fit(
+            base_model_predictions=self._get_base_model_predictions_array(filtered_predictions),
```
shchur: Original model predictions contain columns `["mean"] + [str(q) for q in quantile_levels]`. It looks like we are not explicitly treating the `mean` column right now, should we do that?

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor.py
```diff
@@ -0,0 +1,66 @@
...
+        """
+        Parameters
+        ----------
+        base_model_predictions
+            Predictions of base models. Array of shape
+            (windows, items, prediction_length, quantiles, model)
+
+        labels
+            Ground truth array of shape
+            (windows, items, prediction_length)
```
shchur: More of nit: Should we use `num_windows, num_items, prediction_length, num_quantiles` instead?

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor.py
```diff
@@ -0,0 +1,66 @@
...
+            (1, items, prediction_length, quantiles, model)
+
+        Returns
+        -------
+        ensemble_predictions
+            Array of shape (1, items, prediction_length, quantiles)
+        """
+        pass
+
+    def predict_cold(self, base_model_predictions: np.ndarray) -> np.ndarray:
```
shchur: Maybe let's postpone the addition of this method until we have some ensemble method that relies on the specific set of items?

As far as I understand, `predict` currently assumed that the ordering and set of items exactly matches those in the training data, but this is not always the case. Even if we call
```python
predictor.fit(train_data)
predictor.predict(train_data)
```
some items might be dropped before they reach the ensemble (e.g., if too few observations are available for training).

I think it will be easier to discuss the specific API / checks that need to be in place once we have some models that are item-specific.

timeseries/src/autogluon/timeseries/models/ensemble/array_based/models.py
```diff
@@ -0,0 +1,7 @@
+from .abstract import ArrayBasedTimeSeriesEnsembleModel
+from .regressor import SimpleAverageEnsembleRegressor
+
+
+class ArrayBasedSimpleAverageEnsemble(ArrayBasedTimeSeriesEnsembleModel):
+    _default_model_name = "ArrayBasedSimpleAverageEnsemble"
+    _regressor_type = SimpleAverageEnsembleRegressor
```
shchur: Nit: Since we already have `SimpleAverageEnsemble`, maybe we can use `SimpleMedianEnsemble` as a very simple representative of the array-based ensembles?

shchur (commented):
Thanks @canerturkmen! Left a few comments about the handling of edge cases

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor.py
canerturkmen: I'll remove `predict_cold` for now and move this intention to `predict` for now, however it's not easy to enforce globality at the API level.

timeseries/src/autogluon/timeseries/models/ensemble/array_based/models.py
canerturkmen: nice!

timeseries/src/autogluon/timeseries/models/ensemble/array_based/abstract.py
canerturkmen: added a todo 👍

timeseries/src/autogluon/timeseries/models/ensemble/array_based/abstract.py
canerturkmen: Let me leave this in for now. I agree with what you point out but I can't make out if this will make things easier or harder when the same class is being used in L3. Also, I'm not sure how much earlier we can make this anyway?

timeseries/src/autogluon/timeseries/models/ensemble/array_based/abstract.py
canerturkmen: Parallel to the other comment, I already think Trainer is too big of a class right now. At least in this case, I think the ensemble should be responsible for curating its own data and making a workable model much like our other model classes. 

For "good items" I think the only good items are those that have "prediction_length" time steps. I am not sure if one can provide data lacking this through the Predictor, but I agree we should have the check here anyway.

timeseries/src/autogluon/timeseries/models/ensemble/array_based/abstract.py
canerturkmen: Also note here "failed model" refers to models that have 10*median in loss.

timeseries/src/autogluon/timeseries/models/ensemble/array_based/abstract.py
```diff
@@ -0,0 +1,236 @@
...
+
+    def _get_default_hyperparameters(self) -> dict[str, Any]:
+        return {
+            "isotonization": "sort",
+            "detect_and_ignore_failures": True,
+        }
+
+    @staticmethod
+    def to_array(df: TimeSeriesDataFrame) -> np.ndarray:
+        """Given a TimeSeriesDataFrame object, or a list or dict of such objects, return
```
shchur: It looks like `a list or dict of such objects` does not apply anymore?

timeseries/src/autogluon/timeseries/models/ensemble/array_based/abstract.py
```diff
@@ -0,0 +1,236 @@
...
+
+        Parameters
+        ----------
+        df
+            TimeSeriesDataFrame to convert to an array.
+
+        Returns
+        -------
+        array
+            of shape (num_items, num_timesteps, num_quantiles).
```
shchur: How about we explicitly replace `num_timesteps` with `prediction_length` here? As I understand, there is an implicit assumption that `df` is a TSDF of ground truth values for the forecast horizon (or the predictions), where each item has `prediction_length` values (otherwise reshaping to an array won't work).

timeseries/src/autogluon/timeseries/models/ensemble/array_based/abstract.py
```diff
@@ -0,0 +1,236 @@
...
+    ) -> None:
+        # get the last prediction_length steps from each item
+        ground_truth_per_window = [y.slice_by_timestep(-self.prediction_length, None) for y in data_per_window]
+
+        # drop items which have less than prediction_length steps as ground truth
+        item_ids_per_window = []
+        for i, ground_truth in enumerate(ground_truth_per_window):
+            item_ids = ground_truth.item_ids[ground_truth.groupby(level="item_id").size() == self.prediction_length]
+            item_ids_per_window.append(item_ids)
+            ground_truth_per_window[i] = ground_truth.loc[item_ids]
```
shchur: We can safely assume that the items in `data_per_window` that reach `AbstractTimeSeriesEnsembleModel.fit` have at least `prediction_length` observations for each window. This assumption in currently enforced [inside the predictor](https://github.com/autogluon/autogluon/blob/c4382621dadfb6c7cf69f9bb6b696f0ffb610497/timeseries/src/autogluon/timeseries/predictor.py#L750C31-L750C41) and I think that many other parts of the code would break if this was not the case.

timeseries/src/autogluon/timeseries/models/ensemble/array_based/abstract.py
```diff
@@ -0,0 +1,236 @@
...
+        ----------
+        df
+            TimeSeriesDataFrame to convert to an array.
+
+        Returns
+        -------
+        array
+            of shape (num_items, num_timesteps, num_quantiles).
+        """
+        df = df.sort_index()
```
shchur: We can safely skip this check since it's currently guaranteed that all data that reaches learner + goes deeper is sorted by index due to https://github.com/autogluon/autogluon/blob/c4382621dadfb6c7cf69f9bb6b696f0ffb610497/timeseries/src/autogluon/timeseries/predictor.py#L275C9-L275C38. I understand that making assumptions is bad in general, but in this case it should be fine. 

If we really want to keep a check, maybe `assert df.index.is_monotonic_increasing` is sufficient?

timeseries/src/autogluon/timeseries/models/ensemble/array_based/abstract.py
```diff
@@ -0,0 +1,236 @@
...
+        array = df.to_numpy()
+        num_items = df.num_items
+        shape = (
+            num_items,
+            df.shape[0] // num_items,  # timesteps per item
+            df.shape[1],  # num_quantiles
```
shchur: We sometimes call `.to_array` on the TSDF of ground truth future values where there is no concept of `num_quantiles`. Should we change this to `len(df.columns)` or `num_outputs` here (and also in the method signature)?

timeseries/src/autogluon/timeseries/models/ensemble/array_based/abstract.py
```diff
@@ -0,0 +1,236 @@
...
+
+    @staticmethod
+    def to_array(df: TimeSeriesDataFrame) -> np.ndarray:
+        """Given a TimeSeriesDataFrame object, or a list or dict of such objects, return
+        a single array composing the values contained in the data frames.
+
+        Parameters
+        ----------
+        df
+            TimeSeriesDataFrame to convert to an array.
```
shchur: Let's explicitly state here that `df` must contain exactly `prediction_length` values for each item for this method to work, and that the `df` can correspond to ground truth values or predictions.

timeseries/src/autogluon/timeseries/models/ensemble/array_based/abstract.py
```diff
@@ -0,0 +1,236 @@
...
+        assert prediction_array.shape[0] == 1
+
+        prediction_array = self._isotonize(prediction_array)
+
+        output = list(input_data.values())[0].copy()
+        num_folds, num_items, num_timesteps, num_outputs = prediction_array.shape
+        assert (num_folds, num_timesteps) == (1, self.prediction_length)
+        assert len(output.columns) == num_outputs
+
+        output[output.columns] = prediction_array.reshape((num_items * num_timesteps, -1))
```
shchur: Nit: explicitly reshape to `(num_items * num_timesteps, num_outputs)`?

timeseries/src/autogluon/timeseries/models/ensemble/array_based/abstract.py
shchur: All right, I guess getting `inf` loss elsewhere is not a common issue since some users would have reported it.

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor.py
```diff
@@ -0,0 +1,51 @@
...
+        pass
+
+    @abstractmethod
+    def fit(self, base_model_predictions: np.ndarray, labels: np.ndarray, **kwargs) -> Self:
+        """
+        Parameters
+        ----------
+        base_model_predictions
+            Predictions of base models. Array of shape
+            (num_windows, num_items, prediction_length, num_quantiles, num_models)
```
shchur: Let's consistently replace `num_quantiles` with `num_outputs` in all methods?

timeseries/src/autogluon/timeseries/models/ensemble/array_based/abstract.py
```diff
@@ -0,0 +1,224 @@
...
+            Array of shape (num_windows, num_items, prediction_length, num_outputs)
+
+        Returns
+        -------
+        isotonized_array
+            Array with same shape but quantiles sorted along last dimension
+        """
+        isotonization = self.get_hyperparameters()["isotonization"]
+        if isotonization == "sort":
+            return np.sort(prediction_array, axis=-1)
```
shchur, line 143: I just now realized that this will put the `mean` column in the incorrect location in the output dataframe. The correct solution is to sort only the columns `predictions_array[..., 1:]` since we assume that `"mean"` is always the first column.

timeseries/tests/unittests/models/ensemble/test_abstract.py
```diff
@@ -0,0 +1,71 @@
...
+
+    @pytest.fixture()
+    def ensemble_data(self):
+        yield {
+            "predictions_per_window": {
+                "dummy_model": [PREDICTIONS_FOR_DUMMY_TS_DATAFRAME],
+                "dummy_model_2": [PREDICTIONS_FOR_DUMMY_TS_DATAFRAME],
+            },
+            "data_per_window": [DUMMY_TS_DATAFRAME],
+            "model_scores": {"dummy_model": 25.0, "dummy_model_2": 15.0},
```
shchur: Nit: scores are always negative for time series, let's keep these negative too in case this becomes relevant in the future?
```suggestion
            "model_scores": {"dummy_model": -25.0, "dummy_model_2": -15.0},
```

PR #5335, opened 2025-10-07
Display GPU info in System Info

prateekdesai04:
*Description of changes:*

Now on every `fit()` call we get the info below if machine has GPU/s
image
Checked for GPU usage leakage on all modules.
Yet to test on CPU only instance.

### DIFF
--- a/common/src/autogluon/common/utils/system_info.py
+++ b/common/src/autogluon/common/utils/system_info.py
@@ -99,7 +99,6 @@ def get_ag_system_info(*, path: str = None, include_gpu_count=False, include_pyt
                 gpu_memory_info.append(f"GPU {i}: {free_memory_gb:.2f}/{total_memory_gb:.2f} GB")
 
             gpu_memory_str = " | ".join(gpu_memory_info)
-            msg_list.append(f"GPU Count:          {system_num_gpus}")
             msg_list.append(f"GPU Memory:         {gpu_memory_str}")
             msg_list.append(
                 f"Total GPU Memory:   Free: {combined_free_memory:.2f} GB, Allocated: {total_allocated_memory:.2f} GB, Total: {combined_gpu_memory:.2f} GB"
@@ -107,6 +106,7 @@ def get_ag_system_info(*, path: str = None, include_gpu_count=False, include_pyt
 
         except Exception as e:
             system_num_gpus = f"WARNING: Exception was raised when calculating GPU count ({e.__class__.__name__})"
+        msg_list.append(f"GPU Count:          {system_num_gpus}")
 
     msg_list.append(
         f"Memory Avail:       {available_mem:.2f} GB / {total_mem:.2f} GB ({mem_avail_percent * 100:.1f}%)"


common/src/autogluon/common/utils/system_info.py
```diff
@@ -79,6 +78,28 @@ def get_ag_system_info(*, path: str = None, include_gpu_count=False, include_pyt
     if include_gpu_count:
         try:
             system_num_gpus = resource_manager.get_gpu_count_torch()
+            gpu_memory_info = []
```
Innixma, line 84: Need to add `import torch` to the try block. It isn't guaranteed to be imported yet.

common/src/autogluon/common/utils/system_info.py
```diff
@@ -79,6 +78,28 @@ def get_ag_system_info(*, path: str = None, include_gpu_count=False, include_pyt
...
+            for i in range(system_num_gpus):
+                total_memory_gpu = torch.cuda.get_device_properties(i).total_memory
+                total_memory_gb = total_memory_gpu / (1024**3)  # Convert bytes to GB
+                allocated_memory = torch.cuda.memory_allocated(i)
+                allocated_memory_gb = allocated_memory / (1024**3)
+                free_memory_gb = total_memory_gb - allocated_memory_gb
+
+                total_free_memory += free_memory_gb
+                total_allocated_memory += allocated_memory_gb
+                total_gpu_memory += total_memory_gb
```
Innixma: rename `total` -> `combined` to be less vague (currently `total` is overloaded as it is also present in `total_memroy_gb`

common/src/autogluon/common/utils/system_info.py
```diff
@@ -79,6 +78,28 @@ def get_ag_system_info(*, path: str = None, include_gpu_count=False, include_pyt
     if include_gpu_count:
         try:
             system_num_gpus = resource_manager.get_gpu_count_torch()
+            gpu_memory_info = []
+            total_free_memory = 0
+            total_allocated_memory = 0
+            total_gpu_memory = 0
+            for i in range(system_num_gpus):
```
Innixma, line 88: Is this fast to execute? Can you test the runtime with a `time.time()` call to see how long this takes?

common/src/autogluon/common/utils/system_info.py
```diff
@@ -78,7 +78,33 @@ def get_ag_system_info(*, path: str = None, include_gpu_count=False, include_pyt
...
+                free_memory_gb = total_memory_gb - allocated_memory_gb
+
+                combined_free_memory += free_memory_gb
+                total_allocated_memory += allocated_memory_gb
+                combined_gpu_memory += total_memory_gb
+
+                gpu_memory_info.append(f"GPU {i}: {free_memory_gb:.2f}/{total_memory_gb:.2f} GB")
+
+            gpu_memory_str = " | ".join(gpu_memory_info)
+            msg_list.append(f"GPU Count:          {system_num_gpus}")
```
Innixma: GPU count is logged twice? At least it is in your PR description

prateekdesai04: fixed this

common/src/autogluon/common/utils/system_info.py
prateekdesai04: tested on CPU only and GPU instances, cannot be measured in seconds as the value too small hence insignificant

common/src/autogluon/common/utils/system_info.py
```diff
@@ -78,10 +78,35 @@ def get_ag_system_info(*, path: str = None, include_gpu_count=False, include_pyt
...
+            msg_list.append(f"GPU Memory:         {gpu_memory_str}")
+            msg_list.append(
+                f"Total GPU Memory:   Free: {combined_free_memory:.2f} GB, Allocated: {total_allocated_memory:.2f} GB, Total: {combined_gpu_memory:.2f} GB"
+            )
+
         except Exception as e:
             system_num_gpus = f"WARNING: Exception was raised when calculating GPU count ({e.__class__.__name__})"
-        msg_list.append(f"GPU Count:          {system_num_gpus}")
```
Innixma, line 84: We should log the message in the exception case. With this line removed now the exception is silent, we don't want exceptions to be silent.

Innixma (approved):
LGTM!

Issue #5341, opened 2025-10-14
Add confusion_matrix method to TabularPredictor


celestinoxp:
## Description
Add a new method to TabularPredictor to compute and optionally plot the confusion matrix for classification tasks.
Includes basic unit tests and supports normalized output, display, and saving plots.

## References
[scikit-learn ](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.confusion_matrix.html)

PR #5345, opened 2025-10-14
[tabular] add confusion_matrix method to TabularPredictor

celestinoxp:
*Closes:* #5341

*Description of changes:*
- Add confusion_matrix() method for easy visualization of classification results
- Support normalization options (true, pred, all)
- Support custom label ordering
- Option to display and/or save confusion matrix plots
- Comprehensive test suite with 8 test cases covering edge cases
- Automatic data loading from training/validation sets when data not provided


tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -5811,6 +5811,109 @@ def _assert_is_fit(self, message_suffix: str = None):
...
+                    "or ensure the predictor was trained with data persistence."
+                )
+
+            # Prefer validation data if available
+            if X_val is not None and y_val is not None:
+                data = X_val.copy()
+                data[self.label] = y_val
+            else:
+                data = X.copy()
+                data[self.label] = y
```
Innixma: In this case we are in bagged mode, and need to get the out-of-fold predictions. We don't want to predict on the train data, as that would be misleading.

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -5811,6 +5811,109 @@ def _assert_is_fit(self, message_suffix: str = None):
...
+        # Validate problem type
+        if self.problem_type not in ["binary", "multiclass"]:
+            raise ValueError(
+                f"Confusion matrix is only applicable to classification problems. "
+                f"Current problem type: '{self.problem_type}'."
+            )
+
+        # Load or prepare data
+        if data is None:
+            X, y, X_val, y_val = self._trainer.load_data()
```
Innixma: This data is already preprocessed, so we need to skip preprocessing in the `predict` call.

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -5811,6 +5811,109 @@ def _assert_is_fit(self, message_suffix: str = None):
...
+
+        # Validate problem type
+        if self.problem_type not in ["binary", "multiclass"]:
+            raise ValueError(
+                f"Confusion matrix is only applicable to classification problems. "
+                f"Current problem type: '{self.problem_type}'."
+            )
+
+        # Load or prepare data
+        if data is None:
```
Innixma: `if data is None` we need to check if the model used for prediction is a `refit_full` model. If it is, we need to get the parent model's out-of-fold or validation predictions.

I can update the code sometime in the next few weeks to do this.

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -5811,6 +5811,109 @@ def _assert_is_fit(self, message_suffix: str = None):
                 error_message = f"{error_message} `.{message_suffix}`."
             raise AssertionError(error_message)
 
+    def confusion_matrix(
+        self, data=None, normalize=None, labels=None, display=True, save_path=None, **kwargs
```
Innixma: we will want to add `model` as a parameter, as done in `predict`.

tabular/src/autogluon/tabular/predictor/predictor.py
```diff
@@ -5811,6 +5811,109 @@ def _assert_is_fit(self, message_suffix: str = None):
...
+            plt.title(f"Confusion Matrix - {self.label}")
+
+            if save_path:
+                plt.savefig(save_path, bbox_inches="tight", dpi=150)
+            if display:
+                plt.show()
+            plt.close()
+
+        if not display:
+            return cm
```
Innixma: Why only return cm if not display? I think it is better to have consistent outputs.

tabular/tests/unittests/metrics/test_confusion_matrix.py
```diff
@@ -0,0 +1,173 @@
...
+    """Test basic confusion matrix computation."""
+    train_data = TabularDataset({
+        'feature1': [1, 2, 3, 4, 5, 6, 7, 8],
+        'label':    [0, 1, 0, 1, 0, 1, 0, 1],
+    })
+
+    predictor = TabularPredictor(
+        label='label', 
+        path=tmp_path
+    ).fit(train_data, presets='medium_quality')
```
Innixma: I'll edit these tests later so it runs faster. No need to fit full medium quality, can fit a single dummy model instead.

.gitattributes
```diff
@@ -0,0 +1,3 @@
+# Normalize line endings to LF (Unix-style) across all platforms.
+# Prevents issues with mixed line endings when developers use Windows, Linux, or Mac.
+* text=auto eol=lf
```
Innixma: I looked this up since I didn't know what it was, seems reasonable, thanks for adding!

tabular/src/autogluon/tabular/predictor/predictor.py
Innixma: Adding `decision_threshold` here will be nice, it allows the user to see how the confusion matrix changes given different decision thresholds for binary classification. I can add.

Innixma (commented):
Thanks for the contrib! I've added an initial review. Most of my suggestions are related to nuances for handling the cases where `data=None` properly, which I can update the PR for since they require some knowledge of how out-of-fold preds are stored.

I'll try to find time to update the code to resolve the comments sometime in the next few weeks.

Innixma:
Thanks for the contribution, and sorry this sat — the ball was on our side. @Innixma offered to make these edits and hasn't had time, so I've pushed them (`7673b05f`). Your last commit had already handled `model`, `decision_threshold` and the bagged-OOF logic; three points were still open, and all three were silent, so each gets a test that fails without the fix.

## Do not re-transform internal data

`load_data_internal` returns data the feature generator has already transformed, and `predict` transformed it again — datetime columns re-expanded, categories re-encoded. On a frame with a category and a datetime:

```
internal val data, n=40: predictions differ = True  (14 of 40 rows)
internal columns: ['num', 'cat', 'when', 'when.year', 'when.month', 'when.day', 'when.dayofweek']
```

So `confusion_matrix()` with no data described predictions the model does not make. Fixed with `transform_features=False` on both internal-data branches.

## Score a refit_full model through its parent

A refit_full model is trained on the training *and* validation data, so scoring it on either is in-sample — and `refit_full` promotes it to `model_best`, which is exactly what `confusion_matrix()` with no arguments picks up:

```
has_val after refit:  True   model_best=LightGBM_FULL
confusion_matrix(data=None) -> [[22, 0], [2, 16]]
```

Worth noting the other half of that review comment needed nothing: `predict_oof("LightGBM_BAG_L1_FULL")` already returns the parent's out-of-fold predictions (verified equal to the parent's), so only the `has_val` branch was affected. Resolving the parent once up front covers both and keeps the reported model honest, with a log line pointing at passing `data` explicitly.

## Always return the matrix

It was returned only when `display=False`, so `cm = predictor.confusion_matrix(data=df)` — the default — silently gave `None`. The docstring already promised an ndarray unconditionally, so this just makes the code match its own documentation.

## Test runtime

Ten fits were `presets="medium_quality"` or full default fits. The confusion matrix depends on the shape of the predictions, not on model quality, so those now use `hyperparameters={"DUMMY": {}}`. 13 tests run in about 12 seconds.

Two are deliberate exceptions. A constant predictor cannot tell a re-transformed frame from the original, nor a refit_full model from its parent — so converting *those* to DUMMY made them pass regardless of the fix, which I only noticed because they kept passing against your original code. They now fit a real model over a learnable signal, and each carries a guard assertion that the two paths actually differ on the frame in question. That guard earned its keep immediately: my first attempt used a GBM on a random label, which predicts a constant, and the guard caught the test passing vacuously.

## Verification

All three new tests fail against the previous state of this branch and pass now. Full file: 13 passed. `ruff check --select I` clean.

One thing left for a maintainer: the branch is ~194 commits behind master. It merges cleanly (`mergeable: MERGEABLE`), but a `git rebase` trips on an intermediate commit that touched `.gitattributes` and is reverted by a later one — squashing on merge avoids it entirely.

Innixma (approved):
LGTM, thanks for the contrib!

Issue #4878, opened 2025-02-11
[timeseries] Add a method for backtesting/rolling evaluation to TimeSeriesPredictor


shchur:
### Discussed in https://github.com/autogluon/autogluon/discussions/4863

<div type='discussions-op-text'>

<sup>Originally posted by **smrutib2017** February  4, 2025</sup>
Hi, a couple of similar questions have been asked before but the answers didn't quite get me where I wanted to be. So want to ask again.

I am using the Time Series module for a prediction task with a prediction length of say 5. I have trained the models using TimeSeriesPredictor and data till say 31/12/2020. My test data starts from 10/1/2021. I want to generate 5 day forward forecasts for each day starting from 10/1/2021. Is there a more efficient way of doing this other than looping through all the dates of the test data and calling predict every time. If I pass test data to predict then it generates 5 day forward predictions only for the last date in test data.

Thanks</div>

atahanoezer:
Hi,

I also need this exact feature. Looping through the entire dataset currently takes a lot of time due to repeated function calls and internal bookkeeping.
I’m using another library (X) for comparison (same complexity), and it only takes a few seconds to generate all predictions for the test set. In contrast, with AutoGluon, the same process takes more than 20 minutes, which makes it practically unusable for my use case.
I believe adding a rolling forecast feature should be relatively straightforward and would significantly improve performance and usability.

shchur:
Complete in #5356

PR #5356, opened 2025-10-24
[timeseries] Add backtesting method for the TimeSeriesPredictor

shchur:
*Issue #, if available:* #4878

*Description of changes:*
- Add method `backtest_predictions` and `backtest_targets` methods for getting rolling predictions/targets with a trained `TimeSeriesPredictor`
- Add assertion that checks that `TimeSeriesPredictor` is fit before methods like `predict`, `backtest_targets` can be used.
- Make it possible to cache the OOF predictions for ensemble models by adding a public method `cache_oof_predictions`

*To do:*
- [x] Update documentation to show how the backtest predictions can be visualized / used to compute metrics for individual windows.
- [x] Make sure that the predictions returned by `backtest_predictions` and `backtest_targets` can be easily combined with other methods such as `TimeSeriesPredictor.plot`.

### DIFF
--- a/docs/tutorials/timeseries/forecasting-indepth.ipynb
+++ b/docs/tutorials/timeseries/forecasting-indepth.ipynb
@@ -918,7 +918,7 @@
     "Multi-window backtesting typically results in more accurate estimation of the forecast quality on unseen data.\n",
     "However, this strategy decreases the amount of training data available for fitting models, so we recommend using single-window backtesting if the training time series are short.\n",
     "\n",
-    "#### Comparing model performance across windows\n",
+    "### Comparing model performance across windows\n",
     "\n",
     "We can compare the performance of different models across the time windows by collecting the [`leaderboard()`](https://auto.gluon.ai/stable/api/autogluon.timeseries.TimeSeriesPredictor.leaderboard.html) scores for each cutoff:"
    ]
@@ -945,7 +945,7 @@
    "id": "ad137e0e",
    "metadata": {},
    "source": [
-    "#### Visualizing backtest predictions\n",
+    "### Visualizing backtest predictions\n",
     "\n",
     "To inspect the predictions and the corresponding targets from each backtest window, use the [`backtest_predictions()`](https://auto.gluon.ai/stable/api/autogluon.timeseries.TimeSeriesPredictor.backtest_predictions.html)\n",
     "and [`backtest_targets()`](https://auto.gluon.ai/stable/api/autogluon.timeseries.TimeSeriesPredictor.backtest_targets.html) methods:"


timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -868,6 +868,76 @@ def predict(
         )
         return cast(TimeSeriesDataFrame, predictions.reindex(original_item_id_order, level=ITEMID))
 
+    def backtest(
```
shchur: Some questions about the design:
1. Should this method support making predictions with multiple models instead of one? Using one model is more aligned with other methods like `predict` / `evaluate`. However, if the user wants to backtest multiple models, data splitting will be repeated, making the implementation less efficient.
2. Should the methods be called `backtest`, `cross_validate` or something else?
3. The returned DataFrame with cols `[item_id, timestamp, cutoff, target, mean, *quantiles]` is quite different from the format used internally to store out-of-fold predictions (`dict[str, list[TimeSeriesDataFrame]`). If we even decide to expose the `get_oof_predictions` method, should we think ahead about how to make the API consistent with `backtest`?

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -868,6 +868,76 @@ def predict(
...
+            This argument is ignored if ``cache_predictions`` was set to False when creating the ``TimeSeriesPredictor``.
+
+        Returns
+        -------
+        backtest_results : pd.DataFrame
+            DataFrame with MultiIndex ``["item_id", "timestamp"]`` and columns ``["cutoff", target, "mean"] + quantiles``,
+            where:
+
+            - "cutoff": timestamp of the last observation used to generate the forecast
+            - target: actual observed values
```
canerturkmen: nit: I understand that not having double quotes signals this is variable, but I would maybe highlight this?

timeseries/src/autogluon/timeseries/trainer/trainer.py
```diff
@@ -18,6 +18,7 @@
 from autogluon.core.utils.loaders import load_pkl
 from autogluon.core.utils.savers import save_pkl
 from autogluon.timeseries import TimeSeriesDataFrame
+from autogluon.timeseries.dataset.ts_dataframe import ITEMID, TIMESTAMP
```
canerturkmen: one PR to send ITEMID, TIMESTAMP to within `TimeSeriesDataFrame` class scope, yay or nay?

timeseries/src/autogluon/timeseries/predictor.py
canerturkmen: 1. I think consistency alone is not the concern as `str | list[str]` is not that different than `str`, and it exists (e.g., `persist`). However I think this being primarily a method of the Predictor, I think not exposing this function is OK. Finally, I don't think splitting data again is a huge penalty (?).
2. `backtest` sounds very nice, except this method does not do a full backtest but sort of `get_oof_predictions`. i.e., it's not `evaluate` but `predict`. `get_backtest_predictions` or `predict_for_backtest` maybe?
3. For consistency, I would definitely prefer `list[TimeSeriesDataFrame]` or `dict[CutoffT, TimeSeriesDataFrame]`. This introduces yet another pd.DataFrame format this class understands, and I think it's the only plain DataFrame object that contains time series (others are leaderboard and feature importance). Imho the bar for introducing this should be higher. 

Yet another point about consistency is that the actual and predictions live in the same data frame. I think this happens nowhere else in the library and our metric APIs kind of hint these should be two different data frames. I think it wouldn't hurt to split them or maybe drop target.

canerturkmen (commented):
Beautiful. Dropped comments on the API discussion. Thanks!!

timeseries/src/autogluon/timeseries/predictor.py
shchur: What do you think about the following API (name subject to change)?
```python
def predict_oof(
    data: TimeSeriesDataFrame | None = None, 
    model: str | list[str] | None = None, 
    ...
) -> dict[str, list[TimeSeriesDataFrame]]
```
The keys in the return dictionary are model names and the values are the lists of predictions for each validation window. This format is more aligned with the API of the metrics in AG-TS (but there is still manual work that needs to be done to get the relevant part of the ground truth to compute the metrics).

One downside of this approach is that the cutoff information is not available, unless we add it as an extra column to each TSDF in the values. We cannot use cutoff as key in the dict since each item might have a different cutoff (if timestamps are not aligned). We can similarly optionally add the target as an extra column for easier analysis:
```python
def predict_oof(
    ...,
    include_target: bool = False,
    include_cutoff: bool = False,
):   
```

If `data is None`, the method can return validation predictions for all models stored by the trainer. This is close the design of [`TabularPredictor.predict_oof`](https://auto.gluon.ai/stable/api/autogluon.tabular.TabularPredictor.predict_oof.html).

Another option is to overload the method to return a single `list[TSDF]` if `model: str`, otherwise return `dict[str, list[TSDF]]` if `model: list[str]`.

timeseries/src/autogluon/timeseries/predictor.py
shchur: We will likely change this API, so I will address this later.

timeseries/src/autogluon/timeseries/trainer/trainer.py
shchur: Great idea, we should definitely do that since the imports all over the codebase are quite annoying.

timeseries/src/autogluon/timeseries/predictor.py
shchur: Some relevant methods:
- nixtla: https://nixtlaverse.nixtla.io/statsforecast/src/core/core.html#statsforecast-cross-validation
- sktime: https://www.sktime.net/en/latest/api_reference/auto_generated/sktime.forecasting.model_evaluation.evaluate.html
- darts: https://unit8co.github.io/darts/generated_api/darts.models.forecasting.baselines.html#darts.models.forecasting.baselines.NaiveDrift.historical_forecasts

The nixtla version looks cleanest to me (it was the inspiration for the original design), but I agree that being somewhat consistent with our implementation of metrics is a more important consideration, so `list[TSDF]` seems to be the way to go.

canerturkmen: I'm leaning more towards "Another option is to overload the method to return a single list[TSDF] if model: str, otherwise return dict[str, list[TSDF]] if model: list[str]." which can clearly be marked with `overload`. I am also OK with not doing `model: list[str]` for now as I think this is anyway not too much overhead to perform the splitting again. 

I am also quite OK not being explicit with cutoffs. Here, it appears that the user does not explicitly provide the cutoffs, so they will anyway not be holding onto the keys to index into the returned dict / TSDF. They can infer time information from the timestamps of the TSDF.

For naming, I would avoid surfacing `oof`, as I'm always inclined to pronounce it as spelled :)

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -856,6 +856,121 @@ def predict(
         )
         return cast(TimeSeriesDataFrame, predictions.reindex(original_item_id_order, level=TimeSeriesDataFrame.ITEMID))
 
+    @overload
```
shchur, line 868: @canerturkmen what do you think about the current design? This makes it easy to compute metrics for each window, is more aligned with the rest of the API, and allows to seamlessly switch between loading the OOF predictions generated during training and backtest predictions on new data.

timeseries/setup.py
```diff
@@ -60,7 +60,6 @@
     ],
 }
 
-# chronos-openvino and chronos-onnx are deprecated, and will be removed in a future version
```
canerturkmen, line 67: thanks!

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -856,6 +856,121 @@ def predict(
...
+            val_step_size=val_step_size,
+            use_cache=use_cache,
+        )
+
+        if isinstance(model, list):
+            return result
+        else:
+            return result[model_names[0]]
+
+    def backtest_targets(
```
canerturkmen, line 998: In both of these I still read `backtest` as the verb and `_*` as the object of the verb. maybe is `get_backtest_*` better?

timeseries/src/autogluon/timeseries/predictor.py
canerturkmen: looks good to me!

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -856,6 +856,121 @@ def predict(
...
+        """Generate predictions for multiple validation windows using expanding window splits.
+
+        Parameters
+        ----------
+        data : TimeSeriesDataFrame, optional
+            Data to backtest on. If None, uses training data.
+        model : str | list[str] | None, default = None
+            Model(s) to use. If None, uses best model. If str, returns list. If list[str], returns dict.
+        num_val_windows : int, optional
+            Number of validation windows. If None, uses training configuration or defaults to 1.
```
canerturkmen: training configuration will potentially be `tuple[int]` now.

canerturkmen (commented):
cool! I assume there will need to be some rebasing now, and we could add tests. otherwise, LGTM!

timeseries/src/autogluon/timeseries/predictor.py
shchur: I think it's okay to only allow `int | None` for this method. I could not think of a reasonable return type for this method if `num_val_windows: tuple[int]` is provided. What do you think?

timeseries/src/autogluon/timeseries/predictor.py
shchur: In v1.0 we made a refactor removing `get_` from all method names in `TabularPredictor` and `TimeSeriesPredictor` (e.g., replacing `get_model_names` -> `model_names`, `get_model_best` -> `model_best`). This was a minor change for TS, but in Tabular there were ~20 methods starting with `get_` making the API a bit cluttered. 

I am unsure about reintroducing the prefix now and drifting away from the previous decision. I spend some time talking to LLMs but could not find an option that is clearly better than the current one. Some potential alternatives:
1. `validation_predictions` / `validation_targets`
2. `evaluation_predictions` / `evaluation_targets`

Do you like these / do you have any alternative suggestions without the `get_*` prefix?

timeseries/src/autogluon/timeseries/predictor.py
shchur: The internal OOF predictions for different models might start looking "weird" for the ensemble models trained on different windows, but I think that's a discussion that we should have on the ensembles PR.

canerturkmen: sure. just a heads up. anyway `sum(num_val_windows) == num_val_windows` will hold.

timeseries/src/autogluon/timeseries/predictor.py
canerturkmen: If it's for consistency, I'm OK with this. But I have a strong opinion about not methods being named only with verbs. We violated this before when the verb was obviously implicit, but now it's ambiguous.  Let's just do as is then.

I think `get_` is the only natural recommendation I would have as these behave most like accessors. Other verbs also don't make much sense to me.

timeseries/tests/unittests/test_predictor.py
```diff
@@ -1958,3 +1958,59 @@ def test_when_seasonal_period_is_provided_to_predictor_then_eval_metric_uses_it_
         predictor.evaluate(DUMMY_TS_DATAFRAME)
         predictor.leaderboard(DUMMY_TS_DATAFRAME)
         assert mock_save_past_metrics.call_args[1]["seasonal_period"] == expected_seasonal_period
+
+
+@pytest.mark.parametrize(
```
shchur, line 1963: @canerturkmen I have rebased the PR and updated the implementation. Let me know if you have any comments

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
```diff
@@ -143,6 +143,11 @@ def set_contexts(self, path_context):
         self.path = path_context
         self.path_root = self.path.rsplit(self.name, 1)[0]
 
+    def cache_oof_predictions(self, predictions: Union[TimeSeriesDataFrame, list[TimeSeriesDataFrame]]) -> None:
```
canerturkmen, line 146: nice!

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -852,6 +865,192 @@ def predict(
...
+        data: Optional[TimeSeriesDataFrame] = None,
+        *,
+        model: Optional[Union[str, list[str]]] = None,
+        num_val_windows: Optional[int] = None,
+        val_step_size: Optional[int] = None,
+        use_cache: bool = True,
+    ) -> Union[list[TimeSeriesDataFrame], dict[str, list[TimeSeriesDataFrame]]]:
+        """Generate predictions for multiple validation windows using expanding window backtesting.
+
+        This method performs backtesting by creating multiple validation windows from the provided data using an
```
canerturkmen: As previously discussed, I would not refer to this as "performing backtesting"

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -852,6 +865,192 @@ def predict(
...
+        expanding window strategy. For each window, predictions are generated for the next ``prediction_length``
+        time steps using models that were already trained during :meth:`~autogluon.timeseries.TimeSeriesPredictor.fit`.
+
+        The ground truth targets for each validation window can be obtained using
+        :meth:`~autogluon.timeseries.TimeSeriesPredictor.backtest_targets`.
+
+        Parameters
+        ----------
+        data : TimeSeriesDataFrame, optional
+            Time series data to perform backtesting on. If ``None``, uses the out-of-fold predictions generated
```
canerturkmen: rather, _returns_ the oof predictions?

canerturkmen (approved):
Couple of nits, otherwise looks good to me!

PR #5363, opened 2025-10-29
[timeseries] move ITEMID and TIMESTAMP to dataset namespace

canerturkmen:
*Issue #, if available:*

*Description of changes:*

Clean up ITEMID and TIMESTAMP from imports, move them under TimeSeriesDataFrame namespace.

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/gluonts/abstract.py
+++ b/timeseries/src/autogluon/timeseries/models/gluonts/abstract.py
@@ -21,7 +21,7 @@ from autogluon.core.hpo.constants import RAY_BACKEND
 from autogluon.tabular.models.tabular_nn.utils.categorical_encoders import (
     OneHotMergeRaresHandleUnknownEncoder as OneHotEncoder,
 )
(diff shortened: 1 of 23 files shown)


timeseries/src/autogluon/timeseries/models/gluonts/abstract.py
```diff
@@ -21,7 +21,7 @@
 from autogluon.tabular.models.tabular_nn.utils.categorical_encoders import (
     OneHotMergeRaresHandleUnknownEncoder as OneHotEncoder,
 )
-from autogluon.timeseries.dataset.ts_dataframe import ITEMID, TimeSeriesDataFrame
+from autogluon.timeseries.dataset.ts_dataframe import TimeSeriesDataFrame
```
shchur: One nit that mildly bothers me: can we also replace all occurrences of 
```python
from autogluon.timeseries.dataset.ts_dataframe import TimeSeriesDataFrame
```
with 
```python
from autogluon.timeseries.dataset import TimeSeriesDataFrame
```
in this PR?

PR #5371, opened 2025-10-31
[tabular] Fix build_all_docs CI step

shchur:
*Issue #, if available:*

*Description of changes:*
- Current `build_all_docs` CI step is failing because of the latest [`huggingface_hub` 1.0 release](https://pypi.org/project/huggingface-hub/#history) on Oct 27, 2025 that leads to a weird behavior during version resolution.
 For example, check https://github.com/autogluon/autogluon/actions/runs/18946183749/job/54113894600. 

image

This then leads to selection of an ancient transformers/tokenizers versions, leading to failure

image

This problem can be reproduced locally by installing all packages sequentially from source
```
python -m pip install --upgrade -e common/
python -m pip install --upgrade -e "core/[all]"
python -m pip install --upgrade -e features/
python -m pip install --upgrade -e "tabular/[all]"
```
as this currently happens in https://github.com/autogluon/autogluon/blob/98309df678b9f8bfee3e22e1dffa9b3761707278/.github/workflow_scripts/env_setup.sh#L75

Capping `huggingface_hub<1.0` on our side seems like a reasonable solution given that the latest release of `transformers` also requires hub < 1.0:
https://github.com/huggingface/transformers/blob/8cb5963cc22174954e7dca2c0a3320b7dc2f4edc/setup.py#L118

### DIFF
--- a/core/src/autogluon/core/_setup_utils.py
+++ b/core/src/autogluon/core/_setup_utils.py
@@ -32,7 +32,7 @@ DEPENDENT_PACKAGES = {
     "lightning": ">=2.5.1,<2.8",  # Major version cap
     "async_timeout": ">=4.0,<6",  # Major version cap
     "transformers[sentencepiece]": ">=4.38.0,<4.50",  # there is a breaking change in 4.50 for model config saving
-    "huggingface_hub": "<1.0",
+    "huggingface_hub[torch]": "<1.0",
     "accelerate": ">=0.34.0,<2.0",
     "typing-extensions": ">=4.0,<5",
     "joblib": ">=1.2,<1.7",  # <{N+1} upper cap


core/src/autogluon/core/_setup_utils.py
```diff
@@ -32,12 +32,17 @@
     "lightning": ">=2.5.1,<2.8",  # Major version cap
     "async_timeout": ">=4.0,<6",  # Major version cap
     "transformers[sentencepiece]": ">=4.38.0,<4.50",  # there is a breaking change in 4.50 for model config saving
+    "huggingface_hub": "<1.0",
```
shchur:
```suggestion
    "huggingface_hub[torch]": "<1.0",
```

Issue #5362, opened 2025-10-28
[BUG][timeseries] RecursiveTabular fails when known covariate is named "hour_of_day"


azzurracasali:
### Bug report: RecursiveTabular fails when known covariate is named "hour_of_day"

- [x] I provided code that demonstrates a minimal reproducible example.
- [x] I confirmed bug exists on the latest stable version of AutoGluon (1.4.0).
- [ ] I did not test mainline source install.

---

#### Describe the bug
When training `TimeSeriesPredictor` with `RecursiveTabular`, the pipeline fails with a LightGBM feature shape mismatch during recursive prediction if a known covariate column is named `hour_of_day`. The exact same values succeed when the column is renamed (e.g., to `hour`). This indicates name-specific preprocessing behavior that leads to inconsistent feature sets between training and prediction.

Typical error:
```
[LightGBM] [Fatal] The number of features in data (N) is not the same as it was in training data (N-1).
```

Other models (e.g., `DirectTabular`) do not exhibit this issue with the same data.

---

#### Expected behavior
- Preprocessing should produce identical feature sets between training and recursive prediction, regardless of column names. Using a column named `hour_of_day` should not trigger special handling or cause feature mismatch errors.

---

#### Minimal reproducible example
```python
import pandas as pd
import numpy as np
from autogluon.timeseries import TimeSeriesPredictor, TimeSeriesDataFrame

# Minimal hourly single-series data
rng = pd.date_range("2024-01-01", periods=240, freq="h")
df = pd.DataFrame({
    "item_id": "s0",
    "timestamp": rng,
    "target": np.random.randn(len(rng)).astype("float32")
})

df_1 = df.copy()
df_2 = df.copy()

def run(ts, kn):
    p = TimeSeriesPredictor(prediction_length=24, eval_metric="MASE", verbosity=3, known_covariates_names=kn)
    p.fit(train_data=ts,
          enable_ensemble=False,
          presets="medium_quality",
          hyperparameters={"RecursiveTabular": {}},
          time_limit=30)

# Fails (feature mismatch)
df_1["hour_of_day"] = df_1["timestamp"].dt.hour.astype("float32")
kn_1 = ["hour_of_day"]
ts_1 = TimeSeriesDataFrame.from_data_frame(df_1, id_column="item_id", timestamp_column="timestamp")
run(ts_1, kn_1)

# Succeeds (same data, renamed column)
[...]

PR #5373, opened 2025-11-01
[timeseries] Add covariate column name mapping for MLForecast models

alvaroborras:
This closes https://github.com/autogluon/autogluon/issues/5362

*Description of changes:*
  - Add name mapping to handle known covariates with sanitized names
  - Prefix known covariates with '__known_covariate__' to avoid conflicts
  - Update scaling logic to use original column names while working with sanitized names
  - Add test for recursive tabular model with hour_of_day covariate
  - Fix column name handling in _prepare_mlforecast_df method

This ensures proper handling of covariate columns when they have names that might conflict with internal MLForecast column names.

### DIFF
--- a/timeseries/tests/unittests/models/test_mlforecast.py
+++ b/timeseries/tests/unittests/models/test_mlforecast.py
@@ -160,9 +160,7 @@ def test_recursive_tabular_handles_hour_of_day_covariate(temp_model_path):
                 }
             )
 
-    index = pd.MultiIndex.from_tuples(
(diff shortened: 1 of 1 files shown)


shchur:
Hi @alvaroborras, I suppose this PR was motivated by the following issue #5362? I believe the solution in #5414 is more concise. 

Are there any specific scenarios besides the one described in #5362 that you want to address with this PR?

alvaroborras:
Hi @shchur 

I also think that the other PR is more concise. I'll close this one then :)

PR #5375, opened 2025-11-03
[timeseries] refactor trainer methods with ensemble logic

canerturkmen:
*Issue #, if available:*

*Description of changes:*

Refactor trainer logic, collecting ensemble fit logic behind a single method.
- send `num_val_windows` and `val_step_size` down to Trainer, making it the sole responsible of `Splitter` creation. also switch this to lazy creation (as its stateless).
- remove unnecessary indirection (`_train_single` and `_train_multi`). These carried over from tabular but are no longer relevant.
- move some logic related to ensemble fitting behind `fit_ensemble` which will be factored out for multilayer stacking.


timeseries/src/autogluon/timeseries/trainer/trainer.py
```diff
@@ -393,16 +382,38 @@ def _log_scores_and_times(
         if predict_time is not None:
             logger.info(f"\t{predict_time:<7.2f} s".ljust(15) + "= Validation (prediction) runtime")
 
-    def _train_multi(
```
abdulfatir, line 396: Is there additional context on what `_train_single` and `_train_multi` were previously useful for?

canerturkmen: Tabular has different variations on the theme of `_train_single` and `_train_multi` which it switches between. This was initially to mirror that structure. Since we factored out the core abstract trainer now, they are redundant.

PR #5376, opened 2025-11-04
[timeseries] add TabularPredictor stacker model

canerturkmen:
*Issue #, if available:*

*Description of changes:*

- add single quantile regressor variant of TabularPredictor-based stacking

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/tabular.py
+++ b/timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/tabular.py
@@ -19,10 +19,12 @@ class TabularEnsembleRegressor(EnsembleRegressor):
 
     def __init__(
         self,
+        path: str,
         quantile_levels: list[float],
         tabular_hyperparameters: Optional[dict] = None,
     ):
         super().__init__()
+        self.path = path
         self.quantile_levels = quantile_levels
         self.tabular_hyperparameters = tabular_hyperparameters or {}
         self.predictor: Optional[TabularPredictor] = None
@@ -36,6 +38,7 @@ class TabularEnsembleRegressor(EnsembleRegressor):
         **kwargs,
     ) -> Self:
         self.predictor = TabularPredictor(
+            path=self.path,
             label="target",
             problem_type="quantile",
             quantile_levels=self.quantile_levels,
@@ -64,7 +67,10 @@ class TabularEnsembleRegressor(EnsembleRegressor):
         base_model_quantile_predictions: np.ndarray,
     ) -> tuple[np.ndarray, np.ndarray]:
         if self.predictor is None:
-            raise ValueError("Model must be fitted before prediction")
+            try:
+                self.predictor = TabularPredictor.load(self.path)
+            except FileNotFoundError:
+                raise ValueError("Model must be fitted before prediction")
 
         df = self._get_feature_df(base_model_mean_predictions, base_model_quantile_predictions)
 
--- a/timeseries/src/autogluon/timeseries/models/ensemble/array_based/abstract.py
+++ b/timeseries/src/autogluon/timeseries/models/ensemble/array_based/abstract.py
@@ -175,7 +175,10 @@ class ArrayBasedTimeSeriesEnsembleModel(AbstractTimeSeriesEnsembleModel, ABC):
 
     def _predict(self, data: dict[str, TimeSeriesDataFrame], **kwargs) -> TimeSeriesDataFrame:
         if self.ensemble_regressor is None:
-            raise ValueError("Ensemble model has not been fitted yet.")
+            if not self._model_names:
+                raise ValueError("Ensemble model has not been fitted yet.")
+            # Try to recreate the regressor (for loaded models)
+            self.ensemble_regressor = self._get_ensemble_regressor()
 
         input_data = {}
         for m in self.model_names:
--- a/timeseries/src/autogluon/timeseries/models/ensemble/array_based/models.py
+++ b/timeseries/src/autogluon/timeseries/models/ensemble/array_based/models.py
@@ -1,11 +1,12 @@
+import os
 from typing import Any
 
 from .abstract import ArrayBasedTimeSeriesEnsembleModel
-from .regressor import EnsembleRegressor, MedianEnsembleRegressor, TabularEnsembleRegressor
+from .regressor import MedianEnsembleRegressor, TabularEnsembleRegressor
 
 
 class MedianEnsemble(ArrayBasedTimeSeriesEnsembleModel):
-    def _get_ensemble_regressor(self) -> EnsembleRegressor:
+    def _get_ensemble_regressor(self) -> MedianEnsembleRegressor:
         return MedianEnsembleRegressor()
 
 
@@ -25,9 +26,10 @@ class TabularEnsemble(ArrayBasedTimeSeriesEnsembleModel):
         )
         return default_hps
 
-    def _get_ensemble_regressor(self) -> EnsembleRegressor:
+    def _get_ensemble_regressor(self) -> TabularEnsembleRegressor:
         """Create and return a TabularEnsembleRegressor instance."""
         return TabularEnsembleRegressor(
+            path=os.path.join(self.path, "ensemble_regressor"),
             quantile_levels=list(self.quantile_levels),
             tabular_hyperparameters=self.get_hyperparameters()["tabular_hyperparameters"],
         )
--- a/timeseries/tests/unittests/models/ensemble/array_based/test_models.py
+++ b/timeseries/tests/unittests/models/ensemble/array_based/test_models.py
@@ -1,4 +1,5 @@
 import math
+import os
 from itertools import chain
 
 import numpy as np
@@ -126,9 +127,9 @@ class TestTabularEnsemble:
         ],
     )
     def test_given_quantile_levels_when_get_median_quantile_index_called_then_correct_index_returned(
-        self, quantile_levels, expected_median_idx
+        self, quantile_levels, expected_median_idx, tmp_path
     ):
-        regressor = TabularEnsembleRegressor(quantile_levels=quantile_levels)
+        regressor = TabularEnsembleRegressor(path=str(tmp_path), quantile_levels=quantile_levels)
         median_idx = regressor._get_median_quantile_index()
         assert median_idx == expected_median_idx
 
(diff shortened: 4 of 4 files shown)


timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/tabular.py
```diff
@@ -0,0 +1,127 @@
...
+        **kwargs,
+    ) -> Self:
+        time_limit = kwargs.get("time_limit", None)
+
+        self.predictor = TabularPredictor(
+            label="target",
+            problem_type="quantile",
+            quantile_levels=self.quantile_levels,
+            verbosity=1,
+        )
```
Innixma, line 46: Missing `path` param to store the artifacts in the correct location?

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/tabular.py
```diff
@@ -0,0 +1,127 @@
...
+        # get labels
+        num_windows, num_items, prediction_length = base_model_mean_predictions.shape[:3]
+        label_series = labels.reshape(num_windows * num_items * prediction_length)
+        df["target"] = label_series
+
+        self.predictor.fit(
+            df,
+            hyperparameters=self.tabular_hyperparameters,
+            time_limit=time_limit,  # type: ignore
+        )
```
Innixma, line 60: This leads to TabularPredictor doing a random train/val split as if the data was IID because `tuning_data` wasn't specified. Do we want to instead pre-define the split? Alternatively we could enable bagging for better results if we aren't worried about validation leakage.

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/tabular.py
```diff
@@ -0,0 +1,127 @@
...
+
+    def __init__(
+        self,
+        quantile_levels: list[float],
+        tabular_hyperparameters: Optional[dict] = None,
+    ):
+        super().__init__()
+        self.quantile_levels = quantile_levels
+        self.tabular_hyperparameters = tabular_hyperparameters or {}
+        self.predictor: Optional[TabularPredictor] = None
```
Innixma, line 30: To enable portability, you'll need to implement a save/load logic in TabularEnsembleRegressor that loads the predictor from a saved relative path like models are saved in TabularPredictor. Otherwise this will break if you move the cached files to a new directory or machine.

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/tabular.py
```diff
@@ -0,0 +1,127 @@
...
+        base_model_quantile_predictions: np.ndarray,
+    ) -> tuple[np.ndarray, np.ndarray]:
+        if self.predictor is None:
+            raise ValueError("Model must be fitted before prediction")
+
+        df = self._get_feature_df(base_model_mean_predictions, base_model_quantile_predictions)
+
+        pred = self.predictor.predict(df)
+        if isinstance(pred, pd.DataFrame):
+            pred = pred.to_numpy()
```
Innixma: You can instead call `self.predictor.predict(df, as_pandas=False)` which is faster.

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/tabular.py
```diff
@@ -0,0 +1,127 @@
...
+        self,
+        base_model_mean_predictions: np.ndarray,
+        base_model_quantile_predictions: np.ndarray,
+    ) -> tuple[np.ndarray, np.ndarray]:
+        if self.predictor is None:
+            raise ValueError("Model must be fitted before prediction")
+
+        df = self._get_feature_df(base_model_mean_predictions, base_model_quantile_predictions)
+
+        pred = self.predictor.predict(df)
```
Innixma: If you plan to call predict multiple times, you may want to persist the models into memory with `self.predictor.persist()` for faster throughput (be aware of memory usage though)

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/tabular.py
```diff
@@ -0,0 +1,127 @@
...
+
+class TabularEnsembleRegressor(EnsembleRegressor):
+    """TabularPredictor ensemble regressor using AutoGluon-Tabular as a single
+    quantile regressor for the target.
+    """
+
+    def __init__(
+        self,
+        quantile_levels: list[float],
+        tabular_hyperparameters: Optional[dict] = None,
```
Innixma, line 24: This can be made more generic by having `tabular_init_kwargs` and `tabular_fit_kwargs`.

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/tabular.py
```diff
@@ -0,0 +1,127 @@
...
+        self.predictor: Optional[TabularPredictor] = None
+
+    def fit(
+        self,
+        base_model_mean_predictions: np.ndarray,
+        base_model_quantile_predictions: np.ndarray,
+        labels: np.ndarray,
+        **kwargs,
+    ) -> Self:
+        time_limit = kwargs.get("time_limit", None)
```
Innixma: May as well put `time_limit` as an arg rather than a kwarg

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/tabular.py
```diff
@@ -0,0 +1,127 @@
...
+        self.quantile_levels = quantile_levels
+        self.tabular_hyperparameters = tabular_hyperparameters or {}
+        self.predictor: Optional[TabularPredictor] = None
+
+    def fit(
+        self,
+        base_model_mean_predictions: np.ndarray,
+        base_model_quantile_predictions: np.ndarray,
+        labels: np.ndarray,
+        **kwargs,
```
Innixma, line 38: Should we have `**kwargs` here? Is there a reason it is present?

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/tabular.py
canerturkmen: IID is fine in this case (different rows are different items, and are assumed iid). This is really just porting what we explored with Nathanael so let's punt enhancements maybe.

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/tabular.py
canerturkmen: I don't think we are in this case. Thanks!

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/tabular.py
canerturkmen: Because without it it's an LSP violation. If `superclass.method(.., **kwargs)` then all `subclass.method(...,**kwargs)` also or the type checker complains. 

One could get rid of `**kwargs` and make `time_limit` explicit everywhere. But then this signals time limit will be handled by all `EnsembleRegressor`s, which is not the case. I think I'm a bit more comfortable with keeping `**kwargs` around for now but we can come back to trim later.

canerturkmen: i.e., we will add a couple more of these classes.

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/tabular.py
canerturkmen: I purposely hide these from the user. Even exposing hyperparameters is debatable imo. We only really had a good experience with GBM iirc and I would avoid exposing too many knobs that would break. For example, we don't want to expose the ability to set `problem_type="classification"` or turn bagging up to eleven, etc.

Innixma: It is less about the user and more about the developers having the flexibility to toggle options. Is this supposed to be a primarily user-facing class?

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/tabular.py
Innixma: sounds good, I also keep **kwargs pretty much everywhere in Tabular's models

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/tabular.py
Innixma: If we can benchmark it easily, we should try toggling `num_bag_folds=8` and seeing how it compares (of course after the initial merge to get things working). I suspect it would improve.

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/tabular.py
canerturkmen: No. Less so than a model class, the hyperparameters are maybe configurable for power users, but not really recommended. I understand you have the intuition also from the bagging discussion that one can drive accuracy further by tuning this bit. One probably can, but I would really like to look at that further down the road. 

For now, what we really want to achieve with ensemble classes is reliability and almost 0 fit_time.

In other words, this is already too generic. I think trying to one-shot with GBM is already the maximum one should do for forecast combination, and we're using TabularPredictor as a wrapper for that not because it's more flexible or more accurate, but because it's more stable. Users can already use/tune/get the most out of TabularPredictor for doing the actual forecasting elsewhere in the library. This is not that bit.

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/tabular.py
canerturkmen: Thanks for the save here! I think we might be able to get away without custom save/load. I updated the PR. Will check the tests and keep this in mind going forward.

timeseries/tests/unittests/models/ensemble/test_array_based.py
```diff
@@ -144,19 +145,6 @@ def test_given_unfitted_model_when_predict_called_then_error_raised(self, model)
         with pytest.raises(ValueError, match="Ensemble model has not been fitted yet"):
             model._predict(data)
 
-    def test_given_model_when_fit_called_with_hyperparameters_then_regressor_receives_hyperparameters(
-        self, model, ensemble_data
-    ):
-        model.prediction_length = 5  # Match the prediction data
-        model.hyperparameters = {"isotonization": False}
```
abdulfatir, line 151: What was `isotonization`? 😄

abdulfatir (approved):
Missed this one somehow. Looks good!

PR #5380, opened 2025-11-05
[timeseries] remove Chronos source code, depend on PyPI chronos

canerturkmen:
*Issue #, if available:*

*Description of changes:*

- Remove Chronos source code, factoring out the Pipeline creation logic and quantile updates for Bolt finetuning.


timeseries/src/autogluon/timeseries/models/chronos/pipeline_factory.py
```diff
@@ -0,0 +1,58 @@
...
+
+class PipelineFactory:
+    dtypes = {
+        "bfloat16": torch.bfloat16,
+        "float32": torch.float32,
+        "float64": torch.float64,
+    }
+
+    @classmethod
+    def get_pipeline(
```
abdulfatir: The logic in this function (loading from S3 and checks) is identical to what `BaseChronosPipeline` [provides](https://github.com/amazon-science/chronos-forecasting/blob/c23d34cd887b889c302ca7b6df3fa0bca96d78a9/src/chronos/base.py#L210). Should we just call `BaseChronosPipeline.from_pretrained`?

abdulfatir:
See also: https://github.com/amazon-science/chronos-forecasting/pull/366

PR #5388, opened 2025-11-11
[Tabular] GPU optimization, scheduling for parallel_local fitting strategy

prateekdesai04:
*Description of changes:*
Distribute load all folds on all available GPUs equally for simplicity, this is used when `fold_fitting_strategy` is set to `parallel_local`.

We notice that Ray's scheduler sometimes places unequal load on all GPUs, i.e., when `num_gpus = 3` and `num_bag_folds = 17` then by default all GPUs get 3 tasks each but for the remaining two tasks ray sometimes schedules them on one GPU for example GPU 0 instead of giving GPU 0 and GPU 1 the two tasks, causing uneven load.

This change addresses this issue by doing exactly what is mentioned above.

How to test:

1. Fit a small tabular model with `num_bag_folds` and `num_gpus` to any value
2. Set model hyperparameters to `ag.ens.fold_fitting_strategy':'parallel_local`
3. Simultaneously run `watch -n 0.1 nvidia-smi` on a new window
4. See ray process spin up evenly on GPUs 

### DIFF
--- a/core/src/autogluon/core/models/ensemble/fold_fitting_strategy.py
+++ b/core/src/autogluon/core/models/ensemble/fold_fitting_strategy.py
@@ -410,7 +410,7 @@ def _ray_fit(
     y_pseudo: Union[str, pd.DataFrame],
     task_id: int,
     fold_ctx: Dict[str, Any],
-    task_to_gpu_ids_list: List[int],
+    task_gpu_ids: List[int],
     time_limit_fold: float,
     save_bag_folds: bool,
     resources: Dict[str, Any],
@@ -419,10 +419,10 @@ def _ray_fit(
     model_sync_path: Optional[str] = None,
 ):
     import ray  # ray must be present
-    if task_to_gpu_ids_list:
+    if task_gpu_ids:
         # Set CUDA_VISIBLE_DEVICES to the assigned GPU IDs
-        os.environ['CUDA_VISIBLE_DEVICES'] = ','.join(map(str, task_to_gpu_ids_list))
-        logger.debug(f"Set CUDA_VISIBLE_DEVICES to {task_to_gpu_ids_list}")
+        os.environ['CUDA_VISIBLE_DEVICES'] = ','.join(map(str, task_gpu_ids))
+        logger.debug(f"Set CUDA_VISIBLE_DEVICES to {task_gpu_ids}")
 
     reset_logger_for_remote_call(verbosity=kwargs_fold.get("verbosity",2))
 
@@ -860,7 +860,7 @@ class ParallelFoldFittingStrategy(FoldFittingStrategy):
             y_pseudo=y_pseudo_ref,
             task_id=task_id,
             fold_ctx=fold_ctx_ref,
-            task_to_gpu_ids_list=gpu_assignments[task_id],
+            task_gpu_ids=gpu_assignments[task_id],
             time_limit_fold=time_limit_fold,
             save_bag_folds=save_bag_folds,
             resources=resources_model,


core/src/autogluon/core/models/ensemble/fold_fitting_strategy.py
```diff
@@ -726,10 +748,23 @@ def _run_pseudo_sequential(self, X, y, X_pseudo, y_pseudo, model_base_ref, time_
...
 
+    def _calculate_gpu_assignment(self, gpu_assignments: Dict, task_id: int, gpus_per_task: int, total_gpus: int):
+        if gpus_per_task >= 1:
+            gpu_id = task_id * gpus_per_task
+            assigned_gpus = []
+            for i in range(gpus_per_task):
+                assigned_gpus.append(gpu_id + i % total_gpus)
+            gpu_assignments[task_id] = assigned_gpus
+        else:
+            gpu_id = task_id % self.num_gpus
```
Innixma: ensure consistent usage of either `total_gpus` or `self.num_gpus`

tabular/tests/unittests/resource_allocation/test_gpu_assignment.py
```diff
@@ -0,0 +1,490 @@
...
+        time_start=time.time(),
+        time_limit=None,
+        models=[],
+        oof_pred_proba=np.array([]),
+        oof_pred_model_repeats=np.array([]),
+        save_folds=True,
+        num_cpus=num_cpus,
+        num_gpus=num_gpus,
+    )
+    return ParallelLocalFoldFittingStrategy(**args)
```
Innixma, line 134: Lets actually make some of the more generic parameters here have defaults in the actual class so we don't need to specify so many things here in the test explicitly.

For example, `X_pseudo, y_pseudo, `sample_weight, time_start, time_limit, save_folds` can all be given default values equal to what you have here.

tabular/tests/unittests/resource_allocation/test_gpu_assignment.py
```diff
@@ -0,0 +1,490 @@
...
+    """
+    Create a Ray remote function that reports GPU information.
+
+    This function is created outside the test class so it can be serialized
+    and sent to Ray remote workers.
+    """
+    try:
+        import ray
+    except ImportError:
+        return None
```
Innixma: Why are we allowing None / ImportError? Feels like this doesn't help us with tests.

tabular/tests/unittests/resource_allocation/test_gpu_assignment.py
```diff
@@ -0,0 +1,490 @@
...
+        gpu_assignments = {}
+
+        gpu_assignments = strategy._calculate_gpu_assignment(
+            gpu_assignments=gpu_assignments,
+            task_id=0,
+            gpus_per_task=0,
+            total_gpus=0
+        )
+
+        assert gpu_assignments[0] == []
```
Innixma: we should verify more than just this from the output, such as len(gpu_assignments).

tabular/tests/unittests/resource_allocation/test_gpu_assignment.py
```diff
@@ -0,0 +1,490 @@
...
+            gpu_assignments = strategy._calculate_gpu_assignment(
+                gpu_assignments=gpu_assignments,
+                task_id=task_id,
+                gpus_per_task=0.5,
+                total_gpus=1
+            )
+
+        # Both tasks should be assigned to GPU 0
+        assert gpu_assignments[0] == [0]
+        assert gpu_assignments[1] == [0]
```
Innixma: verify there are only 2 tasks

tabular/tests/unittests/resource_allocation/test_gpu_assignment.py
```diff
@@ -0,0 +1,490 @@
...
+
+        expected = [0, 1, 0, 1]
+        for task_id, expected_gpu in enumerate(expected):
+            assert gpu_assignments[task_id] == [expected_gpu]
+
+        # Submit tasks and collect reports
+        task_refs = []
+        for task_id in range(4):
+            ref = ray_task_report_gpu_info.remote(task_id, gpu_assignments[task_id])
+            task_refs.append(ref)
```
Innixma: can we verify that a task waits to be scheduled until its GPU is available?

can we check the available GPUs as jobs finish one by one?

tabular/tests/unittests/resource_allocation/test_gpu_assignment.py
```diff
@@ -0,0 +1,490 @@
...
+            gpu_assignments = strategy._calculate_gpu_assignment(
+                gpu_assignments=gpu_assignments,
+                task_id=task_id,
+                gpus_per_task=0.5,
+                total_gpus=2
+            )
+
+        expected = [0, 1, 0, 1]
+        for task_id, expected_gpu in enumerate(expected):
+            assert gpu_assignments[task_id] == [expected_gpu]
```
Innixma: Need to test extra scenario where we have 4 tasks, minimum 1 GPU per task, and only 2 GPUs available. In this scenario:

If task 2 finishes before task 1, shouldn't task 3 launch with GPU 1 instead of task 4 launching with GPU 1?

Doing it this way means the GPU used for a task isn't pre-determined before the task is ran, but rather is dynamically allocated based on available resources.

This seems like it would be most efficient.

tabular/tests/unittests/resource_allocation/test_gpu_assignment.py
```diff
@@ -0,0 +1,490 @@
...
+        ray.shutdown()
+
+    @pytest.mark.multi_gpu
+    def test_ray_gpu_assignment_multiple_gpus_round_robin_task_execution(self):
+        """
+        Integration Test: Ray tasks verify round-robin GPU assignment
+        This test should run on continuous_integration_multigpu.yaml
+
+        Scenario: 2 GPUs, 4 tasks
+        num_gpus_per_task = 2 / 4 = 0.5 (fractional, round-robin)
```
Innixma: Test edge cases, ensure logic works as expected

- 2 GPUs, 3 tasks, per-task = 2/3 GPU
- 2 GPUs, 3 tasks, per-task = 1 GPU
- 3 GPUs, 2 tasks, per-task = 1.5 GPUs (assert raises?)

core/src/autogluon/core/models/ensemble/fold_fitting_strategy.py
```diff
@@ -726,10 +752,27 @@ def _run_pseudo_sequential(self, X, y, X_pseudo, y_pseudo, model_base_ref, time_
...
+        if gpus_per_task >= 1:
+            gpu_id = task_id * gpus_per_task
+            assigned_gpus = []
+            for i in range(gpus_per_task):
+                assigned_gpus.append(gpu_id + i % total_gpus)
+            gpu_assignments[task_id] = assigned_gpus
+        else:
+            gpu_id = task_id % total_gpus
+            gpu_assignments[task_id] = [gpu_id]
+        return gpu_assignments
```
Innixma: Why are we doing in-place edits to gpu_assignments?
Why not just return the value we want to set it to, and then

```
gpu_assignments[task_id] = self._calculate_gpu_assignment(task_id, gpus_per_task, total_gpus)
```

core/src/autogluon/core/models/ensemble/fold_fitting_strategy.py
```diff
@@ -726,10 +752,27 @@ def _run_pseudo_sequential(self, X, y, X_pseudo, y_pseudo, model_base_ref, time_
...
+    def _calculate_gpu_assignment(self, gpu_assignments: Dict, task_id: int, gpus_per_task: int, total_gpus: int):
+        if total_gpus == 0:
+            logger.debug(f"No GPUs available, CPU-only mode for task {task_id}")
+            gpu_assignments[task_id] = []
+            return gpu_assignments   
+        if gpus_per_task >= 1:
+            gpu_id = task_id * gpus_per_task
+            assigned_gpus = []
+            for i in range(gpus_per_task):
+                assigned_gpus.append(gpu_id + i % total_gpus)
```
Innixma: Is this a bug?

Shouldn't it be:

```
assigned_gpus.append((gpu_id + i) % total_gpus)
```

core/src/autogluon/core/models/ensemble/fold_fitting_strategy.py
```diff
@@ -726,10 +752,27 @@ def _run_pseudo_sequential(self, X, y, X_pseudo, y_pseudo, model_base_ref, time_
...
+            gpu_assignments[task_id] = assigned_gpus
```
Innixma: Should we sort the GPU IDs?

`assigned_gpus = sorted(assigned_gpus)`

core/src/autogluon/core/models/ensemble/fold_fitting_strategy.py
```diff
@@ -726,10 +752,27 @@ def _run_pseudo_sequential(self, X, y, X_pseudo, y_pseudo, model_base_ref, time_
 
         self._update_bagged_ensemble_times()
 
+    def _calculate_gpu_assignment(self, gpu_assignments: Dict, task_id: int, gpus_per_task: int, total_gpus: int):
+        if total_gpus == 0:
+            logger.debug(f"No GPUs available, CPU-only mode for task {task_id}")
+            gpu_assignments[task_id] = []
+            return gpu_assignments   
+        if gpus_per_task >= 1:
```
Innixma, line 765: Should we add an assert that it is not a float if >=1?

core/src/autogluon/core/models/ensemble/fold_fitting_strategy.py
```diff
@@ -726,10 +752,27 @@ def _run_pseudo_sequential(self, X, y, X_pseudo, y_pseudo, model_base_ref, time_
 
         self._update_bagged_ensemble_times()
 
+    def _calculate_gpu_assignment(self, gpu_assignments: Dict, task_id: int, gpus_per_task: int, total_gpus: int):
```
Innixma: gpus_per_task: int | float

core/src/autogluon/core/models/ensemble/fold_fitting_strategy.py
```diff
@@ -726,10 +752,27 @@ def _run_pseudo_sequential(self, X, y, X_pseudo, y_pseudo, model_base_ref, time_
...
+        if total_gpus == 0:
```
Innixma, line 762: assert total_gpus >= 0
assert gpus_per_task >= 0
assert task_id >= 0

core/src/autogluon/core/models/ensemble/fold_fitting_strategy.py
```diff
@@ -417,6 +419,14 @@ def _ray_fit(
     model_sync_path: Optional[str] = None,
 ):
     import ray  # ray must be present
+    import torch
```
Innixma: What if torch isn't installed?

tabular/tests/unittests/resource_allocation/test_gpu_assignment.py
```diff
@@ -0,0 +1,490 @@
...
+
+        # Verify each task sees its assigned GPU
+        for result in results:
+            task_id = result['task_id']
+            expected_gpu = task_id
+            assert result['CUDA_VISIBLE_DEVICES'] == str(expected_gpu)
+            # Each task should see exactly 1 GPU
+            if result['torch_gpu_count'] != 'torch not available':
+                assert result['torch_gpu_count'] == 1, \
+                    f"Task {task_id} expected 1 GPU, saw {result['torch_gpu_count']}"
```
Innixma: I notice a lot of code dupe between these tests. Try to re-use logic into functions as much as possible. I think we can make this code a lot shorter by re-using verification logic.

Innixma (commented):
Added first round of comments, awesome contribution!

tabular/tests/unittests/resource_allocation/test_gpu_assignment.py
prateekdesai04: removed import to fail if `ray` is not installed or detected

tabular/tests/unittests/resource_allocation/test_gpu_assignment.py
prateekdesai04: Will look at dynamic allocation in a follow-up PR as it seems to be a little more complex than this approach.
But added more tests where available GPUs = n, tasks = m, and gpus_per_task = n where (n > m)

tabular/tests/unittests/resource_allocation/test_gpu_assignment.py
prateekdesai04: converted verification logic into separate functions

core/src/autogluon/core/models/ensemble/fold_fitting_strategy.py
```diff
@@ -417,6 +419,11 @@ def _ray_fit(
     model_sync_path: Optional[str] = None,
 ):
     import ray  # ray must be present
+    gpu_ids = assignments.get(task_id, [])
+    if gpu_ids:
+        # Set CUDA_VISIBLE_DEVICES to the assigned GPU IDs
+        os.environ['CUDA_VISIBLE_DEVICES'] = ','.join(map(str, gpu_ids))
+        logger.debug(f"Set CUDA_VISIBLE_DEVICES to {gpu_ids}")
```
Innixma: What if gpu_ids is `[]`?
This will treat `if gpu_ids` as False, which means we won't actually edit `CUDA_VISIBLE_DEVICES`.
Would it perhaps be better to do `gpu_ids = assignments.get(task_id, None)` and `if gpu_ids is not None:`?

core/src/autogluon/core/models/ensemble/fold_fitting_strategy.py
```diff
@@ -408,7 +408,9 @@ def _ray_fit(
     y: Union[str, pd.DataFrame],
     X_pseudo: Union[str, pd.DataFrame],
     y_pseudo: Union[str, pd.DataFrame],
+    task_id: int,
     fold_ctx: Dict[str, Any],
+    assignments: Dict[int, List[int]],
```
Innixma: `assignments` is too generic of a name. Consider something like `task_to_gpu_ids_map`.

Also, why are we sending in a dict when this is a fit call for a specific task_id.
Why not just pass in `gpu_ids` directly?

core/src/autogluon/core/models/ensemble/fold_fitting_strategy.py
```diff
@@ -661,9 +679,10 @@ def _update_bagged_ensemble_child_resources(self):
     def _run_parallel(self, X, y, X_pseudo, y_pseudo, model_base_ref, time_limit_fold, head_node_id):
         job_refs = []
         job_fold_map = {}
+        gpu_assignments = {}
```
Innixma, line 684: Why are we passing an empty dictionary?
Why not have `gpu_assignments` be a variable of the class: `self.gpu_assignments`?

core/src/autogluon/core/models/ensemble/fold_fitting_strategy.py
```diff
@@ -705,7 +726,9 @@ def _run_pseudo_sequential(self, X, y, X_pseudo, y_pseudo, model_base_ref, time_
         a job could start fitting a model while the results are processed; resulting in the fit running out of memory
         due to the overhead of processing and storing the result.
         """
-        for job in self.jobs:
+        gpu_assignments = {}
```
Innixma, line 731: Why are we passing an empty dictionary?
Why not have `gpu_assignments` be a variable of the class: `self.gpu_assignments`?

core/src/autogluon/core/models/ensemble/fold_fitting_strategy.py
```diff
@@ -807,7 +855,9 @@ def _fit(
             y=y_ref,
             X_pseudo=X_pseudo_ref,
             y_pseudo=y_pseudo_ref,
+            task_id=task_id,
             fold_ctx=fold_ctx_ref,
+            assignments=gpu_assignments,
```
Innixma: Don't pass a task more information than it needs. It doesn't need to know about other task's GPU assignments.

core/src/autogluon/core/models/ensemble/fold_fitting_strategy.py
prateekdesai04: Changed the variable name.
Sending in the whole dict may help us achieve 2 things
1. If we have to use the whole dict for dynamic allocation logic in the future
2. Monitor the state of the dict on all ray workers (better for logging and debugging)

core/src/autogluon/core/models/ensemble/fold_fitting_strategy.py
prateekdesai04: `_calculate_gpu_assignment()` returns `[]` in CPU only mode hence the current implementation works, would replacing `[]` with `None` make any difference ?

core/src/autogluon/core/models/ensemble/fold_fitting_strategy.py
prateekdesai04: My assumption/understanding may be wrong, but what if the ensemble starts training another model before it finishes training the first model or basically another instance of FoldFittingStrategy is called in the same process?

core/src/autogluon/core/models/ensemble/fold_fitting_strategy.py
prateekdesai04: ditto as above

core/src/autogluon/core/models/ensemble/fold_fitting_strategy.py
prateekdesai04: same comment as to why we pass the dict - for dynamic allocation and logging/monitoring in the future for debugging states across multiple ray processes

Innixma: A task doesn't need to know the state of other processes. And it isn't necessary for dynamic allocation, that is handled in the main process, not the workers.

core/src/autogluon/core/models/ensemble/fold_fitting_strategy.py
```diff
@@ -425,6 +432,17 @@ def _ray_fit(
...
+    try:
+        import torch
+        visible_gpus = os.environ.get("CUDA_VISIBLE_DEVICES", "not set")
+        num_gpus = torch.cuda.device_count()
+        current_gpu = torch.cuda.current_device() if torch.cuda.is_available() else "N/A"
+        print(f"[GPU DEBUG] CUDA_VISIBLE_DEVICES={visible_gpus}, Torch sees {num_gpus} GPUs, Using GPU {current_gpu}", flush=True)
+    except ImportError:
+        pass
+    except Exception as e:
+        print(f"[GPU DEBUG] Could not get GPU info: {e}", flush=True)
```
Innixma: This is too heavy of a debug statement, we don't want to import torch unless we have to. You can add a debug flag for this, but don't import torch by default in workers or it slows everything down.

PR #5389, opened 2025-11-11
[timeseries] add per-quantile TabularPredictor based ensemble

canerturkmen:
*Issue #, if available:*

*Description of changes:*

- Add the second version of the `TabularPredictor` based ensemble that fits an independent model for every quantile + the point forecast. 
- Some more minor refactors.

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/per_quantile_tabular.py
+++ b/timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/per_quantile_tabular.py
@@ -29,6 +29,9 @@ class PerQuantileTabularEnsembleRegressor(EnsembleRegressor):
         self.quantile_predictors: list[TabularPredictor] = []
         self.mean_predictor: Optional[TabularPredictor] = None
 
+    def set_path(self, path: str) -> None:
+        self.path = path
+
     def fit(
         self,
         base_model_mean_predictions: np.ndarray,
@@ -37,6 +40,7 @@ class PerQuantileTabularEnsembleRegressor(EnsembleRegressor):
         **kwargs,
     ) -> Self:
         """Fit separate TabularPredictor for mean and each quantile level."""
+        # TODO: implement time_limit
 
         num_windows, num_items, prediction_length = base_model_mean_predictions.shape[:3]
         target = labels.reshape(num_windows * num_items * prediction_length).ravel()
@@ -48,6 +52,7 @@ class PerQuantileTabularEnsembleRegressor(EnsembleRegressor):
             label="target",
             path=os.path.join(self.path, "mean"),
             verbosity=1,
+            problem_type="regression",
         ).fit(
             mean_df,
             hyperparameters=self.tabular_hyperparameters,
--- a/timeseries/src/autogluon/timeseries/models/ensemble/array_based/abstract.py
+++ b/timeseries/src/autogluon/timeseries/models/ensemble/array_based/abstract.py
@@ -1,7 +1,9 @@
+import os
 from abc import ABC, abstractmethod
 from typing import Any, Optional, Sequence, Union
 
 import numpy as np
+from typing_extensions import Self
 
 from autogluon.timeseries.dataset import TimeSeriesDataFrame
 from autogluon.timeseries.metrics.abstract import TimeSeriesScorer
@@ -58,6 +60,15 @@ class ArrayBasedTimeSeriesEnsembleModel(AbstractTimeSeriesEnsembleModel, ABC):
             "isotonization": "sort",
             "detect_and_ignore_failures": True,
         }
+
+    @classmethod
+    def load(cls, path: str, reset_paths: bool = True, load_oof: bool = False, verbose: bool = True) -> Self:
+        model = super().load(path=path, reset_paths=reset_paths, load_oof=load_oof, verbose=verbose)
+
+        if reset_paths and model.ensemble_regressor is not None:
+            model.ensemble_regressor.set_path(os.path.join(model.path, "ensemble_regressor"))
+
+        return model
 
     @staticmethod
     def to_array(df: TimeSeriesDataFrame) -> np.ndarray:
--- a/timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/abstract.py
+++ b/timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/abstract.py
@@ -8,6 +8,9 @@ class EnsembleRegressor(ABC):
     def __init__(self, *args, **kwargs):
         pass
 
+    def set_path(self, path: str) -> None:
+        pass
+
     @abstractmethod
     def fit(
         self,
--- a/timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/tabular.py
+++ b/timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/tabular.py
@@ -29,6 +29,9 @@ class TabularEnsembleRegressor(EnsembleRegressor):
         self.tabular_hyperparameters = tabular_hyperparameters or {}
         self.predictor: Optional[TabularPredictor] = None
 
+    def set_path(self, path: str) -> None:
+        self.path = path
+
     def fit(
         self,
         base_model_mean_predictions: np.ndarray,


timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/per_quantile_tabular.py
```diff
@@ -0,0 +1,128 @@
...
+            verbosity=1,
+        ).fit(
+            mean_df,
+            hyperparameters=self.tabular_hyperparameters,
+        )
+
+        # fit quantile predictors, each quantile predictor is based on the
+        # estimates of that quantile from base models
+        for i, quantile in enumerate(self.quantile_levels):
+            q_df = self._get_feature_df(base_model_quantile_predictions, i)
```
abdulfatir, line 64: Is it a given that `quantile_levels` will always be in a given (sorted?) order?

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/per_quantile_tabular.py
```diff
@@ -0,0 +1,128 @@
...
+        num_windows, num_items, prediction_length = base_model_mean_predictions.shape[:3]
+        target = labels.reshape(num_windows * num_items * prediction_length).ravel()
+
+        # fit mean predictor, based on mean predictions of base models
+        mean_df = self._get_feature_df(base_model_mean_predictions, 0)
+        mean_df["target"] = target
+        self.mean_predictor = TabularPredictor(
+            label="target",
+            path=os.path.join(self.path, "mean"),
+            verbosity=1,
```
abdulfatir, line 54: Maybe also specify problem_type here to be explicit?

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/per_quantile_tabular.py
```diff
@@ -0,0 +1,128 @@
...
+            self.quantile_predictors.append(predictor)
+
+        return self
+
+    def _get_feature_df(self, predictions: np.ndarray, index: int) -> pd.DataFrame:
+        num_windows, num_items, prediction_length, _, num_models = predictions.shape
+        num_tabular_items = num_windows * num_items * prediction_length
+
+        df = pd.DataFrame(
+            predictions[:, :, :, index].reshape(num_tabular_items, num_models),
```
abdulfatir, line 82: The order is fixed I presume? mean first and then quantiles?

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/per_quantile_tabular.py
```diff
@@ -0,0 +1,128 @@
...
+            except FileNotFoundError:
+                raise ValueError("Model must be fitted before loading for prediction")
+
+    def predict(
+        self, base_model_mean_predictions: np.ndarray, base_model_quantile_predictions: np.ndarray
+    ) -> tuple[np.ndarray, np.ndarray]:
+        self.load_predictors()
+
+        num_windows, num_items, prediction_length, _, _ = base_model_mean_predictions.shape
+        assert num_windows == 1, "Prediction expects a single window to be provided"
```
abdulfatir, line 107: Why is this the case? I presume there will be models later where num_windows can be > 1?

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/per_quantile_tabular.py
```diff
@@ -0,0 +1,128 @@
...
+            quantile_predictions_list.append(
+                predictor.predict(self._get_feature_df(base_model_quantile_predictions, i), as_pandas=False).reshape(
+                    num_windows, num_items, prediction_length
+                )
+            )
+        quantile_predictions = np.stack(quantile_predictions_list, axis=-1)
+
+        return mean_predictions, quantile_predictions
+
+    def __getstate__(self):
```
abdulfatir, line 128: What happens if someone pickles after calling fit?

timeseries/tests/unittests/models/ensemble/array_based/test_models.py
```diff
@@ -226,3 +225,157 @@ def test_given_fitted_tabular_ensemble_when_deleted_and_loaded_then_can_predict(
...
+        assert model.ensemble_regressor.path == regressor_path
+
+        # Check mean predictor path
+        assert os.path.exists(os.path.join(regressor_path, "mean"))
+
+        # Check quantile predictor paths
+        for quantile in model.quantile_levels:
+            assert os.path.exists(os.path.join(regressor_path, f"quantile_{quantile}"))
+
+    def test_given_per_quantile_ensemble_when_fitted_then_separate_predictors_created(self, ensemble_data, tmp_path):
```
abdulfatir, line 286: Is this a bit overkill? 😄

abdulfatir (approved):
Looks good! Thanks @canerturkmen. Left some minor comments.

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/per_quantile_tabular.py
canerturkmen: yup, enforced by the predictor: https://github.com/autogluon/autogluon/blob/35328b48062f0c6e8cedcad3d0e7b55b6a499172/timeseries/src/autogluon/timeseries/predictor.py#L211

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/per_quantile_tabular.py
canerturkmen: in this method I believe it doesn't matter. as it's called separately with mean and quantile arrays.

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/per_quantile_tabular.py
canerturkmen: all of the user-facing methods have the contract for predicting single-window at a time. so I don't think this constraint will go away very easily.

timeseries/tests/unittests/models/ensemble/array_based/test_models.py
canerturkmen: preparing for an agentic future!!

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/per_quantile_tabular.py
canerturkmen: `TabularPredictor` anyway saves itself with all of its constituent models. This class constructs the paths to the predictors again on the fly and lazy-loads them in `load_predictors`. So it doesn't need to unnecessarily re-pickle predictor objects.

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/per_quantile_tabular.py
```diff
@@ -0,0 +1,133 @@
...
+        """Fit separate TabularPredictor for mean and each quantile level."""
+        # TODO: implement time_limit
+
+        num_windows, num_items, prediction_length = base_model_mean_predictions.shape[:3]
+        target = labels.reshape(num_windows * num_items * prediction_length).ravel()
+
+        # fit mean predictor, based on mean predictions of base models
+        mean_df = self._get_feature_df(base_model_mean_predictions, 0)
+        mean_df["target"] = target
+        self.mean_predictor = TabularPredictor(
```
shchur, line 51: Sorry for being late to the party, but could we consider using a single `AbstractModel` here instead of the `TabularPredictor`? I would follow the `model_name: str` + `model_hyperparameters: dict[str, Any]` pattern for configuring the individual model like we do in tabular models (since v1.4) and in the covariate regressor (https://auto.gluon.ai/stable/tutorials/timeseries/forecasting-model-zoo.html#autogluon.timeseries.models.DirectTabularModel). I would store the model object as an attribute of the ensemble model ([like here](https://github.com/autogluon/autogluon/blob/7496bbe4475efb2f3b6baeaff1b5081157057602/timeseries/src/autogluon/timeseries/regressor.py#L139-L150)) and set `path=""` to avoid creating new folders on the disk.

Tabular predictor does a lot of stuff besides just training the model and saves many files to the disk. I think it's an overkill, especially if we train multiple models for different quantiles.

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/per_quantile_tabular.py
```diff
@@ -0,0 +1,133 @@
...
+        # estimates of that quantile from base models
+        for i, quantile in enumerate(self.quantile_levels):
+            q_df = self._get_feature_df(base_model_quantile_predictions, i)
+            q_df["target"] = target
+
+            predictor = TabularPredictor(
+                label="target",
+                path=os.path.join(self.path, f"quantile_{quantile}"),
+                verbosity=1,
+                problem_type="regression",
```
shchur, line 71: I'm a bit confused here, shouldn't this be `problem_type="quantile"`?

shchur: Also, if we switch to individual models instead of using the full predictor, we will need to set the `quantile_levels` hyperparameter via 
https://github.com/autogluon/autogluon/blob/7496bbe4475efb2f3b6baeaff1b5081157057602/timeseries/src/autogluon/timeseries/models/autogluon_tabular/mlforecast.py#L628-L630

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/per_quantile_tabular.py
canerturkmen: Sure. Pushed it to the backlog. Thanks!

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/per_quantile_tabular.py
canerturkmen: https://github.com/autogluon/autogluon/pull/5406/files

PR #5399, opened 2025-11-15
Decouple and enable module-wise installation

prateekdesai04:
*Issue #, if available:*

*Description of changes:*


.github/workflows/pythonpublish.yml
```diff
@@ -36,10 +36,26 @@ jobs:
           for v in core common features tabular multimodal timeseries autogluon
           do
             cd "$v"/
+
```
Innixma, line 39: revert so it is `for v in common core features` ...

PR #5401, opened 2025-11-17
[timeseries] add linear stacker, refactor timer

canerturkmen:
*Issue #, if available:*

*Description of changes:*

- add linear stacker model
- refactor array based ensemble model tests to increase coverage and add new tests
- refactor timer

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/linear_stacker.py
+++ b/timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/linear_stacker.py
@@ -31,10 +31,8 @@ class LinearStackerEnsembleRegressor(EnsembleRegressor):
         Learning rate for Adam optimizer. Defaults to 0.1.
     max_epochs
         Maximum number of training epochs. Defaults to 10000.
-    tolerance_change
-        Convergence tolerance for loss change between epochs. Defaults to 1e-7.
-    tolerance_grad
-        Convergence tolerance for maximum gradient magnitude. Defaults to 1e-6.
+    relative_tolerance
+        Convergence tolerance for relative loss change between epochs. Defaults to 1e-7.
     """
 
     def __init__(
--- a/timeseries/tests/unittests/models/ensemble/array_based/test_models.py
+++ b/timeseries/tests/unittests/models/ensemble/array_based/test_models.py
@@ -22,7 +22,7 @@ MODEL_TO_REGRESSOR = {
 
 
 class TestEnsembleModels:
-    @pytest.fixture(params=[MedianEnsemble, TabularEnsemble, PerQuantileTabularEnsemble, LinearStackerEnsemble])
+    @pytest.fixture(params=list(MODEL_TO_REGRESSOR.keys()))
     def model_class(self, request):
         yield request.param
 


timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/linear_stacker.py
```diff
@@ -0,0 +1,180 @@
...
+        def _wql(
+            labels_tensor: torch.Tensor,
+            ensemble_predictions: torch.Tensor,
+        ) -> torch.Tensor:
+            """Compute the weighted quantile loss on predictions and ground truth (labels).
+            Considering that the first dimension of predictions is the mean, we treat
+            mean predictions on the same footing as median (0.5) predictions as contribution
+            to the overall weighted quantile loss.
+            """
+            quantile_levels = torch.tensor([0.5] + self.quantile_levels, dtype=torch.float64)
```
shchur: Very nice solution for effectively using the same weight for the `"mean"` prediction!

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/linear_stacker.py
```diff
@@ -0,0 +1,180 @@
...
+    def fit(
+        self,
+        base_model_mean_predictions: np.ndarray,
+        base_model_quantile_predictions: np.ndarray,
+        labels: np.ndarray,
+        time_limit: Optional[float] = None,
+    ) -> Self:
+        import torch
+
+        def _wql(
```
shchur: This is not really wql (no normalization by total target sum - and we don't need that). Should we rename this to just `ql` or `quantile_loss`?

Related, how about we normalize the loss per item by the average target value during the horizon (clipped e.g. to at least 1)? Otherwise I can imagine that having terms with extremely different scales can mess up the optimization.

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/linear_stacker.py
```diff
@@ -0,0 +1,180 @@
...
+            return torch.sum(quantile_loss, dim=-1).mean()
+
+        timer = Timer(time_limit).start()
+
+        base_model_predictions = torch.tensor(
+            np.concatenate(
+                [base_model_mean_predictions, base_model_quantile_predictions],
+                axis=3,
+            ),
+            dtype=torch.float64,
```
shchur: Let's use float32 everywhere instead of float64? I don't think there is any benefit for using double here.

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/linear_stacker.py
```diff
@@ -0,0 +1,180 @@
...
+            optimizer.zero_grad()
+
+            ensemble_predictions = weighted_average(base_model_predictions)
+
+            loss = _wql(labels_tensor, ensemble_predictions)
+            loss.backward()
+            optimizer.step()
+
+            loss_change = abs(last_loss - loss.item()) / loss.item()
+            if loss_change < self.tolerance_change and max_grad() < self.tolerance_grad:
```
shchur: Do we really need the `max_grad()` check here? I am a bit concerned that gradients might be out of date / missing after we called `optimizer.step()`. Also not sure about using the gradient magnitude as a stopping criterion in general.

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/linear_stacker.py
```diff
@@ -0,0 +1,180 @@
...
+        for _ in range(self.max_epochs):
+            optimizer.zero_grad()
+
+            ensemble_predictions = weighted_average(base_model_predictions)
+
+            loss = _wql(labels_tensor, ensemble_predictions)
+            loss.backward()
+            optimizer.step()
+
+            loss_change = abs(last_loss - loss.item()) / loss.item()
```
shchur: Let's add an epsilon to avoid division by zero?
```python
            loss_change = abs(last_loss - loss.item()) / (loss.item() + 1e-8)
```

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/linear_stacker.py
```diff
@@ -0,0 +1,180 @@
...
+            if loss_change < self.tolerance_change and max_grad() < self.tolerance_grad:
+                break
+            last_loss = loss.item()
+
+            if timer.timed_out():
+                break
+
+        # store final weights as numpy array
+        with torch.no_grad():
+            self.weights = weighted_average.get_normalized_weights().detach().numpy()
```
shchur, line 146: Let's at least add a TODO to perform sparsification of the models based on the assigned weights? Otherwise the inference can be pretty slow if we need to predict with all L1 models.

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/linear_stacker.py
```diff
@@ -0,0 +1,180 @@
...
+        1e-7.
+    tolerance_grad
+        Convergence tolerance for maximum gradient magnitude. Defaults to
+        1e-6.
+    """
+
+    def __init__(
+        self,
+        quantile_levels: list[float],
+        weights_per: str = "m",
```
shchur: Nit: Literal?

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/linear_stacker.py
```diff
@@ -0,0 +1,180 @@
...
+        ) -> torch.Tensor:
+            """Compute the weighted quantile loss on predictions and ground truth (labels).
+            Considering that the first dimension of predictions is the mean, we treat
+            mean predictions on the same footing as median (0.5) predictions as contribution
+            to the overall weighted quantile loss.
+            """
+            quantile_levels = torch.tensor([0.5] + self.quantile_levels, dtype=torch.float64)
+            error = labels_tensor - ensemble_predictions  # (num_windows, num_items, num_time, num_outputs)
+            quantile_loss = torch.maximum(quantile_levels * error, (quantile_levels - 1) * error)
+            return torch.sum(quantile_loss, dim=-1).mean()
```
shchur: Why not just `torch.mean(quantile_loss)`?

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/linear_stacker.py
```diff
@@ -0,0 +1,180 @@
...
+    prediction.
+
+    Parameters
+    ----------
+    quantile_levels
+        List of quantile levels for quantile predictions (e.g.,
+        [0.1, 0.5, 0.9]).
+    weights_per
+        Weight configuration specifying which dimensions to learn weights
+        for:
```
shchur: Nitissimo: We use line-length=119 in AG-TS, could you remove the line breaks to make them docstrings follow this line length too?

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/linear_stacker.py
```diff
@@ -0,0 +1,180 @@
...
+from .abstract import EnsembleRegressor
+
+
+class LinearStackerEnsembleRegressor(EnsembleRegressor):
+    """Linear stacker ensemble regressor using PyTorch optimization with
+    softmax weights.
+
+    Implements weighted averaging of base model predictions with learnable
+    weights optimized via gradient descent. Uses PyTorch during training
+    for optimization, then stores weights as numpy arrays for efficient
```
shchur: Nice!

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/linear_stacker.py
shchur: If we drop the max_grad check, maybe also rename `tolerance_change` -> `relative_tolerance`?

timeseries/src/autogluon/timeseries/utils/timer.py
```diff
@@ -4,23 +4,53 @@
 from typing_extensions import Self
 
 
-class SplitTimer:
+class Timer:
```
shchur, line 7: We should have added this class a long time ago 😄

timeseries/src/autogluon/timeseries/utils/timer.py
```diff
@@ -4,23 +4,53 @@
...
+            return None
+        return self.time_limit - (time.monotonic() - self.start_time)
+
+    def timed_out(self) -> bool:
+        if self.time_limit is None:
+            return False
+        return self.time_elapsed() >= self.time_limit
+
+
+class SplitTimer(Timer):
```
shchur, line 38: As a side note, I think a docstring for `SplitTimer` would be nice since it's not 100% clear from the method names / API how this object should be used.

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/linear_stacker.py
canerturkmen: @nathanaelbosch !

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/linear_stacker.py
canerturkmen: added a todo. however, if the weights aren't sparse for the same models across different quantiles/horizon/etc., does this really help?

timeseries/src/autogluon/timeseries/utils/timer.py
canerturkmen: sure. I'll come back to this later.

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/linear_stacker.py
canerturkmen: thanks! let's also punt the exploration of other variations to another PR maybe.

timeseries/tests/unittests/models/ensemble/array_based/test_linear_stacker.py
```diff
@@ -0,0 +1,108 @@
...
+            quantile_levels=sample_data["quantile_levels"],
+            weights_per=weights_per,
+            max_epochs=5,
+        )
+
+        regressor.fit(
+            base_model_mean_predictions=sample_data["mean_predictions"],
+            base_model_quantile_predictions=sample_data["quantile_predictions"],
+            labels=sample_data["labels"],
+        )
```
shchur, line 62: Not sure how long these tests are taking, but should we set a time limit (e.g. 5 seconds) to make sure that they don't take up too much time?

timeseries/tests/unittests/models/ensemble/array_based/test_models.py
```diff
@@ -1,301 +1,49 @@
...
+    PerQuantileTabularEnsemble: PerQuantileTabularEnsembleRegressor,
+    LinearStackerEnsemble: LinearStackerEnsembleRegressor,
+}
 
 
 class TestEnsembleModels:
-    @pytest.fixture(params=[MedianEnsemble, TabularEnsemble, PerQuantileTabularEnsemble])
-    def model(self, request):
-        yield request.param()
+    @pytest.fixture(params=[MedianEnsemble, TabularEnsemble, PerQuantileTabularEnsemble, LinearStackerEnsemble])
```
shchur: Nit: How about the following so that it's easier to add new models in the future?

```suggestion
    @pytest.fixture(params=list(MODEL_TO_REGRESSOR.keys()))
```

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/linear_stacker.py
```diff
@@ -0,0 +1,169 @@
...
+        - "mtq": Per-model, per-time, and per-quantile weights
+          (shape: prediction_length, num_quantiles+1, num_models)
+    lr
+        Learning rate for Adam optimizer. Defaults to 0.1.
+    max_epochs
+        Maximum number of training epochs. Defaults to 10000.
+    tolerance_change
+        Convergence tolerance for loss change between epochs. Defaults to 1e-7.
+    tolerance_grad
+        Convergence tolerance for maximum gradient magnitude. Defaults to 1e-6.
```
shchur: The docstring is out-of-date

timeseries/tests/unittests/models/ensemble/array_based/test_linear_stacker.py
canerturkmen: controlled by epochs above. time is negligible.

PR #5409, opened 2025-11-18
[WIP] Add AbstractTorchModel

### LINKED ISSUE #5381: [BUG] Mitra fails to load when loading a Linux-trained model on Windows
**Bug Report Checklist**
I confirmed the bug exists on the latest stable version (AutoGluon 1.4.0) 

**Describe the bug**
A model trained on Linux cannot be loaded and used for prediction on Windows.
Even after manually editing learner.pkl to fix the model path, AutoGluon still raises a FileNotFoundError pointing to the original Linux path.

FileNotFoundError: [Errno 2] No such file or directory: '/root/AutogluonModels/ag-20251105_091414/models/Mitra_BAG_L1/S1F1\\model.pt'

**Expected behavior**
AutoGluon should be able to correctly load a model trained on Linux and run predictions on Windows,
or provide an official way to reset model paths when moving between environments.

**Installed Versions**
AutoGluon Version: 1.4.0

Innixma:
*Description of changes:*

- Add AbstractTorchModel to support training on GPU -> loading on a CPU-only machine seamlessly by avoiding saving weights/tensors with `torch.device("cuda")` which crashes on load for CPU-only instances.
- Update TabM, TabPFNv2, & Mitra to use AbstractTorchModel
- Fixed bug in MItraModel where `reset_paths=False` was set instead of `reset_paths=True` for `.load`. This caused cross-machine loading to fail. ( Resolves #5381 )
- Fixed TabM editing torch threads during fit.
- Added global state consistency checker to unit tests to verify that the outer scope isn't being edited when fitting models.

Fixed GPU -> CPU model load:
- [x] TabM
- [x] TabPFNv2
- [x] Mitra
- [x] TabICL

TODO:
- [ ] RealMLP -> Possibly challenging, David Holzmueller plans to investigate if this can be supported in pytabkit.
- [ ] Stretch: Consider clearing cuda cache after fit to avoid vram leaks (suggestion from Lennart).
- [ ] Stretch: Special VRAM memory estimate logic?


Innixma:
Recreating with #5434 in `autogluon` repo in order to ensure CI works properly with version dependencies

PR #5416, opened 2025-11-22
[timeseries] document timer utility and update tests

canerturkmen:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/timeseries/src/autogluon/timeseries/utils/timer.py
+++ b/timeseries/src/autogluon/timeseries/utils/timer.py
@@ -87,7 +87,7 @@ class SplitTimer(Timer):
 
     Extends Timer to divide the total time limit across a specified number of rounds,
     useful for allocating time budgets to sequential operations. At each call of
-    ``round_time_remaining``, the timer re-distributes the remaining time evenly among
+    ``next_round``, the timer re-distributes the remaining time evenly among
     the remaining rounds.
 
     Parameters
@@ -100,14 +100,16 @@ class SplitTimer(Timer):
 
     Examples
     --------
-    Split time equally across 3 rounds:
+    Split time across 3 rounds:
 
-    >>> timer = SplitTimer(time_limit=10.0, rounds=2).start()
-    >>> time_round_1 = timer.round_time_remaining()  # Returns ~5.0
-    >>> # Do work for round 1, e.g., for 7 seconds
-    >>> timer.split()
-    >>> time_round_2 = timer.round_time_remaining()  # Returns remaining time = ~3 seconds
+    >>> timer = SplitTimer(time_limit=10.0, rounds=3).start()
+    >>> time_round_1 = timer.round_time_remaining()  # Returns ~3.33
+    >>> # Do work for round 1
+    >>> timer.next_round()
+    >>> time_round_2 = timer.round_time_remaining()  # Returns remaining time divided by 2
     >>> # Do work for round 2
+    >>> timer.next_round()
+    >>> time_round_3 = timer.round_time_remaining()  # Returns all remaining time
     """
 
     def __init__(


timeseries/src/autogluon/timeseries/utils/timer.py
```diff
@@ -14,57 +48,101 @@ def __init__(
...
+        Total time limit to split across all rounds. If None, `get` returns None.
+    rounds
+        Number of rounds to split the time across. Default is 1.
+
+    Examples
+    --------
+    Split time equally across 3 rounds:
+
+    >>> timer = SplitTimer(time_limit=30.0, rounds=3).start()
+    >>> for time_for_round in timer:
```
shchur: Should this example be removed from the docstring now that `__iter__` method is removed?

timeseries/src/autogluon/timeseries/utils/timer.py
```diff
@@ -14,57 +48,101 @@ def __init__(
...
         self,
         time_limit: Optional[float],
         rounds: int = 1,
-        strict_stop: bool = False,
     ):
         super().__init__(time_limit)
         self.rounds = rounds
-        self.strict_stop = strict_stop
 
         self.round_index = 0
```
shchur: Nit: Maybe current_round is more clear?

timeseries/src/autogluon/timeseries/utils/timer.py
```diff
@@ -14,57 +48,101 @@ def __init__(
...
+        """Whether the timer has timed out. If ``time_limit`` is None, this method
+        always returns False."""
+        if self.start_time is None:
+            raise RuntimeError("Timer has not been started")
         if self.time_limit is None:
             return False
         return self.time_elapsed() >= self.time_limit
 
 
 class SplitTimer(Timer):
```
shchur, line 85: I am not sure if inheritance makes sense here given that the `SplitTimer` is only used via `get()` and `split()` methods, and the meaning of `timed_out()` and `time_remaining()` methods is ambiguous for `SplitTimer` ("timed out for the current round or overall?"). 

What do you think about just defining a separate class for `SplitTimer` that only offers two methods: 
- get time left for the current round
- increment the current round

Something like:

```python
class Timer:
    def __init__(self, time_limit: Optional[float]):
        self.time_limit = time_limit
        self.start_time = time.monotonic()

    def time_elapsed(self) -> float:
        return time.monotonic() - self.start_time

    def time_remaining(self) -> Optional[float]:
        if self.time_limit is None:
            return None
        return self.time_limit - (time.monotonic() - self.start_time)

    def timed_out(self) -> bool:
        if self.time_limit is None:
            return False
        return self.time_elapsed() >= self.time_limit


class SplitTimer:
    def __init__(self, time_limit: Optional[float], rounds: int = 1):
        self.current_round = 0
        self.rounds = rounds
        self._timer = Timer(time_limit=time_limit)

    def time_per_round(self) -> Optional[float]:
        if self._timer.time_limit is None:
            return None
        remaining_rounds = self.rounds - self.current_round
        if remaining_rounds <= 0:
            return self._timer.time_remaining()
        return self._timer.time_remaining() / remaining_rounds

    def next_round(self) -> Self:
        self.current_round += 1
        return self
```

shchur: Some other nits/optional items to consider:
1. Set `start_time = time.monotonic()` at creation of the timer, since it looks like currently we are anyway always calling `start()` every time we create the object.
2. Rename the methods `get` -> `time_per_round` and `split` -> `next_round`, I was initially a bit confused by the current names since they sound rather generic.
3. Make the naming consistent for "round" or "split" (e.g. `time_per_split`, `current_split`, `next_split` or `time_per_round`, `current_round`, `next_round`).

timeseries/src/autogluon/timeseries/utils/timer.py
canerturkmen: thanks!

PR #5418, opened 2025-11-24
Fix permissions for platform_tests action

shchur:
*Issue #, if available:*

*Description of changes:*
- Undo the error introduced in #5351 that excessively narrowed down the permissions for the platform tests action

### DIFF
--- a/.github/workflows/platform_tests-command.yml
+++ b/.github/workflows/platform_tests-command.yml
@@ -20,11 +20,6 @@ on:
         description: 'Get info of forked repository and branch'
         required: false
 
-permissions:
-  contents: read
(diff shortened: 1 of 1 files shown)


.github/workflows/platform_tests-command.yml
```diff
@@ -20,10 +20,6 @@ on:
         description: 'Get info of forked repository and branch'
         required: false
 
-permissions:
-  contents: read
-  pull-requests: write
-
```
shchur, line 26: I want to merge this for testing purposes. If having this change committed to `master` does not fix the platform tests, this will mean that the problem is with the permission scope on our security tokens.

PR #5422, opened 2025-11-25
[timeseries] extend EnsembleComposer to multilayer stacking

canerturkmen:
*Issue #, if available:*

*Description of changes:*

Extends EnsembleComposer to handle multilayer stacking. 

TODOS:
- [x] discussion on naming format for ensemble models, and implementation. currently, names are set as they would be set if every ensemble model was layer 1.
- [x] implement time limit with SplitTimer. discussion: the plan is to divide split time equally among all ensembles. If time limit is exceeded, we have fit the models that are in lower layers. 
- [x] ensure that double calls to `.save()` per ensemble model do not cause issues with serde.
- [x] revise and update docstrings
- [x] compute `val_score`s on the fly during training, on the correct window.
- [x] rename all references to `level` to `layer`
- [x] raise if performance weighted ensemble is fit before the last layer.

### DIFF
--- a/timeseries/src/autogluon/timeseries/trainer/ensemble_composer.py
+++ b/timeseries/src/autogluon/timeseries/trainer/ensemble_composer.py
@@ -2,7 +2,8 @@ import logging
 import os
 import time
 import traceback
-from typing import Iterator
+from pathlib import Path
+from typing import Any, Iterator
 
 import networkx as nx
 import numpy as np
@@ -10,7 +11,12 @@ from typing_extensions import Self
 
 from autogluon.timeseries import TimeSeriesDataFrame
 from autogluon.timeseries.metrics import TimeSeriesScorer
-from autogluon.timeseries.models.ensemble import AbstractTimeSeriesEnsembleModel, get_ensemble_class
+from autogluon.timeseries.models.ensemble import (
+    AbstractTimeSeriesEnsembleModel,
+    PerformanceWeightedEnsemble,
+    get_ensemble_class,
+)
+from autogluon.timeseries.utils.timer import SplitTimer
 from autogluon.timeseries.utils.warning_filters import warning_filter
 
 from .utils import log_scores_and_times
@@ -19,17 +25,59 @@ logger = logging.getLogger("autogluon.timeseries.trainer")
 
 
 class EnsembleComposer:
-    """Helper class for TimeSeriesTrainer to build multi-layer stack ensembles."""
+    """Helper class for TimeSeriesTrainer to build multi-layer stack ensembles.
+
+    This class depends on the trainer to provide the necessary initialization parameters, training
+    and validation data, as well as having fit the base (non-ensemble) models and persisted their
+    out-of-fold predictions which will be used for ensemble training.
+
+    Parameters
+    ----------
+    path
+        Path of the calling TimeSeriesTrainer. EnsembleComposer finds the model objects and their
+        out-of-fold prediction artifacts with respect to this path. EnsembleComposer only saves
+        ensemble models and their out-of-fold predictions to this folder (i.e., does not pickle
+        itself).
+    prediction_length
+        Number of time steps to forecast.
+    eval_metric
+        Metric used to evaluate ensemble performance.
+    target
+        Name of the target column in the time series data.
+    num_windows_per_layer
+        Number of windows used for training each ensemble layer. Length must match the number of layers
+        in ensemble_hyperparameters. Example: (3, 2) means first layer uses 3 windows, second layer uses
+        2 windows.
+
+        Base models must have OOF predictions saved for all sum(num_windows_per_layer) windows, prior
+        to this class being called.
+    ensemble_hyperparameters
+        Ensemble configuration. A list of dicts, one per layer. If an ensemble model should be fitted
+        with multiple hyperparameter configurations, a list of dicts may be provided as the value.
+        Each layer's dict maps ensemble names to either a single hyperparameter dict or a list of
+        hyperparameter dicts.
+
+        Examples:
+        - ``[{"GreedyEnsemble": {}}, {"GreedyEnsemble": {}}]`` for 2 layers of greedy ensembles.
+        - ``[{"GreedyEnsemble": [{"ensemble_size": 10}, {"ensemble_size": 20}]}]`` for a single layer of
+          two greedy ensembles, with differing ensemble sizes.
+    quantile_levels
+        Quantile levels for probabilistic forecasting.
+    model_graph
+        Directed graph containing base models and their metadata (val_score, fit_time, etc.). Only
+        base models (nodes without predecessors) are used for ensemble training.
+    """
 
     def __init__(
         self,
-        path,
+        path: str,
         prediction_length: int,
         eval_metric: TimeSeriesScorer,
         target: str,
+        num_windows_per_layer: tuple[int, ...],
+        ensemble_hyperparameters: list[dict[str, dict | list[dict]]],
         quantile_levels: list[float],
         model_graph: nx.DiGraph,
-        ensemble_hyperparameters: dict,
     ):
         self.eval_metric = eval_metric
         self.path = path
@@ -37,6 +85,15 @@ class EnsembleComposer:
         self.target = target
         self.quantile_levels = quantile_levels
 
+        self.num_windows_per_layer = num_windows_per_layer
+        self.num_layers = len(num_windows_per_layer)
+
+        if len(ensemble_hyperparameters) != self.num_layers:
+            raise ValueError(
+                "Number of ensemble_hyperparameters must match the number of layers. "
+                f"Received {len(ensemble_hyperparameters)} ensemble_hyperparameters, "
+                f"but {self.num_layers} layers."
+            )
         self.ensemble_hyperparameters = ensemble_hyperparameters
 
         self.banned_model_names = list(model_graph.nodes)
@@ -44,8 +101,9 @@ class EnsembleComposer:
 
     @staticmethod
     def _get_base_model_graph(source_graph: nx.DiGraph) -> nx.DiGraph:
-        """Return a model graph by copying only base models (nodes without predecessors)
-        This ensures we start fresh for ensemble building.
+        """Return a model graph by copying only base models (nodes without predecessors).
+
+        This ensures we start fresh for training ensembles.
         """
         rootset = EnsembleComposer._get_rootset(source_graph)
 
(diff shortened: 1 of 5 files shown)


timeseries/src/autogluon/timeseries/trainer/ensemble_composer.py
```diff
@@ -19,33 +20,85 @@
...
-        ensemble_hyperparameters: dict,
     ):
         self.eval_metric = eval_metric
         self.path = path
         self.prediction_length = prediction_length
         self.target = target
         self.quantile_levels = quantile_levels
 
+        self.num_windows_per_layer = num_windows_per_layer
+        self.num_stack_levels = len(num_windows_per_layer)
```
shchur: Do we plan to expose `num_stack_levels` in the `predictor.fit` API, as we discussed in the doc? Imo it's not necessary since `ensemble_hyperparameters` and `num_val_windows: tuple[int]` provide a sufficient sanity check to prevent accidental misconfiguration (just wanted to flag this here).

timeseries/src/autogluon/timeseries/trainer/ensemble_composer.py
```diff
@@ -59,12 +112,44 @@ def _get_base_model_graph(source_graph: nx.DiGraph) -> nx.DiGraph:
     def _get_rootset(graph: nx.DiGraph) -> list[str]:
         return [n for n in graph.nodes if not list(graph.predecessors(n))]
 
+    def _load_model(self, model_name: str) -> Any:
+        """Load a model from the graph by name."""
+        attrs = self.model_graph.nodes[model_name]
+        model_path = os.path.join(self.path, *attrs["path"])
+        return attrs["type"].load(path=model_path)
+
+    def _iter_models(self, layer: int) -> Iterator[tuple[str, Any]]:
```
shchur, line 126: Let's rename `layer` -> `level` here and in the rest of the methods for consistency with `TimeSeriesTraienr`
https://github.com/autogluon/autogluon/blob/dca6e576b9d5396dff6add318ba18085e70564c4/timeseries/src/autogluon/timeseries/trainer/trainer.py#L230-L234

and the tabular trainer
https://github.com/autogluon/autogluon/blob/dca6e576b9d5396dff6add318ba18085e70564c4/tabular/src/autogluon/tabular/trainer/abstract_trainer.py#L2071

timeseries/src/autogluon/timeseries/trainer/ensemble_composer.py
```diff
@@ -73,17 +158,14 @@ def iter_ensembles(self) -> Iterator[tuple[int, AbstractTimeSeriesEnsembleModel,
...
+        for layer_idx, layer in enumerate(nx.traversal.bfs_layers(self.model_graph, rootset)):
+            if layer_idx == 0:  # we don't need base models
                 continue
 
             for model_name in layer:
-                attrs = self.model_graph.nodes[model_name]
-                model_path = os.path.join(self.path, *attrs["path"])
-                model = attrs["type"].load(path=model_path)
-
+                model = self._load_model(model_name)
```
shchur: This logic seems to be very similar to `_iter_models` above. Could we instead do something like the following?
```python
for level in range(1, self.num_stack_levels + 1):
    for model_name, model in self._iter_models(level=level):
        yield (level, model, list(self.model_graph.predecessors(model_name)))
```

timeseries/src/autogluon/timeseries/trainer/trainer.py
```diff
@@ -52,7 +52,7 @@ def __init__(
         skip_model_selection: bool = False,
         enable_ensemble: bool = True,
         verbosity: int = 2,
-        num_val_windows: Optional[int] = None,
+        num_val_windows: int = 1,
```
shchur, line 55: We can also modify this method since `num_val_windows` is now always an `int`
https://github.com/autogluon/autogluon/blob/dca6e576b9d5396dff6add318ba18085e70564c4/timeseries/src/autogluon/timeseries/trainer/trainer.py#L547-L549

timeseries/src/autogluon/timeseries/trainer/trainer.py
```diff
@@ -555,17 +568,42 @@ def _get_val_splitter(self) -> AbstractWindowSplitter:
...
+        the different layers of stack ensembling when calling EnsembleComposer.
+
+        The key rule is that when ``use_val_data=True`` is provided, it is always used to both train
+        and produce scores for the last layer of ensembles. Let's break down what happens in each
+        scenario, depending on ``num_val_windows`` and ``use_val_data``.
+
+        - If ``num_val_windows=(3,)`` and ``use_val_data=False``, then three windows at the end of
+          ``train_data`` is held out for validation. These same three windows are used to fit L2
+          ensembles by EnsembleComposer.
+        - If the original ``num_val_windows=(3,)`` and ``use_val_data=True``, then no windows from
```
shchur, line 589: Should this be an impossible combination? I.e. if user calls `predictor.fit(train_data, val_data, num_val_windows=3)` (or `num_val_windows=(3,)`), the predictor will automatically set `num_val_windows=0` and forward this value to the trainer / ensemble composer? So base models will be trained up to the end of `train_data` and the ensemble will be fit on the OOF predictions generated on the last window of `val_data`.

timeseries/src/autogluon/timeseries/trainer/trainer.py
```diff
@@ -555,17 +568,42 @@ def _get_val_splitter(self) -> AbstractWindowSplitter:
...
+          the training window will be scored by base models. The base models will instead be scored on
+          the last one window of ``val_data``, and the ensemble will also be fit on this window.
+        - If ``num_val_windows=(3, 2)`` and ``use_val_data``, then the base models will score 5 windows
```
shchur: Typo?
```suggestion
        - If ``num_val_windows=(3, 2)`` and ``use_val_data=False``, then the base models will score 5 windows
```

timeseries/src/autogluon/timeseries/trainer/trainer.py
```diff
@@ -555,17 +568,42 @@ def _get_val_splitter(self) -> AbstractWindowSplitter:
...
+          at the end of ``train_data``, L2 ensembles will be fit on the first 3, and L3 ensembles will
+          be fit on the last 2.
+        - If the original ``num_val_windows=(3, 2)`` and ``use_val_data=True``, base models will record
```
shchur, line 595: After reading this comment I think our understanding is 100% aligned, the only question is what should be documented in this docstring and how to split the operations between `predictor` and `trainer`.

How about we do the following?
- update the signature of `TimeSeriesTrainer.__init__` to expect `num_val_windows: tuple[int, ...]` 
- modify `num_val_windows` in `TimeSeriesPredictor.fit` depending on the provided values of `num_val_windows` and `val_data`, and forward the correct values to `learner.fit` (following the logic described in this docstring)

timeseries/src/autogluon/timeseries/trainer/ensemble_composer.py
```diff
@@ -94,73 +176,173 @@ def fit(
...
+
+        def get_ground_truth_for_layer(layer_idx):
+            window_start = sum(self.num_windows_per_layer[: layer_idx - 1])
+            window_end = window_start + self.num_windows_per_layer[layer_idx - 1]
+            return data_per_window[window_start:window_end]
+
+        for layer_idx in range(1, self.num_stack_levels + 1):
+            layer_input_model_names = [name for name, _ in self._iter_models(layer=layer_idx - 1)]
+            layer_input_model_scores = {
+                name: self.model_graph.nodes[name]["val_score"] for name in layer_input_model_names
```
shchur, line 219: I guess we will also need to update how the `val_score` is computed for the base models and which values are reported in the predictor log (currently it's the average of the scores across `sum(num_windows_per_layer)` windows). Should we store the score per window as model's attribute and compute the correct average here?

timeseries/src/autogluon/timeseries/trainer/ensemble_composer.py
```diff
@@ -94,73 +176,173 @@ def fit(
...
-                # TODO: add ensemble's own time to predict_time
-                ensemble.predict_time = self._calculate_base_models_predict_time(ensemble.model_names)
-
-                log_scores_and_times(
-                    ensemble.val_score,
-                    ensemble.fit_time,
-                    ensemble.predict_time,
-                    eval_metric_name=self.eval_metric.name_with_sign,
-                )
+                    # if this is not the last layer, ensembles should also persist out-of-fold predictions
```
shchur: With the previous code we persisted all OOF predictions for ensembles. E.g. for `num_val_windows=3` and the final `WeightedEnsemble` we re-computed its predictions on the 3 windows and saved them with `cache_oof_predictions` + computed `val_score` on them. 

This way when we called `predictor.backtest_predictions(model="WeightedEnsemble")`, we go the prediction of `WeightedEnsemble` on the `num_val_windows=3` windows that were used for training the `WeightedEnsemble`.

In light of the new logic, would the correct approach be to always score all ensemble models on the last `num_windows_per_layer[-1]` windows?

timeseries/src/autogluon/timeseries/trainer/ensemble_composer.py
```diff
@@ -19,33 +20,85 @@
...
+        to this class being called.
+    ensemble_hyperparameters
+        Ensemble configuration. A list of dicts, one per layer. If an ensemble model should be fitted
+        with multiple hyperparameter configurations, a list of dicts may be provided as the value.
+        Each layer's dict maps ensemble names to either a single hyperparameter dict or a list of
+        hyperparameter dicts.
+
+        Examples:
+        - ``[{"GreedyEnsemble": {}}, {"GreedyEnsemble": {}}]`` for 2 layers of greedy ensembles.
+        - ``[{"GreedyEnsemble": [{"ensemble_size": 10}, {"ensemble_size": 20}]}]`` for a single layer of
```
shchur, line 62: How will the models be named currently with such a config? `GreedyEnsemble` and `GreedyEnsemble_2` or something else?

Regarding the discussion point in the PR description: I think that `GreedyEnsemble_L2` and `GreedyEnsemble_L3` would be a good and consistent way to label L2/L3 ensemble models, but I'm not sure how well this plays with the `_2`, `_3`, ... suffixes for different variants of the same model. I also see that we cannot force "single config per ensemble class in a level" because of the different variants of the `LinearEnsemble` that we might want to combine.

timeseries/src/autogluon/timeseries/trainer/ensemble_composer.py
canerturkmen: makes sense. we can skip this.

timeseries/src/autogluon/timeseries/trainer/ensemble_composer.py
canerturkmen: I'm not sure. Should we rename the trainer instead? We did call it multilayer in the paper, layer is more descriptive.

timeseries/src/autogluon/timeseries/trainer/ensemble_composer.py
canerturkmen: thanks!

timeseries/src/autogluon/timeseries/trainer/ensemble_composer.py
canerturkmen: Let's punt this to the trainer maybe. I wouldn't have this class touch the base models preferably.

timeseries/src/autogluon/timeseries/trainer/trainer.py
canerturkmen: again, this PR is for getting the ensemble composer in place. So I'll punt these items as well.

timeseries/src/autogluon/timeseries/trainer/ensemble_composer.py
canerturkmen: How about
- if it's a single layer ensemble, don't use layer indices
- if not, use `GreedyEnsemble_2_L2` ?

timeseries/src/autogluon/timeseries/trainer/trainer.py
canerturkmen: Yes, this doc will move to predictor at one point. `(3,), False` as well as `(3,2), False` should be impossible combinations at this point.

timeseries/src/autogluon/timeseries/trainer/ensemble_composer.py
canerturkmen: Yes. So we should refactor this to compute `val_score` on the fly and report it in the training log?

timeseries/src/autogluon/timeseries/trainer/ensemble_composer.py
shchur: I have a slight preference towards `level` since it's used consistently across Tabular as well, and semantically both words feel very similar to me

timeseries/src/autogluon/timeseries/trainer/ensemble_composer.py
```diff
@@ -59,108 +117,252 @@ def _get_base_model_graph(source_graph: nx.DiGraph) -> nx.DiGraph:
...
+                        for window_idx in range(self.num_windows_per_layer[pred_layer_idx - 1]):
+                            prediction = ensemble.predict(
+                                {n: pred_base_predictions[n][window_idx] for n in ensemble.model_names}
+                            )
+                            predictions.append(prediction)
+
+                        predict_time = time.monotonic() - predict_time_start
+
+                    # prediction time is last layer's time + base models
+                    ensemble.predict_time = predict_time + self._calculate_base_models_predict_time(
```
shchur, line 257: Should we store the predict time for a single window, not all windows? I think that's the current interpretation.

So something like
```python
ensemble.predict_time = predict_time / len(predictions) + ...
```

timeseries/src/autogluon/timeseries/trainer/ensemble_composer.py
```diff
@@ -59,108 +117,252 @@ def _get_base_model_graph(source_graph: nx.DiGraph) -> nx.DiGraph:
...
-                ensemble.fit_time = time.monotonic() - time_start
 
-                ensemble_predictions_per_window = []
-                score_per_window = []
-                for window_idx, data in enumerate(data_per_window):
-                    predictions = ensemble.predict(
-                        {n: predictions_per_window[n][window_idx] for n in ensemble.model_names}
+                inputs = {}
+                for model_name in model_names:
+                    oof_predictions = self._get_model_oof_predictions(model_name)
```
shchur, line 203: Could we just use `predictions_per_window` instead of accessing the model attributes?

canerturkmen: the original `predictions_per_window` will only contain base model predictions. here, we're actually looking for outputs of ensembles from previous layers.

PR #5431, opened 2025-11-28
[timeseries] upgrade type annotations to Python3.10

canerturkmen:
*Issue #, if available:*

*Description of changes:*

- update ruff language specification to `py310`
- replace all `Union` with the pipe syntax
- replace all `Optional` with `| None`

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/chronos/utils.py
+++ b/timeseries/src/autogluon/timeseries/models/chronos/utils.py
@@ -93,7 +93,7 @@ class ChronosFineTuningDataset(IterableDataset):
         target_column: str = "target",
         context_length: int = 512,
         prediction_length: int = 64,
-        tokenizer: ChronosTokenizer | None = None,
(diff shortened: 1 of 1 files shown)


timeseries/src/autogluon/timeseries/metrics/__init__.py
```diff
@@ -1,7 +1,7 @@
 from __future__ import annotations
```
shchur, line 1: Should we remove these statements as well?

canerturkmen: I think that and Self come into play with 3.11

shchur: Not sure about `Self`, but indeed `from __future__ import annotations` has uses besides using `|` in python <= 3.9 https://peps.python.org/pep-0563. TIL

shchur: Feel free to merge once the CI passes

PR #5436, opened 2025-12-01
[timeseries] fix predict_time computation in ensembles

canerturkmen:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/timeseries/src/autogluon/timeseries/trainer/ensemble_composer.py
+++ b/timeseries/src/autogluon/timeseries/trainer/ensemble_composer.py
@@ -412,7 +412,7 @@ class EnsembleComposer:
     def _calculate_predict_time(self, model: AbstractTimeSeriesEnsembleModel) -> float:
         """Calculate ensemble predict time as sum of base model predict times."""
         assert model.predict_time_marginal is not None
-        predict_time = model.predict_time_marginal or 0
+        predict_time = model.predict_time_marginal
         for model_name in nx.ancestors(self.model_graph, model.name):
             ancestor = self._load_model(model_name)
             if isinstance(ancestor, AbstractTimeSeriesEnsembleModel):


timeseries/src/autogluon/timeseries/trainer/ensemble_composer.py
```diff
@@ -409,9 +409,19 @@ def _get_ensemble_model_name(self, name: str, layer_idx: int) -> str:
             name = f"{base_name}_{increment}" + layer_suffix
         return name
 
-    def _calculate_base_models_predict_time(self, model_names: list[str]) -> float:
+    def _calculate_predict_time(self, model: AbstractTimeSeriesEnsembleModel) -> float:
         """Calculate ensemble predict time as sum of base model predict times."""
-        return sum(self.model_graph.nodes[name]["predict_time"] for name in model_names)
+        assert model.predict_time_marginal is not None
+        predict_time = model.predict_time_marginal or 0
```
shchur: After the assertion above, can we keep this as `predict_time = model.predict_time_marginal`?

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
```diff
@@ -122,12 +122,16 @@ def __init__(
...
-        self.predict_1_time: float | None = (
-            None  # Time taken to predict 1 row of data in seconds (with batch size `predict_1_batch_size`)
-        )
-        self.val_score: float | None = None  # Score with eval_metric (Validation data)
+        # Time taken to fit in seconds (Training data)
+        self.fit_time: float | None = None
+        # Time taken to predict in seconds, for a single prediction horizon on validation data
+        self.predict_time: float | None = None
+        # Time taken to predict 1 row of data in seconds (with batch size `predict_1_batch_size`)
+        self.predict_1_time: float | None = None
```
shchur, line 130: We never use this for time series models, should we just avoid setting this attribute?

canerturkmen: iirc, HPO machinery depends nontrivially on this and therefore I could not remove it.

timeseries/src/autogluon/timeseries/trainer/ensemble_composer.py
canerturkmen: thanks!

timeseries/src/autogluon/timeseries/models/abstract/abstract_timeseries_model.py
shchur: All right, let's keep it then

PR #5441, opened 2025-12-02
Add new preprocessors for tabular data

atschalz:
*Issue #, if available:*

*Description of changes:*
- Add five new generators to improve performance on tabular datasets.
- Each generator inherits from AbstractFeatureGenerator and implementes _fit_transform and _transform functions along with internal logic to apply the actual transformations
- In addition, each preprocessor contains an estimate_no_of_new_features function that is meant to be used for memory estimation

Requires clarification:
- Is it appropriate to return an empty dict() for type_group_map_special in fit_transform?
- Do we need unit tests for the preprocessors and what should be covered?
- The generators are not fully aligned with how other generators work. Currently, each of the new preprocessors returns a concatenation of the unaffected, original data with the new, generated features. In addition some preprocessors have a keep_original parameter if the base features for the encoded ones should be kept. This differs from the current AutoGluon generators, where get_default_infer_features_in_args is used to apply transformations on affected features and generators return only the new features. Instead, the current implementation of the new preprocessors directly returns a dataframe that is ready-to-use for modeling and contains unaffected old as well as new transformed features. Therefore, some changes would be required for integration with the current generators and pipelines logic in AutoGluon.
- Is it appropriate to return an empty dict() in get_default_infer_features_in_args?
- estimate_no_of_new_features (used for memory estimation) currently is pessimistic and computes the worst case scenario, because some preprocessors contain filters making it hard to estimate the actual no. of features without actually computing them. In the worst case, the estimation of how many new features are added can be off by a few hundred features. Could there be downsides to that? 

Todos: 
- Currently only the behavior with default hyperparameters was tested and considered in estimate_no_of_new_features. Need to make sure that all the functionalities work as expected or remove them
- Improve estimate_no_of_new_features functions for preprocessors to return expected new data shape and feature dtypes instead of just the no. of new features
- There are still leftovers from how the preprocessors worked before rewriting them as AG classes. I.e., target_type is a parameter although its not used for some preprocessors.

### DIFF
--- a/common/tests/test_check_style.py
+++ b/common/tests/test_check_style.py
@@ -1,15 +1,57 @@
+import json
 import logging
 import warnings
+from collections import Counter
 from subprocess import PIPE, Popen
 
 
 def test_check_style():
     logging.getLogger().setLevel(logging.INFO)
-    logging.info("PEP8 Style check")
-    flake8_proc = Popen(["flake8", "--count", "--max-line-length", "300"], stdout=PIPE)
-    flake8_out = flake8_proc.communicate()[0]
-    lines = flake8_out.splitlines()
-    count = int(lines[-1].decode())
-    if count > 0:
-        warnings.warn(f"{count} PEP8 warnings remaining")
-    assert count < 1000, "Too many PEP8 warnings found, improve code quality to pass test."
+    logging.info("Ruff style check")
+
+    ruff_proc = Popen(
+        ["ruff", "check", ".", "--exit-zero", "--output-format", "json"],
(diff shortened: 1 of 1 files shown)


Innixma (changes requested):
MAJOR BLOCKER:

We need to ensure we don't go OOM during the memory usage estimate. Currently we make a very large pandas DataFrame during the memory estimate, but we probably have to avoid doing this or be extra conservative and return an extremely large number in case we think doing this operation will lead to OOM.

PR #5447, opened 2025-12-03
[timeseries] Chronos-2: Add support for cross learning

abdulfatir:
*Issue #, if available:*

*Description of changes:*

- Adds `cross_learning` hyperparameter which defaults to `True`.
- Update default fine-tuning batch size. 


timeseries/src/autogluon/timeseries/models/chronos/chronos2.py
```diff
@@ -198,6 +204,7 @@ def _predict(
             quantile_levels=self.quantile_levels,
             batch_size=batch_size,
             validate_inputs=False,
+            predict_batches_jointly=cross_learning,
```
shchur, line 207: Let's also update this after we add `cross_learning` to chronos-forecasting. For now we can merge to unblock benchmarking

PR #5452, opened 2025-12-03
[timeseries] Fix broken link to custom models tutorial in the advanced page

colesussmeier:
When on the advanced page of the time series tutorials (https://auto.gluon.ai/stable/tutorials/timeseries/advanced/index.html), the link in the card to the custom models page is advanced/forecasting-custom-model.html.

This duplicates advanced/ and points to https://auto.gluon.ai/stable/tutorials/timeseries/advanced/advanced/forecasting-custom-model.html. This does not work and leads to an AccessDenied error.

I removed advanced/ from the link in the card in the index.md so now it should point to the right thing.


shchur:
Nice, thanks a lot for the fix @colesussmeier!

shchur (approved):

PR #5456, opened 2025-12-05
[timeseries] Enable multilayer stacking in trainer

canerturkmen:
*Issue #, if available:*

*Description of changes:*

- Update how `val_score` is computed for base models in the Trainer. 
- Update how `MultiWindowBacktestingModel` deals with `val_data`
- Tests, etc.

TODO:
- [x] add tests

### DIFF
--- a/timeseries/tests/unittests/trainer/test_trainer.py
+++ b/timeseries/tests/unittests/trainer/test_trainer.py
@@ -938,3 +938,17 @@ class TestMultilayerEnsembleTraining:
             parents = list(graph.predecessors(ensemble_name))
             assert len(parents) > 0
             assert all(parent in trainer.get_model_names() for parent in parents)
+
+    def test_when_trainer_fit_then_predictions_are_consistent(self, trainer_and_params):
+        trainer, _ = trainer_and_params
+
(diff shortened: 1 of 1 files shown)


timeseries/src/autogluon/timeseries/models/multi_window/multi_window_model.py
```diff
@@ -213,12 +214,25 @@ def score_and_cache_oof(
         store_predict_time: bool = False,
         **predict_kwargs,
     ) -> None:
-        # self.val_score, self.predict_time, self._oof_predictions already saved during _fit()
```
shchur: I had a look at the rest of the code but still not sure why we need to modify the `score_and_cache_oof` method. What is the new case when we need to pass `val_data` to `MultiWindowBacktestingModel.score_and_cache_oof`? 

As I understand, previously there were two branches:
1. Custom `val_data` is provided by the user, `num_val_windows=0` -> in this case `MultiWindowBacktestingModel` is not used.
2. User only provides `train_data` and `num_val_windows > 0` -> in this case `MultiWindowBacktestingModel` is used and it stores the `val_score` during `fit`; `score_and_cache_oof` is a no-op.

PR #5458, opened 2025-12-06
[timeseries] Add 'revision' argument to Chronos models for reproducib…

ryuichi-ichinose:
*Description of changes:*

[timeseries] Add 'revision' argument to Chronos models for reproducibility
This commit adds a revision parameter to both ChronosModel and
Chronos2Model. This allows users to pin specific model versions(commit hashes) from Hugging Face Hub.

Key changes:

- Added revision argument to init and load_model_pipeline.
- Ensured backward compatibility for pickled models using getattr.
- Added unit tests for revision passing and persistence.

This enhancement is crucial for production reproducibility and mitigating supply chain risks.
Even with safetensors ensuring code safety, pinning the exact commit hash is the only way to guarantee model integrity and deterministic behavior against upstream changes.

### DIFF
--- a/timeseries/tests/unittests/models/chronos/test_model.py
+++ b/timeseries/tests/unittests/models/chronos/test_model.py
@@ -507,21 +507,12 @@ def test_when_chronos_bolt_no_fine_tune_with_custom_quantiles_then_original_quan
 def test_when_revision_provided_then_from_pretrained_is_called_with_revision(chronos_model_path):
     model_revision = "my-test-branch"
     model = ChronosModel(
-        hyperparameters={
-            "model_path": chronos_model_path,
-            "revision": model_revision,
-            "device": "cpu",
-        },
+        hyperparameters={"model_path": chronos_model_path, "revision": model_revision, "device": "cpu"},
     )
-
-    # Speed optimization:
-    # We manually set attributes to simulate a configured model state.
-    # Calling fit() is unnecessary overhead just to verify serialization logic.
-    model.device = "cpu"
-    model.torch_dtype = "auto"
 
     with mock.patch("chronos.BaseChronosPipeline.from_pretrained") as mock_from_pretrained:
         mock_from_pretrained.return_value = mock.MagicMock()
+        model.fit(train_data=DUMMY_TS_DATAFRAME)
         model.load_model_pipeline()
 
     mock_from_pretrained.assert_called_once()


shchur:
Hi @ryuichi-ichinose, thanks a lot for your contribution! I have slightly updated the PR to simplify the logic. AutoGluon does not offer compatibility between different released (e.g., predictor trained in AG-v1.4 cannot be loaded in AG-v1.5), so I removed the code required to handle such operations in a backward-compatible manner.

ryuichi-ichinose:
Thanks for refining the PR! I was a bit concerned about backward compatibility, but I agree that simplifying the logic is the better approach for maintainability. Looks much cleaner now.

ryuichi-ichinose:
I apologize for the confusion regarding the test failure.
The test used a "speed optimization" hack (manually setting attributes to skip fit()), which likely made the dependency on revision unclear during the simplification process. The failure is due to this non-standard setup interacting with the refactored logic.
I can submit a quick fix to explicitely handle revision in this test case.

timeseries/tests/unittests/models/chronos/test_model.py
```diff
@@ -502,3 +502,18 @@ def test_when_chronos_bolt_no_fine_tune_with_custom_quantiles_then_original_quan
...
+
+def test_when_revision_provided_then_from_pretrained_is_called_with_revision(chronos_model_path):
+    model_revision = "my-test-branch"
+    model = ChronosModel(
+        hyperparameters={"model_path": chronos_model_path, "revision": model_revision, "device": "cpu"},
+    )
+
+    with mock.patch("chronos.BaseChronosPipeline.from_pretrained") as mock_from_pretrained:
+        mock_from_pretrained.return_value = mock.MagicMock()
+        model.fit(train_data=DUMMY_TS_DATAFRAME)
```
shchur, line 515: This should not introduce any major overhead, and it's safer in case we change what happens inside `fit` in the future.

abdulfatir:
@shchur Overall this looks good to me. I am only wondering about the following scenario:
- User loads the model with a revision specified.
- The model is fine-tuned and saved.
- Upon reloading, will this cause problems when `from_pretrained` is called with a local path and `revision`?

ryuichi-ichinose:
Agreed. Using fit() is indeed more robust against future changes. Thanks for the update.

PR #5459, opened 2025-12-06
[timeseries] enable multi layer ensembles in TimeSeriesPredictor

canerturkmen:
*Issue #, if available:*

*Description of changes:*

depends on #5456 

Enables TimeSeriesPredictor to perform multilayer ensembling.

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/ensemble/__init__.py
+++ b/timeseries/src/autogluon/timeseries/models/ensemble/__init__.py
@@ -6,25 +6,21 @@ from .weighted import GreedyEnsemble, PerformanceWeightedEnsemble, SimpleAverage
 
 def get_ensemble_class(name: str):
     mapping = {
-        "GreedyEnsemble": GreedyEnsemble,
-        "PerItemGreedyEnsemble": PerItemGreedyEnsemble,
-        "PerformanceWeightedEnsemble": PerformanceWeightedEnsemble,
-        "SimpleAverageEnsemble": SimpleAverageEnsemble,
-        "WeightedEnsemble": GreedyEnsemble,  # old alias for this model
-        "MedianEnsemble": MedianEnsemble,
-        "TabularEnsemble": TabularEnsemble,
-        "PerQuantileTabularEnsemble": PerQuantileTabularEnsemble,
-        "LinearStackerEnsemble": LinearStackerEnsemble,
+        "Greedy": GreedyEnsemble,
+        "PerItemGreedy": PerItemGreedyEnsemble,
+        "PerformanceWeighted": PerformanceWeightedEnsemble,
+        "SimpleAverage": SimpleAverageEnsemble,
+        "Weighted": GreedyEnsemble,  # old alias for this model
+        "Median": MedianEnsemble,
+        "Tabular": TabularEnsemble,
+        "PerQuantileTabular": PerQuantileTabularEnsemble,
+        "LinearStacker": LinearStackerEnsemble,
     }
 
-    for full_name, ensemble_class in list(mapping.items()):
-        if full_name.endswith("Ensemble"):
-            alias = full_name[:-8]  # Remove "Ensemble" suffix
-            mapping[alias] = ensemble_class
-
-    if name not in mapping:
+    name_clean = name.removesuffix("Ensemble")
+    if name_clean not in mapping:
         raise ValueError(f"Unknown ensemble type: {name}. Available: {list(mapping.keys())}")
-    return mapping[name]
+    return mapping[name_clean]
 
 
 __all__ = [
--- a/timeseries/src/autogluon/timeseries/trainer/ensemble_composer.py
+++ b/timeseries/src/autogluon/timeseries/trainer/ensemble_composer.py
@@ -227,18 +227,6 @@ class EnsembleComposer:
                     # train the ensemble model
                     time_start = time.monotonic()
 
-                    fit_log_message = f"Training ensemble model {ensemble_name}. "
-                    if self.num_layers > 1:
-                        fit_log_message = fit_log_message[:-2] + (
-                            f" in layer {layer_idx} of {self.num_layers} stacked ensemble layers. "
-                        )
-                    if main_loop_timer.time_remaining() is not None:
-                        fit_log_message += (
-                            f"Training for up to {main_loop_timer.round_time_remaining():.1f}s of the "
-                            f"{main_loop_timer.time_remaining():.1f}s of remaining time. "
-                        )
-                    logger.info(fit_log_message)
-
                     ensemble = self._fit_single_ensemble(
                         model_name=ensemble_name,
                         hyperparameters=ensemble_hp_dict,
@@ -359,6 +347,11 @@ class EnsembleComposer:
         if ensemble.name != old_name:
             path_obj = Path(ensemble.path)
             ensemble.path = str(path_obj.parent / ensemble.name)
+
+        fit_log_message = f"Training ensemble model {ensemble.name}. "
+        if time_limit is not None:
+            fit_log_message += f"Training for up to {time_limit:.1f}s."
+        logger.info(fit_log_message)
 
         with warning_filter():
             ensemble.fit(


timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -619,13 +595,36 @@ def fit(
...
-        num_val_windows : int, default = 1
+        ensemble_hyperparameters : dict or list of dict, optional
+            Hyperparameters for ensemble models. Can be a single dict for one ensemble layer, or a list of dicts
+            for multiple ensemble layers (multi-layer stacking).
+
+            For single-layer ensembling (default)::
+
+                predictor.fit(
+                    ...,
+                    ensemble_hyperparameters={"GreedyEnsemble": {"ensemble_size": 10}},
```
shchur: We discussed this before, but I feel like renaming `WeightedEnsemble` -> `GreedyEnsemble` can be a massive breaking change (and, as a minor nit, departure from Tabular). I could be okay with this change in v2.0, but I am a bit concerned about doing it v1.5. To name a few examples, the users could be accessing the `WeightedEnsemble` in leaderboard, when extracting the ensemble weights from `fit_summary`, using model name in `predictor.predict`.  Should we keep the `WeighedEnsemble` name around for a bit longer?

Does not need to be addressed in this PR, but just something to keep in mind before v1.5.

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -619,13 +595,36 @@ def fit(
...
+                )
+
+            For multi-layer ensembling, provide a list where each element configures one ensemble layer::
+
+                predictor.fit(
+                    ...,
+                    num_val_windows=(2, 3),
+                    ensemble_hyperparameters=[
+                        {"GreedyEnsemble": {"ensemble_size": 5}},  # Layer 1
```
shchur: Imo an L3 model only makes sense if there are > 1 L2 models. Let's update the example here to multiple L2 models?

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -619,13 +595,36 @@ def fit(
...
+                        {"PerformanceWeightedEnsemble": {}},       # Layer 2
+                    ],
+                )
+
+            When using multi-layer ensembling, ``len(ensemble_hyperparameters)`` must match ``len(num_val_windows)``.
```
shchur:
```suggestion
            When using multi-layer ensembling, ``num_val_windows`` must be a tuple of integers, and  ``len(ensemble_hyperparameters)`` must match ``len(num_val_windows)``.
```

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -774,6 +796,99 @@ def fit(
...
+            f"Time series in train_data are too short for chosen num_val_windows={num_val_windows}. "
+            f"Reducing num_val_windows to {max_allowed} total windows."
+        )
+
+        result = list(num_val_windows)
+        for i in range(len(result) - 1, -1, -1):
+            while result[i] > 1 and sum(result) > max_allowed:
+                result[i] -= 1
+            if sum(result) <= max_allowed:
+                break
```
shchur, line 890: Took me some time to understand what's going on here, but after understanding, this is brilliant! 

I would add a comment like the following
> Starting from the last group of windows, reduce number of windows in each group by 1, until `sum(num_val_windows) <= max_allowed` is satisfied.

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -774,6 +796,99 @@ def fit(
...
+
+        if sum(result) > max_allowed:
```
shchur, line 895: I would also explain what's going on here, something like:
> It is possible that the above for loop reduced the number of windows in each group to 1 (i.e. `result = [1] * len(num_val_windows)`), but still `sum(result) > max_allowed`. In this case we set `result = [1] * max_allowed`

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -719,35 +740,35 @@ def fit(
         if val_step_size is None:
             val_step_size = self.prediction_length
 
-        if num_val_windows > 0:
-            num_val_windows = self._reduce_num_val_windows_if_necessary(
-                train_data, original_num_val_windows=num_val_windows, val_step_size=val_step_size
-            )
+        num_val_windows, ensemble_hyperparameters = self._validate_and_normalize_validation_and_ensemble_inputs(
```
shchur, line 743: This method should somehow be aware of whether `tuning_data` is available - in this case the last entry of `num_val_windows` will be set to 1 and will be irrelevant for the length check.

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -774,6 +796,99 @@ def fit(
...
+        val_step_size: int,
+        median_timeseries_length: float,
+    ) -> tuple[int, ...]:
+        if isinstance(num_val_windows, int):
+            num_val_windows = (num_val_windows,)
+        if not isinstance(num_val_windows, tuple):
+            raise TypeError(f"num_val_windows must be int or tuple[int, ...], got {type(num_val_windows)}")
+        if len(num_val_windows) == 0:
+            raise ValueError("num_val_windows tuple cannot be empty")
+        if not all(isinstance(n, int) and n > 0 for n in num_val_windows):
```
shchur, line 846: This is an extremely minor breaking change: the following code that worked previously will break now
```python
predictor.fit(train_data, tuning_data, num_val_windows=0)
```
Feel free to ignore it if the fix requires extra complexity - but maybe this can be combined with https://github.com/autogluon/autogluon/pull/5459/files#r2599398145 and we can patch `num_val_windows` based on the availability of `tuning_data`

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -756,9 +777,10 @@ def fit(
             val_data=tuning_data,
             hyperparameter_tune_kwargs=hyperparameter_tune_kwargs,
             excluded_model_types=excluded_model_types,
+            ensemble_hyperparameters=ensemble_hyperparameters,
             time_limit=time_left,
             verbosity=verbosity,
-            num_val_windows=num_val_windows,
+            num_val_windows=(num_val_windows,) if isinstance(num_val_windows, int) else num_val_windows,
```
shchur: I thought that at this point `num_val_windows` is guaranteed to be a `tuple[int, ...]`. Do we need this check here?

timeseries/src/autogluon/timeseries/predictor.py
canerturkmen: Not a problem. I think just changing the docstring is enough here. It was anyway aliased, and always showed up as `WeightedEnsemble` in the leaderboard (even when specified as GreedyEnsemble)

timeseries/src/autogluon/timeseries/predictor.py
canerturkmen: Changed the logic and the order of normalization operations significantly.

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -776,6 +790,113 @@ def fit(
...
+
+        if ensemble_hyperparameters is not None:
+            if isinstance(ensemble_hyperparameters, dict):
+                ensemble_hyperparameters = [ensemble_hyperparameters]
+
+            if len(ensemble_hyperparameters) != len(original_num_val_windows):
+                raise ValueError(
+                    f"Length mismatch: num_val_windows has {len(original_num_val_windows)} layers but "
+                    f"ensemble_hyperparameters has {len(ensemble_hyperparameters)} layers. "
+                    f"These must match for multilayer ensembling."
```
shchur: Let's use `multi-layer` consistently in the user-facing API of the predictor?

timeseries/src/autogluon/timeseries/predictor.py
```diff
@@ -776,6 +790,113 @@ def fit(
...
+            if sum(result) <= max_allowed:
+                break
+
+        # It is possible that the above for loop reduced the number of windows in each group to 1
+        # (i.e. result = [1] * len(num_val_windows)), but still sum(result) > max_allowed. In this
+        # case we set result = [1] * max_allowed
+        if sum(result) > max_allowed:
+            result = [1] * max_allowed
+
+        return tuple(result)
```
shchur, line 898: This PR looks great, taking some notes about what we should improve for multi-layer stacking:
1. Better logging when training ensemble models. Log ensemble names, etc.
<details> 

```python

full = TimeSeriesDataFrame('https://autogluon.s3.amazonaws.com/datasets/timeseries/m4_hourly/test.csv')
h = 24
train, test = full.train_test_split(h)
predictor = TimeSeriesPredictor(prediction_length=h, eval_metric="SQL").fit(
    train,
    num_val_windows=(2, 2),
    hyperparameters={"Naive": {}, "SeasonalNaive": {}, "Theta": {}, "Chronos": {}},
    ensemble_hyperparameters=[{"WeightedEnsemble": {}, "MedianEnsemble": {}, "LinearStackerEnsemble": {"weights_per": "mq"}}, {"WeightedEnsemble": {}}],
)
```
Output:
```
Starting training. Start time is 2025-12-10 10:10:04
Models that will be trained: ['Naive', 'SeasonalNaive', 'Theta', 'Chronos[autogluon__chronos-bolt-small]']
Training timeseries model Naive. 
	-9.0523       = Validation score (-SQL)
	2.67    s     = Training runtime
	0.25    s     = Validation (prediction) runtime
Training timeseries model SeasonalNaive. 
	-0.8341       = Validation score (-SQL)
	1.00    s     = Training runtime
	0.24    s     = Validation (prediction) runtime
Training timeseries model Theta. 
	-1.9149       = Validation score (-SQL)
	3.31    s     = Training runtime
	1.93    s     = Validation (prediction) runtime
Training timeseries model Chronos[autogluon__chronos-bolt-small]. 
	-0.6527       = Validation score (-SQL)
	3.20    s     = Training runtime
	0.89    s     = Validation (prediction) runtime
Fitting 4 ensemble(s), in 2 layers.
	Ensemble weights: {'Chronos[autogluon__chronos-bolt-small]': 0.85, 'SeasonalNaive': 0.15}
	-0.6387       = Validation score (-SQL)
	1.29    s     = Training runtime
	1.13    s     = Validation (prediction) runtime
	-0.8757       = Validation score (-SQL)
	0.02    s     = Training runtime
	3.33    s     = Validation (prediction) runtime
	-0.6407       = Validation score (-SQL)
	0.74    s     = Training runtime
	3.31    s     = Validation (prediction) runtime
	Ensemble weights: {'LinearStackerEnsemble_L2': 0.39, 'MedianEnsemble_L2': 0.03, 'WeightedEnsemble_L2': 0.58}
	-0.6373       = Validation score (-SQL)
	1.08    s     = Training runtime
	3.35    s     = Validation (prediction) runtime
Training complete. Models trained: ['Naive', 'SeasonalNaive', 'Theta', 'Chronos[autogluon__chronos-bolt-small]', 'MedianEnsemble_L2', 'LinearStackerEnsemble_L2', 'WeightedEnsemble_L2', 'WeightedEnsemble_L3']
Total runtime: 17.32 s
Best model: WeightedEnsemble_L3
Best model score: -0.6373
```
</details>


2. Validate ensemble names (and maybe hyperparameters) before training any models. Currently it's very easy to misconfigure the `ensemble_hyperparameters`, but the error only gets raised after all base models are trained - this can be very annoying. We could apply this logic to base models too since now we have the `allowed_hyperparameters` property for all models IIRC.

3. Allow specifying the ensemble models without the `Ensemble` suffix? E.g. currently `hyperparameters={"DeepAR": {}}` and `hyperparameters={"DeepARModel": {}}` are allowed, but only `hyperparameters={"MedianEnsemble": {}}` is allowed and `hyperparameters={"Median": {}}` raises an error.

canerturkmen: thanks!! I took care of 1 and 3. Ensemble names are validated by the trainer before fitting. I'll leave the hyperparameter validation as future work.

timeseries/src/autogluon/timeseries/models/ensemble/__init__.py
```diff
@@ -16,6 +16,12 @@ def get_ensemble_class(name: str):
         "PerQuantileTabularEnsemble": PerQuantileTabularEnsemble,
         "LinearStackerEnsemble": LinearStackerEnsemble,
     }
+
+    for full_name, ensemble_class in list(mapping.items()):
+        if full_name.endswith("Ensemble"):
+            alias = full_name[:-8]  # Remove "Ensemble" suffix
+            mapping[alias] = ensemble_class
```
shchur: Adding new aliases to the list will lead to duplicated names reported in L26 below in case of an error.

How about we do the following?

```python
    mapping = {
        "Greedy": GreedyEnsemble,
        "PerItemGreedy": PerItemGreedyEnsemble,
        "PerformanceWeighted": PerformanceWeightedEnsemble,
        "SimpleAverage": SimpleAverageEnsemble,
        "Weighted": GreedyEnsemble,  # old alias for this model
        "Median": MedianEnsemble,
        "Tabular": TabularEnsemble,
        "PerQuantileTabular": PerQuantileTabularEnsemble,
        "LinearStacker": LinearStackerEnsemble,
    }

    name_clean = name.removesuffix("Ensemble")
    if name_clean not in mapping:
        raise ValueError(f"Unknown ensemble type: {name}. Available: {list(mapping.keys())}")
    return mapping[name_clean]
```

timeseries/src/autogluon/timeseries/trainer/ensemble_composer.py
```diff
@@ -227,6 +227,18 @@ def get_ground_truth_for_layer(layer_idx):
                     # train the ensemble model
                     time_start = time.monotonic()
 
+                    fit_log_message = f"Training ensemble model {ensemble_name}. "
+                    if self.num_layers > 1:
+                        fit_log_message = fit_log_message[:-2] + (
+                            f" in layer {layer_idx} of {self.num_layers} stacked ensemble layers. "
```
shchur: I'm not sure if the verbosity really helps here, maybe we could just report the model name with the `_L{layer_idx}` suffix?

Current: 
```
Training ensemble model WeightedEnsemble in layer 1 of 2 stacked ensemble layers. 
	Ensemble weights: {'Chronos[autogluon__chronos-bolt-small]': 0.85, 'SeasonalNaive': 0.15}
	-0.6387       = Validation score (-SQL)
	1.25    s     = Training runtime
	1.07    s     = Validation (prediction) runtime
Training ensemble model MedianEnsemble in layer 1 of 2 stacked ensemble layers. 
	-0.8757       = Validation score (-SQL)
	0.02    s     = Training runtime
	3.36    s     = Validation (prediction) runtime
Training ensemble model LinearStackerEnsemble in layer 1 of 2 stacked ensemble layers. 
	-0.6406       = Validation score (-SQL)
	0.73    s     = Training runtime
	3.34    s     = Validation (prediction) runtime
Training ensemble model WeightedEnsemble in layer 2 of 2 stacked ensemble layers. 
	Ensemble weights: {'LinearStackerEnsemble_L2': 0.39, 'MedianEnsemble_L2': 0.03, 'WeightedEnsemble_L2': 0.58}
	-0.6373       = Validation score (-SQL)
	1.05    s     = Training runtime
	3.37    s     = Validation (prediction) runtime
```

Proposed:
```
Training ensemble model WeightedEnsemble_L2.
	Ensemble weights: {'Chronos[autogluon__chronos-bolt-small]': 0.85, 'SeasonalNaive': 0.15}
	-0.6387       = Validation score (-SQL)
	1.25    s     = Training runtime
	1.07    s     = Validation (prediction) runtime
Training ensemble model MedianEnsemble_L2.
	-0.8757       = Validation score (-SQL)
	0.02    s     = Training runtime
	3.36    s     = Validation (prediction) runtime
Training ensemble model LinearStackerEnsemble_L2.
	-0.6406       = Validation score (-SQL)
	0.73    s     = Training runtime
	3.34    s     = Validation (prediction) runtime
Training ensemble model WeightedEnsemble_L3. 
	Ensemble weights: {'LinearStackerEnsemble_L2': 0.39, 'MedianEnsemble_L2': 0.03, 'WeightedEnsemble_L2': 0.58}
	-0.6373       = Validation score (-SQL)
	1.05    s     = Training runtime
	3.37    s     = Validation (prediction) runtime
```

This way the names reported during training will be consistent with the ones displayed in leaderboard/accepted by `predict(..., model="...")`.

PR #5465, opened 2025-12-08
[timeseries] add sparsification to linear stacker

canerturkmen:
*Issue #, if available:*

*Description of changes:*
- add sparsification to linear stacker
- fix seed in random fixture generation
- fix ensemble data fixture

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/linear_stacker.py
+++ b/timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/linear_stacker.py
@@ -149,19 +149,16 @@ class LinearStackerEnsembleRegressor(EnsembleRegressor):
         with torch.no_grad():
             self.weights = weighted_average.get_normalized_weights().detach().numpy()
 
+        assert self.weights is not None
         if self.prune_below > 0.0:
-            importances = self.weights.squeeze()
-            if importances.ndim > 1:
-                importances = importances.mean(axis=tuple(range(importances.ndim - 1)))
+            importances = self.weights.mean(axis=tuple(range(self.weights.ndim - 1)))  # shape (num_models,)
 
-            kept_indices = [i for i in range(len(importances)) if importances[i] >= self.prune_below]
-            if not kept_indices:
-                kept_indices = [int(importances.argmax())]
+            mask = importances >= self.prune_below
+            if not mask.any():
+                mask[importances.argmax()] = True
 
-            if len(kept_indices) < len(importances):
-                self.kept_indices = kept_indices
-                mask = np.zeros(len(importances), dtype=bool)
-                mask[kept_indices] = True
+            if not mask.all():
+                self.kept_indices = np.where(mask)[0].tolist()
                 self.weights = self.weights[..., mask]
                 self.weights = self.weights / self.weights.sum(axis=-1, keepdims=True)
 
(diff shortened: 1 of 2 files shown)


timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/linear_stacker.py
```diff
@@ -140,11 +146,25 @@ def _ql(
...
+            kept_indices = [i for i in range(len(importances)) if importances[i] >= self.prune_below]
+            if not kept_indices:
+                kept_indices = [int(importances.argmax())]
+
+            if len(kept_indices) < len(importances):
+                self.kept_indices = kept_indices
+                mask = np.zeros(len(importances), dtype=bool)
+                mask[kept_indices] = True
+                self.weights = self.weights[..., mask]
+                self.weights = self.weights / self.weights.sum(axis=-1, keepdims=True)
```
shchur, line 163: Could this be vectorized, something like the following? I find vectorized code a bit easier to read
```python
if self.prune_below > 0.0:
    importances = self.weights.mean(axis=tuple(range(self.weights.ndim - 1)))  # shape (num_models,)
    
    mask = importances >= self.prune_below
    if not mask.any():
        mask[importances.argmax()] = True
    
    if not mask.all():
        self.kept_indices = np.where(mask)[0].tolist()
        self.weights = self.weights[..., mask]
        self.weights = self.weights / self.weights.sum(axis=-1, keepdims=True)
```

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/linear_stacker.py
```diff
@@ -155,10 +175,11 @@ def predict(
         if self.weights is None:
             raise ValueError("Model must be fitted before prediction")
 
-        # combine base model predictions
         all_predictions = np.concatenate([base_model_mean_predictions, base_model_quantile_predictions], axis=3)
 
-        # predict
+        if self.kept_indices is not None and all_predictions.shape[-1] > len(self.kept_indices):
```
shchur: If I understand correctly, after we updated
```python
            self._model_names = [original_names[i] for i in self.ensemble_regressor.kept_indices]
```
in `LinearStackerEnsemble._fit`, the `predict` method should always receive the predictions corresponding to the kept models. Maybe a better sanity check here is an assertion?
```python
if self.kept_indices is not None:
    assert all_predictions.shape[-1] == len(self.kept_indices)
```

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/linear_stacker.py
```diff
@@ -42,16 +46,18 @@ def __init__(
...
         self.quantile_levels = quantile_levels
         self.weights_per = weights_per
         self.lr = lr
         self.max_epochs = max_epochs
         self.relative_tolerance = relative_tolerance
+        self.prune_below = prune_below
 
-        # Learned weights (stored as numpy arrays)
         self.weights: np.ndarray | None = None
+        self.kept_indices: list[int] | None = None
```
shchur, line 60: Should we make this an attribute of the abstract class `EnsembleRegressor` for future-proofing? We could use similar logic in e.g. a GBDT-based TabularRegressor by looking at `model.feature_importances_`.

shchur (commented):
Thanks! Just a few comments

timeseries/src/autogluon/timeseries/models/ensemble/array_based/regressor/linear_stacker.py
canerturkmen: will skip for now as EnsembleRegressor has a dummy constructor.

PR #5469, opened 2025-12-09
[AutoMM] Fix false positive document detection for images with incidental text

FANGAreNotGnu:
https://github.com/autogluon/autogluon/issues/5326

Fixes `KeyError: 'document_transformer_text_valid_length'` when fitting MultiModalPredictor with non-document images (e.g., maps) that contain some text. The `is_document_image_column` function was incorrectly classifying regular images as documents when they contained any text (watermarks, copyright notices, labels). This caused the DocumentTransformer to be incorrectly applied.

*Changes*:
- Added text density heuristic (characters per pixel) to distinguish documents from images with scattered text
- Added minimum line count check to filter out single-line watermarks/captions
- Increased default text length threshold from 100 to 200 characters
- Fixed missing return statement in failure branch
- Improved error handling and logging

### DIFF
--- a/multimodal/src/autogluon/multimodal/data/infer_types.py
+++ b/multimodal/src/autogluon/multimodal/data/infer_types.py
@@ -125,7 +125,12 @@ def is_rois_input(sample):
     -------
     bool, whether a sample is rois for object detection
     """
-    return isinstance(sample, list) and len(sample) and isinstance(sample[0], list) and len(sample[0]) == 5
+    return (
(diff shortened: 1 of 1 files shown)


FANGAreNotGnu:
@Innixma @cheungdaven The updated document detection still works on our sample ducument dataset and should reduce the rate of images with a few texts or watermarks being misclassified as documents. We can adjust the threshold further if needed.

PR #5470, opened 2025-12-10
Fix Windows memory detection bug (#5433)

### LINKED ISSUE #5433: [BUG] PerformanceWarning: Available Memory reported incorrectly on Windows (3.2 TB)
During training with TabularPredictor on Windows, I received the following memory warning:

`PerformanceWarning: DataFrame is highly fragmented...
Available Memory: 3203341.44 MB
Train Data (Original) Memory Usage: 3185574.16 MB (99.4% of available memory)
`
Image

My computer has approximately 32 GB of RAM, so the reported value (~3.2 PB) is incorrect.

Large dataset (~85k rows and 1097 columns).

I am using AutoGluon version 1.4.1b20250924 and Python 3.12.11.

celestinoxp:
## Summary

Fixes #5433 - Incorrect memory reporting on Windows where available memory was reported as **3.2 TB** instead of actual RAM (~32 GB).

## Problem

During training with large datasets on Windows, the system reported unrealistic memory values:
- **Reported**: `Available Memory: 3203341.44 MB` (3.2TB) ❌
- **Actual**: ~32 GB RAM ✅

This occurred because `psutil.virtual_memory()` occasionally returns incorrect values on Windows.

## Solution

Implemented a **simplified, platform-specific approach** that uses the most reliable memory detection method for each platform:

### Windows
```python
# Use native Windows API (most reliable)
if platform.system() == "Windows":
    try:
        total_mem, _ = GlobalMemoryStatusEx()  # ctypes wrapper
        return total_mem
    except:
        return psutil.virtual_memory().total  # Fallback
```

### Linux/Mac
```python
# psutil works well on these platforms
return psutil.virtual_memory().total
```

### Why This Approach

1. **Simple**: No complex validation thresholds or magic numbers
2. **Reliable**: Always uses the most trusted source per platform
3. **Fast**: Single API call instead of multiple
4. **Maintainable**: ~70 lines less code than cross-validation approach
5. **Industry standard**: How other projects handle platform-specific APIs
6. **Future-proof**: Even if psutil improves, Windows API remains the best choice

### Modified Files
- `common/src/autogluon/common/utils/resource_utils.py` - Simplified platform-specific memory detection
- `common/tests/unittests/test_memory_validation.py` - Unit tests for validation logic

### Code Changes Summary
- **Removed**: 76 lines of complex cross-validation code
- **Added**: 14 lines of simple, clear platform-specific code
- **Net**: -62 lines, much cleaner implementation

### Key Functions

1. **`_get_memory_size_windows()`** - Windows API wrapper using `GlobalMemoryStatusEx`
   - Returns `(total_memory, available_memory)` in bytes
   - More reliable than psutil per Microsoft documentation

2. **`_validate_memory_size()`** - Basic sanity check
   - Validates memory is between 512 MB and 2 TB
   - Used for logging warnings, not primary detection

3. **`_get_memory_size()` and `_get_available_virtual_mem()`**
   - Windows: Use Windows API first, fallback to psutil
   - Other platforms: Use psutil directly

## Testing

All tests pass:
- ✅ Windows: Correct memory detection (31.42 GB total, ~17 GB available)
- ✅ Unit tests: Validation logic works correctly
- ✅ No breaking changes

**Test output:**
```
Test PASSED: Total=31.42 GB, Available=17.08 GB
```

## Impact

**Before**: Training logs showed 3.2 TB causing confusion and incorrect OOM warnings  
**After**: Correct memory detection (32 GB) with accurate monitoring

## Notes

- The bug is specific to Windows; Linux/Mac use psutil without issues
- Windows API (`GlobalMemoryStatusEx`) is recommended by Microsoft as the most reliable method
- Even latest psutil (7.1.3) can exhibit this issue intermittently
- Solution is permanent, not a temporary workaround


common/src/autogluon/common/utils/resource_utils.py
```diff
@@ -230,13 +319,23 @@ def _get_memory_rss() -> float:
...
             p = psutil.Process()
             return total_memory - p.memory_info().rss
 
+        # On Windows, prefer native Windows API (more reliable than psutil)
+        if platform.system() == "Windows":
+            try:
+                _, available_mem = ResourceManager._get_memory_size_windows()
+                return available_mem
+            except Exception as e:
+                logger.debug(f"Windows API unavailable, falling back to psutil: {e}")
```
Innixma, line 336: Might also need to adjust the above code:

```
     if os.environ.get("AG_MEMORY_LIMIT_IN_GB", None) is not None:
            total_memory = ResourceManager._get_custom_memory_size()
            p = psutil.Process()
            return total_memory - p.memory_info().rss
```

so that it uses the right value for `p.memory_info().rss`?

common/src/autogluon/common/utils/resource_utils.py
```diff
@@ -212,14 +212,103 @@ def _get_custom_memory_size():
...
+            
+        Returns
+        -------
+        bool
+            True if memory size seems realistic, False otherwise
+        """
+        # Most systems have between 512 MB and 2 TB of RAM
+        # Values outside this range are likely errors
+        MIN_REALISTIC_MEMORY = 512 * 1024 * 1024  # 512 MB
+        MAX_REALISTIC_MEMORY = 2 * 1024 * 1024 * 1024 * 1024  # 2 TB
```
Innixma: 2 TB is highly realistic in industry. Better to do something like 20 TB.

common/src/autogluon/common/utils/resource_utils.py
```diff
@@ -212,14 +212,103 @@ def _get_custom_memory_size():
...
+            Source of the memory value (for logging)
+            
+        Returns
+        -------
+        bool
+            True if memory size seems realistic, False otherwise
+        """
+        # Most systems have between 512 MB and 2 TB of RAM
+        # Values outside this range are likely errors
+        MIN_REALISTIC_MEMORY = 512 * 1024 * 1024  # 512 MB
```
Innixma: Because this is available memory, it is possible that the system is running on very little available memory, and thus 512 MB is actually a reasonable value. IMO better to do 32 MB.

Innixma:
@celestinoxp 3203341.44 MB is actually 3.2 TB, not 3.2 PB.

Innixma:
It seems like it is overestimating available memory by a factor of ~100, which seems like an oddly specific number.

common/src/autogluon/common/utils/resource_utils.py
```diff
@@ -212,14 +212,103 @@ def _get_custom_memory_size():
...
+            mem_status = MEMORYSTATUSEX()
+            mem_status.dwLength = ctypes.sizeof(MEMORYSTATUSEX)
+
+            if ctypes.windll.kernel32.GlobalMemoryStatusEx(ctypes.byref(mem_status)):
+                return float(mem_status.ullTotalPhys), float(mem_status.ullAvailPhys)
+            else:
+                raise RuntimeError("GlobalMemoryStatusEx API call failed")
+        except Exception as e:
+            logger.warning(f"Failed to get memory size using Windows API: {e}")
+            raise
```
Innixma, line 252: @celestinoxp do you have a references to others who have experienced this issue more generally (psutil on Windows)?

common/src/autogluon/common/utils/resource_utils.py
```diff
@@ -212,14 +212,103 @@ def _get_custom_memory_size():
...
+                    ("ullTotalVirtual", ctypes.c_ulonglong),
+                    ("ullAvailVirtual", ctypes.c_ulonglong),
+                    ("ullAvailExtendedVirtual", ctypes.c_ulonglong),
+                ]
+
+            mem_status = MEMORYSTATUSEX()
+            mem_status.dwLength = ctypes.sizeof(MEMORYSTATUSEX)
+
+            if ctypes.windll.kernel32.GlobalMemoryStatusEx(ctypes.byref(mem_status)):
+                return float(mem_status.ullTotalPhys), float(mem_status.ullAvailPhys)
```
Innixma: are these values integers? Probably best to return them as integers if they are in `bytes` form. I don't remember why the other methods are returning as floats, but they probably can return integers too if they are returning bytes I'd guess.

common/tests/unittests/test_memory_validation.py
```diff
@@ -0,0 +1,116 @@
...
+    if platform.system() == "Windows":
+        try:
+            total_mem, avail_mem = ResourceManager._get_memory_size_windows()
+            
+            # Validate the results
+            assert ResourceManager._validate_memory_size(total_mem, "Windows API test"), \
+                "Windows API total memory is unrealistic"
+            assert ResourceManager._validate_memory_size(avail_mem, "Windows API test"), \
+                "Windows API available memory is unrealistic"
+            assert avail_mem <= total_mem, "Available memory cannot exceed total"
```
Innixma, line 88: Maybe good to look at discrepancy in values between this and the normal `ResourceManager._get_available_virtual_mem` call?

common/src/autogluon/common/utils/resource_utils.py
```diff
@@ -230,13 +319,23 @@ def _get_memory_rss() -> float:
...
 
+        # On Windows, prefer native Windows API (more reliable than psutil)
+        if platform.system() == "Windows":
+            try:
+                _, available_mem = ResourceManager._get_memory_size_windows()
+                return available_mem
+            except Exception as e:
+                logger.debug(f"Windows API unavailable, falling back to psutil: {e}")
+        
+        # On other platforms or if Windows API failed, use psutil
```
Innixma, line 338: Maybe can make this its own mini method `_get_available_virtual_mem_psutil`, so we can call it in an isolated fashion in tests.

Innixma (commented):
Thanks for the PR! This is very nice and I wasn't aware Windows can sometimes do this. I've added some comments.

celestinoxp:
Thanks for the review @Innixma! I've addressed your feedback:
- Updated [_validate_memory_size](cci:1://file:///c:/Users/celes/autogluon/common/src/autogluon/common/utils/resource_utils.py:253:4-290:19) ranges to 32MB (min) and 20TB (max).
- Changed Windows API return types to `int`.
- Extracted [_get_available_virtual_mem_psutil](cci:1://file:///c:/Users/celes/autogluon/common/src/autogluon/common/utils/resource_utils.py:340:4-343:48) into its own method to isolate the logic.
- Added [test_psutil_fallback_direct](cci:1://file:///c:/Users/celes/autogluon/common/tests/unittests/test_memory_validation.py:95:0-107:51) to explicitly test the fallback mechanism.

Ready for another look!

Innixma:
Thanks! Generally LGTM, will fix the linting when I get a chance and will then merge

benhoumine-Abdelkhalek:
Hello,
We have observed a similar issue when running workloads on Linux nodes in a Kubernetes cluster, where the application reports severely inflated available memory values. We have encountered the same behavior in our environment.
Could you please advise whether the existing solution can be adapted for a Kubernetes environment (e.g., by relying on pod limits or cgroup-based memory reporting), or if we should open a separate issue to propose a Kubernetes-specific solution?
Thank you for your guidance.

celestinoxp:
@benhoumine-Abdelkhalek Thanks for the details!
From my experience, this issue seems to appear only in specific scenarios, such as with large datasets, and I’ve had difficulty reproducing it on my side. Have you been able to reproduce this consistently?
@Innixma, do you have insights on whether the current PR approach makes sense for Kubernetes environments, or should this be investigated separately?
It seems the main culprit may still be the psutil package, but further investigation is needed.

Innixma:
Hi @benhoumine-Abdelkhalek, I suspect that what you are experiencing would require a separate fix, as this PR seems specific to Windows.

In general we need a reproducible example, or at least where the user can tell us a specific setup that causes it to happen (and in what way it is incorrect)

benhoumine:
Hi,
Just a quick update.
We are currently testing the AutoGluon library in different VM environments (Windows and Linux) to better understand and confirm the root cause of the issue.
Once the testing phase is completed and the behavior is validated, I will open a PR with the proposed fix.

celestinoxp:
I wasn't able to reproduce the issue on my side, even after testing multiple versions. At this point, I suspect it may have been caused by a dependency (or a transitive dependency), which makes it hard to isolate a clear root cause.

I initially suspected psutil, but testing older versions didn’t reveal any issues either.

I have a couple of questions to move this forward:
• @benhoumine — were you able to find a reproducible example or any concrete evidence of the issue?
• @Innixma — do you have any suggestions on how to proceed?

If no one can currently reproduce the problem, I think it might make sense to close this PR for now and revisit it if the issue reappears with a reproducible case.

celestinoxp:
One additional thought on this:

Even if the original issue was real at the time, without a reproducible case it’s very hard to justify maintaining Windows-specific memory detection logic inside AutoGluon. This introduces long-term maintenance cost as Windows evolves, whereas relying on psutil delegates that responsibility to a library that actively tracks OS-level changes.

Given that the issue seems transient and potentially caused by a transitive dependency that may already have been updated, I’m not confident this PR provides a clear net benefit today.

Happy to follow your guidance on whether it’s better to keep this open or close it until a reproducible case exists.

benhoumine:
We are able to consistently reproduce the issue when running on Kubernetes.
However, when running the same code on Windows and on Ubuntu (VM), we aren't able to reproduce the error at all. The behavior is stable in both of those environments.
We believe the issue is environment-specific and strongly correlated with Kubernetes. At this stage, our main suspicion is psutil, which may be handling memory incorrectly in a containerized environment (cgroups / resource limits / Node / Pod). This could explain why the problem only manifests on Kubernetes and not on Windows or a standalone Ubuntu VM.
Given that the issue appears only under Kubernetes and involves memory behavior, it is likely related to how psutil interacts with container-level memory reporting or cleanup rather than an application-level bug.

celestinoxp:
@benhoumine Thank you for the details.

Just to clarify an important point: although the problem is currently being reproduced in Kubernetes, it doesn't seem to be exclusive to that environment. In the past, I myself observed the same behavior in Windows, which suggests that Kubernetes may only be making the problem more visible, and not necessarily being the root cause.

One factor that concerns me here is the complete dependency tree, including transitive dependencies. Between the time the problem originally occurred and now, several dependencies may have been updated—which would explain why we can no longer reproduce it today, even when forcing older direct versions (for example, of psutil).

To move forward more objectively, could you check if all package versions (direct and transitive) are exactly the same between:

the Kubernetes environment where the problem occurs

and the Windows/Ubuntu environments where it does not occur?

If we can fully align these versions and the problem still only occurs in Kubernetes, this would strengthen the likelihood of environment-specific behavior. Otherwise, differences in dependencies could explain the inconsistent behavior we are observing.

Any additional details about package versions, Python, and how the incorrect memory value is being observed would be very helpful in continuing the investigation.

Innixma:
Following up on this after digging into #5433 — the root cause turned out to be somewhere else entirely, and there's now a fix in #5863.

**psutil was reporting memory correctly.** The two log lines in the issue give this away:

```
Available Memory: 3203341.44 MB
Train Data (Original) Memory Usage: 3185574.16 MB (99.4% of available memory)
```

We all read line 1 as psutil's answer, but at `features/src/autogluon/features/generators/pipeline.py:126` that line is a **sum**:

```python
available_mem_mb = ResourceManager.get_available_virtual_mem(format="MB")
pre_memory_usage_percent = pre_memory_usage_mb / (available_mem_mb + pre_memory_usage_mb)
self._log(20, f"\tAvailable Memory:  {(pre_memory_usage_mb + available_mem_mb):.2f} MB")
```

So psutil's actual value is the difference between the two reported numbers:

```
3203341.44 - 3185574.16 = 17767.28 MB = 17.35 GiB
```

17.35 GiB available on a 32 GB machine — correct. Cross-check: feeding that back through the percentage formula gives `3185574.16 / 3203341.44 = 99.4%`, matching the issue exactly, which confirms the reading rather than just fitting it. And @celestinoxp, your own "Test PASSED: Total=31.42 GB, Available=17.08 GB" is essentially that same number — the value this PR set out to fix is the one psutil was already returning.

**The actual bug** is in `get_approximate_df_mem_usage`, which estimated **3.04 TiB** for an 85k x 1097 frame that cannot exceed ~0.7 GiB dense. `memory_usage(deep=True)` charges `sys.getsizeof(obj)` once per *cell*, so a value shared across many cells is counted once per cell. pandas' C `read_csv` parser reuses one `str` object per distinct value, so any low-cardinality string column read from disk is fully shared — measured at 99.9% of cells pointing at the same object. Against real process memory, on 200k x 5:

| case | real | before | after |
|---|---|---|---|
| `read_csv`, 20 B strings | 7.9 MiB | 66.8 MiB (8.5x) | 7.6 MiB (0.97x) |
| `read_csv`, 2 KB strings | 8.8 MiB | 1955 MiB (223x) | 8.0 MiB (0.91x) |
| `read_csv`, all distinct | 84.7 MiB | 83.4 MiB (0.98x) | 82.7 MiB (0.98x) |

None of it is platform-specific. @benhoumine-Abdelkhalek @benhoumine — that may well cover your Linux/Kubernetes reports too; worth re-checking against #5863 before opening a separate fix, since the conclusion back in December that it needed a Windows-independent investigation was based on the same misreading.

@celestinoxp — your instinct on 2026-02-04 was right, and for a better reason than "can't reproduce": there was nothing to reproduce on the psutil side. Your argument for not carrying Windows-specific memory code ("relying on psutil delegates that responsibility to a library that actively tracks OS-level changes") holds up. Thanks for the work here and for pushing back on your own PR — that's what made it worth re-examining the numbers instead of merging a plausible-looking fix.

Suggest closing this in favor of #5863, and keeping #5433 open but retitled away from "on Windows".

Innixma:
Thanks for the PR @celestinoxp and especially for highlighting the problem. I did a deep dive on this one (refer to the above message) and found the true root cause which is a bug in how the DF size is being calculated for object dtypes with shared values and is platform agnostic. The fix is in #5863. Closing this PR.

PR #5475, opened 2025-12-11
[timeseries] fixes to Toto and add to smoke tests

canerturkmen:
*Issue #, if available:*

*Description of changes:*

- disables compilation by default in Toto, for added stability
- computes inference batch size dynamically (previous logic forced batch size to be multiple of 32)
- add to smoke tests

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/toto/hf_pretrained_model.py
+++ b/timeseries/src/autogluon/timeseries/models/toto/hf_pretrained_model.py
@@ -1,4 +1,7 @@
+import json
 import logging
+import os
+from pathlib import Path
 
 from transformers import PretrainedConfig, PreTrainedModel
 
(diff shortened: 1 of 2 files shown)


timeseries/tests/smoketests/test_all_models.py
```diff
@@ -88,6 +88,7 @@ def generate_train_and_test_data(
     "SimpleFeedForward": DUMMY_MODEL_HPARAMS,
     "TemporalFusionTransformer": DUMMY_MODEL_HPARAMS,
     "TiDE": DUMMY_MODEL_HPARAMS,
+    "Toto": {"num_samples": 20, "device": "cpu"},
```
shchur: Maybe we can keep this even lower for better efficiency? Like `"num_samples": 5"` even

shchur (approved):

PR #5477, opened 2025-12-11
Refactor TabularNeuralNetTorch: Fix bugs, optimize memory, and improve performance

celestinoxp:
**Description:** This PR refactors TabularNeuralNetTorch  to fix critical logic bugs, optimize memory usage, and enable parallel data loading on non-MacOS systems.

**Critical Fixes**
Early Stopping: Fixed off-by-one error where epoch-1 was passed instead of epoch, causing misalignment between logging and internal logic.
Time Estimation: Corrected time-per-epoch calculation (removed incorrect divisor epoch + 1), which was underestimating training time by up to 50% in early epochs.
Stability: Fixed a ValueError crash in _predict_tabular_data when predicting on empty datasets. Ensures returned empty array has correct shape (0, num_outputs) to maintain downstream compatibility.

**Optimizations**
Memory & Speed: Changed internal checkpointing to save model.state_dict() instead of pickling the entire model object. This reduces memory pressure and serialization overhead during validation.
Data Loading: Enabled parallel data loading (num_dataloading_workers > 0) for Windows and Linux systems. It remains disabled on MacOS due to platform-specific issues.

**Code Quality**
Replaced non-functional Warning(...) instantiation with logger.log(15, ...) for consistent and effective logging.
Fixed typo in error message ("must of of" -> "must be of").
Fixed typo in docstrings.

### DIFF
--- a/tabular/src/autogluon/tabular/models/tabular_nn/torch/tabular_nn_torch.py
+++ b/tabular/src/autogluon/tabular/models/tabular_nn/torch/tabular_nn_torch.py
@@ -664,7 +664,8 @@ class TabularNeuralNetTorchModel(AbstractNeuralNetworkModel):
             new_data = self._process_test_data(new_data)
         if not isinstance(new_data, TabularTorchDataset):
             raise ValueError("new_data must be of type TabularTorchDataset if process=False")
-        val_dataloader = new_data.build_loader(self.max_batch_size, self.num_dataloading_workers, is_test=True)
+        # Ensure single-process loading for inference to guarantee order
+        val_dataloader = new_data.build_loader(self.max_batch_size, 0, is_test=True)
         preds_dataset = []
         for data_batch in val_dataloader:
             preds_batch = self.model.predict(data_batch)
--- a/tabular/src/autogluon/tabular/models/tabular_nn/torch/tabular_torch_dataset.py
+++ b/tabular/src/autogluon/tabular/models/tabular_nn/torch/tabular_torch_dataset.py
@@ -154,6 +154,12 @@ class TabularTorchDataset(torch.utils.data.IterableDataset):
         if self.shuffle:
             np.random.shuffle(idxarray)
         indices = range(0, self.num_examples, self.batch_size)
+
+        worker_info = torch.utils.data.get_worker_info()
+        if worker_info is not None:
+            # Split data across workers
+            indices = indices[worker_info.id :: worker_info.num_workers]
+
         for idx_start in indices:
             # Drop last batch
             if self.drop_last and (idx_start + self.batch_size) > self.num_examples:


celestinoxp:
cc @Innixma regarding lines 203: we noticed sample_weight support is pending. Is this something we should tackle soon?

tabular/src/autogluon/tabular/models/tabular_nn/torch/tabular_nn_torch.py
```diff
@@ -207,9 +207,12 @@ def _fit(
             self.num_dataloading_workers = max(1, int(num_cpus / 2.0))
         else:
             self.num_dataloading_workers = 1
+        import platform
+
         if self.num_dataloading_workers == 1:
             self.num_dataloading_workers = 0  # TODO: verify 0 is typically faster and uses less memory than 1 in pytorch
-        self.num_dataloading_workers = 0  # TODO: >0 crashes on MacOS
+        if platform.system() == "Darwin":
+            self.num_dataloading_workers = 0  # TODO: >0 crashes on MacOS
```
Innixma, line 216: Interesting, I forgot about this being hard-coded to 0 in the past. I'll need to run a few tests to see how this impacts training time on Linux.

tabular/src/autogluon/tabular/models/tabular_nn/torch/tabular_nn_torch.py
```diff
@@ -464,10 +467,10 @@ def _train_net(
                         is_best = True
                     best_val_metric = val_metric
                     io_buffer = io.BytesIO()
-                    torch.save(self.model, io_buffer)  # nosec B614
+                    torch.save(self.model.state_dict(), io_buffer)
                     best_epoch = epoch
                     best_val_update = total_updates
-                early_stop = early_stopping_method.update(cur_round=epoch-1, is_best=is_best)
+                early_stop = early_stopping_method.update(cur_round=epoch, is_best=is_best)
```
Innixma: I think `epoch-1` is correct after looking through the code, since we want to have cur_round be 0-indexed for consistency, so the first epoch should be 0. This is relevant for early stopping methods that use the `cur_round` value to dynamically determine `patience`, such as in the `AdaptiveES` class.

tabular/src/autogluon/tabular/models/tabular_nn/torch/tabular_nn_torch.py
```diff
@@ -495,7 +498,7 @@ def _train_net(
 
             if time_limit is not None:
                 time_elapsed = time.time() - start_fit_time
-                time_epoch_average = time_elapsed / (epoch + 1)
+                time_epoch_average = time_elapsed / epoch
```
Innixma, line 513: Nice find!!

tabular/src/autogluon/tabular/models/tabular_nn/torch/tabular_nn_torch.py
```diff
@@ -648,13 +651,17 @@ def _predict_tabular_data(self, new_data, process=True):
...
         preds_dataset = []
         for data_batch in val_dataloader:
             preds_batch = self.model.predict(data_batch)
             preds_dataset.append(preds_batch)
-        preds_dataset = np.concatenate(preds_dataset, 0)
+        if len(preds_dataset) > 0:
+            preds_dataset = np.concatenate(preds_dataset, 0)
+        else:
+            num_net_outputs = self._get_num_net_outputs()
+            preds_dataset = np.zeros((0, num_net_outputs), dtype=np.float32)
```
Innixma, line 677: Is it common practice to support predicting on an empty dataset?
I think in general AutoGluon models crash in this scenario, so I'm unsure if we want to be supporting this case.

Happy to hear your thoughts.

tabular/src/autogluon/tabular/models/tabular_nn/torch/tabular_nn_torch.py
```diff
@@ -207,9 +207,12 @@ def _fit(
             self.num_dataloading_workers = max(1, int(num_cpus / 2.0))
         else:
             self.num_dataloading_workers = 1
+        import platform
```
Innixma: I think platform import is safe to put at top of file alongside the other imports

Innixma (commented):
Added initial review. Really nice changes, thanks for this!!

Innixma:
> cc @Innixma regarding lines 203: we noticed sample_weight support is pending. Is this something we should tackle soon?

Regarding sample_weight, this is something not well supported in most models currently in AutoGluon, but is definitely something that would be good to tackle. Generally speaking the sample weights are given in the `_fit` call if specified by the user, and it has to be implemented in a model-specific fashion since each model handles sample weights differently.

tabular/src/autogluon/tabular/models/tabular_nn/torch/tabular_nn_torch.py
celestinoxp: Fixed the epoch indexing to be 0-based (epoch - 1)

tabular/src/autogluon/tabular/models/tabular_nn/torch/tabular_nn_torch.py
celestinoxp: Moved import platform to top-level.

celestinoxp:
Thanks for the review!

Fixed the epoch indexing to be 0-based (epoch - 1).
Moved import platform to top-level.
Regarding empty datasets: I added it to prevent crashes in edge cases (e.g., during some split operations or unique validation scenarios), but if you prefer strict failure for empty inputs to align with other models, I can remove it. For now, it returns an empty prediction with the correct shape instead of crashing.

celestinoxp:
FYI @Innixma Closing this PR because I found a bug in the workers that breaks the functionality of nn_torch. To avoid creating a snowball effect of more commits and bugs, I will later create individual PRs for each individual problem.

PR #5479, opened 2025-12-12
Fix: Raise ValueError when package version is None

celestinoxp:
This PR prevents misleading error messages. The issue is likely caused by package conflicts (e.g. conda vs pip).

Fixes https://github.com/autogluon/autogluon/issues/5464

### DIFF
--- a/common/src/autogluon/common/utils/utils.py
+++ b/common/src/autogluon/common/utils/utils.py
@@ -7,7 +7,7 @@ import sys
 from datetime import datetime, timezone
 from hashlib import md5
 from pathlib import Path
-from typing import Any, Dict, Optional
+from typing import Any, Optional
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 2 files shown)


common/src/autogluon/common/utils/utils.py
```diff
@@ -132,7 +132,16 @@ def get_package_versions() -> Dict[str, str]:
...
-    package_version_dict = {dist.metadata["Name"].lower(): dist.version for dist in importlib.metadata.distributions()}
+    package_version_dict = {}
+    for dist in importlib.metadata.distributions():
+        name = dist.metadata["Name"]
+        version = dist.version
+        if version is None:
+            raise ValueError(
+                f"Package '{name}' has version None. This usually indicates that multiple versions of the package are installed. "
+                f"Please check your environment for conflicting packages."
+            )
```
Innixma: Perhaps better to log a warning and convert `version` to the string `"None"`?

Without knowing when None can appear, I'm concerned with blindly raising a ValueError.

celestinoxp: updated with a more robust code...

celestinoxp:
Udated with more robust code

Innixma:
Hi @celestinoxp, while the new code is more detailed, it does not address my original comment

> Perhaps better to log a warning and convert version to the string "None"?
> Without knowing when None can appear, I'm concerned with blindly raising a ValueError.

The code is still able to raise ValueError's which is what I was concerned about, and it is not converting the string to "None" / "unknown" when we can't identify it.

Innixma (approved):
Thanks for the contrib @celestinoxp!

I added unit tests + updated the code to avoid logging and instead included the invalid package exceptions as part of the saved json file.

PR #5482, opened 2025-12-14
[Tabular] Pass label cleaner to model for semantic encodings

LennartPurucker:
Add support for passing the label cleaner to the model, such that models and preprocessing can invert the label encoding and ensure the semantics of the label. 

*Description of changes:*
* add label cleaner as a `self.label_cleaner` variable in AbstractModel and pass it to the `_fit` call. 


Innixma (approved):
LGTM!

PR #5484, opened 2025-12-16
Fix time_epoch_average calculation in TabularNeuralNetTorch

celestinoxp:
*Description of changes:*
Fix time_epoch_average calculation in TabularNeuralNetTorch


tabular/src/autogluon/tabular/models/tabular_nn/torch/tabular_nn_torch.py
```diff
@@ -495,7 +495,7 @@ def _train_net(
 
             if time_limit is not None:
                 time_elapsed = time.time() - start_fit_time
-                time_epoch_average = time_elapsed / (epoch + 1)
+                time_epoch_average = time_elapsed / epoch
```
Innixma:
```suggestion
                time_epoch_average = time_elapsed / max(epoch, 1)  # avoid divide by 0
```

PR #5486, opened 2025-12-16
[timeseries] Add option to disable covariates in Chronos2

shchur:
*Issue #, if available:*

*Description of changes:*
- Add hyperparameters `disable_known_covariates` and `disable_past_covariates` to Chronos2 (similar to the same hyperparameters in GluonTS models)

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/gluonts/models.py
+++ b/timeseries/src/autogluon/timeseries/models/gluonts/models.py
@@ -41,7 +41,6 @@ class DeepARModel(AbstractGluonTSModel):
         Number of steps to unroll the RNN for before computing predictions
     disable_static_features : bool, default = False
         If True, static features won't be used by the model even if they are present in the dataset.
-        If False, static features will be used by the model if they are present in the dataset.
     disable_known_covariates : bool, default = False
         If True, known covariates won't be used by the model even if they are present in the dataset.
     num_layers : int, default = 2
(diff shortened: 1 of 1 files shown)


timeseries/src/autogluon/timeseries/models/chronos/chronos2.py
```diff
@@ -76,6 +76,12 @@ class Chronos2Model(AbstractTimeSeriesModel):
         Extra keyword arguments passed to ``transformers.TrainingArguments``
     revision : str, default = None
         Model revision to use (branch name or commit hash). If None, the default branch (usually "main") is used.
+    disable_known_covariates : bool, default = False
+        If True, known covariates won't be used by the model even if they are present in the dataset.
+        If False, known covariates will be used by the model if they are present in the dataset.
```
abdulfatir: nit: Isn't the `False` docstring redundant?

timeseries/tests/unittests/models/chronos/test_chronos2.py
```diff
@@ -10,6 +10,18 @@
 from ..common import CHRONOS2_MODEL_PATH
 
 
+@pytest.fixture()
```
abdulfatir: Why do we need another fixture? 
https://github.com/autogluon/autogluon/blob/2581d8ca2a53fd459d2e396f36eb2a083b69b951/timeseries/tests/conftest.py#L95

timeseries/src/autogluon/timeseries/models/chronos/chronos2.py
shchur: Thanks, fixed

timeseries/tests/unittests/models/chronos/test_chronos2.py
shchur: My bad, removed this

PR #5492, opened 2025-12-17
[timeseries] Make Chronos-2 respect `time_limit` during `predict`

### LINKED ISSUE #5488: [timeseries] Make sure that Chronos-2 respects the time_limit during predict
Currently the model completely ignores the time_limit, potentially leading to very long runtimes.

We can address it similarly to how Chronos-Bolt does it with a time limit callback
https://github.com/autogluon/autogluon/blob/656b1ff4460615c55fe07fd15478b9d087420dca/timeseries/src/autogluon/timeseries/models/chronos/model.py#L691-L697

Additionally, we could reduce the default batch_size when running on CPU to speed up the inference.

abdulfatir:
*Issue #, if available:* Fixes #5488

*Description of changes:*

- Ensures that Chronos-2 respects time limit during predict.
- Renames `on_batch` to `after_batch` in callback.

### DIFF
--- a/timeseries/tests/unittests/models/chronos/test_chronos2.py
+++ b/timeseries/tests/unittests/models/chronos/test_chronos2.py
@@ -124,7 +124,7 @@ class TestChronos2Inference:
         mock_from_pretrained.assert_called_once()
         assert mock_from_pretrained.call_args.kwargs.get("revision") == model_revision
 
-    def test_when_chronos_scores_oof_and_time_limit_is_exceeded_then_exception_is_raised(self, chronos2_model):
+    def test_when_chronos2_scores_oof_and_time_limit_is_exceeded_then_exception_is_raised(self, chronos2_model):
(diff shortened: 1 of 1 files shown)


timeseries/src/autogluon/timeseries/models/chronos/chronos2.py
```diff
@@ -228,6 +228,8 @@ def _predict(
         known_covariates: TimeSeriesDataFrame | None = None,
```
shchur, line 229: Should we also bump the version range for `chronos-forecasting` in `setup.py`?

shchur: Actually, let's do this after fixing the predict_df issue

abdulfatir: Yes, but let's wait for any other efficiency improvements before release.

PR #5493, opened 2025-12-17
[Tabular] Fix XGBoost crashing on eval metric name in HPs

LennartPurucker:
There seems to have been a bug that if you pass the name of an eval metric to the HPs of XGBoost, the `eval_metric_name` variable was never assigned. This PR fixes this. 


Innixma (approved):
LGTM!

PR #5494, opened 2025-12-17
[timseries] Update presets for v1.5

shchur:
*Issue #, if available:*

*Description of changes:*

Requires #5492 to be merged!

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/chronos/chronos2.py
+++ b/timeseries/src/autogluon/timeseries/models/chronos/chronos2.py
@@ -83,7 +83,7 @@ class Chronos2Model(AbstractTimeSeriesModel):
     """
 
     ag_model_aliases = ["Chronos-2"]
-    ag_priority = 90
+    ag_priority = 75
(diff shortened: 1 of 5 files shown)


timeseries/src/autogluon/timeseries/models/chronos/chronos2.py
```diff
@@ -83,7 +83,7 @@ class Chronos2Model(AbstractTimeSeriesModel):
     """
 
     ag_model_aliases = ["Chronos-2"]
-    ag_priority = 90
+    ag_priority = 75
```
abdulfatir, line 86: Just to confirm: this is intended?

abdulfatir (approved):

timeseries/src/autogluon/timeseries/models/chronos/chronos2.py
shchur: Yes, making sure that Chronos-2 is trained after Direct/RecursiveTabular

PR #5505, opened 2025-12-19
[tabular] Add v1.5.0 Presets 

Innixma:
*Issue #, if available:*

*Description of changes:*

- Add new tabular presets for the v1.5.0 release.


tabular/setup.py
```diff
@@ -61,7 +61,7 @@
         "torch",  # version range defined in `core/_setup_utils.py`
     ],
     "tabpfn": [
-        "tabpfn>=6.1.0,<6.1.1",  # <{N+1} upper cap, where N is the latest released minor version
+        "tabpfn>=6.2.0,<6.2.1",  # <{N+1} upper cap, where N is the latest released minor version
```
prateekdesai04, line 64: why do we not allow 6.1 anymore ?

Innixma: they added `.to(device)` support in 6.2, so we want to depend only on that version for CPU loading.

Issue #4756, opened 2024-12-27
How to Use Focal Loss Function in Classification Problems with Autogluon?


lovechang1986:
First of all, thanks for such an excellent open-source project.

Currently, I'm using Autogluon for small-sample classification tasks and I've found that the Focal loss function is very helpful for learning with imbalanced samples. 

So, how can I use the Focal loss function in classification problems with Autogluon?

celestinoxp:
you shuld use "eval_metric" to define the metric, and value "log_loss" to use loss function...
Example:
```python
time_limit = 60  # for quick demonstration only, you should set this to longest time you are willing to wait (in seconds)
metric = 'roc_auc'  # specify your evaluation metric here
predictor = TabularPredictor(label, eval_metric=metric).fit(train_data, time_limit=time_limit, presets='best_quality')
```
you can see more: https://auto.gluon.ai/stable/index.html

FANGAreNotGnu:
The tutorial for implementing Focal Loss in AutoGluon can be found here:
https://auto.gluon.ai/stable/tutorials/multimodal/advanced_topics/focal_loss.html

I'm closing this issue since the tutorial provides the implementation details. Feel free to reopen if you have any additional questions!

dai-ichiro:
[The provided tutorial](https://auto.gluon.ai/stable/tutorials/multimodal/advanced_topics/focal_loss.html) is stuck with an error.

My System Info is here.
```
=================== System Info ===================
AutoGluon Version:  1.4.0
Python Version:     3.12.12
Operating System:   Linux
Platform Machine:   x86_64
Platform Version:   #1 SMP Tue Nov 5 00:21:55 UTC 2024
CPU Count:          20
Pytorch Version:    2.7.1+cu126
CUDA Version:       12.6
GPU Count:          1
Memory Avail:       20.87 GB / 23.47 GB (88.9%)
Disk Space Avail:   793.41 GB / 1006.85 GB (78.8%)
===================================================
```

This was all I needed.
```python
weights = [float(w) for w in weights]
```

FANGAreNotGnu:
@dai-ichiro Thanks for the findings! Fixing this in PR https://github.com/autogluon/autogluon/pull/5496

dai-ichiro:
I upgraded Autogluon to version 1.5, but this error in [tutorial](https://auto.gluon.ai/stable/tutorials/multimodal/advanced_topics/focal_loss.html) persists.

[...]

PR #5526, opened 2026-01-07
Fix focal loss alpha parsing for numpy float types

FANGAreNotGnu:
#4756 

### Problem

When users compute class weights using numpy operations, the resulting list contains `np.float64` values instead of native Python floats:

```python
weights = list(np.array(weights) / np.sum(weights))  # Creates list of np.float64
```

When passed through OmegaConf, these `np.float64` values get serialized as strings like `'np.float64(0.0239...)'`, causing `_parse_alpha()` to fail with:

```
ValueError: too many dimensions 'str'
```

### Fix

Updated `_parse_alpha()` to handle numpy float string representations:

```python
if isinstance(alpha, (list, tuple)):
    alpha = [float(str(val).split("(")[-1].rstrip(")")) for val in alpha]
```

### Test Update

Added parametrized test case covering both Python floats and numpy floats:

```python
@pytest.mark.parametrize(
    "alpha",
    [
        [1.0, 0.25, 0.35, 0.16],  # Python float
        list(np.array([1.0, 0.25, 0.35, 0.16])),  # np.float64
    ],
)
def test_focal_loss_multiclass(alpha):
    ...
```


shchur (approved):
Thank you for the fix!

PR #5538, opened 2026-01-10
Support np.ndarray in cv_splitter

Innixma:
*Issue #, if available:*

*Description of changes:*

- Support np.ndarray in splitter
- Also support X=None for splitter
- Used in TabArena

### DIFF
--- a/common/src/autogluon/common/utils/cv_splitter.py
+++ b/common/src/autogluon/common/utils/cv_splitter.py
@@ -101,7 +101,7 @@ class CVSplitter:
         else:
             raise AssertionError(f"{splitter_cls} is not supported as a valid `splitter_cls` input to CVSplitter.")
 
-    def split(self, X: pd.DataFrame | None, y: pd.Series) -> list[tuple[np.ndarray, np.ndarray]]:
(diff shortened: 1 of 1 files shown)


common/src/autogluon/common/utils/cv_splitter.py
```diff
@@ -101,7 +101,11 @@ def _get_splitter(self, splitter_cls) -> BaseCrossValidator:
         else:
             raise AssertionError(f"{splitter_cls} is not supported as a valid `splitter_cls` input to CVSplitter.")
 
-    def split(self, X: pd.DataFrame, y: pd.Series) -> list[tuple[np.ndarray, np.ndarray]]:
+    def split(self, X: pd.DataFrame | None, y: pd.Series) -> list[tuple[np.ndarray, np.ndarray]]:
```
shchur: Should the annotation then be as follows?
```suggestion
    def split(self, X: pd.DataFrame | None, y: pd.Series | np.ndarray) -> list[tuple[np.ndarray, np.ndarray]]:
```

Innixma: Good catch!

PR #5539, opened 2026-01-11
[timeseries] Fix environment-dependent errors in PredictionCache

ryuichi-ichinose:
Description

### Current Issues
The current PredictionCache implementation has a critical design flaw: it defines the cache storage location based on the absolute path of the model at training time. This causes severe issues when the model is moved to a different environment:

- Environment Leakage: A model trained in a specific environment (like /opt/ml) persistently attempts to create cache files at that same absolute path, leading to unintended file generation in arbitrary host locations.
- Permission Denied: Processes fail with PermissionError if the training-time absolute path is read-only or missing in the execution environment.
- Silent Cache Pollution: Because it relies on absolute paths rather than content/state, different models can incorrectly share the same cache. For example, if two models have different internal parameters but share the same name and training path, the system will return the same cached results. This occurs when comparing same-named models with different parameters in the same script, or when a previous cache remains and a new model with the same name is used for prediction, leading to silent data corruption and incorrect evaluation.

### Changes
I have refactored the cache to be execution-environment agnostic and robust:

- Relative Path Storage: Cache is now strictly contained within the predictions_cache directory relative to the current predictor's root path.
- Context-Aware Model Hashing: Introduced compute_model_hash, which anchors the cache key to the model's relative path, file size, and modification time (mtime). This ensures cache invalidation if the model is updated or replaced by a different version.
- Atomic Writes & Granular Structure: Switched to a hierarchical file-per-model structure using os.replace for atomic updates, preventing file corruption during concurrent writes.
- Decoupled Trainer Logic: Updated TimeSeriesTrainer to pass a model_path_map, ensuring the cache system always knows where models reside relative to the root.

### Results
- Encapsulation: All cache files are now fully contained within the predictor's own directory.
- Zero Permission Issues: Eliminates PermissionError at fixed system paths (like /opt/ml) by using the predictor's local environment.
- Data Integrity: Strict cache integrity is guaranteed. Even if model names collide, differences in file size or modification time will trigger a cache miss, ensuring correct results for different model instances.
- Robustness: Verified via unit tests that model updates correctly invalidate the cache.

### DIFF
--- a/timeseries/src/autogluon/timeseries/trainer/prediction_cache.py
+++ b/timeseries/src/autogluon/timeseries/trainer/prediction_cache.py
@@ -1,7 +1,10 @@
+import hashlib
 import logging
+import os
+import shutil
+import uuid
 from abc import ABC, abstractmethod
 from pathlib import Path
-from typing import Any
 
 from autogluon.common.utils.utils import hash_pandas_df
 from autogluon.core.utils.loaders import load_pkl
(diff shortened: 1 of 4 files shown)


shchur:
We will deprecate prediction caching in v1.7, so I will close this PR for now.

Issue #5541, opened 2026-01-13
[CRITICAL BUG] [tabular] TabDPT crashes when predicting a single sample for binary classification.


Innixma:
TabDPT crashes at inference time when predicting on a single sample for binary classification. This impacts all cases where TabDPT is used in AutoGluon, such as in the v1.5 `extreme` preset, causing AutoGluon to crash in this scenario.

This bug impacts `tabdpt<1.1.12`

I have created an issue (https://github.com/layer6ai-labs/TabDPT-inference/issues/63) and proposed a fix (https://github.com/layer6ai-labs/TabDPT-inference/pull/64).

As a temporary workaround, I suggest users who wish to predict on single samples to duplicate the sample before predicting with AutoGluon when using the `extreme` setting until TabDPT releases with the fix (hopefully with 1.1.12).

Innixma:
I have additionally created #5542 to add unit tests so we catch these kinds of bugs early on before we release.

Innixma:
This has been fixed with the release of TabDPT v1.1.12

PR #5542, opened 2026-01-13
[tabular] Add check that models can predict single samples

Innixma:
*Issue #, if available:*

*Description of changes:*

- Add check that models can predict single samples. -> TabDPT crashes in this case
- Add check that models return the same predictions when predicting twice in a row on the same data
- Add check that models return the same prediction on sample A when predicting on samples [A, B, C] vs predicting on A only. -> Note: TabDPT, Mitra, and TabPFN fail this check, so manually disabling the check for them.

Note: TabDPT test will fail until #5541 is resolved. This PR makes sure that the bug is visible and caught by our tests.

### DIFF
--- a/tabular/tests/unittests/models/test_tabicl.py
+++ b/tabular/tests/unittests/models/test_tabicl.py
@@ -12,4 +12,6 @@ def test_tabicl():
         model_cls=model_cls,
         model_hyperparameters=model_hyperparameters,
         verify_load_wo_cuda=True,
+        # TabICL returns different predictions when predicting on an individual sample
+        verify_single_prediction_equivalent_to_multi=False,
     )


tabular/src/autogluon/tabular/testing/fit_helper.py
```diff
@@ -252,6 +254,26 @@ def fit_and_validate_dataset(
...
+            pred_proba_repeat = predictor.predict_proba(test_data)
+            are_close = np.isclose(pred_proba, pred_proba_repeat).all()
+            if not are_close:
+                raise AssertionError(
+                    "Predictions differ when predicting on the same data multiple times\n"
+                    f"First Predict:\n{pred_proba}\n"
+                    f"Second Predict:\n{pred_proba_repeat}\n"
+                )
+
+            pred_proba_1 = predictor.predict_proba(test_data.head(1))  # Verify model can predict on a single sample
```
shchur, line 267: Why is this currently not failing for TabDPT?

shchur (approved):
LGTM!

tabular/src/autogluon/tabular/testing/fit_helper.py
Innixma: Because TabDPT authors released v1.1.12 which fixes the issue, it was failing yesterday when it was using v1.1.11

PR #5548, opened 2026-01-17
[Tabular][Maint] Allow fit helper to use large test datasets 

LennartPurucker:
For some models, we have to create several internal splits, which does not work with using just 4 samples from the default toy tasks. This add a toggle to the fit helper to run on the large toy datasets. 


Innixma (approved):
LGTM!

PR #5556, opened 2026-01-26
[Core] Refactor greedy ensemble selection to use np.isclose for floating point comparisons

celestinoxp:
*Description of changes:*

Refactors the strike degree selection to use np.isclose in floating-point comparison instead of direct comparison (==). This improves the robustness of the algorithm when working with close floating-point scores.

Currently, the direct comparison method of floating-point values ​​can lead to inconsistent results in tie scenarios or when values ​​are very close, due to the nature of floating-point numbers in Python. Using np.isclose solves this by ensuring more stable numerical comparisons.

### DIFF
--- a/core/src/autogluon/core/models/greedy_ensemble/ensemble_selection.py
+++ b/core/src/autogluon/core/models/greedy_ensemble/ensemble_selection.py
@@ -164,7 +164,7 @@ class EnsembleSelection(AbstractWeightedEnsemble):
                                 y_true=labels, y_pred_proba=fant_ensemble_prediction, metric=secondary_metric
                             )
                         all_best_tiebreak = np.argwhere(
-                            np.isclose(scores_tiebreak, np.nanmin(scores_tiebreak), atol=0)
+                            np.isclose(scores_tiebreak, np.nanmin(scores_tiebreak), atol=0, rtol=1e-12)
                         ).flatten()
                         all_best = [index_map[index] for index in all_best_tiebreak]
 


celestinoxp:
Great point, Nick! I completely agree. The standard absolute tolerance (atol=1e-08) can really cause problems in regression tasks with very small values.

Your suggestion to use atol=0 makes perfect sense because:

It maintains only the relative tolerance (rtol=1e-05), which scales appropriately with the magnitude of the values.
It avoids the problem of very small, but significantly different, scores being considered equal.
It continues to solve the original problem of floating-point numerical imprecision.

I will update the PR to use np.isclose(scores, np.nanmin(scores), atol=0). 

Thank you for the careful review!

core/src/autogluon/core/models/greedy_ensemble/ensemble_selection.py
```diff
@@ -137,7 +137,7 @@ def _fit(self, predictions: List[np.ndarray], labels: np.ndarray, time_limit=Non
                 if round_scores:
                     scores[j] = scores[j].round(round_decimals)
 
-            all_best = np.argwhere(scores == np.nanmin(scores)).flatten()
+            all_best = np.argwhere(np.isclose(scores, np.nanmin(scores), atol=0)).flatten()
```
Innixma:
```suggestion
            all_best = np.argwhere(np.isclose(scores, np.nanmin(scores), atol=0, rtol=1e-12)).flatten()
```

core/src/autogluon/core/models/greedy_ensemble/ensemble_selection.py
```diff
@@ -163,7 +163,9 @@ def _fit(self, predictions: List[np.ndarray], labels: np.ndarray, time_limit=Non
                             scores_tiebreak[k] = self._calculate_regret(
                                 y_true=labels, y_pred_proba=fant_ensemble_prediction, metric=secondary_metric
                             )
-                        all_best_tiebreak = np.argwhere(scores_tiebreak == np.nanmin(scores_tiebreak)).flatten()
+                        all_best_tiebreak = np.argwhere(
+                            np.isclose(scores_tiebreak, np.nanmin(scores_tiebreak), atol=0)
+                        ).flatten()
```
Innixma, line 168:
```suggestion
                        all_best_tiebreak = np.argwhere(
                            np.isclose(scores_tiebreak, np.nanmin(scores_tiebreak), atol=0, rtol=1e-12)
                        ).flatten()
```

Innixma:
Sounds good! Given we generally compute metrics in float64, I've also made rtol=1e-12 so it is truly only for numerical precision level differences rather than actual differences.

PR #5559, opened 2026-01-26
Add .gitattributes to enforce LF for Python files

celestinoxp:
*Description of changes:*

It adds the .gitattributes file to enforce Unix-style (LF) line endings for Python and Shell scripts, ensuring consistency between Windows and Linux environments and avoiding large differences caused by CRLF conversions. It also explicitly marks binary files to prevent corruption.

Context: This setting was previously introduced in an older version, but the branch was closed and this file was never incorporated into the main version.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -13,8 +13,8 @@
 *.hxx text eol=lf
 *.py text eol=lf
 *.sh text eol=lf
-*.md text
-*.txt text
+*.md  text eol=lf
+*.txt text eol=lf
 
 # Declare files that will always have CRLF line endings on checkout.
 *.sln text eol=crlf


.gitattributes
```diff
@@ -0,0 +1,29 @@
+# Set the default behavior, in case people don't have core.autocrlf set.
+* text=auto
```
Innixma:
```suggestion
* text=auto eol=lf
```

.gitattributes
```diff
@@ -0,0 +1,29 @@
...
+
+# Explicitly declare text files you want to always be normalized and converted
+# to native line endings on checkout.
+*.c text
+*.h text
+*.c++ text
+*.h++ text
```
Innixma:
```suggestion
*.c   text eol=lf
*.h   text eol=lf
*.cc  text eol=lf
*.cpp text eol=lf
*.cxx text eol=lf
*.hh  text eol=lf
*.hpp text eol=lf
*.hxx text eol=lf
```

.gitattributes
```diff
@@ -0,0 +1,29 @@
...
+*.py text eol=lf
+*.sh text eol=lf
+*.md text
+*.txt text
```
Innixma:
```suggestion
*.md  text eol=lf
*.txt text eol=lf
```

PR #5561, opened 2026-01-27
[Common][Core][Tabular] Replace bare except blocks with specific exceptions

celestinoxp:
*Description of changes:*

This PR addresses the "bare except" anti-pattern by replacing generic `except:` blocks with specific exceptions (e.g., `ValueError`, `TypeError`) or `except Exception:` where a catch-all is truly intended but system interrupts should pass through.

### DIFF
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -1303,7 +1303,7 @@ class AbstractModel(ModelBase, Tunable):
             try:
                 import torch
             except ImportError:
-                # torch missing â†’ nothing to restore
+                # torch missing -> nothing to restore
                 pass
             else:
                 if reset_torch_threads:


core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -1303,7 +1303,7 @@ def fit(
             try:
                 import torch
             except ImportError:
-                # torch missing → nothing to restore
+                # torch missing â†’ nothing to restore
```
Innixma:
```suggestion
                # torch missing -> nothing to restore
```

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -1317,7 +1317,7 @@ def fit(
                 out = self
             out = out._post_fit(**kwargs)
         finally:
-            # Always executed — even if _fit or _post_fit raise
+            # Always executed â€” even if _fit or _post_fit raise
```
Innixma:
```suggestion
            # Always executed even if _fit or _post_fit raise
```

Innixma (approved):
LGTM!

LennartPurucker:
Heyho, this PR seems to trigger a bug when using code with categorical features, as the infer types functions now throw un-caught exceptions. 

In the function `get_bool_true_val` we are missing `AttributeError`. Not sure what else yet. I will comment on it if I spot more

PR #5564, opened 2026-01-27
[Core] Optimize augment_rare_classes with single concat

celestinoxp:
*Description of changes:*

This PR optimizes the [augment_rare_classes] utility function by replacing iterative DataFrame concatenation inside a loop with a single list collection and final implementation of `pd.concat`.

**Changes:**
- Replaced the loop that appended to `aug_df` on every iteration (O(N^2) complexity behavior) with a list accumulation strategy followed by a single `pd.concat` (O(N)).
- Simplified the logic for duplicating rows using list multiplication `[df] * n`.
**Impact:**
- Improves performance, especially when there are many rare classes or large augmentation factors.
- Reduces memory overhead by avoiding the creation of multiple intermediate DataFrame copies.
- Code is cleaner and more idiomatic.

### DIFF
--- a/core/src/autogluon/core/utils/utils.py
+++ b/core/src/autogluon/core/utils/utils.py
@@ -1,4 +1,4 @@
-﻿from __future__ import annotations
+from __future__ import annotations
 
 import logging
 import math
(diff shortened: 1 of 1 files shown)


Innixma (approved):
LGTM! I tested the logic with some unit tests and it matches the old logic, thanks!

PR #5567, opened 2026-01-29
[Tabular] Optimize memory usage in CatBoostModel preprocessing

celestinoxp:
*Description of changes:*

CatBoostModel._preprocess previously performed X = X.copy() to safely handle missing values in categorical columns.

Optimization: Refactored to use X = X.copy(deep=False) combined with explicit copying of only the categorical columns being modified.

### DIFF
--- a/tabular/src/autogluon/tabular/models/catboost/catboost_model.py
+++ b/tabular/src/autogluon/tabular/models/catboost/catboost_model.py
@@ -70,8 +70,10 @@ class CatBoostModel(AbstractModel):
         if self._category_features is None:
             self._category_features = list(X.select_dtypes(include="category").columns)
         if self._category_features:
-            X = X.copy()
+            X = X.copy(deep=False)
             for category in self._category_features:
+                # Changes are made in place, so we need to copy the category
+                X[category] = X[category].copy()
                 current_categories = X[category].cat.categories
                 if "__NaN__" in current_categories:
                     X[category] = X[category].fillna("__NaN__")
--- a/tabular/src/autogluon/tabular/models/catboost/catboost_model.py
+++ b/tabular/src/autogluon/tabular/models/catboost/catboost_model.py
@@ -70,10 +70,11 @@ class CatBoostModel(AbstractModel):
         if self._category_features is None:
             self._category_features = list(X.select_dtypes(include="category").columns)
         if self._category_features:
+            # Only the categorical columns below are rewritten, and each is replaced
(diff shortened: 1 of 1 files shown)


Innixma:
Measured this to see what the shallow copy buys, on pandas 2.3.3 with the real `_preprocess` path. It is both a memory and a throughput win, and both scale with how many columns are not categorical.

Peak allocation, 200k rows with 30 numeric and 3 categorical columns (48.6 MB input): **58.6 MB to 10.6 MB**.

Time per `_preprocess` call, 200k rows:

| columns | before | after |
|---|---|---|
| 100 numeric + 2 categorical | 24.1 ms | 11.5 ms (2.10x) |
| 30 numeric + 3 categorical | 18.9 ms | 14.4 ms (1.31x) |
| 10 numeric + 3 categorical | 14.9 ms | 14.9 ms (1.00x) |
| 1 numeric + 20 categorical | 110.0 ms | 111.0 ms (0.99x) |

End to end, `predict_proba` on 200k rows with 30 numeric and 3 categorical columns goes from 35.6 ms to 29.6 ms. Frames dominated by categorical columns see nothing, which is expected: there the deep copy was already copying almost nothing extra.

Correctness checks, since the shallow copy shares buffers with the caller: predictions are bit identical to master (same md5) on a fit with categorical NaNs and unseen test categories; the caller's frame and dtypes are unchanged; and that holds with `copy_on_write` both off and on, which matters because AutoGluon pins `pandas<2.4.0` where copy on write is not the default. Also checked on the sliced frames AutoGluon actually passes in (`X = X[self.features]`, row slices, views) with no chained assignment warnings.

I pushed one commit to the branch removing the per-column copy:

```python
# Changes are made in place, so we need to copy the category
X[category] = X[category].copy()
```

It is discarded on the very next line, since both branches reassign the column with the result of `fillna` or `add_categories().fillna()`, so nothing ever reads the copy. It also cannot be protecting the caller: no operation in this method edits data in place, they all build a new Series and rebind the column, which is exactly what makes `copy(deep=False)` safe. Removing it leaves the caller's frame equally unmutated in every case above, with peak allocation unchanged at 10.6 MB. The comment now records the invariant the shallow copy depends on, rather than describing in place mutation that does not happen.

The branch still merges cleanly with master, though it is 182 commits behind, so it may be worth updating the branch before merging so CI runs against current master.

Innixma (approved):
Thanks for the contrib!!

PR #5569, opened 2026-01-29
[Tabular] Optimize memory usage estimation in XGBoostModel

celestinoxp:
*Description of changes:*

XGBoostModel._estimate_memory_usage_static used a hardcoded multiplier of 7x input data size, regardless of configuration. This was overly conservative and inaccurate.

Solution: Updated the estimation to dynamically account for enable_categorical.

If enable_categorical=True, overhead is estimated at 2x (DMatrix quantization + loading).
If enable_categorical=False (default, using OHE), overhead is estimated at 4x (OHE sparse expansion + DMatrix + loading).
This provides a tighter bounds on memory usage, allowing better resource utilization.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,33 @@
+# Set the default behavior, in case people don't have core.autocrlf set.
+* text=auto eol=lf
+
+# Explicitly declare text files you want to always be normalized and converted
+# to native line endings on checkout.
+*.c   text eol=lf
+*.h   text eol=lf
+*.cc  text eol=lf
+*.cpp text eol=lf
+*.cxx text eol=lf
+*.hh  text eol=lf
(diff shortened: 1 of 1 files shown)


Innixma:
Definitely appreciate enhancements to memory estimation, but currently without a benchmarking mechanism to check accuracy of the estimate, it is unclear if this new estimate is better than the existing memory estimate. For that reason, I'll avoid merging this change for now and instead wait for us to provide a benchmark that tracks memory usage so we can better empirically verify the memory estimate logic. This is planned as part of an extension of the functionality present in TabArena, so that we can naturally track memory usage as we fit models, and then we can use that information across many datasets to write a memory estimate function that aligns with the real observed memory usage.

PR #5572, opened 2026-02-03
[multimodal] Cache HF datasets during CI

shchur:
*Issue #, if available:*

  Multimodal tests download models and datasets from HuggingFace Hub. We already cache HF models in S3, but two tests (`test_backbones.py` and `test_metrics.py`) also load HF datasets (`glue/mrpc` and `SetFit/stsb`) which weren't cached. This causes HF rate limiting failures in CI.

*Description of changes:*
  - Extend the existing HF mirror infrastructure to also cache datasets
  - Add missing model (`microsoft/deberta-v3-small`) used by `test_ner.py`
  - Set `HF_HUB_OFFLINE=1` during tests to guarantee no network requests

*After merging:*
- The `CI/hf_mirror/` job needs to be run after this PR is merged to populate S3 with the new model and datasets. This job clones from master, so it must be triggered separately after merge:
  1. Rebuild and push the Docker image via `CI/hf_mirror/deploy.sh`
  2. Trigger the AWS Batch job to download the new artifacts to S3 

### DIFF
--- a/multimodal/src/autogluon/multimodal/data/preprocess_dataframe.py
+++ b/multimodal/src/autogluon/multimodal/data/preprocess_dataframe.py
@@ -615,9 +615,9 @@ class MultiModalFeaturePreprocessor(TransformerMixin, BaseEstimator):
                 processed_data = col_value.apply(lambda ele: ele if isinstance(ele, list) else [ele]).tolist()
             elif col_type == IMAGE_BASE64_STR:
                 processed_data = col_value.apply(
-                    lambda ele: [base64.b64decode(e) for e in ele]
-                    if isinstance(ele, list)
-                    else [base64.b64decode(ele)]
+                    lambda ele: (
+                        [base64.b64decode(e) for e in ele] if isinstance(ele, list) else [base64.b64decode(ele)]
+                    )
                 ).tolist()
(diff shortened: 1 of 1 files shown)


multimodal/tests/hf_model_list.yaml
```diff
@@ -20,6 +20,7 @@ others:
 - huawei-noah/TinyBERT_General_4L_312D
 - microsoft/deberta-base
 - microsoft/deberta-v3-base
+- microsoft/deberta-v3-small
```
Innixma, line 23: I've also seen failures in tabular tests due to downloading Mitra (although far less common than AutoMM failures). We may want to also do a separate PR which pre-downloads Mitra (and maybe TabPFN/TabDPT/TabICL/etc.?)

prateekdesai04 (approved):

multimodal/tests/hf_model_list.yaml
shchur: Sure. Here is an example of how caching is done on the TS side https://github.com/autogluon/autogluon/blob/8ab07861f260b29ad78d42c1b606531b47adf63e/timeseries/tests/conftest.py#L7-L19

PR #5584, opened 2026-02-10
[Core] Fix feature importance p-value calculation and optimize batch size

celestinoxp:
Description of changes:
* Fixed p-value calculation for feature importance when standard deviation is 0 and mean is negative (now correctly returns 1.0 instead of 0.5).
* Replaced fixed max_feature_batch_count default (200) with an adaptive heuristic that limits rows*features per batch to ~2,500,000 (capped at 10,000), balancing memory usage and vectorization speed across datasets of different sizes.


Innixma (approved):
Batching logic sounds reasonable, thanks for the contrib!

PR #5585, opened 2026-02-10
[CI] Ensure consistent Ruff version across local and CI environments

celestinoxp:
**Description of changes:**

Currently, the CI environment installs the latest version of Ruff (pip install ruff), while local development relies on varying versions. This can lead to situations where code passes locally but fails in CI due to new rules introduced in a recent Ruff release.

This PR pins the Ruff version to 0.15.0 in both .pre-commit-config.yaml and .github/workflow_scripts/env_setup.sh.

**Benefits:**
Developer Experience: Ensures all contributors use the exact same linter and formatter version, preventing "works on my machine" conflicts.
Consistency: Ensures that pre-commit run locally produces the exact same results as the CI pipeline.
Stability: Prevents sudden CI failures caused by upstream Ruff updates.
Reproducibility: Older commits will remain valid even as Ruff evolves.

**Future Work:** To keep the Ruff version up-to-date without manual intervention, I recommend configuring Dependabot (or similar tools like Renovate/pre-commit-ci) to automatically open PRs when new versions become available. This would combine the stability of pinned versions with the benefits of staying current. @Innixma I'd like to know what you think about this.

### DIFF
--- a/.github/workflow_scripts/env_setup.sh
+++ b/.github/workflow_scripts/env_setup.sh
@@ -4,6 +4,7 @@ function setup_build_env {
     python -m pip install flake8
     python -m pip install bandit
     python -m pip install packaging
+    # Read the Ruff version from .pre-commit-config.yaml to keep a single source of truth
     RUFF_VERSION=$(grep -A5 'astral-sh/ruff-pre-commit' .pre-commit-config.yaml | grep 'rev:' | head -n1 | sed 's/.*v//')
     python -m pip install ruff=="$RUFF_VERSION"
 }


.github/workflow_scripts/env_setup.sh
```diff
@@ -4,7 +4,7 @@ function setup_build_env {
     python -m pip install flake8
     python -m pip install bandit
     python -m pip install packaging
-    python -m pip install ruff
+    python -m pip install ruff==0.15.0
```
Innixma: not required, but I wonder if we can have a single source of truth for the ruff version so we don't have `0.15.0` repeated in multiple places?

.pre-commit-config.yaml
```diff
@@ -1,9 +1,14 @@
...
-        always_run: true
+      # Run the formatter.
+      - id: ruff-format
+        name: ruff-format
+        files: ^(multimodal/|timeseries/|common/|core/|features/|tabular/)
+      # Run the linter.
+      - id: ruff
+        name: ruff-lint
+        args: ["--select", "I", "--fix"]
+        files: ^(multimodal/|timeseries/|common/|core/|features/|tabular/)
```
Innixma, line 15: @shchur This looks to be force formatting the code on commit.

Is this aligned with the direction you want to head in, or is this a step too far?

Innixma:
@celestinoxp Regarding

> To keep the Ruff version up-to-date without manual intervention, I recommend configuring Dependabot (or similar tools like Renovate/pre-commit-ci) to automatically open PRs when new versions become available.

Do you have an example PR opened by Dependabot in a similar package I can refer to so I can better understand the process?

celestinoxp:
@Innixma Dependabot doesn't support .pre-commit-config.yaml natively.
For the simplest solution, pre-commit.ci would handle this automatically — it runs the hooks on every PR and opens weekly autoupdate PRs. It just requires enabling it for the repo.
Alternatively, a scheduled GitHub Action can do the same without any external service:

```yaml
# .github/workflows/update-pre-commit.yml
name: Update pre-commit hooks
on:
  schedule:
    - cron: '0 9 * * 1'  # every monday
  workflow_dispatch:
jobs:
  autoupdate:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - uses: actions/setup-python@v5
        with:
          python-version: '3.11'
      - run: pip install pre-commit
      - run: pre-commit autoupdate
      - uses: peter-evans/create-pull-request@v7
        with:
          token: ${{ secrets.GITHUB_TOKEN }}
          commit-message: 'chore: update pre-commit hooks'
          title: 'chore: auto-update pre-commit hooks'
          branch: auto/pre-commit-update         
```

.github/workflow_scripts/env_setup.sh
celestinoxp: @Innixma Good point! We can read the version directly from .pre-commit-config.yaml in the CI script, so it's only defined in one place:
```bash
RUFF_VERSION=$(grep -A1 'astral-sh/ruff-pre-commit' .pre-commit-config.yaml | grep 'rev:' | sed 's/.*v//')
python -m pip install ruff==$RUFF_VERSION
```
This way, updating the version in .pre-commit-config.yaml is all that's needed — the CI script picks it up automatically.

.pre-commit-config.yaml
celestinoxp: @Innixma @shchur After reviewing the change more carefully, the switch to --fix and auto-formatting was unintentional. Updated to preserve the original behavior: the hooks now check and block on failure without modifying files automatically.

Innixma:
@celestinoxp would this script only create a PR if there are new ruff releases? I guess we have the ability to define the script and therefore how it works.

celestinoxp:
@Innixma Yes — peter-evans/create-pull-request only opens a PR if pre-commit autoupdate actually modifies .pre-commit-config.yaml. If the version is already up-to-date, no PR is created.

Innixma:
I'm happy to give it a try, maybe if you'd like you can send a PR after this is merged as a follow-up to enable it

celestinoxp:
Sounds good! I'll open a follow-up PR with the workflow after this is merged.

Innixma (approved):
LGTM!

shchur:
Thanks for the PR @celestinoxp!

PR #5586, opened 2026-02-10
[CI] Fix mmengine failure

prateekdesai04:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/.github/workflow_scripts/setup_mmcv.sh
+++ b/.github/workflow_scripts/setup_mmcv.sh
@@ -6,7 +6,7 @@ function setup_mmcv {
   # Install MMEngine from PyPI wheel to avoid setuptools>=82 removing pkg_resources
   # FIXME: https://github.com/open-mmlab/mmcv/issues/3325, revert mmcv installation to use mim once fixed
   python3 -m pip install "setuptools<82"
-  python3 -m pip install "mmengine==0.10.5"
   python3 -m pip install "mmcv==2.1.0" --no-build-isolation --timeout 60
-  python3 -m pip install "mmdet==3.2.0"
+  python3 -m pip install "mmengine==0.10.7"
+  python3 -m pip install "mmdet==3.3.0"
 }
--- a/core/setup.py
+++ b/core/setup.py
@@ -60,6 +60,7 @@ extras_require = {
         "hyperopt>=0.2.7,<0.2.8",  # This is needed for the bayes search to work.
         # 'GPy>=1.10.0,<1.11.0'  # TODO: Enable this once PBT/PB2 are supported by ray lightning
         "stevedore<5.5",
+        "setuptools<82",
     ],
 }
 


.github/workflow_scripts/env_setup.sh
```diff
@@ -85,7 +85,7 @@ function install_all_no_tests {
 }
 
 function build_pkg {
-    pip install --upgrade setuptools wheel
+    pip install --upgrade "setuptools<82" wheel
```
Innixma, line 89: add reasoning for version limit as a comment in this file

prateekdesai04: still trying it out, will add it once I find the best resolution for this.
I would ideally not like to touch setuptools as its a very basic yet important package which constantly gets updated, but if there is no other way we may have to.

.github/workflow_scripts/setup_mmcv.sh
```diff
@@ -3,8 +3,9 @@ function setup_mmcv {
...
   fi
-  # Install MMEngine from git with the fix for torch 2.5
-  python3 -m pip install "git+https://github.com/open-mmlab/mmengine.git@2e0ab7a92220d2f0c725798047773495d589c548"
-  mim install "mmcv==2.1.0" --timeout 60
+  # Install MMEngine from PyPI wheel to avoid setuptools>=82 removing pkg_resources
+  python3 -m pip install "setuptools<82"
+  python3 -m pip install "mmengine==0.10.5"
+  python3 -m pip install "mmcv==2.1.0" --no-build-isolation --timeout 60
   python3 -m pip install "mmdet==3.2.0"
 }
```
celestinoxp, line 12: @prateekdesai04 @Innixma investigated an alternative that avoids pinning setuptools entirely. Since mmdet 3.3.0 supports mmengine>=0.7.1, <1.0.0, we can upgrade the full stack:

```bash
python3 -m pip install "mmengine==0.11.0rc2"
python3 -m pip install "mmcv==2.1.0" --no-build-isolation --timeout 60
python3 -m pip install "mmdet==3.3.0"
```
mmengine 0.11.0rc2 replaces pkg_resources with importlib natively ([open-mmlab/mmengine#1676](https://github.com/open-mmlab/mmengine/pull/1676)), so no setuptools<82 pin needed at all.
Worth testing — if it passes CI, the only follow-up would be updating the AutoGluon docs which currently reference mmdet==3.2.0.

prateekdesai04: Thanks for the suggestion @celestinoxp , updating mmengine and mmdet doesn't affect the mmcv build process at all.
Hence we have to pin the setuptools version for now as it is a fresh issue on the mmcv side (opened yesterday): https://github.com/open-mmlab/mmcv/issues/3325.

Here is the CI run for the changes you suggested in case you're curious: https://github.com/autogluon/autogluon/pull/5587

PR #5588, opened 2026-02-11
[tabular] Improve logic for specifying custom model feature generators

Innixma:
*Issue #, if available:*

*Description of changes:*

1. Adds support for custom feature generator pipelines for any model.
2. Ensure all models pass `y` into their training `preprocess` call

Code example:

```python

data_root = 'https://autogluon.s3.amazonaws.com/datasets/Inc/'
train_data = TabularDataset(data_root + 'train.csv')
test_data = TabularDataset(data_root + 'test.csv')

prep_params = {
        "ag.model_specific_feature_generator_kwargs": {
            "feature_generators": [
                [
                    [["ArithmeticFeatureGenerator", {}]],
                    [
                        ["CategoricalInteractionFeatureGenerator", {"passthrough": True}],
                        ["OOFTargetEncodingFeatureGenerator", {}],
                    ],
                ]
            ],
            "passthrough_types": {"invalid_raw_types": ["category", "object"]},
        },
    }

predictor = TabularPredictor(label='class').fit(
    train_data,
    hyperparameters={
        "GBM": [
            {},  # Normal
            {**prep_params},  # With custom preprocessing
        ]},
)
predictor.leaderboard(display=True)

```

### DIFF
--- a/features/src/autogluon/features/registry/parse_custom_generator.py
+++ b/features/src/autogluon/features/registry/parse_custom_generator.py
@@ -47,10 +47,7 @@ def resolve_fg_class(
             target = parse_import_target(name)
             cls = import_by_target(target)
         else:
-            raise KeyError(
-                f"Unknown feature generator {name!r}. "
-                f"Known registry keys: {sorted(registry or {})}... "
-            )
+            raise KeyError(f"Unknown feature generator {name!r}. Known registry keys: {sorted(registry or {})}... ")
 
     if not inspect.isclass(cls):
         raise TypeError(f"Resolved {name!r} to non-class object: {cls!r}")


core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -1472,7 +1467,7 @@ def _fit(
         Refer to `fit` method for documentation.
         """
 
-        X = self.preprocess(X)
```
prateekdesai04, line 1475: Don't have full context but were we just preprocessing X earlier, if yes why so?

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -2554,38 +2579,61 @@ def estimate_memory_usage(self, X: pd.DataFrame, **kwargs) -> int:
...
+                    num_samples=len(X),
+                    num_features=len(new_feature_metadata.get_features()),
+                    hyperparameters=hyperparameters,
+                    problem_type=self.problem_type,
+                    num_classes=self.num_classes,
+                )
+                self._memory_usage_estimate = memory_usage_estimate
+                return memory_usage_estimate
+            else:
+                # FIXME: This is expensive
```
prateekdesai04, line 2602: is this expensive just because of addition of y ?

features/src/autogluon/features/generators/arithmetic/preprocessor.py
```diff
@@ -81,8 +79,6 @@ class ArithmeticFeatureGenerator(AbstractFeatureGenerator):
         Whether to transform categorical features to numeric and use them as base features in arithmetic interaction generation.
     min_cardinality : int, default = 3
         Minimum cardinality for a feature to be considered as a base feature in arithmetic interaction generation.
-    random_state : int, default = 42
```
prateekdesai04, line 84: why are we removing this default value?

core/src/autogluon/core/models/abstract/abstract_model.py
Innixma: preprocessing generally occurs only on the features (X). We pass y now to enable fitting feature preprocessors that use the target labels to inform the feature preprocessing, which were only recently added.

core/src/autogluon/core/models/abstract/abstract_model.py
Innixma: No, it is because we are fitting the entire feature preprocessing pipeline in order to estimate the memory usage. We do this to get the preprocessed X to pass into the memory estimate function for a model, which may be much larger than the original X.

An ideal implementation would avoid fitting the preprocessing pipeline for estimating memory, but this is complicated and not yet implemented.

features/src/autogluon/features/generators/arithmetic/preprocessor.py
Innixma: It has been moved to AbstractFeatureGenerator so `random_state` is standardized to all preprocessors.

PR #5589, opened 2026-02-11
[Core] Update typing-extensions constraint to >=4.14.0 for cattrs compatibility

celestinoxp:
## Description
This PR updates the dependency lower bound for `typing-extensions` in [core/src/autogluon/core/_setup_utils.py](cci:7://file:///c:/Users/celes/autogluon/core/src/autogluon/core/_setup_utils.py:0:0-0:0) to resolve an installation conflict.
### Problem
The previous constraint `">=4.0,<5"` allowed older versions (e.g., 4.13.2) to be installed. However, `cattrs` (v25.3.0), pulled in by test dependencies, requires `typing-extensions>=4.14.0`.
### Solution
Updated the constraint to `">=4.14.0,<5"`. This ensures the installed version satisfies both AutoGluon and `cattrs` requirements.


Innixma (approved):
LGTM!

PR #5594, opened 2026-02-14
[multimodal] Update HF model list and enforce offline mode in CI

celestinoxp:
Update HF model list and enforce offline mode in CI

### DIFF
--- a/.github/workflow_scripts/test_multimodal.sh
+++ b/.github/workflow_scripts/test_multimodal.sh
@@ -6,6 +6,14 @@ function test_multimodal {
 
     setup_build_env
     setup_hf_model_mirror "$SUB_FOLDER"
+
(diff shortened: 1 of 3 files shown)


celestinoxp:
@shchur The CI failure was due to missing system-level dependencies (poppler-utils and tesseract-ocr) required for the PDF/OCR tests in the Linux environment. I've added them to the Dockerfile and test script, and the tests are passing now.

I noticed the documentation (pdf_classification.ipynb) instructs users to manually install these tools on Windows. Just checking: is there any long-term plan to handle these dependencies more automatically (e.g. conda packages or setup checks), or should we continue to rely on manual user installation as the standard approach?

shchur:
Manual installation should be fine for now

PR #5597, opened 2026-02-17
Add test_pandas_utils.py

Innixma:
*Issue #, if available:*

*Description of changes:*

Add test_pandas_utils.py


Innixma:
> Why do we need this test, I see it is a memory estimate test. Do we have any OOM failures that we were not catching earlier in the CI ?

This is related to sanity checks in case we decide to change how the memory estimate works in future.
Currently the memory estimate seems to underestimate memory usage for categorical features in several cases that still need further deep dives, so I want to ensure that we track exactly how the logic is changed when I attempt to fix it later.

PR #5601, opened 2026-02-19
[tabular] Add TabICLv2 support

### LINKED ISSUE #5603: Add support for TABICLv2 (with regression support) in AutoGluon Tabular
## Description

This proposal refers to the **`tabular`** module.

I would like to request support for **TABICLv2** in AutoGluon Tabular.

TABICLv2 is the successor of TabICL and introduces architectural improvements and stronger performance. Importantly, **v2 now supports both classification and regression**, whereas earlier versions focused only on classification. 

## References

* Official repository (includes TABICLv2 implementation and pretrained weights):
  [https://github.com/soda-inria/tabicl](https://github.com/soda-inria/tabicl)

* TABICLv2 paper:
  [https://arxiv.org/abs/2602.11139](https://arxiv.org/abs/2602.11139)

* Original TabICL paper:
  [https://arxiv.org/abs/2303.00997](https://arxiv.org/abs/2303.00997)

Innixma:
Resolves #5603 

New model: Add TabICLv2 support

Paper: TabICLv2: A better, faster, scalable, and open tabular foundation model
Authors: Jingang Qu, David Holzmüller, Gaël Varoquaux, Marine Le Morvan
https://arxiv.org/abs/2602.11139

I did initial testing on TabArena and was able to perfectly replicate the results obtained by running the TabArena implementation of TabICLv2 (roughly equivalent code).

This PR will make the AutoGluon TabICL model default to using the TabICLv2 model weights. If users want to still use TabICLv1, they will need to specify the checkpoint as a model hyperparameter.

TODO:
- [ ] Add quantile support in a follow-up PR.

### DIFF
--- a/tabular/src/autogluon/tabular/models/tabicl/tabicl_model.py
+++ b/tabular/src/autogluon/tabular/models/tabicl/tabicl_model.py
@@ -152,6 +152,7 @@ class TabICLModel(AbstractTorchModel):
                 # TODO: Instead of caps, should we subsample for large datasets?
                 "max_rows": 1000000,  # TODO: What should be the cap? 1 million rows works, but unsure if it is good
                 "max_features": 2000,  # TODO: What should be the cap? 10k features works, but unsure if it is good
+                "max_batch_size": 1024,  # avoid excessive VRAM usage
             }
         )
         return default_auxiliary_params


tabular/src/autogluon/tabular/models/tabicl/tabicl_model.py
```diff
@@ -115,15 +149,17 @@ def _get_default_auxiliary_params(self) -> dict:
         default_auxiliary_params = super()._get_default_auxiliary_params()
         default_auxiliary_params.update(
             {
-                "max_rows": 30000,
-                "max_features": 2000,
+                # TODO: Instead of caps, should we subsample for large datasets?
+                "max_rows": 1000000,  # TODO: What should be the cap? 1 million rows works, but unsure if it is good
+                "max_features": 2000,  # TODO: What should be the cap? 10k features works, but unsure if it is good
+                "max_batch_size": 1024,  # avoid excessive VRAM usage
```
prateekdesai04, line 155: how do we determine the batch_size here ?

Innixma: This is an outer loop of the batch size, basically lowest value that doesn't slow down inference speed

PR #5613, opened 2026-03-06
[Tabular] Workaround for not enough sampels per class

LennartPurucker:
It seems that our code does not handle cases where we do not have enough samples for a minority class, leading to a crash when evaluating a model during cross-validation.

Here is the code that fails with the current mainline, but not the code from the PR:

```python
import pandas as pd
import numpy as np

# tiny synthetic binary dataset: 45 majority (0) and 5 minority (1)
rng = np.random.default_rng(0)
X = rng.normal(size=(555, 5))
labels = np.array([0] * 550 + [1] * 5)
df = pd.DataFrame(X, columns=[f"f{i}" for i in range(5)])
df["target"] = labels


predictor = TabularPredictor(
    label="target", problem_type="binary", eval_metric="roc_auc"
)
predictor.fit(
    train_data=df,
    presets="best",
    raise_on_model_failure=True,
	adapt_num_bag_folds_to_n_classes=True,
)
predictor.leaderboard(display=True)
```

The changes in this PR ensure we adjust the number of folds to avoid failures. Moreover, we raise an error when we detect insufficient data for cross-validation. 

I recall that this should not be an issue given our oversampling during cross-validation. But this does not seem to be the case. Happy to dive deeper into this later. But I think a fix at an early stage would anyhow be a better idea. 


Innixma:
Indeed this shouldn't be the case, I need to look a little closer here

Innixma (approved):
LGTM! This is a tricky scenario, we probably will want to revisit at some point but this solution at least catches most of the failure cases.

PR #5631, opened 2026-03-31
refactor: updated torch range version and timeseries/setup.py to separate catboost dependencies 

DorotaDR:
Assisted by: Cursor

***Issue** #, if available:*
PyTorch update request: https://github.com/autogluon/autogluon/issues/5611 

***Description of changes:***

The proposed PR solves dependency issues with torch 2.10.1. The torch 2.10.1 contains CVE fix as reported in https://github.com/autogluon/autogluon/issues/5611 . 

The PR also removes the CatBoost from the main install_requires and adds it to extras_require.  The CatBoost only publishes wheels for x86_64 and aarch64. This makes the entire timeseries module uninstallable on s390x or ppc64le, even though the module does not use CatBoost directly; it only inherits it via the tabular ML backends. Making CatBoost optional corrects an architectural inconsistency between the Tabular and Timeseries modules.

The lightning less restrictive requirements are helpful with solving installation conflicts. The AutoGluon uses a very stable API surface of Lightning (e.g., EarlyStopping, Timer, and basic logging). Analysis of the Lightning 2.6 changelog confirms no breaking changes to these specific APIs, meaning the < 2.6 bound is unnecessarily restrictive.

I've added all changes in one PR. If you want I can split into smaller, separate PRs. Let me just know. 

_By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice._


DorotaDR:
Some CI tests for timeseries failed I will investigate those changes to see if they are caused by the dependency changes introduced by me.

shchur:
Hi @DorotaDR, thanks for the PR and sorry for the slow response. I have addressed the torch version range bump in https://github.com/autogluon/autogluon/pull/5724. We fixed the checkpoint loading issue upstream in `gluonts`, so the fix on AutoGluon's side is not required.

Regarding CatBoost / unavailability of `autogluon.timeseries` on some architectures: I have opened a separate issue to track this: https://github.com/autogluon/autogluon/issues/5732. Most likely just adding a `; platform_machine in "x86_64 aarch64"` marker on the `tabular[catboost]` requirement is sufficient.

Issue #5641, opened 2026-04-13
[BUG] LightGBMPrep internal feature engineering makes feature_importance() unusable for all L2+ models when using best_quality_v150


celestinoxp:
**Description**
Calling predictor.feature_importance() crashes with a MemoryError for any model at L2 or above when the best_quality_v150 preset is used. This includes the default best model (WeightedEnsemble_L3) and all stack ensemble models.

The root cause is the LightGBMPrep models introduced in v1.5.0. These models contain an internal prep_params pipeline (ArithmeticFeatureGenerator, CategoricalInteractionFeatureGenerator, OOFTargetEncodingFeatureGenerator) that expands features combinatorially — O(n²) for numeric features.

The crash does not only happen when calling feature_importance() directly on a LightGBMPrep model. It happens on any L2+ model that has a LightGBMPrep among its L1 base models, because feature_importance() must re-execute the full prediction pipeline — including all L1 base models — to reconstruct stacking features. When any LightGBMPrep is invoked in this process, its internal preprocessor expands the feature space and causes the crash.

With best_quality_v150, multiple LightGBMPrep models are included at L1, and virtually every L2+ model depends on them. This makes feature_importance() effectively broken for all practical use of this preset.
Steps to reproduce

```python
from autogluon.tabular import TabularPredictor

predictor = TabularPredictor(label="target").fit(
    train_data=train_data,    # ~1780 rows, 625 features
    presets="best_quality_v150",
    num_bag_folds=10,
    num_stack_levels=2,
)

# Crashes even on a non-Prep model at L2
predictor.feature_importance(
    data=train_data,
    model="NeuralNetTorch_r144_BAG_L2",
    subsample_size=200,
    num_shuffle_sets=1,
)
# MemoryError: Unable to allocate 1.86 GiB for an array
# with shape (4005, 125000) and data type float32

# Also crashes on the default best model
predictor.feature_importance(data=train_data)
# MemoryError: Unable to allocate 11.8 GiB for an array
# with shape (4005, 793880) and data type float32
```

Root cause — full call stack analysis
The NeuralNetTorch_r144_BAG_L2 model has 644 features — 625 original + 19 stacking features from 7 L1 base models. Its input space is completely normal. However, its base models include:

```python
stacker_info: {
  'num_base_models': 7,
  'base_model_names': [
    'CatBoost_c1_BAG_L1', 'LightGBM_r177_BAG_L1', 'LightGBM_r72_BAG_L1',
[...]

PR #5642, opened 2026-04-15
[Tabular] Add short-term feature_importance safeguard for LightGBMPrep pipelines

bluehyena:
*Issue*  
Addresses #5641.

*Description of changes:*  
This PR adds a short-term safeguard for `predictor.feature_importance()` when the requested model depends on `LightGBMPrep` in the prediction pipeline with active `prep_params`.

Instead of proceeding into a likely raw `MemoryError`, AutoGluon now emits a warning and raises a clearer explicit error with suggested workarounds.

This is intentionally a narrow short-term mitigation rather than a full solution for feature importance support on these pipelines.

*Validation:*
- `conda run -n ag-supported-311 python -m pytest tabular/tests/unittests/models/test_lightgbm_prep.py -q -k feature_importance` -> `2 passed, 1 deselected`
- `conda run -n ag-supported-311 python -m pytest tabular/tests/unittests/models/test_lightgbm_prep.py -q` -> `3 passed`

### DIFF
--- a/tabular/src/autogluon/tabular/trainer/abstract_trainer.py
+++ b/tabular/src/autogluon/tabular/trainer/abstract_trainer.py
@@ -63,6 +63,7 @@ from autogluon.core.utils.exceptions import (
 from autogluon.core.utils.feature_selection import FeatureSelector
 from autogluon.core.utils.loaders import load_pkl
 from autogluon.core.utils.savers import save_pkl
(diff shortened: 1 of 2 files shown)


Innixma:
Thanks for the contrib @bluehyena. I've decided to address this problem directly by fixing the cause of the OOM in #5645. Because #5645 should fix almost all cases where OOM would occur, I will close this PR.

PR #5644, opened 2026-04-15
Disable text and datetime feature types in unrelated preprocessors

Innixma:
*Issue #, if available:*

*Description of changes:*

- Disable text and datetime feature types in unrelated preprocessors
- For these preprcoessors, processing text and datetime features is out of scope and harms performance.
- Also exclude stacker features from these preprocessors, since it also wouldn't work well.


prateekdesai04 (approved):
LGTM!

PR #5645, opened 2026-04-16
[tabular] Reduce feature importance mem usage by 25x

### LINKED ISSUE #5641: [BUG] LightGBMPrep internal feature engineering makes feature_importance() unusable for all L2+ models when using best_quality_v150

Innixma:
*Issue #, if available:*

Replaces #5642
Resolves #5641

*Description of changes:*

- Previously, TabularPredictor.feature_importance would generate batches of up to 2.5M samples for inference at a time, causing a large memory usage in cases where later in the pipeline the data is given many additional features, such as with PrepLightGBM. This caused OOM errors for users.
- Added max_rows_per_batch as an argument to the feature_importance method so users can control this value. Also made the default 100,000 (25x lower), for a 25x reduction in memory usage by default. Also added max_memory_usage argument and reduced default to 0.1 from 0.2. This should now require <5GB memory even in extreme cases.

This change actually speeds up feature_importance by ~50% in my testing, because 2.5M rows was actually slower to process in some cases than 25 batches of 100k rows, while reducing peak memory usage by 25x. 

### DIFF
--- a/core/src/autogluon/core/utils/utils.py
+++ b/core/src/autogluon/core/utils/utils.py
@@ -1099,7 +1099,14 @@ def _compute_mean_stddev_and_p_value(values: list):
     return mean, stddev, p_value, n
 
 
-def _get_safe_fi_batch_count(X, num_features, X_transformed=None, max_memory_ratio=0.1, max_feature_batch_count=None, max_rows_per_batch=100_000):
+def _get_safe_fi_batch_count(
+    X,
+    num_features,
+    X_transformed=None,
+    max_memory_ratio=0.1,
+    max_feature_batch_count=None,
+    max_rows_per_batch=100_000,
+):
     if max_rows_per_batch is None:
         max_rows_per_batch = 100000
     # calculating maximum number of features that are safe to process in parallel
--- a/core/src/autogluon/core/utils/utils.py
+++ b/core/src/autogluon/core/utils/utils.py
@@ -770,6 +770,9 @@ def compute_permutation_feature_importance(
     log_prefix="",
     importance_as_list=False,
     random_state=0,
+    max_memory_ratio=0.1,
+    max_feature_batch_count=None,
+    max_rows_per_batch=None,
     **kwargs,
 ) -> pd.DataFrame:
     """
@@ -925,10 +928,21 @@ def compute_permutation_feature_importance(
                     )
 
                 if transform_func is None:
-                    feature_batch_count = _get_safe_fi_batch_count(X=X, num_features=num_features)
+                    feature_batch_count = _get_safe_fi_batch_count(
+                        X=X,
+                        num_features=num_features,
+                        max_memory_ratio=max_memory_ratio,
+                        max_feature_batch_count=max_feature_batch_count,
+                        max_rows_per_batch=max_rows_per_batch,
+                    )
                 else:
                     feature_batch_count = _get_safe_fi_batch_count(
-                        X=X, num_features=num_features, X_transformed=X_transformed
+                        X=X,
+                        num_features=num_features,
+                        X_transformed=X_transformed,
+                        max_memory_ratio=max_memory_ratio,
+                        max_feature_batch_count=max_feature_batch_count,
+                        max_rows_per_batch=max_rows_per_batch,
                     )
 
             # creating copy of original data N=feature_batch_count times for parallel processing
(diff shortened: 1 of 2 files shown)


core/src/autogluon/core/utils/utils.py
```diff
@@ -925,10 +928,21 @@ def compute_permutation_feature_importance(
                     )
 
                 if transform_func is None:
-                    feature_batch_count = _get_safe_fi_batch_count(X=X, num_features=num_features)
```
abdulfatir, line 928: Just to confirm: Previously `max_memory_ratio` and `max_feature_batch_count` weren't actually being used?

abdulfatir (approved):
Thanks @Innixma! Overall looks good to me, but I am wondering how this would affect other models than LGBMTabPrep? Would this lead to any regressions in other models which do not add a lot of additional features?

core/src/autogluon/core/utils/utils.py
Innixma: max_memory_ratio was `0.2` in _get_safe_fi_batch_count by default, and `max_feature_batch_count` was 2_500_000

Innixma:
@abdulfatir this is equivalent logic, it produces the exact same feature importance values, and the feature batching was just a way to pass larger row counts into the models for faster throughput. I tested also on a non-TabPrep model and I also still saw a 50% speedup. My guess is making a massive pandas DataFrame via many concat calls starts to get a large overhead at a certain point, so the speedup is likely due to reducing this concat overhead.

PR #5650, opened 2026-04-19
[tabular] Fix TabPrep running the feature generator twice

Innixma:
*Issue #, if available:*

Bug originated in #5588 (Feb 10th)

*Description of changes:*

As part of the logic post-1.5 to enable TabPrep for any model, the method `get_preprocessor` was created for AbstractModel which creates the TabPrep feature generator.

This silently changed PrepLightGBM's behavior because PrepLightGBM wraps `ModelAgnosticPrepMixin` which also has a `get_preprocessor` method used to create the TabPrep feature generator.

This caused the call in AbstractModel to call ModelAgnosticPrepMixin.get_preprocessor for PrepLightGBM, leading to a scenario where the original data is transformed by TabPrep, and then the transformed data is transformed again, leading to a huge increase in feature count and worse performance.

This PR fixes this bug by renaming ModelAgnosticPrepMixin.get_preprocessor to `get_preprocessor_tabprep` so it doesn't overwrite the `get_preprocessor` method.

### DIFF
--- a/tabular/src/autogluon/tabular/models/tabprep/prep_mixin.py
+++ b/tabular/src/autogluon/tabular/models/tabprep/prep_mixin.py
@@ -163,7 +163,7 @@ class ModelAgnosticPrepMixin:
         else:
             raise ValueError(f"Invalid value for prep_param: {prep_param}")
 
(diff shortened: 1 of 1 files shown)


LennartPurucker (approved):
LGTM!

PR #5654, opened 2026-04-20
Add support for Pandas 3.0

celestinoxp:
## Description
This PR enables full functional compatibility for AutoGluon with Pandas 3.0.2. 
Extensive benchmarking confirms that AutoGluon Tabular maintains performance parity (and shows slight improvements in inference) under the new Pandas version.
### Key Changes
- **Frequency Alias Mapping**: Added automatic mapping for legacy frequency aliases (`M`, `Q`, `Y`, `H`, etc.) to new Pandas standard formats (`ME`, `QE`, `YE`, `h`) in `autogluon.common`.
- **Downcast Argument Removal**: Patched `fillna` calls across Tabular and Feature generators to avoid errors from the removed `downcast` parameter in Pandas 3.0.
- **Datetime Resolution & Timezone Fix**: Updated `DatetimeFeatureGenerator` to be resolution-aware (`us` vs `ns`) and strictly handle localized bounds (UTC), preventing `TypeErrors` during naive vs aware timezone comparisons in Pandas 3.0.
- **Type Check Updates**: Replaced legacy `np.issubdtype` checks with `pd.api.types` for improved compatibility with new Pandas/Numpy types.
- **TimeSeries Compatibility**: Fixed `pd.Period` construction and frequency handling in the TimeSeries module.
- **Test Suite Robustness**: Improved `conftest.py` settings in `common` and `core` to gracefully skip missing `mypy` plugin arguments, ensuring tests run smoothly in different environments.
- **Dependency Update**: Updated `core` setup to officially allow `pandas < 3.1.0`.
### Benchmark Results (50k rows, 50 features)
*Note: Results obtained on a personal laptop; absolute times may vary across different hardware and environments.*
| Metric | Pandas 2.3.3 | Pandas 3.0.2 | Difference |
| :--- | :--- | :--- | :--- |
| **Fit Time (avg)** | 69.29s | 64.81s | -6.5% |
| **Inf Time** | 2.40s | 2.17s | -9.6% |


Innixma:
Thanks for the PR @celestinoxp! I am busy traveling this week, but will try to review when I can.

features/src/autogluon/features/generators/datetime.py
```diff
@@ -55,11 +56,18 @@ def normalize_timeseries(self, X: pd.DataFrame, feature: str, is_fit: bool) -> p
         #   Alternatives like Polars do not offer the same datetime conversion logic, and thus aren't valid to use.
         #   The runtime is approximately 0.08 seconds per 1000 rows in worst case.
         series = pd.to_datetime(X[feature].copy(), utc=True, errors="coerce", format="mixed")
+        if PANDAS_V3_OR_NEWER:
+            # Force ns resolution to match legacy behavior (coerce out-of-bounds to NaT)
+            # Year 2700 is valid in 'us' but not in 'ns'.
```
shchur, line 61: Why is this needed? Are there some downstream users that will break on timestamps outside of the `min_ts, max_ts` range?

common/src/autogluon/common/utils/pandas_utils.py
```diff
@@ -2,11 +2,18 @@
...
+import pandas as pd
+from packaging import version
 from pandas import DataFrame
 
 from ..features.infer_types import get_type_map_raw
 from ..features.types import R_CATEGORY, R_FLOAT, R_INT
 
+PANDAS_VERSION = version.parse(pd.__version__)
+PANDAS_V2_OR_NEWER = PANDAS_VERSION >= version.parse("2.0.0")
+PANDAS_V2_1_OR_NEWER = PANDAS_VERSION >= version.parse("2.1.0")
```
shchur, line 14: I can't seem to find any occurrences for PANDAS_V2_OR_NEWER or PANDAS_V2_1_OR_NEWER. Why do we need these variables?

timeseries/tests/unittests/common.py
```diff
@@ -91,6 +91,20 @@ def to_supported_pandas_freq(freq: str) -> str:
...
+            "Y": "YE",
+            "A": "YE",
+            "SM": "SME",
+            "H": "h",
+            "T": "min",
+            "S": "s",
+            "L": "ms",
+            "U": "us",
+            "N": "ns",
+        }.get(freq, freq)
```
shchur, line 107: This is unnecessary, the inputs to the method are only 2.2+ compatible strings

features/src/autogluon/features/generators/text_ngram.py
```diff
@@ -229,10 +229,10 @@ def _generate_ngrams(self, X, downsample_ratio: int = None):
         X_nlp_features_combined = []
         for nlp_feature, vectorizer_fit in zip(self.vectorizer_features, self.vectorizers):
             if nlp_feature == "__nlp__":
-                X_str = X.astype(str)
+                X_str = X.astype(object).fillna('nan').astype(str)
                 text_data = [". ".join(row) for row in X_str.values]
             else:
-                nlp_feature_str = X[nlp_feature].astype(str)
+                nlp_feature_str = X[nlp_feature].astype(object).fillna('nan').astype(str)
```
shchur, line 235: Let's maybe move this to a private helper method in this file?

features/src/autogluon/features/generators/fillna.py
```diff
@@ -62,14 +62,18 @@ def _transform(self, X: DataFrame) -> DataFrame:
                 #  but we need it to avoid incorrect type conversion.
                 #  Pandas authors may have not considered our edge-case.
                 #  We specifically want to have an object dtype not be converted to a numeric dtype,
```
shchur, line 64: Is this comment still true for >=3.0?

features/tests/features/generators/test_fillna.py
```diff
@@ -2,25 +2,41 @@
...
+                "int": np.nan,
+                "float": np.nan,
+                "obj": "",
+                "cat": np.nan,
+                "datetime": np.nan,
+                "text": "",
+                "datetime_as_object": "",
+            },
+            downcast=False,
+        )
```
shchur, line 39: This would be cleaner with something like
```python
  fill_map = {
      "int": np.nan,
      "float": np.nan,
      "obj": "",
      "cat": np.nan,
      "datetime": np.nan,
      "text": "",
      "datetime_as_object": "",
  }
  # pandas 3 removed `downcast=`; silent downcasting is off by default, so object
  # columns stay object without it — same intended behaviour.
  extra_kwargs = {} if PANDAS_V3_OR_NEWER else {"downcast": False}
  expected_output_data = input_data.fillna(fill_map, **extra_kwargs)
```

tabular/src/autogluon/tabular/models/fastainn/tabular_nn_fastai.py
```diff
@@ -239,7 +239,12 @@ def _fill_missing(self, df: pd.DataFrame) -> pd.DataFrame:
                 # TODO: Remove warning later as a follow up to https://github.com/autogluon/autogluon/pull/3734
                 with warnings.catch_warnings():
                     warnings.simplefilter(action="ignore", category=FutureWarning)
-                    df = df.fillna(column_fills, inplace=False, downcast=False)
```
shchur, line 242: fillna with/without downcast seems to show up in multiple places. Should we put it into some helper method in `pandas_utils` to avoid the repeated imports of `PANDAS_V3_OR_NEWER`?

timeseries/src/autogluon/timeseries/models/gluonts/dataset.py
```diff
@@ -67,10 +68,13 @@ def _get_freq_for_period(freq: str) -> str:
             # Replace unsupported frequency "SME" with "2W"
             return "2W"
         elif freq_name == "bh":
-            # Replace unsupported frequency "bh" with dummy value "Y"
-            return "Y"
+            # Replace unsupported frequency "bh" with dummy value "Y"/"YE"
+            return "YE" if PANDAS_V3_OR_NEWER else "Y"
```
shchur, line 72: This is incorrect, `pd.Period` still takes `Y` not `YE` (unlike `pd.Timestamp`)

timeseries/src/autogluon/timeseries/models/gluonts/dataset.py
```diff
@@ -67,10 +68,13 @@ def _get_freq_for_period(freq: str) -> str:
...
         else:
-            freq_name_for_period = {"YE": "Y", "QE": "Q", "ME": "M"}.get(freq_name, freq_name)
+            if PANDAS_V3_OR_NEWER:
+                freq_name_for_period = freq_name
+            else:
+                freq_name_for_period = {"YE": "Y", "QE": "Q", "ME": "M"}.get(freq_name, freq_name)
```
shchur, line 77: Same as above, this is incorrect

timeseries/tests/unittests/test_ts_dataset.py
```diff
@@ -262,8 +263,11 @@ def test_when_dataset_constructed_from_iterable_with_freq_then_freq_is_inferred(
 @pytest.mark.parametrize("start_time, freq", FREQ_TEST_CASES)
 def test_when_dataset_constructed_via_constructor_with_freq_then_freq_is_inferred(start_time, freq):
     freq = to_supported_pandas_freq(freq)
-    # Period requires freq=M for ME frequency
-    start_period = pd.Period(start_time, freq={"ME": "M"}.get(freq))
+    # Period requires freq=M for ME frequency in pandas < 3.0
+    if PANDAS_V3_OR_NEWER:
+        start_period = pd.Period(start_time, freq=freq)
+    else:
+        start_period = pd.Period(start_time, freq={"ME": "M"}.get(freq))
```
shchur, line 270: This is incorrect, `pd.Period` expects `M` not `ME`.

common/src/autogluon/common/utils/pandas_utils.py
```diff
@@ -58,3 +65,41 @@ def get_approximate_df_mem_usage(df: DataFrame, sample_ratio=0.2):
             )
             memory_usage = memory_usage_inexact.combine_first(memory_usage)
         return memory_usage
+
+
+def pandas_freq_alias(freq: str) -> str:
```
shchur, line 70: I think this method can be removed completely. Currently it's used in `timeseries` only, where the `freq` is automatically inferred from the data with `pd.infer_freq`, which already returns the correct version of the aliases. The only reason to add it would be in a few methods like `TimeSeriesDataFrame.convert_frequency` which will error in `pandas>=3.0` for an unsupported alias. I think that's reasonable since we don't want to forever support the deprecated aliases.

common/src/autogluon/common/features/infer_types.py
```diff
@@ -18,11 +18,11 @@ def get_type_family_raw(dtype) -> str:
             return "category"
         if "datetime" in dtype.name:
             return "datetime"
-        if "string" in dtype.name:
+        if "string" in dtype.name or dtype.name == "str":
```
shchur, line 21: Maybe just `if "str" in dtype.name:`?

shchur (commented):
Thanks a lot @celestinoxp for the detailed PR! I left a few comments, please ping me to approve the CI run once they are addressed.

shchur:
Hi @celestinoxp, looks like there are several upstream dependencies that are blocking the pandas-3 upgrade for AutoGluon (StatsForecast, MLForecast, GluonTS), so we won't even be able to properly test pandas-3 in our CI unless 1/ the upstream dependencies release a new version with pandas support or 2/ we remove them from our dependencies. 

Based on this, I think it makes sense to wait until after the AutoGluon v1.6.0 release for the pandas-3 upgrade.

celestinoxp:
@shchur Any news?

common/src/autogluon/common/utils/pandas_utils.py
celestinoxp: Initially, the program was designed to support all versions, meaning it was prepared for the new features of each version.

shchur:
@celestinoxp Tracking this in #5730, it looks like https://github.com/Nixtla/statsforecast/issues/1182 is still unresolved, which blocks us

Issue #5660, opened 2026-04-28
[TimeSeries] Accept tabular-style full metric names in eval_metric


Mateusz-Switala:
## Description
Users switching between `TabularPredictor` and `TimeSeriesPredictor` must learn two different naming conventions for the same underlying metrics. Tabular uses full lowercase names (`mean_absolute_error`, `root_mean_squared_error`), while TimeSeries uses uppercase acronyms (`MAE`, `RMSE`). This creates unnecessary friction and is a common source of confusion for users who work across both modules.

## Current behavior

```python
# Tabular — full lowercase names
TabularPredictor(...).fit(..., eval_metric="mean_absolute_error")

# TimeSeries — must know the acronym; full name raises a confusing error
TimeSeriesPredictor(...).fit(..., eval_metric="mean_absolute_error")
# ValueError: Time series metric mean_absolute_error not supported.
# Available metrics are: ['MAE', 'MAPE', 'MASE', ...]
```

## Proposed behavior

```python
# Both now accepted by TimeSeriesPredictor
TimeSeriesPredictor(...).fit(..., eval_metric="mean_absolute_error")    # resolves to MAE
TimeSeriesPredictor(...).fit(..., eval_metric="root_mean_squared_error") # resolves to RMSE
TimeSeriesPredictor(...).fit(..., eval_metric="weighted_quantile_loss")  # resolves to WQL
```

All existing acronym inputs (`"MAE"`, `"RMSE"`, etc.) continue to work without any change.

## Implementation

Add a `METRIC_ALIASES` constant to `autogluon.timeseries.metrics` that maps full lowercase names to their canonical acronyms, and consult it at the start of `check_get_evaluation_metric()` — before the existing `.upper()` conversion:

```python
METRIC_ALIASES: dict[str, str] = {
    "mean_absolute_error": "MAE",
    "mean_squared_error": "MSE",
    "root_mean_squared_error": "RMSE",
    "root_mean_squared_logarithmic_error": "RMSLE",
    "mean_absolute_percentage_error": "MAPE",
    "symmetric_mean_absolute_percentage_error": "SMAPE",
    "mean_absolute_scaled_error": "MASE",
    "root_mean_squared_scaled_error": "RMSSE",
    "weighted_absolute_percentage_error": "WAPE",
    "weighted_quantile_loss": "WQL",
    "scaled_quantile_loss": "SQL",
}
```

One-line change in `check_get_evaluation_metric()`:

```python
# Before
[...]

PR #5661, opened 2026-04-28
feat: align timeseries eval_metric format with tabular

Mateusz-Switala:
Assisted-by: Claude Code

*Issue # , if available:*
#5660 

*Description of changes:*
Adds `METRIC_ALIASES` to `autogluon.timeseries.metrics` — a dict that maps full lowercase metric names (as used in AutoGluon-Tabular) to their canonical uppercase timeseries acronyms. `check_get_evaluation_metric()` now consults this dict first, so users can pass either style interchangeably.

### Changes

**`timeseries/src/autogluon/timeseries/metrics/__init__.py`**

- Added `METRIC_ALIASES: dict[str, str]` constant mapping 11 full lowercase names to their acronyms:

  | Full name (tabular style) | Acronym (timeseries style) |
  |---|---|
  | `mean_absolute_error` | `MAE` |
  | `mean_squared_error` | `MSE` |
  | `root_mean_squared_error` | `RMSE` |
  | `root_mean_squared_logarithmic_error` | `RMSLE` |
  | `mean_absolute_percentage_error` | `MAPE` |
  | `symmetric_mean_absolute_percentage_error` | `SMAPE` |
  | `mean_absolute_scaled_error` | `MASE` |
  | `root_mean_squared_scaled_error` | `RMSSE` |
  | `weighted_absolute_percentage_error` | `WAPE` |
  | `weighted_quantile_loss` | `WQL` |
  | `scaled_quantile_loss` | `SQL` |

  When multiple timeseries metrics share the same tabular equivalent (e.g. `MAE`, `MASE`, and `WAPE` all correspond to `mean_absolute_error`), the simplest non-scaled metric is chosen as the alias target.

- One-line change in `check_get_evaluation_metric()` to consult `METRIC_ALIASES` before the existing `.upper()` + `DEPRECATED_METRICS` lookup:

  ```python
  # Before
  metric_name = DEPRECATED_METRICS.get(eval_metric, eval_metric).upper()

  # After
  metric_name = METRIC_ALIASES.get(eval_metric, DEPRECATED_METRICS.get(eval_metric, eval_metric)).upper()
  ```

- `METRIC_ALIASES` added to `__all__` so it is part of the public API.

**`timeseries/tests/unittests/test_metrics.py`**

- Added `test_given_tabular_style_full_name_check_get_eval_metric_resolves_to_correct_acronym` — 11 parametrized cases, one per alias entry, asserting that each full name resolves to the expected acronym.

### Why there is no regression

- `METRIC_ALIASES` keys are exclusively lowercase full names. None of them were valid inputs before this change (all raised `ValueError`), so no existing call site is affected.
- Existing acronym inputs (`"MAE"`, `"mae"`, `"Mae"`) miss the `METRIC_ALIASES` lookup and fall through to the unchanged `.upper()` path.
- `DEPRECATED_METRICS` is still consulted in the same position in the lookup chain.

### DIFF
--- a/timeseries/src/autogluon/timeseries/metrics/__init__.py
+++ b/timeseries/src/autogluon/timeseries/metrics/__init__.py
@@ -24,6 +24,8 @@ __all__ = [
     "WAPE",
     "WCD",
     "WQL",
+    "AVAILABLE_METRICS",
+    "METRIC_ALIASES",
 ]
 
 DEFAULT_METRIC_NAME = "WQL"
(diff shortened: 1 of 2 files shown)


timeseries/src/autogluon/timeseries/metrics/__init__.py
```diff
@@ -42,6 +44,23 @@
     "MAE": MAE,
 }
 
+# Maps full lowercase metric names (as used in autogluon.tabular) to their canonical uppercase acronyms.
+# When multiple time series metrics share the same tabular equivalent (e.g. MAE/MASE/WAPE all map to
+# mean_absolute_error), the simplest non-scaled metric is preferred here.
```
shchur: This seems overly verbose. How about something like this
```suggestion
# Provide lowercase aliases for metrics for consistency with autogluon.tabular
```

shchur (approved):
Thanks!

PR #5663, opened 2026-04-30
feat: relax strict requirements for patch version when loading predictor

Mateusz-Switala:
Assisted-by: Claude Code

*Issue #, if available:*
https://github.com/autogluon/autogluon/issues/5662

*Description of changes:*

`check_saved_predictor_version `previously raised AssertionError on any version mismatch, including patch-only differences (e.g. 1.2.0 →
   1.2.1). This forced users to choose between retraining after every patch release or disabling all version protection with
  `require_version_match=False`.

  The check is now two-tier, consistent with the existing Python version check:

  - Patch-only difference between two stable releases (same major.minor, no pre-release or dev marker) — logs at INFO, load proceeds.
  - Major/minor difference, pre-release, dev, or unparseable version — WARNING banner + AssertionError as before.
  - require_version_match=False — bypasses all checks as before.

  compare_autogluon_metadata applies the same logic: patch-only version differences are logged at INFO (level 20) rather than WARNING
  (level 30), and the summary header level is now the maximum of the individual entry levels rather than always WARNING.

  Files changed:
  - common/src/autogluon/common/utils/utils.py — updated check_saved_predictor_version and compare_autogluon_metadata
  - tabular/src/autogluon/tabular/predictor/predictor.py — updated require_version_match docstring in TabularPredictor.load()
  - timeseries/src/autogluon/timeseries/predictor.py — updated require_version_match docstring in TimeSeriesPredictor.load()
  - common/tests/unittests/utils/test_compare_autogluon_metadata.py — added CheckSavedPredictorVersionTestCase covering exact match,
  patch-only (both directions), minor/major/pre-release/unparseable version mismatches, and metadata log-level assertions

### DIFF
--- a/common/src/autogluon/common/utils/utils.py
+++ b/common/src/autogluon/common/utils/utils.py
@@ -11,6 +11,7 @@ from typing import Any, Optional
 
 import numpy as np
 import pandas as pd
+from packaging.version import InvalidVersion, Version
 
 from ..version import __version__
 from .random import get_numpy_seed as _get_numpy_seed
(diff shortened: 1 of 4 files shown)


Innixma:
Closing this PR as we do not promise cross-version compatibility in AutoGluon. For more details refer to https://github.com/autogluon/autogluon/issues/5662#issuecomment-5078284798.

PR #5668, opened 2026-05-15
[Tabular] Small Fixes to AutoGluon Code from TabArena Benchmarks

LennartPurucker:
The fixes include:

* Maint: some documentation changes.
* Add: allow to change verbosity only for the model, not for all of AutoGluon.
* Add: support for disabling ray retries on model failure during cross-validation.

Also, some changes to existing models:
* CAT: add support for using CatBoost without AutoGluon's preprocessing when unseen categories exist; minor changes to its memory estimation.
* XGB: add support for using XGBoost without AutoGluon's preprocessing when unseen categories exist.
* LR: improved memory estimation for large data.
* RF: improved memory estimation for large data.

---


core/src/autogluon/core/models/ensemble/fold_fitting_strategy.py
```diff
@@ -638,7 +638,8 @@ def __init__(
         self.fit_num_cpus = None
         self.fit_num_gpus = None
         # max_calls to guarantee release of gpu resource
-        self._ray_fit = self.ray.remote(max_calls=1)(_ray_fit)
+        max_retries = 0 if os.getenv("RAY_DISABLE_RETRIES") == "1" else 3 # 3 is ray default
+        self._ray_fit = self.ray.remote(max_calls=1, max_retries=max_retries)(_ray_fit)
```
Innixma: Would be better to avoid hard-coding 3 here in case the env var is not set.

tabular/src/autogluon/tabular/models/catboost/catboost_model.py
```diff
@@ -126,7 +145,7 @@ def _estimate_memory_usage_static(
         histogram_effective_depth = max(min(depth + 1, 7), depth)
 
         # Formula based on manual testing, aligns with LightGBM histogram sizes
-        histogram_mem_usage_bytes = 24 * math.pow(2, histogram_effective_depth) * len(X.columns) * border_count
+        histogram_mem_usage_bytes = 24 * math.pow(2, histogram_effective_depth) * (1.1 * len(X.columns)) * border_count
```
Innixma, line 147: Should this 1.1 be put somewhere else?

LennartPurucker: The problem was that the mem estimate was wrong for very large feature counts (like 10k+), so this worked well. Not sure which of the other parameters would be better, but I also don't know all of the logic behind this for Catboost (besides the code being from LGBM)

tabular/src/autogluon/tabular/models/xgboost/xgboost_model.py
```diff
@@ -73,9 +76,28 @@ def _preprocess(self, X, is_train=False, max_category_levels=None, **kwargs):
...
+        else:
+            # FIXME: same code as in RealMLP, make it a general function in the future.
+            # Avoid bad dtype for cat categories in later ordinal encoding.
+            # Maps unseen categories to a new high integer.
+            if self._cat_col_names is not None:
+                if self._category_mapping is None:
+                    self._category_mapping = {}
+                    for col in self._cat_col_names:
+                        cats = X[col].cat.categories
+                        self._category_mapping[col] = {cat: code for code, cat in enumerate(cats)}
```
Innixma: Better to move so all the is_train logic is in the same block?

Innixma (approved):
LGTM!

tabular/src/autogluon/tabular/models/catboost/catboost_model.py
Innixma: I'm more saying that 1.1 could be multiplied in the 24 to be a single number

tabular/src/autogluon/tabular/models/xgboost/xgboost_model.py
LennartPurucker: Done

tabular/src/autogluon/tabular/models/catboost/catboost_model.py
LennartPurucker: Ah yes, I did that before, just looked weird / was hard to interpret so I would want to stick with this

LennartPurucker (commented):
Also LGTM if test pass

PR #5677, opened 2026-06-04
perf: lazy-import networkx, requests, tqdm to speed up `import TabularPredictor` (~25% faster)

Innixma:
## Summary

Several heavy/utility third-party libraries were imported at module top-level in the `TabularPredictor` import chain and loaded on **every** `from autogluon.tabular import TabularPredictor`, even though each is only needed in a narrow, rarely-hit code path. This PR defers them to function-local (lazy) imports, cutting cold-import time by **~0.7 s (~22%)** with no behavioral change.

Profiling was done with `python -X importtime` and verified with fresh-subprocess wall-clock benchmarks.

### 1. `networkx` — the dominant win (~0.6 s)

Imported at top-level in `TabularPredictor` and both the core and tabular abstract trainers, but only used in fit-time / model-graph paths.

- `tabular/.../predictor/predictor.py` — lazy import in `plot_ensemble_model` (next to the existing lazy `pygraphviz`).
- `tabular/.../trainer/abstract_trainer.py` — lazy import in the 5 methods that use `nx`: `_construct_model_pred_order`, `_construct_model_pred_order_with_pred_dict`, `get_models_attribute_dict`, `leaderboard`, `delete_models`.
- `core/.../trainer/abstract_trainer.py` — lazy import in `__init__` and `get_minimum_model_set`.

### 2. `requests` and its networking stack (~0.07 s)

`common/loaders/load_pkl.py` and `common/loaders/_utils.py` imported `requests` at top-level, dragging in `urllib3` + `idna` + `charset_normalizer` + `certifi` + `simplejson`, but `requests` is only used to fetch a pickle/file over an `http(s)` URL. (`load_pkl` already lazy-imports `boto3`; this matches that pattern.)

- `common/.../loaders/load_pkl.py` — lazy import in `_load_pickle_from_url`.
- `common/.../loaders/_utils.py` — lazy import in `download`.

### 3. `tqdm` (~0.04 s)

Imported at top-level in four files but only used for progress bars in download / HPO loops.

- `common/.../utils/s3_utils.py` — lazy import in `download_s3_files`.
- `core/.../utils/files.py` — lazy import in `download` (next to the existing lazy `requests`).
- `core/.../scheduler/seq_scheduler.py` — lazy import in `run` (`tqdm.auto`).
- `common/.../loaders/_utils.py` — lazy import in `download` (off the bare-import path; fixed for consistency).

The common paths — importing the predictor, loading a local file, or loading from S3 (which uses `boto3`, not `requests`) — no longer pay for any of these.

## Verification

All four stages were benchmarked **back-to-back in a single session** (20 fresh-subprocess runs per stage, `from autogluon.tabular import TabularPredictor`), so the rows are directly comparable and monotonic:

| stage | min | median |
|---|---|---|
| baseline (all eager) | 3.090 s | 3.242 s |
| after networkx fix | 2.512 s | 2.601 s |
| after requests fix | 2.449 s | 2.534 s |
| after tqdm fix | 2.409 s | 2.526 s |
| **total saved** | **0.68 s (~22%)** | 0.72 s (~22%) |

`networkx` dominates (~0.6 s); `requests` (~0.07 s) and `tqdm` (~0.04 s) are smaller. The `tqdm` step is near the wall-clock noise floor, so its real proof is that `tqdm` is now deterministically **absent** from the `-X importtime` profile.

After all three fixes, `networkx`, the entire `requests` networking stack, and `tqdm` are absent from the `-X importtime` profile. The remaining import cost is scipy/pandas/sklearn/numpy, which are genuinely required at import time. A sweep confirmed all other heavy libraries (`torch`, `lightgbm`, `xgboost`, `catboost`, `tabpfn`, `transformers`, `matplotlib`, `ray`, `boto3`, …) are already lazy and never loaded by a bare import.

**Functional smoke tests pass:**
- networkx: fit a small predictor + weighted ensemble exercising `leaderboard(extra_info=True)`, `get_minimum_model_set`, `get_models_attribute_dict`, `delete_models`.
- requests: local pickle round-trip still works; the URL path raises a `requests.ConnectionError` on an unreachable host, confirming the lazy import resolves.
- tqdm: `download_s3_files([], dry_run=True)` exercises the lazy `from tqdm import tqdm` (no network); all touched modules import cleanly.

Note: these defer, not eliminate, the loads — `networkx` still loads when a trainer is created during `fit`, `requests`/`tqdm` when downloading. The win is specifically on time-to-`TabularPredictor`-importable (CLI responsiveness, `--help`, lazy-loading callers).

### DIFF
--- a/common/src/autogluon/common/loaders/_utils.py
+++ b/common/src/autogluon/common/loaders/_utils.py
@@ -10,7 +10,6 @@ from typing import Optional
 
 import boto3
 import numpy as np
-import requests
 import tqdm
 
 S3_PREFIX = "s3://"
@@ -136,6 +135,8 @@ def download(
     fname
         The file path of the downloaded file.
     """
+    import requests
+
     is_s3 = url.startswith(S3_PREFIX)
     if is_s3:
         s3 = boto3.resource("s3")
--- a/common/src/autogluon/common/loaders/_utils.py
+++ b/common/src/autogluon/common/loaders/_utils.py
@@ -10,7 +10,6 @@ from typing import Optional
 
 import boto3
 import numpy as np
-import tqdm
 
 S3_PREFIX = "s3://"
 
@@ -136,6 +135,7 @@ def download(
         The file path of the downloaded file.
     """
     import requests
+    import tqdm
 
     is_s3 = url.startswith(S3_PREFIX)
     if is_s3:
--- a/common/src/autogluon/common/loaders/load_pkl.py
+++ b/common/src/autogluon/common/loaders/load_pkl.py
@@ -6,8 +6,6 @@ import pickle
 from typing import Any
 from urllib.parse import urlparse
 
-import requests
-
 from ..utils import compression_utils, s3_utils
 
 logger = logging.getLogger(__name__)
@@ -86,6 +84,8 @@ def _is_web_url(path: str) -> bool:
 
 
 def _load_pickle_from_url(url: str):
+    import requests
+
     response = requests.get(url)
     response.raise_for_status()  # Raise an error for bad status codes
     return pickle.loads(response.content)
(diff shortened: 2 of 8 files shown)


shchur (approved):
Nice!

PR #5679, opened 2026-06-05
cleanup: remove unsupported "lite" mode

Innixma:
## Summary

Removes all "lite" mode (a.k.a. `autogluon.lite` / web-browser mode) logic, which has been unsupported for a long time. This is a **behavior-preserving** cleanup for normal installs and a foundation for simplifying the packaging/build setup.

Why it's a no-op for the supported path: `disable_if_lite_mode(...)` returned the wrapped function **unchanged** whenever not in lite mode, so removing the decorators changes nothing for the only supported (non-lite) path.

## Removed

- **`common/utils/lite.py`** and the `disable_if_lite_mode` decorator — 15 call sites across `common`/`core`/`features`/`tabular`, plus their imports.
- **`LITE_MODE` / `__lite__`** in `common/utils/utils.py`, and the `lite` key from `get_autogluon_metadata()`.
- **`core/_setup_utils.py`:**
  - the `AUTOGLUON_PACKAGE_NAME` env var (only existed to build under the `autogluon.lite` name) and `LITE_MODE`;
  - the lite-only `DEPENDENT_PACKAGES` filter (`psutil`/`Pillow`/`timm`);
  - the `__lite__` line written into the generated `version.py`.
  - `PACKAGE_NAME` is retained as the constant `autogluon`.
- The `if not ag.LITE_MODE else [...]` `install_requires` branches in `autogluon`/`common`/`core`/`features` `setup.py`.

**15 files, +37/−128.**

## Verification

- All touched modules import cleanly; `from autogluon.tabular import TabularPredictor` works.
- `setup.py` metadata still builds (`egg_info`); generated `version.py` no longer carries `__lite__`.
- `get_autogluon_metadata()` keys: `system, version, py_version, py_version_micro, packages, packages_invalid` (no `lite`).
- Pre-existing ruff findings (15× F541, 1× F401, 1× F841 — all confirmed present on `master`) were left untouched to keep the diff focused.

## Note

`get_autogluon_metadata()` no longer emits a `lite` field — the one externally observable change. Everything else is internal.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5684, opened 2026-06-05
[Prototype/RFC] PEP 621 + uv workspace for the full monorepo

Innixma:
> **Prototype/RFC, stacked on #5683 → … → #5679.** Base auto-retargets as the chain merges. Not for merge as-is — see open questions.

## Summary
Converts the **entire monorepo** to PEP 621 `pyproject.toml` + a **uv workspace**, building on the versioning simplification (#5683) that unblocked it.

All 7 distributions — `common`, `core`, `features`, `tabular`, `timeseries`, `multimodal`, and the meta `autogluon` — now:
- carry their metadata/deps/extras in `pyproject.toml` (PEP 420 `autogluon.*` namespace discovery), and
- keep a **thin `setup.py`** that supplies only the dynamic version (`base VERSION + AUTOGLUON_VERSION_SUFFIX`, single-sourced) and writes `version.py` for runtime — so the nightly `setup.py sdist` path and the version model are preserved.

### Highlights
- **tabular**: all 16 extras converted; the aggregates (`all`/`tabarena`/`tests`) use **self-referential extras** (`autogluon.tabular[...]`) instead of duplicated unions.
- **multimodal**: keeps its `configs/*.yaml` via `[tool.setuptools.package-data]`.
- **meta `autogluon`**: depends on the full stack + the `tabarena` extra.
- **Centralized caps**: the former `DEPENDENT_PACKAGES` caps move to one `[tool.uv] constraint-dependencies` at the workspace root; packages list those shared deps **bare** (package-specific caps like `gluonts`/`lightgbm` stay inline). Removes per-package cap duplication.
- **Root**: `[tool.uv.workspace]` members = all 7; `[tool.uv.sources]` maps the 6 sibling packages to the workspace.

## Verification
- **`uv lock` resolves the whole workspace in one pass — 301 packages, all 7 editable members**, self-referential extras, and constraints (incl. the full torch/transformers/gluonts/tensorrt tree).
- `uv sync --package autogluon.tabular` → `common`/`core`/`features`/`tabular` import; `TabularPredictor` works; `numpy` resolves to `2.1.3` (constraint `<2.5` enforced via constraint-dependencies).
- Full-workspace `uv sync` installs all 7 editable; `multimodal`/`timeseries`/meta + `MultiModalPredictor`/`TimeSeriesPredictor` all import.

## Open questions (for the real migration)
1. **Sibling pinning** — bare siblings resolve via workspace sources for dev; static PEP 621 can't carry the exact dynamic `==<version>` pin used today. Publish path needs a decision (range + `--pre` for nightlies, or dynamic deps). **Main thing to settle.**
2. **constraint-dependencies govern the lock, not published wheels' `Requires-Dist`.** The pip-from-PyPI path would lose the shared-dep caps unless re-injected at build (e.g. a thin-setup.py step) — decision needed.
3. **`uv.lock`** is gitignored upstream; a real migration should commit it.
4. **Thin `setup.py`** could later be dropped via a build-backend version plugin (pure PEP 621).
5. **`packages/` move** intentionally deferred to its own PR (purely cosmetic; rename-heavy).


Innixma:
## Decision needed: sibling pinning & where the caps live

Open questions #1 (sibling pins) and #2 (caps not in published wheels) are **the same underlying problem**, so I'm writing them up together.

### Root cause
Static PEP 621 `[project]` metadata can't carry **build-time-computed** values into a published wheel's `Requires-Dist`:
- sibling deps need the **exact dynamic build version** today — `autogluon.tabular==1.5.1b20260605` → `autogluon.core==1.5.1b20260605`;
- shared third-party caps need to land in `Requires-Dist` for the **`pip install` from PyPI** path (today `_setup_utils.get_dependency_version_ranges` bakes them in at build).

`[tool.uv.sources]` (workspace) and `[tool.uv] constraint-dependencies` solve both **for the dev/uv path** (editable workspace + `uv.lock`) — but **not** for what a normal PyPI user gets. The current prototype is therefore correct for `uv sync` but, as-is, would publish wheels with **bare/uncapped** third-party deps and **unpinned** siblings.

A hard setuptools constraint that shapes the options: `[project].dependencies` is **all-static or all-dynamic** — you can't keep third-party deps static in `pyproject` while injecting *only* the siblings dynamically.

### The product call that drives the choice
**How strict must inter-package pinning be on PyPI?** Today it's exact (`==<same build>`), matching AutoGluon's "no cross-version compatibility" stance. A range can't reproduce same-build/same-date matching:
- **stable**: `autogluon.core==1.5.*` works.
- **nightly**: nightlies publish together daily, so `==1.5.*` + `--pre` resolves to the *latest* (normally same-day) sibling — fine for "install the latest nightly", but pinning an *old* nightly date would pull a newer sibling (mismatch).
- ⚠️ PEP 440 gotcha: `~=1.5.1` / `>=1.5.1` **exclude** pre-releases, so a range must be `==1.5.*` (or a `.dev0` floor) **with `--pre`** to match nightlies at all.

If exact same-build pinning must hold → **Option B**. If `==1.5.* --pre` is acceptable → **Option A**.

### Options

| | **A — static deps (full PEP 621)** | **B — dynamic deps via thin setup.py** | **C — build hook** |
|---|---|---|---|
| Deps live in | `pyproject` (static) | `setup.py` (computed) | `pyproject` + custom hook |
| Sibling pins | range `==1.5.*` (relaxed) | exact `==<version>` (unchanged) | exact (injected) |
| Caps in published wheels | inline per package | via `get_dependency_version_ranges` | injected from one list |
| Cap de-duplication | duplicated (or a pre-commit sync); `constraint-dependencies` helps **dev only** | central in `_setup_utils` | central |
| uv workspace / `uv sync` | ✅ | ✅ (uv builds to resolve deps) | ✅ |
| Publish behavior change | **yes** (pins relaxed) | **none** | none |
| "Pure PEP 621" deps | ✅ | ✗ (deps stay in setup.py) | ~ |
| Effort / risk | low / medium (pin semantics) | low / low | high |

- **A** is what this prototype is closest to. To make it *publishable*, restore **inline caps** on the third-party deps (so wheels stay capped); `constraint-dependencies` then becomes a dev convenience rather than the cap source. Cost: cap duplication across packages (a `ruff`/pre-commit or CI check can keep them in sync) and relaxed sibling pins.
- **B** declares `dynamic = ["dependencies", "optional-dependencies"]` and lets the (already thin) `setup.py` supply `install_requires` via the existing cap+pin logic; `pyproject` keeps the static metadata + namespace config. **Zero publish-behavior change**, still gets the uv workspace and the single dynamic version. Cost: deps don't move into `pyproject` (it's "PEP 621 metadata shell + computed deps").
- **C** bakes caps + exact pins into static deps via a custom build-backend/setuptools hook — most machinery; only worth it if we want both pure-PEP-621 deps *and* exact pins.

### Recommendation
- Want **exact pinning + capped wheels with zero risk**, and OK with deps staying in a (much smaller) `setup.py`? → **Option B.**
- Willing to **relax inter-package pins to `==1.5.* --pre`** for the clean full-PEP-621 end state? → **Option A** (inline caps; `constraint-dependencies` optional; add a sync check).

Independent of the choice: **commit `uv.lock`** (currently gitignored), and keep the `packages/` move as its own follow-up PR.

I can implement whichever you pick on this branch — A is a small adjustment to the current state; B is a re-shape of the per-package `setup.py`/`pyproject` split.

PR #5688, opened 2026-06-05
PEP 621 + uv workspace for the full monorepo

Innixma:
## Summary

Converts the **entire monorepo** to PEP 621 `pyproject.toml` + a **uv workspace**, realizing the approach prototyped in #5684 (closed RFC). Dependencies stay *dynamic* and are supplied by a thin per-package `setup.py`, so published wheels keep their exact `Requires-Dist`: **zero publish-behavior change**.

All 7 distributions (`common`, `core`, `features`, `tabular`, `timeseries`, `multimodal`, and the meta `autogluon`) now:
- carry **static metadata** (name, description, SPDX `license`, `requires-python`, `authors`, `[project.urls]`) + setuptools namespace config in `pyproject.toml`, and
- declare `version` / `dependencies` / `optional-dependencies` / `classifiers` / `readme` as **dynamic**, supplied at build time by a thin `setup.py`.

### Why keep dependencies dynamic (what it preserves)
`setup.py` computes deps via `core/_setup_utils.py`:
- shared third-party **caps** from `DEPENDENT_PACKAGES` (single source of truth), and
- exact `==<version>` **sibling pins** (e.g. `autogluon.tabular==X` → `autogluon.core==X`).

So `pip install` from PyPI yields the same capped, pinned wheels as today. `classifiers` (incl. the `RELEASE`→Development Status toggle) and the root `README` long-description are also supplied dynamically — robust for dev/`uv` builds and the publish workflow that copies `README`/`LICENSE`/`NOTICE` into each package dir at build time.

### Highlights
- **Root**: `[tool.uv.workspace]` (7 members) + `[tool.uv.sources]` (siblings → workspace) for editable dev. The prototype's `constraint-dependencies` was removed (caps are single-sourced in `_setup_utils` again).
- **`uv.lock` committed** (301 packages) for reproducible dev/CI; un-ignored in `.gitignore`.
- **`_setup_utils.py`**: replaced the now-unused `default_setup_args` with focused `load_readme()` / `get_classifiers()`; dropped dead `PYTHON_REQUIRES` and the `License :: OSI Approved` classifier (SPDX `license` replaces it; setuptools ≥77 disallows mixing). Source/editable/`uv` installs now report a `.dev0` version suffix (e.g. `1.5.1.dev0`) so a from-source install is distinguishable from a published release; nightlies (`bDATE`) and stable releases (no suffix) are unchanged.

## Verification
- `uv lock` resolves the full workspace — **301 packages**.
- Built `autogluon.common` wheel `METADATA` carries the capped `Requires-Dist`, the exact extras, `Author`, `Project-URL`s, classifiers, and the README long-description.
- `ruff check` / `ruff format --check` clean.

## Local install
`git clone` + `uv sync --all-extras` installs all 7 members editable from the lockfile. Documentation for this (and updates to the source-install docs/scripts) is in the stacked docs PR #5689, which targets this branch and should be merged into it before this PR merges to `master`.

## Open follow-ups (separate PRs)
- Drop the thin `setup.py` via a build-backend version plugin (pure PEP 621).
- The cosmetic `packages/` directory move.

---


autogluon/setup.py
```diff
@@ -1,42 +1,46 @@
 #!/usr/bin/env python
+# Option B thin setup.py for the meta package (submodule=None): see common/setup.py. Supplies the
```
shchur: Nit: Let's remove Option B phrasing from the code & comments (around 9 occurrences in the diff)

autogluon/setup.py
```diff
@@ -1,42 +1,46 @@
...
 spec = importlib.util.spec_from_file_location("ag_min_dependencies", filepath_import)
 ag = importlib.util.module_from_spec(spec)
-# Identical to `from autogluon.core import _setup_utils as ag`, but works without `autogluon.core` being installed.
 spec.loader.exec_module(ag)
 ###########################
 
-version = ag.load_version_file()
-version = ag.update_version(version)
+submodule = None  # meta package
+version = ag.update_version(ag.load_version_file())
```
shchur, line 21: When I run the following command on the PR branch
```bash
uv pip install -e autogluon/
```
in a fresh venv, I get package versions
```
+ autogluon==1.5.1 (from file:///local/home/shchuro/workspace/autogluon/autogluon)
+ autogluon-common==1.5.1 (from file:///local/home/shchuro/workspace/autogluon/common)
+ autogluon-core==1.5.1 (from file:///local/home/shchuro/workspace/autogluon/core)
+ autogluon-features==1.5.1 (from file:///local/home/shchuro/workspace/autogluon/features)
+ autogluon-multimodal==1.5.1 (from file:///local/home/shchuro/workspace/autogluon/multimodal)
+ autogluon-tabular==1.5.1 (from file:///local/home/shchuro/workspace/autogluon/tabular)
+ autogluon-timeseries==1.5.1 (from file:///local/home/shchuro/workspace/autogluon/timeseries)
```
without the usual suffix. Is this intended?

shchur (approved):
Thanks! Just two minor comments

autogluon/setup.py
Innixma: Yes, we now use the date only for the pre-release

Innixma: This avoids a lot of nasty edge cases like the day changing half-way through multi-package installs which would break people installing remotely one at a time the submodules from source.

Innixma: If we want we could add a suffix to it still, maybe a consistent one that doesn't change by date though? Open to suggestions

shchur: Makes sense. Is it possible to change the version to something like `1.6.0b0` or `1.6.0.dev0` after the `1.5.0` release? So install from source will use the `b0` suffix and the pre-release will keep the usual `b20260608` suffix?

autogluon/setup.py
Innixma: updated

autogluon/setup.py
Innixma: Implemented in 7ec8e0a — source installs now carry a static `.dev0` suffix, computed by context in `update_version()`:

| Context | Trigger | Version |
|---|---|---|
| source / editable / `uv` | (no env vars) | `1.5.1.dev0` |
| nightly pre-release | `AUTOGLUON_VERSION_SUFFIX=bDATE` | `1.5.1b20260608` (unchanged) |
| stable release | `RELEASE=1` | `1.5.1` (unchanged) |

PEP 440 orders these `dev < beta < final` (`1.5.1.dev0` < `1.5.1b20260608` < `1.5.1`), so a from-source install is always distinguishable from a published release, and published wheels are unaffected.

One note on the exact approach: I compute the suffix rather than baking it into the `VERSION` file. If `VERSION` literally held `1.6.0.dev0`, the nightly (which appends `bDATE` to the base) would produce `1.6.0.dev0b20260608`, which is invalid PEP 440 — a `.dev` segment must come last. Computing it keeps `VERSION` the plain release target (so the existing "bump `VERSION` after release" step is unchanged) while giving source/nightly/release their distinct suffixes.

I went with `.dev0` over `b0` since it's the conventional marker for an unreleased from-source build and sorts *before* the nightly betas (a working checkout is "earlier" than a published nightly). Happy to switch to `b0` if you'd prefer — it's a one-token change.

shchur: Sounds good, thanks @Innixma!

PR #5700, opened 2026-06-18
Fix IndexError in GroupByFeatureGenerator on all-NaN categorical columns

Innixma:
## Description

`GroupByFeatureGenerator` raises `IndexError: cannot do a non-empty take from an empty axes` when a grouping categorical column has no non-null values.

## Root cause

In `_fit`, a column is selected as a grouping categorical when `nunique() < num_as_cat_cardinality_thresh`. Because `nunique()` ignores NaN, an **all-NaN** column has `0` unique values and is selected. But `groupby(observed=True)` then produces **zero groups**, so `group_values_[(cat, num)]` holds empty arrays.

In `_transform`, the guard
```python
if idx is not None and vals_dict is not None and len(vals_dict) > 0:
    for agg, vals in vals_dict.items():
        col = vals.take(safe_codes).astype(float, copy=False)
```
checks the number of *aggregations* (`len(vals_dict)`), not the number of *groups*. With an empty group table it still enters the branch and runs `vals.take(safe_codes)` on an empty `vals` while `safe_codes` has one entry per row → `IndexError`.

This shows up in practice on datasets with (near-)all-missing categorical columns — e.g. such a column landing entirely NaN in the 1000-row subsample AutoGluon uses for model memory estimation (`_estimate_dtypes_after_preprocessing_cheap`).

## Changes

- **`_fit`**: drop grouping categoricals with no non-null values. They carry no grouping signal; constant columns (one observed value → one group) are kept and handled normally.
- **`_transform`**: add `len(idx) > 0` to the group-agg guard so an empty group table falls through to the existing NaN/fill branch instead of raising (defensive).
- **Tests**: new `features/tests/features/generators/test_groupby.py` covering all-NaN categorical (no crash + column dropped), the sole-categorical-all-NaN case (empty output), the normal happy path, and the `_transform` empty-group-table guard.

## Test plan

- `pytest features/tests/features/generators/test_groupby.py` → 4 passed
- `pytest features/tests/features/generators/` → 57 passed (no regressions)
- `ruff check` clean on changed files

### DIFF
--- a/features/src/autogluon/features/generators/groupby.py
+++ b/features/src/autogluon/features/generators/groupby.py
@@ -187,6 +187,14 @@ class GroupByFeatureGenerator(AbstractFeatureGenerator):
         self.categorical_features += X.select_dtypes(include=[R_CATEGORY, R_OBJECT]).columns.tolist()
         self.categorical_features = np.unique(self.categorical_features).tolist()
 
+        # Drop all-NaN grouping categoricals: `groupby(observed=True)` yields zero groups for a
+        # column with no non-null values (and such a column is even *selected* above, since
+        # `nunique()` ignores NaN -> 0 < num_as_cat_cardinality_thresh). Keeping it would build an
+        # empty group table that `_transform` then indexes into, raising "cannot do a non-empty
+        # take from an empty axes". An all-NaN column carries no grouping signal anyway; constant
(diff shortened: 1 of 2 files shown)


LennartPurucker (approved):
LGTM

PR #5701, opened 2026-06-18
Silence spurious warnings on degenerate (constant / all-NaN) columns during fit

Innixma:
## Description

Fitting feature generators / the learner on data with constant or all-NaN columns emitted three benign-but-noisy warnings. These show up in practice on datasets with (near-)all-missing categorical columns — e.g. when such a column lands entirely NaN in the 1000-row subsample AutoGluon uses for model memory estimation. All three fixes are behavior-preserving (the produced values are unchanged); they only remove the warnings.

## Warnings fixed

1. **`OOFTargetEncodingFeatureGenerator`** — `RuntimeWarning: Mean of empty slice`
   An all-NaN categorical column has zero categories after `pd.factorize` (`n_cat == 0`), so the per-category mean matrix has shape `(0, n_targets)` and `np.nanmean(mean_all, axis=0)` averages an empty slice. The global mean is NaN for such a column regardless; compute it directly when `n_cat == 0`.

2. **`SpearmanFeatureSelector`** — `scipy ConstantInputWarning: An input array is constant; the correlation coefficient is not defined`
   Spearman correlation is undefined for a constant column (scipy returns NaN). Drop columns with `<= 1` distinct non-null value before `corrwith(method="spearman")` — the subsequent `.dropna()` already discarded them, so selection is unchanged.

3. **`DefaultLearner._check_for_non_finite_values`** — `FutureWarning: Downcasting behavior in `replace` is deprecated`
   Append `.infer_objects(copy=False)` to `replace([np.inf, -np.inf], np.nan)` to retain the pre-pandas-2.2 dtype behavior explicitly (the migration-recommended idiom).

## Test plan

- New `features/tests/features/generators/test_selection.py` (constant / all-NaN columns excluded without warning; happy path).
- New cases in `features/tests/features/generators/test_oof_target_encoder.py` (all-NaN categorical → NaN global mean, no empty-slice warning; normal categorical finite, no warning).
- `pytest features/tests/features/generators/` → 57 passed (no regressions).
- Verified each warning is silenced under `warnings.simplefilter("error", ...)` / capture.

Related: same all-NaN-categorical root cause as #5700 (GroupByFeatureGenerator crash); this PR is warning hygiene only.

### DIFF
--- a/features/src/autogluon/features/generators/oof_target_encoder.py
+++ b/features/src/autogluon/features/generators/oof_target_encoder.py
@@ -179,7 +179,13 @@ class OOFTargetEncodingFeatureGenerator(AbstractFeatureGenerator):
 
             # global_mean as in original _transform:
             # mean of per-category means
-            global_mean = np.nanmean(mean_all, axis=0)  # (n_targets,)
+            if n_cat == 0:
+                # All-NaN categorical column -> zero categories, so `mean_all` has shape
+                # (0, n_targets) and `np.nanmean(..., axis=0)` averages an empty slice (a
+                # RuntimeWarning, returning NaN). Produce the same all-NaN global mean directly.
+                global_mean = np.full(self.n_targets, np.nan)
+            else:
+                global_mean = np.nanmean(mean_all, axis=0)  # (n_targets,)
 
             # Smoothed per-category encodings used at inference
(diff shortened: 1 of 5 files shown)


LennartPurucker (approved):
LGTM

PR #5702, opened 2026-06-19
[experimental] Prototype GPU support for fit_strategy='parallel' (behind AG_PARALLEL_GPU)

Innixma:
## Summary

**Experimental / draft.** Prototype GPU support for `TabularPredictor.fit(..., fit_strategy="parallel")`, which today only schedules CPU model-workers. The entire change is gated behind the `AG_PARALLEL_GPU=True` environment variable and is a **no-op when unset** — existing parallel-fit (CPU) behavior is unchanged.

With the flag on, GPU models that declare a per-model `num_gpus` (via `ag_args_fit`) are scheduled across the available GPUs and fit concurrently, instead of `fit_strategy="parallel"` falling back to sequential whenever `num_gpus > 0`.

## Motivation

`fit_strategy="parallel"` (Ray model-workers via `ParallelFitManager`) parallelizes across bagged model fits, but GPU resources were hardcoded to 0 for the parallel path, so any GPU run silently degraded to sequential. On multi-GPU machines this leaves most of the hardware idle when fitting several GPU models (foundation models like TabPFN/TabICL, torch models like TabM/RealMLP).

## What's in this PR (all behind `AG_PARALLEL_GPU`)

- **Flag + plumbing** — `gpu_parallel_fit_enabled()`; the trainer's `num_gpus > 0 → sequential` fallback is gated by the flag; `ParallelFitManager` reads each model's declared per-fold `num_gpus` and caps concurrent children by available GPUs.
- **`invalid device ordinal` fix** — the bagged orchestrator / refit fit no longer multiplies its GPU count by `num_parallel` (Ray only makes `num_gpus` visible; models that select devices by absolute index, e.g. TabPFN-3, crashed otherwise).
- **Strategy-aware reservation** — `sequential_local` (folds fit in-process in the single model-worker) reserves the per-fold GPU on the model-worker and **no** nested-fold-worker GPUs, so total reserved GPUs == the per-fold count (not `num_gpus × num_splits`); `parallel_local` keeps the nested-fold-worker reservation.
- **Concurrency-aware CPU/GPU reservation** — `sequential_local`'s concurrent footprint is 1 fold (not `num_splits`), so it reserves one fold's CPUs/GPUs and is no longer deferred waiting to "fit all folds in parallel." `parallel_local`/`auto` are unchanged.
- **Logging** — the scheduler line now reports `Allocated GPUS` alongside `Allocated CPUS` (only when GPUs > 0; fractional-safe formatting).

## Validation

Exercised empirically on a 8× GPU machine across: TabPFN-3, TabICLv2, TabM, RealMLP (GPU) mixed with LightGBM / XGBoost / CatBoost (CPU); bagged and non-bagged; `sequential_local` and `parallel_local` fold fitting; and fractional per-fold GPU (e.g. RealMLP at `0.25` GPU/fold packing several folds per device — note foundation-model wrappers floor `num_gpus` at 1 via `get_minimum_resources`, so they can't be packed fractionally). Unit tests added in `core/tests/unittests/ray/test_parallel_gpu_prototype.py` for the flag gate, per-child GPU sizing, strategy-aware reservation, and the concurrency logic.

## Status / open questions

- Draft — the flag name, default-off posture, and resource-reservation heuristics are up for discussion before this would be merge-ready.
- `parallel_local` GPU models reserve a per-fold GPU for each of `num_splits` folds, so a single such model effectively wants the whole GPU pool — worth deciding whether that should be capped.
- No CI coverage of the end-to-end Ray-on-GPU path (requires real GPUs); only the scheduler/reservation logic is unit-tested.

### DIFF
--- a/core/src/autogluon/core/ray/distributed_jobs_managers.py
+++ b/core/src/autogluon/core/ray/distributed_jobs_managers.py
@@ -180,6 +180,26 @@ class ParallelFitManager:
         else:
             return self.num_splits
 
+    def _num_concurrent_children(self, model: AbstractModel | str, num_children: int) -> int:
+        """Number of folds resident simultaneously, for CPU/GPU/memory reservation & gating.
+
+        For the EXPERIMENTAL GPU-parallel prototype, ``sequential_local`` fits every fold (and the
+        refit) one-at-a-time IN the single model-worker -- there are no nested fold-workers, so at
+        most ONE fold is ever resident. Its concurrent footprint is therefore 1 fold, not
+        ``num_children`` (= num_splits). Reporting 1 stops the scheduler from reserving
+        ``num_splits x`` the per-fold CPUs/GPUs and from deferring the model to "fit all folds in
+        parallel" -- neither applies when folds are sequential. Every other strategy fits
+        ``num_children`` folds in parallel, so the count is unchanged (also a no-op when the flag
+        is off, preserving historical behavior).
+        """
+        if (
+            gpu_parallel_fit_enabled()
+            and isinstance(model, AbstractModel)
+            and model._user_params.get("fold_fitting_strategy") == "sequential_local"
+        ):
+            return 1
+        return num_children
+
     def _num_gpus_per_child(self, model: AbstractModel | str) -> float:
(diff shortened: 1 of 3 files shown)


LennartPurucker (approved):
LGTM

PR #5715, opened 2026-07-12
Add AGENTS.md + .agents/ machine-readable repo intelligence for AI coding agents

topprismdata:
## What & Why

A concise, AST-derived map of the repository for AI coding agents (Claude Code, Copilot, Cursor, Gemini CLI, etc.), plus machine-readable JSON artifacts under `.agents/` that agents can parse directly.

AutoGluon is a 7-member `uv` workspace monorepo, but that structure — and several non-obvious traps — are invisible from the root layout. These three facts are not documented elsewhere and are exactly what an agent needs to avoid breaking the build:

1. **It is NOT a single package.** Six independently-publishable PEP 621 members (`common`/`core`/`features`/`multimodal`/`tabular`/`timeseries`) + a meta-package. An agent that doesn't know this will edit code in the wrong place (e.g. the top-level `autogluon/` dir, which is packaging-only).
2. **Versions and dependency caps are DYNAMIC.** Each member's `pyproject.toml` declares them as `dynamic`; they are injected at build time by `setup.py` from `core/_setup_utils.py`. Editing `pyproject.toml` version/deps directly is a no-op.
3. **The inter-package dependency graph has a back-edge.** `common` (a base layer) imports `tabular`/`multimodal` (upper layers) — measured by AST, surfaced in `dependency_graph.json`. Worth knowing before any refactoring.

## Contents

| file | what |
|------|-----|
| `AGENTS.md` | concise root index (progressive disclosure) |
| `.agents/analyze.py` | AST-based generator — re-run after structural changes |
| `.agents/dependency_graph.json` | inter-package `import autogluon.X` edges |
| `.agents/packages.json` | per-package stats (files/classes/functions/tests) |
| `.agents/api_surface.json` | public class/function inventory with `file:line` |
| `.agents/entry_points.json` | Predictor/Space/data classes with `file:line` |
| `.agents/README.md` | provenance + regeneration instructions |

No existing files are modified. Purely additive.

## How to keep it current

```bash
python3 .agents/analyze.py   # regenerates the 4 JSON files deterministically
```

Run it after structural changes and commit the result alongside the change. The `AGENTS.md` prose is curated; every number and edge in `.agents/*.json` is machine-verified.

## Methodology

Informed by [RepoAgent (OpenBMB, EMNLP 2024)](https://github.com/openbmb/repoagent) — repository-level documentation driven by AST dependency analysis — and [Anthropic's context-engineering guidance](https://www.anthropic.com/engineering/effective-context-engineering-for-ai-agents) on progressive disclosure. The distinction vs. hosted services (e.g. Google Code Wiki) is that these assets live in-repo, so they're available to any agent offline, in CI, and in private forks.

Happy to adjust scope, naming, or placement per maintainer feedback. If a nested `AGENTS.md` per package would be welcome, I can follow up in a separate PR.


shchur:
Thanks for the PR. This isn't something we're planning to add right now, so I'm going to close it. We may revisit it in the future if it becomes a better fit for the project.

PR #5720, opened 2026-07-13
EnsembleSelection hygiene: no caller mutation, predict problem_type override

Innixma:
## Description

Two small fixes for external drivers of `EnsembleSelection` (e.g. TabArena's ensemble simulation, which fits it thousands of times over cached predictions):

1. **`_fit` no longer mutates the caller's `predictions` list.** When `subsample_size` triggers, the subsampled predictions were assigned back into the passed-in list (`predictions[i] = predictions[i][subsample_indices]`), silently corrupting any caller that reuses the list — a footgun next to prediction caches. The subsampled list is now rebuilt locally.

2. **`AbstractWeightedEnsemble.predict` takes an optional `problem_type` override** for the proba -> label conversion. Callers that fit on a transformed prediction space (e.g. TabArena feeds ground-truth-class probabilities with `problem_type="binary"` for log_loss ensembling) previously had to reassign `ensemble.problem_type` around the call; now they can pass the original problem type explicitly. Default behavior is unchanged.

Both changes are backwards compatible for the internal consumers (`GreedyWeightedEnsembleModel`, tabular `abstract_learner`, multimodal ensemble learner, timeseries `TimeSeriesEnsembleSelection`): existing call signatures and behavior are preserved.

## Verification

Verified directly that (a) with `subsample_size` set, the caller's list and arrays are unchanged after `fit` while producing the same weights as before, and (b) `predict(X)` and `predict(X, problem_type=<same>)` return identical outputs.

### DIFF
--- a/core/src/autogluon/core/models/greedy_ensemble/ensemble_selection.py
+++ b/core/src/autogluon/core/models/greedy_ensemble/ensemble_selection.py
@@ -16,9 +16,18 @@ logger = logging.getLogger(__name__)
 
 
 class AbstractWeightedEnsemble:
-    def predict(self, X):
+    def predict(self, X, problem_type: str | None = None):
+        """Predict labels from the weighted prediction probabilities.
+
+        problem_type overrides self.problem_type for the proba -> label conversion,
+        for callers that fit on a transformed prediction space (e.g. TabArena's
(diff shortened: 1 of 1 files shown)


LennartPurucker (approved):
LGTM

PR #5721, opened 2026-07-13
[Security Fix] Add warning docstrings on pickle.loads() usage

prateekdesai04:
*Description of changes:*

- Disable loading pickle from URL unless an environment variable is set
- Add warning to the internal helper method `load_pkl.load`
- Emit warning when loading pickle from S3

### DIFF
--- a/common/src/autogluon/common/loaders/load_pkl.py
+++ b/common/src/autogluon/common/loaders/load_pkl.py
@@ -11,21 +11,15 @@ from ..utils import compression_utils, s3_utils
 
 logger = logging.getLogger(__name__)
 
-# Loading a pickle file executes arbitrary code during deserialization (see CWE-502).
-# Fetching a pickle from a remote HTTP(S) URL is therefore unsafe: anyone able to supply
-# or intercept the URL achieves remote code execution. Loading from a URL is blocked by
-# default and must be explicitly opted into by users who fully trust the source.
-_ALLOW_REMOTE_PICKLE_ENV_VAR = "AG_ALLOW_REMOTE_PICKLE"
+# Block loading pickle from remote URLs by default; opt in via this env var.
+_ALLOW_PICKLE_FROM_URL_ENV_VAR = "AG_ALLOW_PICKLE_FROM_URL"
 
 
-def _is_remote_pickle_allowed() -> bool:
-    return os.environ.get(_ALLOW_REMOTE_PICKLE_ENV_VAR, "False").lower() in ("true", "1")
+def _url_pickle_allowed() -> bool:
+    return os.environ.get(_ALLOW_PICKLE_FROM_URL_ENV_VAR, "False").lower() in ("true", "1")
 
 
-# S3 loading is a core, widely-used feature (e.g. loading a predictor from an ``s3://`` path), so it
-# cannot be blocked by default without breaking legitimate workflows. It remains unsafe, however:
-# ``pickle.loads`` executes arbitrary code during deserialization, so a compromised or misconfigured
-# bucket results in code execution on load (CWE-502). We surface this risk with a one-time warning.
+# Loading from S3 is widely used, so we warn instead of blocking.
 _warned_s3_pickle = False
 
 
@@ -33,14 +27,13 @@ def _warn_s3_pickle_load(path: str) -> None:
     global _warned_s3_pickle
     if not _warned_s3_pickle:
         logger.warning(
-            f"Loading a pickle file from S3 ({path}). Deserializing a pickle file executes arbitrary "
-            "code, so a compromised or misconfigured bucket can lead to code execution. Only load "
-            "objects from S3 locations you control and trust. This warning is shown once per process."
+            f"Loading pickle from S3 ({path}). Unpickling executes arbitrary code; only load from "
+            "buckets you trust."
         )
         _warned_s3_pickle = True
 
 
-def load(path: str, format: str | None = None, verbose: bool = True, **kwargs) -> Any:
+def load(path: str, format: str | None = None, verbose: bool = True, trust_remote: bool = False, **kwargs) -> Any:
     """
 
     Parameters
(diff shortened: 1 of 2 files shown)


common/src/autogluon/common/loaders/load_pkl.py
```diff
@@ -10,6 +11,34 @@
...
+_warned_s3_pickle = False
+
+
+def _warn_s3_pickle_load(path: str) -> None:
+    global _warned_s3_pickle
+    if not _warned_s3_pickle:
+        logger.warning(
+            f"Loading a pickle file from S3 ({path}). Deserializing a pickle file executes arbitrary "
+            "code, so a compromised or misconfigured bucket can lead to code execution. Only load "
+            "objects from S3 locations you control and trust. This warning is shown once per process."
```
shchur: Let's make this more concise? The user is more likely to ignore long messages. For example:
```python
f"Loading pickle from S3 ({path}). Unpickling executes arbitrary code; only load from "
"buckets you trust."
```

common/src/autogluon/common/loaders/load_pkl.py
```diff
@@ -10,6 +11,34 @@
...
+
+
+def _is_remote_pickle_allowed() -> bool:
+    return os.environ.get(_ALLOW_REMOTE_PICKLE_ENV_VAR, "False").lower() in ("true", "1")
+
+
+# S3 loading is a core, widely-used feature (e.g. loading a predictor from an ``s3://`` path), so it
+# cannot be blocked by default without breaking legitimate workflows. It remains unsafe, however:
+# ``pickle.loads`` executes arbitrary code during deserialization, so a compromised or misconfigured
+# bucket results in code execution on load (CWE-502). We surface this risk with a one-time warning.
```
shchur: The comments here and above are really verbose, could we shorten them? 

```python
# Block loading pickle from remote URLs by default; opt in via this env var.
...
# Loading from S3 is widely used, so we warn instead of blocking
```

common/src/autogluon/common/loaders/load_pkl.py
```diff
@@ -10,6 +11,34 @@
 
 logger = logging.getLogger(__name__)
 
+# Loading a pickle file executes arbitrary code during deserialization (see CWE-502).
+# Fetching a pickle from a remote HTTP(S) URL is therefore unsafe: anyone able to supply
+# or intercept the URL achieves remote code execution. Loading from a URL is blocked by
+# default and must be explicitly opted into by users who fully trust the source.
+_ALLOW_REMOTE_PICKLE_ENV_VAR = "AG_ALLOW_REMOTE_PICKLE"
```
shchur: This reads like it also blocks S3 access, but actually only blocks URLs. Should we rename it to something like `AG_ALLOW_PICKLE_FROM_URL` to be more precise?

common/src/autogluon/common/loaders/load_pkl.py
```diff
@@ -84,8 +114,23 @@ def _is_web_url(path: str) -> bool:
...
+            f"you fully trust the source and connection, set the environment variable "
+            f"{_ALLOW_REMOTE_PICKLE_ENV_VAR}=True to allow loading pickle files from URLs."
+        )
+
     import requests
 
+    logger.warning(
+        f"Loading a pickle file from a remote URL ({url}). This executes arbitrary code during "
+        "deserialization; only do this for sources you fully trust."
+    )
```
shchur: The user already explicitly opted in to allow loading from URLs, do we really need to warn here?

common/src/autogluon/common/loaders/load_pkl.py
```diff
@@ -84,8 +114,23 @@ def _is_web_url(path: str) -> bool:
...
 def _load_pickle_from_url(url: str):
+    if not _is_remote_pickle_allowed():
+        raise ValueError(
+            f"Refusing to load a pickle file from a remote URL: {url}\n"
+            "Loading a pickle file executes arbitrary code during deserialization, so fetching one "
+            "from an HTTP(S) URL allows anyone who can supply or intercept that URL to run arbitrary "
+            "code on this machine (remote code execution).\n"
+            "Download the file to local storage and verify its integrity before loading it, or, only if "
+            f"you fully trust the source and connection, set the environment variable "
+            f"{_ALLOW_REMOTE_PICKLE_ENV_VAR}=True to allow loading pickle files from URLs."
```
shchur: More concise
```suggestion
            f"Refusing to load a pickle file from a remote URL: {url}\n"
            "Unpickling executes arbitrary code, so loading from a URL lets anyone who controls it run "
            "code on this machine.\n"
            "Download the file locally and verify it before loading, or set "
            f"{_ALLOW_REMOTE_PICKLE_ENV_VAR}=True if you trust the source."
```

common/src/autogluon/common/loaders/load_pkl.py
```diff
@@ -10,6 +11,34 @@
...
+    if not _warned_s3_pickle:
+        logger.warning(
+            f"Loading a pickle file from S3 ({path}). Deserializing a pickle file executes arbitrary "
+            "code, so a compromised or misconfigured bucket can lead to code execution. Only load "
+            "objects from S3 locations you control and trust. This warning is shown once per process."
+        )
+        _warned_s3_pickle = True
+
 
 def load(path: str, format: str | None = None, verbose: bool = True, **kwargs) -> Any:
```
shchur: Should we also add a bool flag to this method (and this method only) to allow the users to load from URLs without touching the env variable? E.g. if someone wants to download a file from a safe URL, they can call
```python
data = load_pkl.load("...", trust_remote=True)
```
instead of 
```python
import os
os.environ["AG_ALLOW_REMOTE_PICKLE"] = "True"
data = load_pkl.load("...")
```
I wouldn't propagate this to `*Predictor.load`, but for one-off loading with `load_pkl.load` this seems convenient.

common/src/autogluon/common/loaders/load_pkl.py
shchur, line 55: This needs to be updated since URL loading is not supported out of the box anymore.

common/src/autogluon/common/loaders/load_pkl.py
shchur: If we add this flag, will also need to update the helper method as
```python
  def _load_pickle_from_url(url, trust_remote):
      if not (trust_remote or _url_pickle_allowed()):
          raise ValueError(...)
```

common/src/autogluon/common/loaders/load_pkl.py
shchur, line 70: Let's add some unit tests that ensure that URL loading is blocked unless the env variable is set?

jalencato:
I do think we add too much for the doc string. We do not need that much output log here.

shchur (approved):

PR #5722, opened 2026-07-13
Validate zip entries before extraction

Atharva-Rajan-Kale:
*Issue #, if available:*
- Adds path traversal and symlink validation before zip extraction at all extractall call sites

### DIFF
--- a/common/src/autogluon/common/loaders/_utils.py
+++ b/common/src/autogluon/common/loaders/_utils.py
@@ -2,6 +2,7 @@ import functools
 import hashlib
 import logging
 import os
+import stat
 import sys
 import uuid
 import warnings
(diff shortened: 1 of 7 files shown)


core/src/autogluon/core/utils/files.py
```diff
@@ -7,6 +7,8 @@
 import zipfile
 from pathlib import Path
 
+from autogluon.common.loaders._utils import safe_extractall
```
shchur: It's a bit awkward to import from a private submodule `_utils`. If the method `safe_extractall` is now used across the codebase, let's move it to an appropriate public submodule.

common/src/autogluon/common/loaders/_utils.py
```diff
@@ -251,6 +252,18 @@ def path_expander(path, base_folder):
     return ";".join([os.path.join(base_folder, path) for path in path_l])
 
 
+def safe_extractall(zf, dest_dir):
```
shchur: Let's add some unit tests for this method testing normal behavior + all the cases that should result in an error.

shchur (approved):

PR #5716, opened 2026-07-12
Generate agent-friendly llms.txt / llms-full.txt from the Sphinx build

topprismdata:
## What

Adds the [`sphinx_llm.txt`](https://github.com/NVIDIA/sphinx-llm) extension to the docs build, so a normal `sphinx-build -b html` also emits a **markdown mirror** of the documentation alongside the HTML:

- `llms.txt` — markdown index of all doc pages (the [llms.txt standard](https://llmstxt.org/))
- `llms-full.txt` — all pages concatenated, for long-context models
- `<page>.html.md` — per-page markdown render

Once deployed, these become available at `auto.gluon.ai/llms.txt` and `auto.gluon.ai/llms-full.txt`, giving AI coding agents (Claude Code, Copilot, Cursor, Gemini CLI) a token-efficient, navigation-noise-free view of the docs instead of forcing them to scrape HTML.

## Why sphinx-llm (not sphinx-llms-txt)

AutoGluon's API reference under `docs/api/*.rst` is generated **dynamically** via `autosummary` / `automodule` / `autoclass` (see e.g. `api/tabular.rst`). The simpler `sphinx-llms-txt` extension reads the raw `.rst` source files and therefore leaves those directives **unexpanded** — an agent would see literal `.. autoclass:: TabularPredictor` placeholders instead of the rendered API.

`sphinx-llm` runs the full Sphinx pipeline (autodoc + napoleon + myst-nb) and renders the *resolved* content to markdown, so docstrings, parameter tables, and type info are fully expanded in the agent-facing output.

## Changes

| file | change |
|------|--------|
| `docs/requirements_doc.txt` | add `sphinx-llm` |
| `docs/conf.py` | register `sphinx_llm.txt` extension + `llms_txt_description` |
| `docs/llms-txt.md` | new — maintainer/agent-facing doc explaining the output |
| `docs/index.md` | add `llms-txt.md` to the Resources toctree |

4 files, +90 lines, no existing behavior changed. The markdown output is purely additive — the HTML site is unaffected.

## Relationship to #5715

This complements #5715 (which adds code-level agent assets: `AGENTS.md` + `.agents/`). Together they cover both sides:
- **#5715** → the *repository structure* (packages, dependency graph, AST chunks)
- **this PR** → the *human-authored documentation* (tutorials, install guides, API refs)

Happy to merge them if preferred, but keeping them separate keeps each review focused.

## Verification

[...]

PR #5723, opened 2026-07-14
fix(docs): disable sphinx-llm in per-tutorial CI builds (#5716 followup)

topprismdata:
## What
Fixes the failing docs CI jobs in PR #5716 by disabling the sphinx-llm
markdown build in the per-tutorial CI jobs. The `build_all_docs` job,
which produces the deployed llms.txt artifacts, is unaffected — it gets
the markdown build back via a new `AUTOGLUON_BUILD_ALL_DOCS=1` env
var that `build_all_docs.sh` sets.

## Why PR #5716 fails
sphinx-llm launches a markdown `sphinx-build` subprocess in parallel
with the main HTML build. The per-tutorial CI jobs (tabular / cloud /
timeseries) all run with notebook execution enabled (no `-D
nb_execution_mode=off`), so the parent process is busy executing
tutorials/*.ipynb while the markdown subprocess is also building the
docs. On the constrained CI runners the two processes race for CPU
and memory and the markdown subprocess times out, leaving the parent
waiting forever:

```
Waiting for markdown build subprocess to finish...
================================================================================
Job [AutoGluon-BuildTimeseries-PR5716] FAILED
##[error]Process completed with exit code 1.
```

This is purely a CI scheduling problem; the `build_all_docs` job
which is the one that actually deploys llms.txt, succeeds because it
sets `sphinx-build -D nb_execution_mode=off` so the parent has no
work to do while the markdown subprocess runs.

## Fix
- `docs/conf.py`: detect the env var and set
  `llms_txt_enabled = False` when it is not present.
- `.github/workflow_scripts/build_all_docs.sh`: set
  `AUTOGLUON_BUILD_ALL_DOCS=1` so the all-docs build keeps emitting
  llms.txt / llms-full.txt.
- `docs/llms-txt.md`: document when llms.txt is generated and how
  to opt back in for one-off builds.

## Verification
- Local sphinx-build with the env var unset: no 'Waiting for markdown
  build subprocess to finish' line, no `llms.txt` / `llms-full.txt`
  emitted (matches the per-tutorial CI behaviour).
- Local sphinx-build with the env var set: markdown subprocess runs,
  `llms.txt` and `llms-full.txt` are emitted (matches the
  build_all_docs behaviour).
- Minimal sphinx-llm test project (single .rst + .py): builds llms.txt
  end-to-end, confirming the extension itself is healthy.

## Relationship to #5716
This is a followup to #5716. The first PR introduced the sphinx-llm
extension; this PR fixes the CI integration so it can actually merge.
Once this lands, the auto.gluon.ai/llms.txt artifacts will start being
deployed.

### DIFF
--- a/.github/workflow_scripts/build_all_docs.sh
+++ b/.github/workflow_scripts/build_all_docs.sh
@@ -54,6 +54,11 @@ else
 fi
 
 setup_build_contrib_env
+# Enable sphinx-llm (emits llms.txt + llms-full.txt). Disabled by default in
+# conf.py to avoid the parallel markdown build racing tutorial CI jobs that
+# run jupyter notebook execution. This is the only job that produces the
+# deployed llms.txt artifacts (auto.gluon.ai/llms.txt).
+export AUTOGLUON_BUILD_ALL_DOCS=1
 install_all_no_tests
 
 LOCAL_DOC_PATH=_build/html
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -40,6 +40,20 @@ llms_txt_description = (
     "text and multimodal data."
 )
 
+# sphinx-llm runs a second sphinx-build subprocess (markdown builder) in
+# parallel with the main HTML build. On CI runners with constrained memory /
+# CPU this can race the parent process and either timeout or OOM during the
+# expensive jupyter notebook execution phase (tutorials/*.ipynb).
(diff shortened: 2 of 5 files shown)


shchur:
Thanks for the PR. This isn't something we're planning to add right now, so I'm going to close it. We may revisit it in the future if it becomes a better fit for the project.

PR #5737, opened 2026-07-20
Fix/tar slip detection dataset

Atharva-Rajan-Kale:
*Issue #, if available:*

- Validate tar/zip entries before extraction to prevent path traversal in dataset preparation CLI

### DIFF
--- a/multimodal/src/autogluon/multimodal/cli/prepare_detection_dataset.py
+++ b/multimodal/src/autogluon/multimodal/cli/prepare_detection_dataset.py
@@ -1,8 +1,11 @@
 import argparse
 import os
-import shutil
+import tarfile
+import zipfile
 
 import requests
+
+from autogluon.common.loaders.load_zip import safe_extractall
 
 
 def get_root_dir(output_dir=None, new_folder_name=None):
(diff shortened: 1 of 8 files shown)


multimodal/src/autogluon/multimodal/cli/prepare_detection_dataset.py
```diff
@@ -49,11 +52,30 @@ def download_urls(urls, root_dir, fnames=[]):
...
+        if member.issym() or member.islnk():
+            raise ValueError(f"Tar contains link: {member.name}")
+    tf.extractall(dest_dir)  # nosec B202 - members validated above
+
+
 def unpack(archived_file_paths, root_dir):
     for archived_file_path in archived_file_paths:
         fname = get_fname_from_path_or_url(archived_file_path)
         print(f"extracting {fname}...")
-        shutil.unpack_archive(archived_file_path, root_dir)
```
shchur, line 56: I think a cleaner design would be to create a single helper method `safe_unpack_archive` (mirroring `shutil.unpack_archive`) somewhere in `autogluon.common.loaders` (e.g. new file `load_archive.py`) and use this method across the codebase. Something like the following:

```python
def safe_unpack_archive(path: str | os.PathLike, dest_dir: str | os.PathLike) -> None:
    """Safely extract a zip or tar archive, rejecting path traversal and links.

    The archive format is auto-detected; supports zip and tar (incl. gz/bz2/xz).
    """
```
The method operates directly on paths instead of open file handles, so it's also easier to use.

We can then replace the method `load_zip.safe_extractall` that you introduced in #5722 across the entire codebase to also use this helper. This won't be a breaking change since we haven't made a stable release since `safe_extractall` was introduced.

We can then test the `safe_unpack_archive` method in the tests for `common`.

What do you think?

multimodal/src/autogluon/multimodal/cli/prepare_detection_dataset.py
```diff
@@ -49,11 +52,30 @@ def download_urls(urls, root_dir, fnames=[]):
...
+        if archived_file_path.endswith((".tar", ".tar.gz", ".tgz")):
```
shchur, line 71: If we go with the new design described above, a cleaner/safer option is to use `zipfile.is_zipfile` and `tarfile.is_tarfile` instead of manually checking the extension suffix.

multimodal/src/autogluon/multimodal/cli/prepare_detection_dataset.py
```diff
@@ -49,11 +52,30 @@ def download_urls(urls, root_dir, fnames=[]):
...
 
+def _safe_tar_extractall(tf: tarfile.TarFile, dest_dir: str) -> None:
+    """Extract tar file with path traversal and symlink validation."""
+    dest_dir = os.path.realpath(dest_dir)
+    for member in tf.getmembers():
+        member_path = os.path.realpath(os.path.join(dest_dir, member.name))
+        if not member_path.startswith(dest_dir + os.sep) and member_path != dest_dir:
+            raise ValueError(f"Tar Slip detected: {member.name} would extract outside {dest_dir}")
+        if member.issym() or member.islnk():
+            raise ValueError(f"Tar contains link: {member.name}")
```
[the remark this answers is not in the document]
shchur: @Atharva-Rajan-Kale this looks like a reasonable suggestion

multimodal/src/autogluon/multimodal/cli/prepare_detection_dataset.py
Atharva-Rajan-Kale: Makes sense, I'll make the appropriate changes

PR #5738, opened 2026-07-21
Re-commit uv.lock for Dependabot; enforce freshness via CI + pre-commit

shchur:
Re-commits `uv.lock` (reverting #5736): Dependabot can't read our version ranges from the dynamic `setup.py` / `_setup_utils.py`, so it needs the resolved lockfile to scan.

- **`pyproject.toml`**: `[tool.uv] exclude-newer` pins resolution to a date so a newly published dep can't make the lock stale on its own. Bump it when refreshing the lock.
- **`.python-version`**: pins Python 3.12 so uv (incl. Dependabot's runner) resolves a compatible interpreter instead of defaulting to one outside `requires-python` (`>=3.10, <3.14`), which was making `uv lock` fail.
- **`.github/workflows/lockfile_check.yml`**: runs `uv lock --check` on PRs/push. Standalone `pull_request` workflow (read-only, no secrets) rather than a job in `continuous_integration.yml`, which runs on `pull_request_target` and would check out the base branch (no-op) or run untrusted `setup.py` with a privileged token.
- **`.pre-commit-config.yaml`**: `uv-lock` hook keeps the lock in sync locally.
- **docs / CONTRIBUTING**: committed-lock workflow, `uv sync --frozen`, and the `exclude-newer` bump.


Atharva-Rajan-Kale (approved):
Looks good to me!

PR #5751, opened 2026-07-23
Fix/aiohttp 3.14.2

HuawenShen:
Upgrade transitive dependency aiohttp 3.14.0 -> 3.14.2 to resolve 8 open
Dependabot alerts (CVE-2026-54275/54278/54280 ...，patched in 3.14.1).

- Regenerated uv.lock via `uv lock --upgrade-package aiohttp`
- uv lock --check passes
- Only uv.lock changed


shchur:
Superseded by #5759

PR #5754, opened 2026-07-25
Add TabPFN-3 and TabDPT-Turbo models

Innixma:
# Add TabPFN-3 and TabDPT-Turbo models

## Summary

Adds two tabular foundation models, plus supporting infrastructure fixes. No preset or
zeroshot-portfolio changes — both models are opt-in via explicit `hyperparameters`.

### TabPFN-3 (`TABPFN-3`)
- New `TabPFN3Model` following the TabPFN-2.6 subclass pattern: v3 default checkpoints,
  fixed random state for CV/refit consistency, v3 auxiliary caps (`max_rows` 500k,
  `max_features` 2500, `max_classes` 160, `max_batch_size` 150k), and a v3-specific
  memory estimate (the v2 layer/embedding heuristic does not transfer to the v3
  architecture).
- Quantile regression is supported out of the box (inherited machinery; the v3
  regressor natively serves distribution quantiles).
- `tabpfn` extra bumped to `>=8.0,<8.1` (required for the v3 checkpoints; 8.x retains
  the `tabpfn.model.loading` import path the existing 2.5/2.6 code uses).

### TabDPT-Turbo (`TABDPT-TURBO`)
- New `TabDPTTurboModel` (TabDPT v1.2): pins the `tabdpt1_2.safetensors` checkpoint,
  defaults to subsampled context reduction (`context_reduction="subsample"`,
  `clip_sigma=8`) and exposes the v1.2 predict knobs (`n_ensembles`, `batch_size`).
  ~120x faster fit+inference than v1.1 on TabArena with improved accuracy
  (https://openreview.net/pdf?id=Y00pwFyrHR).
- `TabDPTModel` refactored to a class-attribute surface (`_checkpoint_filename`,
  `_constructor_defaults`, `_predict_param_names`) that versions override; its own
  behavior is unchanged for a fixed tabdpt package version.

### Base-class fixes (TabPFN family)
- Multi-GPU fits: `TabPFNModel` resolves the tabpfn `device` argument via
  `_get_tabpfn_device` — a device list when more than one GPU is assigned, clamped to
  the CUDA-visible device count (resource grants can be counted via NVML, which
  ignores `CUDA_VISIBLE_DEVICES`; unclamped this crashes with
  `CUDA error: invalid device ordinal` in the refit path). Default resources now
  request up to `max_gpus` (8).
- `custom_model_dir` can be passed as a hyperparameter (popped before estimator kwargs
  are built, threaded into `_resolve_model_path`).
- TabICL: quantile regression support (TabICL v2's regressor natively serves
  distribution quantiles via `predict(output_type="quantiles", alphas=...)`).
- Flattened the TabPFN version-class hierarchy: 2.6 and 3 no longer subclass the
  concrete `RealTabPFNv25Model` (whose only genuinely shared behavior was the license
  notice, while inheriting concrete state created traps such as silently sweeping 2.5
  checkpoints during HPO). All version classes now subclass the abstract `TabPFNModel`
  and declare their version facts fully; the noncommercial license notice moves into
  the base behind a `license_noncommercial` class flag.

### Core
- `ag.min_rows` fit constraint (`AbstractModel.validate_fit_args`): raises at fit time
  when `len(X) < min_rows`, for parity with `ag.max_rows`. Covered in
  `test_raise_on_fit_args`.

## ⚠️ Behavior change: `TABDPT` results differ from 1.5.x

The `tabdpt` extra moves from `>=1.1.11,<1.2` to `>=1.2,<1.3`. The tabdpt package can
only load checkpoints of its own version (tabdpt 1.2 fails on the v1.1 checkpoint with
an omegaconf schema error), so under the new pin the existing `TabDPTModel` runs
**TabDPT v1.2 weights and v1.2 code defaults** (subsampled context reduction instead of
FAISS retrieval, `clip_sigma` 8 instead of 4). Predictions will not reproduce
previously benchmarked `TABDPT` results; v1.2 is more accurate and much faster, but
exact 1.5.x reproduction requires `tabdpt<1.2` (which cannot coexist with
`TABDPT-TURBO` in the same environment). Freezing the old behavior alongside Turbo is
not possible with a single installable package version.

## Testing

- `test_tabpfn3.py`, `test_tabdpt_turbo.py` (new), `test_tabpfnv2.py`, `test_tabdpt.py`,
  `test_tabicl.py`: full `FitHelper.verify_model` runs (fit / predict / save-load /
  load-without-CUDA / refit across all supported problem types, including quantile)
  pass on GPU.
- `test_model_registry.py`: 176 passed.
- `test_tabpfn3.py` is skipped by default: the v3 weights are not publicly
  downloadable without accepting a license agreement, so CI cannot fetch them. The
  test passed when run manually with the checkpoints in the tabpfn cache.


LennartPurucker (approved):
LGTM!

PR #5756, opened 2026-07-27
Add validation_structure fit kwarg for grouped/temporal validation splits

Innixma:
## Summary

Adds a `validation_structure` fit kwarg for leakage-free validation on grouped and temporal data:

```python
predictor = TabularPredictor(label="churned").fit(
    train_data,
    validation_structure={
        "group_on": "customer_id",     # str | list[str]: group-disjoint splits
        "time_on": "timestamp",        # datetime/numeric: contiguous time blocks; forward holdout
        "group_time_on": "session_id", # groups ordered in time: disjoint AND forward
        "stratify_on": "region",       # optional stratification
    },
)
```

Users declare only the semantic columns; block boundaries, clamping (to group counts / stratification minority classes), and the resulting splits are derived. Referenced columns remain features.

## Design

**New**: `autogluon.common.utils.validation_structure.ValidationStructure` — the declarative spec plus model-agnostic resolvers: `custom_splits()` (bagged path), `holdout_split_indices()` (non-bagged path), and `fold_ids()` (per-row labels). It sits in `common` beside `cv_splitter`, which consumes the splits it produces; the problem-type constants it needs moved to a new `autogluon.common.constants` (re-exported from `autogluon.core.constants`, so existing imports are unaffected) since `common` does not depend on `core`.

### Bagged path

The learner computes explicit `(train_idx, val_idx)` splits on the raw cleaned frame — before feature transformation can alter the referenced columns — and delivers them to the models through `core_kwargs["ag_args_ensemble"]["custom_splits"]`, which `CVSplitter` already consumes.

Explicit splits rather than the `groups` channel, because that channel forces `LeaveOneGroupOut` with `n_repeats == 1` (`CVSplitter` raises `AssertionError: n_repeats must be 1 when split groups are specified`) and silently overrides the fold count to the group count. Repeated grouped cross-validation now works.

- `group_on`: group-disjoint folds. When every group carries a single stratification value, the *group table* (one row per group) is split with `Repeated(Stratified)KFold` and expanded back to rows, so group sizes do not skew stratification; otherwise `StratifiedGroupKFold` / `GroupKFold` over samples.
- `time_on`: contiguous, tie-preserving blocks of the observed time values (never equal-width spans), each cut placed where the cumulative row count is closest to that block's share.
- `group_time_on`: whole groups blocked in time order, so splits are simultaneously group-disjoint and forward in time. `group_on` + `time_on` together points here rather than raising a bare `NotImplementedError`.

**Clamping**: folds drop to the group count, and to the rarest stratification value, each forcing `num_repeats = 1` since the partition is then deterministic. Temporal partitions are likewise single-repeat. Callers must adopt the returned counts rather than assume their originals held.

### Holdout path

The holdout has no splitting algorithm of its own — it is derived from the bagged construction, so tie handling, cut placement, stratification and the per-group/per-sample pathway each exist in exactly one place and cannot drift:

- **grouped**: one fold of the split the bagged path would build. (`GroupShuffleSplit` cannot stratify, so a task declaring both `group_on` and `stratify_on` previously dropped its stratification silently.)
- **temporal**: the latest block of the same blocking, which is also strictly better at honoring the requested size than cutting at the fraction and walking the boundary back over ties — that walk moves in one direction only, so a large tie block overshoots badly (23,697 rows against a 17,588-row target on the worst benchmark task, versus 16,143 for the nearest block boundary).
- **stratify-only**: returns `None`; AutoGluon's default label-stratified holdout needs no correction.

`holdout_frac` selects the fold/block count and is honored closely when the structure supports it (0.05 -> 0.050, 0.2 -> 0.200 on 200 groups). With coarse grouping the realised size is larger, because whole groups cannot be subdivided — inherent to group-disjoint holdout, and documented.

### Dynamic stacking

Both DyStack validation procedures honor the declared structure, so the stacked-overfitting detector is not blind on the data where structure-aware validation matters most: its cross-validation procedure uses `custom_splits()`, and its holdout procedure resolves `holdout_split_indices()` up front and passes it through the `train_indices`/`val_indices` channel the CV procedure already used (the sub-fit itself is unchanged). The sub-fits' own bagging inherits the structure through `ag_fit_kwargs`, so both the outer split and the inner folds are covered.

Other notes: `stratify_on` may name the label rather than a feature column; `groups` (predictor init) and `validation_structure` are mutually exclusive; without `validation_structure` nothing is injected and the default path is untouched. Fold and repeat *counts* are deliberately not derived from data size — that policy is left to the caller, pending more customizable logic in a follow-up.

## Validation

Compared split-for-split against the grouped/temporal validation protocol TabArena applies to its non-IID benchmark tasks, on real BeyondArena data, with fold/repeat counts and seed matched:

| path | exact match |
|---|---|
| bagged | **39 / 39** |
| holdout | **33 / 39** (all 21 temporal exact) |

The 39 tasks cover every distinct configuration: both `group_labels` modes, stratified and unstratified, binary / multiclass / regression, grouped and temporal, from 906 to 1,000,350 rows.

Reaching bagged parity fixed several things a code read alone would not surface: the per-group and per-sample pathways are different algorithms; `stratify_on` can name the label; `GroupKFold` needs shuffling with the seed; and temporal fold *indices* come from routing block labels through `GroupKFold` rather than emitting blocks chronologically (identical partition set either way, so OOF predictions are unaffected — only fold indices differ).

The 6 remaining holdout differences are all grouped+stratified, where the reference calls a dedicated single-split routine (a seed-42 `train_test_split` over the group table for `per_group`, or a fold count derived from average group size for `per_sample`) instead of taking a fold of the k-fold split. Both constructions are group-disjoint, stratified, and clamp identically where whole groups cannot be subdivided; measured against the `1/num_folds` target neither is more accurate (mean |deviation| 23.3% for both, medians 0.9% vs 0.7%, opposite signs per task), so this is treated as equivalent-in-kind rather than a defect.

**IID behavior is unchanged.** The reference builds no custom splits for IID tasks even when `stratify_on` is set, deferring to AutoGluon's built-in stratified splitting; this PR likewise leaves the default path untouched when no structure is declared.

**Performance.** Building the group table originally scanned the full group column once per group, making construction quadratic in group count. With a single `pd.factorize` plus a boolean group->row mask per fold, construction on the largest grouped task (1,000,350 rows / 82,952 groups) drops from ~676s to 0.2s. Across all 39 tasks construction totals ~10s (median 40ms).

## Testing

`tabular/tests/unittests/test_validation_structure.py` (18 tests, ~17s): group-disjointness of folds, fold clamping to group counts, time-block contiguity + tie safety, forward-holdout ordering, stratify-only deferring to the default split, repeated grouped bagging, `group_time_on` disjoint-and-forward blocks, the mutual-exclusion errors, and end-to-end `TabularPredictor.fit` integrations.

Coverage includes the three paths that previously ignored or broke on the declared structure: `use_bag_holdout` (asserting the split indices address the reduced frame the trainer bags over) and both DyStack validation procedures (asserting their sub-fit splits are group-disjoint).

Several tests assert on the splits a fitted bagged model was *actually* fit with, rather than on child-model counts. That distinction matters: plain KFold produces the same child count for a given `num_bag_folds x num_bag_sets`, so splits that are computed and then dropped look identical from the outside.

## Limitations / follow-ups

- Fold/repeat counts are not derived from data size (deliberately deferred).
- `stratify_on` alone builds explicit stratified splits; deferring to AutoGluon's built-in label stratification when it resolves to the label would match the holdout path, which already defers.
- No benchmark task uses multi-column `group_on` or sets `group_on` with `time_on`, so `group_time_on` is covered by unit tests only.

### DIFF
--- a/common/src/autogluon/common/utils/validation_structure.py
+++ b/common/src/autogluon/common/utils/validation_structure.py
@@ -6,7 +6,7 @@ from dataclasses import dataclass, fields
 import numpy as np
 import pandas as pd
 
-from autogluon.core.constants import BINARY, MULTICLASS
+from autogluon.common.constants import BINARY, MULTICLASS
 
 logger = logging.getLogger(__name__)
 
--- a/core/src/autogluon/core/utils/validation_structure.py
+++ b/core/src/autogluon/core/utils/validation_structure.py
@@ -314,11 +314,14 @@ class ValidationStructure:
                 )
             return order[:cut], order[cut:]
 
-        from sklearn.model_selection import GroupShuffleSplit
-
-        splitter = GroupShuffleSplit(n_splits=1, test_size=holdout_frac, random_state=random_state)
-        train_idx, val_idx = next(splitter.split(X, y, groups=self._group_values(X)))
-        return train_idx, val_idx
+        # Grouped holdout: one fold of the same split the bagged path would build, so the
+        # holdout inherits the group-disjointness *and* the stratification handling rather
+        # than re-deriving them (GroupShuffleSplit cannot stratify at all). The fold count
+        # is chosen so a single fold is ~holdout_frac of the data; with coarse grouping the
+        # realised size can differ, because whole groups cannot be subdivided.
+        n_splits = max(2, int(round(1 / max(holdout_frac, 1e-9))))
+        splits, _, _ = self.custom_splits(X, y, num_folds=n_splits, num_repeats=1, random_state=random_state)
+        return splits[0]
 
 
 def _splits_from_labels(labels: pd.Series) -> list[tuple[np.ndarray, np.ndarray]]:
--- a/core/src/autogluon/core/utils/validation_structure.py
+++ b/core/src/autogluon/core/utils/validation_structure.py
@@ -15,10 +15,10 @@ logger = logging.getLogger(__name__)
 class ValidationStructure:
     """Declarative description of a dataset's validation-relevant structure.
 
-    Users specify only the semantic columns; fold counts, interval blocking, and
-    clamping are derived. Consumed by the learner (bagged path: fold-id labels routed
-    through the existing ``groups`` channel) and the trainer (holdout path: a
-    group-disjoint or temporally-forward train/validation split).
+    Users specify only the semantic columns; fold counts, interval blocking, and clamping
+    are derived. The learner resolves the bagged splits (:meth:`custom_splits`) and the
+    trainer the non-bagged holdout (:meth:`holdout_split_indices`), both on the raw cleaned
+    frame, before feature transformation can alter the columns named here.
 
     Parameters
     ----------
@@ -33,8 +33,12 @@ class ValidationStructure:
         to the label for classification when combined with ``group_on``; for time-based
         splits it only influences how the contiguous blocks are assigned to fold indices,
         never the block boundaries themselves.
-
-    Specifying both ``group_on`` and ``time_on`` is not supported.
+    group_time_on : str, optional
+        Column identifying groups that are *also* ordered in time (e.g. a session id whose
+        sessions arrive in sequence). Whole groups are blocked in time order, so the splits
+        are simultaneously group-disjoint and forward in time, and the non-bagged holdout is
+        the latest groups. Use this for data that is both grouped and temporal; combining
+        ``group_on`` with ``time_on`` is not supported, as their semantics would be ambiguous.
     """
 
     group_on: str | list[str] | None = None
@@ -69,8 +73,20 @@ class ValidationStructure:
             return cls(**value)
         raise ValueError(f"`validation_structure` must be a dict or ValidationStructure, got: {type(value)}")
 
-    def _is_per_group(self, X: pd.DataFrame, y: pd.Series, problem_type: str | None = None) -> bool:
-        """True when each group has a single stratification value (group-level labelling)."""
+    def _can_split_group_table(self, X: pd.DataFrame, y: pd.Series, problem_type: str | None = None) -> bool:
+        """Whether the folds can be built from the group table rather than from the samples.
+
+        True only when a stratification signal exists *and* is constant within every group:
+        the group table holds one row per group, so it can only carry one stratification
+        value per group. Splitting it then balances the stratification across folds without
+        letting group sizes skew it.
+
+        This is therefore always False without a stratification signal -- notably for
+        regression, which does not stratify -- and such tasks fall through to the
+        sample-level ``GroupKFold`` / ``StratifiedGroupKFold`` path. That matches the
+        grouped-validation protocol this mirrors, whose group-table branch likewise requires
+        a stratification column.
+        """
         stratify = self._resolve_stratify_for_folds(X, y, problem_type)
         if stratify is None:
             return False
@@ -155,7 +171,7 @@ class ValidationStructure:
                     )
                     num_folds, num_repeats = max(2, minority), 1
 
-        if self.group_on is not None and self._is_per_group(X, y, problem_type):
+        if self.group_on is not None and self._can_split_group_table(X, y, problem_type):
             # Every group carries one stratification value: split the *group table* (one row
             # per group) and expand back to rows, so group sizes do not skew stratification.
             splits = _per_group_splits(
--- a/core/src/autogluon/core/utils/validation_structure.py
+++ b/core/src/autogluon/core/utils/validation_structure.py
@@ -344,9 +344,10 @@ def _validate_splits(splits: list[tuple[np.ndarray, np.ndarray]], n_rows: int, s
         if train_idx.max(initial=-1) >= n_rows or val_idx.max(initial=-1) >= n_rows:
             raise ValueError("validation_structure produced out-of-range split indices.")
     if stratify is not None:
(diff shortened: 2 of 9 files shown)


LennartPurucker:
Should I review this state of the PR or wait?

LennartPurucker (approved):
LGTM, I assume the divergences from the BeyondArena split structure (e.g., each no-label per-group cases and UX support, but not implemented temporal-grouped cases) are intentional.

Innixma:
> LGTM, I assume the divergences from the BeyondArena split structure (e.g., each no-label per-group cases and UX support, but not implemented temporal-grouped cases) are intentional.

I've addressed the remaining divergences, so it is ready for a 2nd pass

common/src/autogluon/common/utils/validation_structure.py
```diff
@@ -0,0 +1,467 @@
...
+        Column(s) identifying groups that must not span both a training and a
+        validation split (e.g. a customer or session id).
+    time_on : str, optional
+        Column encoding time (datetime or numeric). Validation splits become
+        contiguous time blocks; the non-bagged holdout is the latest block.
+    stratify_on : str, optional
+        Column to stratify splits on (may name the label rather than a feature). Defaults
+        to the label for classification when combined with ``group_on``; for time-based
+        splits it only influences how the contiguous blocks are assigned to fold indices,
+        never the block boundaries themselves.
```
LennartPurucker, line 35: doc for group_time_on missing

core/src/autogluon/core/utils/validation_structure.py
```diff
@@ -0,0 +1,467 @@
...
+                    f"Invalid `validation_structure` keys: {sorted(invalid)}. Valid keys: {sorted(valid)}"
+                )
+            return cls(**value)
+        raise ValueError(f"`validation_structure` must be a dict or ValidationStructure, got: {type(value)}")
+
+    def _is_per_group(self, X: pd.DataFrame, y: pd.Series) -> bool:
+        """True when each group has a single stratification value (group-level labelling)."""
+        stratify = self._resolve_stratify_for_folds(X, y)
+        if stratify is None:
+            return False
```
LennartPurucker: I am not sure I get this part, would this not always be False for regression? As we dont stratify for regression

core/src/autogluon/core/utils/validation_structure.py
```diff
@@ -0,0 +1,467 @@
...
+        if n_folds < n_splits:
+            logger.log(
+                20,
+                f"validation_structure: data supports only {n_folds} folds (requested {n_splits}).",
+            )
+        return pd.Series(labels, index=X.index, name="__fold_id__")
+
+    def _resolve_stratify_for_folds(self, X: pd.DataFrame, y: pd.Series) -> pd.Series | None:
+        stratify = self._stratify_values(X, y)
+        if stratify is None and y.nunique() <= max(20, int(np.sqrt(len(y)))):
```
LennartPurucker: That is a pretty big point here, if you want to use this, document it better and make it a function. 
Moreover, we should test this, so enalbe us to On/off the function

LennartPurucker (changes requested):

core/src/autogluon/core/utils/validation_structure.py
Innixma: Good catch, I also didn't notice this. Claude tried to reproduce the problem type detection logic, but we can just pass the problem_type directly since it is known

Innixma: fixed

common/src/autogluon/common/utils/validation_structure.py
Innixma: added

core/src/autogluon/core/utils/validation_structure.py
Innixma: Claude:

Correct, it is always `False` for regression, and that's intended — but the name made that look like a bug, so I've renamed it and documented the reasoning.

The method isn't really asking "is this data labelled per group?" (a property of the grouping). It's asking "can the folds be built from the *group table* instead of from the samples?" That's only possible when a stratification signal exists **and** is constant within every group, because the group table holds one row per group and so can carry only one stratification value per group. No stratification signal means nothing to balance at the group level, so there's no reason to prefer it over splitting the samples.

Regression therefore falls through to the sample-level `GroupKFold` path, which is the correct outcome: group-disjoint folds, no stratification. Nothing is lost, and the group-table branch isn't dead code — it's reachable for classification, which is where it matters, since splitting the group table stops uneven group sizes from skewing the class balance across folds.

This also matches the protocol we're mirroring: data_foundry's `per_group` branch requires a stratification column too, so grouped regression tasks go through its `GroupKFold` path as well. That's part of why split parity holds on all 39 grouped/temporal BeyondArena tasks.

Renamed to `_can_split_group_table` with the above in the docstring (94c9eecc).

One genuine open question your comment surfaces, which I've deliberately left alone: for regression with *very uneven* group sizes, splitting the group table would balance group counts rather than row counts, which is arguably better. I didn't change it because it would diverge from the reference protocol — happy to file it as a follow-up if you think it's worth pursuing.

LennartPurucker (approved):
LGTM, I think time and in production/benchmark use will tell us all that is misisng from the design

PR #5757, opened 2026-07-27
GPU memory estimation plumbing + calibrated CPU/GPU memory estimates for foundation models

Innixma:
## Summary

Gives models a first-class way to declare **GPU memory (VRAM) estimates**, mirroring the existing CPU memory machinery, and ships **measurement-calibrated CPU/GPU estimates** for every GPU model in `tabular` (TabPFN-3, TabPFN-2.6, TabICL, TabDPT-Turbo, TabM, RealMLP, Nori) plus a fix to LightGBM's CPU estimate. Groundwork for safely parallelizing GPU models without VRAM OOMs.

Estimates are calibrated against **776 measured fits on 136 real tasks** (51 TabArena + 85 BeyondArena, 100 to 1M rows, fold 0, AutoGluon-preprocessed) on top of synthetic sweeps. Peak VRAM is the NVML per-process peak over fit + predict, measured in an isolated subprocess.

membench_estimate_accuracy_2

**Estimate vs measured across all 136 real tasks** (ratio = estimate ÷ measured peak; >1 is conservative, the safe direction):

| model | tasks | min | median | max | below parity |
|---|---|---|---|---|---|
| TabDPT-Turbo | 129 | 1.01 | 1.12 | 1.85 | none |
| TabPFN-2.6 | 113 | 1.00 | 1.25 | 3.79 | none |
| Nori-30M | 30 | 1.03 | 1.24 | 2.20 | none |
| TabPFN-3 | 132 | 1.02 | 1.31 | 4.28 | none |
| TabM | 136 | 1.00 | 1.61 | 7.25 | none |
| RealMLP | 136 | 1.04 | 2.30 | 2.46 | none |

No model underestimates on any task. TabICL is excluded from the table: its estimate is deliberately a *minimum requirement*, not expected usage (see below).

## GPU memory plumbing

CPU-parity API on `AbstractModel`:
- `_estimate_gpu_memory_usage` / `estimate_gpu_memory_usage_static` (+ `can_estimate_gpu_memory_usage_static` class tag). Default: no estimate, checks skipped — existing models unaffected.
- `_validate_fit_gpu_memory_usage`: fit-time check against available VRAM via the new `ResourceManager.get_available_vram()` (`torch.cuda.mem_get_info` with `nvidia-smi` fallback), governed by `ag.max_gpu_memory_usage_ratio` (default 1.0; None disables). Raises `NotEnoughCudaMemoryError` (already handled by the trainer as a graceful skip). Only runs when the model is assigned a GPU **and** defines an estimate.

## Calibrated GPU estimates

`R` = train + prediction rows, `f` = features, `cells` = rows x features.

| model | VRAM estimate |
|---|---|
| **TabPFN-3** | `1.1GB + 40KB·n_train + p + b + 18MB·min(f,230)·[1.5 if reg]`, where `p` = `26KB·n_test` (classification) or `110KB·min(n_test,100k) + 40KB·(n_test−100k)₊` (regression), `b` = `min(2.9GB, 1MB·n_train)` for regression else 0 |
| **TabPFN-2.6** | `0.85GB + m·(r + 3.8KB·min(cells,5M) + 0.8KB·(cells−5M)₊ + 0.4KB·R·(min(f,1000)−300)₊)`, `cells = R·min(f,300)`, `r` = `25KB·R` (`40KB·R` regression), `m` = 2.5 for regression |
| **TabDPT-Turbo** | `1.3GB + 25KB·(n_train + n_test)` — feature-independent |
| **TabM** | `1.2GB + 10.5MB·f_eff + 140B·min(cells,100M) + 40B·(cells−100M)₊ + 1MB·Σcat_levels`, `f_eff = min(f + #numeric-cols-with-NaN, 2000)` |
| **RealMLP** | `1.65GB + 660B·n_train + 50KB·f_eff + 28B·n_train·f_eff`, `f_eff` = post-encoding width (pytabkit one-hots cardinality ≤ `max_one_hot_cat_size`, else `embedding_size` columns) |
| **Nori** | `1GB + 32KB·min(cells,1M) + 8KB·(cells−1M)₊ + 0.5MB·n_test` |
| **TabICL** | `0.7GB + 250B·total_cells` — **minimum requirement, not usage** |

Non-obvious behaviors the measurements forced into the formulas:

- **TabPFN-2.6's per-cell cost is piecewise** because tabpfn itself switches to a memory-saving execution mode above a total-cell threshold (`tabpfn.architectures.base.memory.should_save_peak_mem`, ~6M cells on an 80GB device, scaled by free VRAM), cutting per-cell cost ~5×. Modeling that knee is what makes same-row-count tasks with different widths fit the same formula.
- **TabM's and Nori's per-cell costs saturate.** Linear-in-cells extrapolated to 359GB (TabM) on a 1M×1903 task measuring 79GB; past the knee, batches cover a shrinking fraction of the data and activations stop scaling with it.
- **TabM and RealMLP need post-encoding width, not raw column count** — categorical expansion and NaN-indicator columns change effective width by up to 30×, and the two models expand differently (a 131-row × 12.6k-feature frame uses ~12GB in TabM but ~0.9GB in RealMLP).
- **Nori is memory-heavy for a small model**: measured peaks reach 95GB on an ordinary 24k-row × 387-feature task (79GB on 12.8k × 739), well inside its own 50k-row cap. It previously had no GPU estimate at all.
- **TabICL plans batches from free VRAM**: it uses ~30GB idle but completes in <1.5GB free at ~2× runtime, so its estimate is the requirement floor — estimating usage would needlessly skip it on busy devices.

## Calibrated CPU estimates

| model | estimate |
|---|---|
| TabPFN-3 | `2.8GB + 10 float64 copies of (train + prediction batch) × min(f,500)` (replaces a flat 10GB guess) |
| TabPFN-2.6 | `2.25GB + 8.5 copies × R·f` |
| TabM | existing structural formula, with the feature count capped like the GPU estimate and the baseline raised 0.3 → 1.6GB (measured torch-process floor). Was 0.44–148× of measured (a 22k-feature frame estimated 318GB for a model that fits in 30); now 1.1–17×, no underestimates |
| LightGBM | +50MB fixed-overhead intercept: small fits were underestimated 3–11× by the size-proportional formula; worst case now ≤2.9× |

## Behavior changes

- **Prediction-row proxy is now the fit-time batch.** These estimates bound *fit* memory, where predictions are on held-out folds of the training data, so the proxy is `min(ag.max_batch_size, n_train)` rather than `max(100k, n_train)`. This was the single largest accuracy win (TabPFN-3 median 2.96× → 1.36×, TabDPT-Turbo 2.89× → 1.12×). Inference on a test set far larger than the training data is not covered — that is inference-time memory, which AutoGluon's fit-time checks never bounded.
- **TabPFN family `max_batch_size` default is `"auto"`** → resolved at fit time to `min(1M, max(max_batch_size_min, n_train))`. The floor is a per-class attribute: **1k for TabPFN-2.5/2.6**, which re-process the joint train+batch sequence per chunk (a 10k-train/50k-test regression fit peaks at 12.9GB instead of 40.6GB), and **100k for TabPFN-3**, whose KV-cached context makes small chunks a predict-time loss for little memory. Explicit `None` genuinely disables chunking.
- **TabICL's `max_batch_size=1024` is removed** (`None`): each 1024-row chunk re-encoded the full training context, making a 100k-row prediction ~50× slower (1462s → 28s measured) while saving no VRAM — tabicl batches internally, sizing to free device memory with OOM-halving.
- **Nori gains `ag.max_batch_size=10000`** (as a class constant shared with its estimate): an unchunked 50k-query predict against a 50k-row context fails with a CUDA kernel-configuration error after ~76GB; 10k-row chunks on the same context run in ~58GB.

## Testing

`tabular/tests/unittests/test_gpu_memory_estimate.py` (7 tests): estimate defaults / None-skip, VRAM check raise/pass/skip paths, ratio-None disable, static `NotImplementedError`, VRAM detection. Existing model constraint tests and `test_nori.py` pass. Lint matches the pinned pre-commit ruff (v0.16.0): `ruff check --select I` and `ruff format` clean.

### Known limitations

- **RealMLP's 2.30× median is floor-bound, not fixable by coefficients**: it measures 0.65–2GB on 125 of 136 tasks, and the 1.65GB base cannot drop without underestimating the tasks that do sit near 1.5GB.
- **TabM's 7.25× tail** is small-row/wide-feature tasks (~114–380 rows × ~1000 features), where the per-feature term is needed at 2000+ features but overshoots around 1000.
- **Nori's calibration is regression-only** (all it supports) and bounded by its 50k-row cap; its tightest task measured 94.8GB on a 96GB device, so if that peak was clipped by available VRAM the true requirement there may be higher.

### DIFF
--- a/common/src/autogluon/common/utils/resource_utils.py
+++ b/common/src/autogluon/common/utils/resource_utils.py
@@ -85,6 +85,26 @@ class ResourceManager:
         return num_gpus
 
     @staticmethod
+    def get_available_vram(device: int = 0) -> float | None:
+        """Available GPU memory (VRAM) of `device` in bytes, or None when it cannot be determined.
+
+        Uses `torch.cuda.mem_get_info` when torch with CUDA is available, falling back to
+        `nvidia-smi`. The GPU counterpart of `get_available_virtual_mem`.
+        """
+        try:
+            import torch
+
+            if torch.cuda.is_available() and device < torch.cuda.device_count():
+                free, _total = torch.cuda.mem_get_info(device)
+                return float(free)
+        except Exception:
+            pass
+        memory_free_values = ResourceManager.get_gpu_free_memory()  # MiB per device
+        if device < len(memory_free_values):
+            return float(memory_free_values[device]) * 1024**2
+        return None
+
+    @staticmethod
     def get_gpu_free_memory():
         """Grep gpu free memory from nvidia-smi tool.
         This function can fail due to many reasons(driver, nvidia-smi tool, envs, etc) so please simply use
--- a/common/src/autogluon/common/utils/resource_utils.py
+++ b/common/src/autogluon/common/utils/resource_utils.py
@@ -88,15 +88,38 @@ class ResourceManager:
     def get_available_vram(device: int = 0) -> float | None:
         """Available GPU memory (VRAM) of `device` in bytes, or None when it cannot be determined.
 
-        Uses `torch.cuda.mem_get_info` when torch with CUDA is available, falling back to
-        `nvidia-smi`. The GPU counterpart of `get_available_virtual_mem`.
+        The GPU counterpart of `get_available_virtual_mem`. Three effects make this more
+        than a `torch.cuda.mem_get_info` call:
+
+        1. `mem_get_info` reports memory free on the *device*, which excludes memory
+           PyTorch's caching allocator already holds. That memory is reusable by this
+           process without a new device allocation, so it is added back
+           (`memory_reserved - memory_allocated`); ignoring it under-reports what a fit
+           can actually use and needlessly skips models.
+        2. `torch.cuda.set_per_process_memory_fraction` caps this process below the
+           device total. The cap is not visible to `mem_get_info`, so it is applied here —
+           a process allocating past its fraction OOMs even with the device free.
+        3. Without torch/CUDA, `nvidia-smi` gives device-level free memory only (no
+           allocator or fraction information available).
         """
         try:
             import torch
 
             if torch.cuda.is_available() and device < torch.cuda.device_count():
-                free, _total = torch.cuda.mem_get_info(device)
-                return float(free)
+                device_free, device_total = torch.cuda.mem_get_info(device)
+                cached_unused = torch.cuda.memory_reserved(device) - torch.cuda.memory_allocated(device)
+                available = float(device_free + cached_unused)
+
+                # Respect a per-process cap when one is set (used to partition a GPU
+                # across processes). The getter exists from torch 2.9; older versions
+                # expose no way to read it back, so the cap is simply not applied.
+                get_fraction = getattr(torch.cuda, "get_per_process_memory_fraction", None)
+                if get_fraction is not None:
+                    fraction = float(get_fraction(device))
+                    if fraction < 1.0:
+                        process_headroom = fraction * device_total - torch.cuda.memory_allocated(device)
+                        available = min(available, max(process_headroom, 0.0))
+                return min(available, float(device_total))
         except Exception:
             pass
         memory_free_values = ResourceManager.get_gpu_free_memory()  # MiB per device
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -56,7 +56,7 @@ from ...utils import (
     infer_problem_type,
     normalize_pred_probas,
 )
-from ...utils.exceptions import NotEnoughMemoryError, NoValidFeatures, TimeLimitExceeded
+from ...utils.exceptions import NotEnoughCudaMemoryError, NotEnoughMemoryError, NoValidFeatures, TimeLimitExceeded
 from ...utils.loaders import load_json, load_pkl
 from ...utils.savers import save_json, save_pkl
 from ...utils.time import sample_df_for_time_func, time_func
@@ -516,6 +516,7 @@ class AbstractModel(ModelBase, Tunable):
         """
         default_auxiliary_params = dict(
             max_memory_usage_ratio=1.0,  # Ratio of memory usage allowed by the model. Values > 1.0 have an increased risk of causing OOM errors. Used in memory checks during model training to avoid OOM errors.
+            max_gpu_memory_usage_ratio=1.0,  # GPU counterpart of max_memory_usage_ratio: ratio of available VRAM the model is allowed to use. Only checked for models that define a GPU memory estimate (see `_estimate_gpu_memory_usage`). If None, GPU memory checks are skipped.
             # TODO: Add more params
             # max_memory_usage=None,
             # max_disk_usage=None,
@@ -1267,6 +1268,7 @@ class AbstractModel(ModelBase, Tunable):
         self._register_fit_metadata(**kwargs)
         self.validate_fit_resources(**kwargs)
         approx_mem_size_req, available_mem = self._validate_fit_memory_usage(**kwargs)
+        self._validate_fit_gpu_memory_usage(**kwargs)
         if "time_limit" in kwargs and kwargs["time_limit"] is not None:
             time_start_fit = time.time()
             kwargs["time_limit"] -= time_start_fit - time_start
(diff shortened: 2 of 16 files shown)


core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -2706,6 +2708,84 @@ def estimate_memory_usage_child(self, X: pd.DataFrame, **kwargs) -> int:
...
+        Estimates the peak GPU memory (VRAM) usage in bytes during model fitting.
+
+        Default: None (no estimate; VRAM checks are skipped). Models that fit on GPU
+        should override this to enable VRAM safety checks and, eventually, safe
+        parallel scheduling of GPU models.
+        """
+        return None
+
+    @classmethod
+    def can_estimate_gpu_memory_usage_static(cls) -> bool:
```
LennartPurucker, line 2743: Can we make this a property instead of a function, or similar? let us try to reduce the function overload for a user to implement, rather a bunch of flags we can toggle

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -2706,6 +2708,84 @@ def estimate_memory_usage_child(self, X: pd.DataFrame, **kwargs) -> int:
...
+        Returns
+        -------
+        int | None: estimated peak VRAM usage in bytes during training, or None if the model defines no estimate
+        """
+        assert self.is_initialized(), "Only estimate GPU memory usage after the model is initialized."
+        gpu_memory_usage_estimate = self._estimate_gpu_memory_usage(X=X, **kwargs)
+        self._gpu_memory_usage_estimate = gpu_memory_usage_estimate
+        return gpu_memory_usage_estimate
+
+    def _estimate_gpu_memory_usage(self, X: pd.DataFrame, **kwargs) -> int | None:
```
LennartPurucker, line 2732: Can we merge this somehow, with just a boolean flag make the difference (again thinking of API design here for end user adding a model).

common/src/autogluon/common/utils/resource_utils.py
```diff
@@ -84,6 +84,26 @@ def get_gpu_count_torch(cuda_only: bool = False) -> int:
...
+        """Available GPU memory (VRAM) of `device` in bytes, or None when it cannot be determined.
+
+        Uses `torch.cuda.mem_get_info` when torch with CUDA is available, falling back to
+        `nvidia-smi`. The GPU counterpart of `get_available_virtual_mem`.
+        """
+        try:
+            import torch
+
+            if torch.cuda.is_available() and device < torch.cuda.device_count():
+                free, _total = torch.cuda.mem_get_info(device)
```
LennartPurucker: Is this from Claude? I remember one needs to be a bit more clever, but maybe this has already been fixed in Torch. Might need to spend one prompt on making sure this is working well

core/src/autogluon/core/models/abstract/abstract_model.py
```diff
@@ -2706,6 +2708,84 @@ def estimate_memory_usage_child(self, X: pd.DataFrame, **kwargs) -> int:
...
+    @classmethod
+    def can_estimate_gpu_memory_usage_static(cls) -> bool:
+        """
+        True if `estimate_gpu_memory_usage_static` is implemented for this model.
+        If False, calling `estimate_gpu_memory_usage_static` will raise a NotImplementedError.
+        """
+        return cls._get_class_tags().get("can_estimate_gpu_memory_usage_static", False)
+
+    @classmethod
+    def estimate_gpu_memory_usage_static(
```
LennartPurucker, line 2751: dito, if we could merge this all into one function that would be nice (even if we cannot refactor the other code now)

tabular/src/autogluon/tabular/models/tabicl/tabicl_model.py
```diff
@@ -236,9 +240,38 @@ def _get_default_ag_args_ensemble(cls, **kwargs) -> dict:
...
+        total_cells = (n_train + n_test) * n_features
+        return int(0.7e9 + 250 * total_cells)
+
     @classmethod
     def _class_tags(cls) -> dict:
-        return {"can_estimate_memory_usage_static": True}
+        return {
+            "can_estimate_memory_usage_static": True,
+            "can_estimate_gpu_memory_usage_static": True,
+        }
```
LennartPurucker, line 276: In my perfect API, this should just be a class-level variable / toggle, not yet another function (e.g. more like we do it for data classes)

LennartPurucker (changes requested):
LGTM so far, some thoughts:

* Maybe not in this PR, but in the future, we should redesign the API to be less class-based, as it becomes a lot of code to read for new users. Here we could start the API revolution more or less :D 
* How does the fallback logic work? Do the cases we used so far, where RAM is a proxy for GPU, still work? 
* This logic is also much more important for TabM and RealMLP? Likewise, if this works well, we could parallelize folds across one GPU, etc

common/src/autogluon/common/utils/resource_utils.py
Innixma: Good catch! Claude indeed found there was a bug here and fixed it.

core/src/autogluon/core/models/abstract/abstract_model.py
Innixma: Lets leave this for a follow-up PR. I can ask Claude to try to refactor it and see if it runs into any issues. Ditto for the other comments

Innixma:
@LennartPurucker 

> How does the fallback logic work? Do the cases we used so far, where RAM is a proxy for GPU, still work?

Both memory estimates are now ran if present, and are independent of eachother.

For TabArena, we will need to update the fake GPU mem logic to also override for this GPU logic I think.

LennartPurucker:
I don't think we will need the fake GPU memory estimate anymore once this is merged and we migrate the TabArena models. 

But I would suggest: (1) fix API design / remove functions -> (2) Migrate models -> (3) Remove logic for fake memory estimate, introduce logic for GPU VRAM limit.

tabular/src/autogluon/tabular/models/tabm/tabm_model.py
```diff
@@ -258,8 +273,10 @@ def _estimate_tabm_ram(
 
         mem_ds = n_samples * (4 * n_numerical + 8 * len(cat_sizes))
 
-        # some safety constants and offsets (the 5 is probably excessive)
-        mem_total = 5 * mem_ds + 1.2 * mem_forward_backward + 1.2 * mem_params + 0.3 * (1024**3)
+        # some safety constants and offsets (the 5 is probably excessive); the
+        # baseline is the measured torch-process floor (~1.6 GB), which dominates
+        # on small datasets.
+        mem_total = 5 * mem_ds + 1.2 * mem_forward_backward + 1.2 * mem_params + 1.6e9
```
LennartPurucker, line 279: Mhm, did you check all the logic I had to add to make the mem est on BeyondArena better/not wrong in TabArena? Also, we could forward port the new implementation of TabM at some point

LennartPurucker (approved):
LGTM

LennartPurucker:
We should find a better way to avoid diverging model implementations between TabArena and AutoGluon

tabular/src/autogluon/tabular/models/tabm/tabm_model.py
Innixma: re the BeyondArena mem estimation, can you link it?

LennartPurucker: It us the current main branch state of TabArena, not sure what exaclty changed over the course of the benchmark dev

LennartPurucker: But I had to adjust it and RealMLP qutie a bit

PR #5758, opened 2026-07-28
Fix master lint_check: pin explicit ruff rule selection (ruff 0.16 default expansion)

Innixma:
## Summary

Master's `lint_check` CI job is failing on every branch since #5755 (the automated ruff pre-commit bump to v0.16.0). Root cause: **ruff 0.16 expanded its default `lint.select`**, and the root `pyproject.toml` relies on the implicit defaults — so `ruff check timeseries/` (the one full-rule lint in `lint_check.sh`) went from clean under 0.15.x to **247 errors across 15+ rule families** (UP, RUF, C4, TRY, SIM, BLE, PLR, PYI, S, …) that the project never opted into, with zero code changes.

## Fix

- Pin the pre-0.16 defaults explicitly: `lint.select = ["E4", "E7", "E9", "F"]`, with a comment so future rule-set expansion happens deliberately rather than via version bumps. This makes `lint_check` robust to the automated ruff updates.
- Fix the 2 genuine `--select I` import-order errors in `multimodal/tests` that 0.16 also surfaced (autofix).

## Verification

All `lint_check.sh` commands (`ruff format --diff` + `ruff check --select I` per package, `ruff check timeseries/`, run from repo root) pass under **both** ruff 0.16.0 (the CI-pinned version) and 0.15.17:

```
$ ruff check timeseries/   # ruff 0.16.0, after this PR
All checks passed!
```

An alternative would be adopting the new 0.16 defaults and fixing all 247 findings (~110 autofixable, the rest manual incl. behavior-adjacent ones like TRY004 exception-type changes) — deliberately not done here to unblock CI with a minimal, no-behavior-change diff; happy to follow up with an opt-in rule expansion if desired.

### DIFF
--- a/multimodal/tests/unittests/others/test_backbones.py
+++ b/multimodal/tests/unittests/others/test_backbones.py
@@ -2,12 +2,12 @@ import os
 import shutil
 
 import pytest
-from datasets import load_dataset
 
 from autogluon.multimodal import MultiModalPredictor
 from autogluon.multimodal.constants import IA3_LORA
 from autogluon.multimodal.models import HFAutoModelForTextPrediction, TimmAutoModelForImagePrediction
 from autogluon.multimodal.utils import download
+from datasets import load_dataset
 
 from ..utils import PetFinderDataset, get_home_dir, verify_no_redundant_model_configs, verify_predictor_save_load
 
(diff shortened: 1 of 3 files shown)


LennartPurucker (approved):
LGTM

PR #5760, opened 2026-07-29
[timeseries] Vendor Toto-2 instead of depending on the `toto-2` package

shchur:
*Issue #, if available:*

*Description of changes:*

`toto-2` Python package depends on Python 3.12 (https://github.com/DataDog/toto/issues/126), which is a major annoyance for its use in AutoGluon's default presets. This PR addresses this problem by vendoring the `toto-2` code inside AutoGluon.

## Summary
Replaces the optional `toto-2` PyPI dependency with a minimal, inference-only vendored implementation under `timeseries/src/autogluon/timeseries/models/toto2/_internal/`, following the same pattern already used for Toto 1.0. Toto-2 now works out of the box with `pip install autogluon.timeseries` — no `[toto]` extra and no Python 3.12+ constraint.

### Correctness verification

`_internal/unit_scaling.py` vendors the u-μP (unit-scaled maximal update parametrization) primitives that upstream pulls from `unit_scaling` and `dd_unit_scaling`. These fixed forward/backward multipliers are **part of the forward computation**, so any deviation would silently corrupt predictions rather than fail loudly. The vendored model was checked against `toto-2` 2.0.0 in two harnesses:

- **Real weights** (`Datadog/Toto-2.0-4m`): state-dict keys match exactly and all 7 scenarios are **bitwise identical** (`maxabs = 0.0`) — single-pass with and without missing values, block decoding, multi-patch block decoding, multivariate, multivariate block, and `horizon < patch_size`.
- **Randomly-initialized models** sharing weights via `load_state_dict`, covering config paths no published checkpoint uses: all 11 variants **bitwise identical** — `qk_norm`, `qk_norm_include_weight`, `norm_include_weight`, plain RoPE (no xPos), no `per_dim_scale`, `variate_layer_first`, grouped-query attention, `mlp_bias`, no `attn_bias`, and 2 variate layers per group.

### Reductions vs. upstream

- Inference only: no optimizers, training-time/distributed scale factors, dropout, `mup_type` tagging, `DepthModuleList` depth tagging, or general constraints machinery.
- `Toto2ModelConfig` rejects `pre_norm=False` and `num_output_patches != 1`. The latter is **broken upstream** (`RuntimeError: The size of tensor a (16) must match the size of tensor b (8)` in the quantile rescaling step) and every published checkpoint uses `num_output_patches=1`.
- `from_pretrained` loads `config.json` + `model.safetensors` directly instead of via `PyTorchModelHubMixin`. All 5 published repos are single-file, so the sharded-index branch is dropped.
- The three `ResidualMLP` `layer_type` variants collapse to a single `is_readout: bool`.
- Added a `context_length % patch_size` validation that upstream lacked.

### Other changes

- Moves the AutoGluon wrapper from `toto/toto2.py` to `toto2/model.py`, and `Toto2DataLoader` / `Toto2InferenceDataset` from `toto/dataloader.py` to `toto2/dataloader.py`.
- Removes the `[toto]` extra from `timeseries/setup.py` and declares `huggingface_hub[torch]` + `safetensors`, which are now direct (previously transitive via `transformers`) imports.
- Drops the `toto-2`-availability gating from the tests (`is_toto2_available`, the `skipif` mark, and the `find_spec` guard in the smoke tests), so Toto-2 is now always exercised in CI.


Innixma (approved):
LGTM!

PR #5761, opened 2026-07-29
Disable prediction caching by default in TimeSeriesPredictor

shchur:
## Description of changes

Most forecasting models now have acceptable inference speed, so caching predictions to disk is no longer worth the added complexity it introduces. This PR sets `cache_predictions=False` by default and marks prediction caching as deprecated.

- Flip `cache_predictions` to `False` in `TimeSeriesPredictor`, `TimeSeriesLearner` and `TimeSeriesTrainer` (the value chains through all three, so all must change together).
- Raise a `FutureWarning` when `cache_predictions=True` is passed explicitly. `FutureWarning` rather than `DeprecationWarning` because the latter is silenced by default outside `__main__`, so library users would never see it. The warning only fires on explicit opt-in, so nobody is affected by the default flip alone.
- Update docstrings for `cache_predictions` and the four `use_cache` arguments. The old note ("ignored if `cache_predictions` was set to False") now describes the default path, so it is reworded.

`use_cache` on `predict` / `backtest_predictions` / `evaluate` / `leaderboard` keeps its current default. It already means "use the cache *if* caching is enabled", so it is now a silent no-op via `NoOpPredictionCache`; flipping it would only break users who deliberately set `cache_predictions=True`. Both arguments, along with `trainer/prediction_cache.py`, are intended for removal in a future release.

### Behavior change

`leaderboard()` or `evaluate()` followed by `predict()` on the same data now re-runs inference instead of reusing cached predictions. Within a single call nothing changes: `get_model_pred_dict` still shares base model predictions through the in-memory dict, so ensembles are unaffected. `feature_importance` already passed `use_cache=False`.

## Testing

- Five existing tests that assert `FileBasedPredictionCache` now pass `cache_predictions=True` explicitly.
- Added tests covering the new default (`NoOpPredictionCache`), the `FutureWarning` on opt-in, and the absence of a warning when the argument is omitted or `False`.

Ran `pytest timeseries/tests/unittests/trainer/test_trainer.py -k "cache or warning"` and `timeseries/tests/unittests/trainer/test_prediction_cache.py` locally (20 passed). Broader suites left to CI.

### DIFF
--- a/timeseries/src/autogluon/timeseries/learner.py
+++ b/timeseries/src/autogluon/timeseries/learner.py
@@ -29,7 +29,7 @@ class TimeSeriesLearner(AbstractLearner):
         trainer_type: Type[TimeSeriesTrainer] = TimeSeriesTrainer,
         eval_metric: str | TimeSeriesScorer | None = None,
         prediction_length: int = 1,
-        cache_predictions: bool = True,
+        cache_predictions: bool = False,
         ensemble_model_type: Type | None = None,
         **kwargs,
     ):
--- a/timeseries/src/autogluon/timeseries/predictor.py
+++ b/timeseries/src/autogluon/timeseries/predictor.py
@@ -4,6 +4,7 @@ import math
 import os
 import pprint
 import time
+import warnings
 from pathlib import Path
 from typing import Any, Literal, Type, cast, overload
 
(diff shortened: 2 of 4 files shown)


Innixma (approved):
LGTM!

PR #5762, opened 2026-07-29
Update semantic_segmentation.py

eyh002:
*Description of changes:*

Fixed a parameter mapping bug where `predictor.evaluate()` was passing positional arguments incorrectly to `evaluate_semantic_segmentation()`. Explicitly pass keyword arguments `(data, metrics, return_pred, realtime)` so that `return_pred` is correctly forwarded, preventing a `ValueError` during tuple unpacking when evaluating semantic segmentation tasks.


shchur (approved):
Thanks for the fix!

PR #5764, opened 2026-07-30
[timeseries] Respect user-provided device for Chronos models

### LINKED ISSUE #5529: Support Non-CUDA Devices (e.g., MPS, XPU) for Chronos Models by Allowing Device Overrides
https://github.com/autogluon/autogluon/blob/0e21eb62cf69019cff2e709ec3aef42e3443e2fe/timeseries/src/autogluon/timeseries/models/chronos/model.py#L317

#### Summary
In AutoGluon Time Series (v1.5+), the device handling for Chronos models (including Chronos-2) forces a fallback to `'cpu'` if `torch.cuda.is_available()` is `False`, overriding custom `device` settings like `'mps'` (Apple Silicon) or `'xpu'` (Intel GPUs via IPEX). This prevents acceleration on non-CUDA hardware. Earlier versions (e.g., v1.4) allowed overrides.

#### Affected Code
In `timeseries/src/autogluon/timeseries/models/chronos/model.py`, the line in `load_model_pipeline`:

```python
device = (self.device or "cuda") if gpu_available else "cpu"
```

Where `gpu_available` checks only `torch.cuda.is_available()`.

In v1.4, it was:

```python
device = self.device or ("cuda" if gpu_available else "cpu")
```

#### Problem Details
- **Impact**: No acceleration for MPS/XPU users, despite PyTorch support via `device_map`. Defaults to CPU.
- **Reproduction**:
  1. Install AutoGluon >=1.5 with Time Series.
  2. Use hyperparameters: `{"Chronos": {"device": "mps"}}` (or `"xpu"`).
  3. Fit `TimeSeriesPredictor`—falls back to CPU without CUDA.

#### Proposed Solution
Revert to v1.4 logic:

```python
device = self.device or ("cuda" if gpu_available else "cpu")
```

Or expand detection in `_is_gpu_available`:

```python
def _is_gpu_available(self) -> bool:
    import torch
    return torch.cuda.is_available() or torch.mps.is_available() or torch.xpu.is_available()
```

Then:

```python
device = self.device or ("cuda" if torch.cuda.is_available() else "mps" if torch.mps.is_available() else "xpu" if torch.xpu.is_available() else "cpu")
```

This respects user devices without affecting CUDA users.

#### Benefits
- Enables 1.5-3x speedup on MPS/XPU for non-NVIDIA hardware.
- Aligns with PyTorch's multi-backend support.

Guidance on patches or workarounds appreciated!

shchur:
Fixes #5529

`device` provided by the user was overridden with `"cpu"` whenever `torch.cuda.is_available()` returned `False`, so non-CUDA backends like `"mps"` and `"xpu"` could not be used. This restores the pre-v1.5 resolution order in both `ChronosModel` and `Chronos2Model`:

```python
device = self.device or ("cuda" if gpu_available else "cpu")
```

The `device=None` default is unchanged (CUDA if available, else CPU), so CUDA and CPU users are unaffected. Requesting `"cuda"` without CUDA now raises from torch instead of silently falling back to CPU.

Added a parametrized test over `(device, cuda_available, expected_device)` for both models, asserting the `device_map` passed to `from_pretrained`.

Note: `ChronosModel`'s existing GPU guard is CUDA-only, so `chronos-t5-small/base/large` (`num_gpus=1`) still raise without CUDA. Left as-is; the fix applies to Chronos-Bolt, `t5-tiny/mini`, and Chronos-2.

### DIFF
--- a/timeseries/src/autogluon/timeseries/models/chronos/chronos2.py
+++ b/timeseries/src/autogluon/timeseries/models/chronos/chronos2.py
@@ -275,7 +275,7 @@ class Chronos2Model(AbstractTimeSeriesModel):
     def load_model_pipeline(self):
         from chronos.chronos2.pipeline import Chronos2Pipeline
 
-        device = (self.get_hyperparameter("device") or "cuda") if self._is_gpu_available() else "cpu"
+        device = self.get_hyperparameter("device") or ("cuda" if self._is_gpu_available() else "cpu")
 
         assert self.model_path is not None
         pipeline = Chronos2Pipeline.from_pretrained(
--- a/timeseries/src/autogluon/timeseries/models/chronos/model.py
+++ b/timeseries/src/autogluon/timeseries/models/chronos/model.py
@@ -308,7 +308,7 @@ class ChronosModel(AbstractTimeSeriesModel):
                 "`import torch; torch.cuda.is_available()` returns `True`."
             )
 
-        device = (self.device or "cuda") if gpu_available else "cpu"
+        device = self.device or ("cuda" if gpu_available else "cpu")
 
         assert self.model_path is not None
         pipeline = BaseChronosPipeline.from_pretrained(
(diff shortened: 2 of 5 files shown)


Innixma (approved):
LGTM!

PR #5705, opened 2026-06-30
[tabular] Add Nori

minkyu-choi07:
*Issue #, if available:*

*Description of changes:*

- Add Nori model (a small, regression-only tabular foundation model via in-context learning)
- Register `NoriModel` in the model registry (key `NORI`, name `Nori`, priority 40)
- Add `synthefy-nori` as an optional dependency (`autogluon.tabular[nori]`)

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #5195, opened 2025-07-08
[tabular] Add Mitra Model

xiyuanzh:
*Issue #, if available:*

*Description of changes:*

* Add Mitra Model to AutoGluon 

To test Mitra Model:

1. Create conda virtual environment and install packages

```
conda create -n ag python=3.11
conda activate ag
cd autogluon/tabular
bash full_install.sh
```

2. Install additional package (openml) to run the benchmarking, which is not required for AG users so it's not listed in setup.py

```
pip install openml
```

3. Run benchmarking on 10-fold TabRepo-66

```
python -m examples.tabular.example_mitra
```

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #5765, opened 2026-07-30
[tabular] Support loading Mitra from a local checkpoint path

shchur:
## Description

Mitra could only load pretrained weights from HuggingFace, so a local checkpoint path passed via `hf_cls_model` / `hf_reg_model` was always treated as a repo id and failed. These hyperparameters now accept a local directory too.

Three changes:

1. **Local paths** (`_internal/models/tab2d.py`) — `Tab2D.from_pretrained` called `hf_hub_download` unconditionally. It now reads `config.json` / `model.safetensors` from a local directory when given one, else falls back to HuggingFace. The `os.path.isdir` dispatch matches what HuggingFace's own `ModelHubMixin.from_pretrained` does. An incomplete directory raises `FileNotFoundError` naming the missing file, instead of a confusing 404.

2. **Remove the dead `state_dict` path** (`sklearn_interface.py`) — loading a raw `.pt` checkpoint was gated behind a module-level `USE_HF = True` that could not be set, so it has never executed since Mitra was added in #5195. Since (1) covers the use case — `Tab2D.save_pretrained(<dir>)`, then pass the directory — the branch and its `state_dict` parameter are removed rather than revived. `state_dict`, `state_dict_classification` and `state_dict_regression` are still popped in `_fit` so existing hyperparameter dicts keep working, but now log a warning pointing at `hf_model`; they were already no-ops.

3. **Fix `hf_*` alias leak** (`mitra_model.py`) — fallback aliases were only popped when no more specific alias matched, so an unused key stayed in the hyperparameters dict and was splatted into the model constructor. Passing an override plus a fallback crashed the fit:

   ```python
   hyperparameters={MitraModel: {"hf_cls_model": "org/cls", "hf_general_model": "org/gen"}}
   # TypeError: MitraClassifier.__init__() got an unexpected keyword argument 'hf_general_model'
   ```

   Alias resolution is now a small pure helper, `MitraModel._resolve_hf_model`, which pops all four aliases before applying precedence (problem-specific > general > generic).

Usage:

```python
predictor.fit(train_data, hyperparameters={
    MitraModel: {"hf_cls_model": "/path/to/local/ckpt"},   # config.json + model.safetensors
    # or: {"hf_cls_model": "some-org/some-mitra"}          # HuggingFace repo id
})
```

HuggingFace defaults are unchanged. No new hyperparameters.

Also included: a formatting-only lint fix for `nori_model.py` / `test_nori.py` (added in #5705), which currently fail `ruff format` / `ruff check --select I` on every branch. Happy to split this out.

### DIFF
--- a/tabular/src/autogluon/tabular/models/mitra/_internal/models/tab2d.py
+++ b/tabular/src/autogluon/tabular/models/mitra/_internal/models/tab2d.py
@@ -34,6 +34,18 @@ from ..._internal.models.embedding import (
 logger = logging.getLogger(__name__)
 
 
+def _resolve_pretrained_file(path_or_repo_id: str, filename: str) -> str:
+    """Locate `filename` inside a local checkpoint directory, falling back to a HuggingFace repo download."""
+    if os.path.isdir(path_or_repo_id):
+        local_path = os.path.join(path_or_repo_id, filename)
+        if not os.path.isfile(local_path):
+            raise FileNotFoundError(
+                f"Local checkpoint directory '{path_or_repo_id}' is missing the required file '{filename}'."
+            )
+        return local_path
+    return hf_hub_download(repo_id=path_or_repo_id, filename=filename)
+
+
 class Tab2D(BaseModel):
     def __init__(
         self,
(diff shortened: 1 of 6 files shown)


xiyuanzh (commented):
Thanks looks good to me!

tabular/src/autogluon/tabular/models/mitra/_internal/models/tab2d.py
xiyuanzh, line 1: Looks good to me!

xiyuanzh (approved):

Innixma (approved):
LGTM! Merging to fix CI

PR #5767, opened 2026-07-31
[timeseries] Add method for exporting fine-tuned models

shchur:
*Issue #, if available:*

*Description of changes:*

Adds `TimeSeriesPredictor.export_model(path, model)` to export a trained model as a standalone Hugging Face checkpoint that can be loaded directly with `chronos`, independent of AutoGluon.

- Supported for Chronos, Chronos-Bolt and Chronos-2 (including fine-tuned models). For fine-tuned Chronos-2, the LoRA adapter is merged into the base weights so the exported checkpoint is self-contained.
- `model` is a required argument; export raises with an actionable list of exportable models otherwise.
- Refuses to export a model that relies on AutoGluon-side data transforms (target/covariate scaler, covariate regressor), since these are applied outside the exported artifact and would silently change forecasts.

### DIFF
--- a/timeseries/src/autogluon/timeseries/learner.py
+++ b/timeseries/src/autogluon/timeseries/learner.py
@@ -1,6 +1,7 @@
 import logging
 import reprlib
 import time
+from pathlib import Path
 from typing import Any, Literal, Type
 
 import pandas as pd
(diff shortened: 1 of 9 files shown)


Innixma (approved):
LGTM!

PR #5768, opened 2026-07-31
Simplify the memory-estimate API: derive capability tags, default the instance wrappers

Innixma:
## Summary

Follow-up to #5757, addressing @LennartPurucker's review comments about the memory-estimate API surface ("let us try to reduce the function overload for a user to implement, rather a bunch of flags we can toggle" / "In my perfect API, this should just be a class-level variable / toggle, not yet another function").

Enabling a memory estimate currently takes **three** edits:

1. implement `_estimate_memory_usage_static` (or the GPU / `_lite` variant),
2. write an instance wrapper that forwards to it,
3. declare a `can_estimate_*` class tag.

Steps 2 and 3 are mechanical, and a tag that disagrees with the implementation fails at call time rather than at definition. This PR removes both, so **adding an estimate is a single method**.

No behavior changes.

## Changes

**Capability is derived, not declared.** `can_estimate_memory_usage_static`, `can_estimate_memory_usage_static_lite` and `can_estimate_gpu_memory_usage_static` now report whether the model overrides the corresponding `_estimate_*` method:

```python
@classmethod
def can_estimate_gpu_memory_usage_static(cls) -> bool:
    return cls._overrides("_estimate_gpu_memory_usage_static")
```

The `_overrides` helper compares the underlying functions rather than the attributes — accessing a classmethod yields a fresh bound object every time, so an identity check on the attributes would always report an override. The two CPU variants also become classmethods, matching the GPU one.

**The instance wrappers become the base default.** `AbstractModel._estimate_memory_usage` and `_estimate_gpu_memory_usage` delegate to the static estimate when the model defines one, so models no longer write:

```python
def _estimate_memory_usage(self, X, **kwargs) -> int:
    return self.estimate_memory_usage_static(
        X=X, problem_type=self.problem_type, num_classes=self.num_classes,
        hyperparameters=self._get_model_params(), **kwargs)
```

Net: **25 tag declarations and 14 wrappers removed across 19 model files**; 8 `_class_tags` overrides disappear entirely (including TabICL's, called out in the review). −214 / +58 lines.

## What was deliberately left alone

- **The static/instance signature split stays.** The review also suggested merging them behind a boolean flag. They differ for a reason: the static form takes `hyperparameters` / `problem_type` / `num_classes` explicitly because it is called *before* the model exists, while the instance form reads `self`. Collapsing them would make every caller pass either self-state or explicit params, trading two clear signatures for one ambiguous one. Defaulting the wrapper gets the practical benefit (no copied boilerplate) without that.
- **Three instance wrappers are kept** — Mitra, EBM and tabprep's `PrepMixin` — because they pass extra state the default cannot supply (`y`, `features`, post-preprocessing dtype counts).

## Testing

- **Capability-identical to the parent commit for all 35 registered models** (`can_estimate_memory_usage_static` / `_gpu_` / `_lite`): zero differences, checked before and after formatting.
- `tabular/tests/unittests/test_gpu_memory_estimate.py` and `models/test_nori.py`: 10 passed, 1 skipped.
- `tabular/tests/unittests/registry` + `resource_allocation`: 298 passed.
- Lint clean with the pinned pre-commit ruff (v0.16.0): `ruff check --select I` and `ruff format`.

Pre-existing flake, unrelated to this PR: `resource_allocation/test_gpu_assignment.py::TestRayGpuAssignmentIntegration::test_ray_gpu_assignment_no_gpu_integration` fails when run after the registry suite (ray state leaking between suites) and passes in isolation — it fails identically on the parent commit.

### DIFF
--- a/tabular/src/autogluon/tabular/models/catboost/catboost_model.py
+++ b/tabular/src/autogluon/tabular/models/catboost/catboost_model.py
@@ -97,16 +97,6 @@ class CatBoostModel(AbstractModel):
                     X[col] = X[col].map(mapping).fillna(unseen_code).astype(int).astype("category")
         return X
 
-    def _estimate_memory_usage(self, X: pd.DataFrame, **kwargs) -> int:
-        hyperparameters = self._get_model_params()
-        return self.estimate_memory_usage_static(
-            X=X,
-            problem_type=self.problem_type,
-            num_classes=self.num_classes,
-            hyperparameters=hyperparameters,
-            **kwargs,
-        )
-
     @classmethod
     def _estimate_memory_usage_static(
         cls,
@@ -449,9 +439,7 @@ class CatBoostModel(AbstractModel):
 
     @classmethod
     def _class_tags(cls):
-        return {
-            "can_estimate_memory_usage_static": True,
-        }
+        return {}
 
     def _more_tags(self):
         # `can_refit_full=True` because iterations is communicated at end of `_fit`
--- a/tabular/src/autogluon/tabular/models/catboost/catboost_model.py
+++ b/tabular/src/autogluon/tabular/models/catboost/catboost_model.py
@@ -437,10 +437,6 @@ class CatBoostModel(AbstractModel):
     def supported_problem_types(cls) -> list[str] | None:
         return ["binary", "multiclass", "regression", "quantile", "softclass"]
 
-    @classmethod
-    def _class_tags(cls):
-        return {}
-
     def _more_tags(self):
         # `can_refit_full=True` because iterations is communicated at end of `_fit`
         return {"can_refit_full": True}
--- a/tabular/src/autogluon/tabular/models/knn/knn_model.py
+++ b/tabular/src/autogluon/tabular/models/knn/knn_model.py
@@ -111,16 +111,6 @@ class KNNModel(AbstractModel):
                 X=X, y=y, model_params=params, time_limit=time_limit - (time.time() - time_start)
             )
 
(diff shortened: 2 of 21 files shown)


tabular/src/autogluon/tabular/models/catboost/catboost_model.py
```diff
@@ -449,9 +439,7 @@ def supported_problem_types(cls) -> list[str] | None:
 
     @classmethod
     def _class_tags(cls):
```
LennartPurucker: Just remove empty class here at this point?

tabular/src/autogluon/tabular/models/knn/knn_model.py
```diff
@@ -308,9 +298,7 @@ def supported_problem_types(cls) -> list[str] | None:
 
     @classmethod
     def _class_tags(cls):
```
LennartPurucker: Dito

LennartPurucker (approved):
If you want to push for more simplification, please do. I think we could remove a lot of functions and also create some new general purpose functions (like parsing num_gpus under cuda) so that we can keep the code new models have to add to an absoulte minimum. I think one big Claude hunt/job could do some really good work here if we are open to it.

tabular/src/autogluon/tabular/models/catboost/catboost_model.py
Innixma: yep, also noticed this. Fixed

tabular/src/autogluon/tabular/models/knn/knn_model.py
Innixma: ditto

PR #5769, opened 2026-07-31
Allow scikit-learn v1.9

### LINKED ISSUE #5766: Upgrade AutoGluon's scikit-learn dependency to version 1.9+ across relevant modules.
Upgrade AutoGluon's scikit-learn dependency to version 1.9+ across relevant modules.  scikit-survival requires scikit-learn 1.9+ and it would be good to be able to use scikit-survival and Autogluon tabular in the same notebook.

- This proposal affects multiple modules: tabular, multimodal, and timeseries as scikit-learn is a core dependency.
- Benefits include access to new algorithms, performance improvements, and bug fixes available in the latest scikit-learn versions.
- This would require updating minimum version requirements in setup files and addressing any deprecated API usage.
- 

## References

[release notes for scikit-learn 1.9](https://scikit-learn.org/stable/whats_new/v1.9.html)

- [https://scikit-learn.org/stable/whats_new/v1.9.html](https://scikit-learn.org/stable/whats_new/v1.9.html) (Release notes for scikit-learn 1.9)
- https://scikit-learn.org/stable/auto_examples/release_highlights/plot_release_highlights_1_9_0.html (release highlights)

shchur:
*Issue #, if available:* Fixes #5766

*Description of changes:*
- Increase version range to `"scikit-learn": ">=1.4.0,<1.10.0"`
- Introduce a method `_check_targets_compat` that handles the breaking change to `_check_targets` introduced in v1.9
- Bump uv.lock

### DIFF
--- a/core/src/autogluon/core/metrics/classification_metrics.py
+++ b/core/src/autogluon/core/metrics/classification_metrics.py
@@ -4,6 +4,7 @@ from typing import Union
 import numpy as np
 import pandas as pd
 import sklearn
+from packaging.version import parse as parse_version
 from scipy.sparse import coo_matrix
 from sklearn.metrics import cohen_kappa_score
 from sklearn.utils import check_consistent_length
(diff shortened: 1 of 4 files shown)


Innixma (commented):
Need to update the uv lock file

core/src/autogluon/core/metrics/classification_metrics.py
```diff
@@ -24,8 +25,17 @@
 logger = logging.getLogger(__name__)
 
 
+def _check_targets_compat(solution, prediction):
```
shchur, line 28: @Innixma looks like sklearn is really punishing the users for touching the internal `_check_targets` method. They broke it in v1.8.0 and now again in v1.9.0 🤪

Innixma: ouch! Good catch

Innixma (approved):
LGTM, thanks!

PR #5771, opened 2026-07-31
Declare supported problem types via _supported_problem_types class attribute

Innixma:
## Description

Every `supported_problem_types()` override in the codebase was a trivial constant return, while the same classes already declare `ag_key` / `ag_name` / `ag_priority` as plain class attributes. Models now declare their supported problem types as data, alongside that metadata:

```python
_supported_problem_types = ["binary", "multiclass", "regression"]
```

one line instead of three, for 24 model classes (+32/−76 lines overall).

## API is unchanged

`supported_problem_types()` remains the public classmethod all callers use — the base implementation returns `cls._supported_problem_types`. Third-party custom models that override the classmethod keep working untouched (both through direct calls and `_get_default_ag_args()`), which is why this is not a plain attribute conversion: with a mixed ecosystem, attribute-style access on a classmethod-overriding subclass would return a truthy bound method and silently corrupt `ag_args["problem_types"]`.

## Guard for first-party models

`FitHelper.verify_model` now fails when the tested model class overrides the classmethod, with an error message showing the attribute style to use instead — so new in-repo models converge on the attribute while third-party overrides stay supported.

## Verification

- All 35 registry models return identical values via the attribute and the classmethod, and `_get_default_ag_args()` produces a correct `problem_types` list for attribute-style, legacy classmethod-override, and unspecified models.
- Model unit tests exercising `verify_model` pass (test_linear, test_dummy, test_lightgbm).
- The new guard raises on a legacy-style override and on a model with no declaration, each with the intended message.

### DIFF
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -228,6 +228,9 @@ class AbstractModel(ModelBase, Tunable):
         None  # the name of the hyperparameter that controls the model's random seed, or None if no random seed exists.
     )
     seed_name_alt: list[str] = []  # alternative names for the random seed hyperparameter.
+    _supported_problem_types: list[str] | None = (
+        None  # problem types the model supports; None = unspecified (never filtered by problem type). Read via `supported_problem_types()`.
+    )
 
     model_file_name = "model.pkl"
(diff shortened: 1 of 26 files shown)


LennartPurucker (approved):
LGTM

PR #5772, opened 2026-07-31
Declare auxiliary-param overrides via _default_auxiliary_params_extra class attribute

Innixma:
## Description

Follow-up to #5771, applying the same declare-data-not-methods pattern to `_get_default_auxiliary_params`.

Every `_get_default_auxiliary_params()` override in the codebase was a constant dict wrapped in the identical 6–14 line boilerplate:

```python
def _get_default_auxiliary_params(self) -> dict:
    default_auxiliary_params.update({...})
    return default_auxiliary_params
```

Models now declare the overrides with their other class attributes:

```python
_default_auxiliary_params_extra = {"max_rows": 50000, "max_features": 2000, ...}
```

26 model classes converted (+162/−306 lines). Inline comments inside the dicts are preserved verbatim.

## Semantics

The base `_get_default_auxiliary_params()` builds the base defaults, then merges each class's own `_default_auxiliary_params_extra` (via `__dict__`, so each class contributes exactly once) in reverse MRO order — base-most first, subclass wins. This reproduces the historical chained-`super().update()` behavior exactly, including multi-level chains (TabPFN-3 → TabPFN-2.5, FTTransformer → MultiModalPredictor).

Method overrides remain fully supported for dynamic cases and compose with declarative parents: `super()._get_default_auxiliary_params()` already includes the merged extras.

Also removes two no-op `__init__` pass-throughs (`FTTransformerModel`, mitra's internal `BaseModel`).

## Verification

- Auxiliary params for all 35 registry models plus `BaggedEnsembleModel` and `GreedyWeightedEnsembleModel` are bit-identical to a snapshot captured before the change.
- A legacy method-override subclass of a declarative parent composes correctly; a declarative child correctly overrides its declarative parent's keys.
- Model unit tests exercising the full fit path pass (test_linear, test_dummy, test_lightgbm).

### DIFF
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -231,6 +231,9 @@ class AbstractModel(ModelBase, Tunable):
     _supported_problem_types: list[str] | None = (
         None  # problem types the model supports; None = unspecified (never filtered by problem type). Read via `supported_problem_types()`.
     )
+    _default_auxiliary_params_extra: dict | None = (
+        None  # auxiliary-param overrides merged onto the base defaults (base-most class first); see `_get_default_auxiliary_params`.
+    )
 
     model_file_name = "model.pkl"
     model_info_name = "info.pkl"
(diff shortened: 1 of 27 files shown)


LennartPurucker (approved):
Love it!

Thought: could we already move towards a data class for this here, or in the future?

Innixma:
> Love it!
> 
> Thought: could we already move towards a data class for this here, or in the future?

Claude likes the idea, but suggests as a dedicated PR.

Also, might be a bit complicated, will have to see.

PR #5773, opened 2026-07-31
Validate _default_auxiliary_params_extra keys in verify_model

Innixma:
## Description

Follow-up to #5772, implementing the first step toward @LennartPurucker's dataclass suggestion: closing the typo hole in the auxiliary-param key space.

A typo'd key in `_default_auxiliary_params_extra` (e.g. `"max_rowz"`) is silently ignored at runtime. `FitHelper.verify_model` now validates that every declared key is actually consumed by something:

- **Known keys** = base auxiliary-param defaults ∪ `_ag_params_common()` ∪ the model's `_ag_params()`.
- An unknown key fails with the sorted unknown/known key lists and the fix: correct the typo, or register a wrapper-consumed private key in the model's `_ag_params()`.

This makes `_ag_params()` the single registration point for model-private auxiliary params, which is the schema a future `AuxiliaryParams` dataclass can be generated from — declarations stay ergonomic partial dicts, the merged runtime object can become typed later without touching model code again.

## Registrations the check surfaced

Three keys were consumed but registered nowhere:

- `max_batch_size` → `_ag_params_common()`: consumed by `AbstractModel` itself (chunked prediction via `_predict_proba_batch`), available to every model.
- `drop_unique` → `_ag_params_common()`: also consumed by `AbstractModel` (unique-value feature dropping).
- `model_telemetry` → `TabPFNModel._ag_params()`: its only consumer.

Both `_ag_params_common()` additions are documented in its docstring. No behavior change: consumers read these via `params_aux.get(...)` and continue to; registration only makes them visible to `_get_ag_params()` and the validator.

## Verification

- All registry models plus `BaggedEnsembleModel` / `GreedyWeightedEnsembleModel` pass the new check.
- The guard raises on a typo'd key and on an unregistered model-private key, and passes once that key is registered in `_ag_params()`.
- Model unit tests pass (test_linear, test_dummy, test_lightgbm — all routed through `verify_model`).

### DIFF
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -3556,6 +3556,11 @@ class AbstractModel(ModelBase, Tunable):
             If specified, raises an AssertionError at fit time if self.problem_type not in problem_types
         ignore_constraints: bool
             If True, ignores the values of `min_rows`, `max_rows`, `max_features`, `max_classes` and `problem_types`.
+        max_batch_size: int
+            If specified, predictions on more than `max_batch_size` rows are computed in chunks of at most
+            `max_batch_size` rows each (see `_predict_proba_batch`), bounding prediction-time memory usage.
+        drop_unique: bool
+            Whether to drop features that have only 1 unique value (default True).
 
         """
         return {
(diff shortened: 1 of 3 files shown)


shchur (approved):
LGTM

PR #5775, opened 2026-07-31
AuxiliaryParams typed schema; make params_aux immutable after construction (deprecation, raises in 1.7)

Innixma:
## Description

Follow-up to #5773 and the next step toward @LennartPurucker's dataclass suggestion. Four commits:

### 1. `AuxiliaryParams` dataclass schema

The single source of truth for the auxiliary-param key space:
- Its *materialized* fields generate the `_get_default_auxiliary_params` base defaults (replacing the 30-line literal dict; the per-key documentation moved to the schema fields).
- Its *wrapper-only* fields cover the constraint/wrapper params (`max_rows`, `max_batch_size`, `drop_unique`, ...) with the same fallback defaults their consumers used.
- `known_keys()` drives the `verify_model` declaration check from #5773.
- New `AbstractModel.aux_params` property: the typed runtime view, built from `params_aux`; non-schema keys (model-private params registered via `_ag_params()`) land in `.extra`. Base consumers converted as demonstration (`_validate_fit_args`, `drop_unique`, prediction chunking).

Declarations are unchanged: models keep partial `_default_auxiliary_params_extra` dicts, users keep `ag_args_fit` dicts — the dataclass is the *merged view*, which sidesteps the partial-declaration problem a per-class dataclass would have.

### 2–3. Remove the two in-repo `params_aux` mutations

Both were runtime state stored in resolved configuration:
- TabPFN's `max_batch_size="auto"` resolution → instance state, read through a new `_get_max_batch_size()` hook on `AbstractModel` (the documented pattern for sentinel-resolving models).
- The calibration temperature the trainer wrote into the fitted model's `params_aux` → `temperature_scalar` property backed by instance state (mirroring how conformalization already works), falling back to the user-configured `ag.temperature_scalar`.

Behavior unchanged in both: values persist through pickling, refit templates re-resolve as before, explicit user settings are honored.

### 4. Deprecate post-construction mutation — **raises starting in AutoGluon 1.7**

With the above, nothing in AutoGluon mutates `params_aux` after construction, so the invariant is now enforced for external code: `params_aux` is a `ParamsAuxDict` whose mutating operations emit a `DeprecationWarning` naming the replacement patterns. The gate is version-based (same convention as the `Deprecated` decorator): **once the installed version reaches 1.7, mutation raises a `TypeError`** — matching the behavior of the read-only mapping `params_aux` will eventually become, at which point `aux_params` can be cached and frozen.

Notes:
- `copy()` returns a plain mutable dict, so copy-and-edit code (e.g. `get_params_aux_info`) is unaffected.
- `__reduce__` rebuilds through the constructor — pickle's default dict-subclass protocol repopulates via `__setitem__` and would otherwise trip the deprecation on every model load.
- `setdefault` on a present key is a read and does not warn.

## Verification

- Merged auxiliary params for all registry models plus the ensemble models are bit-identical to a snapshot taken before any of this work.
- 13 new unit tests (`test_params_aux_mutation.py`): every mutating op warns, reads don't, copy stays plain, pickle roundtrips warning-free, the 1.7 gate raises `TypeError`.
- Calibrate + model suites pass, and run clean under `-W error::DeprecationWarning` (remaining errors there are third-party pyparsing deprecations).

### DIFF
--- a/core/src/autogluon/core/models/abstract/_auxiliary_params.py
+++ b/core/src/autogluon/core/models/abstract/_auxiliary_params.py
@@ -28,12 +28,12 @@ from packaging import version
 
 from ...version import __version__
 
-#: Field metadata marking params that are consumed by wrapper code but intentionally
-#: absent from the `_get_default_auxiliary_params` defaults dict.
 _WRAPPER_ONLY = {"materialized": False}
+"""Field metadata marking params that are consumed by wrapper code but intentionally
+absent from the `_get_default_auxiliary_params` defaults dict."""
 
-#: AutoGluon version from which mutating `params_aux` raises instead of warning.
 _MUTATION_RAISE_MIN_VERSION = "1.7"
+"""AutoGluon version from which mutating `params_aux` raises instead of warning."""
 _MUTATION_SHOULD_RAISE = version.parse(__version__) >= version.parse(_MUTATION_RAISE_MIN_VERSION)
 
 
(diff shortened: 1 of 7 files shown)


core/src/autogluon/core/models/abstract/_auxiliary_params.py
```diff
@@ -0,0 +1,194 @@
...
+
+
+@dataclass
+class AuxiliaryParams:
+    """Typed view of a model's auxiliary params. See the module docstring."""
+
+    # -- materialized defaults: seed `AbstractModel._get_default_auxiliary_params` --
+    # Ratio of memory usage allowed by the model. Values > 1.0 have an increased risk of causing OOM errors.
+    # Used in memory checks during model training to avoid OOM errors.
+    max_memory_usage_ratio: float = 1.0
```
LennartPurucker, line 114: Stupid nit, but isn't this the wrong doc string format for a class variable? 

Should it not be below it with """...."""?

LennartPurucker (approved):
Nice, the really cool thing that this will allow is that we can support passing this as input to the fit call or a model in the future!

Innixma:
Good catch — converted the schema fields (and the module-level constants) to attribute docstrings below the assignment, which doc tooling actually picks up. Verified the schema is behaviorally unchanged (base defaults and all registry models still bit-identical to the pre-change snapshot).

core/src/autogluon/core/models/abstract/_auxiliary_params.py
Innixma: updated

PR #5776, opened 2026-07-31
Make params immutable after construction (deprecation, raises in 1.7)

Innixma:
## Description

Follow-up to #5775, applying the same treatment to `params` (the model hyperparameters). Two commits:

### 1. Relocate the one real mutator

`BaggedEnsembleModel` forced `save_bag_folds=True` for children without `refit_full` support by writing into `self.params` mid-fit — the only post-construction mutation of model hyperparameters in the codebase. The forcing is now the `_save_bag_folds_forced` instance flag, and readers of the *effective* setting (fold pruning, the fold-fitting strategy's model stripping, `can_infer`, copy-based refit's assertion, and the trainer's refit-failure fallback) go through the new `save_bag_folds` property, which prefers the forced state over the configured hyperparameter. The configured value in `params` is never overwritten.

### 2. Deprecate post-construction mutation — raises starting in AutoGluon 1.7

The mutation-deprecating dict machinery from #5775 is generalized into `MutationDeprecatedDict` (version gate, mutating dunders, the `__reduce__` fix that keeps pickle/deepcopy from tripping the deprecation; `ParamsAuxDict` now only supplies its message), and `params` becomes a `ParamsDict` wrapping the final assignment at the end of `initialize()`. Its message points at the `params`-specific remedies:

- work on the copy returned by `_get_model_params()` (the documented model-author pattern — untouched),
- record hyperparameter values resolved during fit in `params_trained`,
- store other runtime values as instance state.

As with `params_aux`, mutation warns today and raises a `TypeError` once the installed version reaches **1.7**.

Design note: `params` may hold search spaces pre-HPO — no conflict, since HPO resolves into freshly constructed instances rather than mutating in place.

## Verification

- 6 new unit tests (`test_params_mutation.py`) plus the `params_aux` suite (18 total): mutating ops warn, reads/copies don't, `_get_model_params()` copies stay plain and mutable, pickle roundtrips, the 1.7 gate raises.
- dummy / lightgbm / linear / calibrate suites pass (bagged, refit, and calibration flows exercise the `save_bag_folds` property) and emit **zero** of these warnings under `-W error::DeprecationWarning`.

### DIFF
--- a/core/src/autogluon/core/models/abstract/_auxiliary_params.py
+++ b/core/src/autogluon/core/models/abstract/_auxiliary_params.py
@@ -20,36 +20,24 @@ not fields (model-private params registered via ``_ag_params()``) land in ``extr
 from __future__ import annotations
 
 import dataclasses
-import warnings
 from dataclasses import dataclass, field
 from typing import Any
 
-from packaging import version
-
-from ...version import __version__
+from ._mutation_deprecated_dict import _MUTATION_RAISE_MIN_VERSION, MutationDeprecatedDict
 
 _WRAPPER_ONLY = {"materialized": False}
 """Field metadata marking params that are consumed by wrapper code but intentionally
 absent from the `_get_default_auxiliary_params` defaults dict."""
 
-_MUTATION_RAISE_MIN_VERSION = "1.7"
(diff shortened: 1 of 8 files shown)


LennartPurucker (approved):
LGTM

PR #5777, opened 2026-07-31
Declare ensemble-args overrides and fit-resource defaults as class attributes

Innixma:
## Description

Three declare-data-not-methods conversions in the model layer, following #5771/#5772.

### `_default_ag_args_ensemble_extra`

Every `_get_default_ag_args_ensemble` override except `RFModel`'s was a constant dict wrapped in the `super() → update() → return` boilerplate. Models now declare:

```python
_default_ag_args_ensemble_extra = {
    "fold_fitting_strategy": "sequential_local",
    "refit_folds": True,  # Better to refit the model for faster inference and similar quality as the bag.
}
```

with their other class attributes (10 models converted; inline comments and method docstrings preserved as attribute docstrings). The base classmethod merges each class's declaration in reverse MRO order — same semantics and machinery as `_default_auxiliary_params_extra`. `RFModel` keeps its method override (it branches on `problem_type`), which remains fully supported.

Each foundation model keeps its own declaration rather than inheriting a shared default — the payloads are per-model decisions, even where they currently coincide.

### `minimum_num_gpus` / `gpu_required`

11 `get_minimum_resources` overrides were three constant shapes. They reduce to two class attributes read by the base method:

- `minimum_num_gpus: float = 0` — the smallest GPU share the model fits with when a GPU is available (`0.5` = can share a GPU with another job).
- `gpu_required: bool = False` — the minimum applies even without a detected GPU, for GPU-mandatory models (RAPIDS mixin, MultiModalPredictor; `FTTransformerModel` opts back out of its parent's requirement).

Shape note: models that previously returned an explicit `"num_gpus": 0` now omit the key, matching the method's contract ("if a resource is not part of the output, it is considered unnecessary") and the majority shape.

### `default_resources_physical_cores_only` / `default_num_gpus`

`_get_default_resources` had 16 overrides that were two constant bodies copy-pasted (physical cores + 0 GPUs; physical cores + up to 1 CUDA GPU, with the same benchmark comment). Models now declare the two attributes and the base assembles the tuple. TabPFN's multi-GPU default is `default_num_gpus = max_gpus`; KNN's override was identical to base behavior and is deleted. MultiModalPredictor and the RAPIDS mixin keep their method overrides — they count non-CUDA GPUs (`get_gpu_count_torch()` without `cuda_only=True`), which the attribute-driven base deliberately does not model.

## Verification

- `_get_default_ag_args_ensemble()` (± `problem_type`), `get_minimum_resources()` (± GPU), and `_get_default_resources()` for all 35 registry models are effectively identical to snapshots captured before the change (`_get_default_resources` verified on an 8-GPU machine, so the GPU axis discriminates 0 / 1 / 8; the physical-cores attribute mapping was additionally checked per class against the pre-change inventory).
- Model unit tests pass (test_dummy, test_lightgbm, test_linear).

### DIFF
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -239,6 +239,12 @@ class AbstractModel(ModelBase, Tunable):
     _default_ag_args_ensemble_extra: dict | None = (
         None  # ag_args_ensemble overrides merged base-most class first; see `_get_default_ag_args_ensemble`.
     )
+    default_resources_physical_cores_only: bool = False
+    """Whether `_get_default_resources` counts only physical CPU cores (faster for
+    training-bound models based on benchmarks) instead of logical cores."""
+    default_num_gpus: int = 0
+    """num_gpus in `_get_default_resources`: use up to this many CUDA GPUs by default when
+    available (0 = train on CPU by default)."""
     minimum_num_gpus: float = 0
     """num_gpus in `get_minimum_resources` when a GPU is available: the smallest GPU share the
     model can fit with (fractional = the model can share a GPU with other jobs; 0 = fits on CPU)."""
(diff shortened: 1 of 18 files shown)


LennartPurucker (approved):
Nice!

PR #5778, opened 2026-07-31
Share the categorical label-encoding preprocess step across models

Innixma:
## Description

Six models — RF, Nori, TabDPT, TabPFN, TabPFNMix, Mitra — carried the identical stateful label-encoding block in `_preprocess`. It is now a base-class helper:

```python
X = self._label_encode_categoricals(X)                    # lazy: fit on first call
X = self._label_encode_categoricals(X, is_train=is_train) # refit at fit time (Mitra/TabPFN variant)
```

- The encoder lives on `self._label_encoder` (a specific name replacing the generic per-model `_feature_generator`), initialized by the base constructor; the six per-model `__init__` assignments (and a dead one in iModels) are removed — Nori's `__init__` becomes empty and is deleted.
- Per-model output handling (float32 cast, `fillna(0)`, `.to_numpy()`) stays in each `_preprocess`.
- TabPFN's cat-feature/index bookkeeping is restructured around the helper call with identical results.
- This block was already spreading by copy-paste (TabPFNMix's carried a "copy pasted from TabPFNModel, never tried removing" FIXME).

The companion `_fit` device-resolution helper is #5780.

## Verification

- Helper output frame-identical to the original block on mixed categorical/NaN data; lazy vs `is_train`-refit semantics match both prior variants; input never mutated; numeric-only frames pass through uncopied.
- TabPFN's `_cat_features` / `_cat_indices` identical through the rewritten `_preprocess`.
- RF (full fits with categoricals through the helper) + dummy unit tests pass; the one RF failure (`test_rf_binary_compile_onnx_as_ag_arg`) fails identically without this change (environment).

### DIFF
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -296,6 +296,8 @@ class AbstractModel(ModelBase, Tunable):
         # (see `temperature_scalar`); runtime state, distinct from the user-configured
         # `ag.temperature_scalar` it falls back to.
         self._temperature_scalar: float | None = None
+        # stateful categorical label encoder fitted by `_label_encode_categoricals`
+        self._label_encoder = None
         self.label_cleaner: LabelCleaner | None = None
 
         if eval_metric is not None:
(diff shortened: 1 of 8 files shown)


LennartPurucker (approved):
LGTM

PR #5780, opened 2026-07-31
Share the _fit device-resolution block on AbstractTorchModel

Innixma:
## Description

Companion to #5778 (this commit was pushed to that branch after it merged, so it never landed — reopened here on a clean branch).

Six torch models — TabM, TabICL, Nori, TabDPT, RealMLP, TabPFN — carried the same `_fit` block: derive `"cuda"`/`"cpu"` from the allocated `num_gpus` and raise when a GPU was allocated but CUDA is unavailable. Now:

```python
device = self._resolve_fit_device(num_gpus=num_gpus)                       # TabM/TabICL/Nori/TabDPT
device = self._resolve_fit_device(num_gpus=num_gpus, gpu_device="cuda:0")  # RealMLP (single-visible-GPU assumption)
```

- The helper also logs the `gpu_strongly_recommended` CPU-fallback warning — a no-op for models without the flag (TabM/RealMLP, which previously skipped the warning call), and free coverage if they're ever flagged.
- TabPFN calls the helper for the warning + availability check and keeps `_get_tabpfn_device` for its multi-GPU device list.
- Mitra is untouched: its device flows through hyperparameters with separate suggestion logic.
- TabM's `# FIXME: warn instead and switch to CPU` moves into the helper as a general TODO — it applies to every model equally.

## Verification

- Helper returns cpu / cuda / cuda:0 for `num_gpus` 0 / 1 / fractional exactly as the original conditionals did (verified on a CUDA machine, including the `!= 0` vs `== 0` inversions).
- TabM and RealMLP fit tests pass end to end; orphaned `from torch.cuda import is_available` imports removed (remaining try-block F401s are pre-existing availability probes).

### DIFF
--- a/tabular/src/autogluon/tabular/models/abstract/abstract_torch_model.py
+++ b/tabular/src/autogluon/tabular/models/abstract/abstract_torch_model.py
@@ -28,6 +28,27 @@ class AbstractTorchModel(AbstractModel):
         self.device = None
         self.device_train = None
 
+    def _resolve_fit_device(self, num_gpus: int | float, gpu_device: str = "cuda") -> str:
+        """Resolve the torch device for `_fit` from the allocated `num_gpus`.
+
+        Logs the CPU-fallback warning (see `_log_cpu_fallback_warning`) and raises when a
+        GPU was allocated but CUDA is unavailable. `gpu_device` is the device string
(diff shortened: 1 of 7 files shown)


LennartPurucker (approved):
LGTM!

shchur (approved):

PR #5782, opened 2026-07-31
[timeseries] Add MAEB and WAPEB bias-penalized metrics to timeseries

shchur:
*Description of changes:*

Adds three forecast-bias metrics to `autogluon.timeseries`:

- **MAEB** (mean absolute error with bias): `MAE + |mean(f − y)|`. Scale-dependent; equals `MAE` when the forecast is unbiased.
- **WAPEB** (weighted absolute percentage error with bias): `(Σ|f − y| + |Σ(f − y)|) / Σ|y|`. Scale-free counterpart of MAEB (mirrors how `WAPE` relates to `MAE`); equals `WAPE` when unbiased. This is the metric used to rank submissions in the [VN1 Forecasting Accuracy Challenge](https://www.datasource.ai/en/home/data-science-competitions-for-startups/vn1-forecasting-accuracy-challenge-phase-1/description).
- **BIAS** (signed mean forecast bias, `mean(f − y)`): a diagnostic reported by `predictor.evaluate()`. Since its optimum is 0 and both signs indicate a worse forecast, it cannot be used as an `eval_metric` for model selection.

The bias term in MAEB/WAPEB discourages models that achieve low error by consistently under-forecasting — a common failure mode when optimizing `MAE`/`WAPE` on intermittent (sparse) time series. For that reason MAEB/WAPEB target a forecast between the median and the mean rather than the median.

All three are documented in the forecasting metrics tutorial and support `horizon_weight`. Added tests covering the unbiased-equals-base-metric property, the sign of BIAS, and that evaluate-only metrics are rejected as `eval_metric`.


Innixma (approved):
LGTM!

PR #5784, opened 2026-08-01
Support xgboost v3.3: shared fit/predict category space for enable_categorical

Innixma:
## Description

Follow-up to #5783, which capped xgboost at `<3.3` because 3.3.0 crashed `XGBoostModel`'s `enable_categorical=True` path (used by the 2025 zeroshot portfolio configs). This fixes the path and raises the cap to `<3.4`.

**Note: stacked on #5783** — merge that first; this diff includes it until then.

### The fix

xgboost 3.3 records the fit-time categories of each categorical column and rejects predict-time values outside them (probed: declared-but-unobserved levels and NaN are fine; only values absent from the fit frame's *declared* category list raise). The `enable_categorical` ordinal encoding built each frame's category set from the values that frame happened to observe, so a predict frame containing a level unobserved at fit crashed with `Found a category not in the training set`.

The encoding now declares the full code range of the fit-time category mapping on every frame — fit and predict share one category space:

```python
codes = X[col].astype(object).map(mapping).fillna(-1).astype(int)
X[col] = pd.Categorical(codes, categories=range(len(mapping)))
```

Categories unseen at fit and missing values become NaN (missing), matching the previous effective behavior.

### Version handling

- Cap raised to `<3.4`; `uv.lock` forks xgboost by Python version (3.2.0 for `<3.12`, 3.3.0 for `>=3.12`) since 3.3.0 requires Python ≥3.12.
- The now-default `enable_categorical` in 3.3 doesn't affect AutoGluon's default path (OHE — xgboost never sees category dtypes there); AutoGluon's own `enable_categorical` hyperparameter keeps controlling the behavior explicitly.

## Verification

- **No behavior change on ≤3.2**: the new encoding is bit-identical to the old one on xgboost 3.2.0 across four small TabArena datasets (blood-transfusion, credit-g, anneal, QSAR_fish_toxicity) in both `enable_categorical` modes (8/8 cells).
- **3.3.0 works**: all 8 cells run, including the previously-crashing anneal case. Comparing 3.2.0 vs 3.3.0 with identical code in the same Python 3.12 environment: blood-transfusion and anneal are bit-identical; QSAR and credit-g drift slightly (~0.02 log-loss/RMSE in a no-early-stopping regime) — QSAR is numeric-only, so this is upstream 3.3 core-algorithm drift, unrelated to categoricals or this change.
- **New unit test** (`test_xgboost_enable_categorical_predict_time_categories`) covers the predict-time cases that previously had no coverage: declared-but-unobserved level, NaN, and a novel category (treated as missing). Tests pass on 3.2.0 (py3.11) and 3.3.0 (py3.12).

### DIFF
--- a/tabular/setup.py
+++ b/tabular/setup.py
@@ -39,7 +39,7 @@ extras_require = {
         "catboost>=1.2,<1.3",
     ],
     "xgboost": [
-        "xgboost>=2.0,<3.3",  # <{N+1} upper cap, where N is the latest released minor version
+        "xgboost>=2.0,<3.4",  # <{N+1} upper cap, where N is the latest released minor version
     ],
     "realmlp": [
         "pytabkit>=1.7.2,<1.8",
--- a/tabular/src/autogluon/tabular/models/xgboost/xgboost_model.py
+++ b/tabular/src/autogluon/tabular/models/xgboost/xgboost_model.py
@@ -5,7 +5,6 @@ import math
 import os
 import time
 
-import numpy as np
 import pandas as pd
 
 from autogluon.common.features.types import R_BOOL, R_CATEGORY, R_FLOAT, R_INT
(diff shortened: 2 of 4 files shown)


LennartPurucker (approved):
LGTM, so this should only change metadata not the data itself, right?

Innixma:
> LGTM, so this should only change metadata not the data itself, right?

Yep

PR #5788, opened 2026-08-03
Allow tabpfn v8.1-8.2 and tabicl v2.1

Innixma:
## Summary

Bumps `tabpfn` to `<8.3` and `tabicl` to `<2.2`, following the `<{N+1}` upper-cap convention (latest released minors are 8.2 and 2.1).

`fastai` is deliberately **not** bumped — see below.

## Testing

Installed tabpfn 8.2.0 and tabicl 2.1.1, then ran the affected model tests: `test_tabicl`, `test_tabpfn3`, and `test_tabular_nn_fastai` — **15 passed, 1 skipped**.

`test_tabpfnv2` fails locally, but identically on the currently-supported tabpfn 8.0.8, so it is a pre-existing environment failure (a CUDA-hidden subprocess check) rather than a regression from this bump. Verified by downgrading and re-running the same test.

## Lock changes

`uv.lock` gains tabpfn 8.2.0 / tabicl 2.1.1 and *loses* four packages: tabpfn 8.2.0 drops its `tabpfn-common-utils[telemetry-interactive]` dependency, which was pulling in `backoff`, `distro`, `nvidia-ml-py` and `posthog`.

## Why fastai stays at `<2.8.8`

fastai 2.8.8 requires `fastcore>=1.14.6`, but the highest released fastcore 1.x is **1.14.5** — so 2.8.8 is unreachable under the existing `fastcore<2` cap. That cap is load-bearing: fastcore v2 removed `L.starmap`, which the fastai models use.

Supporting fastai 2.8.8 is therefore a code change rather than a version bump — the `L.starmap` usage has to be replaced before `fastcore<2` can be lifted. Worth noting that a plain `pip install -U fastai` silently upgrades fastcore to 2.x to satisfy the requirement, so this is easy to break accidentally in a dev environment.

The reasoning is recorded as a comment next to the pin so the next person does not retry the bump blind.

### DIFF
--- a/tabular/setup.py
+++ b/tabular/setup.py
@@ -50,6 +50,9 @@ extras_require = {
     "fastai": [
         "spacy<3.9",
         "torch",  # version range defined in `core/_setup_utils.py`
+        # Held below 2.8.8: that release requires fastcore>=1.14.6, but the highest fastcore 1.x
+        # is 1.14.5, so it is unreachable under the fastcore<2 cap below. Lifting that cap needs
+        # the removed L.starmap usage replaced first.
         "fastai>=2.3.1,<2.8.8",  # Cap for major version
         "fastcore<2",  # Breaking change in v2: removed L.starmap, which breaks fastai models
     ],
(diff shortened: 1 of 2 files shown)


shchur (approved):
LGTM

PR #5789, opened 2026-08-03
Express the auto-selected validation method as size curves

Innixma:
## Summary

The validation method AutoGluon picks when the user specifies nothing was encoded as arithmetic — `min(8, max(5, rows / 10))` for folds, a bare `1` for repeats, a `>= 1e6` comparison for the extra holdout, and a compound condition with an embedded 750-row threshold for stack depth.

Those thresholds are real policy decisions, but they were buried in expressions a reader has to re-derive, and retuning any of them per size regime meant editing the module. Each knob is now a **size curve**: either a fixed value or `[[rows, value], ..., fallback]`, read as "use `value` at or below `rows`".

```python
{"num_bag_sets": [[2_000, 5], 1]}                  # repeated CV on small data
{"holdout_frac": [[1_000, 0.2], 0.1]}              # hold out 20% on small data, 10% above
{"use_bag_holdout": [[10_000, False], True]}       # move the CV/holdout switch
{"num_stack_levels": [[749, 0], [100_000, 1], 2]}  # a second layer on large data
```

This is the same declarative shape as the count-based fit callbacks. `_get_validation_preset` takes an optional `validation_size_curves` override, so one knob can be retuned without restating the rest — `num_bag_sets` in particular could not previously express anything but `1`.

## Usage

What is reachable from `fit()` today is the group-based sizing, via a key of `validation_structure`:

```python
import numpy as np
import pandas as pd

# 900 rows, but only 60 independent patients: large by rows, small by groups.
rng = np.random.default_rng(0)
n = 900
train_data = pd.DataFrame({
    "feature": rng.normal(size=n),
    "patient_id": np.repeat(np.arange(60), 15),
    "outcome": rng.integers(0, 2, n),
})

predictor = TabularPredictor(label="outcome").fit(
    train_data,
    auto_stack=True,  # so the auto-selected validation method is used
    validation_structure={
        "group_on": "patient_id",            # group-disjoint folds
        "size_validation_on_groups": True,   # ...and size the method on 60 groups, not 900 rows
    },
)
```

That fit uses **6 folds and no stack layer** (sized on 60 groups). Leaving
`size_validation_on_groups` at its default `False` gives **8 folds and a stack layer** (sized on
900 rows), which is today's behavior.

Retuning how the method scales with size is a `validation_size_curves` kwarg (marked **[EXPERIMENTAL]**: the curve format and the set of tunable knobs may still change):

```python
from autogluon.tabular.configs.pipeline_presets import ValidationSizeCurves

predictor = TabularPredictor(label="outcome").fit(
    train_data,
    auto_stack=True,
    # repeated CV at or below 2000 rows; a plain dict is accepted too
    validation_size_curves=ValidationSizeCurves(num_bag_sets=[[2_000, 5], 1]),
)
```

Unset knobs keep their defaults, and unknown knob names are rejected rather than silently ignored.

Only the entries given are overridden, and an explicitly passed `num_bag_folds` /
`num_bag_sets` / `use_bag_holdout` / `num_stack_levels` still wins over any curve. The two
compose: with `size_validation_on_groups` the curves are read at the group count, so
`validation_size_curves={"num_bag_folds": [[100, 5], 8], "num_bag_sets": [[100, 5], 5]}` on the
60-group frame above gives 5x5 cross-validation.

## How this composes with structure-aware clamping

The curves are **policy**: what to ask for at a given data size. `ValidationStructure.custom_splits` is **feasibility**: it reduces the requested fold and repeat counts where the data cannot support them (fewer groups than folds, a stratification value rarer than the fold count, a deterministic temporal partition), never raises them, and returns what it used for the caller to adopt.

So a caller may treat its request as an upper bound, and neither half needs to know the other's rules. Documented on both sides.

## Behavior is unchanged

`DEFAULT_VALIDATION_SIZE_CURVES` reproduce the old numbers exactly. Asserted against the replaced expressions:

- folds / repeats / holdout: every size from 1 row to past the 1M threshold;
- stack depth: all **324** combinations of rows × `auto_stack` × `dynamic_stacking` × `use_bag_holdout` × problem type, since the row threshold gates only part of that condition and dynamic stacking is exempt from it.

Whether repeats actually help on small data is a separate question this makes answerable by experiment rather than by editing code.

## Optional group-based sizing

Rows and group counts can imply different regimes. A benchmark task with 4,672 rows across 68 groups is large by rows and small by groups, and the grouped protocol treats it as small — 5x5 CV where row-based sizing gives 8x1.

`validation_structure={"group_on": ..., "size_validation_on_groups": True}` reads the curves at the group count instead. It is **off by default**: a declared grouping alone changes nothing, because when rows within a group are not independent the group count is often the sample size that matters, but whether that applies is a judgement about the data rather than something to infer from the presence of a grouping. The flag is a key of `validation_structure` so it is validated and documented with the columns beside it.

`ValidationStructure` exposes `num_group_instances(X)` (None without a grouping) and `TabularPredictor.fit` passes it through; the structure is now resolved before the validation method is chosen, since the method can depend on it. `holdout_frac` stays sized on rows either way, being a fraction of the rows actually held out.

End to end on 900 rows across 60 groups with `auto_stack=True`: rows give 8 folds and a stack layer, groups give 6 folds and none.

## Also

A temporal structure silently corrected requested repeats to 1 (the partition is deterministic, so the resolver returns 1 and the caller adopts it). Correct, but invisible — now logged, with a test pinning the correction end to end: a fit requesting 3x5 on `time_on` must produce 3 child models, not 15, which only works if the reduced count reaches the trainer.

## Testing

`tabular/tests/unittests/configs/test_validation_size_curves.py` (new) plus additions to `test_validation_structure.py` — 140 tests pass. Coverage: curve resolution and malformed-curve rejection, equivalence with the replaced arithmetic, each knob's override, opt-in group sizing including that a known group count alone changes nothing, and the temporal repeat collapse.

### DIFF
--- a/common/src/autogluon/common/utils/validation_structure.py
+++ b/common/src/autogluon/common/utils/validation_structure.py
@@ -142,6 +142,11 @@ class ValidationStructure:
         channel) are what make repeated grouped cross-validation possible: the ``groups``
         channel forces leave-one-group-out with ``n_repeats == 1``.
 
+        These clamps are the *feasibility* half of choosing a validation method: the caller's
+        requested counts are the policy (see AutoGluon's validation size curves), and this
+        reduces them only where the data cannot support what was asked. It never raises them,
+        so a caller may treat the request as an upper bound.
+
         Clamping rules, each of which also forces ``num_repeats = 1`` because the resulting
         partition is deterministic and repeating it would only duplicate work:
 
--- a/common/src/autogluon/common/utils/validation_structure.py
+++ b/common/src/autogluon/common/utils/validation_structure.py
@@ -33,6 +33,13 @@ class ValidationStructure:
         to the label for classification when combined with ``group_on``; for time-based
         splits it only influences how the contiguous blocks are assigned to fold indices,
         never the block boundaries themselves.
+    size_validation_on_groups : bool, default False
+        If True, the automatically selected validation method (fold count, repeats, extra
+        holdout, stack depth) is chosen from the number of groups rather than the number of
+        rows. Rows within a group are not independent, so the group count is often the sample
+        size that matters -- a task with 4,672 rows across 68 groups is large by rows and small
+        by groups -- but which applies is a judgement about the data, so this is opt-in. Has no
+        effect unless ``group_on`` or ``group_time_on`` is set.
     group_time_on : str, optional
         Column identifying groups that are *also* ordered in time (e.g. a session id whose
         sessions arrive in sequence). Whole groups are blocked in time order, so the splits
@@ -45,6 +52,7 @@ class ValidationStructure:
     time_on: str | None = None
     stratify_on: str | None = None
     group_time_on: str | None = None
+    size_validation_on_groups: bool = False
 
     def __post_init__(self):
         if self.group_on is not None and self.time_on is not None:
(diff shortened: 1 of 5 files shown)


LennartPurucker (changes requested):
Mhm, can we make this a data class right away? While it does not need to be used as one, I think we can make it more explicilt now

LennartPurucker (approved):
LGTM

PR #5791, opened 2026-08-03
Account for GPU memory when budgeting parallel bagging folds

Innixma:
## Summary

Parallel folds share one device, so co-scheduling them multiplies VRAM use — but the fold budget never saw it. `folds_to_fit_in_parallel_with_mem` compares the model's memory estimate against **host** memory, and only a fraction of the GPU is reserved per fold, so on a RAM-rich node every fold could land on one card and exhaust it.

That surfaced as a CUDA OOM mid-fit rather than as a scheduling decision. In practice the workaround has been to misreport the memory limit *as* the card's VRAM (`memory_limit`/`fake_memory_for_estimates` set to the GPU's capacity) so the existing host-memory arithmetic happens to land on a safe fold count — which also makes the real host budget wrong, and silently does nothing for models whose estimate is not VRAM-shaped.

## What changed

The pieces already existed and were simply unconsumed by the fold-fitting strategy: `AbstractModel.estimate_gpu_memory_usage` / `can_estimate_gpu_memory_usage_static`, and `ResourceManager.get_available_vram`.

1. **A VRAM cap on parallel folds**: free VRAM divided by the per-fold estimate, with the same `max_memory_usage_ratio` applied, and the **stricter** of that and the existing host-memory cap wins.
2. **A skip when even one fold cannot fit.** The cap bottoming out at 1 is ambiguous on its own — it means either "run folds one at a time" or "not even one fold fits". The host-memory path already distinguishes these by calling the model's memory check at its floor; this mirrors that with `_validate_fit_gpu_memory_usage`, the documented GPU counterpart, which raises `NotEnoughCudaMemoryError` and lets the trainer skip the model gracefully instead of reaching a CUDA OOM.

## Multi-GPU: capacity and placement

Review of the first commit surfaced that it was single-device-minded: the cap read device 0 only (a quarter of a 4-GPU node's real capacity), and fold placement was round-robin with fractional requests truncated to zero.

- **Capacity** now sums per-device: `sum(device_vram / per_fold_estimate)` over the devices the bag may use — per device, because with mixed cards the smallest bounds what a fold placed there can use.
- **Placement** goes through a packing planner (`plan_gpu_assignments`). Devices are selected on two priorities: progress toward **1.0 utilization** first — the device whose assigned GPU fraction is already highest and still fits the request is filled before the next is touched — then **VRAM headroom** as the tiebreak, which is spent by the per-fold estimate as requests are placed.

  Requests `[0.5, 1, 0.25, 0.5]` on 4 GPUs go to devices `0, 1, 0, 2`, leaving utilizations `0.75 / 1.0 / 0.5 / 0`.

- The planner accepts an arbitrary request sequence (not just the homogeneous fractions one bagged model produces), and a request that fits nowhere falls back to the least-utilized device rather than failing.

## Inert wherever VRAM cannot constrain the decision

GPU memory estimates are **opt-in per model**, so the overwhelming majority of fits must behave exactly as before. Each of these returns infinity, leaving the host-memory budget to decide alone:

- CPU-only fits (`num_gpus == 0`);
- models that do not implement `_estimate_gpu_memory_usage_static` (most of them);
- an estimate that comes back unusable;
- free VRAM that cannot be read.

## No double raise

`AbstractModel.fit` also runs `_validate_fit_gpu_memory_usage`, but per **child fold**, after scheduling, against the fractional `num_gpus` that fold was assigned. On the error path the budget-time raise aborts before any child is fit, so the failure is reported once and earlier; on the passing path both checks simply pass.

This is also why the existing per-child check never caught the original problem: it compares one fold's estimate against *total* free VRAM, so N folds each individually "fit" while collectively exhausting the card. The budget-time check is what accounts for the co-scheduling.

## Testing

`core/tests/unittests/models/test_parallel_fold_vram.py` (new, 8 tests): the cap arithmetic (12 GB/fold on 96 GB free → 8 folds; ratio 0.8 → 6; 40 GB/fold → 2), each of the four inert paths, and the rejection of a model larger than the card — including that the check receives the single-fold estimate and the bag's GPU count rather than a fold fraction.

Regression-checked against the dummy-model and validation-structure suites (37 tests).

## Follow-up

`fake_memory_for_estimates` is not obsolete: it remains the only lever for models where `can_estimate_gpu_memory_usage_static()` is False. Adding `_estimate_gpu_memory_usage_static` to the GPU foundation models that lack one would let them benefit from this automatically.

### DIFF
--- a/core/src/autogluon/core/models/ensemble/fold_fitting_strategy.py
+++ b/core/src/autogluon/core/models/ensemble/fold_fitting_strategy.py
@@ -438,6 +438,69 @@ class SequentialLocalFoldFittingStrategy(FoldFittingStrategy):
         return fold_model
 
 
+def plan_gpu_assignments(
+    requests: List[Union[int, float]],
+    total_gpus: int,
+    per_device_vram: Optional[List[float]] = None,
+    vram_est_per_task: Optional[float] = None,
+) -> List[List[int]]:
+    """Assign each GPU request to device(s), packing devices toward full utilization.
+
+    Requests are processed in order; each holds its assignment for as long as it runs, so
+    the plan describes concurrent load. A request is a GPU fraction (< 1, one device) or a
+    whole-device count (int >= 1). Device selection sorts candidates on two priorities:
+
+    1. progress toward 1.0 utilization -- the device whose assigned fraction is already
+       highest (and still fits this request) is filled first, so one device reaches exactly
+       1.0 before the next is touched. E.g. requests ``[0.5, 1, 0.25, 0.5]`` on 4 GPUs go to
+       devices ``0, 1, 0, 2``, leaving utilizations ``0.75, 1.0, 0.5, 0``.
+    2. available VRAM -- among equally-utilized devices, the most VRAM headroom, which is
+       spent by ``vram_est_per_task`` as requests are placed (when both are known).
+
+    A request that fits nowhere by these constraints falls back to the least-utilized
+    device(s): the fraction/VRAM budgets upstream should prevent that, but assignment must
+    never fail where round-robin would have produced something.
+    """
+    fraction_used = [0.0] * total_gpus
+    vram_used = [0.0] * total_gpus
+
+    def headroom(device: int) -> float:
+        if not per_device_vram or device >= len(per_device_vram):
+            return 0.0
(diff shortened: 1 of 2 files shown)


LennartPurucker (approved):
LGTM

PR #5796, opened 2026-08-04
Add forward-chaining temporal validation

Innixma:
## Summary

Temporal validation is currently leave-one-block-out: each contiguous time block is validated once and trained on for every *other* fold. That covers every row, but most folds train partly on data from **after** the window they are scored on. On a 1173-row temporal task (`garments_worker_productivity`, `time_on='date'`):

| fold | validation window | train rows *after* the val window |
|---|---|---|
| 0 | Feb 16 – Mar 10 | 0 |
| 1 | Jan 1 – Jan 22 | 780 of 780 |
| 2 | Jan 24 – Feb 15 | 388 |

At 8 folds, 7 of 8 folds train on the future.

This adds `ValidationStructure.temporal_forward_only`: an expanding window where fold *i* validates block *i+1* and trains only on earlier blocks, so no fold is ever trained on data from after its validation window. Same task, same requested fold counts:

| | default | `temporal_forward_only=True` |
|---|---|---|
| train rows after the val window | 780 / 388 / 0 | **0 on every fold** |
| train sizes (3 folds) | 785 / 780 / 781 | 299 → 580 → 884 |
| rows with out-of-fold predictions | 1173 / 1173 | 874 / 1173 |

The cost is inherent to forward-chaining: the earliest block is training data only and is never validated, so those rows have no out-of-fold prediction. `uncovered_rows()` reports them. Folds also train on less data, and only the last sees nearly all of it. Requires at least 3 blocks (2 folds plus the initial training block).

## Making partial coverage honest

Supporting this exposed a pre-existing hazard. A bagged model's out-of-fold accumulator starts at zero, and `_predict_proba_oof` divided rows with no fold prediction by a substituted 1 — so an unvalidated row emerged as a confident prediction of **0**. Validation *scoring* already excluded such rows (`score_with_oof` masks on `_oof_pred_model_repeats > 0`), but the weighted ensemble did not: it optimized weights against those rows and was scored on them.

The result was two leaderboard numbers computed over different row sets and then compared to pick `model_best`. Measured with 61 of 300 rows unvalidated:

| | score_val |
|---|---|
| `LightGBM_BAG_L1` (reported, masked to its 239 covered rows) | −1.318491 |
| `WeightedEnsemble_L2` of that one model at weight 1.0 (before) | **−1.693790** |
| that same model's OOF scored over all 300 rows, zeros included | −1.693790 |

An ensemble that is one model at weight 1.0 must score identically to that model. It did not, and the discrepancy was exactly the contamination.

Two changes fix it:

- `BaggedEnsembleModel._predict_proba_oof` emits **NaN** for rows no fold validated, so "missing" is representable instead of being spelled `0`. NaN rather than dropping rows keeps positional row alignment intact — `custom_splits` are positional indices into the full frame, so dropping upstream would misalign them.
- `stack_new_level_aux` drops rows whose base-model out-of-fold predictions are NaN before fitting the weighted ensemble, so it optimizes and is scored on the same rows the base models are scored on. The single-model ensemble now reports −1.318491, matching its base exactly.

**Full-coverage runs are unaffected**: no row is uncovered, so no NaN is ever produced and the arithmetic is unchanged. Partial coverage was already reachable before this PR via incremental fold fitting (`k_fold_start` / `k_fold_end`), where the same fabricated zeros applied.

A side benefit: feeding NaN to `EnsembleSelection` now raises instead of silently optimizing against fabricated zeros, so this class of bug cannot recur quietly.

## Not supported: stacking above layer 1

`num_stack_levels > 0` and `dynamic_stacking` raise when combined with `temporal_forward_only`. A higher layer would train on rows with no real lower-layer features, and its own validation score would cover rows it has no features for. Supporting it means teaching the trainer to drop uncovered rows per level — deliberately left out of this PR, as it changes the shared stacking path for every configuration.

## Testing

`common` / `core`:
- out-of-fold predictions are NaN exactly for rows no fold validated, and scoring still succeeds
- no NaN when every row is validated (the ordinary case)

`ValidationStructure` (`tabular/tests/unittests/test_validation_structure.py`):
- every fold's training rows precede its validation window; train sizes grow; validation windows advance
- the earliest block is unvalidated and `uncovered_rows` agrees with the splits
- the default mode still covers every row *and* exhibits the lookahead (the trade-off pinned from both sides)
- `group_time_on` forward-chains over whole groups while staying group-disjoint
- `temporal_forward_only` without a time column raises; fewer than 3 blocks raises

End-to-end through `TabularPredictor`:
- weighted ensemble score equals its single base model's score, and out-of-fold predictions contain NaN
- stacking alongside forward-chaining raises

Suites run: 6 core bagged-ensemble tests, 104 tabular validation tests (`test_validation_structure.py` + `configs/test_validation_size_curves.py`), 15 `test_model_random_seed.py` tests (the existing suite exercising bagging/OOF).

## Also included

A `.gitignore` entry for the `datasets/` directories the tabular tests write fitted predictors into. Their location follows pytest's working directory, so they appear as hundreds of untracked pickles at the repo root or under `tabular/`; only `tabular/tests/unittests/datasets` was ignored. Unanchored, since no tracked file lives under a `datasets` directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -135,3 +135,8 @@ multimodal/src/autogluon/multimodal/version.py
 
 AutogluonModels
 lightning_logs
+
+# Datasets and fitted predictors written by the test suite. Left unanchored because the location
+# depends on the directory pytest was run from (repo root, `tabular/`, ...), and no tracked file
+# lives under a `datasets` directory.
+datasets/
--- a/common/src/autogluon/common/utils/validation_structure.py
+++ b/common/src/autogluon/common/utils/validation_structure.py
@@ -46,6 +46,22 @@ class ValidationStructure:
         are simultaneously group-disjoint and forward in time, and the non-bagged holdout is
         the latest groups. Use this for data that is both grouped and temporal; combining
         ``group_on`` with ``time_on`` is not supported, as their semantics would be ambiguous.
+    temporal_forward_only : bool, default False
+        Restrict temporal validation to forward-chaining (an expanding window): fold *i*
+        validates time block *i+1* and trains only on the blocks before it, so a model is never
+        trained on data from after the window it is scored on.
+
+        The default (False) is leave-one-block-out: every block is validated exactly once and
+        trained on for every other fold, which covers all rows but trains most folds partly on
+        the future. Forward-chaining removes that lookahead at two costs:
+
+        1. **The earliest block is never validated.** It is training data only, so those rows
+           get no out-of-fold prediction. Callers that consume out-of-fold predictions must
+           handle the gap -- see :meth:`uncovered_rows`.
+        2. **Folds train on less data.** Fold 0 sees one block; only the last sees nearly all.
+
+        Requires at least 3 blocks (2 folds plus the initial training block), and has no effect
+        unless ``time_on`` or ``group_time_on`` is set.
     """
 
     group_on: str | list[str] | None = None
@@ -53,6 +69,7 @@ class ValidationStructure:
     stratify_on: str | None = None
     group_time_on: str | None = None
     size_validation_on_groups: bool = False
+    temporal_forward_only: bool = False
 
     def __post_init__(self):
         if self.group_on is not None and self.time_on is not None:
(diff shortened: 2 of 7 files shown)


LennartPurucker (approved):
LGTM

PR #5798, opened 2026-08-04
Report constraint misses as skips, and add min_features / min_cells / max_cells

Innixma:
Two related changes to the fit-constraint system: constraint misses are reported as skips rather than crashes, and three new bounds are added (`ag.min_features`, `ag.min_cells`, `ag.max_cells`).

## 1. A constraint miss is a skip, not a crash

A model that hits one of its fit constraints raised a bare `AssertionError`. The trainer classifies exceptions by type to choose between a one-line skip and a failure with a traceback, and `AssertionError` fell through to the failure branch.

**Before:**

```
	Warning: Exception caused Dummy to fail during training... Skipping this model.
		ag.max_rows=5 for model 'Dummy', but found 80 rows.
Detailed Traceback:
  File ".../abstract_trainer.py", line 2376, in _train_and_save
    model = self._train_single(**model_fit_kwargs)
  ...
AssertionError: ag.max_rows=5 for model 'Dummy', but found 80 rows.
```

**After:**

```
	Skipping Dummy because a fit constraint is not satisfied: ag.max_rows=5, but the data has 80 rows.
```

Nothing went wrong in this case — the model was configured not to run on data like this. Presenting it as a crash is misleading on its own, and it also buries real failures: a run with several foundation models past their row or feature limits fills the log with tracebacks that all need to be ignored.

A constraint miss is now a skip in every sense:

| | before | after |
|---|---|---|
| log output | failure warning + `Detailed Traceback` | one `Skipping ...` line |
| `_models_failed_to_train_errors` | records it as a failure | `{}` — recorded in `_models_skipped_by_constraint` instead |
| `raise_on_model_failure=True` | aborts the run | skips, run continues |
| every model skipped | `RuntimeError` | `RuntimeError` (unchanged) |

The `raise_on_model_failure` change matters for sweeps: that flag exists to surface real breakage, and a benchmark over mixed dataset sizes could not enable it at all if a foundation model past its row limit killed the run. "Nothing was trained" stays loud via `raise_on_no_models_fitted`.

### What changed

- New `ConstraintViolationError` in `core/utils/exceptions.py`, raised by all five checks in `validate_fit_args`.
- The trainer recognizes it and logs a single line at level 20 with no traceback, alongside the existing branches for `TimeLimitExceeded`, `NotEnoughMemoryError`, `NoGPUError` and friends.
- The exception carries a `reason` without the model name (the skip line already names it), while its own message stays self-contained for when it surfaces directly, e.g. under `raise_on_model_failure=True`.

`ConstraintViolationError` derives from `AssertionError` as well as `AutoGluonException`, so anything catching `AssertionError` keeps working.

`test_raise_on_fit_args` asserted the old aborting behavior, so its six gating cases now assert the skip instead: each pairs the constrained model with one that trains, keeps `raise_on_model_failure=True` on, and checks the skip was recorded with the expected constraint and not counted as a failure. The bagging case stays an assertion that nothing is trained, since 4 folds over 4 rows leaves no room for a companion model.

Unchanged: which models are skipped, and `ag.ignore_constraints`.

## 2. New bounds: `min_features`, `min_cells`, `max_cells`

Row and feature bounds cannot express total table size: a feature limit wide enough for a short-and-wide table also admits a long-and-wide one many times larger. `ag.min_cells` / `ag.max_cells` bound `rows * features`, for models whose cost tracks the cell count rather than either axis alone. `ag.min_features` fills in the lower bound that `ag.max_features` lacked.

The full set is now `min_rows` / `max_rows`, `min_features` / `max_features`, `min_cells` / `max_cells`, plus `max_classes` and `problem_types`, all honoring `ag.ignore_constraints`.

Both cell bounds reuse the feature count the feature bounds already compute — post-preprocessing where that can be estimated — and that estimate (which fits a feature generator) now runs at most once per fit for all four bounds, where before it ran only for `max_features`.

The message reports the shape behind the count, since a cell total alone does not say which axis was responsible:

```
	Skipping Dummy because a fit constraint is not satisfied: ag.max_cells=200, but the data has 640 cells (80 rows x 8 features).
```

The case cell bounds exist for, from the verification script below: an 80x8 fit passes `max_rows=1000` and `max_features=1000` on each axis alone, yet `max_cells=500` catches it.

## Verifying the output

`tmp_verify_constraint_messages.py` (not committed) fits a predictor once per constraint with two models — a `DummyModel` that violates it and a LightGBM that trains normally — so each skip line appears in the context it really occurs in:

```
	Skipping Dummy because a fit constraint is not satisfied: ag.min_rows=10000, but the data has only 80 rows.
	Skipping Dummy because a fit constraint is not satisfied: ag.max_rows=5, but the data has 80 rows.
	Skipping Dummy because a fit constraint is not satisfied: ag.min_features=50, but the data has only 8 features.
	Skipping Dummy because a fit constraint is not satisfied: ag.max_features=3, but the data has 8 features.
	Skipping Dummy because a fit constraint is not satisfied: ag.min_cells=100000, but the data has only 640 cells (80 rows x 8 features).
	Skipping Dummy because a fit constraint is not satisfied: ag.max_cells=200, but the data has 640 cells (80 rows x 8 features).
	Skipping Dummy because a fit constraint is not satisfied: ag.max_classes=2, but the data has 5 classes.
	Skipping Dummy because a fit constraint is not satisfied: ag.problem_types=['regression'], but the problem type is 'binary'.
	Skipping Dummy because a fit constraint is not satisfied: ag.max_cells=500, but the data has 640 cells (80 rows x 8 features).
```

Zero occurrences of `Detailed Traceback` across all nine runs, LightGBM trains in each, and the `ag.ignore_constraints=True` case still trains the model despite violating its constraint.

(The row count is the model's own fit data — 80 of 100 rows, the rest being the holdout split — which is what the constraint is checked against.)

## Testing

- `test_constraint_violation_is_a_clean_skip` — `ConstraintViolationError` is still an `AssertionError`, its message names the model while `reason` does not, and end to end the skip is recorded with the model-naming form.
- `test_constraint_violation_is_not_a_model_failure` — the constrained model is skipped, others still train, the failures dict stays empty, and the skip is introspectable via `_models_skipped_by_constraint`.
- `test_constraint_violation_still_raises_when_nothing_is_trained` — skipping everything still raises.
- `test_cell_and_feature_bounds` — the new bounds pass when satisfied and raise when not, `ignore_constraints` covers them, the message reports the shape behind the cell count, and a table within generous row and feature bounds is still caught by a cell bound.
- `tabular/tests/unittests/models/test_dummy.py`: 16 passed, including the 13 pre-existing tests unmodified.

### DIFF
--- a/core/src/autogluon/core/models/abstract/_auxiliary_params.py
+++ b/core/src/autogluon/core/models/abstract/_auxiliary_params.py
@@ -100,15 +100,27 @@ class AuxiliaryParams:
     """If specified, raises an AssertionError at fit time if len(X) < min_rows."""
     max_rows: int | None = field(default=None, metadata=_WRAPPER_ONLY)
     """If specified, raises an AssertionError at fit time if len(X) > max_rows."""
+    min_features: int | None = field(default=None, metadata=_WRAPPER_ONLY)
+    """If specified, raises an AssertionError at fit time if len(X.columns) < min_features."""
     max_features: int | None = field(default=None, metadata=_WRAPPER_ONLY)
     """If specified, raises an AssertionError at fit time if len(X.columns) > max_features."""
+    min_cells: int | None = field(default=None, metadata=_WRAPPER_ONLY)
+    """If specified, raises an AssertionError at fit time if len(X) * len(X.columns) < min_cells."""
+    max_cells: int | None = field(default=None, metadata=_WRAPPER_ONLY)
+    """If specified, raises an AssertionError at fit time if len(X) * len(X.columns) > max_cells.
+
+    Cell bounds constrain total table size, for models whose cost tracks the cell count rather than
+    rows or columns alone. Row and feature bounds cannot express that: a feature limit wide enough
+    for a short-and-wide table also admits a long-and-wide one many times larger. Both cell bounds
+    use the same feature count as the feature bounds (post-preprocessing where that can be
+    estimated)."""
     max_classes: int | None = field(default=None, metadata=_WRAPPER_ONLY)
     """If specified, raises an AssertionError at fit time if self.num_classes > max_classes."""
     problem_types: list[str] | None = field(default=None, metadata=_WRAPPER_ONLY)
     """If specified, raises an AssertionError at fit time if self.problem_type not in problem_types."""
     ignore_constraints: bool = field(default=False, metadata=_WRAPPER_ONLY)
-    """If True, ignores the values of `min_rows`, `max_rows`, `max_features`, `max_classes` and
-    `problem_types`."""
+    """If True, ignores the values of `min_rows`, `max_rows`, `min_features`, `max_features`,
+    `min_cells`, `max_cells`, `max_classes` and `problem_types`."""
     max_batch_size: int | str | None = field(default=None, metadata=_WRAPPER_ONLY)
     """If specified, predictions on more than `max_batch_size` rows are computed in chunks of at most
     `max_batch_size` rows each (see `_predict_proba_batch`), bounding prediction-time memory usage.
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -1412,7 +1412,10 @@ class AbstractModel(ModelBase, Tunable):
             ag.problem_types
             ag.min_rows
             ag.max_rows
+            ag.min_features
             ag.max_features
+            ag.min_cells
+            ag.max_cells
             ag.max_classes
             ag.ignore_constraints
         """
@@ -1425,7 +1428,10 @@ class AbstractModel(ModelBase, Tunable):
         max_classes: int | None = aux_params.max_classes
         min_rows: int | None = aux_params.min_rows
         max_rows: int | None = aux_params.max_rows
+        min_features: int | None = aux_params.min_features
         max_features: int | None = aux_params.max_features
+        min_cells: int | None = aux_params.min_cells
+        max_cells: int | None = aux_params.max_cells
         ignore_constraints: bool = aux_params.ignore_constraints
 
         if ignore_constraints:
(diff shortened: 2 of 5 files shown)


LennartPurucker (approved):
This avoid them being counted as failure, right? Or do they still counts as failed model>

Innixma:
They still counted as failures — you were right to ask. The previous commit only changed the logging, so a skipped model was still recorded in `_models_failed_to_train_errors`, which is what callers read to decide whether a run went wrong. A healthy run that skips a foundation model past its row limit looked like a run with failures.

Fixed in the follow-up commit: constraint skips now go to a separate `_models_skipped_by_constraint` map, so they stay introspectable without being failures. Measured before/after on a fit with one constrained model plus a LightGBM:

| | `_models_failed_to_train_errors` |
|---|---|
| before | `{"Dummy": "ConstraintViolationError"}` |
| after | `{}` |

Two related behaviours I deliberately left alone:

- `raise_on_model_failure=True` still aborts on a constraint skip. That flag is documented as "raise instead of skipping to next model", and `test_raise_on_fit_args` pins constraints raising under it, so whether a *configured* skip should abort a strict run seems like a separate call from whether it counts as a failure. Happy to change it if you think a strict run should tolerate them — it matters for benchmark sweeps where some models legitimately do not apply to some datasets.
- A run where *every* model is constraint-skipped still raises `RuntimeError: No models were trained successfully`, via `raise_on_no_models_fitted`.

Also in this PR now (per a follow-up request): `ag.min_features`, `ag.min_cells` and `ag.max_cells`, the last two bounding `rows * features` for models whose cost tracks table size rather than either axis alone.

Innixma:
Made constraint skips bypass `raise_on_model_failure` as well, so a miss is now a skip in every sense:

| | before | after |
|---|---|---|
| log output | failure warning + `Detailed Traceback` | one `Skipping ...` line |
| `_models_failed_to_train_errors` | records it as a failure | `{}` (recorded in `_models_skipped_by_constraint`) |
| `raise_on_model_failure=True` | aborts the run | skips, run continues |
| every model skipped | `RuntimeError` | `RuntimeError` (unchanged) |

Note this **changes `test_raise_on_fit_args`**, which asserted the old aborting behaviour — flagging it since that was a deliberate existing contract rather than an oversight. Its six gating cases now pair the constrained model with one that trains, keep `raise_on_model_failure=True` on, and assert the skip was recorded with the expected constraint and not counted as a failure. The bagging case still asserts nothing is trained, since 4 folds over 4 rows leaves no room for a companion model.

The motivation is sweeps: that flag exists to catch real breakage, and a benchmark over mixed dataset sizes cannot enable it at all if a foundation model past its row limit kills the run. "Nothing was trained" stays loud through `raise_on_no_models_fitted`.

PR #5799, opened 2026-08-04
Resolve the structure holdout before feature generation

Innixma:
## Summary

`validation_structure`'s **bagged** path resolves its splits in the learner, on the raw cleaned frame, with the explicit intent that feature transformation cannot alter the columns the structure names:

```python
# Structure-aware bagging: explicit (train_idx, val_idx) splits, computed on the
# raw cleaned frame before feature transformation can alter the referenced columns.
```

The **non-bagged holdout** path did not. It left the split to the trainer (`generate_train_test_split`), which runs *after* feature generation — so a generator that consumes a structure column and does not emit it made the fit die:

```
KeyError: "`group_on` column 'mouse_id' not found in the training data."
```

This is not a hypothetical generator. TabArena's `tabarena_default` pipeline does exactly that:

```
GroupAggregationFeatureGenerator: generating groupby aggregation features and then drop group columns ['mouse_id'].
```

So every grouped task fit through that pipeline on the holdout path failed — found while running the AutoGluon 1.6 preset over BeyondArena, where it killed the whole fit (zero models trained) on `pancreatic_cancer_mouse_detection`.

## Fix

Carve the holdout in the learner alongside the bagged path, reusing the same mechanism `use_bag_holdout` already used to hand explicit validation rows to the trainer. The trainer then never needs to split, and never needs the structure columns to still exist. Verified on the failing task, with the ordering now visible in the log:

```
Structure-aware holdout: 23 validation rows (holdout_frac=0.2); training on 92 rows.
        GroupAggregationFeatureGenerator: ... and then drop group columns ['mouse_id'].
SUCCESS — models: ['LightGBM']
```

The resolved holdout is group-disjoint (verified separately on the same task: 97 train / 18 validation rows, no `mouse_id` spanning both).

A stratify-only structure still returns no split from `holdout_split_indices`, so AutoGluon's default holdout is left untouched — unchanged behavior for tasks with nothing to correct.

`bag_holdout` is renamed to `structure_holdout`, since the same variable now carries either the bag-holdout (when bagging) or the validation data (on the holdout path).

## Testing

`test__holdout__group_column_dropped_by_feature_generation` fits on the holdout path with a feature generator that emits only the non-group columns, asserting a model trains, the group column is absent from the feature set, and the resolved holdout is group-disjoint. It reproduces the failure without depending on TabArena.

`tabular/tests/unittests/test_validation_structure.py`: 33 passed (32 pre-existing, unmodified).

### DIFF
--- a/tabular/src/autogluon/tabular/learner/default_learner.py
+++ b/tabular/src/autogluon/tabular/learner/default_learner.py
@@ -336,14 +336,41 @@ class DefaultLearner(AbstractTabularLearner):
         X, w = extract_column(X, self.sample_weight)
         X, groups = extract_column(X, self.groups)
         structure_splits = None
-        bag_holdout = None
+        structure_holdout = None
+        if validation_structure is not None and groups is not None:
+            raise ValueError("Specify either `groups` or `validation_structure`, not both.")
+        if validation_structure is not None and num_bag_folds < 2 and X_val is None:
+            # Non-bagged holdout, resolved here for the same reason the bagged path below is:
+            # feature transformation can drop the very columns the structure names (TabArena's
+            # `tabarena_default` pipeline builds aggregate features from the group column and then
+            # drops it), and the trainer's own holdout split runs *after* that. Carving the rows
+            # here, off the raw cleaned frame, hands the trainer explicit validation data so it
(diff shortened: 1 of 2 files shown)


LennartPurucker (approved):
LGTM

PR #5805, opened 2026-08-04
[timeseries] Silence noisy model loading logs for Chronos and TFT

shchur:
*Issue #, if available:*

*Description of changes:*

Suppresses noisy output printed while loading/fitting some time series models, without hiding anything useful.

- **Chronos & Chronos-2** (`models/chronos/model.py`, `chronos2.py`): wrap the `from_pretrained(...)` call in the existing `warning_filter(all_warnings=True)`. This silences the `TqdmWarning: IProgress not found ...` warning and the "sending unauthenticated requests to the HF Hub" warning. The actual download/weight-loading progress bar is left intact.
- **TFT** (`models/gluonts/abstract.py`): add `enable_model_summary=False` to the Lightning trainer kwargs (next to the existing `enable_progress_bar=False`). Lightning 2.7 started printing a large rich model-summary table on every fit; this disables it.


Innixma (approved):
LGTM!

PR #5806, opened 2026-08-04
Do not give up the requested fold count over a rare stratification value

Innixma:
Two fold-construction paths in `validation_structure.py` treated a rare stratification value
as a reason to give up on the requested fold count. One aborted the fit; the other silently
shrank the bagging.

## 1. Grouped folds crashed instead of splitting

```
File "autogluon/common/utils/validation_structure.py", line 534, in _per_group_splits
File "sklearn/model_selection/_split.py", line 800, in _make_test_folds
ValueError: n_splits=8 cannot be greater than the number of members in each class.
```

sklearn raises that only when **every** class holds fewer than `n_splits` members:

```python
if np.all(self.n_splits > y_counts):   # sklearn/model_selection/_split.py
    raise ValueError("n_splits=%d cannot be greater than the number of members in each class.")
if self.n_splits > min_groups:
    warnings.warn("The least populated class in y has only %d members ...")
```

One class above the bar turns it into a warning. Measured:

| class counts | n_splits | result |
|---|---:|---|
| 8 classes × 7 | 8 | **RAISE** |
| 7 classes × 7 + one class of 8 | 8 | OK, 8 folds, warning only |
| 8 classes × 2 | 8 | **RAISE** |
| 7 classes × 2 + one class of 8 | 8 | **OK, 8 folds**, warning only |

The last row is the point: 2-member classes split into 8 folds fine. The refusal is an
artifact of the `np.all` guard, not a property of the data. k-fold needs only two members of
a class, in two *different* folds, for every training split to retain one of them.

The group table is where this bites. It holds one row per group with the class read off the
group's first row, so a class with plenty of rows can still have only a handful of groups.

**Fix:** route the split through `CVSplitter`, which has carried a workaround for this
sklearn behaviour for years:

```python
# FIXME: There is a bug in sklearn that causes an incorrect ValueError if performing
#  stratification and all classes have fewer than n_splits samples.
#  This is hacked by adding a dummy class with n_splits samples, performing the kfold
#  split, then removing the dummy samples from all resulting indices.
```

It otherwise constructs the same `Repeated{,Stratified}KFold` with the same arguments, so
**splits are unchanged wherever sklearn succeeds today** — verified bit-identical across
stratified and unstratified, string / categorical / integer targets, uneven group sizes, and
repeated bagging (7/7 configurations).

## 2. Stratified folds were capped at the rarest value's count

```python
n_splits = min(n_splits, int(stratify.value_counts().min()))   # removed
```

Same argument: a value occurring twice supports any number of folds. The cap stood in for the
sklearn refusal above, so with `CVSplitter` handling that, the bound becomes the row count.
Requesting 8 folds:

| | old | new |
|---|---:|---:|
| minority=3 of 3 classes | 3 folds | **8 folds** |
| minority=2, all classes < 8 | 2 folds | **8 folds** |
| minority=2 + one class of 8 | 2 folds | **8 folds** |

each with every row validated exactly once and every value present in every training split.
Bit-identical wherever the cap did not bite.

A value occurring **once** is the case two members cannot rescue — the fold that validates it
trains on none of it. That is what `augment_rare_classes` duplicates upstream, so it stays
rejected, now checked directly rather than as a side-effect of the cap collapsing below 2.

## 3. Rarity was counted in rows while folds are built from groups

Two rows inside one group are **one** independent unit: they always land in the same fold, so
the fold validating them trains on none of that value. Row duplication — what
`augment_rare_classes` does — cannot change that. Demonstrated before this change:

```
'rare' rows=2, groups holding it=1   (threshold=2 is satisfied on rows)
folds=2 | folds whose TRAIN split has no 'rare' row: [0]
warnings raised: none
```

`ValidationStructure.rare_stratification_values` is the shared answer: occupancy counted in
groups where grouping is declared, in rows otherwise. Four call sites now use it or the
principle behind it:

**`default_learner`** reports values still spanning fewer than 2 groups after augmentation,
instead of implying `threshold=2` settled it.

**`holdout_split_indices`** takes `min_cls_count_train` — the per-class floor
`generate_train_test_split` has always enforced on the *unstructured* path, which the
structured path silently skipped — and honours it by moving **whole groups** across the
boundary so the split stays group-disjoint. A forward holdout cannot be repaired that way
(it would move the time boundary) and is reported instead. When repair would consume the
entire holdout, the original split is kept: an unlearnable rare value degrades a fit, an
empty holdout has nothing to validate at all.

**`augment_rare_classes` now runs for the `groups` channel too.** Skipping it there left that
path dropping single-row classes in the cleaner while `validation_structure` duplicated them —
same input, different training data. Measured on a 60-row frame with one single-row class:

```
before:  Train Data Class Count: 2     (no "Duplicated" line — the class was dropped)
after:   Duplicated 1 samples from 1 rare classes ...
         Train Data Class Count: 3
```

**`custom_splits` no longer shrinks `num_folds`/`num_repeats`** when a value is scarcer than
the fold count. A 5×3 request on such data returned 2×1 before and returns 5×3 now.

**`_validate_splits` checks the training side of every fold, in every problem type.** It
previously warned only about validation folds and only for binary, so the serious direction —
a fold whose model never sees a value — was silent.

## Why reducing the fold count is (and isn't) the right response

The reductions removed here were protecting the wrong thing. The test is whether a reduction
protects *constructibility* or *scorability*:

- **`n_groups < num_folds` — required, and kept.** Folds are group-disjoint, so groups are the
  indivisible unit. More folds than groups means at least one fold gets no group at all, i.e.
  an empty validation split. The request cannot be honoured.
- **`n_splits` > a value's member count — not required, removed.** The folds are still
  constructible and every training split still keeps every value once it spans two groups
  (asserted in the new tests). Nothing about the *partition* is invalid.

What degrades in the second case is only per-fold scorability: with 6 groups of a class over 8
folds, at least 2 folds validate on no example of it. The aggregate OOF is unaffected — each
row is still validated exactly once, so the OOF metric the ensemble weights on sees every
value. Only an individual child's validation score narrows, and for binary metrics a fold
holding one value is unscorable outright (`roc_auc` undefined). So the right response there is
to build the folds and report, which is what sklearn itself does once its `np.all` guard
passes — and what `_validate_splits` now does in both directions.

## Verification

On the shape that failed — 720 rows, 48 groups, 8 classes, 6 groups per class, 8 folds:

```
row-level min class count=90   GROUP-level min class count=6
old: ValueError: n_splits=8 cannot be greater than the number of members in each class.
new: 8 folds | every row validated exactly once: True | group-disjoint: True
     | every class present in every train split: True
     val fold sizes (rows): [90, 90, 90, 90, 90, 90, 90, 90]
```

An end-to-end `TabularPredictor.fit` on that shape with
`validation_structure={"group_on": ..., "stratify_on": ...}`, `num_bag_folds=8` now fits 8
child models with full OOF coverage (0 NaN rows) where it previously raised.

Non-numeric targets are encoded to first-appearance integer codes before splitting (shared
`_split_target`): the sentinel class the workaround appends cannot be ordered against string
labels (`TypeError: '<' not supported between instances of 'int' and 'str'`), and
first-appearance codes are the encoding sklearn derives internally
(`np.unique(y_idx, return_inverse=True)` ranks classes by first occurrence), so the folds do
not move — verified. Value counting also moved off the Series, because a categorical dtype
reports unused categories as zero-count entries.

- 6 new tests covering rare-group counting, fold-count preservation, the holdout repair, the
  holdout-would-be-emptied fallback, and both crash shapes
- `tabular/tests/unittests/test_validation_structure.py`: 40 passed
- `common/tests/unittests/utils/test_cv_splitter.py`: 15 passed

## How it was found

An AutoGluon 1.6 preset run over grouped BeyondArena tasks.
`mice_protein_trisomy_discriminant` (720 train rows, 48 mouse groups, 8 classes, 6 groups per
class) failed at 8 folds, and because the failure aborted its worker it also took the four
tasks queued behind it down — 5 of 40 tasks lost to one avoidable split error.

### DIFF
--- a/common/src/autogluon/common/utils/validation_structure.py
+++ b/common/src/autogluon/common/utils/validation_structure.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import logging
 from dataclasses import dataclass, fields
+from typing import Any
 
 import numpy as np
 import pandas as pd
@@ -122,6 +123,30 @@ class ValidationStructure:
         groups = self._group_values(X)
         return bool(pd.Series(np.asarray(stratify)).groupby(np.asarray(groups)).nunique().max() == 1)
 
+    def rare_stratification_values(
+        self, X: pd.DataFrame, y: pd.Series, min_units: int = 2, problem_type: str | None = None
+    ) -> dict[Any, int]:
+        """Stratification values spanning fewer than ``min_units`` independent units.
+
+        The independent unit is the group where grouping is declared, and the row otherwise.
+        The distinction matters because the guarantee callers want from "at least two members"
+        is that *every* training split retains one of them, and only two members in two
+        different folds deliver it. Two rows inside one group are one unit: they always land in
+        the same fold, so the fold validating them trains on none of that value. Duplicating
+        rows -- what ``augment_rare_classes`` does -- cannot fix that, which is why the count
+        has to be taken in units rather than rows.
+
+        Returns ``{value: n_units}`` for the offending values only, empty when there is no
+        stratification signal.
+        """
+        stratify = self._resolve_stratify_for_folds(X, y, problem_type)
+        if stratify is None:
+            return {}
+        values = np.asarray(stratify)
+        units = np.asarray(self._group_values(X)) if self.group_on is not None else np.arange(len(values))
+        units_per_value = pd.Series(units).groupby(values, observed=True).nunique()
+        return {value: int(count) for value, count in units_per_value.items() if count < min_units}
+
     def num_group_instances(self, X: pd.DataFrame) -> int | None:
         """Number of distinct groups, or None when this structure declares no grouping.
 
@@ -237,16 +262,19 @@ class ValidationStructure:
                     f"setting num_folds={n_groups} and num_repeats=1.",
                 )
                 num_folds, num_repeats = n_groups, 1
-            stratify = self._resolve_stratify_for_folds(X, y, problem_type)
-            if stratify is not None:
-                minority = int(stratify.value_counts().min())
-                if minority < num_folds:
-                    logger.log(
-                        20,
-                        f"validation_structure: rarest stratification value occurs {minority} times "
-                        f"< num_folds ({num_folds}); setting num_folds={minority} and num_repeats=1.",
-                    )
-                    num_folds, num_repeats = max(2, minority), 1
+            # A stratification value scarcer than num_folds does *not* reduce the fold count.
+            # Groups are the indivisible unit, so too few groups makes folds unbuildable and is
+            # corrected above; too few members of a value does not -- the partition stays valid
+            # and every training split keeps the value as long as it spans two groups. What
+            # suffers is per-fold scorability, which `_validate_splits` reports.
+            rare = self.rare_stratification_values(X, y, min_units=num_folds, problem_type=problem_type)
+            if rare:
+                logger.log(
+                    20,
+                    f"validation_structure: stratification value(s) {sorted(map(str, rare))} span fewer "
+                    f"than num_folds ({num_folds}) groups, so some folds validate on none of them; "
+                    f"keeping num_folds={num_folds}.",
+                )
 
         if self.group_on is not None and self._can_split_group_table(X, y, problem_type):
             # Every group carries one stratification value: split the *group table* (one row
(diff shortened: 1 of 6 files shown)


LennartPurucker (approved):
LGTM

PR #5807, opened 2026-08-04
[timeseries] Move chronos import inside warning_filter to silence TqdmWarning

shchur:
*Issue #, if available:*

*Description of changes:*

Follow-up to #5805. That PR wrapped Chronos `from_pretrained(...)` in `warning_filter`, but the `TqdmWarning: IProgress not found` still appeared.

Root cause: the warning is emitted **at import time** of `tqdm.auto` (in `tqdm/autonotebook.py`), not when a progress bar is drawn. Importing `chronos` transitively imports `tqdm.auto`, so the warning fires during the `from chronos ...` import statement — which was *outside* the `warning_filter` block.

Fix: move the lazy `chronos` import inside the `warning_filter(all_warnings=True)` block in both `ChronosModel` and `Chronos2Model`, so the import-time warning is suppressed. The download progress bar itself is unaffected.


Innixma (approved):
LGTM!

PR #5809, opened 2026-08-05
Stop losing missing categorical values in the TabPFN wrappers

Innixma:
## Problem

`TabPFNModel._preprocess` label-encodes categorical columns before handing them to TabPFN:

```python
X = self._label_encode_categoricals(X, is_train=is_train)
```

which runs `LabelEncoderFeatureGenerator`:

```python
def convert_category_to_int(X: DataFrame) -> DataFrame:
    X = X.apply(lambda x: x.cat.codes)   # pandas maps MISSING to -1
```

TabPFN then casts every column named in `categorical_features_indices` back to `category` itself:

```python
# tabpfn/preprocessing/clean.py, fix_dtypes
X[cat_indices] = X[cat_indices].astype("category")
```

so the `-1` becomes an ordinary category level. TabPFN sees **no missing values** in those
columns and its missing-value handling never runs; on a column that is mostly missing it instead
reads a highly informative category. Confirmed with instrumentation: both wrappers produce
identical modality decisions (all columns `CATEGORICAL`, thresholds
`max_unique_for_category=30`, `min_unique_for_numerical=4`) and identical
`categorical_features_indices` — only the values differ.

## Isolated measurement

120-row frame, one reported categorical column with 25 missing values, partition / reported
indices / seed held fixed:

| frame | missing represented as | max \|Δ predict_proba\| |
|---|---|---:|
| `category` dtype | 25 NaN | reference |
| `.cat.codes` | 25 rows at −1, 0 NaN | **0.096397** (v3), **0.033440** (v2.6) |
| `.cat.codes`, −1 restored to NaN | 25 NaN | **0.000000** |

Label-encoding by itself is harmless — the third row proves that. Only the lost missingness
moves predictions.

## Fix

Record the categorical indices and leave the values as `category` dtype. TabPFN needs the
indices, not encoded values, and ordinal-encodes from the dtype itself. This is also cheaper
than any encoded form: `category` stores one code byte per row plus a level table, against
float64's eight bytes per row (measured at 100k rows / 4 levels: 100,416 B vs 800,000 B).

## End-to-end verification

BeyondArena's `audiology_diagnosis` — 69 categorical columns, 7 with missing values, `bser`
missing in 196 of 199 rows. Same experiment with only the wrapper swapped, all four core splits:

| split | before | after | TabArena's own wrapper |
|---:|---:|---:|---:|
| 0 | 0.20912 | 0.21011 | 0.21011 |
| 1 | 0.50491 | 0.46785 | 0.46785 |
| 2 | 0.31384 | 0.25668 | 0.25668 |
| 3 | 0.24806 | 0.24438 | 0.24438 |

After the change AutoGluon reproduces TabArena's `TA-TabPFN-3` wrapper **exactly** — that
wrapper has passed the dtype through from the start, which is what made the discrepancy
findable. Mean error on this dataset improves by 0.024 (8.2%).

- New test `test_tabpfn_preprocess_preserves_missing_categoricals`, covering TabPFN-3 and
  TabPFN-2.6 (dtype preserved, NaN count preserved, no `-1` level, indices still correct,
  numeric column untouched).
- `tabular/tests/unittests/models/test_tabpfn3.py`: 2 passed, 1 skipped (the skip is the
  pre-existing license-gated fit test).
- End-to-end bagged fit + predict on a frame with missing categoricals still works.

## Scope

Shared by every TabPFN version through this base class: **TabPFN-2.5, RealTabPFN-2.5,
TabPFN-2.6, TabPFN-3**.

## Same defect elsewhere, not fixed here

`_label_encode_categoricals` is documented as *"preserving missing values"*, which is not what
`.cat.codes` does. Two other foundation-model wrappers rely on it and lose missingness the same
way, with no `categorical_features_indices` to soften the landing — the `-1` arrives as an
ordinary value inside a continuous feature:

- `nori_model.py`: `X = self._label_encode_categoricals(X)` then `to_numpy(float32)`, under a
  docstring that says *"with NaN preserved (handled natively)"*.
- `tabdpt_model.py`: the same pattern.

`TabICLv2` is unaffected (no `_preprocess`, values pass through) and `RealMLP` is unaffected
(its `_category_mapping` restores NaN via a mask). I have left Nori and TabDPT alone rather than
change models this PR has not measured; happy to follow up.

### DIFF
--- a/tabular/src/autogluon/tabular/models/tabpfnv2/tabpfnv2_5_model.py
+++ b/tabular/src/autogluon/tabular/models/tabpfnv2/tabpfnv2_5_model.py
@@ -86,7 +86,6 @@ class TabPFNModel(AbstractTorchModel):
 
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
-        self._cat_features = None
         self._cat_indices = None
         # `ag.max_batch_size="auto"` resolved against the training size during `_fit`.
         self._max_batch_size_resolved: int | None = None
@@ -111,17 +110,18 @@ class TabPFNModel(AbstractTorchModel):
         ordinary level, so TabPFN sees no missing values in those columns and its missing-value
         handling never runs. Passing the dtype through keeps the missingness and is cheaper
         (`category` stores one code byte per row plus a level table, against float64's eight).
+
+        The indices are still needed. TabPFN infers a column's modality from its dtype, and a
+        `category` column whose levels are integers reads as numeric (`_is_numeric_pandas_series`
+        coerces it), so without them anything above `min_unique_for_numerical` levels is treated
+        as NUMERICAL rather than CATEGORICAL. AutoGluon's default `CategoryFeatureGenerator`
+        minimizes memory by re-coding levels to integers, which makes that the common case.
         """
         X = super()._preprocess(X, **kwargs)
 
         if is_train:
-            # Detect/set cat features and indices
-            self._cat_indices = []
             categorical_features = X.select_dtypes(include=["category"]).columns.tolist()
-            if categorical_features:
-                if self._cat_features is None:
-                    self._cat_features = categorical_features
-                self._cat_indices = [X.columns.get_loc(col) for col in self._cat_features]
+            self._cat_indices = [X.columns.get_loc(column) for column in categorical_features]
 
         return X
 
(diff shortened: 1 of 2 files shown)


LennartPurucker (changes requested):
I would rather completely drop the label encoding for TabPFN in AutoGluon as done for the v3 in TabArena. This is on the more bad-than-good side of diverging implementations.

Innixma:
> I would rather completely drop the label encoding for TabPFN in AutoGluon as done for the v3 in TabArena. This is on the more bad-than-good side of diverging implementations.

Removed everything that was removable, the rest is needed

LennartPurucker (approved):

PR #5810, opened 2026-08-05
Keep missing values and lift the class cap for TabDPT

Innixma:
Two independent limits in the same wrapper. Both apply to **TabDPT and TabDPT-Turbo** (which
inherits this class).

## 1. Missing categorical values were erased

`_preprocess` label-encoded categoricals before handing over the numpy array, and `.cat.codes`
maps missing to `-1`, so the array reached TabDPT with no NaN in those columns.

That matters more here than in a model which merely tolerates NaN. TabDPT appends a **binary
missing-indicator column per affected feature** and then mean-imputes:

```python
# tabdpt/estimator.py
inds = np.isnan(X)
self.has_missing_indicator = inds.any(axis=0)
X = np.hstack((X, inds[:, self.has_missing_indicator].astype(float)))
X = self.imputer.fit_transform(X)          # SimpleImputer(strategy="mean")
```

A `-1` suppresses both mechanisms *and* puts an out-of-range value on the feature's numeric axis,
one step below the first real code, which then goes through the scaler. The missing mask is
restored after encoding. Numeric columns were never affected — their NaN always reached
`np.isnan`.

### Measured impact

BeyondArena datasets under 10k train rows whose categorical columns contain missing values (17 of
45 such datasets — the rest have none, so both variants are identical by construction), first
split, 2×1 bags:

| dataset | before | after | rel Δ |
|---|---:|---:|---:|
| heart_disease_hungary | 0.08526 | 0.07800 | **−8.5%** |
| regensburg_pediatric_appendicitis (32/33 cols) | 0.09176 | 0.08476 | **−7.6%** |
| audiology_diagnosis | 0.32755 | 0.30661 | **−6.4%** |
| ljubljana_primary_tumor | 1.26923 | 1.21368 | **−4.4%** |
| heart_disease_va_long_beach | 0.21882 | 0.21059 | −3.8% |
| ljubljana_breast_cancer | 0.23726 | 0.22954 | −3.3% |
| horse_colic_survival | 0.74084 | 0.71801 | −3.1% |
| heart_disease_cleveland | 0.07154 | 0.07036 | −1.7% |
| in_vehicle_coupon_recommendation | 0.16459 | 0.16196 | −1.6% |
| coffee_rating_prediction | 1.28476 | 1.27421 | −0.8% |
| ghanas_indigenous_intel | 0.27141 | 0.26971 | −0.6% |
| credit_approval | 0.06419 | 0.06411 | −0.1% |
| wine_world_cost | 0.39657 | 0.39904 | +0.6% |
| immoscout_german_house_prices (8/8 cols) | 0.44111 | 0.44754 | +1.5% |
| cirrhosis_patient_survival_prediction | 0.62728 | 0.63799 | +1.7% |
| mic (93/95 cols) | 0.46009 | 0.49859 | +8.4% |
| homeq_default_prediction | 0.00026 | 0.00029 | +12.5% |

**12 of 17 improve, 5 regress, median −1.2%.** The clear wins are small clinical tables with
pervasive missingness. The regressions sit where preserving missingness inflates the feature
count sharply: `mic` appends 93 indicator columns to a 95-column frame at 1,132 rows, and
`immoscout` appends 8 of 8. `homeq_default_prediction`'s +12.5% is a ratio artifact on errors of
0.00026 vs 0.00029.

So this is a **correctness fix with a mixed accuracy effect**, not a uniform win. Worth knowing
before merging: if the indicator-column inflation is judged the greater evil on wide categorical
frames, the alternative is to preserve missingness only when the affected column count is small,
which I have not implemented because it adds a threshold with no principled value.

## 2. `max_classes` 10 → 160

The old value carried its own caveat: `"max_classes": 10,  # TODO: Test >10 classes`.

Above the checkpoint's output-head width the library encodes the label in base `max_num_classes`,
runs one forward pass per digit, and sums the per-digit log-probabilities
(`tabdpt.classifier._predict_large_cls`, dispatched from `predict_proba`). The head width
therefore bounds a forward pass, not the usable class count.

Sanity checked end to end — fit, then `predict_proba` returning the right column count with rows
summing to 1:

| n_classes | proba shape | row sum | fit time |
|---:|---|---:|---:|
| 11 | (20, 11) | 1.0000 | 7s |
| 20 | (20, 20) | 1.0000 | 4s |
| 100 | (20, 100) | 1.0000 | 6s |
| **160** | (20, 160) | 1.0000 | 5s |

and on real data: BeyondArena's 11-class `ljubljana_primary_tumor` fits and scores (it is the
dataset the old cap excluded, and it appears in the table above because of this change).

Independent evidence: TabArena's own TabDPT wrapper sets **no** class cap, and its published
BeyondArena results include TabDPT fits at 11 classes (`ljubljana_primary_tumor`) and 20 classes
(`micro_mass`) — real scores, not imputed placeholders.

Cost is `ceil(log_head(n_classes))` prediction passes, so 2 passes to head² classes and 3 to
head³. 160 mirrors the cap TabPFN-3 already uses.

## Tests

- `test_tabdpt_preprocess_preserves_missing_categoricals` — NaN count preserved in the emitted
  array, no `-1` among the remaining values.
- `test_tabdpt_allows_more_than_ten_classes` — the cap on both TabDPT and TabDPT-Turbo.
- `tabular/tests/unittests/models/test_tabdpt.py`: 3 passed.

## Related

The same `.cat.codes` defect was fixed for the TabPFN wrappers in #5809. `nori_model.py` has it
too, under a docstring claiming *"with NaN preserved (handled natively)"*; measured there it is
worth ~0.05% mean on regression under 10k rows (20 of 25 datasets byte-identical), so I have left
it alone. `TabICLv2` is unaffected and `RealMLP` was fixed incidentally by #5808.

### DIFF
--- a/tabular/src/autogluon/tabular/models/tabdpt/tabdpt_model.py
+++ b/tabular/src/autogluon/tabular/models/tabdpt/tabdpt_model.py
@@ -2,12 +2,11 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING, ClassVar
 
-import numpy as np
-
 from autogluon.core.constants import BINARY, MULTICLASS, REGRESSION
 from autogluon.tabular.models.abstract.abstract_torch_model import AbstractTorchModel
 
 if TYPE_CHECKING:
+    import numpy as np
     import pandas as pd
 
 
@@ -212,20 +211,13 @@ class TabDPTModel(AbstractTorchModel):
     def _preprocess(self, X: pd.DataFrame, **kwargs) -> np.ndarray:
         """TabDPT requires a numpy array as input, with missing values left as NaN.
 
-        `.cat.codes` maps a missing categorical value to -1, which would arrive as an ordinary
-        value on the same numeric axis as the real codes. TabDPT handles NaN itself and does more
-        with it than impute: it appends a binary missing-indicator column per affected feature
-        before mean-imputing (`tabdpt.estimator`), so a -1 both hides the missingness and skews
-        the feature. The mask restores it after encoding.
+        TabDPT handles NaN itself and does more with it than impute: it appends a binary
+        missing-indicator column per affected feature before mean-imputing (`tabdpt.estimator`),
+        so the encoding's default -1 for missing would both hide the missingness and put an
+        out-of-range value on the feature's numeric axis.
         """
         X = super()._preprocess(X, **kwargs)
-        categorical_features = X.select_dtypes(include=["category"]).columns.tolist()
-        missing = X[categorical_features].isna() if categorical_features else None
-        X = self._label_encode_categoricals(X)
-        if categorical_features:
-            X = X.copy()
-            for column in categorical_features:
-                X.loc[missing[column].to_numpy(), column] = np.nan
+        X = self._label_encode_categoricals(X, preserve_missing=True)
         return X.to_numpy()
 
     def _more_tags(self) -> dict:
--- a/tabular/src/autogluon/tabular/models/tabdpt/tabdpt_model.py
+++ b/tabular/src/autogluon/tabular/models/tabdpt/tabdpt_model.py
@@ -2,11 +2,12 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING, ClassVar
 
+import numpy as np
+
 from autogluon.core.constants import BINARY, MULTICLASS, REGRESSION
 from autogluon.tabular.models.abstract.abstract_torch_model import AbstractTorchModel
 
 if TYPE_CHECKING:
-    import numpy as np
     import pandas as pd
 
 
@@ -61,7 +62,12 @@ class TabDPTModel(AbstractTorchModel):
     _default_auxiliary_params_extra = {
         "max_rows": 100000,  # TODO: Test >100k rows
         "max_features": 2500,  # TODO: Test >2500 features
-        "max_classes": 10,  # TODO: Test >10 classes
+        # TabDPT decomposes a label into base-`max_num_classes` digits above the checkpoint's
+        # output-head width and sums the per-digit log-probabilities
+        # (`tabdpt.classifier._predict_large_cls`), so the head width bounds a forward pass, not
+        # the usable class count. Verified to fit and predict a calibrated distribution at 11, 20,
+        # 100 and 160 classes; the cost is ceil(log_head(n_classes)) passes per prediction.
+        "max_classes": 160,
     }
     minimum_num_gpus = 0.5
     _default_ag_args_ensemble_extra = {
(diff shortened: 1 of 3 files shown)


tabular/src/autogluon/tabular/models/tabdpt/tabdpt_model.py
```diff
@@ -203,10 +209,23 @@ def _predict_proba(self, X, **kwargs) -> np.ndarray:
...
+        `.cat.codes` maps a missing categorical value to -1, which would arrive as an ordinary
+        value on the same numeric axis as the real codes. TabDPT handles NaN itself and does more
+        with it than impute: it appends a binary missing-indicator column per affected feature
+        before mean-imputing (`tabdpt.estimator`), so a -1 both hides the missingness and skews
+        the feature. The mask restores it after encoding.
+        """
         X = super()._preprocess(X, **kwargs)
+        categorical_features = X.select_dtypes(include=["category"]).columns.tolist()
+        missing = X[categorical_features].isna() if categorical_features else None
         X = self._label_encode_categoricals(X)
```
LennartPurucker: Maybe move this logic into the _label_encode_categoricals function and then add a flag for keeping missings?

LennartPurucker (changes requested):

tabular/src/autogluon/tabular/models/tabdpt/tabdpt_model.py
Innixma: Done in be902a03 — good call, it belongs in the shared helper.

`_label_encode_categoricals` now takes `preserve_missing: bool = False`. The default keeps the
`-1` convention, so the four other callers (Mitra, RandomForest, TabPFNMix, Nori) are untouched,
and TabDPT opts in with one argument:

```python
X = self._label_encode_categoricals(X, preserve_missing=True)
```

Two things fell out of moving it:

- The helper's docstring already claimed it converted categoricals "preserving missing values",
  which it never did. It now documents the real behaviour, what the flag changes, and that the
  affected columns become float (an integer dtype cannot hold NaN).
- The TabDPT wrapper no longer needs `numpy` at runtime, so that import goes back under
  `TYPE_CHECKING`.

Verified unchanged end to end: a standalone fit with 40 missing categorical and 10 missing
numeric cells still emits 24 NaN cells to TabDPT (per-fold slice of a 2-fold bag). Added
`test_label_encode_categoricals_preserve_missing_flag` covering both settings of the flag —
default gives 2 cells at `-1` and no NaN, `preserve_missing=True` gives 2 NaN and no `-1`, with
the non-missing codes identical either way. `test_tabdpt.py`: 4 passed.

Worth flagging while you are here: `nori_model.py` has the same defect and its docstring makes
the same false claim ("with NaN preserved (handled natively)"). I left it out because the measured
impact there is ~0.05% mean on BeyondArena regression under 10k rows, with 20 of 25 datasets
byte-identical — it is a one-word change now that the flag exists, if you would like it in this PR.

LennartPurucker (approved):

PR #5816, opened 2026-08-05
[timeseries] Fix UndefinedVariableError in update() on Python 3.10/3.11

shchur:
## Description

`TimeSeriesPredictor.update()` raised `pandas.errors.UndefinedVariableError: local variable 'items' is not defined` on Python 3.10 and 3.11, while passing on 3.12.

### Root cause

In `TimeSeriesTrainer.update()`:

```python
items = data_per_window[0].item_ids
for window in data_per_window[1:]:
    items = items.intersection(window.item_ids)
data_per_window = [window.query("item_id in @items") for window in data_per_window]
```

`DataFrame.query("... @items")` resolves the `@items` reference by inspecting the **caller's stack frame** for a local named `items`. On Python 3.10/3.11 a list comprehension runs in its **own** stack frame (locals: only `window` and `.0`), so `items` — which lives in the enclosing method frame — is invisible, and pandas raises `UndefinedVariableError`.

[PEP 709](https://peps.python.org/pep-0709/) inlined comprehensions in Python 3.12, collapsing the separate frame, which is why the tests passed on 3.12 but failed on 3.10/3.11. Not flaky — deterministic per interpreter version.

### Fix

Replace the `query` call with `.loc[items]` index selection. Since `items` is an `Index` of item ids and `item_id` is the leading `MultiIndex` level, `.loc[items]` selects exactly those items with no stack-frame introspection, so it works on all supported Python versions (and avoids the `query` expression-engine overhead).

## Testing

`tests/unittests/test_predictor.py -k update` — all 11 tests pass.

### DIFF
--- a/timeseries/src/autogluon/timeseries/trainer/trainer.py
+++ b/timeseries/src/autogluon/timeseries/trainer/trainer.py
@@ -1325,7 +1325,7 @@ class TimeSeriesTrainer(AbstractTrainer[TimeSeriesModelBase]):
         items = data_per_window[0].item_ids
         for window in data_per_window[1:]:
             items = items.intersection(window.item_ids)
-        data_per_window = [window.query("item_id in @items") for window in data_per_window]
+        data_per_window = [window.loc[items] for window in data_per_window]
 
         def update_model(model: TimeSeriesModelBase, oof_predictions: list[TimeSeriesDataFrame]) -> None:
             model.cache_oof_predictions(oof_predictions)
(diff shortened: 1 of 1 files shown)


Innixma (approved):

shchur:
/platform_tests f626248afbe438be4fb059af8ede82e4578b14c4
[Platform Tests Output][1]

[1]: https://github.com/autogluon/autogluon/actions/runs/31019338644

shchur:
/platform_tests 20988115d16e2d46cee25a17a6e112ccafaef67f
[Platform Tests Output][1]

[1]: https://github.com/autogluon/autogluon/actions/runs/31020921411

PR #5818, opened 2026-08-05
[tabular] Silence the replace-downcast FutureWarning in label cleaning

Innixma:
`_check_for_non_finite_values` already chained `.infer_objects(copy=False)` onto the inf-to-NaN `replace`, and its comment claimed that avoids the pandas FutureWarning:

```
FutureWarning: Downcasting behavior in `replace` is deprecated and will be removed in a future version.
```

It does not: the warning is emitted by `replace` itself when it downcasts internally, which happens whenever the label arrives as a numeric-valued object column (common for raw CSV data), before `infer_objects` runs. A 121-dataset benchmark sweep logged it 25 times.

Wrapping the call in `pd.option_context("future.no_silent_downcasting", True)` stops `replace` from downcasting, so it has nothing to warn about, and leaves the explicit `infer_objects` to apply the same downcast. This is the documented migration for the deprecation. Verified through the learner: an object label of floats still resolves to `float64` with identical values and zero warnings, and an inf label is still rejected with the same `ValueError`.

### DIFF
--- a/tabular/src/autogluon/tabular/learner/default_learner.py
+++ b/tabular/src/autogluon/tabular/learner/default_learner.py
@@ -569,9 +569,12 @@ class DefaultLearner(AbstractTabularLearner):
             X = copy.deepcopy(X)
 
             # treat None, NaN, INF, NINF as NA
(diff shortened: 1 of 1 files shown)


shchur (approved):
LGTM!

PR #5824, opened 2026-08-06
[docs] Fix versioned docs publishing (build /X.Y/ from release tags)

shchur:
## Problem

Versioned doc pages stopped publishing. `auto.gluon.ai/1.6/` doesn't exist, and older pages use inconsistent schemes (`/1.4.0/`, `/v1.5.0/`).

1. **#5728** scoped the push trigger to `branches: [master, stable]`. Once `branches:` is set, tag pushes stop triggering CI — the `v1.6.0`/`v1.6.1` tags built nothing.
2. On push the concurrency key collapsed to `Continuous Integration-` (both operands empty), so any push cancelled any other in-flight push build. This is why `/dev/` went stale.
3. Publish path was `basename $github.ref`, so tags and branches produced duplicate schemes.

## Fix

- **`continuous_integration.yml`** — add `tags: ['v[0-9]+.[0-9]+.[0-9]+']`; make concurrency per-ref. The pattern is narrow on purpose: `'v*'` would match `v1.7.0rc1`, and the sync uses `--delete`.
- **`env_setup.sh`** — new `docs_publish_path`: `master`→`dev`, `vX.Y.Z`→`X.Y`, else verbatim. Now shared by `copy_docs.sh` and `build_all_docs.sh`, which had drifting copies of this logic.
- **`conf.py` + `build_doc.sh`** — `/1.6/` is built from `v1.6.1`, so the title said "1.6.1". CI now exports `AG_DOCS_VERSION_LABEL` for release tags; exported for both sphinx-build invocations since tutorials build separately from the rest of the site.
- **`install.md`** — nightly dropdown points readers at the stable instructions.
- **`versions.rst`** — fix dev/stable versions, add `/1.6/` `/1.5/`, drop nonexistent `/1.3.0/`. These two links 404 until rollout.
- **`ReleaseInstructions.md`** — document tag-based publishing; drop the `X.Y.Z` release branch and the `docs/README.md` space commit (unnecessary — `stable` builds fine at a commit identical to a tag). The `stable` branch step stays; it publishes `/stable/`.

## Rollout (after merge)

No tagging — v1.6.1 is the last release; 1.6.2 is the dev version.

1. Copy `s3://autogluon.mxnet.io/stable/` → `/1.6/` (`stable` is identical to the `v1.6.1` tree).
2. Copy `/v1.5.0/` → `/1.5/`, keeping the original. Pre-1.5 are already bare.
3. Verify both return 200, then reconcile `versions.rst`.

Copied pages keep their original title; `/X.Y/` gets the clean one at the next tagged release. Note a tag push runs the full CI suite (~24 jobs), not just docs — restricting that is a follow-up.

### DIFF
--- a/.github/workflow_scripts/build_all_docs.sh
+++ b/.github/workflow_scripts/build_all_docs.sh
@@ -45,6 +45,7 @@ fi
 
 setup_build_contrib_env
 install_all_no_tests
+export_docs_version_label "$BRANCH"
 
 LOCAL_DOC_PATH=_build/html
 
--- a/.github/workflow_scripts/build_all_docs.sh
+++ b/.github/workflow_scripts/build_all_docs.sh
@@ -19,17 +19,7 @@ then
     if [[ -n $PR_NUMBER ]]; then path=$PR_NUMBER; else path=$BRANCH; fi
     site=$bucket.s3-website-us-west-2.amazonaws.com/$path/$COMMIT_SHA  # site is the actual bucket location that will serve the doc
 else
-    if [[ $BRANCH == 'master' ]]
-    then
-        path='dev'
-    else
-        if [[ $BRANCH == 'dev' ]]
-        then
-            path='dev-branch'
-        else
-            path=$BRANCH
-        fi
-    fi
+    path=$(docs_publish_path "$BRANCH")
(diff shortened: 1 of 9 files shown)


Innixma (approved):
LGTM!

PR #5826, opened 2026-08-07
[docs] Use clean /X.Y/ scheme in versions.rst and freeze dev/stable labels

shchur:
## Description

Follow-up to #5824. Now that the `/X.Y/` doc prefixes exist in the bucket (`/1.0/` … `/1.6/`, each mapped to the latest patch in its minor line, plus `stable`/`dev`), update `versions.rst` to match.

Changes:
- List only `dev`, `stable`, and `1.0`–`1.6` using the clean `/X.Y/` scheme (dropping the old patch-level `/1.4.0/`, `/1.3.1/`, etc. and pre-1.0 entries).
- Drop the patch numbers from the `dev`/`stable` **labels** (`AutoGluon 1.6.2 (dev)` → `AutoGluon dev`).

### Why drop the labels' version numbers

The intent is to copy this `versions.rst` verbatim into every old version's site so old docs always link to the latest release via `stable`/`dev` — without ever having to touch old versions again. For that to be maintenance-free, the copied file must not name anything that changes over time. The `1.0`–`1.6` labels are fixed minor versions (safe to freeze), but `stable`/`dev` version numbers would go stale the moment the next release ships. Naming them plainly keeps the frozen copies correct forever.

### Rollout note

A reader landing on an old `/1.1/` page won't see links to versions released *after* that page was frozen — but `stable` and `dev` always carry the latest, so they're never stranded. This is intentional: old sites are frozen, and `stable`/`dev` are the evergreen path forward.

All links verified to return 200.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@
 
 </div>
 
-AutoGluon, developed by AWS AI, automates machine learning tasks enabling you to easily achieve strong predictive performance in your applications.  With just a few lines of code, you can train and deploy high-accuracy machine learning and deep learning models on image, text, time series, and tabular data.
+AutoGluon automates machine learning tasks enabling you to easily achieve strong predictive performance in your applications.  With just a few lines of code, you can train and deploy high-accuracy machine learning and deep learning models for time series and tabular data.
 
 
 ## 💾 Installation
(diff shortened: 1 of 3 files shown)


Innixma (approved):
LGTM!

PR #4384, opened 2024-08-13
[tabular] Fix exception in HPO when only specifying one of `num_cpus`/`num_gpus`

Innixma:
*Issue #, if available:*

Resolves #4375

*Description of changes:*

- Fix exception in HPO when only specifying one of `num_cpus` or `num_gpus` in `ag_args_fit`.
- Fix exception in HPO for FASTAI and NN_TORCH when `num_cpus` specified in `ag_args_fit`, caused by RayTune converting int to float for `num_cpus`, which leads to a failure in a downstream type assertion as reported in #4375.

## Example

For example, in mainline the following with raise an exception:

```python
import pandas as pd
from autogluon.tabular import TabularPredictor


if __name__ == '__main__':
    label = 'class'
    train_data = pd.read_csv('https://autogluon.s3.amazonaws.com/datasets/Inc/train.csv')
    subsample_size = 1000  # subsample subset of data for faster demo, try setting this to much larger values
    if subsample_size is not None and subsample_size < len(train_data):
        train_data = train_data.sample(n=subsample_size, random_state=0)
    test_data = pd.read_csv('https://autogluon.s3.amazonaws.com/datasets/Inc/test.csv')
    hyperparameters = {'FASTAI': {}, 'NN_TORCH': {}, 'CAT': {}, 'GBM': {}, }
    predictor = TabularPredictor(
        label=label,
    )

    predictor = predictor.fit(
        train_data=train_data,
        hyperparameters=hyperparameters,
        dynamic_stacking=False,
        presets="best_quality",
        time_limit=45,
        ag_args_fit={"num_cpus": 1},
        hyperparameter_tune_kwargs='auto',
    )

    predictor.leaderboard(test_data, display=True)

```

This results in the following exception:

```
Hyperparameter tuning model: LightGBM_BAG_L1 ... Tuning model for up to 10.11s of the 44.95s of remaining time.
/opt/conda/envs/ag-111-310v2/lib/python3.10/site-packages/torch/cuda/__init__.py:628: UserWarning: Can't initialize NVML
  warnings.warn("Can't initialize NVML")
Warning: Exception caused LightGBM_BAG_L1 to fail during hyperparameter tuning... Skipping this model.
Traceback (most recent call last):
  File "/home/ubuntu/workspace/code/autogluon/core/src/autogluon/core/trainer/abstract_trainer.py", line 2278, in _train_single_full
    hpo_models, hpo_results = model.hyperparameter_tune(
[...]

PR #5832, opened 2026-08-10
Fix ZeroDivisionError in HPO resource allocation when num_gpus=0

mario-koddenbrock:
`HpoExecutor.register_resources()` divides `num_gpus // user_specified_trial_num_gpus` without checking for zero. Setting `ag_args_fit={"num_gpus": 0}` crashes with `ZeroDivisionError`.

This is the trial-level counterpart to #4384, which fixed the same issue in the per-fold block further down. That fix didn't cover this block, which runs for every model regardless of bagging.

Fix: when the unspecified resource would divide by zero, skip the division and give the trial all of that resource instead. Symmetric for `num_cpus=0`.

Repro:
```python
TabularPredictor(label="class").fit(
    train_data=train_data,
    hyperparameters={"CAT": {"ag_args_fit": {"num_gpus": 0}}},
    hyperparameter_tune_kwargs="auto",
)
```

Adds two regression tests to `test_hpo_resource_allocation.py`, parametrized over both executors like their neighbors.

### DIFF
--- a/core/src/autogluon/core/hpo/executors.py
+++ b/core/src/autogluon/core/hpo/executors.py
@@ -118,6 +118,16 @@ class HpoExecutor(ABC):
                 "num_gpus", None
             )
             if user_specified_trial_num_cpus is not None or user_specified_trial_num_gpus is not None:
+                # Validate what the user asked for before deriving the other resource from it:
+                # a request larger than the total makes the trial count 0 further down.
+                if user_specified_trial_num_cpus is not None:
+                    assert user_specified_trial_num_cpus <= num_cpus, (
+                        f"Detected trial level cpu requirement = {user_specified_trial_num_cpus} > total cpu granted to AG predictor = {num_cpus}"
+                    )
+                if user_specified_trial_num_gpus is not None:
+                    assert user_specified_trial_num_gpus <= num_gpus, (
+                        f"Detected trial level gpu requirement = {user_specified_trial_num_gpus} > total gpu granted to AG predictor = {num_gpus}"
+                    )
                 num_trials_in_parallel_with_gpu = math.inf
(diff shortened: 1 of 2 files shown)


Innixma:
Pushed a commit to this branch covering the two remaining divide-by-zero paths in the same block, since they are the mirror image of the ones you fixed.

The divisor there is not the user's value but the derived trial count, which becomes 0 whenever the request exceeds the total: `num_gpus: 1` on a predictor granted no gpus, or `num_cpus: 100` on 8 cpus. Both cases already had asserts written for them a few lines below, but the division ran first, so users got `ZeroDivisionError` instead of the message.

The fix is a reorder rather than new handling: the two `<= total` asserts now run before the derivation, guarded by `is not None`. The `>= minimum` asserts stay put because they also have to validate the derived value, while the `<= total` asserts do not, since a derived value is `total // n` with `n >= 1`. Combined with your zero guards, the divisor can no longer be zero in either branch.

Verified against both executors: the target cases keep their behavior (`num_gpus: 0` gives `{'num_cpus': 8, 'num_gpus': 0}`), the over-request cases now raise `Detected trial level gpu requirement = 1 > total gpu granted to AG predictor = 0` and the cpu equivalent, and cases with nothing specified are untouched. Suite goes from 28 to 32 passing with two new parametrized regression tests, which fail with `ZeroDivisionError` if the reorder is reverted.

One thing I deliberately left out of scope for a follow-up: specifying only `num_gpus: 0` now gives the trial every cpu, so Ray runs one trial at a time, where the same predictor with nothing specified would run four in parallel. Worth revisiting separately, not a blocker here.

Innixma (approved):
Thanks for the fix!!

Issue #4505, opened 2024-10-01
[tabular] Kaggle Feedback


Innixma:
From tilli's very nice kaggle post: https://www.kaggle.com/competitions/playground-series-s4e9/discussion/536980

1. Investigate adding LAMA's NN model to AutoGluon.
2. Investigate if hill climbing leads to performance improvement in ensembling. Can verify via [TabRepo]
(https://github.com/autogluon/tabrepo).

Both of these tasks are things that could be implemented by members in the community if they want to give it at try.

- For LAMA's NN: Refer to the [custom model tutorial](https://auto.gluon.ai/stable/tutorials/tabular/advanced/tabular-custom-model.html)
- For hill climbing, refer to the experiment running code in TabRepo, you can replace the ensemble selection algorithm with a hill climbing algorithm and see if it improves the simulation score: https://github.com/autogluon/tabrepo/blob/main/scripts/baseline_comparison/evaluate_baselines.py

mdlakic:
A very good and easy to understand hill climbing approach is already available:

https://github.com/Matt-OP/hillclimbers/

LennartPurucker:
To clarify, if we are talking about hill climbing here, do we mean numerical black-box optimization (like https://arxiv.org/abs/2307.00286)? 

The hill climbing in the linked code also reminded me a lot of https://arxiv.org/abs/1502.04759

PR #5834, opened 2026-08-10
[tabular] Add DenseLight (LAMA-style) neural net model

a11to1n3:
*Issue #, if available:* Addresses #4505 (NN half)

## Description

Kaggle feedback (tilii7 / #4505): AutoGluon NNs are diverse vs trees but not competitive enough with **LAMA-based NNs**, so users run LightAutoML beside AG. This PR adds LAMA’s strong tabular NN (**DenseLight**) **natively** so that path lives in one package.

Architecture is from LightAutoML (Apache-2.0); training, preprocess, bagging, and device handling are AutoGluon-native. There is **no** `lightautoml` dependency.

Follows the [custom model tutorial](https://auto.gluon.ai/stable/tutorials/tabular/advanced/tabular-custom-model.html) contract pointed at in #4505 (`AbstractModel` / `AbstractTorchModel`, `_fit`, `_preprocess`, defaults, `valid_raw_types`), then registers in-tree as `DENSELIGHT`.

## Implementation notes

- New model key `DENSELIGHT` (`DenseLightModel` → `AbstractTorchModel`).
- Vendored DenseLight net: MLP blocks with `concat_input` (input concatenated into later hidden blocks), LAMA-leaning defaults (`hidden_size=[512, 750]`, quantile numericals, early stopping).
- Auxiliary params: `valid_raw_types` for bool/int/float/category (tutorial-style).
- Fit contract: val holdout if needed, early stopping, `time_limit`, CPU/GPU, save/load via torch base.
- Optional extra: `pip install "autogluon.tabular[denselight]"` (torch only).
- Opt-in only — not added to default presets/portfolios yet.
- Unit tests: `FitHelper.verify_model` + auxiliary-params check.

**Usage**

```python
from autogluon.tabular.models import DenseLightModel

# Registered key
predictor.fit(train_data, hyperparameters={"DENSELIGHT": {}})

# Custom-model tutorial style (class as key)
predictor.fit(train_data, hyperparameters={DenseLightModel: {"n_epochs": 50}})
```

**Out of scope (follow-ups)**

- Other LAMA algos (NODE, AutoInt, …).
- Ensembling / hill climbing (#4505 item 2).
- Default portfolio inclusion — needs diversity/competitiveness evidence first.


a11to1n3:
Rebased onto current `master` (includes #5892 Darwin `xgboost` markers).

The only real conflict was `uv.lock`. Kept master's lock and re-added the `denselight` extra (torch-only). Dropped `[ci] Write uv.lock exclude-newer in UTC` — that timestamp is already on master.

`uv lock --check` and ruff on the DenseLight sources are green locally.

Innixma:
Will plan to take a look at this at some point, but probably not until we are close to the next release. Thanks for the contrib!

LennartPurucker:
FYI: we investigated DenseLight on the TabArena side before, and it did not look very promising; see https://github.com/autogluon/tabarena/issues/101 for the thread. RealMLP and TabM seem to be more competitive choices. 

Maybe you can pick up this issue and see whether DenseLight can be improved on TabArena first, before integrating it into AutoGluon. It could be that the code and implementation I used back then were sub-optimal.

a11to1n3:
Hi @LennartPurucker, this is really informative. I will have a look at your issue and test DenseLight against TabArena first and we will see.

a11to1n3:
Hi @LennartPurucker — we followed up on this.

We re-ran **TabArena-Lite** with a **vendored LAMA DenseLight** (CatEmbedder + SWA trainer, LAMA `tabular_config.yml` defaults: `[512, 256]`, 50 epochs, `lr=3e-4`), not the ordinal-cat MLP from [`add_denselight_mlp`](https://github.com/autogluon/tabarena/tree/add_denselight_mlp). Protocol: first split of the 51 datasets, official 8×1 bagging, default + 25 HPO. 1326/1326 bags completed, 0 failures, 0% imputed.

TabArena-Lite Elo with vendored LAMA DenseLight

Lite Elo:

| method | Elo |
|---|---|
| DenseLight default | 974 |
| DenseLight tuned | 1124 |
| **DenseLight T+E** | **1140** |
| FastaiMLP T+E | 1168 |
| TorchMLP T+E | 1247 |
| TabM T+E | 1394 |
| RealMLP T+E | 1480 |

Your earlier read still holds: even with the real LAMA stack, DenseLight is **not competitive** as a TabArena method — well below TabM / RealMLP, and also below AutoGluon’s existing TorchMLP / FastAI on this Lite board.

[#4505](https://github.com/autogluon/autogluon/issues/4505) was a different question (Kaggle: LAMA NNs helping a *blend* vs AG’s NNs). This PR is still opt-in and not in default presets.

@Innixma @LennartPurucker — given that, is it still worth keeping DenseLight as a useful extra NN in AutoGluon for ensemble diversity, or should we close / hold this until there’s evidence it actually helps `WeightedEnsemble` vs TorchMLP/FastAI?

LennartPurucker:
> https://github.com/autogluon/autogluon/issues/4505 was a different question (Kaggle: LAMA NNs helping a blend vs AG’s NNs). This PR is still opt-in and not in default presets. [...] or should we close / hold this until there’s evidence it actually helps WeightedEnsemble vs TorchMLP/FastAI?

With TabArena and your local results, you can simulate this as well and see whether blending would ever pick it up. @Innixma might be able to share more information on this, or Claude can find it, hopefully.

On a side note, IMO, we could always add it to TabArena, which would then exist as a tested and benchmarked model that someone could easily pull into AutoGluon as a custom model via the hyperparameters. I usually add models to TabArena regardless of final performance. But IMO, it would be important if we could pip install the standalone model for this. I got the impression that its usage would otherwise always be a bit obscure. Do you know if this would be possible?

PR #5568, opened 2026-01-29
[Tabular] Optimize memory usage in RealTabPFNv2.5 preprocessing

celestinoxp:
*Description of changes:*
TabPFNModel._preprocess (used by RealTabPFNv2.5) was performing a deep copy of the entire input DataFrame X to safely transform categorical features. For datasets with many numerical features, this unnecessarily duplicated large amounts of data.

Solution: Switched to X = X.copy(deep=False). This allows feature_generator.transform to assign new transformed columns to the DataFrame without duplicating the potentially large numerical columns that remain unchanged.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #5848, opened 2026-08-19
[core] Stop copying unencoded columns in _label_encode_categoricals

Innixma:
Carries over the optimization from #5568 by @celestinoxp, credited as co-author. That PR applied it to the TabPFN 2.5 wrapper, which #5809 has since removed label encoding from entirely, so the change no longer has anywhere to land there. The same copy now lives in the shared helper, where five wrappers benefit from it instead of one: mitra, rf, nori, tabpfnmix and tabdpt.

## What

`AbstractModel._label_encode_categoricals` deep-copied the entire frame in order to rewrite only the categorical columns, duplicating numeric columns it never touches. A shallow copy is enough.

## Why it is safe

The encoded columns are replaced wholesale by `X[features_in] = self._label_encoder.transform(X=X)`, which rebinds them rather than editing shared buffers. The `preserve_missing` path does write in place, with `X.loc[missing[column], column] = np.nan`, but it writes into those replacements, which are already fresh by that point. Unencoded columns are shared and never written.

Verified with `copy_on_write` both off and on, which matters because AutoGluon pins `pandas<2.4.0` where copy on write is not the default: the caller's frame and dtypes are unchanged, output is identical to the deep copy version, and no chained assignment warnings appear, under both `preserve_missing` settings.

## Effect

Per call, 200k rows with 30 numeric and 3 categorical columns:

| | peak allocation | time |
|---|---|---|
| before | 49.8 MB | 4.3 ms |
| after | 1.2 MB | 0.4 ms |

The gain scales with how many columns are not categorical, and is nil for frames that are almost entirely categorical, where the deep copy was already copying about what gets rewritten.

## Tests

Adds a parametrized test that the caller's frame and dtypes survive both `preserve_missing` modes, and that an unencoded column is shared rather than copied. The sharing assertion is what pins the optimization: it fails if the deep copy comes back. Both cases fail on master and pass here.

Also verified end to end that a RandomForest fit over categoricals with missing values and unseen test categories produces bit identical predictions to master (same md5) without mutating the inputs, and `test_rf.py` plus `test_tabdpt.py` pass (10 tests).

### DIFF
--- a/core/src/autogluon/core/models/abstract/abstract_model.py
+++ b/core/src/autogluon/core/models/abstract/abstract_model.py
@@ -735,7 +735,10 @@ class AbstractModel(ModelBase, Tunable):
         features_in = self._label_encoder.features_in
         if features_in:
             missing = X[features_in].isna() if preserve_missing else None
-            X = X.copy()
+            # Only the encoded columns are written, and they are replaced wholesale on the
+            # next line, so every other column stays shared with the caller rather than
+            # being copied. The `preserve_missing` write below lands on those replacements.
+            X = X.copy(deep=False)
             X[features_in] = self._label_encoder.transform(X=X)
             if preserve_missing:
                 for column in features_in:
--- a/tabular/tests/unittests/models/test_rf.py
+++ b/tabular/tests/unittests/models/test_rf.py
@@ -1,5 +1,9 @@
 import copy
 
+import numpy as np
+import pandas as pd
+import pytest
+
 from autogluon.tabular.models.rf.rf_model import RFModel
 from autogluon.tabular.testing import FitHelper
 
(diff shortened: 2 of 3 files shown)


Innixma:
Moved the new test from `test_tabdpt.py` to `test_rf.py`. The helper lives on `AbstractModel` and five wrappers call it, so the test should not sit behind an optional GPU stack: `tabdpt` is a setup.py extra, and the `verify_model` fit beside it needs that package installed, so a shared-core test placed there only runs in the narrower lane. RandomForest is a caller too and needs nothing beyond sklearn.

The `preserve_missing` flag test stays where it is, since TabDPT is the only caller that passes it.

Same coverage as before: both parametrizations fail against master's deep copy and pass here, and `test_rf.py` passes in full (6 tests).

LennartPurucker (approved):

PR #3507, opened 2023-09-06
[eda][bugfix] fixed IndexError: positional indexers are out-of-bounds

gradientsky:
See https://github.com/autogluon/autogluon/issues/3401

*Issue #, if available:*
https://github.com/autogluon/autogluon/issues/3401

*Description of changes:*
Applied suggested changes

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Issue #3401, opened 2023-07-15
[BUG] [EDA] IndexError: positional indexers are out-of-bounds


YifeiYang210:
**Bug Report Checklist**

- [x] I provided code that demonstrates a minimal reproducible example.  
- [?] I confirmed bug exists on the latest mainline of AutoGluon via source install.  
- [x] I confirmed bug exists on the latest stable version of AutoGluon.  

I'm not sure if the stable version changes and I installed v0.8 of autogluon in June, 2023.

**Describe the bug**

Problem: Function `autogluon.eda.auto.detect_anomalies()` reports the train_data and test_data anomalies for 2-sigma outlier scores figures, but ends when it reports the dataframe by `IndexError: positional indexers are out-of-bounds`.

Solve: In the file of `eda.auto.simple` and `eda.analysis.anomaly`, changing `iloc` to `loc` can fix it.

In Pandas, both iloc and loc are used to access data in a DataFrame. 

iloc is used for selecting data using integer-based indexing. It accesses rows and columns of the DataFrame by their integer positions. It takes integers or integer slices as input and returns the corresponding data. For example, df.iloc[0] selects the first row of the DataFrame, and df.iloc[:, 0] selects the first column of the DataFrame. 

loc is used for selecting data using label-based indexing. It accesses rows and columns of the DataFrame by their labels. It takes labels or label slices as input and returns the corresponding data. For example, df.loc[0] selects the row with the label 0, and df.loc[:, 'column_name'] selects the column with the label 'column_name'.

The difference between iloc and loc is the way they handle the indexing. iloc uses integer-based indexing, while loc uses label-based indexing.

When we split the dataframe which has 10 rows to train_df and test_df randomly, the train_df index may range in [0,1,2,4,5,7,9,10] and the test_df index may range in [3,6,8]. And we call `detect_anomalies()` function to get the anomaly_idx,  the result of anomaly_idx may be [4, 10] in train_df. Using train_df.iloc[anomaly_idx] equals the 4th and 10th rows of train_df then cause IndexError. And train_df.loc[anomaly_idx] equals the index of "4" and index of "10", which can fix it.

Source file:
```python
// simple.py line:1667
_state.anomaly_detection.anomalies[ds] = df.iloc[anomaly_idx].join(anomaly_scores)

// anomaly.py line:346
rows=args[dataset].iloc[dataset_row_ids],
```

Fixed file:
```python
// simple.py line:1667
[...]

PR #5851, opened 2026-08-21
[features] Fix DatetimeFeatureGenerator IndexError on non-RangeIndex

zhihao2020:
*Issue #, if available:*
Related to #3401 / #3507 (same `.iloc` vs `.loc` bug class). Those fixed AutoGluon EDA; `DatetimeFeatureGenerator.normalize_timeseries` was not updated.

*Description of changes:*
- In `normalize_timeseries`, select invalid rows with `series.loc[broken_idx]` instead of `series.iloc[broken_idx]`. `broken_idx` is taken from `series.index` (labels). Assignment `series[broken_idx] = fill` is already label-based; `.iloc` treats those labels as positions and raises `IndexError: positional indexers are out-of-bounds` when the index is not a RangeIndex (bag/split leftover labels).
- Tests: `fit_transform` on a non-contiguous index, and `fit` on RangeIndex then `transform` a held-out frame that kept leftover labels. Mean-fill values are unchanged.


Innixma (approved):
Thanks for the PR! I did a small adjustment so it is faster and doesn't care about the index via does a mask via series.isna(). LGTM!

PR #5852, opened 2026-08-21
[core] Fix EnsembleSelection crash when every candidate regret is NaN

zhihao2020:
*Issue #, if available:*

*Description of changes:*
- `EnsembleSelection._fit` crashed with `ValueError: 'a' cannot be empty unless no samples are taken` when every candidate regret was NaN. `np.nanmin` is NaN, `np.isclose(..., equal_nan=False)` yields no matches, then `RandomState.choice([])` raises. The same `isclose`/`nanmin` pattern is used for `second_metric` tie-break. A follow-on `trajectory.index(min_score)` also cannot find NaN.
- Non-finite regrets (NaN and ±inf) are now treated as dead (`+inf`). If no finite score remains, raise `ValueError("all ensemble scores non-finite")` before `choice` / writing `best_score` into `trajectory`. Do not invent a winner and do not use `equal_nan=True`.
- Note: all-inf scores used to `choice` among themselves (`isclose(inf, inf)`). They now error as non-finite. Mixed finite + NaN/inf still picks the finite model.
- Adds fast `EnsembleSelection` unit tests that fail on current master for all-NaN regrets/preds and pass after the patch.

### DIFF
--- a/core/src/autogluon/core/models/greedy_ensemble/ensemble_selection.py
+++ b/core/src/autogluon/core/models/greedy_ensemble/ensemble_selection.py
@@ -226,13 +226,13 @@ class EnsembleSelection(AbstractWeightedEnsemble):
         """Return sanitized regrets and the indices of the lowest finite score.
 
         Non-finite regrets (NaN / +/-inf) are treated as dead (mapped to +inf) so they
-        never beat a finite score. If every score is non-finite, fall back to index 0
-        instead of ``RandomState.choice([])`` (all-NaN + ``isclose(..., equal_nan=False)``)
-        or ``equal_nan=True`` (which later dies on ``list.index(nan)``).
+        never beat a finite score. If every score is non-finite, raise instead of
+        ``RandomState.choice([])`` or ``equal_nan=True`` (the latter later dies on
+        ``list.index(nan)`` when writing ``trajectory``).
         """
         scores = np.where(np.isfinite(scores), scores, np.inf)
         if not np.isfinite(scores).any():
-            return scores, np.array([0], dtype=int)
+            raise ValueError("all ensemble scores non-finite")
         all_best = np.argwhere(np.isclose(scores, np.nanmin(scores), atol=0, rtol=1e-12)).flatten()
         return scores, all_best
 
--- a/core/src/autogluon/core/models/greedy_ensemble/ensemble_selection.py
+++ b/core/src/autogluon/core/models/greedy_ensemble/ensemble_selection.py
@@ -146,7 +146,7 @@ class EnsembleSelection(AbstractWeightedEnsemble):
                 if round_scores:
                     scores[j] = scores[j].round(round_decimals)
 
-            all_best = np.argwhere(np.isclose(scores, np.nanmin(scores), atol=0, rtol=1e-12)).flatten()
+            scores, all_best = self._select_best_by_score(scores)
 
             if (len(all_best) > 1) and used_models:
                 # If tie, prioritize models already in ensemble to avoid unnecessarily large ensemble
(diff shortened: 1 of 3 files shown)


Innixma:
Could you add a reachability case for this? I tried to reproduce the all-NaN state through a predictor fit and couldn't, because the trainer appears to rule it out upstream.

An all-NaN regret needs the metric to be NaN on `(y_val, model predictions)`. But that same metric on that same data is what produced the model's `val_score`, and `_add_model` refuses to register a model with a NaN `val_score` (`tabular/.../abstract_trainer.py:2658`):

```python
if model.val_score is not None and np.isnan(model.val_score):
```

So a model whose regret would be NaN never becomes an ensemble candidate.

I went after the most plausible route I could construct — `eval_metric="spearmanr"`, which does return NaN when `y_true` is constant, with a user-supplied `tuning_data` whose target is constant (plausible for a heavily-tied or zero-inflated regression target):

```
WARNING: LightGBM has a val_score of nan (NaN)! This should never happen. The model will not be saved...
WARNING: LightGBM_2 has a val_score of nan (NaN)! ...
WARNING: XGBoost has a val_score of nan (NaN)! ...
FIT RAISED: ValueError: Model does not exist: (model=LightGBM)
```

The fit fails earlier, for an unrelated reason, and `EnsembleSelection` is never reached.

The `second_metric` tie-break site looks further out of reach: `tie_breaker` defaults to `"random"`, I can't find anywhere in the repo that sets `"second_metric"`, and that branch's secondary metric is hardcoded to `log_loss`, which stays finite for finite predictions.

None of that means the fix is wrong — `RandomState.choice([])` is clearly a bad failure mode and the guard reads fine. It's that I can't tell from the PR which workflow produces it, and that determines how it should be tested. The unit tests construct `EnsembleSelection` directly, which will pass whether or not the state is reachable from a fit.

`EnsembleSelection` has three consumers besides the tabular weighted ensemble, and the `val_score` guard above is specific to the tabular trainer, so a real case is most likely in one of them:

- `tabular/learner/abstract_learner.py:672` — the `compute_oracle` path, which already carries a `# TODO: Only fit non-nan` on the `fit` call. Note `compute_oracle` defaults to `False` and nothing in the repo passes `True`, so this looks dead today.
- `timeseries` (`TimeSeriesEnsembleSelection`) — no NaN filtering at the call site, and forecasting metrics do go non-finite on degenerate series (MASE divides by the naive-forecast error, which is zero for a constant series; sMAPE is 0/0 on all-zero actuals). Intermittent or zero demand is an ordinary workflow, so this seems the most likely origin.
- `multimodal/learners/ensemble.py:231`.

If it came from one of those, could you say which, and add a test at that level? That would pin the behavior where it actually occurs rather than only at the `EnsembleSelection` API, and it would show whether the right fix belongs here or in that caller's candidate filtering.

zhihao2020:
Thanks — your `TabularPredictor.fit` reconstruction is correct. `_add_model` drops a NaN `val_score`, so a normal tabular fit never reaches `EnsembleSelection._fit` with all-NaN regrets. The spearmanr + constant `tuning_data` case dies even earlier. `second_metric` / `compute_oracle` are unused in-repo, as you said.

The reachable consumer is timeseries greedy ensemble selection, not `TimeSeriesPredictor.fit`. `EnsembleComposer.fit` wraps each ensemble in `try/except` and skips it, so a predictor-level fit does not crash. What does hit `_fit` is `GreedyEnsemble.fit` / `PerItemGreedyEnsemble.fit` on an ordinary zero-demand or constant item: default WQL is `nansum(err)/nansum(|y|)` (0/0 on an all-zero horizon) and MASE `_safemean` is NaN when the in-sample seasonal scale is 0. There is no NaN-`val_score` filter on that path.

I added `timeseries/tests/unittests/models/ensemble/test_greedy_nonfinite_scores.py` for that. On current master those cases raise `ValueError: 'a' cannot be empty unless no samples are taken` (`RandomState.choice([])`). After the guard they raise `ValueError: all ensemble scores non-finite`. The existing `EnsembleSelection` unit tests stay as the class-API backstop.

The `_fit` guard is still the right place for the crash; whether a caller should filter candidates instead is a separate question (and `EnsembleComposer` already chooses to skip).

Innixma (approved):
LGTM, thanks for the contrib!

PR #5853, opened 2026-08-21
[tabular] Fix DyStack holdout leakage when using groups

### LINKED ISSUE #5533: [BUG] [tabular] Dynamic stacking and using groups is incompatible
Bug report from an internal user

> I know it’s an experimental feature, but I think there is an incompatibility between dynamic stacking and using groups.  It looks like the holdout set that dynamic stacking uses does not respect the groups, which can lead to massive data leakage (on the set I just tried, the validation f1s are around .8 and the holdout f1s are .9998).  This could potentially be fixed by setting aside one of the groups as a holdout, but I actually recommend just raising a warning or perhaps even an error if you try to do both at once.

a11to1n3:
*Issue #, if available:*

Fixes #5533

*Description of changes:*

`groups=` already makes **bagging** leave-one-group-out. Dynamic stacking's default holdout did not: it used `generate_train_test_split_combined` (a row split), so the same group could sit in the DyStack sub-fit train set and the holdout. Bagged OOF (`score_val`) stayed honest; the DyStack holdout (`score_test`) could look nearly perfect — the `.8` vs `.9998` F1 in the report.

This reuses the existing `ValidationStructure(group_on=)` splitter for DyStack holdout and CV when `groups=` is set (the two APIs stay mutually exclusive for the user). The full fit after DyStack still trains on all rows.

Guards so the sub-fit can still bag:
- at least 3 unique groups (one held out, two left for leave-one-group-out)
- if a coarse split would leave only 1 train group (few groups + large `holdout_frac`), move the smallest whole holdout groups back until train has 2

`ds_args["holdout_data"]` is still the caller's split. No `groups=` is unchanged (row holdout).

---

### Also fixed here: DyStack CV + `groups` crashes on master

`CVSplitter` expects the group *vector*, but master passed the column *name*, and
`_get_splitter_cls` calls `.unique()` on it. So `ds_args={"validation_procedure": "cv"}`
together with `groups=` is unusable today, not merely leaky:

```
### master ###    DyStack CV + groups -> AttributeError: 'str' object has no attribute 'unique'
### this PR ###   DyStack CV + groups: completed
```

### When the check cannot run, stacking is disabled rather than the fit failing

Too few groups to build a group-disjoint holdout means the stacked-overfitting question is
*unanswered*, not answered "clean". Erroring would turn fits that work today into failures
(`groups` with 2 unique values plus `dynamic_stacking=True` succeeds on master), so instead
stacking is turned off and the fit continues, with `_stacked_overfitting_occurred` set to
`None` to distinguish "not measured" from "measured, clean".

### Known follow-up (not in this PR)

The infeasibility guard is gated on `groups is not None`, but the constraint belongs to
group-disjoint holdouts generally. Reached through `validation_structure={"group_on": ...}`
on master, with no `groups` involved:

| n_groups | holdout_frac | train groups left | sub-fit |
|---|---|---|---|
| 2 | any | 1 | fails -> stacking stays **enabled** |
| 3 | >=0.5 | 1 | fails -> stacking stays **enabled** |
| 3 | 0.2 | 2 | ok |
| 4+ | >=0.2 | 2+ | ok |

DyStack swallows the sub-fit failure and keeps stacking on, so the leakage check silently does
not run. Un-gating the guard would apply the same disable-stacking policy to that path.

### DIFF
--- a/tabular/src/autogluon/tabular/predictor/predictor.py
+++ b/tabular/src/autogluon/tabular/predictor/predictor.py
@@ -160,6 +160,11 @@ class TabularPredictor:
         It is not recommended to use this option unless it is required for very specific situations.
         Bugs may arise from edge cases if the provided groups are not valid to properly train models, such as if not all classes are present during training in multiclass classification. It is up to the user to sanitize their groups.
 
+        Dynamic stacking holdouts and DyStack CV splits are group-disjoint as well (whole groups
+        are held out), so the stacked-overfitting check cannot leak across groups. That requires
+        at least 3 unique group ids. For grouped *and* repeated bagging, prefer
+        `validation_structure={"group_on": ...}` instead of `groups`.
+
         As an example, if you want your data folds to preserve adjacent rows in the table without shuffling, then for 3 fold bagging with 6 rows of data, the groups column values should be [0, 0, 1, 1, 2, 2].
     positive_class : str or int, default = None
         Used to determine the positive class in binary classification.
@@ -1672,6 +1677,14 @@ class TabularPredictor:
             enable_ray_logging=enable_ray_logging,
         )
 
+        features_X = X.drop(self.label, axis=1)
+        y = X[self.label]
+        # `validation_structure` already encodes grouped/temporal holdouts. The older
+        # `groups=` bagging channel does not: a random row holdout leaks across groups
+        # (https://github.com/autogluon/autogluon/issues/5533). Reuse the same
+        # group-disjoint splitter when `groups` is set.
+        validation_structure = self._dystack_validation_structure(X=X, ag_fit_kwargs=ag_fit_kwargs, X_val=X_val)
+
         # -- Validation Method
         if validation_procedure == "holdout":
             if holdout_data is None:
@@ -1681,13 +1694,12 @@ class TabularPredictor:
                 # train_indices/val_indices channel the CV procedure already goes through, so
                 # the sub-fit itself needs no change. `holdout_split_indices` returns None for
                 # stratify-only structures, where the default split needs no correction.
-                validation_structure = ag_fit_kwargs.get("validation_structure")
                 structure_holdout = (
                     None
                     if validation_structure is None
                     else validation_structure.holdout_split_indices(
-                        X.drop(self.label, axis=1),
-                        X[self.label],
+                        features_X,
+                        y,
                         holdout_frac=holdout_frac,
                         random_state=42,
                         problem_type=self.problem_type,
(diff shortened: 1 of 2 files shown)


Innixma:
Thanks for this — the diagnosis is right and reusing `ValidationStructure(group_on=)` for the DyStack splits is the direction I'd want. I pushed one commit (`c2d555f5`) and extended the description; details below so nothing is a surprise.

## Changed: an unrunnable check disables stacking rather than failing the fit

The two `ValueError`s turned configurations that work today into hard errors:

```
### master ###    2 groups + dystack: OK  models=['LightGBM_BAG_L1', 'LightGBM_BAG_L2']
### this PR ###   2 groups + dystack: ValueError: dynamic_stacking with `groups` needs at
                  least 3 unique groups...
```

Too few groups to build a group-disjoint holdout means the stacked-overfitting question is *unanswered*, not answered "clean" — so the conservative response is to not stack, not to stop. After the change that case fits fine and simply produces no L2:

```
2 groups + dystack: OK  models=['LightGBM_BAG_L1']
```

- `_dystack_skip_reason` makes the decision up front, before anything is popped off `ag_fit_kwargs`, so the early exit needs no unwinding.
- `_dystack_disable_stacking` centralises the warning and sets `_stacked_overfitting_occurred = None` — `None` rather than `False`, so callers can distinguish "not measured" from "measured, clean".
- `_dystack_keep_logo_feasible` returns `None` instead of raising.
- `stacked_overfitting` is initialised before the branches, since either can now skip its sub-fit.

`test_..._too_few_groups_raises` became `test_..._too_few_groups_disables_stacking`, asserting no `_L2` model and `_stacked_overfitting_occurred is None`.

I also dropped a test I'd briefly added for a large `holdout_frac` — with 3 groups the repair always succeeds, so that case is the one `test_..._large_holdout_frac_keeps_logo_feasible` already covers, not an infeasible one. Good news for the guard: `_dystack_keep_logo_feasible` returning `None` is currently unreachable via `groups=`.

## Description: added the crash fix

Worth surfacing, because it's arguably the more urgent half — `ds_args={"validation_procedure": "cv"}` with `groups=` is an `AttributeError` on master (the column *name* reaches `CVSplitter`, which calls `.unique()` on it), so that path is unusable rather than merely leaky.

## One follow-up I did not do here

The guard is gated on `groups is not None`, but the constraint belongs to group-disjoint holdouts in general. Reached through `validation_structure={"group_on": ...}` on master, no `groups` involved:

| n_groups | holdout_frac | train groups left | sub-fit |
|---|---|---|---|
| 2 | any | 1 | fails → stacking stays **enabled** |
| 3 | ≥0.5 | 1 | fails → stacking stays **enabled** |
| 3 | 0.2 | 2 | ok |
| 4+ | ≥0.2 | 2+ | ok |

End to end that shows up as:

```
validation_structure: 1 groups < num_folds (3); setting num_folds=1 and num_repeats=1.
Warning: Exception encountered during DyStack sub-fit:
...
models=['LightGBM_BAG_L1', 'LightGBM_BAG_L2']     ← L2 present, stacking retained
```

So the leakage check silently didn't run and stacking stayed on. Un-gating the guard would apply the same disable-stacking policy there, but it's a behavior change to the existing API and belongs in its own PR — happy to take it.

## Verification

`tabular/tests/unittests/dynamic_stacking` — 15 passed. `ruff format --check` and `ruff check --select I` clean on `tabular/`.

Innixma (approved):
LGTM! I added a few adjustments + will make a follow-up PR to collapse groups to be identical to group_on so we don't need to worry about two separate group logic implementations in the codebase.

PR #5868, opened 2026-08-26
fix(core): include static params in grid search configs

mameikagou:
*Issue #, if available:* N/A

*Description of changes:*

`LocalGridSearcher` returned only sampled parameters, so configs containing static search-space values failed validation during `update`. This change adds the static values to each returned config, matching `LocalRandomSearcher`, and adds regression coverage for enumeration, result updates, and exhaustion.

Validation: `6 passed` in the focused grid/base searcher tests; Ruff and `git diff --check` pass.


Innixma (approved):
LGTM, thanks for the contrib!

PR #5887, opened 2026-09-04
[features] Reuse categorical codes in OOFTargetEncodingFeatureGenerator

Innixma:
Stacked on #5885 (which is stacked on #5884) — this PR's diff is the final commit only.

## Summary

Two redundancies in `OOFTargetEncodingFeatureGenerator`'s per-column encoding loop.

**Sorting categories nobody reads.** `pd.factorize(col_values, sort=True)` sorts every column's
uniques. The order is never read — codes only index `enc_matrix`, and `categories` is stored at fit
so `transform` reproduces the same codes — so sorting only relabels, at the cost of argsorting
every column's uniques.

**Discarding codes it was handed.** `X_cat = X[self.cols_].astype("object")` cast every column to
object, including category columns that already carry exactly the codes the loop is about to
recompute.

This takes the codes from a category column directly and drops the sort. Object columns keep the
previous path apart from the sort.

## Correctness

Unused categories are dropped before the codes are read. `factorize` only ever yields *observed*
categories, so a declared-but-unobserved category would otherwise take a zero count and — with
`alpha=0` — a 0/0 `NaN` encoding instead of the global mean. There is a test for exactly that.

Encodings are unchanged up to floating-point noise (~1e-16). They are not bit-identical: category
order changes the order in which the `bincount` accumulations are summed. Two tests pin the
invariants that matter — that object and category inputs encode identically, and that the result
does not depend on category order.

## Benchmarks

Per-column category resolution, 4994 rows x 1500 category columns:

| | time |
|---|---|
| `astype(object)` + `factorize(sort=True)` (before) | 1.218s |
| `factorize(sort=False)` | 0.486s |
| reuse the existing codes (after) | **0.035s** |

plus the 0.263s `astype("object")` that is no longer needed.

End to end for the generator that motivated this, together with #5885 (4994 rows, 86 base
features, one key column per subset):

| n_subsets | fit before | after | | transform before | after | |
|---|---|---|---|---|---|---|
| 50 | 0.31s | 0.08s | 4.1x | 0.10s | 0.03s | 3.4x |
| 200 | 1.44s | 0.25s | 5.7x | 0.41s | 0.11s | 3.8x |
| 500 | 3.43s | 0.67s | 5.1x | 1.06s | 0.27s | 4.0x |
| 1500 | 10.24s | 1.95s | 5.3x | 3.07s | 0.79s | 3.9x |

## Testing

3 new tests: object vs category inputs encode identically, unused categories do not produce NaN
encodings (at fit or at transform), and encodings are independent of category order.

`pytest features/tests` passes (13/13 in `test_oof_target_encoder.py`); the one failure in
`test_check_style.py` is pre-existing on master and unrelated to these files.

### DIFF
--- a/features/src/autogluon/features/generators/oof_target_encoder.py
+++ b/features/src/autogluon/features/generators/oof_target_encoder.py
@@ -1,5 +1,6 @@
 import numpy as np
 import pandas as pd
+from pandas.api.types import CategoricalDtype
 from sklearn.model_selection import KFold, StratifiedKFold
 
 from autogluon.common.features.types import (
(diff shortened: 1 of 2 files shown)


LennartPurucker (approved):
Check linting

PR #5891, opened 2026-09-07
[common] Remove `ValidationStructure.group_time_on`

Innixma:
Removes the `group_time_on` field of `ValidationStructure`, along with its handling in the predictor and its tests.

The field blocked whole groups in the order of their first appearance in the training frame and treated that sequence as time, so the resulting splits and the "latest groups" holdout depended on row order rather than on a time column. It was a mistake made when translating the BeyondArena API, where a column of that name is a time *within* a group and does not take part in splitting. Grouped and temporal validation are unchanged and remain available separately via `group_on` and `time_on`; combining the two still raises `NotImplementedError`.


LennartPurucker (approved):
So this removes the metadata fully and never allows models to use it?

Many models likely want to exploit the order of samples for, e.g. feature engineering?

Innixma:
> So this removes the metadata fully and never allows models to use it?
> 
> Many models likely want to exploit the order of samples for, e.g. feature engineering?

Yes. I plan to add it back later, but for now I want the incorrect logic removed.

PR #5892, opened 2026-09-08
[tabular] Install xgboost on macOS; xgboost-cpu has no Darwin wheels

### LINKED ISSUE #5881: [BUG] pip install autogluon fails on macOS since 1.6.0 as xgboost-cpu has no macOS wheels
**Bug Report Checklist**

**Describe the bug**

The documented macOS CPU install cannot succeed on any Mac with AutoGluon 1.6.0 or later.

`autogluon.tabular[all]` requires `xgboost-cpu`, which has **never** published a macOS wheel — not for any version, not for any Python... however if we attempted to install `xgboost` it would work just fine on mac!

This affects every macOS user of the umbrella `autogluon` package, since `autogluon` depends on `autogluon.tabular[all]`.

The change landed in 1.6.0 (2026-08-05). From PyPI metadata for `autogluon.tabular`:

| Version | Released | Requirement |
|---|---|---|
| 1.5.0 | 2025-12-19 | `xgboost<3.2,>=2.0` |
| 1.6.0b20260805 | 2026-08-05 | `xgboost<3.4,>=2.0` |
| **1.6.0** | **2026-08-05** | **`xgboost-cpu<3.4,>=2.1.1`** |
| 1.6.1 | 2026-08-06 | `xgboost-cpu<3.4,>=2.1.1` |

Every nightly up to and including `1.6.0b20260805` still used plain `xgboost`; the 1.6.0 stable cut the same day carries `xgboost-cpu`. 1.5.x and earlier install fine.

**Expected behavior**

`pip install autogluon` on macOS installs prebuilt wheels and completes, as it did through 1.5.x.

**To Reproduce**

On any Mac without `cmake` on PATH, run exactly what `docs/install-mac-cpu.md` says:

```console
pip install -U pip
pip install -U setuptools wheel
pip install autogluon --extra-index-url https://download.pytorch.org/whl/cpu
```

**Pip install Logs**

```
Collecting xgboost-cpu<3.4,>=2.1.1 (from autogluon.tabular[all])
  Downloading xgboost_cpu-3.2.0.tar.gz (1.3 MB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... error
  error: subprocess-exited-with-error

  INFO:xgboost.packager.build_libxgboost:Building libxgboost.dylib from the C++ source files in ...
  INFO:xgboost.packager.build_libxgboost:CMake args: ['cmake', ..., '-DUSE_OPENMP=ON', '-DUSE_CUDA=OFF', ...]
  ...
    File ".../packager/nativelib.py", line 57, in _build
      subprocess.check_call(cmake_cmd, cwd=build_dir)
  FileNotFoundError: [Errno 2] No such file or directory: 'cmake'

error: metadata-generation-failed
```

---

**Root cause**

`xgboost` and `xgboost-cpu` are different distributions with different wheel coverage:

| Distribution | Wheel platforms (3.2.0, and every other release) |
|---|---|
| `xgboost` | macosx_12_0_arm64, macosx_10_15_x86_64, manylinux_2_28_x86_64, manylinux_2_28_aarch64, win_amd64, sdist |
| `xgboost-cpu` | manylinux_2_28_x86_64, manylinux_2_28_aarch64, win_amd64, win_arm64, sdist |

This is not a transient publishing gap, and upstream is unlikely to close it, because these sizes make a macOS `xgboost-cpu` wheel pointless:

```
  2.3 MB   xgboost-3.2.0-py3-none-macosx_12_0_arm64.whl
131.7 MB   xgboost-3.2.0-py3-none-manylinux_2_28_x86_64.whl
  5.6 MB   xgboost_cpu-3.2.0-py3-none-manylinux_2_28_x86_64.whl
```

`xgboost-cpu` exists to strip bundled CUDA out of the large Windows and Linux wheels. There is no CUDA build of XGBoost for macOS, so the ordinary macOS `xgboost` wheel is *already* the CPU-only build at 2.3 MB. A macOS `xgboost-cpu` wheel would duplicate a wheel that already ships. Unless we can persude `xgboost-cpu` maintainers (unlikely I think!) to include a duplicate mac wheel, we are stuck working around it ourselves.

**One possible solution?**

I'm not an expert in pip or conda but I think we can give the requirement a platform marker, so that macOS installs the ordinary `xgboost` wheel while every other platform keeps `xgboost-cpu` exactly as today. In the `"xgboost"` entry of `tabular/setup.py`, the single requirement `"xgboost-cpu>=2.1.1,<3.4"` becomes a pair: `"xgboost-cpu>=2.1.1,<3.4; platform_system != 'Darwin'"` alongside `"xgboost>=2.1.1,<3.4; platform_system == 'Darwin'"`. Evaluating those markers gives `xgboost-cpu` on Linux and on both Windows architectures, and plain `xgboost` only on Darwin.

a11to1n3:
*Issue #, if available:*

Fixes #5881

*Description of changes:*

`autogluon.tabular[xgboost]` (and therefore `pip install autogluon`) has required `xgboost-cpu` since 1.6.0. That extra package has never published a macOS wheel — it exists to strip CUDA out of the large Linux/Windows wheels. pip then falls through to the sdist and fails with `FileNotFoundError: cmake`.

The ordinary macOS `xgboost` wheel is already CPU-only (~2.3 MB) and does not depend on `nvidia-nccl-cu12` (that pin is `platform_system == "Linux"`), so the NCCL clash that motivated `xgboost-cpu` does not apply on Darwin.

This keeps `xgboost-cpu` on Linux and Windows and selects `xgboost` only on Darwin:

```python
"xgboost-cpu>=2.1.1,<3.4; platform_system != 'Darwin'",
"xgboost>=2.1.1,<3.4; platform_system == 'Darwin'",
```

Same marker style as the existing torch / onnxruntime-gpu Darwin pins.

**Verification**
- `uv lock --check` is green (lockfile now forks `xgboost` on `sys_platform == 'darwin'` and `xgboost-cpu` otherwise).
- `uv export --package autogluon.tabular --extra xgboost --frozen` on this Mac resolves to `xgboost==3.3.0 ; sys_platform == 'darwin'` (Python 3.12+).
- `pip download 'xgboost>=2.1.1,<3.4'` fetches `xgboost-3.3.0-py3-none-macosx_12_0_arm64.whl` (2.3 MB). `pip download xgboost-cpu==3.2.0` still hits the sdist/`cmake` failure, which is why the marker is required.

### DIFF
--- a/tabular/setup.py
+++ b/tabular/setup.py
@@ -50,7 +50,13 @@ extras_require = {
         # CPU even when GPUs are allocated (it warns and falls back); no GPU-targeted portfolio
         # contains an XGB config. Revisit if xgboost ships a cu13 wheel or makes NCCL optional
         # (https://github.com/dmlc/xgboost/issues/10729).
-        "xgboost-cpu>=2.1.1,<3.4",  # >=2.1.1 is the earliest xgboost-cpu release; <{N+1} upper cap
+        #
+        # `xgboost-cpu` has never published a macOS wheel (CUDA is why the extra package
+        # exists, and there is no CUDA XGBoost on Darwin). The ordinary macOS `xgboost`
+        # wheel is already CPU-only and has no NCCL pin, so use that there.
+        # https://github.com/autogluon/autogluon/issues/5881
+        "xgboost-cpu>=2.1.1,<3.4; platform_system != 'Darwin'",  # >=2.1.1 is the earliest xgboost-cpu release; <{N+1} upper cap
+        "xgboost>=2.1.1,<3.4; platform_system == 'Darwin'",
     ],
     "realmlp": [
         "pytabkit>=1.7.2,<1.8",
--- a/uv.lock
+++ b/uv.lock
@@ -605,8 +605,10 @@ all = [
     { name = "torch", version = "2.10.0", source = { registry = "https://pypi.org/simple" }, marker = "sys_platform == 'darwin'" },
     { name = "torch", version = "2.13.0", source = { registry = "https://pypi.org/simple" }, marker = "sys_platform != 'darwin'" },
     { name = "transformers" },
-    { name = "xgboost-cpu", version = "3.2.0", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version < '3.12'" },
-    { name = "xgboost-cpu", version = "3.3.0", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version >= '3.12'" },
+    { name = "xgboost", version = "3.2.0", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version < '3.12' and sys_platform == 'darwin'" },
+    { name = "xgboost", version = "3.3.0", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version >= '3.12' and sys_platform == 'darwin'" },
+    { name = "xgboost-cpu", version = "3.2.0", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version < '3.12' and sys_platform != 'darwin'" },
+    { name = "xgboost-cpu", version = "3.3.0", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version >= '3.12' and sys_platform != 'darwin'" },
 ]
 catboost = [
     { name = "catboost" },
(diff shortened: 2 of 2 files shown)


Innixma:
Nice find! I verified it and reproduced. Thanks for the contrib!

Innixma (approved):

PR #5931, opened 2026-09-18
[common] Fall back to spawn when the multiprocessing start method is unavailable

MohammedAlkindi:
*Issue #, if available:* none

*Description of changes:*

`execute_multiprocessing` defaults `multiprocessing_method="forkserver"`, which is POSIX-only. On Windows `multiprocessing.get_all_start_methods()` returns `['spawn']` alone, so `get_context()` raises `ValueError: cannot find context for 'forkserver'`. That reaches users through the public loaders — `load_pd.load([path_a, path_b])` and multipart `save_pd.save` both take the default.

This substitutes `'spawn'` when the requested method is unavailable on the platform, and leaves it untouched wherever it is available.

Measured on Windows 11, Python 3.13.5, at e0bd2e0. Before: 2 failed, both `ValueError: cannot find context for 'forkserver'`. After: 2 passed, 1 skipped, and `load_pd.load([f1, f2])` returns the concatenated frame instead of raising.

The skipped case asserts an available method is still used unchanged, so the fallback cannot fire where `forkserver` is native; it runs on Linux CI, not on my box. `ruff format --diff` and `ruff check --select I` are clean at the pinned v0.16.4.


Innixma (approved):
LGTM, thanks for the contrib!

PR #5933, opened 2026-09-20
fix ignoring L1 in linear model for classification

davidbonet:
Fix `LinearModel` ignoring `penalty="L1"` for binary and multiclass classification.

L1 can use `saga` by default, while L2 continues to use `lbfgs` (it only supports L2 regularization). Regression continues to select `Lasso` or `Ridge` correctly according to the requested penalty. Which is a concern if the user wants to use L1 regularization for proper sparse feature selection.

## Details and validation

AutoGluon previously removed `penalty` before constructing `LogisticRegression`, causing both L1 and L2 requests to use scikit-learn's default L2 regularization, ignoring requesting L1 completely.

This change forwards the requested regularization using the appropriate API for the installed scikit-learn version. It also fixes penalty forwarding in the RAPIDS wrapper.

I reproduced it in a synthetic binary classification example using 800 samples, 3 informative features, and 17 independent noise features. All fits using identical data, preprocessing, and `C=0.03`.

| Configuration | Nonzero coefficients | Exactly zero |
| --- | ---: | ---: |
| Before fix, requesting L1 | 20 | 0 |
| After fix, requesting L1 | 3 | 17 |
| Explicit L2 control | 20 | 0 |

Before the fix, requesting L1 produces exactly the same coefficients as the L2 control. After the fix, all 17 noise coefficients are exactly zero.

All 24 linear-model tests pass on scikit-learn 1.7.2 and 1.9.1, covering:

- Binary and multiclass classification, including coefficient sparsity.
- Lasso/Ridge selection for regression.
- Explicit solver overrides and incompatible solver/penalty combinations.
- Invalid classification penalties.
- RAPIDS parameter forwarding, tested with mocks.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution under the terms of your choice.


LennartPurucker:
I'm surprised this bug has survived this long. It might also change the quality of linear models for stacking @Innixma.

Could this be the result of version drift in sklearn that AutoGluon didn't follow?

Innixma (approved):
Looks great! I added a small cleanup to the penalty type check logic. Thanks for the contrib!

PR #5934, opened 2026-09-22
[core][tabular] TabularPredictor(save_to_disk=False): keep the fitted predictor in memory

Innixma:
*Issue #, if available:*

*Description of changes:*

Adds `TabularPredictor(save_to_disk=False)`: the predictor writes nothing under its path during fit. The learner and trainer keep the fitted models, the cached training data, the holdout predictions and a bagged model's fold models (with their template and out-of-fold predictions) in memory instead of pickling them, and `save()` / `unpersist()` become no-ops. An in-memory bag fits its folds sequentially, since the parallel strategies hand fold models back through files. The default behaviour is unchanged.

Tests in `tabular/tests/unittests/test_save_to_disk.py` cover binary, multiclass and regression fits, a holdout fit with a weighted ensemble and a bagged fit (predict, leaderboard, info, persist, OOF predictions and the simulation artifact work with no files written), and that the default predictor still saves and reloads.


LennartPurucker (approved):
LGTM, nice feature, maybe we should remove low_memory now?

